PR #3732, opened 2026-05-25
feat: seed Orchard shielded pool at genesis with fast, observable sync

shumkov:
## Issue being fixed or feature implemented

Continues the work from #3714 / #3769: make it possible to exercise **wallet shielded-note sync at scale** (≈1,000,000 notes) on a local devnet, and make that sync both fast and observable.

When `drive-abci` is built with `--cfg create_sdk_test_data`, the Orchard shielded pool is pre-populated at genesis with ≈1,000,000 notes (a handful owned across two deterministic ZIP-32 test wallets). Standing that data up surfaced two further needs that this PR also addresses: cold sync of a million notes had to be made fast enough to iterate on, and the long-running sync needed real progress feedback in the wallet UI.

## What was done?

**Genesis seeding of the shielded pool** (`drive-abci`, `dashmate`, `Dockerfile`, `scripts/`)
- `InitChain` pre-populates the Orchard pool under `create_sdk_test_data`. Filler notes (random-valid `cmx` + opaque ciphertext) plus a few real owned notes encrypted to fixed ZIP-32 test wallets, all driven by a single seeded RNG for byte-identical re-runs.
- Bulk-seed path rewritten onto GroveDB `append_many_raw` with a single end-of-bake `commit_mmr`, batched for fast bake at N=1M.
- **Pre-baked snapshot**: a `snapshot-bake` CLI dumps the seeded shielded subtree to a file at image-build time; `InitChain` applies it via the `DRIVE_SHIELDED_SNAPSHOT` env var (fast genesis), with a runtime fallback to live seeding when the snapshot is absent.
- Allow `create_sdk_test_data` on Devnet, not just Regtest.
- dashmate gains a generic `buildArgs` map on the docker-build schema (with a config migration to backfill it); `setup_local_network.sh` sets `SDK_TEST_DATA=true` and `CARGO_BUILD_PROFILE=release`; env generation forwards build args into compose.

**Faster note sync** (`platform-wallet`, `rs-sdk`)
- Multi-chunk fetch (`max_query_chunks`, v1 = 4) — up to ~8192 notes per request, ~122 requests for 1M notes.
- `WAL` + `synchronous=NORMAL` + `temp_store=MEMORY` PRAGMAs on the file-backed commitment-tree store, removing the per-`cmx` fsync bottleneck.
- Pull-based streaming sync that interleaves SDK network fetch with wallet tree-append, fetching up to `max_concurrent = 16` chunks concurrently across the network's nodes.

[...]

PR #3733, opened 2026-05-25
fix(swift-sdk): tie shielded completion suppression to a sync generation

QuantumExplorer:

Follow-up to #3603 (shielded send end-to-end), which merged with a
boolean `suppressShieldedCompletionEvents` gate guarding the shielded
sync completion event after `stop`/`clear`.

A reviewer flagged that the boolean is bypassable: the FFI completion
callback re-dispatches onto the `@MainActor`, so a final, already-enqueued
event can land just after `stop`/`clear` returns. Because `start` cleared
the flag, a caller that stops (or clears) and restarts shielded sync in
the **same actor turn** re-opens the gate before the stale, queued
completion task runs — so the previous run's completion leaks into the
new run and is published as if it belonged to it.

## What was done?

Replaced the boolean with a lock-guarded monotonic generation counter
(`ShieldedSyncGenerationCounter`):

- The FFI completion callback snapshots the generation on the callback
  thread, *before* enqueueing the event onto the main actor.
- `stopShieldedSync` / `clearShielded` bump the generation after the Rust
  drain returns.
- `handleShieldedSyncCompleted(_:generation:)` drops any event whose
  snapshot no longer matches the current generation.
- `startShieldedSync` / `syncShieldedNow` no longer reset anything — a new
  run's events carry the current generation and pass the guard, while a
  trailing event from a prior, stopped run still carries the older
  generation and is dropped, even on a same-turn restart.

This is a pure SDK-layer change (no FFI/Rust signature changes); the
example app's `guard isBound` remains as harmless defense-in-depth.

## How Has This Been Tested?

Built the unified iOS framework + SwiftExampleApp for the simulator
(`./build_ios.sh --target sim`) — build succeeded. The change is
behaviorally identical for the common case (events from the live run
publish normally); the generation guard only differs from the old boolean
on the same-actor-turn stop→restart race, which is the bug being fixed.

## Breaking Changes

None. `handleShieldedSyncCompleted` is an internal method; its new
`generation:` parameter is not part of the public API.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of platform wallet shielded sync by preventing stale completion events from being processed after sync is stopped, cleared, or restarted, closing a race that could surface incorrect/old statuses.
* **Tests**
  * Added regression tests validating correct handling of sync completion, restart, and stale-event drop scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dashpay/platform/pull/3733?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManager.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManager.swift
@@ -15,6 +15,19 @@ import DashSDKFFI
 /// Use as a root `@StateObject` and pass via `.environmentObject(_:)`.
 /// Views observe `@Published` properties directly — no coordinator
 /// class in the middle.
+/// Lock-guarded monotonic generation counter, safe to read and bump from
+/// any thread. Used to drop shielded sync completion events that belong
+/// to a generation already superseded by a `stop`/`clear`, even when a
+/// restart happens in the same `@MainActor` turn (a plain boolean gate
+/// can't, because the restart re-opens the gate before the stale,
+/// previously-enqueued completion task runs).
+final class ShieldedSyncGenerationCounter: @unchecked Sendable {
+    private let lock = NSLock()
+    private var value: UInt64 = 0
+    func current() -> UInt64 { lock.withLock { value } }
+    @discardableResult func bump() -> UInt64 { lock.withLock { value &+= 1; return value } }
+}
+
 @MainActor
 public class PlatformWalletManager: ObservableObject {
     // MARK: - Published observables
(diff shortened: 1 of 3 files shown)


thepastaclaw (commented):
## Code Review

The generation-counter approach fixes the stale shielded-sync completion race at the Swift/Rust callback boundary without changing the FFI ownership contract. I did not find a functional or security regression in the implementation, but the PR still needs focused regression coverage for the exact stop/clear versus queued-completion ordering this fix relies on.

🟡 1 suggestion(s)

### Suggestion: Add regression coverage for stale queued shielded completion events

`packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerShieldedSync.swift:68`

This fix depends on a precise ordering guarantee: the FFI callback snapshots `shieldedSyncGeneration` before enqueueing onto `@MainActor`, then `stopShieldedSync` or `clearShielded` bumps the generation before that queued completion is delivered. I verified the implementation follows that pattern, but there is still no Swift test coverage exercising shielded-sync completion delivery or the stop/clear race. A later refactor could reintroduce the original boolean-gate problem while still compiling cleanly. Please add a focused regression test that enqueues a completion under generation N, bumps to N+1 before delivery and asserts the stale event is dropped, then verifies a completion captured under the current generation is published.

_Inline posting hit GitHub HTTP 422, so this verified finding is posted as a top-level review body._

thepastaclaw (commented):
## Code Review

I verified the worktree at `a70eb9d2edb4f1e18610368235c4463343d802e6` and confirmed the only delta since `07421abc64ef74739de6d12347371e62234987f9` is the new `ShieldedSyncGenerationTests.swift` regression suite. The underlying generation-based stale-event guard remains correctly implemented in the production shielded-sync path, and the new tests exercise the same-turn restart and late-straggler cases that previously needed coverage. I did not find any remaining correctness issues or false-positive review findings to carry forward.

QuantumExplorer:
Superseded by #3732. The generation-counter fix here (the `ShieldedSyncGenerationCounter`, the `handleShieldedSyncCompleted(_:generation:)` guard, the stop/clear `bump()`s, the FFI-callback generation snapshot, and `ShieldedSyncGenerationTests`) has been ported onto the `test/sheilded_test_data` branch in commit 7b36f29bb0, reconciled with that branch's dual-progress `resetCurrentShieldedProgress()` changes (both touch the same shielded lifecycle methods). Closing this PR to avoid a merge collision; the fix ships with #3732.

PR #3692, opened 2026-05-20
feat(platform-wallet): load watch-only wallet state from local storage at startup

Claudius-Maginificent:
## Why this PR exists

- **Problem**: `rs-platform-wallet`'s in-memory wallet is a pure cache — empty every time the process starts. Before this PR, the only way to refill it from local storage was to hand back a real `Wallet`, which meant handing back key material. A client that wants to show a previously-synced wallet's balance and history at launch had no way to do that without the seed already unlocked.
- **What breaks without it**: on iOS the Keychain is locked at launch, so a seed-requiring load isn't just slow, it's impossible until the user unlocks — "show balance on launch" fails structurally, not just slowly. Even where the seed is available, skipping this and re-deriving from scratch means minutes of chain rescan on every restart, for a wallet that was fully synced one launch ago. dash-evo-tool's own backend rewrite (#860) hit this directly: reloading a wallet in watch-only mode used to fail with a missing-private-key error the moment a user tried to fund something from its balance, and its cold-start migration path needs previously-synced wallets loaded back automatically after a restart, with no seed prompt.
- **Blocking relationship**: #3968 (the SQLite persistence-reader implementation) is stacked on top of this branch and depends on the final shape of the types introduced here — it isn't mergeable standalone until this lands. (Correcting this PR's own earlier text, which called #3968 "independent" — it's a dependent, stacked PR.)

### Why `ClientWalletStartState` gained new fields — and why dashwallet-ios never needed them

The old shape of this struct carried a real `Wallet` — actual key material — alongside a `ManagedWalletInfo` for everything else. That made "load without the seed" a contradiction: the type itself required a seed-bearing object to exist. This PR replaces the `Wallet` field with `network` + `birth_height` (plain facts) and an `account_manifest` (the account list as public xpubs only) — enough for `Wallet::new_watch_only()` to rebuild a fully-functional watch-only wallet with no key material anywhere on the load path. Signing keys are derived later, on demand, only when something actually needs to sign. The old info field is kept (renamed `core_wallet_info`) since it never held key material to begin with.

[...]

PR #3672, opened 2026-05-19
feat(platform-wallet): keyring_core secret backends — encrypted-file + OS keyring (secrets feature)

Claudius-Maginificent:
## Summary

Adds the `secrets` feature to `rs-platform-wallet-storage` — a `SecretStore` SPI exposing two backends behind the `keyring_core` trait:

- **`EncryptedFileStore`** — passphrase-protected vault file on disk (Argon2id KDF + XChaCha20-Poly1305 AEAD, atomic write via `NamedTempFile::persist`, in-process + cross-process locking via `fd-lock`)
- **OS-native keyring stores** — Linux (libsecret/dbus + linux-keyutils), macOS (apple-native Keychain), Windows (windows-native), via crate features

Stacks on `feat/platform-wallet-sqlite-persistor` (PR #3625, which now includes the storage hardening landed via #3743).

## Public surface

- `SecretStore` trait (re-exports `keyring_core::Credential`-shaped API for callers)
- `SecretBytes` / `SecretString` — `Drop`-zeroizing buffers with `Display` redaction and `forbid PartialEq` (use `subtle::ConstantTimeEq` for comparison); deliberately omits `Deref`/`AsRef` to prevent accidental widening
- `FileStoreError` — typed taxonomy (`WrongPassphrase`, `Corruption`, `Busy`, `MalformedVault`, `VaultTooLarge { found, max }`, `InsecurePermissions`, …) with no raw bytes in `Display`
- `OsKeyringErrorKind` — security-enforcing payload-stripping projection of `keyring_core::Error` (prevents `BadEncoding`/`BadDataFormat` raw bytes from propagating, CWE-209/CWE-532)

## File vault format

Three core types (post the secrets/file collapse + Vec→BTreeMap change):
- `Vault { version, kdf: KdfParams, salt: [u8; SALT_LEN], verify_nonce: [u8; NONCE_LEN], verify_ct: Vec<u8>, entries: BTreeMap<String, EntryBody> }`
- `EntryBody { nonce: [u8; NONCE_LEN], ciphertext: Vec<u8> }` (label is the BTreeMap key — no Vec linear scans)
- `KdfParams { id, m_kib, t, p }` (Argon2id; id-field validated in `enforce_bounds`)

Serde-validated via const-generic `hex_array` adapter on fixed-size byte fields; length mismatches surface as `MalformedVault` at parse time. AEAD AAD is the typed `(format_version, wallet_id, label)` triple — never serialized bytes (C1 invariant).

## Crate features

[...]

PR #3625, opened 2026-05-11
feat(platform-wallet)!: add platform-wallet-storage crate (sqlite persister)

Claudius-Maginificent:
## Why this PR exists

Wallets on Dash Platform need a durable, local home for their **public** state — UTXOs, transactions, account registrations, address pools, identities and their public keys, contacts, asset locks, token balances, DashPay overlays, and platform-address sync snapshots — so a client can restart and pick up where it left off without re-scanning the chain from genesis. Today that storage is bespoke per integrator.

This PR adds **`platform-wallet-storage`**: a ready-to-use, embeddable SQLite persistence backend for `platform-wallet`, plus a small set of operational tools around it. Integrators (DET, the iOS/Android SDKs, any host app) get one `.db` file that holds many wallets, durable across restarts, with online backup, restore, and migration handled for them — and a clear contract that **no private-key material is ever written to that file.**

What ships:

- **SQLite persister** implementing `platform-wallet`'s `PlatformWalletPersistence` — one database file, many wallets, every per-wallet row keyed by `wallet_id`. Backups use SQLite's online backup API (safe under a concurrent writer); restores run under `BEGIN EXCLUSIVE` so peers back off instead of racing the swap.
- **Per-object KV metadata** (`KvStore` / `ObjectId`) for stashing app-managed metadata (aliases, flags, notes, sync hints) alongside wallet objects, with a no-foreign-key soft cascade so a wallet delete cleans up its metadata even when written ahead of sync.
- **Secret storage** (`secrets` module): an in-house encrypted-file vault (Argon2id + XChaCha20-Poly1305) and an OS-keyring backend behind one `SecretStore` front door, with zeroizing wrappers and a typed, secret-free error surface. Fully implemented and **on by default.**
- **Maintenance CLI** (`platform-wallet-storage`): `migrate`, `backup`, `restore`, and `prune`.

For the full API, schema, and threat model, read the crate docs rather than this description:

- `packages/rs-platform-wallet-storage/README.md` — library + CLI usage, flush/load semantics, Cargo features.
- `packages/rs-platform-wallet-storage/SCHEMA.md` — the 23-table SQLite schema, cascade triggers, enum-domain checks, orphan-metadata semantics.
- `packages/rs-platform-wallet-storage/SECRETS.md` — the private-key boundary and the `secrets` backends.

## What integrators get

- **Durable multi-wallet storage** in a single SQLite file, with `prepare_cached` writers and WAL journaling by default.
[...]

PR #3549, opened 2026-04-27
test(platform-wallet): e2e framework + full test suite — triage pins, Found-*/PA-* guards, fail-closed persist, Stage-2 merge

Claudius-Maginificent:
## Dependency status — tracks rust-dashcore's `dash-evo-tool` branch (tip, not a pinned rev)

rust-dashcore's 8 workspace deps now track **`branch = "dash-evo-tool"`** (current tip `18c68d4c`) instead of a fixed `rev`, so future `cargo update` runs follow that branch as it moves. The branch combines PR **#909** (out-of-order UTXO spend fix, #649) and PR **#818** (address-reservation lifecycle: `AddressState` enum, `AddressPool::next_unused_and_reserve`/`release_reservation`/`sweep_expired_reservations`) — sibling forks off `dev`, neither a superset of the other. Re-pin to `dev` directly once both land there. Base branch: **`v4.1-dev`**.

### Recent updates (2026-07-27 — PA-005c stale-doc correction)

- **TEST_SPEC.md's PA-005c status was two days stale — corrected (`d692f2de3f`).** While checking whether rust-dashcore#818's reservation lifecycle had actually been exercised by the last e2e run, audited the verification ledger directly rather than trusting the doc: a full-suite run on **2026-07-25 14:09–14:33 UTC** (189 tests, **148 passed / 41 failed**, zero `trailing bytes` version-skew errors — that blocker had already cleared) shows `pa_005c_receive_address_reservation_lifecycle ... ok`. No commit had updated `TEST_SPEC.md` to reflect it, so it still read "compile-only / not yet live-verified." Fixed the status and recorded the run's evidence (ledger key, log path, counts).
- **Flagged, not fixed: the "145 passed/44 failed" full-suite claim in `b309d640aa`'s message has no matching ledger record.** That commit's own Codex session was briefed to run only the isolated `tk_001b` test ("do NOT run the full suite... a full validation run will happen separately afterward"), and its tool output shows exactly that one test executing. The number doesn't match the real 2026-07-25 148/41 run either. The stack-size fix itself remains independently verified via the isolated `tk_001b` pass — only the specific 145/44 full-suite figure is unbacked.

### Recent updates (2026-07-25 — protocol-version auto-detect fix + stack-size right-sizing)

[...]

PR #3735, opened 2026-05-25
fix(platform-wallet): wallet_id gate on resolver-fed sign entrypoints

Claudius-Maginificent:
## Why this PR exists

**The gap:** Four FFI entrypoints in `rs-platform-wallet-ffi` accept a `MnemonicResolverHandle` plus a caller-supplied `wallet_id_bytes`. None of them verifies the resolver-returned mnemonic actually corresponds to that `wallet_id`. The resolver is host-controlled opaque state — there is no in-process cross-check.

**Threat scenario:** A host (Swift/iOS or any FFI consumer) calls one of these entrypoints with `wallet_id = W_A`, but the registered resolver — whether by host wiring bug, compromised resolver, or malicious resolver — returns the mnemonic for wallet `W_B`. Before this PR, all four entrypoints silently derive from W_B and proceed.

**Per-site impact (with the gate removed) — confirmed by code audit:**

| Entrypoint | Downstream detection | Worst-case state on wrong-seed | Test demonstrating the gate |
|---|---|---|---|
| `sign_with_mnemonic_resolver_and_path` | External only (Platform `InvalidSignature` on submit, or L1 mempool reject on asset-lock TX) | Wrong-key 65-byte sig returned with `rc = 0`; caller ships it to the network; failure surfaces only after round-trip | `sign_with_mnemonic_resolver::tests::wrong_wallet_id_fails_closed_with_wrong_seed_tag` |
| `derive_and_persist_identity_keys` | **NONE** | Persister callback writes Keychain rows tagged `wallet_id = W_A` but holding W_B's pubkey/privkey/path. Subsequent registration succeeds with internally-consistent W_B keys, creating an identity W_A's seed cannot sign for. **Irrecoverable without manual Keychain surgery.** | `identity_derive_and_persist::tests::wrong_wallet_id_fails_closed_before_persisting_anything` |
| `derive_identity_key_at_slot_with_resolver` | NONE inside FFI | 32-byte W_B private scalar + WIF crosses the FFI boundary with `rc = Success`; caller uses it freely | `derive_identity_key_at_slot::tests::wrong_wallet_id_fails_closed_with_wrong_seed_tag` |
| `platform_wallet_manager_bind_shielded` (shielded feature) | **NONE** — `PlatformWallet::bind_shielded` (`platform_wallet.rs:338-427`) uses `self.wallet_id` and the caller-supplied seed with zero cross-check | W_B's Orchard FVK/IVK/OVK durably installed under `SubwalletId { wallet_id: W_A }`. Sync flushes W_B's notes into W_A's persister rows; spend paths can authorize W_B's notes under W_A. **Cross-wallet fund visibility and spend leakage.** | `shielded_sync::tests::wrong_wallet_id_fails_closed` (requires `--features shielded`) |

Note that `rs-platform-wallet` itself does not catch this: pure-Rust callers instantiate `Wallet` objects whose `wallet_id` is derived from their own root — the mismatch is unreachable from Rust. The vulnerability is exclusively expressible across the FFI trust boundary, which is why the gate lives in `rs-platform-wallet-ffi`.

Each row's test fails on `v3.1-dev` without this PR's gate and passes once the gate is applied — verified locally against the rebased commit.

**Prerequisite/blocking relationship:** This PR is the logical prerequisite for PR #3692 (seedless watch-only rehydration), which removes the existing load-time wrong-seed gate in `rs-platform-wallet`. Without #3735 landing first, #3692 would briefly create a window where any seed could sign for any `wallet_id`. Decoupled here so the security fix ships to `v3.1-dev` on its own merits.

Four FFI sign entrypoints in `rs-platform-wallet-ffi` take a `MnemonicResolverHandle` plus a `wallet_id_bytes` arg and derive signing material from the mnemonic the resolver returns. **None of them previously checked that the seed actually belongs to that wallet_id.** A host that called any of them with a mismatched `(wallet_id, mnemonic)` pair — whether by host bug or by a malicious resolver — would happily derive and sign with the wrong key.

This patch closes that gap.

It also unblocks the seedless-load model in PR #3692: that PR removes a load-time wrong-seed gate, so the sign-time gate has to exist before that change can land without regressing wrong-seed protection. Decoupled here so the security fix can ship to `v3.1-dev` on its own merits, ahead of the rehydration model rework.

### New helper: `crate::sign_gate::verify_seed_matches_wallet_id`

`packages/rs-platform-wallet-ffi/src/sign_gate.rs` (new):

```rust
#[inline]
#[must_use]
pub fn verify_seed_matches_wallet_id(
    root_pub: &ExtendedPubKey,
    expected_wallet_id: &[u8; 32],
) -> bool {
    let root_extended = RootExtendedPubKey::from_extended_pub_key(root_pub);
    let derived = Wallet::compute_wallet_id_from_root_extended_pub_key(&root_extended);
    bool::from(derived.ct_eq(expected_wallet_id))
}
```

- **Constant-time** comparison via `subtle::ConstantTimeEq` — no early-return on byte mismatch.
- **`#[must_use]`** so a caller can't silently drop the verdict.
- Operates on the public half of the root extended key — no private material flows into the comparator.
- New dep: `subtle = "2"` (well-known constant-time-comparison crate, single line added to `Cargo.lock`, no transitive bumps).

### New result code: `ErrorWrongSeedForWallet = 15`

`PlatformWalletFFIResultCode::ErrorWrongSeedForWallet = 15` in `error.rs`. Newly allocated slot 15. Slot 13 is taken by `ErrorArithmeticOverflow` (#3549), slot 14 by `ErrorNoSelectableInputs`; 15 was the smallest free integer.

For the byte-tagged `sign_with_mnemonic_resolver_and_path` path, a new u8 tag `SIGN_WITH_RESOLVER_ERR_WRONG_SEED = 11` is also added in that module.

### Gate wired into 4 entrypoints

| Entrypoint | File | Gate location |
|---|---|---|
| `dash_sdk_sign_with_mnemonic_resolver_and_path` | `sign_with_mnemonic_resolver.rs` | After master derivation, before `derive_priv` |
| `dash_sdk_derive_and_persist_identity_keys` | `identity_derive_and_persist.rs` | Before persister callback ever fires |
| `dash_sdk_derive_identity_key_at_slot_with_resolver` | `derive_identity_key_at_slot.rs` | Before `derive_at_slot_inner` |
| `platform_wallet_manager_bind_shielded` (`shielded` feature) | `shielded_sync.rs` | Before `PLATFORM_WALLET_MANAGER_STORAGE` touch |

At each site, on a mismatch:
- `master.private_key.non_secure_erase()` — derived master xpriv zeroed
- Caller-owned output buffer wiped (`sig_buf` zeroed via `ptr::write_bytes`; `out_row` left at `IdentityKeyPreviewFFI::empty()`; `*out_signature_len = 0`)
- Typed error returned (`ErrorWrongSeedForWallet` for structured paths, `SIGN_WITH_RESOLVER_ERR_WRONG_SEED` for the u8-tagged path)
- Persister callback / coordinator handoff never reached

Two of the four sites (`derive_identity_key_at_slot`, `shielded_sync`) `non_secure_erase` the gate-master unconditionally — defense in depth, since the gate-master is throwaway whether the verdict is true or false. The other two keep the master alive past the gate because the same master is then used for `derive_priv` and is erased downstream via the existing pattern.

## How Has This Been Tested?

```bash
cargo fmt --all -- --check
cargo clippy -p platform-wallet-ffi --all-targets -- -D warnings
cargo clippy -p platform-wallet-ffi --all-targets --features shielded -- -D warnings
cargo check --workspace
cargo test -p platform-wallet-ffi
cargo test -p platform-wallet-ffi --features shielded shielded_sync
cargo test -p platform-wallet           # workspace regression sanity
```

All green. **83 unit + 26 comprehensive + 5 integration tests pass**; 132 platform-wallet tests still pass; new gate tests included:

| Test | Asserts |
|---|---|
| `sign_gate::tests::matching_seed_passes` | CT comparator returns true on matching seed |
| `sign_gate::tests::mismatched_seed_fails` | CT comparator returns false on distinct seeds |
| `sign_with_mnemonic_resolver::tests::happy_path_signs_and_returns_signature` | `sig_len == 65` (compact-recoverable secp256k1) |
| `sign_with_mnemonic_resolver::tests::wrong_wallet_id_fails_closed_with_wrong_seed_tag` | rc=-1, tag=11, `sig_len==0`, every byte of 128-byte `sig_buf` is zero |
| `identity_derive_and_persist::tests::happy_path_persists_three_keys_and_returns_pubkeys` | gate doesn't false-positive on matching seed |
| `identity_derive_and_persist::tests::wrong_wallet_id_fails_closed_before_persisting_anything` | code=`ErrorWrongSeedForWallet`, `out.count==0`, `out.items.is_null()`, persister callback row count is zero |
| `derive_identity_key_at_slot::tests::matching_seed_returns_derived_key` | DIP-9 path, 33-byte pubkey, non-null WIF, non-zero private scalar |
| `derive_identity_key_at_slot::tests::wrong_wallet_id_fails_closed_with_wrong_seed_tag` | all output fields null/zero, every byte of `private_key_bytes` is zero |
| `shielded_sync::tests::wrong_wallet_id_fails_closed` | code=`ErrorWrongSeedForWallet` via `NULL_HANDLE` — gate short-circuits before manager-storage touch |
| `shielded_sync::tests::null_resolver_handle_rejected` | `check_ptr!` ordering locked in |

Every wrong-seed test asserts on **substance** — typed tag + output-buffer state + side-effect counters — not "did the function run".

## Breaking Changes

None for shipped consumers. The change is purely additive:
- New helper module (does not modify existing public APIs).
- New result code variant at a previously unoccupied enum slot.
- Gate insertions return new error tags only on the wrong-seed mismatch path — flows with correct `(wallet_id, mnemonic)` pairs behave identically to before.

No `!` in the title.

## Security note

This patch is logically the prerequisite for PR #3692 (seedless watch-only rehydration), which removes a load-time wrong-seed gate. Without this PR landed first, #3692 would briefly create a window where the same wallet_id could be signed by any seed. Once this lands, the merge-up cycle (`v3.1-dev → #3625 → #3672 → #3692`) will dedupe the duplicate sign-gate commits sitting on #3692 naturally — no force-push to that PR needed.

## Checklist:

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4869,6 +4869,7 @@ dependencies = [
  "platform-wallet",
  "rs-sdk-ffi",
  "serde_json",
+ "subtle",
  "tempfile",
  "tokio",
  "tracing",
--- a/packages/rs-platform-wallet-ffi/Cargo.toml
+++ b/packages/rs-platform-wallet-ffi/Cargo.toml
@@ -54,6 +54,10 @@ bs58 = "0.5"
 # Zeroize intermediate key material crossing the FFI boundary.
 zeroize = { version = "1", features = ["derive"] }
 
+# Constant-time `wallet_id` compare in `sign_gate`, mirroring the
+# upstream gate previously in `manager::rehydrate`.
+subtle = "2"
+
 [dev-dependencies]
 tempfile = "3.8"
 dpp = { path = "../rs-dpp", features = ["fixtures-and-mocks"] }
--- a/packages/rs-platform-wallet-ffi/src/derive_identity_key_at_slot.rs
+++ b/packages/rs-platform-wallet-ffi/src/derive_identity_key_at_slot.rs
@@ -12,7 +12,10 @@ use zeroize::Zeroizing;
 use crate::error::*;
 use crate::identity_key_preview::IdentityKeyPreviewFFI;
 use crate::identity_keys_from_mnemonic::parse_mnemonic_any_language;
+use crate::sign_gate::verify_seed_matches_wallet_id;
 use crate::{check_ptr, unwrap_result_or_return};
+use dashcore::secp256k1::Secp256k1;
+use key_wallet::bip32::ExtendedPubKey;
 use rs_sdk_ffi::{
     mnemonic_resolver_result, MnemonicResolverHandle, MNEMONIC_RESOLVER_BUFFER_CAPACITY,
 };
@@ -182,6 +185,32 @@ pub unsafe extern "C" fn dash_sdk_derive_identity_key_at_slot_with_resolver(
 
     let mnemonic_str = unwrap_result_or_return!(std::str::from_utf8(&mnemonic_buf[..mnemonic_len]));
 
+    // Fail-closed wrong-seed gate before any signing-relevant material
+    // is materialized for the caller. Derive the root xpub from the
+    // resolver-supplied seed in its own scope so the master xpriv +
+    // its SecretKey are non-secure-erased before the inner derivation
+    // path runs (which re-derives from the seed via the shared helper).
+    {
+        let mnemonic = unwrap_result_or_return!(parse_mnemonic_any_language(mnemonic_str));
+        let seed: Zeroizing<[u8; 64]> = Zeroizing::new(mnemonic.to_seed(""));
+        drop(mnemonic);
+        let kw_network: Network = network.into();
+        let mut master =
+            unwrap_result_or_return!(ExtendedPrivKey::new_master(kw_network, seed.as_ref()));
+        let secp = Secp256k1::new();
+        let root_xpub = ExtendedPubKey::from_priv(&secp, &master);
+        let mut wallet_id_expected = [0u8; 32];
+        std::ptr::copy_nonoverlapping(wallet_id_bytes, wallet_id_expected.as_mut_ptr(), 32);
+        let gate_ok = verify_seed_matches_wallet_id(&root_xpub, &wallet_id_expected);
+        master.private_key.non_secure_erase();
+        if !gate_ok {
+            return PlatformWalletFFIResult::err(
+                PlatformWalletFFIResultCode::ErrorWrongSeedForWallet,
+                "wrong seed for wallet (derive_identity_key_at_slot_with_resolver gate)",
+            );
+        }
+    }
+
     derive_at_slot_inner(
         mnemonic_str,
         "",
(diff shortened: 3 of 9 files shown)


QuantumExplorer (changes requested):
I believe this should be closed, as it not a real issue. The client application (iOS) that has the mnemonic is not going to compromise itself. Like if it has the seed phrase what more attack is there?

PR #3742, opened 2026-05-25
fix(platform-wallet): fix spv client deadlocking himself when trying to stop

### LINKED ISSUE #3741: SpvRuntime::stop() deadlocks against running SpvRuntime::run()
## Summary

`SpvRuntime::stop()` cannot terminate a running `SpvRuntime::run()`. The two methods deadlock against each other on the `RwLock<Option<SpvClient>>` guard. **`stop()` must become synchronous and non-locking** (or the lock layout must change) so it can signal termination without contending for the write lock that `run()` is blocking on.

## Location

`packages/rs-platform-wallet/src/spv/runtime.rs:132-159`

```rust
pub async fn run(&self, config: ClientConfig) -> Result<(), PlatformWalletError> {
    self.start(config).await?;
    let client_guard = self.client.read().await;          // ① read lock acquired
    let client = client_guard.as_ref()
        .ok_or(PlatformWalletError::SpvNotRunning)?;
    let result = client.run().await.map_err(...);          // ② held across the entire sync loop
    drop(client_guard);
    let mut client = self.client.write().await;           // ③ only after run() returns
    let _ = client.take();
    result
}

pub async fn stop(&self) -> Result<(), PlatformWalletError> {
    let mut client = self.client.write().await;           // ④ needs write lock
    if let Some(c) = client.take() {
        c.stop().await...;                                 // ⑤ never reached if step ② is alive
    }
}
```

## Why it deadlocks

1. `run()` holds the **read guard** on `self.client` for the full lifetime of `client.run().await` (the SPV sync loop — runs for hours).
2. A concurrent call to `stop()` queues a write-lock waiter on `self.client`.
3. `tokio::sync::RwLock` is writer-priority: a queued writer also blocks new readers, so even other readers cannot ride along.
4. The only way to make `client.run()` return — and let the read guard drop — is to flip `DashSpvClient`'s internal `Arc<watch::Sender<bool>>` via `client.stop()`.
5. The only place that calls `client.stop()` is line ⑤, gated on acquiring the write lock that step 2 is blocked on.

**Circular wait. Any concurrent `stop()` while sync is running hangs forever.** Same applies to any other lifecycle operation that needs the write lock (e.g. a re-`start()`).

## Where the regression came from

| | |
|---|---|
| **Introducing PR** | #3730 — `fix(platform-wallet): spv client deadlocking when sending a tx` |
| **Commit** | `9f3bfa3ca3` |
| **Merged** | 2026-05-22 |

The original shape (commit `bee80e93bf` by Ivan Shumkov, shipped via #3531 on 2026-04-24) was deadlock-free by construction: `run()` accepted an outer `CancellationToken` parameter; cancellation propagated *inward* into `client.run(token)`, which returned cleanly, dropped the read guard at the end of an inner block, and *then* called `self.stop().await` outside the guard scope.

PR #3730 had to adapt to rust-dashcore [#772](https://github.com/dashpay/rust-dashcore/pull/772) (Kevin Rombach, merged 2026-05-18) which removed the external `CancellationToken` parameter from `DashSpvClient::run()` and replaced it with an internal `Arc<watch::Sender<bool>>` flipped via `client.stop()`. The adaptation dropped the outer `cancel_token` from `SpvRuntime::run()` **without replacing the termination mechanism** — the only remaining way to make `client.run()` return is now `client.stop()`, which is reachable only after `stop()` acquires the write lock, which can never be acquired while `run()` holds the read guard.

So: a 3-day-old regression from fixing a *different* SPV deadlock.

## What we need

**`SpvRuntime::stop()` must be synchronous and non-locking** — or, equivalently, must be able to signal `client.run()` to terminate *without* needing to acquire the `RwLock` write guard on `self.client`. Once `client.run()` returns and the read guard drops, the lifecycle teardown (take the client out of the `Option`, drop it) can happen anywhere — but the signaling step must not contend with the read guard.

The signaling needs to be reachable from synchronous contexts (e.g. a `std::panic::set_hook` callback — the original `cancel_background()` use case in PR #3549's e2e harness needs this same property).

## Proposed fix shapes

| # | Approach | Sketch |
|---|---|---|
| **A** | **`Arc<DashSpvClient>` instead of `Option<DashSpvClient>` under the lock.** Both `run()` and `stop()` `Arc::clone()` from the lock and then operate without holding it. `stop()` becomes near-synchronous: clone the Arc, call `client.stop()` (which only flips an internal watch — already designed to be cheap/sync-ish), done. The platform-wallet write lock is then only used for swap operations, not held across awaits. | Smallest patch, idiomatic for tokio-based SPV/blockchain clients. |
| **B** | **Split the lock**: one cell for the client handle, one parallel `watch::Sender<RuntimeState>` for lifecycle signaling. `stop()` flips the watch directly; `run()` selects on the watch alongside `client.run()`. | More machinery; closest to the campaign-tree (#3549) cancel-token design. |
| **C** | **`take()`-and-restore**: `run()` removes the client from the `Option` at the start, owns it directly during the sync loop, puts it back when sync exits. `stop()` only needs to flip a separate watch. | Conceptually clearest; loses concurrent state queries unless they also extract. |

Recommended: **A**, with the constraint that `stop()` must be callable synchronously (just `client_arc.stop()` if dash-spv exposes a sync flip, otherwise a fire-and-forget `tokio::spawn`).

## Test plan

- [ ] Unit test in `runtime.rs`: spawn `run()` on a mocked `SpvClient`, then call `stop()` from another task — assert `stop()` returns within a timeout (currently hangs forever).
- [ ] Unit test: panic-hook-style sync caller can fire termination without an async context (relevant for PR #3549's e2e harness).
- [ ] `cargo test -p platform-wallet --lib` passes after the fix.

## Severity rationale (high)

This makes graceful shutdown impossible whenever sync is running. SwiftExampleApp + the FFI surface in `packages/rs-platform-wallet-ffi/src/spv.rs:269-275` cannot reliably stop the SPV client between app sessions; the e2e harness in PR #3549 can't drive its panic-hook lifecycle teardown; the campaign tree pins an older rust-dashcore rev specifically because the post-#772 platform-wallet shape strands them.

## Related

- rust-dashcore [#772](https://github.com/dashpay/rust-dashcore/pull/772) — upstream API change that triggered the adaptation.
- dashpay/platform PR #3730 — the introducing commit.
- dashpay/platform PR #3549 (e2e framework) — pins an older rust-dashcore rev because of this and other shape mismatches; its `SpvRuntime` has the deadlock-free outer-token design preserved.
- dashpay/platform PR #3651 — was carving cancellation out of #3549 onto v3.1-dev; stripped because this issue's resolution belongs to the SPV runtime owner, not a backport author.

ZocoLini:
The run method in the spa client wrapper was holding a read() lock over the inner spv client while the client was running, and stop tried to take a write() lock, what I do in this PR is clonning the Client, Client fields are Arc so we preserve reference to the same memory space.

Other way to solve the issue would be to take a read() lock on stop, and stop the client, then run() would continue and take a write() lock to replace de Some with a None, maybe it sounds like Chinese without having the code in front but yeah, two ways to solve it and I went with the one its the easiest to read and understand. Anyway, I am trying to refactor the SpvClient usage in platform wallet but first I need my integrations tests to get in so I can make sure I don't make any regression

Closes https://github.com/dashpay/platform/issues/3741

## Checklist:

## Summary by CodeRabbit

* **Refactor**
  * Improved SPV wallet runtime client handling to enhance resource management efficiency.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dashpay/platform/pull/3742?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/rs-platform-wallet/src/spv/runtime.rs
+++ b/packages/rs-platform-wallet/src/spv/runtime.rs
@@ -138,14 +138,15 @@ impl SpvRuntime {
         let client_guard = self.client.read().await;
         let client = client_guard
             .as_ref()
-            .ok_or(PlatformWalletError::SpvNotRunning)?;
+            .ok_or(PlatformWalletError::SpvNotRunning)?
+            .clone();
+        drop(client_guard);
 
         let result = client
             .run()
             .await
             .map_err(|e| PlatformWalletError::SpvError(e.to_string()));
 
-        drop(client_guard);
         let mut client = self.client.write().await;
         let _ = client.take();
 


thepastaclaw (changes requested):
## Code Review

The deadlock fix in `SpvRuntime::run()` is directionally correct, but the new clone-and-drop pattern introduces a real lifecycle race on stop-then-restart. `SpvRuntime` can now lose track of a newly started SPV client while that client keeps running, which makes the runtime report `SpvNotRunning` even though the background sync loop is still alive.

🔴 1 blocking

### Review details

<details>

```

- [BLOCKING] lines 138-151: Exiting `run()` can clear a newer client installed after `stop()`
  `run()` now clones the current `DashSpvClient`, drops the read lock, awaits `client.run()`, and then unconditionally executes `self.client.write().await.take()`. That cleanup is no longer tied to the generation that originally started the run loop. `stop()` removes the current client from `self.client` before the old `client.run()` future has necessarily returned, so a caller can immediately `start()` again and install a fresh client in the slot. When the old run loop finally exits, its unconditional `take()` removes that new client, leaving the restarted SPV sync running only through the cloned `Arc` state inside `dash-spv` while `SpvRuntime::is_started()` becomes false and later `stop()` calls cannot reach it. The upstream `DashSpvClient::clone()` is a shallow clone over shared `Arc` fields, and the same crate already documents this stale-cleanup pattern as a real bug in `identity_sync` and `shielded_sync`, so `SpvRuntime` needs the same generation/ownership check before clearing the slot.

```
</details>

*Reviewed commit: `da12a71a`*

_GitHub rejected inline review comments for this update, so the findings are posted body-only._

lklimek:
@thepastaclaw please describe the race condition with an example flow.

thepastaclaw:
Example flow I had in mind:

1. Task A calls `SpvRuntime::run(config_a)`. `start()` installs client A in `self.client`.
2. `run()` clones client A, drops the read guard, then awaits `client_a.run()`.
3. Task B calls `stop()`. It takes `self.client` (so the slot becomes `None`) and calls `client_a.stop()`. The old `client_a.run()` future may not have returned yet.
4. Before that old future returns, Task C calls `run(config_b)` / `start(config_b)`, which installs client B in `self.client`.
5. The old `client_a.run()` future finally returns. Task A resumes and unconditionally runs `self.client.write().await.take()`.

At step 5, the slot can now contain client B, not client A, so the cleanup for the old run removes the newly-started client. After that, `SpvRuntime::is_started()` reports false and later `stop()` cannot reach client B, even though client B may still be alive/running through its own shared state.

So the cleanup needs to be ownership/generation-aware: only clear the slot if it still contains the same client/generation that this `run()` cloned. If concurrent restart is meant to be invalid instead, then `start()`/`run()` should probably enforce that explicitly rather than allowing the stale cleanup path.

QuantumExplorer (approved):

PR #3748, opened 2026-05-27
feat(wasm-sdk): first-class devnet support with trusted-context prefetch

PastaPastaPasta:

The WASM SDK has no first-class concept of devnets. To reach a devnet today, consumers like `native-platform-explorer` must masquerade as testnet with `{ network: 'testnet', trusted: false, addresses: [...] }`, which disables proof verification entirely. As soon as a proof-bearing response comes back, `WasmContext::get_quorum_public_key` hard-errors with:

> "Non-trusted mode is not supported in WASM. Please use the trusted SDK builders (new_mainnet_trusted or new_testnet_trusted) instead."

(The error message itself is also stale — it references builder names that no longer exist.)

After v3.1.0-dev.6's gRPC trailer-transport fix lands, this error becomes the dominant failure mode on every devnet state query.

## What was done?

Adds devnets as a first-class network across three packages:

- **`rs-sdk-trusted-context-provider`**: Devnet quorum URLs now follow the testnet/mainnet shape — `https://quorums.<name>.networks.dash.org` (was `https://quorums.devnet.<name>.networks.dash.org`). All validation rules for devnet names are unchanged.
- **`wasm-sdk`**:
  - `WasmTrustedContext::prefetchDevnet(devnet_name)` — async factory parallel to `prefetchTestnet`. URL derived from the devnet name.
  - `WasmTrustedContext::prefetchDevnetWithUrl(base_url)` — escape hatch for when the public DNS isn't deployed yet.
  - `WasmSdkBuilder::newDevnet()` factory + a `"devnet"` arm on `withAddresses`.
  - Updated the stale `WasmContext::get_quorum_public_key` error message to reference current method names (`prefetchMainnet/Testnet/Devnet/Local`).
- **`js-evo-sdk`**:
  - Extends `EvoSDKOptions.network` to `'mainnet' | 'testnet' | 'local' | 'devnet'`.
  - New `devnetName` and `quorumUrl` options.
  - `EvoSDK.devnet(name, options)` and `EvoSDK.devnetTrusted(name, options)` static factories.
  - Constructor-time validation rejects misconfigurations: missing `devnetName` and `addresses`, non-trusted devnet without addresses, `quorumUrl` without `trusted`, `quorumUrl` on non-devnet networks.

Once published, the migration for the explorer is `new EvoSDK({ addresses, network: 'testnet', trusted: false })` → `EvoSDK.devnetTrusted('paloma')` (with proof verification restored).

## How Has This Been Tested?

- `cargo test -p rs-sdk-trusted-context-provider`: URL-builder tests pass with the new pattern (`test_get_quorum_base_url`, `test_devnet_without_name_returns_error`, `test_invalid_devnet_names`). One pre-existing DNS-dependent test (`test_known_contracts`) fails the same way without these changes; not a regression.
- `yarn workspace @dashevo/evo-sdk test:unit`: **200 passing** (was 198), including 9 new tests covering both happy paths and every new validation rule for `EvoSDK.devnet` / `EvoSDK.devnetTrusted` (including \"devnet with only addresses, no devnetName\" and \"reject quorumUrl when trusted is false\").
- `cargo check -p wasm-sdk` / wasm32 build: could not run on the dev machine — local Apple clang lacks the wasm32 backend. Rust changes are small (~50 lines) and copy-of-existing-patterns from `prefetch_testnet` / `prefetch_local_with_url`. An independent code-review pass flagged a stale doc comment and a silently-discarded \`quorumUrl\` on non-trusted devnet; both addressed. CI will exercise the full wasm build.

## Breaking Changes

- The default devnet quorum URL pattern changes from \`https://quorums.devnet.<name>.networks.dash.org\` → \`https://quorums.<name>.networks.dash.org\`. This is a behavior change for any existing caller of \`TrustedHttpContextProvider::new(Network::Devnet, Some(name), ...)\` that was relying on the old hostname. No public devnet quorum DNS has been deployed yet, so in practice no live caller depends on the old shape — but flagging here for transparency. The escape hatch (\`new_with_url\` / \`prefetchDevnetWithUrl\`) lets callers pin a different URL if needed.

## Checklist:

- [ ] I have added \"!\" to the title and described breaking changes in the corresponding section if my code contains any

## Summary by CodeRabbit

* **New Features**
  * Devnet support: named devnets and quorum-URL override; devnet usable in address-based or trusted modes. New convenience helpers for devnet and trusted-devnet SDK creation; WASM builder and trusted-context prefetch helpers for devnet and explicit URLs.

* **Documentation**
  * Updated devnet quorum endpoint URL format and examples.

* **Tests**
  * Added unit tests for devnet behaviors, validations, and trusted-context handling.

* **Bug Fixes / Validation**
  * Stronger devnet name and quorum-URL validations (including reserved names and HTTPS rules).

[Review Change Stack](https://app.coderabbit.ai/change-stack/dashpay/platform/pull/3748?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/js-evo-sdk/src/sdk.ts
+++ b/packages/js-evo-sdk/src/sdk.ts
@@ -30,16 +30,26 @@ export interface ConnectionOptions {
 }
 
 export interface EvoSDKOptions extends ConnectionOptions {
-  network?: 'testnet' | 'mainnet' | 'local';
+  network?: 'testnet' | 'mainnet' | 'local' | 'devnet';
   trusted?: boolean;
-  // Custom masternode addresses. When provided, network and trusted options are ignored.
+  // Custom masternode addresses to seed the SDK with. `network` still
+  // controls which Network enum the underlying builder uses (and, for
+  // trusted mode, which quorums endpoint is prefetched); the addresses
+  // here replace the network's built-in defaults at seed time.
   // Example: ['https://127.0.0.1:1443', 'https://192.168.1.100:1443']
   addresses?: string[];
+  // Short name of the devnet (e.g. 'paloma'). Required when network === 'devnet'
+  // unless explicit addresses are provided AND trusted is false.
+  devnetName?: string;
+  // Optional override for the trusted devnet quorum base URL. When omitted,
+  // the URL is derived as `https://quorums.<devnetName>.networks.dash.org`.
+  // Only consulted when trusted === true && network === 'devnet'.
+  quorumUrl?: string;
 }
 
 export class EvoSDK {
   private wasmSdk?: wasm.WasmSdk;
-  private options: Required<Pick<EvoSDKOptions, 'network' | 'trusted'>> & ConnectionOptions & { addresses?: string[] };
+  private options: Required<Pick<EvoSDKOptions, 'network' | 'trusted'>> & ConnectionOptions & { addresses?: string[]; devnetName?: string; quorumUrl?: string };
 
   public addresses!: AddressesFacade;
   public documents!: DocumentsFacade;
@@ -56,8 +66,27 @@ export class EvoSDK {
   public shielded!: ShieldedFacade;
   constructor(options: EvoSDKOptions = {}) {
     // Apply defaults while preserving any future connection options
-    const { network = 'testnet', trusted = false, addresses, ...connection } = options;
-    this.options = { network, trusted, addresses, ...connection };
+    const { network = 'testnet', trusted = false, addresses, devnetName, quorumUrl, ...connection } = options;
+
+    if (network === 'devnet') {
+      const hasAddresses = !!(addresses && addresses.length > 0);
+      if (!devnetName && !hasAddresses) {
+        throw new Error("EvoSDK: network 'devnet' requires either devnetName or explicit addresses");
+      }
+      if (trusted && !devnetName) {
+        throw new Error("EvoSDK: trusted devnet requires devnetName (used to derive the quorum URL)");
+      }
+      if (!trusted && !hasAddresses) {
+        throw new Error("EvoSDK: non-trusted devnet requires explicit addresses (no addresses can be discovered without a trusted context)");
+      }
+      if (quorumUrl && !trusted) {
+        throw new Error("EvoSDK: quorumUrl is only meaningful when trusted === true");
+      }
+    } else if (quorumUrl) {
+      throw new Error("EvoSDK: quorumUrl is only valid when network === 'devnet'");
+    }
+
+    this.options = { network, trusted, addresses, devnetName, quorumUrl, ...connection };
 
     this.addresses = new AddressesFacade(this);
     this.documents = new DocumentsFacade(this);
@@ -96,7 +125,7 @@ export class EvoSDK {
     }
     await initWasm();
 
-    const { network, trusted, version, proofs, settings, logs, addresses } = this.options;
+    const { network, trusted, version, proofs, settings, logs, addresses, devnetName, quorumUrl } = this.options;
 
     // Prefetch trusted context only when trusted mode is requested
     let context: wasm.WasmTrustedContext | undefined;
@@ -107,6 +136,13 @@ export class EvoSDK {
         context = await wasm.WasmTrustedContext.prefetchTestnet();
       } else if (network === 'local') {
         context = await wasm.WasmTrustedContext.prefetchLocal();
+      } else if (network === 'devnet') {
+        if (!devnetName) {
+          throw new Error("EvoSDK: trusted devnet requires devnetName");
+        }
+        context = quorumUrl
+          ? await wasm.WasmTrustedContext.prefetchDevnetWithUrl(quorumUrl)
+          : await wasm.WasmTrustedContext.prefetchDevnet(devnetName);
       } else {
         throw new Error(`Unknown network: ${network}`);
       }
@@ -122,6 +158,8 @@ export class EvoSDK {
       builder = wasm.WasmSdkBuilder.testnet();
     } else if (network === 'local') {
       builder = wasm.WasmSdkBuilder.local();
+    } else if (network === 'devnet') {
+      builder = wasm.WasmSdkBuilder.newDevnet();
     } else {
       throw new Error(`Unknown network: ${network}`);
     }
@@ -182,10 +220,41 @@ export class EvoSDK {
   static localTrusted(options: ConnectionOptions = {}): EvoSDK { return new EvoSDK({ network: 'local', trusted: true, ...options }); }
 
   /**
+   * Create an EvoSDK instance configured for a devnet, without trusted-context
+   * proof verification. Requires either `devnetName` or explicit `addresses`
+   * in `options`. Proof-bearing queries will fail unless paired with a
+   * trusted context — for proof verification on devnet, use
+   * `EvoSDK.devnetTrusted` instead.
+   */
+  static devnet(devnetName: string, options: ConnectionOptions & { addresses?: string[] } = {}): EvoSDK {
+    return new EvoSDK({ network: 'devnet', devnetName, ...options });
+  }
+
+  /**
+   * Create an EvoSDK instance configured for a devnet with a trusted context.
+   *
+   * The trusted context is prefetched from
+   * `https://quorums.<devnetName>.networks.dash.org` by default. Pass
+   * `quorumUrl` to override (useful when the public DNS is not yet deployed).
+   *
+   * @example
+   * ```typescript
+   * const sdk = EvoSDK.devnetTrusted('paloma');
+   * await sdk.connect();
+   * ```
+   */
+  static devnetTrusted(
+    devnetName: string,
+    options: ConnectionOptions & { quorumUrl?: string } = {},
+  ): EvoSDK {
+    return new EvoSDK({ network: 'devnet', devnetName, trusted: true, ...options });
+  }
+
+  /**
    * Create an EvoSDK instance configured with specific masternode addresses.
    *
    * @param addresses - Array of HTTPS URLs to masternodes (e.g., ['https://127.0.0.1:1443'])
-   * @param network - Network identifier: 'mainnet', 'testnet' (default: 'testnet')
+   * @param network - Network identifier: 'mainnet', 'testnet', 'devnet', or 'local' (default: 'testnet')
    * @param options - Additional connection options
    * @returns A configured EvoSDK instance (not yet connected - call .connect() to establish connection)
    *
@@ -195,7 +264,7 @@ export class EvoSDK {
    * await sdk.connect();
    * ```
    */
-  static withAddresses(addresses: string[], network: 'mainnet' | 'testnet' | 'local' = 'testnet', options: ConnectionOptions = {}): EvoSDK {
+  static withAddresses(addresses: string[], network: 'mainnet' | 'testnet' | 'local' | 'devnet' = 'testnet', options: ConnectionOptions & { devnetName?: string } = {}): EvoSDK {
     return new EvoSDK({ addresses, network, ...options });
   }
 }
--- a/packages/js-evo-sdk/src/sdk.ts
+++ b/packages/js-evo-sdk/src/sdk.ts
@@ -82,6 +82,10 @@ export class EvoSDK {
       } else if (!hasAddresses) {
         throw new Error("EvoSDK: non-trusted devnet requires explicit addresses (no addresses can be discovered without a trusted context)");
       }
+    } else if (devnetName) {
+      // Surface a likely typo (e.g. network: 'testent' + devnetName: 'paloma')
+      // — devnetName has no effect outside network === 'devnet'.
+      throw new Error("EvoSDK: devnetName is only valid when network === 'devnet'");
     }
     if (quorumUrl && !trusted) {
       throw new Error("EvoSDK: quorumUrl is only meaningful when trusted === true");
--- a/packages/js-evo-sdk/src/sdk.ts
+++ b/packages/js-evo-sdk/src/sdk.ts
@@ -39,7 +39,9 @@ export interface EvoSDKOptions extends ConnectionOptions {
   // Example: ['https://127.0.0.1:1443', 'https://192.168.1.100:1443']
   addresses?: string[];
   // Short name of the devnet (e.g. 'paloma'). Required when network === 'devnet'
-  // unless explicit addresses are provided AND trusted is false.
+  // AND trusted === true (used to derive the quorum URL). When trusted === false,
+  // explicit `addresses` are mandatory and `devnetName` alone is not sufficient
+  // — no masternode addresses can be discovered without a trusted context.
   devnetName?: string;
   // Optional override for the trusted devnet quorum base URL. When omitted,
   // the URL is derived as `https://quorums.<devnetName>.networks.dash.org`.
@@ -221,9 +223,10 @@ export class EvoSDK {
 
   /**
    * Create an EvoSDK instance configured for a devnet, without trusted-context
-   * proof verification. Requires either `devnetName` or explicit `addresses`
-   * in `options`. Proof-bearing queries will fail unless paired with a
-   * trusted context — for proof verification on devnet, use
+   * proof verification. Requires explicit `addresses` in `options` —
+   * `devnetName` alone is not sufficient in non-trusted mode, since no
+   * masternode addresses can be discovered without a trusted context.
+   * Proof-bearing queries will fail; for proof verification on devnet, use
    * `EvoSDK.devnetTrusted` instead.
    */
   static devnet(devnetName: string, options: ConnectionOptions & { addresses?: string[] } = {}): EvoSDK {
--- a/packages/js-evo-sdk/src/sdk.ts
+++ b/packages/js-evo-sdk/src/sdk.ts
@@ -43,9 +43,12 @@ export interface EvoSDKOptions extends ConnectionOptions {
   // explicit `addresses` are mandatory and `devnetName` alone is not sufficient
   // — no masternode addresses can be discovered without a trusted context.
   devnetName?: string;
-  // Optional override for the trusted devnet quorum base URL. When omitted,
-  // the URL is derived as `https://quorums.<devnetName>.networks.dash.org`.
-  // Only consulted when trusted === true && network === 'devnet'.
+  // Optional override for the trusted-context quorum base URL. When omitted,
+  // the URL is the network's default (e.g.
+  // `https://quorums.<devnetName>.networks.dash.org` for devnet,
+  // `https://quorums.testnet.networks.dash.org` for testnet, etc.).
+  // Only consulted when trusted === true. Useful for pointing at a staging,
+  // self-hosted, or not-yet-deployed quorums endpoint.
   quorumUrl?: string;
 }
 
@@ -72,20 +75,16 @@ export class EvoSDK {
 
     if (network === 'devnet') {
       const hasAddresses = !!(addresses && addresses.length > 0);
-      if (!devnetName && !hasAddresses) {
-        throw new Error("EvoSDK: network 'devnet' requires either devnetName or explicit addresses");
-      }
-      if (trusted && !devnetName) {
-        throw new Error("EvoSDK: trusted devnet requires devnetName (used to derive the quorum URL)");
-      }
-      if (!trusted && !hasAddresses) {
+      if (trusted) {
+        if (!devnetName && !quorumUrl) {
+          throw new Error("EvoSDK: trusted devnet requires devnetName (to derive the quorum URL) or an explicit quorumUrl");
+        }
+      } else if (!hasAddresses) {
         throw new Error("EvoSDK: non-trusted devnet requires explicit addresses (no addresses can be discovered without a trusted context)");
       }
-      if (quorumUrl && !trusted) {
-        throw new Error("EvoSDK: quorumUrl is only meaningful when trusted === true");
-      }
-    } else if (quorumUrl) {
-      throw new Error("EvoSDK: quorumUrl is only valid when network === 'devnet'");
+    }
+    if (quorumUrl && !trusted) {
+      throw new Error("EvoSDK: quorumUrl is only meaningful when trusted === true");
     }
 
     this.options = { network, trusted, addresses, devnetName, quorumUrl, ...connection };
@@ -133,18 +132,25 @@ export class EvoSDK {
     let context: wasm.WasmTrustedContext | undefined;
     if (trusted) {
       if (network === 'mainnet') {
-        context = await wasm.WasmTrustedContext.prefetchMainnet();
-      } else if (network === 'testnet') {
-        context = await wasm.WasmTrustedContext.prefetchTestnet();
-      } else if (network === 'local') {
-        context = await wasm.WasmTrustedContext.prefetchLocal();
-      } else if (network === 'devnet') {
-        if (!devnetName) {
-          throw new Error("EvoSDK: trusted devnet requires devnetName");
-        }
         context = quorumUrl
-          ? await wasm.WasmTrustedContext.prefetchDevnetWithUrl(quorumUrl)
-          : await wasm.WasmTrustedContext.prefetchDevnet(devnetName);
+          ? await wasm.WasmTrustedContext.prefetchMainnetWithUrl(quorumUrl)
+          : await wasm.WasmTrustedContext.prefetchMainnet();
+      } else if (network === 'testnet') {
+        context = quorumUrl
+          ? await wasm.WasmTrustedContext.prefetchTestnetWithUrl(quorumUrl)
+          : await wasm.WasmTrustedContext.prefetchTestnet();
+      } else if (network === 'local') {
+        context = quorumUrl
+          ? await wasm.WasmTrustedContext.prefetchLocalWithUrl(quorumUrl)
+          : await wasm.WasmTrustedContext.prefetchLocal();
+      } else if (network === 'devnet') {
+        if (quorumUrl) {
+          context = await wasm.WasmTrustedContext.prefetchDevnetWithUrl(quorumUrl);
+        } else if (devnetName) {
+          context = await wasm.WasmTrustedContext.prefetchDevnet(devnetName);
+        } else {
+          throw new Error("EvoSDK: trusted devnet requires devnetName or quorumUrl");
+        }
       } else {
         throw new Error(`Unknown network: ${network}`);
       }
--- a/packages/rs-sdk-trusted-context-provider/src/provider.rs
+++ b/packages/rs-sdk-trusted-context-provider/src/provider.rs
@@ -86,44 +86,6 @@ struct MasternodeDiscoveryResponse {
 }
 
 impl TrustedHttpContextProvider {
-    /// Reject plaintext HTTP unless the host is a loopback address.
-    ///
-    /// The base URL is the SDK's root of trust for proof verification — quorum
-    /// public keys and discovered masternode addresses are fetched from it.
-    /// Allowing plaintext on a non-loopback link would let an on-path attacker
-    /// replace both, so HTTPS is required for any host outside `127.0.0.1` /
-    /// `localhost` / `[::1]`.
-    fn verify_secure_or_loopback(url: &str) -> Result<(), TrustedContextProviderError> {
-        let parsed_url = Url::parse(url).map_err(|e| {
-            TrustedContextProviderError::NetworkError(format!("Invalid URL: {}", e))
-        })?;
-        let scheme = parsed_url.scheme();
-        if scheme == "https" {
-            return Ok(());
-        }
-        if scheme != "http" {
-            return Err(TrustedContextProviderError::NetworkError(format!(
-                "Unsupported URL scheme '{}' (expected https, or http for loopback)",
-                scheme
-            )));
-        }
-        let host = parsed_url.host_str().ok_or_else(|| {
-            TrustedContextProviderError::NetworkError("URL has no host".to_string())
-        })?;
-        let host_lc = host.to_ascii_lowercase();
-        let is_loopback = host_lc == "localhost"
-            || host_lc == "127.0.0.1"
-            || host_lc == "[::1]"
-            || host_lc == "::1";
-        if !is_loopback {
-            return Err(TrustedContextProviderError::NetworkError(format!(
-                "Plaintext HTTP base URL is only permitted for loopback hosts; got '{}'",
-                host
-            )));
-        }
-        Ok(())
-    }
-
     /// Verify that a URL's domain resolves
     #[cfg(all(not(target_arch = "wasm32"), not(target_os = "ios")))]
     fn verify_domain_resolves(url: &str) -> Result<(), TrustedContextProviderError> {
@@ -177,12 +139,6 @@ impl TrustedHttpContextProvider {
         base_url: String,
         cache_size: NonZeroUsize,
     ) -> Result<Self, TrustedContextProviderError> {
-        // The base URL is the SDK's root of trust for proof verification
-        // (quorum public keys) and network connectivity (discovered masternode
-        // addresses). Require HTTPS unless the host is loopback — plaintext
-        // over a non-loopback link lets an on-path attacker replace both.
-        Self::verify_secure_or_loopback(&base_url)?;
-
         // Verify the domain resolves before proceeding (skip on WASM and iOS)
         #[cfg(all(not(target_arch = "wasm32"), not(target_os = "ios")))]
         Self::verify_domain_resolves(&base_url)?;
@@ -903,50 +859,6 @@ mod tests {
         }
     }
 
-    #[test]
-    fn test_new_with_url_rejects_plaintext_non_loopback() {
-        // Plaintext HTTP to a non-loopback host is rejected on every network.
-        let result = TrustedHttpContextProvider::new_with_url(
-            Network::Devnet,
-            "http://devnet.example.com".to_string(),
-            NonZeroUsize::new(10).unwrap(),
-        );
-        assert!(matches!(
-            result,
-            Err(TrustedContextProviderError::NetworkError(_))
-        ));
-    }
-
-    #[test]
-    fn test_new_with_url_accepts_plaintext_loopback() {
-        // Plaintext HTTP to loopback is acceptable (used by local dashmate).
-        for host in ["127.0.0.1", "localhost", "[::1]"] {
-            let result = TrustedHttpContextProvider::new_with_url(
-                Network::Regtest,
-                format!("http://{}:22444", host),
-                NonZeroUsize::new(10).unwrap(),
-            );
-            assert!(
-                result.is_ok(),
-                "expected http://{} loopback to be accepted, got {:?}",
-                host,
-                result.err()
-            );
-        }
-    }
-
-    #[test]
-    fn test_new_with_url_rejects_unknown_scheme() {
-        let result = TrustedHttpContextProvider::new_with_url(
-            Network::Devnet,
-            "ftp://example.com".to_string(),
-            NonZeroUsize::new(10).unwrap(),
-        );
-        assert!(matches!(
-            result,
-            Err(TrustedContextProviderError::NetworkError(_))
-        ));
-    }
 
     #[test]
     fn test_known_contracts() {
--- a/packages/rs-sdk-trusted-context-provider/src/provider.rs
+++ b/packages/rs-sdk-trusted-context-provider/src/provider.rs
@@ -86,6 +86,44 @@ struct MasternodeDiscoveryResponse {
 }
 
 impl TrustedHttpContextProvider {
+    /// Reject plaintext HTTP unless the host is a loopback address.
+    ///
+    /// The base URL is the SDK's root of trust for proof verification — quorum
+    /// public keys and discovered masternode addresses are fetched from it.
+    /// Allowing plaintext on a non-loopback link would let an on-path attacker
+    /// replace both, so HTTPS is required for any host outside `127.0.0.1` /
+    /// `localhost` / `[::1]`.
+    fn verify_secure_or_loopback(url: &str) -> Result<(), TrustedContextProviderError> {
+        let parsed_url = Url::parse(url).map_err(|e| {
+            TrustedContextProviderError::NetworkError(format!("Invalid URL: {}", e))
+        })?;
+        let scheme = parsed_url.scheme();
+        if scheme == "https" {
+            return Ok(());
+        }
+        if scheme != "http" {
+            return Err(TrustedContextProviderError::NetworkError(format!(
+                "Unsupported URL scheme '{}' (expected https, or http for loopback)",
+                scheme
+            )));
+        }
+        let host = parsed_url.host_str().ok_or_else(|| {
+            TrustedContextProviderError::NetworkError("URL has no host".to_string())
+        })?;
+        let host_lc = host.to_ascii_lowercase();
+        let is_loopback = host_lc == "localhost"
+            || host_lc == "127.0.0.1"
+            || host_lc == "[::1]"
+            || host_lc == "::1";
+        if !is_loopback {
+            return Err(TrustedContextProviderError::NetworkError(format!(
+                "Plaintext HTTP base URL is only permitted for loopback hosts; got '{}'",
+                host
+            )));
+        }
+        Ok(())
+    }
+
     /// Verify that a URL's domain resolves
     #[cfg(all(not(target_arch = "wasm32"), not(target_os = "ios")))]
     fn verify_domain_resolves(url: &str) -> Result<(), TrustedContextProviderError> {
@@ -139,6 +177,12 @@ impl TrustedHttpContextProvider {
         base_url: String,
         cache_size: NonZeroUsize,
     ) -> Result<Self, TrustedContextProviderError> {
+        // The base URL is the SDK's root of trust for proof verification
+        // (quorum public keys) and network connectivity (discovered masternode
+        // addresses). Require HTTPS unless the host is loopback — plaintext
+        // over a non-loopback link lets an on-path attacker replace both.
+        Self::verify_secure_or_loopback(&base_url)?;
+
         // Verify the domain resolves before proceeding (skip on WASM and iOS)
         #[cfg(all(not(target_arch = "wasm32"), not(target_os = "ios")))]
         Self::verify_domain_resolves(&base_url)?;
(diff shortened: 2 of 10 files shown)


packages/js-evo-sdk/src/sdk.ts
```diff
@@ -181,11 +219,42 @@ export class EvoSDK {
...
+  /**
+   * Create an EvoSDK instance configured for a devnet, without trusted-context
+   * proof verification. Requires either `devnetName` or explicit `addresses`
+   * in `options`. Proof-bearing queries will fail unless paired with a
+   * trusted context — for proof verification on devnet, use
+   * `EvoSDK.devnetTrusted` instead.
+   */
+  static devnet(devnetName: string, options: ConnectionOptions & { addresses?: string[] } = {}): EvoSDK {
+    return new EvoSDK({ network: 'devnet', devnetName, ...options });
+  }
```
thepastaclaw, line 244: **💬 Nitpick: JSDoc on `EvoSDK.devnet` misstates the precondition**

The doc says "Requires either `devnetName` or explicit `addresses` in `options`." But `devnetName` is the required first positional argument, not an `options` field, and the constructor at sdk.ts:71-84 enforces `!trusted && !hasAddresses` → throw. So the effective rule for this non-trusted factory is *always*: callers must pass `addresses` in `options`, regardless of `devnetName`. As written, a user calling `EvoSDK.devnet('paloma')` will hit the runtime throw covered by the constructor's `non-trusted devnet requires explicit addresses` check.

Reword to something like: "Requires explicit `addresses` in `options` — non-trusted devnet cannot discover addresses without a trusted context. For proof verification on devnet, use `EvoSDK.devnetTrusted` instead."

<sub>source: ['claude']</sub>

packages/wasm-sdk/src/context_provider.rs
```diff
@@ -166,6 +167,72 @@ impl WasmTrustedContext {
...
+            .await
+            .map_err(|e| WasmSdkError::generic(format!("Failed to prefetch quorums: {}", e)))?;
+
+        let discovered_addresses = Self::fetch_addresses_from(&inner).await?;
+
+        Ok(WasmTrustedContext {
+            inner,
+            discovered_addresses,
+        })
+    }
```
thepastaclaw, line 194: **💬 Nitpick: `prefetchDevnetWithUrl` takes `String` while sibling `prefetchLocalWithUrl` takes `&str`**

Pure binding-surface asymmetry: `prefetch_devnet_with_url(base_url: String)` at line 210 differs from `prefetch_local_with_url(base_url: &str)` at line 250. Both wasm-bindgen forms copy the JS string into Rust, so there's no correctness or performance difference; the inconsistency is a minor surprise for callers who treat these as a pair. Either pick `&str` everywhere or `String` everywhere.

thepastaclaw (changes requested):
## Code Review

Verified the PR against the worktree at 3abe91993d. Three confirmed in-scope issues: a blocking test/code drift in the WASM builder spec (the suite still asserts devnet must be rejected), a security/footgun in the new flat devnet URL shape (reserved names like 'mainnet' silently resolve to production quorum endpoints), and a defense-in-depth gap where the newly-exposed prefetchDevnetWithUrl entry point accepts plaintext HTTP for trust-root material. Two consistency/doc nits also kept. Several out-of-scope adjacent items are tracked separately.

<details>

```

In `packages/wasm-sdk/tests/unit/builder-with-addresses.spec.ts`:
- [BLOCKING] packages/wasm-sdk/tests/unit/builder-with-addresses.spec.ts:73-95: WASM unit spec still asserts `withAddresses(..., 'devnet')` is rejected after the implementation began accepting it
  `packages/wasm-sdk/src/sdk.rs:205-216` now maps `"devnet" => Network::Devnet` and the error string was updated to `"Expected: mainnet, testnet, devnet, or local"`. But `builder-with-addresses.spec.ts:73-83` still has `it('should reject devnet', ...)` asserting both that the call throws and that the message contains `'mainnet, testnet or local'`. The `'should reject invalid network name'` case at lines 85-95 also asserts the old error string and will fail against the new message. Once the WASM artifact is rebuilt for tests, this suite breaks and no longer describes the new API. Update the spec to assert a positive devnet case (build succeeds) and update the invalid-network assertion to the new error wording.

In `packages/rs-sdk-trusted-context-provider/src/lib.rs`:
- [SUGGESTION] packages/rs-sdk-trusted-context-provider/src/lib.rs:28-53: Devnet-name validation does not reject reserved names that now alias the mainnet/testnet quorum hostnames
  The PR flattens the devnet URL shape from `https://quorums.devnet.<name>.networks.dash.org` to `https://quorums.<name>.networks.dash.org`. The old shape was structurally namespaced; the new shape shares the namespace with mainnet/testnet/local. The validator at lines 30-46 only checks character class and edge hyphens, so `name = "mainnet" | "testnet" | "local" | "regtest"` all pass and produce the production-network quorum URL while the SDK is configured with `Network::Devnet`.

  Via the new `EvoSDK.devnetTrusted('mainnet')` / `WasmTrustedContext.prefetchDevnet('mainnet')` surface, a typo or attacker-influenced devnet name silently pulls real mainnet quorum public keys (and via `withTrustedContext`, real mainnet masternode addresses) into a session labeled `network: 'devnet'`. Proof verification then runs devnet payloads against mainnet quorum keys — a cross-network mismatch instead of an early, loud configuration error. The previous URL pattern made this collision impossible; the new pattern removes the structural guardrail and the validator is the only line of defense.

  Reject `mainnet | testnet | devnet | local | regtest` (case-insensitive) in `get_quorum_base_url`. The escape hatch (`new_with_url` / `prefetchDevnetWithUrl`) still lets callers point at any URL explicitly.

In `packages/rs-sdk-trusted-context-provider/src/provider.rs`:
- [SUGGESTION] packages/rs-sdk-trusted-context-provider/src/provider.rs:137-145: `new_with_url` (now reachable from JS via `prefetchDevnetWithUrl`) does not require HTTPS for trust-root fetches
  Before this PR, `new_with_url` was only reachable from native Rust callers and from `prefetch_local_with_url` (loopback). This PR introduces `WasmTrustedContext::prefetch_devnet_with_url` (context_provider.rs:209-234), exposing `new_with_url` to JS callers with an arbitrary `base_url` for `Network::Devnet`. The constructor does not validate the scheme.

  The documented use case is exactly the dangerous one: "Use this when the default `https://quorums.<devnet_name>.networks.dash.org` URL is not yet deployed." A developer pointing at an early-stage devnet sidecar over `http://` lets any on-path attacker tamper with `/quorums` (the SDK's root of trust for proof verification) and `/masternodes` (the addresses `withTrustedContext` injects into the builder). The UI then shows "proof verified" on attacker-controlled responses signed by attacker-controlled quorum keys.

  Reject non-HTTPS URLs for `Network::Devnet` (and other non-loopback networks) at construction. `prefetch_local_with_url`/`prefetch_local` are scoped to 127.0.0.1 and remain acceptable plaintext.
```
</details>

<details>
<summary>Out-of-scope follow-up suggestions (4)</summary>

These are valid observations, but they are outside this PR's scope and should be handled in separate issues or author/maintainer-requested PRs rather than blocking this review.

- **`WasmSdkBuilder.newDevnet()` doesn't enforce the 'must pair with addresses or trusted context' contract at build time** — The new `new_devnet()` constructs the inner `SdkBuilder` with `AddressList::default()` (empty). Callers are expected to add addresses or attach a trusted context; nothing fails fast if they don't. EvoSDK enforces this at the JS layer (the primary intended consumer), so direct WASM consumers are the only ones who could hit a confusing downstream connectivity failure. Pre-existing builder-pattern convention; not introduced by this PR but a new factory that inherits it.
  - Follow-up: Consider a follow-up to validate in `WasmSdkBuilder.build()` that a Devnet builder has either a trusted context or a non-empty address list.
- **Devnet name validation accepts Unicode alphanumerics** — `get_quorum_base_url` uses `c.is_alphanumeric()`, which is Unicode-aware (Cyrillic etc. pass). The resulting hostname would need IDN handling before `reqwest`. Pre-existing and unchanged by this PR; the URL flattening just expands the surface where this validation is the only check. Worth tightening to ASCII alphanumerics across the devnet-name validators.
  - Follow-up: Open a follow-up to restrict devnet-name validation to ASCII alphanumerics + hyphens across `get_quorum_base_url` and any JS-side mirrors.
- **`Network::Devnet` falls back to port 443 in `fetch_masternode_addresses`** — `provider.rs:309-313` selects the DAPI default port via `match self.network { Mainnet => 443, Testnet => 1443, _ => 443 }`. Devnet now flows through this path as a first-class network; entries without `platformHTTPPort` will fall back to 443, but devnets typically use 1443 like testnet. Pre-existing match arm not introduced by this PR; in practice the discovery payload usually carries `platformHTTPPort`.
  - Follow-up: Track a separate issue to revisit Devnet's default DAPI port fallback (1443 vs 443).
- **Boilerplate duplication across the five `prefetch_*` methods** — After this PR, `prefetch_mainnet`, `prefetch_testnet`, `prefetch_devnet`, `prefetch_devnet_with_url`, and `prefetch_local_with_url` are five copies of the same ~14-line body. The duplication is pre-existing; this PR follows the established pattern faithfully, but adding two more copies makes consolidation more worthwhile.
  - Follow-up: Open a separate cleanup PR to extract the shared post-construction body into a private helper.

</details>

packages/wasm-sdk/src/context_provider.rs
thepastaclaw, line 194: **💬 Nitpick: `prefetchDevnetWithUrl` takes `String` while sibling `prefetchLocalWithUrl` takes `&str`**

The wasm-bindgen surface is inconsistent: `prefetch_devnet_with_url(base_url: String)` vs `prefetch_local_with_url(base_url: &str)`. Both forms copy the JS string into Rust regardless, so this is not a correctness or memory issue — purely API-surface noise that future maintainers may misread as conveying ownership intent. Align the parameter shapes. Verified STILL VALID at head df019bb7.

```suggestion
    #[wasm_bindgen(js_name = "prefetchDevnetWithUrl")]
    pub async fn prefetch_devnet_with_url(
        base_url: &str,
    ) -> Result<WasmTrustedContext, WasmSdkError> {
        let inner = rs_sdk_trusted_context_provider::TrustedHttpContextProvider::new_with_url(
            dash_sdk::dpp::dashcore::Network::Devnet,
            base_url.to_string(),
            std::num::NonZeroUsize::new(100).unwrap(),
        )
        .map_err(|e| WasmSdkError::generic(format!("Failed to create context provider: {}", e)))?
        .with_refetch_if_not_found(false);

        let inner = Arc::new(inner);

        inner
            .update_quorum_caches()
            .await
            .map_err(|e| WasmSdkError::generic(format!("Failed to prefetch quorums: {}", e)))?;

        let discovered_addresses = Self::fetch_addresses_from(&inner).await?;

        Ok(WasmTrustedContext {
            inner,
            discovered_addresses,
        })
    }
```

thepastaclaw (commented):
## Code Review

Latest delta (`3abe9199..df019bb7`) fixes the stale WASM unit test and the `EvoSDK.devnet` JSDoc precondition. I found no new defects in that delta.

Prior finding reconciliation for `df019bb7`:
- **FIXED:** `builder-with-addresses.spec.ts` now accepts devnet and expects the updated invalid-network message.
- **FIXED:** `EvoSDK.devnet` JSDoc now says non-trusted devnet requires explicit `addresses`.
- **STILL VALID:** devnet names can still use reserved aliases like `mainnet`/`testnet`.
- **STILL VALID:** `prefetchDevnetWithUrl` / `new_with_url` still allows plaintext HTTP for devnet trust-root fetches.
- **STILL VALID:** `prefetchDevnetWithUrl` still takes `String` while sibling `prefetchLocalWithUrl` takes `&str`.

🟡 2 suggestion(s) | 💬 1 nitpick

_3 findings are carried forward from the prior review because they are still valid at this head._

<details>

```

In `packages/rs-sdk-trusted-context-provider/src/lib.rs`:
- [SUGGESTION] lines 28-53: Devnet-name validator accepts reserved names that alias mainnet/testnet/local quorum hostnames
  `get_quorum_base_url` formats devnet URLs as `https://quorums.<name>.networks.dash.org`, which shares the namespace used by mainnet/testnet/local. The validator only checks character set and hyphen placement, so `mainnet`, `testnet`, `local`, and `regtest` are accepted as devnet names and resolve to the production quorum endpoints while the configured `Network` enum is `Devnet`. Reject reserved names up front.

In `packages/rs-sdk-trusted-context-provider/src/provider.rs`:
- [SUGGESTION] lines 137-177: `new_with_url` reachable from JS via `prefetchDevnetWithUrl` does not require HTTPS for non-local trust-root fetches
  `WasmTrustedContext.prefetchDevnetWithUrl(base_url)` forwards a JS-supplied string into `TrustedHttpContextProvider::new_with_url` with `Network::Devnet`. The constructor performs DNS resolution only on native targets and never enforces the URL scheme, so callers can pass `http://...` and fetch quorum keys / masternode discovery data over plaintext HTTP. Require HTTPS for mainnet/testnet/devnet and preserve HTTP only for local/regtest loopback use cases.

In `packages/wasm-sdk/src/context_provider.rs`:
- [NITPICK] lines 209-234: `prefetchDevnetWithUrl` takes `String` while sibling `prefetchLocalWithUrl` takes `&str`
  The wasm-bindgen surface is inconsistent: `prefetch_devnet_with_url(base_url: String)` vs `prefetch_local_with_url(base_url: &str)`. Both forms copy the JS string into Rust, so this is API-surface noise rather than a correctness issue.
```
</details>

*Reviewed commit: `df019bb7`*

packages/wasm-sdk/src/context_provider.rs
```diff
@@ -183,11 +210,35 @@ impl WasmTrustedContext {
...
-        let inner = rs_sdk_trusted_context_provider::TrustedHttpContextProvider::new_with_url(
+        Self::prefetch_for(
             dash_sdk::dpp::dashcore::Network::Regtest,
-            base_url.to_string(),
-            std::num::NonZeroUsize::new(100).unwrap(),
+            None,
+            Some(base_url.to_string()),
         )
+        .await
+    }
```
thepastaclaw, line 219: **💬 Nitpick: `prefetch*WithUrl` signatures are inconsistent across the four URL-taking factories**

After the `prefetch_for` refactor there are four `prefetch*_with_url` async factories exposed across the wasm-bindgen boundary: `prefetch_mainnet_with_url(base_url: String)` (line 125), `prefetch_testnet_with_url(base_url: String)` (line 149), `prefetch_devnet_with_url(base_url: String)` (line 185), and `prefetch_local_with_url(base_url: &str)` (line 210). wasm-bindgen copies the JS string into Rust either way, so this has no correctness or memory impact, but the surface is noticeably asymmetric — three siblings use `String`, only `prefetch_local_with_url` still uses `&str` (which then internally calls `.to_string()` at line 216). Align all four on one shape (`String` is the simpler fix, matches the other three, and removes the internal `.to_string()`). Carried forward from prior finding `wasm-prefetch-devnet-string-asymmetry` — now inverted, with local being the odd one out.

thepastaclaw (commented):
## Code Review

Prior finding reconciliation for 8daa3d5a: carried forward STILL VALID prior findings: #2:trusted-context-custom-url-https, #4:wasm-prefetch-devnet-string-asymmetry; fixed prior findings: #1:trusted-context-devnet-reserved-name-alias-primary, #3:trusted-context-devnet-reserved-name-alias-duplicate. Prior reserved-name items #1 and #3 are duplicates and are consolidated into one public finding when still valid. Latest-delta findings are listed separately when present.

At head 8daa3d5a12, the prior reserved-devnet-name finding is FIXED (validator + tests at lib.rs:47-60 and provider.rs:836-859). Two prior findings carry forward: (1) `new_with_url` still accepts arbitrary URL schemes including plaintext `http://` for trusted contexts, and this PR widens the surface by adding `prefetchMainnetWithUrl`/`prefetchTestnetWithUrl` plus `quorumUrl` overrides in `EvoSDK`; (2) the `prefetch*WithUrl` parameter shapes are inconsistent (`String` for mainnet/testnet/devnet, `&str` for local). No new defects introduced by the latest delta; the `prefetch_for` refactor is clean.

<details>

```

In `packages/rs-sdk-trusted-context-provider/src/provider.rs`:
- [SUGGESTION] packages/rs-sdk-trusted-context-provider/src/provider.rs:137-177: Plaintext-scheme trust-root URLs are still accepted; this PR widens the surface to mainnet/testnet
  `TrustedHttpContextProvider::new_with_url` performs no URL-scheme check, and on wasm/iOS it skips the DNS pre-resolution as well (provider.rs:142-144). The HTTPS-required guard added in 586df73c33 was explicitly reverted in 8daa3d5a12 to preserve an escape hatch for early devnets lacking certs. That rationale is reasonable for `Network::Devnet`/`Network::Regtest`, but this same PR newly exposes the same plaintext sink for production networks: commit d2a7c1b810 added `WasmTrustedContext::prefetchMainnetWithUrl` / `prefetchTestnetWithUrl` (context_provider.rs:124-158), and `js-evo-sdk` now forwards `quorumUrl` for `network: 'mainnet'` / `'testnet'` (sdk.ts). The base URL is the SDK's root of trust for proof verification: an attacker who can MITM a plaintext fetch can substitute quorum public keys (forging proofs the SDK will accept) and replace the discovered masternode address list (funneling later DAPI traffic through attacker-controlled nodes), while the SDK still reports proofs as verified. For `Mainnet`/`Testnet` there is no `no-cert-yet` workflow that justifies plaintext. Recommend enforcing `https` when `network` is `Mainnet` or `Testnet`, keeping the deferred behavior for `Devnet`/`Regtest`. This is carried forward from prior finding `trusted-context-custom-url-https`; the devnet/regtest scope is intentionally deferred per the revert commit, but the new mainnet/testnet surface is in scope for this PR and not covered by the cited rationale.
```
</details>

<details>
<summary>Out-of-scope follow-up suggestions (3)</summary>

These are valid observations, but they are outside this PR's scope and should be handled in separate issues or author/maintainer-requested PRs rather than blocking this review.

- **Reserved-name validator can be sidestepped by non-ASCII alphanumeric look-alikes** — `get_quorum_base_url` at lib.rs:36 uses `char::is_alphanumeric()`, which accepts non-ASCII alphabetics; the reserved-name guard at lib.rs:52-60 lowercases via `to_ascii_lowercase` and compares against ASCII strings. A devnet name like `mаinnet` (Cyrillic `а`) passes the alphanumeric check, is unchanged by `to_ascii_lowercase`, and slips past the reserved-name match. The IDN-encoded hostname differs from the real `quorums.mainnet.networks.dash.org`, so this is not a direct alias bypass, but it is a UI homograph risk for tools displaying the chosen devnet name. The alphanumeric rule predates this PR, so out of scope; a follow-up tightening to ASCII alphanumeric + hyphen would close it cleanly.
  - Follow-up: Open a separate issue to tighten devnet-name validation to `is_ascii_alphanumeric()` + hyphen.
- **`EvoSDK.mainnetTrusted` / `testnetTrusted` JSDoc and option types don't surface the new `quorumUrl` override** — Commit d2a7c1b810 made `quorumUrl` meaningful on every trusted network, but only `EvoSDK.devnetTrusted` documents the option. `mainnetTrusted`/`testnetTrusted` accept only `ConnectionOptions` per their typings, so callers cannot discover the override from types alone. Documentation/typing gap only.
  - Follow-up: Add `quorumUrl?: string` to the option type of `mainnetTrusted`/`testnetTrusted` and document the staging-endpoint use case.
- **`WasmContext` silently returns `Ok(None)` for `get_token_configuration`/`get_data_contract` in non-trusted mode** — Pre-existing in `WasmContext` (context_provider.rs:43-68): the non-trusted path returns `Ok(None)` and only logs a tracing warning when a token configuration is requested. JS callers get a downstream proof-verification failure instead of a clear 'trusted context required' error. Unchanged by this PR.
  - Follow-up: Track separately to surface a structured error earlier when token/data-contract lookups are requested without a trusted context.

</details>

### Review details

<details>

```

In `packages/rs-sdk-trusted-context-provider/src/provider.rs`:
- [SUGGESTION] lines 137-177: Plaintext-scheme trust-root URLs are still accepted; this PR widens the surface to mainnet/testnet
  `TrustedHttpContextProvider::new_with_url` performs no URL-scheme check, and on wasm/iOS it skips the DNS pre-resolution as well (provider.rs:142-144). The HTTPS-required guard added in 586df73c33 was explicitly reverted in 8daa3d5a12 to preserve an escape hatch for early devnets lacking certs. That rationale is reasonable for `Network::Devnet`/`Network::Regtest`, but this same PR newly exposes the same plaintext sink for production networks: commit d2a7c1b810 added `WasmTrustedContext::prefetchMainnetWithUrl` / `prefetchTestnetWithUrl` (context_provider.rs:124-158), and `js-evo-sdk` now forwards `quorumUrl` for `network: 'mainnet'` / `'testnet'` (sdk.ts). The base URL is the SDK's root of trust for proof verification: an attacker who can MITM a plaintext fetch can substitute quorum public keys (forging proofs the SDK will accept) and replace the discovered masternode address list (funneling later DAPI traffic through attacker-controlled nodes), while the SDK still reports proofs as verified. For `Mainnet`/`Testnet` there is no `no-cert-yet` workflow that justifies plaintext. Recommend enforcing `https` when `network` is `Mainnet` or `Testnet`, keeping the deferred behavior for `Devnet`/`Regtest`. This is carried forward from prior finding `trusted-context-custom-url-https`; the devnet/regtest scope is intentionally deferred per the revert commit, but the new mainnet/testnet surface is in scope for this PR and not covered by the cited rationale.

In `packages/wasm-sdk/src/context_provider.rs`:
- [NITPICK] lines 124-219: `prefetch*WithUrl` signatures are inconsistent across the four URL-taking factories
  After the `prefetch_for` refactor there are four `prefetch*_with_url` async factories exposed across the wasm-bindgen boundary: `prefetch_mainnet_with_url(base_url: String)` (line 125), `prefetch_testnet_with_url(base_url: String)` (line 149), `prefetch_devnet_with_url(base_url: String)` (line 185), and `prefetch_local_with_url(base_url: &str)` (line 210). wasm-bindgen copies the JS string into Rust either way, so this has no correctness or memory impact, but the surface is noticeably asymmetric — three siblings use `String`, only `prefetch_local_with_url` still uses `&str` (which then internally calls `.to_string()` at line 216). Align all four on one shape (`String` is the simpler fix, matches the other three, and removes the internal `.to_string()`). Carried forward from prior finding `wasm-prefetch-devnet-string-asymmetry` — now inverted, with local being the odd one out.

```
</details>

*Reviewed commit: `8daa3d5a`*

packages/wasm-sdk/src/context_provider.rs
thepastaclaw, line 219: **💬 Nitpick: `prefetch*WithUrl` factories mix `String` and `&str` parameter shapes**

`prefetch_mainnet_with_url` (l.125), `prefetch_testnet_with_url` (l.149), and `prefetch_devnet_with_url` (l.185) take `base_url: String`; `prefetch_local_with_url` (l.210-211) takes `base_url: &str` and calls `.to_string()` internally. wasm-bindgen accepts both for JS-string-backed arguments, so the JS-visible signatures are identical — this is API-surface noise rather than a correctness issue. Since the PR is already touching this family of exports, converging on one shape (`String`, to match the other three) would keep the Rust wrapper layer consistent and make future maintenance uniform.

```suggestion
    #[wasm_bindgen(js_name = "prefetchLocalWithUrl")]
    pub async fn prefetch_local_with_url(
        base_url: String,
    ) -> Result<WasmTrustedContext, WasmSdkError> {
        Self::prefetch_for(
            dash_sdk::dpp::dashcore::Network::Regtest,
            None,
            Some(base_url),
        )
        .await
    }
```

thepastaclaw (commented):
## Code Review

Prior finding reconciliation for 1051e4e3: carried forward STILL VALID prior findings: #1:trusted-context-plaintext-url-surface-mainnet-testnet, #2:wasm-prefetch-with-url-signature-inconsistency; fixed prior findings: none. Latest-delta findings are listed separately when present.

Latest delta (1051e4e) is whitespace-only (cargo fmt) and introduces no new behavior. Both prior findings are STILL VALID at current head and carried forward: plaintext-scheme trust-root URLs are accepted by `new_with_url` for mainnet/testnet/devnet (the HTTPS guard was explicitly reverted in 8daa3d5a as an intentional escape hatch for early devnets), and the four `prefetch*WithUrl` wasm-bindgen factories use inconsistent `String` vs `&str` parameter shapes. No new defects in the latest delta.

<details>

```

In `packages/rs-sdk-trusted-context-provider/src/provider.rs`:
- [SUGGESTION] packages/rs-sdk-trusted-context-provider/src/provider.rs:137-177: Plaintext trust-root URLs accepted across mainnet/testnet/devnet via new URL-taking factories
  `TrustedHttpContextProvider::new_with_url` performs no URL-scheme validation, and on `wasm32`/`ios` builds it also skips `verify_domain_resolves` (provider.rs:142-144). The HTTPS-required guard added in 586df73c was deliberately reverted in 8daa3d5a to preserve an escape hatch for early devnets without deployed DNS. This PR widens the reach of that escape hatch by exposing URL-taking factories not just for devnet but for mainnet (`prefetchMainnetWithUrl`) and testnet (`prefetchTestnetWithUrl`) as well (context_provider.rs:124-194), and routes the JS-side `EvoSDKOptions.quorumUrl` through the same path. The trust-root response bootstraps quorum public keys used directly in proof verification, so a network-positioned attacker who can MITM a plaintext fetch can substitute attacker-controlled quorum keys and forge proofs that the SDK will accept. The browser/iOS branches are the most exposed because they don't even pre-resolve the host.

  Given the revert was intentional, this may be an accepted-risk choice — but at minimum the surface should be narrowed (require `https` for non-loopback hosts on mainnet/testnet, allow `http://127.0.0.1`/`localhost`/`::1` for the local sidecar use case) or the policy explicitly documented on `new_with_url` so future callers and JS integrators understand HTTPS is their responsibility. Both Claude and Codex independently flagged this and agree on the mitigation shape.
```
</details>

### Review details

<details>

```

In `packages/rs-sdk-trusted-context-provider/src/provider.rs`:
- [SUGGESTION] lines 137-177: Plaintext trust-root URLs accepted across mainnet/testnet/devnet via new URL-taking factories
  `TrustedHttpContextProvider::new_with_url` performs no URL-scheme validation, and on `wasm32`/`ios` builds it also skips `verify_domain_resolves` (provider.rs:142-144). The HTTPS-required guard added in 586df73c was deliberately reverted in 8daa3d5a to preserve an escape hatch for early devnets without deployed DNS. This PR widens the reach of that escape hatch by exposing URL-taking factories not just for devnet but for mainnet (`prefetchMainnetWithUrl`) and testnet (`prefetchTestnetWithUrl`) as well (context_provider.rs:124-194), and routes the JS-side `EvoSDKOptions.quorumUrl` through the same path. The trust-root response bootstraps quorum public keys used directly in proof verification, so a network-positioned attacker who can MITM a plaintext fetch can substitute attacker-controlled quorum keys and forge proofs that the SDK will accept. The browser/iOS branches are the most exposed because they don't even pre-resolve the host.

Given the revert was intentional, this may be an accepted-risk choice — but at minimum the surface should be narrowed (require `https` for non-loopback hosts on mainnet/testnet, allow `http://127.0.0.1`/`localhost`/`::1` for the local sidecar use case) or the policy explicitly documented on `new_with_url` so future callers and JS integrators understand HTTPS is their responsibility. Both Claude and Codex independently flagged this and agree on the mitigation shape.

In `packages/wasm-sdk/src/context_provider.rs`:
- [NITPICK] lines 124-219: `prefetch*WithUrl` factories mix `String` and `&str` parameter shapes
  `prefetch_mainnet_with_url` (l.125), `prefetch_testnet_with_url` (l.149), and `prefetch_devnet_with_url` (l.185) take `base_url: String`; `prefetch_local_with_url` (l.210-211) takes `base_url: &str` and calls `.to_string()` internally. wasm-bindgen accepts both for JS-string-backed arguments, so the JS-visible signatures are identical — this is API-surface noise rather than a correctness issue. Since the PR is already touching this family of exports, converging on one shape (`String`, to match the other three) would keep the Rust wrapper layer consistent and make future maintenance uniform.

```
</details>

*Reviewed commit: `1051e4e3`*

packages/js-evo-sdk/src/sdk.ts
```diff
@@ -56,8 +71,23 @@ export class EvoSDK {
...
+        if (!devnetName && !quorumUrl) {
+          throw new Error("EvoSDK: trusted devnet requires devnetName (to derive the quorum URL) or an explicit quorumUrl");
+        }
+      } else if (!hasAddresses) {
+        throw new Error("EvoSDK: non-trusted devnet requires explicit addresses (no addresses can be discovered without a trusted context)");
+      }
+    }
+    if (quorumUrl && !trusted) {
+      throw new Error("EvoSDK: quorumUrl is only meaningful when trusted === true");
+    }
```
thepastaclaw, line 92: **🟡 Suggestion: `devnetName` is silently accepted on non-devnet networks**

The constructor guards `quorumUrl` against `trusted === false` but does not guard `devnetName` against non-devnet networks. `new EvoSDK({ network: 'testnet', devnetName: 'paloma' })` is accepted and the field is stored on `this.options`, but `connect()` only consults `devnetName` when `network === 'devnet'`. The misconfig is benign at runtime but masks typos like `network: 'testent'`. Add a symmetric check — `if (devnetName && network !== 'devnet') throw …` — to match the validation surface already established for `quorumUrl`.

<sub>source: ['claude']</sub>

packages/rs-sdk-trusted-context-provider/src/provider.rs
```diff
@@ -139,6 +139,22 @@ impl TrustedHttpContextProvider {
...
+        // plaintext escape hatch (early-stage devnets without a cert yet,
+        // local sidecars on loopback).
+        if matches!(network, Network::Mainnet | Network::Testnet)
+            && !base_url.starts_with("https://")
+        {
+            return Err(TrustedContextProviderError::NetworkError(format!(
+                "Custom quorum URL for {:?} must use https://; got '{}'",
+                network, base_url
+            )));
+        }
```
thepastaclaw: **💬 Nitpick: HTTPS gate uses case-sensitive prefix match**

`base_url.starts_with("https://")` is byte-prefix and case-sensitive, so RFC-3986-valid inputs like `"HTTPS://..."` are rejected with a misleading error, and a literal `"https://"` with an empty authority slips this check and only fails downstream. The failure mode is fail-closed (over-rejection, never under-acceptance) so it is not a security defect, but parsing with `url::Url::parse(...).scheme()` and comparing case-insensitively would make the rule schema-driven and let a future test assert against the parsed scheme rather than a substring.

packages/rs-sdk-trusted-context-provider/src/provider.rs
```diff
@@ -833,6 +849,50 @@ mod tests {
...
+                "http://example.com".to_string(),
+                NonZeroUsize::new(10).unwrap(),
+            );
+            assert!(
+                matches!(result, Err(TrustedContextProviderError::NetworkError(_))),
+                "expected http:// to be rejected for {:?}",
+                network
+            );
+        }
+    }
```
thepastaclaw, line 936: **💬 Nitpick: HTTPS-gate test only covers the negative path**

`test_new_with_url_rejects_plaintext_for_production_networks` pins the rejection case but does not assert the (intentionally permitted) complements: that `http://...` for `Network::Devnet`/`Network::Regtest` and `https://...` for mainnet/testnet are not rejected by this new gate. A future refactor that broadened the gate to all networks would pass the existing test silently. Because `verify_domain_resolves` is cfg-gated to non-wasm/non-iOS, the positive assertions need to either target a wasm/iOS cfg path or assert that any returned error is not the new `NetworkError("... must use https://...")` variant.

<sub>source: ['claude']</sub>

packages/wasm-sdk/src/context_provider.rs
thepastaclaw, line 219: **💬 Nitpick: `prefetch*WithUrl` exports use inconsistent Rust string ownership**

Three of the four URL-taking wasm-bindgen factories take `base_url: String` (`prefetchMainnetWithUrl`, `prefetchTestnetWithUrl`, `prefetchDevnetWithUrl`) while `prefetchLocalWithUrl` takes `&str` and immediately calls `.to_string()`. wasm-bindgen marshals both shapes identically from JS, so this is not a correctness or memory-safety issue across the boundary, but it is needless inconsistency on the Rust side and makes the four sibling factories harder to maintain uniformly. Normalize to `String` to match the other three.

thepastaclaw (commented):
## Code Review

Cumulative incremental review of PR #3748 at head 3752ed72. Prior findings status: (1) stale `prefetchDevnet` doc comment — FIXED; (2) silently-discarded `quorumUrl` on non-trusted devnet — FIXED; (3) mainnet/testnet plaintext quorum URLs — FIXED in this delta via the new HTTPS guard in `TrustedHttpContextProvider::new_with_url` plus a regression test; (4) devnet plaintext escape hatch — INTENTIONALLY DEFERRED per the in-source comment justifying early-stage devnets and loopback sidecars (recorded out-of-scope, not blocking). New latest-delta findings: none blocking. Remaining items are low-confidence nits and one symmetry suggestion in `js-evo-sdk`.

🟡 1 suggestion(s) | 💬 3 nitpick(s)

<details>

```

In `packages/js-evo-sdk/src/sdk.ts`:
- [SUGGESTION] packages/js-evo-sdk/src/sdk.ts:76-88: `devnetName` is silently accepted on non-devnet networks
  The constructor guards `quorumUrl` against `trusted === false` but does not guard `devnetName` against non-devnet networks. `new EvoSDK({ network: 'testnet', devnetName: 'paloma' })` is accepted and the field is stored on `this.options`, but `connect()` only consults `devnetName` when `network === 'devnet'`. The misconfig is benign at runtime but masks typos like `network: 'testent'`. Add a symmetric check — `if (devnetName && network !== 'devnet') throw …` — to match the validation surface already established for `quorumUrl`.
```
</details>

thepastaclaw (commented):
## Code Review

Cumulative incremental review of PR #3748 at head 7c362c4070. The latest delta (3752ed72..7c362c4070) is a 9-line constructor guard in js-evo-sdk that rejects `devnetName` on non-devnet networks; it is well-scoped, has a regression test, and introduces no new defects. Prior-finding reconciliation: FIXED — `devnetName` typo guard (verified at packages/js-evo-sdk/src/sdk.ts:85-89 with test at tests/unit/sdk.spec.ts:293-296); `quorumUrl` discarded on non-trusted (verified at sdk.ts:90-92); plaintext mainnet/testnet trust roots (verified at provider.rs:149-156); stale prefetchDevnet rustdoc; reserved devnet names. STILL VALID (carried forward) — case-sensitive HTTPS prefix match; HTTPS-gate test covers only the negative path; prefetch*WithUrl String vs &str inconsistency; book and js-evo-sdk README still document only three networks and omit the new devnet factories this PR exposes. INTENTIONALLY DEFERRED — devnet/regtest plaintext trust-root escape hatch (documented design, scope-gated to non-production networks). DROPPED — codex's `newDevnet()` empty-AddressList blocking claim: the raw wasm `WasmSdkBuilder::new_devnet()` rustdoc at packages/wasm-sdk/src/sdk.rs:250-255 documents the required pairing, and the JS-facing EvoSDK constructor (sdk.ts:76-89) enforces it for the first-class entry points this PR adds, so the documented contract is not violated through this PR's stated surface. No new findings in the latest delta.

_Note: Inline posting failed (command failed (1): python3 scripts/review_poster.py dashpay/platform 3748 7c362c4070ae89d62ec65ec98c31a125938c1adb
STDOUT:

STDERR:
Traceback (most recent call last):
  File "/Users/claw/.openclaw/workspace/scripts/review_poster.py", line 221, in <module>
    result = post_review(repo, pr_number, h), so I posted the same verified findings as a top-level review body._

*Reviewed commit: 7c362c40*

🟡 2 suggestion(s) | 💬 3 nitpick(s)

### Verified findings

<details>
<summary><b>suggestion:</b> [carried-forward] README documents `network` as only 'testnet' | 'mainnet' | 'local' and omits the new devnet factories</summary>

`packages/js-evo-sdk/README.md` (line 48)

This PR adds first-class devnet support via `EvoSDK.devnet(name, opts)` and `EvoSDK.devnetTrusted(name, opts)` (packages/js-evo-sdk/src/sdk.ts:242, 259) and a corresponding `network: 'devnet'` branch in the constructor (sdk.ts:76-89). The npm-facing README still types `network` as `'testnet' | 'mainnet' | 'local'` and the preset-factories sentence at line 55 lists only the six pre-existing factories, so consumers reading the package's top-level usage guide are told `network: 'devnet'` is invalid and will not discover the new factories. Since this is the public-facing surface this PR adds, the README should reflect it.

</details>
<details>
<summary><b>suggestion:</b> [carried-forward] Book page still claims 'three built-in networks' and does not document the new devnet factories</summary>

`book/src/evo-sdk/networks-and-environments.md` (line 3)

The Evo SDK book page asserts "three built-in network configurations" and the table only enumerates Testnet/Mainnet/Local. This PR introduces `EvoSDK.devnet()` and `EvoSDK.devnetTrusted()` as first-class entry points, plus the `prefetchDevnet` / `prefetchDevnetWithUrl` WASM factories and the `devnetName` / `quorumUrl` config options. The user-facing docs for the very feature this PR adds remain blank, which materially undermines discoverability of the new API.

</details>
<details>
<summary><b>nitpick:</b> [carried-forward] `prefetchLocalWithUrl` takes `&str` while the three sibling factories take `String`</summary>

`packages/wasm-sdk/src/context_provider.rs` (line 209)

Verified at head 7c362c4070: `prefetch_mainnet_with_url` (line 125), `prefetch_testnet_with_url` (line 149), and `prefetch_devnet_with_url` (line 185) all take `base_url: String`, but `prefetch_local_with_url` at line 210 takes `&str` and immediately calls `.to_string()` at line 216. wasm-bindgen marshals both shapes identically from JS, so this is not a correctness or memory-safety defect at the FFI boundary — purely a Rust-side inconsistency across four sibling factories. Normalizing to `String` removes the redundant allocation hop and makes the four factories trivially uniform.

</details>
<details>
<summary><b>nitpick:</b> [carried-forward] HTTPS gate uses byte-prefix, case-sensitive match</summary>

`packages/rs-sdk-trusted-context-provider/src/provider.rs` (line 149)

Verified at head 7c362c4070: the guard is `base_url.starts_with("https://")`. RFC-3986 schemes are case-insensitive, so `"HTTPS://..."` is rejected with a misleading 'must use https://' error, and a literal `"https://"` with an empty authority slips this check and fails downstream with a less actionable error. Fail-closed (over-rejection, never under-acceptance), so not a security defect. Parsing with `url::Url::parse(...)` and comparing `.scheme().eq_ignore_ascii_case("https")` would make the gate schema-driven.

</details>
<details>
<summary><b>nitpick:</b> [carried-forward] HTTPS-gate test pins only the rejection path</summary>

`packages/rs-sdk-trusted-context-provider/src/provider.rs` (line 852)

`test_new_with_url_rejects_plaintext_for_production_networks` covers the negative arm of the new guard but does not assert the (intentionally permitted) complements: `http://...` for `Network::Devnet` / `Network::Regtest`, and `https://...` for mainnet/testnet, must not be rejected by the new gate. A future refactor that broadened the gate to all networks would pass the existing test silently. Because `verify_domain_resolves` is cfg-gated to non-wasm/non-iOS, the positive assertions need to either target a wasm/iOS cfg path or assert that any returned error is not the new `NetworkError("... must use https://...")` variant.

</details>

<details>

```

- [SUGGESTION] In `packages/js-evo-sdk/README.md`:48-55: [carried-forward] README documents `network` as only 'testnet' | 'mainnet' | 'local' and omits the new devnet factories
  This PR adds first-class devnet support via `EvoSDK.devnet(name, opts)` and `EvoSDK.devnetTrusted(name, opts)` (packages/js-evo-sdk/src/sdk.ts:242, 259) and a corresponding `network: 'devnet'` branch in the constructor (sdk.ts:76-89). The npm-facing README still types `network` as `'testnet' | 'mainnet' | 'local'` and the preset-factories sentence at line 55 lists only the six pre-existing factories, so consumers reading the package's top-level usage guide are told `network: 'devnet'` is invalid and will not discover the new factories. Since this is the public-facing surface this PR adds, the README should reflect it.
- [SUGGESTION] In `book/src/evo-sdk/networks-and-environments.md`:3-12: [carried-forward] Book page still claims 'three built-in networks' and does not document the new devnet factories
  The Evo SDK book page asserts "three built-in network configurations" and the table only enumerates Testnet/Mainnet/Local. This PR introduces `EvoSDK.devnet()` and `EvoSDK.devnetTrusted()` as first-class entry points, plus the `prefetchDevnet` / `prefetchDevnetWithUrl` WASM factories and the `devnetName` / `quorumUrl` config options. The user-facing docs for the very feature this PR adds remain blank, which materially undermines discoverability of the new API.
- [NITPICK] In `packages/wasm-sdk/src/context_provider.rs`:209-219: [carried-forward] `prefetchLocalWithUrl` takes `&str` while the three sibling factories take `String`
  Verified at head 7c362c4070: `prefetch_mainnet_with_url` (line 125), `prefetch_testnet_with_url` (line 149), and `prefetch_devnet_with_url` (line 185) all take `base_url: String`, but `prefetch_local_with_url` at line 210 takes `&str` and immediately calls `.to_string()` at line 216. wasm-bindgen marshals both shapes identically from JS, so this is not a correctness or memory-safety defect at the FFI boundary — purely a Rust-side inconsistency across four sibling factories. Normalizing to `String` removes the redundant allocation hop and makes the four factories trivially uniform.
- [NITPICK] In `packages/rs-sdk-trusted-context-provider/src/provider.rs`:149-156: [carried-forward] HTTPS gate uses byte-prefix, case-sensitive match
  Verified at head 7c362c4070: the guard is `base_url.starts_with("https://")`. RFC-3986 schemes are case-insensitive, so `"HTTPS://..."` is rejected with a misleading 'must use https://' error, and a literal `"https://"` with an empty authority slips this check and fails downstream with a less actionable error. Fail-closed (over-rejection, never under-acceptance), so not a security defect. Parsing with `url::Url::parse(...)` and comparing `.scheme().eq_ignore_ascii_case("https")` would make the gate schema-driven.
- [NITPICK] In `packages/rs-sdk-trusted-context-provider/src/provider.rs`:852-868: [carried-forward] HTTPS-gate test pins only the rejection path
  `test_new_with_url_rejects_plaintext_for_production_networks` covers the negative arm of the new guard but does not assert the (intentionally permitted) complements: `http://...` for `Network::Devnet` / `Network::Regtest`, and `https://...` for mainnet/testnet, must not be rejected by the new gate. A future refactor that broadened the gate to all networks would pass the existing test silently. Because `verify_domain_resolves` is cfg-gated to non-wasm/non-iOS, the positive assertions need to either target a wasm/iOS cfg path or assert that any returned error is not the new `NetworkError("... must use https://...")` variant.
```
</details>

thepastaclaw (commented):
## Code Review

Cumulative incremental review of PR #3748 at head dd8cad309001c84cc6c37e7a9932a852e9a1535e. Latest delta (7c362c40..dd8cad30) is a focused hardening pass on the trusted-context provider: the HTTPS gate now uses `Url::parse` + `eq_ignore_ascii_case("https")`, three positive-path tests were added, and `prefetch_local_with_url` was normalized to `base_url: String` to match its three sibling factories. No new defects in the latest delta. Prior-finding reconciliation: FIXED — case-sensitive HTTPS prefix check; HTTPS-gate test pinning only the negative path; `prefetchLocalWithUrl` &str-vs-String inconsistency; `devnetName` silently accepted on non-devnet; `quorumUrl` discarded when non-trusted; reserved devnet-name aliasing; stale wasm doc/error strings. STILL VALID (carried forward) — js-evo-sdk README still types `network` as 'testnet'|'mainnet'|'local' and omits the new devnet factories; book Networks page still claims 'three built-in networks'. INTENTIONALLY DEFERRED — devnet/regtest plaintext quorum URL escape hatch (prior reviewer accepted tradeoff; out-of-scope follow-up). Codex flagged two items as 'blocking' that are actually pre-existing documented behavior (trusted-context replacing builder addresses; discovered-DAPI hosts rewritten to https://); both are out of this PR's scope and not new defects. Net: only two carried-forward documentation suggestions remain.

🟡 2 suggestion(s)

<details>

```

In `packages/js-evo-sdk/README.md`:
- [SUGGESTION] packages/js-evo-sdk/README.md:48-55: [carried-forward] README omits the new devnet factories and types `network` without `'devnet'`
  Verified STILL VALID at head dd8cad3090. This PR adds first-class devnet support — `EvoSDK.devnet()`, `EvoSDK.devnetTrusted()`, and a `network: 'devnet'` constructor branch with `devnetName` / `quorumUrl` options — but the npm-facing README at L48 still types `network` as `'testnet' | 'mainnet' | 'local'`, and the preset-factories paragraph at L55 lists only the six pre-existing factories. Consumers reading the package's top-level usage guide will not discover the new API surface this PR introduces.

In `book/src/evo-sdk/networks-and-environments.md`:
- [SUGGESTION] book/src/evo-sdk/networks-and-environments.md:3-12: [carried-forward] Book page still claims 'three built-in networks' and does not document devnet factories
  Verified STILL VALID at head dd8cad3090. L3-4 still asserts "three built-in network configurations" and the table at L8-12 only enumerates Testnet/Mainnet/Local. This PR promotes devnet to a first-class entry point (`EvoSDK.devnetTrusted`, `prefetchDevnet`/`prefetchDevnetWithUrl`, `devnetName`/`quorumUrl` config options). The user-facing book page for the very feature this PR adds remains blank.
```
</details>

thepastaclaw (commented):
## Code Review

Both mandatory prior documentation findings are **fixed** at `c32f2466`: `js-evo-sdk/README.md` now documents the `devnet` network value, `devnetName`/`quorumUrl`/`addresses`, and the `EvoSDK.devnet` / `EvoSDK.devnetTrusted` factories; `book/src/evo-sdk/networks-and-environments.md` now says “four built-in network configurations”, adds Devnet, and includes a dedicated Devnets section.

Two non-blocking issues remain: one suggestion around `addresses` being silently overwritten in trusted devnet mode, and one README wording nit.

<details>

```

In `packages/js-evo-sdk/src/sdk.ts`:
- [SUGGESTION] packages/js-evo-sdk/src/sdk.ts:165-182: Trusted devnet drops caller-supplied `addresses` — docs/validation should match the actual override semantics
  In `connect()`, the builder is first configured from explicit `addresses` via `WasmSdkBuilder.withAddresses(addresses, network)`, then `builder.withTrustedContext(context)` is applied whenever `trusted: true`. The Rust side (`packages/wasm-sdk/src/sdk.rs:296-313`) unconditionally replaces the address list with `context.discovered_addresses()` whenever the trusted context exposes any addresses — so the caller's explicit addresses are silently discarded before `build()`.
  
  The new devnet API makes this user-visible: `EvoSDK.devnetTrusted(name, { quorumUrl, addresses })` reads naturally as 'trust this quorum, but pin DAPI to these specific endpoints', and the README's `addresses` row says it 'replaces built-in defaults'. In practice it can only override for non-trusted mode, or for trusted setups where the context happens to discover no addresses.
  
  Either (a) document explicitly that `addresses` is ignored under `trusted: true` when the context provides any discovered addresses (and ideally warn/throw at construction when both are supplied), or (b) thread explicit `addresses` through `withTrustedContext` so they take precedence over discovered ones. Pre-existing for trusted mainnet/testnet, but this PR is what exposes it as a public option on the devnet factory.

In `packages/js-evo-sdk/README.md`:
- [NITPICK] packages/js-evo-sdk/README.md:51-51: README row for `devnetName` is internally inconsistent
  Two small inaccuracies vs. the constructor in `packages/js-evo-sdk/src/sdk.ts`:
  
  1. 'Required when `network: "devnet"` and `trusted: true`' — the constructor (sdk.ts L78-81) accepts trusted devnet with `devnetName` *or* `quorumUrl`; either is sufficient. The very next table row documents `quorumUrl` as an override, and tests assert trusted devnet works with only `quorumUrl`.
  2. 'ignored otherwise' — for non-devnet networks the constructor (sdk.ts L85-89) throws `EvoSDK: devnetName is only valid when network === 'devnet'`, not silently ignored. The trailing clause 'only valid when `network === "devnet"`' already captures this; 'ignored' contradicts it.

  Suggested wording:
  ```suggestion
  | `devnetName` | `string` | — | Short name of the devnet (e.g. `'paloma'`). Used to derive the trusted-context quorum URL; required for `trusted: true` devnet unless `quorumUrl` is supplied. Only valid when `network === 'devnet'` (rejected on other networks). |
  ```

```
</details>

*Reviewed commit: `c32f2466`*

QuantumExplorer (approved):

PR #3755, opened 2026-05-27
feat(swift-sdk,rs-sdk-ffi): wire devnet SDK config + auto-discover masternodes

shumkov:

Until now, getting the SwiftExampleApp talking to a devnet required manually entering three pieces of information in the Options screen: SPV peers, DAPI URL, and the quorum-list URL. The DAPI URL was load-bearing — without it the SDK had no addresses to talk to — but it was also redundant with the quorum-list service, which already exposes every active masternode's address. As nodes churn the manually-entered DAPI URL would silently go stale and proof verification would start failing with \"no available addresses to use\". The Swift side had no plumbing to feed the quorum URL into `TrustedHttpContextProvider`, and the cached `PlatformWalletManager` continued referencing the previous Sdk clone after a network / endpoint switch.

## What was done?

- `DashSDKConfig` (rs-sdk-ffi) gains a `quorum_url` field plumbed into `TrustedHttpContextProvider::new_with_url` from `dash_sdk_create_trusted`.
- `SDK.swift` adds `discoverActiveMasternodes(quorumBase:)` — fetches `{base}/masternodes` and returns `[(spvPeer, dapiUrl)]`. Devnet `SDK.init` now always auto-discovers fresh on every build, so node churn self-heals.
- `OptionsView` devnet UX shrinks from three inputs to one — Quorum URL only.
- `CoreContentView.spvPeerOverride` devnet branch derives SPV peers from the same `/masternodes` response (verbatim `ip:CoreP2PPort` per node — paloma reports `:20001`).
- `WalletManagerStore.activate` rebuilds the cached manager when the SDK handle changes, so endpoint overrides take effect on the next activation.
- `SwiftExampleApp/Info.plist` is now hand-managed (peer to the .xcodeproj so `PBXFileSystemSynchronizedRootGroup` doesn't auto-bundle it), gains an ATS exception so the HTTP `/masternodes` fetch isn't blocked. Test-only banner in the plist makes the dev intent explicit.
- `SDKLogger.log` mirrors to `NSLog` in addition to `Swift.print` so `simctl spawn booted log stream` captures it without an attached debugger.

## How Has This Been Tested?

- `cargo check -p rs-sdk-ffi` — clean.
- `xcodebuild -scheme SwiftExampleApp -destination 'platform=iOS Simulator,name=iPhone 17'` — builds successfully (only a benign `__eh_frame` linker warning).
- Manual end-to-end on paloma devnet from the parent branch (`test/sheilded_test_data`): launching the app cold, switching to devnet, and entering only the quorum URL produces a successful SPV sync and ShieldedNetworkSummary lookup with no further configuration.

## Breaking Changes

`DashSDKConfig` adds a new field. C/Swift callers that construct the struct literal must initialize `quorum_url` (use `ptr::null()` / `nil` when not configuring a quorum URL).

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Quorum URL configuration support for SDK initialization
  * Public helper to discover active masternodes from a quorum server
  * Devnet-specific UI to enter quorum server URL in the example app

* **Improvements**
  * Trusted setup now honors provided quorum URL and clarifies DAPI handling for devnet/regtest
  * Logger now mirrors messages to system logs for improved visibility
  * Wallet manager detects and refreshes when SDK configuration changes

* **Tests**
  * Test helpers updated to exercise the quorum URL path

[Review Change Stack](https://app.coderabbit.ai/change-stack/dashpay/platform/pull/3755?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/rs-sdk-ffi/src/sdk.rs
+++ b/packages/rs-sdk-ffi/src/sdk.rs
@@ -334,10 +334,47 @@ pub unsafe extern "C" fn dash_sdk_create_trusted(config: *const DashSDKConfig) -
         "dash_sdk_create_trusted: creating trusted context provider"
     );
 
-    // Create trusted context provider
-    // For regtest, use the quorum sidecar at localhost:22444 (dashmate Docker default)
-    let is_local = matches!(network, Network::Regtest);
-    let trusted_provider = if is_local {
+    // Create trusted context provider. Resolution order for the quorum
+    // lookup base URL:
+    //   1. Caller-provided `config.quorum_url` (highest priority — required
+    //      for devnet, also usable for non-default mainnet/testnet shards).
+    //   2. Regtest fallback to the local quorum sidecar `127.0.0.1:22444`
+    //      (the dashmate Docker default).
+    //   3. Network-derived default (mainnet/testnet only).
+    let explicit_quorum_url: Option<String> = if config.quorum_url.is_null() {
+        None
+    } else {
+        match unsafe { CStr::from_ptr(config.quorum_url) }.to_str() {
+            Ok(s) if !s.is_empty() => Some(s.to_string()),
+            Ok(_) => None,
+            Err(e) => {
+                return DashSDKResult::error(DashSDKError::new(
+                    DashSDKErrorCode::InvalidParameter,
+                    format!("Invalid quorum URL string: {}", e),
+                ))
+            }
+        }
+    };
+    let trusted_provider = if let Some(quorum_url) = explicit_quorum_url {
+        info!(
+            quorum_url = %quorum_url,
+            "dash_sdk_create_trusted: using caller-provided quorum URL"
+        );
+        match rs_sdk_trusted_context_provider::TrustedHttpContextProvider::new_with_url(
+            network,
+            quorum_url,
+            std::num::NonZeroUsize::new(100).unwrap(),
+        ) {
+            Ok(provider) => Arc::new(provider),
+            Err(e) => {
+                error!(error = %e, "dash_sdk_create_trusted: failed to create context provider from override URL");
+                return DashSDKResult::error(DashSDKError::new(
+                    DashSDKErrorCode::InternalError,
+                    format!("Failed to create context provider: {}", e),
+                ));
+            }
+        }
+    } else if matches!(network, Network::Regtest) {
         info!("dash_sdk_create_trusted: using local quorum sidecar for regtest");
         match rs_sdk_trusted_context_provider::TrustedHttpContextProvider::new_with_url(
             network,
@@ -376,10 +413,11 @@ pub unsafe extern "C" fn dash_sdk_create_trusted(config: *const DashSDKConfig) -
         }
     };
 
-    // Parse DAPI addresses - for trusted setup, we always need real addresses
+    // Parse DAPI addresses - for trusted setup, we always need real addresses.
+    // Devnet/regtest have no built-in defaults; callers must supply
+    // `dapi_addresses` (and typically `quorum_url`) for those networks.
     let builder = if config.dapi_addresses.is_null() {
         info!("dash_sdk_create_trusted: no DAPI addresses provided, using defaults for network");
-        // Use default addresses for the network
         match network {
             Network::Testnet => SdkBuilder::new_testnet(),
             Network::Mainnet => SdkBuilder::new_mainnet(),
@@ -600,6 +638,7 @@ pub unsafe extern "C" fn dash_sdk_create_with_callbacks(
             skip_asset_lock_proof_verification: config_ref.skip_asset_lock_proof_verification,
             request_retry_count: config_ref.request_retry_count,
             request_timeout_ms: config_ref.request_timeout_ms,
+            quorum_url: config_ref.quorum_url,
             platform_version: config_ref.platform_version,
         },
         context_provider: context_provider_handle,
--- a/packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
@@ -172,22 +172,22 @@ public final class SDK: @unchecked Sendable {
       let address: String          // "ip:CoreP2PPort"
       let status: String
       // Optional to match the Rust trusted-context provider, which
-      // tolerates entries missing `platform_http_port` and substitutes
+      // tolerates entries missing `platformHTTPPort` and substitutes
       // a per-network default. Requiring this would make a single
       // misbehaving JSON entry fail the whole decode (Decodable is
       // all-or-nothing per object), nuking devnet auto-discovery.
+      //
+      // Note JSON wire keys are camelCase (`platformHTTPPort`,
+      // `versionCheck`) — Rust renames its snake_case fields with
+      // `#[serde(rename = ...)]` to produce that on the wire. Swift's
+      // default `Decodable` synthesis matches property name → JSON
+      // key literally, so no `CodingKeys` is needed here as long as
+      // these property names match the wire keys verbatim.
       let platformHTTPPort: UInt16?
-      // Same `version_check` field the Rust provider filters on.
+      // Same `versionCheck` field the Rust provider filters on.
       // Optional because older quorum-list-server builds may omit it;
       // callers below treat missing as "not success" (i.e. excluded).
       let versionCheck: String?
-
-      enum CodingKeys: String, CodingKey {
-        case address
-        case status
-        case platformHTTPPort
-        case versionCheck = "version_check"
-      }
     }
 
     guard
--- a/packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
@@ -89,6 +89,114 @@ public final class SDK: @unchecked Sendable {
     return "http://127.0.0.1:2443"
   }
 
+  /// Optional caller-provided base URL for the trusted-context-provider's
+  /// quorum lookups. Read from UserDefaults key `platformQuorumURL`.
+  /// Required to connect to devnets (no built-in default exists on the
+  /// Rust side); also usable to override mainnet/testnet for staging
+  /// shards. Returns nil when unset/empty.
+  private static var platformQuorumURL: String? {
+    guard
+      let value = UserDefaults.standard.string(forKey: "platformQuorumURL"),
+      !value.isEmpty
+    else { return nil }
+    return value
+  }
+
+  /// Synchronously fetch `{quorumBase}/masternodes` and return the
+  /// raw `data` array. Both the DAPI list and the SPV peer list are
+  /// derived from this — DAPI takes `<ip>:<platformHTTPPort>`, SPV
+  /// takes the verbatim `address` field (`<ip>:<CoreP2PPort>`).
+  ///
+  /// Returns nil on any failure (timeout, JSON shape mismatch, etc.).
+  /// Filters to `status == "ENABLED"`.
+  ///
+  /// `public` because both the SDK init (DAPI fan-out) and the
+  /// SwiftExampleApp's SPV start path call this against the same
+  /// endpoint — keeping it in one place means a single round-trip
+  /// per build instead of two, but callers must handle their own
+  /// caching if needed.
+  public static func discoverActiveMasternodes(
+    quorumBase: String
+  ) -> [(spvPeer: String, dapiUrl: String)]? {
+    guard
+      var components = URLComponents(string: quorumBase),
+      let scheme = components.scheme,
+      !scheme.isEmpty
+    else { return nil }
+    if components.path.hasSuffix("/") {
+      components.path = String(components.path.dropLast())
+    }
+    components.path += "/masternodes"
+    guard let url = components.url else { return nil }
+
+    var request = URLRequest(url: url)
+    request.timeoutInterval = 5.0
+    request.httpMethod = "GET"
+
+    // Reference-typed box for the response so the completion
+    // handler can safely store into it from URLSession's worker
+    // thread without violating Swift 6 strict-concurrency capture
+    // rules (which forbid mutating a captured `var Data?` from a
+    // concurrently-executing closure). The semaphore guarantees
+    // we only read `box.data` after the closure has run to
+    // completion, so the cross-thread access is data-race-free.
+    final class ResponseBox: @unchecked Sendable {
+      var data: Data?
+    }
+    let box = ResponseBox()
+    let semaphore = DispatchSemaphore(value: 0)
+    let task = URLSession.shared.dataTask(with: request) { data, _, _ in
+      box.data = data
+      semaphore.signal()
+    }
+    task.resume()
+    _ = semaphore.wait(timeout: .now() + .seconds(6))
+    guard let data = box.data else {
+      task.cancel()
+      return nil
+    }
+
+    struct Envelope: Decodable {
+      let success: Bool
+      let data: [Masternode]
+    }
+    struct Masternode: Decodable {
+      let address: String          // "ip:CoreP2PPort"
+      let status: String
+      let platformHTTPPort: UInt16
+    }
+
+    guard
+      let env = try? JSONDecoder().decode(Envelope.self, from: data),
+      env.success
+    else { return nil }
+
+    let active: [(String, String)] = env.data.compactMap { mn in
+      guard mn.status == "ENABLED" else { return nil }
+      let host = mn.address.split(separator: ":").first.map(String.init) ?? mn.address
+      return (mn.address, "https://\(host):\(mn.platformHTTPPort)")
+    }
+    return active.isEmpty ? nil : active
+  }
+
+  /// Synchronously fetch `{quorumBase}/masternodes` and build a
+  /// comma-separated DAPI URL list (`https://<ip>:<platformHTTPPort>,…`).
+  /// Returns nil on any error (network failure, JSON shape mismatch,
+  /// timeout). Used by `init(network:)` to auto-populate the DAPI
+  /// fan-out list on devnet when the user hasn't supplied one
+  /// manually — saves the "you must paste 13 URLs" UX.
+  ///
+  /// Filters to `status == "ENABLED"` so down / banned nodes don't
+  /// pollute the AddressList (the DAPI client would ban them on
+  /// first request anyway, but skipping them up front speeds the
+  /// first sync).
+  private static func discoverDAPIAddresses(quorumBase: String) -> String? {
+    guard let active = discoverActiveMasternodes(quorumBase: quorumBase) else {
+      return nil
+    }
+    return active.map(\.dapiUrl).joined(separator: ",")
+  }
+
   /// Create a new SDK instance with trusted setup
   ///
   /// This uses a trusted context provider that fetches quorum keys and
@@ -98,34 +206,83 @@ public final class SDK: @unchecked Sendable {
     var config = DashSDKConfig()
     config.network = network.ffiValue
     config.dapi_addresses = nil
+    config.quorum_url = nil
     config.skip_asset_lock_proof_verification = false
     config.request_retry_count = 1
     config.request_timeout_ms = 8000 // 8 seconds
     config.platform_version = platformVersion // 0 = SDK default (auto-detect)
 
-    // Create SDK with trusted setup — Rust side auto-detects local/regtest
-    // and uses the quorum sidecar at localhost:22444 instead of remote endpoints.
+    // Create SDK with trusted setup. DAPI / quorum-URL overrides come from
+    // UserDefaults and apply on:
     //
-    // Regtest has no remote DAPI defaults on the Rust side, so it
-    // *must* be constructed with a local DAPI address regardless of
-    // the user-facing `useDockerSetup` toggle. Without this, building
-    // a regtest SDK from a context where the toggle has been
-    // auto-disabled (e.g. orphan-mnemonic recovery routing wallets to
-    // their original network from a non-regtest active state) fails
-    // with `DAPI addresses not available for network: Regtest` and
-    // the recovery loop stalls.
+    //   * Regtest unconditionally — the Rust side has no built-in DAPI
+    //     defaults for it, so we must supply addresses every time
+    //     (otherwise SDK creation panics with `DAPI addresses not
+    //     available for network: Regtest`, which would stall orphan-
+    //     mnemonic recovery if it ran from a non-regtest active state).
+    //   * Devnet unconditionally — same reason; additionally needs an
+    //     explicit `quorum_url` because the default quorum endpoint
+    //     `https://quorums.devnet.<name>.networks.dash.org` is template-
+    //     interpolated from a devnet name we don't carry across FFI.
+    //   * Mainnet/testnet only when the user opted in via
+    //     `useDockerSetup` (existing dashmate-on-localhost flow). When
+    //     that toggle is off, the Rust side picks the canonical seed
+    //     addresses for the network.
+    //
+    // `quorum_url` is forwarded whenever the UserDefaults override is
+    // set, regardless of network — supports custom mainnet/testnet
+    // shards and any future deployment that needs a non-default
+    // endpoint.
     let result: DashSDKResult
-    let forceLocal = network == .regtest
+    let useOverrideAddresses = network == .regtest
+        || network == .devnet
         || UserDefaults.standard.bool(forKey: "useDockerSetup")
-    if forceLocal {
-      let localAddresses = Self.platformDAPIAddresses
-      result = localAddresses.withCString { addressesCStr -> DashSDKResult in
-        var mutableConfig = config
-        mutableConfig.dapi_addresses = addressesCStr
-        return dash_sdk_create_trusted(&mutableConfig)
+    let overrideQuorumURL: String? = Self.platformQuorumURL
+
+    // Resolve the DAPI address list. Two paths:
+    //
+    //   * Devnet → ALWAYS auto-discover from `{quorumURL}/masternodes`
+    //     fresh on every SDK build. The user input surface for devnet
+    //     is just the quorum URL — DAPI nodes are an implementation
+    //     detail of which masternodes happen to be ENABLED right now.
+    //     Doing this every init is what makes the path self-healing
+    //     when a node goes down on the chain. Cheap: one HTTP round-
+    //     trip (~200ms) at network-switch cadence, which the user
+    //     pays for explicitly anyway.
+    //
+    //   * Regtest / `useDockerSetup` → respect the existing
+    //     `platformDAPIAddresses` UserDefaults override (default
+    //     `http://127.0.0.1:2443`). This is the dashmate-local flow
+    //     that's been stable; it has no /masternodes service to
+    //     consult.
+    //
+    //   * Mainnet/testnet without overrides → Rust side picks seeds.
+    let overrideAddresses: String?
+    if network == .devnet {
+      if let quorum = overrideQuorumURL,
+         let discovered = Self.discoverDAPIAddresses(quorumBase: quorum) {
+        overrideAddresses = discovered
+      } else {
+        // Quorum URL unset, or /masternodes unreachable / wrong shape.
+        // Fall through with nil; Rust will refuse to build the SDK
+        // and the resulting error surfaces in the iOS UI as
+        // "Disconnected", prompting the user to fix the Quorum URL.
+        overrideAddresses = nil
       }
+    } else if useOverrideAddresses {
+      overrideAddresses = Self.platformDAPIAddresses
     } else {
-      result = dash_sdk_create_trusted(&config)
+      overrideAddresses = nil
+    }
+
+    result = SDK.withOptionalCStrings(
+      overrideAddresses,
+      overrideQuorumURL
+    ) { addressesCStr, quorumCStr in
+      var mutableConfig = config
+      if let addressesCStr { mutableConfig.dapi_addresses = addressesCStr }
+      if let quorumCStr { mutableConfig.quorum_url = quorumCStr }
+      return dash_sdk_create_trusted(&mutableConfig)
     }
 
     // Check for errors
@@ -146,6 +303,34 @@ public final class SDK: @unchecked Sendable {
     // Store the handle and network
     handle = result.data?.assumingMemoryBound(to: SDKHandle.self)
     self.network = network
+  }
+
+  /// Run `body` with two optional C-string pointers. Each input string,
+  /// when non-nil, is materialized into a NUL-terminated C buffer that is
+  /// valid for the duration of the call; nil inputs pass through as nil
+  /// pointers. Mirrors `String.withCString` for the two-optional-string
+  /// case so the SDK init can hand both `dapi_addresses` and
+  /// `quorum_url` into a single FFI call without nested withCString
+  /// closures.
+  private static func withOptionalCStrings<R>(
+    _ a: String?,
+    _ b: String?,
+    _ body: (UnsafePointer<CChar>?, UnsafePointer<CChar>?) -> R
+  ) -> R {
+    switch (a, b) {
+    case (nil, nil):
+      return body(nil, nil)
+    case (.some(let sa), nil):
+      return sa.withCString { body($0, nil) }
+    case (nil, .some(let sb)):
+      return sb.withCString { body(nil, $0) }
+    case (.some(let sa), .some(let sb)):
+      return sa.withCString { aPtr in
+        sb.withCString { bPtr in
+          body(aPtr, bPtr)
+        }
+      }
+    }
   }
 
   /// Load known contracts into the trusted context provider
--- a/packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
@@ -108,13 +108,21 @@ public final class SDK: @unchecked Sendable {
   /// takes the verbatim `address` field (`<ip>:<CoreP2PPort>`).
   ///
   /// Returns nil on any failure (timeout, JSON shape mismatch, etc.).
-  /// Filters to `status == "ENABLED"`.
+  /// Filters to `status == "ENABLED" && version_check == "success"`
+  /// to match the Rust trusted-context provider's active-node policy
+  /// (see `rs-sdk-trusted-context-provider/src/provider.rs`). Without
+  /// the `version_check` filter, nodes the quorum service has
+  /// already flagged as incompatible would be seeded into both the
+  /// DAPI fan-out and the SPV peer list, undermining the
+  /// self-healing rebuild this enables.
   ///
   /// `public` because both the SDK init (DAPI fan-out) and the
-  /// SwiftExampleApp's SPV start path call this against the same
-  /// endpoint — keeping it in one place means a single round-trip
-  /// per build instead of two, but callers must handle their own
-  /// caching if needed.
+  /// SwiftExampleApp's SPV start path call it independently against
+  /// the same endpoint — each caller pays its own round-trip, with
+  /// no shared cache. An SDK rebuild on devnet therefore performs
+  /// two `/masternodes` fetches; if that becomes a problem, the
+  /// expectation is that callers add a short-lived cache locally
+  /// (or refactor to share one through the SDK).
   public static func discoverActiveMasternodes(
     quorumBase: String
   ) -> [(spvPeer: String, dapiUrl: String)]? {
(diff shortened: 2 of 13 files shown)


packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
```diff
@@ -89,6 +89,114 @@ public final class SDK: @unchecked Sendable {
...
+      let env = try? JSONDecoder().decode(Envelope.self, from: data),
+      env.success
+    else { return nil }
+
+    let active: [(String, String)] = env.data.compactMap { mn in
+      guard mn.status == "ENABLED" else { return nil }
+      let host = mn.address.split(separator: ":").first.map(String.init) ?? mn.address
+      return (mn.address, "https://\(host):\(mn.platformHTTPPort)")
+    }
+    return active.isEmpty ? nil : active
```
thepastaclaw, line 207: **🔴 Blocking: Swift masternode discovery is stricter than the Rust provider it shadows**

`discoverActiveMasternodes` declares `platformHTTPPort: UInt16` as a required Decodable field and filters only on `status == "ENABLED"`. The equivalent Rust trusted-context provider at `packages/rs-sdk-trusted-context-provider/src/provider.rs:316-327` treats `platform_http_port` as optional with a per-network default and additionally filters on `version_check == "success"`. Two practical consequences for the devnet path this PR enables:

1. A single masternode entry in the JSON missing `platformHTTPPort` makes the whole decode fail (`try?` returns nil), so `init(.devnet)` falls through with `overrideAddresses = nil` and the SDK refuses to build — even though the Rust-side discovery would have produced a usable list.
2. Nodes the quorum service has already flagged as `versionCheck != success` get seeded into both the DAPI fan-out and `spvPeerOverride`, undermining the self-healing behavior this PR cites as the reason for re-discovering on every build.

Align the Swift decoder with the Rust source of truth.

```suggestion
    struct Envelope: Decodable {
      let success: Bool
      let data: [Masternode]
    }
    struct Masternode: Decodable {
      let address: String          // "ip:CoreP2PPort"
      let status: String
      let versionCheck: String?
      let platformHTTPPort: UInt16?
    }

    guard
      let env = try? JSONDecoder().decode(Envelope.self, from: data),
      env.success
    else { return nil }

    let defaultDapiPort: UInt16 = 443
    let active: [(String, String)] = env.data.compactMap { mn in
      guard mn.status == "ENABLED", mn.versionCheck == "success" else { return nil }
      let host = mn.address.split(separator: ":").first.map(String.init) ?? mn.address
      let dapiPort = mn.platformHTTPPort ?? defaultDapiPort
      return (mn.address, "https://\(host):\(dapiPort)")
    }
    return active.isEmpty ? nil : active
```

<sub>source: ['codex']</sub>

packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
```diff
@@ -89,6 +89,114 @@ public final class SDK: @unchecked Sendable {
...
+    let semaphore = DispatchSemaphore(value: 0)
+    let task = URLSession.shared.dataTask(with: request) { data, _, _ in
+      box.data = data
+      semaphore.signal()
+    }
+    task.resume()
+    _ = semaphore.wait(timeout: .now() + .seconds(6))
+    guard let data = box.data else {
+      task.cancel()
+      return nil
```
thepastaclaw, line 164: **🟡 Suggestion: Devnet discovery blocks the main actor for up to 6 s per SDK build**

`discoverActiveMasternodes` issues an async `URLSession` request and then waits on a `DispatchSemaphore` with a 6 s ceiling on top of the 5 s `URLRequest.timeoutInterval`. `SDK.init(network:)` calls this synchronously for devnet, and `init` is invoked from `AppState.initializeSDK` / `switchNetwork` on a `@MainActor` class, plus a second call from `CoreContentView.spvPeerOverride()` on the SPV start path. A slow or unreachable quorum-list service will freeze the UI exactly when the user is trying to fix connectivity, with no loading affordance. Since this path is now mandatory for devnet, expose an `async` variant (or run discovery off the main thread before constructing the SDK) so the synchronous shape doesn't bleed into UI-driven code.

packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
```diff
@@ -89,6 +89,114 @@ public final class SDK: @unchecked Sendable {
...
+    guard
+      var components = URLComponents(string: quorumBase),
+      let scheme = components.scheme,
+      !scheme.isEmpty
+    else { return nil }
+    if components.path.hasSuffix("/") {
+      components.path = String(components.path.dropLast())
+    }
+    components.path += "/masternodes"
+    guard let url = components.url else { return nil }
```
thepastaclaw, line 138: **🟡 Suggestion: Devnet discovery accepts cleartext quorum URLs, enabling MITM of trusted SDK setup**

`discoverActiveMasternodes` accepts any URL scheme, and the same `platformQuorumURL` is passed to both auto-discovery and `dash_sdk_create_trusted` (used for `TrustedHttpContextProvider`'s quorum-key fetch). With the new `NSAllowsArbitraryLoads = true` plist, an attacker on the network — or a hostile DNS resolver — that intercepts an `http://...` quorum URL can rewrite the `/masternodes` payload to redirect SPV/DAPI traffic at attacker hosts, and serve matching quorum keys to the trusted provider so forged platform state verifies as authentic. There is also no validation on the parsed `address`: `mn.address.split(separator: ":").first` is used verbatim in URL construction, so an entry like `"address": "evil.example:1@victim:443"` could parse surprisingly. At minimum, reject non-`https` quorum URLs unless the host is a loopback literal (so a devnet `http://127.0.0.1:8080` still works), and validate that the parsed host is a plain IP literal or DNS label before reusing it.

```suggestion
    guard
      var components = URLComponents(string: quorumBase),
      let scheme = components.scheme?.lowercased(),
      let host = components.host?.lowercased(),
      scheme == "https"
        || (scheme == "http" && ["localhost", "127.0.0.1", "::1"].contains(host))
    else { return nil }
    if components.path.hasSuffix("/") {
      components.path = String(components.path.dropLast())
    }
    components.path += "/masternodes"
    guard let url = components.url else { return nil }
```

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp.xcodeproj/project.pbxproj
```diff
@@ -460,7 +461,8 @@
 				CURRENT_PROJECT_VERSION = 1;
 				DEVELOPMENT_TEAM = 44RJ69WHFF;
 				ENABLE_PREVIEWS = YES;
-				GENERATE_INFOPLIST_FILE = YES;
+				GENERATE_INFOPLIST_FILE = NO;
+				INFOPLIST_FILE = Info.plist;
```
thepastaclaw: **🟡 Suggestion: Release config also points at Info.plist with `NSAllowsArbitraryLoads = true`**

`GENERATE_INFOPLIST_FILE = NO` + `INFOPLIST_FILE = Info.plist` is set in both Debug (~line 432) and Release (~line 461) configs, and the new `Info.plist` enables `NSAllowsArbitraryLoads`, globally disabling ATS. Every `URLSession` request in the app — not just the documented devnet `/masternodes` fetch — can then use plaintext HTTP. The plist's own comment says this MUST be removed before shipping, but nothing in the build system enforces it: a Release archive built today ships with ATS off. Gate the exception to Debug-only via a separate `Info.plist` / `xcconfig` per configuration, or scope it through `NSExceptionDomains` to the quorum-list host(s) only.

<sub>source: ['claude']</sub>

packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
```diff
@@ -89,6 +89,114 @@ public final class SDK: @unchecked Sendable {
...
+  /// takes the verbatim `address` field (`<ip>:<CoreP2PPort>`).
+  ///
+  /// Returns nil on any failure (timeout, JSON shape mismatch, etc.).
+  /// Filters to `status == "ENABLED"`.
+  ///
+  /// `public` because both the SDK init (DAPI fan-out) and the
+  /// SwiftExampleApp's SPV start path call this against the same
+  /// endpoint — keeping it in one place means a single round-trip
+  /// per build instead of two, but callers must handle their own
+  /// caching if needed.
```
thepastaclaw: **💬 Nitpick: Doc claim of "single round-trip per build" contradicts call sites**

The doc comment says "keeping it in one place means a single round-trip per build instead of two, but callers must handle their own caching if needed." The two known call sites — `SDK.init(network:)` via `discoverDAPIAddresses` and `CoreContentView.spvPeerOverride()` (`CoreContentView.swift:636`) — each invoke `discoverActiveMasternodes` with no shared cache, so an SDK rebuild on devnet performs two HTTP fetches against `/masternodes` rather than one. Either add a short-lived cache as the comment implies, or rephrase to make clear that each caller pays its own round-trip.

<sub>source: ['claude']</sub>

packages/rs-sdk-ffi/src/sdk.rs
```diff
@@ -600,6 +638,7 @@ pub unsafe extern "C" fn dash_sdk_create_with_callbacks(
             skip_asset_lock_proof_verification: config_ref.skip_asset_lock_proof_verification,
             request_retry_count: config_ref.request_retry_count,
             request_timeout_ms: config_ref.request_timeout_ms,
+            quorum_url: config_ref.quorum_url,
             platform_version: config_ref.platform_version,
         },
```
thepastaclaw, line 643: **💬 Nitpick: `quorum_url` forwarded into extended config but never consumed on callback path**

`dash_sdk_create_with_callbacks` copies `config_ref.quorum_url` into `DashSDKConfigExtended.base_config.quorum_url`, but `dash_sdk_create_extended` only reads `base_config.dapi_addresses`, `network`, and `platform_version` — the callback-based path uses `CallbackContextProvider`, not `TrustedHttpContextProvider`, so `quorum_url` is silently dropped. As written, callers can pass a quorum URL through `dash_sdk_create_with_callbacks` and it appears to be wired up. Either drop the forwarding (and document that quorum_url is only honored on the `dash_sdk_create_trusted` path) or surface a clear error when it is non-null here.

<sub>source: ['claude']</sub>

packages/swift-sdk/SwiftExampleApp/Info.plist
```diff
@@ -0,0 +1,68 @@
...
+
+		`NSAllowsArbitraryLoads` lets the app fetch from plain `http://`
+		endpoints. Required so `SDK.discoverActiveMasternodes` can hit
+		`http://<host>:8080/masternodes` (the dash quorum-list-server
+		runs HTTP by default — see
+		`/Users/ivanshumkov/Projects/dashpay/quorum-list-server`).
+
+		On a production build this MUST be removed and the
+		quorum-list-server fronted by HTTPS. Tracked alongside the rest
+		of the devnet-test scaffolding (dashpay/platform#3714).
```
thepastaclaw: **💬 Nitpick: Bundled developer absolute path leaks build-environment info**

The plist comment references `/Users/ivanshumkov/Projects/dashpay/quorum-list-server`. That string ships inside every `.app` built with this config and is recoverable from a binary plist dump — it discloses the developer's local username and project layout, and it's the kind of leftover that tends to outlive the test-only scaffolding it documents. Replace with a repo-relative pointer (e.g. the GitHub URL of `quorum-list-server`).

<sub>source: ['claude']</sub>

thepastaclaw (changes requested):
## Code Review

Two blocking issues confirmed: an integration test struct literal in rs-sdk-ffi is missing the new `quorum_url` field (test target won't compile), and the Swift `discoverActiveMasternodes` decoder is materially stricter than the Rust trusted-context provider it shadows — it requires `platformHTTPPort` on every entry and skips the `versionCheck == "success"` filter, so a single quirky masternode row can break devnet SDK creation entirely. Remaining items are devnet-scaffolding concerns: synchronous fetch on the main actor, broad ATS bypass that bleeds into Release builds, cleartext-HTTP trust in the discovery path, and a couple of doc / dead-code nits.

🔴 2 blocking | 🟡 3 suggestion(s) | 💬 3 nitpick(s)

<details>

```

In `packages/rs-sdk-ffi/tests/context_provider_test.rs`:
- [BLOCKING] packages/rs-sdk-ffi/tests/context_provider_test.rs:82-89: Integration test struct literal missing new `quorum_url` field — won't compile
  `DashSDKConfig` in `packages/rs-sdk-ffi/src/types.rs:88` gained a required `quorum_url: *const c_char` field. The in-crate token unit tests were updated, but this integration test under `tests/` was missed. Because integration tests are compiled as a separate crate, `cargo check -p rs-sdk-ffi` (which the PR description cites as clean) does not type-check them — only `cargo check -p rs-sdk-ffi --tests` / `cargo test -p rs-sdk-ffi` does. CI for the FFI crate's test target is broken on this commit and will fail with `error[E0063]: missing field 'quorum_url' in initializer of 'DashSDKConfig'`.

In `packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift`:
- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift:159-179: Swift masternode discovery is stricter than the Rust provider it shadows
  `discoverActiveMasternodes` declares `platformHTTPPort: UInt16` as a required Decodable field and filters only on `status == "ENABLED"`. The equivalent Rust trusted-context provider at `packages/rs-sdk-trusted-context-provider/src/provider.rs:316-327` treats `platform_http_port` as optional with a per-network default and additionally filters on `version_check == "success"`. Two practical consequences for the devnet path this PR enables:

  1. A single masternode entry in the JSON missing `platformHTTPPort` makes the whole decode fail (`try?` returns nil), so `init(.devnet)` falls through with `overrideAddresses = nil` and the SDK refuses to build — even though the Rust-side discovery would have produced a usable list.
  2. Nodes the quorum service has already flagged as `versionCheck != success` get seeded into both the DAPI fan-out and `spvPeerOverride`, undermining the self-healing behavior this PR cites as the reason for re-discovering on every build.

  Align the Swift decoder with the Rust source of truth.
- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift:147-156: Devnet discovery blocks the main actor for up to 6 s per SDK build
  `discoverActiveMasternodes` issues an async `URLSession` request and then waits on a `DispatchSemaphore` with a 6 s ceiling on top of the 5 s `URLRequest.timeoutInterval`. `SDK.init(network:)` calls this synchronously for devnet, and `init` is invoked from `AppState.initializeSDK` / `switchNetwork` on a `@MainActor` class, plus a second call from `CoreContentView.spvPeerOverride()` on the SPV start path. A slow or unreachable quorum-list service will freeze the UI exactly when the user is trying to fix connectivity, with no loading affordance. Since this path is now mandatory for devnet, expose an `async` variant (or run discovery off the main thread before constructing the SDK) so the synchronous shape doesn't bleed into UI-driven code.
- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift:121-130: Devnet discovery accepts cleartext quorum URLs, enabling MITM of trusted SDK setup
  `discoverActiveMasternodes` accepts any URL scheme, and the same `platformQuorumURL` is passed to both auto-discovery and `dash_sdk_create_trusted` (used for `TrustedHttpContextProvider`'s quorum-key fetch). With the new `NSAllowsArbitraryLoads = true` plist, an attacker on the network — or a hostile DNS resolver — that intercepts an `http://...` quorum URL can rewrite the `/masternodes` payload to redirect SPV/DAPI traffic at attacker hosts, and serve matching quorum keys to the trusted provider so forged platform state verifies as authentic. There is also no validation on the parsed `address`: `mn.address.split(separator: ":").first` is used verbatim in URL construction, so an entry like `"address": "evil.example:1@victim:443"` could parse surprisingly. At minimum, reject non-`https` quorum URLs unless the host is a loopback literal (so a devnet `http://127.0.0.1:8080` still works), and validate that the parsed host is a plain IP literal or DNS label before reusing it.

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp.xcodeproj/project.pbxproj`:
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp.xcodeproj/project.pbxproj:461-465: Release config also points at Info.plist with `NSAllowsArbitraryLoads = true`
  `GENERATE_INFOPLIST_FILE = NO` + `INFOPLIST_FILE = Info.plist` is set in both Debug (~line 432) and Release (~line 461) configs, and the new `Info.plist` enables `NSAllowsArbitraryLoads`, globally disabling ATS. Every `URLSession` request in the app — not just the documented devnet `/masternodes` fetch — can then use plaintext HTTP. The plist's own comment says this MUST be removed before shipping, but nothing in the build system enforces it: a Release archive built today ships with ATS off. Gate the exception to Debug-only via a separate `Info.plist` / `xcconfig` per configuration, or scope it through `NSExceptionDomains` to the quorum-list host(s) only.
```
</details>

packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
thepastaclaw, line 207: **🔴 Blocking: Swift /masternodes decoder diverges from Rust trusted-provider contract**

`discoverActiveMasternodes` declares `platformHTTPPort: UInt16` (non-optional) and filters only on `status == "ENABLED"`. The Rust source of truth at `packages/rs-sdk-trusted-context-provider/src/provider.rs:309-336` treats `platform_http_port` as optional with a per-network default (443 mainnet / 1443 testnet / 443 otherwise) and additionally requires `version_check == "success"`. Two consequences on this PR's mandatory devnet path:

1. A single `/masternodes` row missing `platformHTTPPort` makes the whole decode fail (`try?` → nil), so `SDK.init(.devnet)` ends up with `overrideAddresses = nil` and devnet SDK creation fails even though Rust would have produced a usable list.
2. Nodes already flagged `versionCheck != success` get seeded into both the DAPI fan-out and `spvPeerOverride`, undercutting the self-healing claim in the PR description.

Align the Swift decoder with Rust.

```suggestion
    struct Envelope: Decodable {
      let success: Bool
      let data: [Masternode]
    }
    struct Masternode: Decodable {
      let address: String          // "ip:CoreP2PPort"
      let status: String
      let versionCheck: String?
      let platformHTTPPort: UInt16?
    }

    guard
      let env = try? JSONDecoder().decode(Envelope.self, from: data),
      env.success
    else { return nil }

    let defaultDapiPort: UInt16 = 443
    let active: [(String, String)] = env.data.compactMap { mn in
      guard mn.status == "ENABLED", mn.versionCheck == "success" else { return nil }
      let host = mn.address.split(separator: ":").first.map(String.init) ?? mn.address
      let dapiPort = mn.platformHTTPPort ?? defaultDapiPort
      return (mn.address, "https://\(host):\(dapiPort)")
    }
    return active.isEmpty ? nil : active
```

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp.xcodeproj/project.pbxproj
thepastaclaw: **🔴 Blocking: Release builds inherit a global ATS bypass via shared Info.plist**

Both Debug (project.pbxproj:435-436) and Release (project.pbxproj:464-465) set `GENERATE_INFOPLIST_FILE = NO; INFOPLIST_FILE = Info.plist;`, and that `Info.plist` enables `NSAllowsArbitraryLoads = true` (Info.plist:62-66). The plist comment explicitly says this exception is test-only and MUST be removed before shipping, but the build system does not enforce that: a Release archive built from this branch disables ATS for every `URLSession` in the app, not just the devnet `/masternodes` probe. Combined with the unrestricted-scheme quorum URL accepted by `discoverActiveMasternodes`, this widens the MITM surface for any user running the example app on an untrusted network. Gate the exception to Debug only (separate per-config Info.plist / xcconfig) or scope it through `NSExceptionDomains` to the quorum-list host(s) only.

thepastaclaw (changes requested):
## Code Review

Prior compile-break (missing quorum_url in context_provider_test) is resolved at this SHA. Two convergent blocking issues remain: the Swift /masternodes decoder diverges from the Rust trusted-provider contract (non-optional platformHTTPPort, missing version_check filter), and the Info.plist that disables ATS is wired into both Debug and Release configs. Additional suggestions cover main-actor blocking discovery, cleartext quorum URLs feeding the trusted setup, and a dropped quorum_url on the callback-based create path.

<details>

```

In `packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift`:
- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift:159-179: Swift /masternodes decoder diverges from Rust trusted-provider contract
  `discoverActiveMasternodes` declares `platformHTTPPort: UInt16` (non-optional) and filters only on `status == "ENABLED"`. The Rust source of truth at `packages/rs-sdk-trusted-context-provider/src/provider.rs:309-336` treats `platform_http_port` as optional with a per-network default (443 mainnet / 1443 testnet / 443 otherwise) and additionally requires `version_check == "success"`. Two consequences on this PR's mandatory devnet path:

  1. A single `/masternodes` row missing `platformHTTPPort` makes the whole decode fail (`try?` → nil), so `SDK.init(.devnet)` ends up with `overrideAddresses = nil` and devnet SDK creation fails even though Rust would have produced a usable list.
  2. Nodes already flagged `versionCheck != success` get seeded into both the DAPI fan-out and `spvPeerOverride`, undercutting the self-healing claim in the PR description.

  Align the Swift decoder with Rust.

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp.xcodeproj/project.pbxproj`:
- [BLOCKING] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp.xcodeproj/project.pbxproj:464-465: Release builds inherit a global ATS bypass via shared Info.plist
  Both Debug (project.pbxproj:435-436) and Release (project.pbxproj:464-465) set `GENERATE_INFOPLIST_FILE = NO; INFOPLIST_FILE = Info.plist;`, and that `Info.plist` enables `NSAllowsArbitraryLoads = true` (Info.plist:62-66). The plist comment explicitly says this exception is test-only and MUST be removed before shipping, but the build system does not enforce that: a Release archive built from this branch disables ATS for every `URLSession` in the app, not just the devnet `/masternodes` probe. Combined with the unrestricted-scheme quorum URL accepted by `discoverActiveMasternodes`, this widens the MITM surface for any user running the example app on an untrusted network. Gate the exception to Debug only (separate per-config Info.plist / xcconfig) or scope it through `NSExceptionDomains` to the quorum-list host(s) only.

In `packages/rs-sdk-ffi/src/sdk.rs`:
- [SUGGESTION] packages/rs-sdk-ffi/src/sdk.rs:633-646: `quorum_url` propagated into callback-based config but silently ignored
  `dash_sdk_create_with_callbacks` copies `config_ref.quorum_url` into `DashSDKConfigExtended.base_config` at sdk.rs:634-643, but the callback path uses `CallbackContextProvider` and `dash_sdk_create_extended` never reads `base_config.quorum_url`. A C/Swift caller populating the field gets no error and no effect — the API surface implies support the Rust side does not provide. Either reject non-null `quorum_url` here with `InvalidParameter`, or zero it and document that `quorum_url` is honored only by `dash_sdk_create_trusted`.
```
</details>

packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
```diff
@@ -89,6 +89,142 @@ public final class SDK: @unchecked Sendable {
...
+      // Optional because older quorum-list-server builds may omit it;
+      // callers below treat missing as "not success" (i.e. excluded).
+      let versionCheck: String?
+
+      enum CodingKeys: String, CodingKey {
+        case address
+        case status
+        case platformHTTPPort
+        case versionCheck = "version_check"
+      }
```
thepastaclaw: **🔴 Blocking: versionCheck CodingKey maps to the wrong JSON key, breaking every devnet SDK build**

The new `CodingKeys` enum declares `case versionCheck = "version_check"`, telling `JSONDecoder` to look for snake_case `version_check` in the `/masternodes` response. The Rust source of truth (`packages/rs-sdk-trusted-context-provider/src/provider.rs:73-80`) renames the Rust field with `#[serde(rename = "versionCheck")]`, i.e. the actual JSON key the quorum-list-server emits is camelCase `versionCheck`. Because Swift now looks for the wrong key, every decoded `Masternode.versionCheck` is `nil`, so the new filter `mn.versionCheck == "success"` (line 202) is false for every entry, `active` is always empty, and `discoverActiveMasternodes` returns `nil`. `SDK.init(network: .devnet)` mandatorily routes through this helper (lines 289-299) and falls through to `overrideAddresses = nil`, which the Rust side then refuses to build — so every cold devnet boot at this SHA fails SDK creation, exactly the regression the version_check filter was meant to prevent. The PR's manual paloma test would have passed before this rename was added (no filter at all, so all ENABLED nodes passed). Fix by mapping the case to the camelCase key the server actually emits, or drop the rename entirely so Swift's default property-name match handles it.

```suggestion
      enum CodingKeys: String, CodingKey {
        case address
        case status
        case platformHTTPPort
        case versionCheck
      }
```

thepastaclaw (changes requested):
## Code Review

Carried-forward: two convergent suggestion-level issues remain — cleartext quorum URLs accepted at the trusted-setup entry, and synchronous main-actor blocking during devnet discovery. New in this delta: the freshly added `versionCheck` CodingKey maps to `"version_check"` (snake_case) while the Rust source of truth and quorum-list-server emit camelCase `versionCheck`. This makes the new filter universally false, so every devnet SDK build returns nil from `discoverActiveMasternodes` and fails — a hard regression that wipes out the PR's mandatory devnet path. Prior fixes for the decoder divergence, Release-config ATS bleed, leaked developer path, and stale doc claim are confirmed.

🔴 1 blocking

<details>

```

In `packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift`:
- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift:185-190: versionCheck CodingKey maps to the wrong JSON key, breaking every devnet SDK build
  The new `CodingKeys` enum declares `case versionCheck = "version_check"`, telling `JSONDecoder` to look for snake_case `version_check` in the `/masternodes` response. The Rust source of truth (`packages/rs-sdk-trusted-context-provider/src/provider.rs:73-80`) renames the Rust field with `#[serde(rename = "versionCheck")]`, i.e. the actual JSON key the quorum-list-server emits is camelCase `versionCheck`. Because Swift now looks for the wrong key, every decoded `Masternode.versionCheck` is `nil`, so the new filter `mn.versionCheck == "success"` (line 202) is false for every entry, `active` is always empty, and `discoverActiveMasternodes` returns `nil`. `SDK.init(network: .devnet)` mandatorily routes through this helper (lines 289-299) and falls through to `overrideAddresses = nil`, which the Rust side then refuses to build — so every cold devnet boot at this SHA fails SDK creation, exactly the regression the version_check filter was meant to prevent. The PR's manual paloma test would have passed before this rename was added (no filter at all, so all ENABLED nodes passed). Fix by mapping the case to the camelCase key the server actually emits, or drop the rename entirely so Swift's default property-name match handles it.
```
</details>

packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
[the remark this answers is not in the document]
shumkov: ATS fix shipped in 7cfede7b6b: Release config flipped back to GENERATE_INFOPLIST_FILE=YES, so Release archives no longer ship with the permissive plist (only Debug uses it for the /masternodes HTTP probe). Scoping further via NSExceptionDomains would fight the self-healing-on-node-churn property this PR is built around, since the host set is not known at build time. Closing.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Views/CoreContentView.swift
```diff
@@ -622,6 +622,21 @@ var body: some View {
...
+        // verbatim values rather than guessing the canonical 29999
+        // port (paloma reports 20001 per masternode, for example).
+        // No manual SPV input on devnet — the quorum URL is the
+        // single source of truth (see `OptionsView`'s devnet branch).
+        if platformState.currentNetwork == .devnet {
+            guard
+                let quorum = UserDefaults.standard.string(forKey: "platformQuorumURL"),
+                !quorum.isEmpty,
+                let active = SDK.discoverActiveMasternodes(quorumBase: quorum)
+            else { return [] }
```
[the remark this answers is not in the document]
shumkov: Not addressing in this PR. The fix is real (URLSession+DispatchSemaphore.wait blocks the @MainActor caller for up to ~6s), but fixing it properly requires making SDK.init async, which cascades through AppState.initializeSDK / switchNetwork / spvPeerOverride — an API-breaking refactor far beyond this PR scope. Filing as a separate follow-up.

packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
shumkov: Done in 7cfede7b6b + a6a9c5a9df: platformHTTPPort is now optional (default 443 fallback) and versionCheck=="success" is part of the active-node filter. a6a9c5a9df fixed a wire-format typo on the JSON key (camelCase versionCheck, not snake_case version_check — Rust uses #[serde(rename = "versionCheck")] so that camelCase is what the server actually emits). Closing.

packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
shumkov: Not addressing in this PR — see CoreContentView:637 thread for full rationale (async SDK.init is an API-breaking refactor across multiple call sites). Filing as separate follow-up.

packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
shumkov: Not addressing in this PR. The suggested https-or-loopback guard would actively break the documented use case (paloma quorum-list service at http://44.238.203.84:8080 — neither https nor loopback). The PR is wired to reach real devnet infra; tightening this needs a separate decision about how to authenticate devnet HTTP endpoints (cert pinning to the quorum-list server, etc).

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp.xcodeproj/project.pbxproj
shumkov: Done in 7cfede7b6b — Release config now uses GENERATE_INFOPLIST_FILE=YES and no longer references the permissive Info.plist. Only Debug uses the hand-managed plist (verified xcodebuild -configuration Release builds clean). Closing.

packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
shumkov: Done in 7cfede7b6b — rewrote the doc to say each caller pays its own round-trip and that a shared cache is a future addition (no callers added one in this PR). Closing.

packages/rs-sdk-ffi/src/sdk.rs
shumkov: Done in 7cfede7b6b — documented on DashSDKConfig::quorum_url that the field is only honored on the dash_sdk_create_trusted path; the callback-based path uses CallbackContextProvider and silently drops non-null values. Forwarding kept because the struct field is required, but the doc now flags it as a dead field on that path. Closing.

packages/swift-sdk/SwiftExampleApp/Info.plist
shumkov: Done in 7cfede7b6b — replaced /Users/ivanshumkov/Projects/dashpay/quorum-list-server with https://github.com/dashpay/quorum-list-server. Closing.

packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
shumkov: Already fixed in 7cfede7b6b + a6a9c5a9df. See parallel thread for SDK.swift:207 — platformHTTPPort is now optional, versionCheck filter is wired up, and the JSON key uses the correct camelCase wire format. Closing as duplicate.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp.xcodeproj/project.pbxproj
shumkov: Already fixed in 7cfede7b6b. Release now uses GENERATE_INFOPLIST_FILE=YES and the permissive Info.plist no longer ships in Release archives. See parallel pbxproj:465 thread. Closing as duplicate.

packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
shumkov: Fixed in a6a9c5a9df. Dropped the CodingKeys enum entirely — Swift Decodable defaults match property names to JSON keys literally, and the property names (platformHTTPPort, versionCheck) already match the camelCase wire keys that Rust emits via #[serde(rename = ...)]. Closing.

packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
```diff
@@ -89,6 +89,142 @@ public final class SDK: @unchecked Sendable {
...
+    guard
+      var components = URLComponents(string: quorumBase),
+      let scheme = components.scheme,
+      !scheme.isEmpty
+    else { return nil }
+    if components.path.hasSuffix("/") {
+      components.path = String(components.path.dropLast())
+    }
+    components.path += "/masternodes"
+    guard let url = components.url else { return nil }
```
thepastaclaw, line 138: **🟡 Suggestion: Carried-forward: `discoverActiveMasternodes` accepts cleartext quorum URLs that feed the trusted SDK setup**

Carried forward from the 7cfede7b review — STILL VALID at a6a9c5a9; the latest delta does not touch this path. `discoverActiveMasternodes` validates only that `URLComponents.scheme` is non-empty (SDK.swift:130-133), so plain `http://` is accepted. The same `platformQuorumURL` is then forwarded across the FFI as `config.quorum_url` into `dash_sdk_create_trusted` (SDK.swift:308-313), where the Rust `TrustedHttpContextProvider` uses it as the base URL for quorum-public-key lookups (rs-sdk-trusted-context-provider/src/provider.rs:309-330). Because the Debug `Info.plist` still ships `NSAllowsArbitraryLoads = true` and the OptionsView placeholder encourages `http://<host>:8080`, an on-path attacker on a devnet/coffee-shop network can rewrite `/masternodes` to substitute attacker-controlled DAPI/SPV peers AND serve matching forged quorum public keys to the trusted provider — collapsing the trust boundary `dash_sdk_create_trusted` is supposed to give. Require `https` unless the host is a loopback literal so dashmate-local `http://127.0.0.1:8080` keeps working.

```suggestion
    guard
      var components = URLComponents(string: quorumBase),
      let scheme = components.scheme?.lowercased(),
      let host = components.host?.lowercased(),
      scheme == "https"
        || (scheme == "http" && ["localhost", "127.0.0.1", "::1"].contains(host))
    else { return nil }
    if components.path.hasSuffix("/") {
      components.path = String(components.path.dropLast())
    }
    components.path += "/masternodes"
    guard let url = components.url else { return nil }
```

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

The latest delta (7cfede7b..a6a9c5a9) is a 9-line, single-purpose fix: it removes the wrong `CodingKeys` enum on Swift's `Masternode` decoder so that the property names `platformHTTPPort`/`versionCheck` now match the camelCase wire keys emitted by the Rust trusted-context provider via `#[serde(rename = ...)]` (provider.rs:76-79). Verified — the fix is correct and resolves the prior blocking issue; no new defects introduced. Prior-finding reconciliation: `version-check-coding-key-wrong` = FIXED; `release-ats-bypass`, `developer-path-leak`, `single-roundtrip-doc` = FIXED in earlier pushes and still resolved; cleartext quorum URLs accepted into trusted SDK setup (SDK.swift:129-138) = STILL VALID; main-actor sync semaphore on devnet discovery (SDK.swift:140-165) = STILL VALID; testnet default DAPI port divergence = STILL VALID but only reachable on the devnet code path today (dropped as not-in-scope nitpick); callback-path `quorum_url` silently dropped = INTENTIONALLY DEFERRED by author via doc comment (kept out of the public review). Carrying forward two suggestion-level findings.

🟡 1 suggestion(s)

<details>

```

In `packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift`:
- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift:129-138: Carried-forward: `discoverActiveMasternodes` accepts cleartext quorum URLs that feed the trusted SDK setup
  Carried forward from the 7cfede7b review — STILL VALID at a6a9c5a9; the latest delta does not touch this path. `discoverActiveMasternodes` validates only that `URLComponents.scheme` is non-empty (SDK.swift:130-133), so plain `http://` is accepted. The same `platformQuorumURL` is then forwarded across the FFI as `config.quorum_url` into `dash_sdk_create_trusted` (SDK.swift:308-313), where the Rust `TrustedHttpContextProvider` uses it as the base URL for quorum-public-key lookups (rs-sdk-trusted-context-provider/src/provider.rs:309-330). Because the Debug `Info.plist` still ships `NSAllowsArbitraryLoads = true` and the OptionsView placeholder encourages `http://<host>:8080`, an on-path attacker on a devnet/coffee-shop network can rewrite `/masternodes` to substitute attacker-controlled DAPI/SPV peers AND serve matching forged quorum public keys to the trusted provider — collapsing the trust boundary `dash_sdk_create_trusted` is supposed to give. Require `https` unless the host is a loopback literal so dashmate-local `http://127.0.0.1:8080` keeps working.
```
</details>

packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
shumkov: Duplicate of the SDK.swift:138 thread already addressed earlier in this round (https://github.com/dashpay/platform/pull/3755#discussion_r3312523522). Same rationale: not addressing in this PR because https-or-loopback would actively break the documented use case (paloma quorum-list service at http://44.238.203.84:8080 — neither https nor loopback). Filing as a separate decision about how to authenticate devnet HTTP endpoints. Closing as duplicate.

PR #3756, opened 2026-05-27
feat(drive-abci, sdk)!: allow shielded-notes queries to span 4 MMR chunks

QuantumExplorer:

Cold-syncing a 1M-note shielded pool from iPhone against paloma took ~96 s at `max_concurrent=16`. Per-chunk timing (via a throwaway bench against the live chain) showed the cost is **server-side fixed overhead per request** — ~1.5 s/chunk regardless of payload size, dominated by GroveDB walk + proof generation + envoy hop. Trial decrypt is ~1.3 s for the **entire 1M notes** (rayon-irrelevant), proof verify is ~4 ms/chunk. Bandwidth was nowhere near saturated.

The on-chain MMR chunks hold 2048 notes each (`SHIELDED_NOTES_CHUNK_POWER = 11`), and the query handler restricts every request to exactly one chunk by reading `max_encrypted_notes_per_query: 2048` for both the count clamp **and** the alignment requirement. But `BulkAppendTreeProof::generate` and `MmrTreeProof::generate` in grovedb already accept arbitrary multi-chunk ranges — the proof format has no single-chunk assumption. The 2048 cap is purely a query-handler convention.

## What was done?

Decouple the two coupled concerns and let a single query span up to 4 adjacent MMR chunks (8192 notes per response):

- **rs-platform-version**: rename `max_encrypted_notes_per_query: u16` → `max_query_chunks: u8` on `DriveAbciQueryShieldedVersions`. Expressed in chunks (not raw notes) so the MMR-shape coupling is explicit. v0 = 1 (preserves legacy single-chunk behaviour); v1 = 4.
- **drive-abci query handler**: alignment uses `1 << SHIELDED_NOTES_CHUNK_POWER` (the on-chain MMR chunk size, never coupled to the cap); count clamp uses `max_query_chunks × mmr_chunk_size`.
- **dash-sdk fetch loop**: same split — `mmr_chunk_size` for alignment, `fetch_size = max_query_chunks × mmr_chunk_size` for the per-request `count`. `next_chunk_index` advances by `fetch_size`.
- **drive-proof-verifier + mocks**: updated to the new constant + multiplication.

Projected impact: 488 chunks → 122 multi-chunk requests. At observed per-request cost of ~1.5 s fixed + ~200 ms wire/chunk, cold sync should drop from ~96 s to ~16 s on the same iPhone against the same paloma deployment.

## How Has This Been Tested?

- All 14 existing `query::shielded::encrypted_notes` unit tests pass unchanged (alignment + clamp exercised through the new decoupled helpers).
- Workspace check passes on the affected crates: `cargo check -p drive-abci -p drive-proof-verifier -p dash-sdk -p platform-version -p platform-wallet --features platform-wallet/shielded`.
- The bench that drove the diagnosis (\`shielded_chunk_timing_bench\` on \`test/sheilded_test_data\`) showed that response time barely changed for count=64 vs count=2048 (1.57 vs 1.52 s avg), confirming the bottleneck is per-request fixed cost — which this PR amortizes 4×.

End-to-end iPhone verification will land on a follow-up branch once a paloma rebuild is deployed.

## Breaking Changes

**Yes — consensus-breaking** (marked `feat!`).

- The query handler now accepts `count > 2048` (up to 8192 on v1). The on-chain MMR is unchanged — no rebake / snapshot regen.
- Old SDK clients (still sending `count = 2048`) keep working unchanged — they just won't realise they could ask for more.
- Old drive-abci nodes still on `max_query_chunks = 1` (v0) would silently clamp at 2048 against new clients. Mixed-version clusters degrade gracefully to single-chunk throughput, no correctness issue.

## Checklist:

- [x] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published in downstream modules

## Summary by CodeRabbit

* **Refactor**
  * Revised shielded encrypted-notes handling to use chunk-based sizing: query limits are now expressed as max query chunks × MMR chunk size, and start indexes must align to chunk boundaries. This affects querying, verification, and sync behaviors.
* **Documentation**
  * Clarified chunk-size invariants and alignment requirements in sync code comments.
* **Tests**
  * Updated unit tests and mocks to match the new chunk/limit semantics.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dashpay/platform/pull/3756?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/rs-drive-proof-verifier/src/proof.rs
+++ b/packages/rs-drive-proof-verifier/src/proof.rs
@@ -2550,7 +2550,8 @@ impl FromProof<platform::GetShieldedEncryptedNotesRequest> for ShieldedEncrypted
             .drive_abci
             .query
             .shielded_queries
-            .max_encrypted_notes_per_query as u32;
+            .max_query_chunks as u32
+            * (1u32 << drive::drive::shielded::paths::SHIELDED_NOTES_CHUNK_POWER);
 
         let (root_hash, notes) = Drive::verify_shielded_encrypted_notes(
             &proof.grovedb_proof,
--- a/packages/rs-sdk/src/platform/shielded/notes_sync/sync_shielded_notes.rs
+++ b/packages/rs-sdk/src/platform/shielded/notes_sync/sync_shielded_notes.rs
@@ -40,20 +40,38 @@ pub async fn sync_shielded_notes(
 ) -> Result<ShieldedSyncResult, Error> {
     let config = config.unwrap_or_default();
 
-    let chunk_size = sdk
+    // `mmr_chunk_size` is the on-chain MMR chunk size — the unit that
+    // `start_index` must align to. `fetch_size` is how many notes we
+    // pull per request; under `max_query_chunks=4` the server packs 4
+    // MMR chunks into one proof, so each request advances by 4× the
+    // MMR chunk size. Decoupling the two means the SDK can opportunistically
+    // request larger spans without touching the on-chain tree shape.
+    let mmr_chunk_size: u64 = 1u64 << drive::drive::shielded::paths::SHIELDED_NOTES_CHUNK_POWER;
+    let max_query_chunks = sdk
         .version()
         .drive_abci
         .query
         .shielded_queries
-        .max_encrypted_notes_per_query as u64;
+        .max_query_chunks as u64;
+    let fetch_size = mmr_chunk_size
+        .saturating_mul(max_query_chunks)
+        .max(mmr_chunk_size);
 
-    // Validate alignment
-    if chunk_size > 0 && !start_index.is_multiple_of(chunk_size) {
+    // Validate alignment against the MMR chunk size (NOT the multi-chunk
+    // fetch size). The server only requires per-MMR-chunk alignment; any
+    // multiple of `mmr_chunk_size` is a legal start. A `start_index`
+    // produced by a previous sync pass will be a multiple of `fetch_size`
+    // (and therefore of `mmr_chunk_size`) so this check is normally a
+    // no-op — but a hand-built resume point could land on a non-fetch
+    // boundary and still be valid.
+    if mmr_chunk_size > 0 && !start_index.is_multiple_of(mmr_chunk_size) {
         return Err(Error::Generic(format!(
             "start_index {} is not chunk-aligned; must be a multiple of {}",
-            start_index, chunk_size
+            start_index, mmr_chunk_size
         )));
     }
+
+    let chunk_size = fetch_size;
 
     let max_concurrent = config.max_concurrent.max(1);
     let settings = config.request_settings;
--- a/packages/rs-drive-abci/src/query/shielded/encrypted_notes/mod.rs
+++ b/packages/rs-drive-abci/src/query/shielded/encrypted_notes/mod.rs
@@ -92,8 +92,8 @@ mod tests {
     fn test_query_shielded_encrypted_notes_non_aligned_start_index_rejected() {
         let (platform, state, version) = setup_platform(None, Network::Testnet, None);
 
-        // chunk size is `max_encrypted_notes_per_query` (2048 in v1). Anything
-        // that isn't a multiple of that should be rejected as unaligned.
+        // Alignment is to the MMR chunk size (`1 << SHIELDED_NOTES_CHUNK_POWER`
+        // = 2048). Anything that isn't a multiple of that is rejected.
         let request = GetShieldedEncryptedNotesRequest {
             version: Some(RequestVersion::V0(GetShieldedEncryptedNotesRequestV0 {
                 start_index: 1,
@@ -151,9 +151,10 @@ mod tests {
 
     #[test]
     fn test_query_shielded_encrypted_notes_count_zero_uses_max() {
-        // When count == 0, the limit falls back to `max_encrypted_notes_per_query`.
-        // This exercises the "count == 0 || count > max" branch of the limit
-        // derivation without needing notes to be stored.
+        // When count == 0, the limit falls back to the per-query max
+        // (`max_query_chunks × MMR_CHUNK_SIZE`). This exercises the
+        // "count == 0 || count > max" branch of the limit derivation
+        // without needing notes to be stored.
         let (platform, state, version) = setup_platform(None, Network::Testnet, None);
 
         let request = GetShieldedEncryptedNotesRequest {
@@ -188,11 +189,8 @@ mod tests {
     #[test]
     fn test_query_shielded_encrypted_notes_count_exceeds_max_clamped() {
         let (platform, state, version) = setup_platform(None, Network::Testnet, None);
-        let max = version
-            .drive_abci
-            .query
-            .shielded_queries
-            .max_encrypted_notes_per_query as u32;
+        let max = version.drive_abci.query.shielded_queries.max_query_chunks as u32
+            * (1u32 << drive::drive::shielded::paths::SHIELDED_NOTES_CHUNK_POWER);
 
         let request = GetShieldedEncryptedNotesRequest {
             version: Some(RequestVersion::V0(GetShieldedEncryptedNotesRequestV0 {
--- a/packages/rs-drive-abci/src/query/shielded/encrypted_notes/v0/mod.rs
+++ b/packages/rs-drive-abci/src/query/shielded/encrypted_notes/v0/mod.rs
@@ -49,9 +49,9 @@ impl<C> Platform<C> {
             .query
             .shielded_queries
             .max_query_chunks as u32;
-        let max_notes = max_query_chunks
-            .saturating_mul(mmr_chunk_size as u32)
-            .min(u32::MAX);
+        // `saturating_mul` on u32 already caps at u32::MAX — no extra
+        // clamp needed.
+        let max_notes = max_query_chunks.saturating_mul(mmr_chunk_size as u32);
 
         if start_index % mmr_chunk_size != 0 {
             return Ok(QueryValidationResult::new_with_error(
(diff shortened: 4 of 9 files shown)


thepastaclaw (changes requested):
## Code Review

This PR correctly separates MMR alignment from the per-request cap on the server side, but one client proof-verification path still treats the widened cap as the alignment unit. That creates a real v1 regression for valid non-zero chunk offsets. Separately, the SDK's public resume token now rewinds farther than necessary after a partial multi-chunk fetch, which wastes bandwidth for direct SDK callers but does not break correctness.

<details>

```

In `packages/rs-drive-proof-verifier/src/proof.rs`:
- [BLOCKING] packages/rs-drive-proof-verifier/src/proof.rs:2549-2560: Proof verification now rejects valid v1 chunk-aligned requests at offsets 2048/4096/6144
  This PR changed `max_elements` from the MMR chunk size to the per-request cap (`max_query_chunks * 2048` on v1), and this caller now passes that widened value into `Drive::verify_shielded_encrypted_notes`. The verifier still uses `max_elements` as the alignment unit in `verify_shielded_encrypted_notes_v0` (`start_index.is_multiple_of(max_elements as u64)`), so a legal v1 request starting at `2048`, `4096`, or `6144` is rejected client-side even though the server accepts it and proves that range. Wallet sync can legitimately resume from those offsets because alignment is now per MMR chunk, not per full fetch window. The proof verifier needs a separate MMR alignment size instead of reusing the request cap for both roles.

In `packages/rs-sdk/src/platform/shielded/notes_sync/sync_shielded_notes.rs`:
- [SUGGESTION] packages/rs-sdk/src/platform/shielded/notes_sync/sync_shielded_notes.rs:148-165: Resume token rewinds to the start of the whole fetch window instead of the mutable tail chunk
  `next_start_index` treats any short response as a reason to restart from the beginning of that entire `fetch_size` window. Under v1, a short 8192-note response can still contain multiple complete 2048-note MMR chunks plus only one mutable tail chunk. Returning the window start forces direct SDK callers that use `ShieldedSyncResult.next_start_index` to re-download and re-verify immutable chunks on every incremental pass. The resume point should advance to the last fully sealed MMR boundary inside the partial window, and if the short response ends exactly on an MMR boundary it can advance to the end of the returned data.
```
</details>

QuantumExplorer (commented):
Approved

packages/rs-drive-proof-verifier/src/proof.rs
```diff
@@ -2550,7 +2550,8 @@ impl FromProof<platform::GetShieldedEncryptedNotesRequest> for ShieldedEncrypted
             .drive_abci
             .query
             .shielded_queries
-            .max_encrypted_notes_per_query as u32;
+            .max_query_chunks as u32
+            * (1u32 << drive::drive::shielded::paths::SHIELDED_NOTES_CHUNK_POWER);
```
thepastaclaw, line 2554: **💬 Nitpick: Carried-forward (STILL VALID): verifier uses plain `*`; server uses `saturating_mul` — same invariant, two computations**

After the latest delta the server computes `max_notes` with `saturating_mul`. The verifier still computes the same quantity with plain `*` and a non-saturating `1u32 << SHIELDED_NOTES_CHUNK_POWER`. Today's constants (`max_query_chunks ≤ 4`, `SHIELDED_NOTES_CHUNK_POWER = 11`) cannot overflow either form, so this is purely a consistency / future-proofing point: a future bump of `SHIELDED_NOTES_CHUNK_POWER` past ~22 with a sizeable `max_query_chunks` would make the server clamp to `u32::MAX` and the verifier panic in debug / wrap in release — a verification mismatch from an arithmetic asymmetry. Mirror the server-side formula. Independent of the alignment-unit bug above; this is just about making both sides compute the same value the same way.

```suggestion
        let max_elements = (platform_version
            .drive_abci
            .query
            .shielded_queries
            .max_query_chunks as u32)
            .saturating_mul(1u32 << drive::drive::shielded::paths::SHIELDED_NOTES_CHUNK_POWER);
```

<sub>source: ['claude']</sub>

packages/rs-sdk/src/platform/shielded/notes_sync/sync_shielded_notes.rs
```diff
@@ -40,21 +40,39 @@ pub async fn sync_shielded_notes(
...
+    // boundary and still be valid.
+    if mmr_chunk_size > 0 && !start_index.is_multiple_of(mmr_chunk_size) {
         return Err(Error::Generic(format!(
             "start_index {} is not chunk-aligned; must be a multiple of {}",
-            start_index, chunk_size
+            start_index, mmr_chunk_size
         )));
     }
 
+    let chunk_size = fetch_size;
```
thepastaclaw, line 74: **💬 Nitpick: Carried-forward (STILL VALID): redundant `let chunk_size = fetch_size` rebinding**

After the PR there are two legitimate 'chunk sizes' in this function — the MMR chunk size (alignment unit) and the per-request fetch window. The variable still named `chunk_size` is the fetch-window one, kept only so that downstream uses didn't need to be renamed. This is a magnet for future confusion (and is implicated in the `is_partial` / rewind findings, where the wrong 'chunk size' is being used). Either rename the downstream uses to `fetch_size` and drop the alias, or rename `fetch_size` to `chunk_size` and drop the alias — pick one term per quantity.

<sub>source: ['claude']</sub>

thepastaclaw (changes requested):
## Code Review

Incremental review at head 1a9283504. The latest delta is a one-line cleanup that drops a redundant `.min(u32::MAX)` after `saturating_mul`; correct and inconsequential. Prior-finding reconciliation: (a) FIXED — redundant `.min(u32::MAX)` clamp; (b) STILL VALID — verifier alignment unit divergence (blocking); (c) STILL VALID — SDK resume token rewinds to start of whole multi-chunk window (suggestion); (d) STILL VALID — SDK truncates / pins against any node that returns fewer notes than the local fetch_size (suggestion, partially intentional under `feat!`); (e) STILL VALID — verifier uses non-saturating `*` while server uses `saturating_mul` (nitpick); (f) STILL VALID — redundant `let chunk_size = fetch_size` rebinding (nitpick). No new defects in the latest delta itself. REQUEST_CHANGES because the cross-crate alignment mismatch (verifier vs. server) is a real correctness regression for any v1 client resuming at an MMR-aligned non-fetch boundary.

<details>

```

In `packages/rs-drive-proof-verifier/src/proof.rs`:
- [BLOCKING] packages/rs-drive-proof-verifier/src/proof.rs:2549-2560: Carried-forward (STILL VALID): verifier alignment unit is the per-query cap (8192), not the MMR chunk size (2048) — rejects valid server proofs
  The proof verifier computes `max_elements = max_query_chunks * (1 << SHIELDED_NOTES_CHUNK_POWER)` (= 8192 on v1) and passes that value to `Drive::verify_shielded_encrypted_notes`. The verifier core in `packages/rs-drive/src/verify/shielded/verify_shielded_encrypted_notes/v0/mod.rs:26-31` then uses the SAME value as the alignment unit (`!start_index.is_multiple_of(chunk_size)` with `chunk_size = max_elements`). The server-side handler in `packages/rs-drive-abci/src/query/shielded/encrypted_notes/v0/mod.rs:46-63` only requires `start_index` to be a multiple of `mmr_chunk_size` (= 2048). Concretely: a v1 client that issues a legitimate request at `start_index = 2048`, `4096`, or `6144` (e.g. on resume — `rs-platform-wallet/src/wallet/shielded/sync.rs` already rounds to 2048-note boundaries, and the SDK comment at `sync_shielded_notes.rs:60-66` explicitly documents these as valid resume points) will receive a valid proof from the server and then reject it locally with `CorruptedElementType('start_index is not chunk-aligned ...')`. Fix by passing the MMR chunk size as the alignment unit (independent of the per-query cap), either by changing the verifier's signature to take `chunk_size` separately, or by deriving `chunk_size = 1 << SHIELDED_NOTES_CHUNK_POWER` inside `verify_shielded_encrypted_notes_v0` and keeping `max_elements` solely as the upper bound on returned notes.

In `packages/rs-sdk/src/platform/shielded/notes_sync/sync_shielded_notes.rs`:
- [SUGGESTION] packages/rs-sdk/src/platform/shielded/notes_sync/sync_shielded_notes.rs:148-165: Carried-forward (STILL VALID): resume index rewinds to the start of the whole 8192-note fetch window, re-fetching sealed MMR chunks
  `chunk_size` here equals `fetch_size = max_query_chunks * mmr_chunk_size` (= 8192 on v1). When the last non-empty response is shorter than `chunk_size`, `next_start_index` is set to that response's `chunk_start` — i.e. the start of a 4-MMR-chunk window. A partial response of, say, 5000 notes contains two fully sealed MMR chunks (positions 0..4095 inside the window) plus a mutable tail chunk; the next sync re-downloads and re-verifies those two sealed chunks needlessly. Worse, a response that ends on an exact 2048-note boundary (e.g. 4096 or 6144 notes) is still shorter than `fetch_size`, so the cursor still rewinds even though every returned note is already sealed. The mutability boundary is the 2048-note MMR chunk, not the larger fetch window. Advance to the last *fully sealed MMR boundary* within the partial window — fall back to `chunk_start` only when no full MMR chunk is present.
- [SUGGESTION] packages/rs-sdk/src/platform/shielded/notes_sync/sync_shielded_notes.rs:102-121: Carried-forward (STILL VALID): `is_partial = notes.len() < fetch_size` truncates sync when talking to any node that returns fewer notes than the local fetch_size
  `is_partial` is derived from the local `fetch_size = max_query_chunks * mmr_chunk_size` only. A node that clamps the response to a smaller value (e.g. one MMR chunk, 2048 notes) — whether due to a version skew during rollout, a stale node in rotation, or a future protocol-version downgrade — will cause every response to be flagged partial, set `reached_end = true`, and abort forward scheduling. With `max_concurrent = 1` the next-sync cursor will be set to the same `chunk_start` it just fetched (see the related rewind finding), pinning sync indefinitely against that node. With higher concurrency, pre-scheduled requests still complete and silently introduce gaps (positions 2048..8191 of each window are never re-fetched). The PR is marked `feat!`, so requiring matched versions is a defensible posture, but the truncation mode is silent and proof-verified — wallets cannot detect it. Treat a response as 'end of tree' only when `notes.len() < mmr_chunk_size`, since the server always returns whole-chunk multiples on the in-tree path, and bound forward scheduling by an explicit end-of-tree signal rather than 'shorter than what I asked for'.
```
</details>

PR #3758, opened 2026-05-27
feat(rs-platform-wallet-ffi): expose devnet name and LLMQ_DEVNET override in spv_start

QuantumExplorer:
SPV-syncing a Dash devnet from the platform wallet currently doesn't work — the FFI's `platform_wallet_manager_spv_start` has no way to pass the two devnet-specific knobs that rust-dashcore PR [dashpay/rust-dashcore#784](https://github.com/dashpay/rust-dashcore/pull/784) recently exposed on `ClientConfig` (already pinned at `58d61ea` in our `Cargo.toml`):

1. **Devnet user-agent substring.** Dash Core peers running a devnet gate inbound handshakes on a `devnet.devnet-<name>` substring in the peer's user agent (`net_processing.cpp:3957`). Without it, the SPV client cannot connect to any devnet peer.
2. **`LLMQ_DEVNET` size/threshold override.** Mirrors Dash Core's `-llmqdevnetparams=<size>:<threshold>`. Without matching params, the SPV client reconstructs the wrong quorum and ChainLock / legacy InstantSend signatures fail to verify.

This PR wires both through the FFI and the Swift wrapper as pure plumbing.

## What was done?

**Rust FFI (`rs-platform-wallet-ffi`)** — `platform_wallet_manager_spv_start` gains three trailing params:
- `devnet_name: *const c_char` — `NULL` means "not devnet".
- `llmq_devnet_size: u32`, `llmq_devnet_threshold: u32` — both `0` means "no override".

Validations (returned as `ErrorInvalidParameter`):
- `devnet_name` must be set iff `network == Devnet`.
- `llmq_devnet_size` and `llmq_devnet_threshold` must both be set or both zero.
- Non-zero LLMQ params are rejected on non-devnet networks (matches `ClientConfig::validate()` wording for consistency).

When `devnet_name` is supplied, the user agent is rebuilt as `/{base}(devnet.devnet-{name})/`, falling back to a `platform-wallet-ffi:<crate version>` base when the caller didn't pass one — mirrors the format the `dash-spv` binary constructs in `main.rs`. The LLMQ override is written into `ClientConfig.llmq_devnet_params`; `DashSpvClient::new` applies it via `apply_global_overrides` (idempotent for identical values).

**Swift wrapper (`SwiftDashSDK`)** — `PlatformSpvStartConfig` gains `devnetName: String?`, `llmqDevnetSize: UInt32`, `llmqDevnetThreshold: UInt32`, all defaulted so existing call sites (e.g. `CoreContentView.swift`) compile unchanged. `startSpv(config:)` marshals `devnetName` to a C string with the same `strdup`/`defer free` pattern as `userAgent`, then passes the new trailing args to the FFI. Pure marshalling — no decisions on the Swift side (complies with `packages/swift-sdk/CLAUDE.md`).

Diff stat: 2 files changed, +93 / -3.

## How Has This Been Tested?
- `cargo check -p platform-wallet-ffi` — clean.
- `cargo fmt --all -- --check` — clean.
- `cd packages/swift-sdk && ./build_ios.sh --target sim` — succeeded end-to-end (regenerated the C header inside `DashSDKFFI.xcframework`, then ran `xcodebuild build` on `SwiftExampleApp` with `-warnings-as-errors`; reports `Swift/Xcode build succeeded`).

The regenerated header inside the xcframework carries the new signature:
```c
struct PlatformWalletFFIResult platform_wallet_manager_spv_start(Handle handle, const char *data_dir, FFINetwork network, const char *user_agent, const char *const *peers, uintptr_t peer_count, bool restrict_to_configured_peers, uint32_t start_from_height, const char *devnet_name, uint32_t llmq_devnet_size, uint32_t llmq_devnet_threshold);
```

## Breaking Changes
None at the Swift level — every new field on `PlatformSpvStartConfig` is defaulted (`devnetName: nil`, `llmqDevnetSize: 0`, `llmqDevnetThreshold: 0`), so existing call sites compile unchanged.

The Rust `extern "C"` signature does gain three trailing params, which is a C-ABI source-level break for any out-of-tree caller calling `platform_wallet_manager_spv_start` directly. Per `feedback_breaking_change_scope`, `feat!:` is reserved for consensus-breaking changes in this repo, so this lands under plain `feat:`.

## Checklist:

## Release Notes

* **New Features**
  * SPV wallet manager now supports devnet network configuration with custom devnet naming capabilities
  * Added customizable LLMQ (Quorum) parameters for devnet environments, including size and threshold override options

[Review Change Stack](https://app.coderabbit.ai/change-stack/dashpay/platform/pull/3758?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/rs-platform-wallet-ffi/src/spv.rs
+++ b/packages/rs-platform-wallet-ffi/src/spv.rs
@@ -3,6 +3,7 @@
 use std::ffi::CStr;
 use std::os::raw::c_char;
 
+use dashcore::sml::llmq_type::LlmqDevnetParams;
 use platform_wallet::spv::{ClientConfig, ProgressPercentage, SyncProgress, SyncState};
 
 use crate::error::*;
@@ -197,6 +198,17 @@ pub unsafe extern "C" fn platform_wallet_manager_spv_tip_unix_seconds(
 /// Hardcoded to `true` here; if a future caller has a real reason to
 /// run SPV without masternode sync, it can construct the wallet
 /// manager without the asset-lock path instead.
+///
+/// Devnet support:
+/// - `devnet_name` must be set (non-null) when `network == Devnet` and
+///   must be null on every other network. When set, the SPV user agent
+///   is rewritten to embed the `devnet.devnet-<name>` substring Dash
+///   Core peers gate inbound devnet handshakes on (matches the format
+///   the `dash-spv` binary uses).
+/// - `llmq_devnet_size` / `llmq_devnet_threshold` mirror Dash Core's
+///   `-llmqdevnetparams=<size>:<threshold>`. Both zero means "no
+///   override". Setting one without the other, or setting them on a
+///   non-devnet network, is rejected.
 #[no_mangle]
 #[allow(clippy::field_reassign_with_default)]
 pub unsafe extern "C" fn platform_wallet_manager_spv_start(
@@ -208,16 +220,62 @@ pub unsafe extern "C" fn platform_wallet_manager_spv_start(
     peer_count: usize,
     restrict_to_configured_peers: bool,
     start_from_height: u32,
+    devnet_name: *const c_char,
+    llmq_devnet_size: u32,
+    llmq_devnet_threshold: u32,
 ) -> PlatformWalletFFIResult {
     check_ptr!(data_dir);
     let data_dir_str = unwrap_result_or_return!(CStr::from_ptr(data_dir).to_str()).to_string();
-    let user_agent_str = if user_agent.is_null() {
+    let mut user_agent_str = if user_agent.is_null() {
         None
     } else {
         Some(unwrap_result_or_return!(CStr::from_ptr(user_agent).to_str()).to_string())
     };
 
     let net: crate::types::Network = network.into();
+
+    let devnet_name_str = if devnet_name.is_null() {
+        None
+    } else {
+        Some(unwrap_result_or_return!(CStr::from_ptr(devnet_name).to_str()).to_string())
+    };
+
+    if net == crate::types::Network::Devnet && devnet_name_str.is_none() {
+        return PlatformWalletFFIResult::err(
+            PlatformWalletFFIResultCode::ErrorInvalidParameter,
+            "devnet_name is required when network=Devnet",
+        );
+    }
+    if net != crate::types::Network::Devnet && devnet_name_str.is_some() {
+        return PlatformWalletFFIResult::err(
+            PlatformWalletFFIResultCode::ErrorInvalidParameter,
+            "devnet_name is only valid on devnet",
+        );
+    }
+    if (llmq_devnet_size > 0) ^ (llmq_devnet_threshold > 0) {
+        return PlatformWalletFFIResult::err(
+            PlatformWalletFFIResultCode::ErrorInvalidParameter,
+            "llmq_devnet_size and llmq_devnet_threshold must both be set or both zero",
+        );
+    }
+    if llmq_devnet_size > 0 && net != crate::types::Network::Devnet {
+        return PlatformWalletFFIResult::err(
+            PlatformWalletFFIResultCode::ErrorInvalidParameter,
+            "llmq_devnet_params is only valid on devnet",
+        );
+    }
+
+    // Dash Core devnet peers disconnect any inbound connection whose
+    // user agent doesn't contain `devnet.devnet-<name>`. Rebuild the
+    // user agent to embed the substring while keeping whatever base
+    // the caller (or our default) provided. Mirrors the dash-spv
+    // binary's `/rust-dash-spv:VERSION(devnet.devnet-NAME)/` format.
+    if let Some(name) = devnet_name_str.as_deref() {
+        let base = user_agent_str
+            .clone()
+            .unwrap_or_else(|| format!("platform-wallet-ffi:{}", env!("CARGO_PKG_VERSION")));
+        user_agent_str = Some(format!("/{base}(devnet.devnet-{name})/"));
+    }
 
     let mut peer_list: Vec<String> = Vec::new();
     if !peers.is_null() && peer_count > 0 {
@@ -255,6 +313,12 @@ pub unsafe extern "C" fn platform_wallet_manager_spv_start(
             if let Ok(addr) = p.parse() {
                 config.peers.push(addr);
             }
+        }
+        if llmq_devnet_size > 0 {
+            config.llmq_devnet_params = Some(LlmqDevnetParams {
+                size: llmq_devnet_size,
+                threshold: llmq_devnet_threshold,
+            });
         }
 
         let _guard = runtime().enter();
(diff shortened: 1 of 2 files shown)


QuantumExplorer (commented):
Reviewed

packages/rs-platform-wallet-ffi/src/spv.rs
```diff
@@ -208,17 +220,63 @@ pub unsafe extern "C" fn platform_wallet_manager_spv_start(
...
+    } else {
+        Some(unwrap_result_or_return!(CStr::from_ptr(devnet_name).to_str()).to_string())
+    };
+
+    if net == crate::types::Network::Devnet && devnet_name_str.is_none() {
+        return PlatformWalletFFIResult::err(
+            PlatformWalletFFIResultCode::ErrorInvalidParameter,
+            "devnet_name is required when network=Devnet",
+        );
+    }
```
thepastaclaw, line 248: **🟡 Suggestion: Empty `devnet_name` passes the required-on-devnet check**

The check on line 243 only rejects `NULL`. A caller passing an empty (but non-null) C string for `devnet_name` while `network == Devnet` satisfies validation, and the user agent is then rebuilt as `/{base}(devnet.devnet-)/`. That trailing `devnet.devnet-` substring will not match any real devnet's gating prefix, so the SPV client returns `ok()` but silently fails the handshake — precisely the failure mode this PR exists to prevent. Because the Swift wrapper takes `devnetName: String?` (not a validated enum), the FFI is the right place to enforce non-empty.

```suggestion
    let devnet_name_str = if devnet_name.is_null() {
        None
    } else {
        Some(unwrap_result_or_return!(CStr::from_ptr(devnet_name).to_str()).to_string())
    };

    if net == crate::types::Network::Devnet
        && devnet_name_str.as_deref().map_or(true, str::is_empty)
    {
        return PlatformWalletFFIResult::err(
            PlatformWalletFFIResultCode::ErrorInvalidParameter,
            "devnet_name is required (non-empty) when network=Devnet",
        );
    }
    if net != crate::types::Network::Devnet && devnet_name_str.is_some() {
        return PlatformWalletFFIResult::err(
            PlatformWalletFFIResultCode::ErrorInvalidParameter,
            "devnet_name is only valid on devnet",
        );
    }
```

packages/rs-platform-wallet-ffi/src/spv.rs
```diff
@@ -208,17 +220,63 @@ pub unsafe extern "C" fn platform_wallet_manager_spv_start(
...
+    // user agent doesn't contain `devnet.devnet-<name>`. Rebuild the
+    // user agent to embed the substring while keeping whatever base
+    // the caller (or our default) provided. Mirrors the dash-spv
+    // binary's `/rust-dash-spv:VERSION(devnet.devnet-NAME)/` format.
+    if let Some(name) = devnet_name_str.as_deref() {
+        let base = user_agent_str
+            .clone()
+            .unwrap_or_else(|| format!("platform-wallet-ffi:{}", env!("CARGO_PKG_VERSION")));
+        user_agent_str = Some(format!("/{base}(devnet.devnet-{name})/"));
+    }
```
thepastaclaw, line 278: **💬 Nitpick: User-agent rebuild double-wraps an already-BIP14-formatted base**

If the caller supplies a base already in BIP14 form (e.g. `"/myapp:1.0/"`), the rewrite emits `"//myapp:1.0/(devnet.devnet-name)/"` — extra slashes that deviate from the `/rust-dash-spv:VERSION(devnet.devnet-NAME)/` format the comment claims to mirror. Dash Core's substring filter still matches, so this is cosmetic, but normalizing the base avoids surprising peer-log fingerprints.

```suggestion
    if let Some(name) = devnet_name_str.as_deref() {
        let base = user_agent_str
            .clone()
            .unwrap_or_else(|| format!("platform-wallet-ffi:{}", env!("CARGO_PKG_VERSION")));
        let base = base.trim_matches('/');
        user_agent_str = Some(format!("/{base}(devnet.devnet-{name})/"));
    }
```

<sub>source: ['claude']</sub>

packages/rs-platform-wallet-ffi/src/spv.rs
```diff
@@ -208,17 +220,63 @@ pub unsafe extern "C" fn platform_wallet_manager_spv_start(
...
+            PlatformWalletFFIResultCode::ErrorInvalidParameter,
+            "llmq_devnet_size and llmq_devnet_threshold must both be set or both zero",
+        );
+    }
+    if llmq_devnet_size > 0 && net != crate::types::Network::Devnet {
+        return PlatformWalletFFIResult::err(
+            PlatformWalletFFIResultCode::ErrorInvalidParameter,
+            "llmq_devnet_params is only valid on devnet",
+        );
+    }
```
thepastaclaw, line 266: **💬 Nitpick: No `threshold <= size` check at the FFI boundary**

The XOR check ensures both LLMQ params are zero or both nonzero, but accepts `size=3, threshold=10`. Downstream `ClientConfig::validate()` is expected to catch it, but enforcing at the FFI surface produces a direct `ErrorInvalidParameter` instead of a less direct downstream failure. Low priority — confirm whether `spawn_in_background` actually invokes `validate()` before applying the config; if it does, this is purely cosmetic.

```suggestion
    if (llmq_devnet_size > 0) ^ (llmq_devnet_threshold > 0) {
        return PlatformWalletFFIResult::err(
            PlatformWalletFFIResultCode::ErrorInvalidParameter,
            "llmq_devnet_size and llmq_devnet_threshold must both be set or both zero",
        );
    }
    if llmq_devnet_threshold > llmq_devnet_size {
        return PlatformWalletFFIResult::err(
            PlatformWalletFFIResultCode::ErrorInvalidParameter,
            "llmq_devnet_threshold must be <= llmq_devnet_size",
        );
    }
    if llmq_devnet_size > 0 && net != crate::types::Network::Devnet {
        return PlatformWalletFFIResult::err(
            PlatformWalletFFIResultCode::ErrorInvalidParameter,
            "llmq_devnet_params is only valid on devnet",
        );
    }
```

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Pure FFI plumbing for two devnet knobs (devnet_name and LLMQ_DEVNET size/threshold). The validation matrix is sensible and mirrors the documented contract. Three small input-edge-cases are worth tightening at the boundary; nothing blocks merge. The codex 'blocking' truncation claim could not be verified against the pinned dash-spv revision (`58d61ea`) and is dropped.

<details>

```

- [SUGGESTION] packages/rs-platform-wallet-ffi/src/spv.rs:237-248: Empty `devnet_name` passes the required-on-devnet check
  The check on line 243 only rejects `NULL`. A caller passing an empty (but non-null) C string for `devnet_name` while `network == Devnet` satisfies validation, and the user agent is then rebuilt as `/{base}(devnet.devnet-)/`. That trailing `devnet.devnet-` substring will not match any real devnet's gating prefix, so the SPV client returns `ok()` but silently fails the handshake — precisely the failure mode this PR exists to prevent. Because the Swift wrapper takes `devnetName: String?` (not a validated enum), the FFI is the right place to enforce non-empty.
```
</details>

Issue #788, opened 2023-02-17
feat(dashmate): refactor reindex behaviour


pshenmic:
## Expected Behavior

Reindex command should not change the config and ask user for additional command executions. Instead, command arguments should be used to turn local dashcore container into reindex mode. You don't actually need to edit any dashmate or dashcore configs.

## Current Behavior

Reindex command actually changes dashmate config and re-renders all the configs.

## Possible Solution

As a more redundant solution, we can use a normal running node to the reindex process and just keep it running when reindex is finished. It will keep Dashmate architecture simpler and we won't need to keep reindex state.

Two commands should be updated:

### Reindex command
1. if the Core service is started we need to inform a user that during reindexing the node won't operate for some time. Core should be stopped.
3. Write `reindex=1` into the Core config
4. Start the Core service
5. When Core service is started we update the Core config again and remove reindex flag
6. If `--detach` flag is not passed, we show the progress and estimation in reindex command output. The command should exit when reindex is done. We also should warn users that they can exit from the command (CTRL - C) and the process will continue in the background. In both cases, we should tell users that can monitor the process with `dashmate status core` command.

### Dashmate status
Add sync/reindex estimation to `dashmate status` and `dashmate status core`
Possibly to detect that reindex is required and show an error with suggestion to run reindex command

## Alternatives Considered

Build a statfull soltuion with background reindexing but it will overcomplectate the system and require much more work which doesn't give much benefits

pshenmic:
https://github.com/dashpay/platform/pull/1212

PR #781, opened 2023-02-16
feat(dasmate): pack release script

shuplenkov:
## Issue being fixed or feature implemented

Dashmate must be distributed to users as self- installable package. User must be able to use dashmate without installing NodeJS and other stuff on their computers.

## What was done?

Bash script which packs dashmate into the installable package. This is achieved by using `oclif pack` command.

## How Has This Been Tested?

Manual

## Breaking Changes

No

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added or updated relevant unit/integration/functional/e2e tests
- [x] I have made corresponding changes to the documentation

**For repository code-owners and collaborators only**
- [x] I have assigned this pull request to a milestone

PR #3761, opened 2026-05-28
feat(swift-example-app): enable SPV on devnet

shumkov:

The example app could already create an SDK against a devnet, but `Start Sync` would either fail or connect to the wrong peers because the SPV start path never told `dash-spv` it was running against a devnet. This PR enables SPV on devnet end-to-end.

## What was done?

**Example app — devnet identity is now part of the SPV start config**
- `OptionsView`: new \"Devnet Name\" text field next to the existing Quorum URL field, backed by `platformDevnetName` UserDefaults.
- `CoreContentView.startSync`: when `currentNetwork == .devnet`, reads `platformDevnetName` and passes it to `PlatformSpvStartConfig.devnetName`. Without this the FFI rejects the start with \"devnet_name is required when network=Devnet\".

**FFI — migrated to the new `dash-spv::DevnetConfig` surface**
- `packages/rs-platform-wallet-ffi/src/spv.rs`: replaced the removed `ClientConfig::llmq_devnet_params` field with `ClientConfig.devnet = Some(DevnetConfig::new(name).with_llmq_params(...))`.
- `packages/rs-platform-wallet/src/wallet/platform_wallet_traits.rs`: implemented the new `WalletInfoInterface::monitored_script_pubkeys` required method by delegating to the inner `ManagedWalletInfo`.
- `packages/rs-platform-wallet/src/spv/mod.rs`: re-exported `DevnetConfig` from `dash_spv` so the FFI layer can build it without depending on `dash-spv` directly.

**Dependency bump (the enabler)**
- Bumped `rust-dashcore` pin `58d61ea5` → `eb889af1` (current dev HEAD), picking up:
  - [`eb889af1`](https://github.com/dashpay/rust-dashcore/commit/eb889af13f667ed39c35e8e8a0830eeedf523476) feat(dash-spv): consolidate devnet config into `DevnetConfig` struct (#788)
  - [`9655ea38`](https://github.com/dashpay/rust-dashcore/commit/9655ea38) refactor: avoid rebuilding `script_pubkey`s for BIP158 filter matching (#781)

## How Has This Been Tested?

- `cargo check -p platform-wallet -p platform-wallet-ffi` against the bumped revision is clean.
- Manual UAT to follow: enter `platformQuorumURL` + `platformDevnetName` in Options on devnet, hit \"Start Sync\" on the Wallet tab, confirm SPV reaches the masternodes auto-discovered from `{quorumURL}/masternodes` and that headers / filters / masternodes progress all advance.

## Breaking Changes

None for downstream consumers. The Swift `PlatformSpvStartConfig.devnetName` parameter was already public (added in #3758); this PR just starts using it.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a persistent "Devnet Name" setting and UI field for configuring devnet identity.
  * SPV startup now accepts and validates an optional devnet name, enabling devnet-specific behavior and user-agent embedding on next sync.

* **Bug Fixes**
  * Wallet info now exposes monitored script pubkeys for improved wallet visibility.

* **Chores**
  * Updated internal dependency revisions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dashpay/platform/pull/3761?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/rs-platform-wallet-ffi/src/spv.rs
+++ b/packages/rs-platform-wallet-ffi/src/spv.rs
@@ -254,6 +254,26 @@ pub unsafe extern "C" fn platform_wallet_manager_spv_start(
             "devnet_name is only valid on devnet",
         );
     }
+    // Reject empty or `/`-containing names synchronously here rather
+    // than letting `DevnetConfig::validate` surface them asynchronously
+    // from `spawn_in_background`. Mirrors `DevnetConfig::validate` and
+    // saves callers without a pre-filter (other language bindings,
+    // integration tests) from a `(devnet.devnet-)` user agent that
+    // Dash Core peers silently drop.
+    if let Some(name) = devnet_name_str.as_deref() {
+        if name.is_empty() {
+            return PlatformWalletFFIResult::err(
+                PlatformWalletFFIResultCode::ErrorInvalidParameter,
+                "devnet_name must not be empty",
+            );
+        }
+        if name.contains('/') {
+            return PlatformWalletFFIResult::err(
+                PlatformWalletFFIResultCode::ErrorInvalidParameter,
+                "devnet_name must not contain '/'",
+            );
+        }
+    }
     if (llmq_devnet_size > 0) ^ (llmq_devnet_threshold > 0) {
         return PlatformWalletFFIResult::err(
             PlatformWalletFFIResultCode::ErrorInvalidParameter,
--- a/packages/rs-platform-wallet-ffi/src/spv.rs
+++ b/packages/rs-platform-wallet-ffi/src/spv.rs
@@ -4,7 +4,9 @@ use std::ffi::CStr;
 use std::os::raw::c_char;
 
 use dashcore::sml::llmq_type::LlmqDevnetParams;
-use platform_wallet::spv::{ClientConfig, ProgressPercentage, SyncProgress, SyncState};
+use platform_wallet::spv::{
+    ClientConfig, DevnetConfig, ProgressPercentage, SyncProgress, SyncState,
+};
 
 use crate::error::*;
 use crate::handle::*;
@@ -314,11 +316,15 @@ pub unsafe extern "C" fn platform_wallet_manager_spv_start(
                 config.peers.push(addr);
             }
         }
-        if llmq_devnet_size > 0 {
-            config.llmq_devnet_params = Some(LlmqDevnetParams {
-                size: llmq_devnet_size,
-                threshold: llmq_devnet_threshold,
-            });
+        if let Some(name) = devnet_name_str.as_deref() {
+            let mut devnet = DevnetConfig::new(name);
+            if llmq_devnet_size > 0 {
+                devnet = devnet.with_llmq_params(LlmqDevnetParams {
+                    size: llmq_devnet_size,
+                    threshold: llmq_devnet_threshold,
+                });
+            }
+            config.devnet = Some(devnet);
         }
 
         let _guard = runtime().enter();
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1132,7 +1132,7 @@ version = "3.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "faf9468729b8cbcea668e36183cb69d317348c2e08e994829fb56ebfdfbaac34"
 dependencies = [
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -1550,7 +1550,7 @@ dependencies = [
 [[package]]
 name = "dash-network"
 version = "0.43.0"
-source = "git+https://github.com/dashpay/rust-dashcore?rev=58d61ea50ac4c217c8e1350b57363b351a0b99d4#58d61ea50ac4c217c8e1350b57363b351a0b99d4"
+source = "git+https://github.com/dashpay/rust-dashcore?rev=eb889af13f667ed39c35e8e8a0830eeedf523476#eb889af13f667ed39c35e8e8a0830eeedf523476"
 dependencies = [
  "bincode",
  "bincode_derive",
@@ -1561,7 +1561,7 @@ dependencies = [
 [[package]]
 name = "dash-network-seeds"
 version = "0.43.0"
-source = "git+https://github.com/dashpay/rust-dashcore?rev=58d61ea50ac4c217c8e1350b57363b351a0b99d4#58d61ea50ac4c217c8e1350b57363b351a0b99d4"
+source = "git+https://github.com/dashpay/rust-dashcore?rev=eb889af13f667ed39c35e8e8a0830eeedf523476#eb889af13f667ed39c35e8e8a0830eeedf523476"
 dependencies = [
  "dash-network",
 ]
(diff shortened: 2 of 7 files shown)


thepastaclaw (changes requested):
## Code Review

Verified codex-general's blocker against upstream source at eb889af1 and against `cargo check`. The workspace does not compile against the pinned revision: dash-spv's `ClientConfig` no longer exposes `llmq_devnet_params` (replaced by `devnet: Option<DevnetConfig>`), and key-wallet's `WalletInfoInterface` trait gained a required `monitored_script_pubkeys` method that `PlatformWalletInfo` does not implement. The PR cannot land as-is without follow-up callsite updates. Both Claude agents missed these regressions.

🔴 2 blocking

<details>

```

- [BLOCKING] packages/rs-platform-wallet-ffi/src/spv.rs:317-322: `platform-wallet-ffi` references `ClientConfig::llmq_devnet_params` removed at eb889af1
  Verified against the upstream source at rev `eb889af1` (cargo git checkout `rust-dashcore-c6b13647c01f74b9/eb889af`): `dash_spv::ClientConfig` no longer exposes a `llmq_devnet_params` field. The devnet surface was consolidated into a single `pub devnet: Option<DevnetConfig>` field (see `dash-spv/src/client/config.rs:25` and `:76`, with `with_devnet(DevnetConfig)` as the builder API at `:191`). This file still does `config.llmq_devnet_params = Some(LlmqDevnetParams { size, threshold })`, which fails to compile against the pinned revision and will block the entire workspace build (and any transitive consumer such as the Swift SDK). The import of `dashcore::sml::llmq_type::LlmqDevnetParams` at line 6 is now also dead code at this callsite. Callsite needs to be migrated to construct a `DevnetConfig` (e.g. `DevnetConfig::new(name).with_llmq_params(LlmqDevnetParams { size, threshold })`) and assign it via `config.devnet = Some(...)` / `with_devnet(...)`. Note: the codex agent flagged this; the Claude agents missed it.

In `packages/rs-platform-wallet/src/wallet/platform_wallet_traits.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/wallet/platform_wallet_traits.rs:33: `PlatformWalletInfo` is missing `monitored_script_pubkeys` required by updated `WalletInfoInterface`
  Confirmed by running `cargo check -p platform-wallet-ffi` against the pinned revision: `error[E0046]: not all trait items implemented, missing: 'monitored_script_pubkeys'`. At `eb889af1`, `key_wallet::wallet::managed_wallet_info::wallet_info_interface::WalletInfoInterface` adds a required method `fn monitored_script_pubkeys(&self) -> Vec<ScriptBuf>` (see `key-wallet/src/wallet/managed_wallet_info/wallet_info_interface.rs:87`). `PlatformWalletInfo` only implements the older surface, so `platform-wallet` fails to compile — which prevents reaching the `llmq_devnet_params` error above. The natural fix is to delegate to the inner `ManagedWalletInfo` (which already implements it at `:310`): `fn monitored_script_pubkeys(&self) -> Vec<ScriptBuf> { self.core_wallet.monitored_script_pubkeys() }`. Treating as a co-symptom of the same revision bump.
```
</details>

thepastaclaw (commented):
## Code Review

Cumulative incremental review at head 58117611. Both prior blocking findings from the 056ab3ec review are FIXED at this head: (1) `platform-wallet-ffi` now constructs `DevnetConfig::new(name).with_llmq_params(...)` and assigns it via `config.devnet = Some(devnet)` at spv.rs:319-328, replacing the removed `ClientConfig::llmq_devnet_params`; (2) `PlatformWalletInfo` now implements the required `monitored_script_pubkeys` by delegating to `core_wallet` at platform_wallet_traits.rs:100-102. The new latest-delta surface (Swift `platformDevnetName` UserDefaults → `PlatformSpvStartConfig.devnetName` → FFI `devnet_name` C string) is wired end-to-end and the FFI already enforces network/devnet pairing and matched llmq params. One in-scope hardening suggestion remains: the FFI accepts empty (`""`) and BIP14-reserved-character devnet names from C callers, which would produce a malformed `(devnet.devnet-)` user agent. The Swift example app filters empty strings to nil before calling, so this PR's immediate path is unaffected; the public C ABI is the chokepoint where future callers (other language bindings, integration tests) cannot pre-filter.

🟡 1 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet-ffi/src/spv.rs:239-256: FFI accepts empty/whitespace `devnet_name` and produces malformed devnet user agent
  `platform_wallet_manager_spv_start` validates that `devnet_name` is non-null and that network/llmq pairing is consistent, but it does not reject an empty C string. A caller passing a valid pointer to `""` (or whitespace-only) enters the devnet branch, constructs `DevnetConfig::new("")`, and at line 279 rewrites the user agent to `/<base>(devnet.devnet-)/` — a substring Dash Core devnet peers will not accept, which is the exact failure mode this PR aims to eliminate. The Swift caller in `CoreContentView.startSync` happens to trim and convert empty to `nil` (CoreContentView.swift:591-596), so the example-app path is safe, but this C ABI is the only chokepoint for other consumers (future bindings, Rust integration tests, the `dash-spv` binary equivalent). Reject empty names at the same validation step that rejects null/wrong-network so the failure is synchronous on the FFI return rather than asynchronous after `spawn_in_background`. Optionally also reject BIP14-reserved characters (`/`, `(`, `)`) since they would similarly corrupt the user-agent string.
```
</details>

packages/rs-platform-wallet-ffi/src/spv.rs
```diff
@@ -252,6 +254,26 @@ pub unsafe extern "C" fn platform_wallet_manager_spv_start(
...
         );
     }
+    // Reject empty or `/`-containing names synchronously here rather
+    // than letting `DevnetConfig::validate` surface them asynchronously
+    // from `spawn_in_background`. Mirrors `DevnetConfig::validate` and
+    // saves callers without a pre-filter (other language bindings,
+    // integration tests) from a `(devnet.devnet-)` user agent that
+    // Dash Core peers silently drop.
+    if let Some(name) = devnet_name_str.as_deref() {
+        if name.is_empty() {
```
thepastaclaw, line 264: **🟡 Suggestion: FFI still accepts whitespace-only `devnet_name` values and produces a malformed devnet identity**

This change only rejects `name.is_empty()` and `name.contains('/')`. A whitespace-only C string such as `"   "` survives `CStr::to_str()?.to_string()`, is not empty, contains no slash, and therefore passes validation. It is then appended into the devnet user agent and passed to `DevnetConfig::new(name)`, so the prior finding's explicit whitespace-only case is still reproducible at the exported FFI surface even though the empty-string case is fixed. Trim before validating, or treat `name.trim().is_empty()` as invalid so whitespace-only input is rejected synchronously like the prior finding requested.

<sub>source: unknown</sub>

thepastaclaw (commented):
## Code Review

The prior FFI validation finding was only partially fixed at head 9036567073561ff1e303e7d83538cfe5efb90ed3. Empty-string `devnet_name` is now rejected, but the explicit whitespace-only half of the prior finding remains valid because the new check only uses `name.is_empty()` and does not trim before building the devnet user agent and `DevnetConfig`.

🟡 1 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet-ffi/src/spv.rs:264: FFI still accepts whitespace-only `devnet_name` values and produces a malformed devnet identity
  This change only rejects `name.is_empty()` and `name.contains('/')`. A whitespace-only C string such as `"   "` survives `CStr::to_str()?.to_string()`, is not empty, contains no slash, and therefore passes validation. It is then appended into the devnet user agent and passed to `DevnetConfig::new(name)`, so the prior finding's explicit whitespace-only case is still reproducible at the exported FFI surface even though the empty-string case is fixed. Trim before validating, or treat `name.trim().is_empty()` as invalid so whitespace-only input is rejected synchronously like the prior finding requested.
```
</details>

shumkov:
Superseded by #3763 — the dashcore bump + Rust API migrations identical to my first two commits already landed via #3762, so the unique value of this PR (Swift example-app devnet wiring + FFI hardening with whitespace trim) was carved out and reopened as a clean PR on top of v3.1-dev.

PR #3764, opened 2026-05-28
feat(dashmate): configure docker build args via config

shumkov:

Adds a `dockerBuild.buildArgs` map to the dashmate config schema so per-image Docker build args (e.g. `CARGO_BUILD_PROFILE`, `SDK_TEST_DATA`) can be pinned **per config** and forwarded to the image build. Previously the only way to set these at build time was to export shell envs on every invocation or hand-edit the compose YAML — neither survives `yarn setup` cleanly.

## What was done?

- **Schema** (`configJsonSchema.js`): `dockerBuild.buildArgs` is `{ [string]: string }`, with keys constrained to env-var identifiers (`^[A-Za-z_][A-Za-z0-9_]*$`).
- **Config helpers** (`Config.js`, `commands/config/set.js`): schema-aware path validation (`Config.isSchemaPathAllowed`) so map-shaped paths like `platform.drive.abci.docker.build.buildArgs.SDK_TEST_DATA` are accepted — the old existence-based pre-check rejected any key that didn't exist yet. Paths with empty segments are rejected.
- **Forwarding via dynamic-compose** (`templates/dynamic-compose.yml.dot`): each config's `buildArgs` render directly into that config's generated `dynamic-compose.yml` as `build.args` entries, with values `JSON.stringify`-escaped so a quote/backslash/newline stays valid YAML. **Dashmate config is the single source of truth — there is no shell-env passthrough.**
- **Default config + migration**: `buildArgs: {}` placeholder on both the `drive.abci` and `dapi.rsDapi` build blocks; a 3.1.0 migration backfills the field on pre-existing configs.

Yarn-script side (`setup_local_network.sh`): pins `SDK_TEST_DATA="true"` per `local_N` (the genesis test-data flag). `CARGO_BUILD_PROFILE` is intentionally **not** pinned — operators set it per config when they want a release bake:

```
dashmate config set --config=local_1 platform.drive.abci.docker.build.buildArgs.CARGO_BUILD_PROFILE '"release"'
```

Release is strongly recommended when `SDK_TEST_DATA` is on: the genesis shielded-pool seeding runs at **InitChain (runtime)**, and a release-compiled binary runs that Sinsemilla work ~10× faster than a debug build.

> ⚠️ Build args are sourced **only** from dashmate config. A shell-only `CARGO_BUILD_PROFILE=release yarn start` is **not** forwarded to the build — persist it with `dashmate config set …buildArgs.CARGO_BUILD_PROFILE` instead.

## How Has This Been Tested?

- **`Config.spec.js`** — unit tests for schema-aware path traversal (typed `properties`, `additionalProperties` maps, `$ref` descent, leaf-primitive and empty-segment rejection) plus a regression test for the original `buildArgs.SDK_TEST_DATA` failure.
- **`dynamicComposeBuildArgs.spec.js`** — renders the template and asserts `build.args` are emitted per service, omitted when `buildArgs` is empty, and YAML-escaped (a js-yaml round-trip on a value containing a quote, backslash and newline).
- **Manual end-to-end**: `yarn setup local --node-count=3 …`, then a release bake via `dashmate config set …CARGO_BUILD_PROFILE '"release"'`, produces a release `drive:local` image with SDK_TEST_DATA seeding active.

## Breaking Changes

None — the schema change is additive. Existing dashmate configs without `buildArgs` are backfilled to an empty map by the 3.1.0 migration and continue to work unchanged.

## Checklist:


thepastaclaw (commented):
## Code Review

PR adds a generic `buildArgs` map to dashmate's dockerBuild schema and wires it through to docker compose, plus a `set.js` fix to support map-shaped (additionalProperties) schema paths. The implementation is sound — coerceTypes in AJV correctly handles boolean→string coercion for `config set ... true` flows, and tests for `isSchemaPathAllowed` are thorough. The remaining issues are documentation inconsistencies introduced by this PR: contradictory claims about shell-env precedence and an invalid config-path reference.

<details>

```

In `docs/shielded-seeder-performance.md`:
- [SUGGESTION] docs/shielded-seeder-performance.md:44: Doc references a config path that does not exist
  Line 44 says the release profile is plumbed through dashmate as `platform.drive.abci.docker.build.cargoBuildProfile`, but the schema in `configJsonSchema.js` only adds a generic `buildArgs` map — there is no `cargoBuildProfile` field. The same document uses the correct path (`buildArgs.CARGO_BUILD_PROFILE`) later on lines 113 and 121. An operator following the line-44 instruction will hit `InvalidOptionPathError` or set an inert field.

In `packages/dashmate/src/config/configJsonSchema.js`:
- [SUGGESTION] packages/dashmate/src/config/configJsonSchema.js:35-48: Schema comment claims shell env overrides config, but the opposite is true
  The schema comment says `Shell env overrides any value set here (so CARGO_BUILD_PROFILE=release yarn start wins per-invocation)`. The actual precedence is the reverse: `DockerCompose.js#createOptionsWithEnvs` (lines 676–679) constructs the spawn env as `{ ...process.env, ...envs }`, so the values `generateEnvsFactory` injects from `buildArgs` shadow same-named shell env vars. The implementation comment in `generateEnvsFactory.js` (lines 112–116) and the PR design (dashmate config = single source of truth) confirm dashmate wins. Shell env still flows through as a fallback for keys NOT set in `buildArgs`, but it does not override. Reconcile the schema comment with the implementation, or callers will be surprised when `CARGO_BUILD_PROFILE=release yarn start` is silently dropped.

In `packages/dashmate/src/config/generateEnvsFactory.js`:
- [SUGGESTION] packages/dashmate/src/config/generateEnvsFactory.js:112-116: Comment overstates the single-source-of-truth claim
  The comment says `do NOT fall back to process.env[key]. Operators who need a per-invocation override should yarn dashmate config set ... rather than FOO=bar yarn start.` In practice `DockerCompose.js#createOptionsWithEnvs` still spreads `process.env` first into the spawn env, so shell-provided values do reach docker-compose for any buildArg key the dashmate config has not set. The accurate semantics are: dashmate config wins on collision; shell env still works as a fallback for unset keys. Stating this precisely prevents a future contributor from removing the `process.env` spread in the name of single-source enforcement, which would silently break existing `FOO=bar yarn start` patterns.
```
</details>

docs/shielded-seeder-performance.md
```diff
@@ -0,0 +1,132 @@
...
+| 1 000 000 | dev | ~60–120 min | unusable |
+
+## Options researched, ranked by impact / cost
+
+### 1. Release build (`CARGO_BUILD_PROFILE=release`) — **recommended first step**
+
+- **Speedup:** ~20–50× for the Sinsemilla phase; ~10× overall.
+- **Effort:** one config option in dashmate (see below) + one Docker build arg.
+- **Tradeoff:** slower image builds (release optimizations take longer to compile), larger debug-info-stripped binary remains ~50–100 MB vs ~700 MB for debug.
+- **Status:** plumbed through dashmate as `platform.drive.abci.docker.build.cargoBuildProfile` — set to `release` when you want fast seeding.
```
thepastaclaw: **🟡 Suggestion: Doc references a config path that does not exist**

Line 44 says the release profile is plumbed through dashmate as `platform.drive.abci.docker.build.cargoBuildProfile`, but the schema in `configJsonSchema.js` only adds a generic `buildArgs` map — there is no `cargoBuildProfile` field. The same document uses the correct path (`buildArgs.CARGO_BUILD_PROFILE`) later on lines 113 and 121. An operator following the line-44 instruction will hit `InvalidOptionPathError` or set an inert field.

```suggestion
- **Status:** plumbed through dashmate as `platform.drive.abci.docker.build.buildArgs.CARGO_BUILD_PROFILE` — set to `release` when you want fast seeding.
```

<sub>source: ['claude']</sub>

packages/dashmate/src/config/configJsonSchema.js
```diff
@@ -32,6 +32,24 @@ export default {
...
+        // any value set here (so `CARGO_BUILD_PROFILE=release yarn start`
+        // wins per-invocation).
+        //
+        // Common keys (when meaningful for the image's target):
+        // - CARGO_BUILD_PROFILE: "dev" | "release" — Rust profile for
+        //   drive-abci / rs-dapi. Release is required for SDK_TEST_DATA
+        //   shielded seeding at N > a few thousand; see
+        //   `docs/shielded-seeder-performance.md`.
+        // - SDK_TEST_DATA: "true" — enable the SDK test-data cfg flag in
+        //   the binary at compile time.
```
thepastaclaw: **🟡 Suggestion: Schema comment claims shell env overrides config, but the opposite is true**

The schema comment says `Shell env overrides any value set here (so CARGO_BUILD_PROFILE=release yarn start wins per-invocation)`. The actual precedence is the reverse: `DockerCompose.js#createOptionsWithEnvs` (lines 676–679) constructs the spawn env as `{ ...process.env, ...envs }`, so the values `generateEnvsFactory` injects from `buildArgs` shadow same-named shell env vars. The implementation comment in `generateEnvsFactory.js` (lines 112–116) and the PR design (dashmate config = single source of truth) confirm dashmate wins. Shell env still flows through as a fallback for keys NOT set in `buildArgs`, but it does not override. Reconcile the schema comment with the implementation, or callers will be surprised when `CARGO_BUILD_PROFILE=release yarn start` is silently dropped.

packages/dashmate/src/config/generateEnvsFactory.js
```diff
@@ -103,6 +103,32 @@ export default function generateEnvsFactory(configFile, homeDir, getConfigProfil
...
+    // (`platform.drive.abci.docker.build.buildArgs`,
+    // `platform.dapi.rsDapi.docker.build.buildArgs`) as env vars under the
+    // arg name. `docker-compose.build.*.yml` reads them via `${NAME}`
+    // substitution and forwards them into the Dockerfile `ARG`.
+    //
+    // Dashmate config is the single source of truth for build args — do NOT
+    // fall back to `process.env[key]`. Operators who need a per-invocation
+    // override should `yarn dashmate config set ...` rather than `FOO=bar
+    // yarn start`. Keeping this single-source matches the `yarn setup` flow
+    // that writes SDK_TEST_DATA into the local config automatically.
```
thepastaclaw: **🟡 Suggestion: Comment overstates the single-source-of-truth claim**

The comment says `do NOT fall back to process.env[key]. Operators who need a per-invocation override should yarn dashmate config set ... rather than FOO=bar yarn start.` In practice `DockerCompose.js#createOptionsWithEnvs` still spreads `process.env` first into the spawn env, so shell-provided values do reach docker-compose for any buildArg key the dashmate config has not set. The accurate semantics are: dashmate config wins on collision; shell env still works as a fallback for unset keys. Stating this precisely prevents a future contributor from removing the `process.env` spread in the name of single-source enforcement, which would silently break existing `FOO=bar yarn start` patterns.

<sub>source: ['claude']</sub>

scripts/setup_local_network.sh
```diff
@@ -16,3 +16,18 @@ yarn run dashmate setup local --verbose \
...
 # enable insight
 yarn dashmate config set core.insight.enabled true --config local_seed
+
+# Bake SDK_TEST_DATA=true into the drive-abci docker build for each masternode
+# so the genesis shielded-pool seeder + identity/contract fixtures run on every
+# local devnet bring-up. Production / release builds explicitly do NOT set this.
+#
+# Note: the seeder runs at `docker build` time inside the Dockerfile's bake
+# stage, so a release-mode binary is highly recommended when SDK_TEST_DATA is
+# on (see `docs/shielded-seeder-performance.md`). Set it per-invocation via
```
thepastaclaw: **💬 Nitpick: Comment says seeder runs at `docker build` time, but it runs at InitChain**

The comment claims `the seeder runs at docker build time inside the Dockerfile's bake stage`. Per `docs/shielded-seeder-performance.md` (line 1) the SDK_TEST_DATA seeder runs at `InitChain` — at runtime in drive-abci, not during image build. What happens at build time is the cfg-gated compilation of the seeder code into the binary (controlled by the `SDK_TEST_DATA` build arg). The release-profile recommendation is still right, but the stated reason is wrong and will confuse anyone debugging why their image build is slow versus why their node's first block is slow.

<sub>source: ['claude']</sub>

packages/dashmate/src/config/generateEnvsFactory.js
```diff
@@ -103,6 +103,32 @@ export default function generateEnvsFactory(configFile, homeDir, getConfigProfil
...
+    const getBuildArgs = (configPath) => (
+      config.has(configPath) ? (config.get(configPath) || {}) : {}
+    );
+    const mergedBuildArgs = {
+      ...getBuildArgs('platform.dapi.rsDapi.docker.build.buildArgs'),
+      ...getBuildArgs('platform.drive.abci.docker.build.buildArgs'),
+    };
+    for (const [key, value] of Object.entries(mergedBuildArgs)) {
+      envs[key] = value;
+    }
```
thepastaclaw: **💬 Nitpick: No unit test for buildArgs forwarding / merge precedence**

`Config.spec.js` covers `isSchemaPathAllowed` thoroughly, but the new buildArgs-merge behavior in `generateEnvsFactory` has no unit coverage. A small spec — drive-abci wins on collision with rs-dapi (matches the comment), absent buildArgs does not shadow `convertObjectToEnvs` output, empty/undefined buildArgs is tolerated — would lock in the documented behavior and protect against a future regression that flipped the spread order.

<sub>source: ['claude']</sub>

packages/dashmate/templates/dynamic-compose.yml.dot
```diff
@@ -13,13 +13,23 @@ services:
...
       {{~ driveLogs :logger }}
       {{ [name, settings] = logger; }}
       - {{=settings.destination}}:/var/log/dash/drive/{{=name}}/{{=settings.destination.split('/').reverse()[0]}}
       {{~}}
+    {{?}}
+    {{? driveBuildArgs.length > 0 }}
+    build:
+      args:
+        {{~ driveBuildArgs :pair }}
+        {{=pair[0]}}: "{{=pair[1]}}"
```
thepastaclaw: **🟡 Suggestion: buildArgs values are interpolated into YAML without escaping**

Lines 30 and 58 render each pair as `{{=pair[0]}}: "{{=pair[1]}}"`. doT's `{{=...}}` is raw interpolation and the schema only constrains values to `type: 'string'` with no character restriction, so a value containing `"`, `\`, or a newline produces invalid or altered YAML that fails the next dashmate render with a confusing parse error rather than at `config set` time. The PR opens this surface as a public, open-ended `Record<string, string>` API, so worth either YAML-escaping (single-quoted form with doubled single quotes) or constraining values via a safe-character regex in the schema.

thepastaclaw (changes requested):
## Code Review

Latest delta moves buildArgs forwarding from env passthrough into the per-config `dynamic-compose.yml` template and rewrites the schema/comments accordingly. All five prior findings are resolved. One blocking regression remains: `dashmate config set ...buildArgs.* true` (used by this PR's own `scripts/setup_local_network.sh`) JSON-parses `true` into a boolean, which fails schema validation that now requires string-typed values. A secondary suggestion covers unescaped YAML interpolation of buildArgs values.

<details>

```

In `packages/dashmate/src/commands/config/set.js`:
- [BLOCKING] packages/dashmate/src/commands/config/set.js:54-60: `dashmate config set ...buildArgs.SDK_TEST_DATA true` is rejected because the CLI JSON-parses the value into a boolean
  `configJsonSchema.js` lines 40–42 now requires every `docker.build.buildArgs.*` entry to be `type: 'string'`, but `set.js` unconditionally runs `JSON.parse(optionValue)` before calling `config.set(...)`. For the inputs introduced by this PR — `scripts/setup_local_network.sh` line 24 (`... buildArgs.SDK_TEST_DATA "true"`) and the analogous `CARGO_BUILD_PROFILE release` example — the shell strips the quotes and dashmate receives the bare token `true`/`release`. `JSON.parse("true")` returns boolean `true`, `Config.set` runs AJV validation (Config.js:151), and AJV rejects the boolean against the string-typed `additionalProperties`, throwing `InvalidOptionError`. The setup script in this same PR therefore fails at line 23–24, and the documented operator workflow is broken end-to-end. Skip JSON-parsing for the open-string-map buildArgs path (or, equivalently, fall back to the raw string when the parsed value's type doesn't match the schema).

In `packages/dashmate/templates/dynamic-compose.yml.dot`:
- [SUGGESTION] packages/dashmate/templates/dynamic-compose.yml.dot:30: buildArgs values are interpolated into YAML without escaping
  Lines 30 and 58 render each pair as `{{=pair[0]}}: "{{=pair[1]}}"`. doT's `{{=...}}` is raw interpolation and the schema only constrains values to `type: 'string'` with no character restriction, so a value containing `"`, `\`, or a newline produces invalid or altered YAML that fails the next dashmate render with a confusing parse error rather than at `config set` time. The PR opens this surface as a public, open-ended `Record<string, string>` API, so worth either YAML-escaping (single-quoted form with doubled single quotes) or constraining values via a safe-character regex in the schema.
```
</details>

thepastaclaw:
✅ Review complete (commit 59cceebd)

packages/dashmate/src/config/generateEnvsFactory.js
```diff
@@ -103,39 +103,10 @@ export default function generateEnvsFactory(configFile, homeDir, getConfigProfil
...
-      // don't let buildArgs clobber reserved compose/runtime envs
-      if (reservedEnvKeys.has(key)) {
-        continue;
-      }
-      envs[key] = value;
-    }
+    // NOTE: per-image docker `build.args` (`…docker.build.buildArgs`) are NOT
+    // forwarded as env vars here. They are rendered directly into the per-config
+    // `dynamic-compose.yml` as `build.args` entries (see
+    // `templates/dynamic-compose.yml.dot`), which is the single source of truth.
```
thepastaclaw, line 109: **🟡 Suggestion: Shell-only `CARGO_BUILD_PROFILE=release yarn start` overrides no longer reach Docker builds**

This PR removes both halves of the previous shell-env passthrough: `generateEnvsFactory.js` explicitly stops forwarding `docker.build.buildArgs` into the compose environment, and `docker-compose.build.drive_abci.yml` / `docker-compose.build.rs-dapi.yml` drop their `CARGO_BUILD_PROFILE: ${CARGO_BUILD_PROFILE:-dev}` entries (confirmed in the diff against v3.1-dev). After this change, `CARGO_BUILD_PROFILE=release yarn start` is silently a no-op — the only working path is `dashmate config set …buildArgs.CARGO_BUILD_PROFILE '"release"'`. The setup script's own comment correctly tells operators to use `dashmate config set`, so the script is internally consistent, but the PR description still advertises the shell-env override. Either restore the `${CARGO_BUILD_PROFILE:-dev}` fallback in the two build compose files (so `dynamic-compose.yml` overrides win when set but the shell env still works otherwise), or update the PR description to drop the shell-env claim and point users at `dashmate config set` exclusively.

packages/dashmate/templates/dynamic-compose.yml.dot
thepastaclaw: **🟡 Suggestion: [Carried-forward — STILL VALID] buildArgs values are raw-interpolated into YAML without escaping or schema char restrictions**

Both render sites (`drive_abci` line 30, `rs_dapi` line 58) emit `{{=pair[0]}}: "{{=pair[1]}}"` with no escaping. `configJsonSchema.js` restricts keys via `propertyNames.pattern: '^[A-Za-z_][A-Za-z0-9_]*$'` but values are just `{ type: 'string' }` with no character restriction. A value containing `"`, `\`, or a newline (settable via `dashmate config set …buildArgs.FOO 'a"b'`) renders into malformed YAML and breaks the next compose invocation — there is no subsequent validation step over the generated `dynamic-compose.yml`. Operator-controlled input, so it's a robustness/UX concern rather than security. The cleanest fix is to mirror the key restriction onto values in the schema (any printable-ASCII regex that excludes `"`, `\`, and control chars easily covers realistic build args like `dev`/`release`/`true`/version/path strings). Alternatively, render values via `JSON.stringify` in the .dot template so YAML-quoting handles edge cases automatically.

packages/dashmate/configs/getConfigFileMigrationsFactory.js
```diff
@@ -1446,6 +1447,7 @@ export default function getConfigFileMigrationsFactory(homeDir, defaultConfigs)
               options.platform.dapi.rsDapi.docker.build.buildArgs = {};
             }
 
+
```
thepastaclaw: **💬 Nitpick: Stray double blank line in 3.1.0 migration block**

Two consecutive blank lines between the new `buildArgs` backfill and the existing `allowlistOnly` backfill. Cosmetic only.

```suggestion
            if (options.platform?.dapi?.rsDapi?.docker?.build
              && typeof options.platform.dapi.rsDapi.docker.build.buildArgs === 'undefined') {
              options.platform.dapi.rsDapi.docker.build.buildArgs = {};
            }

            if (options.platform?.drive?.tenderdash?.p2p
```

<sub>source: ['claude']</sub>

thepastaclaw (changes requested):
## Code Review

Latest delta removes the CARGO_BUILD_PROFILE=release pin from the local setup script and JSON-quotes SDK_TEST_DATA so the path/value combo passes validation. Two prior findings carry forward (set.js still JSON-parses values before schema validation; buildArgs values still render into YAML unescaped). One new regression: with env forwarding and `${CARGO_BUILD_PROFILE:-dev}` both removed, the previously-documented `CARGO_BUILD_PROFILE=release yarn start` per-invocation override no longer reaches Docker builds.

<details>

```

In `packages/dashmate/src/commands/config/set.js`:
- [BLOCKING] packages/dashmate/src/commands/config/set.js:54-60: [Carried-forward — STILL VALID] `dashmate config set` JSON-parses build-arg values, so bare string values that look like JSON literals are rejected
  The path precheck was correctly swapped to `Config.isSchemaPathAllowed`, but `runWithDependencies` still unconditionally `JSON.parse`s every CLI value. Because `configJsonSchema.js` types `buildArgs` values as `{ type: 'string' }` (with `additionalProperties: { type: 'string' }`), any CLI invocation like `dashmate config set …buildArgs.CARGO_BUILD_PROFILE release` parses `release` fine (string) — but `dashmate config set …buildArgs.SDK_TEST_DATA true` parses to boolean `true` and is rejected by AJV. The new setup script works only because it passes `'"true"'` (a JSON-quoted string). This is the PR's new public API; documented use cases (`SDK_TEST_DATA`, plus future flag-style args like `1`/`0`/`null`) all collide with this typing. Either short-circuit JSON.parse for `…docker.build.buildArgs.…` paths, or fall back to the raw string when JSON.parse yields a value that fails schema validation as a string.

In `packages/dashmate/src/config/generateEnvsFactory.js`:
- [SUGGESTION] packages/dashmate/src/config/generateEnvsFactory.js:106-109: Shell-only `CARGO_BUILD_PROFILE=release yarn start` overrides no longer reach Docker builds
  This PR removes both halves of the previous shell-env passthrough: `generateEnvsFactory.js` explicitly stops forwarding `docker.build.buildArgs` into the compose environment, and `docker-compose.build.drive_abci.yml` / `docker-compose.build.rs-dapi.yml` drop their `CARGO_BUILD_PROFILE: ${CARGO_BUILD_PROFILE:-dev}` entries (confirmed in the diff against v3.1-dev). After this change, `CARGO_BUILD_PROFILE=release yarn start` is silently a no-op — the only working path is `dashmate config set …buildArgs.CARGO_BUILD_PROFILE '"release"'`. The setup script's own comment correctly tells operators to use `dashmate config set`, so the script is internally consistent, but the PR description still advertises the shell-env override. Either restore the `${CARGO_BUILD_PROFILE:-dev}` fallback in the two build compose files (so `dynamic-compose.yml` overrides win when set but the shell env still works otherwise), or update the PR description to drop the shell-env claim and point users at `dashmate config set` exclusively.

In `packages/dashmate/templates/dynamic-compose.yml.dot`:
- [SUGGESTION] packages/dashmate/templates/dynamic-compose.yml.dot:30: [Carried-forward — STILL VALID] buildArgs values are raw-interpolated into YAML without escaping or schema char restrictions
  Both render sites (`drive_abci` line 30, `rs_dapi` line 58) emit `{{=pair[0]}}: "{{=pair[1]}}"` with no escaping. `configJsonSchema.js` restricts keys via `propertyNames.pattern: '^[A-Za-z_][A-Za-z0-9_]*$'` but values are just `{ type: 'string' }` with no character restriction. A value containing `"`, `\`, or a newline (settable via `dashmate config set …buildArgs.FOO 'a"b'`) renders into malformed YAML and breaks the next compose invocation — there is no subsequent validation step over the generated `dynamic-compose.yml`. Operator-controlled input, so it's a robustness/UX concern rather than security. The cleanest fix is to mirror the key restriction onto values in the schema (any printable-ASCII regex that excludes `"`, `\`, and control chars easily covers realistic build args like `dev`/`release`/`true`/version/path strings). Alternatively, render values via `JSON.stringify` in the .dot template so YAML-quoting handles edge cases automatically.
```
</details>

packages/dashmate/src/config/generateEnvsFactory.js
thepastaclaw, line 109: **🟡 Suggestion: [Carried-forward — STILL VALID] PR description still advertises shell-env CARGO_BUILD_PROFILE override, but env passthrough was removed**

`generateEnvsFactory.js` now explicitly stops forwarding `docker.build.buildArgs` into the compose environment (replaced with a NOTE comment), and `docker-compose.build.drive_abci.yml` / `docker-compose.build.rs-dapi.yml` no longer carry `CARGO_BUILD_PROFILE: ${CARGO_BUILD_PROFILE:-dev}`. Only values persisted into dashmate config land in `dynamic-compose.yml`, so a shell-only invocation like `CARGO_BUILD_PROFILE=release yarn start` is silently ignored. This contradicts the PR description's stated workflow ("per-invocation via CARGO_BUILD_PROFILE=release yarn start or persist via dashmate config set"). `scripts/setup_local_network.sh` already reflects the new reality by using `dashmate config set ...buildArgs.CARGO_BUILD_PROFILE '"release"'`. Either update the PR description to match (config-only workflow), or restore a narrowly-scoped env override for ergonomics. Documenting the operator workflow consistently matters because this is the headline UX of the change.

thepastaclaw (commented):
## Code Review

Incremental + cumulative review at 6f083eb093 (no new delta vs prior reviewed SHA). Reconciled all four prior findings: the migration blank-line was fixed; the YAML-escaping and shell-env-passthrough findings remain valid as suggestions; the set.js JSON.parse issue is downgraded — set.js is not in this PR's cumulative diff against origin/v3.1-dev and the PR ships an explicit workaround. No blocking issues; review action COMMENT.

🟡 2 suggestion(s)

<details>

```

In `packages/dashmate/src/config/generateEnvsFactory.js`:
- [SUGGESTION] packages/dashmate/src/config/generateEnvsFactory.js:106-109: [Carried-forward — STILL VALID] PR description still advertises shell-env CARGO_BUILD_PROFILE override, but env passthrough was removed
  `generateEnvsFactory.js` now explicitly stops forwarding `docker.build.buildArgs` into the compose environment (replaced with a NOTE comment), and `docker-compose.build.drive_abci.yml` / `docker-compose.build.rs-dapi.yml` no longer carry `CARGO_BUILD_PROFILE: ${CARGO_BUILD_PROFILE:-dev}`. Only values persisted into dashmate config land in `dynamic-compose.yml`, so a shell-only invocation like `CARGO_BUILD_PROFILE=release yarn start` is silently ignored. This contradicts the PR description's stated workflow ("per-invocation via CARGO_BUILD_PROFILE=release yarn start or persist via dashmate config set"). `scripts/setup_local_network.sh` already reflects the new reality by using `dashmate config set ...buildArgs.CARGO_BUILD_PROFILE '"release"'`. Either update the PR description to match (config-only workflow), or restore a narrowly-scoped env override for ergonomics. Documenting the operator workflow consistently matters because this is the headline UX of the change.

In `packages/dashmate/src/commands/config/set.js`:
- [SUGGESTION] packages/dashmate/src/commands/config/set.js:54-60: [Carried-forward — STILL VALID, downgraded] `dashmate config set` JSON-parses values; new buildArgs string-only schema rejects bare `true`/`false`/`1`/`null`
  `runWithDependencies` still runs `JSON.parse(optionValue)` with a string-fallback before `config.set(...)`. For the new `...docker.build.buildArgs.<KEY>` paths, `configJsonSchema.js` requires every value to be a string, so bare CLI inputs that happen to be valid JSON non-strings (`true`, `false`, `null`, `1`, `[…]`, `{…}`) are converted to non-strings and then rejected by AJV. Operators have to know to wrap such values in JSON quotes (e.g. `'"true"'`), which `scripts/setup_local_network.sh` already does. Note that `set.js` itself is NOT modified in this PR's cumulative diff against `origin/v3.1-dev`, so the JSON.parse logic predates this branch; however, the new buildArgs schema is what exposes the sharp edge for the PR's intended workflow. Typical CARGO_BUILD_PROFILE values (`dev`, `release`) parse cleanly via the catch-fallback, so this is an edge case rather than a workflow-breaker — hence suggestion rather than blocking. Consider bypassing JSON.parse for `…buildArgs.*` paths since the schema is string-only.
```
</details>

packages/dashmate/src/config/generateEnvsFactory.js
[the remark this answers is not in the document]
shumkov: Resolved by the merge with v3.1-dev (800583d5): the buildArgs env-passthrough block in `generateEnvsFactory.js` was removed entirely. buildArgs now render directly into the per-config `dynamic-compose.yml` (`build.args`) via `dynamic-compose.yml.dot`, so there is no longer an env-merge path that could clobber reserved runtime keys. Stale — resolving.

docs/shielded-seeder-performance.md
shumkov: `docs/shielded-seeder-performance.md` was removed in 868ae52122 (the dynamic-compose pivot) and is no longer referenced from the schema or defaults. The flagged config path no longer exists in any doc. Stale — resolving.

packages/dashmate/src/config/configJsonSchema.js
shumkov: The "shell env overrides any value set here" wording was removed during the merge resolution (800583d5). The schema comment now only describes the dynamic-compose render path and makes no precedence claim. Stale — resolving.

packages/dashmate/src/config/generateEnvsFactory.js
shumkov: The single-source-of-truth comment block was removed entirely in the merge (env-passthrough dropped in favor of dynamic-compose). Stale — resolving.

packages/dashmate/src/config/generateEnvsFactory.js
shumkov: The buildArgs merge logic in `generateEnvsFactory.js` was removed; forwarding moved to `dynamic-compose.yml.dot`, covered by `dynamicComposeBuildArgs.spec.js`. No merge-precedence code remains to test. Stale — resolving.

packages/dashmate/configs/getConfigFileMigrationsFactory.js
shumkov: Fixed in 6f083eb093 — the double blank line was collapsed to a single blank line. Resolving.

packages/dashmate/templates/dynamic-compose.yml.dot
shumkov: Fixed in 3d461d8a9a — `build.args` values now render via `JSON.stringify`, so a value containing a quote, backslash or newline stays valid YAML. Added a js-yaml round-trip test. Resolving.

shumkov: Fixed in 3d461d8a9a — values now render via `JSON.stringify` (valid-YAML escaping) with a js-yaml round-trip test. Resolving.

packages/dashmate/src/config/Config.js
```diff
@@ -45,6 +45,74 @@ export default class Config {
...
+   *   schema object — `additionalProperties: false` blocks the descent),
+   * - `$ref` references into `#/definitions/...`.
+   *
+   * @param {string} path - dot-separated option path (e.g.
+   *   `'platform.drive.abci.docker.build.buildArgs.SDK_TEST_DATA'`).
+   * @return {boolean} true when the path is allowed by the schema.
+   */
+  static isSchemaPathAllowed(path) {
+    if (typeof path !== 'string' || path.length === 0) return false;
+
```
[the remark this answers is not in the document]
shumkov: Fixed in 3d461d8a9a — `isSchemaPathAllowed` now rejects paths with empty segments (leading/trailing/double dots), so an empty key can't slip through a map's `additionalProperties` descent. Test added. Resolving.

scripts/setup_local_network.sh
shumkov: Fixed in 3d461d8a9a — corrected the comment: SDK_TEST_DATA seeding runs at genesis (InitChain, runtime), not at docker build. Release matters because the compiled binary runs that seeding ~10x faster. Resolving.

packages/dashmate/test/unit/templates/dynamicComposeBuildArgs.spec.js
```diff
@@ -0,0 +1,65 @@
...
+import path from 'path';
+import { fileURLToPath } from 'url';
+import dot from 'dot';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+const TEMPLATE_PATH = path.resolve(
+  __dirname,
+  '../../../templates/dynamic-compose.yml.dot',
+);
```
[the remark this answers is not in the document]
shumkov: Fixed in 3d461d8a9a — renamed the dangle-underscore locals to `currentDir` (they weren't a repo convention; `src/constants.js` uses the inline `fileURLToPath(import.meta.url)` form). Resolving.

packages/dashmate/src/config/generateEnvsFactory.js
shumkov: Intentional design: this PR makes dashmate config (`buildArgs`) the single source of truth and drops the shell-env passthrough. Operators set the profile via `dashmate config set ...buildArgs.CARGO_BUILD_PROFILE '"release"'` rather than `CARGO_BUILD_PROFILE=release yarn start`. Resolving as won't-fix.

shumkov: Updated the PR description to match the shipped behavior: dashmate config is now the single source of truth for build args (no shell-env passthrough), and the operator workflow uses `dashmate config set …buildArgs.CARGO_BUILD_PROFILE '"release"'` rather than `CARGO_BUILD_PROFILE=release yarn start`. Added an explicit note that shell-only invocations are not forwarded. Resolving.

thepastaclaw (approved):
## Code Review

Latest delta resolves the YAML escaping issue (dynamic-compose.yml.dot now wraps buildArg values via JSON.stringify with a js-yaml round-trip regression test) and tightens isSchemaPathAllowed with an empty-segment guard plus unit coverage. The carried-forward 'dashmate config set foot-gun' finding from both agents does not survive verification: Config.ajv is constructed with coerceTypes: true, so bare CLI inputs true/false/1/null are silently coerced into strings before AJV validates against the buildArgs string schema. No in-scope issues remain.

thepastaclaw (approved):
## Code Review

Cumulative review of PR #3764 against v3.1-dev (head 59cceebd). The latest delta is a no-shadow lint rename in `Config.isSchemaPathAllowed` — zero behavior change. The full PR cleanly wires per-image docker build-args into `dynamic-compose.yml` for `drive_abci`/`rs_dapi`, with safe JSON.stringify rendering, env-var-identifier propertyNames constraint, empty-segment path rejection, and matching tests. No in-scope blockers or suggestions. One follow-up worth tracking separately about the shared `dockerBuild` schema exposing `buildArgs` on two sites that the renderer ignores.

PR #3771, opened 2026-05-31
test(wasm-sdk): fix flaky tokenPaymentInfo document balance assertions

QuantumExplorer:
The nightly `Tests` workflow (`Packages functional tests / Run functional tests` → `Run WASM SDK functional tests`) has been failing intermittently for several scheduled runs, e.g. [run 26697658517](https://github.com/dashpay/platform/actions/runs/26697658517/job/78685260617).

The failures are in the `tokenPaymentInfo document flow` tests in `packages/wasm-sdk/tests/functional/transitions/documents.spec.ts`. They are flaky, not deterministic — different nightly runs fail on different lines with different stale values:

- 2026-05-30: line 592 `expected 911n to equal 916n`, line 619 `expected 916n to equal 917n`
- 2026-05-29: line 539 `expected 45n to equal 41n`, line 619 `expected 905n to equal 917n`

These tests pass on push-triggered runs only because the changed-package gating skips the WASM SDK functional suite there; the scheduled run exercises everything.

## What was done?
Root cause: each balance check was a single proof-verified `getIdentityTokenBalances` read taken **immediately** after a document state transition. That balance proof can be served by a node whose view briefly lags the block that committed the transition, so an immediate read intermittently returns a stale balance. The existing `waitForPlatform()` sleep sat *after* the assertions, so it never protected them.

Replaced the immediate `expect(await getSingleTokenBalance(...)).to.equal(x)` assertions with a new `expectTokenBalance(...)` helper that polls the balance until it converges to the expected value (500ms interval, 30s timeout) before asserting. On a clean read it returns immediately; under lag it waits instead of failing.

## How Has This Been Tested?
Change is test-only and mechanical (assertion sites rewritten to the polling helper; the expected values are unchanged). The expected balances were already correct — the suite passes on most runs — so converging on them removes the read-after-write race. Verified the full diff; no production code is touched.

## Breaking Changes
None.

## Checklist:

## Summary by CodeRabbit

* **Tests**
  * Improved reliability of token-balance checks in document/payment-flow tests by switching to a polling-based verification to handle timing variations after state transitions.
  * Updated test assertions and expected post-transition token amounts for all involved parties (seller, buyer, third-party) to reflect finalized balances.

### DIFF
--- a/packages/wasm-sdk/tests/functional/transitions/documents.spec.ts
+++ b/packages/wasm-sdk/tests/functional/transitions/documents.spec.ts
@@ -30,6 +30,23 @@ describe('Document State Transitions', function describeDocumentStateTransitions
     const balances = await client.getIdentityTokenBalances(identityId, [tokenId]);
     return balances.get(tokenId);
   };
+  // Token balance proofs are served by nodes whose view can briefly lag the block
+  // that committed a state transition, so poll until the balance converges instead
+  // of asserting on a single immediate (and racy) read.
+  const expectTokenBalance = async (
+    identityId: string,
+    tokenId: string,
+    expected: bigint,
+    { timeoutMs = 30000, intervalMs = 500 } = {},
+  ) => {
+    const deadline = Date.now() + timeoutMs;
+    let actual = await getSingleTokenBalance(identityId, tokenId);
+    while (actual !== expected && Date.now() < deadline) {
+      await new Promise((resolve) => { setTimeout(resolve, intervalMs); });
+      actual = await getSingleTokenBalance(identityId, tokenId);
+    }
+    expect(actual).to.equal(expected);
+  };
   const buildSimpleTokenConfiguration = (baseSupply: bigint, newTokensDestinationIdentity: string) => {
     const contractOwner = sdk.AuthorizedActionTakers.ContractOwner();
     const contractOwnerChangeRules = new sdk.ChangeControlRules({
@@ -395,7 +412,7 @@ describe('Document State Transitions', function describeDocumentStateTransitions
 
       await waitForPlatform();
 
-      expect(await getSingleTokenBalance(testData.identityId, tokenPaidTokenId)).to.equal(1000n);
+      await expectTokenBalance(testData.identityId, tokenPaidTokenId, 1000n);
 
       await client.tokenTransfer({
         dataContractId: tokenPaidContractId,
@@ -419,9 +436,9 @@ describe('Document State Transitions', function describeDocumentStateTransitions
 
       await waitForPlatform();
 
-      expect(await getSingleTokenBalance(testData.identityId, tokenPaidTokenId)).to.equal(900n);
-      expect(await getSingleTokenBalance(testData.identityId2, tokenPaidTokenId)).to.equal(50n);
-      expect(await getSingleTokenBalance(testData.identityId3, tokenPaidTokenId)).to.equal(50n);
+      await expectTokenBalance(testData.identityId, tokenPaidTokenId, 900n);
+      await expectTokenBalance(testData.identityId2, tokenPaidTokenId, 50n);
+      await expectTokenBalance(testData.identityId3, tokenPaidTokenId, 50n);
     });
 
     it('should reject create when tokenPaymentInfo is omitted', async () => {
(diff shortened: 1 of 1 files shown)


thepastaclaw:
✅ Review complete (commit 96dc5f60)

packages/wasm-sdk/tests/functional/transitions/documents.spec.ts
```diff
@@ -30,6 +30,23 @@ describe('Document State Transitions', function describeDocumentStateTransitions
...
+    { timeoutMs = 30000, intervalMs = 500 } = {},
+  ) => {
+    const deadline = Date.now() + timeoutMs;
+    let actual = await getSingleTokenBalance(identityId, tokenId);
+    while (actual !== expected && Date.now() < deadline) {
+      await new Promise((resolve) => { setTimeout(resolve, intervalMs); });
+      actual = await getSingleTokenBalance(identityId, tokenId);
+    }
+    expect(actual).to.equal(expected);
+  };
```
thepastaclaw, line 55: **💬 Nitpick: expectTokenBalance timeout failure lacks debugging context**

When the poll loop exits because the deadline elapsed, the final `expect(actual).to.equal(expected)` produces a plain equality mismatch with no indication that polling timed out, nor any of the data needed to triage a flake (identityId, tokenId, expected/actual, timeoutMs, elapsed). Since the whole purpose of this helper is to ride out a transient lag, a timeout failure here is the most likely failure mode and the one that most needs context in CI logs. Attach an assertion message that includes identityId, tokenId, expected, actual, timeoutMs, and elapsed time; keep the polling loop unchanged.

thepastaclaw (commented):
## Code Review

Test-only change replaces immediate token-balance assertions with a 30s/500ms polling helper to address a documented read-after-write race in token balance proofs. Both agents found no in-scope defects. CodeRabbit's nitpick about the unhelpful timeout failure message is valid but low-priority and does not block.

💬 1 nitpick(s)

packages/wasm-sdk/tests/functional/transitions/documents.spec.ts
```diff
@@ -395,7 +412,7 @@ describe('Document State Transitions', function describeDocumentStateTransitions
 
       await waitForPlatform();
```
shumkov, line 419: We just need to increase timeout in waitForPlatform

shumkov (changes requested):

thepastaclaw (approved):
## Code Review

Latest delta is a one-commit, test-only follow-up that adds triage context to the `expectTokenBalance` timeout assertion in `packages/wasm-sdk/tests/functional/transitions/documents.spec.ts`. The prior nitpick from beced7f1 is FIXED in substance: timeout failures now include identityId, tokenId, expected value, elapsed ms, and last observed value. No new in-scope defects were identified at HEAD 96dc5f60; both Claude and Codex agree.

packages/wasm-sdk/tests/functional/transitions/documents.spec.ts
shumkov: Addressed in `96dc5f60cc`: the timeout assertion now reports the identity, token, expected value, elapsed time, and last read — so a flake surfaces with full triage context (including the `undefined`-balance case). The polling loop itself is unchanged.

packages/wasm-sdk/tests/functional/transitions/documents.spec.ts
shumkov: Changed approach after digging in. The SDK already waits for state-transition execution — `put_to_platform_and_wait_for_response` → `WaitForStateTransitionResult` only returns once the ST is committed in a block. The flake is a *post-commit* read lag: the local network runs 3 masternodes, and the follow-up proof-verified `getIdentityTokenBalances` can be served by a node whose view briefly trails the committed block. A larger `waitForPlatform` sleep would pay the full wait on every (usually fresh) read and still be racy when the lag exceeds it. The platform test suite mitigates the same read-after-write race with a polling loop (`waitForBalanceToChange`, 500ms/20s), so this helper mirrors that precedent. Keeping the polling helper.

shumkov (approved):
Kept the polling helper (mirrors the platform test suite's `waitForBalanceToChange`) over a larger `waitForPlatform` sleep — the SDK already waits for ST execution, so the flake is a post-commit cross-node read lag that polling rides out without paying a fixed wait on every read. Nitpick about timeout triage context addressed in 96dc5f60cc. Approving.

PR #3777, opened 2026-06-01
fix(swift-sdk): fixed mempool tx categorization after restart

ZocoLini:
After a mid-flight restart between mempool sighting and block confirmation, a self-send tx was misclassified as Incoming with netAmount = +sum_of_outputs. The persister flipped PersistentTxo.isSpent = true on the mempool sighting, and loadWalletList filters by isSpent == false, so the input TXO was dropped from the restored set — the catch-up classifier on the next launch saw an "unknown" input and emitted Incoming.

Fix: gate the isSpent write on the spending tx being in a block (context >= inBlock). Mempool sightings still link the spendingTransaction relationship but keep the row in the unspent set. The next upsert with a confirmed context flips isSpent then.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed spend-state tracking so mempool-only (unconfirmed) spends are no longer incorrectly marked as confirmed.
  * Improved confirmation detection to reliably distinguish unconfirmed spending transactions from those confirmed in a block.
  * Enhanced pending-transaction resolution and linkage handling to keep UTXO spent/unspent status more accurate and stable during context changes, reducing false positives and state flips.

### DIFF
--- a/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
@@ -627,6 +627,18 @@ public class PlatformWalletPersistenceHandler {
         }
     }
 
+    /// `true` if the spending tx has reached a confirmed context. Used
+    /// to gate `isSpent` writes so a mempool-sighting alone — which is
+    /// reversible by RBF or mempool eviction — doesn't permanently flip
+    /// the input TXO out of the unspent set. The TXO becomes truly spent
+    /// only when the spending tx lands in a block; until then the
+    /// persisted state reflects "still spendable from this row's POV",
+    /// and the catch-up classifier on the next launch reloads the
+    /// row and recognises it as ours when the block arrives.
+    private static func spendIsInBlock(_ tx: PersistentTransaction) -> Bool {
+        tx.context >= TransactionContextType.inBlock.rawValue
+    }
+
     /// Mark the `PersistentTxo` whose 36-byte `outpoint` matches the
     /// given input as spent and link it to `spendingTransaction`.
     /// If no matching TXO exists yet (in-Swift out-of-order, or
@@ -644,16 +656,21 @@ public class PlatformWalletPersistenceHandler {
             predicate: #Predicate { $0.outpoint == outpoint }
         )
         if let txo = try? backgroundContext.fetch(txoDescriptor).first {
-            // Only touch the row if the linkage actually changes —
-            // an idempotent re-upsert of the same tx must not
-            // gratuitously bump `lastUpdated` and trigger a follow-on
-            // changeset emit.
+            // `isSpent` only flips once the spending tx is in a block
+            // (see `spendIsInBlock`'s doc) — a mempool sighting
+            // alone links the spending relationship but keeps the
+            // row in the unspent set so a `restartWalletManager()`
+            // load can hand the TXO back to Rust for the post-restart
+            // catch-up classifier to recognise as ours. The next
+            // upsert of this same tx with a confirmed context flips
+            // `isSpent` then.
+            let expectedIsSpent = Self.spendIsInBlock(spendingTransaction)
             let linkageChanged =
-                !txo.isSpent
+                txo.isSpent != expectedIsSpent
                 || txo.spendingTransaction?.txid != spendingTxid
                 || txo.spendingInputIndex != inputIndex
             if linkageChanged {
-                txo.isSpent = true
+                txo.isSpent = expectedIsSpent
                 if txo.spendingTransaction?.txid != spendingTxid {
                     txo.spendingTransaction = spendingTransaction
                 }
@@ -822,7 +839,24 @@ public class PlatformWalletPersistenceHandler {
             // visible spendingTransaction matches the most recent
             // observation; the rest are dropped.
             let chosen = pendingRows.max(by: { $0.createdAt < $1.createdAt }) ?? pendingRows[0]
-            record.isSpent = true
+
+            // Resolve the spending tx (prefer the relationship; fall
+            // back to a txid lookup if the row wasn't faulted in).
+            // We need its `context` to gate `isSpent` — same rule as
+            // `resolveInputOutpoint`: mempool sighting links the
+            // spendingTransaction but doesn't flip `isSpent` until
+            // the spending tx is in a block.
+            let resolvedSpending: PersistentTransaction?
+            if let spending = chosen.spendingTransaction {
+                resolvedSpending = spending
+            } else {
+                let spendingTxid = chosen.spendingTxid
+                let txDescriptor = FetchDescriptor<PersistentTransaction>(
+                    predicate: #Predicate { $0.txid == spendingTxid }
+                )
+                resolvedSpending = try? backgroundContext.fetch(txDescriptor).first
+            }
+
             // Carry the vin index forward so the spending tx's
             // detail view can render its inputs in the canonical
             // serialized order. Same source as the linkage write
@@ -830,21 +864,12 @@ public class PlatformWalletPersistenceHandler {
             // pending rows captures the index from FFI's
             // `input_outpoints` slice, which mirrors `tx.input.iter()`.
             record.spendingInputIndex = chosen.inputIndex
-            if let spending = chosen.spendingTransaction {
-                if record.spendingTransaction?.txid != spending.txid {
-                    record.spendingTransaction = spending
-                }
-            } else {
-                // Pending row's parent tx wasn't faulted in; fall
-                // back to a txid lookup so the linkage still lands.
-                let spendingTxid = chosen.spendingTxid
-                let txDescriptor = FetchDescriptor<PersistentTransaction>(
-                    predicate: #Predicate { $0.txid == spendingTxid }
-                )
-                if let spending = try? backgroundContext.fetch(txDescriptor).first,
-                   record.spendingTransaction?.txid != spending.txid {
-                    record.spendingTransaction = spending
-                }
+            if let spending = resolvedSpending,
+               record.spendingTransaction?.txid != spending.txid {
+                record.spendingTransaction = spending
+            }
+            if let spending = resolvedSpending {
+                record.isSpent = Self.spendIsInBlock(spending)
             }
             record.lastUpdated = Date()
             for row in pendingRows {
@@ -864,7 +889,6 @@ public class PlatformWalletPersistenceHandler {
         guard let txo = try? backgroundContext.fetch(descriptor).first else {
             return
         }
-        txo.isSpent = true
         // Link the spending transaction. The FFI now carries
         // `spending_txid` alongside the outpoint (the txid of the
         // `TransactionRecord` whose inputs included this outpoint),
@@ -875,15 +899,29 @@ public class PlatformWalletPersistenceHandler {
         // next flush carrying that tx triggers another upsert
         // round and eventually catches up.
         let spendingTxid = hashData(entry.spending_txid)
-        if !spendingTxid.isEmpty,
-           !spendingTxid.allSatisfy({ $0 == 0 }),
-           txo.spendingTransaction?.txid != spendingTxid {
-            let txDescriptor = FetchDescriptor<PersistentTransaction>(
-                predicate: #Predicate { $0.txid == spendingTxid }
-            )
-            if let spendingTx = try? backgroundContext.fetch(txDescriptor).first {
-                txo.spendingTransaction = spendingTx
+        var spendingTx: PersistentTransaction? = nil
+        if !spendingTxid.isEmpty, !spendingTxid.allSatisfy({ $0 == 0 }) {
+            if txo.spendingTransaction?.txid == spendingTxid {
+                spendingTx = txo.spendingTransaction
+            } else {
+                let txDescriptor = FetchDescriptor<PersistentTransaction>(
+                    predicate: #Predicate { $0.txid == spendingTxid }
+                )
+                spendingTx = try? backgroundContext.fetch(txDescriptor).first
+                if let spending = spendingTx {
+                    txo.spendingTransaction = spending
+                }
             }
+        }
+        // Gate the `isSpent` flip on the spending tx being in a
+        // block — same rule as `resolveInputOutpoint`. When the
+        // spending tx isn't resolved this flush, leave `isSpent`
+        // alone instead of writing `false`: the next upsert round
+        // carrying the spending tx will run `resolveInputOutpoint`
+        // and set it then. Writing `false` here would flap a
+        // previously-true `isSpent` on every reordered emit.
+        if let spending = spendingTx {
+            txo.isSpent = Self.spendIsInBlock(spending)
         }
         txo.lastUpdated = Date()
         // The spend signal landed both via the legacy


thepastaclaw:
✅ Review complete (commit d20ddee3)

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -627,6 +627,19 @@ public class PlatformWalletPersistenceHandler {
...
+    /// writes so a mempool-sighting alone — which is reversible by
+    /// RBF or mempool eviction — doesn't permanently flip the input
+    /// TXO out of the unspent set. The TXO becomes truly spent only
+    /// when the spending tx lands in a block; until then the
+    /// persisted state reflects "still spendable from this row's POV",
+    /// and the catch-up classifier on the next launch reloads the
+    /// row and recognises it as ours when the block arrives.
+    private static func spendIsInBlock(_ tx: PersistentTransaction) -> Bool {
+        tx.context >= TransactionContextType.inBlock.rawValue
+    }
```
thepastaclaw, line 640: **🔴 Blocking: InstantSend-locked spends regress to unspent across a restart**

`TransactionContextType` ordering is `mempool=0, instantSend=1, inBlock=2, inChainLockedBlock=3`, so `tx.context >= inBlock.rawValue` excludes `instantSend`. The stated reason for the new gate — "reversible by RBF or mempool eviction" — does not apply to InstantSend: by Dash design it cannot be reorged out and cannot be replaced. Treating it the same as mempool produces two concrete regressions:

1. While the spending tx is `instantSend` but not yet in a block, the input TXO keeps `isSpent == false`, so the wallet UI/queries that read SwiftData see the row as still spendable.
2. Across a restart in that window, `loadWalletList` (line 2948) filters TXOs by `isSpent == false` and hands the row back to Rust as a live UTXO. Nothing in the restore path rebuilds ordinary spending transactions into Rust's in-memory graph, so the spent input is silently re-credited until a later block update arrives, which can overstate spendable funds and enable a duplicate-spend attempt.

This contradicts the project's own contract — `ManagedCoreWallet.CoreBalance.confirmed`'s doc literally says "Confirmed balance (in a block or InstantSend-locked)" — so the persistence layer should follow the same definition.

```suggestion
    private static func spendIsInBlock(_ tx: PersistentTransaction) -> Bool {
        tx.context >= TransactionContextType.instantSend.rawValue
    }
```

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -875,16 +898,27 @@ public class PlatformWalletPersistenceHandler {
...
+                }
             }
         }
+        // Gate the `isSpent` flip on the spending tx being in a
+        // block — same rule as `resolveInputOutpoint`. When the
+        // spending tx isn't found yet, conservatively assume not
+        // confirmed; the next upsert round with the spending tx
+        // landed will run `resolveInputOutpoint` and flip the
+        // flag if appropriate.
+        txo.isSpent = spendingTx.map(Self.spendIsInBlock) ?? false
```
thepastaclaw: **🟡 Suggestion: `markUtxoSpent` can now clear a previously-set `isSpent`**

Before this PR, `markUtxoSpent` was an authoritative spend signal: once the TXO was located, `isSpent` was set to `true` unconditionally. After this change the tail-write is `txo.isSpent = spendingTx.map(Self.spendIsInBlock) ?? false`, which can flip `isSpent` from `true` back to `false` in three paths:

1. The guard at line 902 leaves `spendingTx == nil` when the FFI emits an empty/all-zero `spending_txid`. The fact that this guard exists implies Rust can emit that case — and the legacy semantic there was "mark as spent anyway".
2. The spending tx hasn't been persisted yet this flush. The comment promises self-healing on the next flush, but until then the TXO observably toggles back to spendable.
3. A TXO already confirmed-spent in an earlier flush is re-emitted here with the spending tx's record evicted from this `backgroundContext` — `isSpent` goes `true → false`.

The PR's stated intent ("don't permanently flip `isSpent` from a mempool-only sighting") only requires the negative case for mempool. An upgrade-only rule — never clear `isSpent` here, only set `false → true` when the spending tx is confirmed — preserves the fix while keeping the existing authoritative-signal semantics. Same shape should be applied at the deferred-pending drain (line 872) for consistency.

<sub>source: ['claude']</sub>

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
thepastaclaw: **💬 Nitpick: No regression test for the restart misclassification**

The fix targets a subtle ordering bug (self-send mempool sighting → restart → `loadWalletList` drops the input TXO because `isSpent==true` → catch-up classifier misclassifies as Incoming) that's easy to silently regress later. A regression test that (1) upserts a tx in mempool context spending a persisted TXO, (2) asserts `PersistentTxo.isSpent == false` with `spendingTransaction` linked, (3) re-runs the same predicate `loadWalletList` uses, (4) re-upserts the tx with `inBlock` context, and (5) asserts `isSpent == true` would lock the contract in place.

<sub>source: ['claude']</sub>

thepastaclaw (changes requested):
## Code Review

The PR correctly identifies and partially fixes the mempool self-send restart-misclassification bug. However, `spendIsInBlock` is too strict: it excludes `instantSend`, which this codebase already treats as confirmed (`CoreBalance.confirmed` doc string says "in a block or InstantSend-locked"). Combined with `loadWalletList`'s `isSpent == false` filter, this regresses InstantSend spends across a restart, re-feeding the spent UTXO to Rust as unspent. A second, narrower concern: `markUtxoSpent` now flips `isSpent` back to `false` whenever the spending tx can't be resolved, including the empty/zero `spending_txid` guard path, which previously was an authoritative spend signal.

<details>

```

- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift:639-641: InstantSend-locked spends regress to unspent across a restart
  `TransactionContextType` ordering is `mempool=0, instantSend=1, inBlock=2, inChainLockedBlock=3`, so `tx.context >= inBlock.rawValue` excludes `instantSend`. The stated reason for the new gate — "reversible by RBF or mempool eviction" — does not apply to InstantSend: by Dash design it cannot be reorged out and cannot be replaced. Treating it the same as mempool produces two concrete regressions:

  1. While the spending tx is `instantSend` but not yet in a block, the input TXO keeps `isSpent == false`, so the wallet UI/queries that read SwiftData see the row as still spendable.
  2. Across a restart in that window, `loadWalletList` (line 2948) filters TXOs by `isSpent == false` and hands the row back to Rust as a live UTXO. Nothing in the restore path rebuilds ordinary spending transactions into Rust's in-memory graph, so the spent input is silently re-credited until a later block update arrives, which can overstate spendable funds and enable a duplicate-spend attempt.

  This contradicts the project's own contract — `ManagedCoreWallet.CoreBalance.confirmed`'s doc literally says "Confirmed balance (in a block or InstantSend-locked)" — so the persistence layer should follow the same definition.
- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift:915-921: `markUtxoSpent` can now clear a previously-set `isSpent`
  Before this PR, `markUtxoSpent` was an authoritative spend signal: once the TXO was located, `isSpent` was set to `true` unconditionally. After this change the tail-write is `txo.isSpent = spendingTx.map(Self.spendIsInBlock) ?? false`, which can flip `isSpent` from `true` back to `false` in three paths:

  1. The guard at line 902 leaves `spendingTx == nil` when the FFI emits an empty/all-zero `spending_txid`. The fact that this guard exists implies Rust can emit that case — and the legacy semantic there was "mark as spent anyway".
  2. The spending tx hasn't been persisted yet this flush. The comment promises self-healing on the next flush, but until then the TXO observably toggles back to spendable.
  3. A TXO already confirmed-spent in an earlier flush is re-emitted here with the spending tx's record evicted from this `backgroundContext` — `isSpent` goes `true → false`.

  The PR's stated intent ("don't permanently flip `isSpent` from a mempool-only sighting") only requires the negative case for mempool. An upgrade-only rule — never clear `isSpent` here, only set `false → true` when the spending tx is confirmed — preserves the fix while keeping the existing authoritative-signal semantics. Same shape should be applied at the deferred-pending drain (line 872) for consistency.
```
</details>

ZocoLini:
I will be taking a look into pasta claw complains to be able to preperly answer them

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
ZocoLini: The threshold here is intentional and not a regression of the IS semantics elsewhere — these are 
  two different predicates serving two different goals:
  
   - CoreBalance.confirmed is display semantics ("can the UI show this as confirmed funds?"). Treating
   IS as confirmed there is correct.
  - spendIsInBlock is classifier-load semantics for loadWalletList ("should this TXO drop out of the 
  unspent set we hand back to key-wallet's catch-up?"). Until the spending tx is in a block, 
  key-wallet still needs to see the input in utxos so it can recognise the spend as ours on the 
  post-restart pass. If the input is missing, the classifier emits the tx as Incoming with netAmount 
  = +sum_of_outputs — that's the bug this gate fixes (caught by an integration test in 
  feat/swift-sdk-integration-tests).
  
  Cost of inBlock instead of instantSend: between the IS lock and the block landing, a restart 
  briefly shows the spent UTXO as available again. That's a transient display overstatement — the 
  network rejects any conflicting broadcast under the IS lock, so it's not a re-spend risk. Switching
   the threshold to instantSend would reintroduce the persistent row-level misclassification, which 
  is the worse failure mode.

ZocoLini: Regression test exists — testSelfSendClassificationSurvivesMidFlightRestart in PersisterRestartClassificationIntegrationTests on the feat/swift-sdk-integration-tests branch. It reproduces the exact mid-flight scenario (self-send sighting → restart → mine → catch-up classifier) and was what surfaced this bug originally. The test lands when that branch merges; happy to cherry-pick it ahead if you'd rather not wait.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -627,6 +627,18 @@ public class PlatformWalletPersistenceHandler {
         }
     }
 
+    /// `true` if the spending tx has reached a confirmed context. Used
+    /// to gate `isSpent writes so a mempool-sighting alone — which is
```
thepastaclaw: **💬 Nitpick: Unterminated backtick in `spendIsInBlock` doc comment**

The opening backtick on `` `isSpent `` is never closed before `writes`, which leaves the rest of the sentence rendered as inline code in Xcode quick-help and DocC output.

```suggestion
    /// `true` if the spending tx has reached a confirmed context. Used
    /// to gate `isSpent` writes so a mempool-sighting alone — which is
    /// reversible by RBF or mempool eviction — doesn't permanently flip
    /// the input TXO out of the unspent set. The TXO becomes truly spent
    /// only when the spending tx lands in a block; until then the
    /// persisted state reflects "still spendable from this row's POV",
    /// and the catch-up classifier on the next launch reloads the
    /// row and recognises it as ours when the block arrives.
```

<sub>source: ['claude']</sub>

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -627,6 +627,18 @@ public class PlatformWalletPersistenceHandler {
...
+    /// reversible by RBF or mempool eviction — doesn't permanently flip
+    /// the input TXO out of the unspent set. The TXO becomes truly spent
+    /// only when the spending tx lands in a block; until then the
+    /// persisted state reflects "still spendable from this row's POV",
+    /// and the catch-up classifier on the next launch reloads the
+    /// row and recognises it as ours when the block arrives.
+    private static func spendIsInBlock(_ tx: PersistentTransaction) -> Bool {
+        tx.context >= TransactionContextType.inBlock.rawValue
+    }
```
thepastaclaw, line 640: **💬 Nitpick: Call out InstantSend exclusion in `spendIsInBlock` doc**

`TransactionContextType.instantSend` (raw value 1) is intentionally excluded by the `>= inBlock` gate, but the doc only motivates the mempool/RBF case. Since IS-locked funds are treated as confirmed for balance display, a future contributor could reasonably read this as an oversight and widen the gate to `>= instantSend`, which would re-introduce the restart-misclassification bug for IS-locked self-sends. A single line in the comment noting that IS-locked-but-not-yet-mined spends are deliberately treated like mempool here — so the catch-up classifier can re-identify the TXO after restart, and the next `inBlock` upsert flips it — would lock that intent in.

thepastaclaw (commented):
## Code Review

The PR's gate on `isSpent` writes is intentionally scoped to block-confirmed contexts so mempool/IS-locked spends can be re-identified by the catch-up classifier on restart — this is explicitly stated in the doc comment and the linkage comment. Codex's blocking flag on InstantSend exclusion contradicts this documented design and is dropped as a false positive. The remaining valid feedback is a real doc-comment typo (unterminated backtick) plus low-priority suggestions around test coverage and clarifying the IS exclusion rationale.

💬 3 nitpick(s)

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -644,16 +656,21 @@ public class PlatformWalletPersistenceHandler {
...
             let linkageChanged =
-                !txo.isSpent
+                txo.isSpent != expectedIsSpent
                 || txo.spendingTransaction?.txid != spendingTxid
                 || txo.spendingInputIndex != inputIndex
             if linkageChanged {
-                txo.isSpent = true
+                txo.isSpent = expectedIsSpent
                 if txo.spendingTransaction?.txid != spendingTxid {
                     txo.spendingTransaction = spendingTransaction
```
thepastaclaw, line 675: **🟡 Suggestion: New write path can flip `isSpent` from true back to false on a stale/conflicting mempool spender**

The pre-PR code in `resolveInputOutpoint` unconditionally set `txo.isSpent = true`. The new code computes `expectedIsSpent = spendIsInBlock(spendingTransaction)` and writes it whenever `linkageChanged` is true. This is the intended fix when an unconfirmed spender arrives first — but it also means: if a TXO is already linked to a confirmed spender A (`isSpent == true`), and a later stale or conflicting mempool observation B for the same outpoint flows through this path, then `expectedIsSpent` is false, `linkageChanged` is true, and the code clears `isSpent` and reassigns `spendingTransaction` to B. Because `loadWalletList()` restores rows where `isSpent == false`, that TXO becomes re-eligible as spendable, leading to balance overcount and a possible user-driven double-spend attempt. The same shape exists in the pending-input drain path at line 871–872. Consider only allowing the spender to be replaced (and `isSpent` lowered) when the existing linked spender is also unconfirmed, the same txid is changing context, or an explicit reorg downgrade signal is present. Pre-PR behavior masked this because `true` was always rewritten; this PR makes the downward flip newly reachable.

<sub>source: ['codex']</sub>

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -627,6 +627,18 @@ public class PlatformWalletPersistenceHandler {
...
+    /// to gate `isSpent` writes so a mempool-sighting alone — which is
+    /// reversible by RBF or mempool eviction — doesn't permanently flip
+    /// the input TXO out of the unspent set. The TXO becomes truly spent
+    /// only when the spending tx lands in a block; until then the
+    /// persisted state reflects "still spendable from this row's POV",
+    /// and the catch-up classifier on the next launch reloads the
+    /// row and recognises it as ours when the block arrives.
+    private static func spendIsInBlock(_ tx: PersistentTransaction) -> Bool {
+        tx.context >= TransactionContextType.inBlock.rawValue
+    }
```
thepastaclaw, line 640: **🟡 Suggestion: [Carried forward] Document the deliberate InstantSend exclusion in `spendIsInBlock`**

`TransactionContextType` is `mempool=0, instantSend=1, inBlock=2, inChainLockedBlock=3` (KeyWallet/TransactionContext.swift:4-8). The threshold `tx.context >= inBlock.rawValue` therefore groups `instantSend` with `mempool` as 'don't flip `isSpent`'. The justification given in the comment is reversibility via RBF or mempool eviction — but an InstantSend-locked tx is precisely the case that is NOT reversible by RBF or eviction. Two consequences:

* For the bug being fixed (restart classifier needs the input TXO back as `isSpent == false`), lumping InstantSend with mempool is actually safe and arguably required, since the catch-up classifier still needs the input handed back until the spending tx lands in a block.
* For UI surfaces that read `isSpent == false` as 'spendable', an InstantSend-locked spend will appear spendable until the block confirms — a spendable-balance overcount window.

A future maintainer reading only the comment's RBF/eviction justification could reasonably widen the predicate to `>= instantSend.rawValue` and silently reintroduce the restart-classifier bug. Either widen the comment to explicitly call out InstantSend (intentional, and why), or — if the intent really is 'is this spend safely final' — change the threshold to `> mempool`. Recommend the doc-clarification approach so the restart-classifier contract stays intact.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -822,29 +839,37 @@ public class PlatformWalletPersistenceHandler {
...
-                    record.spendingTransaction = spending
-                }
+            if let spending = resolvedSpending,
+               record.spendingTransaction?.txid != spending.txid {
+                record.spendingTransaction = spending
+            }
+            if let spending = resolvedSpending {
+                record.isSpent = Self.spendIsInBlock(spending)
             }
             record.lastUpdated = Date()
```
thepastaclaw, line 874: **💬 Nitpick: Pending-resolution branch lacks the same flap-protection comment as `markUtxoSpent`**

`markUtxoSpent` (line 916–925) explicitly explains why it does NOT mirror the symmetric write: 'Writing `false` here would flap a previously-true `isSpent` on every reordered emit.' The pending-resolution branch in `upsertUtxo` does the symmetric `record.isSpent = Self.spendIsInBlock(spending)` unconditionally when `resolvedSpending != nil` — which can in principle reset a previously-true `isSpent` to false on a reorder. Either make the rule symmetric across the three write sites or add a one-line comment here noting the asymmetry is intentional (because in this path `chosen` is the freshest pending row, so it is the most recent observation by construction). Otherwise the next reader will conclude one of the three sites has the rule wrong.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Latest delta (95b219dd→d20ddee3) is a single-character doc fix (closing backtick on `isSpent`). The prior 'backtick typo' finding is FIXED; the prior 'document/clarify InstantSend exclusion' and 'add restart regression test' findings are STILL VALID against the head — they are carried forward below. The de novo concern surfaced this round is the symmetric `isSpent = expectedIsSpent` write in `resolveInputOutpoint`, which can now flip a previously-true `isSpent` to false if a stale/conflicting mempool spender for the same outpoint is observed after a confirmed spender — this is a behavioral regression introduced by this PR (pre-PR code unconditionally wrote `true`). It does not justify blocking on its own because it depends on an unusual conflict/reorg emit sequence, but it is the most important new observation in the cumulative state. Recommending COMMENT.

<details>

```

- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift:667-675: New write path can flip `isSpent` from true back to false on a stale/conflicting mempool spender
  The pre-PR code in `resolveInputOutpoint` unconditionally set `txo.isSpent = true`. The new code computes `expectedIsSpent = spendIsInBlock(spendingTransaction)` and writes it whenever `linkageChanged` is true. This is the intended fix when an unconfirmed spender arrives first — but it also means: if a TXO is already linked to a confirmed spender A (`isSpent == true`), and a later stale or conflicting mempool observation B for the same outpoint flows through this path, then `expectedIsSpent` is false, `linkageChanged` is true, and the code clears `isSpent` and reassigns `spendingTransaction` to B. Because `loadWalletList()` restores rows where `isSpent == false`, that TXO becomes re-eligible as spendable, leading to balance overcount and a possible user-driven double-spend attempt. The same shape exists in the pending-input drain path at line 871–872. Consider only allowing the spender to be replaced (and `isSpent` lowered) when the existing linked spender is also unconfirmed, the same txid is changing context, or an explicit reorg downgrade signal is present. Pre-PR behavior masked this because `true` was always rewritten; this PR makes the downward flip newly reachable.
- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift:630-640: [Carried forward] Document the deliberate InstantSend exclusion in `spendIsInBlock`
  `TransactionContextType` is `mempool=0, instantSend=1, inBlock=2, inChainLockedBlock=3` (KeyWallet/TransactionContext.swift:4-8). The threshold `tx.context >= inBlock.rawValue` therefore groups `instantSend` with `mempool` as 'don't flip `isSpent`'. The justification given in the comment is reversibility via RBF or mempool eviction — but an InstantSend-locked tx is precisely the case that is NOT reversible by RBF or eviction. Two consequences:

  * For the bug being fixed (restart classifier needs the input TXO back as `isSpent == false`), lumping InstantSend with mempool is actually safe and arguably required, since the catch-up classifier still needs the input handed back until the spending tx lands in a block.
  * For UI surfaces that read `isSpent == false` as 'spendable', an InstantSend-locked spend will appear spendable until the block confirms — a spendable-balance overcount window.

  A future maintainer reading only the comment's RBF/eviction justification could reasonably widen the predicate to `>= instantSend.rawValue` and silently reintroduce the restart-classifier bug. Either widen the comment to explicitly call out InstantSend (intentional, and why), or — if the intent really is 'is this spend safely final' — change the threshold to `> mempool`. Recommend the doc-clarification approach so the restart-classifier contract stays intact.
- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift:628-936: [Carried forward] No regression test for the restart + mempool sighting + classifier sequence
  The PR replaces three previously unconditional `isSpent = true` writes with a context-gated path. The behavior change is meaningful in at least four flows: (1) mempool-only sighting then restart then block (the bug being fixed), (2) confirmed sighting in-order, (3) pending-row resolution in `upsertUtxo` when spending tx is mempool vs in-block, (4) `markUtxoSpent` with a missing spending-tx record (now leaves `isSpent` untouched instead of writing `true`). All four are realistic; (4) in particular silently changes a previously-loud signal into a no-op. There is no Swift test in `packages/swift-sdk` exercising the persist→mempool-sighting→restart→block sequence (no `testSelfSendClassificationSurvivesMidFlightRestart` or equivalent). Given the bug being fixed was a production-observed `netAmount` direction flip with inflated value, a regression test asserting `isSpent == true` and correct classification after that sequence would meaningfully harden the fix.
```
</details>

QuantumExplorer (approved):

PR #3177, opened 2026-03-04
feat(dpp): shielded state transitions and Orchard bundle types (Medusa)

QuantumExplorer:
## Summary

Split from #3076 — this PR contains the **DPP layer** of the shielded credit pool (Medusa) feature:

- **5 new state transition types**: Shield (15), ShieldedTransfer (16), Unshield (17), ShieldFromAssetLock (18), ShieldedWithdrawal (19)
- **Orchard bundle integration**: Halo 2 ZK proof types, Sinsemilla commitment tree structures, note encryption/decryption
- **Platform versioning**: New version fields across `rs-platform-version` for shielded operations in Drive, Drive-ABCI, and DPP
- **Consensus errors**: New error variants for shielded transaction validation failures
- **Platform sighash**: Signing infrastructure for shielded state transitions
- **grovedb update**: Aligned to rev `7ecb8465` which adds commitment tree support (`sinsemilla_hash_calls`, new `TreeType` variants, `Option<KeyInfo>`)

Dependent packages (`rs-drive`, `rs-drive-abci`, `wasm-dpp`, `wasm-dpp2`, `rs-sdk-ffi`) include `todo!()` stubs so the workspace compiles. These will be replaced with real implementations in subsequent split PRs.

## Packages Changed

| Package | Scope |
|---|---|
| `rs-dpp` | 103 files — full shielded ST types, Orchard bundles, consensus errors |
| `rs-platform-version` | 47 files — version fields for shielded operations |
| `rs-drive` | Compilation stubs + grovedb version bump |
| `rs-drive-abci` | Compilation stubs for ST processor traits |
| `wasm-dpp` / `wasm-dpp2` | Compilation stubs for new ST variants |
| `rs-sdk-ffi` | Compilation stub for new Element variants |

## Test plan

- [ ] `cargo check --workspace` passes (verified locally)
- [ ] `cargo test -p dpp` passes
- [ ] CI green on Rust checks
- [ ] Dependent PRs (Drive, Drive-ABCI) will add real implementations

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2866, opened 2025-11-22
feat: addresses on the Platform chain

QuantumExplorer:
This PR is made to merge the main bulk of the work into a branch holding the work of multiple contributors.

PR #3782, opened 2026-06-02
fix(platform-version): gate shielded-pool block methods to protocol v12

shumkov:

`drive 4.0.0-beta.1` cannot process **any** protocol-v11 block. A node syncing testnet on this build crash-loops in `ProcessProposal` with:

```
storage: grovedb: path parent layer not found: could not get key 4d for parent [ [hex: 34, str: 4],  ] of subtree: path key not found: key not found in Merk for get from storage: 4d
```

`0x34` = `RootTree::ShieldedBalances` (52); `0x4d` = `MAIN_SHIELDED_CREDIT_POOL_KEY` (`"M"`). The shielded credit pool subtree `[52, "M"]` is created only from protocol **v12** onward (`transition_to_version_12`, or a v12 genesis via init structure v3). On a protocol-v11 state it does not exist.

**Root cause:** protocol v11 and v12 both reference the shared `DRIVE_ABCI_METHOD_VERSIONS_V7`. The Medusa shielded-pool PR (#3177, `b33f8ec75b`) added four shielded block-processing methods to V7 as `Some(0)`:

- `store_nullifiers_to_recent_block_storage`
- `cleanup_recent_block_storage_nullifiers`
- `record_shielded_pool_anchor`
- `prune_shielded_pool_anchors`

Because V7 is shared, this silently turned them on for **v11** too. They run on every block (independent of transactions) and unconditionally read `[52, "M"]`, aborting consensus.

This was **not** introduced by #2920 (which switched v11 from V6→V7 for checkpoints): at release 3.0.1, V7 had no shielded fields, so v11 was fine. The rest of testnet runs `dashpay/drive:3.0.1` and is unaffected — only the lone `4.0.0-beta.1` node broke.

Observed on testnet `hp-masternode-7` at height **354892** — the first block the beta binary processed after it was deployed (the previous binary had committed cleanly up to 354891).

## What was done?

- Restored `DRIVE_ABCI_METHOD_VERSIONS_V7` (the protocol-v11 method set) to its 3.0.1 behavior: the four shielded methods are back to `None`.
- Added `DRIVE_ABCI_METHOD_VERSIONS_V8` (= V7 plus the four shielded methods `Some(0)`) and pointed `v12.rs` at it.
- `v11.rs` is unchanged (still references V7).

Net effect: shielded-pool block processing is gated to protocol **v12+**, where the `[52, "M"]` subtree is guaranteed to exist. The legitimately-v11 parts of V7 (checkpoints, address recent-block-storage from #2866) are untouched.

## How Has This Been Tested?

New unit test `platform-version::shielded_pool_gating_tests` pins the invariant with a red→green transition against this exact bug:

- ✖ **before fix:** `PlatformVersion::get(11).drive_abci.methods.block_end.record_shielded_pool_anchor == Some(0)` → test FAILS (`left: Some(0)`, `right: None`)
- ✔ **after fix:** v11 → `None` for all four methods; v12 → `Some(0)` → PASSES

```
cargo test -p platform-version --lib   # 7 passed, 0 failed
cargo fmt -p platform-version -- --check   # clean
```

Ground truth confirmed against the live network: every other testnet masternode runs `dashpay/drive:3.0.1`, whose `V7` has no shielded fields — this change restores exact v11 parity with them.

## Breaking Changes

None. This is a consensus-relevant version-map correction, but no chain state was ever produced under the buggy v11 rules (a v11 node with the bug crashes before it can commit), and the fix restores parity with the released 3.0.1 network. The shielded pool remains a v12 feature; the v12 upgrade migration (`transition_to_version_12`) is idempotent (`grove_insert_if_not_exists` per node) and creates `[52, "M"]` and its children when v12 activates.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added shielded pool block-processing functionality for Protocol v12.

* **Tests**
  * Added verification tests for shielded pool feature gating to Protocol v12.

* **Chores**
  * Updated version management configuration to support Protocol v12 requirements.

### DIFF
--- a/packages/rs-platform-version/src/version/drive_abci_versions/drive_abci_method_versions/mod.rs
+++ b/packages/rs-platform-version/src/version/drive_abci_versions/drive_abci_method_versions/mod.rs
@@ -7,6 +7,7 @@ pub mod v4;
 pub mod v5;
 pub mod v6;
 pub mod v7;
+pub mod v8;
 
 #[derive(Clone, Debug, Default)]
 pub struct DriveAbciMethodVersions {
--- a/packages/rs-platform-version/src/version/drive_abci_versions/drive_abci_method_versions/v8.rs
+++ b/packages/rs-platform-version/src/version/drive_abci_versions/drive_abci_method_versions/v8.rs
@@ -12,18 +12,18 @@ use crate::version::drive_abci_versions::drive_abci_method_versions::{
     DriveAbciVotingMethodVersions,
 };
 
-// Introduced in Protocol version 11 (3.0.0) for checkpoints.
+// Introduced in Protocol version 12 for the shielded credit pool.
 //
-// The four shielded-pool block-processing methods
-// (`store_nullifiers_to_recent_block_storage`,
-// `cleanup_recent_block_storage_nullifiers`, `record_shielded_pool_anchor`,
-// `prune_shielded_pool_anchors`) are deliberately `None` here: they read the
-// shielded credit pool subtree `[ShieldedBalances (52), "M"]`, which is only
-// created from protocol v12 onward. Activating them on a v11 state opens a
-// subtree that does not exist and panics consensus with
-// "path parent layer not found: could not get key 4d for parent [52]".
-// The shielded activation lives in DRIVE_ABCI_METHOD_VERSIONS_V8 (protocol v12).
-pub const DRIVE_ABCI_METHOD_VERSIONS_V7: DriveAbciMethodVersions = DriveAbciMethodVersions {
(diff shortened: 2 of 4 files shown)


thepastaclaw:
✅ Review complete (commit 36f7f9ed)

thepastaclaw (approved):
## Code Review

Targeted, correct consensus-safety fix. V7 (the method set used by PLATFORM_V11) is restored to its 3.0.1 behavior with the four shielded-pool block-processing methods set to None, and the new V8 struct (used only by PLATFORM_V12) carries the Some(0) activations. V7 vs V8 differ in exactly those four fields, no other protocol version references V8, and the added tests pin the invariant at both v11 and v12.

QuantumExplorer (approved):

PR #3785, opened 2026-06-02
feat(swift-sdk): iOS simluator writes logs to disk

ZocoLini:
This PR makes two things:

  - Exposes a method in platform-wallet-ffi to enable file logging and the example app always calls it
  - Adds a script (get_logs.sh) in the swift-sdk package to be able to extract all the logs written to disk by the iOS Simulators

the logs structure is as follows:
```bash
.
├── build_info.txt
├── dash_sdk
│   └── run.log
├── dash_spv
│   └── run.log
├── grpc
│   └── run.log
├── key_wallet
│   └── run.log
└── platform_wallet
    └── run.log

```

build_info contains the state of the repo at compile time and each folder the logs written by those crates

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * File-based logging capability for configurable log levels with per-component output files.
  * New utility script to extract logs from iOS Simulator app instances with session management.
  * Conditional file logging for iOS Simulator builds with timestamped session directories.
  * Build information tracking including git commit hash and dirty state detection.


thepastaclaw:
✅ Review complete (commit 33b7e997)

packages/swift-sdk/Sources/SwiftDashSDK/Core/Services/SDKLogger.swift
```diff
@@ -29,7 +29,13 @@ public enum LoggingPreferences {
...
         let enableSwiftVerbose: Bool
 
-        SDK.enableLogging(level: .info)
+        // First subscriber install wins, so try files before
+        // falling back to stdout-only.
+        if let sessionRoot = launchLogPaths(),
+           SDK.enableFileLogging(level: .info, sessionRoot: sessionRoot) {
+        } else {
+            SDK.enableLogging(level: .info)
+        }
```
thepastaclaw: **🔴 Blocking: File logging is enabled on physical iOS devices, not just the simulator**

`LoggingPreferences.configure()` calls `SDK.enableFileLogging` unconditionally, and `launchLogPaths()` only checks that a Library directory exists — there is no `#if targetEnvironment(simulator)` guard. As a result, physical iOS devices (and the macOS build) will also start writing per-launch session directories at `<sandbox>/Library/Logs/SwiftDashSDK/<timestamp>/` containing `info`-level logs from `dash_sdk`, `platform_wallet`, `dash_spv`, `key_wallet`, and `grpc`. Two concrete problems on devices: (a) no rotation or cap, so disk usage grows monotonically across the app's lifetime, and (b) `Library/Logs` is included in iTunes/iCloud backups by default, meaning logs containing identity IDs, balances, endpoints, and asset-lock traces can leave the device. The PR title ("iOS simluator writes logs to disk") and the accompanying `extract_simulator_logs.sh` make it clear the feature is intended for simulator debugging. Either gate the file-logging path with `#if targetEnvironment(simulator)` or document and accept the on-device behavior with an explicit retention/backup policy.

packages/rs-platform-wallet-ffi/src/logging.rs
```diff
@@ -0,0 +1,170 @@
...
+        .with(l_grpc)
+        .try_init()
+        .is_err()
+    {
+        return false;
+    }
+
+    if fs::write(path.join("build_info.txt"), build_info_string()).is_err() {
+        return false;
+    }
```
thepastaclaw: **🟡 Suggestion: FFI returns `false` after the global tracing subscriber is already installed**

`tracing_subscriber::registry()...try_init()` at lines 94-102 installs the global subscriber for the lifetime of the process — it cannot be replaced afterwards. The subsequent `fs::write(path.join("build_info.txt"), ...)` at line 107 returning `Err` causes the function to return `false`, but the file subscriber is already active. The docstring (lines 14-18) tells Swift that `false` means "no subscriber was installed by this call", so `SDK.enableFileLogging` propagates `false` to `LoggingPreferences.configure()`, which then falls through to `SDK.enableLogging(level: .info)` — that path calls `dash_sdk_enable_logging`, which silently no-ops because the global is taken. Net result: the documented fallback never runs, the caller has no signal that file logging is in fact active, and a callsite that branches on the return value to pick alternate routing will be misled. Either return `true` once `try_init()` succeeds and downgrade the `build_info.txt` write failure to a `tracing::warn!`, or write `build_info.txt` *before* `try_init()` so the contract holds.

packages/rs-platform-wallet-ffi/build.rs
```diff
@@ -9,6 +10,21 @@ fn main() {
...
+        Some(out) if !out.is_empty() => "1",
+        Some(_) => "0",
+        None => "unknown",
+    };
+    println!("cargo:rustc-env=PLATFORM_WALLET_GIT_DIRTY={dirty}");
+
+    if let Some(workspace_root) = run_git(&["rev-parse", "--show-toplevel"]) {
+        println!("cargo:rerun-if-changed={workspace_root}/.git/HEAD");
+        println!("cargo:rerun-if-changed={workspace_root}/.git/index");
+    }
```
thepastaclaw, line 28: **🟡 Suggestion: Build-info `git-dirty` is monorepo-wide and the rerun hooks are broken in worktrees**

Two issues in the same plumbing that make the metadata baked into `build_info.txt` unreliable:

1. `run_git(&["status", "--porcelain"])` (line 16) has no pathspec and no `--untracked-files=no`, so it scans the entire workspace. Any untracked or modified file in any crate of the monorepo marks this crate's build as dirty. In practice `git-dirty: yes` will be the common case whenever a developer has anything in flight, even when `rs-platform-wallet-ffi/` itself is clean. Scoping the status check to the files this crate's binary actually depends on (e.g. `packages/rs-platform-wallet-ffi`, `packages/rs-platform-wallet`, `packages/key-wallet`) would make the flag meaningful.

2. `cargo:rerun-if-changed={workspace_root}/.git/HEAD` / `/.git/index` (lines 24-25) assumes `.git` is a directory. This repo is commonly checked out as a Git worktree (in fact this verifier is running from `/Users/claw/.openclaw/workspace/worktrees/review-platform-3785`), where `.git` is a text file pointing at the real Git dir and neither `.git/HEAD` nor `.git/index` exists at that path. Cargo therefore never invalidates the build script when the commit or index changes, leaving stale `PLATFORM_WALLET_GIT_COMMIT` / `PLATFORM_WALLET_GIT_DIRTY` baked into the binary across rebuilds. Resolve the actual paths with `git rev-parse --git-path HEAD` / `--git-path index` and emit those instead.

packages/swift-sdk/get_logs.sh
```diff
@@ -0,0 +1,238 @@
...
+# `LoggingPreferences.configure()` (SDKLogger.swift) calls
+# `platform_wallet_enable_file_logging` at app startup, which lays
+# out one file per `tracing` bucket under a session-stamped root:
+#
+#     <sandbox>/Library/Logs/SwiftDashSDK/<YYYY-MM-DDTHH-mm-ss>/
+#         dash-sdk/run.log
+#         platform-wallet/run.log
+#         dash-spv/run.log
+#         key-wallet/run.log
+#         grpc/run.log
```
thepastaclaw, line 15: **💬 Nitpick: Header comment lists hyphenated bucket names but Rust writes underscored names**

The script header advertises `dash-sdk/run.log`, `platform-wallet/run.log`, `dash-spv/run.log`, `key-wallet/run.log`, `grpc/run.log`, but `rs-platform-wallet-ffi/src/logging.rs:40-54` actually writes to `dash_sdk/`, `platform_wallet/`, `dash_spv/`, `key_wallet/`, `grpc/`. The `cp -R` still copies whatever's there, but anyone reading this comment to grep a specific bucket file (or to extend the script with a per-bucket check) will look at the wrong paths. Update the comment to match the real directory names.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet-ffi/src/logging.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    tracing::info!(level = log_level, "file logging enabled");
+    true
+}
```
thepastaclaw, line 113: **💬 Nitpick: Partial filesystem state is left behind when init fails midway**

`open_file` calls `fs::create_dir_all` and `OpenOptions::new().create(true).append(true).open(...)`, so each successful call leaves a directory plus an empty `run.log`. If, say, the 4th `open_file` fails, the first 3 buckets remain as empty dir+file pairs under the session root. Likewise if `try_init` returns `Err` (or `fs::write(build_info.txt)` returns `Err`), all 5 buckets plus the timestamped session directory remain on disk and the function returns `false`. Combined with the lack of any rotation/cleanup, this quietly accumulates orphan session directories under `Library/Logs/SwiftDashSDK/` across launches. Either pre-flight writability of the session root before creating any per-bucket files, or unwind the partial filesystem state on the failure path.

<sub>source: ['claude']</sub>

thepastaclaw (changes requested):
## Code Review

PR adds an optional per-bucket file-logging FFI plus a log extraction script. Verified two notable in-scope issues: (1) the Swift entry point unconditionally enables on-device file logging despite the PR scoping the feature to the simulator, with attendant privacy/retention implications; (2) the Rust FFI's boolean contract is violated when `build_info.txt` write fails after the global subscriber has already been installed, which causes the Swift caller to silently take a now-noop fallback. Several smaller issues exist in the build-info plumbing and extraction-script documentation.

<details>

```

In `packages/swift-sdk/Sources/SwiftDashSDK/Core/Services/SDKLogger.swift`:
- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/Core/Services/SDKLogger.swift:32-38: File logging is enabled on physical iOS devices, not just the simulator
  `LoggingPreferences.configure()` calls `SDK.enableFileLogging` unconditionally, and `launchLogPaths()` only checks that a Library directory exists — there is no `#if targetEnvironment(simulator)` guard. As a result, physical iOS devices (and the macOS build) will also start writing per-launch session directories at `<sandbox>/Library/Logs/SwiftDashSDK/<timestamp>/` containing `info`-level logs from `dash_sdk`, `platform_wallet`, `dash_spv`, `key_wallet`, and `grpc`. Two concrete problems on devices: (a) no rotation or cap, so disk usage grows monotonically across the app's lifetime, and (b) `Library/Logs` is included in iTunes/iCloud backups by default, meaning logs containing identity IDs, balances, endpoints, and asset-lock traces can leave the device. The PR title ("iOS simluator writes logs to disk") and the accompanying `extract_simulator_logs.sh` make it clear the feature is intended for simulator debugging. Either gate the file-logging path with `#if targetEnvironment(simulator)` or document and accept the on-device behavior with an explicit retention/backup policy.

In `packages/rs-platform-wallet-ffi/src/logging.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/logging.rs:94-109: FFI returns `false` after the global tracing subscriber is already installed
  `tracing_subscriber::registry()...try_init()` at lines 94-102 installs the global subscriber for the lifetime of the process — it cannot be replaced afterwards. The subsequent `fs::write(path.join("build_info.txt"), ...)` at line 107 returning `Err` causes the function to return `false`, but the file subscriber is already active. The docstring (lines 14-18) tells Swift that `false` means "no subscriber was installed by this call", so `SDK.enableFileLogging` propagates `false` to `LoggingPreferences.configure()`, which then falls through to `SDK.enableLogging(level: .info)` — that path calls `dash_sdk_enable_logging`, which silently no-ops because the global is taken. Net result: the documented fallback never runs, the caller has no signal that file logging is in fact active, and a callsite that branches on the return value to pick alternate routing will be misled. Either return `true` once `try_init()` succeeds and downgrade the `build_info.txt` write failure to a `tracing::warn!`, or write `build_info.txt` *before* `try_init()` so the contract holds.

In `packages/rs-platform-wallet-ffi/build.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/build.rs:13-26: Build-info `git-dirty` is monorepo-wide and the rerun hooks are broken in worktrees
  Two issues in the same plumbing that make the metadata baked into `build_info.txt` unreliable:

  1. `run_git(&["status", "--porcelain"])` (line 16) has no pathspec and no `--untracked-files=no`, so it scans the entire workspace. Any untracked or modified file in any crate of the monorepo marks this crate's build as dirty. In practice `git-dirty: yes` will be the common case whenever a developer has anything in flight, even when `rs-platform-wallet-ffi/` itself is clean. Scoping the status check to the files this crate's binary actually depends on (e.g. `packages/rs-platform-wallet-ffi`, `packages/rs-platform-wallet`, `packages/key-wallet`) would make the flag meaningful.

  2. `cargo:rerun-if-changed={workspace_root}/.git/HEAD` / `/.git/index` (lines 24-25) assumes `.git` is a directory. This repo is commonly checked out as a Git worktree (in fact this verifier is running from `/Users/claw/.openclaw/workspace/worktrees/review-platform-3785`), where `.git` is a text file pointing at the real Git dir and neither `.git/HEAD` nor `.git/index` exists at that path. Cargo therefore never invalidates the build script when the commit or index changes, leaving stale `PLATFORM_WALLET_GIT_COMMIT` / `PLATFORM_WALLET_GIT_DIRTY` baked into the binary across rebuilds. Resolve the actual paths with `git rev-parse --git-path HEAD` / `--git-path index` and emit those instead.
```
</details>

packages/rs-platform-wallet-ffi/src/logging.rs
thepastaclaw, line 113: **🟡 Suggestion: `platform_wallet_enable_file_logging` returns `false` after the global subscriber is already installed**

`tracing_subscriber::registry()...try_init()` at line 101 installs the process-wide subscriber for the lifetime of the program — once it succeeds, no subsequent `try_init` call (including the stdout fallback `SDK.enableLogging` from Swift) can register another subscriber. If the subsequent `fs::write(path.join("build_info.txt"), ...)` at line 107 fails (read-only volume, sandbox quirk, transient FS error), the function still returns `false`. The Swift caller then takes the `else` branch in `LoggingPreferences.configure()` and calls `SDK.enableLogging(level: .info)`, which is now a silent no-op because the registry subscriber is already global. Result: file logging is in fact active, but the caller has been told it isn't — confusing both for diagnostics and for future attempts to attach extra layers.

Either write `build_info.txt` *before* `try_init` (so a write failure is observable and prevents installation), or treat the write as best-effort: log a warning and still return `true`.

packages/rs-platform-wallet-ffi/build.rs
thepastaclaw, line 28: **🟡 Suggestion: Git-dirty detection is monorepo-wide and rerun hints reference paths that don't exist in worktrees**

Two related problems whose outputs are baked into `build_info.txt` and shipped with every session:

1. `run_git(&["status", "--porcelain"])` (line 16) runs with no pathspec and no `--untracked-files=no`, so any unrelated edit anywhere in the ~44-crate monorepo — including untracked files outside this crate — flips `PLATFORM_WALLET_GIT_DIRTY` to `1`. Every iOS log session then carries the `WARNING: this build had uncommitted changes` blurb even when `packages/rs-platform-wallet-ffi/` and its deps are clean. Scope the check to this crate (and its workspace dependencies if needed): `git status --porcelain --untracked-files=no -- <crate-dir>`.

2. `cargo:rerun-if-changed={workspace_root}/.git/HEAD` and `.git/index` (lines 24-25) assume a non-worktree checkout. In a git worktree (this very review is running in one: `/Users/claw/.openclaw/workspace/worktrees/review-platform-3785`), the worktree root's `.git` is a *file* pointing into `<main>/.git/worktrees/<name>/`, so `HEAD`/`index` live there, not at the workspace root. The hints point to non-existent paths and Cargo will not invalidate the build info when HEAD advances or branches switch inside the worktree — leading to stale `GIT_COMMIT` values in the shipped binary. Use `git rev-parse --git-path HEAD` / `--git-path index` (or `--git-dir`) to resolve the real paths.

thepastaclaw (commented):
## Code Review

Re-verified prior review at head 479795ed. All five prior findings remain valid: the latest delta only refactored `dash_sdk_enable_logging` in `rs-sdk-ffi/src/lib.rs` and did not touch any of the previously flagged code. One additional convergent finding was introduced by the refactor: the new in-process EnvFilter string dropped `dash_spv` and `key_wallet` directives that the sibling `broad_env_filter` (and the prior `build_env_filter` helper) still carry, causing SPV and key-wallet traces to fall through to EnvFilter's default `warn` level on the stdout-only path.

🟡 3 suggestion(s)

<details>

```

In `packages/rs-sdk-ffi/src/lib.rs`:
- [SUGGESTION] packages/rs-sdk-ffi/src/lib.rs:112-117: Refactor of `dash_sdk_enable_logging` dropped `dash_spv` and `key_wallet` directives
  The 14b641bb..479795ed delta inlined the prior `build_env_filter` helper into `dash_sdk_enable_logging` (a legitimate fix that avoids the `env::set_var("RUST_LOG", ...)` data-race documented in the surrounding comment). However, the new `filter_string` no longer lists `dash_spv={log_level},key_wallet={log_level}` — both of which were present in the previous implementation and are still present in the sibling `broad_env_filter` at `packages/rs-platform-wallet-ffi/src/logging.rs:127-138`. Consequences:

  - On the stdout fallback path (when `enable_file_logging` returns false, when `targetEnvironment(simulator)` is gated off after the blocking finding is applied, on macOS host runs, in unit-test harnesses, and for any FFI consumer that only calls `dash_sdk_enable_logging`), SPV and key-wallet traces will silently fall through to EnvFilter's default `warn` level instead of honoring the requested level.
  - The two filter sets now disagree about which crates' logs the user gets, which is confusing for downstream consumers and likely unintentional given the inlined comment only justifies adding `platform_wallet`/`platform_wallet_ffi`, not removing the other two.

  Restore the dropped directives so the stdout filter matches `broad_env_filter`.

In `packages/rs-platform-wallet-ffi/src/logging.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/logging.rs:94-113: `platform_wallet_enable_file_logging` returns `false` after the global subscriber is already installed
  `tracing_subscriber::registry()...try_init()` at line 101 installs the process-wide subscriber for the lifetime of the program — once it succeeds, no subsequent `try_init` call (including the stdout fallback `SDK.enableLogging` from Swift) can register another subscriber. If the subsequent `fs::write(path.join("build_info.txt"), ...)` at line 107 fails (read-only volume, sandbox quirk, transient FS error), the function still returns `false`. The Swift caller then takes the `else` branch in `LoggingPreferences.configure()` and calls `SDK.enableLogging(level: .info)`, which is now a silent no-op because the registry subscriber is already global. Result: file logging is in fact active, but the caller has been told it isn't — confusing both for diagnostics and for future attempts to attach extra layers.

  Either write `build_info.txt` *before* `try_init` (so a write failure is observable and prevents installation), or treat the write as best-effort: log a warning and still return `true`.

In `packages/rs-platform-wallet-ffi/build.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/build.rs:13-26: Git-dirty detection is monorepo-wide and rerun hints reference paths that don't exist in worktrees
  Two related problems whose outputs are baked into `build_info.txt` and shipped with every session:

  1. `run_git(&["status", "--porcelain"])` (line 16) runs with no pathspec and no `--untracked-files=no`, so any unrelated edit anywhere in the ~44-crate monorepo — including untracked files outside this crate — flips `PLATFORM_WALLET_GIT_DIRTY` to `1`. Every iOS log session then carries the `WARNING: this build had uncommitted changes` blurb even when `packages/rs-platform-wallet-ffi/` and its deps are clean. Scope the check to this crate (and its workspace dependencies if needed): `git status --porcelain --untracked-files=no -- <crate-dir>`.

  2. `cargo:rerun-if-changed={workspace_root}/.git/HEAD` and `.git/index` (lines 24-25) assume a non-worktree checkout. In a git worktree (this very review is running in one: `/Users/claw/.openclaw/workspace/worktrees/review-platform-3785`), the worktree root's `.git` is a *file* pointing into `<main>/.git/worktrees/<name>/`, so `HEAD`/`index` live there, not at the workspace root. The hints point to non-existent paths and Cargo will not invalidate the build info when HEAD advances or branches switch inside the worktree — leading to stale `GIT_COMMIT` values in the shipped binary. Use `git rev-parse --git-path HEAD` / `--git-path index` (or `--git-dir`) to resolve the real paths.
```
</details>

packages/rs-platform-wallet-ffi/src/logging.rs
thepastaclaw: Resolved in this update — _Partial filesystem state is left behind when init fails midway_ no longer present.

packages/rs-platform-wallet-ffi/src/logging.rs
thepastaclaw: Resolved in this update — _FFI returns `false` after the global tracing subscriber is already installed_ no longer present.

packages/rs-platform-wallet-ffi/build.rs
thepastaclaw: Resolved in this update — _Build-info `git-dirty` is monorepo-wide and the rerun hooks are broken in worktrees_ no longer present.

packages/swift-sdk/get_logs.sh
thepastaclaw: Resolved in this update — _Header comment lists hyphenated bucket names but Rust writes underscored names_ no longer present.

packages/swift-sdk/Sources/SwiftDashSDK/Core/Services/SDKLogger.swift
thepastaclaw: Resolved in this update — _File logging is enabled on physical iOS devices, not just the simulator_ no longer present.

packages/rs-platform-wallet-ffi/src/logging.rs
```diff
@@ -0,0 +1,170 @@
...
+    let Some(f_grpc) = open_file(path.join("grpc").join("run.log")) else {
+        return false;
+    };
+
+    let l_sdk = tracing_subscriber::fmt::layer()
+        .with_writer(Mutex::new(f_sdk))
+        .with_ansi(false)
+        .with_filter(tracing_subscriber::EnvFilter::new(format!(
+            "dash_sdk={log_level},rs_sdk_ffi={log_level}"
+        )));
```
thepastaclaw, line 61: **🟡 Suggestion: dash_sdk bucket file filter omits rs_sdk crate**

The `l_sdk` layer routes records to `dash_sdk/run.log` and its EnvFilter only enables `dash_sdk={log_level},rs_sdk_ffi={log_level}`. The shared `broad_env_filter` in the same file (line 129) explicitly treats `rs_sdk` as part of the SDK surface, and so does the stdout fallback in rs-sdk-ffi. Anyone collecting `dash_sdk/run.log` via `get_logs.sh` for a simulator triage will be missing every `rs_sdk` event even though file logging reported success.

```suggestion
    let l_sdk = tracing_subscriber::fmt::layer()
        .with_writer(Mutex::new(f_sdk))
        .with_ansi(false)
        .with_filter(tracing_subscriber::EnvFilter::new(format!(
            "dash_sdk={log_level},rs_sdk={log_level},rs_sdk_ffi={log_level}"
        )));
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/build.rs
```diff
@@ -9,6 +10,23 @@ fn main() {
...
     println!("cargo:rerun-if-changed=src/");
 
+    let commit = run_git(&["rev-parse", "HEAD"]).unwrap_or_else(|| "unknown".to_string());
+    println!("cargo:rustc-env=PLATFORM_WALLET_GIT_COMMIT={commit}");
+
+    let dirty = match run_git(&["status", "--porcelain"]) {
+        Some(out) if !out.is_empty() => "1",
+        Some(_) => "0",
+        None => "unknown",
+    };
```
thepastaclaw, line 20: **🟡 Suggestion: Git-dirty flag reflects monorepo-wide status, not the wallet crate**

`run_git(&["status", "--porcelain"])` runs with no pathspec and default untracked handling, so any unrelated edit anywhere in the worktree flips `PLATFORM_WALLET_GIT_DIRTY` to 1. Simulator log bundles produced by this PR will then claim the wallet build was dirty even when only an unrelated package was touched, making `build_info.txt` actively misleading for the debugging workflow this PR is adding.

```suggestion
    let dirty = match run_git(&[
        "status",
        "--porcelain",
        "--untracked-files=no",
        "--",
        "packages/rs-platform-wallet-ffi",
    ]) {
        Some(out) if !out.is_empty() => "1",
        Some(_) => "0",
        None => "unknown",
    };
    println!("cargo:rustc-env=PLATFORM_WALLET_GIT_DIRTY={dirty}");
```

packages/rs-platform-wallet-ffi/src/logging.rs
```diff
@@ -0,0 +1,170 @@
...
+        .with(l_grpc)
+        .try_init()
+        .is_err()
+    {
+        return false;
+    }
+
+    tracing::info!(level = log_level, "file logging enabled");
+    true
+}
```
thepastaclaw, line 113: **💬 Nitpick: Failed file-log init leaves a partial session directory behind**

`open_file` creates parent directories before opening each `run.log`. If a later bucket fails (or `try_init` returns Err because a subscriber is already installed), the FFI returns `false` and Swift falls back to `SDK.enableLogging`, but the timestamped session root is left on disk with empty `run.log` files (and possibly `build_info.txt`). `get_logs.sh` will happily list those orphan sessions as if they were real captures. Either best-effort `fs::remove_dir_all(path)` on the failure paths, or document that the caller owns cleanup.

thepastaclaw (commented):
## Code Review

Carried-forward prior findings: file-logging filter `l_sdk` and the stdout fallback in rs-sdk-ffi both still omit components, `build.rs` still reports monorepo-wide dirtiness, the new `get_logs.sh` header still documents hyphenated bucket names that Rust never creates, and a failed `enable_file_logging` still leaves a partial session directory behind. Fixed in this delta: the simulator-only gate in Swift, and writing `build_info.txt` before `try_init`. One new latest-delta-adjacent finding: the per-file `dash_sdk` bucket filter is missing the `rs_sdk` directive that the broad filter and the new Swift API contract both imply.

<details>

```

In `packages/rs-sdk-ffi/src/lib.rs`:
- [SUGGESTION] packages/rs-sdk-ffi/src/lib.rs:112-117: Stdout fallback filter omits dash_spv and key_wallet
  `dash_sdk_enable_logging` is the fallback path used on every non-simulator build and whenever `platform_wallet_enable_file_logging` returns false on the simulator (see SDKLogger.swift:33-41). Its filter string lists dash_sdk/rs_sdk/rs_sdk_ffi/platform_wallet/platform_wallet_ffi/dapi_grpc/h2/tower/hyper/tonic but not dash_spv or key_wallet, while the sibling file-logging path (`broad_env_filter` in platform-wallet-ffi/src/logging.rs:127-134) does include them. Result: SPV-sync and key-wallet traces silently fall through to EnvFilter's default (`warn`) on every device build, regressing observability that the file-logging path is careful to preserve.

In `packages/rs-platform-wallet-ffi/src/logging.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/logging.rs:56-61: dash_sdk bucket file filter omits rs_sdk crate
  The `l_sdk` layer routes records to `dash_sdk/run.log` and its EnvFilter only enables `dash_sdk={log_level},rs_sdk_ffi={log_level}`. The shared `broad_env_filter` in the same file (line 129) explicitly treats `rs_sdk` as part of the SDK surface, and so does the stdout fallback in rs-sdk-ffi. Anyone collecting `dash_sdk/run.log` via `get_logs.sh` for a simulator triage will be missing every `rs_sdk` event even though file logging reported success.

In `packages/rs-platform-wallet-ffi/build.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/build.rs:16-20: Git-dirty flag reflects monorepo-wide status, not the wallet crate
  `run_git(&["status", "--porcelain"])` runs with no pathspec and default untracked handling, so any unrelated edit anywhere in the worktree flips `PLATFORM_WALLET_GIT_DIRTY` to 1. Simulator log bundles produced by this PR will then claim the wallet build was dirty even when only an unrelated package was touched, making `build_info.txt` actively misleading for the debugging workflow this PR is adding.
```
</details>

packages/rs-platform-wallet-ffi/build.rs
```diff
@@ -9,6 +10,23 @@ fn main() {
...
+    println!("cargo:rustc-env=PLATFORM_WALLET_GIT_DIRTY={dirty}");
```
thepastaclaw, line 21: **🟡 Suggestion: Dirty-state detection is monorepo-wide and includes untracked files**

`run_git(&["status", "--porcelain"])` has no pathspec and no `--untracked-files=no`, so any edit anywhere in the ~44-crate monorepo — including stray untracked files such as the `.review/` artifacts review tooling creates in this very worktree — flips `PLATFORM_WALLET_GIT_DIRTY` to `1`. The `build_info.txt` emitted into each simulator session then claims the wallet binary is dirty (and appends the 'commit hash above does NOT fully describe the source state' warning) even when nothing in the wallet crate or its dependencies actually changed. The rerun hints were correctly tightened to use `git rev-parse --git-path` in this delta; the dirty probe should be scoped to match. At minimum exclude untracked files; ideally restrict to paths the wallet crate actually depends on.

```suggestion
    let dirty = match run_git(&["status", "--porcelain", "--untracked-files=no"]) {
        Some(out) if !out.is_empty() => "1",
        Some(_) => "0",
        None => "unknown",
    };
    println!("cargo:rustc-env=PLATFORM_WALLET_GIT_DIRTY={dirty}");
```

packages/swift-sdk/get_logs.sh
thepastaclaw, line 15: **💬 Nitpick: Script header documents hyphenated bucket directories that don't exist on disk**

`enable_file_logging` creates `dash_sdk/`, `platform_wallet/`, `dash_spv/`, `key_wallet/`, and `grpc/` under the session root (`packages/rs-platform-wallet-ffi/src/logging.rs:40-54`), but the script header advertises hyphenated names (`dash-sdk/`, `platform-wallet/`, `dash-spv/`, `key-wallet/`). The runtime behavior is correct because the script copies the session directory verbatim, but the header — which `--help` prints by `sed -n '3,30p'` — is the first thing a developer sees and does not match the on-disk layout.

```suggestion
#     <sandbox>/Library/Logs/SwiftDashSDK/<YYYY-MM-DDTHH-mm-ss>/
#         dash_sdk/run.log
#         platform_wallet/run.log
#         dash_spv/run.log
#         key_wallet/run.log
#         grpc/run.log
```

packages/rs-platform-wallet-ffi/src/logging.rs
thepastaclaw, line 113: **💬 Nitpick: Partial bucket directories remain on disk when initialization fails midway**

`enable_file_logging` opens the five bucket files sequentially via `open_file`, each `create_dir_all`-ing its parent before opening `run.log`, then writes `build_info.txt`, then calls `try_init`. If any later step fails the function returns `false` and Swift falls back to stdout logging, but the session root already contains whichever bucket directories and empty `run.log` files were created before the failure. `get_logs.sh` enumerates sessions purely by timestamp directory name, so abandoned partial sessions become selectable from the picker and confuse session ↔ run mapping. Either pre-validate all paths before opening any files, clean up the session root on failure, or document partial artifacts as expected best-effort behavior.

thepastaclaw (commented):
## Code Review

Carried forward review of PR #3785 at 33b7e997. The two highest-impact prior issues are resolved: file logging is now correctly gated to the iOS Simulator (prior-1), and build_info.txt is written before try_init so the boolean return value no longer misreports state after the subscriber is installed (prior-3). Four prior findings still reproduce on the current head and are surfaced here; one of them (the stdout fallback filter omitting dash_spv/key_wallet) is now strictly worse because the new simulator gate routes all device/macOS callers exclusively through the narrower filter. No new in-scope issues introduced by the delta beyond these carried-forward items.

<details>

```

In `packages/rs-sdk-ffi/src/lib.rs`:
- [SUGGESTION] packages/rs-sdk-ffi/src/lib.rs:112-117: Stdout-fallback EnvFilter omits dash_spv and key_wallet, diverging from file-logging path
  `dash_sdk_enable_logging`'s inlined filter string lists dash_sdk, rs_sdk, rs_sdk_ffi, platform_wallet, platform_wallet_ffi, dapi_grpc, h2, tower, hyper, and tonic — but not dash_spv or key_wallet. The companion file-logging path (`broad_env_filter` in `packages/rs-platform-wallet-ffi/src/logging.rs:127-138`, plus the per-bucket layers for dash_spv and key_wallet at lines 70-82) does include both. After the prior-1 fix introduced `#if targetEnvironment(simulator)` in `SDKLogger.swift:33-41`, every non-simulator Swift caller — physical iOS devices and the fallback path when file logging fails — now routes exclusively through `dash_sdk_enable_logging`. The result is that SPV and key-wallet info-level traces silently fall back to EnvFilter's `warn` default on devices, while the same call on the simulator captures them. The two FFI entry points should expose the same logical coverage; the prior refactor that inlined this filter clearly intended to.

In `packages/rs-platform-wallet-ffi/build.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/build.rs:16-21: Dirty-state detection is monorepo-wide and includes untracked files
  `run_git(&["status", "--porcelain"])` has no pathspec and no `--untracked-files=no`, so any edit anywhere in the ~44-crate monorepo — including stray untracked files such as the `.review/` artifacts review tooling creates in this very worktree — flips `PLATFORM_WALLET_GIT_DIRTY` to `1`. The `build_info.txt` emitted into each simulator session then claims the wallet binary is dirty (and appends the 'commit hash above does NOT fully describe the source state' warning) even when nothing in the wallet crate or its dependencies actually changed. The rerun hints were correctly tightened to use `git rev-parse --git-path` in this delta; the dirty probe should be scoped to match. At minimum exclude untracked files; ideally restrict to paths the wallet crate actually depends on.
```
</details>

packages/rs-platform-wallet-ffi/src/logging.rs
thepastaclaw: Resolved in this update — _dash_sdk bucket file filter omits rs_sdk crate_ no longer present.

packages/rs-platform-wallet-ffi/src/logging.rs
thepastaclaw: Correction: this was **not** resolved in the 33b7e997 review. The partial-session-directory issue is still present and was carried forward in review https://github.com/dashpay/platform/pull/3785#pullrequestreview-4414073659.

<sub>This replaces an incorrect auto-resolved reconciliation reply from the review system.</sub>

packages/rs-platform-wallet-ffi/build.rs
thepastaclaw: Correction: this was **not** resolved in the 33b7e997 review. The monorepo-wide dirty-state issue is still present and was carried forward in review https://github.com/dashpay/platform/pull/3785#pullrequestreview-4414073659.

<sub>This replaces an incorrect auto-resolved reconciliation reply from the review system.</sub>

QuantumExplorer (approved):

PR #759, opened 2023-01-27
fix(dashmate): migrations fixes

pshenmic:
## Issue being fixed or feature implemented

Migration fixes for pulls #752 and #755

## What was done?

* Added missing migration for dapi envoy grpc port
* Fixed invalid config name pointer

## How Has This Been Tested?

Locally + devnet, by manipulating local and dashmate versions and testing the result

## Breaking Changes

No

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added or updated relevant unit/integration/functional/e2e tests
- [x] I have made corresponding changes to the documentation

**For repository code-owners and collaborators only**
- [x] I have assigned this pull request to a milestone

PR #758, opened 2023-01-26
feat: varint protocol version

QuantumExplorer:
## Issue being fixed or feature implemented
We used to using a u32 prepended to each message for the version. This PR changes this to a varint. This is to save space.

## Checklist:

- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have added or updated relevant unit/integration/functional/e2e tests
- [ ] I have made corresponding changes to the documentation

**For repository code-owners and collaborators only**
- [ ] I have assigned this pull request to a milestone

PR #752, opened 2023-01-24
feat(dapi)!: use single envoy port for all connections

shuplenkov:
## Issue being fixed or feature implemented

Replace two ports DAPI addresses with single port ones

## What was done?

Changes envoy config and dapi client to introduce single port support

## How Has This Been Tested?

With tests

## Breaking Changes

Will not support double port urls

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added or updated relevant unit/integration/functional/e2e tests
- [x] I have made corresponding changes to the documentation

**For repository code-owners and collaborators only**
- [x] I have assigned this pull request to a milestone

PR #3789, opened 2026-06-04
fix: grovedb incompatibilty issues

shumkov:
## What

Update grovedb to current `develop` (`40521b83`) and fix all resulting backward-compat breaks, so platform tracks the latest grovedb. The rev brings:

- **dashpay/grovedb#759** — *version-gated* `add_element_on_transaction`: `GROVE_V1`/`GROVE_V2` (protocol ≤ v11) write `CountSumTree`/`ProvableCountTree`/`ProvableCountSumTree` as plain values (`Op::Put`); `GROVE_V3` (protocol v12) uses the layered subtree. This is the **protocol-v11 consensus fix** (below).
- **dashpay/grovedb#758** — migrates the commitment tree's `orchard` dependency 0.13.1 → 0.14.0 (an Orchard **circuit soundness fix**). Forces a small orchard 0.14 migration in platform's shielded verifier.
- **dashpay/grovedb#760** — re-exports `ProofSizeEnforcement` + `ActionFromPartsError` from `grovedb-commitment-tree`, so platform needs **no direct `orchard` dependency**.

## Issue 1 — protocol-v11 consensus divergence (the urgent fix)

`drive 4.0.0-beta.2` had grovedb `a18f7929`, where grovedb #752 changed the non-batch insert dispatch so the v11 address-pool trees (`empty_provable_count_sum_tree` / `empty_count_sum_tree`, created by `transition_to_version_11`) were written as layered subtrees (`Op::PutLayeredReference`) instead of plain values (`Op::Put`), changing the parent `value_hash` and thus the app-hash. A from-genesis beta.2 node **got stuck at the v11 activation block** (testnet height **245,344**), computing `98DD9B…` where the canonical header is `29B639…`. grovedb #759 restores `Op::Put` for the v11 grove version.

**Validation:** built the fixed image, deployed to testnet **hp-masternode-7**, replayed 245,344 from the saved 245,343 state → `run_block_proposal(245344)` = **`29B639…`** (canonical; beta.2 produced `98DD9B…`); node then synced to the network tip and tracks live blocks **in lockstep with the canonical 3.0.1 masternodes** (same height, 0 errors).

Regression test `provable_count_sum_tree_insert_preserves_v11_consensus_root` (rs-drive) pins **both** sides of the #759 version gate: v11/`GROVE_V2` → canonical `Op::Put` golden roots (incl. the `empty_sum_tree` control on `GOLDEN_1`); v12/`GROVE_V3` → a different (layered) root. Green on the new rev.

## Issue 2 — orchard 0.14 backward-compat (shielded verifier)

The orchard 0.14 bump changed the bundle-reconstruction API. The migration is confined to **one file** — the consensus verifier `rs-drive-abci/.../shielded_common/mod.rs` — and platform carries **no direct `orchard` dependency**: with grovedb #760, `ProofSizeEnforcement` and `ActionFromPartsError` are imported from `grovedb_commitment_tree` alongside the other orchard types it already re-exports. All other shielded code (rs-dpp builder, rs-platform-wallet, wasm-dpp2) compiles unchanged.

Changes, preserving verification semantics:
- `Action::from_parts` now returns `Result` (was `Option`). The `IdentityRk` rejection keeps its original message; the **new `InvalidEpk` rejection is kept** (orchard 0.14's circuit-soundness reject); `ActionFromPartsError` is `#[non_exhaustive]`, so unknown future variants default to **reject**, never accept.
- `Bundle::from_parts` → `Bundle::try_from_parts(.., ProofSizeEnforcement::Strict)`. **`Strict`** rejects any proof whose byte-length is not canonical for the action count (`NonCanonicalProofSize`) — the sound, anti-malleability policy. v12 shielded is not yet activated on the canonical chain, so there is no live acceptance set to preserve; we set the strict policy from the start rather than inheriting orchard 0.13's missing proof-size check.
- Per-action spend-auth signatures stay attached inline to each `Action`, so action↔signature pairing/ordering is preserved structurally.

**Validation (orchard 0.14):**
- `shield_from_asset_lock` (+ transfer/withdrawal/unshield) real round-trips: valid canonical proofs **accepted** under `Strict`; mutated value-balance, zeroed spend-auth/binding sigs, and tampered proofs **rejected** (145 shielded tests, 0 failed).
- New `from_parts` rejection tests: `test_identity_rk_returns_error` (`IdentityRk`) and `test_identity_epk_returns_invalid_epk_error` (`InvalidEpk`) pin both arms so a future refactor cannot silently flip a soundness reject to accept.

> **Consensus-review note.** Two intended acceptance *tightenings* vs orchard 0.13, both safe because v12 shielded is not yet activated (no live acceptance set to preserve): (1) orchard 0.14's new identity/undecodable-`epk` rejection (#758's soundness fix), and (2) `ProofSizeEnforcement::Strict` rejecting non-canonical proof sizes. Flagging for conscious sign-off.

## Root-cause method (v11)

Replayed block 245,344 from the saved 245,343 rocksdb under both grovedb revs with per-grovedb-op root logging; the divergence localized to the single insert of `empty_provable_count_sum_tree` at `[AddressBalances, CLEAR_ADDRESS_POOL]`. Details in dashpay/grovedb#759 and `.claude/skills/debug-wrong-app-hash`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced validation of shielded transactions with stricter enforcement of security requirements and consensus rules to prevent processing of invalid transactions.

* **Tests**
  * Added regression test coverage for shielded transaction validation to ensure correct behavior across protocol versions.

* **Chores**
  * Updated multiple package dependencies to newer versions for improved stability and performance.

### DIFF
--- a/packages/rs-drive-abci/Cargo.toml
+++ b/packages/rs-drive-abci/Cargo.toml
@@ -82,14 +82,7 @@ derive_more = { version = "1.0", features = ["from", "deref", "deref_mut"] }
 async-trait = "0.1.77"
 console-subscriber = { version = "0.4", optional = true }
 bls-signatures = { git = "https://github.com/dashpay/bls-signatures", rev = "0842b17583888e8f46c252a4ee84cdfd58e0546f", optional = true }
-grovedb-commitment-tree = { git = "https://github.com/dashpay/grovedb", rev = "41786daa51495743fc5d7223065d848c36b481f0" }
-# orchard 0.14 made the only constructor for an `Authorized` bundle
-# (`Bundle::try_from_parts`) require a `ProofSizeEnforcement` argument, which
-# `grovedb-commitment-tree` does not re-export. The shielded verifier needs to
-# name that enum to reconstruct the bundle, so orchard is pinned here to the
-# exact same git tag that `grovedb-commitment-tree` resolves to (cargo unifies
-# this to a single orchard build; no second copy is introduced).
-orchard = { git = "https://github.com/dashpay/orchard.git", tag = "dashified-0.14.0", features = ["circuit"] }
+grovedb-commitment-tree = { git = "https://github.com/dashpay/grovedb", rev = "40521b831ca0792e4d63f4d8949c45f07517641e" }
 nonempty = "0.11"
 # Shielded-pool snapshot needs raw RocksDB SstFileWriter + ingest_external_file_cf
 # bindings, and blake3 for the snapshot-file checksum.
@@ -114,7 +107,7 @@ dpp = { path = "../rs-dpp", default-features = false, features = [
 drive = { path = "../rs-drive", features = ["fixtures-and-mocks"] }
 drive-proof-verifier = { path = "../rs-drive-proof-verifier" }
 strategy-tests = { path = "../strategy-tests" }
-grovedb-commitment-tree = { git = "https://github.com/dashpay/grovedb", rev = "41786daa51495743fc5d7223065d848c36b481f0", features = ["client"] }
+grovedb-commitment-tree = { git = "https://github.com/dashpay/grovedb", rev = "40521b831ca0792e4d63f4d8949c45f07517641e", features = ["client"] }
 assert_matches = "1.5.0"
 drive-abci = { path = ".", features = ["testing-config", "mocks", "shielded_test_data"] }
 bls-signatures = { git = "https://github.com/dashpay/bls-signatures", rev = "0842b17583888e8f46c252a4ee84cdfd58e0546f" }
@@ -128,8 +121,8 @@ integer-encoding = { version = "4.0.0" }
 
 # For dump_only_default_and_aux_cfs_under_shielded_subtree_prefix — same
 # subtree-prefix algorithm grovedb uses internally.
-grovedb-path = { git = "https://github.com/dashpay/grovedb", rev = "41786daa51495743fc5d7223065d848c36b481f0" }
-grovedb-storage = { git = "https://github.com/dashpay/grovedb", rev = "41786daa51495743fc5d7223065d848c36b481f0" }
+grovedb-path = { git = "https://github.com/dashpay/grovedb", rev = "40521b831ca0792e4d63f4d8949c45f07517641e" }
+grovedb-storage = { git = "https://github.com/dashpay/grovedb", rev = "40521b831ca0792e4d63f4d8949c45f07517641e" }
 
 [features]
 default = ["bls-signatures"]
--- a/packages/rs-drive-abci/Cargo.toml
+++ b/packages/rs-drive-abci/Cargo.toml
@@ -82,7 +82,7 @@ derive_more = { version = "1.0", features = ["from", "deref", "deref_mut"] }
 async-trait = "0.1.77"
 console-subscriber = { version = "0.4", optional = true }
 bls-signatures = { git = "https://github.com/dashpay/bls-signatures", rev = "0842b17583888e8f46c252a4ee84cdfd58e0546f", optional = true }
-grovedb-commitment-tree = { git = "https://github.com/dashpay/grovedb", rev = "a18f7929460ef9c5d814f61ff84d8805b2a1761b" }
+grovedb-commitment-tree = { git = "https://github.com/dashpay/grovedb", rev = "e5a904d70b2b7cdf078abcf47a51f076d8d468a4" }
 nonempty = "0.11"
 # Shielded-pool snapshot needs raw RocksDB SstFileWriter + ingest_external_file_cf
 # bindings, and blake3 for the snapshot-file checksum.
@@ -107,7 +107,7 @@ dpp = { path = "../rs-dpp", default-features = false, features = [
 drive = { path = "../rs-drive", features = ["fixtures-and-mocks"] }
 drive-proof-verifier = { path = "../rs-drive-proof-verifier" }
 strategy-tests = { path = "../strategy-tests" }
-grovedb-commitment-tree = { git = "https://github.com/dashpay/grovedb", rev = "a18f7929460ef9c5d814f61ff84d8805b2a1761b", features = ["client"] }
+grovedb-commitment-tree = { git = "https://github.com/dashpay/grovedb", rev = "e5a904d70b2b7cdf078abcf47a51f076d8d468a4", features = ["client"] }
 assert_matches = "1.5.0"
 drive-abci = { path = ".", features = ["testing-config", "mocks", "shielded_test_data"] }
 bls-signatures = { git = "https://github.com/dashpay/bls-signatures", rev = "0842b17583888e8f46c252a4ee84cdfd58e0546f" }
@@ -121,8 +121,8 @@ integer-encoding = { version = "4.0.0" }
 
 # For dump_only_default_and_aux_cfs_under_shielded_subtree_prefix — same
 # subtree-prefix algorithm grovedb uses internally.
-grovedb-path = { git = "https://github.com/dashpay/grovedb", rev = "a18f7929460ef9c5d814f61ff84d8805b2a1761b" }
-grovedb-storage = { git = "https://github.com/dashpay/grovedb", rev = "a18f7929460ef9c5d814f61ff84d8805b2a1761b" }
+grovedb-path = { git = "https://github.com/dashpay/grovedb", rev = "e5a904d70b2b7cdf078abcf47a51f076d8d468a4" }
+grovedb-storage = { git = "https://github.com/dashpay/grovedb", rev = "e5a904d70b2b7cdf078abcf47a51f076d8d468a4" }
 
 [features]
 default = ["bls-signatures"]
--- a/packages/rs-drive-abci/Cargo.toml
+++ b/packages/rs-drive-abci/Cargo.toml
@@ -82,7 +82,7 @@ derive_more = { version = "1.0", features = ["from", "deref", "deref_mut"] }
 async-trait = "0.1.77"
 console-subscriber = { version = "0.4", optional = true }
 bls-signatures = { git = "https://github.com/dashpay/bls-signatures", rev = "0842b17583888e8f46c252a4ee84cdfd58e0546f", optional = true }
-grovedb-commitment-tree = { git = "https://github.com/dashpay/grovedb", rev = "e5a904d70b2b7cdf078abcf47a51f076d8d468a4" }
+grovedb-commitment-tree = { git = "https://github.com/dashpay/grovedb", rev = "41786daa51495743fc5d7223065d848c36b481f0" }
 nonempty = "0.11"
 # Shielded-pool snapshot needs raw RocksDB SstFileWriter + ingest_external_file_cf
 # bindings, and blake3 for the snapshot-file checksum.
@@ -107,7 +107,7 @@ dpp = { path = "../rs-dpp", default-features = false, features = [
 drive = { path = "../rs-drive", features = ["fixtures-and-mocks"] }
 drive-proof-verifier = { path = "../rs-drive-proof-verifier" }
 strategy-tests = { path = "../strategy-tests" }
-grovedb-commitment-tree = { git = "https://github.com/dashpay/grovedb", rev = "e5a904d70b2b7cdf078abcf47a51f076d8d468a4", features = ["client"] }
+grovedb-commitment-tree = { git = "https://github.com/dashpay/grovedb", rev = "41786daa51495743fc5d7223065d848c36b481f0", features = ["client"] }
 assert_matches = "1.5.0"
 drive-abci = { path = ".", features = ["testing-config", "mocks", "shielded_test_data"] }
 bls-signatures = { git = "https://github.com/dashpay/bls-signatures", rev = "0842b17583888e8f46c252a4ee84cdfd58e0546f" }
@@ -121,8 +121,8 @@ integer-encoding = { version = "4.0.0" }
 
 # For dump_only_default_and_aux_cfs_under_shielded_subtree_prefix — same
 # subtree-prefix algorithm grovedb uses internally.
-grovedb-path = { git = "https://github.com/dashpay/grovedb", rev = "e5a904d70b2b7cdf078abcf47a51f076d8d468a4" }
-grovedb-storage = { git = "https://github.com/dashpay/grovedb", rev = "e5a904d70b2b7cdf078abcf47a51f076d8d468a4" }
+grovedb-path = { git = "https://github.com/dashpay/grovedb", rev = "41786daa51495743fc5d7223065d848c36b481f0" }
+grovedb-storage = { git = "https://github.com/dashpay/grovedb", rev = "41786daa51495743fc5d7223065d848c36b481f0" }
 
 [features]
 default = ["bls-signatures"]
--- a/packages/rs-drive-abci/Cargo.toml
+++ b/packages/rs-drive-abci/Cargo.toml
@@ -83,6 +83,13 @@ async-trait = "0.1.77"
 console-subscriber = { version = "0.4", optional = true }
 bls-signatures = { git = "https://github.com/dashpay/bls-signatures", rev = "0842b17583888e8f46c252a4ee84cdfd58e0546f", optional = true }
 grovedb-commitment-tree = { git = "https://github.com/dashpay/grovedb", rev = "41786daa51495743fc5d7223065d848c36b481f0" }
+# orchard 0.14 made the only constructor for an `Authorized` bundle
+# (`Bundle::try_from_parts`) require a `ProofSizeEnforcement` argument, which
+# `grovedb-commitment-tree` does not re-export. The shielded verifier needs to
+# name that enum to reconstruct the bundle, so orchard is pinned here to the
+# exact same git tag that `grovedb-commitment-tree` resolves to (cargo unifies
+# this to a single orchard build; no second copy is introduced).
+orchard = { git = "https://github.com/dashpay/orchard.git", tag = "dashified-0.14.0", features = ["circuit"] }
 nonempty = "0.11"
 # Shielded-pool snapshot needs raw RocksDB SstFileWriter + ingest_external_file_cf
 # bindings, and blake3 for the snapshot-file checksum.
--- a/packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/shielded_common/mod.rs
+++ b/packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/shielded_common/mod.rs
@@ -14,18 +14,10 @@ use drive::fees::op::LowLevelDriveOperation;
 use drive::grovedb::TransactionArg;
 use drive::state_transition_action::StateTransitionAction;
 use grovedb_commitment_tree::{
-    redpallas, Action, Anchor, Authorized, BatchValidator, Bundle, DashMemo,
-    ExtractedNoteCommitment, Flags, NoteBytesData, Nullifier, Proof, TransmittedNoteCiphertext,
-    ValueCommitment, VerifyingKey,
+    redpallas, Action, ActionFromPartsError, Anchor, Authorized, BatchValidator, Bundle, DashMemo,
+    ExtractedNoteCommitment, Flags, NoteBytesData, Nullifier, Proof, ProofSizeEnforcement,
+    TransmittedNoteCiphertext, ValueCommitment, VerifyingKey,
 };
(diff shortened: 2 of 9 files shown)


thepastaclaw:
✅ Review complete (commit 6ef45ed1)

thepastaclaw (approved):
## Code Review

All agents agree this is a minimal, well-scoped consensus fix: grovedb rev bump across all manifests plus a regression test pinning the v11 post-insert root hash. No in-scope findings from any reviewer; out-of-scope notes are speculative hardening/upstream audit suggestions that should not gate the PR.

packages/rs-drive/src/util/grove_operations/grove_insert_if_not_exists/mod.rs
```diff
@@ -51,3 +51,107 @@ impl Drive {
...
+    use crate::util::test_helpers::setup::setup_drive;
+    use dpp::version::PlatformVersion;
+    use grovedb::Element;
+    use grovedb_path::SubtreePath;
+
+    /// Protocol-v11 (`GROVE_V2`) consensus regression guard — testnet block 245,344.
+    ///
+    /// `transition_to_version_11` inserts an `empty_provable_count_sum_tree`
+    /// (CLEAR_ADDRESS_POOL) and an `empty_count_sum_tree` (ADDRESS_BALANCES) via
+    /// `grove_insert_if_not_exists`. Under the v11 grove version (`GROVE_V2`) grovedb
```
thepastaclaw: **💬 Nitpick: Doc-comment names wrong tree type for ADDRESS_BALANCES**

The doc-comment pairs ADDRESS_BALANCES with `empty_count_sum_tree`, but the test it documents inserts `Element::empty_sum_tree()` at `[56u8]` (line 100), and the inline comment at line 91 confirms `Insert empty_sum_tree at [56] (AddressBalances, the control)`. The upper doc-comment should match the actual inserted element — `empty_sum_tree` — to avoid misleading future readers about what `transition_to_version_11` is being modelled to do here.

```suggestion
    /// `transition_to_version_11` inserts an `empty_provable_count_sum_tree`
    /// (CLEAR_ADDRESS_POOL) and an `empty_sum_tree` (ADDRESS_BALANCES) via
    /// `grove_insert_if_not_exists`. Under the v11 grove version (`GROVE_V2`) grovedb
```

packages/rs-drive/src/util/grove_operations/grove_insert_if_not_exists/mod.rs
```diff
@@ -51,3 +51,107 @@ impl Drive {
...
+        // v12 / GROVE_V3 — intentionally layered (grovedb #759 version gate): the
+        // provable_count_sum_tree root MUST differ from the v11 Op::Put golden.
+        let (_v12_root_1, v12_root_2) =
+            insert_v11_address_trees(PlatformVersion::get(12).expect("protocol v12"));
+        eprintln!("v12 root_2 (provable_count_sum_tree) = {v12_root_2:?}");
+        assert_ne!(
+            v12_root_2, GOLDEN_2,
+            "GROVE_V3 (protocol v12) must use the layered-subtree dispatch (grovedb #759) — \
+             a root matching the v11 Op::Put value means the version gate was lost"
+        );
```
thepastaclaw, line 163: **💬 Nitpick: v12 control root is discarded instead of pinned against GOLDEN_1**

The doc-comment claims `empty_sum_tree` is the unchanged control ("layered always") and that grovedb #759's gate is scoped to `CountSumTree` / `ProvableCount[Sum]Tree`. The v12 arm binds the control root as `_v12_root_1` and only asserts `v12_root_2 != GOLDEN_2`. Pinning `v12_root_1 == GOLDEN_1` is a free additional guard: if the v3 gate ever accidentally widens to also re-layer the plain `sum_tree`, this assertion would catch it. The existing test still detects the block-245,344 regression and the intended v12 divergence, so this is purely a coverage tightening.

```suggestion
        // v12 / GROVE_V3 — intentionally layered (grovedb #759 version gate): the
        // provable_count_sum_tree root MUST differ from the v11 Op::Put golden, while
        // the empty_sum_tree control MUST stay on GOLDEN_1 (the gate is scoped to
        // CountSumTree / ProvableCount[Sum]Tree only).
        let (v12_root_1, v12_root_2) =
            insert_v11_address_trees(PlatformVersion::get(12).expect("protocol v12"));
        eprintln!("v12 root_1 (control sum_tree)        = {v12_root_1:?}");
        eprintln!("v12 root_2 (provable_count_sum_tree) = {v12_root_2:?}");
        assert_eq!(
            v12_root_1, GOLDEN_1,
            "v12 control sum_tree root changed — the grovedb #759 version gate must \
             only affect CountSumTree / ProvableCount[Sum]Tree, not empty_sum_tree"
        );
        assert_ne!(
            v12_root_2, GOLDEN_2,
            "GROVE_V3 (protocol v12) must use the layered-subtree dispatch (grovedb #759) — \
             a root matching the v11 Op::Put value means the version gate was lost"
        );
    }
}
```

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Cumulative re-review at e84b9907 — carried-forward prior findings: none (c350fd0e approved with zero findings). Delta from c350fd0e adds a strengthened bidirectional regression test (v11/GROVE_V2 pinned to Op::Put golden roots, v12/GROVE_V3 asserted to diverge), correctly shaped against the grovedb #759 version gate. Two non-blocking documentation/test-coverage nitpicks in the new test block: a doc-comment that mispairs ADDRESS_BALANCES with empty_count_sum_tree (it actually uses empty_sum_tree), and a discarded v12 control root that could pin against GOLDEN_1 for free additional gate-scope coverage. No security, consensus, or scope concerns in the delta.

💬 2 nitpick(s)

packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/shielded_common/mod.rs
```diff
@@ -142,8 +159,17 @@ pub fn reconstruct_and_verify_bundle(
...
+                InvalidShieldedProofError::new("action has identity randomizer key".to_string())
+            }
+            ActionFromPartsError::InvalidEpk => InvalidShieldedProofError::new(
+                "action has invalid ephemeral public key (identity or undecodable epk)".to_string(),
+            ),
+            // `ActionFromPartsError` is `#[non_exhaustive]`. Any future
+            // rejection variant added upstream MUST also be rejected here —
+            // defaulting to acceptance would weaken consensus soundness.
+            other => InvalidShieldedProofError::new(format!("malformed orchard action: {other}")),
         })?;
```
thepastaclaw, line 165: **🟡 Suggestion: New InvalidEpk rejection path is not pinned by a direct unit test**

The migration makes `Action::from_parts` the only place that surfaces two consensus-critical rejections: `IdentityRk` (carried over from 0.13) and `InvalidEpk` (the orchard 0.14 soundness fix this PR explicitly highlights as an intended acceptance change). The existing `reconstruct_and_verify_bundle_tests` cover encrypted-note size, empty actions, invalid `rk` (decoded before `Action::from_parts`), and invalid flags — none of them reach the `map_err` arms on lines 162-173. Without a direct test, a future orchard bump or refactor could silently map `InvalidEpk` to acceptance and the comment's load-bearing invariant ("we must not weaken it back to acceptance") would not be caught by CI. A focused test constructing a `SerializedAction` whose epk bytes decode to the identity point would lock the rejection semantics.

thepastaclaw (commented):
## Code Review

Re-validated both prior findings against head 954b40d0: both are STILL VALID (untouched in this push). The Orchard 0.13→0.14 verifier migration in `shielded_common/mod.rs` is sound — `ProofSizeEnforcement::Unenforced` preserves byte-exact acceptance, `InvalidEpk` rejection is the intended soundness change, and the `#[non_exhaustive]` catch-all defaults to reject. Two new latest-delta nits: orchard pinned by mutable `tag` (neighbors use `rev`), and no direct unit test pins the new `InvalidEpk` rejection path.

<details>

```

In `packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/shielded_common/mod.rs`:
- [SUGGESTION] packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/shielded_common/mod.rs:162-173: New InvalidEpk rejection path is not pinned by a direct unit test
  The migration makes `Action::from_parts` the only place that surfaces two consensus-critical rejections: `IdentityRk` (carried over from 0.13) and `InvalidEpk` (the orchard 0.14 soundness fix this PR explicitly highlights as an intended acceptance change). The existing `reconstruct_and_verify_bundle_tests` cover encrypted-note size, empty actions, invalid `rk` (decoded before `Action::from_parts`), and invalid flags — none of them reach the `map_err` arms on lines 162-173. Without a direct test, a future orchard bump or refactor could silently map `InvalidEpk` to acceptance and the comment's load-bearing invariant ("we must not weaken it back to acceptance") would not be caught by CI. A focused test constructing a `SerializedAction` whose epk bytes decode to the identity point would lock the rejection semantics.
```
</details>

packages/rs-drive-abci/Cargo.toml
```diff
@@ -82,7 +82,14 @@ derive_more = { version = "1.0", features = ["from", "deref", "deref_mut"] }
...
 bls-signatures = { git = "https://github.com/dashpay/bls-signatures", rev = "0842b17583888e8f46c252a4ee84cdfd58e0546f", optional = true }
-grovedb-commitment-tree = { git = "https://github.com/dashpay/grovedb", rev = "a18f7929460ef9c5d814f61ff84d8805b2a1761b" }
+grovedb-commitment-tree = { git = "https://github.com/dashpay/grovedb", rev = "41786daa51495743fc5d7223065d848c36b481f0" }
+# orchard 0.14 made the only constructor for an `Authorized` bundle
+# (`Bundle::try_from_parts`) require a `ProofSizeEnforcement` argument, which
+# `grovedb-commitment-tree` does not re-export. The shielded verifier needs to
+# name that enum to reconstruct the bundle, so orchard is pinned here to the
+# exact same git tag that `grovedb-commitment-tree` resolves to (cargo unifies
+# this to a single orchard build; no second copy is introduced).
+orchard = { git = "https://github.com/dashpay/orchard.git", tag = "dashified-0.14.0", features = ["circuit"] }
```
thepastaclaw: **💬 Nitpick: orchard is pinned by mutable `tag` while neighbors pin by immutable `rev`**

Every other git dep in this Cargo.toml (`bls-signatures`, `grovedb-commitment-tree`, `rocksdb`) pins to a 40-char commit SHA via `rev`. The new `orchard` entry uses `tag = "dashified-0.14.0"`. Tags are force-pushable; revs are content-addressed. Cargo.lock currently records the resolved commit (`f05557390a5843bc4eb04c66d8140bc9ef0fe9b7`), so reproducible builds hold as long as the lockfile is honored, but a `cargo update -p orchard` (or lockfile regeneration after a tag move) would silently shift orchard. For a consensus-critical verifier, matching the existing `rev` convention is more robust and consistent with the rest of the file.

```suggestion
orchard = { git = "https://github.com/dashpay/orchard.git", rev = "f05557390a5843bc4eb04c66d8140bc9ef0fe9b7", features = ["circuit"] }
```

<sub>source: ['claude']</sub>

packages/rs-drive/src/util/grove_operations/grove_insert_if_not_exists/mod.rs
thepastaclaw: Resolved in this update — _v12 control root is discarded instead of pinned against GOLDEN_1_ no longer present.

packages/rs-drive/src/util/grove_operations/grove_insert_if_not_exists/mod.rs
thepastaclaw: Resolved in this update — _Doc-comment names wrong tree type for ADDRESS_BALANCES_ no longer present.

shumkov: Fixed in d8474dc04c — the `[56]` control is now documented as `empty_sum_tree`, and the doc no longer implies the control is `Op::Put` (only the provable-count-sum tree is; `sum_tree` is layered in every version). Thanks.

packages/rs-drive/src/util/grove_operations/grove_insert_if_not_exists/mod.rs
shumkov: Done in d8474dc04c — added `assert_eq!(v12_root_1, GOLDEN_1)` and verified it holds (GROVE_V3 keeps the `sum_tree` control on GOLDEN_1), so grovedb #759's gate cannot silently widen to re-layer plain `sum_tree`. Good call, free guard.

packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/shielded_common/mod.rs
shumkov: Added in d8474dc04c: `test_identity_rk_returns_error` and `test_identity_epk_returns_invalid_epk_error` pin both `Action::from_parts` rejection arms (identity `rk`/`epk` = `[0u8;32]` — they decode but are rejected inside `from_parts`). The base action is minted from a real unauthorized Orchard bundle with precondition guards, so it fails loud if it ever stops reaching `from_parts`. `reconstruct_and_verify_bundle` suite: 6 passed (4 existing + 2 new).

packages/rs-drive-abci/Cargo.toml
shumkov: Addressing this at the root rather than switching tag to rev: the direct `orchard` dep should not exist at all — `grovedb-commitment-tree` already re-exports the orchard verification API, it just missed these two symbols. Opened dashpay/grovedb#760 to re-export `ProofSizeEnforcement` + `ActionFromPartsError`. Once it merges, this PR bumps to the new develop rev and drops the direct `orchard` dependency entirely (no tag-vs-rev question). Leaving this thread open until that lands.

thepastaclaw: Resolved in this update — _orchard is pinned by mutable `tag` while neighbors pin by immutable `rev`_ no longer present.

thepastaclaw: Correction: the auto-resolve note above was wrong. This finding is **still valid at d8474dc0**: `packages/rs-drive-abci/Cargo.toml` still pins the direct `orchard` dependency by mutable `tag = "dashified-0.14.0"`.

Ivan’s plan to fix this at the root via dashpay/grovedb#760 and then drop the direct `orchard` dependency here is a good resolution path; this thread should stay open until that lands. The latest review did not find any new issues beyond this carried-forward nit.

shumkov: Resolved in d40bbdf26a. grovedb#760 (merged) now re-exports `ProofSizeEnforcement` + `ActionFromPartsError` from `grovedb-commitment-tree`, so the **direct `orchard` dependency is removed entirely** — the verifier imports both via `grovedb_commitment_tree` alongside the other orchard types it already re-exports. There is no longer any direct orchard dep, so the tag-vs-rev concern no longer applies. Cargo.lock keeps a single orchard `0.14.0` (now only transitive via `grovedb-commitment-tree`; the grovedb rev was bumped 41786daa→40521b83 to pick up #760).

thepastaclaw (commented):
## Code Review

Latest delta is a clean mechanical refactor: rs-drive-abci drops its direct orchard dependency and now consumes ProofSizeEnforcement/ActionFromPartsError through grovedb-commitment-tree (grovedb#760), which itself is SHA-pinned. Prior nitpick about pinning orchard by tag is FIXED. One in-scope suggestion remains: the PR description asks reviewers to sign off on ProofSizeEnforcement::Strict, but the shipped code uses Unenforced with an in-code comment explicitly arguing the opposite — description and code must agree before reviewer sign-off is meaningful.

🟡 1 suggestion(s)

<details>

```

In `packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/shielded_common/mod.rs`:
- [SUGGESTION] packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/shielded_common/mod.rs:186-216: PR description claims `ProofSizeEnforcement::Strict`, code ships `Unenforced` — reconcile before consensus sign-off
  The PR description's "Consensus-review note" explicitly asks reviewers for a conscious sign-off on two acceptance tightenings vs orchard 0.13: (1) the new `InvalidEpk` rejection, and (2) `ProofSizeEnforcement::Strict` rejecting non-canonical proof sizes, characterized as "the sound, anti-malleability policy." The validation section likewise claims valid canonical proofs are accepted under `Strict`.

  The shipped code does the opposite at `reconstruct_and_verify_bundle`: `Bundle::try_from_parts(.., ProofSizeEnforcement::Unenforced)`, with an in-tree comment that explicitly forbids the policy the description advertises: "Choosing `Strict` here would introduce a NEW rejection (`BundleError::NonCanonicalProofSize`) that did not exist before, changing consensus behavior for an edge case — which is forbidden." `git log -S 'ProofSizeEnforcement::Strict'` confirms `Strict` was never present; the code has been `Unenforced` since the orchard-0.14 migration commit (954b40d09a).

  Why this matters:

  1. Reviewer integrity — a consensus reviewer signing off on the description's prompt believes they are approving an anti-malleability tightening; the patch they would actually merge preserves 0.13's accept-anything proof sizing.
  2. Acceptance surface — under `Unenforced` the wire-level proof-size check is skipped. The verifier is still sound (non-canonical proof bytes must still verify under the Halo 2 batch validator downstream, so forged proofs cannot be silently accepted), but a non-canonical encoding can survive `try_from_parts`, change the serialized state transition bytes, and therefore the `transaction_id()` hash. Once shielded transitions activate, that opens a txid-malleability surface for relayers/wallets keying off the transition hash, even though the state effect (nullifiers, signatures) is unchanged. The PR description's "tightening over 0.13" claim for this leg is therefore inaccurate.
  3. No test pins the decision — `test_identity_epk_returns_invalid_epk_error` pins the `InvalidEpk` arm, but no test distinguishes `Unenforced` from `Strict`. Existing positive tests use canonical proof sizes, so a future refactor could silently flip either way.

  Resolution: either (a) update the PR description (and the consensus-review note) to reflect that only `InvalidEpk` is the new tightening and that `Unenforced` was deliberately chosen to preserve 0.13 acceptance semantics — which matches the in-tree comment and is the conservative choice if v12 shielded acceptance is already load-bearing; or (b) switch the code to `Strict` and add a `NonCanonicalProofSize` rejection test. Description and code must agree before merge so the requested consensus sign-off is meaningful.

  The `InvalidEpk` half of the description is correctly wired up and tested — this finding is solely about the proof-size enforcement half.
```
</details>

shumkov:
@thepastaclaw addressed — thanks for the careful re-review.

1. **Description/code mismatch is resolved.** That review ran against a pre-`Strict` commit (it notes `git log -S 'ProofSizeEnforcement::Strict'` finds nothing, citing `954b40d09a`). The current head (`6ef45ed191`) uses `ProofSizeEnforcement::Strict` in `reconstruct_and_verify_bundle`, and the description's consensus-review note now matches the code — both list the two intended acceptance-tightenings (`InvalidEpk` + non-canonical proof size).

2. **The proof-size decision is now pinned by a test** (your point #3): `test_noncanonical_proof_size_rejected_under_strict` — a bundle with a valid action but a non-canonical proof byte-length is rejected by `try_from_parts(.., Strict)` (`NonCanonicalProofSize`). This is exactly what distinguishes `Strict` from `Unenforced`; the positive round-trips use canonical proofs and pass under either, so a future flip would now break a test. (7 `reconstruct_and_verify_bundle` tests pass.)

Your txid-malleability analysis was the rationale for choosing `Strict`. The orchard-tag nitpick is also fully resolved — there is no longer any direct `orchard` dependency (imported via `grovedb_commitment_tree`, grovedb#760).

thepastaclaw (approved):
## Code Review

Cumulative incremental review of PR #3789 at head 6ef45ed1. Prior finding status: 'PR description vs code mismatch on ProofSizeEnforcement (Strict vs Unenforced)' — FIXED in commits 25d0b4426c (code flip Unenforced→Strict at shielded_common/mod.rs:197) and 6ef45ed191 (regression test test_noncanonical_proof_size_rejected_under_strict at lines 665-686 pinning the policy). Prior orchard pinning concern remains FIXED (no direct orchard dep; imports come via grovedb-commitment-tree). No new in-scope blocking issues in the latest delta. One low-priority nitpick on test assertion strength is recorded but does not affect review action.

QuantumExplorer (approved):

PR #3799, opened 2026-06-04
fix(dpp): return error instead of panicking on storage-fee refund div-by-zero

QuantumExplorer:
> [!NOTE]
> ## Two-layer fix: panic removed **and** liveness restored (updated after adversarial review)
>
> An adversarial review found that converting the panic to an `Err` (the first commit) **relocates** the halt rather than preventing it. The `Err` propagates up the consensus path — `restore_original_removed_credits_amount` → `refund_storage_fee_to_epochs_map` → `add_distribute_storage_fee_to_epochs_operations` → `process_block_fees_and_validate_sum_trees` → `run_block_proposal_v0` (the `?` returns the **outer** `Err`; it is **never** folded into `ValidationResult.errors`) → `process_proposal` (the `?` short-circuits **before** the graceful `if !run_result.is_valid()` reject path) → `ResponseException`, which CometBFT/Tenderdash **panics** on. So returning `Err` still halts every validator at the same block.
>
> **And it is reachable** — via a temporal decoupling that the original "needs a ~50-year-old document" framing understated. The refund **amount** is computed at the *removal* epoch (`FeeRefunds::from_storage_removal`, keyed by the original write epoch — the removal epoch is then discarded), but the **clawback** and its `1 / ratio_used` multiplier are recomputed one epoch later, at the next epoch change, against the *then-current* epoch (`add_distribute_storage_fee_to_epochs_operations_v0`; natural `c_consume = c_store + 1` lag). So data **removed just before** its window expires produces a legitimately non-zero refund that is then **distributed just after** the boundary is crossed, where `current_era == PERPETUAL_STORAGE_ERAS` and `ratio_used == 0`.
>
> **Both layers are now included:**
> 1. **Liveness (the actual fix):** a `start_era >= PERPETUAL_STORAGE_ERAS` guard in `refund_storage_fee_to_epochs_map`. Once the window has fully elapsed at distribution time there are no future epoch pools to claw back from — the per-era loop is already empty and the multiplier would be unused — so the entire refund is routed through the existing leftovers path to the current epoch and the block **processes**. Credits are conserved, and for the `+1`-lag boundary the current epoch *is* the data's final in-window epoch, so debiting it is exactly correct. Chosen over "return a multiplier of `1`" because there should be no future-epoch redistribution at all once the window has elapsed.
> 2. **Defense-in-depth:** `original_removed_credits_multiplier_from` still returns `Err(DivideByZero)` rather than panicking, so the helper is safe even if a future caller reaches it at the boundary (now unreachable from the production path).
>
> **Remaining, optional (not halts on the production path):** the unchecked `skip_until_epoch_index - start_epoch_index` subtraction (can't underflow in practice, since `skip_until = current + 1 ≥ start`); the near-boundary multiplier explosion (`current_era == 49` → ~32000×), now unreachable since the guard fires at `>= 50` but which could in principle overflow `to_u64()` for enormous refunds; and the `epochs_per_era == 0` divide-by-zeros (`u16` from config with no `NonZeroU16` validation) — a misconfiguration concern, separate from this path.

---

**Latent chain-halt in storage-fee refund accounting.**

`original_removed_credits_multiplier_from` (`fee/epoch/distribution.rs`) computed `dec!(1) / ratio_used`. `FEE_DISTRIBUTION_TABLE` has exactly `PERPETUAL_STORAGE_ERAS` (50) entries, so once the repayment epoch is at least that whole window past the original storage epoch (`current_era >= 50`), every table era compares `Ordering::Less`, the iterator yields nothing, and `ratio_used` sums to **zero**.

`rust_decimal::Decimal`'s `/` operator **panics** on a zero divisor — unlike integer/`f64` division, and unlike its own `checked_div`. This call runs on the consensus path (`process_block_fees_and_validate_sum_trees` at epoch change, via `restore_original_removed_credits_amount` → `refund_storage_fee_to_epochs_map`). A panic there aborts every node simultaneously → chain halt. It is reached via the temporal decoupling described in the note above: data **removed just before** its window expires (a non-zero refund) is **distributed one epoch later**, just after the boundary is crossed. It needs the chain to have aged ~50 eras, so it is not a near-term risk, but it is a deterministic liveness failure as the chain ages, with no graceful path.

## What was done?

**1. Liveness guard (keeps the block processing).** `refund_storage_fee_to_epochs_map` now returns early when `start_era >= PERPETUAL_STORAGE_ERAS`: with the window fully elapsed at distribution time there are no future epoch pools to claw back from, so the entire refund is treated as leftovers and removed from the current epoch (the existing leftovers path), and the block processes normally. The would-be-divide-by-zero `1 / ratio_used` multiplier is never computed, because it would be unused (the per-era loop is empty).

**2. Panic → error (defense-in-depth).** `original_removed_credits_multiplier_from` now returns `Result<Decimal, ProtocolError>`:
- Returns `ProtocolError::DivideByZero` when `ratio_used == 0` instead of panicking.
- Guards the `paid_epochs = start_repayment - start` subtraction with `checked_sub` → `ProtocolError::Overflow`.

The sole production caller `restore_original_removed_credits_amount` returns a `Result` and propagates with `?`; test call sites updated to `.expect(...)`. With layer 1 in place this helper is no longer reached at the boundary on the production path, but the guard remains as defense-in-depth.

## How Has This Been Tested?

`cargo test -p dpp --lib distribution` — **456 passed, 0 failed**, including three new tests:
- `should_not_halt_when_refund_distributed_as_window_boundary_is_crossed`: reproduces the **decoupled-epoch** path — a non-zero refund computed at the removal epoch (`c_store = 1998`, write epoch 0, 40 epochs/era) via `calculate_storage_fee_refund_amount_and_leftovers`, then distributed at `c_consume = 1999` via `subtract_refunds_from_epoch_credits_collection`. Without the guard this divides by zero (panic on the original code, `Err(DivideByZero)` after the first commit); with it, the call succeeds and the full refund is clawed back from the current epoch. Every pre-existing `subtract_refunds` test fed the *same* epoch to both calls (the self-consistent case), which is why this was never caught.
- `should_return_error_instead_of_panicking_when_window_fully_elapsed`: asserts `original_removed_credits_multiplier_from(0, 1000, 20)` returns `Err(DivideByZero)` and that `restore_original_removed_credits_amount` propagates it.
- `should_return_error_when_repayment_epoch_precedes_start`: asserts the underflow guard returns `Err(Overflow)`.

`cargo fmt --all` clean.

## Breaking Changes

No API breakage. This does change behavior on the consensus path at one boundary: a fully-elapsed-window refund that previously **halted** every node (panic, and after the first commit a propagated `Err` → Tenderdash `ResponseException`) now **processes** — the refund is removed from the current epoch and the block continues. Because the prior behavior was a deterministic halt (not a divergent state), this is a liveness fix rather than a fork risk; if it ships to a live network the activation should still be coordinated like any consensus change.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened validation in credit calculation logic to properly detect and handle edge cases that could cause calculation failures
  * Enhanced error handling in credit restoration operations to prevent unexpected failures from invalid parameter combinations
  * Added safeguards against mathematical edge cases that could produce invalid results or unexpected behavior

### DIFF
--- a/packages/rs-dpp/src/fee/epoch/distribution.rs
+++ b/packages/rs-dpp/src/fee/epoch/distribution.rs
@@ -310,6 +310,26 @@ where
 
     let start_era: u16 = (skip_until_epoch_index - start_epoch_index) / epochs_per_era;
 
+    // The perpetual storage window for this data ends `PERPETUAL_STORAGE_ERAS`
+    // eras after `start_epoch_index`. Once the distribution epoch reaches or
+    // passes that end (`start_era >= PERPETUAL_STORAGE_ERAS`) there are no future
+    // epoch pools left to claw the refund back from: the per-era loop below is
+    // empty, and `original_removed_credits_multiplier_from` returns
+    // `DivideByZero` (`ratio_used == 0`) computing a multiplier that is never
+    // used. Returning that error here still halts the chain — it propagates up
+    // the consensus path to a Tenderdash `ResponseException`. Instead treat the
+    // whole refund as leftovers so the caller removes it from the current
+    // epoch's pool, keeping the chain live.
+    //
+    // This is reachable on the consensus path: the refund amount is computed at
+    // the removal epoch (`FeeRefunds::from_storage_removal`) but this clawback
+    // runs at least one epoch later, at the next epoch change, against the
+    // then-current epoch. So a non-zero refund for data removed just before
+    // expiry can be distributed just after the window boundary is crossed.
+    if start_era >= PERPETUAL_STORAGE_ERAS {
+        return Ok(storage_fee);
+    }
+
     // Let's imagine that we are refunding something from epoch 5
     // We are at Epoch 12
     // The refund amount is from Epoch 13 (current + 1) to Epoch 1005 (5 + 1000)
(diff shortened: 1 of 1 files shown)


thepastaclaw:
✅ Review complete (commit b4f2b1c5)

thepastaclaw (commented):
## Code Review

Targeted, well-scoped fix that converts a latent Decimal divide-by-zero panic in `original_removed_credits_multiplier_from` into a propagable `ProtocolError::DivideByZero`. The core fix is sound and consensus-safe. One suggestion: the companion defensive `checked_sub` for `paid_epochs` is bypassed by an unchecked subtraction one frame up in `refund_storage_fee_to_epochs_map` (line 311), so the underflow guard is unreachable on the production path. The deeper semantic question of whether aged-out refunds should be no-ops rather than errors is acknowledged as out-of-scope by the PR description.

🟡 1 suggestion(s)

<details>

```

In `packages/rs-dpp/src/fee/epoch/distribution.rs`:
- [SUGGESTION] packages/rs-dpp/src/fee/epoch/distribution.rs:311: New `checked_sub` underflow guard is bypassed by an earlier unchecked subtraction on the same call path
  The PR adds a `checked_sub` in `original_removed_credits_multiplier_from` (lines 174-178) explicitly to ensure corrupted/unexpected inputs return `ProtocolError::Overflow` rather than panicking. The sole production caller, `refund_storage_fee_to_epochs_map`, performs the same subtraction one frame up at line 311: `let start_era: u16 = (skip_until_epoch_index - start_epoch_index) / epochs_per_era;`. This runs before `restore_original_removed_credits_amount` (line 317) and therefore before the new guard ever sees the inputs. If `skip_until_epoch_index < start_epoch_index` ever occurred, line 311 would panic (debug) or wrap (release) first, so the guard is unreachable for the underflow case it's meant to defend. Either guard line 311 with `checked_sub` for a consistent defensive posture, or drop the underflow guard (and its test) since the invariant is structurally enforced. The divide-by-zero portion of the fix is unaffected by this and remains correct.
```
</details>

packages/rs-dpp/src/fee/epoch/distribution.rs
```diff
@@ -287,6 +310,26 @@ where
 
     let start_era: u16 = (skip_until_epoch_index - start_epoch_index) / epochs_per_era;
```
thepastaclaw, line 311: **🟡 Suggestion: Unchecked `skip_until_epoch_index - start_epoch_index` bypasses (and now interacts badly with) the new defensive guards**

This PR deliberately hardened `original_removed_credits_multiplier_from` with `checked_sub` (lines 174-178) so corrupted/unexpected inputs return `ProtocolError::Overflow` rather than panic on the consensus path. The same subtraction is performed unchecked one frame higher at line 311 while computing `start_era`, so if `skip_until_epoch_index < start_epoch_index` ever held, debug builds still panic and release builds wrap before the helper's checked subtraction can run — defeating the hardening on its only production caller.

The new liveness guard at line 329 (`start_era >= PERPETUAL_STORAGE_ERAS`) actually makes this slightly worse in release builds: a wrapped `start_era` divided by `epochs_per_era` will almost certainly land >= 50, so the guard returns `Ok(storage_fee)` and silently clawback-routes a fictitious refund from the current epoch instead of erroring. Not reachable on the production path today (the sole caller passes `current_epoch_index + 1 >= start_epoch_index` by construction, and `FeeRefunds::from_storage_removal` enforces the invariant), and the PR description explicitly defers it. Flagging it so the two paths stay symmetric and so the new guard cannot silently mask future corruption.

```suggestion
    let start_era: u16 = skip_until_epoch_index
        .checked_sub(start_epoch_index)
        .ok_or(ProtocolError::Overflow(
            "skip_until_epoch_index is before the original storage epoch",
        ))?
        / epochs_per_era;
```

thepastaclaw (commented):
## Code Review

Latest delta (b4f2b1c5) replaces the prior relocated-panic with the correct liveness fix: a `start_era >= PERPETUAL_STORAGE_ERAS` early-return in `refund_storage_fee_to_epochs_map` plus a focused regression test for the decoupled-epoch boundary. Credits conservation holds and the new test exercises the previously-uncovered c_store != c_consume path. One carried-forward finding from 34fe2613 remains STILL VALID: line 311 still computes `skip_until_epoch_index - start_epoch_index` unchecked, bypassing the `checked_sub` defense-in-depth this PR adds in the helper one frame down. Author explicitly deferred it; suggestion-level only.

🟡 1 suggestion(s)

<details>

```

In `packages/rs-dpp/src/fee/epoch/distribution.rs`:
- [SUGGESTION] packages/rs-dpp/src/fee/epoch/distribution.rs:311: Unchecked `skip_until_epoch_index - start_epoch_index` bypasses (and now interacts badly with) the new defensive guards
  This PR deliberately hardened `original_removed_credits_multiplier_from` with `checked_sub` (lines 174-178) so corrupted/unexpected inputs return `ProtocolError::Overflow` rather than panic on the consensus path. The same subtraction is performed unchecked one frame higher at line 311 while computing `start_era`, so if `skip_until_epoch_index < start_epoch_index` ever held, debug builds still panic and release builds wrap before the helper's checked subtraction can run — defeating the hardening on its only production caller.

  The new liveness guard at line 329 (`start_era >= PERPETUAL_STORAGE_ERAS`) actually makes this slightly worse in release builds: a wrapped `start_era` divided by `epochs_per_era` will almost certainly land >= 50, so the guard returns `Ok(storage_fee)` and silently clawback-routes a fictitious refund from the current epoch instead of erroring. Not reachable on the production path today (the sole caller passes `current_epoch_index + 1 >= start_epoch_index` by construction, and `FeeRefunds::from_storage_removal` enforces the invariant), and the PR description explicitly defers it. Flagging it so the two paths stay symmetric and so the new guard cannot silently mask future corruption.
```
</details>

QuantumExplorer:
Reviewed

PR #3804, opened 2026-06-07
fix(dashmate): ssl obtain --force populates context before CSR generation (fixes #3803)

### LINKED ISSUE #3803: dashmate ssl obtain --force fails with "Attribute value not specified" (--force skips externalIp init, breaking CSR generation)
## Summary

`dashmate ssl obtain --provider zerossl --force` always fails with `[FAILED] Attribute value not specified.` at the **Generate certificate request** step. The `--force` flag is effectively unusable for the ZeroSSL provider.

This is particularly impactful because several of dashmate's own SSL error messages explicitly instruct the user to re-run **with `--force`** (e.g. on `PRIVATE_KEY_IS_NOT_PRESENT`, `EXTERNAL_IP_MISMATCH`, `CSR_FILE_IS_NOT_PRESENT`). Users following the tool's own guidance hit a dead end.

## Environment

- dashmate `3.0.1` (also confirmed present on the current `v3.1-dev` default branch by source inspection — see line refs below)
- SSL provider: ZeroSSL
- OS: Ubuntu 24.04

## Steps to reproduce

1. A node that already has a valid ZeroSSL certificate configured (valid `externalIp` and ZeroSSL API key in config).
2. Run:
   ```
   dashmate ssl obtain --provider zerossl --config mainnet --force --verbose
   ```

## Actual behavior

```
[STARTED] Obtain ZeroSSL certificate
[STARTED] Check if certificate already exists and not expiring soon
[SKIPPED] Check if certificate already exists and not expiring soon
[STARTED] Generate a keypair
[SUCCESS] Generate a keypair
[STARTED] Generate certificate request
[FAILED] Attribute value not specified.
```

The certificate is not renewed. (`--expiration-days N` on the same node works fine, which is the clue — see Root cause.)

## Expected behavior

`--force` issues a fresh certificate even when the current one is still valid (that is what the flag documents: *"renew even if certificate is valid"*).

## Root cause

In `packages/dashmate/src/listr/tasks/ssl/zerossl/obtainZeroSSLCertificateTaskFactory.js` (line numbers against `v3.1-dev`):

- The first subtask, *"Check if certificate already exists and not expiring soon"*, is skipped entirely under `--force`:
  ```js
  skip: (ctx) => ctx.force,            // ~line 50
  ```
- But that subtask's body is the **only** place `ctx.externalIp` (and `ctx.apiKey`, `ctx.isCsrFilePresent`, `ctx.csr`) are populated:
  ```js
  const { error, data } = await validateZeroSslCertificate(config, ctx.expirationDays);
  lodash.merge(ctx, data);             // ~line 54  -> sets ctx.externalIp, ctx.apiKey, ctx.isCsrFilePresent, ...
  ```
- So with `--force`, `ctx.externalIp` and `ctx.isCsrFilePresent` are `undefined`. The keypair + CSR subtasks then run, because their guard is `enabled: (ctx) => !ctx.isCsrFilePresent` → `!undefined === true`:
  ```js
  ctx.csr = await generateCsr(ctx.keyPair, ctx.externalIp);   // ~line 129-131, ctx.externalIp === undefined
  ```
- `generateCsr` (`packages/dashmate/src/ssl/zerossl/generateCsr.js`) passes the undefined value straight into node-forge's subject:
  ```js
  csr.setSubject([{ name: 'commonName', value: externalIp }]); // ~line 20-21, value === undefined
  ```
  node-forge then throws **`Attribute value not specified.`**

## Workaround

Use `--expiration-days N` (with `N` greater than the certificate's remaining validity) instead of `--force`. This keeps the check subtask (so `ctx.externalIp` etc. are initialized) but still triggers re-issuance via the `CERTIFICATE_EXPIRES_SOON` path. Verified working.

## Suggested fix

Do not skip the ctx-initialization under `--force`. For example, either:

1. Always run `validateZeroSslCertificate` + `lodash.merge(ctx, data)` to populate `ctx` (externalIp / apiKey / isCsrFilePresent / csr), and let `--force` only override the *"certificate is still valid → stop"* early-return; **or**
2. Split the first subtask into (a) an always-run *"load SSL context"* task and (b) a *"short-circuit if certificate already valid"* task that is the one skipped by `--force`.

## Note (possibly a separate issue)

The CLI `ssl obtain` writes the new `bundle.crt` but does **not** reload the gateway — envoy keeps serving the old certificate until the gateway container receives `kill -SIGHUP 1`. The scheduled-renewal helper does send the SIGHUP (`scheduleRenewZeroSslCertificateFactory.js`), but the manual CLI path does not, so a successful manual `ssl obtain` appears to have "no effect" on the served certificate until the operator restarts/HUPs the gateway. It may be worth either documenting this or having the CLI command reload the gateway on success.

---

*This bug was discovered and root-caused by **Claude** (Anthropic's Claude Code agent) while performing SSL certificate maintenance on a dashmate node, and is filed on behalf of the node operator. Happy to open a PR with the preferred fix shape if a maintainer confirms which approach they'd like.*

DashBot-0001:
## What

Fixes #3803 — `dashmate ssl obtain --provider zerossl --force` always fails with `[FAILED] Attribute value not specified.` at the *Generate certificate request* step, which makes the `--force` flag unusable for ZeroSSL.

## Root cause

In `packages/dashmate/src/listr/tasks/ssl/zerossl/obtainZeroSSLCertificateTaskFactory.js`, the first subtask *"Check if certificate already exists and not expiring soon"* was skipped entirely under `--force`:

```js
skip: (ctx) => ctx.force,
```

But that subtask is the **only** place the obtain context is populated:

```js
const { error, data } = await validateZeroSslCertificate(config, ctx.expirationDays);
lodash.merge(ctx, data); // sets ctx.externalIp, ctx.apiKey, ctx.isCsrFilePresent, ...
```

So under `--force`, `ctx.externalIp` is `undefined`, and the later CSR subtask calls `generateCsr(ctx.keyPair, ctx.externalIp)` → node-forge `csr.setSubject([{ name: 'commonName', value: undefined }])` → throws **`Attribute value not specified`**.

## Fix

Run `validateZeroSslCertificate` + `lodash.merge(ctx, data)` **unconditionally** so the context is always populated, then handle `--force` explicitly *after* the merge:

- still abort on genuine prerequisites (`API_KEY_IS_NOT_SET`, `EXTERNAL_IP_IS_NOT_SET`);
- otherwise force full regeneration (`ctx.isCsrFilePresent = false; ctx.certificate = null`), which preserves the original intent of `--force` (a brand-new keypair, CSR and certificate).

The non-`--force` `switch (error)` path is left **unchanged**, so normal issuance and the scheduled auto-renewal behavior are untouched.

## Testing

- Reproduced on a real node (dashmate `3.0.1`): `ssl obtain --provider zerossl --force` → `Attribute value not specified`; the same node with `ssl obtain --provider zerossl --expiration-days N` succeeds — which is the clue, since `--expiration-days` keeps the context-loading subtask while `--force` skipped it.
- Verified the code paths against the current `v3.1-dev` source.
- `node --check` passes on the patched file.
- I was **not** able to run the full dashmate test suite in my environment. Happy to add/adjust unit tests for the `--force` path per your preference.

---

*Root-caused and fixed by **Claude** (Anthropic's Claude Code agent) while performing SSL certificate maintenance on a dashmate node, submitted on behalf of the node operator. Full investigation in #3803.*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SSL certificate validation and force renewal to ensure comprehensive validation occurs before regeneration while protecting against configuration errors.

### DIFF
--- a/packages/dashmate/src/listr/tasks/ssl/zerossl/obtainZeroSSLCertificateTaskFactory.js
+++ b/packages/dashmate/src/listr/tasks/ssl/zerossl/obtainZeroSSLCertificateTaskFactory.js
@@ -46,8 +46,6 @@ export default function obtainZeroSSLCertificateTaskFactory(
     const tasks = new Listr([
       {
         title: 'Check if certificate already exists and not expiring soon',
-        // Skips the check if force flag is set
-        skip: (ctx) => ctx.force,
         task: async (ctx, task) => {
           const { error, data } = await validateZeroSslCertificate(config, ctx.expirationDays);
 
@@ -55,6 +53,28 @@ export default function obtainZeroSSLCertificateTaskFactory(
 
           // Ensure we have config dir created
           fs.mkdirSync(ctx.sslConfigDir, { recursive: true });
+
+          // With --force we always (re)create the certificate. We must still run
+          // validateZeroSslCertificate above so the context (externalIp, apiKey, file
+          // paths) is populated for the keypair/CSR/creation steps below — otherwise
+          // generateCsr() is called with an undefined externalIp and fails with
+          // "Attribute value not specified". Only genuine prerequisite configuration
+          // errors should still abort under --force.
+          if (ctx.force) {
+            if (error === ERRORS.API_KEY_IS_NOT_SET) {
+              throw new Error('ZeroSSL API key is not set. Please set it in the config file');
+            }
+
+            if (error === ERRORS.EXTERNAL_IP_IS_NOT_SET) {
+              throw new Error('External IP is not set. Please set it in the config file');
+            }
+
+            // Force full regeneration of keypair, CSR and certificate
+            ctx.isCsrFilePresent = false;
+            ctx.certificate = null;
+
+            return;
+          }
 
           switch (error) {
             case undefined:


thepastaclaw:
✅ Review complete (commit 885d69f2)

packages/dashmate/src/listr/tasks/ssl/zerossl/obtainZeroSSLCertificateTaskFactory.js
```diff
@@ -56,6 +54,28 @@ export default function obtainZeroSSLCertificateTaskFactory(
...
+              throw new Error('ZeroSSL API key is not set. Please set it in the config file');
+            }
+
+            if (error === ERRORS.EXTERNAL_IP_IS_NOT_SET) {
+              throw new Error('External IP is not set. Please set it in the config file');
+            }
+
+            // Force full regeneration of keypair, CSR and certificate
+            ctx.isCsrFilePresent = false;
+            ctx.certificate = null;
```
thepastaclaw, line 74: **🔴 Blocking: --force leaves stale file-presence flags, skipping key/bundle writes**

When `validateZeroSslCertificate` returns `error === undefined` (the certificate is currently valid — the common --force case), it sets `data.isPrivateKeyFilePresent = true` and `data.isBundleFilePresent = true` (validateZeroSslCertificateFactory.js:67-69, with the bundle flag only cleared in the expires-soon / not-validated / not-valid branches). After `lodash.merge(ctx, data)`, the new --force branch clears only `ctx.isCsrFilePresent` and `ctx.certificate`. The downstream tasks then create a new keypair, CSR, and ZeroSSL certificate, and write the config pointing at the new certificate ID — but `Save certificate private key file` (enabled: `!isPrivateKeyFilePresent`), `Download certificate file` (skip: `isBundleFilePresent`), and `Save certificate file` (skip: `isBundleFilePresent`) are all skipped. The end state is a config that references a brand-new ZeroSSL certificate while `private.key` and `bundle.crt` on disk still belong to the previous one — precisely the breakage `--force` exists to fix. Reset all three flags (and clear `ctx.certificateFile` for symmetry) when forcing regeneration.

```suggestion
            // Force full regeneration of keypair, CSR and certificate
            ctx.isCsrFilePresent = false;
            ctx.isPrivateKeyFilePresent = false;
            ctx.isBundleFilePresent = false;
            ctx.certificate = null;
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):
## Code Review

One real blocking bug: the --force branch resets only `isCsrFilePresent` and `certificate` but leaves `isPrivateKeyFilePresent`/`isBundleFilePresent` truthy when validation returned a valid cert. The downstream Save private key, Download bundle, and Save bundle tasks then skip themselves, so a new certificate is registered in config while the old on-disk key/bundle remain — the very state --force is supposed to refresh. Other reported items are low-value (pre-existing test gap, intentional --force semantics) and dropped.

🔴 1 blocking

<details>

```

In `packages/dashmate/src/listr/tasks/ssl/zerossl/obtainZeroSSLCertificateTaskFactory.js`:
- [BLOCKING] packages/dashmate/src/listr/tasks/ssl/zerossl/obtainZeroSSLCertificateTaskFactory.js:72-74: --force leaves stale file-presence flags, skipping key/bundle writes
  When `validateZeroSslCertificate` returns `error === undefined` (the certificate is currently valid — the common --force case), it sets `data.isPrivateKeyFilePresent = true` and `data.isBundleFilePresent = true` (validateZeroSslCertificateFactory.js:67-69, with the bundle flag only cleared in the expires-soon / not-validated / not-valid branches). After `lodash.merge(ctx, data)`, the new --force branch clears only `ctx.isCsrFilePresent` and `ctx.certificate`. The downstream tasks then create a new keypair, CSR, and ZeroSSL certificate, and write the config pointing at the new certificate ID — but `Save certificate private key file` (enabled: `!isPrivateKeyFilePresent`), `Download certificate file` (skip: `isBundleFilePresent`), and `Save certificate file` (skip: `isBundleFilePresent`) are all skipped. The end state is a config that references a brand-new ZeroSSL certificate while `private.key` and `bundle.crt` on disk still belong to the previous one — precisely the breakage `--force` exists to fix. Reset all three flags (and clear `ctx.certificateFile` for symmetry) when forcing regeneration.
```
</details>

PR #3793, opened 2026-06-04
feat(drive): shielded fees for Shield/ShieldFromAssetLock + shield credit conservation

QuantumExplorer:
## Issue being fixed or feature implemented

This PR does **two things** to the shielded pool (protocol v12 — still pre-release, behind feature gates):

**1. Fixes a bug in transparent `Shield` that destroys credits and halts the chain.**
A `Shield` moves credits from transparent platform addresses into the shielded pool. A wallet authorizes a *maximum* contribution per input, and only part of it is actually shielded — the rest is supposed to stay in the address. The old code debited the **full** authorized amount from each address but credited the pool only the shielded amount, so the difference was **destroyed**. The platform runs a credit-conservation check at the end of every block, so a single ordinary shield (the wallet routinely authorizes more than it shields) makes every validator reject the block at the same height → **the chain stops**.

**2. Completes and corrects the shielded fee model so every shielded operation pays for the resources it uses.**
Every shielded operation should pay for its expensive zero-knowledge proof verification and its storage. Transfers/unshields/withdrawals got that in #3800; this PR finishes the job for the two *entry* transitions (`Shield`, `ShieldFromAssetLock`), charges `Shield` the **real metered cost** instead of a flat estimate, and prices the **Core withdrawal document** into `ShieldedWithdrawal` so the proposer is properly compensated.

### 1. The credit-conservation fix

Debit only what is actually shielded, and leave the rest in the source addresses (the documented behavior). The shield action transform reallocates the per-input debits so the total consumed equals exactly the shielded amount; the unused remainder stays put. Nothing about the Orchard proof depends on how the transparent side is split, so this is safe. Net result: addresses lose `amount + fee`, the pool gains `amount`, the fee pools gain `fee` — **credits conserved**.

### 2. The shielded fees

- **`Shield`** (transparent → pool) is charged like any other address-funded transition: GroveDB **meters** the real storage and processing of the note/nullifier writes, and only the ZK **compute** fee `compute_shielded_verification_fee(num_actions)` (Halo 2 proof verification + per-action processing, **no storage term**) is added on top — via `additional_fixed_fee_cost`, the same mechanism `IdentityCreateFromAddresses` uses. Storage is never double-counted, and conservation (deduct == book) holds by the **standard machinery** with no special-case override.
[...]

PR #3809, opened 2026-06-08
fix(sdk): default initial protocol version to 10 when unpinned (upgrade-safe ratchet floor)

Claudius-Maginificent:

**Problem:** The SDK can already auto-detect the network's protocol version and follow it — but that behaviour was *opt-in*. You had to know to call `with_initial_version()` to turn it on, and almost nobody did. So an unpinned SDK fell back to `PlatformVersion::latest()`. During an upgrade window that is *higher* than the network, and because version auto-update only ratchets **upward**, a too-high SDK can never come back down — it stays stuck shipping a wire format the network rejects.

**Fix:** Make auto-detect the default. An unpinned SDK now seeds from a safe floor (`DEFAULT_INITIAL_PROTOCOL_VERSION`, currently `10`) and ratchets up to whatever the network reports. No special call required.

**Opt out:** Callers who want a fixed version call `with_version(...)`, which pins the version and disables auto-detect. Explicit configuration behaves exactly as before.

## What was done?

- Added `DEFAULT_INITIAL_PROTOCOL_VERSION` (currently `10`) — a bumpable, upgrade-safe floor.
- `SdkBuilder::default()` now seeds the protocol version from this floor instead of `latest()`, so auto-detect is on by default.
- **Removed `with_initial_version()` from the public API** (now a `#[cfg(test)]` test-only seed). Auto-detect being the default makes the old opt-in redundant; `with_version()` is the sole public way to pin + opt out.
- Verified the auto-detected `protocol_version` is cryptographically trustworthy: it is bound to the quorum BLS threshold signature via the Tenderdash `StateId.app_version`, and the version ratchet runs only after proof + signature verification succeeds. Added a guard comment at the verify-before-ratchet call site plus a regression test asserting the ratchet rejects unknown / non-upward inputs.
- Count/`group_by` mock tests (which need a fixed V1-documents wire version) now pin via `with_version(PV12)` instead of the now-hidden override.
- Docs aligned with the new contract: auto-detect is the default; `with_version()` = pin + opt out.

Scope: `packages/rs-sdk` (the change + its tests).

## How Has This Been Tested?

- `cargo test -p dash-sdk --lib` — passing, including new tests for the floor default, upward-only ratchet, and explicit-pin override.
- `cargo clippy -p dash-sdk -p rs-sdk-ffi --all-targets --locked -- -D warnings` — clean.
- `cargo test -p dash-sdk --doc` and `cargo fmt` — clean.

## Breaking Changes

For end users: none — only the *unpinned* default changes (boots at the floor and ratchets up instead of starting at `latest`); pinning via `with_version()` is unchanged. For SDK source callers: `with_initial_version()` is no longer public (it was the rarely-used opt-in auto-detect seed, now the default). This API-surface reduction lands within the `v3.1-dev` pre-release cycle; migrate any such caller to the default (auto-detect) or `with_version()` to pin.

## Checklist:


thepastaclaw:
✅ Review complete (commit 892365d9)

packages/rs-sdk/src/sdk.rs
```diff
@@ -759,7 +768,8 @@ impl Default for SdkBuilder {
 
             cancel_token: CancellationToken::new(),
 
-            version: PlatformVersion::latest(),
+            version: PlatformVersion::get(DEFAULT_INITIAL_PROTOCOL_VERSION)
+                .expect("DEFAULT_INITIAL_PROTOCOL_VERSION must be a known PlatformVersion"),
```
thepastaclaw, line 795: **🔴 Blocking: Default PV10 seed breaks first PV12-only queries before auto-detect can ratchet**

Seeding every unpinned SDK at PV10 means the first request is encoded with PV10 because `Fetch::fetch_with_metadata_and_proof` (packages/rs-sdk/src/platform/fetch.rs:177) reads `sdk.query_settings()` (sdk.rs:520, returning `self.version()`) and runs the `DocumentQuery -> GetDocumentsRequest` encoder synchronously, while `maybe_update_protocol_version` only fires after response metadata is verified. With PV10, `drive_abci.query.document_query.default_current_version = 0` (drive_abci_query_versions/v0.rs:23), so `encode_v0` rejects any `select` other than `Documents` with a local `Error::Config` at document_query.rs:508. Concretely, `DocumentCount::fetch` (which requires `SelectProjection::count_star()` via `assert_select_is_count`) cannot make a first call against the current PV12 network on an unpinned SDK: it fails before reaching the node, so auto-detect never sees a metadata response to ratchet on, and the caller must either pin via `with_version`/`with_initial_version` or first issue some unrelated ratcheting query. `group_by` / `having` paths fail the same way. The fix needs to either (a) pre-ratchet via a cheap version probe before encoding, (b) gate the V0-only encode on a confirmed-old network rather than the seed alone, or (c) document the requirement to pin when using V1-only document_query surfaces.

<sub>source: ['codex']</sub>

packages/rs-sdk/src/sdk.rs
```diff
@@ -1663,6 +1675,75 @@ mod test {
...
+        let floor = super::DEFAULT_INITIAL_PROTOCOL_VERSION;
+        assert_eq!(sdk.protocol_version_number(), floor);
+
+        // Ratchet up: a newer network version raises the floor.
+        sdk.maybe_update_protocol_version(floor + 2);
+        assert_eq!(
+            sdk.protocol_version_number(),
+            floor + 2,
+            "auto-detect must ratchet upward from the floor"
+        );
```
thepastaclaw: **🟡 Suggestion: Ratchet test uses `floor + 2` and will silently break when the floor is bumped**

`test_default_floor_ratchets_up_but_never_down` exercises the ratchet by calling `sdk.maybe_update_protocol_version(floor + 2)` and asserting the atomic moved to `floor + 2`. Today this works because `floor=10` and `PlatformVersion::latest()=PV12`, so `floor + 2 = 12` is a known version. The doc comment on `DEFAULT_INITIAL_PROTOCOL_VERSION` (sdk.rs:53-58) explicitly invites bumping the floor as the network's supported floor advances. The moment the floor is raised to 11, `floor + 2 = 13` is unknown; `maybe_update_protocol_version` hits the `PlatformVersion::get(received_version).is_err()` guard at sdk.rs:328, silently keeps the seed, and the assertion fails with a message that points at the ratchet rather than the actual cause. Anchor the ratchet target to a known version greater than the floor (e.g. `PlatformVersion::latest().protocol_version` with a precondition `assert!(latest > floor)`) instead of arithmetic on the floor.

<sub>source: ['claude']</sub>

thepastaclaw (changes requested):
## Code Review

The PR seeds unpinned SDKs at PV10 instead of latest() so they can ratchet upward across an upgrade window. The change itself is small, well-tested, and explicit pins are unaffected. However, the new default introduces a real first-request regression for SDK surfaces that require PV12-only document_query dispatch (e.g. DocumentCount), because the wire is encoded synchronously from sdk.query_settings() before any metadata is observed, and encode_v0 rejects non-Documents selects locally. There is also a test that couples `floor + 2` to PlatformVersion::get returning Ok, which silently regresses the moment the floor constant is bumped.

<details>

```

- [BLOCKING] packages/rs-sdk/src/sdk.rs:771-772: Default PV10 seed breaks first PV12-only queries before auto-detect can ratchet
  Seeding every unpinned SDK at PV10 means the first request is encoded with PV10 because `Fetch::fetch_with_metadata_and_proof` (packages/rs-sdk/src/platform/fetch.rs:177) reads `sdk.query_settings()` (sdk.rs:520, returning `self.version()`) and runs the `DocumentQuery -> GetDocumentsRequest` encoder synchronously, while `maybe_update_protocol_version` only fires after response metadata is verified. With PV10, `drive_abci.query.document_query.default_current_version = 0` (drive_abci_query_versions/v0.rs:23), so `encode_v0` rejects any `select` other than `Documents` with a local `Error::Config` at document_query.rs:508. Concretely, `DocumentCount::fetch` (which requires `SelectProjection::count_star()` via `assert_select_is_count`) cannot make a first call against the current PV12 network on an unpinned SDK: it fails before reaching the node, so auto-detect never sees a metadata response to ratchet on, and the caller must either pin via `with_version`/`with_initial_version` or first issue some unrelated ratcheting query. `group_by` / `having` paths fail the same way. The fix needs to either (a) pre-ratchet via a cheap version probe before encoding, (b) gate the V0-only encode on a confirmed-old network rather than the seed alone, or (c) document the requirement to pin when using V1-only document_query surfaces.
- [SUGGESTION] packages/rs-sdk/src/sdk.rs:1709-1715: Ratchet test uses `floor + 2` and will silently break when the floor is bumped
  `test_default_floor_ratchets_up_but_never_down` exercises the ratchet by calling `sdk.maybe_update_protocol_version(floor + 2)` and asserting the atomic moved to `floor + 2`. Today this works because `floor=10` and `PlatformVersion::latest()=PV12`, so `floor + 2 = 12` is a known version. The doc comment on `DEFAULT_INITIAL_PROTOCOL_VERSION` (sdk.rs:53-58) explicitly invites bumping the floor as the network's supported floor advances. The moment the floor is raised to 11, `floor + 2 = 13` is unknown; `maybe_update_protocol_version` hits the `PlatformVersion::get(received_version).is_err()` guard at sdk.rs:328, silently keeps the seed, and the assertion fails with a message that points at the ratchet rather than the actual cause. Anchor the ratchet target to a known version greater than the floor (e.g. `PlatformVersion::latest().protocol_version` with a precondition `assert!(latest > floor)`) instead of arithmetic on the floor.
```
</details>

packages/rs-sdk/src/sdk.rs
Claudius-Maginificent: Addressed via option (c) (intentional design + documented).

The PV10 default is a deliberate **upgrade-safe floor**: starting below the network's version lets ratchet-up auto-detection converge correctly when some nodes still lag during an upgrade window (the inverse — seeding `latest()` — leaves the SDK permanently stuck too high, since the ratchet only moves up).

For the v3.1+-only surfaces you correctly flag (Count / group_by / having), which can't ratchet passively because they fail in the local V0 encoder before producing a response, the `DEFAULT_INITIAL_PROTOCOL_VERSION` rustdoc now documents two unlock paths:
- (a) pin a higher initial version via `with_initial_version(...)`, or
- (b) issue one ratcheting query right after `build()` — e.g. `ExtendedEpochInfo::fetch_current(&sdk).await` — which works at the floor, returns response metadata, and ratchets the protocol version up; subsequent count/group_by/having queries then encode at V1.

The affected tests now pin PV12 explicitly (`dpp::version::v12::PROTOCOL_VERSION_12`). Fixed in 2f455220f4.

packages/rs-sdk/src/sdk.rs
Claudius-Maginificent: Fixed in 2f455220f4. The ratchet test no longer uses `floor + 2`; it targets the hardcoded `dpp::version::v12::PROTOCOL_VERSION_12` with an `assert!(PROTOCOL_VERSION_12 > floor, …)` precondition — so a future floor bump fails loudly with the right message instead of silently tripping the unknown-version guard.

thepastaclaw (commented):
## Code Review

Both prior nitpicks remain STILL VALID at head 2f455220 — the latest delta did not touch the docstring framing in either location. The delta also introduces a new minor docs issue: the warm-up doctest comment shows a stale `SdkBuilder::new_mainnet(addrs)` signature even though `new_mainnet` is now parameterless. No blocking, security, or consensus issues introduced by this PR; the PV10 floor + upward ratchet logic is sound and the test pinning is appropriate.

💬 4 nitpick(s)

packages/rs-sdk/src/sdk.rs
```diff
@@ -50,6 +50,49 @@ pub const DEFAULT_CONTRACT_CACHE_SIZE: usize = 100;
...
+///
+/// Example of option 2 — a parameterless current-state fetch warms up the ratchet:
+///
+/// ```no_run
+/// # use dash_sdk::{Sdk, SdkBuilder};
+/// # use dash_sdk::platform::fetch_current_no_parameters::FetchCurrent;
+/// # use dpp::block::extended_epoch_info::ExtendedEpochInfo;
+/// # async fn warm_up() -> Result<(), dash_sdk::Error> {
+/// // Option 1 alternative: SdkBuilder::new_mainnet(addrs)
+/// //   .with_initial_version(dpp::version::PlatformVersion::latest()).build()?;
```
thepastaclaw: **💬 Nitpick: Warm-up doctest comment uses stale `SdkBuilder::new_mainnet(addrs)` signature**

The doctest example added in this push includes a commented-out alternative: `// Option 1 alternative: SdkBuilder::new_mainnet(addrs)` followed by `//   .with_initial_version(dpp::version::PlatformVersion::latest()).build()?;`. Per the same diff, `new_mainnet()` and `new_testnet()` are parameterless (line 868) and now source bootstrap addresses from `dash_network_seeds`. Because the snippet sits in a `//` comment inside `no_run`, the doctest still compiles, but a reader copying the alternative gets a compile error. Either drop the `(addrs)` argument or drop the comment entirely.

<sub>source: ['claude']</sub>

packages/rs-sdk/src/sdk.rs
```diff
@@ -50,6 +50,49 @@ pub const DEFAULT_CONTRACT_CACHE_SIZE: usize = 100;
...
+/// Under V0 the local query encoder rejects the v3.1+-only surfaces — `Count`
+/// (`SelectProjection::count_star`), `group_by`, and `having` — with an
+/// [`Error::Config`] *before* any network round-trip. To use them, either:
+///
+/// 1. Pin a higher initial version at build time via
+///    [`SdkBuilder::with_initial_version`], or
+/// 2. Issue one ratcheting query right after `build()`. Any normal query works
+///    at the floor and its response metadata ratchets the SDK's protocol version
+///    up to the network's actual version; subsequent `Count` / `group_by` /
+///    `having` queries then encode correctly.
```
thepastaclaw: **💬 Nitpick: "Any normal query works at the floor" overstates floor compatibility**

The new guidance says `Any normal query works at the floor and its response metadata ratchets the SDK's protocol version`. The same push had to pin `test_mock_fetch_data_contract` to PV12 because the v3.1+ mock data contract round-trips differently under PV10. The recommendation is fine in spirit — and the concrete `ExtendedEpochInfo::fetch_current` example below is floor-safe — but the prose understates that the warm-up query must itself be PV10-compatible (no v3.1+ document-type or query-surface dependencies). Reframing the sentence around "issue one floor-compatible ratcheting query, for example the current-state fetch below" matches reality more precisely.

<sub>source: ['codex']</sub>

thepastaclaw (commented):
## Code Review

Latest push (8cea833d) is a merge of v3.1-dev carrying the unrelated shielded-fees PR #3793; the in-scope SDK delta (DEFAULT_INITIAL_PROTOCOL_VERSION=PV10 + test pinning) is unchanged since 2f455220. All four prior documentation nitpicks on packages/rs-sdk/src/sdk.rs are STILL VALID at HEAD and carried forward. One new convergent suggestion: state-transition broadcast/wait does not ratchet from response metadata before proof verification, so a fresh unpinned SDK now starts at the PV10 floor when verifying the first transition proof against a potentially-newer network.

<details>

```

In `packages/rs-sdk/src/platform/transition/broadcast.rs`:
- [SUGGESTION] packages/rs-sdk/src/platform/transition/broadcast.rs:182-188: broadcast_and_wait verifies first-call proof at the PV10 floor without ratcheting from response metadata
  Confirmed at broadcast.rs:182-188: `Drive::verify_state_transition_was_executed_with_proof(..., sdk.version())` is called directly with `sdk.version()` and there is no `sdk.verify_response_metadata(metadata)` call earlier in this wait path to ratchet from `metadata.protocol_version` before proof verification. After this PR, an unpinned SDK seeds at `DEFAULT_INITIAL_PROTOCOL_VERSION` (PV10) instead of `latest()`. If a freshly built, unpinned SDK's very first operation is a `broadcast_and_wait` on a v3.1+ network, the transition-result proof is verified against PV10 while the network proof shape may correspond to a newer PV — this is the same first-call bootstrap window that sdk.rs:394-408 documents for query proof parsing, but the broadcast/wait path has no equivalent ratchet hook before `verify_state_transition_was_executed_with_proof`. Either ratchet from `metadata.protocol_version` before proof verification (mirroring `verify_response_metadata`), or extend the bootstrap-limitation doc to explicitly cover broadcast_and_wait and recommend `with_initial_version` for callers whose first op is a transition.
```
</details>

packages/rs-sdk/src/sdk.rs
```diff
@@ -50,6 +50,38 @@ pub const DEFAULT_CONTRACT_CACHE_SIZE: usize = 100;
...
+/// an upgrade window. Bump this constant as the network's supported floor advances.
+///
+/// # v3.1+-only query surfaces
+///
+/// At this floor (PV10, V0 documents wire) the local encoder rejects the
+/// v3.1+-only surfaces — `Count` (`SelectProjection::count_star`), `group_by`,
+/// and `having` — with [`Error::Config`] *before* any network round-trip. To use
+/// them either pin a higher initial version via [`SdkBuilder::with_initial_version`],
+/// or issue one floor-compatible ratcheting query (no v3.1+ surfaces) right after
+/// `build()` — e.g. the `ExtendedEpochInfo::fetch_current` current-state fetch below.
```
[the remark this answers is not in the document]
Claudius-Maginificent: Done in `ea76193c` — propagated the new floor contract to the other public docs:

- `with_initial_version()` no longer claims the `version()` fallback is `latest()`; it now documents that an unpinned SDK seeds to `DEFAULT_INITIAL_PROTOCOL_VERSION` (the floor) and ratchets up via `fetch_max` once response metadata is observed.
- `rs-sdk-ffi` `DashSDKConfig.platform_version = 0` now reads "auto-detect seeded at the default initial protocol-version floor, ratcheting up to the network's version" instead of "auto-detect / latest".

packages/rs-sdk/src/sdk.rs
```diff
@@ -50,6 +50,38 @@ pub const DEFAULT_CONTRACT_CACHE_SIZE: usize = 100;
...
+/// Its response metadata lifts the SDK to the network's version, after which `Count` /
+/// `group_by` / `having` encode correctly.
```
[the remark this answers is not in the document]
Claudius-Maginificent: Done in `ea76193c` — dropped the literal `(PV10, V0 documents wire)` from the bumpable-floor doc. It now reads "At the default floor the local encoder rejects the v3.1+-only surfaces …", so the exported docs stay correct when the floor constant is bumped.

packages/rs-sdk/tests/fetch/document_count.rs
```diff
@@ -28,24 +28,40 @@
...
 };
 use dpp::data_contract::document_type::accessors::DocumentTypeV0Getters;
 use dpp::platform_value::Value;
+use dpp::version::PlatformVersion;
 use drive::query::conditions::{WhereClause, WhereOperator};
 use drive::query::ordering::OrderClause;
 use drive::query::SelectProjection;
 use drive_proof_verifier::{DocumentCount, DocumentSplitCounts, SplitCountEntry};
 
+/// Mock SDK seeded to PV12 — the first release wiring `DRIVE_ABCI_QUERY_VERSIONS_V1`
```
lklimek: Remove that hardcoded v12 and regen test vectors

packages/rs-sdk/src/sdk.rs
```diff
@@ -879,9 +912,11 @@ impl SdkBuilder {
 
     /// Configure platform version.
     ///
-    /// Select specific version of Dash Platform to use.
+    /// Select specific version of Dash Platform to use. This pins the version and
+    /// disables auto-detection.
     ///
-    /// Defaults to [PlatformVersion::latest()].
+    /// When unset, the SDK starts at [`DEFAULT_INITIAL_PROTOCOL_VERSION`] and
```
lklimek, line 918: We need to check if the version discovered by our version update mechanism is signed, and correctly verified. I think it should be part of the StateID.

If verification will succeed,  make the with_initial_version() function non-pub, as we don't want to expose it anymore. We just automatically detect and upgrade version after first request. User can use with_version() to enforce his preferred version. Update doc comments accordingly.

thepastaclaw (commented):
## Code Review

Three prior findings (#1 suggestion, #2 and #3 nitpicks) are STILL VALID at head e9f9323a and carried forward unchanged. Prior nitpicks #4 (stale warm-up doctest) and #5 (overbroad floor-compatibility claim) are FIXED in the latest delta. The agents converged on all three carried-forward items; no new defects were introduced by the most recent push.

<details>

```

In `packages/rs-sdk/src/platform/transition/broadcast.rs`:
- [SUGGESTION] packages/rs-sdk/src/platform/transition/broadcast.rs:182-188: broadcast_and_wait verifies the first execution proof at the PV10 floor without ratcheting from response metadata
  At broadcast.rs:153-188, `metadata` is extracted (line 154) and used only for `block_info`. The proof is then passed to `Drive::verify_state_transition_was_executed_with_proof(..., sdk.version())` with no prior `sdk.verify_response_metadata(metadata)` call (or equivalent) to feed `metadata.protocol_version` into `maybe_update_protocol_version`. Because this PR makes the default unpinned seed `DEFAULT_INITIAL_PROTOCOL_VERSION` (PV10), a fresh auto-detect SDK whose first network interaction is a state-transition broadcast will verify the execution proof at the PV10 floor even when the response advertises a newer network protocol. If proof-verification dispatch differs between PV10 and the network's actual PV, the first broadcast-and-wait may spuriously fail before the SDK ever ratchets — the same bootstrap risk this PR's `parse_proof_with_metadata_and_proof` doc calls out, but applied to a path that has no ratchet step. Ratcheting from `metadata.protocol_version` before computing the verifier's version would close the gap and align this path with the documented bootstrap contract.
```
</details>

lklimek:
@shumkov — re: your question on whether the auto-detected protocol version is signed and correctly verified (*"I think it should be part of the StateID"*): investigated it, and the answer is **yes — it already is.** Verdict: **VERIFIED.**

`ResponseMetadata.protocol_version` is cryptographically bound to the quorum's BLS threshold signature via the Tenderdash `StateId.app_version` field. A malicious or MITM DAPI node cannot forge it — a tampered value produces an invalid signature, which fails verification and aborts the response *before* the version ratchet ever runs.

### Trust-model trace (file:line)

- **`rs-drive-proof-verifier/src/verify.rs:39,56`** — `metadata.protocol_version` is read into `StateId.app_version`; that `StateId` is hashed (`:63-67`), embedded in the `CanonicalVote` (`:75`), and the BLS sign-digest is verified against the quorum public key (`:114`, returns `Err(InvalidSignature)` on mismatch).
- **`tenderdash-abci v1.5.1` `signatures.rs:271-279`** — `StateId::sign_bytes` prost-encodes the *entire* `StateId` (including `app_version`), then SHA-256s it. Primitive-level proof the field is genuinely inside the signed preimage. Version pinned by git tag + commit in `Cargo.lock:7305-7307`.
- **`rs-drive-proof-verifier/src/proof.rs:316 → :318`** — every `FromProof` impl calls `verify_tenderdash_proof(proof, mtd, &root_hash, …)?` and then returns `mtd.clone()`. The metadata that gets verified *is* the metadata that gets returned and ratcheted — no swap window.
- **`rs-sdk/src/sdk.rs:421` before `:423`** — the ratchet (`maybe_update_protocol_version`) runs **only after** proof + signature verification `?`-succeeds. The only production callers are `fetch.rs:204` / `fetch_many.rs:248`; there is no unverified path that writes the version.

### Threat model

Forging `protocol_version` changes the `StateID` hash → BLS verification fails → the response aborts before the ratchet. To succeed, an attacker needs the quorum threshold private key (i.e. consensus compromise), which is outside the DAPI-node boundary. Even hypothetically, the known-version + upward-only `fetch_max` guards cap the blast radius at DoS — never acceptance of forged data. **Severity: INFO** — a positive finding; the mechanism is sound by construction.

### Conclusion

The gate is satisfied (*"if verification will succeed, make `with_initial_version()` non-pub"*), so it's done — pushed in `bcf3b14`:

- `with_initial_version()` is no longer public (now a `#[cfg(test)]` test-only seed); auto-detect is the default and `with_version()` is the documented opt-out.
- The count mock tests that needed a fixed wire version now pin via `with_version(PV12)` instead of the (now-hidden) override.
- Added a guard comment at the verify-before-ratchet call site plus a regression test asserting the version ratchet rejects unknown / non-upward inputs, locking the "verify before ratchet" ordering into an enforced invariant rather than a convention.

thepastaclaw (commented):
## Code Review

Incremental verification against bcf3b145. Prior finding #2 (with_initial_version doc) is FIXED — `with_initial_version` is now `#[cfg(test)] pub(crate)` with rewritten docs. Prior findings #1 (broadcast.rs verify-before-ratchet missing) and #3 (parse_proof failure-mode paragraph still references latest()) are STILL VALID at current head and are carried forward. No genuinely new in-scope defects found in the e9f9323..bcf3b145 delta: the shielded operations.rs doc nit belongs to merged PR #3843, and the Swift/memo FFI findings are pre-existing behavior unrelated to this PR's auto-detect floor change.

Carried-forward prior findings: #1 remains valid in `packages/rs-sdk/src/platform/transition/broadcast.rs:153-188`: the broadcast wait path verifies the first execution proof using `sdk.version()` but never consumes response metadata to ratchet the SDK afterward. #3 remains valid in `packages/rs-sdk/src/sdk.rs:385-393`: the `parse_proof_with_metadata_and_proof` bootstrap note still describes the old `latest()`-vs-older-network failure direction instead of the new PV10-floor-vs-newer-network direction. Prior finding #2 is fixed.

<details>

```

In `packages/rs-sdk/src/platform/transition/broadcast.rs`:
- [SUGGESTION] packages/rs-sdk/src/platform/transition/broadcast.rs:153-188: broadcast_and_wait never ratchets protocol_version from response metadata
  This is the same shape of bug this PR fixes for query proofs, but the broadcast wait path was missed. At lines 153-160 the response `metadata` is extracted and consumed only to build `block_info`. The execution proof is then verified at lines 182-188 with `sdk.version()` and the function returns — nowhere on this path does `sdk.verify_response_metadata(method_name, &metadata)` (or any other entry into `maybe_update_protocol_version`) get called. A grep of broadcast.rs confirms neither symbol appears.

  For an unpinned SDK whose first traffic is a broadcast — exactly the workflow the new auto-detect default is supposed to make upgrade-safe — `sdk.version()` is permanently pinned at `DEFAULT_INITIAL_PROTOCOL_VERSION` (PV10). On a PV12 network the very first execution proof is verified against the wrong wire format and there is no ratchet path to ever self-correct; mixed workflows hide this only because an earlier fetch ratcheted the SDK first.

  Mirror the sequencing now codified at sdk.rs:421-425: call `sdk.verify_response_metadata(method_name, &metadata)` *after* `Drive::verify_state_transition_was_executed_with_proof` returns `Ok`, so the trusted metadata is consumed only once the proof+signature gate has passed. This preserves the verify-before-ratchet security invariant while extending the PR's auto-detect contract to broadcast-only callers.
```
</details>

thepastaclaw (commented):
## Code Review

Incremental delta (bcf3b145..438ce907) is test-infrastructure only (mock metadata reports LATEST_VERSION + bootstrap helper replacing with_version(PV12) pins via proven fetch_current) and introduces no new defects. Both prior findings reconcile as STILL VALID at current head and are carried forward: (A1) broadcast_and_wait still verifies the execution proof with the cached sdk.version() and never invokes verify_response_metadata, so the auto-detect ratchet this PR establishes for the query path is absent on the write path; (A2) the parse_proof_with_metadata_and_proof failure-mode paragraph still describes the pre-PR latest()-based direction, inconsistent with the rewritten lead-in that names DEFAULT_INITIAL_PROTOCOL_VERSION as the floor.

<details>

```

In `packages/rs-sdk/src/platform/transition/broadcast.rs`:
- [SUGGESTION] packages/rs-sdk/src/platform/transition/broadcast.rs:153-209: broadcast_and_wait never ratchets protocol_version from response metadata (carried forward, STILL VALID)
  Verified at HEAD 438ce907 — unchanged. The response `metadata` is extracted at lines 154-157 and consumed solely to build `block_info` (lines 158-160). The execution proof is then verified at lines 182-188 with `sdk.version()` and `context_provider.as_contract_lookup_fn(sdk.version())`, with no call to `Sdk::verify_response_metadata` or `maybe_update_protocol_version` anywhere on this path (confirmed by reading lines 140-220 and absence of the symbol in this file).

  This is the same shape of bug the PR fixes for the query path in `Sdk::parse_proof_with_metadata_and_proof` (sdk.rs:422-429), but the broadcast wait path was not updated. With this PR's new default — unpinned SDKs seed at `DEFAULT_INITIAL_PROTOCOL_VERSION` (currently 10) and ratchet upward only via proven query metadata — a caller whose first network interaction is `broadcast_and_wait` against a newer-version network will (a) verify the execution proof with the floor version (risking proof misinterpretation when wire formats diverge across versions), and (b) never learn the network version from this response, so the next call still starts from the floor.

  Fix must respect the PR's documented security invariant (sdk.rs:422-425): proof+signature verification (`Drive::verify_state_transition_was_executed_with_proof`'s `Ok(_)` arm) must precede the ratchet so the metadata being consumed is already trusted. Concretely: after the existing match at 182-208 yields `Ok`, call `sdk.verify_response_metadata(request.method_name(), &metadata)` and propagate through `ExecutionResult`, mirroring the query path. Reordering the ratchet before the proof check would break the invariant.
```
</details>

shumkov (approved):

thepastaclaw (commented):
## Code Review

PR #3809 makes auto-detect the SDK default by seeding unpinned SDKs from DEFAULT_INITIAL_PROTOCOL_VERSION and removing with_initial_version() from the public API. The latest delta 438ce907..892365d9 is overwhelmingly an upstream merge of v3.1-dev bringing in unrelated already-merged PRs (#3811, #3863, #3865, #3862, etc.); no new in-scope defects were introduced. Both prior findings remain STILL VALID at HEAD: broadcast_and_wait still does not ratchet protocol_version from verified response metadata, and the sdk.rs bootstrap-limitation paragraph still describes the pre-PR latest()-based failure direction.

<details>

```

In `packages/rs-sdk/src/platform/transition/broadcast.rs`:
- [SUGGESTION] packages/rs-sdk/src/platform/transition/broadcast.rs:153-209: broadcast_and_wait never ratchets protocol_version from verified response metadata
  Carried forward — STILL VALID at HEAD 892365d9. The wait path extracts `metadata` at lines 154-157 and consumes it solely to build `block_info` (158-160). The execution proof is then verified at lines 182-188 using `sdk.version()` and `context_provider.as_contract_lookup_fn(sdk.version())`, with no call to `Sdk::verify_response_metadata` or `maybe_update_protocol_version` anywhere on this path. The only callers of `verify_response_metadata` outside tests are on the query path (`parse_proof_with_metadata_and_proof` at sdk.rs:426).

  With this PR seeding unpinned SDKs at `DEFAULT_INITIAL_PROTOCOL_VERSION` (PV10) instead of `latest()`, the broadcast path becomes a real bootstrap hazard: a first `broadcast_and_wait` against a newer-version network will verify/interpret the execution proof with the floor version and never learn the network's actual version from response metadata. A broadcast-only client (e.g., a wallet whose first call is a submit) never converges, while query-side calls correctly ratchet. The right fix is symmetrical to the query path: after the proof verifies successfully (around line 211), call `sdk.verify_response_metadata(method_name, &metadata)` to perform the same upward-only `maybe_update_protocol_version` against the trusted, signature-bound metadata. The verify-before-ratchet ordering the PR's regression test enforces must be preserved.
```
</details>

lklimek:
Write path will be addresses in a separate follow-up PR.

PR #3712, opened 2026-05-20
test(swift-sdk): first swift sdk integration tests with local network

ZocoLini:
What

Adds a Swift SDK integration-test suite that runs against a local dashmate devnet (Core in regtest), exercising the Core (SPV/wallet) and Platform layers end-to-end. Includes a run_integration_tests.sh runner that spins up a fresh
devnet, builds the FFI, runs the tests under ASan, and tears the devnet down afterward. Gated by RUN_INTEGRATION_TESTS=1, so normal CI/builds are unaffected.

Coverage  
  - Wallet-to-wallet sends over SPV
  - TXO / tx-classification survival across SPV and persistor restarts
  - Locked-data-dir error surfacing
  - Core-funding → Platform identity registration

Each test runs against a shared, expensive env (devnet endpoints, RPC clients, pre-synced SPV cache) but with per-test state isolation (fresh in-memory ModelContainer, restored SPV-cache clone, swept keychain), asserted before every test.
  
Note: WalletStorage.keychainService is now overridable via DASH_KEYCHAIN_SERVICE for test isolation; no behavior change in production (defaults unchanged).

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added or updated relevant unit/integration/functional/e2e tests
- [ ] I have added "!" to the title and described breaking changes in the corresponding section if my code contains any
- [ ] I have made corresponding changes to the documentation if needed

**For repository code-owners and collaborators only**
- [ ] I have assigned this pull request to a milestone

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end integration tests for wallet send flows, SDK bootstrap and initialization, plus supporting test harness and RPC/devnet helpers.
  * Introduced utilities for subprocess/shell runs, test wallets, and shared integration environment to run suites against a local devnet.

* **Chores**
  * Reorganized and removed legacy unit test suites and example artifacts.
  * Updated CI/workflow and local test runner scripts to streamline integration test execution and devnet lifecycle; extended build script deployment target exports.

* **Documentation**
  * Fixed relative links in a PlatformWallet README.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dashpay/platform/pull/3712?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #3810, opened 2026-06-08
fix(platform-wallet): spv error propagation

ZocoLini:
Because start was being called inside the run method, and spawn in background was calling this run method without returning any start error, the spv client could fail to start and nobody was reciting the error. With this PR I force the start method to be called independently, this way start errors are no longer discarded.

Regression tests in PR https://github.com/dashpay/platform/pull/3712

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SPV startup so the run loop only begins after a successful initialization, with startup failures now correctly reported to the caller (including via FFI).
  * Refined SPV error handling: when SPV hasn’t been started, responses now consistently report “SPV Client not started” rather than a missing “SPV not running” case.

* **Refactor**
  * Restructured SPV background lifecycle to separate initialization from the background run loop, aligning runtime behavior with the manager’s SPV control flow.

### DIFF
--- a/packages/rs-platform-wallet-ffi/src/spv.rs
+++ b/packages/rs-platform-wallet-ffi/src/spv.rs
@@ -10,7 +10,7 @@ use platform_wallet::spv::{
 
 use crate::error::*;
 use crate::handle::*;
-use crate::runtime::runtime;
+use crate::runtime::{block_on_worker, runtime};
 use crate::types::FFINetwork;
 use crate::{check_ptr, unwrap_option_or_return, unwrap_result_or_return};
 
@@ -358,10 +358,23 @@ pub unsafe extern "C" fn platform_wallet_manager_spv_start(
             config.devnet = Some(devnet);
         }
 
-        let _guard = runtime().enter();
-        manager.spv_arc().spawn_in_background(config);
+        let spv = manager.spv_arc();
+        let start_result = {
+            let spv = spv.clone();
+            block_on_worker(async move { spv.start(config).await })
+        };
+
+        if start_result.is_ok() {
+            let _guard = runtime().enter();
+            spv.spawn_run_loop();
+        }
+
+        start_result
     });
-    unwrap_option_or_return!(option);
+
+    let start_result = unwrap_option_or_return!(option);
+    unwrap_result_or_return!(start_result);
+
     PlatformWalletFFIResult::ok()
 }
 
--- a/packages/rs-platform-wallet/src/spv/runtime.rs
+++ b/packages/rs-platform-wallet/src/spv/runtime.rs
@@ -97,9 +97,9 @@ impl SpvRuntime {
         tx: &Transaction,
     ) -> Result<(), PlatformWalletError> {
         let client_guard = self.client.read().await;
-        let client = client_guard
-            .as_ref()
-            .ok_or(PlatformWalletError::SpvNotRunning)?;
+        let client = client_guard.as_ref().ok_or(PlatformWalletError::SpvError(
+            "SPV Client not started".to_string(),
+        ))?;
 
         client
             .broadcast_transaction(tx)
@@ -117,9 +117,9 @@ impl SpvRuntime {
         height: u32,
     ) -> Result<[u8; 48], PlatformWalletError> {
         let client_guard = self.client.read().await;
-        let client = client_guard
-            .as_ref()
-            .ok_or(PlatformWalletError::SpvNotRunning)?;
+        let client = client_guard.as_ref().ok_or(PlatformWalletError::SpvError(
+            "SPV Client not started".to_string(),
+        ))?;
 
         let llmq_type = LLMQType::from(quorum_type as u8);
         let qh = QuorumHash::from_byte_array(quorum_hash).reverse();
@@ -132,16 +132,15 @@ impl SpvRuntime {
         Ok(*quorum.quorum_entry.quorum_public_key.as_ref())
     }
 
-    /// Run the SPV sync loop until calling [`stop`]. This blocks the current thread.
-    pub async fn run(&self, config: ClientConfig) -> Result<(), PlatformWalletError> {
-        tracing::info!("SpvRuntime::run() starting client...");
-        self.start(config).await?;
-        tracing::info!("SpvRuntime::run() client started, entering sync loop");
-
+    /// Drive the sync loop of an already-[`start`]ed client until [`stop`]
+    /// is called
+    async fn run(&self) -> Result<(), PlatformWalletError> {
         let client_guard = self.client.read().await;
         let client = client_guard
             .as_ref()
-            .ok_or(PlatformWalletError::SpvNotRunning)?
+            .ok_or(PlatformWalletError::SpvError(
+                "SPV Client not started".to_string(),
+            ))?
             .clone();
         drop(client_guard);
 
@@ -189,10 +188,11 @@ impl SpvRuntime {
         stop_result
     }
 
-    /// Spawn `run()` on the current tokio runtime and return immediately.
+    /// Spawn the sync loop of an already-[`start`]ed client on the current
+    /// tokio runtime and return immediately.
     ///
     /// Call [`stop`] to stop it
-    pub fn spawn_in_background(self: &Arc<Self>, config: ClientConfig) {
+    pub fn spawn_run_loop(self: &Arc<Self>) {
         {
             let existing = self.task.lock().expect("spv task mutex poisoned");
             if existing.is_some() {
@@ -206,8 +206,8 @@ impl SpvRuntime {
         let this = Arc::clone(self);
 
         let handle = tokio::spawn(async move {
-            if let Err(e) = this.run(config).await {
-                tracing::warn!("SpvRuntime background run exited with error: {}", e);
+            if let Err(e) = this.run().await {
+                tracing::warn!("SpvRuntime background run loop exited with error: {}", e);
             }
         });
 
@@ -252,9 +252,10 @@ impl SpvRuntime {
     /// The SPV client must be running to perform this operation.
     pub async fn clear_storage(&self) -> Result<(), PlatformWalletError> {
         let client_guard = self.client.read().await;
-        let client = client_guard
-            .as_ref()
-            .ok_or(PlatformWalletError::SpvNotRunning)?;
+        let client = client_guard.as_ref().ok_or(PlatformWalletError::SpvError(
+            "SPV Client not started".to_string(),
+        ))?;
+
         client
             .clear_storage()
             .await
@@ -266,9 +267,10 @@ impl SpvRuntime {
     /// The network cannot be changed on a running client.
     pub async fn update_config(&self, config: ClientConfig) -> Result<(), PlatformWalletError> {
         let client_guard = self.client.read().await;
-        let client = client_guard
-            .as_ref()
-            .ok_or(PlatformWalletError::SpvNotRunning)?;
+        let client = client_guard.as_ref().ok_or(PlatformWalletError::SpvError(
+            "SPV Client not started".to_string(),
+        ))?;
+
         client
             .update_config(config)
             .await
--- a/packages/rs-platform-wallet/src/error.rs
+++ b/packages/rs-platform-wallet/src/error.rs
@@ -149,9 +149,6 @@ pub enum PlatformWalletError {
     #[error("No wallets configured — add a wallet before starting SPV")]
     NoWalletsConfigured,
 
-    #[error("SPV client is not running")]
-    SpvNotRunning,
-
     #[error("SPV error: {0}")]
     SpvError(String),
 
--- a/packages/rs-platform-wallet/src/manager/accessors.rs
+++ b/packages/rs-platform-wallet/src/manager/accessors.rs
@@ -256,7 +256,7 @@ impl<P: PlatformWalletPersistence + 'static> PlatformWalletManager<P> {
     }
 
     /// Clone the `Arc<SpvRuntime>` so callers (e.g. FFI) can invoke
-    /// [`SpvRuntime::spawn_in_background`] which takes `&Arc<Self>`.
+    /// [`SpvRuntime::spawn_run_loop`] which takes `&Arc<Self>`.
     pub fn spv_arc(&self) -> Arc<SpvRuntime> {
         Arc::clone(&self.spv_manager)
     }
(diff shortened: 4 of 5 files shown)


thepastaclaw:
✅ Review complete (commit 416fabae)

packages/rs-platform-wallet-ffi/src/spv.rs
```diff
@@ -358,10 +358,19 @@ pub unsafe extern "C" fn platform_wallet_manager_spv_start(
...
         }
 
-        let _guard = runtime().enter();
-        manager.spv_arc().spawn_in_background(config);
+        let start_result = runtime().block_on(manager.spv().start(config));
+
+        if start_result.is_ok() {
+            let _guard = runtime().enter();
+            manager.spv_arc().spawn_run_loop();
+        }
```
thepastaclaw: **🟡 Suggestion: Poll `spv().start(config)` on the worker runtime, not the FFI caller thread**

`runtime().block_on(manager.spv().start(config))` polls the new (substantial) startup future — `PeerNetworkManager::new`, `DiskStorageManager::new`, `DashSpvClient::new` — directly on the foreign caller's thread. `packages/rs-platform-wallet-ffi/src/runtime.rs` explicitly documents that iOS dispatch/concurrency threads have ~512 KB stacks and that async FFI work should use `block_on_worker`, which parks the caller on a oneshot and runs the future on the 8 MB-stack worker. The rest of this crate consistently follows that pattern. Previously this work happened inside `tokio::spawn` on the runtime; the PR's new ordering reintroduces the exact stack-sensitive pattern the runtime module was built to avoid. Switch to `block_on_worker` using the `Arc<SpvRuntime>` so the future is `Send + 'static`.

```suggestion
        let spv = manager.spv_arc();
        let start_result = crate::runtime::block_on_worker(async move { spv.start(config).await });

        if start_result.is_ok() {
            let _guard = runtime().enter();
            manager.spv_arc().spawn_run_loop();
        }

        start_result
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/spv/runtime.rs
```diff
@@ -132,16 +132,16 @@ impl SpvRuntime {
...
-    /// Run the SPV sync loop until calling [`stop`]. This blocks the current thread.
-    pub async fn run(&self, config: ClientConfig) -> Result<(), PlatformWalletError> {
-        tracing::info!("SpvRuntime::run() starting client...");
-        self.start(config).await?;
-        tracing::info!("SpvRuntime::run() client started, entering sync loop");
-
+    /// Drive the sync loop of an already-[`start`]ed client until [`stop`]
+    /// is called.
+    /// [`spawn_run_loop`](Self::spawn_run_loop).
+    async fn run(&self) -> Result<(), PlatformWalletError> {
```
thepastaclaw, line 137: **💬 Nitpick: Malformed rustdoc on `run()` leaves an orphan link line**

The doc comment ends with a standalone `/// [`spawn_run_loop`](Self::spawn_run_loop).` line that has no leading sentence — an editing residue from the rename. Rustdoc will render this as an orphan paragraph.

```suggestion
    /// Drive the sync loop of an already-[`start`]ed client until [`stop`]
    /// is called. Typically invoked via
    /// [`spawn_run_loop`](Self::spawn_run_loop).
    async fn run(&self) -> Result<(), PlatformWalletError> {
```

<sub>source: ['claude']</sub>

thepastaclaw (changes requested):
## Code Review

PR correctly propagates SPV start errors to the FFI caller, but breaks the SPV integration test which still uses the now-private `run(config)` signature — `cargo test -p platform-wallet` will fail to compile. The new split-lifecycle also opens a half-success window where `spawn_run_loop` silently no-ops on a stale `JoinHandle`, and the synchronous `start()` is now polled on the (small-stack) FFI caller thread instead of via `block_on_worker`. Several stale references to the old `spawn_in_background` name remain.

🔴 1 blocking | 🟡 2 suggestion(s) | 💬 3 nitpick(s)

_4 additional finding(s) omitted (not in diff)._

<details>

```

In `packages/rs-platform-wallet/tests/spv_sync.rs`:
- [BLOCKING] packages/rs-platform-wallet/tests/spv_sync.rs:223-227: Integration test breaks compilation: `run(config)` no longer exists
  `SpvRuntime::run` is now a private, zero-argument helper, but this `#[ignore]`d test still calls `manager_for_spv.spv().run(config).await`. `#[ignore]` only skips execution — cargo still typechecks the file, so `cargo test -p platform-wallet --test spv_sync` will fail to compile after this PR. Update the test to drive the new lifecycle: `start(config).await` followed by `spv_arc().spawn_run_loop()` (and await `stop()` plus the join handle on teardown).

- [SUGGESTION] packages/rs-platform-wallet/src/spv/runtime.rs:196-216: `spv_start` can return Ok with the client started but no run loop driving it
  The new split lifecycle reintroduces the silent-success failure mode the PR is trying to eliminate. `run()` clears `self.client` when it exits naturally, but never clears `self.task`, so the `JoinHandle` of a previously error-exited run loop stays parked in `self.task`. On the next FFI call:

  1. `start()` succeeds and stores a fresh client.
  2. `spawn_run_loop()` observes `self.task.is_some()`, emits a `warn!`, and returns.
  3. The FFI returns `PlatformWalletFFIResult::ok()` — but no future is driving `client.run()`.

  Additionally, the check-and-set on `self.task` releases the mutex between the `is_some()` check and the `*self.task = Some(handle)` write, so two concurrent spawners can both observe `None` and both spawn run loops on the same client (one handle is then dropped and orphaned).

  Fix both at once by: (a) clearing `self.task` (or replacing with `JoinHandle::is_finished()` + take) at the end of `run()`, and (b) holding the `self.task` mutex across the `tokio::spawn` and store so the check-and-set is atomic. Returning a `Result` from `spawn_run_loop` and surfacing it through the FFI would also let the caller observe the failure instead of relying on logs.

- [SUGGESTION] packages/rs-platform-wallet-ffi/src/spv.rs:361-366: Poll `spv().start(config)` on the worker runtime, not the FFI caller thread
  `runtime().block_on(manager.spv().start(config))` polls the new (substantial) startup future — `PeerNetworkManager::new`, `DiskStorageManager::new`, `DashSpvClient::new` — directly on the foreign caller's thread. `packages/rs-platform-wallet-ffi/src/runtime.rs` explicitly documents that iOS dispatch/concurrency threads have ~512 KB stacks and that async FFI work should use `block_on_worker`, which parks the caller on a oneshot and runs the future on the 8 MB-stack worker. The rest of this crate consistently follows that pattern. Previously this work happened inside `tokio::spawn` on the runtime; the PR's new ordering reintroduces the exact stack-sensitive pattern the runtime module was built to avoid. Switch to `block_on_worker` using the `Arc<SpvRuntime>` so the future is `Send + 'static`.
```
</details>

packages/rs-platform-wallet-ffi/src/spv.rs
thepastaclaw: **🟡 Suggestion: SPV startup is polled on the FFI caller thread instead of block_on_worker**

`runtime().block_on(manager.spv().start(config))` polls `PeerNetworkManager::new`, `DiskStorageManager::new`, and `DashSpvClient::new` (storage + crypto + network setup) directly on the foreign caller thread. The crate's `runtime` module documents that iOS dispatch/concurrency threads have small (~512 KB) stacks and that substantial async FFI work should use `block_on_worker`, which polls on the 8 MB-stacked Tokio worker; the rest of `rs-platform-wallet-ffi` follows that pattern uniformly. Pre-PR, this work ran inside the `tokio::spawn`ed run loop and so was already on a worker stack — moving startup ahead of the spawn re-introduces the stack-sensitive path on the caller thread.

Clone the `Arc<SpvRuntime>` outside the closure, `block_on_worker(async move { spv.start(config).await })`, then on success `runtime().enter()` and call `spv.spawn_run_loop()`.

thepastaclaw (commented):
## Code Review

Verified PR head 5045e85d against prior findings. Prior #1 (broken spv_sync test) and #5 (malformed run() rustdoc orphan link) are FIXED. Prior #2 (spawn_run_loop check/set race + stale finished JoinHandle blocking restart), #3 (start polled on FFI caller thread instead of block_on_worker), #4 (stale spawn_in_background warning log), and #6 (stale FFI comment) are still present at the listed line numbers. One additional latest-delta concern surfaced by codex agents — dash-spv's own internal startup is performed in DashSpvClient::run(), so some startup-class errors will still only surface from the spawned task — is recorded as a suggestion because the PR does correctly propagate errors from PeerNetworkManager/DiskStorageManager/DashSpvClient::new. No in-scope blocking issues.

_4 additional finding(s) omitted (not in diff)._

<details>

```

- [SUGGESTION] packages/rs-platform-wallet/src/spv/runtime.rs:195-215: spawn_run_loop can silently skip spawning after a prior run-loop exit, and races check-and-set
  Two issues remain at the head:

  1. `run()` (137-156) clears `self.client` when the background loop exits but never clears `self.task`. If the prior run loop terminated (naturally or via the `tracing::warn!` error path at 210), `self.task` still holds the finished `JoinHandle`. A subsequent `platform_wallet_manager_spv_start` then: `start(config)` succeeds (client field is `None`), but `spawn_run_loop()` (197-203) sees `self.task.is_some()`, logs a warning, and returns. The FFI returns `ok()` to Swift but no task is driving the just-installed `SpvClient` — the exact silent-success failure mode this PR set out to remove, just shifted one stage later.

  2. `spawn_run_loop` holds the `task` mutex only inside the inner block (197-204), then re-acquires it at 214 to install the new `JoinHandle`. Two concurrent callers can both observe `None`, both `tokio::spawn` a run loop, and the second store overwrites the first handle — leaking a `JoinHandle` (the orphan can no longer be awaited or aborted by `stop()`) and double-driving the same `SpvClient`.

  Hold the `task` mutex across the entire check-and-set, treat `existing.as_ref().map(JoinHandle::is_finished)` as `None` so a finished handle is replaced, and consider making `spawn_run_loop` return a `Result` so FFI can surface the refusal rather than swallowing it.
- [SUGGESTION] packages/rs-platform-wallet/src/spv/runtime.rs:51-87: Some upstream dash-spv startup still happens inside run() and bypasses FFI propagation
  `SpvRuntime::start` returns Ok once `PeerNetworkManager::new`, `DiskStorageManager::new`, and `DashSpvClient::new` succeed — these errors are now properly propagated, which is the PR's stated win. However, `DashSpvClient::run()` (driven from the spawned task at 147-150) is what brings the client to the running state — connecting peers, starting the sync coordinator, etc. Failures in that internal start path still only surface via `tracing::warn!` at 210 and are not visible to the FFI caller, who has already been told `spv_start` succeeded. If the PR's intent is end-to-end startup error propagation, the upstream client's internal start needs to be invoked synchronously from `SpvRuntime::start` (before spawning the run loop) rather than from `run()`.

- [SUGGESTION] packages/rs-platform-wallet-ffi/src/spv.rs:361-366: SPV startup is polled on the FFI caller thread instead of block_on_worker
  `runtime().block_on(manager.spv().start(config))` polls `PeerNetworkManager::new`, `DiskStorageManager::new`, and `DashSpvClient::new` (storage + crypto + network setup) directly on the foreign caller thread. The crate's `runtime` module documents that iOS dispatch/concurrency threads have small (~512 KB) stacks and that substantial async FFI work should use `block_on_worker`, which polls on the 8 MB-stacked Tokio worker; the rest of `rs-platform-wallet-ffi` follows that pattern uniformly. Pre-PR, this work ran inside the `tokio::spawn`ed run loop and so was already on a worker stack — moving startup ahead of the spawn re-introduces the stack-sensitive path on the caller thread.

  Clone the `Arc<SpvRuntime>` outside the closure, `block_on_worker(async move { spv.start(config).await })`, then on success `runtime().enter()` and call `spv.spawn_run_loop()`.
```
</details>

thepastaclaw (commented):
## Code Review

Verified all five prior findings against the current head. Prior finding #2 (FFI start polling on the caller thread) is FIXED — packages/rs-platform-wallet-ffi/src/spv.rs:362-365 now wraps spv.start in block_on_worker, with the runtime entered separately before spawn_run_loop. The other four prior findings are STILL VALID at this head and are carried forward: the spawn_run_loop TOCTOU race + never-cleared stale JoinHandle, DashSpvClient::run() startup phase failures still bypassing FFI propagation, and two stale references to the removed spawn_in_background API in a log message and a comment. No new latest-delta defects beyond the carried-forward issues.

_4 additional finding(s) omitted (not in diff)._

<details>

```

- [SUGGESTION] packages/rs-platform-wallet/src/spv/runtime.rs:195-215: spawn_run_loop has check-and-set race and never recycles a finished JoinHandle
  Two related lifecycle problems remain at this head and undermine the PR's goal that `ok` from `platform_wallet_manager_spv_start` means SPV is actually running:

  1. Stale handle blocks restart silently. When the background loop exits, `run()` clears `self.client` at line 153 but never clears `self.task`. A subsequent FFI call to `platform_wallet_manager_spv_start` will: succeed at recreating the client via `spv.start(config)`, then call `spawn_run_loop`, which observes `existing.is_some()` (the finished JoinHandle from the previous run), logs a warning, and returns without spawning. FFI reports `ok`, `is_started()` returns true, but nothing drives sync.

  2. TOCTOU race between the two `self.task.lock()` acquisitions. The check at line 197 drops the lock before `tokio::spawn`, and the store at line 214 reacquires it. Two concurrent callers can both observe `None`, both spawn, and the second store orphans the first handle so `stop()` will never await or abort it.

  Fix: hold the task mutex across the entire check/cleanup/spawn/store window, and treat `JoinHandle::is_finished()` (or a `take()`) as stale so a finished handle is replaced rather than refused. Returning a `Result` would let FFI distinguish 'already running' from 'silently refused'.
- [SUGGESTION] packages/rs-platform-wallet/src/spv/runtime.rs:137-156: DashSpvClient::run startup failures still bypass FFI propagation
  `SpvRuntime::start` correctly propagates failures from `PeerNetworkManager::new`, `DiskStorageManager::new`, and `DashSpvClient::new` constructor errors up to the FFI caller. But `DashSpvClient::run()` (line 148) is what brings the client to the running state — connecting peers, starting the sync coordinator, and starting masternode/ChainLock/InstantSend managers — and its failure is observed inside the spawned task at lines 209-211 and only logged via `tracing::warn!`. By then `platform_wallet_manager_spv_start` has already returned `ok`.

  The PR description states 'start errors are no longer discarded', and callers may reasonably assume `ok` means the SPV client is genuinely running. Currently a `client.run()` startup failure leaves `self.client = Some(...)` and `self.task = Some(handle)` populated, so `is_started()` returns true while sync is dead — recreating the silent-failure mode the PR title says it is fixing. To close the gap, extract the deterministic upstream start phase into `SpvRuntime::start` before `spawn_run_loop`, or expose run-loop startup failure via a `oneshot` that FFI awaits before declaring success.
```
</details>

QuantumExplorer (approved):

PR #3814, opened 2026-06-08
fix(drive): strict merged-query verification for unshield & shielded withdrawal proofs

### LINKED ISSUE #3812: Tighten Unshield & ShieldedWithdrawal execution proofs to a single strict merged query (parity with ShieldFromAssetLock)
## Summary

The `Unshield` and `ShieldedWithdrawal` execution-proof **verify** paths check their sub-proofs with **subset** verification, which tolerates *extra* subtree data in the proof. A dishonest full node could therefore return a proof padded with unrelated state, and verification would still pass. PR #3793 tightened the sibling `ShieldFromAssetLock` surplus path to a **single strict verification against the exact merged query** the prover builds (commit `63c31f7c6e`); this issue tracks bringing `Unshield` and `ShieldedWithdrawal` to the same, tighter pattern.

This is **defense-in-depth / canonicality**, not a known exploit: in all cases the verified keys are derived from the *signed* transition (not from proof data) and the sub-proofs are bound by a shared root hash, so a padded proof cannot forge or misroute. But a proof verifier should accept exactly the data it asked for and reject the rest.

## Affected code

Both arms live in `packages/rs-drive/src/verify/state_transition/verify_state_transition_was_executed_with_proof/v0/mod.rs`:

- **`Unshield`** (~the `StateTransition::Unshield` arm): verifies nullifiers via `Drive::verify_shielded_nullifiers(...)` and the output-address balance via `Drive::verify_addresses_infos(proof, ..., verify_subset_of_proof = true, ...)`, bound by `root_hash_nf == root_hash_addr`. The prover side (`prove_state_transition/v0/mod.rs`, `Unshield` arm) already merges the nullifier and address sub-queries into one `PathQuery`.
- **`ShieldedWithdrawal`** (~the `StateTransition::ShieldedWithdrawal` arm): same shape, but the second sub-proof is the Core withdrawal **document** (a `SingleDocumentDriveQuery`) rather than an address balance. The prover merges nullifier + document sub-queries.

Both use the looser "verify each sub-query subset + bind by root hash" approach that `ShieldFromAssetLock` *used* to use.

## Proposed fix (mirror `ShieldFromAssetLock`)

For each arm, reconstruct the prover's **exact** merged `PathQuery` (the same sub-queries, same cleared limits, same `PathQuery::merge`) and verify it once with **strict** `GroveDb::verify_query_with_absence_proof`, then partition the returned `proved_key_values` by key/path into the typed results. See the `ShieldFromAssetLock` arm (post-`63c31f7c6e`) as the template.

Gotchas discovered while doing the `ShieldFromAssetLock` version (apply them here):

1. **`PathQuery::merge` clears the limit**, and the strict verifier *requires* a limit — set `merged_pq.query.limit` to a non-`None` value. A value unreachable by the legitimate result count (e.g. `u16::MAX`, or the exact terminal-key count) avoids truncating honest proofs. A limit *smaller* than the true count can early-break before a sibling subtree's succinctness check runs (false-rejects / would skip a check), so do not set it too low.
2. **Strict verification rejects extra subtree branches (lower layers), not extra absent leaf keys** within an already-descended subtree. The returned result set is rebuilt from the query's `terminal_keys`, so the partition can never see an attacker-injected key — but a negative test should pad with a whole extra **subtree** (not an extra leaf) to actually exercise the rejection.
3. Keep liveness: the existing prove/verify roundtrip tests must still pass (a merged-query reconstruction that doesn't match the prover will false-reject honest proofs).

## Tests

Add a padded-proof negative test for each arm, mirroring `test_shield_from_asset_lock_padded_proof_is_rejected_by_strict_verify` (in `shield_from_asset_lock/tests.rs`): prove a *superset* query (e.g. add an unrelated subtree like `Pools`) and assert the strict merged verify **rejects** it while the old subset verify would accept it. The existing nullifier/address/document roundtrip tests must continue to pass.

## References

- PR #3793, commit `63c31f7c6e` — the `ShieldFromAssetLock` strict-merged-query tightening (with an independent proof-soundness audit confirming it sound).
- The `ShieldFromAssetLock` verify arm is the reference implementation; `verify_addresses_infos_v0` shows the balance-element parsing to inline for the `Unshield` arm.

QuantumExplorer:

Closes #3812.

The `Unshield` and `ShieldedWithdrawal` execution-proof **verify** paths checked their sub-proofs with **subset** verification, which tolerates *extra* subtree data in the proof. A dishonest full node could therefore return a proof padded with unrelated state and verification would still pass. PR #3793 (commit `63c31f7c6e`) tightened the sibling `ShieldFromAssetLock` surplus path to a **single strict verification against the exact merged query** the prover builds; this PR brings the two remaining shielded arms to the same, tighter pattern.

This is **defense-in-depth / canonicality**, not a known exploit: in all cases the verified keys are derived from the *signed* transition (not from proof data) and the sub-proofs are bound by a shared root hash, so a padded proof could not forge or misroute. But a proof verifier should accept exactly the data it asked for and reject the rest.

## What was done?

Both arms live in `packages/rs-drive/src/verify/state_transition/verify_state_transition_was_executed_with_proof/v0/mod.rs`. The prover side (`prove_state_transition/v0/mod.rs`) already merges each arm's sub-queries into one `PathQuery`; the verifier now mirrors that exactly:

- **`Unshield`** — reconstruct the prover's merged `{nullifiers} ∪ {output-address}` `PathQuery` (same sub-queries, same cleared limits, same `PathQuery::merge`), set `limit = Some(u16::MAX)` (the strict verifier requires a non-`None` limit; `u16::MAX` is unreachable by the legitimate result set so it never truncates an honest proof), verify it once with the **strict** `GroveDb::verify_query_with_absence_proof`, then partition the returned `proved_key_values` **by path** into nullifier spend-statuses and the output-address balance (inlining the `ItemWithSumItem` decode from `verify_addresses_infos_v0`).
- **`ShieldedWithdrawal`** — same shape, but the second sub-proof is the Core withdrawal **document** (`SingleDocumentDriveQuery`). Reconstruct the merged `{nullifiers} ∪ {withdrawal-document}` query, strict-verify once, partition by path into nullifier statuses + the single withdrawal document, and decode the document element.

Notes carried over from the `ShieldFromAssetLock` work and applied here:
- `PathQuery::merge` clears the limit and the strict verifier requires one; a limit *smaller* than the true result count can early-break before a sibling subtree's succinctness check runs, so an unreachable `u16::MAX` is used.
- Strict verification rejects extra subtree *branches* (lower layers), not extra absent leaf keys within an already-descended subtree — so the negative tests pad with a whole extra **subtree**.
- The existing prove/verify roundtrip tests still pass (a mismatched reconstruction would false-reject honest proofs).

## How Has This Been Tested?

- New `test_unshield_padded_proof_is_rejected_by_strict_verify` and `test_shielded_withdrawal_padded_proof_is_rejected_by_strict_verify` (mirroring `test_shield_from_asset_lock_padded_proof_is_rejected_by_strict_verify`): execute a real transition, prove a **superset** query padded with an unrelated genesis `Pools` subtree, and assert the strict merged verify **rejects** it while the old subset verify **accepts** it — plus a liveness check that an honest proof for the production merged query strict-verifies.
- Existing roundtrip tests (`test_unshield_prove_and_verify_nullifiers_and_address`, `test_shielded_withdrawal_prove_and_verify_nullifiers_and_document`) continue to pass.
- `cargo test -p drive-abci --lib return_proof` — 9 passed (incl. all shielded roundtrips + the 2 new negatives).
- `cargo test -p drive --lib verify_state_transition_was_executed_with_proof` — 60 passed.
- `cargo check -p drive --no-default-features --features verify`, `cargo fmt --all`, and `cargo clippy -p drive --no-default-features --features verify` are clean.

## Breaking Changes

None. The proof *format* is unchanged (the prover already produced the merged proof); only the verifier is tightened to reject padded proofs. Not consensus-breaking — this path is used by light-client proof verification.

## Checklist:

## Summary by CodeRabbit

* **Tests**
  * Added regression tests to ensure padded/superset proofs are rejected by strict verification for shielded-withdrawal and unshield transitions while subset proofs remain accepted.

* **Improvements**
  * Enhanced state-transition verification to use strict merged multi-root proof checks for relevant flows, improving detection of extraneous proof data and returning a verified merged root.

### DIFF
--- a/packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/unshield/tests.rs
+++ b/packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/unshield/tests.rs
@@ -1180,6 +1180,25 @@ mod tests {
                 strict_result
             );
 
+            // And the PRODUCTION entry point — the dispatch site this change actually
+            // rewrote — MUST reject the padded proof too. Asserting only against the GroveDB
+            // primitive above would stay green if the Unshield arm regressed to rebuild a
+            // different merged query or fall back to the subset verifier; routing the padded
+            // proof through `Drive::verify_state_transition_was_executed_with_proof` locks the
+            // real code path into the test.
+            let production_result = Drive::verify_state_transition_was_executed_with_proof(
+                &transition,
+                &BlockInfo::default(),
+                &padded_proof,
+                &|_| Ok(None),
+                platform_version,
+            );
+            assert!(
+                production_result.is_err(),
+                "production unshield verifier must reject a padded proof, got {:?}",
+                production_result
+            );
+
             // Contrast: the SUBSET verifier (the looser primitive this change replaced) tolerates
             // the extra layer and ACCEPTS the same padded proof — the exact hole now closed.
             let subset_result = GroveDb::verify_subset_query_with_absence_proof(
--- a/packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/unshield/tests.rs
+++ b/packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/unshield/tests.rs
@@ -968,5 +968,230 @@ mod tests {
                 output_address
             );
         }
+
+        /// NEGATIVE test for the strict-merged tightening (mirrors
+        /// `test_shield_from_asset_lock_padded_proof_is_rejected_by_strict_verify`): a proof
+        /// that carries EXTRA data beyond `{nullifiers, output-address}` MUST be rejected by the
+        /// production verifier.
+        ///
+        /// The production verify path rebuilds the merged `{nullifiers, output-address}` query
+        /// and verifies it with the STRICT `verify_query_with_absence_proof`, whose succinctness
+        /// check rejects any proof that descends into a subtree (a lower layer) the query did not
+        /// require. Here we execute a real unshield, then have the (honest, but over-broad)
+        /// prover generate a SUPERSET proof for `{nullifiers, output-address, + the
+        /// genesis-populated Pools subtree}` and verify it against the production merged query.
+        /// The strict verifier must reject it because the proof carries an extra root-level lower
+        /// layer (`Pools`) the production query never touches.
+        ///
+        /// For contrast we also confirm the SUBSET verifier (the looser primitive this change
+        /// replaced) would have ACCEPTED the same padded proof — demonstrating exactly the hole
+        /// the strict merged verification closes.
+        #[test]
+        fn test_unshield_padded_proof_is_rejected_by_strict_verify() {
+            use drive::drive::shielded::paths::shielded_credit_pool_nullifiers_path_vec;
+            use drive::drive::RootTree;
+            use drive::grovedb::{GroveDb, PathQuery, Query, SizedQuery};
+
+            let platform_version = PlatformVersion::latest();
+            let platform = setup_platform();
+            insert_dummy_encrypted_notes(&platform, 250);
+            let mut rng = StdRng::seed_from_u64(0);
+            let pk = get_proving_key();
+
+            let spend_amount = 500_000_000u64;
+            let output_amount = 5_000u64;
+
+            // --- Create keys ---
+            let sk = SpendingKey::from_bytes([0u8; 32]).unwrap();
+            let fvk = FullViewingKey::from(&sk);
+            let recipient = fvk.address_at(0u32, Scope::External);
+            let ask = SpendAuthorizingKey::from(&sk);
+
+            // --- Create a spendable note ---
+            let rho_bytes: [u8; 32] = {
+                let mut b = [0u8; 32];
+                b[0] = 1;
+                b
+            };
+            let rho = Rho::from_bytes(&rho_bytes).unwrap();
+            let rseed = RandomSeed::from_bytes([42u8; 32], &rho).unwrap();
+            let note =
+                Note::from_parts(recipient, NoteValue::from_raw(spend_amount), rho, rseed).unwrap();
+
+            // --- Build commitment tree and get anchor + merkle path ---
+            let cmx = ExtractedNoteCommitment::from(note.commitment());
+            let mut tree = ClientMemoryCommitmentTree::new(100);
+            tree.append(cmx.to_bytes(), Retention::Marked).unwrap();
+            tree.checkpoint(0u32).unwrap();
+            let anchor = tree.anchor().unwrap();
+            let merkle_path = tree.witness(Position::from(0u64), 0).unwrap().unwrap();
+
+            // --- Build bundle: spend 500M -> output 5K (value_balance = 499,995,000) ---
+            let mut builder = Builder::<DashMemo>::new(BundleType::DEFAULT, anchor);
+            builder.add_spend(fvk.clone(), note, merkle_path).unwrap();
+            builder
+                .add_output(
+                    None,
+                    recipient,
+                    NoteValue::from_raw(output_amount),
+                    [0u8; 36],
+                )
+                .unwrap();
+
+            let (unauthorized, _) = builder.build::<i64>(&mut rng).unwrap().unwrap();
+
+            let output_address = create_output_address();
+            let unshielding_amount = 499_995_000u64;
+            let extra_sighash_data = dpp::shielded::unshield_extra_sighash_data(
+                &output_address.to_bytes(),
+                unshielding_amount,
+            );
+            let bundle_commitment: [u8; 32] = unauthorized.commitment().into();
+            let sighash = compute_platform_sighash(&bundle_commitment, &extra_sighash_data);
+
+            let proven = unauthorized.create_proof(pk, &mut rng).unwrap();
+            let bundle = proven.apply_signatures(rng, sighash, &[ask]).unwrap();
+
+            let (actions, value_balance, anchor_bytes, proof_bytes, binding_sig) =
+                serialize_authorized_bundle_i64(&bundle);
+            assert_eq!(value_balance, 499_995_000);
+
+            insert_anchor_into_state(&platform, &anchor_bytes);
+            set_pool_total_balance(&platform, 500_000_000);
+
+            let transition = create_unshield_transition(
+                output_address.clone(),
+                actions,
+                value_balance as u64,
+                anchor_bytes,
+                proof_bytes,
+                binding_sig,
+            );
+
+            let transition_bytes = transition
+                .serialize_to_bytes()
+                .expect("should serialize transition");
+
+            let platform_state = platform.state.load();
+            let transaction = platform.drive.grove.start_transaction();
+
+            let processing_result = platform
+                .platform
+                .process_raw_state_transitions(
+                    &vec![transition_bytes],
+                    &platform_state,
+                    &BlockInfo::default(),
+                    &transaction,
+                    platform_version,
+                    false,
+                    None,
+                )
+                .expect("expected to process state transition");
+
+            assert_matches!(
+                processing_result.execution_results().as_slice(),
+                [StateTransitionExecutionResult::SuccessfulExecution { .. }]
+            );
+
+            platform
+                .drive
+                .grove
+                .commit_transaction(transaction)
+                .unwrap()
+                .expect("expected to commit transaction");
+
+            // --- Reconstruct the PRODUCTION merged query exactly as the verifier does ---
+            // {nullifiers} ∪ {output address}, each with cleared limits, then a limit that can
+            // never truncate the legitimate result set.
+            let StateTransition::Unshield(ref st) = transition else {
+                unreachable!();
+            };
+            let nullifier_keys: Vec<Vec<u8>> = st.nullifiers();
+
+            let mut nf_query = Query::new();
+            nf_query.insert_keys(nullifier_keys);
+            let nullifier_pq = PathQuery::new(
+                shielded_credit_pool_nullifiers_path_vec(),
+                SizedQuery::new(nf_query, None, None),
+            );
+
+            let mut address_pq =
+                Drive::balances_for_clear_addresses_query(std::iter::once(&output_address));
+            address_pq.query.limit = None;
+
+            let mut production_pq = PathQuery::merge(
+                vec![&nullifier_pq, &address_pq],
+                &platform_version.drive.grove_version,
+            )
+            .expect("merge production query");
+            production_pq.query.limit = Some(u16::MAX);
+
+            // Sanity: an HONEST proof for the production query verifies strictly (liveness).
+            let honest_proof = platform
+                .drive
+                .grove_get_proved_path_query(
+                    &production_pq,
+                    None,
+                    &mut vec![],
+                    &platform_version.drive,
+                )
+                .expect("honest production proof");
+            GroveDb::verify_query_with_absence_proof(
+                &honest_proof,
+                &production_pq,
+                &platform_version.drive.grove_version,
+            )
+            .expect("strict verify of honest production proof must succeed");
+
+            // --- Build a SUPERSET (padded) proof: {nullifiers, address, + an extra subtree} ---
+            // Pad by ALSO descending the genesis-populated `Pools` root subtree, which the
+            // production query never touches; the padded proof carries an extra root-level layer.
+            let mut pools_top = Query::new();
+            pools_top.insert_key(vec![RootTree::Pools as u8]);
+            pools_top.set_subquery(Query::new_range_full());
+            let pools_pq = PathQuery::new(vec![], SizedQuery::new(pools_top, None, None));
+
+            let mut superset_pq = PathQuery::merge(
+                vec![&nullifier_pq, &address_pq, &pools_pq],
+                &platform_version.drive.grove_version,
+            )
+            .expect("merge superset query");
+            superset_pq.query.limit = Some(u16::MAX);
+
+            let padded_proof = platform
+                .drive
+                .grove_get_proved_path_query(
+                    &superset_pq,
+                    None,
+                    &mut vec![],
+                    &platform_version.drive,
+                )
+                .expect("padded superset proof");
+
+            // The STRICT verifier (production behavior) MUST reject the padded proof.
+            let strict_result = GroveDb::verify_query_with_absence_proof(
+                &padded_proof,
+                &production_pq,
+                &platform_version.drive.grove_version,
+            );
+            assert!(
+                strict_result.is_err(),
+                "strict verifier must reject a proof padded with an extra subtree layer, got {:?}",
+                strict_result
+            );
+
+            // Contrast: the SUBSET verifier (the looser primitive this change replaced) tolerates
+            // the extra layer and ACCEPTS the same padded proof — the exact hole now closed.
+            let subset_result = GroveDb::verify_subset_query_with_absence_proof(
+                &padded_proof,
+                &production_pq,
+                &platform_version.drive.grove_version,
+            );
+            assert!(
+                subset_result.is_ok(),
+                "subset verifier was expected to tolerate the padded proof, got {:?}",
+                subset_result
+            );
+        }
     }
 }
--- a/packages/rs-drive/src/verify/state_transition/verify_state_transition_was_executed_with_proof/v0/mod.rs
+++ b/packages/rs-drive/src/verify/state_transition/verify_state_transition_was_executed_with_proof/v0/mod.rs
@@ -1140,14 +1140,10 @@ impl Drive {
 
                 let nullifier_keys: Vec<Vec<u8>> = st.nullifiers();
 
-                // The prove side merges the nullifier spend-status sub-query with the
-                // output-address balance sub-query into a SINGLE multi-root proof (clearing each
-                // sub-query limit before the merge, because `PathQuery::merge` rejects limited
-                // sub-queries). Reconstruct the byte-identical merged query here and verify it
-                // STRICTLY: the strict verifier accepts a proof that matches the merged query
-                // exactly and REJECTS any proof carrying extra branches, so a malicious/buggy
-                // prover cannot pad the proof with unrelated state. (See the `ShieldFromAssetLock`
-                // arm below for the full rationale on why a single strict merged verify is sound.)
+                // Reconstruct the prove side's merged query — nullifier spend-status ∪
+                // output-address balance — and verify it strictly. See
+                // `verify_merged_query_strict` for why a single strict merged verify is
+                // sound and rejects proofs padded with extra subtree branches.
                 let mut nf_query = grovedb::Query::new();
                 nf_query.insert_keys(nullifier_keys);
                 let nullifier_pq = grovedb::PathQuery::new(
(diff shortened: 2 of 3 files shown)


thepastaclaw:
✅ Review complete (commit f98de59d)

packages/rs-drive/src/verify/state_transition/verify_state_transition_was_executed_with_proof/v0/mod.rs
```diff
@@ -1133,18 +1133,110 @@ impl Drive {
...
+                    &platform_version.drive.grove_version,
                 )?;
 
+                // `verify_query_with_absence_proof` (the STRICT verifier) requires a limit, but
+                // `PathQuery::merge` clears it. The merged query targets a fixed, tiny set of
+                // explicit keys ({nullifiers} ∪ {output address}), so set a limit that can never
+                // be exhausted by the legitimate result set. An unreachable limit guarantees every
+                // layer is fully traversed and every extra branch is caught by the succinctness
+                // check (which is independent of the limit value).
+                merged_pq.query.limit = Some(u16::MAX);
```
thepastaclaw: **🟡 Suggestion: Lost u16-bound check on nullifier count when replacing verify_shielded_nullifiers_v0**

The old `verify_shielded_nullifiers_v0` set `limit = Some(u16::try_from(nullifiers.len())?)` and returned a typed error if the keyset exceeded the verifier's u16 capacity. Both new arms (Unshield here, and ShieldedWithdrawal at line 1289) now unconditionally set `merged_pq.query.limit = Some(u16::MAX)` and rely on the post-loop "all returned are spent" check. If a transition ever carried more than `u16::MAX - 1` nullifiers, the strict verifier would silently truncate the result set and the spent-check would pass on a subset rather than reject the proof. Today upstream validation almost certainly bounds nullifier count well below this, so this is defense-in-depth — but the explicit bound made the `u16::MAX` sentinel locally sound, and re-asserting it costs almost nothing. Note: the existing ShieldFromAssetLock arm at line ~1500 has the same gap; fixing all three together would keep the pattern uniform.

<sub>source: ['codex']</sub>

packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/unshield/tests.rs
```diff
@@ -968,5 +968,230 @@ mod tests {
...
+            let subset_result = GroveDb::verify_subset_query_with_absence_proof(
+                &padded_proof,
+                &production_pq,
+                &platform_version.drive.grove_version,
+            );
+            assert!(
+                subset_result.is_ok(),
+                "subset verifier was expected to tolerate the padded proof, got {:?}",
+                subset_result
+            );
```
thepastaclaw, line 1213: **🟡 Suggestion: Negative tests assert GroveDB behavior, not the Drive verifier entry point this PR changes**

The new test reconstructs `production_pq` and confirms `GroveDb::verify_query_with_absence_proof` rejects the padded proof while the subset verifier accepts it. That demonstrates the GroveDB-level distinction, but it never feeds the padded proof through `Drive::verify_state_transition_was_executed_with_proof` — which is the dispatch site this PR actually changed. If a future regression accidentally restored subset verification in the Drive arm, or built a slightly different merged query than the hand-built `production_pq`, this test would still pass. The sibling ShieldedWithdrawal negative test has the same gap. Adding an assertion that the Drive-level verify call rejects the padded proof would lock the production code path into the test.

<sub>source: ['codex']</sub>

packages/rs-drive/src/verify/state_transition/verify_state_transition_was_executed_with_proof/v0/mod.rs
```diff
@@ -1133,18 +1133,110 @@ impl Drive {
...
+                            })
+                            .transpose()?;
+
+                        balances.insert(address, balance_info);
+                    } else {
+                        return Err(Error::Proof(ProofError::CorruptedProof(
+                            "unshield proof contained an entry outside the nullifier and address subtrees".to_string(),
+                        )));
+                    }
+                }
```
thepastaclaw, line 1227: **💬 Nitpick: Unshield silently overwrites duplicate address entries; ShieldedWithdrawal explicitly rejects them**

Within the same PR, the two arms handle the "can the proof contain more than one entry under the singleton subtree?" case differently: ShieldedWithdrawal at lines 1381-1386 returns `CorruptedProof` on a second document entry, while Unshield's `balances.insert(address, ..)` would silently overwrite if a second address entry slipped through. Under the strict verifier both cases should be impossible (single queried key, no duplicates), so functionally this is fine — but the asymmetry invites confusion for future maintainers about which arm's contract is intended. Either tighten Unshield with a `balances.len()` assertion to match ShieldedWithdrawal, or drop the explicit check from ShieldedWithdrawal and rely on the verifier in both.

<sub>source: ['claude']</sub>

packages/rs-drive/src/verify/state_transition/verify_state_transition_was_executed_with_proof/v0/mod.rs
```diff
@@ -1154,25 +1246,8 @@ impl Drive {
                     }
                 }
```
thepastaclaw, line 1237: **💬 Nitpick: Cross-check returned status set against queried nullifier set**

Both arms iterate `statuses` and reject any unspent entry, but never assert that `statuses.len() == nullifier_keys.len()` or that every queried nullifier appears. The strict `verify_query_with_absence_proof` should return an entry (present or absent) for every queried key, so any missing nullifier should surface as `(key, false)` and be caught by the loop — but cross-checking the set explicitly would catch a class of verifier-completeness regressions that the spent-check loop alone wouldn't. Pure defense-in-depth; no known exploit path.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Defensive hardening PR that correctly extends the strict merged-query pattern from PR #3793 to Unshield and ShieldedWithdrawal. The verifier-side merged query is byte-identical to the prover (sub-queries with cleared limits, same merge order, post-merge u16::MAX limit), path-partitioning is safe against out-of-set entries, and the new negative tests demonstrate the closed gap. No blocking issues; a small bound-check regression and a Drive-level test-coverage gap are worth addressing.

<details>

```

In `packages/rs-drive/src/verify/state_transition/verify_state_transition_was_executed_with_proof/v0/mod.rs`:
- [SUGGESTION] packages/rs-drive/src/verify/state_transition/verify_state_transition_was_executed_with_proof/v0/mod.rs:1141-1173: Lost u16-bound check on nullifier count when replacing verify_shielded_nullifiers_v0
  The old `verify_shielded_nullifiers_v0` set `limit = Some(u16::try_from(nullifiers.len())?)` and returned a typed error if the keyset exceeded the verifier's u16 capacity. Both new arms (Unshield here, and ShieldedWithdrawal at line 1289) now unconditionally set `merged_pq.query.limit = Some(u16::MAX)` and rely on the post-loop "all returned are spent" check. If a transition ever carried more than `u16::MAX - 1` nullifiers, the strict verifier would silently truncate the result set and the spent-check would pass on a subset rather than reject the proof. Today upstream validation almost certainly bounds nullifier count well below this, so this is defense-in-depth — but the explicit bound made the `u16::MAX` sentinel locally sound, and re-asserting it costs almost nothing. Note: the existing ShieldFromAssetLock arm at line ~1500 has the same gap; fixing all three together would keep the pattern uniform.

In `packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/unshield/tests.rs`:
- [SUGGESTION] packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/unshield/tests.rs:1171-1194: Negative tests assert GroveDB behavior, not the Drive verifier entry point this PR changes
  The new test reconstructs `production_pq` and confirms `GroveDb::verify_query_with_absence_proof` rejects the padded proof while the subset verifier accepts it. That demonstrates the GroveDB-level distinction, but it never feeds the padded proof through `Drive::verify_state_transition_was_executed_with_proof` — which is the dispatch site this PR actually changed. If a future regression accidentally restored subset verification in the Drive arm, or built a slightly different merged query than the hand-built `production_pq`, this test would still pass. The sibling ShieldedWithdrawal negative test has the same gap. Adding an assertion that the Drive-level verify call rejects the padded proof would lock the production code path into the test.
```
</details>

packages/rs-drive/src/verify/state_transition/verify_state_transition_was_executed_with_proof/v0/mod.rs
```diff
@@ -1133,18 +1133,110 @@ impl Drive {
...
+                use grovedb::Element;
 
                 let nullifier_keys: Vec<Vec<u8>> = st.nullifiers();
 
-                let (root_hash_nf, statuses) = Drive::verify_shielded_nullifiers(
-                    proof,
-                    &nullifier_keys,
-                    true,
-                    platform_version,
+                // The prove side merges the nullifier spend-status sub-query with the
```
shumkov: Shall we extract this code into a function such as we had Drive::verify_shielded_nullifiers?

shumkov: Same here

shumkov: Done in 92ba30679b — extracted into `Drive::verify_merged_query_strict(proof, sub_queries, platform_version)`, which clears each sub-query's limit, merges, sets the unreachable `u16::MAX` limit, and runs the strict `verify_query_with_absence_proof`. All three arms (Unshield, ShieldedWithdrawal, and the ShieldFromAssetLock surplus branch) now call it, and the soundness rationale lives on the helper instead of being copy-pasted three times.

shumkov: Done — ShieldedWithdrawal now calls the same `Drive::verify_merged_query_strict` helper (92ba30679b).

packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/unshield/tests.rs
shumkov: Done in fcfd1fe596 — both the unshield and shielded_withdrawal padded-proof tests now route `padded_proof` through `Drive::verify_state_transition_was_executed_with_proof` and assert it returns `Err`, so a regression that rebuilt the merged query differently or fell back to the subset verifier inside the arm would fail the test. The shielded_withdrawal case loads the withdrawals system contract for the lookup fn, mirroring its positive round-trip test. (CodeRabbit raised the same point on both tests.)

packages/rs-drive/src/verify/state_transition/verify_state_transition_was_executed_with_proof/v0/mod.rs
shumkov: Done in f98de59d52 — tightened Unshield to reject a second output-address entry with `CorruptedProof`, matching ShieldedWithdrawal's reject-on-second-document. Both arms now make the singleton-subtree invariant explicit. This is strictly defense-in-depth: each address/document sub-query carries a single key, so the strict `verify_query_with_absence_proof` returns at most one entry either way — but the two arms now read consistently.

packages/rs-drive/src/verify/state_transition/verify_state_transition_was_executed_with_proof/v0/mod.rs
shumkov: Verified this is unreachable rather than a live gap, so leaving the `u16::MAX` sentinel as-is. The nullifier query is built from `st.nullifiers()` (one nullifier per Orchard action), and `validate_actions_count` — a `consensus/basic` check (`ShieldedTooManyActionsError`) wired into every shielded transition — caps `actions.len()` at `max_shielded_transition_actions = 16` (`system_limits` v1 and v2). So the merged result set is at most 17 entries against a 65535 limit; truncation cannot occur, and any transition that could reach it would be rejected at basic validation before producing a proof. The old `u16::try_from(len)` guarded the same impossible condition, so re-adding it would be a typed error branch that can never fire. Declining for now — happy to revisit if `max_shielded_transition_actions` ever grows by orders of magnitude.

packages/rs-drive/src/verify/state_transition/verify_state_transition_was_executed_with_proof/v0/mod.rs
shumkov: Verified this is already guaranteed, so not adding the explicit cross-check. `verify_query_with_absence_proof` runs with `absence_proofs_for_non_existing_searched_keys: true`, so it returns one entry per queried key — present (spent) or absent→`None` (unspent). With strict succinctness (no extra entries) and no truncation (≤ 16 nullifiers ≪ the `u16::MAX` limit), `statuses.len() == nullifier_keys.len()` always holds, and any unspent/missing nullifier surfaces as `(key, false)`, which the existing spent-check loop already rejects with `IncorrectProof`. A `statuses.len()` assertion would catch nothing the loop does not, so leaving it out.

thepastaclaw (commented):
## Code Review

Incremental review of fcfd1fe5 over ceca33f1. Latest delta: (1) refactor extracting `verify_merged_query_strict` helper used by all three shielded arms (clean dedup, behavior-preserving); (2) tests now route padded proofs through `Drive::verify_state_transition_was_executed_with_proof` and assert `is_err()`. No new in-scope blocking issues. Prior finding status: #1 (u16 bound check) STILL VALID — the refactor centralized but did not re-add the bound; #2 (test entry point) FIXED by fcfd1fe5; #3 (Unshield silent overwrite vs ShieldedWithdrawal explicit reject) STILL VALID — partition loops unchanged; #4 (status set cross-check) STILL VALID — post-partition checks unchanged. All three carried-forward findings are low-severity defense-in-depth, so the review action is COMMENT.

<details>

```

In `packages/rs-drive/src/verify/state_transition/verify_state_transition_was_executed_with_proof/v0/mod.rs`:
- [SUGGESTION] packages/rs-drive/src/verify/state_transition/verify_state_transition_was_executed_with_proof/v0/mod.rs:1620-1628: [carried-forward] verify_merged_query_strict drops the prior u16-bound check on queried-key cardinality
  Carried forward from the prior review (ceca33f1 → fcfd1fe5: STILL VALID; the refactor relocated this pattern into the new helper but did not address the gap). The replaced `verify_shielded_nullifiers_v0` set `limit = Some(u16::try_from(nullifiers.len())?)` and surfaced a typed error when the queried key set exceeded `u16` capacity. The new `verify_merged_query_strict` unconditionally clears sub-query limits and sets `merged_pq.query.limit = Some(u16::MAX)` for every shielded arm, then relies entirely on the post-loop "every returned entry is spent" check.

  If a transition ever carried more than `u16::MAX` total queried keys (nullifiers ∪ one singleton address/document/outpoint), the strict verifier would silently truncate the result set; the spent-only loop would then pass on a subset and accept the proof. Upstream DPP structure validation (`max_shielded_transition_actions = 16`) keeps honest inputs far below this today, so there is no live exploit — but the helper's safety comment explicitly leans on the limit being unreachable, while the local invariant that makes it unreachable now lives elsewhere and is not asserted at the verifier boundary. Re-asserting it at the top of `verify_merged_query_strict` (or at each caller) via `u16::try_from(expected_count)?` would make the `u16::MAX` sentinel locally sound rather than transitively sound.
```
</details>

packages/rs-drive/src/verify/state_transition/verify_state_transition_was_executed_with_proof/v0/mod.rs
```diff
@@ -1507,6 +1586,53 @@ impl Drive {
...
+    > {
+        for sub_query in &mut sub_queries {
+            sub_query.query.limit = None;
+        }
+
+        let mut merged_pq = grovedb::PathQuery::merge(
+            sub_queries.iter().collect(),
+            &platform_version.drive.grove_version,
+        )?;
+        merged_pq.query.limit = Some(u16::MAX);
```
thepastaclaw, line 1628: **🟡 Suggestion: [carried-forward] verify_merged_query_strict drops the prior u16-bound check on queried-key cardinality**

Carried forward from the prior review (ceca33f1 → fcfd1fe5: STILL VALID; the refactor relocated this pattern into the new helper but did not address the gap). The replaced `verify_shielded_nullifiers_v0` set `limit = Some(u16::try_from(nullifiers.len())?)` and surfaced a typed error when the queried key set exceeded `u16` capacity. The new `verify_merged_query_strict` unconditionally clears sub-query limits and sets `merged_pq.query.limit = Some(u16::MAX)` for every shielded arm, then relies entirely on the post-loop "every returned entry is spent" check.

If a transition ever carried more than `u16::MAX` total queried keys (nullifiers ∪ one singleton address/document/outpoint), the strict verifier would silently truncate the result set; the spent-only loop would then pass on a subset and accept the proof. Upstream DPP structure validation (`max_shielded_transition_actions = 16`) keeps honest inputs far below this today, so there is no live exploit — but the helper's safety comment explicitly leans on the limit being unreachable, while the local invariant that makes it unreachable now lives elsewhere and is not asserted at the verifier boundary. Re-asserting it at the top of `verify_merged_query_strict` (or at each caller) via `u16::try_from(expected_count)?` would make the `u16::MAX` sentinel locally sound rather than transitively sound.

packages/rs-drive/src/verify/state_transition/verify_state_transition_was_executed_with_proof/v0/mod.rs
```diff
@@ -1154,25 +1227,8 @@ impl Drive {
                     }
                 }
```
thepastaclaw, line 1228: **💬 Nitpick: [carried-forward] Returned-status set is not cross-checked against the signed nullifier set**

Carried forward from the prior review (STILL VALID — unchanged in the latest delta; applies symmetrically to ShieldedWithdrawal at L1360–1367). Both arms iterate `statuses` and reject any unspent entry, but never assert `statuses.len() == nullifier_keys.len()` or that every queried nullifier appears in the returned set. The strict `verify_query_with_absence_proof` is expected to return one present/absent entry per queried key, so a missing nullifier should surface as `(key, false)` and trip the spent-check — but the verifier relies entirely on the strict verifier's completeness guarantee. Collecting `statuses` into a `BTreeMap<Vec<u8>, bool>` and asserting key-set equality with `nullifier_keys` would catch a class of verifier-completeness regressions the spent-only loop would silently absorb. Pure defense-in-depth; no known exploit path.

packages/rs-drive/src/verify/state_transition/verify_state_transition_was_executed_with_proof/v0/mod.rs
shumkov: Carried-forward duplicate of the u16-bound thread declined above — same reasoning. `validate_actions_count` (a `consensus/basic` check) caps `actions.len()` at `max_shielded_transition_actions = 16`, so the merged key set (nullifiers ∪ one singleton address/document/outpoint) is at most 17 against the 65535 limit and cannot truncate; any transition that could exceed it is rejected at basic validation before a proof exists. Re-asserting `u16::try_from(expected_count)?` inside `verify_merged_query_strict` would be a typed error branch that can never fire, so declining. Will revisit if the action cap ever grows by orders of magnitude.

packages/rs-drive/src/verify/state_transition/verify_state_transition_was_executed_with_proof/v0/mod.rs
shumkov: Carried-forward duplicate of the statuses cross-check thread declined above — same reasoning. `verify_query_with_absence_proof` returns one present/absent entry per queried key, and with strict succinctness plus no truncation (≤ 16 nullifiers ≪ the limit), `statuses.len() == nullifier_keys.len()` already holds; any unspent/missing nullifier surfaces as `(key, false)` and is caught by the existing spent-check loop. A key-set equality assertion would add nothing reachable, so declining.

thepastaclaw (commented):
## Code Review

The latest 9-line delta resolves prior finding #2 by adding the duplicate-output-address rejection in the Unshield arm (lines 1212-1220), mirroring ShieldedWithdrawal's singleton-subtree invariant. Two prior defense-in-depth findings remain valid at f98de59d: the strict merged-query helper still uses an unconditional u16::MAX limit without enforcing the queried-key bound, and neither shielded arm explicitly cross-checks the returned status keys against the signed nullifier set. No new defects in the latest delta; the fix is sound and well-aligned with the symmetric ShieldedWithdrawal pattern.

_2 carried-forward finding(s) are still valid at this commit and were already raised as open threads on this PR, so I am not re-posting duplicate inline comments._

### Carried-forward prior findings still valid

- **suggestion** — [carried-forward] verify_merged_query_strict relies on but does not enforce the u16 cardinality bound

- **nitpick** — [carried-forward] Returned nullifier-status set is not cross-checked against the signed nullifier set (Unshield)

### Prior findings fixed in this push

- **FIXED** — Unshield silently overwrites duplicate address entries; ShieldedWithdrawal explicitly rejects them: FIXED at f98de59d. Lines 1212-1220 now explicitly reject a second entry under the clear-address pool with `ProofError::CorruptedProof('unshield proof contained more than one output-address entry')`, mirroring ShieldedWithdrawal's singleton-subtree invariant. Matches the suggested fix from the prior review.

### New findings in latest delta

None.

shumkov (approved):

PR #3817, opened 2026-06-08
feat(platform-wallet): sweep + recover CoinJoin mixed coins for the DashSync→SDK migration

llbartekll:
## What

Adds platform-wallet support for recovering and offloading CoinJoin ("mixed") coins for wallets migrating from DashSync (legacy SPV) to the SwiftDashSDK Core wallet. Post-migration CoinJoin is being dropped, so this lets users find their scattered mixed coins and sweep them back into spendable BIP44 balance.

Four commits:

1. **feat(platform-wallet): CoinJoin sweep + one-time recovery gap-limit** — `sweep_coinjoin_to_address` (empties the CoinJoin account to a destination address) + `set_coinjoin_gap_limit` for a one-time wide recovery scan that finds deep mixed coins scattered across large address holes (DashSync used a gap of 400 vs the SDK default of 30).
2. **feat(coinjoin): chunk the CoinJoin sweep across multiple transactions** — partitions the spendable UTXO set into balanced chunks of ≤500 inputs and returns `Vec<Transaction>`, so a heavy mixer whose UTXOs exceed a single relayable tx still sweeps (broadcast with partial-failure tolerance).
3. **chore(deps): re-pin rust-dashcore to dev (#804 merged)** — see the dependency note below.
4. **fix(coinjoin): sign internal-chain ("change") mixed coins in the sweep** — DashSync CoinJoin uses both external (`…/0/i`) and internal (`…/1/i`, mixing change) chains; the SDK's CoinJoin account is external-only. The sweep now builds a dual-chain `address → DerivationPath` resolver from the CoinJoin account xpub so internal-chain UTXOs sign correctly (otherwise a single internal input blocks the whole all-or-nothing chunk).

Also includes the FFI (`core_wallet/broadcast.rs`) and swift-sdk wrapper (`ManagedCoreWallet.swift`) surface for the above.

## ⚠️ Dependency bump (intentional)

Commit 3 bumps the workspace rust-dashcore pin from `eb889af1` → `7ff6b246` (8 crates). `7ff6b246` is `eb889af1` + 5 commits and includes the **merged** dashpay/rust-dashcore#804, which the internal-chain signing (commit 4) depends on. This is a forward bump, not a revert.

## Verification

- `cargo build -p platform-wallet` — clean (branch rebased onto current `v3.1-dev`).
- `cargo test -p platform-wallet` — dual-chain resolver unit tests pass (`resolves_external_and_internal_chain_addresses`, `errors_when_an_input_address_is_unresolvable`).
- `DashSDKFFI.xcframework` rebuilt; the SwiftExampleApp and the downstream dashwallet-ios app both build against it (iOS simulator).
- Runtime: confirmed on a ~2k-address testnet CoinJoin wallet — a 278-input sweep spanning both chains confirmed on-chain.

## Merge order

The dashwallet-ios app branch links the SwiftDashSDK built from this branch, so **this PR should merge into `v3.1-dev` before** the app-side PR.

## Summary by CodeRabbit

* **New Features**
  * Added CoinJoin sweep APIs to consolidate CoinJoin UTXOs into a destination address, returning the resulting transaction IDs (including multi-transaction sweeps for larger balances).
  * Added ability to configure the CoinJoin address pool gap limit for recovery scenarios via the Swift SDK.
* **Bug Fixes**
  * CoinJoin sweeps now broadcast sweep transactions in chunks and succeed as long as at least one transaction is successfully broadcast (partial failures are tolerated).


thepastaclaw:
✅ Review complete (commit 8760508e)

packages/rs-platform-wallet/src/wallet/core/coinjoin_recovery.rs
```diff
@@ -0,0 +1,108 @@
...
+                .next()
+                .ok_or_else(|| {
+                    PlatformWalletError::AddressOperation(
+                        "CoinJoin account has no address pool".to_string(),
+                    )
+                })?;
+
+            pool.gap_limit = gap_limit.min(MAX_GAP_LIMIT);
+            pool.maintain_gap_limit(&key_source)
+                .map_err(|e| PlatformWalletError::AddressOperation(e.to_string()))?;
```
thepastaclaw: **🔴 Blocking: Reject zero CoinJoin gap limit before maintaining the pool**

`set_coinjoin_gap_limit` takes `gap_limit: u32` from the Swift FFI and writes it directly into `pool.gap_limit` (only clamped on the upper end via `MAX_GAP_LIMIT`). If `gap_limit == 0`, `AddressPool::maintain_gap_limit` evaluates `self.gap_limit - 1` when `highest_used` is `None` (verified in the pinned key-wallet at `address_pool.rs:888-891`). That panics in debug builds and wraps to `u32::MAX` in release, causing the recovery helper to try generating ~4 billion addresses while holding the wallet write lock. Since this is a brand-new public API (FFI does not validate either), reject the zero case at the boundary before touching the pool.

```suggestion
            let gap_limit = gap_limit.min(MAX_GAP_LIMIT);
            if gap_limit == 0 {
                return Err(PlatformWalletError::AddressOperation(
                    "CoinJoin gap limit must be greater than zero".to_string(),
                ));
            }
            pool.gap_limit = gap_limit;
            pool.maintain_gap_limit(&key_source)
                .map_err(|e| PlatformWalletError::AddressOperation(e.to_string()))?;
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
```diff
@@ -145,4 +168,428 @@ impl<B: TransactionBroadcaster + ?Sized> CoreWallet<B> {
...
+            let batch = pool
+                .generate_addresses(BATCH, key_source, true)
+                .map_err(|e| PlatformWalletError::AddressOperation(e.to_string()))?;
+            for addr in &batch {
+                if needed.remove(addr) {
+                    if let Some(info) = pool.address_info(addr) {
+                        path_map.insert(addr.clone(), info.path.clone());
+                    }
+                }
+            }
```
thepastaclaw: **🟡 Suggestion: Remove address from `needed` only after `path_map.insert` actually succeeds**

`coinjoin_sweep_path_map` does `if needed.remove(addr) { if let Some(info) = pool.address_info(addr) { path_map.insert(...) } }`. The `remove` happens unconditionally on a match, but the `path_map.insert` is gated on `address_info` returning `Some`. The resolver's documented contract is "errors if any input can't be resolved, rather than failing mid-sign" — but if `address_info` ever returns `None` for an address `generate_addresses` just produced (an invariant violation in `AddressPool`), the address is silently dropped from `needed`, `needed.is_empty()` becomes true so the defensive check at line 462 does not fire, and `signer.sign_tx` later fails with the very "no derivation path for input address" error the resolver was added to prevent. Today `generate_address_at_index` inserts into both `addresses` and `address_index` atomically, so this can't trigger — but inverting the gating preserves the "removed from needed ⇒ inserted into path_map" invariant by construction, so a future address-pool refactor cannot silently break this code path.

```suggestion
            for addr in &batch {
                if needed.contains(addr) {
                    if let Some(info) = pool.address_info(addr) {
                        path_map.insert(addr.clone(), info.path.clone());
                        needed.remove(addr);
                    }
                }
            }
```

<sub>source: ['claude']</sub>

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
```diff
@@ -145,4 +168,428 @@ impl<B: TransactionBroadcaster + ?Sized> CoreWallet<B> {
...
+        if broadcast.is_empty() {
+            return Err(last_err.unwrap_or_else(|| {
+                PlatformWalletError::TransactionBuild(
+                    "CoinJoin sweep produced no broadcastable transactions".to_string(),
+                )
+            }));
+        }
+
+        Ok(broadcast)
+    }
```
thepastaclaw, line 255: **🟡 Suggestion: Partial-broadcast failures are swallowed silently when at least one chunk succeeds**

When some chunks broadcast and others fail, the function returns `Ok(broadcast)` with only the successful subset; `last_err` and the count of failures are dropped. The docstring (lines 194–198) calls this out as intentional — the caller is expected to refresh balance and re-run — but the result type currently carries no signal that the sweep was partial. The Swift caller's `sweepCoinJoinAccount` therefore can't surface "swept K of N chunks" telemetry or warn the user before claiming success. At minimum, log dropped errors via `tracing::warn!` inside the loop so the failures are observable in logs; longer term, consider returning a richer outcome type so the FFI consumer can distinguish full vs partial success.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
```diff
@@ -145,4 +168,428 @@ impl<B: TransactionBroadcaster + ?Sized> CoreWallet<B> {
...
+                let signed = signer
+                    .sign_tx(unsigned, chunk_utxos, |addr| path_map.get(&addr).cloned())
+                    .await
+                    .map_err(|e| PlatformWalletError::TransactionBuild(e.to_string()))?;
+
+                debug_assert_eq!(
+                    signed.input.len(),
+                    input_count,
+                    "CoinJoin sweep chunk must consume every UTXO in the chunk"
+                );
```
thepastaclaw: **💬 Nitpick: Promote `debug_assert_eq!` on the signer's input-count invariant to a runtime error**

`debug_assert_eq!(signed.input.len(), input_count, ...)` is compiled out in release. The invariant being checked — that the signer consumed every UTXO in the chunk — is what guarantees `total_input - fee` is a valid single-output amount with no residual UTXOs. If a future signer implementation drops an unsigned input, the release build will silently broadcast a sweep that fails to consume every UTXO in the chunk. Since the check is O(1), convert it to a returned `PlatformWalletError::TransactionBuild` so a broken invariant aborts the chunk instead of broadcasting an incomplete sweep.

```suggestion
                if signed.input.len() != input_count {
                    return Err(PlatformWalletError::TransactionBuild(format!(
                        "CoinJoin sweep chunk must consume every UTXO in the chunk: signed {} inputs, expected {}",
                        signed.input.len(),
                        input_count
                    )));
                }
                signed_txs.push(signed);
```

<sub>source: ['claude']</sub>

thepastaclaw (changes requested):
## Code Review

PR adds a CoinJoin recovery + chunked sweep flow that is well-scoped and faithfully mirrors key-wallet's fee/sizing conventions. One blocking issue: the new `set_coinjoin_gap_limit` public API does not reject `gap_limit == 0`, which underflows `self.gap_limit - 1` inside `maintain_gap_limit` (debug panic / release wrap to u32::MAX → effectively unbounded address generation under the wallet write lock). A few non-blocking defensive-coding improvements in the sweep path as well.

<details>

```

In `packages/rs-platform-wallet/src/wallet/core/coinjoin_recovery.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/wallet/core/coinjoin_recovery.rs:96-98: Reject zero CoinJoin gap limit before maintaining the pool
  `set_coinjoin_gap_limit` takes `gap_limit: u32` from the Swift FFI and writes it directly into `pool.gap_limit` (only clamped on the upper end via `MAX_GAP_LIMIT`). If `gap_limit == 0`, `AddressPool::maintain_gap_limit` evaluates `self.gap_limit - 1` when `highest_used` is `None` (verified in the pinned key-wallet at `address_pool.rs:888-891`). That panics in debug builds and wraps to `u32::MAX` in release, causing the recovery helper to try generating ~4 billion addresses while holding the wallet write lock. Since this is a brand-new public API (FFI does not validate either), reject the zero case at the boundary before touching the pool.

In `packages/rs-platform-wallet/src/wallet/core/broadcast.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/core/broadcast.rs:451-457: Remove address from `needed` only after `path_map.insert` actually succeeds
  `coinjoin_sweep_path_map` does `if needed.remove(addr) { if let Some(info) = pool.address_info(addr) { path_map.insert(...) } }`. The `remove` happens unconditionally on a match, but the `path_map.insert` is gated on `address_info` returning `Some`. The resolver's documented contract is "errors if any input can't be resolved, rather than failing mid-sign" — but if `address_info` ever returns `None` for an address `generate_addresses` just produced (an invariant violation in `AddressPool`), the address is silently dropped from `needed`, `needed.is_empty()` becomes true so the defensive check at line 462 does not fire, and `signer.sign_tx` later fails with the very "no derivation path for input address" error the resolver was added to prevent. Today `generate_address_at_index` inserts into both `addresses` and `address_index` atomically, so this can't trigger — but inverting the gating preserves the "removed from needed ⇒ inserted into path_map" invariant by construction, so a future address-pool refactor cannot silently break this code path.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/core/broadcast.rs:370-388: Partial-broadcast failures are swallowed silently when at least one chunk succeeds
  When some chunks broadcast and others fail, the function returns `Ok(broadcast)` with only the successful subset; `last_err` and the count of failures are dropped. The docstring (lines 194–198) calls this out as intentional — the caller is expected to refresh balance and re-run — but the result type currently carries no signal that the sweep was partial. The Swift caller's `sweepCoinJoinAccount` therefore can't surface "swept K of N chunks" telemetry or warn the user before claiming success. At minimum, log dropped errors via `tracing::warn!` inside the loop so the failures are observable in logs; longer term, consider returning a richer outcome type so the FFI consumer can distinguish full vs partial success.
```
</details>

packages/rs-platform-wallet/src/wallet/core/coinjoin_recovery.rs
llbartekll: Fixed in c65905f8a1 — set_coinjoin_gap_limit now rejects gap_limit == 0 (returns an AddressOperation error) before touching the pool, so the gap_limit - 1 underflow in maintain_gap_limit cannot be reached. Verified the underflow against the pinned key-wallet at address_pool.rs:894.

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
llbartekll: Done in c65905f8a1 — inverted the gating so an address is removed from needed only after its path is recorded, preserving the removed-implies-resolved invariant by construction. The unresolved-address check now fires even if a future AddressPool refactor desyncs generate_addresses and address_info.

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
llbartekll: Added tracing::warn! on each dropped chunk error in c65905f8a1, so partial sweeps are observable in logs. Deferring the richer outcome type to a follow-up: it would cascade through the FFI, swift-sdk, and app, and the current contract is intentional and documented — the Swift caller already records the returned txids and re-runs against the refreshed balance to sweep any remainder. Can file an issue for a structured partial-success type if you prefer.

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
llbartekll: Done in c65905f8a1 — promoted the debug_assert_eq! to a runtime TransactionBuild error so the consumed-every-UTXO fund-safety invariant is enforced in release builds too.

packages/rs-platform-wallet-ffi/src/core_wallet/broadcast.rs
```diff
@@ -127,6 +127,116 @@ pub unsafe extern "C" fn core_wallet_send_to_addresses(
...
+        // dropped before the function returns.
+        let signer = unsafe {
+            MnemonicResolverCoreSigner::new(
+                signer_addr as *mut MnemonicResolverHandle,
+                wallet_id,
+                network,
+            )
+        };
+        runtime().block_on(wallet.sweep_coinjoin_to_address(account_index, dest, &signer))
+    });
```
thepastaclaw, line 198: **🔴 Blocking: Validate the sweep destination against the wallet's network**

`core_wallet_sweep_coinjoin` parses `dest_address` with `dashcore::Address::from_str(...).assume_checked()` and then passes the resulting `Address` straight into `sweep_coinjoin_to_address`, which builds and broadcasts a transaction that drains every spendable CoinJoin UTXO to that script. The wallet's network is read a few lines below (line 175) but is never used to verify the destination. A mainnet `CoreWallet` will therefore accept a testnet/devnet/regtest address (or vice versa), and the resulting all-input → single-output sweep will be broadcast on the wallet's network with the wrong-network script — the entire CoinJoin account becomes unrecoverable. This is especially severe because this entry point is the all-funds sweep used by the DashSync migration, and the address comes from arbitrary Swift caller input. Parse as `Address<NetworkUnchecked>` and `require_network(wallet.network())` inside the storage closure so the failure surfaces as a `PlatformWalletFFIResult` error before any tx is built.

```suggestion
    let dest_str = unwrap_result_or_return!(std::ffi::CStr::from_ptr(dest_address).to_str());
    let dest_unchecked = unwrap_result_or_return!(dashcore::Address::from_str(dest_str));

    let signer_addr = core_signer_handle as usize;

    let option = CORE_WALLET_STORAGE.with_item(handle, |wallet| {
        let wallet_id = wallet.wallet_id();
        let network = wallet.network();
        let dest = dest_unchecked
            .clone()
            .require_network(network)
            .map_err(|e| platform_wallet::PlatformWalletError::AddressOperation(e.to_string()))?;
        // SAFETY: the resolver handle is pinned alive for the duration of
        // this FFI call (see fn-level safety doc). The
        // `MnemonicResolverCoreSigner` lives on this stack frame and is
        // dropped before the function returns.
        let signer = unsafe {
            MnemonicResolverCoreSigner::new(
                signer_addr as *mut MnemonicResolverHandle,
                wallet_id,
                network,
            )
        };
        runtime().block_on(wallet.sweep_coinjoin_to_address(account_index, dest, &signer))
    });
```

<sub>source: ['codex', 'claude']</sub>

thepastaclaw (changes requested):
## Code Review

All four prior findings are fixed at head c65905f8. The latest commit cleanly applies the suggested fixes (zero gap-limit guard, insert-then-remove ordering in path_map, tracing::warn on per-chunk broadcast failures, runtime input-count invariant). One new in-scope blocking issue surfaces in the cumulative review: the new sweep FFI entry point `core_wallet_sweep_coinjoin` parses the caller-supplied destination with `assume_checked()` and never matches it against the wallet's network, so a wrong-network address would silently drain the entire CoinJoin account to a script on the wrong chain.

🔴 1 blocking

<details>

```

- [BLOCKING] packages/rs-platform-wallet-ffi/src/core_wallet/broadcast.rs:168-188: Validate the sweep destination against the wallet's network
  `core_wallet_sweep_coinjoin` parses `dest_address` with `dashcore::Address::from_str(...).assume_checked()` and then passes the resulting `Address` straight into `sweep_coinjoin_to_address`, which builds and broadcasts a transaction that drains every spendable CoinJoin UTXO to that script. The wallet's network is read a few lines below (line 175) but is never used to verify the destination. A mainnet `CoreWallet` will therefore accept a testnet/devnet/regtest address (or vice versa), and the resulting all-input → single-output sweep will be broadcast on the wallet's network with the wrong-network script — the entire CoinJoin account becomes unrecoverable. This is especially severe because this entry point is the all-funds sweep used by the DashSync migration, and the address comes from arbitrary Swift caller input. Parse as `Address<NetworkUnchecked>` and `require_network(wallet.network())` inside the storage closure so the failure surfaces as a `PlatformWalletFFIResult` error before any tx is built.
```
</details>

packages/rs-platform-wallet-ffi/src/core_wallet/broadcast.rs
llbartekll: Fixed in f52644b7e8 — core_wallet_sweep_coinjoin now parses the destination as Address<NetworkUnchecked> and calls require_network(wallet.network()) inside the storage closure, so a wrong-network address errors before any tx is built. Mirrors the withdrawal FFI (require_network at identity_withdrawal.rs:48). Reachability note: the only app caller always passes the wallet own current-network receive address, but this hardens the public fund-moving FFI boundary regardless.

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
```diff
@@ -145,4 +168,452 @@ impl<B: TransactionBroadcaster + ?Sized> CoreWallet<B> {
...
+                let input_count = chunk_utxos.len();
+                let total_input: u64 = chunk_utxos.iter().map(|u| u.value()).sum();
+
+                // Exact fee for (input_count inputs, 1 output, no change).
+                // Mirrors key-wallet's `calculate_base_size()` (8 + input-
+                // varint + output-varint + 34) and the selector's 148 B/input,
+                // so `total_input - fee` yields a single output with zero
+                // change for this chunk.
+                let fee = fee_rate
+                    .calculate_fee(BASE_SIZE_1_OUTPUT_NO_CHANGE + input_count * INPUT_SIZE);
```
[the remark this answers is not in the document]
llbartekll: Fixed in f52644b7e8 — ran cargo fmt on the touched files; rustfmt --check is clean now.

packages/rs-platform-wallet/src/wallet/core/coinjoin_recovery.rs
```diff
@@ -0,0 +1,118 @@
...
+        &self,
+        account_index: u32,
+        gap_limit: u32,
+    ) -> Result<u32, PlatformWalletError> {
+        let mut wm = self.wallet_manager.write().await;
+        let (wallet, info) = wm.get_wallet_and_info_mut(&self.wallet_id).ok_or_else(|| {
+            PlatformWalletError::WalletNotFound(
+                "Wallet not found in wallet manager".to_string(),
+            )
+        })?;
```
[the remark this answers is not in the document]
llbartekll: Fixed in f52644b7e8 — ran cargo fmt; rustfmt --check is clean now.

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
```diff
@@ -145,4 +168,452 @@ impl<B: TransactionBroadcaster + ?Sized> CoreWallet<B> {
...
+                // address stays in `needed` and the check below errors loudly
+                // rather than silently dropping it into a mid-sign failure.
+                if needed.contains(addr) {
+                    if let Some(info) = pool.address_info(addr) {
+                        path_map.insert(addr.clone(), info.path.clone());
+                        needed.remove(addr);
+                    }
+                }
+            }
+            generated += BATCH;
```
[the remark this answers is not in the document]
llbartekll: Declining — this is a false positive for the actual constant. COINJOIN_SWEEP_MAX_INDEX is 20000, an exact multiple of BATCH (500): 20000 / 500 = 40, so the last entering iteration has generated = 19500, produces indices 19500..19999, then generated becomes 20000 and the guard stops. Exactly 20000 indices (0..19999) are searched — no overshoot — and the "within 20000 indices" message is accurate. The only non-multiple caller is a unit test that asserts only is_err() and never inspects the index count. Can add the clamp purely as future-proofing if you prefer, but there is no current inaccuracy to fix.

llbartekll:
Re the CodeRabbit nitpick about releasing the wallet write lock across the sign `.await` loop in the CoinJoin sweep (no inline thread for it): declining for now.

- It is a `tokio::sync::RwLock` — designed to be held across `.await`, so this is not a deadlock/correctness issue, only a throughput observation (already documented in the code comment there).
- The broadcast loop already runs **outside** the lock, and the locked section performs **no** wallet mutation (it snapshots UTXOs into an owned `Vec` and builds an owned `path_map`), so dropping the lock early buys no snapshot→broadcast atomicity.
- The only thing it would save is local, fast signing time during a rare one-time single-user migration sweep, at the cost of re-touching the deliberate `wallet`/`info` borrow ordering.

Tracking it as a possible future optimization if this path is ever reused in a concurrent/server context.

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
```diff
@@ -145,4 +168,455 @@ impl<B: TransactionBroadcaster + ?Sized> CoreWallet<B> {
...
+                let input_count = chunk_utxos.len();
+                let total_input: u64 = chunk_utxos.iter().map(|u| u.value()).sum();
+
+                // Exact fee for (input_count inputs, 1 output, no change).
+                // Mirrors key-wallet's `calculate_base_size()` (8 + input-
+                // varint + output-varint + 34) and the selector's 148 B/input,
+                // so `total_input - fee` yields a single output with zero
+                // change for this chunk.
+                let fee =
+                    fee_rate.calculate_fee(BASE_SIZE_1_OUTPUT_NO_CHANGE + input_count * INPUT_SIZE);
```
thepastaclaw: **🟡 Suggestion: Sweep fee constants are not a true upper bound for large chunks**

`BASE_SIZE_1_OUTPUT_NO_CHANGE = 8 + 1 + 1 + 34` hard-codes the input-count CompactSize as 1 byte and `INPUT_SIZE = 148` is the typical, not maximum, P2PKH input size. Chunks are explicitly allowed up to MAX_INPUTS_PER_SWEEP, and when `input_count >= 253` the CompactSize input-count grows to 3 bytes; compressed P2PKH ECDSA inputs can serialize to 149 bytes at the larger DER signature size. Because the sweep subtracts the fee *before* signing and never recomputes from the signed transaction, the computed fee can land below 1 duff/byte on large chunks — exactly the heavy-CoinJoin case this PR targets. Use upper-bound constants (or compute the varint length from `input_count`) so the fee is always ≥ policy minimum and the all-funds invariant holds.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
```diff
@@ -145,4 +168,455 @@ impl<B: TransactionBroadcaster + ?Sized> CoreWallet<B> {
...
+        account_index: u32,
+        dest: DashAddress,
+        signer: &S,
+    ) -> Result<Vec<Transaction>, PlatformWalletError> {
+        use dashcore::blockdata::witness::Witness;
+        use dashcore::{ScriptBuf, TxIn, TxOut};
+        use key_wallet::wallet::managed_wallet_info::fee::FeeRate;
+        use key_wallet::wallet::managed_wallet_info::transaction_builder::TransactionSigner;
+        use key_wallet::wallet::managed_wallet_info::wallet_info_interface::WalletInfoInterface;
+
```
thepastaclaw: **🟡 Suggestion: Enforce destination network on the Rust sweep API itself**

The wrong-network destination check now lives in the FFI wrapper, but `CoreWallet::sweep_coinjoin_to_address` accepts any `NetworkChecked` `dashcore::Address` and proceeds to drain every spendable CoinJoin UTXO to it. Any non-FFI Rust caller (other consumers of `rs-platform-wallet`, future tests, future wrappers) can pass a mainnet-checked address to a testnet wallet and the sweep will succeed with funds locked to an unspendable script on this chain. Move the `dest.network() == self.network()` check into the sweep method itself so the fund-safety invariant is owned by the API that performs the sweep, not by a single caller.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/core/coinjoin_recovery.rs
```diff
@@ -0,0 +1,116 @@
...
+            if gap_limit == 0 {
+                return Err(PlatformWalletError::AddressOperation(
+                    "CoinJoin gap limit must be greater than zero".to_string(),
+                ));
+            }
+            pool.gap_limit = gap_limit;
+            pool.maintain_gap_limit(&key_source)
+                .map_err(|e| PlatformWalletError::AddressOperation(e.to_string()))?;
+            pool.highest_generated.unwrap_or(0)
+        };
```
thepastaclaw: **🟡 Suggestion: Recovery widens only the external CoinJoin pool, not the internal change branch**

`set_coinjoin_gap_limit` widens the first pool returned by `address_pools_mut()`. The same PR adds dual-chain signing in `coinjoin_sweep_path_map` because DashSync puts mixed change on the internal `/1/i` branch, yet that branch is not materialized by this recovery API — SPV will not watch scripts that have never been pushed into a pool. If migrated wallets are expected to discover new mixed coins on the internal chain via this scan, the recovery is incomplete; if the assumption is that internal-chain coins are only ever imported via an out-of-band UTXO import, that contract is worth documenting at the top of this module so future callers do not silently miss funds.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
```diff
@@ -145,4 +168,455 @@ impl<B: TransactionBroadcaster + ?Sized> CoreWallet<B> {
...
+            }
+        }
+
+        if broadcast.is_empty() {
+            return Err(last_err.unwrap_or_else(|| {
+                PlatformWalletError::TransactionBuild(
+                    "CoinJoin sweep produced no broadcastable transactions".to_string(),
+                )
+            }));
+        }
```
thepastaclaw, line 252: **💬 Nitpick: Only the last broadcast error survives when every chunk fails**

In the partial-failure broadcast loop, `last_err = Some(e)` overwrites every prior chunk error. Each failure is logged at `warn`, but if every chunk fails for different reasons (e.g., one mempool policy reject, others transient network errors), the returned `Err` reflects only the final cause. Consider aggregating distinct error strings or surfacing the failure count alongside the last error so downstream telemetry/tests are not misled.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Latest delta (f52644b7) is the focused FFI network-validation fix plus rustfmt. The prior blocking finding is FIXED at packages/rs-platform-wallet-ffi/src/core_wallet/broadcast.rs:168-185 — `dest_unchecked.require_network(network)` runs inside the storage closure before any tx is built. Cumulative review surfaces no remaining in-scope blockers. Two suggestion-level findings about the sweep path are worth surfacing: (1) the sweep fee constants are not an upper bound for ≥253-input chunks, and (2) the public Rust `sweep_coinjoin_to_address` still trusts its `dest` argument's network. One nitpick on broadcast loop error preservation. The codex-general 'materialize internal CoinJoin chain' concern is downgraded to a suggestion because the SDK's CoinJoin account explicitly models a single (external) pool and the recovery scan is documented as a one-time post-migration helper.

<details>

```

In `packages/rs-platform-wallet/src/wallet/core/broadcast.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/core/broadcast.rs:289-310: Sweep fee constants are not a true upper bound for large chunks
  `BASE_SIZE_1_OUTPUT_NO_CHANGE = 8 + 1 + 1 + 34` hard-codes the input-count CompactSize as 1 byte and `INPUT_SIZE = 148` is the typical, not maximum, P2PKH input size. Chunks are explicitly allowed up to MAX_INPUTS_PER_SWEEP, and when `input_count >= 253` the CompactSize input-count grows to 3 bytes; compressed P2PKH ECDSA inputs can serialize to 149 bytes at the larger DER signature size. Because the sweep subtracts the fee *before* signing and never recomputes from the signed transaction, the computed fee can land below 1 duff/byte on large chunks — exactly the heavy-CoinJoin case this PR targets. Use upper-bound constants (or compute the varint length from `input_count`) so the fee is always ≥ policy minimum and the all-funds invariant holds.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/core/broadcast.rs:198-210: Enforce destination network on the Rust sweep API itself
  The wrong-network destination check now lives in the FFI wrapper, but `CoreWallet::sweep_coinjoin_to_address` accepts any `NetworkChecked` `dashcore::Address` and proceeds to drain every spendable CoinJoin UTXO to it. Any non-FFI Rust caller (other consumers of `rs-platform-wallet`, future tests, future wrappers) can pass a mainnet-checked address to a testnet wallet and the sweep will succeed with funds locked to an unspendable script on this chain. Move the `dest.network() == self.network()` check into the sweep method itself so the fund-safety invariant is owned by the API that performs the sweep, not by a single caller.

In `packages/rs-platform-wallet/src/wallet/core/coinjoin_recovery.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/core/coinjoin_recovery.rs:79-108: Recovery widens only the external CoinJoin pool, not the internal change branch
  `set_coinjoin_gap_limit` widens the first pool returned by `address_pools_mut()`. The same PR adds dual-chain signing in `coinjoin_sweep_path_map` because DashSync puts mixed change on the internal `/1/i` branch, yet that branch is not materialized by this recovery API — SPV will not watch scripts that have never been pushed into a pool. If migrated wallets are expected to discover new mixed coins on the internal chain via this scan, the recovery is incomplete; if the assumption is that internal-chain coins are only ever imported via an out-of-band UTXO import, that contract is worth documenting at the top of this module so future callers do not silently miss funds.
```
</details>

packages/rs-platform-wallet/src/wallet/core/coinjoin_recovery.rs
thepastaclaw: **🟡 Suggestion: Recovery widens only the external CoinJoin pool while sweep expects both chains**

`set_coinjoin_gap_limit` calls `address_pools_mut().into_iter().next()` and widens that single pool. Yet `coinjoin_sweep_path_map` (broadcast.rs:271-287) explicitly resolves signing paths across BOTH external `/0/` and internal `/1/` chains, with a comment that DashSync CoinJoin puts mixing change on the internal chain. SPV only watches scripts that have been materialized into a pool's script index, so a recovery scan that widens only the external pool will not discover deep internal-chain mixed coins; the dual-chain signing fix only helps with internal-chain UTXOs that arrive through some other import path. Migrated users running `setCoinJoinGapLimit` (Swift wrapper presents this as the migration recovery scan) can silently miss change-chain funds. Either materialize the internal CoinJoin branch alongside the external pool, or rename/document the API to make the external-only scope explicit and require an alternate import path for internal-chain coins.

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
```diff
@@ -145,4 +168,457 @@ impl<B: TransactionBroadcaster + ?Sized> CoreWallet<B> {
...
```
thepastaclaw, line 252: **💬 Nitpick: All-fail broadcast returns only the last chunk's error**

`last_err = Some(e)` overwrites every prior chunk's error in the loop. Each failure is logged via `tracing::warn!`, but when every chunk fails for distinct reasons (e.g. one fee-too-low, one missing input, one transient relay error) the `Err` returned at line 397 reflects only the final chunk. That makes all-fail diagnosis brittle in logs-only environments. Preserve the first error with `last_err.get_or_insert(e)` (one-line change, log already covers the rest) or aggregate the errors with a count — both keep tolerate-partial-failure semantics intact while surfacing the more useful root failure to programmatic callers.

thepastaclaw (commented):
## Code Review

Latest delta a0ead209 is a clippy-only cleanup in broadcast.rs (drop a redundant clone on a Copy field, range-contains in a test) — no new latest-delta findings. Re-validation of the four prior findings against current head shows all four are STILL_VALID and carried forward: sweep fee constants under-bound large chunks, the Rust sweep API still defers the destination-network invariant to the FFI wrapper, recovery widens only the external CoinJoin pool while internal-chain mixed change is unmaterialized for SPV, and the all-fail broadcast path returns only the final chunk's error.

<details>

```

In `packages/rs-platform-wallet/src/wallet/core/coinjoin_recovery.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/core/coinjoin_recovery.rs:79-108: Recovery widens only the external CoinJoin pool while sweep expects both chains
  `set_coinjoin_gap_limit` calls `address_pools_mut().into_iter().next()` and widens that single pool. Yet `coinjoin_sweep_path_map` (broadcast.rs:271-287) explicitly resolves signing paths across BOTH external `/0/` and internal `/1/` chains, with a comment that DashSync CoinJoin puts mixing change on the internal chain. SPV only watches scripts that have been materialized into a pool's script index, so a recovery scan that widens only the external pool will not discover deep internal-chain mixed coins; the dual-chain signing fix only helps with internal-chain UTXOs that arrive through some other import path. Migrated users running `setCoinJoinGapLimit` (Swift wrapper presents this as the migration recovery scan) can silently miss change-chain funds. Either materialize the internal CoinJoin branch alongside the external pool, or rename/document the API to make the external-only scope explicit and require an alternate import path for internal-chain coins.
```
</details>

packages/rs-platform-wallet/src/wallet/core/coinjoin_recovery.rs
thepastaclaw: Resolved in this update — _Recovery widens only the external CoinJoin pool, not the internal change branch_ no longer present.

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
thepastaclaw: Resolved in `a0ead20` — _Only the last broadcast error survives when every chunk fails_ no longer present.

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
llbartekll: Fixed in 43ee9af — the sweep fee is now a true upper bound. `FeeRate::normal()` is exactly 1 duff/byte (the relay minimum, no headroom), so the size estimate must not undershoot the serialized tx or the chunk pays below the minimum and is rejected. The input-count CompactSize is now sized per chunk (1 B below 253 inputs, 3 B for `253..=MAX_INPUTS_PER_SWEEP` = 500), and per-input size uses the 149 B maximum compressed-P2PKH input (73-byte low-S DER signature) instead of the 148 B typical. Over-estimating is safe for an all-funds sweep — the single output absorbs the few-duff difference. Confirmed reachable: a heavy mixer with a 253–500-input chunk would otherwise pay below the relay minimum and have the chunk rejected.

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
llbartekll: Done in 43ee9af — `sweep_coinjoin_to_address` now rejects a destination that isn't valid for the wallet's network (`dest.as_unchecked().is_valid_for_network(network)`) before snapshotting UTXOs or building any tx, so the fund-safety invariant is owned by the sweep method itself, not only the FFI boundary. Any non-FFI Rust caller (tests, future wrappers) is covered by construction. It uses `is_valid_for_network` to mirror the FFI's `require_network` semantics (testnet/devnet/regtest equivalence).

packages/rs-platform-wallet/src/wallet/core/coinjoin_recovery.rs
llbartekll: Addressed in 43ee9af by documenting the external-only scope in the module doc of `coinjoin_recovery.rs`, rather than materializing the internal pool — external-only is correct for the discovery model. A migrated wallet's internal-chain ("change") mixed coins arrive as **imported spendable UTXOs**, not via an SPV address scan, so they need no gap widening to be discovered. The dual-chain `coinjoin_sweep_path_map` still signs them, re-deriving both `/0/` and `/1/` from the account xpub. The new doc note flags that if a future migration path instead relied on SPV to *re-discover* internal-chain coins, this recovery would also need to materialize the internal pool. Validated end-to-end against a funded CoinJoin wallet.

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
llbartekll: Done in 43ee9af — changed `last_err = Some(e)` to `last_err.get_or_insert(e)`, so when every chunk fails the returned error is the FIRST (usually root-cause) failure rather than the last. Each chunk error is still logged via `tracing::warn!`, so all causes remain visible; the tolerate-partial-failure semantics are unchanged.

packages/rs-platform-wallet/src/wallet/core/coinjoin_recovery.rs
llbartekll: Superseded by the re-raised thread on this same finding — addressed via the external-only scope doc in 43ee9af. Resolving this duplicate.

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
llbartekll: Superseded by the re-raised thread on this same finding — fixed in 43ee9af via `last_err.get_or_insert(e)`. Resolving this duplicate.

packages/rs-platform-wallet-ffi/src/core_wallet/broadcast.rs
```diff
@@ -127,6 +127,126 @@ pub unsafe extern "C" fn core_wallet_send_to_addresses(
...
+    let option = CORE_WALLET_STORAGE.with_item(handle, |wallet| {
+        runtime().block_on(wallet.set_coinjoin_gap_limit(account_index, gap_limit))
+    });
+    let result = unwrap_option_or_return!(option);
+    let highest = unwrap_result_or_return!(result);
+    *out_highest_index = highest;
+    PlatformWalletFFIResult::ok()
+}
+
 /// Free transaction bytes returned by `core_wallet_send_to_addresses`.
```
thepastaclaw: **💬 Nitpick: Update core_wallet_free_tx_bytes doc — it now also frees the sweep txid buffer**

This PR adds `core_wallet_sweep_coinjoin`, which emits a contiguous `count * 32` byte txid buffer freed by `core_wallet_free_tx_bytes` (passing `count * 32` as `len`). The free function's doc-comment still says only "Free transaction bytes returned by `core_wallet_send_to_addresses`", which makes the sweep free-contract harder for binding authors and downstream consumers to discover from the C header alone. Pure docs — no behavioral defect.

```suggestion
/// Free a heap byte buffer returned across the FFI — either the serialized
/// transaction bytes from `core_wallet_send_to_addresses`, or the contiguous
/// `count * 32` wire-order txid buffer from `core_wallet_sweep_coinjoin` (in
/// that case pass `count * 32` as `len`).
```

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Carried-forward prior findings: all four prior findings are resolved at 43ee9af — fee bound is now a true upper bound (149 B/input + dynamic varint), the wrong-network sweep check is enforced inside the Rust API (not just FFI), the broadcast loop preserves the FIRST chunk error via get_or_insert, and the external-only recovery widening is documented as correct given that internal-chain mixed coins are imported as UTXOs (the sweep path_map still re-derives both chains for signing). New findings in the latest delta: codex-flagged null/network issues in core_wallet_send_to_addresses are about pre-existing code that this PR does not modify and are dropped as out of scope. Only a single in-scope doc nitpick remains on core_wallet_free_tx_bytes whose contract now covers both serialized-tx and txid buffers.

💬 1 nitpick(s)

packages/rs-platform-wallet-ffi/src/core_wallet/broadcast.rs
llbartekll: Done in 41d5e8a — `core_wallet_free_tx_bytes`'s doc now documents both free-contracts: the serialized tx bytes from `core_wallet_send_to_addresses` (pass `out_tx_len` as `len`) and the contiguous `count * 32` txid buffer from `core_wallet_sweep_coinjoin` (pass `out_count * 32`). Pure docs, no behavioral change.

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
```diff
@@ -145,4 +168,487 @@ impl<B: TransactionBroadcaster + ?Sized> CoreWallet<B> {
...
+                    + input_count * MAX_P2PKH_INPUT_SIZE;
+                let fee = fee_rate.calculate_fee(tx_size);
+
+                if total_input <= fee {
+                    return Err(PlatformWalletError::TransactionBuild(format!(
+                        "CoinJoin sweep chunk balance {} is below the chunk fee {}",
+                        total_input, fee
+                    )));
+                }
+                let output_amount = total_input - fee;
```
thepastaclaw: **🟡 Suggestion: Sweep chunk can build a sub-dust output and silently fall back to partial success**

The hand-built sweep tx bypasses `TransactionBuilder` and only guards `total_input <= fee`. If a chunk's `total_input - fee` lands above zero but below the destination script's dust threshold (≈546 duffs for P2PKH), the chunk produces a TxOut that relay policy will reject. Because broadcast loops over chunks and tolerates partial failure (line 408-432), the API can still return `Ok(broadcast)` while a dust chunk is silently dropped and those UTXOs remain unswept — exactly the failure mode the all-input chunking is designed to prevent. CoinJoin denominations are normally well above the bound, but DashSync-imported sub-denomination change can ride in on the internal chain and land alone in a one-input final chunk, which is precisely the path the new dual-chain resolver enables. Validate `output_amount >= dest.script_pubkey().dust_value()` before building the TxOut and return a typed build error, or fold the dust remainder into a sibling chunk so the sweep is all-or-nothing per UTXO.

<sub>source: ['codex']</sub>

thepastaclaw (commented):
## Code Review

Carried-forward prior findings: none — prior-1 (core_wallet_free_tx_bytes doc) is FIXED at packages/rs-platform-wallet-ffi/src/core_wallet/broadcast.rs:250-254, where the doc now explicitly enumerates both the send_to_addresses tx-bytes buffer and the sweep_coinjoin count*32 txid buffer. New findings in the latest delta (43ee9afe..41d5e8ac is docs-only): none introduced by the delta itself. Cumulative review against the current head surfaced one concrete edge case in the hand-built sweep tx assembly (dust-output bound) and a few low-value suggestions; no blocking issues, so APPROVE with a single suggestion comment.

🟡 1 suggestion(s)

<details>

```

In `packages/rs-platform-wallet/src/wallet/core/broadcast.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/core/broadcast.rs:340-346: Sweep chunk can build a sub-dust output and silently fall back to partial success
  The hand-built sweep tx bypasses `TransactionBuilder` and only guards `total_input <= fee`. If a chunk's `total_input - fee` lands above zero but below the destination script's dust threshold (≈546 duffs for P2PKH), the chunk produces a TxOut that relay policy will reject. Because broadcast loops over chunks and tolerates partial failure (line 408-432), the API can still return `Ok(broadcast)` while a dust chunk is silently dropped and those UTXOs remain unswept — exactly the failure mode the all-input chunking is designed to prevent. CoinJoin denominations are normally well above the bound, but DashSync-imported sub-denomination change can ride in on the internal chain and land alone in a one-input final chunk, which is precisely the path the new dual-chain resolver enables. Validate `output_amount >= dest.script_pubkey().dust_value()` before building the TxOut and return a typed build error, or fold the dust remainder into a sibling chunk so the sweep is all-or-nothing per UTXO.
```
</details>

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
llbartekll: Declining — this isn't reachable with the balanced chunking, so the sweep can't silently drop a sub-dust chunk.

**1. No lone-input chunk.** `sweep_chunk_size` ceil-divides so chunks are near-equal (501 → 251 + 250, never 500 + 1) — its doc comment calls this out: *"no chunk is a lone sub-fee dust input."* So the cited "sub-denomination change lands alone in a one-input final chunk" path doesn't exist. Whenever there are ≥2 chunks (N > 500) the smallest chunk still holds ≥250 inputs. A confirmed UTXO can't sit below the dust bound (relay rejects creating sub-dust outputs), so a 250-input chunk totals ≥ ~250 × 546 ≈ 136k duffs and its output (after the ~37k fee) is ~99k duffs — ~180× the dust threshold. No multi-chunk chunk can be sub-dust.

**2. Single-chunk sub-dust is a hard error, not silent.** A sub-dust output therefore requires the chunk total itself to be near-dust, which — with near-equal chunking — means the whole sweep is near-dust, i.e. N ≤ 500 → exactly one chunk. If that lone chunk fails to broadcast, the `if broadcast.is_empty() { return Err(..) }` guard after the broadcast loop returns an error. The silent `Ok(partial)` branch only runs when ≥1 chunk succeeded, which by (1) means none were sub-dust.

So "returns Ok while a dust chunk is silently unswept" can't occur. Happy to add an explicit `output_amount >= dest.script_pubkey().dust_value()` guard as a documented invariant if you'd prefer belt-and-suspenders, but it would only ever guard an unreachable state.

thepastaclaw (commented):
## Code Review

Carried-forward prior findings: none — prior-1 (core_wallet_free_tx_bytes doc) is FIXED at packages/rs-platform-wallet-ffi/src/core_wallet/broadcast.rs:250-254, where the doc now explicitly enumerates both the send_to_addresses tx-bytes buffer and the sweep_coinjoin count*32 txid buffer. New findings in the latest delta (43ee9afe..41d5e8ac is docs-only): none introduced by the delta itself. Cumulative review against the current head surfaced one concrete edge case in the hand-built sweep tx assembly (dust-output bound) and a few low-value suggestions; no blocking issues, so APPROVE with a single suggestion comment.

*Reviewed commit: 41d5e8ac*

Fresh response to a new review request. The PR head is unchanged since the exact-SHA automated review, so this records the same verified result without duplicating inline threads.


## Code Review



🟡 1 suggestion(s)

<details>

```

In `packages/rs-platform-wallet/src/wallet/core/broadcast.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/core/broadcast.rs:340-346: Sweep chunk can build a sub-dust output and silently fall back to partial success
  The hand-built sweep tx bypasses `TransactionBuilder` and only guards `total_input <= fee`. If a chunk's `total_input - fee` lands above zero but below the destination script's dust threshold (≈546 duffs for P2PKH), the chunk produces a TxOut that relay policy will reject. Because broadcast loops over chunks and tolerates partial failure (line 408-432), the API can still return `Ok(broadcast)` while a dust chunk is silently dropped and those UTXOs remain unswept — exactly the failure mode the all-input chunking is designed to prevent. CoinJoin denominations are normally well above the bound, but DashSync-imported sub-denomination change can ride in on the internal chain and land alone in a one-input final chunk, which is precisely the path the new dual-chain resolver enables. Validate `output_amount >= dest.script_pubkey().dust_value()` before building the TxOut and return a typed build error, or fold the dust remainder into a sibling chunk so the sweep is all-or-nothing per UTXO.
```
</details>

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
```diff
@@ -145,4 +168,487 @@ impl<B: TransactionBroadcaster + ?Sized> CoreWallet<B> {
...
+                    .iter()
+                    .map(|u| TxIn {
+                        previous_output: u.outpoint,
+                        script_sig: ScriptBuf::new(),
+                        sequence: 0xffff_ffff, // Dash has no RBF
+                        witness: Witness::new(),
+                    })
+                    .collect();
+                let unsigned = Transaction {
+                    version: 3,
```
ZocoLini: Isn't a better idea to use the TransactionBuilder, doing the needed modifications or does it need to much code to adapt to this use case??

Anyway, this code should live in rust-dashcore, platform-wallet core.rs module is just anwrapper around it (in theory)

packages/rs-platform-wallet/src/wallet/core/coinjoin_recovery.rs
```diff
@@ -0,0 +1,128 @@
+//! One-time CoinJoin recovery: widen a CoinJoin account's address gap limit
```
ZocoLini, line 1: This logic should be in rust-dashcore, and exposed directly though ffi to sdk consumers

ZocoLini (changes requested):

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
llbartekll: Agreed on both counts — moved upstream in dashpay/rust-dashcore#819.

- **TransactionBuilder**: the sweep now goes through it via a new `TransactionBuilder::sweep_to(dest)` *drain* mode (consume every added input into one no-change output `total − fee`, no coin selection). That's exactly why the original couldn't just reuse `build_signed` — its `LargestFirst` selection re-picks a covering subset and would leave small UTXOs behind. The drain mode also gets the dust-output guard for free.
- **Lives in rust-dashcore**: the chunking + dual-chain (`/0/` + `/1/`) path resolution now live in `key-wallet` as `ManagedCoreFundsAccount::build_coinjoin_sweep_txs`. This method here is now a thin wrapper: resolve the account under the wallet lock, delegate, broadcast.

packages/rs-platform-wallet/src/wallet/core/coinjoin_recovery.rs
llbartekll: Done — the gap-limit widening + address materialization now live in `key-wallet` as `ManagedCoreFundsAccount::set_coinjoin_gap_limit` (dashpay/rust-dashcore#819); this file is now a thin wrapper that just resolves the account and delegates.

On the FFI point: for this pass it's still exposed through `platform-wallet-ffi` (unchanged) rather than `key-wallet-ffi` directly — I kept the iOS Core FFI + Swift SDK rewiring out of scope here so the change stays focused. `.

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
```diff
@@ -145,4 +145,112 @@ impl<B: TransactionBroadcaster + ?Sized> CoreWallet<B> {
...
+                        "CoinJoin sweep: a chunk failed to broadcast, continuing \
+                         with remaining chunks (caller can re-run): {}",
+                        e
+                    );
+                    // Keep the FIRST failure (usually the root cause); the later
+                    // chunk errors are already surfaced via the warn! above.
+                    last_err.get_or_insert(e);
+                }
+            }
+        }
```
thepastaclaw, line 244: **💬 Nitpick: `last_err` actually holds the first error — rename to match the comment**

The variable is named `last_err` but is populated via `last_err.get_or_insert(e)` (line 241), which only writes when the slot is `None`. The inline comment at lines 239–240 correctly states 'Keep the FIRST failure (usually the root cause)'. The implementation matches the comment, but the name contradicts it. The common Rust pattern with `last_err` is `last_err = Some(e)` overwriting on each iteration; a reader scanning the variable name will draw the opposite conclusion to what the code does. Rename to `first_err` so the name agrees with `get_or_insert` and the inline comment.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Latest delta refactors CoinJoin sweep + gap-limit recovery into thin wrappers around upstream key-wallet helpers. The prior dust/partial-success finding is resolved: transaction construction (including dust validation) is now delegated to key_wallet::ManagedCoreFundsAccount::build_coinjoin_sweep_txs, and per-chunk broadcast failures emit tracing::warn! and propagate the first error if no chunk succeeds. No in-scope blocking issues remain; one readability nit on variable naming.

💬 1 nitpick(s)

PR #3819, opened 2026-06-08
feat: shielded scan-based spend detection and OVK outgoing-note history

QuantumExplorer:

Reworks how the shielded (Orchard) wallet learns about **spends** and **outgoing payments**, and removes a heavyweight client subsystem in the process. Two related changes:

1. **Scan-based spend detection (replaces the nullifier-sync subsystem).** The wallet already trial-decrypts every action during its note scan, and each scanned note carries the action's public `nullifier`. So a separate "nullifier sync" (trunk/branch tree scan + a compacted/recent nullifier-changes log) was redundant: spends can be detected by matching those already-downloaded nullifiers against the wallet's notes. This removes the entire client-side nullifier-sync path (and, as a bonus, sidesteps the proof-soundness/liveness issues that subsystem had).

2. **OVK outgoing-note history (Zcash-style).** Trial decryption finds notes you *received*; it tells you nothing about notes you *sent*. Recovering sent notes needs Orchard's outgoing-viewing-key (OVK) recovery of `out_ciphertext`, whose key derivation (`derive_ock(ovk, cv, cmx, epk)`) requires the note's value commitment **`cv_net`**. Platform discarded `cv_net` after consensus, so it had to be **retained on-chain**. With it stored, the wallet recovers its outgoing notes (recipient, value, memo) during the scan and persists them.

This is a **pre-release v12 shielded-pool** feature (no deployed network depends on the prior behavior).

## What was done?

**Part A — scan-based spend detection + nullifier-sync removal**
- During `sync_notes_across`, each scanned action's `nullifier` is fed to `ShieldedStore::mark_spent` (a no-op when unowned) after that batch's receipts persist — covering initial/restore and incremental sync.
- Removed the client nullifier-sync path: the SDK `nullifier_sync` module (`sync_nullifiers` + trunk/branch scan), `check_nullifiers_across`, the `Recent{,Compacted}NullifierChanges` `FromProof`/`Fetch`/mock types, and the nullifier-checkpoint plumbing (changeset/store/FFI/Swift). **Server-side** drive structures (the nullifier-changes trees, prove/verify, compaction) and the permanent consensus nullifier set are untouched.

**cv_net retention (depends on grovedb [#761](https://github.com/dashpay/grovedb/pull/761), merged to `develop`)**
- grovedb now stores `cmx ‖ rho ‖ cv_net ‖ payload` in the commitment-tree item. The Sinsemilla **anchor stays cmx-only**, so existing zk spend proofs are unaffected; only the BulkAppendTree (item-data) root changes. Pin bumped to the `develop` merge (`5eafc9e95d`).
- Platform threading: `ShieldedActionNote.cv_net` (from `SerializedAction.cv_net`) → `insert_note_op` storage → the two read sites parse the new offsets (`cv_net [64..96]`, `encrypted_note [96..]`) → `drive_proof_verifier::ShieldedEncryptedNote.cv_net` + the proto `EncryptedNote.cv_net` field → SDK/wasm. cv_net round-trips end-to-end under a new test.
- **Fee:** the +32 B/note is **not** fee-neutral — `SHIELDED_STORAGE_BYTES_PER_ACTION` 312 → 344 and `AVERAGE_NOTE_VALUE_SIZE` 280 → 312. The I5 metering tests confirm `fee ≥ actual cost` and `fee > storage` with ~5–10× margin.

**OVK outgoing recovery + history model**
- `try_recover_outgoing_note` builds a lightweight `ShieldedOutput` from the scan data (`cmx`, `epk`, `enc_ciphertext`) + `cv` from `cv_net`, and calls `orchard::zcash_note_encryption::try_output_recovery_with_ovk` with the wallet's OVK (`orchard` promoted to a direct SDK dep — the fn isn't re-exported by `grovedb-commitment-tree`; `rk` isn't stored, hence the custom output rather than a full `Action`).
- `sync_notes_across` attempts OVK recovery against each subwallet's OVK and records hits into a new `ShieldedOutgoingNote { cmx, recipient, value, memo, block_height }` model (`ShieldedStore` + `ShieldedChangeSet` + start-state rehydration + the Rust FFI persist/load callbacks). Idempotent by `cmx`.
- Swift: a `PersistentShieldedOutgoingNote` SwiftData model + the `on_persist`/`on_load(+free)` callbacks + the FFI struct mirrors (Swift persists/loads/bridges only; OVK logic lives in `rs-platform-wallet`).

## How Has This Been Tested?

- **Spend detection:** a note received then later spent (its nullifier appearing in a subsequent scanned action) is marked `is_spent` purely via the scan path — no nullifier-sync involved (`received_then_later_spent_note_is_marked_spent_via_scan` + 3 others).
- **cv_net round-trip:** insert → store → prove → verify returns the exact non-zero `cv_net` (and the Sinsemilla anchor is asserted unchanged in grovedb #761).
- **OVK recovery:** a real Orchard output encrypted with the wallet's OVK round-trips to the correct recipient/value/memo; a *different* OVK does **not** recover (no false positives); re-scan is idempotent.
- **Fees:** all three I5 `…fee_covers_actual_grovedb_write_cost` tests pass with the 344-byte budget vs the 312-byte note.
- Builds clean across `drive`, `dpp`, `drive-abci`, `drive-proof-verifier`, `dapi-grpc`, `dash-sdk`, `platform-wallet`, `platform-wallet-ffi`, `wasm-sdk` (incl. `--features verify`); **iOS simulator framework + SwiftExampleApp build succeeds** (`** BUILD SUCCEEDED **`). All tests verified against the merged grovedb `develop` rev.

## Breaking Changes

None for any deployed network — this is **pre-release v12** (shielded pool is not active anywhere). The on-chain shielded **note storage format changes** (gains `cv_net`), so the BulkAppendTree / app-hash root differs; dev/test state must be rebuilt (no migration — acceptable pre-release). The Orchard anchor is unchanged. Depends on grovedb #761 (merged). Not consensus-breaking, so no `!` per the consensus-only convention.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Persist and restore outgoing (sent) shielded notes; new UI to browse sent notes.
  * Outgoing-note recovery added so recovered sent notes (memos/recipients) surface in the app.

* **Bug Fixes & Improvements**
  * Note scans now detect spends during regular sync; verification improved.
  * Shielded note payload expanded (additional value field); storage/fee estimates updated.

* **Removed Features**
  * Dedicated nullifier synchronization APIs and related UI elements removed.


thepastaclaw:
✅ Review complete (commit 48e570bf)

QuantumExplorer:
Addressed the CodeRabbit findings:

- **persistence.rs (recipient padding)** — fixed in 562a77b: the outgoing-note FFI persist now **skips** (with a warn) a recipient that isn't exactly 43 bytes instead of zero-padding it into a wrong address.
- **sync.rs (nullifier drop)** — fixed in 562a77b: a malformed-length nullifier during scan replay now **fails fast** (consistent with the cmx check) rather than silently dropping (which would leave a spent note marked unspent).
- **rs-sdk/Cargo.toml (pin existence)** — verified: grovedb rev `5eafc9e95d` (develop, #761 merged) and orchard tag `dashified-0.14.0` both exist and resolve to a single orchard instance.
- **Swift purge on wallet deletion** — fixed in b257ccf: `deleteWalletData` now purges the shielded tables keyed by raw `walletId` (`PersistentShieldedNote`, `PersistentShieldedOutgoingNote`, `PersistentShieldedSyncState`) — none of them cascaded before, so all three leaked; fixed together.
- **Swift recipient zero-pad** — fixed in b257ccf: `loadShieldedOutgoingNotes` skips a row whose recipient isn't 43 bytes instead of zero-padding it.

iOS simulator framework + SwiftExampleApp build succeeds.

packages/dapi-grpc/protos/platform/v0/platform.proto
```diff
@@ -2904,6 +2904,7 @@ message GetShieldedEncryptedNotesResponse {
       bytes nullifier = 1;      // 32-byte nullifier (needed for Rho derivation in trial decryption)
       bytes cmx = 2;            // 32-byte extracted note commitment
       bytes encrypted_note = 3; // encrypted note payload (epk + enc_ciphertext + out_ciphertext)
+      bytes cv_net = 4;         // 32-byte value commitment (stored unencrypted, for OVK recovery of outgoing notes)
```
thepastaclaw, line 2907: **🔴 Blocking: Regenerate DAPI clients for new EncryptedNote.cv_net field**

`platform.proto` now adds `GetShieldedEncryptedNotesResponseV0.EncryptedNote.cv_net = 4`, but the checked-in generated DAPI clients were not regenerated. A repo-wide grep under `packages/dapi-grpc/clients` shows no `cv_net`/`cvNet`/`getCvNet`/`setCvNet`, and the web TS, nodejs JS, and Python generated outputs still expose only `nullifier`, `cmx`, and `encryptedNote`. Those consumers will decode field 4 as unknown and drop it, so they cannot perform OVK outgoing-note recovery even though the server sends the data. Please regenerate and commit the affected generated client artifacts with this proto change.

packages/rs-platform-wallet/src/wallet/shielded/sync.rs
```diff
@@ -374,8 +410,23 @@ pub(super) async fn sync_notes_across<S: ShieldedStore>(
...
         for (i, raw_note) in batch.notes.iter().enumerate() {
+            // A malformed nullifier length means the proven note item is corrupt;
+            // fail fast (consistent with the cmx check below) rather than silently
+            // dropping it — a dropped nullifier would leave a spent note marked
+            // unspent.
+            let nf_bytes = <[u8; 32]>::try_from(raw_note.nullifier.as_slice()).map_err(|_| {
+                PlatformWalletError::ShieldedSyncFailed("Invalid nullifier length".into())
+            })?;
+            scanned_nullifiers.push(nf_bytes);
             let global_pos = batch.start_index + i as u64;
```
thepastaclaw, line 430: **🟡 Suggestion: scanned_nullifiers accumulates for the entire stream — initial-sync memory pressure**

`scanned_nullifiers: Vec<[u8;32]>` collects every action's nullifier across all batches and is only drained into per-subwallet `mark_spent` after the stream completes. For an initial sync against a fully-loaded shielded pool the buffer grows in lockstep with chain history (32 B × N actions plus vector slack), on top of the per-batch `decrypted_by_subwallet` / `recovered_outgoing_by_subwallet` buffers that share the same lifetime. Receipt-before-spend ordering doesn't actually require holding the full nullifier set until end-of-stream — within a batch, receipts persist before `mark_spent` is replayed, and across batches the receipts were persisted on an earlier pass. Drain `scanned_nullifiers` (and ideally the receipt/outgoing buffers) per batch after that batch's saves persist, so client RAM is bounded by batch size rather than chain depth. Not a fund-loss risk (sync can be resumed after OOM) but mitigates a light-client DoS surface as the pool grows.

packages/rs-platform-wallet/src/wallet/shielded/sync.rs
```diff
@@ -374,8 +410,23 @@ pub(super) async fn sync_notes_across<S: ShieldedStore>(
...
+        //    whole scan for the spend-match replay after receipts are
+        //    persisted. These are collected for ALL positions (even
+        //    those gate-skipped for tree append: a re-fetched chunk can
+        //    still surface a spend whose receipt only persisted on a
+        //    later pass), since `mark_spent` is an idempotent no-op for
+        //    nullifiers the wallet doesn't own or already marked spent.
         for (i, raw_note) in batch.notes.iter().enumerate() {
+            // A malformed nullifier length means the proven note item is corrupt;
+            // fail fast (consistent with the cmx check below) rather than silently
+            // dropping it — a dropped nullifier would leave a spent note marked
```
thepastaclaw, line 424: **🟡 Suggestion: Silent drop of malformed-length nullifier can mask real spends**

`if let Ok(nf_bytes) = <[u8; 32]>::try_from(raw_note.nullifier.as_slice()) { scanned_nullifiers.push(nf_bytes); }` quietly discards any nullifier whose wire length isn't exactly 32 bytes. The very next statement (cmx parse) correctly returns `ShieldedSyncFailed` on bad length, so the two adjacent fields use opposite policies for the same class of corruption. A malformed nullifier here silently omits a real spend from the replay set; the wallet would then attempt to spend an already-spent note and only discover it at proof-rejection time. At minimum `warn!` with the action position (matching the cmx error site's verbosity), or hard-fail to align with the cmx behavior — the scan invariant is that every action carries a 32-byte nullifier.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet-ffi/src/persistence.rs
```diff
@@ -1158,16 +1184,46 @@ impl PlatformWalletPersistence for FFIPersister {
...
+                                // from OVK recovery; reject (skip + warn) a
+                                // malformed row rather than silently zero-padding
+                                // it into a wrong address.
+                                let recipient: [u8; 43] = match n.recipient.as_slice().try_into() {
+                                    Ok(r) => r,
+                                    Err(_) => {
+                                        tracing::warn!(
+                                            recipient_len = n.recipient.len(),
+                                            "skipping outgoing-note persist row: \
+                                                 recipient is not the expected 43 bytes"
```
thepastaclaw, line 1211: **🟡 Suggestion: Defensive zero-padding of `recipient` masks corruption silently**

The FFI builder zero-pads `n.recipient` to 43 bytes when the source `Vec` is shorter and truncates when longer. The only producer (`sync_notes_across` building `ShieldedOutgoingNote { recipient: addr.to_raw_address_bytes().to_vec(), .. }`) is statically guaranteed to emit exactly 43 bytes, so either branch firing indicates a real Rust-side bug or corrupted persisted state. Silently zero-padding ships a valid-looking but truncated raw Orchard address to the host (and ultimately to the UI as a corrupted send-history row), which is worse than a loud failure. Replace with `debug_assert_eq!(n.recipient.len(), 43)` plus a release-mode `warn!`+skip (or hard error), so dev catches the divergence and release doesn't propagate garbage.

<sub>source: ['claude']</sub>

thepastaclaw (dismissed):
## Code Review

The core Rust storage/proof/wallet flow looks coherent, but the proto change is missing regenerated DAPI client artifacts, so non-Rust clients will drop cv_net. I also kept three defensive follow-ups around scan memory bounds and malformed persistence/nullifier data handling.

<details>

```

In `packages/dapi-grpc/protos/platform/v0/platform.proto`:
- [BLOCKING] packages/dapi-grpc/protos/platform/v0/platform.proto:2907: Regenerate DAPI clients for new EncryptedNote.cv_net field
  `platform.proto` now adds `GetShieldedEncryptedNotesResponseV0.EncryptedNote.cv_net = 4`, but the checked-in generated DAPI clients were not regenerated. A repo-wide grep under `packages/dapi-grpc/clients` shows no `cv_net`/`cvNet`/`getCvNet`/`setCvNet`, and the web TS, nodejs JS, and Python generated outputs still expose only `nullifier`, `cmx`, and `encryptedNote`. Those consumers will decode field 4 as unknown and drop it, so they cannot perform OVK outgoing-note recovery even though the server sends the data. Please regenerate and commit the affected generated client artifacts with this proto change.

In `packages/rs-platform-wallet/src/wallet/shielded/sync.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/sync.rs:395-430: scanned_nullifiers accumulates for the entire stream — initial-sync memory pressure
  `scanned_nullifiers: Vec<[u8;32]>` collects every action's nullifier across all batches and is only drained into per-subwallet `mark_spent` after the stream completes. For an initial sync against a fully-loaded shielded pool the buffer grows in lockstep with chain history (32 B × N actions plus vector slack), on top of the per-batch `decrypted_by_subwallet` / `recovered_outgoing_by_subwallet` buffers that share the same lifetime. Receipt-before-spend ordering doesn't actually require holding the full nullifier set until end-of-stream — within a batch, receipts persist before `mark_spent` is replayed, and across batches the receipts were persisted on an earlier pass. Drain `scanned_nullifiers` (and ideally the receipt/outgoing buffers) per batch after that batch's saves persist, so client RAM is bounded by batch size rather than chain depth. Not a fund-loss risk (sync can be resumed after OOM) but mitigates a light-client DoS surface as the pool grows.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/sync.rs:421-424: Silent drop of malformed-length nullifier can mask real spends
  `if let Ok(nf_bytes) = <[u8; 32]>::try_from(raw_note.nullifier.as_slice()) { scanned_nullifiers.push(nf_bytes); }` quietly discards any nullifier whose wire length isn't exactly 32 bytes. The very next statement (cmx parse) correctly returns `ShieldedSyncFailed` on bad length, so the two adjacent fields use opposite policies for the same class of corruption. A malformed nullifier here silently omits a real spend from the replay set; the wallet would then attempt to spend an already-spent note and only discover it at proof-rejection time. At minimum `warn!` with the action position (matching the cmx error site's verbosity), or hard-fail to align with the cmx behavior — the scan invariant is that every action carries a 32-byte nullifier.

- [SUGGESTION] packages/rs-platform-wallet-ffi/src/persistence.rs:1206-1211: Defensive zero-padding of `recipient` masks corruption silently
  The FFI builder zero-pads `n.recipient` to 43 bytes when the source `Vec` is shorter and truncates when longer. The only producer (`sync_notes_across` building `ShieldedOutgoingNote { recipient: addr.to_raw_address_bytes().to_vec(), .. }`) is statically guaranteed to emit exactly 43 bytes, so either branch firing indicates a real Rust-side bug or corrupted persisted state. Silently zero-padding ships a valid-looking but truncated raw Orchard address to the host (and ultimately to the UI as a corrupted send-history row), which is worse than a loud failure. Replace with `debug_assert_eq!(n.recipient.len(), 43)` plus a release-mode `warn!`+skip (or hard error), so dev catches the divergence and release doesn't propagate garbage.
```
</details>

QuantumExplorer:
@thepastaclaw responses:

- **🔴 proto / regenerate DAPI clients** — agreed, the committed `dapi-grpc` clients (nodejs/web/python/objc) need regenerating for `EncryptedNote.cv_net`. That codegen runs through the pinned dockerized protoc (`rvolosatovs/protoc:4.0.0` via `yarn workspace @dashevo/dapi-grpc build`), which I can't run in this environment (no Docker). It needs to be run before merge — flagging as a required follow-up. Left unresolved.
- **🟡 nullifier silent-drop** — already fixed in 562a77b (the review was on an earlier head): the malformed-length nullifier now fails fast instead of being dropped. Resolved.
- **🟡 recipient zero-pad** — already fixed in 562a77b: the FFI persist now skips a non-43-byte recipient instead of zero-padding. Resolved.
- **🟡 scanned_nullifiers memory pressure** — valid for very large initial syncs. The accumulation is currently structural: receipts persist at end-of-pass (so a re-fetched buffer chunk can surface a spend whose receipt only lands later in the pass), which is why spends are replayed after the whole pass. A per-batch flush would require moving receipt persistence per-batch — a reasonable follow-up optimization, but it changes the verified ordering, so I'd prefer to do it separately rather than late in this PR. Left open for your call.

thepastaclaw:
I regenerated the `dapi-grpc` clients for `EncryptedNote.cv_net` and pushed the fix here since I don't have permission to push directly to the PR branch:

- branch: https://github.com/thepastaclaw/platform/tree/tracker-1461-dapi-cv-net-clients
- commit: https://github.com/thepastaclaw/platform/commit/f7e8f03d140eef3e0aa62cefe2e8dff1d74b3ddf

What changed:
- adds `cv_net`/`cvNet`/`getCvNet`/`setCvNet` support to regenerated nodejs/web/TS/Python/ObjC artifacts
- `drive_pbjs.js` also updates because it imports `platform.proto`
- the normal generator also caught up stale Java/Python/ObjC service artifacts from existing proto changes

Validation:
- ran `packages/dapi-grpc/scripts/build.sh` through the repo generator path (`rvolosatovs/protoc:4.0.0`, via podman shim because Docker socket denied this user)
- `git diff --check` passed
- node pbjs encode/decode smoke test for `cvNet` passed
- `yarn workspace @dashevo/dapi-grpc test` was not usable on this host due Node v25 ESM loader `EBADF` before tests ran

If you can cherry-pick `f7e8f03d140eef3e0aa62cefe2e8dff1d74b3ddf` (or grant push permission to the PR branch), this should resolve the blocking generated-client item.

QuantumExplorer:
Done — regenerated the DAPI clients (nodejs/web/python/objc/java) via the pinned `rvolosatovs/protoc:4.0.0` codegen in `35e6198`; `EncryptedNote.cv_net` now flows through to all generated clients. Resolves the blocking proto finding.

packages/rs-platform-wallet/src/wallet/shielded/sync.rs
thepastaclaw: Resolved in this update — _scanned_nullifiers accumulates for the entire stream — initial-sync memory pressure_ no longer present.

packages/rs-platform-wallet/src/wallet/shielded/sync.rs
```diff
@@ -539,6 +618,57 @@ pub(super) async fn sync_notes_across<S: ShieldedStore>(
...
+    // against each subwallet — see [`apply_scanned_nullifier_spends`]
+    // for the ordering rationale.
+    let subwallet_ids: Vec<SubwalletId> = subwallets.iter().map(|(id, _)| *id).collect();
+    let newly_spent_per_subwallet = apply_scanned_nullifier_spends(
+        &mut *store,
+        &subwallet_ids,
+        &scanned_nullifiers,
+        &mut changeset,
+    )
+    .map_err(|e| PlatformWalletError::ShieldedStoreError(e.to_string()))?;
```
thepastaclaw, line 670: **🟡 Suggestion: scanned_nullifiers accumulates for the whole sync stream instead of per batch**

Verified on current head d4b8873f (carry-forward from prior review of 35e61983 — unchanged in the latest delta).

`scanned_nullifiers: Vec<[u8; 32]>` is declared at sync.rs:395, before the `while let Some(item) = stream.next().await` loop. Every action in every streamed batch pushes a 32-byte nullifier at :429, and the buffer is only consumed once the stream is fully drained, by `apply_scanned_nullifier_spends` at :664. Resident memory therefore grows with the total scanned chain range (≈32 MB per million actions) rather than being bounded by the configured batch size (~8192 notes).

Why it matters here: this wallet code ships into iOS via the platform-wallet FFI, and a fresh restore over a long-lived shielded pool must scan every action ever produced. Because the shielded action history is influenced by all network participants, the worst-case footprint scales with adoption — a deep restore on a memory-constrained client could OOM.

The receipt-before-spend ordering documented at :386–:394 is already satisfied per-batch (the receipt save runs first within each iteration), so moving the `apply_scanned_nullifier_spends` call inside the loop — operating on just that batch's nullifiers after that batch's receipts persist — preserves correctness while bounding RAM by batch size. `mark_spent` is documented as an idempotent no-op for unowned nullifiers, so per-batch replay is safe.

Not a correctness defect; suggestion-severity bounding fix.

thepastaclaw (commented):
## Code Review

Latest delta (35e61983→d4b8873f) is a clean refactor naming the shielded-note verify result struct and dropping the type_complexity allow; no new defects introduced. One prior carried-forward suggestion remains STILL VALID: scanned_nullifiers buffer in shielded sync grows with chain history rather than batch size during initial restore. A codex-FFI claim about PersistenceCallbacks ABI was dropped as a false positive (this internal SDK has no binary-compat contract; cbindgen header is regenerated per build).

🟡 1 suggestion(s)

<details>

```

In `packages/rs-platform-wallet/src/wallet/shielded/sync.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/sync.rs:395-670: scanned_nullifiers accumulates for the whole sync stream instead of per batch
  Verified on current head d4b8873f (carry-forward from prior review of 35e61983 — unchanged in the latest delta).

  `scanned_nullifiers: Vec<[u8; 32]>` is declared at sync.rs:395, before the `while let Some(item) = stream.next().await` loop. Every action in every streamed batch pushes a 32-byte nullifier at :429, and the buffer is only consumed once the stream is fully drained, by `apply_scanned_nullifier_spends` at :664. Resident memory therefore grows with the total scanned chain range (≈32 MB per million actions) rather than being bounded by the configured batch size (~8192 notes).

  Why it matters here: this wallet code ships into iOS via the platform-wallet FFI, and a fresh restore over a long-lived shielded pool must scan every action ever produced. Because the shielded action history is influenced by all network participants, the worst-case footprint scales with adoption — a deep restore on a memory-constrained client could OOM.

  The receipt-before-spend ordering documented at :386–:394 is already satisfied per-batch (the receipt save runs first within each iteration), so moving the `apply_scanned_nullifier_spends` call inside the loop — operating on just that batch's nullifiers after that batch's receipts persist — preserves correctness while bounding RAM by batch size. `mark_spent` is documented as an idempotent no-op for unowned nullifiers, so per-batch replay is safe.

  Not a correctness defect; suggestion-severity bounding fix.
```
</details>

packages/rs-platform-wallet/src/wallet/shielded/sync.rs
thepastaclaw: Resolved in `a3b5f9a` — _scanned_nullifiers accumulates for the whole sync stream instead of per batch_ no longer present.

packages/rs-drive/src/verify/shielded/verify_shielded_encrypted_notes/mod.rs
```diff
@@ -6,6 +6,21 @@ use crate::error::Error;
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct VerifiedShieldedEncryptedNote {
+    /// 32-byte note commitment.
+    pub cmx: Vec<u8>,
+    /// 32-byte nullifier (the output note's rho).
+    pub nullifier: Vec<u8>,
+    /// 32-byte Orchard value commitment (used for OVK outgoing-note recovery).
+    pub cv_net: Vec<u8>,
+    /// Encrypted note ciphertext (`epk || enc_ciphertext || out_ciphertext`).
+    pub encrypted_note: Vec<u8>,
```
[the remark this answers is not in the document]
QuantumExplorer: Keeping `nullifier` (declined the rename to `rho`) after checking the crypto: these bytes ARE a real published nullifier — the nullifier of the note **spent** in this action — which Orchard reuses as this output note's ρ (the `ρ = nf^old` rule). grovedb's own field doc says exactly this ("Nullifier (`rho`) of the spent note"), and `decrypt.rs` notes "the output note's rho equals the spent note's nullifier in Orchard". So the premise that it is "not a public nullifier" is not quite right — it is one, and it's precisely what scan-based spend detection matches against.

`nullifier` is also consistent across the system: `SerializedAction.nullifier`, the wire `ShieldedEncryptedNote.nullifier`, the `.proto` field, and the Part-A spend-detection path. Renaming only this verify type to `rho` would create a `rho → nullifier` mismatch with the wire type it maps into.

I did tighten the doc to remove the misuse risk you flagged (bec4f5e): it now reads "Nullifier of the note spent in this action — Orchard reuses it as this output note's ρ for trial decryption. Not the nullifier of the note committed by `cmx`."

thepastaclaw (commented):
## Code Review

Latest delta (78b89471→bec4f5e24) is a tight, well-scoped refactor: VerifiedShieldedEncryptedNote moves to fixed-size [u8;32]/[u8;216] fields, rejecting any item not exactly 312 bytes, while the SDK-facing ShieldedEncryptedNote keeps Vec<u8> via .to_vec() so no API breaks. No new in-scope blockers were introduced. The prior wallet-sync memory-bounding finding remains STILL VALID on current head and is carried forward as the only actionable in-scope suggestion.

🟡 1 suggestion(s)

<details>

```

In `packages/rs-platform-wallet/src/wallet/shielded/sync.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/sync.rs:395-670: scanned_nullifiers buffer grows with the full sync range instead of per batch
  Carried forward from the prior review and re-verified on current head bec4f5e24d. `let mut scanned_nullifiers: Vec<[u8; 32]> = Vec::new();` is still declared at line 395 before the `while let Some(item) = stream.next().await` loop, every action of every batch unconditionally pushes a 32-byte entry at line 429, and the buffer is only consumed at line 664 via `apply_scanned_nullifier_spends` after the stream fully drains. Resident memory therefore scales with the total scanned chain range rather than the configured batch size (~8192 notes). During a full restore or large catch-up — particularly on mobile where the OS will kill the wallet under heap pressure — this presents a soft-DoS / memory-pressure surface that forces repeated partial-sync retries. The receipt-before-spend invariant is already satisfied per-batch (this batch's receipts are persisted via `store.save_note` before any nullifier replay would run), and `mark_spent` is documented as idempotent / no-op for unowned nullifiers, so applying `apply_scanned_nullifier_spends` once per batch (with a per-batch Vec, or by draining a reused buffer) preserves correctness while bounding RAM to the batch size.
```
</details>

packages/rs-platform-wallet/src/wallet/shielded/sync.rs
thepastaclaw, line 670: **🟡 Suggestion: scanned_nullifiers buffer grows with the full sync range instead of per batch**

Carried forward from the prior review and re-verified on current head bec4f5e24d. `let mut scanned_nullifiers: Vec<[u8; 32]> = Vec::new();` is still declared at line 395 before the `while let Some(item) = stream.next().await` loop, every action of every batch unconditionally pushes a 32-byte entry at line 429, and the buffer is only consumed at line 664 via `apply_scanned_nullifier_spends` after the stream fully drains. Resident memory therefore scales with the total scanned chain range rather than the configured batch size (~8192 notes). During a full restore or large catch-up — particularly on mobile where the OS will kill the wallet under heap pressure — this presents a soft-DoS / memory-pressure surface that forces repeated partial-sync retries. The receipt-before-spend invariant is already satisfied per-batch (this batch's receipts are persisted via `store.save_note` before any nullifier replay would run), and `mark_spent` is documented as idempotent / no-op for unowned nullifiers, so applying `apply_scanned_nullifier_spends` once per batch (with a per-batch Vec, or by draining a reused buffer) preserves correctness while bounding RAM to the batch size.

thepastaclaw (commented):
## Code Review

The latest delta (bec4f5e2..72f0f1587) is a pure test-file relocation — `ovk_recovery_tests` was extracted from sync.rs into a sibling test file with no production code change. The prior carried-forward finding on `scanned_nullifiers` was re-verified at current head 72f0f1587 and is STILL VALID: the buffer is declared at sync.rs:395 before the stream loop, every action of every batch pushes a 32-byte entry at sync.rs:429, and the buffer is only consumed after the stream fully drains at sync.rs:664-670, so resident memory scales with the total scanned chain range rather than the configured batch size.

Note: the carried-forward finding below matches an already-open review thread from an earlier SHA, so I am not opening a duplicate inline thread here. This top-level review pins the re-validation to the current head `72f0f1587d`.

🟡 1 suggestion(s)

<details>

```

In `packages/rs-platform-wallet/src/wallet/shielded/sync.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/sync.rs:395-670: scanned_nullifiers buffer grows with the full sync range instead of per batch
  Verified on current head 72f0f1587. `let mut scanned_nullifiers: Vec<[u8; 32]> = Vec::new();` is declared at sync.rs:395 before the `while let Some(item) = stream.next().await` loop at sync.rs:397. Inside the loop, every action of every batch unconditionally pushes a 32-byte entry at sync.rs:429 (including positions gate-skipped for tree append at sync.rs:431). The buffer is only consumed after the stream fully drains, via `apply_scanned_nullifier_spends(&mut *store, &subwallet_ids, &scanned_nullifiers, &mut changeset)` at sync.rs:664-670. Resident memory therefore scales linearly with the total scanned chain range (≈32 bytes × total actions) instead of with the configured ~8192-note batch size. On mobile wallet hosts (rs-platform-wallet runs under the Swift FFI surface), a full restore or large catch-up over a heavily populated shielded pool can grow this allocation enough to create memory pressure or OS-level termination, and an adversary growing the public shielded action set widens the surface.

The receipt-before-spend invariant is already satisfied per-batch (the per-batch incoming-receipt save at sync.rs:570-619 runs before any subsequent batch's spend match), and `mark_spent` is documented as idempotent and a no-op for nullifiers the wallet doesn't own. Moving `apply_scanned_nullifier_spends` inside the stream loop — using a per-batch Vec, or `drain(..)` on a reused buffer after each batch's receipt persistence — preserves correctness while bounding RAM to O(batch_size) instead of O(scan_range). Verify the ordering claim documented at sync.rs:380-394 still holds when the call moves per-batch (specifically that batch N's receipts persist before its spend replay in the same iteration).
```
</details>

packages/rs-platform-wallet/src/wallet/shielded/sync.rs
QuantumExplorer: Valid concern — tracked as #3820 and deferred out of this PR.

One correction on the premise though: receipts are **not** persisted per-batch. The loop only accumulates decrypted notes in `decrypted_by_subwallet`; `store.save_note` runs **once at end-of-pass** (sync.rs:613), and `apply_scanned_nullifier_spends` right after (sync.rs:664). So applying the replay per-batch on its own would run against a store that hasn't saved this pass's receipts yet and would miss every same-pass receive-then-spend.

The correct fix is therefore heavier than a one-liner: move the receipt-save **and** spend-apply into the loop so each batch is processed fully (append → save receipts → record outgoing → apply that batch's spends). That bounds all three buffers to ~8192 notes (~256 KB vs ~32 MB/1M notes) and is captured with full rationale + a correctness/behavior note in #3820. Deferring since it's non-blocking and restructures the verified sync loop (which currently has no full-loop integration coverage).

QuantumExplorer (commented):
Reviewed

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -2483,6 +2520,111 @@ public class PlatformWalletPersistenceHandler {
...
+                buf[written] = ShieldedOutgoingNoteRestoreFFI(
+                    wallet_id: walletIdTuple,
+                    account_index: row.accountIndex,
+                    cmx: cmxTuple,
+                    recipient: recipientTuple,
+                    value: row.value,
+                    block_height: row.blockHeight,
+                    memo_ptr: UnsafePointer(memoBuf),
+                    memo_len: UInt(row.memo.count)
+                )
```
thepastaclaw, line 2605: **🟡 Suggestion: Bound restored outgoing-note memo length before handing it to Rust**

`loadShieldedOutgoingNotes` already enforces fixed-length invariants for the other fields on this new restore path — 32-byte `walletId`, 32-byte `cmx`, and explicitly 43-byte `recipient` (lines 2569–2579), skipping malformed rows with a log to match the Rust persist side. The memo field has no equivalent guard: any value of `row.memo.count` is allocated, copied, and passed to Rust as `memo_ptr` / `memo_len`. The Rust contract on `ShieldedOutgoingNoteRestoreFFI.memo_ptr` is a raw Dash memo, documented as always 36 bytes, and the Rust load path trusts `memo_len` and copies that many bytes from the host pointer into a `Vec`. A malformed or migrated SwiftData row therefore bypasses the size contract and can either (a) rehydrate a wrong-shaped memo into the wallet or (b) force an arbitrarily large copy at boot. Adding a `guard row.memo.count == 36` skip (or `<= 36`) mirrors the recipient guard immediately above and closes the asymmetry on the new outgoing-note restore boundary introduced by this PR.

<sub>source: ['codex']</sub>

thepastaclaw (commented):
## Code Review

Cumulative review at 48e570bf74. The latest delta is test-only (wasm-sdk ShieldedEncryptedNote fixtures cover the new cvNet field). One prior performance suggestion (unbounded scanned_nullifiers buffer in shielded sync) is re-verified STILL VALID on current head and carried forward. One new in-scope suggestion: the Swift-side outgoing-note restore path does not bound the memo length the way it bounds recipient length, leaving a defense-in-depth gap on the new SwiftData → Rust restore boundary. No blocking issues.

🟡 1 suggestion(s)

<details>

```

- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift:2568-2605: Bound restored outgoing-note memo length before handing it to Rust
  `loadShieldedOutgoingNotes` already enforces fixed-length invariants for the other fields on this new restore path — 32-byte `walletId`, 32-byte `cmx`, and explicitly 43-byte `recipient` (lines 2569–2579), skipping malformed rows with a log to match the Rust persist side. The memo field has no equivalent guard: any value of `row.memo.count` is allocated, copied, and passed to Rust as `memo_ptr` / `memo_len`. The Rust contract on `ShieldedOutgoingNoteRestoreFFI.memo_ptr` is a raw Dash memo, documented as always 36 bytes, and the Rust load path trusts `memo_len` and copies that many bytes from the host pointer into a `Vec`. A malformed or migrated SwiftData row therefore bypasses the size contract and can either (a) rehydrate a wrong-shaped memo into the wallet or (b) force an arbitrarily large copy at boot. Adding a `guard row.memo.count == 36` skip (or `<= 36`) mirrors the recipient guard immediately above and closes the asymmetry on the new outgoing-note restore boundary introduced by this PR.
```
</details>

QuantumExplorer:
Reviewed

PR #3821, opened 2026-06-09
fix(swift-sdk): example app ask if the spv is running directly instead of using the sync state

ZocoLini:
Before the arquitectural change the example app was checking if the spv client was running  using the sync progress state, in this PR we are replacing that logic with directly calls to the exposed spv client methods that correctly return if the spv client is running or not

## Checklist:

## Summary by CodeRabbit

* **Refactor**
  * SPV running status handling was redesigned and is now exposed as a stable read-only indicator.
  * The UI now reads this indicator to drive Start/Pause and Clear button enabled/disabled states, improving reliability of those controls.
  * SPV polling was adjusted to update the indicator only when the running state changes, reducing unnecessary UI updates.

### DIFF
--- a/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerSPV.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerSPV.swift
@@ -8,13 +8,6 @@ public enum PlatformSpvSyncState: UInt32, Sendable {
     case syncing = 2
     case synced = 3
     case error = 4
-
-    public var isRunning: Bool {
-        switch self {
-        case .waitingForConnections, .syncing: return true
-        case .waitForEvents, .synced, .error: return false
-        }
-    }
 }
 
 /// Progress for one of the SPV sub-managers (headers, filters, masternodes, etc.).
--- a/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManager.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManager.swift
@@ -39,6 +39,8 @@ public class PlatformWalletManager: ObservableObject {
     /// started in [`configure`].
     @Published public private(set) var spvProgress: PlatformSpvSyncProgress = .empty
 
+    @Published public private(set) var spvIsRunning: Bool = false
+
     /// Block time of the SPV header storage's current tip (if any).
     /// `nil` while SPV isn't running or hasn't stored a header yet.
     /// Useful as a "is core producing blocks?" indicator — when this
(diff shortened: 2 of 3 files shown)


thepastaclaw:
✅ Review complete (commit c59d12be)

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerSPV.swift
```diff
@@ -8,13 +8,6 @@ public enum PlatformSpvSyncState: UInt32, Sendable {
...
     case synced = 3
     case error = 4
-
-    public var isRunning: Bool {
-        switch self {
-        case .waitingForConnections, .syncing: return true
-        case .waitForEvents, .synced, .error: return false
-        }
-    }
 }
```
thepastaclaw, line 11: **🟡 Suggestion: Removing public PlatformSpvSyncState.isRunning is source-breaking**

`PlatformSpvSyncState.isRunning` was a public computed property on a public enum in `SwiftDashSDK`. Downstream apps reading `progress.overallState.isRunning` will fail to compile after this PR. The PR title isn't marked breaking and the breaking-change checkbox is unchecked, but the deletion silently changes the public surface. Either mark the PR breaking, or retain the property with a `@available(*, deprecated, message: "Use PlatformWalletManager.spvIsRunning")` annotation pointing to the new source of truth. Note the new signal (FFI `spv_is_running`) is semantically different from the old (derived from progress state: only `waitingForConnections`/`syncing` returned true), so silently keeping a no-op shim is not equivalent — deprecation with redirect is the right shape.

```suggestion
public enum PlatformSpvSyncState: UInt32, Sendable {
    case waitForEvents = 0
    case waitingForConnections = 1
    case syncing = 2
    case synced = 3
    case error = 4

    @available(*, deprecated, message: "Use PlatformWalletManager.spvIsRunning to observe the SPV client lifecycle.")
    public var isRunning: Bool {
        switch self {
        case .waitingForConnections, .syncing: return true
        case .waitForEvents, .synced, .error: return false
        }
    }
}
```

thepastaclaw (commented):
## Code Review

Small, focused change replacing a derived isRunning enum property with a directly-polled @Published Bool from the SPV FFI. The convergent concern is that removing the public PlatformSpvSyncState.isRunning is technically source-breaking for downstream SDK consumers, though the new spvIsRunning property cleanly replaces it. Treated as a suggestion rather than blocking given the swift-sdk's active-development status and trivial migration path.

<details>

```

In `packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerSPV.swift`:
- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerSPV.swift:5-11: Removing public PlatformSpvSyncState.isRunning is source-breaking
  `PlatformSpvSyncState.isRunning` was a public computed property on a public enum in `SwiftDashSDK`. Downstream apps reading `progress.overallState.isRunning` will fail to compile after this PR. The PR title isn't marked breaking and the breaking-change checkbox is unchecked, but the deletion silently changes the public surface. Either mark the PR breaking, or retain the property with a `@available(*, deprecated, message: "Use PlatformWalletManager.spvIsRunning")` annotation pointing to the new source of truth. Note the new signal (FFI `spv_is_running`) is semantically different from the old (derived from progress state: only `waitingForConnections`/`syncing` returned true), so silently keeping a no-op shim is not equivalent — deprecation with redirect is the right shape.
```
</details>

llbartekll (approved):

PR #3837, opened 2026-06-10
chore(swift-sdk): reduced swift-sdk static library size using Cargo profiles

ZocoLini:
The static/dynamic library used by the swift sdk (and in a future the kotlin sdk) has always been huge, I made a PR some time ago that reduced the size by a 50% and today I am reducing the size, once again, up to a 80% in release builds.

The issue was that, some packages where defining a [profile.release] section in their Cargo.toml but those section were being ignored because Cargo workspaces only cares about profiles defined in the root Cargo.toml. What I did is basically define 2 new profiles (dev-ios and release-ios).

Results:
 - Dev static lib build goes from 2.1GB to 817MB, almost 40% size reduction
 - Release static lib goes from 192MB to 54MB, almost 75% size reduction
 
Dyn libs are even smaller, more or less the same size the static library will be in the final binary after linking, around 29MB for release mode. Shipping a dynamic lib in the swift-sdk is something possible that would reduce even more the size but the build script needs to be updated.

Also, I set opt-level to 3 for runtime performance but If reducing size is a priority we can try opt-level = "z", if the performance loss is acceptable. The size reduction would be a 10% for static libraries and a 50% for dynamic ones (relative to the sizes this PR is generating)

## Checklist:

## Summary by CodeRabbit

* **Chores**
  * Added iOS-specific build profiles to provide optimized debug and release compilation settings for iOS builds.
  * Build tooling updated to derive and use these iOS profile names (e.g., dev → dev-ios, release → release-ios) and to place artifacts in matching output directories.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,6 +59,24 @@ key-wallet-manager = { git = "https://github.com/dashpay/rust-dashcore", rev = "
 dash-network = { git = "https://github.com/dashpay/rust-dashcore", rev = "3d0d5dcd4ad64e2199a726651bca7f8ffac123e6" }
 dashcore-rpc = { git = "https://github.com/dashpay/rust-dashcore", rev = "3d0d5dcd4ad64e2199a726651bca7f8ffac123e6" }
 
+# Size-tuned profile for the iOS `rs-unified-sdk-ffi` staticlib, which
+# otherwise ships huge. Inherits `release` and is ONLY used by the iOS
+# build (`build_ios.sh --profile release`)
+[profile.release-ios]
+inherits = "release"
+panic = "abort"
+strip = "symbols"
+lto = "fat"
+codegen-units = 1
+opt-level = 3
+
+# Debug counterpart for smaller iOS dev builds. Inherits the standard `dev`
+# profile (and is ONLY used by the iOS build (`build_ios.sh --profile dev`)
+[profile.dev-ios]
+inherits = "dev"
+panic = "abort"
+lto = "thin"
+
 # Optimize heavy crypto crates even in dev/test builds so that
 # Halo 2 proof generation and verification run at near-release speed.
 # Without this, ZK operations are 10-100x slower (debug field arithmetic).
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -272,6 +272,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "atomic-polyfill"
+version = "1.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8cf2bce30dfe09ef0bfaef228b9d414faaf7e563035494d7fe092dba54b300f4"
+dependencies = [
+ "critical-section",
+]
+
+[[package]]
 name = "atomic-waker"
 version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -279,9 +288,9 @@ checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
 
 [[package]]
 name = "autocfg"
-version = "1.5.0"
+version = "1.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
+checksum = "f2032f911046de80f0a198e0901378627c33f59ea0ac00e363d481118bd70a53"
 
 [[package]]
 name = "aws-lc-rs"
@@ -440,11 +449,10 @@ checksum = "6107fe1be6682a68940da878d9e9f5e90ca5745b3dec9fd1bb393c8777d4f581"
 
 [[package]]
 name = "base58ck"
-version = "0.1.0"
+version = "0.1.100"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2c8d66485a3a2ea485c1913c4572ce0256067a5377ac8c75c4960e1cda98605f"
+checksum = "ec5dc7e09f7bb15f0062da7c03086d6b71a2c84e0af4fccbbc7d8c6559847816"
 dependencies = [
- "bitcoin-internals",
  "bitcoin_hashes",
 ]
 
@@ -537,7 +545,7 @@ dependencies = [
  "quote",
  "regex",
  "rustc-hash 1.1.0",
- "shlex",
+ "shlex 1.3.0",
  "syn 2.0.117",
  "which",
 ]
@@ -548,15 +556,15 @@ version = "0.72.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "993776b509cfb49c750f11b8f07a46fa23e0a1386ffc01fb1e7d343efc387895"
 dependencies = [
- "bitflags 2.11.1",
+ "bitflags 2.13.0",
  "cexpr",
  "clang-sys",
- "itertools 0.10.5",
+ "itertools 0.13.0",
  "proc-macro2",
  "quote",
  "regex",
  "rustc-hash 2.1.2",
- "shlex",
+ "shlex 1.3.0",
  "syn 2.0.117",
 ]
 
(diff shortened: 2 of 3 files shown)


thepastaclaw:
✅ Review complete (commit 1488c57e)

Cargo.toml
```diff
@@ -59,6 +59,24 @@ key-wallet-manager = { git = "https://github.com/dashpay/rust-dashcore", rev = "
...
+lto = "fat"
+codegen-units = 1
+opt-level = 3
+
+# Debug counterpart for smaller iOS dev builds. Inherits the standard `dev`
+# profile (and is ONLY used by the iOS build (`build_ios.sh --profile dev`)
+[profile.dev-ios]
+inherits = "dev"
+lto = "thin"
+codegen-units = 1
```
thepastaclaw: **🟡 Suggestion: `codegen-units = 1` in `dev-ios` will tank incremental rebuilds**

`dev-ios` sets `codegen-units = 1`, which serializes all in-crate codegen onto a single unit and disables parallel codegen. For a workspace this size that's a major hit to incremental dev iteration — and `codegen-units = 1` is normally a *release* knob for size/perf, not a *dev* knob. The PR description motivates size reduction specifically for release; the dev-side justification isn't given. `lto = "thin"` is already pulling cross-unit optimization without the single-unit serialization penalty. Drop `codegen-units = 1` from `dev-ios` (inherit the dev default) unless there's a specific dev-build size constraint you can point to.

```suggestion
[profile.dev-ios]
inherits = "dev"
lto = "thin"
```

<sub>source: ['claude']</sub>

thepastaclaw: **🟡 Suggestion: Panic asymmetry: `dev-ios` inherits `panic = "unwind"`, `release-ios` is `"abort"`**

`release-ios` explicitly sets `panic = "abort"`, which is the right call for a staticlib whose symbols are called from Swift over a C ABI — any Rust panic terminates predictably without trying to unwind through `extern "C"`. `dev-ios` doesn't set it and therefore inherits `dev`'s `panic = "unwind"`. The compiler will still abort on unwind-through-extern-"C" so this isn't UB, but dev and release builds now exhibit different runtime behavior on a panic (forced-abort with drop glue / catch_unwind walk vs. immediate abort). For a staticlib where the Swift host can't catch a panic anyway, the simplest contract is to make dev match release. Add `panic = "abort"` to `dev-ios`, or comment why dev intentionally keeps unwind (e.g. test-harness `catch_unwind`).

```suggestion
[profile.dev-ios]
inherits = "dev"
panic = "abort"
lto = "thin"
```

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Build-config-only PR adding `release-ios`/`dev-ios` Cargo profiles and wiring `build_ios.sh` to use them. The profile contents are sensible for shrinking the iOS staticlib, but a few choices warrant a second look: `codegen-units = 1` in the dev profile, panic-handling asymmetry between dev-ios and release-ios, `strip = "symbols"` impact on Rust crash symbolication, and a large Cargo.lock dependency refresh unrelated to the stated goal.

<details>

```

In `Cargo.lock`:
- [SUGGESTION] Cargo.lock:1: Unrelated dependency upgrades in lockfile bloat this profile-only PR
  This PR only adds two Cargo profiles and tweaks `build_ios.sh`, neither of which should change dependency resolution. But Cargo.lock has 575+/188- lines: patch bumps across `autocfg`, `bitflags` (2.11.1 → 2.13.0), `cbindgen` (0.29.2 → 0.29.4), `cc`, `chrono`, `bumpalo`, `windows-sys`, and notably `reddsa` 0.5.1 → 0.5.2 (now pulling in `frost-core` + `frost-rerandomized`) plus the `bitcoin_hashes`/`bitcoin-io`/`base58ck` family. The dashcore git rev in Cargo.toml is unchanged, so this is a registry `cargo update`, not a transitive consequence. Refreshing crypto-adjacent deps inside a build-system chore broadens the review/test surface to every workspace build and slips consensus-adjacent crate changes past a reviewer who came here for an iOS profile. Revert the lockfile churn and land the dep refresh as its own PR.

In `Cargo.toml`:
- [SUGGESTION] Cargo.toml:75-78: `codegen-units = 1` in `dev-ios` will tank incremental rebuilds
  `dev-ios` sets `codegen-units = 1`, which serializes all in-crate codegen onto a single unit and disables parallel codegen. For a workspace this size that's a major hit to incremental dev iteration — and `codegen-units = 1` is normally a *release* knob for size/perf, not a *dev* knob. The PR description motivates size reduction specifically for release; the dev-side justification isn't given. `lto = "thin"` is already pulling cross-unit optimization without the single-unit serialization penalty. Drop `codegen-units = 1` from `dev-ios` (inherit the dev default) unless there's a specific dev-build size constraint you can point to.
- [SUGGESTION] Cargo.toml:73-78: Panic asymmetry: `dev-ios` inherits `panic = "unwind"`, `release-ios` is `"abort"`
  `release-ios` explicitly sets `panic = "abort"`, which is the right call for a staticlib whose symbols are called from Swift over a C ABI — any Rust panic terminates predictably without trying to unwind through `extern "C"`. `dev-ios` doesn't set it and therefore inherits `dev`'s `panic = "unwind"`. The compiler will still abort on unwind-through-extern-"C" so this isn't UB, but dev and release builds now exhibit different runtime behavior on a panic (forced-abort with drop glue / catch_unwind walk vs. immediate abort). For a staticlib where the Swift host can't catch a panic anyway, the simplest contract is to make dev match release. Add `panic = "abort"` to `dev-ios`, or comment why dev intentionally keeps unwind (e.g. test-harness `catch_unwind`).
```
</details>

Cargo.toml
```diff
@@ -59,6 +59,24 @@ key-wallet-manager = { git = "https://github.com/dashpay/rust-dashcore", rev = "
...
+# Size-tuned profile for the iOS `rs-unified-sdk-ffi` staticlib, which
+# otherwise ships huge. Inherits `release` and is ONLY used by the iOS
+# build (`build_ios.sh --profile release`)
+[profile.release-ios]
+inherits = "release"
+panic = "abort"
+strip = "symbols"
+lto = "fat"
+codegen-units = 1
+opt-level = 3
```
thepastaclaw, line 71: **💬 Nitpick: `strip = "symbols"` loses Rust symbols from production crash reports**

`strip = "symbols"` removes both DWARF and the symbol table from the staticlib, so production iOS crashes touching Rust frames (`dash_unified_sdk_*`, `dash_sdk_*`, `dash_core_sdk_*`) may surface as raw addresses with no demangled names — there is no separate dSYM being produced before the strip. That's a defensible size tradeoff, but it deserves either a comment explaining it or a downgrade to `strip = "debuginfo"`, which keeps the symbol table (cheap, ~bytes) while still dropping DWARF (most of the size). At minimum add a note so the next person knows symbolicated stacks require a different build.

```suggestion
# Size-tuned profile for the iOS `rs-unified-sdk-ffi` staticlib, which
# otherwise ships huge. Inherits `release` and is ONLY used by the iOS
# build (`build_ios.sh --profile release`).
#
# NOTE: `strip = "symbols"` removes the symbol table, so Rust frames in
# production crash reports will not symbolicate. Switch to
# `strip = "debuginfo"` (keeps function names) or `false` if symbolicated
# stack traces are needed.
[profile.release-ios]
inherits = "release"
panic = "abort"
strip = "symbols"
lto = "fat"
codegen-units = 1
opt-level = 3
```

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Carried-forward findings from prior review at 1488c57e: the Cargo.lock dependency churn is still present, and two documentation nitpicks (release-ios strip=symbols symbolication tradeoff, build_ios.sh --profile help text) remain valid. New in the latest delta: dev-ios profile now sets panic="abort" and removed codegen-units=1, resolving the prior panic-asymmetry and dev-iteration findings. No new defects introduced.

<details>

```

In `Cargo.lock`:
- [SUGGESTION] Cargo.lock:1: Unrelated dependency upgrades in lockfile bloat this profile-only PR
  This PR's stated change is adding two Cargo profiles (`release-ios`, `dev-ios`) and tweaking `build_ios.sh`; neither should change dependency resolution. The dashcore git revs in Cargo.toml (`3d0d5dcd…`) are unchanged, so the lockfile delta is a registry-side `cargo update`, not a transitive consequence of a source-rev bump. The churn includes broad patch bumps (`autocfg` 1.5.0→1.5.1, `base58ck` 0.1.0→0.1.100 dropping `bitcoin-internals`, `bitcoin_hashes`, `cbindgen`, `cc`, `chrono`) and crypto-adjacent updates (`reddsa` pulling in `frost-core`/`frost-rerandomized`). Refreshing crypto-adjacent deps inside a build-system chore broadens the review/test surface beyond iOS to every workspace build and slips changes past a reviewer who came here for an iOS profile. Revert the lockfile churn and land the dependency refresh as its own PR with separate validation.
```
</details>

QuantumExplorer (approved):

PR #3838, opened 2026-06-10
feat(swift-sdk): fund identity creation from shielded balance

QuantumExplorer:

The Create Identity screen in SwiftExampleApp could not use the shielded (Orchard) balance as a funding origin, even though the Type-20 `IdentityCreateFromShieldedPool` state transition and its full stack (rs-dpp builder → platform-wallet op → FFI → Swift wrapper) already exist. This PR surfaces that path in the UI and closes a persistence gap in the Rust flow that would have left the newly created identity invisible to the app.

## What was done?

**Rust (platform-wallet)**
- `identity_create_from_shielded_pool` (operations.rs) now captures the SDK broadcast's `StateTransitionProofResult` and extracts the proof-verified `Identity` from `VerifiedIdentityWithShieldedNullifiers`, back-filling `public_keys()` from the submitted set if the result echoes them empty (same defensive pattern as `register_from_addresses`), and returns it alongside the derived id. An unexpected proof-result variant is warn-and-synthesize rather than an error, since the broadcast already succeeded.
- `PlatformWallet::shielded_identity_create_from_pool` takes a new `identity_index: u32` and, after a successful broadcast, registers the verified identity in the local `IdentityManager` via `add_identity` + persister — mirroring `register_from_addresses` Step 3 — so the host persister emits the `IdentityChangeSet`/`IdentityKeysChangeSet` that creates the app's `PersistentIdentity` row. Previously no other layer did this, so the shielded path never produced an identity row. A registration failure after a successful broadcast is logged and swallowed (the identity exists on chain; sync heals).

**FFI / Swift SDK (marshal-only)**
- `platform_wallet_manager_shielded_identity_create_from_pool` gains `identity_index: u32` (after `account`); `shieldedIdentityCreateFromPool` gains the matching `identityIndex: UInt32` and passes it through.

**SwiftExampleApp (CreateIdentityView)**
- New `FundingSelection.shieldedBalance` option: "Shielded Balance — X DASH" appears in the Funding Source picker when the selected wallet is the one bound to `ShieldedService`, the pool balance is positive, and a fallback platform address exists.
- Type-20 spends a fixed denomination, so the free-form amount field is replaced by a denomination picker (0.1 / 0.3 / 0.5 / 1 DASH, mirrored from `shielded_identity_create_denominations` in rs-platform-version v8 with a source-of-truth doc pointer, filtered to denominations the pool covers). Footer explains the exact-equality fee model (fee metered FROM the denomination; excess returns as pool change).
- The REQUIRED `send_to_address_on_creation_failure` (21 raw `PlatformAddress` bytes) is derived from the wallet's lowest-index Platform Payment address — the same `(addressType, hash)` pairing the address-funded path feeds the FFI.
- Submission routes through the `RegistrationCoordinator`-hosted controller (like the Core-funded path) because the Halo 2 proof takes tens of seconds — the registration survives sheet dismissal and stays visible under Pending Registrations. The shared completion path patches the persisted row and marks the DIP-9 slot used.
- DashPay-keys footer made path-aware so the "+N duffs asset-lock minimum" phrasing isn't shown for the shielded path.

## How Has This Been Tested?

- `cargo check -p platform-wallet --all-features --all-targets` and `cargo check -p platform-wallet-ffi --all-features --all-targets` — pass.
- `cargo fmt --all --check` — clean.
- `./build_ios.sh --target sim` rebuilt the xcframework (regenerated header carries `uint32_t identity_index`) and SwiftExampleApp **BUILD SUCCEEDED** against it (iPhone 16 simulator).
- No end-to-end shielded run: strategy/full-block harnesses don't build shielded transitions yet, and exercising the flow needs a devnet with a funded shielded pool.

## Breaking Changes

None (not consensus-breaking). The FFI/Swift signature change to `shielded_identity_create_from_pool` is pre-release public-API churn.

## Checklist:

## Release Notes

* **New Features**
  * Shielded identity creation now allows explicit specification of which identity-registration slot to use (DIP-9 compliant).
  * Added denomination selection for shielded balance funding when creating identities.
  * Identity creation now returns comprehensive verification data for improved validation.

* **Refactor**
  * Enhanced identity creation operations to support explicit slot assignment across wallet implementations.

### DIFF
--- a/packages/rs-platform-wallet-ffi/src/shielded_send.rs
+++ b/packages/rs-platform-wallet-ffi/src/shielded_send.rs
@@ -339,6 +339,11 @@ pub unsafe extern "C" fn platform_wallet_manager_shielded_withdraw(
 /// denomination + full key set) + a per-key proof-of-possession produced via
 /// `signer_identity_handle`. There is NO platform identity signature.
 ///
+/// `identity_index` is the DIP-9 identity-registration slot the new identity occupies. On a
+/// successful broadcast the wallet registers the proof-verified identity at this slot in its local
+/// `IdentityManager` (mirroring address-funded registration), which drives the host persister's
+/// identity-row emit. It carries no decision here — it is marshalled straight through to the wallet.
+///
 /// On success the 32-byte new identity id (`double_sha256(sorted nullifiers)`) is written to
 /// `out_identity_id`. The id is deterministic in the spent notes, so the host can also predict it
 /// independently if needed.
@@ -367,6 +372,7 @@ pub unsafe extern "C" fn platform_wallet_manager_shielded_identity_create_from_p
     handle: Handle,
     wallet_id_bytes: *const u8,
     account: u32,
+    identity_index: u32,
     identity_pubkeys: *const IdentityPubkeyFFI,
     identity_pubkeys_count: usize,
     denomination: u64,
@@ -444,6 +450,7 @@ pub unsafe extern "C" fn platform_wallet_manager_shielded_identity_create_from_p
             .shielded_identity_create_from_pool(
                 &coordinator,
                 account,
+                identity_index,
                 public_keys,
                 denomination,
                 send_to_address_on_creation_failure,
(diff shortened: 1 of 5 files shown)


QuantumExplorer (commented):
Reviewed

PR #3839, opened 2026-06-10
fix(platform): encrypt shielded outputs to the sender's outgoing viewing key

QuantumExplorer:

Every shielded builder passed `ovk: None` to the Orchard bundle builder, so each real output's `out_ciphertext` was encrypted to a per-output random cipher key. As a result, `try_recover_outgoing_note` — the wallet scanner's outgoing-payment recovery path — could never decrypt the wallet's own sends, and the outgoing-payments store (surfaced in the app as sent-payment history with memos) never populated, even though the sync path in `sync.rs` already tried every account's OVK and persisted hits.

## What was done?

**rs-dpp builders** (client-side only — `out_ciphertext` is opaque to consensus validation, so no protocol impact):

- `build_output_only_bundle` takes a new `sender_ovk: Option<OutgoingViewingKey>` parameter, exposed as a new parameter on `build_shield_transition` and both `build_shield_from_asset_lock_transition` variants (these signatures carry no FVK to derive it from).
- Spend-side builders (`build_spend_bundle` — used by unshield, withdrawal, identity-create — and the inline transfer builder) derive `fvk.to_ovk(Scope::External)` internally, so transfer recipient outputs **and** change outputs are recoverable with no public-API change on the spend side.
- Orchard's dummy/padding outputs keep random cipher keys, as before.

**rs-platform-wallet wiring:**

- `operations::shield` passes the account's `outgoing_viewing_key`.
- `shielded_fund_from_asset_lock` resolves the OVK from the bound account whose IVK recognizes the recipient address (falling back to the lowest bound account), on both the normal and the InstantLock→ChainLock retry broadcast paths.

## How Has This Been Tested?

- New dpp bundle-level test `output_built_with_sender_ovk_recovers_under_that_ovk_only`: exactly one action recovers (note, recipient, memo) under the sender's OVK via `try_output_recovery_with_ovk`, and a foreign OVK opens nothing.
- New `ovk_builder_roundtrip_tests::shield_built_note_ovk_recovers_and_persists_as_outgoing` in rs-platform-wallet — the sender-side mirror of `shield_decrypt_tests`: builds a real Shield transition with the wallet's OVK, reassembles each action into the scanner's wire form, asserts `try_recover_outgoing_note` recovers exactly the recipient output with matching recipient/value/memo (this returned `None` before the fix), and persists it end-to-end via `record_outgoing_note`.
- `cargo test -p dpp --all-features shielded` → 172 passed.
- `cargo test -p platform-wallet --features shielded --release` → all green (lib: 218 passed).
- `cargo fmt --all` clean. drive-abci shielded-transfer tests use unchanged signatures and assert only ciphertext sizes, so they are unaffected.

## Breaking Changes

None at the protocol level. `build_shield_transition` and `build_shield_from_asset_lock_transition(_with_signer)` gained a `sender_ovk` parameter (pre-release client API; all in-repo call sites updated).

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added optional outgoing viewing key (OVK) support for shielded outputs so senders can enable recoverable sender-side data (recipient, amount, memo).

* **Tests**
  * Added end-to-end OVK roundtrip tests and updated unit tests to validate OVK recoverability and non-recoverability with mismatched keys.

* **Documentation**
  * Updated examples and docs to show passing the sender OVK.

### DIFF
--- a/packages/rs-dpp/src/shielded/builder/mod.rs
+++ b/packages/rs-dpp/src/shielded/builder/mod.rs
@@ -20,11 +20,12 @@
 //!     &fvk.address_at(0, Scope::External).to_raw_address_bytes(),
 //! );
 //!
-//! // Build a shield transition
+//! // Build a shield transition; pass the sender's OVK so the wallet can
+//! // later recover its own send from chain data (None = unrecoverable)
 //! let pk = ProvingKey::build();
 //! let st = build_shield_transition(
 //!     &recipient, shield_amount, inputs, fee_strategy,
-//!     &signer, 0, &pk, [0u8; 36], platform_version,
+//!     &signer, 0, &pk, [0u8; 36], Some(fvk.to_ovk(Scope::External)), platform_version,
 //! )?;
 //! ```
 
--- a/packages/rs-dpp/src/shielded/builder/mod.rs
+++ b/packages/rs-dpp/src/shielded/builder/mod.rs
@@ -48,7 +48,8 @@ pub use unshield::build_unshield_transition;
 
 use grovedb_commitment_tree::{
     Anchor, Authorized, Builder, Bundle, BundleType, DashMemo, Flags as OrchardFlags,
-    FullViewingKey, MerklePath, Note, NoteValue, PaymentAddress, ProvingKey, SpendAuthorizingKey,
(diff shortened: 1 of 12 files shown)


QuantumExplorer:
Reviewed

PR #806, opened 2023-02-23
feat(wasm dpp): validate state transition key signature

markin-io:
## Issue being fixed or feature implemented

- Refactored validate_state_transition_key_signature to use identity balance instead of identity
- Implemented wasm binding and tests

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added or updated relevant unit/integration/functional/e2e tests
- [ ] I have made corresponding changes to the documentation

**For repository code-owners and collaborators only**
- [x] I have assigned this pull request to a milestone

PR #3842, opened 2026-06-10
feat(swift-sdk): add BIP39 word-validation helpers for the recover flow

llbartekll:
## What

Ports DashSync's `DSBIP39Mnemonic` recover-flow helpers into the platform-wallet FFI so **dashwallet-ios** can drop `DSBIP39Mnemonic` from its wallet-recovery model (part of the ongoing DashSync → SwiftDashSDK migration).

New module `packages/rs-platform-wallet-ffi/src/mnemonic_words.rs` exposes four C functions:

| FFI | Swift wrapper | Purpose |
|---|---|---|
| `platform_wallet_mnemonic_word_is_valid` | `Mnemonic.wordIsValid` | word ∈ any of the 7 bundled BIP39 languages |
| `platform_wallet_mnemonic_word_is_local` | `Mnemonic.wordIsLocal` | word ∈ English wordlist |
| `platform_wallet_mnemonic_normalize_phrase` | `Mnemonic.normalizePhrase` | NFKD + lowercase + whitespace-collapse |
| `platform_wallet_mnemonic_cleanup_phrase` | `Mnemonic.cleanupPhrase` | minimal input cleanup + CJK ideographic auto-split |

## Parity with DashSync

Faithful port of `DSBIP39Mnemonic.m`:

- NFKD + lowercase before every wordlist lookup (DashSync normalizes the phrase, not the per-word lookup), so callers normalize first and membership is exact-byte.
- Exact-byte wordlist membership over the same **7** languages DashSync bundles (en, fr, es, it, ja, ko, zh-Hans) — not all 10 bip39 languages.
- Per-language checksum validation loop, never bip39 autodetect (`language_of` returns `AmbiguousLanguages` and diverges).
- Combining marks (e.g. U+3099 voiced kana, Latin diacritics) preserved in cleanup to match `letterCharacterSet`'s L\*∪M\*.
- `cleanupPhrase`'s CJK ideographic (U+3000) auto-split, including DashSync's global-replace semantics.

**One intentional, inert divergence** (documented in-module): `parse_in_normalized` enforces BIP39's ≥12-word floor, so 3/6/9-word phrases that DashSync's lenient checksum loop would accept are rejected. This only gates `cleanup_phrase`'s early-return; ASCII output is byte-identical, and the recover path enforces a 12-word minimum anyway.

## Tests

**18 unit tests** (`cargo test -p platform-wallet-ffi mnemonic_words`), `cargo clippy` clean:

- English Trezor vectors (12- and 24-word) validate; messy/mixed-case input normalizes
- A valid phrase per bundled non-English language validates through the per-language loop
- NFC/NFKD normalization contract: raw uppercase / NFC forms are non-members; both converge to the same valid member under `nfkd_lower`
- CJK auto-split for Japanese and Chinese no-space phrases; punctuation-strip + split combined; repeated-word wrap-all terminates
- FFI round-trip for both string functions including `into_raw`/free and NULL-pointer handling

## Consumer

dashwallet-ios branch `feature/mnemonic-word-validation` switches the recover-flow call sites to `Mnemonic.{wordIsValid,wordIsLocal,normalizePhrase,cleanupPhrase}` and removes the `DSBIP39Mnemonic`-backed bodies. Only remaining `DSBIP39Mnemonic` use there is phrase-repair, which has no SDK equivalent.

## Summary by CodeRabbit

* **New Features**
  * Multilingual mnemonic support: word validation across bundled languages and language-specific checks.
  * Phrase normalization (NFKD, lowercasing, whitespace collapse) and cleanup with CJK-aware auto-splitting for improved wallet recovery input handling.
  * Public SDK helpers to validate words and normalize/cleanup phrases via platform FFI.

* **Chores**
  * Updated workspace dependencies to a newer upstream commit.


thepastaclaw:
✅ Review complete (commit 2a3f18ae)

ZocoLini (changes requested):
key-wallet already has logic related to mnemonic Bip39 'key-wallet/src/mnemonic.rs', I don't think we should re-implement it. If there is any logic missing I would rather prefer to see a PR in key-wallet than in here, so other dependencies can use the new code. The think is, how should we use it. I can think of two options:

 - Make the wrappers you need in rs-platform-wallet-ffi around key-wallet methods
 - use the KeyWallet module in the SwiftSDK that already wraps the key-wallet module

The idea was to make rs-platform-wallet-ffi a facade for everything needed by the so in theory that option aligns better with @QuantumExplorer vision

llbartekll:
Good call. Moved the generic helpers into key-wallet — dashpay/rust-dashcore#806 adds `normalize_phrase` + `is_word_in_language`. This PR now just wraps those and drops the direct `bip39` dep (it's only transitive via key-wallet now).

I kept the app-specific bits here — the bundled 7-language set and the CJK auto-split in `cleanupPhrase` — since they're not really general wallet logic. Didn't expose them in key-wallet-ffi because we bind key-wallet's Rust API directly here, but happy to add that if you'd rather have them reusable over FFI.

#806 is a draft for now; I'll re-pin to the merge commit once it lands.

packages/rs-platform-wallet-ffi/src/mnemonic_words.rs
```diff
@@ -0,0 +1,599 @@
...
+                    // incl. U+3000 (which `str::trim` also treats as ws).
+                    s = s.trim().to_string();
+                    i += j - 1; // outer `i += 1` advances past the match
+                    break;
+                }
+                j -= 1;
+            }
+            i += 1;
+        }
+    }
```
thepastaclaw: **🔴 Blocking: CJK auto-split fails on precomposed (NFC) Japanese input**

In `cleanup_phrase_impl`, `normalized` is NFKD/lowercase but `s` keeps the original Unicode form — the filter on lines 116-118 only drops non-letter/mark/whitespace, it does not normalize. The bip39 Japanese wordlist is NFKD (e.g. `ご` = `こ` U+3053 + U+3099), and iOS Japanese IMEs typically emit NFC (`ご` = U+3054). On NFC input the loop extracts NFKD candidates from `normalized`, `word_in_any_list(&candidate)` succeeds, but `s.replace(&candidate, &wrapped)` searches the NFC `s` and finds nothing — so `cleanupPhrase` returns the original unsplit no-space phrase and the recover flow still sees one invalid word. The existing CJK tests use `m.words()` from the bip39 crate, which is already NFKD, so they cannot expose this. Fix by either rewriting the NFKD `normalized` form into `s` before the loop, or by replacing against an NFKD'd working buffer and re-NFKD'ing matches found in `s`. Add an explicit NFC-Japanese fixture (e.g. apply `.nfc()` to `m.words()` before concatenating) to the test suite.

<sub>source: ['codex']</sub>

packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/Mnemonic.swift
```diff
@@ -149,4 +149,53 @@ public class Mnemonic {
...
+                try platform_wallet_mnemonic_cleanup_phrase(cPhrase, &outPtr).check()
+            }
+        } catch {
+            return phrase
+        }
+        guard let cStr = outPtr else { return phrase }
+        let result = String(cString: cStr)
+        platform_wallet_free_string(cStr)
+        return result
+    }
```
thepastaclaw: **🟡 Suggestion: Reject embedded NULs before crossing the C boundary**

`withCString` (and Rust's `CStr::from_ptr`) terminate at the first U+0000 byte, so a Swift `String` containing an embedded NUL is silently truncated to its prefix. `wordIsValid("abandon\0notaword")` therefore returns `true`, and `normalizePhrase` / `cleanupPhrase` return a transformation of only the prefix. This is not memory-unsafe but it means the recover-flow helpers can validate or display a different phrase than the caller passed in. `Mnemonic.toSeed(mnemonicUTF8Bytes:)` already treats embedded NUL as invalid; the new helpers should do the same (guard `!s.utf8.contains(0)` early and treat as invalid / passthrough).

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/mnemonic_words.rs
```diff
@@ -0,0 +1,599 @@
...
+
+    s
+}
```
thepastaclaw: **🟡 Suggestion: Cap input length in cleanup/normalize before the CJK loop**

`platform_wallet_mnemonic_cleanup_phrase` and `_normalize_phrase` accept arbitrary FFI input. The CJK auto-split walks every char of every word with `j ≤ 8`, performs full-string `String::replace` per hit, then runs a `while contains("  ")` collapse on each replacement. A long no-space run of valid single-ideograph Chinese words reaches the loop (each `wchars[0] >= U+3000`, not a whole valid phrase) and forces superlinear allocations on the caller thread. Other mnemonic entry points cap material at 1024 bytes; these new helpers should enforce a similar word-count / byte-length floor before the CJK loop fires (or return the input verbatim past the cap) to keep a malicious paste from hanging the wallet recover UI.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/Cargo.toml
```diff
@@ -55,6 +55,14 @@ bs58 = "0.5"
 # Zeroize intermediate key material crossing the FFI boundary.
 zeroize = { version = "1", features = ["derive"] }
 
+# BIP39 wordlist access for the recover-flow word-validation FFI
+# (parity with DashSync DSBIP39Mnemonic). Mirror key-wallet's language
+# feature set so every bundled-language variant is available.
+bip39 = { version = "2.2.0", default-features = false, features = ["std", "chinese-simplified", "chinese-traditional", "czech", "french", "italian", "japanese", "korean", "portuguese", "spanish"] }
```
thepastaclaw: **💬 Nitpick: Bip39 enables 3 language features that are never used**

`LANGS` in `mnemonic_words.rs` is fixed at 7 (en, fr, es, it, ja, ko, zh-Hans). The `czech`, `portuguese`, and `chinese-traditional` bip39 features pull in wordlists that nothing in this crate references. The comment justifies it as mirroring `key-wallet`, but if `key-wallet` does not actually require those, trimming would shrink the iOS framework. Either drop the unused features or expand the comment to point at the specific `key-wallet` site these are mirroring.

<sub>source: ['claude']</sub>

packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/Mnemonic.swift
thepastaclaw, line 210: **💬 Nitpick: normalizePhrase / cleanupPhrase silently swallow FFI errors**

On UTF-8 conversion errors, NULL out-pointer, or allocation failure the Swift wrappers return the input unchanged with no log line. The behavior matches DashSync's `NSString` assumption, but if the FFI ever starts returning meaningful error codes here, debugging on the iOS side will be much harder. At minimum log the error path via `os_log`/`print`; alternatively make these `throws` and force the recover-flow call sites to make the fallback decision explicitly.

<sub>source: ['claude']</sub>

thepastaclaw (changes requested):
## Code Review

One blocking correctness bug found by Codex: the CJK auto-split path compares NFKD candidates against the unnormalized input string, so precomposed (NFC) Japanese input from iOS IMEs will not be auto-split. Otherwise the FFI surface is well-formed, defensively coded, and architecturally aligned with the swift-sdk rules. A few low-severity polish suggestions remain.

<details>

```

In `packages/rs-platform-wallet-ffi/src/mnemonic_words.rs`:
- [BLOCKING] packages/rs-platform-wallet-ffi/src/mnemonic_words.rs:134-174: CJK auto-split fails on precomposed (NFC) Japanese input
  In `cleanup_phrase_impl`, `normalized` is NFKD/lowercase but `s` keeps the original Unicode form — the filter on lines 116-118 only drops non-letter/mark/whitespace, it does not normalize. The bip39 Japanese wordlist is NFKD (e.g. `ご` = `こ` U+3053 + U+3099), and iOS Japanese IMEs typically emit NFC (`ご` = U+3054). On NFC input the loop extracts NFKD candidates from `normalized`, `word_in_any_list(&candidate)` succeeds, but `s.replace(&candidate, &wrapped)` searches the NFC `s` and finds nothing — so `cleanupPhrase` returns the original unsplit no-space phrase and the recover flow still sees one invalid word. The existing CJK tests use `m.words()` from the bip39 crate, which is already NFKD, so they cannot expose this. Fix by either rewriting the NFKD `normalized` form into `s` before the loop, or by replacing against an NFKD'd working buffer and re-NFKD'ing matches found in `s`. Add an explicit NFC-Japanese fixture (e.g. apply `.nfc()` to `m.words()` before concatenating) to the test suite.
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/mnemonic_words.rs:110-177: Cap input length in cleanup/normalize before the CJK loop
  `platform_wallet_mnemonic_cleanup_phrase` and `_normalize_phrase` accept arbitrary FFI input. The CJK auto-split walks every char of every word with `j ≤ 8`, performs full-string `String::replace` per hit, then runs a `while contains("  ")` collapse on each replacement. A long no-space run of valid single-ideograph Chinese words reaches the loop (each `wchars[0] >= U+3000`, not a whole valid phrase) and forces superlinear allocations on the caller thread. Other mnemonic entry points cap material at 1024 bytes; these new helpers should enforce a similar word-count / byte-length floor before the CJK loop fires (or return the input verbatim past the cap) to keep a malicious paste from hanging the wallet recover UI.

In `packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/Mnemonic.swift`:
- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/Mnemonic.swift:158-200: Reject embedded NULs before crossing the C boundary
  `withCString` (and Rust's `CStr::from_ptr`) terminate at the first U+0000 byte, so a Swift `String` containing an embedded NUL is silently truncated to its prefix. `wordIsValid("abandon\0notaword")` therefore returns `true`, and `normalizePhrase` / `cleanupPhrase` return a transformation of only the prefix. This is not memory-unsafe but it means the recover-flow helpers can validate or display a different phrase than the caller passed in. `Mnemonic.toSeed(mnemonicUTF8Bytes:)` already treats embedded NUL as invalid; the new helpers should do the same (guard `!s.utf8.contains(0)` early and treat as invalid / passthrough).
```
</details>

packages/rs-platform-wallet-ffi/src/mnemonic_words.rs
```diff
@@ -0,0 +1,600 @@
...
+    }
+}
+
+/// `true` if `word` is in the default (English) wordlist (DashSync
+/// `wordIsLocal:`). NULL / invalid-UTF-8 input returns `false`.
+///
+/// # Safety
+/// `word` must be NULL or a valid null-terminated UTF-8 C string.
+#[no_mangle]
+pub unsafe extern "C" fn platform_wallet_mnemonic_word_is_local(word: *const c_char) -> bool {
```
ZocoLini: This function doesn't make a lot of sense, only the app knows what is local for the user, and I think it would be more helpful to expose a method that wraps 'is_word_in_language' (I cannot ind a wrapper for it anywhere)

packages/rs-platform-wallet-ffi/src/mnemonic_words.rs
```diff
@@ -0,0 +1,600 @@
...
+                    break;
+                }
+                j -= 1;
+            }
+            i += 1;
+        }
+    }
+
+    s
+}
```
thepastaclaw: **🔴 Blocking: CJK auto-split still mismatches NFKD candidates against an unnormalized buffer (NFC Japanese input fails)**

`cleanup_phrase_impl` builds `s` (lines 110-113) from `phrase.chars().filter(...)`, which preserves the caller's Unicode form — NFC stays NFC. `normalized = normalize_phrase_impl(&s)` (line 128) goes through `key_wallet::Mnemonic::normalize_phrase`, which is NFKD + lowercase. The CJK loop (lines 137-167) iterates `normalized.split(' ')`, extracts `candidate` from NFKD scalar slices, validates it via `word_in_any_list` against the canonical (NFKD) BIP-39 lists, then on line 154 runs `s.replace(&candidate, &wrapped)` against the still-NFC `s`. `String::replace` is exact-byte: NFKD candidate bytes never match NFC `s` bytes and nothing gets wrapped.

iOS Japanese IMEs typically emit NFC (e.g. `ご` = U+3054 as one scalar), while the bip39 Japanese wordlist is NFKD (`こ` U+3053 + U+3099). For a pasted/QR/clipboard no-space NFC Japanese phrase, `platform_wallet_mnemonic_cleanup_phrase` returns the original unsplit phrase and the recover UI surfaces a single giant invalid 'word'. Existing `cjk_passthrough_and_autosplit` / `chinese_no_space_autosplit` tests build fixtures from `key_wallet::Mnemonic::from_entropy(..., L::Japanese).phrase()`, which already emits NFKD, so they cannot expose this.

Fix: either (a) NFKD-normalize `s` after the filter pass so the replace can hit (matches DashSync's normalize-before-split behavior on combining-mark input), or (b) build a parallel NFKD working buffer, run replacements against it, and emit NFKD bytes only when CJK splitting fires. Add an explicit NFC-Japanese fixture (`key_wallet::Mnemonic::from_entropy(.., L::Japanese).unwrap().phrase().nfc().collect::<String>()`) to lock the regression.

packages/rs-platform-wallet-ffi/src/mnemonic_words.rs
```diff
@@ -0,0 +1,600 @@
...
+) -> PlatformWalletFFIResult {
+    check_ptr!(phrase);
+    check_ptr!(out_string);
+    *out_string = std::ptr::null_mut();
+
+    let p = unwrap_result_or_return!(CStr::from_ptr(phrase).to_str());
+    let cleaned = cleanup_phrase_impl(p);
+    let c = unwrap_result_or_return!(CString::new(cleaned));
+    *out_string = c.into_raw();
+    PlatformWalletFFIResult::ok()
```
thepastaclaw, line 127: **🟡 Suggestion: Cap mnemonic helper input before the CJK splitting loop**

`platform_wallet_mnemonic_cleanup_phrase` and `platform_wallet_mnemonic_normalize_phrase` accept arbitrary FFI input with no length cap. `cleanup_phrase_impl` then walks every char of every word with `j ≤ 8`, performs full-string `String::replace` per hit (line 154), and runs `while s.contains(&dbl_ideo)` collapses (lines 155-157) — superlinear in input length. A long no-space run of valid single-ideograph CJK characters (each `wchars[0] >= U+3000`, even if not a valid global phrase) reaches the inner loop and forces large allocations on the iOS recover-UI thread, i.e. an attacker-triggerable local DoS via a crafted clipboard/QR paste. Other mnemonic entry points in this crate cap material at a fixed byte budget; enforce a similar word-count or byte cap (or return input verbatim past the cap) before normalization/cleanup runs.

thepastaclaw (changes requested):
## Code Review

Incremental refactor cleanly delegates BIP-39 primitives to `key-wallet` and removes the direct `bip39` dependency, but the previously-flagged blocking NFC/NFKD mismatch in `cleanup_phrase_impl`'s CJK auto-split is unchanged: NFKD candidates from `normalized` are still searched against the unnormalized buffer `s`, so NFC Japanese input from iOS IMEs still fails to auto-split. Embedded-NUL truncation in the Swift wrappers and unbounded FFI input also remain. All four reviewers converge on the same blocking issue and the same three lower-severity follow-ups.

<details>

```

In `packages/rs-platform-wallet-ffi/src/mnemonic_words.rs`:
- [BLOCKING] packages/rs-platform-wallet-ffi/src/mnemonic_words.rs:104-171: CJK auto-split still mismatches NFKD candidates against an unnormalized buffer (NFC Japanese input fails)
  `cleanup_phrase_impl` builds `s` (lines 110-113) from `phrase.chars().filter(...)`, which preserves the caller's Unicode form — NFC stays NFC. `normalized = normalize_phrase_impl(&s)` (line 128) goes through `key_wallet::Mnemonic::normalize_phrase`, which is NFKD + lowercase. The CJK loop (lines 137-167) iterates `normalized.split(' ')`, extracts `candidate` from NFKD scalar slices, validates it via `word_in_any_list` against the canonical (NFKD) BIP-39 lists, then on line 154 runs `s.replace(&candidate, &wrapped)` against the still-NFC `s`. `String::replace` is exact-byte: NFKD candidate bytes never match NFC `s` bytes and nothing gets wrapped.

  iOS Japanese IMEs typically emit NFC (e.g. `ご` = U+3054 as one scalar), while the bip39 Japanese wordlist is NFKD (`こ` U+3053 + U+3099). For a pasted/QR/clipboard no-space NFC Japanese phrase, `platform_wallet_mnemonic_cleanup_phrase` returns the original unsplit phrase and the recover UI surfaces a single giant invalid 'word'. Existing `cjk_passthrough_and_autosplit` / `chinese_no_space_autosplit` tests build fixtures from `key_wallet::Mnemonic::from_entropy(..., L::Japanese).phrase()`, which already emits NFKD, so they cannot expose this.

  Fix: either (a) NFKD-normalize `s` after the filter pass so the replace can hit (matches DashSync's normalize-before-split behavior on combining-mark input), or (b) build a parallel NFKD working buffer, run replacements against it, and emit NFKD bytes only when CJK splitting fires. Add an explicit NFC-Japanese fixture (`key_wallet::Mnemonic::from_entropy(.., L::Japanese).unwrap().phrase().nfc().collect::<String>()`) to lock the regression.
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/mnemonic_words.rs:218-253: Cap mnemonic helper input before the CJK splitting loop
  `platform_wallet_mnemonic_cleanup_phrase` and `platform_wallet_mnemonic_normalize_phrase` accept arbitrary FFI input with no length cap. `cleanup_phrase_impl` then walks every char of every word with `j ≤ 8`, performs full-string `String::replace` per hit (line 154), and runs `while s.contains(&dbl_ideo)` collapses (lines 155-157) — superlinear in input length. A long no-space run of valid single-ideograph CJK characters (each `wchars[0] >= U+3000`, even if not a valid global phrase) reaches the inner loop and forces large allocations on the iOS recover-UI thread, i.e. an attacker-triggerable local DoS via a crafted clipboard/QR paste. Other mnemonic entry points in this crate cap material at a fixed byte budget; enforce a similar word-count or byte cap (or return input verbatim past the cap) before normalization/cleanup runs.
```
</details>

packages/rs-platform-wallet-ffi/src/mnemonic_words.rs
thepastaclaw: Resolved in `2e9d168` — _CJK auto-split fails on precomposed (NFC) Japanese input_ no longer present.

packages/rs-platform-wallet-ffi/Cargo.toml
thepastaclaw: Resolved in `ca530bb` — _Bip39 enables 3 language features that are never used_ no longer present.

packages/rs-platform-wallet-ffi/src/mnemonic_words.rs
thepastaclaw: Resolved in `2e9d168` — _Cap input length in cleanup/normalize before the CJK loop_ no longer present.

packages/rs-platform-wallet-ffi/src/mnemonic_words.rs
```diff
@@ -0,0 +1,600 @@
...
+            assert!(word_in_any_list(&w), "{lang:?} word should be valid");
+        }
+        assert!(word_in_english(&first_word(L::English)));
+        assert!(!word_in_english(&first_word(L::Japanese)));
+        assert!(!word_in_english(&first_word(L::ChineseSimplified)));
+    }
+
+    #[test]
+    fn valid_phrase_per_bundled_language() {
+        // A real 12-word phrase in each non-English bundled language validates
```
ZocoLini: This feels more like something that should be in key_wallet, since it is testing key_wallet behavior

thepastaclaw:
Coverage correction for the automated review at ca530bb1: the verifier used completed outputs from `claude-general`, `codex-general` (JSON emitted despite nonzero tool exit), `codex-security-auditor`, `claude-ffi-engineer`, and `codex-ffi-engineer`. `claude-security-auditor` stalled before producing JSON and was excluded. CodeRabbit provided no inline findings for this SHA.

packages/rs-platform-wallet-ffi/src/mnemonic_words.rs
```diff
@@ -0,0 +1,600 @@
...
+
+/// `wordIsLocal:` (m:278-281) — member of the default (English) wordlist.
+fn word_in_english(w: &str) -> bool {
+    key_wallet::Mnemonic::is_word_in_language(w, L::English)
+}
+
+/// `normalizePhrase:` (m:387-408) — NFKD + lowercase + whitespace-collapse.
+/// Delegates to key-wallet's lenient input normalization (byte-identical to
+/// the prior in-crate impl).
+fn normalize_phrase_impl(input: &str) -> String {
```
ZocoLini: This wrappers is an overkill

packages/rs-platform-wallet-ffi/src/mnemonic_words.rs
```diff
@@ -0,0 +1,600 @@
...
+            assert!(
+                phrase_is_valid_impl(&normalize_phrase_impl(&phrase)),
+                "{lang:?} phrase should validate"
+            );
+        }
+    }
+
+    #[test]
+    fn normalize_collapses_unicode_forms_like_dashsync() {
+        // Mirrors DSBIP39Tests.m:290-314 (a Czech phrase in NFKD/NFC/NFD all
```
ZocoLini: This tests is basically testing key_wallet logic

packages/rs-platform-wallet-ffi/src/mnemonic_words.rs
```diff
@@ -0,0 +1,600 @@
...
+            .collect();
+        assert!(
+            cleanup_phrase_impl(&nospace).contains(IDEO_SP),
+            "no-space Chinese phrase should get ideographic spaces"
+        );
+    }
+
+    #[test]
+    fn english_trezor_vectors_validate() {
+        // Canonical English BIP-39 vectors from DashSync's own test suite
```
ZocoLini: This is also testing key_wallet logic hidden behind wrappers right??

packages/rs-platform-wallet-ffi/src/mnemonic_words.rs
```diff
@@ -0,0 +1,600 @@
...
+}
+
+/// `cleanupPhrase:` (m:325-384) — minimal cleanup for display/editing, plus
+/// CJK ideographic auto-splitting. Returns the pre-normalize cleaned string.
+///
+/// Index note: DashSync indexes with UTF-16 units (`characterAtIndex:`,
+/// `substringWithRange:`); we use Unicode scalars (`char`). Every BIP-39 CJK
+/// word is in the BMP (1 char = 1 UTF-16 unit), so the two indexings agree.
+fn cleanup_phrase_impl(phrase: &str) -> String {
+    // (1) remove chars not in (letter ∪ mark ∪ whitespace). DashSync uses
```
ZocoLini: This doesn't belong to the FFI layer, any logic should be in a non-FFI crate or in your side, where you use the SDK

ZocoLini (changes requested):
I mentioned a couple of things but in general, FFI crates should have no logic other than converting C types to Rust, calling Rust functions, and converting Rust types to C. 

key_wallet should have the basic functions you need, the ffi crate that wraps it should only provide the export "C" functions, and the Swift consumer should write the logic it needs around the available functionalities and exported types

packages/rs-platform-wallet-ffi/src/mnemonic_words.rs
llbartekll: Fixed in 7176d824da. The CJK split now NFKD-normalizes the working buffer before `String::replace`, so precomposed (NFC) Japanese input from iOS IMEs auto-splits correctly. Added `nfc_japanese_no_space_autosplit`, which asserts an NFC no-space phrase splits byte-identically to the equivalent NFKD input — the prior CJK fixtures were built from `from_entropy().phrase()` (already NFKD), so they could not catch this.

packages/rs-platform-wallet-ffi/src/mnemonic_words.rs
llbartekll: Fixed in 7176d824da. The CJK split now NFKD-normalizes the working buffer before `String::replace`, so precomposed (NFC) Japanese input from iOS IMEs auto-splits correctly. Added `nfc_japanese_no_space_autosplit`, which asserts an NFC no-space phrase splits byte-identically to the equivalent NFKD input — the prior CJK fixtures were built from `from_entropy().phrase()` (already NFKD), so they could not catch this.

packages/rs-platform-wallet-ffi/src/mnemonic_words.rs
llbartekll: Not addressing in this PR. The recover UI bounds phrase length upstream (DW_PHRASE_MIN/MAX_LENGTH) and the seed field is a fixed text input, so the unbounded-paste path is not reachable from the app — a megabyte paste would only self-DoS the user pasting it. The cleanup/normalize logic is also slated to move out of this FFI crate per the architecture review below, where input bounds will be reconsidered. Resolving for now.

packages/rs-platform-wallet-ffi/src/mnemonic_words.rs
llbartekll: Not addressing in this PR. The recover UI bounds phrase length upstream (DW_PHRASE_MIN/MAX_LENGTH) and the seed field is a fixed text input, so the unbounded-paste path is not reachable from the app — a megabyte paste would only self-DoS the user pasting it. The cleanup/normalize logic is also slated to move out of this FFI crate per the architecture review below, where input bounds will be reconsidered. Resolving for now.

packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/Mnemonic.swift
llbartekll: Not addressing in this PR — mirrors DashSync NSString behavior, and the recover-flow input is a UITextField value that cannot carry an embedded NUL. Resolving.

llbartekll: Intentional — these mirror DashSync non-throwing NSString helpers: on any FFI error the recover flow wants a graceful passthrough of the original phrase, not a thrown error. Resolving.

packages/rs-platform-wallet-ffi/Cargo.toml
llbartekll: Obsolete — this PR dropped the direct bip39 dependency entirely (ca530bb); the wordlist, normalize, and validate logic now come from key-wallet. Resolving.

thepastaclaw (commented):
## Code Review

Prior blocking NFC/NFKD CJK auto-split bug is fixed at mnemonic_words.rs:146 with a targeted regression test (lines 358-400). Three lower-severity FFI hygiene findings from prior reviews carry forward against current HEAD: embedded-NUL truncation in Swift wrappers, no input-length cap on the FFI cleanup helpers, and silent error swallow in normalizePhrase/cleanupPhrase.

### 3 additional finding(s)

<details>
<summary><b>suggestion:</b> Reject embedded NULs before crossing the C boundary</summary>

`packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/Mnemonic.swift` (line 158)

`wordIsValid` (158-160), `wordIsLocal` (164-166), `normalizePhrase` (171-184), and `cleanupPhrase` (187-200) all pass the input directly to `withCString`, and the Rust side reads it back with `CStr::from_ptr`. Both terminate at the first U+0000 byte, so a Swift `String` containing an embedded NUL is silently truncated at the FFI boundary: `wordIsValid("abandon\0notaword")` returns `true`, and the string helpers transform only the prefix while the recover UI still displays the full token. This is exploitable by anything that can place a NUL into recovery input (malicious QR code, clipboard from a hostile page, universal-link parameter): a phrase the validator never actually checked can be accepted, or the cleaned text shown back to the user can disagree with the original paste. `Mnemonic.toSeed(mnemonicUTF8Bytes:)` already guards against this at lines 83-86 — mirror that behavior: early-return `false` from the bool helpers when `word.utf8.contains(0)`, and return `phrase` unchanged (or throw) from the string helpers under the same condition. Fixing on the Rust side as well would be preferable so other C consumers also benefit.

</details>
<details>
<summary><b>suggestion:</b> Cap mnemonic helper input before the CJK splitting loop</summary>

`packages/rs-platform-wallet-ffi/src/mnemonic_words.rs` (line 230)

`platform_wallet_mnemonic_normalize_phrase` (230-243) and `platform_wallet_mnemonic_cleanup_phrase` (253-266) accept arbitrary-length C-string input with no cap before `cleanup_phrase_impl` runs. After the early-exit `phrase_is_valid_impl` check (which a long crafted input fails by design), the CJK branch (lines 149-180) walks each word, attempts substrings up to length 8, and on every hit performs a full-string `String::replace` (line 166) followed by a `while s.contains(&dbl_ideo)` collapse pass (167-169) — all superlinear in input length. A maliciously crafted clipboard/QR/universal-link payload consisting of a long no-space run of valid single-ideograph CJK BIP-39 words can therefore force large repeated allocations and freeze the iOS recover-flow thread (Swift's `cleanupPhrase` runs synchronously on the calling thread). BIP-39 mnemonics have a small natural maximum size; enforce a byte- or word-count cap at the FFI entry points (or short-circuit the CJK loop past that cap) before normalization/cleanup runs.

</details>
<details>
<summary><b>nitpick:</b> normalizePhrase / cleanupPhrase silently swallow FFI errors</summary>

`packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/Mnemonic.swift` (line 171)

On UTF-8 conversion error, NULL out-pointer, or allocation failure both wrappers `catch { return phrase }` (lines 177-178, 193-194; the `guard let cStr` branches at 180/196 also silently return the input). Structured `PlatformWalletFFIResult` diagnostics from the Rust side are dropped entirely. The fallback matches DashSync's `NSString` assumption and shouldn't fire for normal Swift `String` input, but if the Rust FFI ever begins returning meaningful error codes — for example the length-cap rejection above — the iOS side has no log line to debug from and the recover UI will silently display the untouched raw input as if cleanup succeeded. Either log via `os_log` on the error branch or make these wrappers `throws` so the recover-flow call sites own the fallback decision explicitly.

</details>

<details>

```

- [SUGGESTION] In `packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/Mnemonic.swift`:158-200: Reject embedded NULs before crossing the C boundary
  `wordIsValid` (158-160), `wordIsLocal` (164-166), `normalizePhrase` (171-184), and `cleanupPhrase` (187-200) all pass the input directly to `withCString`, and the Rust side reads it back with `CStr::from_ptr`. Both terminate at the first U+0000 byte, so a Swift `String` containing an embedded NUL is silently truncated at the FFI boundary: `wordIsValid("abandon\0notaword")` returns `true`, and the string helpers transform only the prefix while the recover UI still displays the full token. This is exploitable by anything that can place a NUL into recovery input (malicious QR code, clipboard from a hostile page, universal-link parameter): a phrase the validator never actually checked can be accepted, or the cleaned text shown back to the user can disagree with the original paste. `Mnemonic.toSeed(mnemonicUTF8Bytes:)` already guards against this at lines 83-86 — mirror that behavior: early-return `false` from the bool helpers when `word.utf8.contains(0)`, and return `phrase` unchanged (or throw) from the string helpers under the same condition. Fixing on the Rust side as well would be preferable so other C consumers also benefit.
- [SUGGESTION] In `packages/rs-platform-wallet-ffi/src/mnemonic_words.rs`:230-266: Cap mnemonic helper input before the CJK splitting loop
  `platform_wallet_mnemonic_normalize_phrase` (230-243) and `platform_wallet_mnemonic_cleanup_phrase` (253-266) accept arbitrary-length C-string input with no cap before `cleanup_phrase_impl` runs. After the early-exit `phrase_is_valid_impl` check (which a long crafted input fails by design), the CJK branch (lines 149-180) walks each word, attempts substrings up to length 8, and on every hit performs a full-string `String::replace` (line 166) followed by a `while s.contains(&dbl_ideo)` collapse pass (167-169) — all superlinear in input length. A maliciously crafted clipboard/QR/universal-link payload consisting of a long no-space run of valid single-ideograph CJK BIP-39 words can therefore force large repeated allocations and freeze the iOS recover-flow thread (Swift's `cleanupPhrase` runs synchronously on the calling thread). BIP-39 mnemonics have a small natural maximum size; enforce a byte- or word-count cap at the FFI entry points (or short-circuit the CJK loop past that cap) before normalization/cleanup runs.
- [NITPICK] In `packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/Mnemonic.swift`:171-200: normalizePhrase / cleanupPhrase silently swallow FFI errors
  On UTF-8 conversion error, NULL out-pointer, or allocation failure both wrappers `catch { return phrase }` (lines 177-178, 193-194; the `guard let cStr` branches at 180/196 also silently return the input). Structured `PlatformWalletFFIResult` diagnostics from the Rust side are dropped entirely. The fallback matches DashSync's `NSString` assumption and shouldn't fire for normal Swift `String` input, but if the Rust FFI ever begins returning meaningful error codes — for example the length-cap rejection above — the iOS side has no log line to debug from and the recover UI will silently display the untouched raw input as if cleanup succeeded. Either log via `os_log` on the error branch or make these wrappers `throws` so the recover-flow call sites own the fallback decision explicitly.
```
</details>

_Inline findings were deduped against existing open review threads; posting this exact-SHA summary so the 7176d824 recheck and prior-finding reconciliation are visible._

packages/rs-platform-wallet-ffi/src/mnemonic_words.rs
llbartekll: Done in 940d251ff2. Dropped platform_wallet_mnemonic_word_is_local and exposed a generic platform_wallet_mnemonic_word_is_in_language(word, language) that wraps key_wallet is_word_in_language — language is a BCP-47-ish code (en, ja, zh-hans, ... covering all 10 key-wallet wordlists). The Swift side is now Mnemonic.wordIsInLanguage(_:language:); deciding what is local is left to the app (the recover flow passes .english explicitly), so that policy is no longer baked into the FFI. Thanks for the catch.

thepastaclaw (commented):
## Code Review

Verified at 940d251f against the worktree. The three FFI-hygiene findings from the prior review (7176d824) remain valid: embedded-NUL truncation at the Swift→C boundary (now covering the new `wordIsInLanguage` wrapper), unbounded input on the normalize/cleanup FFI entry points feeding the superlinear CJK splitter, and silent FFI-error swallow in the Swift string helpers. The prior blocking NFKD CJK auto-split bug is fixed at mnemonic_words.rs:169 and has a regression test. No new blocking issues in this delta.

*Reviewed commit: `940d251f`*

Note: the inline poster deduped these against existing older-SHA review threads, so this same-SHA review carries the verified current-head findings in the body instead of creating duplicate inline threads.

### Findings

1. **[suggestion] Reject embedded NULs before crossing the C boundary**
   - Location: `packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/Mnemonic.swift`:158-206
   - `wordIsValid` (158-160), the new `wordIsInLanguage` (166-172), `normalizePhrase` (177-190), and `cleanupPhrase` (193-206) pass Swift `String` input directly into `withCString`, and the Rust side reads it back via `CStr::from_ptr` (mnemonic_words.rs:223, 248, 252, 278, 301). Both APIs terminate at the first U+0000 byte, so a Swift `String` containing an embedded NUL is silently truncated to its prefix at the FFI boundary: `wordIsValid("abandon\0notaword")` returns `true`, and the string helpers transform only the prefix while the recover UI continues to hold/display the full string. Any source that can place a NUL into recovery input (malicious QR code, hostile clipboard payload, universal-link parameter) can therefore make the validator-visible bytes diverge from what the user sees and what ultimately seeds the wallet. `Mnemonic.toSeed(mnemonicUTF8Bytes:)` already guards against this at lines 83-86 — mirror that policy here: early-return `false` from the bool helpers when `word.utf8.contains(0)` (and `language.code.utf8.contains(0)` for `wordIsInLanguage`), and return `phrase` unchanged (or throw) from the string helpers under the same condition. Defense in depth would also reject NUL on the Rust side so other C consumers benefit.

2. **[suggestion] Cap mnemonic helper input before the CJK splitting loop**
   - Location: `packages/rs-platform-wallet-ffi/src/mnemonic_words.rs`:270-306
   - `platform_wallet_mnemonic_normalize_phrase` (270-283) and `platform_wallet_mnemonic_cleanup_phrase` (293-306) accept arbitrary-length C-string input with no cap before `cleanup_phrase_impl` runs. After the early-exit `phrase_is_valid_impl` check fails (which any sufficiently long crafted input does by design), the CJK branch (169-203) NFKDs the working buffer, walks each whitespace-separated word, attempts substrings up to length 8 per scalar, and on every wordlist hit performs a full-string `String::replace` (189) followed by a `while s.contains(&dbl_ideo)` collapse pass (190-192) — all superlinear in input length. An attacker-supplied no-space run of valid single-ideograph CJK BIP-39 words (each `wchars[0] >= U+3000`, satisfying the gate at 177 even though the global phrase fails checksum validation) reaches the inner loop and forces large repeated allocations on the calling thread. Because Swift's `cleanupPhrase` runs synchronously on whatever thread invokes it (typically the recover-screen UI thread), a crafted clipboard/QR/universal-link payload can freeze the recover flow. BIP-39 mnemonics have a small natural maximum — enforce a byte- or word-count cap at the FFI entry points (or short-circuit the CJK loop past that cap) before normalization/cleanup runs.

3. **[nitpick] normalizePhrase / cleanupPhrase silently swallow FFI errors**
   - Location: `packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/Mnemonic.swift`:177-206
   - On UTF-8 conversion error, NULL out-pointer, or allocation failure both wrappers `catch { return phrase }` (183-185, 199-201); the `guard let cStr` branches at 186/202 also silently return the input. Structured `PlatformWalletFFIResult` diagnostics from the Rust side are dropped entirely. The fallback matches DashSync's `NSString` assumption and shouldn't fire for normal Swift `String` input today, but if the Rust FFI begins returning meaningful error codes (for example the length-cap rejection suggested above), the iOS side has no log line to debug from and the recover UI silently displays the untouched raw input as if cleanup succeeded — while the validator sees a different transform. Log via `os_log` on the error branch, or make these wrappers `throws` so the recover-flow call sites own the fallback decision explicitly.

<details>

```

- [SUGGESTION] In `packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/Mnemonic.swift`:158-206: Reject embedded NULs before crossing the C boundary
  `wordIsValid` (158-160), the new `wordIsInLanguage` (166-172), `normalizePhrase` (177-190), and `cleanupPhrase` (193-206) pass Swift `String` input directly into `withCString`, and the Rust side reads it back via `CStr::from_ptr` (mnemonic_words.rs:223, 248, 252, 278, 301). Both APIs terminate at the first U+0000 byte, so a Swift `String` containing an embedded NUL is silently truncated to its prefix at the FFI boundary: `wordIsValid("abandon\0notaword")` returns `true`, and the string helpers transform only the prefix while the recover UI continues to hold/display the full string. Any source that can place a NUL into recovery input (malicious QR code, hostile clipboard payload, universal-link parameter) can therefore make the validator-visible bytes diverge from what the user sees and what ultimately seeds the wallet. `Mnemonic.toSeed(mnemonicUTF8Bytes:)` already guards against this at lines 83-86 — mirror that policy here: early-return `false` from the bool helpers when `word.utf8.contains(0)` (and `language.code.utf8.contains(0)` for `wordIsInLanguage`), and return `phrase` unchanged (or throw) from the string helpers under the same condition. Defense in depth would also reject NUL on the Rust side so other C consumers benefit.
- [SUGGESTION] In `packages/rs-platform-wallet-ffi/src/mnemonic_words.rs`:270-306: Cap mnemonic helper input before the CJK splitting loop
  `platform_wallet_mnemonic_normalize_phrase` (270-283) and `platform_wallet_mnemonic_cleanup_phrase` (293-306) accept arbitrary-length C-string input with no cap before `cleanup_phrase_impl` runs. After the early-exit `phrase_is_valid_impl` check fails (which any sufficiently long crafted input does by design), the CJK branch (169-203) NFKDs the working buffer, walks each whitespace-separated word, attempts substrings up to length 8 per scalar, and on every wordlist hit performs a full-string `String::replace` (189) followed by a `while s.contains(&dbl_ideo)` collapse pass (190-192) — all superlinear in input length. An attacker-supplied no-space run of valid single-ideograph CJK BIP-39 words (each `wchars[0] >= U+3000`, satisfying the gate at 177 even though the global phrase fails checksum validation) reaches the inner loop and forces large repeated allocations on the calling thread. Because Swift's `cleanupPhrase` runs synchronously on whatever thread invokes it (typically the recover-screen UI thread), a crafted clipboard/QR/universal-link payload can freeze the recover flow. BIP-39 mnemonics have a small natural maximum — enforce a byte- or word-count cap at the FFI entry points (or short-circuit the CJK loop past that cap) before normalization/cleanup runs.
- [NITPICK] In `packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/Mnemonic.swift`:177-206: normalizePhrase / cleanupPhrase silently swallow FFI errors
  On UTF-8 conversion error, NULL out-pointer, or allocation failure both wrappers `catch { return phrase }` (183-185, 199-201); the `guard let cStr` branches at 186/202 also silently return the input. Structured `PlatformWalletFFIResult` diagnostics from the Rust side are dropped entirely. The fallback matches DashSync's `NSString` assumption and shouldn't fire for normal Swift `String` input today, but if the Rust FFI begins returning meaningful error codes (for example the length-cap rejection suggested above), the iOS side has no log line to debug from and the recover UI silently displays the untouched raw input as if cleanup succeeded — while the validator sees a different transform. Log via `os_log` on the error branch, or make these wrappers `throws` so the recover-flow call sites own the fallback decision explicitly.
```
</details>

packages/rs-platform-wallet-ffi/src/mnemonic_words.rs
llbartekll: Done in 6e36a6817e. Dropped the one-line pass-through wrappers normalize_phrase_impl and word_in_language — the FFI entry points and cleanup_phrase_impl now call key_wallet::Mnemonic::normalize_phrase / is_word_in_language directly. Kept the helpers that actually do work: word_in_any_list and phrase_is_valid_impl loop the 7-language bundled union (LANGS), and language_from_code maps the C language code. extern "C" surface unchanged. Thanks.

thepastaclaw (commented):
## Code Review

Verified at 6e36a681. The latest delta is a pure refactor (inlining pass-through wrappers) with no behavioral change to the FFI surface. The prior NFC/NFKD CJK auto-split blocker remains fixed at mnemonic_words.rs:154 with regression coverage. Three prior FFI-hygiene findings carry forward unchanged: embedded-NUL truncation at the Swift→C boundary, unbounded input feeding the superlinear CJK splitter in normalize/cleanup, and silent FFI-error swallow in the Swift string wrappers. No blocking issues; review is COMMENT-only.

Resolved prior finding(s):
- NFC CJK input fails to auto-split because replace operates on the original NFC buffer while candidates are NFKD: Fixed at mnemonic_words.rs:154 by normalizing the working buffer `s` to NFKD before the replace loop; regression test added at lines 417-459 of the same file.

_These are carried-forward findings already represented by existing open review threads, so I am not re-posting duplicate inline comments._

<details>
<summary>🤖 Prompt for all carried-forward findings</summary>

```

- [SUGGESTION] In `packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/Mnemonic.swift`:158-206: Reject embedded NULs before crossing the C boundary
  `wordIsValid` (158-160), `wordIsInLanguage` (166-172), `normalizePhrase` (177-190), and `cleanupPhrase` (193-206) pass Swift `String` input directly into `withCString`, and the Rust side reads it back via `CStr::from_ptr` (mnemonic_words.rs:208, 233, 237, 263, 286). Both APIs terminate at the first U+0000 byte, so a Swift `String` containing an embedded NUL is silently truncated to its prefix at the FFI boundary: `wordIsValid("abandon\0notaword")` returns `true`, and the string helpers transform only the prefix while the recover UI continues to hold/display the full string. Any path that can place a NUL into recovery input (malicious QR code, hostile clipboard payload, universal-link parameter) can therefore make the validator-visible bytes diverge from what the user sees and from what ultimately seeds the wallet — exactly the kind of split-view bug a wallet-recovery UI must not have. `Mnemonic.toSeed(mnemonicUTF8Bytes:)` already guards against this earlier in the same file (lines 83-86); mirror that policy here: early-return `false` from the bool helpers when `word.utf8.contains(0)` (and `language.code.utf8.contains(0)` for `wordIsInLanguage`), and return `phrase` unchanged (or throw) from the string helpers under the same condition. Defense in depth would also reject NUL on the Rust side so other C consumers benefit.
- [SUGGESTION] In `packages/rs-platform-wallet-ffi/src/mnemonic_words.rs`:255-291: Cap mnemonic helper input before the CJK splitting loop
  `platform_wallet_mnemonic_normalize_phrase` (255-268) and `platform_wallet_mnemonic_cleanup_phrase` (278-291) accept arbitrary-length C-string input with no cap before `cleanup_phrase_impl` runs. After the early-exit valid-phrase check fails (which any sufficiently long crafted input does by design), the CJK branch (lines 154-188) NFKDs the working buffer, walks each whitespace-separated word, attempts substrings up to length 8 per scalar, and on every wordlist hit performs a full-string `String::replace` (174) followed by a `while s.contains(&dbl_ideo)` collapse pass (175-177) — all superlinear in input length. An attacker-supplied no-space run of valid single-ideograph CJK BIP-39 words (each `wchars[0] >= U+3000`, satisfying the gate at 162 even though the global phrase fails checksum validation) reaches the inner loop and forces large repeated allocations on the calling thread. Because Swift's `cleanupPhrase` runs synchronously on whatever thread invokes it (typically the recover-screen UI thread), a crafted clipboard/QR/universal-link payload can freeze the recover flow. BIP-39 mnemonics have a small natural maximum (24 words, well under 1 KiB even for CJK); enforce a byte- or word-count cap at the FFI entry points (or short-circuit the CJK loop past that cap) before normalization/cleanup runs.
- [NITPICK] In `packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/Mnemonic.swift`:177-206: normalizePhrase / cleanupPhrase silently swallow FFI errors
  On UTF-8 conversion error, NULL out-pointer, or allocation failure both wrappers `catch { return phrase }` (183-185, 199-201); the `guard let cStr` branches at 186/202 also silently return the input. Structured `PlatformWalletFFIResult` diagnostics from the Rust side are dropped entirely. The fallback matches DashSync's `NSString` assumption and shouldn't fire for normal Swift `String` input today, but if the Rust FFI begins returning meaningful error codes (e.g., the length-cap or embedded-NUL rejection suggested above), the iOS side has no log line to debug from and the recover UI silently displays the untouched raw input as if cleanup succeeded while the validator may see a different transform. Log via `os_log` on the error branch, or make these wrappers `throws` so the recover-flow call sites own the fallback decision explicitly.
```
</details>

packages/rs-platform-wallet-ffi/src/mnemonic_words.rs
llbartekll: Done in 707e1963cd. Moved the recover-flow logic (CJK ideographic auto-split + punctuation cleanup, the 7-language bundled-language union, BCP-47 language mapping) into the non-FFI crate as platform_wallet::mnemonic. The four extern "C" entry points now only marshal C strings and delegate; normalize stays a direct key_wallet::Mnemonic::normalize_phrase call (raw BIP-39 primitive, no platform policy). The extern "C" surface is unchanged, so the cbindgen header and the Swift SDK are untouched.

packages/rs-platform-wallet-ffi/src/mnemonic_words.rs
llbartekll: Agreed. Per-language validate coverage now lives natively in key-wallet via dashpay/rust-dashcore#807 (validate a real 12-word phrase for French, Spanish, Italian, Japanese, Korean, Czech, ChineseSimplified, ChineseTraditional). The union-level check moved with the rest of the recover-flow logic to platform_wallet::mnemonic in 707e1963cd.

packages/rs-platform-wallet-ffi/src/mnemonic_words.rs
llbartekll: Agreed -- that one exercised key_wallet normalize behaviour. Deleted as redundant in 707e1963cd; key-wallet already covers it (test_normalize_phrase + test_normalize_phrase_unicode_forms_converge).

packages/rs-platform-wallet-ffi/src/mnemonic_words.rs
llbartekll: Agreed. Deleted in 707e1963cd: key-wallet test_bip39_test_vectors runs the same Trezor vectors and test_is_word_in_language covers membership, so the facade copy was pure duplication. The recover-flow-specific tests (CJK cleanup, the bundled union) moved to platform_wallet::mnemonic; the FFI crate keeps only the extern "C" boundary test (ffi_roundtrip).

thepastaclaw (commented):
## Code Review

Refactor moves mnemonic recover-flow policy upstream to key_wallet::Mnemonic and reduces the FFI to three thin pass-throughs (word_list, normalize_phrase, cleanup_phrase). Two prior hygiene findings carry forward to the surviving Swift string wrappers (embedded-NUL truncation, swallowed FFI errors); CJK input-cap and bool-wrapper NUL concerns are resolved. No blockers.

Resolved prior finding(s):
- Cap mnemonic helper input before the CJK splitting loop: Fixed upstream — `cleanup_phrase` now delegates to `key_wallet::Mnemonic::cleanup_phrase` at rust-dashcore rev d8fe72cf, which retains the `MAX_CLEANUP_BYTES = 4096` guard before the superlinear CJK scan.
- wordIsValid / wordIsInLanguage NUL truncation at the C boundary: Fixed by removal — the granular `word_is_valid` / `word_is_in_language` FFI entry points and their Swift wrappers were deleted in this delta; the recover UI now composes word-validity checks itself from `wordList(language:)`.

_Inline findings were deduped against existing older-SHA review threads, so this exact-SHA review carries the verified current-head findings in the body instead of creating duplicate inline threads._

<details>
<summary>🤖 Prompt for all carried-forward findings</summary>

```

- [SUGGESTION] In `packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/Mnemonic.swift`:177-206: Reject embedded NULs before crossing the C boundary in normalizePhrase/cleanupPhrase
  `normalizePhrase` (177-190) and `cleanupPhrase` (193-206) pass Swift `String` straight into `withCString`, and Rust reads it back via `CStr::from_ptr(...).to_str()` at `packages/rs-platform-wallet-ffi/src/mnemonic_words.rs:100` and `:123`. Both APIs terminate at the first U+0000 byte, so a Swift string containing an embedded NUL is silently truncated to its prefix at the FFI boundary: the helper transforms only the prefix while the caller continues to hold the full string. That is a real split-view risk in a recovery flow — `normalizePhrase("abandon\0notaword")` returns `"abandon"`, so any caller that does `cleanupPhrase(raw) == raw` checks, echoes the cleaned text back to the user, or feeds the original `raw` to the validator while displaying the cleaned prefix, sees a buffer divergence sourced from a malicious clipboard, QR code, or universal-link payload. `toSeed(mnemonicUTF8Bytes:)` already enforces this policy in the same file at lines 83-86, so the SDK already considers NUL-in-mnemonic an invariant violation. Mirror that here: return `phrase` unchanged (or throw) when `phrase.utf8.contains(0)` before invoking `withCString`. Defense-in-depth NUL rejection on the Rust side belongs in upstream `key_wallet` and is out of scope for this PR.
- [NITPICK] In `packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/Mnemonic.swift`:161-206: wordList / normalizePhrase / cleanupPhrase silently swallow FFI errors
  All three wrappers `catch { return ... }` and the `guard let cStr` branches return a benign fallback (`[]` for `wordList` at 163-168, `phrase` for the string helpers at 183-186 / 199-202). Rust exposes structured `PlatformWalletFFIResult` codes for `ErrorNullPointer`, `ErrorUtf8Conversion`, and interior-NUL `CString::new` failures, but every diagnostic is dropped. The empty-array fallback for `wordList` is the worst case: a caller that builds a `Set` for any-language word-validity checks would classify every typed word as invalid with no log line to diagnose. The error paths cannot fire for normal Swift `String` and the static 2048-entry BIP-39 wordlists today, so this is not a defect now, but if the FFI later returns meaningful errors (e.g. a future hard-error variant of the upstream 4 KiB cleanup cap, or explicit NUL rejection per the finding above), iOS has no signal. At minimum log via `os_log` on the error branch, or make these wrappers `throws` so the recover-flow call sites own the fallback decision explicitly.
```
</details>

thepastaclaw (approved):
## Code Review

Verified at 7a82f2a6. The Swift recover-flow wrappers in Mnemonic.swift now (1) reject embedded NULs in normalizePhrase (line 187) and cleanupPhrase (line 208) before crossing withCString, mirroring the toSeed policy at line 88, and (2) log FFI errors via os.Logger at lines 171, 194, 215 with privacy:.public scoped only to error.localizedDescription (Logger declared at lines 15-17 with an in-source contract that no phrase content is logged). Both prior findings from ed675d4f are FIXED; no new in-scope defects. Three independent reviewers (claude-general, claude-security-auditor, claude-ffi-engineer) and three codex reviewers converge on no remaining in-scope findings. CodeRabbit context contained zero inline findings.

thepastaclaw (approved):
## Code Review

Delta from 7a82f2a6 to d628021c is solely a rust-dashcore dependency pin bump (d8fe72cf → 981e97f1, the merged upstream #806) in Cargo.toml/Cargo.lock with no Swift or Rust source changes. All six agent reports converge: prior recover-flow findings (embedded-NUL guard before withCString in normalizePhrase/cleanupPhrase, os.Logger error logging in the three string wrappers) remain FIXED at Mnemonic.swift:187, :208, :17, :171, :194, :215, and the thin FFI surface in mnemonic_words.rs is untouched. No new in-scope defects; out-of-scope items (repr(C) enum hardening, pre-existing validate(_:) NUL guard) are library-wide concerns unrelated to this delta.

packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/Mnemonic.swift
```diff
@@ -149,4 +154,70 @@ public class Mnemonic {
...
+    /// in the app, not the SDK (callers build `Set`s from these lists). Returns
+    /// `[]` only on the UTF-8/allocation error paths that never occur for a
+    /// real wordlist.
+    public static func wordList(language: MnemonicLanguage) -> [String] {
+        var outPtr: UnsafeMutablePointer<CChar>? = nil
+        do {
+            try platform_wallet_mnemonic_word_list(language.ffiMnemonicValue, &outPtr).check()
+        } catch {
+            Self.log.error("wordList FFI error: \(error.localizedDescription, privacy: .public)")
+            return []
```
ZocoLini: I don't really like the idea of logging errors that cannot really happen, eg, platform_wallet_mnemonic_word_list returns an error if we send outPtr == null or if Cstring::new fails, what is impossible to happen, I rather prefer try! or throwing the exception to the caller

ZocoLini (changes requested):

packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/Mnemonic.swift
llbartekll: Done in 2a3f18ae — word_list / normalize_phrase / cleanup_phrase now use try! (trap on the unreachable FFI error) instead of logging, and the os.Logger import is gone. The only non-trap path left is the explicit embedded-NUL guard on normalize/cleanup, which passes the input through unchanged. Thanks!

thepastaclaw (approved):
## Code Review

Verified at HEAD 2a3f18aeb8da2fce56adaa2db6cc80ba497c7be8. The single-commit delta swaps os.Logger error fallbacks in Mnemonic.wordList/normalizePhrase/cleanupPhrase for try! traps, documented as unreachable post-guard. All six agent runs converged on zero in-scope findings; the embedded-NUL guards at Mnemonic.swift:181 and :199 remain intact, and the Rust FFI error variants in mnemonic_words.rs are not reachable under the Swift call shape (non-NULL out-params via &outPtr, valid UTF-8 from withCString, NUL-free outputs from key_wallet producers). No CodeRabbit inline findings were supplied.

ZocoLini (approved):

QuantumExplorer (approved):

PR #3843, opened 2026-06-10
fix(platform): derive shielded identity-create id from the padded bundle's published nullifiers

QuantumExplorer:

Two client-side shielded bugs found during the paloma E2E run of the full shielded transition matrix (types 15–20), both blocking real-world flows:

1. **Type-20 (`IdentityCreateFromShieldedPool`) was broken for single-note spends.** The builder derived the new identity's id from the *requested spends'* nullifiers, but Orchard's `BundleType::DEFAULT` pads a single-spend bundle to the 2-action minimum — and the padding action's **random dummy nullifier is published on the wire**. Consensus re-derives the id over **all** published action nullifiers, so the builder's id (and the id bound into the Orchard `extra_sighash_data`) diverged from the consensus derivation whenever padding occurred. The builder's own post-proving id-consistency check caught the mismatch and failed the build — but only *after* paying the full Halo 2 proving cost, every time, making every 1-note identity creation unusable.

2. **Unshield was unusable on devnet/regtest.** `PlatformWallet::shielded_unshield_to` compared the parsed output address's network against the SDK network with raw `Network` equality — but bech32m parsing can only ever yield `Mainnet` or `Testnet` (testnet/devnet/regtest all share the `tdash` HRP per DIP-0018), so a `Devnet` wallet rejected every address, including its own.

## What was done?

**Commit 1 — type-20 id derivation from the padded bundle (`rs-dpp`):**
- New `build_spend_bundle_with` / `prove_and_sign_bundle_with` variants that take a closure `FnOnce(&[[u8; 32]]) -> Result<Vec<u8>, ProtocolError>` computing the extra sighash data from the **built** bundle's published action nullifiers — extracted after `Builder::build` fixes the action set (padding included) and before the sighash is bound / proof created. The existing fixed-data functions delegate to the `_with` variants.
- `build_identity_create_from_shielded_pool_transition` now derives the identity id **inside that hook** (`identity_id_from_nullifiers(published_nullifiers)`) and binds the extra sighash data there, so the bound id always equals the consensus re-derivation. The post-build id-consistency assert is kept as a true invariant guard.
- Doc comment corrected (the id is *not* known before the bundle is built when padding occurs) and a stale duplicated comment paragraph deduplicated in drive-abci's `state.rs` type-20 arm (comment-only).

**Commit 2 — unshield address network check by HRP (`rs-platform-wallet`):**
- `shielded_unshield_to` now compares `PlatformAddress::hrp_for_network(addr_network)` against `hrp_for_network(self.sdk.network)` instead of raw networks, matching what the bech32m encoding can actually represent.

## How Has This Been Tested?

New regression tests, all with **real Halo 2 proving** (the cached `TestProver` pattern):

- `single_spend_padded_bundle_derives_id_from_published_nullifiers` — the bug's exact shape: a 1-spend build (witnessed via the same `merkle_path.root(cmx)` anchor pattern production uses) must succeed, be padded to 2 actions, carry the real nullifier among the published set, and produce `identity_id == derive_identity_id_from_actions(actions)` while `!= identity_id_from_nullifiers([real_nullifier])` — proving the dummy participates. Pre-fix this failed the post-proving consistency check.
- `two_spend_unpadded_bundle_id_matches_real_nullifier_derivation` — no padding: the published set equals the real spends', and both derivations agree.
- `prove_and_sign_bundle_with_closure_receives_published_nullifiers` — the closure contract: it runs once and receives exactly the authorized bundle's published nullifiers (one per action, padding included, on-wire order).
- `prove_and_sign_bundle_with_closure_error_short_circuits_before_proving` — closure errors propagate before any proving work.
- `hrp_is_shared_across_all_non_mainnet_networks` — pins the DIP-0018 premise the wallet check now relies on (testnet/devnet/regtest share `tdash`; mainnet differs).

`cargo check --workspace --all-targets` green; `cargo check -p dpp / -p platform-wallet --all-features --tests` green; clippy clean on the touched crates; existing shielded builder suites still pass.

Both fixes were also exercised end-to-end on devnet paloma during the shielded matrix run: post-fix, unshield (type 17) completed pool → platform-address, and the type-20 built, proved, and broadcast successfully with a single-note spend.

## Breaking Changes

None. Client-side builder/wallet fixes only — no consensus, serialization, or protocol-version changes. `build_spend_bundle` / `prove_and_sign_bundle` keep their existing signatures (the `_with` variants are additive, crate-internal).

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed network address validation that was incorrectly rejecting valid testnet/devnet/regtest addresses due to HRP handling

* **Refactor**
  * Refactored shielded bundle builders for improved sighash computation
  * Updated identity derivation logic in shielded pool operations

* **Tests**
  * Added network address validation tests

### DIFF
--- a/packages/rs-dpp/src/shielded/builder/mod.rs
+++ b/packages/rs-dpp/src/shielded/builder/mod.rs
@@ -203,6 +203,44 @@ pub(crate) fn build_spend_bundle<P: OrchardProver>(
     prover: &P,
     extra_sighash_data: &[u8],
 ) -> Result<Bundle<Authorized, i64, DashMemo>, ProtocolError> {
+    let data = extra_sighash_data.to_vec();
+    build_spend_bundle_with(
+        spends,
+        recipient,
+        output_amount,
+        memo,
+        fvk,
+        ask,
+        anchor,
+        prover,
+        move |_| Ok(data),
+    )
+}
+
+/// Like [`build_spend_bundle`], but the extra sighash data is computed by a
+/// closure that receives the built bundle's published action nullifiers (in
+/// on-wire order, INCLUDING any padding actions' dummy nullifiers).
+///
+/// `IdentityCreateFromShieldedPool` needs this: its identity id is
+/// `double_sha256(sorted published nullifiers)`, and `BundleType::DEFAULT`
+/// pads single-spend bundles with a dummy action whose random nullifier only
+/// exists once the bundle is built — deriving the id from the real spends
+/// alone would diverge from the consensus re-derivation.
+#[allow(clippy::too_many_arguments)]
+pub(crate) fn build_spend_bundle_with<P: OrchardProver, F>(
+    spends: Vec<SpendableNote>,
+    recipient: &OrchardAddress,
+    output_amount: u64,
+    memo: [u8; 36],
+    fvk: &FullViewingKey,
+    ask: &SpendAuthorizingKey,
+    anchor: Anchor,
+    prover: &P,
+    extra_sighash_data: F,
+) -> Result<Bundle<Authorized, i64, DashMemo>, ProtocolError>
+where
+    F: FnOnce(&[[u8; 32]]) -> Result<Vec<u8>, ProtocolError>,
+{
     let payment_address = PaymentAddress::from(recipient);
 
     let mut builder = Builder::<DashMemo>::new(BundleType::DEFAULT, anchor);
@@ -224,7 +262,7 @@ pub(crate) fn build_spend_bundle<P: OrchardProver>(
         )
         .map_err(|e| ProtocolError::ShieldedBuildError(format!("failed to add output: {:?}", e)))?;
 
-    prove_and_sign_bundle(
+    prove_and_sign_bundle_with(
         builder,
         prover,
         std::slice::from_ref(ask),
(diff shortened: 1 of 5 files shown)


thepastaclaw:
✅ Review complete (commit 0bd33954)

QuantumExplorer (commented):
Reviewed

packages/rs-dpp/src/shielded/builder/mod.rs
```diff
@@ -550,4 +612,90 @@ mod mod_tests {
...
+            ))
+        });
+
+        match result {
+            Err(ProtocolError::ShieldedBuildError(msg)) => {
+                assert_eq!(msg, "closure rejected", "closure error must pass through");
+            }
+            other => panic!("expected the closure's error to propagate, got {:?}", other),
+        }
+    }
```
thepastaclaw, line 700: **🟡 Suggestion: Short-circuit test does not actually enforce "before proving"**

The test name and intent — that the closure error propagates before Halo 2 proving runs — is the exact invariant motivating the closure refactor (the original bug burned a full ~30 s proof before failing). But the test passes `TestProver`, so it only verifies that the closure error propagates. If `prove_and_sign_bundle_with` were ever reordered to call `create_proof` before invoking the closure and then return the same closure error, this test would still pass. Use a test-only prover whose `proving_key()` panics so any accidental call into proving fails the test deterministically.

<sub>source: ['codex']</sub>

thepastaclaw (commented):
## Code Review

Two narrowly-scoped client-side fixes: (1) derive the shielded identity-create id from the built bundle's published action nullifiers via a post-build closure so it matches the consensus rederivation over padded actions, and (2) compare unshield destination by HRP instead of raw Network so devnet/regtest work. Both changes are correct, well-tested, and have no consensus surface. Only minor test-coverage and documentation nitpicks remain; nothing blocking.

<details>

```

In `packages/rs-dpp/src/shielded/builder/mod.rs`:
- [SUGGESTION] packages/rs-dpp/src/shielded/builder/mod.rs:685-700: Short-circuit test does not actually enforce "before proving"
  The test name and intent — that the closure error propagates before Halo 2 proving runs — is the exact invariant motivating the closure refactor (the original bug burned a full ~30 s proof before failing). But the test passes `TestProver`, so it only verifies that the closure error propagates. If `prove_and_sign_bundle_with` were ever reordered to call `create_proof` before invoking the closure and then return the same closure error, this test would still pass. Use a test-only prover whose `proving_key()` panics so any accidental call into proving fails the test deterministically.
```
</details>

PR #3849, opened 2026-06-10
test(drive-abci): add token supply edge-case coverage

### LINKED ISSUE #3848: Minting past i64::MAX on an uncapped token returns InternalError instead of a consensus error
## Problem

When a token has `max_supply == None`, mint state validation skips its supply-bound check entirely. A mint that would push total supply past `i64::MAX` (the GroveDB sum-item ceiling) is caught only by the low-level Drive `checked_add`, surfacing as an `InternalError` rather than a graceful consensus rejection.

Supply is not corrupted (the write is prevented), but the failure class is wrong.

Observed result:
```
InternalError("storage: drive: corrupted code execution error: trying to add an amount that would overflow total supply")
```

## Where

- `token_mint_transition_action/state_v0/mod.rs` — the supply check is wrapped in `if let Some(max_supply) = token_configuration.max_supply()`, so the `None` case has no validation-layer guard.
- Guard that actually fires: `add_to_token_total_supply/v0/mod.rs` (`checked_add`).

## Expected

Reject at the validation layer with a graceful consensus error (treat `i64::MAX` as an implicit hard ceiling when `max_supply` is `None`).

## Test

Current behavior is pinned by the characterization test `test_token_mint_with_max_i64_base_supply_then_overflow_returns_internal_error_without_mutating_supply` (`batch/tests/token/mint/mod.rs`). Adding the guard should break it — that is the signal to update it to expect the graceful rejection.

Related, distinct gap: `base_supply > max_supply` accepted at creation, documented by the `#[ignore]`d `test_data_contract_creation_with_base_supply_over_max_supply_should_cause_error`.

thephez:

Token supply handling lacked coverage for several edge cases (exact `max_supply` boundary, burning below `base_supply`, burning to zero, destroy-frozen-funds supply accounting, zero/`i64::MAX` `base_supply`, etc.). While adding that coverage, two real gaps surfaced:

1. **`base_supply > max_supply` is accepted at contract creation.** There is no creation-time guard comparing the two (only `base_supply > i64::MAX` is rejected). The token is born already over its cap, `base_supply` is immutable,    so every subsequent mint is permanently blocked by `TokenMintPastMaxSupplyError`. Tracked in dashpay/platform#3850
2. **Minting past `i64::MAX` on a token with `max_supply == None` returns an `InternalError`** instead of a graceful consensus error. The mint validation's supply check is wrapped in `if let Some(max_supply) = ...`, so the uncapped case has no validation-layer guard and is only caught by the low-level Drive `checked_add`. Tracked in dashpay/platform#3848.

This PR is **tests only** — no production behavior changes. The two gaps are documented (see below) for follow-up fixes.

## What was done?

Added ABCI state-transition tests under `.../batch/tests/token/`:

- **Mint:** mint to exactly `max_supply` (succeeds), one over (fails, full `TokenMintPastMaxSupplyError` payload asserted), unbounded mint when `max_supply == None`, mint from zero `base_supply`, mint from zero to exact max.
- **Burn:** burn below `base_supply` (allowed by design), burn entire supply to zero then mint again.
- **Destroy frozen funds:** total supply reduced to zero, and decremented by exactly the destroyed amount (existing tests never asserted total supply).
- **Config update:** set `max_supply` equal to current supply (boundary, succeeds), raise `max_supply` then mint into the new headroom.

Documenting the two gaps:

- `test_data_contract_creation_with_base_supply_over_max_supply_should_cause_error` (in `data_contract_create`) — runs a real `DataContractCreateTransition` and asserts the **intended** rejection. `#[ignore]`d because it fails on current code; it is the regression target for adding the creation-time guard.
- `test_token_mint_with_max_i64_base_supply_then_overflow_returns_internal_error_without_mutating_supply`
  — a **characterization** test pinning the current `InternalError`-without-mutation behavior (gap #2). Named to make clear it is not the desired long-term API; it should break (intentionally) when the validation guard is added.

## How Has This Been Tested?

- `cargo test -p drive-abci` for the `token` and `data_contract_create::tests::tokens` modules: all executable tests pass (200 + 21), with exactly one `#[ignore]`d test (the creation-guard regression target).
- Verified the ignored test fails as intended under `--ignored` (proving the gap is real and the assertion non-vacuous).
- `cargo fmt -p drive-abci` clean.

## Breaking Changes

None — tests only.

## Checklist:

## Summary by CodeRabbit

* **Tests**
  * Added token burn coverage for burns below base supply and for burning all supply then minting again.
  * Added token configuration update tests for updating max supply to current supply and raising the cap before minting.
  * Added token destroy-frozen-funds tests covering zeroing total supply and decrementing total supply while validating account state.
  * Expanded token minting boundary tests for max supply behavior, `max_supply = None`, `base_supply` edge cases, and overflow handling.
  * Added an ignored regression test for data contract creation with inconsistent base/max supply.


thepastaclaw:
✅ Final review complete — no blockers (commit 4496d96a)

thepastaclaw (commented):
## Code Review

Tests-only PR adding well-targeted token supply edge-case coverage and two characterization tests for documented gaps (#3848, #3850). Verified the new tests against the production code: assertions match current behavior and the helpers mirror the established patterns. One in-scope suggestion: the ignored regression test for `base_supply > max_supply` should narrow its assertion to `UnpaidConsensusError(BasicError(_))` so it actually enforces the intended validation stage when the fix lands.

🟡 1 suggestion(s)

<details>

```

In `packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/data_contract_create/mod.rs`:
- [SUGGESTION] packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/data_contract_create/mod.rs:1764-1768: Ignored regression test accepts too broad a class of errors
  `test_data_contract_creation_with_base_supply_over_max_supply_should_cause_error` is the regression target for a future creation-time guard, but the assertion accepts either `UnpaidConsensusError` or any `PaidConsensusError`. A `base_supply > max_supply` violation is a contract-shape problem that the future guard should reject at basic structure validation (alongside the existing `base_supply > i64::MAX` check noted in the comment), which produces `UnpaidConsensusError(BasicError(_))`. As written, this test would still pass even if the fix only rejected the contract later in the paid execution path, defeating the regression intent. The `total_supply == None` assertion below proves the token was not created but not that rejection happened in the right stage.
```
</details>

packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/data_contract_create/mod.rs
```diff
@@ -1680,6 +1680,110 @@ mod tests {
...
+                        None,
+                    )
+                    .expect("expected to process state transition");
+
+                // INTENDED: creation is rejected with a consensus error, not executed.
+                assert_matches!(
+                    processing_result.execution_results().as_slice(),
+                    [StateTransitionExecutionResult::UnpaidConsensusError(_)
+                        | StateTransitionExecutionResult::PaidConsensusError { .. }]
+                );
```
thepastaclaw, line 1916: **🟡 Suggestion: Ignored regression test accepts too broad a class of errors**

`test_data_contract_creation_with_base_supply_over_max_supply_should_cause_error` is the regression target for a future creation-time guard, but the assertion accepts either `UnpaidConsensusError` or any `PaidConsensusError`. A `base_supply > max_supply` violation is a contract-shape problem that the future guard should reject at basic structure validation (alongside the existing `base_supply > i64::MAX` check noted in the comment), which produces `UnpaidConsensusError(BasicError(_))`. As written, this test would still pass even if the fix only rejected the contract later in the paid execution path, defeating the regression intent. The `total_supply == None` assertion below proves the token was not created but not that rejection happened in the right stage.

```suggestion
                assert_matches!(
                    processing_result.execution_results().as_slice(),
                    [StateTransitionExecutionResult::UnpaidConsensusError(
                        ConsensusError::BasicError(_)
                    )]
                );
```

<sub>source: ['codex']</sub>

thepastaclaw (approved):
## Code Review

Tests-only PR adding token supply edge-case coverage in drive-abci. The single prior finding from 8c1fedfa (overly broad assertion in the ignored regression test) is FIXED at 6354960904 — the assertion now requires `UnpaidConsensusError(ConsensusError::BasicError(_))`, locking the future fix to basic-structure validation. No new in-scope issues; documented production gaps (#3848, #3850) are explicitly tracked as out-of-scope follow-ups.

QuantumExplorer (approved):

PR #3851, opened 2026-06-10
fix(wasm-sdk): label getTokenContractInfo parameter as tokenId, not contractId

thephez:

`getTokenContractInfo` / `getTokenContractInfoWithProofInfo` (wasm-sdk) and the `tokens.contractInfo` / `tokens.contractInfoWithProof` facade (js-evo-sdk) labeled their single parameter `dataContractId` / `contractId`. The query is actually keyed by **token ID**, which is a distinct value derived from a data contract ID and the token's position (`token_id = sha256d("dash_token" || contract_id || position)`).

Because the Drive query is a single-key GroveDB lookup keyed by `token_id`, passing a contract ID — as the parameter name invited — matched no record and silently resolved to `undefined` rather than erroring. Developers following the parameter name would conclude the token had no contract info.

The Rust SDK query type (`TokenContractInfoQuery { token_id }`), the gRPC proto field, and the Drive layer were already correctly named `token_id`; the mislabeling existed only at the WASM/JS surface.

## What was done?

- Renamed the parameter to `tokenId` on `getTokenContractInfo` and `getTokenContractInfoWithProofInfo`, corrected the error messages to "Invalid token ID", and added doc comments that state the argument is a token ID and point to `calculateTokenIdFromContract` for deriving one.
- Renamed the parameter to `tokenId` on the `tokens.contractInfo` / `tokens.contractInfoWithProof` facade methods with matching JSDoc.
- Fixed the wasm-sdk functional test, which previously passed a contract ID with no assertion (silently exercising the bug). It now derives a token ID via `calculateTokenIdFromContract` and asserts the returned `contractId` / `tokenContractPosition`.

The generated `dist/*.d.ts` regenerate from the wasm_bindgen `js_name` and TS source on build, so they were not hand-edited.

## How Has This Been Tested?

- `cargo build -p wasm-sdk` — compiles clean.
- `yarn workspace @dashevo/wasm-sdk test:functional` — the rewritten `getTokenContractInfo()` test now asserts the result, so it fails loudly if a contract ID (silent `None`) is ever passed again.
- `yarn workspace @dashevo/evo-sdk test:unit` — facade pass-through tests.

## Breaking Changes

None at runtime. wasm_bindgen generates positional JS arguments, so renaming the parameter does not change the call signature for existing callers — it only corrects the TypeScript parameter label and generated docs. The old `dataContractId` / `contractId` label was actively misleading and is the reason for this change.

## Checklist:

## Summary by CodeRabbit

* **API Changes**
  * Methods returning token contract info now accept token IDs (parameter renamed) in both JS and WASM bindings, aligning public interfaces.
* **Documentation**
  * JSDoc/wasm-bindgen metadata updated to clarify token ID input and how to derive it.
* **Tests**
  * Unit and functional tests updated to use calculated token IDs and to assert behavior with the new parameter name.

### DIFF
--- a/packages/wasm-sdk/src/queries/token.rs
+++ b/packages/wasm-sdk/src/queries/token.rs
@@ -846,8 +846,7 @@ impl WasmSdk {
 
         // Fetch token contract info with proof
         let (info_result, metadata, proof) =
-            TokenContractInfo::fetch_with_metadata_and_proof(self.as_ref(), token_id, None)
-                .await?;
+            TokenContractInfo::fetch_with_metadata_and_proof(self.as_ref(), token_id, None).await?;
 
         let data = info_result
             .map(|info| JsValue::from(TokenContractInfoWasm::from(info)))
--- a/packages/wasm-sdk/src/queries/token.rs
+++ b/packages/wasm-sdk/src/queries/token.rs
@@ -457,21 +457,40 @@ impl WasmSdk {
         Ok(prices_map)
     }
 
+    /// Fetches the contract info for a token (the data contract that defines it
+    /// and the token's position within that contract).
+    ///
+    /// This query is keyed by **token ID**, not by data contract ID. A token ID
+    /// is derived from a data contract ID and the token's position via
+    /// `calculateTokenIdFromContract`. Passing a data contract ID here will not
+    /// match any record and resolves to `undefined`.
+    ///
+    /// # Arguments
+    /// * `token_id` - The token ID in base58 format
+    ///
+    /// # Returns
+    /// The token's contract info, or `undefined` if no token with that ID exists.
+    ///
+    /// # Example
+    /// ```javascript
+    /// const tokenId = sdk.calculateTokenIdFromContract("Hqyu8WcRwXCTwbNxdga4CN5gsVEGc67wng4TFzceyLUv", 0);
+    /// const info = await sdk.getTokenContractInfo(tokenId);
+    /// ```
     #[wasm_bindgen(js_name = "getTokenContractInfo")]
     pub async fn get_token_contract_info(
         &self,
-        #[wasm_bindgen(js_name = "dataContractId")] data_contract_id: IdentifierLikeJs,
+        #[wasm_bindgen(js_name = "tokenId")] token_id: IdentifierLikeJs,
     ) -> Result<Option<TokenContractInfoWasm>, WasmSdkError> {
         use dash_sdk::dpp::tokens::contract_info::TokenContractInfo;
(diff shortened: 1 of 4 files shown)


thepastaclaw:
✅ Review complete (commit 8c28488a)

thepastaclaw (commented):
## Code Review

Small, well-scoped labeling fix that renames the public WASM/JS parameter from contract/dataContract ID to tokenId on getTokenContractInfo and its proof variant. Underlying query and FFI behavior are unchanged. One valid documentation nitpick: the new JSDoc example calls the static token-ID helper as an instance method.

💬 1 nitpick(s)

packages/wasm-sdk/src/queries/token.rs
```diff
@@ -457,21 +457,40 @@ impl WasmSdk {
...
+    ///
+    /// # Arguments
+    /// * `token_id` - The token ID in base58 format
+    ///
+    /// # Returns
+    /// The token's contract info, or `undefined` if no token with that ID exists.
+    ///
+    /// # Example
+    /// ```javascript
+    /// const tokenId = sdk.calculateTokenIdFromContract("Hqyu8WcRwXCTwbNxdga4CN5gsVEGc67wng4TFzceyLUv", 0);
```
thepastaclaw: **💬 Nitpick: Doc example calls the token ID helper as an instance method**

calculate_token_id_from_contract is defined without &self (token.rs:135-139), so wasm-bindgen exposes it as a static associated function. Every existing call site uses sdk.WasmSdk.calculateTokenIdFromContract(...). The new example written as `sdk.calculateTokenIdFromContract(...)` would throw at runtime, defeating the PR's goal of making correct token-ID usage clear. (Note: the same wording exists in the pre-existing example at line 133, but is out of scope for this PR.)

```suggestion
    /// const tokenId = WasmSdk.calculateTokenIdFromContract("Hqyu8WcRwXCTwbNxdga4CN5gsVEGc67wng4TFzceyLUv", 0);
```

<sub>source: ['codex']</sub>

thepastaclaw (approved):
## Code Review

Verified prior-1 against current head: the JSDoc example at packages/wasm-sdk/src/queries/token.rs:476 now correctly uses the static form `WasmSdk.calculateTokenIdFromContract(...)`, matching how wasm-bindgen exposes the helper (it has no &self receiver). Both agents and CodeRabbit converge on no new in-scope findings for the 8c28488a delta (rustfmt + doc nit fix on top of the parameter renames). PR is a clean, well-scoped fix.

QuantumExplorer (approved):

PR #3855, opened 2026-06-11
fix(platform): zero cached platform-address balances absent from state

QuantumExplorer:

After the paloma devnet platform chain was reset, SwiftExampleApp kept displaying a stale platform-address balance (e.g. 294,627,247,940 credits / "2.94627248 DASH") from the old chain, and — worse — offered it as an identity-creation funding source that would fail at broadcast. Clearing Platform Sync Status and re-syncing against the new chain did not help: on the new chain `getAddressInfo` proves the address absent from state (entry without `balance_and_nonce`), but the balance refresh only overwrote cached rows on positive finds, so the stale balance was never zeroed.

The SDK sync engine already reports proven-absent addresses (`AddressSyncResult.absent`), but `rs-platform-wallet` ignored that signal in three places:

1. `PlatformAddressWallet::sync_balances` built the persistence changeset only from `result.found`, so no zeroing entry ever reached the persister (→ FFI → SwiftData).
2. `PlatformPaymentAddressProvider::sync_finished` merged scratch `found` entries but never removed proven-absent addresses from the committed `found` map, so `current_balances()` kept re-seeding the stale balance into every subsequent sync.
3. `on_address_absent` never updated key-wallet's managed account, so the in-memory balance used by spend-path enumeration (identity-creation funding) kept the stale value.

## What was done?

All changes are in `rs-platform-wallet` (per the project convention that the Swift layer only persists what the FFI hands it):

- Extracted the sync diff into a pure helper `compute_address_balance_diff(before, found, absent)` in `wallet/platform_addresses/sync.rs`. It emits found-and-changed entries as before, plus a zeroed entry (balance 0, nonce 0) for every proven-absent address that previously carried non-default cached funds. An absent address has no balance and no nonce in state, so both are reset.
- `sync_finished` in `wallet/platform_addresses/provider.rs` now removes every address proven absent this pass from the committed `found` map (extend-then-remove, so an absent proof wins in the impossible-by-contract overlap case).
- `on_address_absent` now zeroes the managed account's credit balance via `set_address_credit_balance(p2pkh, 0, None)`, mirroring `on_address_found`'s wallet-manager write. The `None` key source is deliberate: gap-limit extension only fires on an unfunded→funded transition, which zeroing never is.

No Swift changes were needed: `PlatformAddressChangeSet::is_empty`, the FFI mapping in `rs-platform-wallet-ffi`, and Swift's `PlatformWalletPersistenceHandler.persistAddressBalances` all pass zero-balance entries through unconditionally, so the `PersistentPlatformAddress` row is updated to 0 end-to-end.

Note: the Platform Sync Status "Clear" button only resets in-memory display state and was left untouched — with this fix the stale balance is zeroed by the next sync pass itself, which is the layer that owns the decision.

## How Has This Been Tested?

New unit tests in `rs-platform-wallet`:

- `sync::tests::absent_previously_funded_address_emits_zero_entry` — the chain-reset case: cached balance + proven absent → single zeroed entry (balance 0, nonce 0)
- `sync::tests::absent_never_funded_address_emits_nothing`
- `sync::tests::absent_zero_balance_nonzero_nonce_emits_zero_entry`
- `sync::tests::absent_default_funds_emits_nothing`
- `sync::tests::found_changed_emitted_unchanged_skipped`
- `sync::tests::found_and_absent_combine`
- `provider::tests::sync_finished_removes_absent_from_committed_found` — absent address evicted from the committed `found` map and no longer yielded by `current_balances()`
- `provider::tests::sync_finished_keeps_found_drops_absent`

Verification: `cargo fmt --all`; `cargo check -p platform-wallet --all-features --tests`; full `cargo test -p platform-wallet` suite passing (228 lib + 8 integration); iOS framework build (`./build_ios.sh --target sim`) and SwiftExampleApp simulator build both succeed.

## Breaking Changes

None. Not consensus-breaking — client-side wallet sync/persistence behavior only.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed stale balance entries appearing in wallet after chain resets or when addresses are proven absent from the blockchain.
  * Enhanced address balance synchronization to properly clear and zero outdated cached balances.
  * Improved consistency between found and absent address tracking during wallet sync operations.

### DIFF
--- a/packages/rs-platform-wallet/src/wallet/platform_addresses/provider.rs
+++ b/packages/rs-platform-wallet/src/wallet/platform_addresses/provider.rs
@@ -384,6 +384,14 @@ impl PlatformPaymentAddressProvider {
     ///
     /// Call before each sync round.
     pub(crate) async fn prepare_for_sync(&mut self) -> Result<(), PlatformWalletError> {
+        // Drop any scratch left over from an aborted pass. The SDK only
+        // calls `sync_finished` on success, so a sync that errored after
+        // staging found/absent entries leaves them here — and a stale
+        // staged absent would make the next successful `sync_finished`
+        // remove a committed balance the new pass never re-proved
+        // absent.
+        self.per_wallet_in_sync.clear();
+
         let wallet_ids: Vec<WalletId> = self.per_wallet.keys().copied().collect();
 
         // Refresh provider-level pending and merge in any new
@@ -719,21 +727,24 @@ impl AddressProvider for PlatformPaymentAddressProvider {
         // proven-absent address that lingers there keeps reporting the
         // old balance forever.
         //
-        // Extend-then-remove order: the SDK proves an address either
-        // present or absent in a single pass, never both, so the two sets
-        // are disjoint and the order is academic. We still extend first
-        // and remove second so that, in the impossible-by-contract case
-        // where an address showed up in both, the absent proof wins —
-        // an address gone from state must not stay funded.
+        // The two scratch sets are NOT guaranteed disjoint: the full
+        // scan can prove an address absent at the trunk/branch
+        // checkpoint, then incremental catch-up re-finds it (a credit op
+        // between checkpoint and tip stages a fresh `found` entry
+        // without retracting the earlier absent proof). The catch-up
+        // find reflects the chain tip, so it wins — such addresses are
+        // dropped from the absent set before it drives any removal.
         let drained = std::mem::take(&mut self.per_wallet_in_sync);
         for (wallet_id, wallet_scratch) in drained {
             let Some(wallet_state) = self.per_wallet.get_mut(&wallet_id) else {
                 continue;
             };
-            for (account_index, account_scratch) in wallet_scratch {
+            for (account_index, mut account_scratch) in wallet_scratch {
                 let Some(account_state) = wallet_state.get_mut(&account_index) else {
                     continue;
                 };
+                let PerAccountInSyncPlatformAddressState { found, absent } = &mut account_scratch;
+                absent.retain(|addr| !found.contains_key(addr));
                 account_state.found.extend(account_scratch.found);
                 for absent_addr in &account_scratch.absent {
                     account_state.found.remove(absent_addr);
@@ -797,17 +808,14 @@ mod tests {
     }
 
     /// Build a provider whose committed `per_wallet` tracks a single
-    /// account with one funded address (index 0). No wallet is registered
-    /// in the manager — `sync_finished` and `current_balances` only touch
-    /// the in-memory `per_wallet` / `per_wallet_in_sync` maps.
-    fn provider_with_one_funded_address(
+    /// account on `wallet_id` with one funded address (index 0), backed
+    /// by the supplied wallet manager.
+    fn provider_tracking_address(
+        wallet_manager: Arc<RwLock<WalletManager<PlatformWalletInfo>>>,
+        wallet_id: WalletId,
         addr: PlatformP2PKHAddress,
         f: AddressFunds,
     ) -> PlatformPaymentAddressProvider {
-        let wallet_manager = Arc::new(RwLock::new(WalletManager::<PlatformWalletInfo>::new(
-            Network::Testnet,
-        )));
-
         let mut account_state = PerAccountPlatformAddressState::from_persisted(
             test_xpub(),
             BiBTreeMap::new(),
@@ -819,10 +827,10 @@ mod tests {
         wallet_state.insert(ACCOUNT, account_state);
 
         let mut per_wallet = BTreeMap::new();
-        per_wallet.insert(WALLET, wallet_state);
+        per_wallet.insert(wallet_id, wallet_state);
 
         let mut pending = BiBTreeMap::new();
-        pending.insert((WALLET, ACCOUNT, 0u32), addr);
+        pending.insert((wallet_id, ACCOUNT, 0u32), addr);
 
         PlatformPaymentAddressProvider {
             wallet_manager,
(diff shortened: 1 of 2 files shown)


thepastaclaw:
✅ Review complete (commit fc96e591)

packages/rs-platform-wallet/src/wallet/platform_addresses/provider.rs
```diff
@@ -673,13 +673,58 @@ impl AddressProvider for PlatformPaymentAddressProvider {
...
+            .platform_payment_managed_account_at_index_mut(account_index)
+        else {
+            tracing::warn!(
+                "on_address_absent: no platform payment account {} in wallet {}",
+                account_index,
+                hex::encode(wallet_id)
+            );
+            return;
+        };
+        account.set_address_credit_balance(p2pkh, 0, None);
```
thepastaclaw, line 712: **🟡 Suggestion: New wallet-manager zeroing in `on_address_absent` has no unit test**

The new `wallet_manager.write()` + `set_address_credit_balance(p2pkh, 0, None)` block in `on_address_absent` is the in-memory fix for the spend-path symptom (stale balance offered to identity-creation funding). None of the new tests exercise it: `sync_finished_removes_absent_from_committed_found` and `sync_finished_keeps_found_drops_absent` use `stage_absent` to write directly into `per_wallet_in_sync` precisely because the fixture has no wallet registered in the manager. Add a test that registers a `WalletManager` with a managed platform account, seeds a non-zero `address_balance`, calls `on_address_absent`, and asserts the account's `credit_balance` and `address_balances` map have been zeroed. Otherwise this mirror of `on_address_found` is unverified by the test suite.

<sub>source: ['claude']</sub>

thepastaclaw (changes requested):
## Code Review

The fix addresses the real chain-reset stale-balance bug, but the new `sync_finished` behavior assumes `result.found` and `result.absent` are disjoint per sync. The SDK's full-scan-then-incremental-catch-up flow can violate that: an address proven absent at the trunk checkpoint can be funded again before the tip and surface via `on_address_found` during catch-up. With the new extend-then-remove order, the legitimate new balance is removed from committed `found` and the diff helper also appends a zeroing persistence entry that wins last. The same exposure exists across syncs because `per_wallet_in_sync` is never cleared on a failed prior pass.

<details>

```

- [BLOCKING] packages/rs-platform-wallet/src/wallet/platform_addresses/provider.rs:707-744: Found/absent overlap within a sync silently wipes a freshly funded address
  The new `sync_finished` extends committed `found` with scratch `found` and then unconditionally removes every scratch `absent` address from committed `found`. The doc comment justifies extend-then-remove on the premise that the SDK proves an address either present or absent per pass and the two sets are disjoint, but that does not hold for the full-scan-plus-incremental-catch-up flow. In `packages/rs-sdk/src/platform/address_sync/mod.rs`, `process_trunk_result` / `process_branch_result` insert into `result.absent` and call `on_address_absent` when an address is missing at the trunk/branch checkpoint, while `incremental_catch_up`'s compacted (line 646) and recent (line 704) phases later call `on_address_found` with the new balance whenever a credit op for that address shows up between the checkpoint and tip. Neither catch-up path removes the address from `result.absent` or from the provider's scratch `absent` set. The order trunk-absent then catchup-found is exactly what happens when an address is empty at the checkpoint and gets funded shortly afterwards (very plausible right after a chain reset, which is the regression this PR targets). With the new extend-then-remove, the freshly inserted `found` balance is deleted from committed `found`, so subsequent `current_balances()` and `from_persisted` no longer see it.

  The symmetric defect is in `compute_address_balance_diff` (sync.rs lines 61-83): the absent loop checks only the pre-sync snapshot, so when `(tag, p2pkh)` is in both `result.found` and `result.absent` and the snapshot had non-default funds, the helper first emits the fresh balance entry and then appends a zero entry. `PlatformAddressChangeSet::addresses` is merged in order downstream, so the zero wins and the persister stores 0 for an address that is funded at the sync tip.

  Fix: filter `account_scratch.absent` against `account_scratch.found` before the removal loop, and skip the zero-emission branch in `compute_address_balance_diff` for any `(tag, p2pkh)` that is also in `found`. The contract for the rest of the persistence path should be that an address can never appear in both buckets simultaneously.
- [BLOCKING] packages/rs-platform-wallet/src/wallet/platform_addresses/provider.rs:386-436: Clear `per_wallet_in_sync` at the start of each sync pass
  `per_wallet_in_sync` is per-pass scratch but `prepare_for_sync` only refreshes `self.pending` and clears the committed `account_state.absent`; it never touches the scratch map. The SDK calls `provider.sync_finished()` only on the success path (`packages/rs-sdk/src/platform/address_sync/mod.rs` line 415), so if a prior `sync_address_balances` returns an error after `on_address_absent` has staged an address, that scratch absent entry survives into the next sync. With this PR the next successful `sync_finished` will now remove that address from committed `found` even if the new pass never re-proved it absent — and if the new pass legitimately found it via catch-up, the freshly committed balance is wiped. Before this PR the leftover scratch was harmless because `sync_finished` only overwrote committed `absent` wholesale; the new destructive removal raises the cost of any aborted sync.

  Fix: `self.per_wallet_in_sync.clear();` at the top of `prepare_for_sync` before refreshing pending.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/platform_addresses/provider.rs:685-704: New wallet-manager zeroing in `on_address_absent` has no unit test
  The new `wallet_manager.write()` + `set_address_credit_balance(p2pkh, 0, None)` block in `on_address_absent` is the in-memory fix for the spend-path symptom (stale balance offered to identity-creation funding). None of the new tests exercise it: `sync_finished_removes_absent_from_committed_found` and `sync_finished_keeps_found_drops_absent` use `stage_absent` to write directly into `per_wallet_in_sync` precisely because the fixture has no wallet registered in the manager. Add a test that registers a `WalletManager` with a managed platform account, seeds a non-zero `address_balance`, calls `on_address_absent`, and asserts the account's `credit_balance` and `address_balances` map have been zeroed. Otherwise this mirror of `on_address_found` is unverified by the test suite.
```
</details>

packages/rs-platform-wallet/src/wallet/platform_addresses/provider.rs
QuantumExplorer: Added `on_address_absent_zeroes_managed_account_balance` in fc96e591 — registers a real WalletManager with a managed platform account, seeds the stale balance, calls `on_address_absent`, and asserts the in-memory balance is zeroed and the scratch absent entry is staged.

thepastaclaw (approved):
## Code Review

All three prior blocking/suggestion findings are fixed at fc96e591: sync_finished now dedupes found-over-absent before applying removals; compute_address_balance_diff skips zero-emission for addresses also in found; prepare_for_sync clears per-pass scratch up front; and a new test exercises on_address_absent against a real wallet manager. Six agents (claude+codex across general/rust-quality/ffi) converge on no new in-scope defects in either the latest delta or the cumulative PR. Approving.

PR #3859, opened 2026-06-11
fix(drive): verify identity-create-from-shielded-pool results without unbounded terminal-key queries

QuantumExplorer:

On devnet (paloma, drive 4.0.0-rc.1, 2026-06-11), fetching the execution-result proof for an `IdentityCreateFromShieldedPool` (state transition type 20) always failed with:

```
Drive error: grovedb: query error: not supported error: terminal keys are not supported with unbounded ranges
```

even though the transition executed successfully on-chain (pool balance dropped by the denomination, notes +2, chain kept advancing). The client's `broadcast_and_wait::<StateTransitionProofResult>` therefore reported the registration as failed and never persisted the new identity — leaving an orphaned-but-live identity.

**Root cause** (client-side verification, not proof generation): the type-20 verifier arm rebuilds the merged {nullifiers, full-identity} query and verified it with `GroveDb::verify_query_with_absence_proof`. That entry point enumerates the query's *terminal keys* to synthesize absence entries — which is impossible for `full_identity_query`, whose all-keys sub-query is an unbounded `RangeFull` over the identity keys tree, so grovedb rejects the query outright for every honest proof. The other shielded spend types (unshield/transfer/withdrawal) only merge explicit-key sub-queries, which is why their result proofs round-tripped fine; the component unique to type 20 is the full-identity query.

## What was done?

- The type-20 arm of `verify_state_transition_was_executed_with_proof` now verifies the same merged query with strict `GroveDb::verify_query` (succinctness on, no absence synthesis). Absence entries were never needed: every queried element (the spent nullifiers and the created identity) must be PRESENT, so each nullifier's spend status is now derived from membership in the proved result set, with an explicit rejection of unrequested nullifier entries. The succinctness check still rejects proofs padded with branches beyond {nullifiers, identity}, preserving the strict guarantee from #3812.
- The prove side is unchanged (it already generated valid proofs); only its comment was updated.
- Added a drive-abci regression test (`executed_transition_result_proof_roundtrips`) that executes a type-20 success action through the real `execute_event` path (synthetic action — no Halo 2 proving), commits, generates the result proof with `prove_state_transition`, and verifies it — failing with the exact devnet error before the fix, passing after.

## How Has This Been Tested?

- New regression test fails pre-fix with `GroveDB(QueryError(NotSupported("terminal keys are not supported with unbounded ranges")))` and passes post-fix (verified both ways by stashing/restoring the fix).
- `cargo test -p drive --features full,verify --lib verify_state_transition_was_executed_with_proof` — 61 passed.
- `cargo test -p drive-abci --lib shielded` — 144 passed, 0 failed (includes the real-Orchard prove/verify round-trips for the other shielded types).
- `cargo check -p drive --no-default-features --features verify` (verify-only build), `cargo check -p drive --all-features --tests`, `cargo check -p drive-abci --all-features --tests`, `cargo fmt --all -- --check`, clippy clean on the changed code.

## Breaking Changes

None — this is a node/SDK query-proof verification fix; no consensus or wire-format change.

## Checklist:

---

**Follow-up note (not fixed here):** SwiftExampleApp misattributes this broadcast-result error to the "Generating Halo 2 proof" progress step in the registration progress view (`packages/swift-sdk/SwiftExampleApp`) — the proof had already been generated and the transition broadcast/executed by the time this error fires. Worth a separate UI fix so result-proof failures surface as their own step.

## Summary by CodeRabbit

* **Tests**
  * Added regression test for state transition result-proof roundtrip verification in identity creation.

* **Bug Fixes**
  * Improved verification logic for identity creation to use stricter query validation.

* **Documentation**
  * Enhanced comments clarifying proof verification behavior and requirements.

### DIFF
--- a/packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/identity_create_from_shielded_pool/tests.rs
+++ b/packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/identity_create_from_shielded_pool/tests.rs
@@ -11,11 +11,15 @@
 //!   path (the fallback Unshield, pool->address minus penalty): the converter ops applied through a
 //!   real Drive keep `calculate_total_credits_balance().ok()` balanced (the end-of-block invariant
 //!   that halts the chain) — the regression guard for the `AddToSystemCredits` over-mint.
+//! - The execute -> prove -> verify result-proof roundtrip for the success path (synthetic action,
+//!   no Halo 2 proving): the regression guard for the devnet bug where the verifier's absence-proof
+//!   verification choked on the merged {nullifiers, full-identity} query ("terminal keys are not
+//!   supported with unbounded ranges"), so every honest type-20 result proof failed client-side.
 //!
-//! The full build->prove->execute->prove/verify happy path (real Orchard proof + the strict merged
-//! nullifier+identity proof roundtrip) is deferred to the shared shielded-strategy harness, a
-//! pre-existing repo-wide TODO that is disabled for every shielded transition (the shielded
-//! `OperationType` build handlers are commented out in `strategy.rs`).
+//! The full build->prove->execute->prove/verify happy path (real Orchard proof) is deferred to the
+//! shared shielded-strategy harness, a pre-existing repo-wide TODO that is disabled for every
+//! shielded transition (the shielded `OperationType` build handlers are commented out in
+//! `strategy.rs`).
 
 use super::state::v0::IdentityCreateFromShieldedPoolStateTransitionStateValidationV0;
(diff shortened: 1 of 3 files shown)


thepastaclaw:
✅ Review complete (commit 65599dfc)

QuantumExplorer (commented):
Reviewed

thepastaclaw (approved):
## Code Review

Narrowly scoped client-side verifier fix for type-20 (IdentityCreateFromShieldedPool) result proofs: swaps `verify_query_with_absence_proof` for `verify_query` because the embedded full-identity query has an unbounded all-keys range that the absence-proof variant cannot enumerate. Strict succinctness is preserved, spent-status is derived from membership in the proved-present set, and a regression test was added. All six agents converged on no in-scope findings.

PR #3860, opened 2026-06-11
fix(platform): derive identity-rescan keys through the wallet signer

QuantumExplorer:

SwiftExampleApp → Identities → Add → "Re-scan for Identities" → pick a mnemonic-restored wallet → "Re-scan Wallet" failed immediately with:

> Couldn't derive identity-authentication keys: Invalid identity data: Failed to derive authentication key: Invalid parameter: External signable wallet has no private key

and discovered 0 identities.

App wallets are loaded into the in-process `WalletManager` as `key_wallet::WalletType::ExternalSignable` — the seed lives in iOS Keychain, not in process. Both the discovery scan (`IdentityWallet::discover`) and its key preview (`platform_wallet_preview_identity_registration_keys`) derived DIP-9 identity-auth keys directly from that in-memory wallet, which holds no private key, so they bottomed out in `root_extended_priv_key()` returning `External signable wallet has no private key`.

Identity *registration* on the same wallets already works because its key derivation never touches the in-memory wallet — it resolves the mnemonic on demand through a Swift-owned `MnemonicResolver` (Keychain-backed), builds a master `ExtendedPrivKey`, and derives from it. This PR routes the rescan / preview through that same mechanism.

## What was done?

**`packages/rs-platform-wallet`**
- `identity_handle.rs`: added `derive_identity_auth_key_hash_from_master`, a master-xpriv sibling of `derive_identity_auth_key_hash` that goes through `derive_ecdsa_identity_auth_keypair_from_master` + `ripemd160_sha256`, so the rescan and registration derivations can never drift.
- `discovery.rs`: factored the gap-limit scan into one shared `discover_inner` with a pluggable `KeyHashSource` (resident wallet vs. supplied master) and added the public `discover_from_master`. `discover()` behavior for key-resident wallets is unchanged; the per-index read lock is now only taken on the wallet-internal derive path. `load_identity_by_index` (same latent flaw) was intentionally left untouched.

**`packages/rs-platform-wallet-ffi`** (signatures changed in place — pre-release)
- `platform_wallet_discover_identities` and `platform_wallet_preview_identity_registration_keys` now take a `MnemonicResolverHandle`. When non-null they resolve the mnemonic keyed by the wallet handle's own `wallet_id` (self-pinned, same rationale as `dash_sdk_derive_identity_key_at_slot_with_resolver`), build the master, and derive via `discover_from_master` / `derive_ecdsa_identity_auth_keypair_from_master`; when null they keep the resident-wallet behavior for key-resident consumers.
- Resolver failures map to distinct NOT_FOUND / BUFFER_TOO_SMALL / other messages. The resolve→parse→seed→master sequence is shared in a new `resolve_master_from_resolver` helper so both entry points stay in lockstep. Mnemonic/seed live in `Zeroizing` buffers; the master's scalar is `non_secure_erase`d before return (`ExtendedPrivKey` has no `Drop`).

**`packages/swift-sdk`**
- `ManagedPlatformWallet.discoverIdentities` / `previewIdentityRegistrationKeys` construct a `MnemonicResolver` (default `WalletStorage`, overridable for tests) and pass its handle to the updated FFI. `discoverIdentities` runs the FFI inside `Task.detached`, so the resolver is captured in the closure and the FFI call is wrapped in `withExtendedLifetime` to keep ARC holding it for the call's duration. Public signatures stay source-compatible (defaulted params only) — the example app's `SearchWalletsForIdentitiesView` compiles unchanged. Swift stays a thin bridge; no derivation/policy moved into Swift (per `swift-sdk/CLAUDE.md`).

## How Has This Been Tested?

- `cargo fmt --all`; `cargo clippy` clean for both `platform-wallet` and `platform-wallet-ffi`.
- `cargo test -p platform-wallet --all-features`: **223 passed, 0 failed** (220 prior + 3 new). New `identity_handle::tests` pin the fix for the "abandon … about" vector:
  - `master_hash_matches_resident_wallet_hash` — master-based hash equals the resident-wallet `derive_identity_auth_key_hash` on a `Mnemonic` wallet across slots/networks.
  - `external_signable_errors_but_master_succeeds` — the resident derive errors on an `ExternalSignable` wallet while the master derive succeeds and reproduces the key-resident hash.
  - `master_hash_matches_registration_keypair_pubkey_hash` — parity with registration's in-creation key #0.
- `cargo test -p platform-wallet-ffi --all-features`: **all passing** (100 + 26 + 5/9, 4 ignored).
- `cargo check --workspace --all-targets`: clean.
- `cd packages/swift-sdk && ./build_ios.sh --target sim --profile release`: succeeded; the regenerated `platform-wallet-ffi.h` shows the new `MnemonicResolverHandle *mnemonic_resolver_handle` parameter on both functions.
- `xcodebuild -project SwiftExampleApp/SwiftExampleApp.xcodeproj -scheme SwiftExampleApp -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 16'`: **BUILD SUCCEEDED** with `SearchWalletsForIdentitiesView` unchanged.

## Breaking Changes

None that warrant a `!` type. The two FFI symbols gained a resolver parameter (pre-release SDK/FFI churn, not consensus-breaking); the only callers are the two Swift SDK methods, whose public signatures remain source-compatible via defaulted parameters.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Identity discovery and registration key preview now support external-signable and watch-only wallets in addition to resident-key wallets.
  * Methods for identity operations now accept wallet storage parameters for enhanced configuration flexibility.

* **Tests**
  * Added validation tests for identity key derivation consistency across wallet types and networks.

### DIFF
--- a/packages/rs-platform-wallet-ffi/src/identity_key_preview.rs
+++ b/packages/rs-platform-wallet-ffi/src/identity_key_preview.rs
@@ -26,16 +26,26 @@
 //! The Swift caller knows nothing about any of those — it just reads
 //! the array back out.
 //!
-//! # Key source: resolver vs. resident wallet
+//! # Key source: chosen by wallet capability
 //!
-//! App wallets are loaded into the in-process `WalletManager` as
-//! `key_wallet::WalletType::ExternalSignable` — their seed lives in iOS
-//! Keychain, NOT in process — so deriving the preview rows from the
-//! resident wallet fails with `External signable wallet has no private
-//! key` (the bug this entry point's resolver path fixes).
+//! The derivation source is selected by the in-process wallet's shape,
+//! NOT by whether a resolver handle was supplied — the resolver is a
+//! *capability* the Rust side consults only when it can't derive
+//! locally, not a command that forces the resolver path:
 //!
-//! - **`mnemonic_resolver_handle` non-null:** resolve the wallet's
-//!   mnemonic on demand via the Swift-owned
+//! - **In-process wallet holds resident private keys** (`WalletType::
+//!   Mnemonic` / `Seed` / `ExtendedPrivKey` — NOT external-signable and
+//!   NOT watch-only): derive every row from the resident wallet under a
+//!   single read lock held for the loop's duration (the historical
+//!   path). The resolver handle is never touched, which also skips a
+//!   pointless iOS Keychain read. This keeps `createWallet(seed:)` /
+//!   raw-seed wallets working even when no BIP-39 mnemonic was ever
+//!   persisted to `WalletStorage`.
+//! - **In-process wallet is external-signable / watch-only:** its seed
+//!   lives in iOS Keychain, NOT in process, so the resident derive
+//!   would fail with `External signable wallet has no private key`. In
+//!   that case, if `mnemonic_resolver_handle` is non-null, resolve the
+//!   wallet's mnemonic on demand via the Swift-owned
 //!   [`MnemonicResolverHandle`] (keyed by the wallet handle's own
 //!   `wallet_id`), build the master `ExtendedPrivKey`, and derive each
 //!   row from that master via
@@ -44,10 +54,10 @@
 //!   The mnemonic / seed / master scalar live in `Zeroizing` buffers
 //!   (the master's `private_key` is explicitly `non_secure_erase`d —
 //!   `ExtendedPrivKey` has no `Drop`) and are scrubbed before this
-//!   function returns. This is the path the iOS app takes.
-//! - **`mnemonic_resolver_handle` null:** derive the rows from the
-//!   resident in-process wallet (the historical path; valid only for
-//!   wallets with in-memory key material).
+//!   function returns. This is the path the iOS app takes. If the
+//!   resolver is null for such a wallet, the call returns an error
+//!   hinting that a mnemonic resolver handle is required for this
+//!   wallet shape.
 
 use std::ffi::CString;
 use std::os::raw::c_char;
@@ -55,6 +65,7 @@ use std::ptr;
 
 use dashcore::PrivateKey as DashPrivateKey;
 use key_wallet::bip32::ExtendedPrivKey;
+use key_wallet::Wallet;
 use platform_wallet::wallet::identity::network::derive_ecdsa_identity_auth_keypair_from_master;
 use platform_wallet::{derive_identity_auth_keypair, IDENTITY_GAP_LIMIT, MASTER_KEY_INDEX};
 use zeroize::Zeroizing;
@@ -151,17 +162,26 @@ impl IdentityKeyPreviewsFFI {
 /// keypairs this wallet would probe during a discovery scan,
 /// starting at identity index `start_index`.
 ///
+/// The derivation source is chosen by the in-process wallet's
+/// capability (see the module docs): resident-key wallets derive
+/// locally and never touch the resolver; external-signable / watch-only
+/// wallets consult the resolver. The resolver is only *needed* for the
+/// latter shape.
+///
 /// # Parameters
 /// - `wallet_handle` — platform-wallet handle.
 /// - `mnemonic_resolver_handle` — Swift-owned
-///   [`MnemonicResolverHandle`]. When **non-null**, the wallet's
-///   mnemonic is resolved on demand (keyed by the wallet handle's own
-///   `wallet_id`), a master `ExtendedPrivKey` is built, and each row is
-///   derived from that master — the path that works for the iOS
-///   Keychain-backed `WalletType::ExternalSignable` shape whose seed is
-///   not in process. When **null**, the rows are derived from the
-///   resident in-process wallet (the historical path; valid only for
-///   wallets with in-memory key material).
+///   [`MnemonicResolverHandle`], consulted **only** when the in-process
+///   wallet lacks resident private keys (external-signable / watch-only
+///   — the iOS Keychain-backed `WalletType::ExternalSignable` shape
+///   whose seed is not in process). For such a wallet, when non-null
+///   the mnemonic is resolved on demand (keyed by the wallet handle's
+///   own `wallet_id`), a master `ExtendedPrivKey` is built, and each row
+///   is derived from that master; when null the call errors with a hint
+///   that a resolver handle is required for this wallet shape. For a
+///   wallet that holds resident private keys this argument is ignored
+///   and the rows are derived from the in-process wallet (the
+///   historical path).
 /// - `start_index` — first identity index to derive.
 /// - `count_or_neg1` — number of consecutive identity indices to
 ///   derive. Pass `< 0` to use the Rust default
@@ -263,16 +283,31 @@ pub unsafe extern "C" fn platform_wallet_preview_identity_registration_keys(
             })
         };
 
-        // Derive one row's material from either the resolved master or
-        // the resident wallet, depending on which key source is active.
+        // Borrowed per-row key source. The two derivation paths are
+        // symmetric in shape: both hand `derive_material` a borrowed
+        // source and get back `RowMaterial`. The resident-wallet
+        // variant borrows the `&Wallet` looked up ONCE before the loop
+        // (under a read guard held for the loop's whole duration — the
+        // derive is pure compute, matching pre-PR behavior); the master
+        // variant borrows the resolved master xpriv.
+        enum DeriveSource<'a> {
+            /// In-process wallet holds resident private keys — derive
+            /// each row directly from it (no per-row lock acquisition).
+            Resident(&'a Wallet),
+            /// External-signable / watch-only wallet — derive each row
+            /// from the master xpriv resolved from the wallet's mnemonic.
+            Master(&'a ExtendedPrivKey),
+        }
+
+        // Derive one row's material from the active borrowed key source.
         let derive_material = |identity_index: u32,
-                               master: Option<&ExtendedPrivKey>|
+                               source: &DeriveSource|
          -> Result<RowMaterial, PlatformWalletFFIResult> {
-            match master {
-                Some(master) => {
-                    // Resolver path: pure derive from the resolved master,
-                    // identical to the registration / rescan-via-resolver
-                    // derive.
+            match source {
+                DeriveSource::Master(master) => {
+                    // External-signable / watch-only path: pure derive
+                    // from the resolved master, identical to the
+                    // registration / rescan-via-resolver derive.
                     let derived = derive_ecdsa_identity_auth_keypair_from_master(
                         master,
                         network,
@@ -285,18 +320,11 @@ pub unsafe extern "C" fn platform_wallet_preview_identity_registration_keys(
                         private_key: derived.private_key,
                     })
                 }
-                None => {
+                DeriveSource::Resident(key_wallet) => {
                     // Resident-wallet path: derive from the in-process
-                    // wallet under a read lock. Errors for
-                    // `WalletType::ExternalSignable` — that shape must
-                    // pass a resolver handle.
-                    let wm = wallet.wallet_manager().blocking_read();
-                    let key_wallet = wm.get_wallet(&wallet.wallet_id()).ok_or_else(|| {
-                        PlatformWalletFFIResult::err(
-                            PlatformWalletFFIResultCode::ErrorInvalidHandle,
-                            "Wallet not found in wallet manager",
-                        )
-                    })?;
+                    // wallet. The read guard + `&Wallet` were acquired
+                    // once before the loop (see below) so this is a
+                    // pure secp256k1 pass with no per-row locking.
                     let (path, ext_priv, public_key) = derive_identity_auth_keypair(
                         key_wallet,
                         network,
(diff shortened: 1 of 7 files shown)


thepastaclaw:
✅ Review complete (commit 74eeb25e)

packages/rs-platform-wallet-ffi/src/identity_key_preview.rs
```diff
@@ -164,77 +205,166 @@ pub unsafe extern "C" fn platform_wallet_preview_identity_registration_keys(
...
+                        identity_index,
+                        MASTER_KEY_INDEX,
+                    )?;
+                    Ok(RowMaterial {
+                        path: path.to_string(),
+                        public_key: public_key.serialize(),
+                        private_key: Zeroizing::new(ext_priv.private_key.secret_bytes()),
+                    })
+                }
+            }
```
thepastaclaw: **🟡 Suggestion: Resident-wallet preview path acquires WalletManager read lock per row**

Before this PR, `platform_wallet_preview_identity_registration_keys` took `wallet.wallet_manager().blocking_read()` and resolved `key_wallet` once at the top of the closure, then reused them for every preview row. After the refactor, the resident-wallet branch of `derive_material` runs per iteration and re-acquires `blocking_read()` plus re-looks-up `get_wallet(&wallet.wallet_id())` on every row. The resolver path correctly hoists the master xpriv once outside the loop, so the resident path is now strictly worse than the original with no upside on the resolver side. Hoist the resident-path lock acquisition (and the `get_wallet` lookup) before the loop — matching how `master_opt` is resolved once — and let `derive_material` consume a borrowed `&Wallet`/`KeyHashSource`-shaped enum per row. Bonus: keeps the two paths symmetric in shape.

<sub>source: ['claude']</sub>

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/ManagedPlatformWallet.swift
```diff
@@ -1138,38 +1160,68 @@ extension ManagedPlatformWallet {
...
-                ids.append(data)
+            try withExtendedLifetime(resolver) {
+                var found = DiscoveredIdentityIdsFFI()
+                let result = platform_wallet_discover_identities(
+                    handle,
+                    resolver.handle,
+                    startArg,
+                    gapArg,
+                    &found
+                )
```
thepastaclaw, line 1229: **🟡 Suggestion: Swift wrapper forces every wallet through the resolver, regressing resident-Mnemonic flows**

The Rust FFI deliberately keeps `mnemonic_resolver_handle` nullable so key-resident wallets keep the historical resident-wallet derivation. Both Swift wrappers (`discoverIdentities` and `previewIdentityRegistrationKeys`) unconditionally construct a `MnemonicResolver` and pass `resolver.handle`, so the null-handle fallback is unreachable from Swift. Concrete regressions: `PlatformWalletManager.createWallet(seed:)` (no BIP-39 mnemonic exists to resolve) and `createWallet(mnemonic:)` callers that do not separately persist the mnemonic via `WalletStorage` — discovery used to derive from the in-process wallet and now fails with resolver `NOT_FOUND`. If the design intent really is "iOS Keychain-backed only," surface that intent by either documenting the new precondition, falling back to a null resolver when no mnemonic is stored, or offering an explicit resident-wallet variant. Otherwise the public API silently loses a previously-working path.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):
## Code Review

PR is well-scoped: it routes identity rescan/preview key derivation through the existing Swift-owned mnemonic resolver so iOS Keychain-backed `ExternalSignable` wallets can derive DIP-9 identity-auth keys without an in-process seed. One blocking issue: the new synchronous preview Swift wrapper does not extend the `MnemonicResolver`'s lifetime across the FFI call (the discovery path does), creating a real use-after-free window because the Rust ctx is `passUnretained`. Two convergent suggestions worth addressing: the preview's resident-wallet branch now takes the `wallet_manager` read lock per row instead of once, and the Swift wrapper unconditionally forces every wallet through the resolver — a regression for resident-Mnemonic wallets created with `createWallet(seed:)` or without a Keychain-stored mnemonic.

<details>

```

In `packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/ManagedPlatformWallet.swift`:
- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/ManagedPlatformWallet.swift:791-798: Keep mnemonic resolver alive across the preview FFI call
  `MnemonicResolver` builds its Rust callback context with `Unmanaged.passUnretained(self)` and frees the underlying handle in `deinit` — Rust only ever sees a raw pointer back to the Swift object, so ARC must keep the resolver alive for the whole synchronous FFI call. In this new preview path, the last Swift-level use of `resolver` is just evaluating `resolver.handle` as a parameter, after which ARC is free to end the object's lifetime. If the resolver is released before or during the call, Rust dereferences a freed `MnemonicResolverHandle` (or invokes the destroyed callback context inside `resolve_master_from_resolver`), causing FFI use-after-free / crash / potential memory corruption on a user-triggered preview. The async `discoverIdentities` path in this same PR already wraps the equivalent call in `withExtendedLifetime(resolver)` for exactly this reason; the synchronous preview needs the same guard.
- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/ManagedPlatformWallet.swift:1199-1210: Swift wrapper forces every wallet through the resolver, regressing resident-Mnemonic flows
  The Rust FFI deliberately keeps `mnemonic_resolver_handle` nullable so key-resident wallets keep the historical resident-wallet derivation. Both Swift wrappers (`discoverIdentities` and `previewIdentityRegistrationKeys`) unconditionally construct a `MnemonicResolver` and pass `resolver.handle`, so the null-handle fallback is unreachable from Swift. Concrete regressions: `PlatformWalletManager.createWallet(seed:)` (no BIP-39 mnemonic exists to resolve) and `createWallet(mnemonic:)` callers that do not separately persist the mnemonic via `WalletStorage` — discovery used to derive from the in-process wallet and now fails with resolver `NOT_FOUND`. If the design intent really is "iOS Keychain-backed only," surface that intent by either documenting the new precondition, falling back to a null resolver when no mnemonic is stored, or offering an explicit resident-wallet variant. Otherwise the public API silently loses a previously-working path.

In `packages/rs-platform-wallet-ffi/src/identity_key_preview.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/identity_key_preview.rs:288-312: Resident-wallet preview path acquires WalletManager read lock per row
  Before this PR, `platform_wallet_preview_identity_registration_keys` took `wallet.wallet_manager().blocking_read()` and resolved `key_wallet` once at the top of the closure, then reused them for every preview row. After the refactor, the resident-wallet branch of `derive_material` runs per iteration and re-acquires `blocking_read()` plus re-looks-up `get_wallet(&wallet.wallet_id())` on every row. The resolver path correctly hoists the master xpriv once outside the loop, so the resident path is now strictly worse than the original with no upside on the resolver side. Hoist the resident-path lock acquisition (and the `get_wallet` lookup) before the loop — matching how `master_opt` is resolved once — and let `derive_material` consume a borrowed `&Wallet`/`KeyHashSource`-shaped enum per row. Bonus: keeps the two paths symmetric in shape.
```
</details>

packages/rs-platform-wallet-ffi/src/identity_key_preview.rs
QuantumExplorer: Done — the wallet-manager read guard and `get_wallet` lookup are now acquired once before the row loop (and held for its duration, matching pre-PR behavior); the per-row derive consumes a borrowed key source (`DeriveSource::Resident(&Wallet)` / `DeriveSource::Master(&ExtendedPrivKey)`) so no per-row lock is taken.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/ManagedPlatformWallet.swift
QuantumExplorer: Done — source selection now happens Rust-side by wallet capability rather than being forced by Swift. Resident-key wallets (not external-signable, not watch-only) keep the historical in-process derive and the resolver is never consulted; it is only used for external-signable / watch-only shapes (null resolver there now returns an enriched error). So `createWallet(seed:)` / raw-seed flows are unaffected even when no mnemonic was persisted to `WalletStorage`.

packages/rs-platform-wallet-ffi/src/identity_key_preview.rs
```diff
@@ -164,77 +225,228 @@ pub unsafe extern "C" fn platform_wallet_preview_identity_registration_keys(
...
+            // `source` borrows `master_opt`, so drop it first to release
+            // that borrow before we mutate the master.
+            drop(source);
+            if let Some(mut master) = master_opt {
+                master.private_key.non_secure_erase();
+            }
+            // `wm` (the read guard) drops here at the closure's end,
+            // releasing the wallet-manager read lock held across the loop.
+            loop_result
+        })();
```
thepastaclaw: **🟡 Suggestion: Preview path holds WalletManager read guard across the Swift resolver callback**

The preview closure acquires `wm = wallet.wallet_manager().blocking_read()` at line 353 and only releases it at the closure's end (line 446). For the external-signable branch this means the guard is held across `resolve_master_from_resolver(...)` at line 391 — a synchronous re-entry into Swift that performs an iOS Keychain read (and can stall on biometric unlock) — and across the entire per-row derive loop. The sibling discovery path in `identity_discovery.rs:165-178` deliberately scopes the read guard to a tight block that only inspects the wallet's shape (`is_external_signable` / `is_watch_only`), drops it before calling the resolver, and never crosses `block_on_worker`. Only the resident branch genuinely needs the guard for the loop's duration (it borrows `&Wallet`); the master branch does not need the guard past the capability check and the `wallet_id` / `network` reads. Today no deadlock occurs because the Swift resolver does not touch the wallet manager, but the wider hold window blocks concurrent wallet-manager writers behind a synchronous Keychain round-trip and introduces a non-obvious cross-FFI reentrancy invariant that the discovery path already avoids. Mirror the discovery shape: capture `wallet_has_resident_keys`, `network`, and `wallet_id` under the short guard, drop `wm`, then resolve the master and either re-acquire the guard for the resident branch's loop or hand the master to the loop directly.

thepastaclaw (commented):
## Code Review

All three prior findings (preview FFI resolver lifetime, per-row WalletManager lock in resident preview, Swift forcing the resolver path on resident wallets) are FIXED on head 2a7d0964. The capability-based dispatch in Rust (is_external_signable || is_watch_only) is the correct invariant, and resident Mnemonic/Seed/ExtendedPrivKey wallets now skip the resolver entirely. One new minor finding: the preview path holds the wallet-manager read guard across the Swift mnemonic-resolver callback (and the full row loop), unlike the discovery path which deliberately scopes the guard to the capability check.

🟡 1 suggestion(s)

<details>

```

In `packages/rs-platform-wallet-ffi/src/identity_key_preview.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/identity_key_preview.rs:353-447: Preview path holds WalletManager read guard across the Swift resolver callback
  The preview closure acquires `wm = wallet.wallet_manager().blocking_read()` at line 353 and only releases it at the closure's end (line 446). For the external-signable branch this means the guard is held across `resolve_master_from_resolver(...)` at line 391 — a synchronous re-entry into Swift that performs an iOS Keychain read (and can stall on biometric unlock) — and across the entire per-row derive loop. The sibling discovery path in `identity_discovery.rs:165-178` deliberately scopes the read guard to a tight block that only inspects the wallet's shape (`is_external_signable` / `is_watch_only`), drops it before calling the resolver, and never crosses `block_on_worker`. Only the resident branch genuinely needs the guard for the loop's duration (it borrows `&Wallet`); the master branch does not need the guard past the capability check and the `wallet_id` / `network` reads. Today no deadlock occurs because the Swift resolver does not touch the wallet manager, but the wider hold window blocks concurrent wallet-manager writers behind a synchronous Keychain round-trip and introduces a non-obvious cross-FFI reentrancy invariant that the discovery path already avoids. Mirror the discovery shape: capture `wallet_has_resident_keys`, `network`, and `wallet_id` under the short guard, drop `wm`, then resolve the master and either re-acquire the guard for the resident branch's loop or hand the master to the loop directly.
```
</details>

packages/rs-platform-wallet-ffi/src/identity_key_preview.rs
QuantumExplorer: Done — the wallet-manager read guard is now scoped to the capability check and dropped before the resolver callback; the resident branch re-acquires it only for the derive loop, mirroring the discovery path so the guard is never held across the synchronous Swift `resolve_master_from_resolver` (Keychain) re-entry.

thepastaclaw (approved):
## Code Review

Incremental + cumulative verification at head 74eeb25e. The prior locking concern is FIXED: the preview path now implements two-phase locking — a short capability-check read guard (lines 370-383) is dropped before the Swift mnemonic resolver call (lines 408-410), and the read guard is only re-acquired for the resident-derive loop (lines 425-439), matching the discovery-path invariant. No new in-scope defects identified across the FFI, Rust quality, security, and FFI-engineer perspectives. The codex-security-auditor finding about an empty SetPrices schedule panic touches an unmodified file (rs-drive token direct-purchase transformer) and is unrelated to this PR's identity-rescan/preview scope, so it is dropped.

PR #3862, opened 2026-06-11
fix(swift-sdk): attribute shielded registration errors to the right step and keep unconfirmed broadcasts safe

QuantumExplorer:

The SwiftExampleApp shielded (Type-20 `IdentityCreateFromShieldedPool`) registration progress view misattributed broadcast-result errors to the "Generating Halo 2 proof" step. Observed live on devnet: a type-20 transition executed successfully on-chain, but the client's `broadcast_and_wait` failed while fetching/verifying the execution-result proof (server side since fixed in #3859). The app showed the registration as failed at the proof step — long after the proof had been generated and the transition broadcast and executed.

Worse than the cosmetic misattribution: on any error the wallet released the spent notes' reservations (`cancel_pending`) and the app treated the identity as unregistered, even though it could be live on-chain — the orphaned-identity hazard. Re-registering the same keys would fail the registered-key-hash stateful check and burn the funds to the fallback address minus a penalty.

## What was done?

The whole flow previously ran as one opaque `broadcast_and_wait`, collapsing build, broadcast, and result-confirmation failures into a single `ShieldedBroadcastFailed` error. It is now staged so each failure class is handled on its own terms:

- **rs-sdk**: new `identity_create_from_shielded_pool_transition` trait method (build + structural validation only); the existing one-shot helper is reimplemented on top of it, so other callers see no behavior change.
- **platform-wallet** (`operations.rs::identity_create_from_shielded_pool`): broadcasts and waits as separate stages.
  - Build errors and broadcast/relay rejections behave as before (reservation released).
  - A wait-stage `StateTransitionBroadcastError` (Platform definitively reported the transition's execution error) also behaves as before.
  - Any **other** wait-stage error (result-proof fetch/verify failure, timeout, transport — the #3859 shape) is ambiguous: the wallet retries fetching the identity by its pre-derived id (4 attempts, ~3 s apart). If found, the flow completes as a normal success (local registration + persister row). If still unknown, it returns the new `ShieldedBroadcastUnconfirmed { identity_id, reason }` error and **leaves the note reservations in place** — `pending_nullifiers` is in-memory only, so the next nullifier sync promotes them to spent if the transition landed, and an app restart frees them if it didn't. Releasing them eagerly would invite double-spend attempts against possibly-consumed notes.
- **rs-platform-wallet-ffi**: new result codes `ErrorShieldedBroadcastFailed = 16` (definitive rejection, notes released, safe to retry) and `ErrorShieldedBroadcastUnconfirmed = 17` (broadcast accepted, result unconfirmed; uniquely, `out_identity_id` is also written on this code so the host can hold the slot).
- **Swift SDK**: mirrors the two codes; `shieldedIdentityCreateFromPool` throws a typed `ShieldedIdentityCreateUnconfirmedError` carrying the derived id on code 17.
- **SwiftExampleApp**:
  - `IdentityRegistrationController` gains an `.unconfirmed(identityId:message:)` phase that is *active* (holds the identity slot; `submit()` and the coordinator refuse re-entry) plus a `FailureStage` for step attribution; `observeController` marks the HD slot used on `.unconfirmed`.
  - The shielded progress step set grows from 4 to 5 with a distinct **"Waiting for platform confirmation"** step. Confirmation failures render an orange warning on that step (steps 1–3 done, register pending) with a "Broadcast succeeded — confirmation pending" banner instead of the red failure; broadcast rejections mark the broadcast step; only genuine pre-broadcast errors land on the Halo 2 step.
  - Pending Registrations renders `.unconfirmed` in orange ("Confirmation pending") and keeps it until the user dismisses it; the coordinator retains it like `.failed` rather than purging it like `.completed`.

All classification, fetch-retry, and reservation logic lives in Rust behind the FFI; the Swift layer only maps the typed error to presentation state, per the swift-sdk layering rules.

## How Has This Been Tested?

- `cargo check --all-targets --all-features --tests` on `dash-sdk`, `platform-wallet`, `platform-wallet-ffi`; `cargo clippy` clean on all three; `cargo fmt --all`.
- `cargo test -p platform-wallet --features shielded --lib`: 220 passed. FFI error-mapping unit tests pass.
- New unit test asserts `Phase.unconfirmed.isActive == true` (slot held against re-submission).
- iOS xcframework rebuilt (regenerated header carries codes 16/17); SwiftExampleApp builds clean for the iPhone 16 simulator, including the test target.

## Breaking Changes

None. The FFI function signature is unchanged; `out_identity_id` gains a documented additional write on the new unconfirmed result code only. The one-shot rs-sdk helper keeps its exact semantics.

## Checklist:

### DIFF
--- a/packages/rs-platform-wallet/src/wallet/shielded/operations.rs
+++ b/packages/rs-platform-wallet/src/wallet/shielded/operations.rs
@@ -757,21 +757,89 @@ where
 
         let identity_id = build.identity_id;
 
-        trace!("IdentityCreateFromShieldedPool: built, broadcasting via SDK helper...");
-        // Broadcast through the SDK helper, which re-assembles the transition from the PoP-signed
-        // keys + bundle params (preserving the per-key signatures) and waits for proven execution.
-        // It returns a `VerifiedIdentityWithShieldedNullifiers` proof result carrying the
-        // proof-verified `Identity` (and the consumed nullifiers).
-        let proof_result = sdk
-            .identity_create_from_shielded_pool(
+        trace!("IdentityCreateFromShieldedPool: built, broadcasting via SDK...");
+        // Stage the broadcast and the result-wait SEPARATELY (instead of one `broadcast_and_wait`)
+        // so the two failure shapes can be told apart:
+        //   - a broadcast-time rejection (relay/CheckTx refused the tx) means it never executed, so
+        //     releasing the note reservations is correct;
+        //   - a post-broadcast wait failure is AMBIGUOUS — the relay accepted the tx and it may well
+        //     have executed; treating it as "unregistered" + releasing the notes is the
+        //     orphaned-identity + double-spend hazard this split exists to avoid.
+        // The transition is built once from the PoP-signed keys + bundle params (preserving the
+        // per-key signatures); the binding signature already committed `identity_id`, the
+        // denomination, and the full key set.
+        let st = sdk
+            .identity_create_from_shielded_pool_transition(
                 build.public_keys,
                 denomination,
                 send_to_address_on_creation_failure,
                 build.bundle,
-                None,
             )
+            .map_err(|e| PlatformWalletError::ShieldedBuildError(e.to_string()))?;
+
+        // Broadcast (relay-ACK only). A failure here is definitive: the tx was NOT accepted, so the
+        // spend never happened — map to `ShieldedBroadcastFailed` and let the outer match release
+        // the reservation, exactly as before.
+        st.broadcast(sdk, None)
             .await
             .map_err(|e| PlatformWalletError::ShieldedBroadcastFailed(e.to_string()))?;
+
+        // Wait for proven execution. Classify the failure:
+        //   - `StateTransitionBroadcastError` = Platform DEFINITIVELY reported the transition's own
+        //     execution error (it ran and was rejected on its merits). The identity does not exist;
+        //     keep today's behavior (release the reservation via `ShieldedBroadcastFailed`).
+        //   - any other error (DriveProofError / Proof / InvalidProvedResponse / TimeoutReached /
+        //     DapiClientError / …) = AMBIGUOUS: the broadcast was accepted and the transition may
+        //     have executed even though we couldn't fetch/verify its result proof (this is exactly
+        //     the #3859 result-proof incident). Fall back to fetching the identity by its
+        //     pre-derived id before deciding it doesn't exist.
+        let proof_result = match st
+            .wait_for_response::<StateTransitionProofResult>(sdk, None)
+            .await
+        {
+            Ok(result) => result,
+            Err(dash_sdk::Error::StateTransitionBroadcastError(e)) => {
+                return Err(PlatformWalletError::ShieldedBroadcastFailed(e.to_string()));
+            }
+            Err(wait_err) => {
+                warn!(
+                    derived_id = %identity_id,
+                    error = %wait_err,
+                    "IdentityCreateFromShieldedPool: broadcast accepted but result confirmation \
+                     failed; falling back to fetching the identity by its derived id"
+                );
+                // Fetch the identity directly. The id is committed in the sighash and derived
+                // deterministically from the spent nullifiers, so if the transition landed the row
+                // is queryable. A freshly-included identity can take a moment to index on the DAPI
+                // node we hit, so retry a few times with a short, fixed backoff before concluding
+                // it's truly absent — long enough to ride out routine indexing/replica lag, short
+                // enough not to wedge the caller's UI for minutes.
+                match fetch_identity_with_retries(sdk, identity_id).await {
+                    Some(mut identity) => {
+                        info!(
+                            derived_id = %identity_id,
+                            "IdentityCreateFromShieldedPool: result confirmation failed but the \
+                             identity was found on chain by its derived id; treating as success"
+                        );
+                        // Same defensive empty-`public_keys` fill as the proven-result path below,
+                        // so downstream auth-key checks see the committed key set immediately.
+                        if identity.public_keys().is_empty() {
+                            identity.set_public_keys(submitted_public_keys.clone());
+                        }
+                        return Ok::<(Identifier, Identity), PlatformWalletError>((
+                            identity.id(),
+                            identity,
+                        ));
+                    }
+                    None => {
+                        return Err(PlatformWalletError::ShieldedBroadcastUnconfirmed {
+                            identity_id,
+                            reason: wait_err.to_string(),
+                        });
+                    }
+                }
+            }
+        };
 
         // Pull the verified `Identity` out of the proof result. The expected variant is
         // `VerifiedIdentityWithShieldedNullifiers`; if drive-abci ever returns a different one the
@@ -846,11 +914,71 @@ where
             );
             Ok((identity_id, identity))
         }
+        // The broadcast was accepted but its result couldn't be confirmed and the identity wasn't
+        // found by a direct fetch. Do NOT `cancel_pending` here: `pending_nullifiers` is in-memory
+        // only (see `SubwalletState`, "never persisted; the next sync after a crash reconciles"),
+        // and `mark_spent` during nullifier sync clears matching reservations. So if the transition
+        // actually executed, the next sync promotes these notes to spent; if it truly never landed,
+        // an app restart drops the in-memory reservation and frees them. Releasing them now would
+        // invite double-spend attempts against notes that may already be consumed on chain — the
+        // very hazard this variant exists to prevent.
+        Err(e @ PlatformWalletError::ShieldedBroadcastUnconfirmed { .. }) => Err(e),
         Err(e) => {
             cancel_pending(store, id, &selected_notes).await;
             Err(e)
         }
     }
+}
+
+/// Number of times [`identity_create_from_shielded_pool`] re-fetches the new identity by its
+/// derived id after a post-broadcast result-confirmation failure, before declaring the broadcast
+/// unconfirmed.
+const IDENTITY_CREATE_FETCH_RETRIES: usize = 4;
+
+/// Fixed backoff between identity fetch attempts. Four attempts ~3 s apart (~9 s of fetch window
+/// total) is enough to ride out routine DAPI indexing / replica lag for a freshly-included identity
+/// without wedging the caller's UI for minutes.
+const IDENTITY_CREATE_FETCH_RETRY_DELAY: std::time::Duration = std::time::Duration::from_secs(3);
+
+/// Fetch an identity by id with a few fixed-interval retries.
+///
+/// Used only on the ambiguous post-broadcast path: the result-proof fetch failed, so we don't know
+/// whether the transition executed. The identity id is derived deterministically from the spent
+/// notes' nullifiers and committed in the transition sighash, so a successful fetch is positive
+/// proof the transition landed. Returns `Some(identity)` on the first hit, or `None` if every
+/// attempt comes back empty or errors (transport hiccup, not-yet-indexed, …) — the caller then
+/// surfaces `ShieldedBroadcastUnconfirmed` rather than a hard failure.
+async fn fetch_identity_with_retries(
+    sdk: &Arc<dash_sdk::Sdk>,
+    identity_id: Identifier,
+) -> Option<Identity> {
+    use dash_sdk::platform::Fetch;
+
+    for attempt in 0..IDENTITY_CREATE_FETCH_RETRIES {
+        match Identity::fetch(sdk, identity_id).await {
+            Ok(Some(identity)) => return Some(identity),
+            Ok(None) => {
+                trace!(
+                    %identity_id,
+                    attempt,
+                    "IdentityCreateFromShieldedPool confirmation fetch: not found yet"
+                );
+            }
+            Err(e) => {
+                trace!(
+                    %identity_id,
+                    attempt,
+                    error = %e,
+                    "IdentityCreateFromShieldedPool confirmation fetch errored; will retry"
+                );
+            }
+        }
+        // Skip the trailing sleep after the final attempt — nothing follows it.
+        if attempt + 1 < IDENTITY_CREATE_FETCH_RETRIES {
+            tokio::time::sleep(IDENTITY_CREATE_FETCH_RETRY_DELAY).await;
+        }
+    }
+    None
 }
 
 // -------------------------------------------------------------------------
--- a/packages/rs-platform-wallet/src/wallet/shielded/operations.rs
+++ b/packages/rs-platform-wallet/src/wallet/shielded/operations.rs
@@ -798,8 +798,10 @@ where
             .await
         {
             Ok(result) => result,
-            Err(dash_sdk::Error::StateTransitionBroadcastError(e)) => {
-                return Err(PlatformWalletError::ShieldedBroadcastFailed(e.to_string()));
+            Err(wait_err) if wait_error_is_definitive_rejection(&wait_err) => {
+                return Err(PlatformWalletError::ShieldedBroadcastFailed(
+                    wait_err.to_string(),
+                ));
             }
             Err(wait_err) => {
                 warn!(
@@ -928,6 +930,21 @@ where
             Err(e)
         }
     }
+}
+
+/// Whether a wait-stage error is a DEFINITIVE platform rejection — the transition executed and was
+/// rejected on its merits, so the identity does not exist and releasing the note reservations is
+/// correct — as opposed to an AMBIGUOUS confirmation failure, where the broadcast was accepted and
+/// the transition may have executed even though we couldn't fetch/verify its result proof (so the
+/// reservation must be retained until a direct id fetch or the next nullifier sync resolves it).
+///
+/// Only `StateTransitionBroadcastError` is definitive: it carries Platform's own report of the
+/// transition's execution error. Everything else — `DriveProofError` / `Proof` /
+/// `InvalidProvedResponse` / `TimeoutReached` / `DapiClientError` / `Config` / … — is ambiguous
+/// (this is exactly the #3859 result-proof incident: the transition landed but the result proof
+/// couldn't be confirmed).
+fn wait_error_is_definitive_rejection(e: &dash_sdk::Error) -> bool {
+    matches!(e, dash_sdk::Error::StateTransitionBroadcastError(_))
 }
 
 /// Number of times [`identity_create_from_shielded_pool`] re-fetches the new identity by its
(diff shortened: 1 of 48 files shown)


thepastaclaw:
✅ Review complete (commit 6df799db)

packages/rs-platform-wallet/src/wallet/shielded/operations.rs
```diff
@@ -757,22 +757,90 @@ where
...
+                    }
+                    None => {
+                        return Err(PlatformWalletError::ShieldedBroadcastUnconfirmed {
+                            identity_id,
+                            reason: wait_err.to_string(),
+                        });
+                    }
+                }
+            }
+        };
```
thepastaclaw, line 875: **🟡 Suggestion: New wait-stage classification has no Rust-level test pinning the safety invariant**

The whole orphaned-identity argument hinges on three branches added in this PR: `Err(StateTransitionBroadcastError)` → `ShieldedBroadcastFailed` (release reservation), other wait error + identity found by retry → success (notes finalized), other wait error + identity still missing → `ShieldedBroadcastUnconfirmed` with reservations intentionally retained. None of those have a unit/integration test in `rs-platform-wallet`. A future reorder of the `match` arms or refactor of `cancel_pending` could silently regress the wallet to the pre-PR behavior the description identifies as a double-spend/orphaned-identity hazard. A test (against a mocked `Sdk`/`Fetch`) that injects each of the three wait-error shapes and asserts which `PlatformWalletError` variant comes out — plus whether `cancel_pending` was called — would lock the invariant in.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

PR cleanly splits broadcast and wait-stage handling for Type-20 shielded identity-create so post-broadcast confirmation failures hold the note reservation and surface an unconfirmed-with-id error, while genuine relay/CheckTx rejections continue to release. The classification, FFI wiring (codes 16/17, conditional out_identity_id), and Swift mirror are internally consistent. No in-scope blocking issues; a few defensive suggestions and one latent FFI-mapping gap worth fixing now.

🟡 3 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet-ffi/src/error.rs:194-214: Blanket From<PlatformWalletError> flattens ShieldedBroadcastFailed/Unconfirmed to ErrorUnknown
  The new `ShieldedBroadcastFailed` and `ShieldedBroadcastUnconfirmed` variants are mapped to the dedicated FFI codes 16/17 only inside `platform_wallet_manager_shielded_identity_create_from_pool`'s explicit match. The blanket `From<PlatformWalletError> for PlatformWalletFFIResult` here still falls through to `ErrorUnknown`, and `unwrap_result_or_return!` routes any `Err(...).into()` through this impl. Today nothing else propagates these variants via `?`, so the bug is latent — but the first time a helper does, the host will see `ErrorUnknown` instead of code 17, `out_identity_id` will be left zeroed (the blanket impl has no out-parameter), and Swift will fall back to `PlatformWalletError.unknown(...)` rather than `ShieldedIdentityCreateUnconfirmedError`. That re-introduces exactly the orphaned-identity/double-spend hazard this PR closes. Add explicit arms here for both variants (the unconfirmed case still cannot carry the id through a blanket conversion, but at least the typed code survives), and/or document on the variants that they must always be matched explicitly at the FFI entry point.

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/operations.rs:786-842: New wait-stage classification has no Rust-level test pinning the safety invariant
  The whole orphaned-identity argument hinges on three branches added in this PR: `Err(StateTransitionBroadcastError)` → `ShieldedBroadcastFailed` (release reservation), other wait error + identity found by retry → success (notes finalized), other wait error + identity still missing → `ShieldedBroadcastUnconfirmed` with reservations intentionally retained. None of those have a unit/integration test in `rs-platform-wallet`. A future reorder of the `match` arms or refactor of `cancel_pending` could silently regress the wallet to the pre-PR behavior the description identifies as a double-spend/orphaned-identity hazard. A test (against a mocked `Sdk`/`Fetch`) that injects each of the three wait-error shapes and asserts which `PlatformWalletError` variant comes out — plus whether `cancel_pending` was called — would lock the invariant in.

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Services/IdentityRegistrationController.swift`:
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Services/IdentityRegistrationController.swift:222-230: FailureStage attribution treats every non-shieldedBroadcastFailed error as pre-broadcast
  The new mapping pins anything that isn't exactly `PlatformWalletError.shieldedBroadcastFailed` to `.beforeBroadcast`. That's correct for the two typed variants this PR introduces, but it also catches unrelated `PlatformWalletError` variants surfacing through the FFI bridge (e.g. wallet-handle / marshalling failures) and confidently attributes them to the Halo 2 build step purely by elimination — the same misattribution this PR is fixing, just for a different error class. Gate `.beforeBroadcast` on errors that actually come from the build/proof path (e.g. `.shieldedBuildError`) and surface anything else with no attribution (or a dedicated `.unknown` stage) so the progress view falls back to its elapsed-time heuristic rather than pointing at the wrong step.
```
</details>

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Services/IdentityRegistrationController.swift
```diff
@@ -150,27 +184,50 @@ final class IdentityRegistrationController: ObservableObject {
...
+                // proof errors) failed before the broadcast.
+                let stage: FailureStage
+                if case PlatformWalletError.shieldedBroadcastFailed = error {
+                    stage = .broadcastRejected
+                } else {
+                    stage = .beforeBroadcast
+                }
                 await MainActor.run {
-                    self?.terminalAt = Date()
+                    self?.failureStage = stage
```
thepastaclaw: **🟡 Suggestion: FailureStage attribution treats every non-shieldedBroadcastFailed error as pre-broadcast**

The new mapping pins anything that isn't exactly `PlatformWalletError.shieldedBroadcastFailed` to `.beforeBroadcast`. That's correct for the two typed variants this PR introduces, but it also catches unrelated `PlatformWalletError` variants surfacing through the FFI bridge (e.g. wallet-handle / marshalling failures) and confidently attributes them to the Halo 2 build step purely by elimination — the same misattribution this PR is fixing, just for a different error class. Gate `.beforeBroadcast` on errors that actually come from the build/proof path (e.g. `.shieldedBuildError`) and surface anything else with no attribution (or a dedicated `.unknown` stage) so the progress view falls back to its elapsed-time heuristic rather than pointing at the wrong step.

<sub>source: ['claude']</sub>

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Services/IdentityRegistrationController.swift
```diff
@@ -168,8 +217,31 @@ final class IdentityRegistrationController: ObservableObject {
...
+                if case PlatformWalletError.shieldedBroadcastFailed = error {
+                    stage = .broadcastRejected
+                } else {
+                    stage = .beforeBroadcast
+                }
                 await MainActor.run {
+                    self?.failureStage = stage
                     self?.terminalAt = Date()
                     self?.phase = .failed(error.localizedDescription)
                 }
```
thepastaclaw, line 259: **🟡 Suggestion: Non-broadcast errors are still attributed as `.beforeBroadcast` by elimination**

Verified at head: the catch-all classifies every error except `PlatformWalletError.shieldedBroadcastFailed` as `.beforeBroadcast`. The `ShieldedIdentityCreateUnconfirmedError` case is caught earlier, but any other typed wallet error surfacing through the FFI bridge (invalid handle, marshalling failure, generic wallet errors) is now confidently attributed to the build/Halo 2 step. That misleads the user about which stage actually failed and, for any failure that escapes after relay acceptance, marks a possibly-live transition as a normal pre-broadcast failure — exactly the misattribution this PR is trying to clean up. Gate `.beforeBroadcast` on errors that actually come from the proof/build path, and route unknown variants to an unattributed/`unknown` stage so the progress view falls back to elapsed-time heuristics.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/PendingRegistrationsList.swift
```diff
@@ -86,11 +90,19 @@ struct PendingRegistrationRow: View {
         }
     }
 
+    private var isDismissable: Bool {
+        switch controller.phase {
+        case .failed, .unconfirmed: return true
+        default: return false
+        }
+    }
```
thepastaclaw, line 142: **🟡 Suggestion: `.unconfirmed` rows are user-dismissable, which can free the identity slot for a fund-burning retry**

Verified at head: `isDismissable` returns true for `.unconfirmed`, and `RegistrationCoordinator.dismiss` removes the controller. The comment frames this as harmless ("dismissing it only drops the Pending row"), but the start-gate at `RegistrationCoordinator` only inspects controllers it currently holds, and `CreateIdentityView.usedIdentityIndices` consults persisted `PersistentIdentity` rows — not the per-slot `isUsed` flag, by explicit design (`CreateIdentityView.swift:2007-2015`). If a user dismisses an `.unconfirmed` row before the identity-sync writes a `PersistentIdentity` for the derived id, the same index becomes selectable again. Re-submitting registers the same keys, hits the registered-key-hash stateful check, and burns the shielded funds to the fallback address minus penalty — the exact orphaned-identity hazard the `.unconfirmed` variant was added to prevent. Gate dismissibility on the persisted-identity row actually being present (or only allow dismiss when `controller.phase == .failed`).

```suggestion
    private var isDismissable: Bool {
        if case .failed = controller.phase { return true }
        return false
    }
```

<sub>source: ['codex']</sub>

thepastaclaw (commented):
## Code Review

PR cleanly splits broadcast and result-wait for the Type-20 shielded identity-create flow, introduces the slot-holding `ShieldedBroadcastUnconfirmed` path, wires `terminalAt` and a fetch-by-derived-id fallback, and adds Drive/ABCI proof + direct-purchase regression tests. Latest delta (2270899→9efa3363) is a v3.1-dev merge plus a small `terminalAt` freeze in the Swift controller. All three prior suggestions are STILL VALID; one cumulative-head suggestion is added around UI dismissibility of the `.unconfirmed` row.

🟡 3 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet-ffi/src/error.rs:194-214: Blanket From<PlatformWalletError> still flattens shielded broadcast codes to ErrorUnknown
  Verified at head: lines 194-214 only special-case `NoSpendableInputs`/`OnlyOutputAddressesFunded`/`OnlyDustInputs` and `WalletAlreadyExists`; the two new variants `ShieldedBroadcastFailed` and `ShieldedBroadcastUnconfirmed` fall through to `ErrorUnknown`. They are mapped to codes 16/17 only inside the dedicated `platform_wallet_manager_shielded_identity_create_from_pool` match. Any future FFI entry point that propagates a `PlatformWalletError` through `?` / `.into()` will silently emit `ErrorUnknown`, defeating the slot-holding contract documented at the top of this file. Add explicit arms here (and keep the `out_identity_id` write at the dedicated entry point — only that path owns the typed id) so the codes survive future plumbing changes.

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Services/IdentityRegistrationController.swift`:
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Services/IdentityRegistrationController.swift:231-247: Non-broadcast errors are still attributed as `.beforeBroadcast` by elimination
  Verified at head: the catch-all classifies every error except `PlatformWalletError.shieldedBroadcastFailed` as `.beforeBroadcast`. The `ShieldedIdentityCreateUnconfirmedError` case is caught earlier, but any other typed wallet error surfacing through the FFI bridge (invalid handle, marshalling failure, generic wallet errors) is now confidently attributed to the build/Halo 2 step. That misleads the user about which stage actually failed and, for any failure that escapes after relay acceptance, marks a possibly-live transition as a normal pre-broadcast failure — exactly the misattribution this PR is trying to clean up. Gate `.beforeBroadcast` on errors that actually come from the proof/build path, and route unknown variants to an unattributed/`unknown` stage so the progress view falls back to elapsed-time heuristics.

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/PendingRegistrationsList.swift`:
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/PendingRegistrationsList.swift:74-98: `.unconfirmed` rows are user-dismissable, which can free the identity slot for a fund-burning retry
  Verified at head: `isDismissable` returns true for `.unconfirmed`, and `RegistrationCoordinator.dismiss` removes the controller. The comment frames this as harmless ("dismissing it only drops the Pending row"), but the start-gate at `RegistrationCoordinator` only inspects controllers it currently holds, and `CreateIdentityView.usedIdentityIndices` consults persisted `PersistentIdentity` rows — not the per-slot `isUsed` flag, by explicit design (`CreateIdentityView.swift:2007-2015`). If a user dismisses an `.unconfirmed` row before the identity-sync writes a `PersistentIdentity` for the derived id, the same index becomes selectable again. Re-submitting registers the same keys, hits the registered-key-hash stateful check, and burns the shielded funds to the fallback address minus penalty — the exact orphaned-identity hazard the `.unconfirmed` variant was added to prevent. Gate dismissibility on the persisted-identity row actually being present (or only allow dismiss when `controller.phase == .failed`).
```
</details>

packages/rs-platform-wallet/src/wallet/shielded/operations.rs
thepastaclaw: Resolved in this update — _New wait-stage classification has no Rust-level test pinning the safety invariant_ no longer present.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Services/IdentityRegistrationController.swift
thepastaclaw: Resolved in `9efa336` — _FailureStage attribution treats every non-shieldedBroadcastFailed error as pre-broadcast_ no longer present.

packages/rs-platform-wallet/src/wallet/shielded/operations.rs
QuantumExplorer: Addressed in 420628a0db: extracted `wait_error_is_definitive_rejection` and added unit tests pinning StateTransitionBroadcastError → definitive (release) vs TimeoutReached/InvalidProvedResponse/Config → ambiguous (retain); a full mocked-Sdk test of the cancel_pending path needs the shared shielded e2e harness, which doesn't exist yet.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Services/IdentityRegistrationController.swift
QuantumExplorer: Addressed in 420628a0db: dropped .beforeBroadcast entirely — only a typed shieldedBroadcastFailed sets a stage (.broadcastRejected); everything else leaves failureStage nil and the progress view falls back to its elapsed-time heuristic instead of claiming a stage by elimination.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Services/IdentityRegistrationController.swift
QuantumExplorer: Addressed in 420628a0db: removed .beforeBroadcast; FailureStage now has only .broadcastRejected (set solely for the typed shieldedBroadcastFailed), all other errors stay unattributed (nil) and the view uses its elapsed-time heuristic.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/PendingRegistrationsList.swift
QuantumExplorer: Addressed in 420628a0db: PendingRegistrationRow now carries a per-slot @Query on PersistentIdentity and .unconfirmed rows are dismissable only once that row exists; while unconfirmed-and-unpersisted the live controller stays as the slot guard. Also corrected the overclaiming markIdentitySlotUsed comment in observeController.

packages/rs-platform-wallet/src/wallet/shielded/operations.rs
```diff
@@ -846,13 +916,88 @@ where
...
+        // actually executed, the next sync promotes these notes to spent; if it truly never landed,
+        // an app restart drops the in-memory reservation and frees them. Releasing them now would
+        // invite double-spend attempts against notes that may already be consumed on chain — the
+        // very hazard this variant exists to prevent.
+        Err(e @ PlatformWalletError::ShieldedBroadcastUnconfirmed { .. }) => Err(e),
         Err(e) => {
             cancel_pending(store, id, &selected_notes).await;
             Err(e)
         }
     }
```
thepastaclaw: **🟡 Suggestion: Outer-match reservation-retention invariant for ShieldedBroadcastUnconfirmed is still untested**

The latest delta adds unit tests for wait_error_is_definitive_rejection (operations.rs:1342-1368), which pin the wait-stage classifier. The complementary safety invariant — that the outer result match returns Err for ShieldedBroadcastUnconfirmed WITHOUT calling cancel_pending, while every other Err arm DOES cancel — is the actual protection against double-spending reserved notes after an ambiguous wait failure, and it has no Rust-level regression test. A small test seam that injects an unconfirmed result and asserts the selected_notes reservations remain held (and conversely, that a ShieldedBroadcastFailed result clears them) would prevent a future refactor from silently moving the unconfirmed arm into the generic cancellation branch. Partial carry-forward of prior finding #2.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):
## Code Review

Verified the latest delta against the prior review at 9efa3363: 3 of 4 prior findings are fixed (typed FFI codes preserved through the blanket From impl, failure-stage no longer attributed by elimination, .unconfirmed dismissal gated on a live PersistentIdentity row); the 4th (Rust-level test for the outer reservation-retention match) is partially addressed via unit tests for wait_error_is_definitive_rejection but the outer-match invariant itself is still untested. One new in-scope blocking issue: hasInFlightRegistrations in RegistrationCoordinator does not include .unconfirmed, so the OptionsView network-switch gate (the only thing keeping the FFI manager / coordinator alive while the slot is in-memory-only) re-enables as soon as the flow becomes .unconfirmed, allowing the same slot to lose its in-memory guard and be re-selected against an identity that is probably live on chain. Two pre-existing structural concerns (broadcast-stage transport ambiguity, persistent slot reservation across app restart) are recorded as out-of-scope follow-ups.

<details>

```

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Services/RegistrationCoordinator.swift`:
- [BLOCKING] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Services/RegistrationCoordinator.swift:44-53: Network-switch gate ignores .unconfirmed and can drop the only in-memory slot guard
  hasInFlightRegistrations returns true only for .preparingKeys and .inFlight, but this PR introduces .unconfirmed precisely because the on-chain identity may already exist and the live IdentityRegistrationController is the only thing keeping the slot from being re-selected before sync writes a PersistentIdentity row. The OptionsView network picker uses this flag (see OptionsView.swift:805 and the wrapper viewmodifier comment at :793-799) as its .disabled(_:) gate, and PlatformWalletManager+RegistrationCoordinator.swift:23-26 documents that switching networks tears down the active PlatformWalletManager (and with it the registrationCoordinator attached as an associated object). As soon as the flow reaches .unconfirmed, the gate releases, the user can switch networks (and switch back), the in-memory controller and pending_nullifiers reservation are gone, and usedIdentityIndices — which consults only PersistentIdentity — will once again surface the same HD slot. A second submission against the same registered key set will be rejected by the registered-key-hash stateful check and route the funded spend to the fallback address minus penalty — exactly the orphaned-identity / fund-burn hazard the .unconfirmed phase was introduced to prevent. The dismissal path in PendingRegistrationsList was hardened against this; this sibling gate must match.

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/operations.rs:919-932: Outer-match reservation-retention invariant for ShieldedBroadcastUnconfirmed is still untested
  The latest delta adds unit tests for wait_error_is_definitive_rejection (operations.rs:1342-1368), which pin the wait-stage classifier. The complementary safety invariant — that the outer result match returns Err for ShieldedBroadcastUnconfirmed WITHOUT calling cancel_pending, while every other Err arm DOES cancel — is the actual protection against double-spending reserved notes after an ambiguous wait failure, and it has no Rust-level regression test. A small test seam that injects an unconfirmed result and asserts the selected_notes reservations remain held (and conversely, that a ShieldedBroadcastFailed result clears them) would prevent a future refactor from silently moving the unconfirmed arm into the generic cancellation branch. Partial carry-forward of prior finding #2.
```
</details>

packages/rs-platform-wallet/src/wallet/shielded/operations.rs
QuantumExplorer: Addressed in d9283d0ce3: the retention decision is now the named error_releases_note_reservation helper (false only for ShieldedBroadcastUnconfirmed) used by the outer match, with unit tests pinning retain-on-unconfirmed and release for ShieldedBroadcastFailed/ShieldedBuildError/ShieldedStoreError — a refactor can no longer fold the unconfirmed arm into the generic cancellation branch.

thepastaclaw (approved):
## Code Review

Both prior findings (network-switch gate excluding .unconfirmed; missing test for the outer-match reservation-retention invariant) are resolved on head d9283d0c: the Swift gate now delegates to phase.isActive which includes .unconfirmed, and the Rust outer match has been refactored to call error_releases_note_reservation, which is directly unit-tested for retain (Unconfirmed) vs release (definitive failures). No new in-scope blocking or suggestion-level defects introduced by the incremental delta.

thepastaclaw (approved):
## Code Review

Prior findings at d9283d0c were revalidated — none existed and none carry forward. Cumulative review against the PR base and the latest delta against d9283d0c found no in-scope issues. The PR cleanly stages shielded `IdentityCreateFromShieldedPool` failures (build / broadcast / result-confirmation) so reservations are released only on definitive rejection, and the new typed FFI codes (16/17/18) plus the `error_releases_note_reservation` helper make the policy explicit and unit-tested. The Swift presentation layer correctly mirrors the new states. Codex's two FFI 'blocking' findings target `deriveIdentityAuthKeyAtSlot` (lines 926-936) and `prePersistIdentityKeysForRegistration` (lines 1056-1067), but those call sites are pre-existing in `ManagedPlatformWallet.swift` and are not touched by this PR — the PR's own new resolver call sites (in `previewIdentityRegistrationKeys` and `discoverIdentities`) already wrap their FFI calls in `withExtendedLifetime(resolver)`, so they're out of scope for this PR.

PR #3865, opened 2026-06-12
fix(dpp): reject empty token pricing schedules to prevent a direct-purchase chain halt

QuantumExplorer:

A chain-halt (liveness) bug in the token direct-purchase path.

An empty `SetPrices(BTreeMap::new())` pricing schedule is a representable, storable value:
`TokenPricingSchedule::SetPrices` wraps a `BTreeMap<TokenAmount, Credits>`, and neither the
set-price **structure** validation nor the set-price **state** validation rejected an empty map.
A token's pricing authority could therefore set `price = Some(SetPrices(empty map))` via a
`TokenSetPriceForDirectPurchase` transition that passed all validation, and the empty schedule
flowed unchanged into GroveDB.

Afterwards, **any** identity submitting a `TokenDirectPurchase` for that token detonated a panic.
In the transformer's `None` arm, the original code built the error via
`set_prices.keys().next().expect("Map is not empty")`, which panics when the map is empty.

This panic is a deterministic, quorum-wide chain halt — confirmed by tracing the call stack:

- The transformer runs during deterministic block execution on every validator:
  `process_proposal` → `run_block_proposal` → `process_raw_state_transitions` →
  `process_state_transition` → `transform_into_action` → the token transformer.
- Per-state-transition `Err`s are caught into an `InternalError` execution result (chain
  continues) at `process_raw_state_transitions/v0/mod.rs`, but **panics are not** — that catch is
  `.unwrap_or_else(...)` on a `Result`, which a panic unwinds straight through.
- There is **no** `catch_unwind` anywhere on the ABCI path (workspace-wide).
- The global panic hook (`install_panic_hook` in `main.rs`) calls `cancel.cancel()`, shutting the
  node down on any panic.

Result: every validator panics on the same block during `process_proposal` → every node shuts
down → chain halt.

## What was done?

Defense in depth across the produce and consume sides:

1. **Reject empty `SetPrices` at set-price structure validation** with a new
   `TokenPricingScheduleEmptyError` consensus error (basic code `10461`). Emptiness is a pure
   property of the transition's `price` field — no state needed — and structure validation
   short-circuits the batch before the transition is ever applied, so an empty schedule can never
   reach state.
2. **Transformer guard (already present in the base branch):** the `None` arm matches on
   `set_prices.keys().next()` and returns `TokenNotForDirectSale` for an empty map instead of
   `.expect()`-ing. This PR keeps it as the consume-side guard and adds a regression test for it.

Wiring for the new error: new error struct under `errors/consensus/basic/token/`, re-exported in
that module, added to the `BasicError` enum (tail-appended to preserve bincode positional
discriminants), assigned error code `10461` in `codes.rs`, and bridged in the `wasm-dpp`
exhaustive consensus-error match.

## How Has This Been Tested?

New regression tests:

- `should_return_error_when_set_prices_is_empty` — set-price structure validation rejects an empty
  `SetPrices` with `TokenPricingScheduleEmptyError` (plus `should_pass_with_non_empty_set_prices`
  and `should_pass_with_single_price` for the happy paths).
- `set_prices_empty_map_does_not_panic_and_is_treated_as_not_for_sale` — the transformer's
  `None`-arm logic resolves an empty map to the not-for-sale branch without panicking.

Verification run locally:

- `cargo test -p dpp --all-features` (new validation tests pass)
- `cargo test -p drive --lib set_prices_empty_map_does_not_panic` (transformer regression passes)
- `cargo check -p dpp --all-features --tests`
- `cargo check -p wasm-dpp --target wasm32-unknown-unknown`
- `cargo fmt --all`

## Breaking Changes

Consensus behavior changes: a `TokenSetPriceForDirectPurchase` transition carrying an empty
`SetPrices` map is now rejected at structure validation (new consensus error code `10461`),
whereas it was previously accepted. This is a consensus-relevant hardening landing pre-release
(`4.0.0-rc.2`) — the protocol version is not finalized/deployed, so there is no released network
to fork against and no migration concern. The new `BasicError` variant is appended at the enum
tail to preserve bincode positional encoding of existing variants.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Validation now rejects empty token pricing schedules and returns a clear error instead of allowing invalid configs.
  * Prevented a panic when processing direct-purchase transitions with empty pricing schedules; these now fail gracefully.
  * Error mapping updated so the new error is surfaced consistently across native and WASM layers.
* **Tests**
  * Added regression and unit tests covering empty and non-empty pricing schedules and direct-purchase edge cases.

### DIFF
--- a/packages/rs-drive/src/state_transition_action/batch/batched_transition/token_transition/token_direct_purchase_transition_action/v0/transformer.rs
+++ b/packages/rs-drive/src/state_transition_action/batch/batched_transition/token_transition/token_direct_purchase_transition_action/v0/transformer.rs
@@ -560,6 +560,36 @@ mod tests {
         assert_eq!(min_threshold, 5);
     }
 
+    /// Regression test for the token direct-purchase chain-halt bug.
+    ///
+    /// An empty `SetPrices` schedule is a representable, storable value. For it,
+    /// `range(..=token_count).next_back()` returns `None`, and in that arm the transformer
+    /// decides which consensus error to emit by inspecting `set_prices.keys().next()`.
+    /// The original code did `*set_prices.keys().next().expect("Map is not empty")` here,
+    /// which panics for an empty map — an uncaught panic during per-state-transition
+    /// processing that deterministically halts the chain across the quorum. The fixed code
+    /// matches on the `Option`: an empty schedule selects the "not for sale" branch instead
+    /// of panicking. This test proves the offending expression is `None`, never a panic.
+    #[test]
+    fn set_prices_empty_map_does_not_panic_and_is_treated_as_not_for_sale() {
+        let set_prices = BTreeMap::<TokenAmount, Credits>::new();
+        let token_count: TokenAmount = 5;
+
+        // The transformer reaches the `None` arm: an empty map has no tier <= token_count.
+        assert!(
+            set_prices.range(..=token_count).next_back().is_none(),
+            "empty schedule has no matching tier"
+        );
+
+        // The transformer then branches on `set_prices.keys().next()`. For an empty map this
+        // is `None` (the old `.expect(\"Map is not empty\")` panicked here), so the transformer
+        // classifies the token as "not for sale" rather than killing the validator.
+        assert!(
+            set_prices.keys().next().is_none(),
+            "empty schedule must resolve to the not-for-sale branch, never a panic"
+        );
+    }
+
     /// SetPrices underpayment: matched_price * token_count computes to required_total but
     /// the user agreed to less, so the transformer returns `TokenDirectPurchaseUserPriceTooLow`.
     #[test]
--- a/packages/rs-drive/src/state_transition_action/batch/batched_transition/token_transition/token_direct_purchase_transition_action/v0/transformer.rs
+++ b/packages/rs-drive/src/state_transition_action/batch/batched_transition/token_transition/token_direct_purchase_transition_action/v0/transformer.rs
@@ -1,3 +1,4 @@
+use std::collections::BTreeMap;
 use std::sync::Arc;
 use grovedb::TransactionArg;
 use dpp::block::block_info::BlockInfo;
@@ -8,6 +9,8 @@ use dpp::consensus::state::token::{TokenAmountUnderMinimumSaleAmount, TokenDirec
 use dpp::identifier::Identifier;
 use dpp::state_transition::batch_transition::token_direct_purchase_transition::v0::TokenDirectPurchaseTransitionV0;
 use dpp::ProtocolError;
+use dpp::balances::credits::TokenAmount;
+use dpp::fee::Credits;
 use crate::drive::contract::DataContractFetchInfo;
 use crate::state_transition_action::batch::batched_transition::token_transition::token_base_transition_action::TokenBaseTransitionAction;
 use crate::state_transition_action::batch::batched_transition::token_transition::token_direct_purchase_transition_action::v0::TokenDirectPurchaseTransitionActionV0;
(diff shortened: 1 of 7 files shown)


thepastaclaw:
✅ Review complete (commit 9eb600da)

thepastaclaw (commented):
## Code Review

Targeted, well-scoped fix for the direct-purchase chain-halt: empty `SetPrices` is now rejected at DPP structure validation with a new `TokenPricingScheduleEmptyError` (code 10461), appended at the tail of `BasicError` so bincode discriminants are preserved, and bridged through the wasm-dpp exhaustive converter. Plumbing, ordering, and gating are correct. The only in-scope concern is that the new transformer-side 'regression' test does not call the transformer — it asserts standard `BTreeMap` semantics on an empty map, so a reintroduction of the original `.expect("Map is not empty")` would not be caught.

🟡 1 suggestion(s)

<details>

```

In `packages/rs-drive/src/state_transition_action/batch/batched_transition/token_transition/token_direct_purchase_transition_action/v0/transformer.rs`:
- [SUGGESTION] packages/rs-drive/src/state_transition_action/batch/batched_transition/token_transition/token_direct_purchase_transition_action/v0/transformer.rs:573-591: Regression test does not exercise the transformer — it tests `BTreeMap`, not the fixed code path
  `set_prices_empty_map_does_not_panic_and_is_treated_as_not_for_sale` is documented as the regression test for the chain-halt bug, but it never invokes `try_from_borrowed_token_direct_purchase_transition_with_contract_lookup` (or any transformer entry point). It only asserts that `BTreeMap::range(..=k).next_back()` and `BTreeMap::keys().next()` return `None` on an empty map — properties of the standard library, not of the transformer. If a later refactor restored the original `*set_prices.keys().next().expect("Map is not empty")` on the `None` arm, this test would still pass green. Drive the transformer directly with a `TokenDirectPurchaseTransition` whose token state carries `Some(TokenPricingSchedule::SetPrices(BTreeMap::new()))` and assert the result is `BumpIdentityDataContractNonce` + `TokenNotForDirectSale` rather than a panic; that fails closed if the `.expect()` is ever reintroduced. Not blocking because the new structure-validation guard prevents an empty schedule from being stored on the happy path, but the consume-side defense-in-depth narrative in the PR description currently isn't backed by a real test of the consume-side branch.
```
</details>

packages/rs-drive/src/state_transition_action/batch/batched_transition/token_transition/token_direct_purchase_transition_action/v0/transformer.rs
```diff
@@ -560,6 +560,36 @@ mod tests {
...
+        );
+
+        // The transformer then branches on `set_prices.keys().next()`. For an empty map this
+        // is `None` (the old `.expect(\"Map is not empty\")` panicked here), so the transformer
+        // classifies the token as "not for sale" rather than killing the validator.
+        assert!(
+            set_prices.keys().next().is_none(),
+            "empty schedule must resolve to the not-for-sale branch, never a panic"
+        );
+    }
```
thepastaclaw, line 591: **🟡 Suggestion: Regression test does not exercise the transformer — it tests `BTreeMap`, not the fixed code path**

`set_prices_empty_map_does_not_panic_and_is_treated_as_not_for_sale` is documented as the regression test for the chain-halt bug, but it never invokes `try_from_borrowed_token_direct_purchase_transition_with_contract_lookup` (or any transformer entry point). It only asserts that `BTreeMap::range(..=k).next_back()` and `BTreeMap::keys().next()` return `None` on an empty map — properties of the standard library, not of the transformer. If a later refactor restored the original `*set_prices.keys().next().expect("Map is not empty")` on the `None` arm, this test would still pass green. Drive the transformer directly with a `TokenDirectPurchaseTransition` whose token state carries `Some(TokenPricingSchedule::SetPrices(BTreeMap::new()))` and assert the result is `BumpIdentityDataContractNonce` + `TokenNotForDirectSale` rather than a panic; that fails closed if the `.expect()` is ever reintroduced. Not blocking because the new structure-validation guard prevents an empty schedule from being stored on the happy path, but the consume-side defense-in-depth narrative in the PR description currently isn't backed by a real test of the consume-side branch.

QuantumExplorer:
@coderabbitai addressed in 9eb600da01. Extracted the `SetPrices` resolution out of the transformer into a pure `resolve_set_prices_direct_purchase_price` helper and routed the transformer through it (behavior unchanged — on `Err` it bumps the nonce and surfaces the same consensus error). Replaced the BTreeMap-replica test with five tests that drive the real helper across every branch: empty → `TokenNotForDirectSale`, below-minimum → `TokenAmountUnderMinimumSaleAmount`, matched tier → required total, underpayment → `TokenDirectPurchaseUserPriceTooLow`, and overflow → `OverflowError`. A regression in the `None` arm now fails a test instead of passing silently.

packages/rs-drive/src/state_transition_action/batch/batched_transition/token_transition/token_direct_purchase_transition_action/v0/transformer.rs
QuantumExplorer: Addressed in 9eb600da01 (this review predates that commit). The `BTreeMap`-only test was replaced: the `SetPrices` `None`-arm logic is now a pure `resolve_set_prices_direct_purchase_price` helper — the exact code the transformer's `None` arm executes — and `resolve_set_prices_empty_map_returns_not_for_sale_without_panicking` calls it with an empty map and asserts `Err(StateError::TokenNotForDirectSale)`. Reintroducing `.expect()` on that arm now panics the test → fails closed. Four sibling tests cover below-minimum, matched-tier, underpayment, and overflow. I took the pure-helper alternative rather than standing up a full `Drive` (the suggestion was marked non-blocking), since the helper is exactly the consume-side branch that had the bug.

shumkov (approved):

thepastaclaw (approved):
## Code Review

Cumulative re-review at head 9eb600da. The single prior finding — that the regression test only asserted BTreeMap stdlib semantics rather than exercising the fixed transformer code path — is FIXED. The latest delta (refactor commit 9eb600da) extracts the SetPrices resolution logic into a pure helper `resolve_set_prices_direct_purchase_price` (transformer.rs:254-308) that the transformer's SetPrices arm now invokes (transformer.rs:199-224). The replacement test `resolve_set_prices_empty_map_returns_not_for_sale_without_panicking` (transformer.rs:571-587) calls that helper directly with an empty BTreeMap and asserts `Err(StateError::TokenNotForDirectSale)`, so reintroducing `.expect("Map is not empty")` in the helper's `None` arm would now fail the test instead of leaving it green. Four additional tests cover below-minimum, matched-tier, underpayment, and overflow branches through the helper. No new in-scope defects identified in the latest delta; the structure-validation guard, tail-appended BasicError variant (code 10461), and wasm-dpp bridge remain correct from the prior pass.

PR #3866, opened 2026-06-12
fix(dpp): guard index property parsing against empty/oversized maps to prevent check_tx panic

QuantumExplorer:

`IndexProperty::from_platform_value` (in `packages/rs-dpp/src/data_contract/document_type/index/mod.rs`) indexed the input slice with `&index_property_map[0]` without first checking that the slice is non-empty. It is called from `Index::try_from(&[(Value, Value)])` (the `"properties"` arm) for each index-property object in a contract's document-type schema.

When a contract's document schema contains an index whose property entry is an empty object `{}`, `property.as_map()` yields an empty slice and `from_platform_value` panics with index-out-of-bounds.

In block execution (Validator mode, `full_validation=true`) this is shadowed by the document meta-schema, which enforces `minProperties: 1` / `maxProperties: 1` on index property objects. But in `check_tx` (`full_validation=false`) the meta-schema is skipped, so a crafted `DataContractCreate`/`Update` reaches `from_platform_value` with an empty slice and panics — a **mempool/check_tx node DoS** (the panic hook cancels the node).

## What was done?

- Replaced the unchecked `[0]` index in `IndexProperty::from_platform_value` with a `len() != 1` guard that returns `DataContractError::InvalidContractStructure` on an empty slice.
- The guard uses `!= 1` rather than only rejecting empty, so it also rejects oversized slices — the meta-schema enforces `maxProperties: 1`, but `check_tx` skips that too.
- The error propagates through the existing `?` in `Index::try_from`, producing a clean validation error instead of a panic in both `check_tx` and block execution.
- Added unit tests for the empty-slice and multiple-entry error paths (`test_index_property_from_platform_value_empty_map_returns_err`, `test_index_property_from_platform_value_multiple_entries_returns_err`).

## How Has This Been Tested?

- `cargo test -p dpp index_property` — all 16 tests pass, including the two new error-path tests.
- `cargo fmt --all` — clean.

## Breaking Changes

None. This change is **not consensus-breaking**. In full validation (block execution) the document meta-schema already rejected an empty/oversized index property object before reaching this code, so such a contract was never accepted. The fix only converts the `check_tx` panic into the same `DataContractError` that full validation would have produced.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced validation to prevent crashes when processing malformed data structures. The system now returns a proper error message instead of potentially panicking on invalid inputs.

### DIFF
--- a/packages/rs-dpp/src/data_contract/document_type/index/mod.rs
+++ b/packages/rs-dpp/src/data_contract/document_type/index/mod.rs
@@ -1016,6 +1016,16 @@ impl IndexProperty {
     pub fn from_platform_value(
         index_property_map: &[(Value, Value)],
     ) -> Result<Self, DataContractError> {
+        // The document meta-schema enforces `minProperties: 1` /
+        // `maxProperties: 1` on each index property object, but that
+        // validation is skipped in check_tx (full_validation=false), so a
+        // crafted contract can reach this point with an empty or oversized
+        // map. Guard explicitly to avoid panicking on an out-of-bounds index.
+        if index_property_map.len() != 1 {
+            return Err(DataContractError::InvalidContractStructure(
+                "index property entry must contain exactly one key/value".to_string(),
+            ));
+        }
         let property = &index_property_map[0];
 
         let key = property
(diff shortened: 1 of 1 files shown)


QuantumExplorer (commented):
Approved

thepastaclaw:
✅ Review complete (commit 01b2b3a3)

thepastaclaw (approved):
## Code Review

Narrow, well-scoped fix that converts an unchecked `[0]` slice index in IndexProperty::from_platform_value into a typed DataContractError when the map is empty or oversized. The author correctly identifies that the document meta-schema (minProperties/maxProperties = 1) shadows this in full validation but is skipped in check_tx, making the panic a real mempool DoS. Block-execution behavior is unchanged, the guard propagates cleanly through the existing `?` in Index::try_from, and tests cover both branches. All six reviewer roles ran and converged on no in-scope issues.

shumkov (approved):

PR #3868, opened 2026-06-12
fix(dpp): enforce byte array encoding stability in data contract updates

QuantumExplorer:

A byte array document property is serialized differently depending on its size bounds: when `minItems == maxItems` it is stored as raw, fixed-length bytes with **no** length prefix; otherwise it is stored with a variable-length (varint) length prefix. Data-contract-update schema-compatibility validation treats widening or removing `maxItems` as a compatible change, so an update could cross that boundary and silently change the on-disk encoding of a property. Documents already stored under the old encoding then become unreadable when re-decoded against the updated type, which is unsafe for any code path that re-reads stored documents.

## What was done?

Added a check to document-type update validation that rejects any byte array size-bound change that would alter its on-disk encoding — i.e. crossing the `minItems == maxItems` (fixed) ⇄ variable boundary, or changing the fixed length. Changes that do not affect the encoding (for example, widening a byte array that is already variable-length) remain allowed.

## How Has This Been Tested?

- **rs-dpp** unit tests: update validation rejects the encoding-changing updates and still accepts safe ones; a serialization round-trip test shows that, absent the check, bytes stored under the old encoding are misread under the changed type.
- **rs-drive-abci** regression covering a path that re-decodes stored documents.

## Breaking Changes

Tightens data contract update validation: a previously accepted class of updates is now rejected. Existing valid contracts and documents are unaffected.

## Checklist

---

Thanks to @DanielDerefaka for the report.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to prevent updates that change how byte-array fields are encoded on disk, rejecting incompatible schema changes that could corrupt stored data.

* **Tests**
  * Added extensive regression and end-to-end tests exercising byte-array encoding flip scenarios, schema-compatibility checks, contract-update edge cases, and contested-index flows.
  * Added a widened contract schema used by tests to simulate on-disk encoding changes.


thepastaclaw:
✅ Review complete (commit 14ba9484)

packages/rs-dpp/src/data_contract/document_type/methods/validate_update/v0/mod.rs
```diff
@@ -1325,4 +1395,103 @@ mod tests {
...
+        #[test]
+        fn accepts_widening_already_variable_byte_array() {
+            // Variable-length on both sides: the on-disk encoding does not change,
+            // so widening the bound stays allowed.
+            assert_accepted(
+                platform_value!({"type":"array","byteArray":true,"minItems":1,"maxItems":32,"position":0}),
+                platform_value!({"type":"array","byteArray":true,"minItems":1,"maxItems":64,"position":0}),
+            );
+        }
+    }
```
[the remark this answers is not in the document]
QuantumExplorer: Done in 180fa0ed — added `rejects_removing_max_items_from_fixed_byte_array` asserting the fixed→unbounded (maxItems removed) flip is rejected by validate_update. (Also addressed the unused-param nitpick in the same commit.)

packages/rs-dpp/src/data_contract/document_type/methods/validate_update/v0/mod.rs
```diff
@@ -33,10 +34,79 @@ impl DocumentTypeRef<'_> {
...
 
+        // Validate that no byte array property changes its on-disk encoding
+        let result = self.validate_byte_array_encoding_stability(new_document_type);
+
+        if !result.is_valid() {
+            return Ok(result);
+        }
+
         // Validate schema compatibility
         self.validate_schema(new_document_type, platform_version)
```
thepastaclaw, line 45: **🔴 Blocking: Tightening validate_update_v0 in place changes consensus behavior for every released protocol version**

validate_update is dispatched via platform_version.dpp.validation.document_type.validate_update, and v1.rs, v2.rs, and v3.rs all pin that field to 0. Inserting validate_byte_array_encoding_stability inside validate_update_v0 therefore retroactively tightens consensus rules for protocol versions that are already deployed on the v3.1 chain this PR targets. During any rolling upgrade or mixed-validator window, a malicious contract owner can submit exactly the update this PR rejects (e.g. byteArray {minItems:32, maxItems:32} → {minItems:32, maxItems:64}): unupgraded nodes accept the state transition while upgraded nodes reject it, producing app-hash divergence / proposal rejection — replacing the chain-halt this PR prevents with a chain-split. Historical replay from genesis is similarly affected. Per the repo's versioning discipline, a consensus tightening must land as a new validate_update_v1 with a match arm in the dispatcher, and the new feature version must be selected only at the next protocol version (v4) in PlatformVersion. v0 behavior should remain bit-exact for v1/v2/v3.

packages/rs-dpp/src/data_contract/document_type/methods/validate_update/v0/mod.rs
```diff
@@ -1325,4 +1395,103 @@ mod tests {
...
+        }
+
+        fn assert_rejected(old_ba: platform_value::Value, new_ba: platform_value::Value) {
+            let platform_version = PlatformVersion::latest();
+            let old = document_type_with_byte_array(old_ba, platform_version);
+            let new = document_type_with_byte_array(new_ba, platform_version);
+            let result = old
+                .as_ref()
+                .validate_update_v0(new.as_ref(), platform_version)
+                .expect("validate_update should not error");
```
thepastaclaw: **🟡 Suggestion: Regression tests bypass the versioned dispatcher**

The new regression tests call validate_update_v0 directly. After the change above, they would still pass even if the public dispatcher or PlatformVersion tables were never wired to the new feature version — exactly the failure mode the versioning convention exists to catch. Once the fix is moved behind validate_update_v1, exercise the public DocumentTypeRef::validate_update entry point with the PlatformVersion that selects the new version, so a missing match arm or stale version constant fails a test instead of slipping through.

<sub>source: ['codex']</sub>

packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/mod.rs
```diff
@@ -3116,4 +3116,424 @@ pub(in crate::execution) mod tests {
...
+        async fn contest_without_widening_resolves_successfully() {
+            let result = run_contest_then_resolve(false).await;
+
+            assert!(
+                result.is_ok(),
+                "control: without the byteArray widening the vote-poll resolver \
+                 must succeed; got {:?}",
+                result
+            );
+        }
```
thepastaclaw, line 3534: **💬 Nitpick: Redundant and fragile substring matching on error formatting**

err_msg.contains("Corrupted") || err_msg.contains("orrupted") || err_msg.contains("erializ") || err_msg.contains("ecod") is doubly redundant ("orrupted" subsumes "Corrupted") and loose enough to match unrelated errors, which weakens the regression. Prefer assert_matches! on the wrapped error variant (DataContractError::CorruptedSerialization through ProtocolError → Error) so the test survives Debug-format renames and asserts the intended decode-failure path.

<sub>source: ['claude']</sub>

thepastaclaw (changes requested):
## Code Review

Targeted fix correctly identifies a real chain-halt vector (byteArray fixed↔variable encoding flip via data-contract update). The validation logic itself is sound. However, the new consensus-critical rejection is added in place to validate_update_v0, while every released protocol version (v1, v2, v3) still dispatches to feature version 0. This risks a consensus split between upgraded and unupgraded validators during rolling rollout on the live v3.1 chain. The fix should be guarded behind a new validation feature version selected by the next protocol version.

<details>

```

In `packages/rs-dpp/src/data_contract/document_type/methods/validate_update/v0/mod.rs`:
- [BLOCKING] packages/rs-dpp/src/data_contract/document_type/methods/validate_update/v0/mod.rs:37-45: Tightening validate_update_v0 in place changes consensus behavior for every released protocol version
  validate_update is dispatched via platform_version.dpp.validation.document_type.validate_update, and v1.rs, v2.rs, and v3.rs all pin that field to 0. Inserting validate_byte_array_encoding_stability inside validate_update_v0 therefore retroactively tightens consensus rules for protocol versions that are already deployed on the v3.1 chain this PR targets. During any rolling upgrade or mixed-validator window, a malicious contract owner can submit exactly the update this PR rejects (e.g. byteArray {minItems:32, maxItems:32} → {minItems:32, maxItems:64}): unupgraded nodes accept the state transition while upgraded nodes reject it, producing app-hash divergence / proposal rejection — replacing the chain-halt this PR prevents with a chain-split. Historical replay from genesis is similarly affected. Per the repo's versioning discipline, a consensus tightening must land as a new validate_update_v1 with a match arm in the dispatcher, and the new feature version must be selected only at the next protocol version (v4) in PlatformVersion. v0 behavior should remain bit-exact for v1/v2/v3.
- [SUGGESTION] packages/rs-dpp/src/data_contract/document_type/methods/validate_update/v0/mod.rs:1434-1437: Regression tests bypass the versioned dispatcher
  The new regression tests call validate_update_v0 directly. After the change above, they would still pass even if the public dispatcher or PlatformVersion tables were never wired to the new feature version — exactly the failure mode the versioning convention exists to catch. Once the fix is moved behind validate_update_v1, exercise the public DocumentTypeRef::validate_update entry point with the PlatformVersion that selects the new version, so a missing match arm or stale version constant fails a test instead of slipping through.
```
</details>

packages/rs-dpp/src/data_contract/document_type/methods/validate_update/v0/mod.rs
QuantumExplorer: Addressed in f71cb817 — moved the check into a version-gated `validate_update_v1`; `validate_update_v0` is restored bit-exact and `validate_update` is bumped to feature version 1 only in `DPP_VALIDATION_VERSIONS_V3` (used solely by protocol 12 / release 3.1.0), so the rule activates atomically at the protocol upgrade rather than on binary rollout. Added a protocol-11 test proving the flip is still accepted pre-activation.

packages/rs-dpp/src/data_contract/document_type/methods/validate_update/v0/mod.rs
QuantumExplorer: Addressed in f71cb817 — the regression tests now go through the public `DocumentTypeRef::validate_update` with `PlatformVersion::latest()` (which selects v1), plus a protocol-11 case, so a missing dispatch arm or stale feature-version constant fails a test instead of slipping through.

packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/mod.rs
QuantumExplorer: Addressed in f71cb817 — replaced the substring matching with `assert_matches!` on the wrapped variant `Error::Protocol(ProtocolError::DataContractError(CorruptedSerialization(_) | DecodingContractError(_)))`.

packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/mod.rs
```diff
@@ -3116,4 +3116,424 @@ pub(in crate::execution) mod tests {
...
+            let err_msg = format!("{:?}", err);
+            assert!(
+                err_msg.contains("Corrupted")
+                    || err_msg.contains("orrupted")
+                    || err_msg.contains("erializ")
+                    || err_msg.contains("ecod"),
+                "expected a serialization/corruption/decoding error from the \
+                 encoding flip, got: {}",
+                err_msg
+            );
```
thepastaclaw: **💬 Nitpick: Regression assertion is too loose to prove the encoding-flip code path**

The added byte-array chain-halt regression accepts any Debug output containing 'Corrupted', 'orrupted', 'erializ', or 'ecod'. That set is broad enough to pass for unrelated decoding or serialization errors anywhere in the contested-vote-poll resolver path, so the test can keep passing even if the encoding-flip-specific code path is no longer exercised (e.g., a refactor that fails earlier with a different error). Match the concrete wrapped error variant — or at least a single, narrower expected message tied to the byte-array decode path — so the regression keeps proving the specific halt it documents.

<sub>source: ['codex']</sub>

thepastaclaw (commented):
## Code Review

Verified the byte-array encoding-stability fix against current code. The validate_update_v0 guard correctly mirrors the encode_value_ref_with_size fixed/variable encoding branch, runs before validate_schema (so it cannot be bypassed by JSON-schema-compat widening), is symmetric and per-property via flattened_properties(), and now has the maxItems-removal test added in the latest delta. The 'tighten v0 in place' concern (raised by 3 Codex agents) was reconciled against repo context: PR targets v3.1-dev (in-development), is explicitly marked breaking with fix(dpp)!:, and matches the immediately preceding chain-halt fix #3865 (ac1cd635a8) on the same branch which also modified validate_structure/v0 in place — this is the established pattern for in-flight v3.1 chain-halt fixes destined for the upcoming protocol-version bump, so it is intentionally deferred rather than a public blocker. Only carryover is a prior nitpick about loose substring matching in the drive-abci regression assertion.

💬 1 nitpick(s)

packages/rs-dpp/src/data_contract/document_type/methods/validate_update/v1/mod.rs
```diff
@@ -0,0 +1,256 @@
...
+        new_document_type: DocumentTypeRef,
+    ) -> SimpleConsensusValidationResult {
+        // `Some(n)` => fixed raw encoding of length `n`; `None` => variable
+        // (varint length-prefixed) encoding.
+        fn fixed_length(sizes: &ByteArrayPropertySizes) -> Option<u16> {
+            match (sizes.min_size, sizes.max_size) {
+                (Some(min), Some(max)) if min == max => Some(min),
+                _ => None,
+            }
+        }
```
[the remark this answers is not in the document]
QuantumExplorer: Verified against the encoder and respectfully not applying this. `encode_value_ref_with_size` (property/mod.rs:1067) uses the raw, no-length-prefix path ONLY for `(Some(min), Some(max)) if min == max`; an omitted `minItems` (`None`) always takes the varint path. So `{maxItems: 0}` (implicit min `None`) is varint-encoded, not fixed — `fixed_length` deliberately mirrors that. Normalizing min to 0 would diverge from the encoder: it would falsely reject safe `{maxItems:0} <-> {maxItems:N}` changes and miss the real `{maxItems:0}` (varint) -> `{minItems:0,maxItems:0}` (raw) flip. Added `accepts_max_items_change_when_min_items_is_omitted` plus a clarifying comment in 18775d81.

packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/mod.rs
QuantumExplorer: Already addressed in f71cb817 — the assertion now uses `assert_matches!` on the wrapped variant `Error::Protocol(ProtocolError::DataContractError(CorruptedSerialization(_) | DecodingContractError(_)))` instead of the substring match.

thepastaclaw (commented):
test dry body

thepastaclaw (commented):
## Code Review

Prior finding (loose substring assertion in the drive-abci regression test) is fully resolved — the test now uses `assert_matches!` against the `DataContractError::CorruptedSerialization | DecodingContractError` variants, which proves the byteArray encoding-flip code path without coupling to Debug-string format. Version gating is correct: `DPP_VALIDATION_VERSIONS_V3` (protocol 12) selects `validate_update: 1`, while v1/v2 keep `validate_update: 0`, so v0 stays bit-exact for already-shipped protocols and the new check is consensus-gated. No new in-scope blocking issues, suggestions, or nitpicks across both review agents.

thepastaclaw (approved):
## Code Review

Latest delta (180fa0ed→18775d81) is purely additive: a clarifying doc comment on `fixed_length` in `validate_update_v1` and a positive test pinning that an omitted `minItems` is not treated as fixed-length (matching the encoder branch at property/mod.rs:1067). Version gating is correct (V1/V2 keep `validate_update: 0`; V3/protocol 12 selects `1`), tests exercise the public dispatcher, and the drive-abci regression uses `assert_matches!` against `CorruptedSerialization | DecodingContractError`. All prior findings remain resolved; no new in-scope issues.

packages/rs-dpp/src/data_contract/document_type/methods/validate_update/v0/mod.rs
QuantumExplorer: Update: per maintainer decision, switched to the in-place approach in 7de91788 — the byte-array encoding-stability check now lives in `validate_update_v0` (running before `validate_schema`, so it cannot be bypassed by a JSON-schema-compatible widening), matching the sibling chain-halt fix #3865 which added its rejection rule to `validate_structure_v0` in place. Reverted `validate_update_v1`, the dispatcher arm, and the `DPP_VALIDATION_VERSIONS_V3` feature-version bump. This aligns with the established in-flight v3.1 halt-fix pattern you noted in the review summary.

packages/rs-dpp/src/data_contract/document_type/methods/validate_update/v0/mod.rs
```diff
@@ -33,10 +34,84 @@ impl DocumentTypeRef<'_> {
...
 
+        // Validate that no byte array property changes its on-disk encoding
+        let result = self.validate_byte_array_encoding_stability(new_document_type);
+
+        if !result.is_valid() {
+            return Ok(result);
+        }
+
         // Validate schema compatibility
         self.validate_schema(new_document_type, platform_version)
```
thepastaclaw, line 45: **🔴 Blocking: Byte-array encoding rule changes validate_update_v0 semantics**

The new byte-array encoding-stability check now runs inside validate_update_v0, and the dispatcher only has the 0 => validate_update_v0 arm. All current DPP validation tables, DPP_VALIDATION_VERSIONS_V1, V2, and V3, still set document_type.validate_update to 0, so this tightens validation for every pinned protocol version instead of only the intended activation version. This method is reached from the consensus data-contract-update state validation path, so changing feature version 0 can make upgraded binaries reject contract updates that the same protocol version previously accepted. Keep v0 bit-exact, restore the separate validate_update_v1 dispatcher arm, and select version 1 only in the platform validation version where this rule is meant to activate.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):
## Code Review

Verified the worktree is exactly at 7de917886b19a9352602a3b1f7ee86ee19adf181. The current delta reintroduces an in-scope consensus-versioning regression: the byte-array encoding rule is now applied through feature version 0 for every DPP validation version, whereas the prior reviewed commit 18775d81 had this behind validate_update_v1 and selected it only from V3.

🔴 1 blocking

<details>

```

In `packages/rs-dpp/src/data_contract/document_type/methods/validate_update/v0/mod.rs`:
- [BLOCKING] packages/rs-dpp/src/data_contract/document_type/methods/validate_update/v0/mod.rs:37-45: Byte-array encoding rule changes validate_update_v0 semantics
  The new byte-array encoding-stability check now runs inside validate_update_v0, and the dispatcher only has the 0 => validate_update_v0 arm. All current DPP validation tables, DPP_VALIDATION_VERSIONS_V1, V2, and V3, still set document_type.validate_update to 0, so this tightens validation for every pinned protocol version instead of only the intended activation version. This method is reached from the consensus data-contract-update state validation path, so changing feature version 0 can make upgraded binaries reject contract updates that the same protocol version previously accepted. Keep v0 bit-exact, restore the separate validate_update_v1 dispatcher arm, and select version 1 only in the platform validation version where this rule is meant to activate.
```
</details>

packages/rs-dpp/src/data_contract/document_type/methods/validate_update/v0/mod.rs
QuantumExplorer: This is intentional, per maintainer decision — not an oversight. The PR deliberately uses the in-place approach to match the sibling chain-halt fix #3865 (ac1cd635a8, already merged on v3.1-dev), which likewise added its new rejection rule to `validate_structure_v0` in place with no feature-version gating. These in-flight v3.1 chain-halt fixes ship together at the upcoming, not-yet-activated protocol-12 bump, so there is no live mixed-validator window for protocol 12; this is the established pattern (and your earlier review summary reconciled exactly this point). A version-gated variant exists in git history (commit 18775d81) if a gated landing is ever preferred.

thepastaclaw: Resolved in `14ba948` — _Byte-array encoding rule changes validate_update_v0 semantics_ no longer present.

thepastaclaw: Correction: the auto-resolve note above was wrong. I rechecked current head `14ba9484`: this finding is **still valid**. The byte-array encoding guard still runs inside `validate_update_v0` (`v0/mod.rs:37-45`), the public dispatcher still only has the `0 => validate_update_v0` arm, and the DPP validation version tables still select `validate_update: 0`. This thread should remain unresolved until the guard is moved behind a new versioned validation path.

thepastaclaw (changes requested):
## Code Review

Correcting the earlier empty same-SHA review / auto-resolve reply: the prior 7de91788 blocking finding is **STILL VALID** at `14ba9484`. The latest delta only adds a reverse-direction unit test; it does not move the new consensus validation rule out of feature version 0 or update the dispatcher/version tables.

🔴 1 blocking

<details>

```

In `packages/rs-dpp/src/data_contract/document_type/methods/validate_update/v0/mod.rs`:
- [BLOCKING] packages/rs-dpp/src/data_contract/document_type/methods/validate_update/v0/mod.rs:37-45: Byte-array encoding guard changes pinned validation v0 semantics
  This PR adds the byte-array encoding-stability rejection directly inside `validate_update_v0`. The public `DocumentTypeRef::validate_update` dispatcher still only accepts feature version 0, and the DPP validation tables still select `document_type.validate_update: 0`, so upgraded binaries would reject a class of contract updates under the same pinned platform validation version that older binaries accepted. Keep v0 bit-exact, add a separate `validate_update_v1` path for this new rule, and select that version only in the platform validation version where the behavior is intended to activate.
```
</details>

thepastaclaw:
Coverage note for the 14ba9484 corrective review: Codex general, Codex security-auditor, Codex rust-quality, and Codex verifier completed. Claude general/security/rust-quality and Claude verifier did not complete because the Claude ACP path hit a usage limit before producing JSON.

QuantumExplorer (commented):
Approved

PR #3875, opened 2026-06-12
fix(sdk): ban node and retry elsewhere on UNIMPLEMENTED responses

QuantumExplorer:

rs-dapi-client classifies gRPC `Unimplemented` as non-retryable, so the error surfaces directly to the caller. For a stable network that's correct, but during a mixed-version rollout (e.g. shielded endpoints newly activated on testnet 4.0.0-rc.2) UNIMPLEMENTED actually means "this particular evonode runs an older build" — the same request succeeds on an upgraded node.

Observed 2026-06-12 on testnet: SwiftExampleApp's shielded sync failed with `Dapi client error: transport error: grpc error: 'Operation is not implemented or not supported'` from one rotation node, while `seed-1.testnet.networks.dash.org:1443` (and subsequent syncs) served all six `getShielded*` methods fine.

## What was done?

Reclassified `Code::Unimplemented` as retryable in `CanRetry for tonic::Status`. In the executor, ban-and-failover *is* the retryable path: `update_address_ban_status` bans the failed address only for retryable errors, and the retry loop then selects a fresh live address. So the stale node gets banned (60s base, exponential backoff on repeat) and the request transparently retries on another node.

This is bounded, not infinitely retryable. How an all-UNIMPLEMENTED network terminates depends on which of `DapiClient::execute`'s two caps trips first:

- `live_addresses <= settings.retries`: the address list is exhausted before the retry budget, so `get_live_address()` runs dry and the executor surfaces `NoAvailableAddressesToRetry` wrapping the last UNIMPLEMENTED status — itself non-retryable, so caller-level retry wrappers (e.g. rs-sdk's `retry`) won't loop on a genuinely unsupported network.
- `live_addresses > settings.retries`: the per-call retry budget (`settings.retries`, default 5) trips first, banning `settings.retries + 1` nodes and surfacing the raw (still-retryable) `Unimplemented`. A caller that honors `CanRetry` then re-enters and bans more nodes each round, converging on the exhausted-addresses case above; rs-sdk additionally bounds this with its own `total_retries` cap. Either way the loop is bounded — it never retries the same node forever.

Caveat (follow-up): bans are address-scoped, not `(address, method)`-scoped, and `Unimplemented` is also returned deliberately for genuinely unsupported / feature-gated endpoints (rs-drive-abci's `broadcast_state_transition` / `wait_for_state_transition_result` / `get_consensus_params`, and any rs-dapi `MethodNotFound`). Calling such an endpoint now bans an otherwise-healthy node across all methods for the ban window. Acceptable for the targeted mixed-version-rollout case; per-method ban granularity is left as a follow-up (documented inline as `TODO(per-method-ban)`).

Server-side uses of `Unimplemented` (rs-dapi error mapping, drive-abci handler codes) are unaffected; nothing else in the workspace relied on the old client-side classification.

## How Has This Been Tested?

- Updated the `can_retry()` classification unit tests in `transport.rs`.
- Integration tests in `tests/unimplemented_failover.rs` drive the real `DapiClient::execute` retry loop with a fake transport (the existing `MockDapiClient` bypasses the executor, so it can't cover this):
  - mixed-version: with two nodes where the first attempt answers UNIMPLEMENTED, the request succeeds with `retries == 1`, the response comes from a different address, and the stale node is banned in the `AddressList`;
  - all-nodes-stale, small pool (`live_addresses <= settings.retries`): both nodes are tried exactly once, the surfaced error is `NoAvailableAddressesToRetry(Unimplemented)` and reports non-retryable;
  - all-nodes-stale, pool larger than the retry budget (`live_addresses > settings.retries`): with 3 nodes and `retries = 1`, the retry cap trips first — exactly `retries + 1 = 2` nodes are banned and the surfaced raw `Unimplemented` is still retryable. This pins the branch where the all-stale collapse does *not* hold, so the inline comment and this description match the actual contract.
- `cargo test -p rs-dapi-client` (default and `--all-features`), `cargo clippy -p rs-dapi-client --all-features --tests`, `cargo fmt` — all clean.

## Breaking Changes

None.

## Checklist:

### DIFF
--- a/packages/rs-dapi-client/src/transport/grpc.rs
+++ b/packages/rs-dapi-client/src/transport/grpc.rs
@@ -114,6 +114,13 @@ impl CanRetry for dapi_grpc::tonic::Status {
                 | Aborted
                 | Internal
                 | Unavailable
+                // During a mixed-version network rollout, Unimplemented means this
+                // particular node runs an older build that doesn't expose the method
+                // yet; another node may serve it. Marking it retryable makes the
+                // executor ban the node and retry elsewhere. If no node implements
+                // the method, all addresses get exhausted and the error still
+                // surfaces as NoAvailableAddressesToRetry.
+                | Unimplemented
         )
     }
 }
--- a/packages/rs-dapi-client/src/transport/grpc.rs
+++ b/packages/rs-dapi-client/src/transport/grpc.rs
@@ -117,9 +117,32 @@ impl CanRetry for dapi_grpc::tonic::Status {
                 // During a mixed-version network rollout, Unimplemented means this
                 // particular node runs an older build that doesn't expose the method
                 // yet; another node may serve it. Marking it retryable makes the
-                // executor ban the node and retry elsewhere. If no node implements
-                // the method, all addresses get exhausted and the error still
-                // surfaces as NoAvailableAddressesToRetry.
+                // executor ban the node (see `update_address_ban_status`) and retry
+                // elsewhere.
+                //
+                // When NO node implements the method, how the failure surfaces depends
+                // on which of `DapiClient::execute`'s two caps trips first:
+                //   * `live_addresses <= settings.retries` — the address list is
+                //     exhausted before the retry budget, so the error surfaces as the
+                //     non-retryable `NoAvailableAddressesToRetry`.
+                //   * `live_addresses > settings.retries` — the per-call retry budget
+                //     trips first, banning `settings.retries + 1` nodes and surfacing
+                //     the raw (still-retryable) `Unimplemented`. A caller that honors
+                //     `CanRetry` then re-enters and bans more nodes each round,
+                //     converging on the exhausted-addresses case above; rs-sdk also
+                //     bounds this via its own `total_retries` cap (see
+                //     `rs-sdk/src/sync.rs`). Either way the loop is bounded — it never
+                //     retries the same node forever.
+                //
+                // Caveat: bans are address-scoped, not (address, method)-scoped, and
+                // Unimplemented is also returned deliberately for genuinely unsupported
+                // / feature-gated endpoints (e.g. rs-drive-abci's
+                // broadcast_state_transition / wait_for_state_transition_result /
+                // get_consensus_params, and any rs-dapi `MethodNotFound`). Calling such
+                // an endpoint bans an otherwise-healthy node across ALL methods for the
+                // ban window. Acceptable for the targeted rollout case; per-method ban
+                // granularity is left as a follow-up so optional endpoints can't poison
+                // the shared address list. TODO(per-method-ban).
                 | Unimplemented
         )
     }
--- a/packages/rs-dapi-client/src/transport.rs
+++ b/packages/rs-dapi-client/src/transport.rs
@@ -166,6 +166,9 @@ mod tests {
             Code::Aborted,
             Code::Internal,
             Code::Unavailable,
+            // Retryable on another node: an old build in a mixed-version network
+            // returns Unimplemented for methods newer nodes already serve.
+            Code::Unimplemented,
         ];
 
         for code in retryable_codes {
@@ -187,7 +190,6 @@ mod tests {
             Code::PermissionDenied,
             Code::FailedPrecondition,
             Code::OutOfRange,
-            Code::Unimplemented,
             Code::Unauthenticated,
         ];
 
--- a/packages/rs-dapi-client/tests/unimplemented_failover.rs
+++ b/packages/rs-dapi-client/tests/unimplemented_failover.rs
@@ -168,3 +168,55 @@ async fn unimplemented_on_all_nodes_still_surfaces_error() {
         other => panic!("expected NoAvailableAddressesToRetry, got: {other:?}"),
     }
 }
+
+#[tokio::test]
+async fn unimplemented_surfaces_retryable_when_pool_exceeds_retry_budget() {
+    // The non-retryable `NoAvailableAddressesToRetry` collapse (above) only holds
+    // when the live address list is exhausted before the retry budget, i.e.
+    // `live_addresses <= settings.retries`. When the pool is LARGER than the retry
+    // budget, the per-call retry cap trips first: the executor surfaces the raw
+    // (still-retryable) `Unimplemented` after banning `retries + 1` nodes, leaving
+    // the rest of the pool live. A caller honoring `CanRetry` re-enters and bans
+    // more nodes each round (rs-sdk additionally caps via `total_retries`), so
+    // termination is still bounded — this test pins the contract for that branch.
+    let settings = RequestSettings {
+        retries: Some(1),
+        ..Default::default()
+    };
+
+    // Three live nodes, retry budget of 1 → at most retries + 1 = 2 attempts, so
+    // the retry cap trips before all three addresses are banned.
+    let request = MixedVersionRequest::with_unimplemented_responses(usize::MAX);
+    let address_list: AddressList =
+        "http://127.0.0.1:10005,http://127.0.0.1:10006,http://127.0.0.1:10007"
+            .parse()
+            .expect("valid address list");
+    let client = DapiClient::new(address_list, settings);
+
+    let error = client
+        .execute(request.clone(), settings)
+        .await
+        .expect_err("request must fail when no node implements the method");
+
+    // retries + 1 = 2 nodes were tried (and banned) before the retry cap tripped;
+    // the third address is never reached.
+    assert_eq!(
+        request.state.lock().unwrap().unimplemented_uris.len(),
+        2,
+        "retry budget (1) caps attempts at 2 before the 3-node pool is exhausted"
+    );
+
+    // Because the retry cap — not address exhaustion — terminated the loop, the
+    // raw Unimplemented surfaces and is still retryable (unlike the exhausted-pool
+    // case, which collapses to the non-retryable NoAvailableAddressesToRetry).
+    assert!(
+        error.can_retry(),
+        "with more live nodes than retries, the surfaced Unimplemented stays retryable"
+    );
+    match error.inner {
+        DapiClientError::Transport(TransportError::Grpc(status)) => {
+            assert_eq!(status.code(), Code::Unimplemented)
+        }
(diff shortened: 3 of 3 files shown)


thepastaclaw:
✅ Review complete (commit cb468d88)

packages/rs-dapi-client/src/transport/grpc.rs
```diff
@@ -114,6 +114,13 @@ impl CanRetry for dapi_grpc::tonic::Status {
                 | Aborted
                 | Internal
                 | Unavailable
+                // During a mixed-version network rollout, Unimplemented means this
+                // particular node runs an older build that doesn't expose the method
+                // yet; another node may serve it. Marking it retryable makes the
+                // executor ban the node and retry elsewhere. If no node implements
+                // the method, all addresses get exhausted and the error still
+                // surfaces as NoAvailableAddressesToRetry.
+                | Unimplemented
```
thepastaclaw, line 146: **🟡 Suggestion: All-stale collapse to NoAvailableAddressesToRetry only holds when live_addresses <= settings.retries**

The new comment and the PR description state that if every node answers UNIMPLEMENTED, the error surfaces as NoAvailableAddressesToRetry (non-retryable). That invariant only holds when the address list is exhausted before the retry budget is. In `DapiClient::execute` (packages/rs-dapi-client/src/dapi_client.rs:681) the inner loop exits via the `retries < max_retries` cap before `get_live_address()` returns None whenever `live_addresses > max_retries`. In that branch the executor returns the raw `DapiClientError::Transport(Grpc(Unimplemented))`, and with this change `can_retry()` is now true on that surfaced error. Direct rs-dapi-client callers that obey `CanRetry` (anyone not going through `rs-sdk/src/sync.rs:131`, which caps via `total_retries > max_retries`) can keep retrying a genuinely unsupported method. The integration test in `tests/unimplemented_failover.rs:143` uses 2 addresses with the default 5 retries and so cannot exercise this branch. Either tighten the executor so an UNIMPLEMENTED that has visited every live address surfaces as a terminal/non-retryable error regardless of which cap triggers first, or add a coverage case with `live_addresses > settings.retries` and adjust the comment/PR claim to match the actual contract.

packages/rs-dapi-client/src/transport/grpc.rs
```diff
@@ -114,6 +114,13 @@ impl CanRetry for dapi_grpc::tonic::Status {
                 | Aborted
                 | Internal
                 | Unavailable
+                // During a mixed-version network rollout, Unimplemented means this
+                // particular node runs an older build that doesn't expose the method
+                // yet; another node may serve it. Marking it retryable makes the
+                // executor ban the node and retry elsewhere. If no node implements
+                // the method, all addresses get exhausted and the error still
+                // surfaces as NoAvailableAddressesToRetry.
```
thepastaclaw: **🟡 Suggestion: Address-scoped ban for UNIMPLEMENTED can degrade healthy nodes serving other methods**

`update_address_ban_status` (packages/rs-dapi-client/src/dapi_client.rs:162) bans by `Address`, not by `(address, method)`. UNIMPLEMENTED is not only emitted by stale rollout nodes: `packages/rs-drive-abci/src/query/service.rs:262,271,446,453` deliberately returns `Status::unimplemented` for `broadcast_state_transition`, `wait_for_state_transition_result`, and `get_consensus_params`, and `packages/rs-dapi/src/error.rs:163` maps every `DapiError::MethodNotFound` to `Status::unimplemented`. After this change, a call to any such endpoint bans that otherwise-healthy node across all methods for the full ban window (60s base, exponential on repeat). On clients with small address pools this can make unrelated supported requests fail with `NoAvailableAddresses` until the ban expires. The targeted mixed-version rollout case justifies the trade-off, but it should at least be documented inline and tracked as a follow-up for per-method ban granularity so optional / feature-gated endpoints do not poison the shared address list.

thepastaclaw (commented):
## Code Review

Small, well-scoped change that correctly routes gRPC UNIMPLEMENTED through the existing ban-and-retry path so stale nodes get rotated out during a mixed-version rollout. Two real concerns survive validation: (1) the PR description's all-stale safety claim only holds when live_addresses <= settings.retries — with more nodes than the retry budget, callers observe a still-retryable Unimplemented; (2) UNIMPLEMENTED is now address-scoped, so calling any genuinely unsupported method (e.g. broadcast_state_transition / wait_for_state_transition_result / get_consensus_params from rs-drive-abci or any MethodNotFound mapped by rs-dapi) bans an otherwise healthy node for the full ban window. Both are bounded by existing rs-sdk total_retries cap and the ban TTL, so I treat them as suggestions rather than blockers.

🟡 2 suggestion(s)

<details>

```

In `packages/rs-dapi-client/src/transport/grpc.rs`:
- [SUGGESTION] packages/rs-dapi-client/src/transport/grpc.rs:114-123: All-stale collapse to NoAvailableAddressesToRetry only holds when live_addresses <= settings.retries
  The new comment and the PR description state that if every node answers UNIMPLEMENTED, the error surfaces as NoAvailableAddressesToRetry (non-retryable). That invariant only holds when the address list is exhausted before the retry budget is. In `DapiClient::execute` (packages/rs-dapi-client/src/dapi_client.rs:681) the inner loop exits via the `retries < max_retries` cap before `get_live_address()` returns None whenever `live_addresses > max_retries`. In that branch the executor returns the raw `DapiClientError::Transport(Grpc(Unimplemented))`, and with this change `can_retry()` is now true on that surfaced error. Direct rs-dapi-client callers that obey `CanRetry` (anyone not going through `rs-sdk/src/sync.rs:131`, which caps via `total_retries > max_retries`) can keep retrying a genuinely unsupported method. The integration test in `tests/unimplemented_failover.rs:143` uses 2 addresses with the default 5 retries and so cannot exercise this branch. Either tighten the executor so an UNIMPLEMENTED that has visited every live address surfaces as a terminal/non-retryable error regardless of which cap triggers first, or add a coverage case with `live_addresses > settings.retries` and adjust the comment/PR claim to match the actual contract.
- [SUGGESTION] packages/rs-dapi-client/src/transport/grpc.rs:114-122: Address-scoped ban for UNIMPLEMENTED can degrade healthy nodes serving other methods
  `update_address_ban_status` (packages/rs-dapi-client/src/dapi_client.rs:162) bans by `Address`, not by `(address, method)`. UNIMPLEMENTED is not only emitted by stale rollout nodes: `packages/rs-drive-abci/src/query/service.rs:262,271,446,453` deliberately returns `Status::unimplemented` for `broadcast_state_transition`, `wait_for_state_transition_result`, and `get_consensus_params`, and `packages/rs-dapi/src/error.rs:163` maps every `DapiError::MethodNotFound` to `Status::unimplemented`. After this change, a call to any such endpoint bans that otherwise-healthy node across all methods for the full ban window (60s base, exponential on repeat). On clients with small address pools this can make unrelated supported requests fail with `NoAvailableAddresses` until the ban expires. The targeted mixed-version rollout case justifies the trade-off, but it should at least be documented inline and tracked as a follow-up for per-method ban granularity so optional / feature-gated endpoints do not poison the shared address list.
```
</details>

thepastaclaw (commented):
## Code Review

Carried-forward prior findings: prior-1 (FIXED — grpc.rs now documents both termination branches and a new test pins the `live_addresses > settings.retries` case); prior-2 (INTENTIONALLY DEFERRED — explicit `TODO(per-method-ban)` and inline caveat in grpc.rs:137-145). New findings in the latest delta: one real bug in rs-sdk error wrapping — `Error::can_retry()` does not delegate to the wrapped `DapiClientError`, so the rs-dapi-client comment's claim that rs-sdk re-enters and converges on `NoAvailableAddressesToRetry` is false for SDK callers; in the `live_addresses > settings.retries` branch, shielded sync can stop before reaching an upgraded node. The fix is a one-line delegation.

🟡 1 suggestion(s)

<details>

```

In `packages/rs-sdk/src/error.rs`:
- [SUGGESTION] packages/rs-sdk/src/error.rs:341-347: rs-sdk Error::can_retry drops retryability of wrapped DapiClientError, breaking the mixed-version failover convergence the PR documents
  The new `Unimplemented` reclassification in `packages/rs-dapi-client/src/transport/grpc.rs:117-146` explicitly relies on callers honoring `CanRetry` to re-enter `DapiClient::execute` and converge on `NoAvailableAddressesToRetry` in the `live_addresses > settings.retries` branch (see the comment block: "A caller that honors `CanRetry` then re-enters and bans more nodes each round … rs-sdk also bounds this via its own `total_retries` cap (see `rs-sdk/src/sync.rs`)").

  But `Error::can_retry()` here only matches `StaleNode | TimeoutReached | Proof`. When `DapiClient::execute` returns a raw retryable `DapiClientError::Transport(Grpc(Unimplemented))` (the `live_addresses > settings.retries` branch), `From<DapiClientError> for Error` wraps it as `Error::DapiClientError(value)` (line 221), and `sync::retry` at `packages/rs-sdk/src/sync.rs:124` evaluates `!error.can_retry()` → terminates immediately. The SDK only bans `settings.retries + 1` stale nodes and surfaces `Unimplemented` to the caller instead of converging on exhausted addresses as the in-tree documentation claims.

  For the PR's stated mixed-version rollout scenario (shielded sync over a pool larger than the per-call retry budget), SDK callers can still fail before reaching an upgraded node. The new `unimplemented_failover` integration test covers the rs-dapi-client boundary but not the SDK wrapper boundary, so this regression is invisible at test time.

  Delegating to the wrapped error's `can_retry()` closes the gap and aligns rs-sdk with the contract the grpc.rs comment promises.
```
</details>

packages/rs-dapi-client/src/transport/grpc.rs
thepastaclaw: Resolved in `cb468d8` — _Address-scoped ban for UNIMPLEMENTED can degrade healthy nodes serving other methods_ no longer present.

packages/rs-dapi-client/src/transport/grpc.rs
thepastaclaw: Resolved in `cb468d8` — _All-stale collapse to NoAvailableAddressesToRetry only holds when live_addresses <= settings.retries_ no longer present.

PR #3800, opened 2026-06-05
fix(drive): charge fees for unshield and shielded withdrawal

QuantumExplorer:
## Issue being fixed or feature implemented

**Unshield and shielded-withdrawal were free (fee bypass / griefing).**

Both transitions hardcoded `fee_amount: 0` (`// TODO`), so validators were paid nothing for verifying the expensive Halo2 proof + storing nullifiers, while the user received the full value back. This makes the operations free and lets an attacker force unbounded proof-verification + storage work on the network at zero cost. Value is conserved (not money-printing), but the sibling `ShieldedTransfer` charges correctly — these two paths just never implemented it.

`UnshieldTransitionV0.unshielding_amount` is documented as the **total** leaving the shielded pool (recipient/net amount + fee), and the builder already sets it to `user_amount + compute_minimum_shielded_fee(...)`, intending the recipient to get `user_amount` and the fee to be the computed minimum. Consensus simply never charged it.

## What was done?

Consensus now charges the same computed minimum fee the builder and the existing `validate_minimum_shielded_fee` already use:

- The transform layer (`unshield`/`shielded_withdrawal` `transform_into_action_v0`) computes `fee_amount = dpp::shielded::compute_minimum_shielded_fee(num_actions, platform_version)` — the exact value that `validate_minimum_shielded_fee` already enforces `unshielding_amount >=`, so the net (`unshielding_amount - fee`) is guaranteed non-negative — and passes it into the action.
- **Unshield** conversion: credits the output address `unshielding_amount - fee` (not the full amount); decrements the pool by `unshielding_amount` (not `amount + fee`). Net: pool `-= amount`, recipient `+= (amount - fee)`, fee pools `+= fee` (via the existing `PaidFromShieldedPool { fees_to_add_to_pool: fee_amount }`) → conserved, and the pool decrement equals the proof's `value_balance`.
- **Shielded withdrawal**: the prepared withdrawal document records the **net** `unshielding_amount - fee` sent to Core; `RemoveFromSystemCredits` uses the net; the pool decrements by `unshielding_amount`. The fee stays in-platform and flows to the fee pools, so the sum trees (`pool -amount`, `fee pools +fee`) and the `total_credits_in_platform` counter (`-(amount - fee)`) both drop by exactly the net.

The misleading `validate_minimum_shielded_fee` comment ("fee = value_balance - amount") is corrected; the check `unshielding_amount >= min_fee` is unchanged (it now correctly guarantees the net amount is non-negative). All checked/guarded against underflow.

[...]

PR #3877, opened 2026-06-13
fix(drive): unify shielded per-action processing fee across all protocol versions

QuantumExplorer:

The shielded per-action processing fee (`shielded_per_action_processing_fee`) was only calibrated to its real value — **22,000,000 credits**, pricing the ~1.1 ms/action Halo 2 verification CPU — in **validation v8** ([#3800](https://github.com/dashpay/platform/pull/3800)). Validation **v1–v7** still carried a **3,000,000** placeholder.

This skew is a live hazard for any client whose protocol-version view lags the network. The SDK auto-detect ratchet seeds at an older protocol version and only advances after it parses a metadata-bearing platform response. A wallet computing the `ShieldFromAssetLock` pool fee while still seeded below v12 uses the **3M** constant, so it reserves `lock_value − (proof_fee + actions × 3M + albc)` instead of the `22M`-priced amount consensus actually charges. The transition then under-funds by `19M × num_actions` and is rejected with `IdentityAssetLockTransactionOutPointNotEnoughBalanceError`.

**Observed on testnet (4.0.0-rc.2):** shielding 0.1 tDASH built a valid 2-action bundle but was rejected — `credits_required: 10,038,000,000` vs `credits_left: 10,000,000,000`. The 38M shortfall is exactly `2 actions × 19M`, the v8 bump the client never saw (its shielded sync queries were failing UNIMPLEMENTED against stale evonodes, so no v12 metadata ever arrived to advance the ratchet).

## What was done?

Set `shielded_per_action_processing_fee` to `22_000_000` in validation versions **v1 through v7** (v8 already had it), making the constant uniform across every protocol version. Each changed site carries a comment explaining why a frozen older version's constant is being updated.

Because the per-action fee is now version-invariant, a client that computes the fee under *any* protocol version reserves the consensus-correct amount — the version-skew can no longer produce a wrong reservation. This is a more robust fix than forcing every fee-sensitive client path to refresh its protocol version before building.

## How Has This Been Tested?

- `cargo check -p platform-version` — clean.
- `cargo test -p dpp --lib compute_minimum_shielded_fee` — all 4 shielded-fee tests pass (incl. `compute_minimum_shielded_fee_v0_equals_historical_formula`, which reads the constant dynamically via `PlatformVersion::latest()`).
- `cargo fmt` applied.
- Verified no test or fixture hardcodes the old `3_000_000` value; the only references are the version constants themselves and the formula that reads them generically.

## Breaking Changes

None to any deployed network. The shielded family does not activate until platform **v12** (validation **v8**), which already used `22M` — so no block under any protocol version has ever priced a shielded action at the old `3M` placeholder. Re-validating historical blocks is unaffected (they contain no shielded transitions), and the live network's fee is unchanged. This is why the change is not marked consensus-breaking (`fix:`, not `fix!:`).

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated shielded action processing fees across protocol versions v1–v7 to ensure correct fee reservation for clients using stale protocol versions.
  * Corrected masternode vote state transition nonce validation in protocol version v5.

### DIFF
--- a/packages/rs-platform-version/src/version/drive_abci_versions/drive_abci_validation_versions/v1.rs
+++ b/packages/rs-platform-version/src/version/drive_abci_versions/drive_abci_validation_versions/v1.rs
@@ -275,7 +275,11 @@ pub const DRIVE_ABCI_VALIDATION_VERSIONS_V1: DriveAbciValidationVersions =
             shielded_anchor_retention_blocks: 1000,
             shielded_anchor_pruning_interval: 100,
             shielded_proof_verification_fee: 100_000_000,
-            shielded_per_action_processing_fee: 3_000_000,
+            // Uniform with v8 (#3800): the shielded family only activates at v12, so no
+            // earlier-version block ever priced a shielded action under the old placeholder.
+            // Pinning every version to the same per-action fee lets a client computing the
+            // fee under a stale protocol version still reserve the consensus-correct amount.
+            shielded_per_action_processing_fee: 22_000_000,
             shielded_implicit_fee_cap: 20_000_000_000,
             shielded_identity_create_denominations: &[],
         },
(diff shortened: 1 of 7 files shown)


thepastaclaw:
✅ Review complete (commit cdb9b30e)

QuantumExplorer (commented):
Approved

thepastaclaw (approved):
## Code Review

Narrow, correct fix unifying the dormant placeholder `shielded_per_action_processing_fee` (3M → 22M) in validation versions v1-v7 to match v8. Shielded transitions activate only at platform v12 (validation v8, already 22M), so no historical block ever priced a shielded action under the old constant — change is not consensus-affecting. The fix closes a real stale-protocol-version client-side under-reservation hazard documented in the commit message with an observed testnet failure mode.

PR #3881, opened 2026-06-13
fix(platform-wallet): build data contract config at the protocol-required version

QuantumExplorer:

Creating ANY data contract through the SwiftExampleApp (and any other client routing through `platform-wallet`) is rejected on protocol v12 (testnet) with:

> Protocol error: Can't update Data Contract <id> config: config version 0 is not supported, minimum version is 1

This blocks all contract creation — both token and document contracts.

**Root cause:** `packages/rs-platform-wallet/src/wallet/identity/network/contract.rs` (`create_data_contract_with_signer`) hardcoded `"$formatVersion": "0"` when tagging the assembled `DataContractConfig`. Since protocol v12, `DataContractConfig` V0 is no longer accepted — the data-contract-create basic-structure validator (`rs-drive-abci/.../data_contract_create/basic_structure/v1/mod.rs`) enforces `config().version() >= platform_version.dpp.contract_versions.config.min_version`, and that minimum is `1` for v12 (see `rs-platform-version/.../dpp_contract_versions/v3.rs` and `v4.rs`: `config.min_version = 1`, "V0 config no longer accepted — lacks sized_integer_types"). So every contract the wallet built carried a V0 config and was rejected.

## What was done?

- In `contract.rs`, the assembled config is now tagged with the running `PlatformVersion`'s `dpp.contract_versions.config.default_current_version` (which equals `min_version` on v12) instead of a hardcoded `"0"`.
- A `config` block is now always inserted — including when the caller passes no `config_json` — so the version tag is explicit rather than relying on the serialization format's `#[serde(default = "DataContractConfigV1::default_with_version")]`.
- A flags-only or empty config dict deserializes into a valid V1: `DataContractConfigV1` is `#[serde(rename_all = "camelCase", default)]`, so every field (including `sized_integer_types: true`) has a serde default. No extra field population is needed.
- Non-object `config_json` now returns a clear `InvalidIdentityData` error instead of being silently dropped.
- Removed the now-duplicate `let platform_version = self.sdk.version();` binding (hoisted earlier for the version lookup).

**Token schema left unchanged (investigated):** `QuickBasicTokenView.swift` emits `"$formatVersion": "0"` for the token schema. `TokenConfiguration` only has a `V0` variant and there is no token-config min-version gate on v12 (`token_versions` carries only `validate_structure_interval`), so V0 is still the only valid token-configuration version. Bumping it would break deserialization, so it was intentionally left alone.

## How Has This Been Tested?

- `cargo check -p platform-wallet` passes (compiles clean, including transitive `dpp`/`drive`/`dash-sdk`).
- `cargo fmt --all --check` passes.
- Full end-to-end iOS contract creation against a v12 testnet node is being verified separately by the orchestrator.

## Breaking Changes

None. This is a client-side fix in the `platform-wallet` crate; it changes the config version the wallet emits to match what the network already requires. No consensus or wire-format change.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed contract creation to properly validate and apply protocol-specified configuration format versions.

* **Tests**
  * Added validation tests for contract configuration handling.

### DIFF
--- a/packages/rs-platform-wallet/src/wallet/identity/network/contract.rs
+++ b/packages/rs-platform-wallet/src/wallet/identity/network/contract.rs
@@ -242,32 +242,12 @@ impl IdentityWallet {
                 serde_json::Value::String(s.to_string()),
             );
         }
-        // `DataContractConfig` is itself a `#[serde(tag =
-        // "$formatVersion")]` enum; the Swift form-builder shape sends
-        // a bare flags dict (or omits config entirely). Tag it with the
-        // protocol-required version so the tagged-enum dispatch picks a
-        // variant the network will accept. `DataContractConfigV1` is
-        // `#[serde(rename_all = "camelCase", default)]`, so a flags-only
-        // (or empty) dict deserializes into a valid V1 — every field,
-        // including `sized_integer_types: true`, has a serde default.
-        //
         // We always insert a `config` block (even when `config_json` is
         // None) so the version tag is explicit rather than relying on
         // the serialization format's `#[serde(default)]`; this keeps the
-        // emitted config aligned with the running protocol version.
-        let mut config_obj = match parse_optional_json("config_json", config_json)? {
-            Some(serde_json::Value::Object(map)) => map,
-            Some(other) => {
-                return Err(PlatformWalletError::InvalidIdentityData(format!(
-                    "config_json must be a JSON object, got: {other}"
-                )));
-            }
-            None => serde_json::Map::new(),
-        };
-        config_obj.insert(
-            "$formatVersion".to_string(),
-            serde_json::Value::String(config_format_version.to_string()),
-        );
+        // emitted config aligned with the running protocol version. See
+        // `build_config_object` for the full assembly contract.
+        let config_obj = build_config_object(config_json, config_format_version)?;
         format_value.insert("config".to_string(), serde_json::Value::Object(config_obj));
 
         // Round-trip through a string instead of `from_value` —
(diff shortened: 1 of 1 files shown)


thepastaclaw:
✅ Review complete (commit 4c31f24c)

packages/rs-platform-wallet/src/wallet/identity/network/contract.rs
```diff
@@ -221,17 +242,33 @@ impl IdentityWallet {
...
             }
-            format_value.insert("config".to_string(), v);
-        }
+            None => serde_json::Map::new(),
+        };
+        config_obj.insert(
+            "$formatVersion".to_string(),
+            serde_json::Value::String(config_format_version.to_string()),
+        );
+        format_value.insert("config".to_string(), serde_json::Value::Object(config_obj));
```
thepastaclaw: **🟡 Suggestion: Add a regression test pinning the config $formatVersion to the protocol's required version**

This PR fixes a real consensus-rejection bug (v12 rejects V0 DataContractConfig because basic-structure validation requires config().version() >= contract_versions.config.min_version). The whole correctness boundary of the change lives in this JSON assembly block, but there's no test in rs-platform-wallet that exercises it, so a future refactor could silently reintroduce the hardcoded "0" or omit the config block without failing CI.

Extracting the config-Map assembly into a small pure helper, e.g. `fn build_config_object(config_json: Option<&str>, version: u16) -> Result<serde_json::Map<String, serde_json::Value>, PlatformWalletError>`, makes this testable without the SDK runtime / signer / wallet_manager dependencies. Three cases are sufficient to lock the fix down:
  - `None` → resulting object has `$formatVersion == default_current_version.to_string()` for the current PlatformVersion (1 on v12).
  - `Some(flags-only object)` → caller fields preserved, version tag set to the protocol default.
  - `Some(non-object)` (array/string/number) → returns `PlatformWalletError::InvalidIdentityData`.

Ideally a fourth case round-trips through `DataContractInSerializationFormat` + `DataContract::try_from_platform_versioned` at the latest PlatformVersion to assert the assembled bytes actually validate. The PR description's own test checklist is unchecked — given this is a fix for a v12 consensus rejection that blocked all contract creation, a guard test here is well worth the few lines.

packages/rs-platform-wallet/src/wallet/identity/network/contract.rs
```diff
@@ -221,17 +242,33 @@ impl IdentityWallet {
...
+                )));
             }
-            format_value.insert("config".to_string(), v);
-        }
+            None => serde_json::Map::new(),
+        };
+        config_obj.insert(
+            "$formatVersion".to_string(),
+            serde_json::Value::String(config_format_version.to_string()),
+        );
```
thepastaclaw: **💬 Nitpick: Caller-supplied `$formatVersion` is now silently overwritten**

Before this PR the code used `entry("$formatVersion").or_insert_with(...)`, preserving any caller-set tag. The new code uses unconditional `insert(...)`, so a caller-supplied `$formatVersion` inside `config_json` is silently dropped — no warning, no error.

In practice this is benign: no caller in this codebase sets the wire-level tag (the whole point of this helper is to abstract it), and on v12+ a caller-supplied `"0"` would have been rejected by the network anyway. But the function now overrides explicit caller intent without signaling it, which makes future debugging harder.

Two reasonable alternatives if this becomes a concern later:
  - Keep the `or_insert_with` shape, but use `config_format_version.to_string()` as the default — still fixes the v12 rejection for the Swift form-builder shape, while honoring an explicit caller tag.
  - If a caller-supplied tag differs from `config_format_version`, return `InvalidIdentityData` with a message that names the protocol's required version, making the override explicit instead of silent.

Not blocking — the current behavior is the most permissive of the three and matches the PR's stated goal of always emitting a network-acceptable tag.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Small, focused fix that correctly tags the assembled DataContractConfig with the running PlatformVersion's contract_versions.config.default_current_version, resolving the v12 testnet rejection where the basic-structure validator enforces config().version() >= contract_versions.config.min_version. Implementation is sound; no blocking issues. Two suggestions: add a regression test pinning the version-tag selection, and decide explicitly how to handle a caller-supplied $formatVersion (currently silently overridden).

<details>

```

In `packages/rs-platform-wallet/src/wallet/identity/network/contract.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/contract.rs:258-271: Add a regression test pinning the config $formatVersion to the protocol's required version
  This PR fixes a real consensus-rejection bug (v12 rejects V0 DataContractConfig because basic-structure validation requires config().version() >= contract_versions.config.min_version). The whole correctness boundary of the change lives in this JSON assembly block, but there's no test in rs-platform-wallet that exercises it, so a future refactor could silently reintroduce the hardcoded "0" or omit the config block without failing CI.

  Extracting the config-Map assembly into a small pure helper, e.g. `fn build_config_object(config_json: Option<&str>, version: u16) -> Result<serde_json::Map<String, serde_json::Value>, PlatformWalletError>`, makes this testable without the SDK runtime / signer / wallet_manager dependencies. Three cases are sufficient to lock the fix down:
    - `None` → resulting object has `$formatVersion == default_current_version.to_string()` for the current PlatformVersion (1 on v12).
    - `Some(flags-only object)` → caller fields preserved, version tag set to the protocol default.
    - `Some(non-object)` (array/string/number) → returns `PlatformWalletError::InvalidIdentityData`.

  Ideally a fourth case round-trips through `DataContractInSerializationFormat` + `DataContract::try_from_platform_versioned` at the latest PlatformVersion to assert the assembled bytes actually validate. The PR description's own test checklist is unchecked — given this is a fix for a v12 consensus rejection that blocked all contract creation, a guard test here is well worth the few lines.
```
</details>

QuantumExplorer:
**QA follow-up (testnet, found while exercising this fix end-to-end):** this change is **necessary but not sufficient** on its own.

It reads the config version from `self.sdk.version()`. On the SwiftExampleApp the wallet's SDK does **not** ratchet to the network's protocol version — it sits at a stale low seed (≤ v8, where `CONTRACT_VERSIONS_V1` makes `config.default_current_version = 0`), so this fix still injects a V0 config and contract creation is still rejected with *"config version 0 is not supported, minimum version is 1"*.

I tried refining both the create and update paths to build the config at the client's compiled `dpp::version::PlatformVersion::latest()` (= v12 → V1) instead of `self.sdk.version()`. That builds a V1 `DataContract` object locally, but the broadcast `put_to_platform_and_wait_for_response(&self.sdk, …)` **serializes against the stale SDK version**, so a V0 config still goes on the wire and is rejected.

**Conclusion:** the real fix is making the wallet's `Sdk` ratchet its `protocol_version` to the network's actual version (v12) — the same stale-SDK-version root cause that blocks shielded identity-create-from-pool (empty denomination set) and originally caused the shielded fee skew (#3877). This PR should still land (correct direction), but contract creation from the app won't actually work until the SDK protocol-version ratchet is fixed. Suggest building config at `latest()` here regardless, and tracking the SDK-ratchet fix as the blocking dependency.

packages/rs-platform-wallet/src/wallet/identity/network/contract.rs
QuantumExplorer: Done — extracted the inline config assembly into a pure, unit-testable helper:

```rust
fn build_config_object(
    config_json: Option<&str>,
    config_format_version: u16,
) -> Result<serde_json::Map<String, serde_json::Value>, PlatformWalletError>
```

`create_data_contract_with_signer` now just calls it with the running `PlatformVersion`'s `contract_versions.config.default_current_version`; behavior is identical. Added `#[cfg(test)] mod tests` with 6 cases that lock the fix down (all keyed off `PlatformVersion::latest()` so they track the real protocol default — 1 on v12 — rather than a hardcoded literal):

- `None` → `$formatVersion == default_current_version.to_string()`
- empty string → behaves like `None`
- `Some(flags-only object)` (`{"canBeDeleted":true}`) → caller field preserved **and** version tag set
- caller-supplied `$formatVersion` → overwritten with the protocol version (see the nitpick thread)
- `Some(non-object)` (`[]` / `"x"` / `5`) → `Err(InvalidIdentityData(_))`
- round-trip: helper output assembled into a minimal serialization-format payload, deserialized via `DataContractInSerializationFormat`, and `DataContract::try_from_platform_versioned(.., full_validation=true, .., PlatformVersion::latest())` → `Ok`. This is the assertion that would have caught the original hardcoded-"0" v12 rejection at unit-test time.

All 193 `platform-wallet` lib tests pass.

packages/rs-platform-wallet/src/wallet/identity/network/contract.rs
QuantumExplorer: Keeping the unconditional overwrite as-is, deliberately. The helper's contract is to always emit a network-acceptable `$formatVersion`: no caller in this codebase sets the wire-level tag (abstracting it is the whole point of the helper), and honoring a caller-supplied `"0"` on v12+ would only earn a network rejection. As you noted, this is the most permissive of the three options and matches the PR's stated goal of always emitting an acceptable tag.

I've documented the overwrite explicitly in the helper's doc comment and added a regression test (`build_config_object_overwrites_caller_supplied_format_version`) that pins the behavior, so the override is no longer silent at the code/test level even though it stays silent at runtime.

QuantumExplorer (commented):
Reviewed

thepastaclaw (commented):
## Code Review

Carried-forward prior findings: prior-1 and prior-2 are both FIXED in the latest delta. New findings in the latest delta: one nitpick — the public Swift docstring for `createDataContract` still claims the library injects `$formatVersion: "0"` if missing, which now contradicts the Rust helper's documented unconditional-overwrite-with-protocol-required-version behavior introduced by this PR. The Rust fix itself looks correct, well-scoped, dependency-direction clean, and the new tests (`build_config_object_*` plus the latest-version `DataContract::try_from_platform_versioned` round-trip) would have caught the original v12 rejection bug.

💬 1 nitpick(s)

PR #3883, opened 2026-06-13
feat(rs-sdk-ffi): add masternode contested-resource vote broadcast (FFI + Swift UI)

QuantumExplorer:

There was **no broadcast FFI for casting votes** on contested resources. `packages/rs-sdk-ffi/src/voting/` and `contested_resource/` contained only *query* functions (vote-poll lookups, vote state, voters, identity votes). As a result the SwiftExampleApp could read contested-resource vote tallies (`ContestDetailView`) but could not cast a vote.

This PR adds the missing **vote write path**: a Rust FFI binding, a thin Swift wrapper, and a minimal UI affordance to cast a contested-resource (DPNS masternode) vote.

### Rust SDK (confirmed, not changed)
The Rust SDK already exposes the broadcast path: `dash_sdk::platform::transition::vote::PutVote` on `Vote`, whose `put_to_platform_and_wait_for_response(voter_pro_tx_hash, voting_public_key, sdk, signer, settings)` builds a `MasternodeVoteTransition` via `try_from_vote_with_signer` and broadcasts it. The transition derives the voter identity from `pro_tx_hash` + `masternode_voting_key.public_key_hash()` and signs via `sign_external`.

### FFI (`packages/rs-sdk-ffi`)
- New `dash_sdk_contested_resource_cast_vote` in `src/contested_resource/transitions/cast_vote.rs`, wired through the new `transitions` submodule and re-exported from `contested_resource/mod.rs`.
- It marshals the contested-document vote poll (`contract_id`, `document_type_name`, `index_name`, `index_values_json`) + a vote-choice discriminant (`0` TowardsIdentity, `1` Abstain, `2` Lock) + optional contender id, assembles a `Vote::ResourceVote`, and calls the existing rs-sdk `PutVote` entry point. Nothing is stitched together that the SDK does not already expose as a single call.
- The masternode voting key + signer are both derived from a supplied 32-byte **voting private key**: a `SingleKeySigner` signs, and the matching `ECDSA_HASH160` / `VOTING` / `HIGH` `IdentityPublicKey` (data = `hash160(pubkey)`, id 0) is constructed to mirror Platform's `get_voter_identity_key_v0`. `SingleKeySigner::can_sign_with` recomputes the same `hash160`, so key and signer agree by construction.
- Index-value parsing (hex → `Value::Bytes`, otherwise `Value::Text`) matches the existing `get_vote_state` query FFI, so a poll readable via the read path is the same poll voted on here.

**FFI signature:**
```c
struct DashSDKResult dash_sdk_contested_resource_cast_vote(
    const struct SDKHandle *sdk_handle,
    const char *contract_id,
    const char *document_type_name,
    const char *index_name,
    const char *index_values_json,
    uint8_t vote_choice,                 // 0=TowardsIdentity, 1=Abstain, 2=Lock
    const char *contender_identity_id,   // base58; required iff vote_choice==0
    const uint8_t *voter_pro_tx_hash,    // 32 bytes
    const uint8_t *voting_private_key,   // 32 bytes
    FFINetwork network);
```

### Swift SDK + UI (`packages/swift-sdk`)
- `SDK.castContestedResourceVote(...)` — a thin bridge in `PlatformQueryExtensions.swift` (marshal in → call → surface success/error out), plus a small `processVoidResult` helper.
- New `ContestedResourceVoteChoice` enum (`towardsIdentity` / `abstain` / `lock`) in `ContestVoteState.swift`, kept in lockstep with the FFI discriminants.
- `ContestDetailView` gains a "Cast a Masternode Vote" section and a sheet to pick a contender / Abstain / Lock and enter masternode credentials (pro_tx_hash + voting private key, both hex), with the masternode-only caveat spelled out in-UI.

## How Has This Been Tested?

- `cargo fmt --all` — clean.
- `cargo check -p rs-sdk-ffi --all-targets` — passes.
- `cargo clippy -p rs-sdk-ffi` — no warnings/errors on the new code.
- `cargo test -p rs-sdk-ffi cast_vote` — 3 new unit tests pass (null SDK handle; TowardsIdentity-requires-contender; invalid choice).
- `cargo build -p rs-sdk-ffi` — confirmed the cbindgen-generated `rs-sdk-ffi.h` exports `dash_sdk_contested_resource_cast_vote` with the exact signature the Swift wrapper calls.
- The iOS app was **not** built here (per instructions; the orchestrator does combined iOS testing).

**Masternode-key caveat (important).** Contested-resource votes are cast by **masternodes** using a masternode voting key tied to a `pro_tx_hash`. The SwiftExampleApp's regular wallets are **not** masternodes and have no voting key, so an end-to-end *accepted* vote is **not testable from this app** — a broadcast from a non-masternode reaches a deterministic **authorization rejection** on the platform side (voter identity / masternode not found / not authorized). The goal of this PR is correct construction + wiring so the transition is assembled, signed, and submitted properly, reaching that deterministic rejection rather than a client-side crash or a "not implemented" stub. The UI sheet accepts a `pro_tx_hash` + voting private key precisely so the path is exercisable end-to-end by someone holding real masternode voting credentials.

## Breaking Changes

None. This is a purely additive FFI + SDK + example-app surface; no consensus behavior and no existing public API is changed. (Per project convention, `feat!`/`fix!` is reserved for consensus-breaking changes; SDK/FFI additions do not qualify.)

## Checklist:

## Release Notes

* **New Features**
  * Added masternode voting capability for contested resources (e.g., DPNS names)
  * Added Swift SDK method to cast votes on contested resources with vote choice options (Towards Identity, Abstain, Lock)
  * Added example UI for casting contested resource votes with masternode credentials

### DIFF
--- a/packages/rs-sdk-ffi/src/contested_resource/mod.rs
+++ b/packages/rs-sdk-ffi/src/contested_resource/mod.rs
@@ -5,3 +5,89 @@ mod transitions;
 // Re-export all public functions
 pub use queries::*;
 pub use transitions::*;
+
+use dash_sdk::dpp::platform_value::Value;
+
+/// Decode a single JSON index-value string into a platform [`Value`] using an
+/// **explicit** type tag rather than guessing by content.
+///
+/// * A string with a `"0x"` prefix is hex-decoded into [`Value::Bytes`] (the
+///   remainder after `0x` must be valid, even-length hex).
+/// * Any other string is taken verbatim as [`Value::Text`].
+///
+/// This is the single source of truth shared by the contested-resource write
+/// path ([`crate::contested_resource::transitions::cast_vote`]) and the read
+/// path query FFIs (`vote_state`, `resources`, `voters_for_identity`) so a poll
+/// a caller can read is the same poll it votes on.
+///
+/// The previous heuristic decoded any even-length all-ASCII-hex string as bytes.
+/// That is unsafe on the signing path: a legitimate DPNS text label that happens
+/// to be even-length all-hex (e.g. `"abcd"`, `"cafe"`, `"deadbeef"`, `"1234"`)
+/// would be silently re-encoded as `Value::Bytes`, targeting a different vote
+/// poll than the operator intended. DPNS index values are text labels, so
+/// typical callers pass plain text (no `0x`) and correctly get `Value::Text`.
+pub(crate) fn parse_index_value(value_str: String) -> Result<Value, String> {
+    match value_str.strip_prefix("0x") {
+        Some(hex_str) => {
+            let bytes = hex::decode(hex_str).map_err(|e| {
+                format!(
+                    "Index value '{}' has a 0x prefix but is not valid hex: {}",
+                    value_str, e
+                )
+            })?;
+            Ok(Value::Bytes(bytes))
+        }
+        None => Ok(Value::Text(value_str)),
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::parse_index_value;
+    use dash_sdk::dpp::platform_value::Value;
+
+    #[test]
+    fn even_length_all_hex_label_without_prefix_is_text() {
+        // A DPNS-style label that is even-length all-hex must stay text on
+        // both the read and write paths — the old content heuristic would have
+        // re-encoded these as bytes.
+        for label in ["abcd", "cafe", "face", "deadbeef", "1234"] {
+            assert_eq!(
+                parse_index_value(label.to_string()).unwrap(),
+                Value::Text(label.to_string()),
+                "{label} should decode as Value::Text"
+            );
+        }
+    }
+
+    #[test]
+    fn zero_x_prefixed_value_is_bytes() {
+        assert_eq!(
+            parse_index_value("0xabcd".to_string()).unwrap(),
+            Value::Bytes(vec![0xab, 0xcd])
+        );
+        assert_eq!(
+            parse_index_value("0xdeadbeef".to_string()).unwrap(),
+            Value::Bytes(vec![0xde, 0xad, 0xbe, 0xef])
+        );
+    }
+
+    #[test]
+    fn plain_text_label_is_text() {
+        assert_eq!(
+            parse_index_value("dash".to_string()).unwrap(),
+            Value::Text("dash".to_string())
+        );
+        assert_eq!(
+            parse_index_value("alice".to_string()).unwrap(),
+            Value::Text("alice".to_string())
+        );
+    }
+
+    #[test]
+    fn zero_x_prefixed_invalid_hex_errors() {
+        assert!(parse_index_value("0xnothex".to_string()).is_err());
+        // odd-length hex after the prefix is rejected too
+        assert!(parse_index_value("0xabc".to_string()).is_err());
+    }
+}
--- a/packages/rs-sdk-ffi/src/contested_resource/mod.rs
+++ b/packages/rs-sdk-ffi/src/contested_resource/mod.rs
@@ -1,5 +1,7 @@
 // Contested resource modules
 mod queries;
+mod transitions;
 
 // Re-export all public functions
 pub use queries::*;
+pub use transitions::*;
--- a/packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
+++ b/packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
@@ -352,15 +352,22 @@ mod tests {
 
     #[test]
     fn test_cast_vote_null_handle() {
+        // Bind every CString to a local so the backing buffer outlives the FFI
+        // call; passing `CString::new(..).unwrap().as_ptr()` inline drops the
+        // temporary before the call reads the pointer (use-after-free).
+        let contract_id = CString::new(DPNS_CONTRACT_ID).unwrap();
+        let document_type_name = CString::new("domain").unwrap();
+        let index_name = CString::new("parentNameAndLabel").unwrap();
+        let index_values = CString::new(r#"["dash","alice"]"#).unwrap();
         unsafe {
             let pro_tx = [1u8; 32];
             let priv_key = [2u8; 32];
             let result = dash_sdk_contested_resource_cast_vote(
                 std::ptr::null(),
-                CString::new(DPNS_CONTRACT_ID).unwrap().as_ptr(),
-                CString::new("domain").unwrap().as_ptr(),
-                CString::new("parentNameAndLabel").unwrap().as_ptr(),
-                CString::new(r#"["dash","alice"]"#).unwrap().as_ptr(),
+                contract_id.as_ptr(),
+                document_type_name.as_ptr(),
+                index_name.as_ptr(),
+                index_values.as_ptr(),
                 ContestedResourceVoteChoiceFFI::Abstain as u8,
                 std::ptr::null(),
                 pro_tx.as_ptr(),
@@ -374,15 +381,22 @@ mod tests {
     #[test]
     fn test_cast_vote_towards_identity_requires_contender() {
         let handle = create_mock_sdk_handle();
+        // Bind every CString to a local so the backing buffer outlives the FFI
+        // call; passing `CString::new(..).unwrap().as_ptr()` inline drops the
+        // temporary before the call reads the pointer (use-after-free).
+        let contract_id = CString::new(DPNS_CONTRACT_ID).unwrap();
+        let document_type_name = CString::new("domain").unwrap();
+        let index_name = CString::new("parentNameAndLabel").unwrap();
+        let index_values = CString::new(r#"["dash","alice"]"#).unwrap();
         unsafe {
             let pro_tx = [1u8; 32];
             let priv_key = [2u8; 32];
             let result = dash_sdk_contested_resource_cast_vote(
                 handle,
-                CString::new(DPNS_CONTRACT_ID).unwrap().as_ptr(),
-                CString::new("domain").unwrap().as_ptr(),
-                CString::new("parentNameAndLabel").unwrap().as_ptr(),
-                CString::new(r#"["dash","alice"]"#).unwrap().as_ptr(),
+                contract_id.as_ptr(),
+                document_type_name.as_ptr(),
+                index_name.as_ptr(),
+                index_values.as_ptr(),
                 ContestedResourceVoteChoiceFFI::TowardsIdentity as u8,
                 std::ptr::null(), // missing contender id
                 pro_tx.as_ptr(),
@@ -399,6 +413,13 @@ mod tests {
     #[test]
     fn test_cast_vote_invalid_choice() {
         let handle = create_mock_sdk_handle();
+        // Bind every CString to a local so the backing buffer outlives the FFI
+        // call; passing `CString::new(..).unwrap().as_ptr()` inline drops the
+        // temporary before the call reads the pointer (use-after-free).
+        let contract_id = CString::new(DPNS_CONTRACT_ID).unwrap();
+        let document_type_name = CString::new("domain").unwrap();
+        let index_name = CString::new("parentNameAndLabel").unwrap();
+        let index_values = CString::new(r#"["dash","alice"]"#).unwrap();
         unsafe {
             let pro_tx = [1u8; 32];
             let priv_key = [2u8; 32];
@@ -407,10 +428,10 @@ mod tests {
             // rejected with `InvalidParameter`.
             let result = dash_sdk_contested_resource_cast_vote(
                 handle,
-                CString::new(DPNS_CONTRACT_ID).unwrap().as_ptr(),
-                CString::new("domain").unwrap().as_ptr(),
-                CString::new("parentNameAndLabel").unwrap().as_ptr(),
-                CString::new(r#"["dash","alice"]"#).unwrap().as_ptr(),
+                contract_id.as_ptr(),
+                document_type_name.as_ptr(),
+                index_name.as_ptr(),
+                index_values.as_ptr(),
                 99, // invalid discriminant
                 std::ptr::null(),
                 pro_tx.as_ptr(),
--- a/packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
+++ b/packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
@@ -93,9 +93,11 @@ pub enum ContestedResourceVoteChoiceFFI {
 ///   (DPNS for username contests).
 /// * `document_type_name` — contested document type (e.g. `"domain"`).
 /// * `index_name` — contested index name (e.g. `"parentNameAndLabel"`).
-/// * `index_values_json` — JSON array of index values. Each element is either
-///   a hex-encoded byte string (decoded to `Value::Bytes`) or a plain string
-///   (`Value::Text`), matching the parsing used by the vote-state query FFI.
+/// * `index_values_json` — JSON array of index values. Each element is decoded
+///   via an explicit type tag: a `"0x"`-prefixed string is hex-decoded to
+///   `Value::Bytes`, and any other string is taken verbatim as `Value::Text`.
+///   DPNS index values are text labels, so typical callers pass plain text
+///   (no `0x`). This matches the parsing used by the vote-state query FFI.
 /// * `vote_choice` — discriminant byte matching
 ///   [`ContestedResourceVoteChoiceFFI`]: `TowardsIdentity` (`0`), `Abstain`
 ///   (`1`) or `Lock` (`2`). Any other value is rejected with
@@ -202,24 +204,21 @@ unsafe fn cast_vote_inner(
         .map_err(|_| invalid("contract id must be exactly 32 bytes"))?;
     let contract_identifier = Identifier::new(contract_id_arr);
 
-    // Same index-value parsing as the vote-state query FFI: hex → Bytes,
-    // otherwise Text. Keeping the two FFIs in lockstep means a poll that a
-    // caller can read with `get_vote_state` is the same poll it votes on here.
+    // Same index-value parsing as the vote-state query FFI, via the shared
+    // `parse_index_value` helper: a `"0x"`-prefixed value is hex-decoded to
+    // `Value::Bytes`, anything else is taken verbatim as `Value::Text`. Using
+    // the single shared helper keeps the read and write paths in lockstep, so a
+    // poll a caller can read with `get_vote_state` is the same poll it votes on
+    // here. An explicit `0x` tag is required (rather than guessing by content)
+    // because this is a signing surface: a legit DPNS text label that happens
+    // to be even-length all-hex must not be silently re-encoded as bytes.
     let index_values_array: Vec<String> = serde_json::from_str(index_values_str)
         .map_err(|e| invalid(&format!("Failed to parse index_values JSON: {}", e)))?;
     let index_values: Vec<Value> = index_values_array
         .into_iter()
-        .map(|value_str| {
-            if value_str.chars().all(|c| c.is_ascii_hexdigit()) && value_str.len() % 2 == 0 {
-                match hex::decode(&value_str) {
-                    Ok(bytes) => Value::Bytes(bytes),
-                    Err(_) => Value::Text(value_str),
-                }
-            } else {
-                Value::Text(value_str)
-            }
-        })
-        .collect();
+        .map(crate::contested_resource::parse_index_value)
+        .collect::<Result<Vec<Value>, String>>()
+        .map_err(|e| invalid(&e))?;
 
     let vote_poll =
         VotePoll::ContestedDocumentResourceVotePoll(ContestedDocumentResourceVotePoll {
@@ -341,7 +340,7 @@ unsafe fn cstr<'a>(ptr: *const c_char, field: &str) -> Result<&'a str, FFIError>
 mod tests {
     use super::*;
     use crate::test_utils::test_utils::create_mock_sdk_handle;
-    use crate::DashSDKErrorCode;
+    use crate::{dash_sdk_error_free, DashSDKErrorCode};
     use std::ffi::CString;
 
     /// Real base58 DPNS data-contract id (32 bytes). Using the genuine id
@@ -375,6 +374,9 @@ mod tests {
                 FFINetwork::Testnet,
             );
             assert!(!result.error.is_null());
+            // Free the DashSDKError (struct + message string) the result owns;
+            // models the caller contract and keeps leak sanitizers happy.
+            dash_sdk_error_free(result.error);
         }
     }
 
@@ -406,6 +408,10 @@ mod tests {
             assert!(!result.error.is_null());
             let err = &*result.error;
             assert_eq!(err.code, DashSDKErrorCode::InvalidParameter);
+            // Free the DashSDKError (struct + message string) the result owns,
+            // after the last use of `err`; models the caller contract and keeps
+            // leak sanitizers happy.
+            dash_sdk_error_free(result.error);
             crate::test_utils::test_utils::destroy_mock_sdk_handle(handle);
         }
     }
@@ -441,6 +447,10 @@ mod tests {
             assert!(!result.error.is_null());
             let err = &*result.error;
             assert_eq!(err.code, DashSDKErrorCode::InvalidParameter);
+            // Free the DashSDKError (struct + message string) the result owns,
+            // after the last use of `err`; models the caller contract and keeps
+            // leak sanitizers happy.
+            dash_sdk_error_free(result.error);
             crate::test_utils::test_utils::destroy_mock_sdk_handle(handle);
         }
     }
--- a/packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
+++ b/packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
@@ -1,0 +1,390 @@
+//! Masternode contested-resource vote broadcast.
+//!
+//! This is the *write* counterpart to the contested-resource query FFIs in
+//! `crate::contested_resource::queries`. It assembles a
+//! [`MasternodeVoteTransition`] from a contested-document vote poll
+//! (`contract_id`, `document_type_name`, `index_name`, `index_values`) plus a
+//! [`ResourceVoteChoice`] and broadcasts it via the existing rs-sdk
+//! [`PutVote`] entry point. Nothing is stitched together here that the SDK
+//! does not already expose as a single call.
+//!
+//! # Who can actually cast a vote
+//!
+//! A contested-resource (DPNS) vote is cast by a **masternode** using its
+//! masternode *voting key* — an `ECDSA_HASH160` key whose 20-byte `data` is
+//! the hash160 of the voting public key, tied to the masternode's
+//! `pro_tx_hash` (see `get_voter_identity_key_v0` in rs-drive-abci). The
+//! voter identity is derived deterministically as
+//! `create_voter_identifier(pro_tx_hash, voting_address)`.
+//!
+//! This FFI therefore takes the raw 32-byte **voting private key** plus the
+//! 32-byte `pro_tx_hash`. From the private key it derives:
+//!   * a [`SingleKeySigner`] that signs the transition, and
+//!   * the matching `ECDSA_HASH160` [`IdentityPublicKey`] (the masternode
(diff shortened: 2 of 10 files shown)


thepastaclaw:
✅ Review complete (commit d85d9fc9)

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/ContestDetailView.swift
```diff
@@ -398,4 +466,172 @@ struct ContestDetailView: View {
...
+        guard let proTxHash = Data(hexString: proTxHashHex), proTxHash.count == 32 else {
+            voteResultMessage = "pro_tx_hash must be 32 bytes (64 hex characters)."
+            voteResultIsError = true
+            return
+        }
+        guard let votingKey = Data(hexString: votingKeyHex), votingKey.count == 32 else {
+            voteResultMessage = "Voting private key must be 32 bytes (64 hex characters)."
+            voteResultIsError = true
+            return
+        }
```
thepastaclaw: **🟡 Suggestion: Hex guards silently accept odd-length input, producing wrong masternode credentials**

`Data(hexString:)` in `PlatformWalletTypes.swift` computes `let len = hexString.count / 2` and walks the string in 2-char steps, so a 65-character `pro_tx_hash` or voting private key passes through as 32 bytes with the trailing nibble dropped — and both `proTxHash.count == 32` / `votingKey.count == 32` guards then succeed. The result is a vote signed with a key that differs from what the masternode operator typed, with no error surfaced. Since this is security-sensitive credential input newly introduced here, validate the trimmed hex length (and even-ness) before decoding rather than relying on the post-decode byte count alone.

```suggestion
let normalizedProTxHashHex = proTxHashHex.trimmingCharacters(in: .whitespacesAndNewlines)
guard normalizedProTxHashHex.count == 64,
      let proTxHash = Data(hexString: normalizedProTxHashHex),
      proTxHash.count == 32 else {
    voteResultMessage = "pro_tx_hash must be 32 bytes (64 hex characters)."
    voteResultIsError = true
    return
}
let normalizedVotingKeyHex = votingKeyHex.trimmingCharacters(in: .whitespacesAndNewlines)
guard normalizedVotingKeyHex.count == 64,
      let votingKey = Data(hexString: normalizedVotingKeyHex),
      votingKey.count == 32 else {
    voteResultMessage = "Voting private key must be 32 bytes (64 hex characters)."
    voteResultIsError = true
    return
}
```

<sub>source: ['codex']</sub>

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
```diff
@@ -0,0 +1,390 @@
...
+use std::ffi::{c_char, CStr};
+use zeroize::Zeroizing;
+
+/// Discriminant for the resource vote choice, mirroring
+/// `ResourceVoteChoice::try_from((i32, Option<Vec<u8>>))` in rs-dpp:
+/// `0` = TowardsIdentity (requires `contender_identity_id`), `1` = Abstain,
+/// `2` = Lock.
+pub const VOTE_CHOICE_TOWARDS_IDENTITY: u8 = 0;
+pub const VOTE_CHOICE_ABSTAIN: u8 = 1;
+pub const VOTE_CHOICE_LOCK: u8 = 2;
```
thepastaclaw: **🟡 Suggestion: Vote-choice discriminant is hand-mirrored across the FFI rather than emitted**

`VOTE_CHOICE_TOWARDS_IDENTITY/ABSTAIN/LOCK` are bare `pub const u8` values that cbindgen does not surface in the generated C header, and `ContestVoteState.swift:68-74` hand-mirrors them as `0/1/2` with only a comment to keep the two sides in lockstep. If a future change to this Rust enum renumbers or inserts a variant, both sides will continue to compile cleanly and silently misinterpret votes (e.g. a `Lock` becoming an `Abstain`). Expose the mapping as a `#[repr(u8)]` enum so cbindgen emits it and Swift imports the canonical values rather than redeclaring them.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

PR adds a well-scoped FFI write path that delegates to the existing rs-sdk `PutVote` flow, with correct null/length checks, a `Zeroizing` private-key buffer, and a derived `ECDSA_HASH160` voter key that matches `SingleKeySigner::can_sign_with` by construction. Two suggestions worth addressing before merge: the Swift hex validators silently truncate odd-length input, and the cross-language vote-choice discriminant is hand-mirrored rather than emitted by cbindgen. One nitpick on test coverage.

<details>

```

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/ContestDetailView.swift`:
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/ContestDetailView.swift:489-498: Hex guards silently accept odd-length input, producing wrong masternode credentials
  `Data(hexString:)` in `PlatformWalletTypes.swift` computes `let len = hexString.count / 2` and walks the string in 2-char steps, so a 65-character `pro_tx_hash` or voting private key passes through as 32 bytes with the trailing nibble dropped — and both `proTxHash.count == 32` / `votingKey.count == 32` guards then succeed. The result is a vote signed with a key that differs from what the masternode operator typed, with no error surfaced. Since this is security-sensitive credential input newly introduced here, validate the trimmed hex length (and even-ness) before decoding rather than relying on the post-decode byte count alone.

In `packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs`:
- [SUGGESTION] packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs:60-62: Vote-choice discriminant is hand-mirrored across the FFI rather than emitted
  `VOTE_CHOICE_TOWARDS_IDENTITY/ABSTAIN/LOCK` are bare `pub const u8` values that cbindgen does not surface in the generated C header, and `ContestVoteState.swift:68-74` hand-mirrors them as `0/1/2` with only a comment to keep the two sides in lockstep. If a future change to this Rust enum renumbers or inserts a variant, both sides will continue to compile cleanly and silently misinterpret votes (e.g. a `Lock` becoming an `Abstain`). Expose the mapping as a `#[repr(u8)]` enum so cbindgen emits it and Swift imports the canonical values rather than redeclaring them.
```
</details>

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
```diff
@@ -0,0 +1,390 @@
...
+                pro_tx.as_ptr(),
+                priv_key.as_ptr(),
+                FFINetwork::Testnet,
+            );
+            assert!(!result.error.is_null());
+            let err = &*result.error;
+            assert_eq!(err.code, DashSDKErrorCode::InvalidParameter);
+            crate::test_utils::test_utils::destroy_mock_sdk_handle(handle);
+        }
+    }
```
thepastaclaw, line 456: **💬 Nitpick: Named test branches are not actually exercised — contract-id length check trips first**

`test_cast_vote_towards_identity_requires_contender` and `test_cast_vote_invalid_choice` both pass `"DPNS"` as `contract_id`. `cast_vote_inner` decodes that via bs58 and rejects with `InvalidParameter("contract id must be exactly 32 bytes")` at lines 178-180 — long before the missing-contender check at line 213 or the `vote_choice` match at line 211. The asserts on `DashSDKErrorCode::InvalidParameter` still succeed, masking that the branches the test names advertise are never reached. Replace `"DPNS"` with the actual base58 DPNS contract id so the boundary code the tests claim to cover is actually executed.

<sub>source: ['claude']</sub>

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/ContestDetailView.swift
QuantumExplorer: Done in 12c8fb34f1: the cast-vote sheet now trims both hex inputs and requires length == 64 before decoding, so an odd-length (65-char) pro_tx_hash or voting key can no longer drop its trailing nibble and pass the count==32 guard.

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
QuantumExplorer: Done in 12c8fb34f1: the three bare `pub const u8` are now a `#[repr(u8)] ContestedResourceVoteChoiceFFI` enum, force-emitted into the C header via cbindgen's `[export] include` list (verified present in the generated header). The `vote_choice` FFI param stays a `u8` on purpose — a C caller can pass any byte, and materializing an out-of-range value as a `#[repr(u8)]` enum is UB — so the match compares the byte against the enum's `as u8` discriminants with a defensive `other` arm. Swift's `ffiTag` now returns the generated `ContestedResourceVoteChoiceFFI` type rather than a bare UInt8 so the two can't silently drift.

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
QuantumExplorer: Done in 12c8fb34f1: both tests now use the real base58 DPNS contract id (GWRSAVFMjXx8HpQFaNJMqBV7MBgMK4br5UESsB4S31Ec, which decodes to 32 bytes), so they pass the bs58 length check and actually reach the missing-contender and invalid-vote-choice branches instead of short-circuiting on "DPNS". Assertions still check DashSDKErrorCode::InvalidParameter; all 3 cast_vote tests pass.

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
```diff
@@ -0,0 +1,426 @@
...
+            if value_str.chars().all(|c| c.is_ascii_hexdigit()) && value_str.len() % 2 == 0 {
+                match hex::decode(&value_str) {
+                    Ok(bytes) => Value::Bytes(bytes),
+                    Err(_) => Value::Text(value_str),
+                }
+            } else {
+                Value::Text(value_str)
+            }
+        })
+        .collect();
```
thepastaclaw: **🟡 Suggestion: Hex-or-text auto-detection silently re-encodes even-length all-hex labels as bytes when signing a vote**

Each `index_values_json` element whose chars are all ASCII hex and whose length is even is decoded as `Value::Bytes`; otherwise `Value::Text`. The read-path FFI uses the same heuristic, but here the result is fed straight into a signed `MasternodeVoteTransition`. A perfectly valid DPNS-style text label that happens to be even-length all-hex — e.g. `"abcd"`, `"cafe"`, `"face"`, `"deadbeef"`, `"1234"` — is encoded as `Value::Bytes` (e.g. `[0xab,0xcd]`) instead of `Value::Text("abcd")`, so the broadcast vote targets a different (likely non-existent) poll than the operator sees in the UI. The failure mode on the read side is "empty results"; on the write side it becomes "signed-and-broadcast vote against the wrong poll, with a deterministic platform rejection" with no signal back to the caller that the value was reinterpreted. Mirroring the query FFI for symmetry is a defensible design choice, but for a signing surface it would be safer to require an explicit type tag — e.g. a `"0x..."` prefix for bytes, or accept index values as `{"text":...}` / `{"bytes_hex":...}` objects — and apply the same encoding to the query FFI so the two stay in lockstep.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/ContestDetailView.swift
```diff
@@ -398,4 +466,183 @@ struct ContestDetailView: View {
...
+            // The FFI expects a base58 identity id; convert from the
+            // hex the contender rows carry.
+            let base58 = Data(hexString: hex)?.toBase58String() ?? hex
+            return .towardsIdentity(base58)
+        case .abstain:
+            return .abstain
+        case .lock:
+            return .lock
+        }
+    }
```
thepastaclaw: **💬 Nitpick: `resolvedChoice` has a silently-wrong `?? hex` fallback if hex decoding ever fails**

`let base58 = Data(hexString: hex)?.toBase58String() ?? hex` passes the raw hex string to the FFI as if it were base58 when `Data(hexString:)` returns nil. The input here is produced by `contender.identityId.toHexString()` in the picker rows, so the fallback path is effectively dead today, but if it ever fired the Rust side would emit a confusing `Failed to decode contender id` from bs58 instead of a clear conversion failure. Either drop the fallback (force-unwrap is defensible because the input came from `Data.toHexString()`) or surface a user-facing error.

<sub>source: ['claude']</sub>

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
```diff
@@ -0,0 +1,426 @@
...
+                pro_tx.as_ptr(),
+                priv_key.as_ptr(),
+                FFINetwork::Testnet,
+            );
+            assert!(!result.error.is_null());
+            let err = &*result.error;
+            assert_eq!(err.code, DashSDKErrorCode::InvalidParameter);
+            crate::test_utils::test_utils::destroy_mock_sdk_handle(handle);
+        }
+    }
```
thepastaclaw, line 456: **💬 Nitpick: Unit tests leak the FFI error allocation they assert against**

All three tests assert `!result.error.is_null()` (two dereference `result.error`) but never call `dash_sdk_error_free(result.error)`. The mock SDK handle is freed via `destroy_mock_sdk_handle`, but the `DashSDKError` struct and its message string produced by `DashSDKResult::error` leak on every test run. Harmless under `cargo test` but it (a) doesn't model the contract callers must follow and (b) would trip a leak sanitizer. Pair each `!result.error.is_null()` assertion with a matching `dash_sdk_error_free(result.error)` after the per-field assertions.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Latest delta (12c8fb34) is a focused hardening pass that resolves all three prior PastaClaw findings: ContestDetailView now requires exact 64-char hex length before decoding (prior #1 FIXED), cast_vote.rs exports a force-included #[repr(u8)] ContestedResourceVoteChoiceFFI enum through cbindgen and matches the FFI boundary u8 against `as u8` discriminants (prior #2 FIXED), and the unit tests now use the real base58 DPNS contract id so the missing-contender / invalid-choice arms are actually reached (prior #3 FIXED). One carried-forward in-scope concern remains: the hex-vs-text heuristic on `index_values_json` (cast_vote.rs:208-222) is mirrored from the read-path FFI, but on this new write/signing path a hex-looking text label is silently signed against a different poll than the user-facing value — flagged independently by both claude-general and codex-security-auditor. Two minor nits (dead `?? hex` fallback in resolvedChoice and leaked FFI error pointers in unit tests) are kept as nitpicks. Lower-value findings around the `cstr` lifetime signature and the remaining Swift literal cast to ContestedResourceVoteChoiceFFI(0/1/2) were dropped — the C-header export is now correct and Swift's Clang importer does not import C `#define`s, so the literal-cast pattern is what's actually expressible.

<details>

```

In `packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs`:
- [SUGGESTION] packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs:208-222: Hex-or-text auto-detection silently re-encodes even-length all-hex labels as bytes when signing a vote
  Each `index_values_json` element whose chars are all ASCII hex and whose length is even is decoded as `Value::Bytes`; otherwise `Value::Text`. The read-path FFI uses the same heuristic, but here the result is fed straight into a signed `MasternodeVoteTransition`. A perfectly valid DPNS-style text label that happens to be even-length all-hex — e.g. `"abcd"`, `"cafe"`, `"face"`, `"deadbeef"`, `"1234"` — is encoded as `Value::Bytes` (e.g. `[0xab,0xcd]`) instead of `Value::Text("abcd")`, so the broadcast vote targets a different (likely non-existent) poll than the operator sees in the UI. The failure mode on the read side is "empty results"; on the write side it becomes "signed-and-broadcast vote against the wrong poll, with a deterministic platform rejection" with no signal back to the caller that the value was reinterpreted. Mirroring the query FFI for symmetry is a defensible design choice, but for a signing surface it would be safer to require an explicit type tag — e.g. a `"0x..."` prefix for bytes, or accept index values as `{"text":...}` / `{"bytes_hex":...}` objects — and apply the same encoding to the query FFI so the two stay in lockstep.
```
</details>

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
```diff
@@ -0,0 +1,426 @@
...
+                CString::new("parentNameAndLabel").unwrap().as_ptr(),
+                CString::new(r#"["dash","alice"]"#).unwrap().as_ptr(),
+                ContestedResourceVoteChoiceFFI::Abstain as u8,
+                std::ptr::null(),
+                pro_tx.as_ptr(),
+                priv_key.as_ptr(),
+                FFINetwork::Testnet,
+            );
+            assert!(!result.error.is_null());
+        }
```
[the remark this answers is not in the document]
QuantumExplorer: Fixed in da40c87cbe. Each test now binds every `CString` (contract_id, document_type_name, index_name, index_values) to a `let` local that outlives the `dash_sdk_contested_resource_cast_vote` call, then passes `.as_ptr()` of the bound variable. Previously the inline `CString::new(..).unwrap().as_ptr()` temporaries were dropped before the FFI function read the pointers, dangling the `*const c_char` args (use-after-free). Test logic and assertions are unchanged; all 3 `cast_vote` tests still pass and clippy is clean.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/ContestDetailView.swift
thepastaclaw: **💬 Nitpick: `resolvedChoice` falls back to raw hex when hex decoding fails**

`let base58 = Data(hexString: hex)?.toBase58String() ?? hex` passes the raw hex string into the FFI as if it were base58 when `Data(hexString:)` returns nil. Today the picker sources `hex` from `contender.identityId.toHexString()`, so the fallback is effectively dead — but it silently masks any future malformed input, and base58-decoding a hex string downstream will land on an unrelated identifier or a decode error far from the source. Prefer a guarded/throwing decode that surfaces the failure (e.g. disable the Submit button or show an error) rather than passing the original hex string through.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Incremental review of da40c87 over 12c8fb34. Latest delta is test-only: every CString is now bound to a local before its pointer is taken, fixing the prior use-after-free in the three unit tests. Three prior findings remain valid against unchanged production/test code and are carried forward as one suggestion and two nitpicks; no new in-scope blockers.

💬 1 nitpick(s)

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
QuantumExplorer: Fixed in d85d9fc96e. The unsafe content heuristic is gone: an index value is now decoded as Value::Bytes only when it carries an explicit "0x" prefix (hex-decode the remainder), otherwise it is taken verbatim as Value::Text. So an even-length all-hex DPNS label like "abcd"/"cafe"/"deadbeef"/"1234" stays text and the signed vote targets the intended poll. The parsing lives in one shared `parse_index_value` helper in contested_resource/mod.rs and is called from the cast-vote write path plus all three read-path query FFIs (vote_state, resources, voters_for_identity) so read and write stay in lockstep. Added unit tests: an even-length all-hex label without 0x decodes as Value::Text, and a "0x...." value decodes as Value::Bytes. Doc comments updated to describe the 0x convention.

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
QuantumExplorer: Fixed in d85d9fc96e. Each of the three cast_vote tests now calls `dash_sdk_error_free(result.error)` after the per-field assertions (for the two that bind `let err = &*result.error`, the free happens after the last use of `err`), so the DashSDKError struct and its message string no longer leak. dash_sdk_error_free is the SDK-exposed free fn for this allocation.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/ContestDetailView.swift
QuantumExplorer: Fixed in d85d9fc96e. `resolvedChoice` in ContestDetailView no longer uses the `?? hex` fallback. It now returns an optional and uses a guarded decode (`guard let base58 = Data(hexString: hex)?.toBase58String() else { return nil }`); the Submit button checks for nil and surfaces a clear error ("Could not decode the selected contender's identity id.") and returns instead of passing un-decodable hex downstream to the FFI as if it were base58.

QuantumExplorer: Fixed in d85d9fc96e. `resolvedChoice` no longer falls back to the raw hex string. It returns an optional via a guarded decode (`guard let base58 = Data(hexString: hex)?.toBase58String() else { return nil }`), and the Submit button surfaces a user-facing error and returns when decoding fails rather than passing the original hex through to the FFI as if it were base58.

packages/rs-sdk-ffi/src/contested_resource/mod.rs
```diff
@@ -1,5 +1,93 @@
...
+                format!(
+                    "Index value '{}' has a 0x prefix but is not valid hex: {}",
+                    value_str, e
+                )
+            })?;
+            Ok(Value::Bytes(bytes))
+        }
+        None => Ok(Value::Text(value_str)),
+    }
+}
```
thepastaclaw, line 42: **🟡 Suggestion: `parse_index_value` silently changes query-side index-value semantics**

The shared helper now backs not only the new `cast_vote` write path but also the existing query FFIs (`vote_state`, `resources`, `voters_for_identity`). Before this PR a bare even-length hex string such as `"deadbeef"` was decoded to `Value::Bytes`; at HEAD the same input produces `Value::Text`, with no error and no compatibility shim. Any existing C/Swift caller that intentionally passed unprefixed hex to query a `Value::Bytes` index will now silently query a different poll and get an empty result. The hardening on the signing path is correct (never auto-encode on a signing surface), but the read path's behavior change deserves either input validation that rejects bare hex on reads with an explanatory error, or an explicit CHANGELOG/migration note plus a deprecation period before flipping query semantics. As written, the only signal an external caller gets is silently-empty result sets.

<sub>source: ['codex', 'claude']</sub>

packages/rs-sdk-ffi/src/contested_resource/mod.rs
```diff
@@ -1,5 +1,93 @@
...
+            let bytes = hex::decode(hex_str).map_err(|e| {
+                format!(
+                    "Index value '{}' has a 0x prefix but is not valid hex: {}",
+                    value_str, e
+                )
+            })?;
+            Ok(Value::Bytes(bytes))
+        }
+        None => Ok(Value::Text(value_str)),
+    }
```
thepastaclaw, line 41: **🟡 Suggestion: `0x` prefix as type tag is ambiguous for valid DPNS text labels starting with `0x`**

Replacing the even-length-all-hex heuristic with an explicit `0x` prefix fixes the `"deadbeef"` case, but the prefix itself is still meaningful text: a DPNS contested label like `"0xdeadbeef"` matches DPNS's alphanumeric label rules and is signable. With the current parser, such a label routed through the Swift `castContestedResourceVote` path (which forwards the user-visible label string verbatim) is signed as `Value::Bytes([0xde, 0xad, 0xbe, 0xef])` instead of `Value::Text("0xdeadbeef")`, so the masternode broadcasts a vote against a different vote-poll encoding than the one displayed in `ContestDetailView`. The vote is rejected or misdirected rather than counted for the displayed contest. A structured tag — e.g. `{ "type": "text"|"bytes", "value": ... }` on the FFI surface, or pre-encoding the Swift caller's text labels with an unambiguous wrapper — removes the residual ambiguity on a signing surface. Treat any prefix-based magic on a signing input as a half-fix.

<sub>source: ['codex']</sub>

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
```diff
@@ -0,0 +1,457 @@
...
+    // exactly the shape Platform assigns to voter identities
+    // (`get_voter_identity_key_v0`). `MasternodeVoteTransition` calls
+    // `masternode_voting_key.public_key_hash()` to derive the voter
+    // identifier, and `SingleKeySigner::can_sign_with` recomputes the same
+    // hash160 from the private key, so the two agree by construction.
+    let secp = Secp256k1::new();
+    let secret_key = SecretKey::from_byte_array(&key_array)
+        .map_err(|e| invalid(&format!("Invalid voting private key: {}", e)))?;
+    let public_key = PublicKey::from_secret_key(&secp, &secret_key);
+    let voting_address = hash160::Hash::hash(&public_key.serialize()).to_byte_array();
```
thepastaclaw, line 296: **💬 Nitpick: Voting public key is derived from a second `SecretKey::from_byte_array` instead of reusing the signer's key**

`SingleKeySigner::new_from_slice` already parses and validates the 32 voting-key bytes as a `dashcore::PrivateKey`. `cast_vote_inner` then reparses the same bytes a second time as `secp256k1::SecretKey` purely to derive the compressed public key for the `ECDSA_HASH160` voting-key `data`. There is no behavioral need for the second parse — the signer already owns an equivalent key and could expose the compressed public key. The duplicate parse keeps a redundant copy of the raw key in scope alongside the `Zeroizing` buffer and creates two independent definitions of "valid voting private key" that can drift if validation in one path tightens later. Reuse the signer's key to keep a single source of truth and reduce raw-key residue on the signing path.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

All three prior findings are resolved at HEAD: the signing-path hex heuristic is replaced with the explicit `0x`-tagged `parse_index_value` helper, `resolvedChoice` returns nil instead of falling back to raw hex, and the three cast_vote unit tests now call `dash_sdk_error_free`. Two new in-scope concerns remain on the index-value parsing change: (1) the read-path FFIs (`get_vote_state`, `get_resources`, `get_voters_for_identity`) now silently reinterpret previously-byte queries as text, and (2) using `0x` as the byte/text discriminator is still ambiguous for valid DPNS text labels that begin with `0x`. A minor nitpick on a redundant `SecretKey` reparse is also kept.

<details>

```

In `packages/rs-sdk-ffi/src/contested_resource/mod.rs`:
- [SUGGESTION] packages/rs-sdk-ffi/src/contested_resource/mod.rs:29-42: `parse_index_value` silently changes query-side index-value semantics
  The shared helper now backs not only the new `cast_vote` write path but also the existing query FFIs (`vote_state`, `resources`, `voters_for_identity`). Before this PR a bare even-length hex string such as `"deadbeef"` was decoded to `Value::Bytes`; at HEAD the same input produces `Value::Text`, with no error and no compatibility shim. Any existing C/Swift caller that intentionally passed unprefixed hex to query a `Value::Bytes` index will now silently query a different poll and get an empty result. The hardening on the signing path is correct (never auto-encode on a signing surface), but the read path's behavior change deserves either input validation that rejects bare hex on reads with an explanatory error, or an explicit CHANGELOG/migration note plus a deprecation period before flipping query semantics. As written, the only signal an external caller gets is silently-empty result sets.
- [SUGGESTION] packages/rs-sdk-ffi/src/contested_resource/mod.rs:30-41: `0x` prefix as type tag is ambiguous for valid DPNS text labels starting with `0x`
  Replacing the even-length-all-hex heuristic with an explicit `0x` prefix fixes the `"deadbeef"` case, but the prefix itself is still meaningful text: a DPNS contested label like `"0xdeadbeef"` matches DPNS's alphanumeric label rules and is signable. With the current parser, such a label routed through the Swift `castContestedResourceVote` path (which forwards the user-visible label string verbatim) is signed as `Value::Bytes([0xde, 0xad, 0xbe, 0xef])` instead of `Value::Text("0xdeadbeef")`, so the masternode broadcasts a vote against a different vote-poll encoding than the one displayed in `ContestDetailView`. The vote is rejected or misdirected rather than counted for the displayed contest. A structured tag — e.g. `{ "type": "text"|"bytes", "value": ... }` on the FFI surface, or pre-encoding the Swift caller's text labels with an unambiguous wrapper — removes the residual ambiguity on a signing surface. Treat any prefix-based magic on a signing input as a half-fix.
```
</details>

PR #3878, opened 2026-06-13
fix(rs-sdk-ffi): stop labeling unclassified SDK errors as "failed to fetch balances"

QuantumExplorer:
## Issue being fixed or feature implemented

The catch-all arm of the `FFIError::SDKError -> DashSDKError` classifier in `packages/rs-sdk-ffi/src/error.rs` prefixed **every** unclassified error with a copy-pasted `"Failed to fetch balances: "` string and mapped it to `NetworkError`. That prefix came from a balance-fetch context but was applied as the generic fallback for all errors that didn't match the earlier `timeout` / `connection` / `DAPI` / `protocol` / `not found` heuristics.

Observed on testnet 2026-06-12: a `getDataContractHistory` query failed and surfaced to the UI as:

> `Internal Error: Failed to fetch balances: Proof verification error: dash drive: proof: corrupted error: we did not get back an element for the correct path for the historical contract`

Both the "Failed to fetch balances" text and a `NetworkError` classification are nonsense for a contract-history proof failure.

## What was done?

- Replaced the catch-all branch so it passes the original SDK error message through **unchanged** and maps to `DashSDKErrorCode::InternalError` instead of `NetworkError`. This matches the existing pass-through arms (`protocol` / `not found` / `timeout`) and correctly classifies genuinely-unknown failures.
- Added a regression test (`unclassified_error_maps_to_internal_error_without_balance_prefix`) that drives a proof-verification error through the catch-all and asserts the code is `InternalError`, the message is the SDK error's `Display` verbatim, and it contains no "Failed to fetch balances" text.

Audit of the surrounding match arms: all other arms are correctly scoped — their messages only fire behind the matching condition that makes them accurate (`"Network connection failed: …"`, `"DAPI error: …"`, the "No DAPI addresses configured" text). The buggy literal appeared in exactly one place repo-wide, and this is the only `From<…> for DashSDKError` impl in the crate.

No Swift changes are required: the Swift side maps the numeric code to a display prefix only and never branches on it, and its retry logic is message-based, so switching the catch-all code from `3` (NetworkError) to `99` (InternalError) is safe.

## How Has This Been Tested?

- `cargo test -p rs-sdk-ffi` — all 5 error-module tests pass, including the new one.
- `cargo check -p rs-sdk-ffi --all-targets --all-features --tests` — clean.
- `cargo clippy -p rs-sdk-ffi --all-targets` — no warnings.
- `cargo fmt --all` applied.

## Breaking Changes

[...]

PR #3884, opened 2026-06-13
fix(drive): return empty history for contracts that don't keep history

QuantumExplorer:

`getDataContractHistory` against a contract with `keepsHistory=false` (e.g. the testnet DPNS contract `GWRSAVFMjXx8HpQFaNJMqBV7MBgMK4br5UESsB4S31Ec`) failed with a proof-verification error instead of returning an empty result:

> `Proof verification error: dash drive: proof: corrupted error: we did not get back an element for the correct path for the historical contract`

A contract that doesn't keep history simply has no history — that should be a graceful empty result, not an error. This also matches a long-standing `it.skip` in `packages/wasm-sdk/tests/functional/contracts.spec.ts`.

## What was done?

**Root cause.** `verify_contract_history_v0` always assumes the queried contract keeps history. It expects every proved key-value at the 3-segment history path `[DataContractDocuments, contract_id, 0]`. But a contract with `keepsHistory=false` stores its single current version as an `Item` at the 2-segment contract **root** path (key `[0]`), one level up. GroveDB therefore proves *that* element, the path is length 2 (not 3), and the verifier rejected a perfectly valid proof as `CorruptedProof`.

**Fix** (`packages/rs-drive/src/verify/contract/verify_contract_history/v0/mod.rs`): detect the non-history shape — *exactly one* proved element at `contract_root_path_vec(id)` (key `[0]`) that is a present item (`is_any_item()`) — and return an empty history map. Any other shape at that path still fails as `CorruptedProof`, mirroring the strict discrimination applied to the history-path entries. Returning `Some(empty)` (not `None`) flows cleanly through `FromProof for DataContractHistory` and the FFI.

**Soundness:** no regression. The proof remains cryptographically bound to the state root (merk hash checks) and the signed block (tenderdash); a server cannot forge an `Item` where a real history `Tree` exists.

> Note: the misleading `"Failed to fetch balances:"` FFI error prefix that compounded this message was a separate issue, already fixed and merged in #3878. This PR was rebased onto current `v3.1-dev` (which already includes both #3878 and the shielded-activity feature), so it now contains only the verifier fix.

## How Has This Been Tested?

- `cargo test -p drive --lib verify_contract_history` — **5 passing**, including two new regression tests (`should_return_empty_history_for_non_history_contract` and a non-default `limit`/`start_at_ms` variant). Existing `keepsHistory=true` tests still pass.
- `cargo check -p drive --no-default-features --features verify` — clean (the verifier builds under the `verify` feature only; confirms `contract_root_path_vec`/`is_any_item` are available there).

**Manual / live testnet** (read-only, no funded identity needed; rebuild the xcframework first):
- No-history: query DPNS `GWRSAVFMjXx8HpQFaNJMqBV7MBgMK4br5UESsB4S31Ec` (limit 10) → graceful empty result, no error.
- Happy path: query `HLY575cNazmc5824FxqaEMEBuzFeE4a98GDRNKbyJqCM` (the known history-keeping testnet contract used by the app's `DiagnosticsView`) → ≥1 entry.

## Breaking Changes

None. Client-side proof-verification bug fix; no consensus or wire-format change. Behavior changes only from "error" to "empty result" for the no-history case.

## Checklist:
- [x] I have added "!" to the title and described breaking changes in the corresponding section if my code contains any (no breaking changes)

### DIFF
--- a/packages/rs-drive/src/verify/contract/verify_contract_history/v0/mod.rs
+++ b/packages/rs-drive/src/verify/contract/verify_contract_history/v0/mod.rs
@@ -1,4 +1,4 @@
-use crate::drive::contract::paths::contract_storage_path_vec;
+use crate::drive::contract::paths::{contract_root_path_vec, contract_storage_path_vec};
 use crate::drive::Drive;
 use crate::error::proof::ProofError;
 use crate::error::Error;
@@ -50,6 +50,29 @@ impl Drive {
 
         let (root_hash, mut proved_key_values) =
             GroveDb::verify_query(proof, &path_query, &platform_version.drive.grove_version)?;
+
+        // A contract that does NOT keep history stores its single current version as an
+        // Item at the contract root path (key [0]), one level above the history subtree.
+        // Querying history for such a contract proves THAT single Item instead of the
+        // (non-existent) history subtree. Treat it as a valid empty history rather than a
+        // corrupted proof. Require exactly one proved element that is a present item, so any
+        // other shape at this path still fails as corrupted (mirroring the strict path
+        // discrimination applied to the history entries below).
+        if proved_key_values.len() == 1 {
+            let (path, key, maybe_element) = &proved_key_values[0];
+            if path == &contract_root_path_vec(&contract_id) && key == &vec![0u8] {
+                let is_item = maybe_element
+                    .as_ref()
+                    .map(|element| element.is_any_item())
+                    .unwrap_or(false);
+                if !is_item {
+                    return Err(Error::Proof(ProofError::CorruptedProof(
+                        "expected a contract item at the contract root path".to_string(),
+                    )));
+                }
+                return Ok((root_hash, Some(BTreeMap::new())));
+            }
+        }
 
         let mut contracts: BTreeMap<u64, DataContract> = BTreeMap::new();
         for (path, key, maybe_element) in proved_key_values.drain(..) {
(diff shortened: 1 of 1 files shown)


thepastaclaw:
✅ Review complete (commit 203ef3a9)

packages/rs-drive/src/verify/contract/verify_contract_history/v0/mod.rs
```diff
@@ -53,6 +53,20 @@ impl Drive {
...
+            // (non-existent) history subtree. This is a valid "no history" result, not a
+            // corrupted proof, so return an empty history map.
+            if path == contract_root_path_vec(&contract_id) && key == vec![0] {
+                if maybe_element.is_none() {
+                    return Err(Error::Proof(ProofError::CorruptedProof(
+                        "expected a contract item at the contract root path".to_string(),
+                    )));
+                }
+                return Ok((root_hash, Some(BTreeMap::new())));
+            }
```
thepastaclaw: **🟡 Suggestion: Tighten the non-history branch: require a single Item element**

The new branch returns `Ok((root_hash, Some(BTreeMap::new())))` on the first proved tuple matching `(contract_root_path_vec(&contract_id), [0])` and only guards `maybe_element.is_none()`. Two robustness gaps in this newly added code:

1. The element type is not validated. The storage invariant is that a non-history contract is stored as an `Item` at the contract root path; if `verify_query` ever returned a `Tree` (or any non-Item) element at that path/key, the verifier would silently accept it as an empty history rather than raising `CorruptedProof`. This is the inverse of the existing strict path discrimination directly below.
2. The function returns from inside the `for (path, key, maybe_element) in proved_key_values.drain(..)` loop without inspecting siblings. If `verify_query` ever produced a mixed result (one matching tuple plus unrelated trailing tuples), the trailing entries would be silently dropped.

In practice GroveDB's contract-history path query produces a single tuple for this shape, so neither is a present bug — but asserting the shape (single entry + `into_item_bytes()` succeeds) before returning would make the discrimination defensible and match the strictness applied to the contract_storage_path arm.

thepastaclaw (commented):
## Code Review

Verified the contract-history verifier fix and the FFI error-message change against source. The fix is correct, well-tested, and cryptographically sound — GroveDB still binds the proved tuple to the authenticated state root. One in-scope robustness suggestion remains on the new branch (validate proof shape more strictly). No blockers.

🟡 1 suggestion(s)

<details>

```

In `packages/rs-drive/src/verify/contract/verify_contract_history/v0/mod.rs`:
- [SUGGESTION] packages/rs-drive/src/verify/contract/verify_contract_history/v0/mod.rs:55-68: Tighten the non-history branch: require a single Item element
  The new branch returns `Ok((root_hash, Some(BTreeMap::new())))` on the first proved tuple matching `(contract_root_path_vec(&contract_id), [0])` and only guards `maybe_element.is_none()`. Two robustness gaps in this newly added code:

  1. The element type is not validated. The storage invariant is that a non-history contract is stored as an `Item` at the contract root path; if `verify_query` ever returned a `Tree` (or any non-Item) element at that path/key, the verifier would silently accept it as an empty history rather than raising `CorruptedProof`. This is the inverse of the existing strict path discrimination directly below.
  2. The function returns from inside the `for (path, key, maybe_element) in proved_key_values.drain(..)` loop without inspecting siblings. If `verify_query` ever produced a mixed result (one matching tuple plus unrelated trailing tuples), the trailing entries would be silently dropped.

  In practice GroveDB's contract-history path query produces a single tuple for this shape, so neither is a present bug — but asserting the shape (single entry + `into_item_bytes()` succeeds) before returning would make the discrimination defensible and match the strictness applied to the contract_storage_path arm.
```
</details>

packages/rs-drive/src/verify/contract/verify_contract_history/v0/mod.rs
QuantumExplorer: Addressed in 203ef3a. The non-history detection now runs before the drain loop, requires `proved_key_values.len() == 1`, and validates the element with `is_any_item()` before returning empty — any other shape (absent, tree, or extra sibling tuples) at the contract root path still raises CorruptedProof, matching the strictness of the history-path arm.

QuantumExplorer:
Heads up for reviewers: this branch was originally cut from an early snapshot of the shielded-activity feature, so the PR initially bundled 16 stale activity commits that have since been merged into `v3.1-dev` independently. I've **rebased onto current `v3.1-dev`**, so the PR now contains a single commit touching one file: the `verify_contract_history` fix.

The earlier CodeRabbit/automated comments on `rs-platform-wallet*`, `PlatformWalletPersistenceHandler.swift`, `ShieldedActivityView.swift`, etc. are against that already-merged activity code and are **out of scope here** — they belong to the activity feature's own history, not this fix, so I've left them unaddressed. The one in-scope review (thepastaclaw, on the verifier branch) has been addressed and resolved. The merge conflict is gone (PR is now mergeable).

QuantumExplorer:
Reviewed

packages/rs-drive/src/verify/contract/verify_contract_history/v0/mod.rs
```diff
@@ -51,6 +51,29 @@ impl Drive {
...
+                    .map(|element| element.is_any_item())
+                    .unwrap_or(false);
+                if !is_item {
+                    return Err(Error::Proof(ProofError::CorruptedProof(
+                        "expected a contract item at the contract root path".to_string(),
+                    )));
+                }
+                return Ok((root_hash, Some(BTreeMap::new())));
+            }
+        }
```
thepastaclaw, line 75: **💬 Nitpick: Compare key as a slice to avoid the Vec allocation**

`key == &vec![0u8]` allocates a one-byte Vec on every verification just to compare against a constant. Use a slice compare instead — it avoids the allocation and reads more idiomatically. Cold path, so no measurable perf impact, but trivial to tighten. The `path == &contract_root_path_vec(&contract_id)` allocation is harder to eliminate without a new helper, so leave it.

```suggestion
        if proved_key_values.len() == 1 {
            let (path, key, maybe_element) = &proved_key_values[0];
            if path == &contract_root_path_vec(&contract_id) && key.as_slice() == [0u8] {
                let is_item = maybe_element
                    .as_ref()
                    .map(|element| element.is_any_item())
                    .unwrap_or(false);
                if !is_item {
                    return Err(Error::Proof(ProofError::CorruptedProof(
                        "expected a contract item at the contract root path".to_string(),
                    )));
                }
                return Ok((root_hash, Some(BTreeMap::new())));
            }
        }
```

thepastaclaw (commented):
## Code Review

Cumulative review: PR scope is a single file (verify_contract_history_v0). The prior PastaClaw finding about tightening the non-history branch (require a single Item element) is FIXED at lines 61-75 — current code requires exactly one proved tuple, validates the contract root path and key [0], and explicitly checks is_any_item(), returning CorruptedProof otherwise. No new in-scope blocking or suggestion-level findings in the latest delta; only a shared nitpick (both Claude and Codex) about a tiny Vec allocation for key comparison in this cold verification path.

💬 1 nitpick(s)

PR #3885, opened 2026-06-13
feat(swift-example-app): send from Core balance to a shielded recipient

QuantumExplorer:

On the SwiftExampleApp **Send Dash** screen, when the recipient was a shielded (Orchard) address the **Send From** list only offered **Shielded** and **Platform** — **Core was never selectable**, even with a spendable Core balance. The `.orchard` branch of `SendViewModel.availableSources` never appended `.core`, so the only way to fund a shielded send from Core money was a manual two-step (shield Core → self, then send shielded → recipient). That two-step was not intentional.

The underlying capability already ships: Type 18 `ShieldFromAssetLock` (`shieldedFundFromAssetLock`) takes Core L1 funds and an **arbitrary** recipient Orchard address. It was simply not wired into the unified Send flow.

## What was done?

Wired Core → shielded into the Send flow via the existing `shieldedFundFromAssetLock` path. No Rust/FFI changes — Swift example-app only.

`SendViewModel.swift`:
- Added a `coreToShielded` `SendFlow` (display "Shield from Core", `lock.shield` icon, fee metadata).
- `availableSources` now offers `.core` for an Orchard recipient, **appended last** so the private shielded → shielded path stays the auto-selected default.
- `updateFlow` maps `(.orchard, .core)` → `.coreToShielded`.
- `executeSend` dispatches the new flow to `shieldedFundFromAssetLock`: parses the recipient's 43-byte Orchard address and funds from the standard BIP44 Core account (`typeTag`/`standardTag == 0`) with the largest confirmed balance (asset locks draw UTXOs from a single account).
- `canSend` gates the flow on a minimum lock size (`minShieldFromCoreDuffs`, mirroring `ShieldedFundFromAssetLockView.minDuffs`) so a sub-fee amount can't kick off a doomed lock-build + ~30s Halo 2 proof.
- `estimateFee` maps the flow to the `.transfer` fee kind (Type 18 carves the same `compute_minimum_shielded_fee` base).

`SendTransactionView.swift`:
- Added `flowColor` for the new flow.
- Added `feeUnit(for:)` to decouple the fee-display unit from the source unit — this flow spends Core **duffs** but its pool fee is denominated in **credits**, which would otherwise render 1000× too large. Behaviour-preserving for every existing flow.

### Design notes for reviewers
- **Amount semantics:** the typed amount is the **L1 lock size in duffs**; the recipient receives `lock_value − pool_fee`. This mirrors `ShieldedFundFromAssetLockView`'s "lock size, not net amount" convention rather than grossing up the fee — Type 18's Orchard `value_balance` is baked into the Halo 2 proof at build time and can't be re-derived afterward.
- **Inline `await`, not staged progress:** matches the other six flows in `executeSend`. Tradeoff: the "Sending…" overlay can sit for a minute+ while the asset-lock pipeline runs (build lock → IS-lock/ChainLock wait → proof → submit). The dedicated `ShieldedFundFromAssetLockView` remains for the rich staged-progress experience. This path bypasses the shield coordinator, but the Rust `shield_guard` mutex still serializes shield-class ops per wallet, so concurrent attempts block rather than corrupt.
- **Single-account funding:** the screen's "Core" balance sums across accounts, but an asset lock draws from one account — so an amount under the displayed total can still fail if the balance is split. Documented inline.

## How Has This Been Tested?

- **Strict simulator build passes.** `./build_ios.sh --target sim` (which builds the example app with `-warnings-as-errors`) completes with `** BUILD SUCCEEDED **`; both `SendViewModel.swift` and `SendTransactionView.swift` compile as part of it, with zero Swift warnings/errors. (A fresh FFI build was required because the prebuilt xcframeworks on disk all predate the `platform_wallet_mnemonic_*` FFI symbols this branch's `SwiftDashSDK` sources need.)
- Self-reviewed all `SendFlow` switch sites for exhaustiveness (`displayName`/`iconName`/`estimatedFee`/`canSend`/`estimateFee`/`executeSend`/`flowColor`/`feeUnit`).
- The new code only uses already-shipping APIs in patterns identical to existing call sites (`shieldedFundFromAssetLock` / `ShieldedFundFromAssetLockRecipient` as used by `ShieldedFundFromAssetLockView`; `accountBalances(for:)`; `DashAddress.parse` `.orchard` extraction as used by the shielded → shielded flow).
- No automated UI tests were added (example-app UI wiring). End-to-end on-device exercise of the new send path (asset-lock → proof → recipient receives the note after sync) was not run as part of this change.

## Breaking Changes

None. Swift example-app changes only; no consensus, protocol, FFI, or SDK public-API changes.

## Checklist:

### DIFF
--- a/packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/ViewModels/SendViewModel.swift
+++ b/packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/ViewModels/SendViewModel.swift
@@ -5,6 +5,7 @@ import SwiftDashSDK
 /// Available send flow types based on source and destination.
 enum SendFlow: Equatable {
     case coreToCore              // Standard L1 payment
+    case coreToShielded          // Shield Core L1 funds to a shielded recipient
     case platformToPlatform      // Platform-address → platform-address transfer
     case platformToShielded      // Shield credits
     case shieldedToShielded      // Private transfer
@@ -14,6 +15,7 @@ enum SendFlow: Equatable {
     var displayName: String {
         switch self {
         case .coreToCore: return "Core Payment"
+        case .coreToShielded: return "Shield from Core"
         case .platformToPlatform: return "Platform Transfer"
         case .platformToShielded: return "Shield Credits"
         case .shieldedToShielded: return "Shielded Transfer"
@@ -25,6 +27,7 @@ enum SendFlow: Equatable {
     var iconName: String {
         switch self {
         case .coreToCore: return "arrow.right"
+        case .coreToShielded: return "lock.shield"
         case .platformToPlatform: return "arrow.right"
         case .platformToShielded: return "lock.shield"
         case .shieldedToShielded: return "arrow.left.arrow.right"
@@ -41,6 +44,7 @@ enum SendFlow: Equatable {
     var estimatedFee: UInt64 {
         switch self {
         case .coreToCore: return 500_000             // ~0.005 DASH
+        case .coreToShielded: return 200_000
         case .platformToPlatform: return 100_000_000 // ~0.001 DASH in credits
         case .platformToShielded: return 200_000
         case .shieldedToShielded: return 300_000
@@ -125,6 +129,13 @@ class SendViewModel: ObservableObject {
     /// (Core uses duffs; Platform / shielded use credits).
     var amountDuffs: UInt64? { amount }
 
+    /// Minimum L1 lock size (in duffs) for a `coreToShielded` send.
+    /// Mirrors `ShieldedFundFromAssetLockView.minDuffs` (1 mDASH): an
+    /// asset lock smaller than the Platform pool fee would build a lock
+    /// and a ~30s Halo 2 proof only to be rejected at submission, so we
+    /// gate it up front rather than burn the work.
+    static let minShieldFromCoreDuffs: UInt64 = 100_000
+
     /// Maximum UTF-8 byte length of a shielded memo (the 32-byte
     /// payload of the 36-byte `DashMemo`; the other 4 bytes are the
     /// kind tag). Mirrors `dpp::shielded::MEMO_PAYLOAD_SIZE`.
@@ -157,6 +168,11 @@ class SendViewModel: ObservableObject {
         switch flow {
         case .coreToCore:
             return (amountDuffs ?? 0) > 0
+        case .coreToShielded:
+            // Funded by an L1 asset lock denominated in duffs; gate on
+            // the lock floor so a doomed (sub-fee) amount can't kick off
+            // the lock-build + proof pipeline.
+            return (amountDuffs ?? 0) >= Self.minShieldFromCoreDuffs
         case .platformToPlatform, .platformToShielded,
              .shieldedToShielded, .shieldedToPlatform, .shieldedToCore:
             return (amountCredits ?? 0) > 0
@@ -177,6 +193,10 @@ class SendViewModel: ObservableObject {
         case .orchard:
             if shieldedBalance > 0 { sources.append(.shielded) }
             if platformBalance > 0 { sources.append(.platform) }
+            // Core L1 → shielded recipient via a Type 18
+            // ShieldFromAssetLock. Appended last so the private
+            // shielded → shielded path stays the auto-selected default.
+            if coreBalance > 0 { sources.append(.core) }
         case .platform:
             if platformBalance > 0 { sources.append(.platform) }
             if shieldedBalance > 0 { sources.append(.shielded) }
@@ -212,6 +232,8 @@ class SendViewModel: ObservableObject {
             detectedFlow = .shieldedToShielded
         case (.orchard, .platform):
             detectedFlow = .platformToShielded
+        case (.orchard, .core):
+            detectedFlow = .coreToShielded
         case (.platform, .platform):
             detectedFlow = .platformToPlatform
         case (.platform, .shielded):
@@ -237,7 +259,11 @@ class SendViewModel: ObservableObject {
     private func estimateFee(for flow: SendFlow) -> UInt64 {
         let kind: PlatformWalletManager.ShieldedFeeKind?
         switch flow {
-        case .shieldedToShielded, .platformToShielded:
+        case .shieldedToShielded, .platformToShielded, .coreToShielded:
+            // Type 18 ShieldFromAssetLock carves the same
+            // `compute_minimum_shielded_fee` base as the Shield/transfer
+            // path; its fee is denominated in credits (deducted from the
+            // locked value), so the view renders it with the credits unit.
             kind = .transfer
         case .shieldedToPlatform:
             kind = .unshield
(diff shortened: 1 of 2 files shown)


thepastaclaw:
✅ Review complete (commit aebecb7d)

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/ViewModels/SendViewModel.swift
```diff
@@ -237,7 +259,11 @@ class SendViewModel: ObservableObject {
     private func estimateFee(for flow: SendFlow) -> UInt64 {
         let kind: PlatformWalletManager.ShieldedFeeKind?
         switch flow {
-        case .shieldedToShielded, .platformToShielded:
+        case .shieldedToShielded, .platformToShielded, .coreToShielded:
+            // Type 18 ShieldFromAssetLock carves the same
+            // `compute_minimum_shielded_fee` base as the Shield/transfer
+            // path; its fee is denominated in credits (deducted from the
+            // locked value), so the view renders it with the credits unit.
             kind = .transfer
```
thepastaclaw, line 267: **🟡 Suggestion: Core-to-shielded fee estimate omits asset_lock_base_cost**

Mapping `.coreToShielded` to `ShieldedFeeKind.transfer` dispatches the FFI to `compute_minimum_shielded_fee` only. The actual Type 18 pool fee that Rust carves (rs-platform-wallet/src/wallet/shielded/fund_from_asset_lock.rs:478) and drive-abci enforces is `compute_minimum_shielded_fee(num_actions) + asset_lock_base_cost`. The inline comment claiming Type 18 'carves the same compute_minimum_shielded_fee base' is misleading — it carves base + asset_lock_base_cost. The Send screen therefore displays a fee lower than the network will deduct from the lock value, and the implied recipient net (`lock_value − pool_fee`) is off by the same amount. The fix likely requires a new `ShieldedFeeKind` variant on the FFI side (e.g. `.shieldFromAssetLock`) since the existing three kinds don't include the asset-lock base cost; until that exists, at minimum drop the misleading comment and flag the displayed fee as a base-only estimate so users aren't led to expect the recipient to receive `amount − base_fee`.

<sub>source: ['codex']</sub>

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/ViewModels/SendViewModel.swift
```diff
@@ -157,6 +168,11 @@ class SendViewModel: ObservableObject {
         switch flow {
         case .coreToCore:
             return (amountDuffs ?? 0) > 0
+        case .coreToShielded:
+            // Funded by an L1 asset lock denominated in duffs; gate on
+            // the lock floor so a doomed (sub-fee) amount can't kick off
+            // the lock-build + proof pipeline.
+            return (amountDuffs ?? 0) >= Self.minShieldFromCoreDuffs
```
thepastaclaw, line 175: **🟡 Suggestion: canSend / availableSources don't gate on the single funding account that executeSend will use**

Two related mismatches against execution reality:

1. `canSend` for `.coreToShielded` only checks `amountDuffs >= minShieldFromCoreDuffs`. But `executeSend` (lines 567–573) funds from a single account — the largest confirmed balance among BIP44 standard accounts (`typeTag == 0 && standardTag == 0`). A user whose Core funds are split across multiple BIP44 accounts can pass `canSend`, kick off the ~30s lock-build + Halo 2 proof pipeline, and only fail at the asset-lock submission step with 'No spendable Core account to fund the shield' or a backend error — exactly the late-failure scenario the `minShieldFromCoreDuffs` floor exists to prevent. The dedicated `ShieldedFundFromAssetLockView` already gates submit on `selectedCoreAccountBalanceDuffs >= amount` for this reason.

2. `availableSources` (and `coreBalanceSnapshot` in SendTransactionView, which sums every entry from `accountBalances`) makes `.core` selectable for Orchard recipients whenever any account has balance, including non-BIP44 / non-standard accounts that the execute-time filter excludes. A wallet whose Core funds live entirely in other `typeTag`/`standardTag` combinations would see the Core source available and then fail with 'No spendable Core account to fund the shield' after the user commits.

Fix: expose the chosen funding account's confirmed balance to the view model and gate `canSend` on `amountDuffs <= funding.confirmed`; mirror the same BIP44-standard predicate in `coreBalanceSnapshot`/`availableSources` so what the UI offers matches what execute time can actually spend, or at minimum do a synchronous pre-check at the top of the `.coreToShielded` case that fails fast before the async asset-lock pipeline starts.

thepastaclaw (commented):
## Code Review

Scoped, well-commented change wiring a new Core-to-shielded source through the existing Type 18 ShieldFromAssetLock API. Two real issues: the fee estimate for the new flow omits the asset_lock_base_cost component (UI under-reports the fee), and canSend/availableSources operate on the cross-account Core total while execution funds from a single BIP44 standard account — an amount within the displayed total but exceeding the chosen account's balance kicks off the ~30s lock+proof pipeline only to fail late.

🟡 2 suggestion(s)

<details>

```

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/ViewModels/SendViewModel.swift`:
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/ViewModels/SendViewModel.swift:262-267: Core-to-shielded fee estimate omits asset_lock_base_cost
  Mapping `.coreToShielded` to `ShieldedFeeKind.transfer` dispatches the FFI to `compute_minimum_shielded_fee` only. The actual Type 18 pool fee that Rust carves (rs-platform-wallet/src/wallet/shielded/fund_from_asset_lock.rs:478) and drive-abci enforces is `compute_minimum_shielded_fee(num_actions) + asset_lock_base_cost`. The inline comment claiming Type 18 'carves the same compute_minimum_shielded_fee base' is misleading — it carves base + asset_lock_base_cost. The Send screen therefore displays a fee lower than the network will deduct from the lock value, and the implied recipient net (`lock_value − pool_fee`) is off by the same amount. The fix likely requires a new `ShieldedFeeKind` variant on the FFI side (e.g. `.shieldFromAssetLock`) since the existing three kinds don't include the asset-lock base cost; until that exists, at minimum drop the misleading comment and flag the displayed fee as a base-only estimate so users aren't led to expect the recipient to receive `amount − base_fee`.
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/ViewModels/SendViewModel.swift:171-175: canSend / availableSources don't gate on the single funding account that executeSend will use
  Two related mismatches against execution reality:

  1. `canSend` for `.coreToShielded` only checks `amountDuffs >= minShieldFromCoreDuffs`. But `executeSend` (lines 567–573) funds from a single account — the largest confirmed balance among BIP44 standard accounts (`typeTag == 0 && standardTag == 0`). A user whose Core funds are split across multiple BIP44 accounts can pass `canSend`, kick off the ~30s lock-build + Halo 2 proof pipeline, and only fail at the asset-lock submission step with 'No spendable Core account to fund the shield' or a backend error — exactly the late-failure scenario the `minShieldFromCoreDuffs` floor exists to prevent. The dedicated `ShieldedFundFromAssetLockView` already gates submit on `selectedCoreAccountBalanceDuffs >= amount` for this reason.

  2. `availableSources` (and `coreBalanceSnapshot` in SendTransactionView, which sums every entry from `accountBalances`) makes `.core` selectable for Orchard recipients whenever any account has balance, including non-BIP44 / non-standard accounts that the execute-time filter excludes. A wallet whose Core funds live entirely in other `typeTag`/`standardTag` combinations would see the Core source available and then fail with 'No spendable Core account to fund the shield' after the user commits.

  Fix: expose the chosen funding account's confirmed balance to the view model and gate `canSend` on `amountDuffs <= funding.confirmed`; mirror the same BIP44-standard predicate in `coreBalanceSnapshot`/`availableSources` so what the UI offers matches what execute time can actually spend, or at minimum do a synchronous pre-check at the top of the `.coreToShielded` case that fails fast before the async asset-lock pipeline starts.
```
</details>

PR #3893, opened 2026-06-14
fix(sdk): refresh protocol version via a proven query, not unproved getStatus

QuantumExplorer:

Follow-up to #3886. In review (@lklimek), the `Sdk::refresh_protocol_version` added there was correctly flagged as **unsafe and redundant**:

- It read the network's protocol version from an **unproved** `getStatus` and fed it into the *same* `maybe_update_protocol_version` ratchet that #3809 had deliberately restricted to **proof + quorum-signature-verified** metadata (the version is bound to the Tenderdash `StateId.app_version`; there's a load-bearing "the ratchet must not consume unverified metadata" invariant at the proof call site). The unproved path bypassed that cryptographic trust boundary.
- It **duplicated** the existing ratchet — the only delta over "issue a normal proven query and let #3809's ratchet do its job" was surviving `UNIMPLEMENTED` on stale evonodes, which is a transient infra issue, not a reason to give up the crypto guarantee.

This PR keeps the genuinely useful, trust-free part of #3886 (the per-network version floor) and removes the unproved trust path — implementing the reviewer's own suggestion ("#3809 + get epoch info").

## What was done?

- **`packages/rs-sdk/src/sdk/refresh.rs`** — `refresh_protocol_version` now issues an ordinary **proven** `getEpochsInfo` query (`ExtendedEpochInfo::fetch_current`) and discards the epoch payload. The protocol version that query carries in its response metadata is ratcheted into the SDK by the same `maybe_update_protocol_version` path every other query uses, and **only after** proof + quorum-signature verification succeeds. Refresh now inherits exactly the same cryptographic trust as ordinary traffic; it adds no second, weaker source of truth — it just runs one proven query eagerly.
  - If the proven query fails (no `ContextProvider`, transport error, `UNIMPLEMENTED` on a stale evonode) the failure is **non-fatal** and we deliberately do **not** fall back to an unverified version. The stored version is left untouched and then clamped to the per-network floor, so it can never sit below the network's known minimum even when the round-trip fails.
  - Removed the `extract_network_protocol_version` helper and all `getStatus` plumbing.
- **Per-network floor unchanged.** The `min_protocol_version` floor (mainnet 11, testnet/devnet/regtest 12), clamped at `build()` and after refresh, is the deterministic, trust-free piece that actually fixed the QA bugs (V0 data-contract config, empty shielded-denomination set, pre-round-trip shielded-fee skew). It stays exactly as in #3886.
- **Tests** (`packages/rs-sdk/src/sdk.rs`) — rewritten to drive refresh through a **proven** `ExtendedEpochInfo` expectation (the mock injects `LATEST_VERSION` into the verified response metadata, exactly as a real quorum-signed response would) and to assert the floor-clamp / no-unverified-fallback behavior when the query is unavailable:
  - `test_refresh_ratchets_up_via_proven_query`
  - `test_refresh_leaves_pinned_sdk_unchanged`
  - `test_refresh_query_unavailable_clamps_to_floor`
  - `test_testnet_refresh_keeps_floor_when_query_unavailable`
- **FFI / Swift** — `dash_sdk_refresh_protocol_version` and `SDK.refreshProtocolVersion()` signatures and call sites are **unchanged**; only doc comments updated to drop the "unproved `getStatus`" framing.

## How Has This Been Tested?

Against `v3.1-dev` (`fix/sdk-refresh-protocol-version-proofed`):

- `cargo fmt --all` — clean.
- `cargo check -p dash-sdk -p rs-sdk-ffi --all-targets` — compiles (tests included).
- `cargo clippy -p dash-sdk -p rs-sdk-ffi --all-targets` — no new warnings.
- `cargo test -p dash-sdk --lib` (refresh / floor / protocol_version) — 13 passed, including the 4 above.
- `cargo test -p rs-sdk-ffi --lib refresh` — `test_refresh_protocol_version_null_handle` passes.

iOS xcframework / SwiftExampleApp not rebuilt here (Swift edits are comment-only; signatures unchanged).

## Breaking Changes

None. The public `refresh_protocol_version` / FFI / Swift signatures are unchanged; only the internal query and its trust properties change. No consensus-affecting behavior.

## Checklist:

## Summary by CodeRabbit

* **Documentation**
  * Updated protocol version refresh documentation across Rust and Swift SDKs to reflect an improved verification approach.

* **Tests**
  * Added comprehensive test coverage validating protocol version refresh behavior, including successful refresh scenarios, pinned version handling, and fallback behavior when queries are unavailable.

### DIFF
--- a/packages/rs-sdk-ffi/src/protocol_version/queries/refresh.rs
+++ b/packages/rs-sdk-ffi/src/protocol_version/queries/refresh.rs
@@ -17,6 +17,10 @@ use std::ffi::CString;
 /// flows (shielded pool shield/unshield/transfer/withdraw) reserve against the
 /// network's actual protocol version instead of the SDK's seed version.
 ///
+/// For an SDK pinned to a fixed protocol version (version updating disabled)
+/// this is a no-op: no network request is made and the pinned version is
+/// returned unchanged.
+///
 /// # Parameters
 /// * `sdk_handle` - Handle to the SDK instance.
 ///
--- a/packages/rs-sdk-ffi/src/protocol_version/queries/refresh.rs
+++ b/packages/rs-sdk-ffi/src/protocol_version/queries/refresh.rs
@@ -5,9 +5,10 @@ use std::ffi::CString;
 
 /// Refresh the SDK's protocol version from the connected network.
 ///
-/// Issues an unproved `getStatus` against the network and ratchets this SDK's
-/// auto-detected protocol version up to the network's current Drive protocol
-/// version (see [`dash_sdk::Sdk::refresh_protocol_version`]). The resulting
+/// Issues an ordinary **proven** `getEpochsInfo` query and ratchets this SDK's
+/// auto-detected protocol version up to the network's version through the same
+/// proof + quorum-signature-verified path every other query uses — no unverified
+/// fallback (see [`dash_sdk::Sdk::refresh_protocol_version`]). The resulting
 /// protocol version number propagates to every clone of this SDK — including
 /// the `Sdk` clone held by a `PlatformWalletManager` — because the version is
 /// stored in a shared `Arc<AtomicU32>`.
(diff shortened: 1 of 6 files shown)


lklimek (dismissed):
Just check if this code path works when version updating is disabled, if not - adjust docs accordingly and/or return error

PR #3900, opened 2026-06-15
refactor(sdk): per-network protocol-version floor + version_pinned unification

Claudius-Maginificent:
## Why this PR exists

- **Problem**: the SDK's protocol-version handling was brittle and inconsistent. The pin flag was modelled two different, opposite-polarity ways (`SdkBuilder::version_explicit` true = pinned, `Sdk::auto_detect_protocol_version` true = *not* pinned), and `refresh_protocol_version` lived in its own module behind a redundant post-build clamp that silently rewrote a caller's pinned version.
- **What breaks without it**: the contradictory pin flags are an easy footgun (a reader has to remember which polarity means "pinned"), and the post-build clamp meant `with_version()` did **not** actually honour the version you pinned — it bumped a sub-floor pin up to the floor behind your back. Cleaning this up is the groundwork for *refreshing the protocol version on SDK init*.
- **Blocking relationship to #3902**: that init wiring (FFI / WASM / JS) ships separately in **#3902**, which builds directly on the co-located refresh primitive and the unified `version_pinned` flag introduced here. The two are intended to land together. (Native `rs-sdk` init is wired by neither PR — `build()` stays synchronous; native callers can call `refresh_protocol_version()` themselves or rely on the lazy first-query ratchet. Tracked as follow-up.)

## What was done

- **`min_protocol_version` is per-network (restored in `02b63c5f0a`).** It is a `const fn` returning the network-specific minimum — **Mainnet → v11; Testnet / Devnet / Regtest → v12** — and is the per-network lower bound an unpinned SDK seeds from in `build()`. An earlier iteration of this PR had collapsed it to a single flat v11 baseline; that has been reverted so an unpinned SDK on a v12 network no longer seeds below the network's live minimum. The rustdoc was condensed to the floor-not-a-pin rationale.
- **The floor is a seed/lower-bound, not a runtime clamp.** `build()` deliberately does **not** apply `.max(min_protocol_version(network))` — the clamp is removed and stays removed (out of scope for the floor restoration). So a pinned-below version is preserved exactly as given (doc: *"the pinned version is used as-is; it is not clamped"*), while an unpinned build seeds from the per-network floor and ratchets upward monotonically via auto-detect.
- **Unified the pin flag as `version_pinned`.** Collapses the opposite-polarity pair (`version_explicit` + `auto_detect_protocol_version`) into one consistently-polarised flag (`true` = pinned, auto-detect off). The builder's `version` field also became `Option<&'static PlatformVersion>` (defaulting to `None` instead of a seeded constant). Internal rename — 21 references, none of the old names remain.
- **Relocated the refresh primitive.** `refresh_protocol_version` is inlined into `sdk.rs` and the dedicated `sdk/refresh.rs` module is deleted (−105 lines). It has no production caller in this PR — the init call sites live in **#3902**; three `#[cfg(test)]` unit tests exercise it here.

## Testing

- `cargo build -p dash-sdk`, `cargo fmt -p dash-sdk --check`, `cargo clippy -p dash-sdk --all-targets --features mocks` — clean (zero warnings).
- `cargo test -p dash-sdk --features mocks --lib` — 143 passed; `cargo test -p dash-sdk --features mocks --test main fetch::document_query_v0_v1` — 10 passed. The floor-targeted unit tests assert against `min_protocol_version(network)` directly (floor-agnostic, survive future floor changes); `sdk_builder_default_seeds_atomic_to_floor` boots a default (Mainnet) SDK and still expects `PROTOCOL_VERSION_11`.
- CI on the PR is the source of truth for the full workspace + JS package suites.
- The refresh-on-init mechanism was validated end-to-end on the paloma devnet on an earlier snapshot of this work — an unpinned SDK seeded at the configured floor and auto-ratcheted to the network's active version on the first proven response. That live refresh path now lives in **#3902**; see its description for the full run.

### CI reconciliation

- `wasm-sdk` `WasmSdkBuilder` builder-chaining tests (`withVersion()` / multiple-method chaining) — `greaterThan(1)` → `equal(1)`, because a pinned version is no longer clamped to the floor (depends on the removed clamp, not on the floor value).
- `tests/fetch/common.rs` and the `sdk_builder_default_seeds_atomic_to_floor` comment — doc-comment wording aligned with the per-network floor semantics (no logic change).

## Breaking changes

- **Removes the public `DEFAULT_INITIAL_PROTOCOL_VERSION` constant** (was `pub`). External code referencing it will no longer compile.
- **Promotes `SdkBuilder::with_initial_version` from `pub(crate)` to `pub`** (additive).
- Behavioural: a pinned version is no longer clamped up to the per-network floor — `with_version()` now honours the version exactly as given, including a pin below the floor. (The per-network seed values themselves are unchanged: Mainnet → v11, others → v12.)
- `with_version()` signature is unchanged; `version_pinned` and `min_protocol_version` remain internal.

## Checklist

- [x] Code builds, formats, and lints clean (`cargo build` / `cargo fmt --check` / `cargo clippy --all-targets --features mocks`).
- [x] Unit + integration tests pass locally (`dash-sdk` lib: 143; `document_query_v0_v1`: 10).
- [x] CI-reconciliation tests updated to match the kept (no-clamp) behaviour.
- [x] Public-API / behavioural breaking changes documented above.
- [ ] Intended to land together with #3902 (refresh-on-init).

## Related

- **#3902** — wires `refresh_protocol_version` into the SDK init paths (FFI / WASM / JS). Provides the production behaviour that eagerly refreshes the version on init; builds on the unified `version_pinned` flag and the co-located refresh primitive from this PR; intended to land together.

## Summary by CodeRabbit

* **New Features**
  * Added ability to manually refresh protocol version on demand with best-effort semantics
  * Made protocol version initialization method publicly available for custom configuration
  * Introduced protocol version pinning to lock SDK behavior to specific versions

* **Refactor**
  * Streamlined SDK protocol version management and default bootstrapping


packages/rs-sdk/src/sdk/refresh.rs
```diff
@@ -66,16 +66,17 @@ impl Sdk {
     /// ## Returns
     ///
     /// The SDK's protocol version number after the (possible) ratchet and the
-    /// per-network floor clamp.
+    /// [`DEFAULT_INITIAL_PROTOCOL_VERSION`] floor clamp.
     ///
     /// [`SdkBuilder::with_version()`]: super::SdkBuilder::with_version
     /// [`ContextProvider`]: crate::platform::ContextProvider
+    /// [`DEFAULT_INITIAL_PROTOCOL_VERSION`]: super::DEFAULT_INITIAL_PROTOCOL_VERSION
     pub async fn refresh_protocol_version(&self) -> Result<u32, Error> {
```
lklimek: I think this file can be deleted, and refresh_protocol_version() moved to sdk.rs and collapsed to 10 lines of code, like (pseudocode):

```
if self.auto_detect_protocol_version { 
if let Err(error) = ExtendedEpochInfo::fetch_current(self).await {
            tracing::warn!(
                target: "dash_sdk::protocol_version",
                %error,
                "proven protocol-version refresh failed; keeping current version \
                 (never falling back to an unverified one)"
            );
        }
}

Ok(self.protocol_version_number)
```

packages/rs-sdk/src/sdk.rs
```diff
@@ -84,41 +96,7 @@ pub const DEFAULT_QUORUM_PUBLIC_KEYS_CACHE_SIZE: usize = 100;
...
-/// - [`Network::Regtest`] → 12
-fn min_protocol_version(network: Network) -> u32 {
-    match network {
-        Network::Mainnet => dpp::version::v11::PROTOCOL_VERSION_11,
-        Network::Testnet => dpp::version::v12::PROTOCOL_VERSION_12,
-        Network::Devnet => dpp::version::v12::PROTOCOL_VERSION_12,
-        Network::Regtest => dpp::version::v12::PROTOCOL_VERSION_12,
-    }
-}
+pub const DEFAULT_INITIAL_PROTOCOL_VERSION: u32 = dpp::version::v11::PROTOCOL_VERSION_11;
```
lklimek: Let's restore fn min_protocol_version, make it const, and remove DEFAULT_INITIAL_PROTOCOL_VERSION. Use it in build().

packages/rs-sdk/src/sdk/refresh.rs
Claudius-Maginificent: Done in 59b17f4ccf. Deleted `packages/rs-sdk/src/sdk/refresh.rs` (and removed its `mod refresh;` declaration), and moved `refresh_protocol_version` into an `impl Sdk` block in `sdk.rs`, collapsed to exactly the shape you sketched:

```rust
pub async fn refresh_protocol_version(&self) -> Result<u32, Error> {
    if self.auto_detect_protocol_version {
        if let Err(error) = ExtendedEpochInfo::fetch_current(self).await {
            tracing::warn!(
                target: "dash_sdk::protocol_version",
                %error,
                "proven protocol-version refresh failed; keeping current version \
                 (never falling back to an unverified one)"
            );
        }
    }
    Ok(self.protocol_version_number())
}
```

The proven `ExtendedEpochInfo::fetch_current` already ratchets the version via the verified metadata path, so I also dropped the now-redundant post-refresh runtime clamp. `cargo test -p dash-sdk --features mocks` is green (291 passed / 0 failed) — the relocated refresh tests live in the `sdk.rs` test module and exercise the real method end to end.

packages/rs-sdk/src/sdk.rs
Claudius-Maginificent: Done in 59b17f4ccf. Restored `min_protocol_version` as a `const fn` and removed `DEFAULT_INITIAL_PROTOCOL_VERSION` entirely:

```rust
const fn min_protocol_version(network: Network) -> u32 {
    match network {
        Network::Mainnet => dpp::version::v11::PROTOCOL_VERSION_11,
        Network::Testnet => dpp::version::v12::PROTOCOL_VERSION_12,
        Network::Devnet => dpp::version::v12::PROTOCOL_VERSION_12,
        Network::Regtest => dpp::version::v12::PROTOCOL_VERSION_12,
    }
}
```

`const fn` + match on the `Network` enum compiles cleanly, no fallback needed. `build()` now floors at it:

```rust
let initial_protocol_version = self
    .version
    .protocol_version
    .max(min_protocol_version(self.network));
```

`min_protocol_version(network)` is the single source of the floor — no `MIN_PROTOCOL_VERSION` const introduced. Every former `DEFAULT_INITIAL_PROTOCOL_VERSION` reference (code, tests, doc-links) is repointed; a grep across `packages/` returns zero hits. Tests are back to per-network reality: testnet/devnet/regtest boot at 12, mainnet at 11.

thepastaclaw:
✅ Review complete (commit c4b795d1)

packages/rs-sdk/src/sdk.rs
```diff
@@ -45,79 +47,20 @@ use tokio::sync::{Mutex, MutexGuard};
...
-fn min_protocol_version(network: Network) -> u32 {
-    match network {
-        Network::Mainnet => dpp::version::v11::PROTOCOL_VERSION_11,
-        Network::Testnet => dpp::version::v12::PROTOCOL_VERSION_12,
-        Network::Devnet => dpp::version::v12::PROTOCOL_VERSION_12,
-        Network::Regtest => dpp::version::v12::PROTOCOL_VERSION_12,
-    }
+const fn min_protocol_version(_network: Network) -> u32 {
+    dpp::version::v10::PROTOCOL_VERSION_10 // TODO: set real per-network floors once the update mechanism is tested
 }
```
thepastaclaw: **🔴 Blocking: min_protocol_version returns v10 for every network — defeats the PR's purpose and breaks the new tests**

`min_protocol_version` discards its `Network` argument and unconditionally returns `dpp::version::v10::PROTOCOL_VERSION_10`, with a `TODO` to set real per-network floors later. This is the helper `SdkBuilder::build()` uses to seed `initial_protocol_version`, so every unpinned SDK now boots at v10 — mainnet boots below its v11 live floor, and testnet/devnet/regtest boot two versions below their v12 live floor.

Concrete consequences in this PR:

1. The three new tests added in the same module fail at runtime:
   - `test_min_protocol_version_mapping` (sdk.rs:1922) asserts Mainnet → `PROTOCOL_VERSION_11` and Testnet/Devnet/Regtest → `PROTOCOL_VERSION_12`.
   - `test_testnet_default_builder_boots_at_per_network_floor` (sdk.rs:1944) asserts the testnet SDK boots at `PROTOCOL_VERSION_12`.
   - `test_testnet_refresh_confirms_floor_via_proven_query` (sdk.rs:1967) asserts `dpp::version::LATEST_VERSION == floor`, which is false when `floor == 10`.
2. The PR description's stated regression fix is undone. The PR explains that an unpinned SDK on testnet/devnet/regtest was 'booting one protocol version low (at 11 instead of those networks' live 12)' and that fee-sensitive flows (shielded shield/unshield/transfer/withdraw) size their reserve off `Sdk::version()`. With the floor pinned at v10 for every network — and the previous runtime `fetch_max` clamp in `refresh_protocol_version` removed in this PR — the under-shoot window is strictly wider than before this PR.
3. The surrounding doc comments (sdk.rs:56–61 and elsewhere) still describe a per-network hard lower bound and claim 'testnet boots directly at 12', which no longer matches the implementation.

Either restore the per-network mapping the rest of the PR (description, tests, doc comments) was built around, or — if a single uniform floor is intentional — revise the PR title, description, tests, and re-evaluate the fee-under-reservation risk for non-mainnet networks before landing.

```suggestion
const fn min_protocol_version(network: Network) -> u32 {
    match network {
        Network::Mainnet => dpp::version::v11::PROTOCOL_VERSION_11,
        Network::Testnet | Network::Devnet | Network::Regtest => {
            dpp::version::v12::PROTOCOL_VERSION_12
        }
    }
}
```

thepastaclaw (changes requested):
## Code Review

All six agents independently flagged the same blocking issue: `min_protocol_version` at packages/rs-sdk/src/sdk.rs:62-64 ignores its `Network` argument and returns `PROTOCOL_VERSION_10` for every network. Verified against the source — this directly contradicts the new in-file tests (test_min_protocol_version_mapping, test_testnet_default_builder_boots_at_per_network_floor, test_testnet_refresh_confirms_floor_via_proven_query) and reintroduces (and widens) the fee-under-reservation window the PR description claims to close. The accompanying doc comments still describe a per-network floor that no longer exists.

🔴 1 blocking

<details>

```

- [BLOCKING] packages/rs-sdk/src/sdk.rs:62-64: min_protocol_version returns v10 for every network — defeats the PR's purpose and breaks the new tests
  `min_protocol_version` discards its `Network` argument and unconditionally returns `dpp::version::v10::PROTOCOL_VERSION_10`, with a `TODO` to set real per-network floors later. This is the helper `SdkBuilder::build()` uses to seed `initial_protocol_version`, so every unpinned SDK now boots at v10 — mainnet boots below its v11 live floor, and testnet/devnet/regtest boot two versions below their v12 live floor.

  Concrete consequences in this PR:

  1. The three new tests added in the same module fail at runtime:
     - `test_min_protocol_version_mapping` (sdk.rs:1922) asserts Mainnet → `PROTOCOL_VERSION_11` and Testnet/Devnet/Regtest → `PROTOCOL_VERSION_12`.
     - `test_testnet_default_builder_boots_at_per_network_floor` (sdk.rs:1944) asserts the testnet SDK boots at `PROTOCOL_VERSION_12`.
     - `test_testnet_refresh_confirms_floor_via_proven_query` (sdk.rs:1967) asserts `dpp::version::LATEST_VERSION == floor`, which is false when `floor == 10`.
  2. The PR description's stated regression fix is undone. The PR explains that an unpinned SDK on testnet/devnet/regtest was 'booting one protocol version low (at 11 instead of those networks' live 12)' and that fee-sensitive flows (shielded shield/unshield/transfer/withdraw) size their reserve off `Sdk::version()`. With the floor pinned at v10 for every network — and the previous runtime `fetch_max` clamp in `refresh_protocol_version` removed in this PR — the under-shoot window is strictly wider than before this PR.
  3. The surrounding doc comments (sdk.rs:56–61 and elsewhere) still describe a per-network hard lower bound and claim 'testnet boots directly at 12', which no longer matches the implementation.

  Either restore the per-network mapping the rest of the PR (description, tests, doc comments) was built around, or — if a single uniform floor is intentional — revise the PR title, description, tests, and re-evaluate the fee-under-reservation risk for non-mainnet networks before landing.
```
</details>

thepastaclaw (changes requested):
## Code Review

Carried-forward prior finding from 8b039376 is STILL VALID at head 42f75e09: `min_protocol_version` at packages/rs-sdk/src/sdk.rs:62-64 still ignores its `Network` argument and returns `PROTOCOL_VERSION_10` for every network, directly contradicting the PR's own tests (`test_min_protocol_version_mapping` asserts v11/v12) and the PR's stated goal of booting mainnet at 11 and testnet/devnet/regtest at 12. The latest-push delta (rename of `auto_detect_protocol_version` to `version_pinned` and boolean polarity inversion) is mechanically clean and introduces no new latest-delta findings. Carried-forward: 1 blocking. New: 0.


packages/rs-sdk/src/sdk.rs
lklimek: this is intended at this point, tracked with todo. We need other team members to integrate and confirm upgrade works before we change anything.

packages/rs-sdk/src/sdk.rs
```diff
@@ -45,79 +47,22 @@ use tokio::sync::{Mutex, MutexGuard};
...
-        Network::Devnet => dpp::version::v12::PROTOCOL_VERSION_12,
-        Network::Regtest => dpp::version::v12::PROTOCOL_VERSION_12,
-    }
+/// This is only the starting point: auto-detect
+/// ([`Sdk::maybe_update_protocol_version`]) ratchets the version *upward* via
+/// `fetch_max` when the network reports a newer one. It is not a hard floor — an
+/// explicitly pinned version below it is preserved as-is.
+const fn min_protocol_version(_network: Network) -> u32 {
+    dpp::version::v10::PROTOCOL_VERSION_10 // TODO: set real per-network floors once the update mechanism is tested
 }
```
thepastaclaw: **🔴 Blocking: min_protocol_version still returns v10 for every network — PR's stated goal is not implemented**

Carried-forward prior finding, STILL VALID at this head. `min_protocol_version` ignores its `_network` argument and unconditionally returns `dpp::version::v10::PROTOCOL_VERSION_10` with a TODO deferring the real per-network mapping. The PR description explicitly commits Mainnet→`PROTOCOL_VERSION_11` and Testnet/Devnet/Regtest→`PROTOCOL_VERSION_12`, and frames the whole change as closing a fee-under-reservation window for shielded flows that size reserves from `Sdk::version()` before the first proven response.

Consequences observable at this SHA:
- `SdkBuilder::build()` (sdk.rs:1081-1084) seeds every unpinned network at v10. Testnet/devnet/regtest under-shoot the live floor by two versions instead of one, which is strictly worse than the cross-network constant the PR claims to have replaced.
- `test_testnet_default_builder_boots_at_per_network_floor` (sdk.rs:1885-1897) asserts `sdk.protocol_version_number() == min_protocol_version(Network::Testnet)`. Because both sides resolve through the same function, the assertion is tautological and passes against any uniform floor; it does NOT verify the PR's stated invariant.
- The PR description claims a `test_min_protocol_version_mapping` test guards the mapping, but no such test exists in this checkout.

Until the per-network mapping is implemented and guarded by a literal-value test (e.g. `assert_eq!(min_protocol_version(Network::Testnet), PROTOCOL_VERSION_12)`), the surrounding refactor (refresh inlining, clamp removal, field rename) is cosmetic and the original regression remains.

```suggestion
const fn min_protocol_version(network: Network) -> u32 {
    match network {
        Network::Dash => dpp::version::v11::PROTOCOL_VERSION_11,
        Network::Testnet | Network::Devnet | Network::Regtest => {
            dpp::version::v12::PROTOCOL_VERSION_12
        }
        _ => dpp::version::v11::PROTOCOL_VERSION_11,
    }
}
```

packages/rs-sdk/src/sdk.rs
```diff
@@ -1104,16 +1078,10 @@ impl SdkBuilder {
...
-        // lower bound, not a pin: it applies to pinned and auto-detect SDKs alike,
-        // and auto-detect still ratchets upward from here via `fetch_max`.
-        let initial_protocol_version = self
-            .version
-            .protocol_version
-            .max(min_protocol_version(self.network));
+        let initial_version = self.version.unwrap_or_else(|| {
+            PlatformVersion::get(min_protocol_version(self.network))
+                .expect("min_protocol_version for a network must be a valid version")
+        });
```
thepastaclaw, line 1101: **🟡 Suggestion: build() does not floor an explicit with_initial_version against the per-network minimum**

The PR description states: "`build()` floors per-network. `initial_protocol_version = self.version.protocol_version.max(min_protocol_version(self.network))`." The actual code uses `self.version.unwrap_or_else(...)`, so a caller invoking `with_initial_version(pv)` where `pv.protocol_version < min_protocol_version(self.network)` is NOT lifted to the floor. `with_version` pins remain intentionally exempt (sdk.rs:949-950, guarded by the pin-below-floor test), but `with_initial_version` is documented as a seed, not a pin — its docstring says auto-detect already starts at the per-network floor, so an explicit seed below the floor contradicts the documented contract. Once the blocking finding above is fixed, this gap is what allows a caller to re-open the under-reservation window. Pair the floor mapping fix with a `max()` clamp here.

```suggestion
        let initial_version = {
            let candidate = self.version.unwrap_or_else(|| {
                PlatformVersion::get(min_protocol_version(self.network))
                    .expect("min_protocol_version for a network must be a valid version")
            });
            if self.version_pinned {
                candidate
            } else {
                let floor = min_protocol_version(self.network);
                if candidate.protocol_version < floor {
                    PlatformVersion::get(floor)
                        .expect("min_protocol_version for a network must be a valid version")
                } else {
                    candidate
                }
            }
        };
```

thepastaclaw (changes requested):
## Code Review

Carried-forward prior finding from 42f75e09 is STILL VALID at 21ea0d9e: min_protocol_version at sdk.rs:64-66 still returns PROTOCOL_VERSION_10 unconditionally with a TODO, while the PR description and title explicitly promise per-network floors (Mainnet→v11, Testnet/Devnet/Regtest→v12). The PR's central stated goal therefore is not implemented; unpinned testnet/devnet/regtest SDKs boot two versions below the live network protocol until the first proven metadata refresh, re-opening the fee-under-reservation window the PR was written to close. A secondary issue from the latest delta: build() no longer clamps caller-supplied with_initial_version against min_protocol_version, so once the per-network mapping is corrected, a sub-floor initial seed would still slip through (with_version pins remain intentionally exempt).

<details>

```

- [BLOCKING] packages/rs-sdk/src/sdk.rs:64-66: min_protocol_version still returns v10 for every network — PR's stated goal is not implemented
  Carried-forward prior finding, STILL VALID at this head. `min_protocol_version` ignores its `_network` argument and unconditionally returns `dpp::version::v10::PROTOCOL_VERSION_10` with a TODO deferring the real per-network mapping. The PR description explicitly commits Mainnet→`PROTOCOL_VERSION_11` and Testnet/Devnet/Regtest→`PROTOCOL_VERSION_12`, and frames the whole change as closing a fee-under-reservation window for shielded flows that size reserves from `Sdk::version()` before the first proven response.

  Consequences observable at this SHA:
  - `SdkBuilder::build()` (sdk.rs:1081-1084) seeds every unpinned network at v10. Testnet/devnet/regtest under-shoot the live floor by two versions instead of one, which is strictly worse than the cross-network constant the PR claims to have replaced.
  - `test_testnet_default_builder_boots_at_per_network_floor` (sdk.rs:1885-1897) asserts `sdk.protocol_version_number() == min_protocol_version(Network::Testnet)`. Because both sides resolve through the same function, the assertion is tautological and passes against any uniform floor; it does NOT verify the PR's stated invariant.
  - The PR description claims a `test_min_protocol_version_mapping` test guards the mapping, but no such test exists in this checkout.

  Until the per-network mapping is implemented and guarded by a literal-value test (e.g. `assert_eq!(min_protocol_version(Network::Testnet), PROTOCOL_VERSION_12)`), the surrounding refactor (refresh inlining, clamp removal, field rename) is cosmetic and the original regression remains.
- [SUGGESTION] packages/rs-sdk/src/sdk.rs:1081-1084: build() does not floor an explicit with_initial_version against the per-network minimum
  The PR description states: "`build()` floors per-network. `initial_protocol_version = self.version.protocol_version.max(min_protocol_version(self.network))`." The actual code uses `self.version.unwrap_or_else(...)`, so a caller invoking `with_initial_version(pv)` where `pv.protocol_version < min_protocol_version(self.network)` is NOT lifted to the floor. `with_version` pins remain intentionally exempt (sdk.rs:949-950, guarded by the pin-below-floor test), but `with_initial_version` is documented as a seed, not a pin — its docstring says auto-detect already starts at the per-network floor, so an explicit seed below the floor contradicts the documented contract. Once the blocking finding above is fixed, this gap is what allows a caller to re-open the under-reservation window. Pair the floor mapping fix with a `max()` clamp here.
```
</details>

packages/rs-sdk/src/sdk.rs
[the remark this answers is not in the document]
lklimek: this is intentional, documented with TODO.

packages/rs-sdk/src/sdk.rs
lklimek: This is intentional, documented with TODO.

packages/rs-sdk/src/sdk.rs
lklimek: PR description updated. When user provides his own floor, we assume he knows what he does and it is intentional.

thepastaclaw (changes requested):
## Code Review

Carried-forward verification: prior finding #1 (flat v10 floor) and #2 (`with_initial_version` not floored) are both classified per the current PR contract — #1 is INTENTIONALLY DEFERRED (TODO at sdk.rs:65 explicitly tracks restoration; docs frame v10 as a seed for auto-detect) and #2 is OUTDATED (with the lowest known PV as the floor and `with_initial_version` documented as a deliberate auto-detect seed, no clamp is required). One new cumulative finding remains: the PR description promises eager `refresh_protocol_version` on every real init path (native, FFI `dash_sdk_create*`, WASM constructor), but the code at f7d9154a does not wire it into any of those paths — only the explicit caller-driven `dash_sdk_refresh_protocol_version` FFI symbol and Swift wrapper invoke it. Because this PR simultaneously lowers the mainnet floor from v11 to v10, the gap the PR claims to close is in fact widened for shielded fee-sensitive flows that compute reserves from `Sdk::version()` before the first proven response ratchets the atomic upward. The latest-push delta itself is test-only and reconciles assertions to the new flat floor and the no-clamp behavior; those test changes are correct.

🔴 1 blocking

<details>

```

- [BLOCKING] packages/rs-sdk/src/sdk.rs:1075-1199: Real SDK init paths return before the proven refresh runs, contradicting the PR's stated safety story
  The PR description justifies collapsing `min_protocol_version` to a flat `PROTOCOL_VERSION_10` (which is below mainnet's live `PROTOCOL_VERSION_12`) by stating that `refresh_protocol_version()` is invoked eagerly on every real init path — native, the FFI `dash_sdk_create*` entry points (via the FFI runtime), and the WASM constructor — so the version is correct before the first real query. That wiring is not present at this head:

  - `SdkBuilder::build()` (sdk.rs:1075-1199) is synchronous; it seeds the atomic with `min_protocol_version` (v10) and returns without invoking refresh.
  - `dash_sdk_create`, `dash_sdk_create_extended`, `dash_sdk_create_trusted`, and `dash_sdk_create_with_callbacks` in `packages/rs-sdk-ffi/src/sdk.rs` all return immediately after `builder.build()` without `runtime.block_on(wrapper.sdk.refresh_protocol_version())`. Only the standalone, caller-driven `dash_sdk_refresh_protocol_version` symbol calls it.
  - `WasmSdkBuilder::build()` (packages/wasm-sdk/src/sdk.rs:315) just calls `self.inner.build()` and returns the wrapper.

  Auto-detect via `verify_response_metadata` → `maybe_update_protocol_version` still ratchets the version after the first proven response, but it cannot help any code path that reads `Sdk::version()` before that first response — which is exactly the pre-query window the PR cites for shielded shield/unshield/transfer/withdraw fee reserves. Before this PR, mainnet seeded at v11, so the gap to a v12 network was one version; with v10 as the new floor and no eager refresh wired, the gap is two versions on first use.

  Either wire `refresh_protocol_version` into the real init paths as the description promises (e.g. an async `build_and_refresh` on `SdkBuilder` plus `runtime.block_on(refresh)` after `builder.build()` in each `dash_sdk_create*` and an `await` in `WasmSdkBuilder::build`), or restore a higher per-network floor (the prior behavior) and amend the description to make clear that eager refresh on init is the caller's responsibility on the FFI/WASM paths. Shipping a v10 floor with the eager-refresh story un-wired is the worst combination for the fee-reserve regression the PR is meant to address.
```
</details>

packages/rs-sdk/tests/fetch/document_query_v0_v1.rs
```diff
@@ -220,13 +219,14 @@ fn encoder_dispatches_v0_via_query_settings_without_sdk() {
...
+    // Auto-detect default uses mainnet, so the atomic seeds to the mainnet
+    // `min_protocol_version` floor, which `version()` returns until the first
+    // response ratchets it upward.  We assert against the actual floor constant
+    // rather than a hardcoded integer so the test survives future floor bumps.
     let sdk_default = SdkBuilder::new_mock().build().expect("mock sdk");
     assert_eq!(
         sdk_default.version().protocol_version,
-        DEFAULT_INITIAL_PROTOCOL_VERSION
+        dpp::version::v10::PROTOCOL_VERSION_10
     );
```
thepastaclaw, line 230: **🔴 Blocking: Floor bumped to PV11 but integration test still asserts PV10 — CI will fail**

Commit 077120681 raised `min_protocol_version()` in `packages/rs-sdk/src/sdk.rs:65` from `PROTOCOL_VERSION_10` to `PROTOCOL_VERSION_11`. `SdkBuilder::new_mock()` defaults to `Network::Mainnet` and `build()` (sdk.rs:1081-1084, 1178) seeds the atomic from `min_protocol_version(self.network)`, so `sdk_default.version().protocol_version` is now 11. But this test still hard-codes the assertion to `dpp::version::v10::PROTOCOL_VERSION_10`, so the assertion will fail every run under `cargo test -p dash-sdk --features mocks`.

The in-crate refresh test `test_refresh_ratchets_up_via_proven_query` was updated in this same delta to use `super::min_protocol_version(Network::Mainnet)` (floor-agnostic), but the equivalent external test here was missed. The comment immediately above the assertion (lines 222-225) explicitly says "We assert against the actual floor constant rather than a hardcoded integer so the test survives future floor bumps" — yet the assertion below it does the opposite by binding to the `PROTOCOL_VERSION_10` symbol.

Note that the PR description still describes the floor as `PROTOCOL_VERSION_10` and lists this exact test as already reconciled to `== 10`; that text is stale relative to the head being reviewed. The PR author should either (a) revert the floor bump in sdk.rs:65 back to PV10 and keep the test as-is to match the stated contract, or (b) keep the bump, update the test to PV11, and reconcile the PR description. Because `min_protocol_version` is private to `sdk.rs`, the cleanest in-scope fix if keeping PV11 is to assert against `dpp::version::v11::PROTOCOL_VERSION_11`. A more durable fix (worth a follow-up) would expose the floor via a `pub` accessor so `tests/` files can track future floor bumps automatically.

```suggestion
    let sdk_default = SdkBuilder::new_mock().build().expect("mock sdk");
    assert_eq!(
        sdk_default.version().protocol_version,
        dpp::version::v11::PROTOCOL_VERSION_11
    );
}
```

thepastaclaw (changes requested):
## Code Review

Latest push (077120681 'chore: initial version 11') bumps min_protocol_version from PV10 to PV11 in sdk.rs:65, but the integration test sdk_builder_default_seeds_atomic_to_floor in tests/fetch/document_query_v0_v1.rs:229 still asserts PROTOCOL_VERSION_10 — CI will fail deterministically. The PR description also still describes the floor as PV10, so code, test, and description are three-way out of sync. Prior finding about real init paths not running refresh_protocol_version is INTENTIONALLY DEFERRED per the updated PR body, which scopes refresh-on-init wiring to #3902 and keeps native rs-sdk caller-driven.

🔴 1 blocking

<details>

```

In `packages/rs-sdk/tests/fetch/document_query_v0_v1.rs`:
- [BLOCKING] packages/rs-sdk/tests/fetch/document_query_v0_v1.rs:226-230: Floor bumped to PV11 but integration test still asserts PV10 — CI will fail
  Commit 077120681 raised `min_protocol_version()` in `packages/rs-sdk/src/sdk.rs:65` from `PROTOCOL_VERSION_10` to `PROTOCOL_VERSION_11`. `SdkBuilder::new_mock()` defaults to `Network::Mainnet` and `build()` (sdk.rs:1081-1084, 1178) seeds the atomic from `min_protocol_version(self.network)`, so `sdk_default.version().protocol_version` is now 11. But this test still hard-codes the assertion to `dpp::version::v10::PROTOCOL_VERSION_10`, so the assertion will fail every run under `cargo test -p dash-sdk --features mocks`.

  The in-crate refresh test `test_refresh_ratchets_up_via_proven_query` was updated in this same delta to use `super::min_protocol_version(Network::Mainnet)` (floor-agnostic), but the equivalent external test here was missed. The comment immediately above the assertion (lines 222-225) explicitly says "We assert against the actual floor constant rather than a hardcoded integer so the test survives future floor bumps" — yet the assertion below it does the opposite by binding to the `PROTOCOL_VERSION_10` symbol.

  Note that the PR description still describes the floor as `PROTOCOL_VERSION_10` and lists this exact test as already reconciled to `== 10`; that text is stale relative to the head being reviewed. The PR author should either (a) revert the floor bump in sdk.rs:65 back to PV10 and keep the test as-is to match the stated contract, or (b) keep the bump, update the test to PV11, and reconcile the PR description. Because `min_protocol_version` is private to `sdk.rs`, the cleanest in-scope fix if keeping PV11 is to assert against `dpp::version::v11::PROTOCOL_VERSION_11`. A more durable fix (worth a follow-up) would expose the floor via a `pub` accessor so `tests/` files can track future floor bumps automatically.
```
</details>

thepastaclaw (commented):
## Code Review

Prior PV10/PV11 test-assertion finding is FIXED at head 1976b6be. No new blocking issues in the 07712068..1976b6be delta. Remaining items are documentation staleness — several doc-comments and one test name still describe the protocol-version floor as 'per-network' even though min_protocol_version now returns a flat PV11 baseline — plus a minor API hygiene nit that refresh_protocol_version returns a Result that is always Ok.

💬 2 nitpick(s)

packages/rs-sdk/src/sdk.rs
```diff
@@ -45,79 +47,22 @@ use tokio::sync::{Mutex, MutexGuard};
...
-        Network::Devnet => dpp::version::v12::PROTOCOL_VERSION_12,
-        Network::Regtest => dpp::version::v12::PROTOCOL_VERSION_12,
-    }
+/// This is only the starting point: auto-detect
+/// ([`Sdk::maybe_update_protocol_version`]) ratchets the version *upward* via
+/// `fetch_max` when the network reports a newer one. It is not a hard floor — an
+/// explicitly pinned version below it is preserved as-is.
+const fn min_protocol_version(_network: Network) -> u32 {
+    dpp::version::v11::PROTOCOL_VERSION_11 // TODO: set real per-network floors once the update mechanism is tested
 }
```
thepastaclaw: **💬 Nitpick: Doc comments still describe the floor as 'per-network' after collapse to a flat baseline**

`min_protocol_version(_network)` now returns `PROTOCOL_VERSION_11` for every network (the `_network` parameter is unused, with a TODO tracking restoration of real per-network floors). Multiple doc strings and one test name still describe this floor as 'per-network', which will mislead readers into expecting mainnet/testnet/devnet/regtest to seed at different versions. Stale spots: the `Sdk::version` doc that says the SDK 'starts at the per-network min_protocol_version', the `parse_proof_with_metadata_and_proof` bootstrap doc, the `with_version` / `with_initial_version` docs, and the test name `test_testnet_default_builder_boots_at_per_network_floor` whose body now asserts both networks agree on a single floor. Either drop 'per-network' from these comments or land the per-network restoration before merge. Not blocking.

<sub>source: ['claude']</sub>

packages/rs-sdk/src/sdk.rs
```diff
@@ -405,6 +351,38 @@ impl Sdk {
...
+                tracing::warn!(
+                    target: "dash_sdk::protocol_version",
+                    %error,
+                    "proven protocol-version refresh failed; keeping current version \
+                     (never falling back to an unverified one)"
+                );
+            }
+        }
+        Ok(self.protocol_version_number())
+    }
```
thepastaclaw, line 395: **💬 Nitpick: `refresh_protocol_version` returns Result<u32, Error> but never returns Err**

`refresh_protocol_version` swallows fetch errors via `tracing::warn!` and unconditionally returns `Ok(self.protocol_version_number())`. The PR description characterizes refresh as 'non-fatal' and the inline doc states 'the stored version is left untouched — we never fall back to an unverified one,' which together mean the `Result` wrapper conveys no real failure state to callers and forces every consumer to handle a `Result` that cannot fail. Either return `u32` directly, or document the future failure mode that justifies keeping the Result so the signature isn't gratuitously wrapping an always-Ok value. Minor API hygiene.

<sub>source: ['claude']</sub>

packages/rs-sdk/tests/fetch/document_query_v0_v1.rs
thepastaclaw: Resolved in `1976b6b` — _Floor bumped to PV11 but integration test still asserts PV10 — CI will fail_ no longer present.

thepastaclaw (commented):
## Code Review

Cumulative review of the rs-sdk protocol-version floor/pinning/refresh refactor at 02b63c5f. The per-network floor is correctly restored (Mainnet→PV11, Testnet/Devnet/Regtest→PV12), `version_pinned` polarity is consistent, the proven-only refresh path is sound, and tests cover floor seeding, sub-floor pin preservation, upward ratchet, and non-fatal refresh failure. Remaining items are documentation/API-shape nitpicks around the newly public surface and the always-`Ok` refresh signature.

💬 4 nitpick(s)

packages/rs-sdk/src/sdk.rs
```diff
@@ -45,73 +47,18 @@ use tokio::sync::{Mutex, MutexGuard};
...
-/// ## Mapping
-///
-/// - [`Network::Mainnet`] → 11
-/// - [`Network::Testnet`] → 12
-/// - [`Network::Devnet`] → 12
-/// - [`Network::Regtest`] → 12
-fn min_protocol_version(network: Network) -> u32 {
+/// A floor, not a pin: auto-detect ([`Sdk::maybe_update_protocol_version`]) still
+/// ratchets the version *upward* via `fetch_max` when the network reports a newer
+/// one.
```
thepastaclaw: **💬 Nitpick: min_protocol_version doc overstates the floor as a runtime lower bound**

The rustdoc says an unpinned SDK "seeds from and never drops below" this value. After this PR the construction-time clamp is gone and there is no refresh-time clamp: `with_initial_version` (now public) can seed an unpinned SDK below the floor, and `with_version` can pin below the floor (see `test_explicit_pin_below_floor_is_preserved`). The accurate framing is that this is the *default seed* for unpinned SDKs without `with_initial_version`, and the stored version is monotonic-upward via `fetch_max` but not bounded below at runtime. Align the doc with the new no-clamp semantics so readers don't infer a guarantee the code no longer enforces.

<sub>source: ['claude']</sub>

packages/rs-sdk/src/sdk.rs
```diff
@@ -967,35 +948,30 @@ impl SdkBuilder {
...
     /// auto-detect that an earlier `with_version` disabled.
-    #[cfg(test)]
-    pub(crate) fn with_initial_version(mut self, version: &'static PlatformVersion) -> Self {
-        self.version = version;
-        self.version_explicit = false;
+    pub fn with_initial_version(mut self, version: &'static PlatformVersion) -> Self {
+        self.version = Some(version);
+        self.version_pinned = false;
         self
     }
```
thepastaclaw: **💬 Nitpick: Public with_initial_version doc still reads as an internal/test helper**

`with_initial_version` was previously `pub(crate)` + `#[cfg(test)]` with a doc framing it as a test-only seed for exercising the upward ratchet. It is now `pub`, but the rustdoc still reads as an internal helper and does not warn callers about the sub-floor seed hazard. Because the construction-time floor clamp was deliberately removed, a public caller passing a version below `min_protocol_version` will operate below the network's live version until a proven response arrives. Restate the intended use (testing / older-network catch-up), call out the caller-owned hazard, and point production callers at `refresh_protocol_version()` for eager correction.

<sub>source: ['claude']</sub>

packages/rs-sdk/src/sdk.rs
```diff
@@ -553,13 +533,10 @@ impl Sdk {
...
-    /// protocol version (`min_protocol_version`), so it is never below the network's
-    /// known live version. With auto-detection (default) the SDK starts at
-    /// `max(DEFAULT_INITIAL_PROTOCOL_VERSION, network floor)` and then tracks the
-    /// network's version — auto-detection only ever ratchets *upward* (`fetch_max`).
-    /// A version pinned via [`SdkBuilder::with_version()`] is returned as pinned,
-    /// except that a pin below the network floor is raised to the floor at build time.
+    /// With auto-detection (default) the SDK starts at the per-network
+    /// [`min_protocol_version`] and then tracks the network's version —
+    /// auto-detection only ever ratchets *upward* (`fetch_max`). A version pinned
+    /// via [`SdkBuilder::with_version()`] is returned as pinned.
```
thepastaclaw, line 549: **💬 Nitpick: Sdk::version() doc omits with_initial_version as a seed override**

The doc on `version()` says "With auto-detection (default) the SDK starts at the per-network `min_protocol_version`". Now that `with_initial_version` is public, an unpinned SDK's starting point can be a caller-supplied value rather than the floor. A one-line mention of `with_initial_version` here keeps `version()`'s contract self-consistent with the newly public surface introduced in this PR.

<sub>source: ['claude']</sub>

packages/rs-sdk/src/sdk.rs
```diff
@@ -1897,41 +1855,12 @@ mod test {
         );
     }
 
+    /// A pin *below* the per-network [`min_protocol_version`] is preserved
     #[test]
-    fn test_explicit_pin_overrides_default_floor() {
+    fn test_explicit_pin_below_floor_is_preserved() {
```
thepastaclaw: **💬 Nitpick: Truncated rustdoc on `test_explicit_pin_below_floor_is_preserved`**

The single-line `///` above the test reads "A pin *below* the per-network [`min_protocol_version`] is preserved" with no terminating period and no explanation of the why-it-matters context that the replaced test's block doc carried. Either finish the sentence or expand it to note that the old behavior clamped this up to the floor and the new behavior intentionally preserves the sub-floor pin, mirroring the no-clamp semantics added by this PR.

```suggestion
    /// A pin *below* the per-network [`min_protocol_version`] is preserved as-is
    /// (no construction-time clamp) and `version_pinned` stays `true`.
```

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Cumulative reconciliation against the prior reviewed commit 02b63c5f. The PR's behavioral core — `version_pinned` unification, per-network seed via `min_protocol_version`, no construction-time clamp, and the inlined `refresh_protocol_version` ratchet — is sound and well-tested. All five prior nitpicks remain unaddressed in the new HEAD (23832dc6); they are carried forward verbatim per verifier policy. One additional nit (truncated test rustdoc) is added. No blocking issues; review action is COMMENT.

💬 2 nitpick(s)

packages/rs-sdk/src/sdk.rs
thepastaclaw: Resolved in this update — _Sdk::version() doc omits with_initial_version as a seed override_ no longer present.

packages/rs-sdk/src/sdk.rs
thepastaclaw: Resolved in this update — _min_protocol_version doc overstates the floor as a runtime lower bound_ no longer present.

packages/rs-sdk/src/sdk.rs
thepastaclaw: Resolved in this update — _Public with_initial_version doc still reads as an internal/test helper_ no longer present.

packages/rs-sdk/src/sdk.rs
thepastaclaw: **💬 Nitpick: `min_protocol_version` rustdoc still claims a runtime lower-bound that the code no longer enforces**

The rustdoc reads "Per-network lower bound an unpinned SDK seeds from and never drops below." After this PR removes the construction-time clamp and promotes `with_initial_version` to `pub`, that invariant is not enforced: an unpinned SDK can be seeded below this value via `with_initial_version`, and `with_version` can pin below it as well. The helper is now the default seed for unconfigured unpinned SDKs, not a runtime lower-bound guarantee — auto-detect is monotonic upward via `fetch_max`, but it is not bounded below by this constant once a caller overrides the seed.

```suggestion
/// Default per-network seed used when an unpinned SDK has no explicit initial version.
///
/// A default seed, not a runtime clamp: [`SdkBuilder::with_initial_version`] can
/// override this seed (including below this value), and auto-detect
/// ([`Sdk::maybe_update_protocol_version`]) only ratchets the stored version
/// *upward* via `fetch_max` when the network reports a newer one.
```

packages/rs-sdk/src/sdk.rs
thepastaclaw: **💬 Nitpick: Public `with_initial_version` rustdoc lacks a sub-floor seed warning**

`with_initial_version` was promoted from `pub(crate)` to `pub` in this PR, but its rustdoc still frames it primarily as an internal seeding override. Because construction no longer clamps, a caller-supplied seed below the per-network `min_protocol_version` is used as-is until a proven response ratchets the SDK upward. For a newly-public method whose behavior deviates from naïve expectations of the word "floor", that caller-owned hazard should be explicit in the docstring, ideally with a pointer to `refresh_protocol_version` for callers who need eager network-version discovery.

```suggestion
    /// Configure the initial protocol version seed while leaving auto-detection enabled.
    ///
    /// By default, unpinned SDKs start at the per-network [`min_protocol_version`].
    /// This overrides that seed and uses the provided version as-is — including
    /// versions below the network floor — until a proven response ratchets the SDK
    /// upward via `fetch_max`. Callers using this on live networks should pair it
    /// with [`Sdk::refresh_protocol_version`] when eager network-version discovery
    /// is required.
    ///
    /// Seeds `self.version` and keeps `version_pinned` `false`, so auto-detect stays
    /// on. Builder chains are last-write-wins: a later `with_initial_version` re-enables
    /// auto-detect that an earlier `with_version` disabled.
    pub fn with_initial_version(mut self, version: &'static PlatformVersion) -> Self {
```

thepastaclaw (commented):
## Code Review

Cumulative reconciliation against prior reviewed commit 23832dc6. The 23832dc6..2dc43e17 delta is dominated by a v3.1-dev merge bringing in unrelated work (swift-sdk, rs-sdk-ffi signer preflight, KeyDisableGate) that does not touch the PR-owned protocol-version surface. All six prior nitpicks were re-verified against the current HEAD and remain present verbatim; none are addressed. The behavioral core (version_pinned unification, per-network seed via min_protocol_version, removal of the construction-time clamp, inlined best-effort refresh) remains sound and well-tested. No blocking issues — review action COMMENT.

💬 3 nitpick(s)

packages/rs-sdk/src/sdk.rs
```diff
@@ -405,6 +353,38 @@ impl Sdk {
...
+        if !self.version_pinned {
+            if let Err(error) = ExtendedEpochInfo::fetch_current(self).await {
+                tracing::warn!(
+                    target: "dash_sdk::protocol_version",
+                    %error,
+                    "proven protocol-version refresh failed; keeping current version \
+                     (never falling back to an unverified one)"
+                );
+            }
+        }
```
thepastaclaw, line 393: **🟡 Suggestion: refresh_protocol_version may panic when SDK is built with proofs disabled**

codex-rust-quality flagged that `ExtendedEpochInfo::fetch_current(self)` flows into `LimitQuery<EpochQuery>::query`, which on proofs-disabled SDKs hits `unimplemented!("queries without proofs are not supported yet")`. The refresh path documents itself as best-effort and non-fatal, so a panic on a supported builder configuration would violate that contract. Verify the proofs=false path: if it does panic, gate it behind an early return that either no-ops (preserving the stored version) or returns a typed error via the existing Result channel.

<sub>source: ['codex']</sub>

packages/rs-sdk/src/sdk.rs
thepastaclaw, line 549: **💬 Nitpick: [Carried-forward STILL VALID] Sdk::version() rustdoc omits with_initial_version as a seed-override path**

Verified at current head. `version()` is the canonical observer for the SDK protocol version, but its rustdoc still enumerates only two paths (auto-detect default vs `with_version` pin), not the now-public third path (`with_initial_version`) that overrides the seed without pinning. The observer doc should match the three constructions a user can now produce with this PR's public surface.

```suggestion
    /// Return [Dash Platform version](PlatformVersion) information used by this SDK.
    ///
    /// With auto-detection (default) the SDK starts at the per-network
    /// [`min_protocol_version`] — unless [`SdkBuilder::with_initial_version()`]
    /// overrides the seed — and then tracks the network's version; auto-detection
    /// only ever ratchets *upward* (`fetch_max`). A version pinned via
    /// [`SdkBuilder::with_version()`] is returned as pinned.
    pub fn version<'v>(&self) -> &'v PlatformVersion {
```

thepastaclaw (commented):
## Code Review

Cumulative verifier review of PR #3900 at 5b0f41ddd3. Latest delta (single rustdoc fix on test_explicit_pin_below_floor_is_preserved) resolves prior finding F6. Four prior nitpicks on rustdoc accuracy after with_initial_version became public and the construction-time clamp was removed remain STILL VALID and are carried forward. One new codex-rust-quality finding (refresh_protocol_version may panic under proofs=false via ExtendedEpochInfo::fetch_current) is preserved as a suggestion. F5 (collapse three non-mainnet match arms) is dropped as out-of-scope style polish. No blocking issues; PR core (version_pinned unification, removed post-build clamp, per-network floor restoration) is coherent and well-tested.

<details>

```

- [SUGGESTION] packages/rs-sdk/src/sdk.rs:375-384: refresh_protocol_version may panic when SDK is built with proofs disabled
  codex-rust-quality flagged that `ExtendedEpochInfo::fetch_current(self)` flows into `LimitQuery<EpochQuery>::query`, which on proofs-disabled SDKs hits `unimplemented!("queries without proofs are not supported yet")`. The refresh path documents itself as best-effort and non-fatal, so a panic on a supported builder configuration would violate that contract. Verify the proofs=false path: if it does panic, gate it behind an early return that either no-ops (preserving the stored version) or returns a typed error via the existing Result channel.
```
</details>

packages/rs-sdk/src/sdk.rs
thepastaclaw: Resolved in `5b0f41d` — _Public `with_initial_version` rustdoc lacks a sub-floor seed warning_ no longer present.

packages/rs-sdk/src/sdk.rs
thepastaclaw: Resolved in `5b0f41d` — _`min_protocol_version` rustdoc still claims a runtime lower-bound that the code no longer enforces_ no longer present.

Claudius-Maginificent: Fixed in `0d7697b` (docs-only, landing on the next push). The rustdoc no longer claims a runtime floor. It now states this is the per-network *default* seed, used only when an unpinned SDK has no explicit initial version; that `SdkBuilder::with_initial_version` can seed *below* it (no construction-time clamp); and that auto-detect (`maybe_update_protocol_version`) only ratchets the stored version *upward* via `fetch_max`.

packages/rs-sdk/src/sdk.rs
Claudius-Maginificent: Fixed in `0d7697b` (docs-only, landing on the next push). The rustdoc now documents that the seed is used verbatim — including versions *below* the per-network floor (no construction-time clamp; supplying a valid seed is the caller's responsibility) — that a sub-floor seed is only corrected once a proven response ratchets the version upward, and that callers needing eager on-init discovery should call `Sdk::refresh_protocol_version` after building.

thepastaclaw (commented):
## Code Review

Latest delta (rustdoc-only on min_protocol_version, with_initial_version, and Sdk::version) resolves the prior floor/seed precision findings (prior-3, prior-4). Three carried-forward prior findings remain valid against head 0d7697b3: refresh_protocol_version still panics via unimplemented! when the SDK is built with proofs disabled, its Result<u32, Error> shape still never produces Err, and Sdk::version()'s rustdoc still omits the now-public with_initial_version seed path. One new related nit: parse_proof_with_metadata_and_proof's bootstrap-limit rustdoc still references latest() as the seed, which no longer matches the per-network floor seed used in practice.

💬 1 nitpick(s)

thepastaclaw (commented):
## Code Review

Incremental review against head 54b8cc71. Latest delta (0d7697b3..54b8cc71) is documentation-only (Swift SwiftExampleApp TEST_PLAN.md removing TOK-17) and does not touch the rs-sdk protocol-version refactor. All four prior findings remain valid against the current sdk.rs and are carried forward unchanged; no new in-scope issues were introduced.

💬 1 nitpick(s)

packages/rs-sdk/src/sdk.rs
thepastaclaw: Resolved in `0d7697b` — _refresh_protocol_version may panic when SDK is built with proofs disabled_ no longer present.

packages/rs-sdk/src/sdk.rs
thepastaclaw: Resolved in `0d7697b` — _[Carried-forward STILL VALID] Sdk::version() rustdoc omits with_initial_version as a seed-override path_ no longer present.

packages/rs-sdk/src/sdk.rs
Claudius-Maginificent: Fixed in `33a69c6`. `refresh_protocol_version` now early-returns `Ok(self.protocol_version_number())` when proofs are disabled (`!self.prove()`), before the epoch query's `unimplemented!()` can abort the process — so the documented non-fatal/best-effort contract now holds on the proofs-off path too. The rustdoc was updated to state the proofs-disabled no-op explicitly. Verified: `cargo test -p dash-sdk --lib refresh` and `--doc` both green; clippy clean.

packages/rs-sdk/src/sdk.rs
Claudius-Maginificent: Fixed in `33a69c6`. `Sdk::version()` rustdoc now names the third seed path: the unpinned starting point is the per-network `min_protocol_version` *or* the seed set via `SdkBuilder::with_initial_version`, then auto-detect ratchets upward via `fetch_max`.

packages/rs-sdk/src/sdk.rs
```diff
@@ -405,6 +355,45 @@ impl Sdk {
...
+        Ok(self.protocol_version_number())
+    }
```
thepastaclaw, line 395: **🟡 Suggestion: New proofs-disabled refresh branch has no test**

The `!self.prove()` early-return at lines 381–383 is the regression-fix the latest delta adds to prevent `refresh_protocol_version` from panicking on a `with_proofs(false)` SDK. The two other branches of this method each have a dedicated unit test in this file (`test_refresh_leaves_pinned_sdk_unchanged` for the pinned arm and `test_refresh_query_unavailable_keeps_current_version` for the fetch-error arm), but the proofs-disabled arm — the exact configuration this commit is in the tree to protect — is not asserted. Adding a tokio test that builds `SdkBuilder::new_mock().with_proofs(false).build()`, calls `refresh_protocol_version()`, and asserts the seeded version is returned without recording any mock expectation would pin the documented contract ("On a proofs-disabled SDK … this is a no-op") so a future refactor cannot silently re-introduce the panic by inverting or removing the guard.

packages/rs-sdk/src/sdk.rs
```diff
@@ -417,14 +406,14 @@ impl Sdk {
...
+    /// first call to this method uses the per-network [`min_protocol_version`] floor as a fallback
+    /// because no network response has been received yet to teach the SDK the real network version.
     ///
     /// The actual network version is learned only *after* proof parsing succeeds, when
     /// [`Self::verify_response_metadata()`] processes `metadata.protocol_version`.  If the
     /// connected network runs an older protocol version **and** proof interpretation differs
-    /// between that version and `latest()`, the very first request may fail before the SDK can
-    /// correct itself.  Subsequent requests will use the correct version.
+    /// between that version and the seeded [`min_protocol_version`], the very first request may
+    /// fail before the SDK can correct itself.  Subsequent requests will use the correct version.
```
thepastaclaw, line 416: **💬 Nitpick: parse_proof bootstrap doc omits the now-public with_initial_version seed path**

The bootstrap-hazard rustdoc on `parse_proof_with_metadata_and_proof` (lines 408–416) was correctly updated away from `latest()` and now talks about the per-network `min_protocol_version` floor as the first-request seed. This PR also promotes `SdkBuilder::with_initial_version` from `#[cfg(test)] pub(crate)` to `pub`, so on a fresh auto-detect SDK the actual first-request seed is whichever of `min_protocol_version(network)` or `with_initial_version` the caller chose — including a value below the floor. The doc should mention that seed-override path alongside the floor so callers using the new public API see the same first-request hazard description that applies to them, and so the cross-reference to `refresh_protocol_version`/eager-init guidance applies to both seeding paths.

<sub>source: ['codex']</sub>

thepastaclaw (commented):
## Code Review

Incremental review at 33a69c64 reconciles four prior findings: the proofs-disabled refresh panic is fixed via the early-return guard, the `Sdk::version()` rustdoc now references `with_initial_version`, and the `parse_proof_with_metadata_and_proof` bootstrap doc is reframed around the per-network floor seed. The single carried-forward concern is API-shape: `refresh_protocol_version` still returns `Result<u32, Error>` while every branch (proofs-off, pinned, fetch error) returns `Ok`. New verified concerns are a missing test for the new proofs-off no-op branch and minor doc gaps (the parse_proof bootstrap doc still omits the `with_initial_version` seed path, and several public rustdoc links target the now-module-private `min_protocol_version`).

<details>

```

- [SUGGESTION] packages/rs-sdk/src/sdk.rs:380-395: New proofs-disabled refresh branch has no test
  The `!self.prove()` early-return at lines 381–383 is the regression-fix the latest delta adds to prevent `refresh_protocol_version` from panicking on a `with_proofs(false)` SDK. The two other branches of this method each have a dedicated unit test in this file (`test_refresh_leaves_pinned_sdk_unchanged` for the pinned arm and `test_refresh_query_unavailable_keeps_current_version` for the fetch-error arm), but the proofs-disabled arm — the exact configuration this commit is in the tree to protect — is not asserted. Adding a tokio test that builds `SdkBuilder::new_mock().with_proofs(false).build()`, calls `refresh_protocol_version()`, and asserts the seeded version is returned without recording any mock expectation would pin the documented contract ("On a proofs-disabled SDK … this is a no-op") so a future refactor cannot silently re-introduce the panic by inverting or removing the guard.
```
</details>

thepastaclaw (commented):
## Code Review

Cumulative review for the rs-sdk protocol-version refactor at ec3d164f. The latest delta (33a69c64..ec3d164f) is a v3.1-dev merge bringing only Swift example-app work (PR #3926); no new Rust changes were introduced. All four prior in-scope findings remain unresolved in the current head and are carried forward unchanged. None are blocking — the only behavioral one (Result with no Err path) is a public-API shape concern best fixed now before #3902 wires this signature into FFI/WASM. Overflow: none (4 of 10 budget used).

💬 1 nitpick(s)

QuantumExplorer (approved):

thepastaclaw (commented):
## Code Review

Verified the four prior findings against head c4b795d1 — all four remain STILL VALID; no new in-scope findings appeared in this delta (which is otherwise dominated by an unrelated v3.1-dev merge bringing in #3917, #3937, #3872). `refresh_protocol_version` still returns `Result<u32, Error>` with no path producing `Err`; the proofs-disabled short-circuit at lines 381-383 still has no dedicated unit test; the `parse_proof_with_metadata_and_proof` bootstrap rustdoc still omits the now-public `SdkBuilder::with_initial_version` seed path; and several public rustdoc items still intra-doc-link to the module-private `min_protocol_version`. None of these are blocking, so the review action is COMMENT.

💬 1 nitpick(s)

PR #3902, opened 2026-06-15
feat(sdk): refresh protocol version on SDK init (FFI/WASM/JS)

Claudius-Maginificent:
## Why this PR exists

**Problem.** A freshly built SDK seeds its protocol version to the per-network `min_protocol_version` floor and only ratchets it upward once a proven query returns the network's real version. Until that first query lands, fee-sensitive flows (e.g. shielded pool shield/unshield/transfer/withdraw) reserve fees against the seed floor, not the network's actual protocol version.

**What breaks without it.** Callers that issue a fee-sensitive operation as their very first action — before any other query has taught the SDK the real version — compute reservations against the floor. This PR makes every real init path eagerly learn the network version up front.

**Blocking relationship.** Compiles independently of #3900: the wiring calls the existing `Sdk::refresh_protocol_version()` (present on `v3.1-dev` with the same signature), so it is based on `v3.1-dev`, not stacked atop #3900. It also adds a small `Sdk::is_mock()` predicate to rs-sdk that references only types already on `v3.1-dev`, so independence still holds. Semantically it **pairs with #3900** — that PR collapses the floor to a flat PV10 baseline and drops the clamp; this PR makes that low floor safe by refreshing on init. The two share `rs-sdk/src/sdk.rs`, but edit non-overlapping regions (the `is_mock` insertion sits where #3900 makes no change), so they merge cleanly. Intended to land together.

Eagerly call the existing best-effort `refresh_protocol_version()` on every network-backed SDK init path so the SDK learns the network's real protocol version at construction instead of on the first downstream query.

## What was done?

A best-effort refresh is wired into every **real** (network-backed) init path. Failures never abort SDK creation — they log at `warn` and proceed with the floor version. Pinned SDKs are unaffected: `refresh_protocol_version()` is a no-op when version updating is disabled.

Mock SDKs are skipped **at the helper level**. A new `Sdk::is_mock(&self) -> bool` predicate (added to rs-sdk; additive, non-breaking) lets `best_effort_refresh` detect a mock from the `Sdk` itself and return early — so the FFI init paths call refresh unconditionally without threading a per-branch flag, and mock SDKs never issue a guaranteed-failing proven query.

| Init path | Layer | Behaviour |
|---|---|---|
| `dash_sdk_create` | FFI | Calls refresh; self-skips when the SDK is a mock (null/empty `dapi_addresses`) |
| `dash_sdk_create_extended` | FFI | Calls refresh; self-skips on mock |
| `dash_sdk_create_trusted` | FFI | Always real → refreshes |
| `dash_sdk_create_with_callbacks` | FFI | Covered indirectly — delegates to `dash_sdk_create_extended` |
| `dash_sdk_create_handle_with_mock` | FFI | Mock only — refresh self-skips (no network round-trip) |
| `WasmSdkBuilder::build` | WASM | Now `async`; awaits a best-effort refresh before returning (always network-real) |
| `EvoSDK::connect` | JS/TS | Awaits the now-async `builder.build()` |

A private `best_effort_refresh(sdk, runtime)` helper centralizes the mock self-skip + FFI `runtime.block_on(...)` + log-and-proceed pattern, reusing the same mechanism as the existing `dash_sdk_refresh_protocol_version` FFI wrapper.

> Note: native `rs-sdk` init (`SdkBuilder::build`) is **not** wired here — it stays synchronous and cannot await the async refresh. Native callers refresh explicitly or rely on the lazy first-query ratchet. Tracked as follow-up.

## How Has This Been Tested?

- `cargo build -p dash-sdk` and `cargo build -p dash-sdk --no-default-features` — clean in both configs (the no-default-features build exercises the `#[cfg(not(feature = "mocks"))]` arm of `is_mock`, since `mocks` is in dash-sdk's default set).
- `cargo check -p rs-sdk-ffi -p wasm-sdk` — clean on `v3.1-dev` (this is what determined the base branch).
- `cargo clippy -p dash-sdk -p rs-sdk-ffi --all-targets -- -D warnings` and `cargo clippy -p wasm-sdk -- -D warnings` — clean.
- `cargo fmt -p dash-sdk -p rs-sdk-ffi -p wasm-sdk -- --check` — clean.
- JS/TS consistency verified: `WasmSdkBuilder::build` is `async`, `EvoSDK::connect` is `async` and awaits it.
- **End-to-end (paloma devnet):** validated as part of the combined floor + refresh-on-init branch before the floor work was split into #3900 — 5/5 identity flows passed (`cr_003`, `id_001`, `id_002`, `id_003`, `id_005`; ~102s; `--test-threads=1`). The SDK seeded at PV10 and auto-ratcheted to the network's active PV12 on the first proven response, confirming the refresh-on-init path lands the client on the correct version.

## Breaking Changes

`WasmSdkBuilder.build()` is now `async` (returns a `Promise<WasmSdk>`). JS/TS callers must `await` it. The in-repo caller (`EvoSDK::connect`) and the wasm-sdk tests already await `build()`, so no internal breakage; downstream JS consumers calling `build()` synchronously must add `await`. The new `Sdk::is_mock()` is purely additive (non-breaking).

## Related

- **#3900** — collapses the protocol-version floor to a flat PV10 baseline, unifies `version_pinned`, drops the post-build clamp, and relocates the `refresh_protocol_version` primitive this PR calls. Intended to land together.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added the ability to detect when the SDK is running in mock mode.

* **Bug Fixes**
  * Improved protocol version initialization handling during SDK setup; version refresh failures are now logged as warnings rather than blocking initialization.

* **Chores**
  * Enhanced SDK initialization to use asynchronous patterns for better resource handling.


thepastaclaw:
✅ Review complete (commit 6b3e934b)

packages/rs-sdk-ffi/src/sdk.rs
```diff
@@ -26,6 +26,23 @@ pub struct DashSDKConfigExtended {
...
+        return;
+    }
+    match runtime.block_on(sdk.refresh_protocol_version()) {
+        Ok(v) => debug!(protocol_version = v, "protocol version refreshed on init"),
+        Err(e) => warn!(
+            error = %e,
+            "protocol version refresh failed on init; proceeding with floor version"
+        ),
+    }
+}
```
thepastaclaw, line 44: **💬 Nitpick: best_effort_refresh's Err arm is unreachable**

Sdk::refresh_protocol_version (packages/rs-sdk/src/sdk/refresh.rs:73-104) already catches ExtendedEpochInfo::fetch_current errors, logs them at warn under target dash_sdk::protocol_version, then unconditionally returns Ok(self.protocol_version_number()). The outer Err(e) => warn!("protocol version refresh failed on init; ...") arm here (and the equivalent in packages/wasm-sdk/src/sdk.rs:317-322) can therefore never fire. The user-visible failure log on init is actually the inner 'proven protocol-version refresh failed; keeping current version' line, not the more descriptive outer message. Either drop the match in favor of letting the inner log speak for itself, or change refresh_protocol_version to surface the inner error so the documented outer warn actually runs.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

PR cleanly wires best-effort refresh_protocol_version into rs-sdk-ffi, wasm-sdk, and js-evo-sdk init paths and adds Sdk::is_mock so mocks self-skip. Verified against HEAD 17428831c5. Two valid concerns: (1) the helper's outer Err arm is unreachable because refresh_protocol_version always returns Ok by design; (2) in dash_sdk_create_trusted the refresh runs synchronously before the trusted provider's quorum cache prefetch, which both blocks the FFI caller on a network round-trip the trusted path deliberately backgrounds AND means the proven refresh likely either falls back to lazy quorum fetching or fails proof verification on cold start. The non-trusted WASM/JS default path also cannot learn the network protocol version because WasmContext::get_quorum_public_key unconditionally errors, but this is a pre-existing limitation of non-trusted WASM mode rather than a regression introduced here.

<details>

```

In `packages/rs-sdk-ffi/src/sdk.rs`:
- [SUGGESTION] packages/rs-sdk-ffi/src/sdk.rs:507-535: Trusted-setup init blocks on network and refreshes before quorum cache prefetch
  dash_sdk_create_trusted was deliberately structured to return quickly: provider_for_prefetch.update_quorum_caches() is spawned on the runtime (line 531) so the FFI caller is not blocked on a network round-trip. Inserting best_effort_refresh(&sdk, &runtime) at line 509 reverses this design intent in two ways. (1) It runs runtime.block_on(sdk.refresh_protocol_version()) synchronously before returning the handle, forcing the caller to wait on a proven getEpochsInfo round-trip. (2) Sdk::refresh_protocol_version issues a proven query whose verification path needs quorum public keys, but TrustedHttpContextProvider's quorum cache has not yet been warmed at this point — the prefetch is spawned afterward. The proven query will therefore either trigger an on-demand quorum fetch inline (defeating the prefetch design) or fall through refresh.rs's warn-and-keep-current-version branch, leaving the seed protocol version at the floor — exactly the symptom this PR aims to eliminate. Move the refresh into the spawned async block after update_quorum_caches().await completes, or await the quorum prefetch before calling best_effort_refresh.
```
</details>

thepastaclaw (changes requested):
## Code Review

The cumulative PR cleanly wires a best-effort proven protocol-version refresh into FFI/WASM/JS init paths and adds the supporting `Sdk::is_mock()` predicate. One new blocking issue: making `WasmSdkBuilder::build` async breaks three pre-existing Rust unit tests in the same file that still call `.build().expect(...)` synchronously — `cargo check -p wasm-sdk --tests` fails with three E0599 errors. Two prior nitpicks (unreachable outer Err arm, default WASM builders unable to actually verify the refresh) and the prior suggestion about trusted setup losing its previously non-blocking startup remain valid against HEAD and are carried forward.

_4 additional finding(s) omitted (not in diff)._

<details>

```

In `packages/wasm-sdk/src/sdk.rs`:
- [BLOCKING] packages/wasm-sdk/src/sdk.rs:542-617: Async `WasmSdkBuilder::build` breaks in-crate unit tests that still call `.build().expect(...)` synchronously
  This PR flips `WasmSdkBuilder::build` from sync to `async`, so it now returns `impl Future<Output = Result<WasmSdk, WasmSdkError>>`. Three pre-existing `#[test]` functions in the same file were not updated and still call `.build().expect("...")`:

  - `with_trusted_context_preserves_user_addresses` (line 550-551)
  - `with_trusted_context_replaces_preset_addresses` (line 590-591)
  - `with_trusted_context_no_discovered_keeps_existing_addresses` (line 615-616)

  Verified locally on HEAD (6b3e934b): `cargo check -p wasm-sdk --tests` fails with three `error[E0599]: no method named 'expect' found for opaque type 'impl Future<Output = ...>'`. The author's earlier `cargo check -p wasm-sdk` and `cargo build -p wasm-sdk` passed because they only build the library target; CI / `cargo test -p wasm-sdk` / `cargo clippy --all-targets` will fail after merge. The tests landed on the branch via the merge from `v3.1-dev`, which is why they were not visible in the initial async-build commit.

  Fix: switch each affected test to an async test attribute (`#[tokio::test]`) and `await` the build call before `.expect(...)`, e.g. `.build().await.expect("...")`. Pinning a protocol version (`.with_version(PlatformVersion::latest().protocol_version)`) is optional and would short-circuit the refresh round-trip for purely structural address-list assertions.
- [SUGGESTION] packages/wasm-sdk/src/sdk.rs:336-348: Default (non-trusted) WASM/JS builds cannot actually refresh the protocol version
  `WasmSdkBuilder::{new_mainnet,new_testnet,new_local,new_devnet,new_with_addresses}` install `WasmContext` as the context provider unless the caller later attaches a `WasmTrustedContext` via `withTrustedContext`. `WasmContext::get_quorum_public_key` unconditionally returns an error (`context_provider.rs:31-41`), so the proven `getEpochsInfo` query that `refresh_protocol_version` issues cannot verify quorum signatures and will always fail. The new `build()` then just logs `"protocol version refresh failed on init; proceeding with floor version"` and returns — so default WASM/JS builds advertise refresh-on-init but never actually leave the floor version. Either document that this code path is a no-op without `withTrustedContext`, or short-circuit the refresh (and skip the warning) when a non-trusted `WasmContext` is in use so the log is not misleading.

In `packages/rs-sdk-ffi/src/sdk.rs`:
- [SUGGESTION] packages/rs-sdk-ffi/src/sdk.rs:507-535: Trusted-setup init now blocks on a network round-trip before the previously-async quorum prefetch
  `dash_sdk_create_trusted` was deliberately structured so that `provider_for_prefetch.update_quorum_caches()` runs on `runtime.spawn(async move { … })` (line 515-535), letting the FFI caller get its SDK handle back without waiting on a network round-trip. Inserting `best_effort_refresh(&sdk, &runtime)` at line 509 — which calls `runtime.block_on(sdk.refresh_protocol_version())` — undoes that property: trusted-setup init now synchronously waits for a proven `getEpochsInfo` query against the seeded address list before returning, bounded only by the SDK's default request timeouts (multi-second on unreachable evonodes). Embedders that previously got a fast handle now see init latency climb on flaky networks, with no inner deadline available to opt out.

  Consider either (a) wrapping the refresh in `tokio::time::timeout(<small>, …)` so a misconfigured or unreachable network degrades within a known budget, or (b) spawning the refresh on the runtime so it runs alongside the existing quorum prefetch instead of in front of it (mirroring the original non-blocking design). Either path preserves the proven ratchet's correctness while keeping trusted-setup init fast.
```
</details>

lklimek:
This solution is not correct. Depending on configuration, proof verification can require SPV to be synced. This means code added in this PR will always fail. Closing.

Issue #3897, opened 2026-06-14
iOS SwiftExampleApp — TEST_PLAN QA run status & blockers


QuantumExplorer:
Living tracker for QA runs of the iOS **SwiftExampleApp** against [`packages/swift-sdk/SwiftExampleApp/TEST_PLAN.md`](https://github.com/dashpay/platform/blob/v3.1-dev/packages/swift-sdk/SwiftExampleApp/TEST_PLAN.md), driven on the booted iOS simulator (via `idb`, the `simulator-control` skill). Updated per build. Distinguishes real pass/fail from environment/tooling blockers.

**Environment:** testnet · fresh `v3.1-dev` build · iPhone 17 simulator · fully-provisioned wallet (4 identities, Core/Platform/shielded balances).

---

## Essential tier — ✅ all passing

All Essential-tier tests pass:
- **Core wallet:** create / restore / backup phrase / receive / send Core L1 / view balance+tx / SPV sync.
- **Identity:** create (Core-funded), load/discover, view, identity→identity credit transfer (production UI added in #3891).
- **DPNS:** register, check availability, resolve, get usernames.
- **Shielded:** sync, view activity, shield from Platform & Core, shielded→shielded transfer, unshield.

> `CORE-08` (QR scan) is classified **Manual** (needs a physical camera — not automatable in the simulator), so it sits in its own tier rather than Essential.

## Shielded (cross-cutting) — ✅ passing

SH-01 sync · SH-02 view · SH-03 shield-from-Platform · SH-04/SH-07 shield-from-Core (asset lock) · SH-05 shielded transfer · SH-06 unshield · SH-08 withdraw→Core · SH-09 prover · SH-13 receive address — all driven green.
An earlier "no available addresses" failure was **diagnosed as an address-list seeding/timing race, not a defect** (the shielded path shares the main SDK; it works once the masternode/evonode list is seeded). See #3890.

## Common tier — 24 / 32 driven

| ID | Action | Result |
|---|---|---|
| CORE-09 | Multiple HD accounts | ✅ |
| CORE-10 | Multi-recipient Core send | ✅ UI added (Send → "Add recipient"); 2-output testnet send verified |
| ID-05 | Top up (asset lock) | ✅ picker fix — driven end-to-end (topped up idx0 from Platform Payment #0; balance increased) |
| ID-06 | Top up (from platform addrs) | ✅ picker fix (`TopUpIdentityView` funds from Platform Payment account addresses — same flow as ID-05) |
[...]

PR #3905, opened 2026-06-15
feat(swift-example-app): one-tap testnet DASH faucet button

QuantumExplorer:

Fresh or drained testnet wallets repeatedly block fixture-dependent iOS QA runs (create-contract, create-doc, token fixtures) tracked in #3897. Today funding means copying the Core address out of the app and pasting it into a web faucet by hand. This adds a **one-tap** "Get testnet DASH" convenience so a wallet can be topped up from inside the app.

## What was done?

Adds a **testnet-only** "Get 1 tDASH — Testnet Faucet" button to the **Receive → Core** tab (next to the existing local-Docker faucet button), gated by `selectedTab == .core && platformState.currentNetwork == .testnet` so it's hidden on mainnet/devnet/regtest. It's tinted orange with a wrench icon and marked as a dev tool.

On tap it funds the wallet's current Core receive address (the address shown on the tab) via `faucet.thepasta.org`:

1. `GET /api/status` — reads the (rotatable) cap endpoint and the faucet amount (no hardcoded cap path).
2. Solves the faucet's `cap.js` **proof-of-work** captcha **on-device, off the main actor** (`POST {capEndpoint}challenge` → derive the sub-challenges via cap.js's seeded PRNG → brute-force the SHA-256 nonces in parallel → `POST {capEndpoint}redeem`).
3. `POST /api/core-faucet {address, capToken}` → on `200` shows the txid; normal SPV sync picks up the inbound funds.

**Failure handling:** on rate-limit (`429`, the faucet allows 3/IP/hour) or any other failure (solve failure, network error, non-200), it falls back to copying the address to the clipboard and opening the web faucet in the browser — never fails silently. The faucet's "hard" captcha (rate-limit bypass, ~100× the work) is intentionally **not** auto-solved on-device.

**Scope / architecture:** this is an app-level dev/testing convenience — a small `URLSession` client plus a pure-Swift `CryptoKit` PoW solver. Per `packages/swift-sdk/CLAUDE.md` it deliberately does **not** route through the wallet/Platform FFI and adds **no** SDK business logic. No Rust/FFI/SDK code changed.

### Files
- **New** `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/TestnetFaucetService.swift` — `CapSolver` (cap.js PRNG + PoW solver) and the `TestnetFaucet` HTTP client (`TestnetFaucetOutcome` = `.sent` / `.rateLimited` / `.failed`).
- **Modified** `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Views/ReceiveAddressView.swift` — the gated button, its status toast, and the clipboard + web-faucet fallback (`@Environment(\.openURL)`).

## How Has This Been Tested?

Built with `./build_ios.sh --target sim` → **BUILD SUCCEEDED** (both changed files compiled, 0 errors), then driven on the iPhone 17 simulator (iOS 26.3):

- **Button gating (positive):** on a **testnet** wallet, the orange "Get 1 tDASH — Testnet Faucet" button (`accessibilityIdentifier receive.testnetFaucetButton`) is present on the Receive → Core tab.
- **Button gating (negative):** switched the app to **devnet** — the Receive → Core sheet ("Your Core Address") shows only *Done* and *Copy Address*; the faucet button is **absent**. (Hidden on mainnet/regtest by the same `== .testnet` guard.)
- **One-tap funding (end-to-end, live testnet):** tapping the button transitioned the label `Solving captcha…` → `Sent 1 tDASH! tx: b0d0f1f9dbf5…`. OSLog network trace confirmed the request flow: `GET faucet.thepasta.org/api/status` then `POST cap.thepasta.org/<id>/challenge` (cap.js PoW solved on-device), then redeem + `/api/core-faucet`.
- **On-chain confirmation:** the funded receive address `yeFECyXWbriKb9XAgPvB6tzRnfZyXqrCeW` received exactly **1 tDASH** in txid `b0d0f1f9dbf592289f454f47ae6a2a9e06fc40cca168bfd4bd8796a105b67d1d` (matches the txid shown in the app) — verified via the testnet insight API.
- **Captcha protocol** independently proven against the live service (challenge → solve → redeem → `/api/core-faucet` accepts the token; the Swift PRNG is bit-identical to `@cap.js/widget`).

**Wallet pickup confirmed:** after Core SPV caught up, the wallet's SwiftData store gained the faucet UTXO — `100000000` duffs (1.00000000 DASH) on `yeFECyXWbriKb9XAgPvB6tzRnfZyXqrCeW`, unspent, tx in-block at height 1,496,641 — so the Core balance rose 1.81172874 → 2.81172874 DASH. No automated tests were added — the behavior is a UI dev-tool button over a live HTTP service; verification was manual + against the real faucet/chain.

## Breaking Changes

None. App-only, testnet-gated dev tool; no consensus, FFI, or SDK API changes.

## Checklist:

## Release Notes

* **New Features**
  * Added a testnet faucet button to the receive address view for Swift SDK users.
  * Testnet faucet displays real-time status updates, including "Solving captcha…" progress.
  * Successfully funded addresses show transaction ID and amount received.
  * Rate limiting and error handling gracefully redirect to the web faucet interface when needed.

### DIFF
--- a/packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/TestnetFaucetService.swift
+++ b/packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/TestnetFaucetService.swift
@@ -1,0 +1,358 @@
+import CryptoKit
+import Foundation
+
+/// **Dev/QA tool only.** App-level client for the testnet DASH faucet at
+/// `faucet.thepasta.org`. This is a plain `URLSession` HTTP client plus a
+/// pure-Swift `cap.js` proof-of-work solver — it deliberately does **not**
+/// touch the wallet / Platform FFI, the `platform-wallet` crate, or any
+/// SDK business logic. It exists purely so QA can top up a fresh/drained
+/// testnet wallet with one tap instead of round-tripping through the web
+/// faucet (issue #3897).
+///
+/// The faucet enforces a server-side captcha (`cap.js`). The "soft"
+/// challenge is a small proof-of-work (≈6.5M SHA-256 ops) that we solve
+/// on-device off the main actor; the "hard" challenge returned on rate
+/// limit is far too expensive to brute-force here, so on rate limit (or
+/// any other failure) the caller falls back to opening the web faucet.
+
+// MARK: - cap.js proof-of-work solver
+
+enum CapSolver {
+    /// cap.js seeded PRNG: FNV-1a-style seed + xorshift32. Emits `length`
+    /// lowercase hex chars.
+    ///
+    /// Verified bit-identical to `@cap.js/widget` against the live server
+    /// (cross-checked with `/tmp/cap_wrap_check.py` — both the wrapping and
+    /// the masked int32 forms agree, and the wrapping form redeems live).
+    static func prng(_ seed: String, _ length: Int) -> String {
+        var state: UInt32 = 2166136261                 // 0x811C9DC5
+        for u in seed.utf16 {                          // seeds are ASCII (hex token + digits [+ "d"])
+            state ^= UInt32(u)
+            let sh = (state << 1) &+ (state << 4) &+ (state << 7) &+ (state << 8) &+ (state << 24)
+            state = state &+ sh
+        }
+        var out = ""
+        while out.count < length {
+            state ^= state << 13
+            state ^= state >> 17
+            state ^= state << 5
+            out += String(format: "%08x", state)        // unsigned 32-bit, 8 hex chars
+        }
+        return String(out.prefix(length))
+    }
+
+    /// A single sub-challenge: brute-force the smallest non-negative
+    /// integer `nonce` such that lowercase-hex(`SHA256(salt + String(nonce))`)
+    /// starts with `target`.
+    ///
+    /// We never materialize the 64-char hex digest per attempt. `target`
+    /// is parsed once into a byte prefix (plus an optional trailing high
+    /// nibble when `target` has odd length) and compared directly against
+    /// the raw `SHA256` digest bytes.
+    static func solve(salt: String, target: String) -> Int {
+        let targetChars = Array(target.utf8)
+        let fullBytes = targetChars.count / 2
+        let hasNibble = targetChars.count % 2 == 1
+
+        // Parse the full leading bytes of the hex target.
+        var prefix = [UInt8]()
+        prefix.reserveCapacity(fullBytes)
+        for i in 0..<fullBytes {
+            let hi = hexNibble(targetChars[i * 2])
+            let lo = hexNibble(targetChars[i * 2 + 1])
+            prefix.append((hi << 4) | lo)
+        }
+        // Optional trailing high-nibble (odd-length target).
+        let nibble: UInt8 = hasNibble ? hexNibble(targetChars[fullBytes * 2]) : 0
+
+        let saltBytes = Array(salt.utf8)
+        var nonce = 0
+        while true {
+            var hasher = SHA256()
+            hasher.update(data: saltBytes)
+            hasher.update(data: Array(String(nonce).utf8))
+            let digest = hasher.finalize()
+
+            if digestHasPrefix(digest, prefix: prefix, nibble: nibble, hasNibble: hasNibble) {
+                return nonce
+            }
+            nonce += 1
+        }
+    }
+
+    private static func digestHasPrefix(
+        _ digest: SHA256.Digest,
+        prefix: [UInt8],
+        nibble: UInt8,
+        hasNibble: Bool
+    ) -> Bool {
+        var idx = 0
+        for byte in digest {
+            if idx < prefix.count {
+                if byte != prefix[idx] { return false }
+                idx += 1
+            } else if hasNibble {
+                // Compare only the high nibble of the next digest byte.
+                return (byte >> 4) == nibble
+            } else {
+                return true
+            }
+            if idx == prefix.count && !hasNibble {
+                return true
+            }
+        }
+        // Digest exhausted: matched iff we consumed the whole prefix and
+        // there's no dangling nibble to check.
+        return idx == prefix.count && !hasNibble
+    }
+
+    private static func hexNibble(_ c: UInt8) -> UInt8 {
+        switch c {
+        case 0x30...0x39: return c - 0x30            // '0'-'9'
+        case 0x61...0x66: return c - 0x61 + 10       // 'a'-'f'
+        case 0x41...0x46: return c - 0x41 + 10       // 'A'-'F'
+        default: return 0
+        }
+    }
+}
+
+// MARK: - Outcome
+
+/// Result of a faucet funding attempt. The view turns `.failed` /
+/// `.rateLimited` into the clipboard + web-faucet fallback.
+enum TestnetFaucetOutcome: Sendable {
+    /// Funds were sent. `txid` is the Core transaction id; `amount` is the
+    /// tDASH amount reported by `/api/status` (`coreFaucetAmount`).
+    case sent(txid: String, amount: Double)
+    /// Server rate-limited the request (429). Carries a user-facing message.
+    case rateLimited(message: String)
+    /// Any other failure (network error, solve failure, non-200 response).
+    case failed(reason: String)
+}
+
+// MARK: - HTTP client
+
+/// Thin async client for the testnet faucet. Stateless; create one per use.
+struct TestnetFaucet {
+    /// Faucet web host. Used both for the JSON API and as the web fallback URL.
+    static let webURL = URL(string: "https://faucet.thepasta.org/")!
+
+    private let host = URL(string: "https://faucet.thepasta.org")!
+    private let session: URLSession
+
+    init(session: URLSession = .shared) {
+        self.session = session
+    }
+
+    /// Request 1 tDASH to `address`. Solves the soft captcha on-device and
+    /// posts to `/api/core-faucet`. Safe to call from the main actor — the
+    /// proof-of-work runs on a detached background-priority task.
+    func requestCoreDash(address: String) async -> TestnetFaucetOutcome {
+        let trimmed = address.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard !trimmed.isEmpty else {
+            return .failed(reason: "No address available")
+        }
+
+        // 1. Read status for the (rotatable) cap endpoint + faucet amount.
+        let status: FaucetStatus
+        do {
+            status = try await fetchStatus()
+        } catch {
+            return .failed(reason: "Faucet status unavailable: \(error.localizedDescription)")
+        }
+
+        guard let capBase = URL(string: status.capEndpoint) else {
+            return .failed(reason: "Invalid captcha endpoint")
+        }
+
+        // 2. Solve the soft captcha proof-of-work.
+        let capToken: String
+        do {
+            capToken = try await solveCaptcha(capBase: capBase)
+        } catch {
+            return .failed(reason: "Captcha solve failed: \(error.localizedDescription)")
+        }
+
+        // 3. Post the funding request.
+        return await postFaucet(
+            address: trimmed,
+            capToken: capToken,
+            amount: status.coreFaucetAmount
+        )
+    }
+
+    // MARK: status
+
+    private struct FaucetStatus: Decodable {
+        let capEndpoint: String
+        let coreFaucetAmount: Double
+    }
+
+    private func fetchStatus() async throws -> FaucetStatus {
+        let url = host.appendingPathComponent("api/status")
+        var request = URLRequest(url: url)
+        request.httpMethod = "GET"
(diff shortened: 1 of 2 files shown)


thepastaclaw:
✅ Review complete (commit c05d58b3)

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/TestnetFaucetService.swift
```diff
@@ -0,0 +1,358 @@
...
+                    let nonce = CapSolver.solve(salt: salt, target: target)
+                    return (i, nonce)
+                }
+            }
+            var collected = Array(repeating: 0, count: c)
+            for try await (i, nonce) in group {
+                collected[i - 1] = nonce   // preserve sub-challenge order
+            }
+            return collected
+        }
```
thepastaclaw, line 285: **🟡 Suggestion: Bound server-controlled captcha parameters before solving**

`challenge.c`, `s`, and `d` are taken directly from `/api/status` → `/challenge` and only `c > 0` is enforced. `c` controls how many concurrent `Task`s are spawned (line 240), `s` controls PRNG salt length, and `d` controls the hex-prefix length the PoW must match — each sub-challenge runs `CapSolver.solve`'s `while true` loop (line 70) with expected work ≈ 16^d and no cancellation/timeout. A misbehaving, compromised, or future-changed faucet response can therefore spawn a huge task group, drive `String.prefix` with a pathological length in `prng`, or make the solve effectively infeasible — pinning the device's cores indefinitely with no in-app way to abort (the user must kill the app). The PR explicitly promises that hard/failed captchas fall back to the web faucet, so unsupported challenge sizes should be rejected before any work starts. Add cheap bounds (e.g. `c <= 100`, `s <= 128`, `d <= 6`) plus a wrapping timeout, throwing `FaucetError.message(...)` so the existing fallback path opens the web faucet.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Views/ReceiveAddressView.swift
```diff
@@ -408,4 +435,57 @@ struct ReceiveAddressView: View {
...
+        switch outcome {
+        case .sent(let txid, let amount):
+            testnetFaucetStatus = "Sent \(formattedAmount(amount)) tDASH! tx: \(txid.prefix(12))…"
+        case .rateLimited(let message):
+            testnetFaucetStatus = message
+            openWebFaucetFallback(address: address)
+        case .failed(let reason):
+            testnetFaucetStatus = reason
+            openWebFaucetFallback(address: address)
+        }
```
thepastaclaw, line 464: **🟡 Suggestion: Failure reason is overwritten by the fallback toast before the user sees it**

In `.rateLimited` and `.failed`, `testnetFaucetStatus` is set to the carried message/reason and then immediately overwritten by `openWebFaucetFallback`, which unconditionally assigns `testnetFaucetStatus = "Opened web faucet — address copied"` (line 476). Both assignments happen in the same synchronous tick on the main actor, so SwiftUI only renders the second value — the actual reason (rate-limit text, captcha-solve failure, network error, non-200 detail) never reaches the UI. This defeats the value of the typed `TestnetFaucetOutcome` and weakens the PR's "never fails silently" guarantee. Fold the reason into the fallback toast (e.g. `"Rate limited — opened web faucet (address copied)"`) or surface it briefly before the fallback message replaces it.

<sub>source: ['claude']</sub>

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/TestnetFaucetService.swift
```diff
@@ -0,0 +1,358 @@
...
+        let status: FaucetStatus
+        do {
+            status = try await fetchStatus()
+        } catch {
+            return .failed(reason: "Faucet status unavailable: \(error.localizedDescription)")
+        }
+
+        guard let capBase = URL(string: status.capEndpoint) else {
+            return .failed(reason: "Invalid captcha endpoint")
+        }
```
thepastaclaw: **🟡 Suggestion: Constrain capEndpoint to HTTPS on the faucet host**

`status.capEndpoint` is fed straight into `URL(string:)` and then `appendingPathComponent("challenge")` / `"redeem"` without any scheme/host check. A faucet response that returns an `http://` URL would silently downgrade the captcha exchange (leaking the redeemed capToken to anyone on path), and a fully attacker-controlled URL would turn the app into a POST client for an arbitrary endpoint that also receives the receive-address-correlated request. The client is dedicated to `faucet.thepasta.org`, so requiring `scheme == "https"` and `host == host.host` before issuing the POSTs is free defense-in-depth.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/TestnetFaucetService.swift
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    private func fetchStatus() async throws -> FaucetStatus {
+        let url = host.appendingPathComponent("api/status")
+        var request = URLRequest(url: url)
+        request.httpMethod = "GET"
+        request.timeoutInterval = 30
+        let (data, _) = try await session.data(for: request)
+        return try JSONDecoder().decode(FaucetStatus.self, from: data)
+    }
```
thepastaclaw, line 220: **💬 Nitpick: fetchStatus ignores HTTP status code**

`fetchStatus` discards the `URLResponse` and immediately JSON-decodes the body, so a non-200 from `/api/status` (5xx error page, 404 HTML, captive-portal interstitial) surfaces as a generic `DataCorrupted` decoder error and the user sees `Faucet status unavailable: The data couldn't be read…`. `postJSON` already checks `http.statusCode != 200`; applying the same check here produces a more actionable fallback toast.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

PR is a Swift example-app-only testnet faucet convenience: a SwiftUI button plus a CryptoKit-based cap.js PoW solver and URLSession client. No FFI/SDK/consensus surface is touched. Findings are all UX or defense-in-depth around blindly trusting server-controlled captcha parameters and endpoints; nothing is consensus-critical or a hard blocker for a testnet dev tool.

<details>

```

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/TestnetFaucetService.swift`:
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/TestnetFaucetService.swift:227-253: Bound server-controlled captcha parameters before solving
  `challenge.c`, `s`, and `d` are taken directly from `/api/status` → `/challenge` and only `c > 0` is enforced. `c` controls how many concurrent `Task`s are spawned (line 240), `s` controls PRNG salt length, and `d` controls the hex-prefix length the PoW must match — each sub-challenge runs `CapSolver.solve`'s `while true` loop (line 70) with expected work ≈ 16^d and no cancellation/timeout. A misbehaving, compromised, or future-changed faucet response can therefore spawn a huge task group, drive `String.prefix` with a pathological length in `prng`, or make the solve effectively infeasible — pinning the device's cores indefinitely with no in-app way to abort (the user must kill the app). The PR explicitly promises that hard/failed captchas fall back to the web faucet, so unsupported challenge sizes should be rejected before any work starts. Add cheap bounds (e.g. `c <= 100`, `s <= 128`, `d <= 6`) plus a wrapping timeout, throwing `FaucetError.message(...)` so the existing fallback path opens the web faucet.
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/TestnetFaucetService.swift:164-166: Constrain capEndpoint to HTTPS on the faucet host
  `status.capEndpoint` is fed straight into `URL(string:)` and then `appendingPathComponent("challenge")` / `"redeem"` without any scheme/host check. A faucet response that returns an `http://` URL would silently downgrade the captcha exchange (leaking the redeemed capToken to anyone on path), and a fully attacker-controlled URL would turn the app into a POST client for an arbitrary endpoint that also receives the receive-address-correlated request. The client is dedicated to `faucet.thepasta.org`, so requiring `scheme == "https"` and `host == host.host` before issuing the POSTs is free defense-in-depth.

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Views/ReceiveAddressView.swift`:
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Views/ReceiveAddressView.swift:460-466: Failure reason is overwritten by the fallback toast before the user sees it
  In `.rateLimited` and `.failed`, `testnetFaucetStatus` is set to the carried message/reason and then immediately overwritten by `openWebFaucetFallback`, which unconditionally assigns `testnetFaucetStatus = "Opened web faucet — address copied"` (line 476). Both assignments happen in the same synchronous tick on the main actor, so SwiftUI only renders the second value — the actual reason (rate-limit text, captcha-solve failure, network error, non-200 detail) never reaches the UI. This defeats the value of the typed `TestnetFaucetOutcome` and weakens the PR's "never fails silently" guarantee. Fold the reason into the fallback toast (e.g. `"Rate limited — opened web faucet (address copied)"`) or surface it briefly before the fallback message replaces it.
```
</details>

QuantumExplorer:
Thanks for the review — addressed all four findings in `1cfb9cb263`:

- **Bound server-controlled captcha params + timeout** (`TestnetFaucetService.swift`): reject `c`/`s`/`d` outside `1...256` / `1...256` / `1...6` before doing any work, wrap the solve in a 30s timeout, and make `CapSolver.solve` honor cancellation (`Task.checkCancellation` in the loop). A misbehaving/changed faucet now throws → the web-faucet fallback opens instead of pinning the device.
- **Constrain capEndpoint**: require `scheme == "https"` and that the endpoint shares the faucet's registrable domain before posting. (Used a registrable-domain check rather than an exact host match, since the cap endpoint is a sibling subdomain — `cap.thepasta.org` vs `faucet.thepasta.org`.)
- **Failure reason overwritten by fallback toast** (`ReceiveAddressView.swift`): `openWebFaucetFallback` now folds the reason into the toast (`"<reason> — opened web faucet (address copied)"`), so the rate-limit/error detail reaches the UI.
- **`fetchStatus` ignores HTTP status** (nitpick): now throws a clear error on non-200 instead of an opaque decode failure. Also added docstrings to lift coverage.

Rebuilt (`build_ios.sh`-produced xcframework, `BUILD SUCCEEDED`) and re-verified end-to-end on the iPhone 17 simulator (testnet): tap → on-device PoW solve → `Sent 1 tDASH! tx: 4526d2dc0eca…`; the wallet's Core balance reflected the credit (→ 2.81172874 DASH).

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/TestnetFaucetService.swift
QuantumExplorer: Fixed in 1cfb9cb263: `c`/`s`/`d` are now rejected unless within `1...256` / `1...256` / `1...6` before any work, the solve is wrapped in a 30s timeout, and `CapSolver.solve` honors `Task.checkCancellation` — out-of-bounds or slow challenges throw `FaucetError` and the web-faucet fallback opens instead of pinning the device.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/TestnetFaucetService.swift
QuantumExplorer: Fixed in 1cfb9cb263: the cap endpoint from `/api/status` is now validated to be `https` and to share the faucet's registrable domain before any POST. (Used a registrable-domain check rather than `host == host.host` because the cap endpoint is a sibling subdomain — `cap.thepasta.org` vs `faucet.thepasta.org`.)

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/TestnetFaucetService.swift
QuantumExplorer: Fixed in 1cfb9cb263: `fetchStatus` now throws on a non-200 (matching `postJSON`) instead of falling through to an opaque JSON-decode error.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Views/ReceiveAddressView.swift
QuantumExplorer: Fixed in 1cfb9cb263: `openWebFaucetFallback(address:reason:)` now folds the reason into the toast (`"<reason> — opened web faucet (address copied)"`) instead of overwriting it, so the rate-limit/error detail reaches the UI.

PR #3906, opened 2026-06-15
feat(swift-example-app): production Withdraw Credits UI on identity detail (ID-10)

QuantumExplorer:

Identity credit withdrawal to an L1 Dash address — iOS test-plan item **ID-10** — was **builder-only**: reachable only via *Settings → Platform State Transitions → Identity Credit Withdrawal*, which calls the low-level `dash_sdk_identity_withdraw` with a *test* signer. There was no production "happy path" button for it. This adds one on the identity detail screen, next to **Transfer Credits** (the production transfer flow added in #3891).

Tracked in QA issue #3897 (ID-10).

## What was done?

The entire **Rust → FFI → Swift-SDK withdraw stack already existed** and mirrors the transfer stack 1:1:
- `platform_wallet_withdraw_credits_with_signer` (rs-platform-wallet-ffi)
- `IdentityWallet::withdraw_credits_with_external_signer` (rs-platform-wallet)
- `ManagedPlatformWallet.withdrawCredits(identityId:amount:toAddress:signer:)` (swift-sdk)

So this PR is **pure Swift UI — no Rust changes**. The broadcast routes through `rs-platform-wallet` using the wallet's keychain-backed signer, per the swift-sdk architecture rule (Swift only collects inputs + bridges; no business logic).

- **New `WithdrawCreditsView.swift`** — mirrors `TransferCreditsView` (DASH→credits parsing, balance validation, `submitGeneration` guard, fresh `KeychainSigner` per submit), swapping the recipient identity picker for a network-validated **destination L1-address** `TextField`. Calls `wallet.withdrawCredits(...)`.
  - Success screen shows the destination + withdrawn amount. It intentionally shows **no L1 txid**, because Platform withdrawals are pooled and paid out asynchronously by the network — there is no txid at broadcast time. The UI says so rather than fabricating one.
  - The credit decrease reflects automatically via the existing persister → SwiftData → `@Query` path (same as transfer).
- **`IdentityDetailView.swift`** — "Withdraw Credits" button + sheet immediately after Transfer Credits, in the same wallet-gated section.
- **`TEST_PLAN.md`** — ID-10 row 🧪 → ✅ (production entry point); updated the entry-point reality-check note.

## How Has This Been Tested?

Built with `./build_ios.sh --target sim` (**BUILD SUCCEEDED**, `-warnings-as-errors`), then driven end-to-end on the booted simulator (`idb`) on **devnet/paloma**:

1. "Withdraw Credits" button renders on identity detail, next to Transfer Credits.
2. Sheet opens; submit button correctly **disabled → enabled** as address + amount validate.
3. Submitted **0.001 DASH** to an L1 address → **"Withdrawal submitted"** success screen.
4. Identity credit balance dropped **0.09659655 → 0.09331446 DASH** (amount + withdrawal fee), confirmed in SwiftData *and* reflected on the detail screen via `@Query`.

**Prerequisite finding:** withdrawal signing needs a TRANSFER/CRITICAL key. A funded identity that predates the keyId-3 derive fix (keys `[AUTH×3, ENC, DEC]`, no transfer key) first failed with a clean *"missing key: no withdrawal public key"* error **and no balance change** — confirming correct error handling and that the FFI bridge runs end-to-end. Adding a TRANSFER key via **ID-07** (Add Key → Purpose = Transfer) made the subsequent withdrawal succeed. Newly-derived identities already get this key.

## Breaking Changes

None. UI-only addition to the example app; no public API or consensus changes.

## Checklist:

- [x] I have added "!" to the title and described breaking changes in the corresponding section if my code contains any (n/a — no breaking changes)
- [x] I have made corresponding changes to the documentation if needed (TEST_PLAN ID-10)

## Summary by CodeRabbit

* **New Features**
  * Users can withdraw platform credits to a Dash L1 address with validation for destination address and available balance.
  * Success confirmation displays withdrawn amount and destination details.
  * New "Withdraw Credits" action added alongside existing credit management options.

* **Documentation**
  * Updated test documentation to reflect the new withdraw credits feature.


thepastaclaw:
✅ Review complete (commit c4ce64c1)

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/WithdrawCreditsView.swift
```diff
@@ -0,0 +1,341 @@
...
+            }
+            .alert(item: $submitError) { err in
+                Alert(
+                    title: Text("Withdrawal failed"),
+                    message: Text(err.message),
+                    dismissButton: .default(Text("OK"))
+                )
+            }
+        }
+    }
```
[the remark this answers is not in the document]
QuantumExplorer: Fixed in c4ce64c18b — added `.interactiveDismissDisabled(isSubmitting)` so the sheet can't be swipe-dismissed mid-broadcast (Cancel was already disabled). Good catch on the duplicate-write risk.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/WithdrawCreditsView.swift
```diff
@@ -0,0 +1,341 @@
...
+    /// value, round to the nearest credit, and reject any result below
+    /// 1 credit or beyond `UInt64.max`.
+    private var parsedCredits: UInt64? {
+        let trimmed = amountDash.trimmingCharacters(in: .whitespaces)
+        guard let dash = Double(trimmed), dash.isFinite, dash > 0 else {
+            return nil
+        }
+        let credits = (dash * Double(Self.creditsPerDash)).rounded()
+        guard credits >= 1, credits <= Double(UInt64.max) else { return nil }
+        return UInt64(credits)
```
thepastaclaw, line 254: **🔴 Blocking: parsedCredits can crash on large amounts: Double(UInt64.max) rounds up to 2^64**

`Double(UInt64.max)` is not exactly representable and rounds up to 2^64, so the `credits <= Double(UInt64.max)` guard admits a value (2^64) that is one above `UInt64`'s representable range. A user typing an amount that scales to ~2^64 credits (DASH input around 1.844e8) makes SwiftUI recompute `parsedCredits` during keystrokes, and `UInt64(credits)` traps before the submit-time re-validation can run — crashing the withdrawal sheet on input rather than producing a friendly error. The same shape was copied from `TransferCreditsView.parsedCredits`; the fix here is to switch the upper bound to a strict `<` so 2^64 is rejected, which keeps the result within UInt64's representable range.

```suggestion
        guard credits >= 1, credits < Double(UInt64.max) else { return nil }
        return UInt64(credits)
```

<sub>source: ['codex']</sub>

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/WithdrawCreditsView.swift
```diff
@@ -0,0 +1,341 @@
...
+    /// `Int64`; clamp any (unexpected) negative value to zero.
+    private var senderBalanceCredits: UInt64 {
+        identity.balance < 0 ? 0 : UInt64(identity.balance)
+    }
+
+    private var canSubmit: Bool {
+        !trimmedAddress.isEmpty
+            && managedWallet != nil
+            && (parsedCredits.map { $0 > 0 && $0 <= senderBalanceCredits } ?? false)
+    }
```
thepastaclaw, line 267: **🟡 Suggestion: Submit enables for any non-empty destination — use DashAddress.parse for local pre-validation**

`canSubmit` requires only that `trimmedAddress` be non-empty, so the Withdraw button lights up for obviously malformed or wrong-network strings and the user only learns it was invalid after the round-trip into Rust. The app already has `DashAddress.parse(_, network:)` (Core/Models/DashAddress.swift) which performs Base58Check + network validation via the same Rust FFI used elsewhere, so a local `.core` check would tighten the UI without duplicating consensus logic — authoritative validation still happens in Rust on submit, matching the file header comment.

```suggestion
    private var isValidDestinationAddress: Bool {
        guard !trimmedAddress.isEmpty else { return false }
        if case .core = DashAddress.parse(trimmedAddress, network: identity.network).type {
            return true
        }
        return false
    }

    private var canSubmit: Bool {
        isValidDestinationAddress
            && managedWallet != nil
            && (parsedCredits.map { $0 > 0 && $0 <= senderBalanceCredits } ?? false)
    }
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):
## Code Review

Pure SwiftUI addition for the production Withdraw Credits flow that mirrors TransferCreditsView shape and correctly defers L1 address + balance validation into Rust. One real crash-class bug carried over from the copied parsedCredits pattern (Double(UInt64.max) rounds up to 2^64, so the guard admits a value that traps on UInt64 cast). One UX suggestion to use the existing network-aware DashAddress.parse so the submit button doesn't enable for obviously invalid destinations.

<details>

```

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/WithdrawCreditsView.swift`:
- [BLOCKING] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/WithdrawCreditsView.swift:253-254: parsedCredits can crash on large amounts: Double(UInt64.max) rounds up to 2^64
  `Double(UInt64.max)` is not exactly representable and rounds up to 2^64, so the `credits <= Double(UInt64.max)` guard admits a value (2^64) that is one above `UInt64`'s representable range. A user typing an amount that scales to ~2^64 credits (DASH input around 1.844e8) makes SwiftUI recompute `parsedCredits` during keystrokes, and `UInt64(credits)` traps before the submit-time re-validation can run — crashing the withdrawal sheet on input rather than producing a friendly error. The same shape was copied from `TransferCreditsView.parsedCredits`; the fix here is to switch the upper bound to a strict `<` so 2^64 is rejected, which keeps the result within UInt64's representable range.
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/WithdrawCreditsView.swift:263-267: Submit enables for any non-empty destination — use DashAddress.parse for local pre-validation
  `canSubmit` requires only that `trimmedAddress` be non-empty, so the Withdraw button lights up for obviously malformed or wrong-network strings and the user only learns it was invalid after the round-trip into Rust. The app already has `DashAddress.parse(_, network:)` (Core/Models/DashAddress.swift) which performs Base58Check + network validation via the same Rust FFI used elsewhere, so a local `.core` check would tighten the UI without duplicating consensus logic — authoritative validation still happens in Rust on submit, matching the file header comment.
```
</details>

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/WithdrawCreditsView.swift
QuantumExplorer: Good catch — confirmed: `Double(UInt64.max)` rounds up to 2^64, so `<=` admitted it and `UInt64(credits)` trapped. Fixed with strict `<` in follow-up #3907 (PR #3906 had already merged). Verified on-sim: a huge amount no longer crashes the sheet. (The same shape in TransferCreditsView is flagged for a separate follow-up.)

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/WithdrawCreditsView.swift
QuantumExplorer: Adopted in #3907 — added `isValidDestinationAddress` via `DashAddress.parse(_, network: identity.network)` (`.core` check) gating `canSubmit`/`submit()`, plus an inline 'not a valid <network> address' hint. Authoritative validation still runs in Rust on submit. Verified on-sim: button stays disabled for malformed input.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/WithdrawCreditsView.swift
```diff
@@ -0,0 +1,346 @@
...
+    /// value, round to the nearest credit, and reject any result below
+    /// 1 credit or beyond `UInt64.max`.
+    private var parsedCredits: UInt64? {
+        let trimmed = amountDash.trimmingCharacters(in: .whitespaces)
+        guard let dash = Double(trimmed), dash.isFinite, dash > 0 else {
+            return nil
+        }
+        let credits = (dash * Double(Self.creditsPerDash)).rounded()
+        guard credits >= 1, credits <= Double(UInt64.max) else { return nil }
+        return UInt64(credits)
```
thepastaclaw, line 259: **🔴 Blocking: [Carried forward — STILL VALID] parsedCredits traps on amounts at the UInt64 boundary**

Verified at current head (c4ce64c18b, lines 258–259). `Double(UInt64.max)` is not exactly representable in IEEE-754 binary64 and rounds up to `2^64` (18446744073709551616.0). The guard `credits <= Double(UInt64.max)` therefore admits a value equal to `2^64`, which then traps in `UInt64(credits)` on line 259. Because SwiftUI recomputes `parsedCredits` on every keystroke (it is read from `amountSection` during render), a single typed character around the boundary crashes the withdrawal sheet before any submit-time backend validation can run. Reaching the bucket requires ~1.84×10^8 DASH, which is unrealistic on mainnet but trivially reachable on regtest/devnet and via paste/UI automation — and the failure mode is a hard process abort on a funds-moving screen. The fix is to make the upper bound strict so values that round to `2^64` are rejected. The same pattern lives in `TransferCreditsView`; that file is unchanged by this PR and is captured as an out-of-scope follow-up.

```suggestion
        let credits = (dash * Double(Self.creditsPerDash)).rounded()
        guard credits >= 1, credits < Double(UInt64.max) else { return nil }
        return UInt64(credits)
```

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/WithdrawCreditsView.swift
```diff
@@ -0,0 +1,346 @@
...
+    /// `Int64`; clamp any (unexpected) negative value to zero.
+    private var senderBalanceCredits: UInt64 {
+        identity.balance < 0 ? 0 : UInt64(identity.balance)
+    }
+
+    private var canSubmit: Bool {
+        !trimmedAddress.isEmpty
+            && managedWallet != nil
+            && (parsedCredits.map { $0 > 0 && $0 <= senderBalanceCredits } ?? false)
+    }
```
thepastaclaw, line 272: **🟡 Suggestion: [Carried forward — STILL VALID] Gate canSubmit on DashAddress.parse for local Core-address pre-validation**

Verified at current head (c4ce64c18b, lines 268–272). `canSubmit` requires only `!trimmedAddress.isEmpty`, so the Withdraw button enables for malformed strings, wrong-network addresses, and fragments. Authoritative Base58Check + network validation does happen in Rust at submit time (and the `addressSection` footer advertises this), so this is not a funds-safety gap — it is a UX/defense-in-depth issue: users currently pay the round trip to discover an obviously-bad destination. The app already exposes `DashAddress.parse(_:network:)` (used throughout `SendViewModel`, `SendTransactionView`, and `QRScannerView`), which routes Core addresses through the Rust FFI for Base58Check and network validation. Requiring a parsed `.core` address matching `identity.network` in `canSubmit` (and the `submit()` pre-check) would catch the common cases locally while keeping the FFI submit path authoritative. This mirrors the precedent already established in the Core send flow.

```suggestion
    private var isValidDestinationAddress: Bool {
        if case .core = DashAddress.parse(trimmedAddress, network: identity.network).type {
            return true
        }
        return false
    }

    private var canSubmit: Bool {
        isValidDestinationAddress
            && managedWallet != nil
            && (parsedCredits.map { $0 > 0 && $0 <= senderBalanceCredits } ?? false)
    }
```

thepastaclaw (changes requested):
## Code Review

Cumulative re-review at c4ce64c18b. Latest delta (c4ce64c) adds `.interactiveDismissDisabled(isSubmitting)` to the sheet — a sound fix closing a real re-entry window during submit; no new findings in this delta. Both prior findings remain STILL VALID at the current head and are carried forward: (1) `parsedCredits` admits `2^64` because `Double(UInt64.max)` rounds up, then traps on `UInt64(credits)`; (2) `canSubmit` only requires a non-empty destination instead of using the available `DashAddress.parse(_, network:)` Rust-backed pre-validation.

<details>

```

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/WithdrawCreditsView.swift`:
- [BLOCKING] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/WithdrawCreditsView.swift:258-259: [Carried forward — STILL VALID] parsedCredits traps on amounts at the UInt64 boundary
  Verified at current head (c4ce64c18b, lines 258–259). `Double(UInt64.max)` is not exactly representable in IEEE-754 binary64 and rounds up to `2^64` (18446744073709551616.0). The guard `credits <= Double(UInt64.max)` therefore admits a value equal to `2^64`, which then traps in `UInt64(credits)` on line 259. Because SwiftUI recomputes `parsedCredits` on every keystroke (it is read from `amountSection` during render), a single typed character around the boundary crashes the withdrawal sheet before any submit-time backend validation can run. Reaching the bucket requires ~1.84×10^8 DASH, which is unrealistic on mainnet but trivially reachable on regtest/devnet and via paste/UI automation — and the failure mode is a hard process abort on a funds-moving screen. The fix is to make the upper bound strict so values that round to `2^64` are rejected. The same pattern lives in `TransferCreditsView`; that file is unchanged by this PR and is captured as an out-of-scope follow-up.
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/WithdrawCreditsView.swift:268-272: [Carried forward — STILL VALID] Gate canSubmit on DashAddress.parse for local Core-address pre-validation
  Verified at current head (c4ce64c18b, lines 268–272). `canSubmit` requires only `!trimmedAddress.isEmpty`, so the Withdraw button enables for malformed strings, wrong-network addresses, and fragments. Authoritative Base58Check + network validation does happen in Rust at submit time (and the `addressSection` footer advertises this), so this is not a funds-safety gap — it is a UX/defense-in-depth issue: users currently pay the round trip to discover an obviously-bad destination. The app already exposes `DashAddress.parse(_:network:)` (used throughout `SendViewModel`, `SendTransactionView`, and `QRScannerView`), which routes Core addresses through the Rust FFI for Base58Check and network validation. Requiring a parsed `.core` address matching `identity.network` in `canSubmit` (and the `submit()` pre-check) would catch the common cases locally while keeping the FFI submit path authoritative. This mirrors the precedent already established in the Core send flow.
```
</details>

PR #3907, opened 2026-06-15
fix(swift-example-app): harden Withdraw Credits amount + address validation

QuantumExplorer:

Follow-up to #3906 (Withdraw Credits production UI, ID-10), which merged before review feedback from `thepastaclaw` was addressed. Two issues in `WithdrawCreditsView`:

- 🔴 **Blocking — `parsedCredits` could crash on large input.** `Double(UInt64.max)` isn't exactly representable and rounds **up** to 2^64, so the `credits <= Double(UInt64.max)` guard admitted 2^64 and `UInt64(credits)` trapped. `parsedCredits` recomputes on every keystroke (before submit-time re-validation), so an amount scaling to ~2^64 credits would crash the sheet on input rather than show a friendly error. (Same shape exists in the merged `TransferCreditsView.parsedCredits` — flagged separately as a follow-up.)
- 🟡 **UX — submit enabled for any non-empty destination.** The button lit up for malformed / wrong-network addresses; the user only learned otherwise after the Rust round-trip.

## What was done?

- **`parsedCredits`:** switched the upper bound to strict `<` so 2^64 is rejected and the `UInt64` cast stays in range.
- **Destination validation:** added `isValidDestinationAddress`, which uses the existing `DashAddress.parse(_, network:)` bridge (Base58Check + network validation in Rust) and accepts a `.core` (L1) result. `canSubmit` and the `submit()` guard now require a valid address, and an inline "Not a valid <network> Dash address" hint explains the disabled state. No consensus logic added to Swift — authoritative validation still runs in Rust on submit, matching the file header.

## How Has This Been Tested?

`./build_ios.sh`-equivalent app build succeeds (0 errors, 0 code warnings). Driven on the simulator (devnet):
- Invalid destination → submit stays **disabled** + inline hint shown.
- Amount of 2×10⁸ DASH (huge) → sheet no longer crashes; button stays disabled.

## Breaking Changes

None. UI-only validation hardening in the example app.

## Checklist:

- [x] I have added "!" to the title and described breaking changes in the corresponding section if my code contains any (n/a)
- [x] I have made corresponding changes to the documentation if needed (n/a)

### DIFF
--- a/packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/WithdrawCreditsView.swift
+++ b/packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/WithdrawCreditsView.swift
@@ -133,6 +133,11 @@ struct WithdrawCreditsView: View {
                     .textInputAutocapitalization(.never)
                     .autocorrectionDisabled(true)
                     .disabled(isSubmitting)
+                if !trimmedAddress.isEmpty && !isValidDestinationAddress {
+                    Text("Not a valid \(identity.network.displayName) Dash address.")
+                        .font(.caption)
+                        .foregroundColor(.red)
+                }
             } else {
                 Text("The wallet that owns this identity isn't loaded.")
                     .font(.subheadline)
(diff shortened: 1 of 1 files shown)


thepastaclaw:
✅ Review complete (commit 6ae7f999)

thepastaclaw (approved):
## Code Review

Tight, well-scoped UI hardening on SwiftExampleApp's Withdraw Credits sheet. Strict `<` against `Double(UInt64.max)` correctly avoids the rounding-to-2^64 trap on `UInt64(credits)`, and the new `isValidDestinationAddress` delegates to the existing Rust-backed `DashAddress.parse` bridge while keeping authoritative validation in Rust on submit. Diff is confined to `WithdrawCreditsView.swift`; no FFI, consensus, or business-logic surface area is affected.

PR #3908, opened 2026-06-15
feat(swift-sdk): add production create-document flow via platform-wallet FFI

QuantumExplorer:

Creating (broadcasting) a document in SwiftExampleApp was only reachable via *Settings → Platform State Transitions → Document Create* (test-signer builder, test-plan **DOC-02** 🧪), and the `DocumentsView` "Create Document" path was a **local-only mock** with no broadcast (**DOC-09** ⚠️). There was no real, production create-document flow signed by the wallet's keychain signer.

Tracked in the iOS QA tracker [#3897](https://github.com/dashpay/platform/issues/3897) (DOC-02).

## What was done?

A real, schema-driven **Create Document** flow that broadcasts a `DocumentCreate` through `rs-platform-wallet` using the wallet's keychain-backed signer (per `packages/swift-sdk/CLAUDE.md` — high-level flows live in `platform-wallet`, Swift only renders/marshals/persists).

- **`rs-platform-wallet`** — `IdentityWallet::create_document_with_signer` (new `wallet/identity/network/document.rs`): fetches the on-chain `DataContract`, builds a rev-1 `Document` via `DocumentType::create_document_from_data`, selects an `AUTHENTICATION` + `ECDSA_SECP256K1` key whose security level satisfies the doc type's `security_level_requirement()` (mirrors the consensus rule — **not** a hardcoded `CRITICAL`), and broadcasts via the SDK's `PutDocument::put_to_platform_and_wait_for_response` on the 8 MB worker stack. Includes unit tests for the security-level selection.
- **`rs-platform-wallet-ffi`** — `platform_wallet_create_document_with_signer` C ABI (new `document.rs`), a faithful mirror of `platform_wallet_create_data_contract_with_signer`.
- **`swift-sdk`** — thin `ManagedPlatformWallet.createDocument(...)` wrapper; schema-driven `CreateDocumentView` (owner-identity picker + reused `DocumentFieldsView`) reachable from **Contracts → contract → document type → New Document** (`DocumentTypeDetailsView`). The confirmed document is persisted to SwiftData so it appears in the documents list. The local-only mock in `DocumentsView` was replaced by this real flow. Per-field `accessibilityIdentifier`s added to `DocumentFieldsView` for QA automation.

**Bonus fix (required for the feature to be usable):** the create form self-dismissed within ~1 s. Background-sync SwiftData writes invalidated the Contracts `@Query` list whose `ForEach` rows *owned* the details sheet, tearing down the contract → doc-type → create navigation every second (compounded by deprecated `NavigationView`/`NavigationLink(destination:)` and auto-refreshing relative-time `Text`). Fixed by hoisting the details `.sheet(item:)` to a stable container, migrating to value-based `navigationDestination`, making relative-time strings static (`AppDate.relative`), and decoupling the root `TabView` from SPV `spvProgress` ticks. This also stabilizes other Contracts drill-downs.

## How Has This Been Tested?

- `cargo fmt --all` clean; `cargo check -p platform-wallet -p platform-wallet-ffi` clean; `cargo test -p platform-wallet` security-level unit tests pass.
- `cd packages/swift-sdk && ./build_ios.sh --target sim` → **BUILD SUCCEEDED** (warnings-as-errors).
- **End-to-end on iPhone 17 simulator (testnet):** created a `preorder` document (32-byte `saltedDomainHash`) on the DPNS contract `GWRSAVFMjXx8HpQFaNJMqBV7MBgMK4br5UESsB4S31Ec` from a funded identity. Result: network-confirmed (`put_to_platform_and_wait_for_response` returned Ok — a rejection would surface as "Create failed"), doc id `7i1hJgvVt8fJms26kGwkEZ6jVZxrfd3BrqfmAfpqXMoG`, `ZPERSISTENTDOCUMENT` count `0 → 1`, document appears in the list. After the nav fix, the form stayed presented 18s+ with no interaction (pre-fix: dismissed within ~1s). (`domain` was intentionally avoided as a create target — it's a contested resource that a plain DocumentCreate can't confirm.)
- Updated `TEST_PLAN.md` (DOC-02 🧪 → ✅, DOC-09 retired) and QA issue #3897.

## Breaking Changes

None. SDK/example-app surface only; no consensus-affecting changes (no `!`).

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added platform wallet support for creating and broadcasting revision-1 identity wallet documents with schema-driven properties.
  * Added Swift SDK API `ManagedPlatformWallet.createDocument(...)` returning the created document identifier and canonical JSON.
* **Improvements**
  * Updated example app document creation to a production broadcast flow, including preset document-type support.
  * Refined navigation and sheet behavior across contract and document detail screens; improved global sync indicator rendering.
  * Enhanced document field UX with accessibility identifiers and smarter boolean/hex input handling.
  * Added unit coverage for signing security-level rules.
* **Documentation**
  * Updated QA test plan with a new “Retired” status and refreshed document catalog actions.


thepastaclaw:
✅ Review complete (commit a4690746)

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentFieldsView.swift
```diff
@@ -100,10 +105,12 @@ struct DocumentFieldsView: View {
                         RoundedRectangle(cornerRadius: 8)
                             .stroke(Color.gray.opacity(0.3), lineWidth: 1)
                     )
+                    .accessibilityIdentifier("createDocument.field.\(property.name)")
```
thepastaclaw, line 113: **🟡 Suggestion: Object-typed properties cross the boundary as JSON strings, not objects**

For schema `type: "object"` properties, the form renders a `TextEditor` bound to `textFields` (String), and `updateFieldValues` stores the edited text as a String in `fieldValues`. `DocumentsView.propertiesJSON(from:)` then emits `"field":"{\"foo\":1}"` instead of `"field":{"foo":1}`. On the Rust side, `serde_json::from_value::<BTreeMap<String, Value>>` parses it as `Value::Text`, and `sanitize_document_properties` only dispatches to `property_type.sanitize_value_mut`, which won't reshape a string into a map. The resulting document fails schema validation for any document type with an object-valued property. Parse object-field text into a JSON object before placing it into `fieldValues` so the existing `JSONSerialization` step emits a nested object.

<sub>source: ['codex']</sub>

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentsView.swift
```diff
@@ -156,136 +158,392 @@ struct DocumentDetailView: View {
...
+        Section {
+            DocumentFieldsView(documentType: docType, fieldValues: $fieldValues)
+        } header: {
+            Text("Fields")
+        } footer: {
+            if let required = docType.requiredFields, !required.isEmpty {
+                Text("Required: \(required.joined(separator: ", "))")
+            }
+        }
+    }
```
thepastaclaw, line 357: **🟡 Suggestion: Switching document types leaves stale field values in the payload**

`DocumentFieldsView` holds its field dictionaries in `@State` and only repopulates them in `.onAppear` via `initializeFields()`. When the user changes `selectedDocumentTypeName`, SwiftUI reuses the same `DocumentFieldsView` instance (same position in the tree), so values typed for the previous type stay in `textFields`/`numberFields`/etc. and flow into `fieldValues`. The next submit can broadcast leftover fields, triggering `additionalProperties` validation failures or sending wrong values. Key the child view by the document type id so SwiftUI tears it down and rebuilds fresh state whenever the schema changes.

```suggestion
            DocumentFieldsView(documentType: docType, fieldValues: $fieldValues)
                .id(docType.id)
```

<sub>source: ['codex']</sub>

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentsView.swift
```diff
@@ -156,136 +158,392 @@ struct DocumentDetailView: View {
...
+            network: network
         )
-        // Link to the parent contract so cascading cleanup works.
-        document.dataContract = contract
+        // Link to the parent contract so cascading cleanup works and the
+        // contract-scoped document list picks it up.
+        document.dataContract = parentContract
         modelContext.insert(document)
+        document.linkToLocalIdentityIfNeeded(in: modelContext)
         try? modelContext.save()
```
thepastaclaw: **🟡 Suggestion: PersistentDocument stores the pre-broadcast form JSON, not the canonical confirmed document**

`persistConfirmedDocument` writes the form-built `propertiesJSON` (hex-encoded byte arrays, base58-encoded identifiers) as `PersistentDocument.data`. The confirmed `Document` returned by `put_to_platform_and_wait_for_response` is discarded at the FFI boundary — `platform_wallet_create_document_with_signer` only returns the 32-byte id. For any document type with byteArray or identifier properties (e.g. DPNS `preorder.saltedDomainHash`), the persisted SwiftData row's body structurally diverges from what Platform actually stores, which will mislead later reads and re-sync logic. Either extend the FFI to return the canonical serialized document, or re-fetch the document by id before persisting.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Production create-document flow is structurally sound: Rust `create_document_with_signer` reproduces the consensus security-level rule, the FFI bindings mirror the established platform-wallet signer pattern with correct lifetime/keepalive handling, and the Swift wrapper is appropriately thin. Three in-scope improvements remain in the SwiftExampleApp form/persistence layer; nothing blocks merge.

🟡 3 suggestion(s)

<details>

```

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentFieldsView.swift`:
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentFieldsView.swift:100-108: Object-typed properties cross the boundary as JSON strings, not objects
  For schema `type: "object"` properties, the form renders a `TextEditor` bound to `textFields` (String), and `updateFieldValues` stores the edited text as a String in `fieldValues`. `DocumentsView.propertiesJSON(from:)` then emits `"field":"{\"foo\":1}"` instead of `"field":{"foo":1}`. On the Rust side, `serde_json::from_value::<BTreeMap<String, Value>>` parses it as `Value::Text`, and `sanitize_document_properties` only dispatches to `property_type.sanitize_value_mut`, which won't reshape a string into a map. The resulting document fails schema validation for any document type with an object-valued property. Parse object-field text into a JSON object before placing it into `fieldValues` so the existing `JSONSerialization` step emits a nested object.

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentsView.swift`:
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentsView.swift:329-339: Switching document types leaves stale field values in the payload
  `DocumentFieldsView` holds its field dictionaries in `@State` and only repopulates them in `.onAppear` via `initializeFields()`. When the user changes `selectedDocumentTypeName`, SwiftUI reuses the same `DocumentFieldsView` instance (same position in the tree), so values typed for the previous type stay in `textFields`/`numberFields`/etc. and flow into `fieldValues`. The next submit can broadcast leftover fields, triggering `additionalProperties` validation failures or sending wrong values. Key the child view by the document type id so SwiftUI tears it down and rebuilds fresh state whenever the schema changes.
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentsView.swift:510-525: PersistentDocument stores the pre-broadcast form JSON, not the canonical confirmed document
  `persistConfirmedDocument` writes the form-built `propertiesJSON` (hex-encoded byte arrays, base58-encoded identifiers) as `PersistentDocument.data`. The confirmed `Document` returned by `put_to_platform_and_wait_for_response` is discarded at the FFI boundary — `platform_wallet_create_document_with_signer` only returns the 32-byte id. For any document type with byteArray or identifier properties (e.g. DPNS `preorder.saltedDomainHash`), the persisted SwiftData row's body structurally diverges from what Platform actually stores, which will mislead later reads and re-sync logic. Either extend the FFI to return the canonical serialized document, or re-fetch the document by id before persisting.
```
</details>

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/ManagedPlatformWallet.swift
```diff
@@ -2559,6 +2559,86 @@ extension ManagedPlatformWallet {
...
+                                    typePtr,
+                                    propsPtr,
+                                    signerHandle,
+                                    outBp.baseAddress!
+                                )
+                            }
+                        }
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
QuantumExplorer: Done in 92a5ce4f — wrapped the full marshalling chain in `withExtendedLifetime(signer)` and removed the `_ = signer` discard, matching the other `*_with_signer` wrappers.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentFieldsView.swift
```diff
@@ -65,17 +66,20 @@ struct DocumentFieldsView: View {
...
                     .keyboardType(.numberPad)
                     .textFieldStyle(RoundedBorderTextFieldStyle())
+                    .accessibilityIdentifier("createDocument.field.\(property.name)")
 
             case "boolean":
                 Toggle(isOn: binding(for: property.name, in: $boolFields)) {
                     Text("")
                 }
                 .labelsHidden()
+                .accessibilityIdentifier("createDocument.field.\(property.name)")
```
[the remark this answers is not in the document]
QuantumExplorer: Done in 92a5ce4f — added `.accessibilityLabel(property.name)` to the boolean Toggle.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentsView.swift
```diff
@@ -156,136 +158,392 @@ struct DocumentDetailView: View {
...
-                        }
-                    }
+    // MARK: - Sections
+
+    private var selectionSection: some View {
+        Section("Document") {
+            Picker("Contract", selection: $selectedContract) {
+                Text("Select a contract").tag(nil as PersistentDataContract?)
+                ForEach(contracts) { contract in
+                    Text(contract.name).tag(contract as PersistentDataContract?)
```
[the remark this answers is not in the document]
QuantumExplorer: Done in 92a5ce4f — added `activeContracts` (filtered by `appState.currentNetwork`); the contract picker now iterates it so an owner can't be paired with an off-network contract.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentsView.swift
```diff
@@ -156,136 +158,392 @@ struct DocumentDetailView: View {
...
+        } footer: {
+            Text("The identity that owns and signs for this document. Signing uses this wallet's keychain-backed signer.")
+        }
+    }
+
+    @ViewBuilder
+    private func schemaSection(for docType: PersistentDocumentType) -> some View {
+        Section {
+            DocumentFieldsView(documentType: docType, fieldValues: $fieldValues)
+        } header: {
```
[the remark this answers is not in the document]
QuantumExplorer: Done in 92a5ce4f — added `.id(docType.id)` + `.onChange(of: docType.id) { fieldValues = [:] }` on the fields view, so switching type can't carry over the previous schema's values.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentsView.swift
```diff
@@ -156,136 +158,392 @@ struct DocumentDetailView: View {
...
+            let wallet = managedWallet
+        else {
+            submitError = .init(message: "Select a document type and an owner identity held by a loaded wallet.")
+            return
+        }
+
+        let propertiesJSON: String
+        do {
+            propertiesJSON = try Self.propertiesJSON(from: fieldValues)
+        } catch {
```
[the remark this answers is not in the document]
QuantumExplorer: Done in 92a5ce4f — `propertiesJSON(from:documentType:)` is now schema-aware and parses `object`-typed fields' editor JSON into nested objects (throws on invalid JSON) instead of double-encoding them as strings.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentsView.swift
```diff
@@ -156,136 +158,392 @@ struct DocumentDetailView: View {
...
+                await MainActor.run {
+                    persistConfirmedDocument(
+                        documentId: documentId,
+                        documentType: typeName,
+                        contractId: contractId,
+                        ownerId: ownerId,
+                        propertiesJSON: propertiesJSON,
+                        network: network,
+                        parentContract: parentContract
+                    )
```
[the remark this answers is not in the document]
QuantumExplorer: Done in 92a5ce4f — `persistConfirmedDocument` now `throws` (propagates `modelContext.save()`). On local-save failure we still report the confirmed broadcast (the document is on-chain/queryable) but surface an orange warning instead of swallowing the error.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentFieldsView.swift
QuantumExplorer: Fixed in 92a5ce4f — `propertiesJSON(from:documentType:)` is now schema-aware: `object`-typed fields are parsed from their editor JSON text into nested objects (throwing on invalid JSON) instead of being emitted as a JSON string. (Same finding CodeRabbit raised.)

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentsView.swift
QuantumExplorer: Fixed in 92a5ce4f — keyed the fields view with `.id(docType.id)` (plus an `.onChange(of: docType.id)` that clears `fieldValues`) so SwiftUI rebuilds fresh state and no stale values from a previous type leak into the payload. (Same finding CodeRabbit raised.)

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentsView.swift
QuantumExplorer: Acknowledged — deferring. `PersistentDocument` is a display-only local cache written solely by this flow (the DOC-01 query view doesn't persist), and nothing reads `.data` for re-sync. It stores the submitted properties in their JSON transport encoding (byteArrays as hex, identifiers as base58) — which is actually a more readable display form than the canonical `Value` JSON (which would render bytes as an int array). Canonical state is authoritative on-chain and queryable via DOC-01. Returning the canonical serialized document from the FFI would add a C-string alloc/free across the boundary for no functional gain here, so I'm leaving this open as a possible future enhancement rather than fixing it in this PR.

QuantumExplorer:
### Note on the failing `Rust workspace tests / Tests (macOS)` check

The one failing test — `fetch::document_query_v0_v1::sdk_builder_default_seeds_atomic_to_floor` — is **pre-existing on `v3.1-dev` and unrelated to this PR.**

- This PR's entire Rust diff is the 4 new `platform-wallet` / `platform-wallet-ffi` files; it makes **zero changes to `rs-sdk`** (and no `Cargo.lock`/`*.toml` changes). The branch is identical to the merge-base for all of `rs-sdk`.
- The test asserts `SdkBuilder::new_mock().version().protocol_version == DEFAULT_INITIAL_PROTOCOL_VERSION` (10), but `version()` is now floored to `max(DEFAULT_INITIAL_PROTOCOL_VERSION, network floor)` and the mainnet floor is 11 — so it resolves to 11. This is a semantic conflict between the protocol-version-floor work (#3886/#3893) and the assertion added in #3809; it reproduces locally on a clean checkout of this branch's (== base) `rs-sdk`.
- The fix is already in flight: **#3900 ("refactor(sdk): single protocol-version floor + version_pinned unification")** modifies both `rs-sdk/src/sdk.rs` and this exact test file.

Fixing an unrelated `rs-sdk` test is out of scope for this PR. Once #3900 lands on `v3.1-dev`, rebasing this branch will turn the check green; alternatively this PR can merge ahead of it since the failure is orthogonal to the change here.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentsView.swift
```diff
@@ -156,136 +158,438 @@ struct DocumentDetailView: View {
...
+                            documentType: typeName,
+                            contractId: contractId,
+                            ownerId: ownerId,
+                            propertiesJSON: propertiesJSON,
+                            network: network,
+                            parentContract: parentContract
+                        )
+                    } catch {
+                        self.persistWarning = "Broadcast confirmed, but saving the local copy failed: \(error.localizedDescription). The document is on-chain and queryable."
+                    }
```
thepastaclaw, line 523: **💬 Nitpick: Failed local-cache save leaves an inserted-but-unsaved PersistentDocument in modelContext**

If `try persistConfirmedDocument(...)` throws at the `try modelContext.save()` step, the `modelContext.insert(document)` call inside that function has already attached the row to the context — only the save failed. The next successful `modelContext.save()` from anywhere else (another document create, a delete, any context mutation that triggers save) will silently flush this row too, which contradicts the `persistWarning` message that tells the user the local copy was *not* saved. Either call `modelContext.delete(document)` on the error path or `modelContext.rollback()` before setting `persistWarning`, so behavior matches what the UI reports.

<sub>source: ['claude']</sub>

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentsView.swift
```diff
@@ -156,136 +158,438 @@ struct DocumentDetailView: View {
...
+                    ForEach(documentTypeNames(for: contract), id: \.self) { type in
+                        Text(type)
+                            .tag(type)
+                            .accessibilityIdentifier("createDocument.docType.\(type)")
                     }
+                }
+                .accessibleFormPicker("createDocument.docTypePicker")
+                .disabled(isSubmitting)
+            }
+        }
```
thepastaclaw, line 296: **💬 Nitpick: Changing the contract picker doesn't clear selectedDocumentTypeName**

In the no-preset flow, when the user changes `selectedContract`, `selectedDocumentTypeName` keeps the previous contract's value. `resolvedDocumentType` returns nil safely if the new contract has no type with that name (so the form just hides), but if both contracts happen to share a type name (e.g. "note"), `resolvedDocumentType` silently resolves to the new contract's type. The `.id(docType.id)` reset on `DocumentFieldsView` does clear field state, but the picker shows a stale selection until the user explicitly retaps it. Adding `.onChange(of: selectedContract) { _, _ in selectedDocumentTypeName = "" }` makes the contract change unambiguous.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Cumulative re-review of head 92a5ce4f. Two of three prior findings are FIXED (object-typed properties now parse as nested JSON via the schema-aware `propertiesJSON`; document-type switches reset field state via `.id(docType.id)` + onChange). Prior #3 (local cache stores form JSON, not canonical confirmed document) is STILL VALID and carried forward — multiple agents independently confirmed the FFI returns only the 32-byte id and the Swift side persists the pre-broadcast form JSON. New latest-delta findings: optional booleans are always serialized as `false` (pre-existing in DocumentFieldsView, but now load-bearing because this PR turns this path into the production broadcast flow), plus two low-severity UX/state-hygiene nitpicks. Codex's `blocking` claim that the preview container can broadcast and persist to a throwaway store is not reachable — `.modelContainer(preview.container)` rebinds the `@Query` scope so `ownerIdentities` is empty and `canSubmit` returns false; dropped as a false positive. No in-scope blockers.

<details>

```

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentFieldsView.swift`:
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentFieldsView.swift:158-238: Optional boolean fields are always serialized as `false` in the broadcast payload
  `initializeFields()` seeds every boolean property with `false` (DocumentFieldsView.swift:168), and `updateFieldValues()` unconditionally copies every entry from `boolFields` into the payload (DocumentFieldsView.swift:236-238). String/number/array/object branches all skip empty values, so an untouched optional field of those types is omitted. Booleans have no equivalent skip, so an optional boolean the user never toggled is broadcast as `false` rather than absent. For schemas where absence and `false` are distinct (audit, default-driven, or storage-cost-sensitive document types), this changes the actual document content and storage cost compared to what the user supplied. Pre-PR this code path fed only the Settings test-signer flow, so the divergence was harmless; this PR makes it the production broadcast path. Track which boolean fields the user actually interacted with (e.g. drop the seed and key on user-touched only), or omit non-required booleans whose state hasn't changed from the schema default.
```
</details>

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentsView.swift
thepastaclaw: Correction: this prior finding is **still valid / intentionally deferred**, not resolved in `92a5ce4`.

The cumulative review kept it carried forward: the FFI still returns only the 32-byte document id, and the SwiftData row still stores the submitted/form JSON rather than a canonical confirmed document. The author has explained this is a display-only local cache and is intentionally leaving it as-is for this PR.

<sub>Edited to correct an automated reconciliation mistake; do not treat the earlier “resolved” wording as accurate.</sub>

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentsView.swift
QuantumExplorer: Fixed in ae6ec62 — `persistConfirmedDocument` now wraps `modelContext.save()` in do/catch and calls `modelContext.delete(document)` to detach the inserted row before rethrowing, so a later save elsewhere can't silently flush it (now consistent with the "not saved locally" warning).

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentsView.swift
QuantumExplorer: Fixed in ae6ec62 — added `.onChange(of: selectedContract) { _, _ in selectedDocumentTypeName = "" }` so changing the contract clears the document-type selection rather than carrying over a stale (or same-named) type.

packages/rs-platform-wallet-ffi/src/document.rs
```diff
@@ -0,0 +1,91 @@
...
+                .map(|document| document.id())
+        });
+        result
+    });
+    let result = unwrap_option_or_return!(option);
+    let document_id = unwrap_result_or_return!(result);
+    let bytes = document_id.to_buffer();
+    let dst = slice::from_raw_parts_mut(out_document_id, 32);
+    dst.copy_from_slice(&bytes);
+    PlatformWalletFFIResult::ok()
```
thepastaclaw: **🟡 Suggestion: [Carried forward — STILL VALID] FFI discards canonical confirmed Document; SwiftData cache stores pre-broadcast form JSON instead of Platform's canonical document**

Re-validated at head ae6ec628. The Rust side of the boundary still throws away the confirmed document and only returns its 32-byte id: `create_document_with_signer` returns the confirmed `Document` from `put_to_platform_and_wait_for_response`, but the FFI collapses it to `.map(|document| document.id())` (packages/rs-platform-wallet-ffi/src/document.rs:81) and writes only 32 bytes through `out_document_id` (lines 87-89). Swift therefore has no canonical bytes to persist and writes the user's form-built `propertiesJSON` (hex-encoded byte arrays, base58-encoded identifiers, raw object subtrees) directly into `PersistentDocument.data` with a hardcoded revision of 1 (packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentsView.swift:546-555). Any Rust/DPP-side sanitization, default-filling, byte-level normalization, revision data, or timestamps Platform actually accepted are invisible to Swift — the SwiftData row reads back the user's submitted strings rather than the canonical document. The newly-added `persistWarning` only covers SwiftData save failures, not representation divergence after a successful save. This is owned by this PR because it is the moment the persistence path went from a Settings test-signer flow to the production broadcast path. Not a security/consensus issue (broadcast bytes are produced inside Rust from the sanitized value map), but a documented local-cache fidelity bug. Fix by either (a) extending the FFI to return canonical document bytes (e.g. a serialized `Document`) or the sanitized properties map, or (b) adding a sibling FFI that refetches by id and returns canonical bytes so the cache row reflects what Platform stored.

thepastaclaw (commented):
## Code Review

Cumulative re-review of ae6ec628. Latest delta is small and targeted: it resolves three of the four prior findings (prior-2 boolean payload gating via touchedBoolFields, prior-3 modelContext.delete on save failure, prior-4 onChange clearing the doc-type picker). Carried forward: prior-1 — the FFI still discards the canonical confirmed Document and surfaces only its 32-byte id, so the SwiftData cache row stores pre-broadcast form JSON rather than Platform's canonical document. No new findings in the latest delta. CodeRabbit returned a rate-limited summary with zero inline findings, so no CR reactions are produced.

🟡 1 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet-ffi/src/document.rs:70-90: [Carried forward — STILL VALID] FFI discards canonical confirmed Document; SwiftData cache stores pre-broadcast form JSON instead of Platform's canonical document
  Re-validated at head ae6ec628. The Rust side of the boundary still throws away the confirmed document and only returns its 32-byte id: `create_document_with_signer` returns the confirmed `Document` from `put_to_platform_and_wait_for_response`, but the FFI collapses it to `.map(|document| document.id())` (packages/rs-platform-wallet-ffi/src/document.rs:81) and writes only 32 bytes through `out_document_id` (lines 87-89). Swift therefore has no canonical bytes to persist and writes the user's form-built `propertiesJSON` (hex-encoded byte arrays, base58-encoded identifiers, raw object subtrees) directly into `PersistentDocument.data` with a hardcoded revision of 1 (packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentsView.swift:546-555). Any Rust/DPP-side sanitization, default-filling, byte-level normalization, revision data, or timestamps Platform actually accepted are invisible to Swift — the SwiftData row reads back the user's submitted strings rather than the canonical document. The newly-added `persistWarning` only covers SwiftData save failures, not representation divergence after a successful save. This is owned by this PR because it is the moment the persistence path went from a Settings test-signer flow to the production broadcast path. Not a security/consensus issue (broadcast bytes are produced inside Rust from the sanitized value map), but a documented local-cache fidelity bug. Fix by either (a) extending the FFI to return canonical document bytes (e.g. a serialized `Document`) or the sanitized properties map, or (b) adding a sibling FFI that refetches by id and returns canonical bytes so the cache row reflects what Platform stored.
```
</details>

packages/rs-platform-wallet-ffi/src/document.rs
QuantumExplorer: Fixed in a469074 — `platform_wallet_create_document_with_signer` now also returns the confirmed `Document` serialized to canonical JSON via a new `out_document_json` C string (DPP `to_json_with_identifiers_using_bytes`: base58 `$id`/`$ownerId` + normalized properties; freed Swift-side with `platform_wallet_string_free`). The Swift wrapper returns `(Identifier, String)` and `persistConfirmedDocument` now stores the canonical JSON in `PersistentDocument.data` instead of the form input. Verified on testnet: a `preorder` doc persisted as `{"$id":"AFLkAA…mnZ2","$ownerId":"BjJz…Dx91","saltedDomainHash":[66,31,249,…]}` — canonical id/owner + normalized byteArray, matching a DOC-01 query.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentsView.swift
QuantumExplorer: Update: now **fixed** in a469074 (superseding my earlier deferral). The FFI returns the confirmed document's canonical JSON (`out_document_json`, DPP `to_json_with_identifiers_using_bytes`) and Swift persists that into `PersistentDocument.data` instead of the form input. Verified on testnet — persisted body is now `{"$id":…,"$ownerId":…,"saltedDomainHash":[…]}` (canonical id/owner + normalized byteArray).

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/ManagedPlatformWallet.swift
```diff
@@ -2559,6 +2559,111 @@ extension ManagedPlatformWallet {
...
+
+            try result.check()
+            // Take ownership of the JSON and release the Rust allocation.
+            defer { if let p = documentJsonPtr { platform_wallet_string_free(p) } }
+            // Defensive: on a successful broadcast the Rust side always
+            // writes the canonical JSON, but fall back to an empty
+            // string rather than crashing if it is unexpectedly null.
+            let canonicalJSON = documentJsonPtr.map { String(cString: $0) } ?? ""
+            return (Data(documentIdBytes), canonicalJSON)
```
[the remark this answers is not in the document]
QuantumExplorer: Fixed in ac0a4f2 — replaced the `?? ""` fallback with a guard that throws `PlatformWalletError.walletOperation(...)` when the canonical-JSON pointer is null after `result.check()`, so an FFI/ABI contract violation fails loudly instead of persisting an empty body. (The `defer` free still runs and no-ops on null.)

QuantumExplorer:
Addressed the remaining CodeRabbit items from the last review in ac0a4f2 (the inline null-JSON one is replied/resolved on its thread; the rest were outside-diff / nitpick with no inline thread):

- **Null canonical JSON → throw** (inline): now `throw PlatformWalletError.walletOperation(...)` instead of `?? ""`.
- **DocumentFieldsView byte-array sanitize** (outside-diff): call `updateFieldValues()` after the `.onChange` hex cleaner rewrites the field, so the submitted value reflects the cleaned hex.
- **CreateDocumentView swipe-dismiss** (outside-diff): added `.interactiveDismissDisabled(isSubmitting)` so the sheet can't be dismissed mid-broadcast.
- **Signer lifetime doc comment** (nitpick): updated to describe `withExtendedLifetime(signer)` rather than the unsafe `_ = signer`.

`build_ios.sh --target sim` → BUILD SUCCEEDED.

thepastaclaw (commented):
## Code Review

Cumulative verification at head a4690746. Prior finding prior-ae6ec628-1 is FIXED: the FFI now returns the canonical confirmed document JSON via an out-param and Swift persists that verbatim into PersistentDocument.data (verified in document.rs:96-145, ManagedPlatformWallet.swift:2599-2665, and DocumentsView.swift propertiesJSON / persistConfirmedDocument). No new blocking issues in the latest delta. Three new low-severity findings on the latest diff: a near-impossible CString::new failure that would silently drop a confirmed broadcast (nitpick), a defensive empty-string fallback in Swift that masks contract drift (nitpick), and a UI-layer limitation where non-byte arrays are always broadcast as string arrays regardless of schema item type (suggestion).

<details>

```

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentFieldsView.swift`:
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DocumentFieldsView.swift:266-272: Non-byte array fields are broadcast as `[String]` regardless of schema item type
  The new production create flow builds `propertiesJSON` from whatever `DocumentFieldsView.fieldValues` emits, and for `array` fields that are not detected as byte arrays the value is always a comma-split `[String]` (line 269). `CreateDocumentView.propertiesJSON` then forwards those strings unchanged into the JSON payload. For any document type whose `items` schema is `integer`, `number`, `boolean`, `object`, or an identifier shape not classified as a byte array, the broadcast carries quoted strings where DPP expects typed elements; DPP's sanitize step does not coerce `"1"` into an integer array element, so otherwise-valid document types become uncreatable through this new production path. Parse array entries per the schema's `items.type` (or fall back to a raw JSON array editor for non-byte arrays so the submitted JSON preserves item types).
```
</details>

packages/rs-platform-wallet-ffi/src/document.rs
```diff
@@ -0,0 +1,146 @@
...
+    let json_cstring = unwrap_result_or_return!(CString::new(document_json));
+
+    let bytes = document_id.to_buffer();
+    let dst = slice::from_raw_parts_mut(out_document_id, 32);
+    dst.copy_from_slice(&bytes);
+    // Transfer ownership of the JSON to the caller (freed via
+    // `platform_wallet_string_free`). Written last so the id out-param
+    // and the JSON are populated together on the success path.
+    *out_document_json = json_cstring.into_raw();
+    PlatformWalletFFIResult::ok()
```
thepastaclaw, line 145: **💬 Nitpick: Successful broadcast is reported as an error if `CString::new` sees an interior NUL; document id is never written**

By the time `CString::new(document_json)` (line 136) runs, the state transition has already been broadcast and the confirmed document returned from Platform. If `CString::new` were to fail (only possible if the serialized JSON contains an interior NUL byte — a `serde_json::to_string` output bug, effectively never), `unwrap_result_or_return!` exits the FFI with an error and the 32-byte id copy on lines 138-140 never runs, so Swift sees a failed result for an on-chain success and has no id to recover from. Reorder so the id is written before the JSON allocation (or write the id even on JSON marshalling failure with a distinct error code), keeping the success/error signal honest in that degenerate case. Not blocking.

<sub>source: ['claude']</sub>

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/ManagedPlatformWallet.swift
thepastaclaw, line 2663: **💬 Nitpick: Silent empty-string fallback masks a future regression in the Rust JSON contract**

After `try result.check()` confirms success, the wrapper returns `documentJsonPtr.map { String(cString: $0) } ?? ""`. The documented Rust post-condition is that `*out_document_json` is non-null on success, so substituting an empty string when it is unexpectedly null hides any future drift in that contract — `persistConfirmedDocument` would then persist an empty `dataBlob` into SwiftData without surfacing anything via the existing `persistWarning` channel. Surface the null case explicitly (precondition, throw, or warning) so a contract violation is caught instead of silently degrading local-cache fidelity. Not blocking; today's Rust side does guarantee the post-condition.

<sub>source: ['claude']</sub>

PR #3910, opened 2026-06-15
feat(contract): on-chain QA framework storage layer (testCase + testRun)

QuantumExplorer:

QA status for the iOS app currently lives only in `packages/swift-sdk/SwiftExampleApp/TEST_PLAN.md` and GitHub issue #3897. This adds a **storage layer** so QA test definitions and test-run results live **on-chain on Dash Platform testnet** — queryable, proof-verifiable, and renderable by a website (the dashboard site is a separate task that consumes the contract id below).

## What was done?

New `qa-contract/` directory: a dedicated data contract plus `@dashevo/evo-sdk` (js-evo-sdk) tooling.

**Schema** (`qa-contract/schema/qa-contract.documents.json`) — two document types:
- `testCase` — a test definition mirroring one TEST_PLAN §4 row. Mutable; unique `testId`; indices `testId` / `tier` / `category`; `additionalProperties:false`.
- `testRun` — an append-only run record. **Immutable + non-deletable** (audit log); indices `testIdCreatedAt` / `resultCreatedAt` / `buildRef`; `$createdAt` = platform-stamped run time; `additionalProperties:false`.
- v1 ownership: a single QA identity owns/creates all docs (`creationRestrictionMode: OwnerOnly`); README documents relaxing `testRun` to allow per-team-member submission later.

**Scripts** (`qa-contract/src/`):
- `register.mjs` — register the contract, write `contract-id.testnet.json`; re-runnable (re-registers when the committed id no longer resolves, e.g. after a testnet reset).
- `seed.mjs` — parse the TEST_PLAN §4 catalog and upsert one `testCase` per row; idempotent (skips existing, `--update` replaces changed).
- `submit-run.mjs` — create a `testRun` from `(testId, result, buildRef, network, device, evidence, notes, blockerReason)`.
- `query.mjs` — read back and a self-check that exercises every index (`--proof` for proof-verified reads).
- `derive-identity-key.mjs` — recover the signing key from a wallet mnemonic (see below).
- `README.md` documents schema, QA identity setup, register/seed/submit, the contract id, and the testnet-reset re-seed procedure. `.env` (mnemonic + key) is gitignored.

**Two Platform constraints baked in:**
- Indexed string properties are capped at `maxLength <= 63`.
- Index sort direction must be `asc` in the contract definition — `drive-abci` rejects `desc` (`JsonSchemaError: "desc" is not one of ["asc"]`), and local `new DataContract({fullValidation:true})` does **not** catch this. Newest-first ordering is done at query time via `orderBy [['$createdAt','desc']]` (reverse traversal).

## How Has This Been Tested?

End-to-end on **public testnet**:
- Registered the contract → **`2qEVUbg4znNgNRs3FJQ4kof4NKpB8q4fGtYa7qBouLzw`** (owner QA identity `85KjYZLZXA7YZBPyFEjiMaH36xcQpBBZisKGBHF3uKuH`).
- Seeded **all 126** `testCase` documents from `TEST_PLAN.md` (per-category counts match the plan exactly); re-runs skip existing (idempotent).
- Submitted a sample `testRun` (CORE-05 = pass); `$createdAt` correctly platform-stamped.
- Queried back through **all 6 indices**, including `orderBy $createdAt desc` and proof-verified reads.

The QA identity was minted via SwiftExampleApp's Core asset-lock flow (js-evo-sdk can't produce an asset-lock proof on its own); the HIGH AUTHENTICATION signing key was re-derived from the wallet mnemonic and confirmed against the on-chain public keys.

## Breaking Changes

None. Additive — a new top-level directory and a new data contract; no changes to existing packages or consensus code.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added a QA contract package with CLIs to register contracts, seed test cases, submit immutable append-only test runs, and query results (with index self-check when type is omitted).
  * Introduced networked contract configuration with a testnet preset, plus schemas and canonical code mappings for apps, tiers, and categories.
  * Added a CLI to derive (and optionally write) a QA identity signing key from a mnemonic.
* **Documentation**
  * Added a detailed README describing the on-chain schema and end-to-end workflows.
* **Chores**
  * Added environment templates, git ignore rules, and a test-plan parser to generate catalog rows.


thepastaclaw:
✅ Review complete (commit fd469b74)

qa-contract/src/query.mjs
```diff
@@ -0,0 +1,130 @@
...
+  // testRun.resultCreatedAt
+  docs = await run(sdk, {
+    dataContractId: contractId, documentTypeName: 'testRun',
+    where: [['result', '==', 'pass']], orderBy: [['$createdAt', 'desc']], limit,
+  }, proof);
+  printDocs("testRun index 'resultCreatedAt'  where result == pass order $createdAt desc", docs,
+    ['testId', 'result', 'buildRef']);
+
+  console.log('\n✅ All indexed queries returned without error — indices are valid.');
+}
```
thepastaclaw: **🟡 Suggestion: selfCheck does not exercise the `testRun.buildRef` index**

The schema defines six indices (`testCase.testId`, `tier`, `category`; `testRun.testIdCreatedAt`, `resultCreatedAt`, `buildRef`), but `selfCheck` only queries the first five and then prints '✅ All indexed queries returned without error — indices are valid.' The README and PR description both advertise that `query.mjs` exercises every index, so a regression that breaks or drops the `buildRef` index would not be caught by the no-arg self-check. Add a sixth query filtering on `buildRef` so the verification actually matches the claim.

qa-contract/schema/qa-contract.documents.json
```diff
@@ -0,0 +1,214 @@
...
+        "position": 0,
+        "description": "Test identifier this run is a result for (matches testCase.testId)."
+      },
+      "result": {
+        "type": "string",
+        "minLength": 1,
+        "maxLength": 16,
+        "position": 1,
+        "description": "Outcome. One of: pass, fail, blocked, skipped."
+      },
```
thepastaclaw, line 238: **🟡 Suggestion: `testRun.result` allowed values enforced only client-side**

`result` is documented as one of `pass | fail | blocked | skipped` and the CLI rejects anything else, but on-chain the schema only enforces `minLength: 1, maxLength: 16`. `testRun` documents are immutable and non-deletable, and the README plans to relax `creationRestrictionMode` to 0 in v2 so any identity can submit runs. At that point any out-of-set string would land permanently and silently break consumers of the `resultCreatedAt` index (e.g. `where result == 'pass'`). Add an `enum: ["pass", "fail", "blocked", "skipped"]` to the schema so platform itself rejects bad values.

<sub>source: ['claude']</sub>

qa-contract/src/submit-run.mjs
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  const { sdk, mod, network } = await connect();
+  const cfg = readConfig(network);
+  if (!cfg?.contractId) throw new Error(`No contract registered for ${network}. Run register.mjs first.`);
+  const contractId = cfg.contractId;
+
+  const { ownerId, signer, identityKey } = await loadOwnerAuth(sdk, mod, network);
+
+  const properties = { testId, result, network: values.network?.trim() || getNetwork(), buildRef };
```
thepastaclaw: **🟡 Suggestion: `--network` flag is recorded in the document but does not select the target network**

`connect()` derives the target network from `getNetwork()` / `NETWORK` env (qa-contract/src/sdk.mjs), and `readConfig(network)` then loads the contract for whichever network was used to connect. But line 51 stamps `properties.network` from `values.network` (the CLI flag) with a fallback to `getNetwork()`. If a user passes `--network testnet` while `NETWORK=devnet`, the run is submitted into the devnet contract but recorded with `network: 'testnet'`. Either drop the CLI flag and always stamp `network` from the connected network, or apply `values.network` before `connect()` so the connection, the loaded config, and the stamped property all agree.

qa-contract/src/query.mjs
```diff
@@ -0,0 +1,130 @@
...
+      tier: { type: 'string' },
+      category: { type: 'string' },
+      result: { type: 'string' },
+      buildRef: { type: 'string' },
+      limit: { type: 'string', default: '50' },
+      proof: { type: 'boolean', default: false },
+      json: { type: 'boolean', default: false },
+    },
+  });
+  const limit = Number(values.limit);
```
thepastaclaw, line 107: **💬 Nitpick: `--limit` is not validated before use**

`Number(values.limit)` silently accepts `NaN`, `0`, negative numbers, and floats and passes them downstream into both `selfCheck` and the ad-hoc query. The same shape appears in `seed.mjs` at line 79 via `rows.slice(0, Number(values.limit))`, where a `NaN` slice yields zero rows and a negative slice is interpreted from the end. Add a `Number.isInteger(limit) && limit > 0` guard at the CLI boundary in both scripts so misuse fails fast.

qa-contract/README.md
```diff
@@ -0,0 +1,246 @@
...
+│   ├── sdk.mjs                          # SDK load, connect, signer, identity-key, config
+│   ├── parse-test-plan.mjs              # TEST_PLAN.md §4 catalog parser
+│   ├── register.mjs                     # register the contract
+│   ├── seed.mjs                         # seed testCases from the plan (idempotent)
+│   ├── submit-run.mjs                   # create one testRun
+│   ├── query.mjs                        # read back + verify indices
+│   └── derive-identity-key.mjs          # recover signing key from a wallet mnemonic
+├── .env.example
+└── README.md
+```
```
thepastaclaw: **💬 Nitpick: Add a language tag to the file-tree fenced block (MD040)**

The file-tree code block has no language identifier, which trips markdownlint MD040. Tag it `text` (or similar) to clear the lint.

thepastaclaw (commented):
## Code Review

Purely additive new `qa-contract/` package: a JSON schema for `testCase`/`testRun` plus Node CLI scripts. No consensus, FFI, or platform-version touch points. Two in-scope suggestions stand: the advertised index self-check skips the `buildRef` index, and `testRun.result` is not constrained to its documented enum at the schema level. A few minor CR nits are valid; the WIF-print finding is by-design for a testnet-only recovery tool.

<details>

```

In `qa-contract/src/query.mjs`:
- [SUGGESTION] qa-contract/src/query.mjs:37-78: selfCheck does not exercise the `testRun.buildRef` index
  The schema defines six indices (`testCase.testId`, `tier`, `category`; `testRun.testIdCreatedAt`, `resultCreatedAt`, `buildRef`), but `selfCheck` only queries the first five and then prints '✅ All indexed queries returned without error — indices are valid.' The README and PR description both advertise that `query.mjs` exercises every index, so a regression that breaks or drops the `buildRef` index would not be caught by the no-arg self-check. Add a sixth query filtering on `buildRef` so the verification actually matches the claim.

In `qa-contract/schema/qa-contract.documents.json`:
- [SUGGESTION] qa-contract/schema/qa-contract.documents.json:158-164: `testRun.result` allowed values enforced only client-side
  `result` is documented as one of `pass | fail | blocked | skipped` and the CLI rejects anything else, but on-chain the schema only enforces `minLength: 1, maxLength: 16`. `testRun` documents are immutable and non-deletable, and the README plans to relax `creationRestrictionMode` to 0 in v2 so any identity can submit runs. At that point any out-of-set string would land permanently and silently break consumers of the `resultCreatedAt` index (e.g. `where result == 'pass'`). Add an `enum: ["pass", "fail", "blocked", "skipped"]` to the schema so platform itself rejects bad values.

In `qa-contract/src/submit-run.mjs`:
- [SUGGESTION] qa-contract/src/submit-run.mjs:44-51: `--network` flag is recorded in the document but does not select the target network
  `connect()` derives the target network from `getNetwork()` / `NETWORK` env (qa-contract/src/sdk.mjs), and `readConfig(network)` then loads the contract for whichever network was used to connect. But line 51 stamps `properties.network` from `values.network` (the CLI flag) with a fallback to `getNetwork()`. If a user passes `--network testnet` while `NETWORK=devnet`, the run is submitted into the devnet contract but recorded with `network: 'testnet'`. Either drop the CLI flag and always stamp `network` from the connected network, or apply `values.network` before `connect()` so the connection, the loaded config, and the stamped property all agree.
```
</details>

qa-contract/src/derive-identity-key.mjs
```diff
@@ -0,0 +1,103 @@
...
+  const signable = matches
+    .filter((m) => m.purpose === 'AUTHENTICATION' && !m.readOnly && (m.securityLevel === 'HIGH' || m.securityLevel === 'CRITICAL'))
+    .sort((a, b) => SEC_RANK[a.securityLevel] - SEC_RANK[b.securityLevel] || a.id - b.id);
+  // Prefer HIGH (matches the platform doc/contract transition precedent), else CRITICAL.
+  const pick = signable.find((m) => m.securityLevel === 'HIGH') || signable[0];
+  if (!pick) throw new Error('No writable HIGH/CRITICAL AUTHENTICATION key matched.');
+
+  console.log(`\nSigning key: id=${pick.id} ${pick.securityLevel} AUTHENTICATION`);
+  console.log(`  WIF: ${pick.wif}`);
+
```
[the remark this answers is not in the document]
thepastaclaw: This script's sole purpose is to recover and surface a testnet-only signing key from a user-supplied mnemonic; printing the WIF is the requested output, and both the README and .env.example mark these as TESTNET-ONLY keys. Masking by default would defeat the tool. Worth revisiting only if this utility is ever reused for mainnet identities.

qa-contract/src/query.mjs
```diff
@@ -0,0 +1,130 @@
...
+
+  const where = []; const orderBy = [];
+  if (values.type === 'testCase') {
+    if (values.testId) where.push(['testId', '==', values.testId]);
+    if (values.tier) { where.push(['tier', '==', values.tier]); orderBy.push(['tier', 'asc']); }
+    if (values.category) { where.push(['category', '==', values.category]); orderBy.push(['category', 'asc']); }
+  } else if (values.type === 'testRun') {
+    if (values.testId) { where.push(['testId', '==', values.testId]); orderBy.push(['$createdAt', 'desc']); }
+    if (values.result) { where.push(['result', '==', values.result]); orderBy.push(['$createdAt', 'desc']); }
+    if (values.buildRef) where.push(['buildRef', '==', values.buildRef]);
```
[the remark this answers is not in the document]
thepastaclaw: The CLI is a small operator tool — the SDK will surface a clear error if an unsupported composite/index combination is requested at runtime. Adding a hand-maintained compatibility matrix here is more code than it saves, and would need to be kept in sync with any future index additions.

qa-contract/README.md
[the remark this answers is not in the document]
QuantumExplorer: Fixed in 563610407d — tagged the file-tree fence as ```text.

qa-contract/src/query.mjs
```diff
@@ -0,0 +1,130 @@
...
+      category: { type: 'string' },
+      result: { type: 'string' },
+      buildRef: { type: 'string' },
+      limit: { type: 'string', default: '50' },
+      proof: { type: 'boolean', default: false },
+      json: { type: 'boolean', default: false },
+    },
+  });
+  const limit = Number(values.limit);
+
```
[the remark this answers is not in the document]
QuantumExplorer: Fixed in 563610407d — query.mjs now throws if --limit isn't a positive integer.

qa-contract/src/seed.mjs
```diff
@@ -0,0 +1,127 @@
...
+  const planCommit = resolvePlanCommit(planPath);
+  let rows = parseTestPlan(planPath, planCommit);
+
+  const idFilter = csv(values.ids);
+  const tierFilter = csv(values.tier)?.map((s) => s.toLowerCase());
+  const catFilter = csv(values.category)?.map((s) => s.toLowerCase());
+  if (idFilter) rows = rows.filter((r) => idFilter.includes(r.testId));
+  if (tierFilter) rows = rows.filter((r) => tierFilter.includes(r.tier.toLowerCase()));
+  if (catFilter) rows = rows.filter((r) => catFilter.includes(r.category.toLowerCase()));
+  if (values.limit) rows = rows.slice(0, Number(values.limit));
```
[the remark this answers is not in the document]
QuantumExplorer: Fixed in 563610407d — seed.mjs validates --limit as a positive integer before slicing.

qa-contract/src/submit-run.mjs
```diff
@@ -0,0 +1,71 @@
...
+  if (values.device) properties.device = values.device;
```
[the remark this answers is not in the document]
QuantumExplorer: Fixed in 563610407d — --network now sets the target NETWORK before connect(), so the connection, the loaded contract config, and the stamped properties.network all agree.

qa-contract/src/query.mjs
QuantumExplorer: Fixed in 563610407d — added a sixth self-check query on testRun.buildRef, so the no-arg selfCheck now exercises all six indices.

qa-contract/src/submit-run.mjs
QuantumExplorer: Fixed in 563610407d — --network is applied before connect() so the target network and the recorded network agree.

qa-contract/src/query.mjs
QuantumExplorer: Fixed in 563610407d — both query.mjs and seed.mjs now guard --limit as a positive integer.

qa-contract/README.md
QuantumExplorer: Fixed in 563610407d — tagged the file-tree fence as ```text.

qa-contract/schema/qa-contract.documents.json
QuantumExplorer: Good call, but deferring: the live contract `2qEVUbg4znNgNRs3FJQ4kof4NKpB8q4fGtYa7qBouLzw` is already pinned by the deployed dashboard site, and a data contract's schema is immutable — adding the enum means re-registering, which changes the contract id and breaks that pin. For v1 the only writer is submit-run.mjs, which already rejects anything outside pass|fail|blocked|skipped. The on-chain `enum: [pass,fail,blocked,skipped]` belongs with the v2 change that relaxes creationRestrictionMode to open submissions to other identities — that requires a fresh contract anyway, so the enum lands there. Tracking it for that re-register.

qa-contract/src/seed.mjs
```diff
@@ -0,0 +1,133 @@
...
+  const props = {};
+  for (const f of CONTENT_FIELDS) {
+    if (row[f] !== undefined && row[f] !== null && row[f] !== '') props[f] = row[f];
+  }
+  return props;
+}
+
+function contentEquals(existing, props) {
+  return CONTENT_FIELDS.every((f) => (existing?.[f] ?? undefined) === (props[f] ?? undefined));
+}
```
thepastaclaw: **🟡 Suggestion: `--update` can drop a previously stamped `planCommit`**

`cleanProps` strips empty/undefined fields and `contentEquals` compares `(existing[f] ?? undefined) === (props[f] ?? undefined)`. When `resolvePlanCommit` returns undefined (git unavailable, detached run, etc.), the new row's `planCommit` is missing, `contentEquals` reports a mismatch, and `sdk.documents.replace` is called with properties that omit `planCommit`. Since the contract has `additionalProperties: false` and `planCommit` isn't `required`, the replace succeeds and the previously seeded SHA is silently dropped from the document. Either skip field-wise comparison when `props[f]` is undefined but `existing[f]` is set, or carry the existing `planCommit` forward when no new one is resolved so re-running `--update` from a context without git history doesn't erase provenance.

<sub>source: ['claude']</sub>

qa-contract/src/submit-run.mjs
```diff
@@ -0,0 +1,75 @@
...
+  if (values.network?.trim()) process.env.NETWORK = values.network.trim();
+
+  const { sdk, mod, network } = await connect();
+  const cfg = readConfig(network);
+  if (!cfg?.contractId) throw new Error(`No contract registered for ${network}. Run register.mjs first.`);
+  const contractId = cfg.contractId;
+
+  const { ownerId, signer, identityKey } = await loadOwnerAuth(sdk, mod, network);
+
+  const properties = { testId, result, network, buildRef };
```
thepastaclaw: **🟡 Suggestion: Validate `--testId` against existing testCases before writing an immutable run**

`testRun` is intentionally immutable and non-deletable. `submit-run.mjs` only checks that `--testId` is non-empty before broadcasting, so an operator typo (`COR-05` vs `CORE-05`) permanently creates an orphan run record that can never join back to a `testCase`. The on-chain audit log then carries garbage rows that consume credits and pollute future dashboard joins. Since `testCase.testId` is a unique index, a single `documents.query` before `documents.create` is cheap and turns this into an immediate CLI error rather than a permanent record. Risk is bounded today because v1 is OwnerOnly, but the README plans to open `testRun` creation to per-team-member identities, which compounds the problem.

<sub>source: ['codex']</sub>

qa-contract/schema/qa-contract.documents.json
thepastaclaw, line 238: **🟡 Suggestion: Constrain `testRun.result` to the documented enum in the contract schema**

`submit-run.mjs` enforces `pass | fail | blocked | skipped`, the schema description documents the same set, and the README treats it as the canonical vocabulary, but the on-chain property only declares `minLength: 1, maxLength: 16`. Because `testRun` is immutable, any direct SDK caller (today the owner, tomorrow any identity if `creationRestrictionMode` is relaxed to 0 per the v2 plan) can permanently store `passed`/`PASS`/`unknown` rows that break every dashboard `result` filter. Putting `enum: ["pass", "fail", "blocked", "skipped"]` in the schema makes the on-chain storage layer self-enforcing, which is the whole point of the contract. Carried forward from prior review: still valid, not addressed in this delta.

qa-contract/src/derive-identity-key.mjs
```diff
@@ -0,0 +1,103 @@
...
+      const re = new RegExp(`^${key}=.*$`, 'm');
+      env = re.test(env) ? env.replace(re, `${key}=${val}`) : `${env}\n${key}=${val}`;
+    };
+    setLine('QA_PRIVATE_KEY', pick.wif);
+    if (Number.isFinite(pick.id)) setLine('QA_IDENTITY_KEY_ID', String(pick.id));
+    writeFileSync(envPath, env.endsWith('\n') ? env : `${env}\n`);
+    console.log(`\nWrote QA_PRIVATE_KEY${Number.isFinite(pick.id) ? ' + QA_IDENTITY_KEY_ID' : ''} to ${envPath}`);
+  } else {
+    console.log('\nRe-run with --write to save QA_PRIVATE_KEY + QA_IDENTITY_KEY_ID into .env.');
+  }
```
thepastaclaw, line 114: **🟡 Suggestion: Recovered signing key is logged to stdout and written to `.env` without restrictive permissions**

`derive-identity-key.mjs` always prints the recovered WIF to stdout (line 85), and `--write` calls `writeFileSync(envPath, ...)` with the process default umask (line 96), leaving `.env` world-readable on many systems. On shared dev workstations or CI runners, the WIF can be captured via terminal scrollback, shell history, journald, or another local user reading the file. The README scopes this to testnet, but `NETWORK=mainnet`/`local` are also supported by the tooling, and the consequence is full control of the QA identity (arbitrary `testCase` updates, immutable `testRun` injection, credit drain). Defense-in-depth: gate the stdout print behind a `--print` flag and `chmod 0o600` the env file after writing.

thepastaclaw (commented):
## Code Review

Incremental review of qa-contract storage-layer PR. The latest delta resolves four of five prior findings (buildRef self-check, --network propagation, --limit validation, README MD040 fence). The remaining prior finding (testRun.result enum) is intentionally deferred but valid given immutable records. Three new in-scope suggestions surface around immutable-record data integrity and a recovered-private-key handling defense-in-depth gap. No blocking issues for a v1 OwnerOnly storage layer.

🟡 4 suggestion(s)

<details>

```

In `qa-contract/src/seed.mjs`:
- [SUGGESTION] qa-contract/src/seed.mjs:31-33: `--update` can drop a previously stamped `planCommit`
  `cleanProps` strips empty/undefined fields and `contentEquals` compares `(existing[f] ?? undefined) === (props[f] ?? undefined)`. When `resolvePlanCommit` returns undefined (git unavailable, detached run, etc.), the new row's `planCommit` is missing, `contentEquals` reports a mismatch, and `sdk.documents.replace` is called with properties that omit `planCommit`. Since the contract has `additionalProperties: false` and `planCommit` isn't `required`, the replace succeeds and the previously seeded SHA is silently dropped from the document. Either skip field-wise comparison when `props[f]` is undefined but `existing[f]` is set, or carry the existing `planCommit` forward when no new one is resolved so re-running `--update` from a context without git history doesn't erase provenance.

In `qa-contract/src/submit-run.mjs`:
- [SUGGESTION] qa-contract/src/submit-run.mjs:37-55: Validate `--testId` against existing testCases before writing an immutable run
  `testRun` is intentionally immutable and non-deletable. `submit-run.mjs` only checks that `--testId` is non-empty before broadcasting, so an operator typo (`COR-05` vs `CORE-05`) permanently creates an orphan run record that can never join back to a `testCase`. The on-chain audit log then carries garbage rows that consume credits and pollute future dashboard joins. Since `testCase.testId` is a unique index, a single `documents.query` before `documents.create` is cheap and turns this into an immediate CLI error rather than a permanent record. Risk is bounded today because v1 is OwnerOnly, but the README plans to open `testRun` creation to per-team-member identities, which compounds the problem.

In `qa-contract/schema/qa-contract.documents.json`:
- [SUGGESTION] qa-contract/schema/qa-contract.documents.json:158-164: Constrain `testRun.result` to the documented enum in the contract schema
  `submit-run.mjs` enforces `pass | fail | blocked | skipped`, the schema description documents the same set, and the README treats it as the canonical vocabulary, but the on-chain property only declares `minLength: 1, maxLength: 16`. Because `testRun` is immutable, any direct SDK caller (today the owner, tomorrow any identity if `creationRestrictionMode` is relaxed to 0 per the v2 plan) can permanently store `passed`/`PASS`/`unknown` rows that break every dashboard `result` filter. Putting `enum: ["pass", "fail", "blocked", "skipped"]` in the schema makes the on-chain storage layer self-enforcing, which is the whole point of the contract. Carried forward from prior review: still valid, not addressed in this delta.

In `qa-contract/src/derive-identity-key.mjs`:
- [SUGGESTION] qa-contract/src/derive-identity-key.mjs:84-100: Recovered signing key is logged to stdout and written to `.env` without restrictive permissions
  `derive-identity-key.mjs` always prints the recovered WIF to stdout (line 85), and `--write` calls `writeFileSync(envPath, ...)` with the process default umask (line 96), leaving `.env` world-readable on many systems. On shared dev workstations or CI runners, the WIF can be captured via terminal scrollback, shell history, journald, or another local user reading the file. The README scopes this to testnet, but `NETWORK=mainnet`/`local` are also supported by the tooling, and the consequence is full control of the QA identity (arbitrary `testCase` updates, immutable `testRun` injection, credit drain). Defense-in-depth: gate the stdout print behind a `--print` flag and `chmod 0o600` the env file after writing.
```
</details>

qa-contract/src/seed.mjs
```diff
@@ -0,0 +1,133 @@
...
+        if (contentEquals(existingJson, props)) { skipped += 1; continue; }
+        const doc = new Document({
+          id: String(existingJson.$id),
+          ownerId,
+          dataContractId: contractId,
+          documentTypeName: 'testCase',
+          properties: props,
+          revision: BigInt(existingJson.$revision ?? 1) + 1n,
+        });
+        await sdk.documents.replace({ document: doc, identityKey, signer });
```
thepastaclaw, line 182: **🟡 Suggestion: `--update` can erase an existing `planCommit`**

`cleanProps` drops fields whose value is undefined/null/'', and `contentEquals` then compares CONTENT_FIELDS with `?? undefined`. When `resolvePlanCommit()` returns undefined (git unavailable, detached run, CI shallow clone, etc.), the new row's `planCommit` is absent, `contentEquals` reports a mismatch against a previously seeded row, and `sdk.documents.replace({ document: doc, ... })` writes `properties: props` without `planCommit` — silently deleting the audit-log provenance the schema/README intend to preserve. Before replacing, carry forward the existing `planCommit` if the current run cannot resolve a new one (e.g., `if (props.planCommit === undefined && existingJson.planCommit) props.planCommit = existingJson.planCommit;`).

qa-contract/schema/qa-contract.documents.json
```diff
@@ -0,0 +1,256 @@
...
+      "network": {
+        "type": "integer",
+        "minimum": 0,
+        "position": 2,
+        "description": "Network id the run executed against: 0=mainnet, 1=testnet, 2=devnet, 3=regtest."
+      },
```
thepastaclaw, line 245: **🟡 Suggestion: Enforce documented testRun enums in the contract schema**

The README and the schema's own description declare `result` as `pass | fail | blocked | skipped` and `network` as `0=mainnet, 1=testnet, 2=devnet, 3=regtest`, but the on-chain schema only constrains `result` by length and `network` by `minimum: 0`. Because `testRun` is immutable and non-deletable, any non-CLI SDK caller today — and any submitter after the documented v2 relaxation of `creationRestrictionMode` to `0` — can permanently store `PASS`, `unknown`, or `999`, breaking result/network filters with rows that can never be corrected. The contract is the trust boundary; add `enum` constraints (`pass`/`fail`/`blocked`/`skipped`, and `0..3` for `network`) so Platform rejects out-of-vocabulary audit-log values before they become non-deletable records.

qa-contract/schema/qa-contract.documents.json
```diff
@@ -0,0 +1,256 @@
...
+            "testId": "asc"
+          },
+          {
+            "network": "asc"
+          },
+          {
+            "$createdAt": "asc"
+          }
+        ]
+      },
```
thepastaclaw, line 203: **🟡 Suggestion: `ownerTestNetwork` is a strict prefix of `ownerTestNetworkCreated` and pays storage cost on every immutable testRun**

`ownerTestNetwork` indexes `[$ownerId, testId, network]` and `ownerTestNetworkCreated` indexes `[$ownerId, testId, network, $createdAt]`. Every equality query the shorter index can serve (e.g., `$ownerId==X AND testId==Y AND network==Z`) is also served by the longer index — with or without an `orderBy('$createdAt')`. Because `testRun` documents are immutable and non-deletable, every submitted run pays storage cost in both index trees forever; on long-running QA streams this compounds without serving any unique query shape. Drop `ownerTestNetwork` and let `ownerTestNetworkCreated` cover those queries. (`ownerTestCreated` is genuinely needed — different middle column.)

<sub>source: ['claude']</sub>

qa-contract/src/query.mjs
```diff
@@ -0,0 +1,174 @@
...
+    } else {
+      if (ownerId) where.push(['$ownerId', '==', ownerId]);
+      if (values.testId) where.push(['testId', '==', values.testId]);
+      if (values.network) where.push(['network', '==', networkId(values.network)]);
+      if (values.result) where.push(['result', '==', values.result]);
+      orderBy.push(['$createdAt', 'desc']);
+    }
+  } else {
+    throw new Error("--type must be 'testCase' or 'testRun'.");
+  }
```
thepastaclaw, line 161: **💬 Nitpick: testRun CLI builds queries no index can serve**

All testRun indices except `buildRefOwner` are `$ownerId`-prefixed and require `testId` immediately after the owner. `--type testRun --result pass` (no `--testId`) produces `where: [['$ownerId','==',X], ['result','==','pass']]`, which no index covers — the server returns an opaque "no matching index" error. Same for `--type testRun --network testnet` alone. Additionally, the `if (values.buildRef) ... else ...` split is mutually exclusive, so `--buildRef X --result Y` silently drops the `result` filter (and `--testId`/`--network`) on the buildRef branch with no warning. Add a small CLI-side check: require `--testId` whenever `--buildRef` is absent, and reject (or at minimum warn about) `--buildRef` combined with `--result`/`--testId`/`--network`.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

PR #3910 adds an isolated qa-contract/ package (schema + Node CLI scripts) that does not touch consensus, drive, dpp, or proof code. Reconciling against 6fa36ea4, all four prior suggestions remain valid (planCommit drop on --update, missing testId existence check before immutable testRun, missing testRun.result enum, plaintext WIF + lax .env perms). Two new in-scope suggestions: a strictly-redundant testRun index and a query CLI that builds index-less testRun queries. No blocking issues; recommending COMMENT.

<details>

```

In `qa-contract/src/seed.mjs`:
- [SUGGESTION] qa-contract/src/seed.mjs:92-108: `--update` can erase an existing `planCommit`
  `cleanProps` drops fields whose value is undefined/null/'', and `contentEquals` then compares CONTENT_FIELDS with `?? undefined`. When `resolvePlanCommit()` returns undefined (git unavailable, detached run, CI shallow clone, etc.), the new row's `planCommit` is absent, `contentEquals` reports a mismatch against a previously seeded row, and `sdk.documents.replace({ document: doc, ... })` writes `properties: props` without `planCommit` — silently deleting the audit-log provenance the schema/README intend to preserve. Before replacing, carry forward the existing `planCommit` if the current run cannot resolve a new one (e.g., `if (props.planCommit === undefined && existingJson.planCommit) props.planCommit = existingJson.planCommit;`).

In `qa-contract/schema/qa-contract.documents.json`:
- [SUGGESTION] qa-contract/schema/qa-contract.documents.json:201-213: Enforce documented testRun enums in the contract schema
  The README and the schema's own description declare `result` as `pass | fail | blocked | skipped` and `network` as `0=mainnet, 1=testnet, 2=devnet, 3=regtest`, but the on-chain schema only constrains `result` by length and `network` by `minimum: 0`. Because `testRun` is immutable and non-deletable, any non-CLI SDK caller today — and any submitter after the documented v2 relaxation of `creationRestrictionMode` to `0` — can permanently store `PASS`, `unknown`, or `999`, breaking result/network filters with rows that can never be corrected. The contract is the trust boundary; add `enum` constraints (`pass`/`fail`/`blocked`/`skipped`, and `0..3` for `network`) so Platform rejects out-of-vocabulary audit-log values before they become non-deletable records.
- [SUGGESTION] qa-contract/schema/qa-contract.documents.json:118-149: `ownerTestNetwork` is a strict prefix of `ownerTestNetworkCreated` and pays storage cost on every immutable testRun
  `ownerTestNetwork` indexes `[$ownerId, testId, network]` and `ownerTestNetworkCreated` indexes `[$ownerId, testId, network, $createdAt]`. Every equality query the shorter index can serve (e.g., `$ownerId==X AND testId==Y AND network==Z`) is also served by the longer index — with or without an `orderBy('$createdAt')`. Because `testRun` documents are immutable and non-deletable, every submitted run pays storage cost in both index trees forever; on long-running QA streams this compounds without serving any unique query shape. Drop `ownerTestNetwork` and let `ownerTestNetworkCreated` cover those queries. (`ownerTestCreated` is genuinely needed — different middle column.)
```
</details>

qa-contract/schema/qa-contract.documents.json
thepastaclaw: Correction: this prior finding is still valid at `6fa36ea`, not resolved. The auto-resolve reply here was posted in error by the review system. The current review carries it forward as: _Constrain `testRun.result` to the documented enum in the contract schema_.

qa-contract/src/submit-run.mjs
thepastaclaw: Correction: this prior finding is still valid at `6fa36ea`, not resolved. The auto-resolve reply here was posted in error by the review system. The current review carries it forward as: _Validate `--testId` against existing testCases before writing an immutable run_.

qa-contract/src/derive-identity-key.mjs
thepastaclaw: Correction: this prior finding is still valid at `6fa36ea`, not resolved. The auto-resolve reply here was posted in error by the review system. The current review carries it forward as: _Recovered signing key is logged to stdout and written to `.env` without restrictive permissions_.

qa-contract/src/seed.mjs
thepastaclaw: Correction: this prior finding is still valid at `6fa36ea`, not resolved. The auto-resolve reply here was posted in error by the review system. The current review carries it forward as: _`--update` can drop a previously stamped `planCommit`_.

QuantumExplorer: Fixed in 30eb6bb415 — seed.mjs --update now merges the new row over the existing content (carry-forward), so an absent planCommit (git unavailable) no longer overwrites/drops a previously stamped one.

qa-contract/src/submit-run.mjs
QuantumExplorer: Fixed in 30eb6bb415 — submit-run.mjs now queries the (testId, app) testCase before writing the immutable run and errors on a miss (orphan-prevention); --force overrides.

qa-contract/src/derive-identity-key.mjs
QuantumExplorer: Fixed in 30eb6bb415 — derive-identity-key.mjs masks the WIF by default (--print to reveal) and chmods the written .env to 0600. (Both bots now align on gating the print; done.)

qa-contract/schema/qa-contract.documents.json
QuantumExplorer: Agreed it belongs on-chain. It needs a fresh registration (a contract's schema is immutable), and the QA identity is down to ~5.4B credits (a register costs ~20-30B), so a 4th re-register isn't feasible this session without a top-up. Deferred to the next re-registration (testnet reset / next schema change); submit-run.mjs enforces pass|fail|blocked|skipped for v1's owner-only writer meanwhile. Documented in the README.

qa-contract/schema/qa-contract.documents.json
```diff
@@ -0,0 +1,284 @@
+{
+  "app": {
+    "type": "object",
+    "documentsMutable": true,
+    "canBeDeleted": true,
+    "creationRestrictionMode": 1,
```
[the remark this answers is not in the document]
QuantumExplorer: Applied in 2a67c3b69e — app/tier/category lookups are now canBeDeleted:false (testCase stays deletable for dropped plan rows). Schema is immutable, so it takes effect at the next re-registration.

qa-contract/schema/qa-contract.documents.json
```diff
@@ -0,0 +1,284 @@
...
+        "maxLength": 16,
+        "position": 2,
+        "description": "Outcome. One of: pass, fail, blocked, skipped."
+      },
+      "network": {
+        "type": "integer",
+        "minimum": 0,
+        "position": 3,
+        "description": "Network id the run executed against: 0=mainnet, 1=testnet, 2=devnet, 3=regtest."
+      },
```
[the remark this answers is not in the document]
QuantumExplorer: Applied in 2a67c3b69e — added maximum:3 to testRun.network. Deploys at the next re-registration.

qa-contract/schema/qa-contract.documents.json
QuantumExplorer: Applied in 2a67c3b69e — result now enum:[pass,fail,blocked,skipped] and network bounded 0..3 on-chain. Deploys at the next re-registration (schema immutable).

qa-contract/schema/qa-contract.documents.json
QuantumExplorer: Applied in 2a67c3b69e — dropped ownerAppTestNetwork; ownerAppTestNetworkCreated serves the same [$ownerId,app,testId,network] equality prefix, so no run pays for two overlapping trees. Kept ownerAppTestCreated (different middle column).

qa-contract/src/query.mjs
QuantumExplorer: Fixed in 2a67c3b69e — testRun CLI now requires --testId unless --buildRef, and rejects --buildRef combined with --testId/--result/--network, so it never builds an index-less query.

qa-contract/src/seed.mjs
QuantumExplorer: Already fixed in 30eb6bb415 (carry-forward merge of existing content into the --update doc) — this review was cut at 11a50f69, before that commit.

qa-contract/schema/qa-contract.documents.json
QuantumExplorer: Now applied on-chain-bound in 2a67c3b69e (result enum added to the schema); deploys at the next re-registration.

qa-contract/src/submit-run.mjs
QuantumExplorer: Fixed in 30eb6bb415 — submit-run.mjs verifies the (testId, app) testCase exists before the immutable write (--force overrides).

qa-contract/src/derive-identity-key.mjs
QuantumExplorer: Fixed in 30eb6bb415 — WIF masked by default (--print to reveal); written .env chmod'd to 0600.

qa-contract/src/seed.mjs
QuantumExplorer: Fixed in 30eb6bb415 — --update carries forward the existing planCommit/content, so it can't be dropped when git history is unavailable.

qa-contract/src/query.mjs
```diff
@@ -0,0 +1,171 @@
...
+  if (!values.type) { await selfCheck(sdk, contractId, ownerId, app, netId, limit, values.proof); return; }
+
+  const where = []; const orderBy = [];
+  if (['app', 'tier', 'category'].includes(values.type)) {
+    orderBy.push(['code', 'asc']); // byCode index
+  } else if (values.type === 'testCase') {
+    where.push(['app', '==', app]);
+    if (values.testId) where.push(['testId', '==', values.testId]);
+    if (values.tier) where.push(['tier', '==', tierCode(values.tier)]);
+    if (values.category) where.push(['category', '==', categoryCode(values.category)]);
```
thepastaclaw, line 135: **🟡 Suggestion: testCase query allows --tier with --category — no index covers app+tier+category**

The testCase indices are `testIdApp`, `appTier`, and `appCategory`. The CLI appends every supplied filter unconditionally, so `--type testCase --tier Essential --category Identity` builds `where: [['app','==',app], ['tier','==',t], ['category','==',c]]`, which no declared index can serve. The branch already rejects unservable combinations on the testRun path (see the `--buildRef` guard at lines 134-137) — apply the same shape here so a documented filter combo fails with a clear CLI error instead of an opaque Platform index error, or add a compound index.

<sub>source: ['codex']</sub>

qa-contract/src/query.mjs
```diff
@@ -0,0 +1,171 @@
...
+    } else {
+      if (!values.testId) {
+        throw new Error('testRun queries need --testId (with optional --result/--network), or --buildRef.');
+      }
+      if (ownerId) where.push(['$ownerId', '==', ownerId]);
+      where.push(['app', '==', app]);
+      where.push(['testId', '==', values.testId]);
+      if (values.network) where.push(['network', '==', networkId(values.network)]);
+      if (values.result) where.push(['result', '==', values.result]);
+      orderBy.push(['$createdAt', 'desc']);
```
thepastaclaw, line 157: **🟡 Suggestion: testRun query allows --network and --result together — no covering index**

The testRun indices `ownerAppTestNetworkCreated` and `ownerAppTestResultCreated` cover network OR result after `[$ownerId, app, testId]`, but no index covers both. The else-branch at lines 140-149 appends each filter independently, so `--testId X --network testnet --result pass` produces `where: [$ownerId, app, testId, network, result]` plus `orderBy $createdAt desc`, which Platform will reject with an unhelpful error. This is the same class of issue the new `--buildRef` guard at lines 134-137 prevents — extend that guard to also reject `--network` + `--result`, or add a covering index for the combination. Partial residual of prior finding #6.

<sub>source: ['codex']</sub>

qa-contract/schema/qa-contract.documents.json
```diff
@@ -0,0 +1,282 @@
...
+        "minLength": 1,
+        "maxLength": 32,
+        "position": 1,
+        "description": "Feature-area / Domain name (Core, Identity, DPNS, Token, Shielded, MultiWallet, ...)."
+      }
+    },
+    "required": ["code", "name"],
+    "additionalProperties": false,
+    "description": "Lookup table of feature-area categories. testCase.category holds the integer `code`."
+  },
```
thepastaclaw, line 97: **🟡 Suggestion: Mutable app/tier/category lookups can silently relabel historical immutable testRuns**

`app`, `tier`, and `category` all set `documentsMutable: true` with `canBeDeleted: false`. The `canBeDeleted` guard prevents orphaning a `code`, but does not prevent updating an existing lookup row's `name` (or `app.platform`). Because `testRun` is immutable and stores only the integer foreign key, replacing `app.code=0`'s name from 'SwiftExampleApp' to something else permanently relabels every historical run for that app — including audit-log entries that were intended to be tamper-evident. Either make the lookup document types `documentsMutable: false` (treat them as a stable code table; add new rows with the next `code` for any change, which the README already advocates), or split immutable `code↔name` mapping from a mutable display-metadata document type.

<sub>source: ['codex']</sub>

qa-contract/src/register.mjs
```diff
@@ -0,0 +1,80 @@
...
+          + 'so re-run with --force to publish a fresh contract (new id), or revert the schema.',
+        );
+      }
+      console.log(`Already registered: ${existing.contractId} (still resolves on ${network}).`);
+      console.log('Pass --force to register a fresh contract.');
+      return;
+    }
+    console.log(`Config has ${existing.contractId} but it no longer resolves on ${network} `
+      + '(testnet reset?). Registering a fresh contract.');
+  }
```
thepastaclaw, line 54: **🟡 Suggestion: Silent catch on contracts.fetch can publish a duplicate contract on transient errors**

`const onChain = await sdk.contracts.fetch(existing.contractId).catch(() => undefined);` collapses every fetch failure (transient gRPC error, DNS hiccup, deserialization issue, auth error) into the same `undefined` state that the next block treats as 'no longer resolves on $network — registering a fresh contract'. Without `--force`, a single transient lookup failure will publish a duplicate data contract under the QA owner, burn credits, overwrite `contract-id.<network>.json` with a new id, and propagate the new id to dashboard consumers via the next commit — none of which is recoverable. Inspect the error and only treat a structured NotFound as 'gone'; re-throw on anything else so transient failures abort instead of silently re-registering.

<sub>source: ['claude']</sub>

qa-contract/src/derive-identity-key.mjs
```diff
@@ -0,0 +1,115 @@
...
+    let env = existsSync(envPath) ? readFileSync(envPath, 'utf8') : '';
+    const setLine = (key, val) => {
+      const re = new RegExp(`^${key}=.*$`, 'm');
+      env = re.test(env) ? env.replace(re, `${key}=${val}`) : `${env}\n${key}=${val}`;
+    };
+    setLine('QA_PRIVATE_KEY', pick.wif);
+    if (Number.isFinite(pick.id)) setLine('QA_IDENTITY_KEY_ID', String(pick.id));
+    writeFileSync(envPath, env.endsWith('\n') ? env : `${env}\n`);
+    chmodSync(envPath, 0o600); // contains a private key + mnemonic — owner-only
+    console.log(`\nWrote QA_PRIVATE_KEY${Number.isFinite(pick.id) ? ' + QA_IDENTITY_KEY_ID' : ''} to ${envPath} (chmod 0600).`);
```
thepastaclaw, line 111: **💬 Nitpick: .env is written under default umask before chmod 0600 — small TOCTOU window on a private key**

`writeFileSync(envPath, ...)` at line 107 creates/overwrites the file under the current umask (typically 0o022 → 0644) BEFORE `chmodSync(envPath, 0o600)` at line 108 tightens it. Between those two syscalls the file containing the WIF private key is world-/group-readable. The key is testnet-only per `.env.example`, so impact is bounded, but a local user on a shared dev/CI host could race the write-then-chmod. Either pass `{ mode: 0o600 }` to `writeFileSync` (relies on umask further restricting), or `openSync(envPath, 'w', 0o600)` + write + close to enforce 0600 from creation.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Reconciled six prior findings against head 2a67c3b6: five are FIXED (planCommit merge in seed.mjs --update, testCase existence guard in submit-run.mjs, on-chain result/network constraints, .env chmod 0600, dropped redundant ownerAppTestNetwork index) and one is PARTIALLY FIXED (testRun CLI now rejects bad --buildRef combinations but still allows --network+--result without a covering index — carried forward). New surviving in-scope items are query-index gaps in testCase (tier+category) and testRun (network+result), a lookup-table mutability concern that can relabel historical runs, the register.mjs catch-all that can spuriously re-register on a transient fetch error, and a small write-then-chmod ordering nit on .env. Codex's blocking schemaSha mismatch is intentionally documented in the README as a pending operational re-registration step, so it is recorded as an out-of-scope follow-up rather than a blocker.

🟡 4 suggestion(s) | 💬 1 nitpick(s)

<details>

```

In `qa-contract/src/query.mjs`:
- [SUGGESTION] qa-contract/src/query.mjs:124-130: testCase query allows --tier with --category — no index covers app+tier+category
  The testCase indices are `testIdApp`, `appTier`, and `appCategory`. The CLI appends every supplied filter unconditionally, so `--type testCase --tier Essential --category Identity` builds `where: [['app','==',app], ['tier','==',t], ['category','==',c]]`, which no declared index can serve. The branch already rejects unservable combinations on the testRun path (see the `--buildRef` guard at lines 134-137) — apply the same shape here so a documented filter combo fails with a clear CLI error instead of an opaque Platform index error, or add a compound index.
- [SUGGESTION] qa-contract/src/query.mjs:140-149: testRun query allows --network and --result together — no covering index
  The testRun indices `ownerAppTestNetworkCreated` and `ownerAppTestResultCreated` cover network OR result after `[$ownerId, app, testId]`, but no index covers both. The else-branch at lines 140-149 appends each filter independently, so `--testId X --network testnet --result pass` produces `where: [$ownerId, app, testId, network, result]` plus `orderBy $createdAt desc`, which Platform will reject with an unhelpful error. This is the same class of issue the new `--buildRef` guard at lines 134-137 prevents — extend that guard to also reject `--network` + `--result`, or add a covering index for the combination. Partial residual of prior finding #6.

In `qa-contract/schema/qa-contract.documents.json`:
- [SUGGESTION] qa-contract/schema/qa-contract.documents.json:42-97: Mutable app/tier/category lookups can silently relabel historical immutable testRuns
  `app`, `tier`, and `category` all set `documentsMutable: true` with `canBeDeleted: false`. The `canBeDeleted` guard prevents orphaning a `code`, but does not prevent updating an existing lookup row's `name` (or `app.platform`). Because `testRun` is immutable and stores only the integer foreign key, replacing `app.code=0`'s name from 'SwiftExampleApp' to something else permanently relabels every historical run for that app — including audit-log entries that were intended to be tamper-evident. Either make the lookup document types `documentsMutable: false` (treat them as a stable code table; add new rows with the next `code` for any change, which the README already advocates), or split immutable `code↔name` mapping from a mutable display-metadata document type.

In `qa-contract/src/register.mjs`:
- [SUGGESTION] qa-contract/src/register.mjs:28-45: Silent catch on contracts.fetch can publish a duplicate contract on transient errors
  `const onChain = await sdk.contracts.fetch(existing.contractId).catch(() => undefined);` collapses every fetch failure (transient gRPC error, DNS hiccup, deserialization issue, auth error) into the same `undefined` state that the next block treats as 'no longer resolves on $network — registering a fresh contract'. Without `--force`, a single transient lookup failure will publish a duplicate data contract under the QA owner, burn credits, overwrite `contract-id.<network>.json` with a new id, and propagate the new id to dashboard consumers via the next commit — none of which is recoverable. Inspect the error and only treat a structured NotFound as 'gone'; re-throw on anything else so transient failures abort instead of silently re-registering.
```
</details>

qa-contract/src/query.mjs
QuantumExplorer: Fixed in e77a17c0bf — testCase query now rejects more than one of --testId/--tier/--category (the indices pair app with exactly one), with a clear CLI error.

qa-contract/src/query.mjs
QuantumExplorer: Fixed in e77a17c0bf — testRun query now rejects --network combined with --result (no covering index; separate ownerAppTest{Network,Result}Created indices).

qa-contract/src/register.mjs
QuantumExplorer: Fixed in e77a17c0bf — register.mjs no longer swallows a thrown contracts.fetch() error to undefined; a transient lookup failure now aborts (only a genuine undefined/absent contract triggers a fresh registration), so a flaky fetch can't publish a duplicate.

qa-contract/src/derive-identity-key.mjs
QuantumExplorer: Fixed in e77a17c0bf — .env is now written with { mode: 0o600 } (plus the existing chmod for the overwrite case), so the WIF is never briefly world-readable.

qa-contract/schema/qa-contract.documents.json
QuantumExplorer: Valid, but deferring rather than churn the live contract id a 5th time: lookups are owner-only and the seed never updates them (it skips by code), so for v1 the relabel risk is effectively nil. Making the lookups documentsMutable:false is the right hardening and is tracked for the next intentional schema re-registration (v2/multi-submitter), where it can't relabel an existing audit log. Not applied in-tree now to keep the committed schema == the deployed v4 contract.

QuantumExplorer: Update: now applied in-tree in 8d0261bb87 — app/tier/category are documentsMutable:false (fully immutable code table). It post-dates the deployed v4 contract and takes effect at the next re-registration; not spinning a 5th contract id solely for this (the register drift-guard + README flag the pending state).

qa-contract/README.md
```diff
@@ -0,0 +1,287 @@
...
+
+The scripts use the recommended **`@dashevo/evo-sdk`** (js-evo-sdk) in trusted
+mode (required so state-transition responses are proof-verified). Node ≥ 18.18.
+
+**Option A — standalone (published SDK):**
+
+```sh
+cd qa-contract
+yarn install        # or npm install
+```
```
thepastaclaw, line 209: **🟡 Suggestion: Documented `cd qa-contract && yarn install` fails: qa-contract is not a workspace**

Option A tells users to `cd qa-contract && yarn install`, but the root `package.json` declares a Yarn 4 project whose `workspaces` list does not include `qa-contract`, and `qa-contract/` has no `yarn.lock` of its own. Yarn 4 walks up to find the project root, sees that the current directory is not a member of it, and aborts with `The nearest package directory ... doesn't seem to be part of the project`, so dependency resolution never runs. `npm install` does work as the documented fallback on the same line, and Option B (workspace build) is unaffected — but the primary Yarn path is broken as written. Either add `qa-contract` to the root `workspaces`, drop the `yarn install` instruction in favour of `npm install`, or commit a `qa-contract/yarn.lock` and make this a real standalone Yarn project.

<sub>source: ['codex']</sub>

thepastaclaw (commented):
## Code Review

All five prior findings reconcile cleanly: the two unservable-query-combo guards (prior-1, prior-2), the transient-fetch-error duplicate-publish guard (prior-4), and the .env permission TOCTOU fix (prior-5) are present in the current head and verified. Prior-3 (mutable app/tier/category lookups silently relabeling immutable testRun history) remains unaddressed and is carried forward. Two new in-scope issues from this delta: the README's primary `cd qa-contract && yarn install` path fails under the root Yarn 4 project, and `parse-test-plan.mjs` still contains a literal NUL sentinel that makes Git treat the parser as binary.

<details>

```

In `qa-contract/README.md`:
- [SUGGESTION] qa-contract/README.md:196-201: Documented `cd qa-contract && yarn install` fails: qa-contract is not a workspace
  Option A tells users to `cd qa-contract && yarn install`, but the root `package.json` declares a Yarn 4 project whose `workspaces` list does not include `qa-contract`, and `qa-contract/` has no `yarn.lock` of its own. Yarn 4 walks up to find the project root, sees that the current directory is not a member of it, and aborts with `The nearest package directory ... doesn't seem to be part of the project`, so dependency resolution never runs. `npm install` does work as the documented fallback on the same line, and Option B (workspace build) is unaffected — but the primary Yarn path is broken as written. Either add `qa-contract` to the root `workspaces`, drop the `yarn install` instruction in favour of `npm install`, or commit a `qa-contract/yarn.lock` and make this a real standalone Yarn project.
```
</details>

qa-contract/README.md
QuantumExplorer: Fixed in f513178f4e — the standalone install now documents `npm install` (with a note that qa-contract is deliberately not a Yarn workspace member, so Yarn 4 aborts from a non-member dir). Also fixed the related parser nitpick: replaced the literal NUL sentinel in parse-test-plan.mjs with a regex split so Git no longer treats it as binary.

qa-contract/src/sdk.mjs
```diff
@@ -0,0 +1,192 @@
...
+
+// Build a PrivateKey + single-key IdentitySigner from a WIF or 64-char hex key.
+export function buildSigner(mod, keyString, network) {
+  const { IdentitySigner, PrivateKey } = mod;
+  const trimmed = (keyString || '').trim();
+  if (!trimmed) throw new Error('Missing private key (set QA_PRIVATE_KEY).');
+  const isHex = /^[0-9a-fA-F]{64}$/.test(trimmed);
+  const privateKey = isHex
+    ? PrivateKey.fromHex(trimmed, network)
+    : PrivateKey.fromWIF(trimmed);
```
thepastaclaw, line 103: **🟡 Suggestion: NETWORK=local with a hex QA_PRIVATE_KEY crashes in PrivateKey.fromHex**

connect() (sdk.mjs:80) explicitly accepts NETWORK=local and the README advertises QA_PRIVATE_KEY as WIF or 64-char hex. For the hex branch, buildSigner passes the raw string 'local' into PrivateKey.fromHex(trimmed, network). The evo-sdk NetworkLike parser accepts mainnet/testnet/devnet/regtest — not 'local' — so register/seed/submit-run all crash on local before reaching the signer. Either normalize the string to 'regtest' here (the same alias already used in NETWORK_IDS at sdk.mjs:62-64) or reject NETWORK=local with a clearer error.

```suggestion
  const keyNetwork = network === 'local' ? 'regtest' : network;
  const privateKey = isHex
    ? PrivateKey.fromHex(trimmed, keyNetwork)
    : PrivateKey.fromWIF(trimmed);
```

<sub>source: ['codex']</sub>

qa-contract/src/derive-identity-key.mjs
```diff
@@ -0,0 +1,117 @@
...
+    keyType: String(k.keyType), readOnly: !!k.isReadOnly, hex: normHex(k.data),
+  }));
+
+  const matches = [];
+  for (let idIdx = 0; idIdx <= 3; idIdx += 1) {
+    for (let kt = 0; kt <= 1; kt += 1) {
+      for (let ki = 0; ki <= 6; ki += 1) {
+        const path = `m/9'/${coin}'/5'/0'/${kt}'/${idIdx}'/${ki}'`;
+        let d;
+        try { d = await mod.wallet.deriveKeyFromSeedWithPath({ mnemonic, path, network }); } catch { continue; }
```
thepastaclaw: **🟡 Suggestion: NETWORK=local breaks derive-identity-key with misleading "No derived key matched"**

derive-identity-key.mjs supports NETWORK=local via connect(), and COIN(net) at line 27 already treats local as a testnet-like coin (1). However, the loop passes the raw 'local' string into mod.wallet.deriveKeyFromSeedWithPath({ network }). evo-sdk's NetworkLike parser rejects 'local', so every candidate derivation is caught and skipped, ending with the misleading "No derived key matched any on-chain key" error at line 78. Pass 'regtest' instead when network is 'local' so the local workflow remains usable.

```suggestion
        try { d = await mod.wallet.deriveKeyFromSeedWithPath({ mnemonic, path, network: network === 'local' ? 'regtest' : network }); } catch { continue; }
```

<sub>source: ['codex']</sub>

thepastaclaw (commented):
## Code Review

Latest delta resolves all three prior findings: app/tier/category lookups are now immutable in schema (commit 8d0261b), README install instructions correctly use npm with an explanation of why (commit f513178), and parse-test-plan.mjs no longer contains a literal NUL byte (commit f513178). Two new suggestions remain — both about the advertised NETWORK=local mode passing the string 'local' into evo-sdk key APIs that expect mainnet/testnet/devnet/regtest. No blocking issues; scope remains storage-layer-only.

🟡 2 suggestion(s)

<details>

```

In `qa-contract/src/sdk.mjs`:
- [SUGGESTION] qa-contract/src/sdk.mjs:96-98: NETWORK=local with a hex QA_PRIVATE_KEY crashes in PrivateKey.fromHex
  connect() (sdk.mjs:80) explicitly accepts NETWORK=local and the README advertises QA_PRIVATE_KEY as WIF or 64-char hex. For the hex branch, buildSigner passes the raw string 'local' into PrivateKey.fromHex(trimmed, network). The evo-sdk NetworkLike parser accepts mainnet/testnet/devnet/regtest — not 'local' — so register/seed/submit-run all crash on local before reaching the signer. Either normalize the string to 'regtest' here (the same alias already used in NETWORK_IDS at sdk.mjs:62-64) or reject NETWORK=local with a clearer error.

In `qa-contract/src/derive-identity-key.mjs`:
- [SUGGESTION] qa-contract/src/derive-identity-key.mjs:70: NETWORK=local breaks derive-identity-key with misleading "No derived key matched"
  derive-identity-key.mjs supports NETWORK=local via connect(), and COIN(net) at line 27 already treats local as a testnet-like coin (1). However, the loop passes the raw 'local' string into mod.wallet.deriveKeyFromSeedWithPath({ network }). evo-sdk's NetworkLike parser rejects 'local', so every candidate derivation is caught and skipped, ending with the misleading "No derived key matched any on-chain key" error at line 78. Pass 'regtest' instead when network is 'local' so the local workflow remains usable.
```
</details>

qa-contract/src/sdk.mjs
QuantumExplorer: Fixed in b2d39890f6 — added a sdkNetwork() helper (local→regtest) and buildSigner now passes it to PrivateKey.fromHex, so NETWORK=local no longer crashes the hex-key path.

qa-contract/src/derive-identity-key.mjs
QuantumExplorer: Fixed in b2d39890f6 — derive-identity-key now passes sdkNetwork(network) (local→regtest) to deriveKeyFromSeedWithPath, so the local workflow matches keys instead of failing with the misleading 'No derived key matched'.

thepastaclaw (approved):
## Code Review

Both prior findings about NETWORK=local crashing the evo-sdk key APIs are FIXED in commit b2d39890 via the new sdkNetwork() helper in sdk.mjs (mapping 'local' -> 'regtest') and its use in buildSigner (sdk.mjs:102) and derive-identity-key.mjs:70. The new codex 'blocking' finding about contract-id.testnet.json schemaSha (28818bca69425d87) not matching the on-disk schema hash (1c2a2055ba414fe4) is verifiable but intentional — README explicitly documents that the committed schema makes lookups immutable beyond the deployed v4 contract and that register.mjs flags the drift, requiring --force to publish a fresh contract. Nitpick-level operator-hardening notes from the security auditor are out of scope.

qa-contract/src/seed.mjs
```diff
@@ -0,0 +1,182 @@
...
+
+  // Retired (➖) rows are historical markers in the plan, not runnable tests —
+  // skip them so the on-chain catalog (and the dashboard) never carries a
+  // confusing "Unspecified / Unknown, no runs" entry. E.g. DOC-09, whose
+  // local-demo mock was folded into the real broadcast flow (see DOC-02).
+  const retired = rows.filter((r) => (r.implStatus || '').trim() === '➖');
+  if (retired.length) {
+    console.log(`Skipping ${retired.length} retired (➖) row(s): ${retired.map((r) => r.testId).join(', ')}`);
+    rows = rows.filter((r) => (r.implStatus || '').trim() !== '➖');
+  }
```
thepastaclaw: **🟡 Suggestion: Retired rows are skipped on seed but not deleted from the catalog**

The new filter drops retired (➖) rows before the upsert loop, so retired testCases that were already seeded before this commit (e.g. DOC-09 on testnet) stay on-chain indefinitely — exactly the 'confusing Unspecified / Unknown, no runs' entry the comment says will never appear. `testCase` is `canBeDeleted: true` / `documentsMutable: true` (schema/qa-contract.documents.json:100-101), so the seeder can actually remove them. Either delete (or tombstone via implStatus update) any existing testCase matching a retired (testId, app) before filtering it out of the upsert loop, or relax the comment to acknowledge that retired rows seeded prior to this change must be cleaned up manually.

<sub>source: ['codex']</sub>

thepastaclaw (commented):
## Code Review

Incremental review reconciliation: the prior reviewed commit b2d39890 had no findings, so nothing is carried forward. The latest delta is the retired-row seeding filter in seed.mjs. One in-scope suggestion: the new filter only prevents fresh seeding of retired rows but does not delete previously-seeded rows from the on-chain catalog, despite the author's comment promising otherwise; the schema supports deletion (`canBeDeleted: true`). No other in-scope issues found.

🟡 1 suggestion(s)

<details>

```

In `qa-contract/src/seed.mjs`:
- [SUGGESTION] qa-contract/src/seed.mjs:116-124: Retired rows are skipped on seed but not deleted from the catalog
  The new filter drops retired (➖) rows before the upsert loop, so retired testCases that were already seeded before this commit (e.g. DOC-09 on testnet) stay on-chain indefinitely — exactly the 'confusing Unspecified / Unknown, no runs' entry the comment says will never appear. `testCase` is `canBeDeleted: true` / `documentsMutable: true` (schema/qa-contract.documents.json:100-101), so the seeder can actually remove them. Either delete (or tombstone via implStatus update) any existing testCase matching a retired (testId, app) before filtering it out of the upsert loop, or relax the comment to acknowledge that retired rows seeded prior to this change must be cleaned up manually.
```
</details>

qa-contract/src/seed.mjs
QuantumExplorer: Fixed in 2fb83faec4 — seed.mjs now deletes any already-seeded testCase for a retired (testId, app) (testCase is deletable), not just skipping the upsert, so retired rows can't linger on-chain. Selection filters are applied before the retired split so cleanup is scoped. (The plan currently marks 0 rows ➖, so this is a no-op on the live v4 catalog until a row is retired.)

thepastaclaw (commented):
## Code Review

The prior retired-row finding from f9e3395a is FIXED: seed.mjs now deletes already-seeded retired (➖) testCase documents on-chain before pruning them from the upsert set, and selection filters now run before retired-cleanup so both phases share the same scope. One in-scope suggestion remains in the latest delta: the retired-delete catch block logs failures but doesn't affect the script's exit status, which can silently leave stale retired entries on-chain — exactly the state this commit was meant to fix.

🟡 1 suggestion(s)

<details>

```

In `qa-contract/src/seed.mjs`:
- [SUGGESTION] qa-contract/src/seed.mjs:144-146: Retired-row delete failures don't affect exit status
  The catch block for retired-row deletes logs the error to stderr but doesn't increment any failure counter. `failed` (declared at line 162) only tracks upsert failures, and the exit gate at line 199 (`if (failed) process.exit(1)`) only considers upserts. A transient network blip on one of several retired deletes therefore leaves the on-chain catalog in the very stale state this commit was added to fix, while seed.mjs still exits 0 and prints `Seed complete ... 0 failed`. CI / re-run automation relying on the exit code will silently miss the regression. Either count retired-delete failures in `failed` (or a sibling counter that also gates the exit) and surface them in the final summary, or fail-fast by re-throwing.
```
</details>

qa-contract/src/seed.mjs
```diff
@@ -0,0 +1,202 @@
...
+          id: String(existing.toJSON().$id), ownerId, dataContractId: contractId, documentTypeName: 'testCase',
+        },
+        identityKey,
+        signer,
+      });
+      deleted += 1;
+      console.log(`  - retired ${r.testId} (deleted on-chain)`);
+    } catch (e) {
+      console.error(`  ! retired ${r.testId} delete failed: ${e?.message || e}`);
+    }
```
thepastaclaw, line 147: **🟡 Suggestion: Retired-row delete failures don't affect exit status**

The catch block for retired-row deletes logs the error to stderr but doesn't increment any failure counter. `failed` (declared at line 162) only tracks upsert failures, and the exit gate at line 199 (`if (failed) process.exit(1)`) only considers upserts. A transient network blip on one of several retired deletes therefore leaves the on-chain catalog in the very stale state this commit was added to fix, while seed.mjs still exits 0 and prints `Seed complete ... 0 failed`. CI / re-run automation relying on the exit code will silently miss the regression. Either count retired-delete failures in `failed` (or a sibling counter that also gates the exit) and surface them in the final summary, or fail-fast by re-throwing.

QuantumExplorer: Fixed in bbf17104e8 — retired-delete failures now increment a retiredFailed counter that's shown in the summary (… N retired-deleted, M failed) and gates the exit (if (failed || retiredFailed) process.exit(1)), so a failed delete no longer exits 0 with stale rows left on-chain.

qa-contract/README.md
```diff
@@ -0,0 +1,297 @@
...
+To let any identity submit runs (while keeping `testCase` owner-controlled),
+change **`testRun`** only:
+
+- set `creationRestrictionMode: 0` (NoRestrictions) on `testRun`, and
+- register the change via a data-contract **update** (or re-register on the next
+  testnet reset).
+
+`testRun` is already immutable + owner-stamped (`$ownerId`/`$createdAt` are
+system fields), so opening creation keeps every run attributable and tamper-proof.
+Leave `testCase` as OwnerOnly so the canonical catalog stays curated.
```
thepastaclaw, line 197: **🟡 Suggestion: Documented testRun v2 upgrade path is rejected by DPP**

The README tells operators to relax `testRun.creationRestrictionMode` to `0` and apply it via a data-contract update, but DPP rejects any change to `creation_restriction_mode` on an existing document type. See `packages/rs-dpp/src/data_contract/document_type/methods/validate_update/v0/mod.rs:119-132` (the unconditional inequality check in `validate_config`) and the dedicated regression test `should_return_invalid_result_when_creation_restriction_mode_is_changed` at lines 405-480 of the same file, which exercises exactly the OwnerOnly→NoRestrictions transition this README recommends. A user following these instructions will get a `DocumentTypeUpdateError` from the network. The README should instead state that opening `testRun` creation requires registering a fresh contract (or doing it before the first registration / on the next testnet reset noted parenthetically), not a contract update.

<sub>source: ['codex']</sub>

thepastaclaw (commented):
## Code Review

Prior retired-delete exit-status finding is FIXED at bbf17104 (seed.mjs:130,145,150,200-201 now track retiredFailed and gate exit). One new in-scope documentation defect: qa-contract/README.md advertises a v2 upgrade path that DPP's validate_update unconditionally rejects, so following the documented steps will fail on-chain.

🟡 1 suggestion(s)

<details>

```

In `qa-contract/README.md`:
- [SUGGESTION] qa-contract/README.md:184-195: Documented testRun v2 upgrade path is rejected by DPP
  The README tells operators to relax `testRun.creationRestrictionMode` to `0` and apply it via a data-contract update, but DPP rejects any change to `creation_restriction_mode` on an existing document type. See `packages/rs-dpp/src/data_contract/document_type/methods/validate_update/v0/mod.rs:119-132` (the unconditional inequality check in `validate_config`) and the dedicated regression test `should_return_invalid_result_when_creation_restriction_mode_is_changed` at lines 405-480 of the same file, which exercises exactly the OwnerOnly→NoRestrictions transition this README recommends. A user following these instructions will get a `DocumentTypeUpdateError` from the network. The README should instead state that opening `testRun` creation requires registering a fresh contract (or doing it before the first registration / on the next testnet reset noted parenthetically), not a contract update.
```
</details>

qa-contract/README.md
QuantumExplorer: Fixed in fd469b7434 — confirmed DPP's validate_update rejects any creationRestrictionMode change (the inequality check + should_return_invalid_result_when_creation_restriction_mode_is_changed test). Reworded the v2 section to require a fresh registration (before first register or at the next re-register/testnet reset), not a data-contract update.

thepastaclaw (approved):
## Code Review

Cumulative review of the qa-contract tooling PR. The latest delta is README-only and correctly resolves the prior finding: `qa-contract/README.md:184-197` now states that relaxing `testRun.creationRestrictionMode` requires a fresh contract registration (citing the DPP `validate_update` rejection), not a data-contract update. No new in-scope issues across schema, CLIs, or docs; PR is isolated additive QA tooling with no consensus surface touched.

PR #3915, opened 2026-06-16
fix(platform-wallet): apply disabled-key flags to local cache after identity update

QuantumExplorer:

The post-broadcast local-cache apply for the **disable-key** side of an identity update was an unimplemented TODO in `update_identity_with_external_signer`: it logged a warning and left the in-memory `ManagedIdentity` cache stale. As a result, a disabled key's `disabled_at` flag only flipped in SwiftData after a full network re-fetch (via `IdentityDetailView.refreshIdentityData()`).

This blocks promoting test case **ID-12 (disable identity key)** to a production UI action, where the Identity Keys list needs to reflect the disabled flag immediately.

## What was done?

Added `ManagedIdentity::disable_keys`, the disable-side counterpart to the existing `add_key`:

- Stamps `disabled_at` on the matching `IdentityPublicKey` records in the cached DPP `Identity`, so every signing / introspection path sees the disabled flag immediately.
- Fires the persister with the same combined changeset shape `add_key` uses (scalar identity snapshot + `IdentityKeysChangeSet` upserts), so the client's `PersistentPublicKey.disabledAt` rows update without a network re-fetch. The FFI derives the Swift `disabledAt` from `entry.public_key.disabled_at()`.
- Each emitted key entry **reuses the `(wallet_id, identity_index, key_index)` derivation breadcrumb** that `add_key` carries — reconstructed from the managed identity's own slot. This matters because the Swift upsert wipes `privateKeyKeychainIdentifier` when `derivationIndices` is absent; carrying the breadcrumb makes the client re-derive idempotently and keep the key's private-key linkage instead of dropping it.
- Skips (with a warning) key ids not present on the identity and emits no changeset when nothing matched.
- Does **not** touch the identity revision — the revision bump is already handled by the caller alongside the update transition.

Wired it into the call site in `update.rs`, replacing the TODO/`tracing::warn!` block. The local wall-clock time is passed as a placeholder `disabled_at`; the next Platform refresh reconciles it to the authoritative on-chain block time.

## How Has This Been Tested?

- `cargo check -p platform-wallet` — passes.
- `cargo test -p platform-wallet --lib managed_identity` — **21 passed, 0 failed**, including the new `test_disable_keys_stamps_disabled_at` (asserts the targeted key is stamped, an untouched key stays enabled, and a non-existent key id is skipped without creating a phantom row).
- `cargo fmt --all` — clean.

## Breaking Changes

None. Adds a new public method to `platform-wallet`; no consensus or wire-format changes.

## Checklist:

### DIFF
--- a/packages/rs-platform-wallet/src/wallet/identity/state/managed_identity/identity_ops.rs
+++ b/packages/rs-platform-wallet/src/wallet/identity/state/managed_identity/identity_ops.rs
@@ -7,6 +7,7 @@ use crate::wallet::persister::WalletPersister;
 use dpp::identity::accessors::IdentityGettersV0;
 use dpp::identity::identity_public_key::accessors::v0::IdentityPublicKeyGettersV0;
 use dpp::identity::Identity;
+use dpp::identity::{KeyID, TimestampMillis};
 use dpp::prelude::Identifier;
 use dpp::util::hash::ripemd160_sha256;
 use std::collections::BTreeMap;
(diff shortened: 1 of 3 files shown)


thepastaclaw:
✅ Review complete (commit 431d63b5)

QuantumExplorer (commented):
Reviewed

packages/rs-platform-wallet/src/wallet/identity/state/managed_identity/mod.rs
```diff
@@ -356,4 +356,52 @@ mod tests {
...
+            pk1.disabled_at(),
+            Some(1_700_000_000),
+            "targeted key must be stamped disabled"
+        );
+        // The skipped key id must not have materialized a phantom row.
+        assert!(
+            managed.identity.get_public_key_by_id(9).is_none(),
+            "non-existent key id must not be created"
+        );
+    }
```
thepastaclaw, line 406: **🟡 Suggestion: Test never exercises the breadcrumb — the load-bearing part of the fix**

The PR's stated purpose is that each emitted `IdentityKeyEntry` carries `(wallet_id, identity_index, key_index)` so the Swift upsert preserves `privateKeyKeychainIdentifier`. `test_disable_keys_stamps_disabled_at` constructs a `ManagedIdentity::new(identity, 0)` (leaving `wallet_id == None`) and uses `noop_persister()`, so the `match (self.wallet_id, self.identity_index)` always falls into the `None` arm and the persister discards any changeset. A regression that drops the breadcrumb, computes the wrong `key_index`, or stops calling `persister.store(...)` entirely would still pass this test. Add a recording persister, set `managed.wallet_id = Some(..)` and `identity_index = Some(..)`, then assert the captured `PlatformWalletChangeSet` contains one upsert with `wallet_id` set, `derivation_indices == Some(IdentityKeyDerivationIndices { identity_index, key_index: key_id })`, and the matching disabled public key. Cover the no-match early-return at lines 388–390 in the same pass.

packages/rs-platform-wallet/src/wallet/identity/state/managed_identity/identity_ops.rs
```diff
@@ -294,4 +295,109 @@ impl ManagedIdentity {
...
+            let (wallet_id, derivation_indices) = match breadcrumb {
+                Some((wallet_id, identity_index)) => (
+                    Some(wallet_id),
+                    Some(crate::changeset::IdentityKeyDerivationIndices {
+                        identity_index,
+                        key_index: key_id,
+                    }),
+                ),
+                None => (None, None),
+            };
```
thepastaclaw, line 371: **🟡 Suggestion: Document the `key_index == key_id` cross-module invariant**

`disable_keys` reconstructs the breadcrumb as `IdentityKeyDerivationIndices { identity_index, key_index: key_id }`, equating the DPP `KeyID` with the DIP-9 hardened derivation index. This holds today because every `add_key` call site passes `key_index = key.id()` (e.g. `update.rs:231`, registration), and `MASTER_KEY_INDEX == 0` happens to equal the master key id. But this is an implicit cross-module invariant, not a typed one — if a future wallet adds keys with non-monotonic IDs (revoke/re-add, imports), the persisted breadcrumb will point the Swift Keychain at the wrong derivation slot, silently defeating the PR's stated goal. Either cache the derivation index per key at `add_key` time and read it back here, or at minimum add a doc-comment and a `debug_assert` recording the invariant and the failure mode.

<sub>source: ['claude']</sub>

thepastaclaw (changes requested):
## Code Review

The PR cleanly mirrors `add_key` for the disable side and is largely correct, but it introduces one consequential side effect: stamping `disabled_at` in the local cache means the existing MASTER-key selector in `update_identity_with_external_signer` can now pick a disabled master after a rotation. Test coverage of the persistence contract (the actual fix payload — derivation breadcrumb + wallet_id) is also missing.

<details>

```

In `packages/rs-platform-wallet/src/wallet/identity/network/update.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/wallet/identity/network/update.rs:138-151: MASTER signer selector must skip disabled keys now that the cache stamps `disabled_at`
  This PR newly stamps `disabled_at` on cached `IdentityPublicKey`s in `disable_keys`, so the cached identity loaded at lines 117–120 can now contain disabled keys. The MASTER selector here iterates `public_keys()` and picks the first match by id without checking `disabled_at`. A valid master-key rotation disables the old MASTER and adds a new MASTER in the same update; after this PR applies that rotation locally, a subsequent update on the same identity will deterministically select the old (lower-id) MASTER even though `sign_external`/DPP reject keys with `disabled_at` set, and the update will fail locally despite the cache containing a valid enabled MASTER that could sign. This is a direct regression caused by making the local cache accurate. Add `&& key.disabled_at().is_none()` to the predicate.

In `packages/rs-platform-wallet/src/wallet/identity/state/managed_identity/mod.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/state/managed_identity/mod.rs:360-406: Test never exercises the breadcrumb — the load-bearing part of the fix
  The PR's stated purpose is that each emitted `IdentityKeyEntry` carries `(wallet_id, identity_index, key_index)` so the Swift upsert preserves `privateKeyKeychainIdentifier`. `test_disable_keys_stamps_disabled_at` constructs a `ManagedIdentity::new(identity, 0)` (leaving `wallet_id == None`) and uses `noop_persister()`, so the `match (self.wallet_id, self.identity_index)` always falls into the `None` arm and the persister discards any changeset. A regression that drops the breadcrumb, computes the wrong `key_index`, or stops calling `persister.store(...)` entirely would still pass this test. Add a recording persister, set `managed.wallet_id = Some(..)` and `identity_index = Some(..)`, then assert the captured `PlatformWalletChangeSet` contains one upsert with `wallet_id` set, `derivation_indices == Some(IdentityKeyDerivationIndices { identity_index, key_index: key_id })`, and the matching disabled public key. Cover the no-match early-return at lines 388–390 in the same pass.

In `packages/rs-platform-wallet/src/wallet/identity/state/managed_identity/identity_ops.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/state/managed_identity/identity_ops.rs:362-371: Document the `key_index == key_id` cross-module invariant
  `disable_keys` reconstructs the breadcrumb as `IdentityKeyDerivationIndices { identity_index, key_index: key_id }`, equating the DPP `KeyID` with the DIP-9 hardened derivation index. This holds today because every `add_key` call site passes `key_index = key.id()` (e.g. `update.rs:231`, registration), and `MASTER_KEY_INDEX == 0` happens to equal the master key id. But this is an implicit cross-module invariant, not a typed one — if a future wallet adds keys with non-monotonic IDs (revoke/re-add, imports), the persisted breadcrumb will point the Swift Keychain at the wrong derivation slot, silently defeating the PR's stated goal. Either cache the derivation index per key at `add_key` time and read it back here, or at minimum add a doc-comment and a `debug_assert` recording the invariant and the failure mode.
```
</details>

PR #3916, opened 2026-06-16
fix(rs-sdk-ffi): classify DAPI transport and timeout errors instead of internalError

QuantumExplorer:

Surfaced via QA test **ID-SYS-03**: in SwiftExampleApp's Platform Queries catalog, both *Get Protocol Version Upgrade State* and *Get Protocol Version Upgrade Vote Status* intermittently returned a misleading **"Internal Error: Failed to fetch protocol version upgrade…"** on testnet (proof-verified / TRUSTED mode).

Root-causing it revealed the query path was **never broken**. The raw error was:

```
Dapi client error: transport error: grpc error: 'service currently unavailable',
invalid peer certificate: certificate expired
```

A cluster of testnet evonodes was serving **expired TLS certs** (25/30 ENABLED nodes at one point). That is an ops issue, since resolved by cert renewal — there is nothing to fix in this repo for the outage itself.

The one genuine defect in this codebase is that this transient **network/transport** failure was classified and displayed as an **"Internal Error"**, which masked the real, retryable cause and actively misdirected diagnosis toward a phantom proof-verification bug. This PR fixes that classification.

## What was done?

- **`packages/rs-sdk-ffi/src/error.rs`** — `FFIError::from` now classifies by the typed `dash_sdk::Error` variant *first*, before the substring fallbacks:
  - `DapiClientError(_)` / `NoAvailableAddressesToRetry(_)` → `NetworkError`
  - `TimeoutReached(..)` → `Timeout`
  
  (Previously these fell through to `InternalError`: `DapiClientError`'s `Display` starts with `"Dapi client error:"`, which matched none of the old `"DAPI"`/`"dapi"`/`"connection"` heuristics.) Single source-of-truth classifier, so **every** platform query benefits — not just the protocol-version ones.
- **`packages/rs-sdk-ffi/src/protocol_version/queries/upgrade_state.rs`** and **`upgrade_vote_status.rs`** — the inner helpers now propagate the typed `dash_sdk::Error` via `?` instead of hand-rolling a `format!(…)` string that bypassed the classifier and forced `InternalError`. Pure bridging; no business logic added.
- **`packages/swift-sdk/Sources/SwiftDashSDK/FFI/PlatformQueryExtensions.swift`** — the JSON / array / string result helpers and the inline state-query error block call the existing `SDKError.fromDashSDKError(_:)` mapper instead of unconditionally throwing `.internalError`, so a transport failure renders as **"Network Error" / "Timeout"** in `QueryDetailView`.

## How Has This Been Tested?

- `cargo fmt --all -- --check`: clean.
- `cargo clippy -p rs-sdk-ffi --lib`: clean.
- `cargo check -p rs-sdk-ffi --all-targets`: clean.
- `cargo test -p rs-sdk-ffi --lib`: **282 passed**, including new `dapi_client_error_maps_to_network_error` and `timeout_reached_maps_to_timeout`.
- `./build_ios.sh --target sim`: BUILD SUCCEEDED (xcframework + SwiftExampleApp).
- **Simulator** (QA-iPhone16Pro, testnet, TRUSTED/proof-verified mode): both queries return valid live results once node rotation lands on a healthy node:
  - *Upgrade State* → `{ "upgrades": [{ "version_number": 12, "vote_count": 15 }] }`
  - *Vote Status* → array of `{ "pro_tx_hash": "05b687…", "version": 12 }`
  
  When a transport failure does occur it now surfaces as a "Network Error" rather than "Internal Error" (covered by the new unit tests).

## Breaking Changes

None. This only changes how existing transport/timeout errors are *classified and labeled* across the FFI boundary; no public API signatures or consensus behavior change.

## Checklist:

## Release Notes

* **Bug Fixes**
  * Improved error classification and reporting across SDK boundaries to provide more specific error details (network, timeout, invalid input) instead of generic internal errors.
  * Enhanced error handling in protocol version queries to preserve underlying error types when communicating between components.
  * Refined input validation with clearer error categorization for protocol queries.

### DIFF
--- a/packages/rs-sdk-ffi/src/error.rs
+++ b/packages/rs-sdk-ffi/src/error.rs
@@ -118,7 +118,25 @@ impl From<FFIError> for DashSDKError {
                     // emit only the inner Drive message so downstream FFI consumers
                     // don't double-render the "Drive internal error: " prefix.
                     (DashSDKErrorCode::DriveInternalError, inner.clone())
-                } else if error_str.contains("timeout") || error_str.contains("Timeout") {
+                } else if matches!(
+                    sdk_err,
+                    dash_sdk::Error::DapiClientError(_)
+                        | dash_sdk::Error::NoAvailableAddressesToRetry(_)
+                ) {
+                    // Transport / connectivity failure (e.g. all DAPI nodes
+                    // unreachable or serving expired TLS certificates). Match the
+                    // typed variant rather than the Display string: the message
+                    // ("Dapi client error: transport error: ...") matches none of
+                    // the substrings below, so it would otherwise fall through to
+                    // InternalError and surface in the UI as a misleading
+                    // "Internal Error" for what is really a network problem.
+                    (DashSDKErrorCode::NetworkError, error_str)
+                } else if matches!(sdk_err, dash_sdk::Error::TimeoutReached(_, _))
+                    || error_str.contains("timeout")
+                    || error_str.contains("Timeout")
+                {
+                    // Typed SDK timeout, plus a substring fallback for timeouts
+                    // surfaced inside other error types' Display strings.
                     (DashSDKErrorCode::Timeout, error_str)
                 } else if error_str.contains("I/O error") || error_str.contains("connection") {
                     (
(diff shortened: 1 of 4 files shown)


QuantumExplorer (commented):
Approved

thepastaclaw:
✅ Review complete (commit 02c190ee)

thepastaclaw (changes requested):
## Code Review

Well-scoped Rust FFI fix that classifies DAPI transport and timeout errors via typed `dash_sdk::Error` variants before substring heuristics, and routes the two protocol-version Swift helpers through `SDKError.fromDashSDKError` so codes survive the boundary. The Rust changes and memory handling are correct. One regression: routing the platform-query helpers through `fromDashSDKError` exposes that the Swift mapper has no case for `DashSDKErrorCode::DriveInternalError = 10`, so Drive-internal errors that previously rendered as `Internal Error` now fall through to `.unknown` (`Unknown Error`). Fixing this is in-scope because the same hunks introduced the regression.

🔴 1 blocking

<details>

```

In `packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift`:
- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift:620-623: Swift `fromDashSDKError` missing case for `DriveInternalError = 10`; routing query helpers through it regresses Drive errors to `.unknown`
  `packages/rs-sdk-ffi/src/error.rs:32` defines `DashSDKErrorCode::DriveInternalError = 10`, and lines 114-120 emit that code for `dash_sdk::Error::DriveInternalError`. The Swift mapper at `SDK.swift:603-625` only handles raw values 1–9 and 99, so code 10 falls into the `default` arm and becomes `.unknown(message)` ("Unknown Error: …").

  Before this PR, `processJSONResult` / `processJSONArrayResult` / `processStringResult` and the inline `getProtocolVersionUpgradeState` error block all threw `SDKError.internalError(message)` regardless of the FFI code, so a `DriveInternalError` surfacing from these queries rendered as "Internal Error: …". After this PR they route through `fromDashSDKError` (PlatformQueryExtensions.swift:24, :59, :84, :1091), so the same FFI error now renders as "Unknown Error: …" — a user-visible classification regression on the same surface the PR is trying to improve. Add a case for raw value 10 (mapping to `.internalError` to preserve prior labeling) so the round-trip stays lossless.
```
</details>

PR #3921, opened 2026-06-16
fix(swift-sdk): make Document transition contract/type pickers idb-drivable

QuantumExplorer:

The SwiftExampleApp Document state-transition builders (Purchase / Transfer / Update Price / Create / Replace / Delete) select their **data contract** and **document type** through SwiftUI `Picker`s that UI automation (idb) cannot drive. This blocked automating **DOC-07 (Purchase Document)** and every other contract-scoped document builder.

`TransitionDetailView` hosts these inputs in a `ScrollView { VStack }`, **not** a `Form`/`List`. In that host:
- `.accessibleInlinePicker` renders as an opaque control that idb sees only as a `Slider` — the option rows are absent from the accessibility tree (`describe-all` and `describe-point` return nothing tappable).
- `.accessibleFormPicker` / `.pickerStyle(.navigationLink)` renders and pushes its list, but **silently drops the selection binding** — navigationLink only commits its selection when it has a `List`/`Form` ancestor. (Verified on an iPhone 17 simulator: the row tap pops back but the checkmark stays on the previous value.)

So neither shared picker style works in this host, which is why the builders couldn't be driven.

## What was done?

Replaced the contract and document-type pickers with an explicit `NavigationLink` to a small selection `List` whose rows are `Button`s that write the binding directly and `dismiss()`:

- `ContractSelectionList` / `DocumentTypeSelectionList` in `TransitionInputView.swift`.
- Each row carries a stable `accessibilityIdentifier` — `transition.<input>.contractPicker.row.<idBase58>` / `transition.<input>.documentTypePicker.row.<name>` — so idb can tap a specific contract/type by id, robustly across localization and layout changes.
- Selection commits in **any** host (no `Form`/`List` dependency), and the trigger row shows the current selection. `contractPicker()` is shared by all six document builders, so all of them become drivable.

Also guarded `TransitionDetailView`'s one-time form setup with a `didInitializeForm` flag: pushing a picker list re-fires `.onAppear` on pop-back, and re-running `clearForm()` there wiped the in-progress form (including the just-picked contract). Setup now runs only on first appearance; a freshly navigated builder is a new view instance and still resets correctly.

## How Has This Been Tested?

Built the SwiftExampleApp for the simulator and drove **DOC-07 end-to-end via idb** on a booted iPhone 17 (testnet), using the existing fixture (contract `5jpKat9U82PGcfmeYm8QZZWX6q7zDo2C32PZMxHpbiGB`, doc type `card`, Knight card `Cdg8JTz6Rbx6Hq8swZLkVCLML3acRTtsShwVfrUpG2mX` priced at 100000 credits, buyer `qagaussqa` ≠ owner):

- Both pickers now push a real list with tappable, id-addressable rows.
- Contract and document-type selections **persist** across the pop-back (clearForm guard).
- The price **auto-fetches (100000 credits)** and the owner resolves — confirming the selections correctly drove the on-chain document/price lookup.

> Note: Execute itself currently fails with a **pre-existing, unrelated signing bug** — `selectSigningKey` (`StateTransitionExtensions.swift`) picks a CRITICAL `TRANSFER` key instead of an `AUTHENTICATION` key for document state transitions, so the platform rejects it (`requires AUTHENTICATION`). This affects all six document operations for identities that own a CRITICAL transfer key and is out of scope for this UI fix; it's tracked separately.

## Breaking Changes

None. UI/testability change in the example app only; no protocol or public-API changes.

## Checklist:

## Bug Fixes
* Fixed an issue where form inputs were unexpectedly reset when navigating back from picker selections

## New Features
* Redesigned contract and document type selection interface with dedicated selection lists for improved navigation flow
* Enhanced accessibility identifiers for better support with automated testing and accessibility tools

### DIFF
--- a/packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/TransitionDetailView.swift
+++ b/packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/TransitionDetailView.swift
@@ -34,6 +34,12 @@ struct TransitionDetailView: View {
   @State private var selectedContractId: String = ""
   @State private var selectedDocumentType: String = ""
   @State private var documentFieldValues: [String: Any] = [:]
+  /// Guards the one-time form setup in `.onAppear`. Contract / document-type
+  /// selection now pushes a child list onto the navigation stack; popping it
+  /// re-fires this view's `.onAppear`, and re-running `clearForm()` there
+  /// would wipe the in-progress form (including the just-picked contract).
+  /// Run setup only on the first appearance of each builder instance.
+  @State private var didInitializeForm = false
 
   // Query for data contracts
   @Query private var dataContracts: [PersistentDataContract]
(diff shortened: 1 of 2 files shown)


thepastaclaw:
✅ Review complete (commit 46f0acfe)

thepastaclaw (approved):
## Code Review

Both agents and CodeRabbit converged on no in-scope findings. The PR is a focused SwiftExampleApp UI/testability fix that swaps inline Pickers for NavigationLink-pushed selection lists with stable per-row accessibility IDs, and guards TransitionDetailView's one-time form setup so popping the pushed picker doesn't wipe the in-progress form. The didInitializeForm guard is correctly scoped to the view-instance lifetime (transitionKey is a let, so a freshly-built builder is a new struct with a fresh @State). No consensus-critical surface area touched.

PR #3922, opened 2026-06-16
fix(swift-sdk): sign document state transitions with an available AUTHENTICATION key

QuantumExplorer:

Every SwiftExampleApp document state transition (create / replace / delete / transfer / update-price / purchase) failed on-chain for any identity that owns a CRITICAL-security TRANSFER key:

> Protocol error: Invalid identity key purpose TRANSFER. This state transition requires AUTHENTICATION

This is now common, because identities are derived with a CRITICAL TRANSFER key (keyId added by the Rust `derive_and_persist` change). Reproduced end-to-end via DOC-07 (Purchase Document) with buyer `qagaussqa`.

**Root cause:** `selectSigningKey(from:operation:)` in `packages/swift-sdk/Sources/SwiftDashSDK/FFI/StateTransitionExtensions.swift`. For every operation except CONTRACT CREATE/UPDATE it fell to a general branch that preferred *any* `.critical` key regardless of purpose — selecting the CRITICAL TRANSFER key. Document-batch state transitions require an AUTHENTICATION-purpose key. The helper is shared by all six document ops, so this broke every document state transition for identities with a CRITICAL TRANSFER key.

## What was done?

Rewrote `selectSigningKey` to delegate to the SDK's existing `KeyManager.findSigningKey`, which:

1. **Filters by purpose** — `.authentication` for document ops (`.critical` + `.authentication` for the contract branch).
2. **Respects Keychain availability** — iterates ranked candidates and skips any whose private key isn't present in the Keychain, the same dual scheme (`identityId+keyIndex`, then `publicKeyHex`) the `KeychainSigner` trampoline uses to resolve the key at sign time.

For `qagaussqa` (keyId 1 AUTH/CRITICAL with no stored private key, keyId 5 AUTH/HIGH with a private key, keyId 6 TRANSFER/CRITICAL with a private key) it now correctly selects keyId 5 instead of keyId 6.

`KeyManager`/`KeychainManager` are `@MainActor`, but five of the six ops previously selected inside an off-main `DispatchQueue.global().async` block. So `selectSigningKey` is now `@MainActor`, and the call is hoisted to the MainActor portion of each op (before the continuation); the chosen `IdentityPublicKey` (which is `Sendable`) is captured into the work block. FFI calls, signer handling, defer/cleanup, method signatures, and the SwiftExampleApp `executeDocument*` handlers are unchanged — the one shared fix covers all six ops.

Swift-only change — no Rust/xcframework rebuild required.

## How Has This Been Tested?

Built (`SwiftExampleApp`, iPhone 17 simulator, `EXCLUDED_ARCHS=x86_64`) → BUILD SUCCEEDED, and driven end-to-end on **testnet**:

- **Purchase Document (DOC-07)** — buyer `qagaussqa` (owns a CRITICAL TRANSFER key), contract `5jpKat9U82PGcfmeYm8QZZWX6q7zDo2C32PZMxHpbiGB`, type `card`, document `Cdg8JTz6Rbx6Hq8swZLkVCLML3acRTtsShwVfrUpG2mX` (price 100000 credits). Signing succeeded (no purpose rejection); on-chain the document's `owner_id` changed from `85KjYZ…` to hex `0A3DA9E619A40A6C…` = `qagaussqa`, revision bumped to 3. Cross-checked against the app's SwiftData store.
- **Update Price** — same identity, document now owned by `qagaussqa`, new price 50000 → `"Document price updated successfully"`. Exercises a different signing call site.

Both ops would have signed with the CRITICAL TRANSFER key and been rejected under the old code; the platform accepting them confirms an AUTHENTICATION key is now used. The remaining ops (create/replace/delete/transfer) run the identical shared selector.

## Breaking Changes

None. This is a Swift SDK behavior fix; it is not consensus-breaking and does not change any public API signature.

## Checklist:

### DIFF
--- a/packages/swift-sdk/Sources/SwiftDashSDK/FFI/StateTransitionExtensions.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/FFI/StateTransitionExtensions.swift
@@ -22,6 +22,21 @@ private final class SendableOpaque: @unchecked Sendable { let p: OpaquePointer; 
 /// AND whose private material is actually available in the Keychain
 /// (delegating to KeyManager.findSigningKey, which mirrors how the
 /// KeychainSigner trampoline resolves the key at sign time).
+///
+/// **Caller contract / known coupling:** availability is checked against
+/// `KeychainManager.shared`, so the `signer` handed to the document ops
+/// below must be backed by the shared Keychain. Every current caller
+/// passes a `KeychainSigner`, which defaults to `KeychainManager.shared`,
+/// so the store queried here is exactly the one the signer will use at
+/// sign time. A signer bound to a different `KeychainManager`, a raw
+/// private-key signer (`KeyManager.createSigner(from:)`), or a
+/// hardware-backed callback signer is NOT consulted here — this preflight
+/// could then reject a key that signer can actually sign. Removing the
+/// coupling would mean ranking candidates by purpose/security in Swift
+/// and delegating the availability check to the supplied signer (which
+/// needs a `dash_sdk_signer_can_sign`-style FFI that does not yet exist;
+/// the `can_sign` vtable slot is only invoked internally by Rust during
+/// signing).
 @MainActor
 private func selectSigningKey(from identity: DPPIdentity, operation: String) -> IdentityPublicKey? {
     print("🔑 [\(operation)] Selecting public key for identity \(identity.id.toBase58String())")
--- a/packages/swift-sdk/Sources/SwiftDashSDK/FFI/StateTransitionExtensions.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/FFI/StateTransitionExtensions.swift
@@ -17,37 +17,38 @@ private final class SendableOpaque: @unchecked Sendable { let p: OpaquePointer; 
 
 // MARK: - Key Selection Helpers
 
-/// Helper to select the appropriate key for signing operations
-/// Returns the key we most likely have the private key for
+/// Helper to select the appropriate key for signing operations.
+/// Picks a key whose purpose matches the state-transition requirement
+/// AND whose private material is actually available in the Keychain
+/// (delegating to KeyManager.findSigningKey, which mirrors how the
+/// KeychainSigner trampoline resolves the key at sign time).
+@MainActor
 private func selectSigningKey(from identity: DPPIdentity, operation: String) -> IdentityPublicKey? {
-    // Select a suitable public key based on policy only. Availability of a
-    // matching private key is enforced by the caller when creating the signer.
     print("🔑 [\(operation)] Selecting public key for identity \(identity.id.toBase58String())")
 
-    let keys = Array(identity.publicKeys.values)
+    let km = KeyManager.withSharedKeychain()
 
-    // For contract create/update, require CRITICAL + AUTHENTICATION
+    // Contract create/update require a CRITICAL + AUTHENTICATION key.
     if operation == "CONTRACT CREATE" || operation == "CONTRACT UPDATE" {
-        if let k = keys.first(where: { $0.securityLevel == .critical && $0.purpose == .authentication }) {
+        if let k = km.findSigningKey(for: identity, purpose: .authentication, minimumSecurityLevel: .critical, preferCritical: true) {
             print("📝 [\(operation)] Selected CRITICAL AUTHENTICATION key #\(k.id)")
             return k
         }
-        print("❌ [\(operation)] No CRITICAL AUTHENTICATION key found")
+        print("❌ [\(operation)] No CRITICAL AUTHENTICATION key with available private material")
         return nil
     }
 
-    // Otherwise prefer CRITICAL, then AUTHENTICATION, then any
-    if let k = keys.first(where: { $0.securityLevel == .critical }) {
-        print("📝 [\(operation)] Selected CRITICAL key #\(k.id)")
+    // Document state transitions require an AUTHENTICATION-purpose key.
+    // Prefer a CRITICAL auth key, but findSigningKey skips any candidate
+    // whose private key is not in the Keychain, so an identity whose only
+    // CRITICAL key is a TRANSFER key (or a CRITICAL auth key with no stored
+    // private material) correctly falls through to its HIGH auth key.
+    if let k = km.findSigningKey(for: identity, purpose: .authentication, minimumSecurityLevel: nil, preferCritical: true) {
+        print("📝 [\(operation)] Selected AUTHENTICATION key #\(k.id) (security \(k.securityLevel.name))")
         return k
     }
-    if let k = keys.first(where: { $0.purpose == .authentication }) {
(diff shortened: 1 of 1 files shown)


thepastaclaw:
✅ Review complete (commit c23ee755)

packages/swift-sdk/Sources/SwiftDashSDK/FFI/StateTransitionExtensions.swift
```diff
@@ -17,37 +17,38 @@ private final class SendableOpaque: @unchecked Sendable { let p: OpaquePointer;
...
+    // private material) correctly falls through to its HIGH auth key.
+    if let k = km.findSigningKey(for: identity, purpose: .authentication, minimumSecurityLevel: nil, preferCritical: true) {
+        print("📝 [\(operation)] Selected AUTHENTICATION key #\(k.id) (security \(k.securityLevel.name))")
         return k
     }
-    let k = keys.first
-    if let k = k { print("📝 [\(operation)] Selected fallback key #\(k.id)") }
-    return k
+    print("❌ [\(operation)] No AUTHENTICATION key with available private material")
+    return nil
```
thepastaclaw, line 66: **🟡 Suggestion: Selector preflights the shared Keychain even though SDK methods accept an arbitrary signer handle**

`selectSigningKey` now goes through `KeyManager.withSharedKeychain().findSigningKey(...)`, which only returns a candidate if its private material is in `KeychainManager.shared`. The six `documentCreate/Replace/Delete/Transfer/UpdatePrice/Purchase` methods still take a generic `signer: OpaquePointer`, and the SDK separately exposes `KeyManager.createSigner(from:)` / `dash_sdk_signer_create_from_private_key`, so a caller can validly pass a non-shared-Keychain signer (raw private key, hardware-backed callback signer, or a Keychain-backed signer bound to a different KeychainManager). For those callers, this preflight will reject a key the supplied signer can actually sign and surface 'No suitable key found' before the signer is ever consulted; conversely, picking a key by shared-Keychain presence can pick a key the supplied signer cannot sign. In this repo every current call site (SwiftExampleApp TransitionDetailView, etc.) passes a `KeychainSigner` backed by the shared `KeychainManager`, so this is a latent design issue rather than an active bug — but the cleanest shape is to rank candidates by purpose/security in Swift and delegate availability to the supplied signer (e.g. a small `can_sign_with` FFI helper) instead of `KeychainManager.shared`. Document the coupling on the SDK method signatures if you intend to keep it.

<sub>source: ['codex']</sub>

thepastaclaw (commented):
## Code Review

Swift-only fix that replaces a buggy ad-hoc 'first CRITICAL key' picker (which could select a TRANSFER key) with a delegation to KeyManager.findSigningKey filtered on AUTHENTICATION purpose and Keychain availability, and hoists the now-@MainActor selector out of the off-actor DispatchQueue closures. The implementation is correct for the SwiftExampleApp flow (the only realistic caller), preserves FFI ownership/lifecycle, and resolves the qagaussqa-style regression. One residual design suggestion about coupling the selector to the shared Keychain.

🟡 1 suggestion(s)

<details>

```

In `packages/swift-sdk/Sources/SwiftDashSDK/FFI/StateTransitionExtensions.swift`:
- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/FFI/StateTransitionExtensions.swift:29-51: Selector preflights the shared Keychain even though SDK methods accept an arbitrary signer handle
  `selectSigningKey` now goes through `KeyManager.withSharedKeychain().findSigningKey(...)`, which only returns a candidate if its private material is in `KeychainManager.shared`. The six `documentCreate/Replace/Delete/Transfer/UpdatePrice/Purchase` methods still take a generic `signer: OpaquePointer`, and the SDK separately exposes `KeyManager.createSigner(from:)` / `dash_sdk_signer_create_from_private_key`, so a caller can validly pass a non-shared-Keychain signer (raw private key, hardware-backed callback signer, or a Keychain-backed signer bound to a different KeychainManager). For those callers, this preflight will reject a key the supplied signer can actually sign and surface 'No suitable key found' before the signer is ever consulted; conversely, picking a key by shared-Keychain presence can pick a key the supplied signer cannot sign. In this repo every current call site (SwiftExampleApp TransitionDetailView, etc.) passes a `KeychainSigner` backed by the shared `KeychainManager`, so this is a latent design issue rather than an active bug — but the cleanest shape is to rank candidates by purpose/security in Swift and delegate availability to the supplied signer (e.g. a small `can_sign_with` FFI helper) instead of `KeychainManager.shared`. Document the coupling on the SDK method signatures if you intend to keep it.
```
</details>

packages/swift-sdk/Sources/SwiftDashSDK/FFI/StateTransitionExtensions.swift
QuantumExplorer: Good catch — agreed this is a real coupling. A few notes on why it's intentional for now, plus what I changed:

- **The availability check is necessary, not incidental.** The bug is precisely that an identity's CRITICAL *authentication* key can have no private material on the device (qagaussqa keyId 1), so a pure purpose+security ranking would still hand back an unsignable key — just failing with a different error instead of the TRANSFER-purpose rejection. So we can't simply "rank in Swift and let the signer fail."

- **Today the coupling is consistent.** Every caller of these six ops passes a `KeychainSigner`, whose backing keychain defaults to `KeychainManager.shared` — the exact store `KeyManager.withSharedKeychain().findSigningKey(...)` queries. So the preflight checks the same store the signer uses at sign time.

- **The cleaner shape needs an FFI that doesn't exist yet.** Delegating availability to the *supplied* signer would require a `dash_sdk_signer_can_sign`-style export that invokes the signer vtable's `can_sign` slot from Swift. That slot is currently only called internally by Rust during signing; adding the export is a Rust/FFI change + xcframework rebuild, out of scope for this Swift-only fix.

I've documented the coupling and the would-be-FFI on `selectSigningKey` (pushed as a follow-up `docs(swift-sdk)` commit). Happy to file a follow-up to add the `can_sign` FFI and switch to signer-delegated availability if you'd like to remove the coupling for real.

thepastaclaw (approved):
## Code Review

The latest delta (4123871d..c23ee755) is documentation only — it extends the selectSigningKey docstring to record the implicit dependency on KeychainManager.shared. The prior shared-Keychain coupling finding is INTENTIONALLY DEFERRED via this documentation (no runtime change, all current callers use shared-Keychain KeychainSigner, fix requires a new dash_sdk_signer_can_sign FFI). No new latest-delta findings; cumulative changes (purpose=AUTHENTICATION key selection, MainActor hoisting, IdentityPublicKey Sendable) remain sound. Approving.

packages/swift-sdk/Sources/SwiftDashSDK/FFI/StateTransitionExtensions.swift
thepastaclaw: Resolved in `c23ee75` — _Selector preflights the shared Keychain even though SDK methods accept an arbitrary signer handle_ no longer present.

PR #3924, opened 2026-06-16
feat(rs-sdk-ffi): expose dash_sdk_signer_can_sign for signer-delegated key preflight

QuantumExplorer:

Follow-up to #3922 (review thread https://github.com/dashpay/platform/pull/3922#discussion_r3423222271).

`selectSigningKey(from:operation:)` in `StateTransitionExtensions.swift` picks the public key for a document state transition by ranking candidates by purpose/security, then checking private-material availability against `KeychainManager.shared` (via `KeyManager.withSharedKeychain().findSigningKey(...)`). The six document SDK ops (documentCreate / Replace / Delete / Transfer / UpdatePrice / Purchase) each accept a generic `signer: OpaquePointer`, so a caller passing a signer **not** backed by the shared Keychain — a raw-key `KeyManager.createSigner(from:)` signer, a hardware-backed callback signer, or a `KeychainSigner` bound to a different `KeychainManager` — would have its key availability mis-judged, and the preflight could reject a key that signer can actually sign with.

It's a latent design coupling, not an active bug: every current caller passes a `KeychainSigner` defaulting to `KeychainManager.shared`. But it's exactly the coupling #3922's own doc-comment flagged for removal.

## What was done?

Removed the coupling by delegating the **availability** decision to the *supplied* signer, while keeping key-selection **policy** (ranking by purpose/security) in Swift.

**`rs-sdk-ffi` (`signer.rs`)** — new export:

```rust
pub unsafe extern "C" fn dash_sdk_signer_can_sign(
    signer: *const SignerHandle,
    pubkey_bytes: *const u8,
    pubkey_len: usize,
    key_type: u8,
) -> bool
```

It reconstructs a minimal `IdentityPublicKey` from `(pubkey_bytes, key_type)` and calls the existing `VTableSigner::can_sign_with` (previously only invoked internally by Rust during signing). This naturally covers both inner variants: `Inner::Callback` forwards `(pubkey_bytes, key_type)` through the vtable's `can_sign_with` slot (same wire encoding as `sign_async`), and `Inner::Native` delegates to the wrapped Rust signer. `KeyType::try_from(key_type)` fails closed on any out-of-range byte (including the `0xFF` platform-address tag); a null `signer` returns `false`. No private-key material crosses the boundary.

**`swift-sdk`**:
- `selectSigningKey` now takes `signer: OpaquePointer?`. It ranks candidates via the new public `KeyManager.rankSigningCandidates(...)` and returns the first key the **supplied** signer reports it can sign with through a thin `signerCanSign(_:_:)` marshalling wrapper over `dash_sdk_signer_can_sign`.
- The shared-Keychain `findSigningKey` path is retained as the **fallback used only when no signer is available** (`signer == nil`).
- All six document ops pass their in-scope `signer` to `selectSigningKey`.
- `KeyManager.rankSigningCandidates(...)` is a public, read-only ranking entry point delegating to the existing private `rankKeys(...)` (no duplication; stays off the `@MainActor` Keychain path).

The same purpose/security policy as before is preserved: contract create/update → AUTHENTICATION + minimum CRITICAL; document ops → AUTHENTICATION, prefer CRITICAL. Existing log markers are unchanged.

## How Has This Been Tested?

- `cargo fmt --all` clean; `cargo check -p rs-sdk-ffi` (incl. `--all-targets`) clean.
- 4 new unit tests in `signer.rs` (`can_sign_ffi_*`): true/false vtable verdict round-trip, null signer → false, out-of-range key_type fails closed before the vtable is consulted. `cargo test -p rs-sdk-ffi can_sign_ffi` → **4 passed**; full `signer` suite → **39 passed**.
- `packages/swift-sdk/build_ios.sh` succeeded; the new `dash_sdk_signer_can_sign` symbol is present in the generated `rs-sdk-ffi.h`. **SwiftExampleApp built to `BUILD SUCCEEDED`**, confirming the Swift wrapper, `rankSigningCandidates`, and rewired `selectSigningKey` compile against the regenerated header.

## Breaking Changes

None. This adds a new FFI export and an additional `signer:` parameter on a `private` Swift helper (all in-tree call sites updated). Not consensus-breaking.

## Checklist:

### DIFF
--- a/packages/rs-sdk-ffi/src/signer.rs
+++ b/packages/rs-sdk-ffi/src/signer.rs
@@ -819,6 +819,102 @@ pub unsafe extern "C" fn dash_sdk_signer_destroy(handle: *mut SignerHandle) {
     let _ = Box::from_raw(handle as *mut VTableSigner);
 }
 
+/// Query whether the supplied signer can sign for a given identity public
+/// key, WITHOUT pulling any private-key material across the FFI.
+///
+/// This exists so the Swift SDK can keep its key-selection *policy* (ranking
+/// candidates by purpose / security level) in Swift while delegating the
+/// *availability* decision — "does this exact signer hold private material
+/// for this key?" — to the signer that will actually be used at sign time.
+/// Before this export existed, the only way to learn availability was to let
+/// Rust invoke the vtable's `can_sign_with` slot internally during signing,
+/// which forced Swift to preflight against a *separate* store
+/// (`KeychainManager.shared`). That coupling silently mis-judged any signer
+/// not backed by the shared Keychain (a raw-key `createSigner`, a
+/// hardware-backed callback signer, or a `KeychainSigner` bound to a
+/// different `KeychainManager`). This export removes that coupling: the
+/// preflight asks the supplied signer directly.
+///
+/// The verdict naturally covers BOTH inner `VTableSigner` variants:
+/// - `Inner::Callback` forwards `(pubkey_bytes, key_type)` through the
+///   vtable's `can_sign_with` slot, identical to the wire encoding used by
+///   `sign_async` (raw pubkey bytes + `KeyType` discriminant byte).
+/// - `Inner::Native` delegates to the wrapped Rust `Signer`, which compares
+///   the key material it holds.
+///
+/// `key_type` is the `dpp::identity::KeyType` discriminant byte (0–4). This
+/// FFI is for *identity* public keys only; the `0xFF` platform-address tag
+/// (and any other out-of-range byte) returns `false`.
+///
+/// Returns `false` (never aborts) on any of: null `signer`, an out-of-range
+/// `key_type`, or a signer that reports it cannot sign.
+///
+/// # Safety
+/// - `signer` must be a valid `*const SignerHandle` previously returned by
+///   this SDK (e.g. `dash_sdk_signer_create*` /
+///   `dash_sdk_signer_create_from_private_key`) and not yet destroyed. It may
+///   be null, in which case this returns `false`.
+/// - `pubkey_bytes` must point to at least `pubkey_len` readable bytes, OR be
+///   null with `pubkey_len == 0`. The bytes are only read for the duration of
+///   this call and are not retained.
+#[no_mangle]
+pub unsafe extern "C" fn dash_sdk_signer_can_sign(
+    signer: *const SignerHandle,
+    pubkey_bytes: *const u8,
+    pubkey_len: usize,
+    key_type: u8,
+) -> bool {
+    if signer.is_null() {
+        return false;
+    }
+
+    // This FFI is for identity public keys only. `KeyType::try_from` rejects
+    // anything outside 0–4 (including the 0xFF platform-address tag), so an
+    // out-of-range byte fails closed.
+    let key_type = match dash_sdk::dpp::identity::KeyType::try_from(key_type) {
+        Ok(kt) => kt,
+        Err(_) => return false,
+    };
+
+    let bytes = if pubkey_bytes.is_null() {
+        Vec::new()
+    } else {
+        // SAFETY: caller guarantees `pubkey_bytes` is valid for `pubkey_len`
+        // readable bytes (documented contract above).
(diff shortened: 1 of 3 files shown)


thepastaclaw:
✅ Review complete (commit 1e884dec)

thepastaclaw (commented):
## Code Review

Both agents converged on no blocking issues. The new `dash_sdk_signer_can_sign` FFI export is sound (null-checks, fail-closed `KeyType::try_from`, no ownership transfer, no private material crossing the boundary), and the six Swift call-site rewires correctly delegate availability to the supplied signer. Two nitpicks remain: one about a documented-but-easy-to-misuse `signer == nil` fallback path in Swift, and one minor doc-contract loosening in the Rust FFI Safety comment.

💬 2 nitpick(s)

packages/swift-sdk/Sources/SwiftDashSDK/FFI/StateTransitionExtensions.swift
```diff
@@ -17,52 +17,104 @@ private final class SendableOpaque: @unchecked Sendable { let p: OpaquePointer;
...
     }
-    print("❌ [\(operation)] No AUTHENTICATION key with available private material")
+
+    if isContractOp {
+        print("❌ [\(operation)] No CRITICAL AUTHENTICATION key with available private material")
+    } else {
+        print("❌ [\(operation)] No AUTHENTICATION key with available private material")
+    }
     return nil
 }
```
thepastaclaw, line 119: **💬 Nitpick: `signer == nil` fallback silently reintroduces the `KeychainManager.shared` coupling this PR removes**

All six in-tree callers of `selectSigningKey` (the document operations updated in this PR) pass a non-nil `signer: OpaquePointer`, so the `guard let signer = signer else { ... }` branch at lines 70–90 is unreachable today. The lines 51–53 docstring already calls out that passing `nil` falls back to `KeyManager.findSigningKey(...)` against `KeychainManager.withSharedKeychain()` — exactly the coupling the rest of the PR removes. Two cleaner options: (a) tighten the parameter to a non-optional `OpaquePointer` and delete the fallback so any future caller can't accidentally opt back into shared-Keychain semantics, or (b) leave the fallback but rename it so the foot-gun is explicit at the call site. Not a correctness defect for this PR, just a maintenance hazard.

<sub>source: ['claude']</sub>

packages/rs-sdk-ffi/src/signer.rs
```diff
@@ -819,6 +819,102 @@ pub unsafe extern "C" fn dash_sdk_signer_destroy(handle: *mut SignerHandle) {
...
+/// `key_type`, or a signer that reports it cannot sign.
+///
+/// # Safety
+/// - `signer` must be a valid `*const SignerHandle` previously returned by
+///   this SDK (e.g. `dash_sdk_signer_create*` /
+///   `dash_sdk_signer_create_from_private_key`) and not yet destroyed. It may
+///   be null, in which case this returns `false`.
+/// - `pubkey_bytes` must point to at least `pubkey_len` readable bytes, OR be
+///   null with `pubkey_len == 0`. The bytes are only read for the duration of
+///   this call and are not retained.
```
thepastaclaw, line 859: **💬 Nitpick: Safety contract under-specifies the `(non-null pubkey_bytes, pubkey_len == 0)` case**

The Safety section says `pubkey_bytes` must "point to at least `pubkey_len` readable bytes, OR be null with `pubkey_len == 0`". The Swift caller (`signerCanSign`) routes through `Data.withUnsafeBytes { ... bindMemory(to: UInt8.self).baseAddress }`, which for an empty `Data` may yield either `nil` or a non-null sentinel depending on Swift runtime. The implementation correctly handles both because `std::slice::from_raw_parts(non_null, 0)` is well-defined for any non-null aligned pointer, but the doc as written excludes the non-null/zero-length case. Loosening to "pointer is ignored when `pubkey_len == 0`" would match actual behavior. Pure documentation polish — no behavior change.

<sub>source: ['claude']</sub>

QuantumExplorer (commented):
Reviewed

Issue #3927, opened 2026-06-17
Document type allowing `documentsKeepHistory: true` + `canBeDeleted: true`: delete fails as an internal error, not a clean rejection


thephez:
## Expected Behavior

A document type that sets both `documentsKeepHistory: true` and `canBeDeleted: true` is self-contradictory: rs-drive unconditionally refuses to delete a document whose type keeps history, so `canBeDeleted: true` advertises a capability the storage layer will always reject.

The contradiction should be caught at the earliest possible point:

1. **Contract creation** — validation should reject a document type that sets both `documentsKeepHistory: true` and `canBeDeleted: true`, so the bad contract can never be deployed. (Mirrors the existing cross-flag rule `ContestedUniqueIndexOnMutableDocumentTypeError`.)
2. **Delete state transition** — even for already-deployed contracts, a delete targeting a keep-history document type should be rejected up front by the delete-transition structure validator as a normal invalid (paid) state transition, with a clear consensus error.

In both cases an SDK user should get an immediate, legible error instead of the delete failing as an internal error.

## Current Behavior

Nothing catches the contradiction up front, and the delete fails as an internal error rather than a clean rejection:

- **Contract creation accepts the combination.** `DocumentType::try_from_schema` parses `documentsKeepHistory: true` + `canBeDeleted: true` with no error. There is no cross-flag validation for this pair. (Live example: testnet contract `5CBPiadGmx3Zsjc26g5onopcx7pdxHPbrRAUD2T2yAbC`, document type `note`, has all of `documentsMutable`, `documentsKeepHistory`, and `canBeDeleted` set to `true`.)
- **The delete-transition structure validator does not check `documents_keep_history()`.** It only checks `documents_can_be_deleted()` (`document_delete_transition_action/advanced_structure_v0/mod.rs`), so the delete passes broadcast/structure validation.
- **At execution, Drive refuses the delete with a `DriveError`, not a consensus error.** `force_delete_document_for_contract_operations_v0` returns `DriveError::InvalidDeletionOfDocumentThatKeepsHistory` when `documents_keep_history()` is true. In the rs-drive-abci processing path this surfaces as `ExecutionResult::InternalError` — the transition is classified as **neither valid nor invalid** (`valid_count == 0`, `invalid_paid_count == 0`).

An internal error means "the node failed to process this", not "this transition is invalid". This is the wrong classification for what is really an invalid transition, and it offers no clean consensus accept/reject for a client to act on.

[...]

PR #3928, opened 2026-06-17
test(dpp,drive-abci): pin keepHistory + canBeDeleted contradiction

thephez:

Reproduces #3927. A document type can set both `documentsKeepHistory: true` and `canBeDeleted: true`, but rs-drive unconditionally refuses to delete a document whose type keeps history — so `canBeDeleted: true` advertises a capability the storage layer always rejects. Nothing catches this contradiction up front: contract creation accepts it, the delete-transition structure validator only checks `documents_can_be_deleted()` (not `documents_keep_history()`), and the delete fails deep in Drive as an `ExecutionResult::InternalError` (neither valid nor invalid) instead of a clean rejection.

This PR adds failing tests that pin the bug and mark exactly where each guard belongs. It does not include the fix — it's a red-test reproduction to anchor the follow-up.

## What was done?

Added failing tests across the three layers where the contradiction should be caught but currently isn't, plus passing companions so a future guard can't be over-broad.

**Contract creation — rs-dpp `try_from_schema/v2/mod.rs`:**
- `doctype_keep_history_with_can_be_deleted_rejected`: explicit `documentsKeepHistory: true` + `canBeDeleted: true` on the document schema must be rejected, naming both flags (fails today).
- Inherited/default-flag variants — because `canBeDeleted` defaults to `true`, a type that only sets `documentsKeepHistory: true` still hits the contradiction. These pin that the guard must inspect the resolved booleans, not just the raw schema keys: `doctype_keep_history_with_can_be_deleted_inherited_from_defaults_rejected` (both from contract defaults), `doctype_keep_history_default_with_explicit_can_be_deleted_rejected` and `doctype_explicit_keep_history_with_can_be_deleted_default_rejected` (each one-sided mix). All fail today.
- `doctype_keep_history_with_can_be_deleted_false_accepted`: `keepHistory + canBeDeleted: false` stays valid (passes — over-broad guard rejected).

**Local transition build — rs-dpp document factories:**
- `DocumentFactoryV0` (`delete_keep_history_document_is_rejected_locally`) and `SpecializedDocumentFactoryV0` (`create_state_transition_delete_with_keep_history_document_returns_error`) must reject a delete transition for a keep-history document type locally, before it can be signed and broadcast (fail today). This is the layer the SDK-side fix would surface at.

**Node delete validation — rs-drive-abci `batch/tests/document/deletion.rs`:**
- `test_document_delete_on_document_type_that_keeps_history_is_rejected` asserts the delete is refused up front as an invalid (paid) transition, mirroring the existing `can_not_be_deleted` test (fails today; the doomed delete is currently classified as an internal error).
- Parametrized so an inherited-defaults variant (`...inherits_keep_history_and_can_be_deleted...`) runs the same path against a contract whose flags come from config defaults — the structure validator only sees the resolved `DocumentTypeRef`, so the guard must cover this too.
- Fixtures `tests/supporting_files/contract/note/note-contract-keep-history-and-can-be-deleted.json` (explicit flags, mirroring live testnet contract `5CBPiadGmx3Zsjc26g5onopcx7pdxHPbrRAUD2T2yAbC` `note` type) and `...-by-default.json` (flags from contract defaults).

Suggested fix (separate PR): reject the resolved flag combination in `DocumentType::try_from_schema`, reject the delete in the DPP document factories / SDK before broadcast, and add a `documents_keep_history()` check next to the `documents_can_be_deleted()` check in the delete-transition structure validator.

## How Has This Been Tested?

Run at `PlatformVersion::latest()`:

- `cargo test -p dpp --lib keep_history` — the six rejection tests above fail as intended; the passing companions and unrelated keep-history tests stay green (no regressions).
- `cargo test -p drive-abci --lib keep_history` — the keep-history delete tests fail on the `invalid_paid_count == 1` assertion, confirming the delete is not rejected at validation today.

Also reproduced end-to-end against testnet via `@dashevo/evo-sdk` (`documents.create` then `documents.delete` against the contract above; the delete returns `InvalidDeletionOfDocumentThatKeepsHistory`).

Note: this PR is intentionally red — the new tests fail until the guards land (tracked in #3927).

## Breaking Changes

None — tests and test fixtures only.

## Checklist:


QuantumExplorer:
Closing in favor of #3930 which contains the complete fix and regression coverage.

PR #3929, opened 2026-06-17
docs(swift-example-app): retire stub builder rows VOTE-07/ID-13

QuantumExplorer:
`VOTE-07` (masternode vote — generic builder entry) and `ID-13` (top up identity — builder path) were marked `🧪` (builder-only), but their generic-builder entries are stubs (`default → notImplemented` / `notImplemented`). The real coverage already lives in `VOTE-01` and `ID-05`/`ID-06`.

Because `🧪` counts as "automatable now" (per §3), these two seeded a confusing **"no runs, never"** tile on the on-chain QA dashboard that looked like overlooked tests rather than intentional stubs — the same clutter we just cleared for the retired `DOC-09`.

## What was done?
- Reclassified both rows `🧪 → ➖` ("Retired — folded into another row", per the status legend), which is more accurate than `🧪` for a `notImplemented` stub.
- Updated the notes to point at the covering tests (`VOTE-01`, `ID-05`/`ID-06`) and record that the rows are kept here to document the stub but are **not seeded** to the QA catalog.
- The companion `seed.mjs` change (in the qa-contract PR #3910) already skips `➖` rows, so these drop out of the seeded catalog on the next re-seed.
- The two on-chain `testCase` documents (`VOTE-07`, `ID-13`) were deleted to match (dashboard already updated).

Net: the plan still documents the builder stubs (so QA doesn't mistake them for defects), but they no longer appear as runnable/seeded tests.

## How Has This Been Tested?
- Verified `VOTE-07`/`ID-13` were the only stub-marked builder rows in the plan.
- Confirmed the `seed.mjs` `➖`-skip guard skips these rows: a targeted `seed --ids VOTE-07,ID-13 --update` against the canonical plan seeds 0 and logs them as skipped.
- Confirmed both on-chain `testCase` docs now query to count 0.

## Breaking Changes
None — documentation/catalog only.

## Checklist:


thepastaclaw:
✅ Review complete (commit e33929c8)

thepastaclaw (approved):
## Code Review

Documentation-only change retiring two stub builder rows (ID-13, VOTE-07) in the Swift example app test plan. Both agents independently confirmed the reclassification is consistent with the legend and points to existing covering rows. No code or consensus impact.

PR #3934, opened 2026-06-18
fix(swift-sdk): update token balance from the transfer/burn proof result (MW-02)

QuantumExplorer:

QA test **MW-02** ("Token transfer between two on-device identities") was failing: after a token transfer A→B, the token-balance UI stayed stale — the sender's balance appeared "stuck" and the recipient never showed the arrived tokens. The on-chain transfer worked; the local `PersistentTokenBalance` rows the UI observes via `@Query` weren't updated.

## What was done?

A token transfer/burn's `broadcast_and_wait` **already returns proof-verified post-op balances** — `StateTransitionProofResult::VerifiedTokenIdentitiesBalances` (both sender *and* recipient) for transfer, `VerifiedTokenBalance` for burn — and `rs-platform-wallet` already returns them (`TransferResult`/`BurnResult`). The FFI was discarding them and Swift returned `Void`. So we **surface and persist the proof balances directly** — no follow-up query:

- **`rs-platform-wallet-ffi`**: `platform_wallet_token_transfer` / `_token_burn` emit the balances via an `out_balances_json` out-param. New `balances_json.rs` serializes `{identityIdBase58: balanceString}` — **u64 amounts as strings** to avoid JSON number precision loss. Non-balance proof variants (historical-document, group-action-with-document, and a group action still awaiting co-signatures) yield `{}`; a *closed* group burn emits the caller's proof-verified balance.
- **SwiftDashSDK**: `tokenTransfer`/`tokenBurn` now return `[Data: UInt64]`; the views persist it via `SDK.persistProvenTokenBalances` (upsert by `(canonicalTokenId, identityId)` + relationship link, using `try` not `try?`).
- **`IdentityDetailView`**: its Tokens section now reloads unconditionally on appear (was gated on an empty `@State` snapshot), so the primary identity-centric view is current after a transfer/burn.
- Added `ProvenTokenBalancePersistTests` (Swift) + 3 Rust unit tests for the serializer.

**One round-trip instead of two, race-free, proof-verified, recipient covered for free.** (This supersedes an earlier approach in this PR that did a separate `getIdentityTokenBalances` fetch; that helper and its off-main wrapper were removed.)

## How Has This Been Tested?

Built for the iOS Simulator and driven on testnet (iPhone 16 Pro): A (`J52PSRr…`) → B (`envSngUJ…`), Coin token `5w52FMyK…`:
- transfer 10 → A **250.01 → 240.01**, B row created **= 10.00**;
- transfer 5 (proof-balance build) → A **240.01 → 235.01**, B **10.00 → 15.00** — both persisted straight from the transfer's proof, UI reflects both.
Recorded `pass` for MW-02 on the [QA dashboard](https://dashpay.github.io/qa-dashboard-site/).

## Breaking Changes

None. (The `platform_wallet_token_transfer`/`_token_burn` FFI signatures gain an `out_balances_json` out-param — internal FFI, not a consensus or public-SDK contract.)

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Token actions (transfer and burn) now return proof-verified post-action balances, enabling automatic balance updates.
  * Token balances automatically refresh immediately after successful transfer and burn operations for wallet-owned identities.
* **Bug Fixes**
  * Token balances in the Identity detail view now reliably reload whenever the view appears, preventing stale displays after token flows.
* **Tests**
  * Added SwiftData persistence tests to ensure token balance rows are correctly upserted and linked.


thepastaclaw:
✅ Review complete (commit 188c27d4)

packages/swift-sdk/Sources/SwiftDashSDK/FFI/TokenBalanceRefresh.swift
```diff
@@ -0,0 +1,171 @@
...
+                    identityId: identityBase58,
+                    tokenIds: [canonicalTokenId]
+                )
+                // The query omits tokens the identity has never held;
+                // default those to 0 so a sender that drained to empty
+                // still gets its row zeroed.
+                result[identityId] = balances[canonicalTokenId] ?? 0
+            }
+            return result
+        }.value
```
thepastaclaw: **🟡 Suggestion: Task.detached still runs the blocking FFI on the main actor**

The docstring (lines 55-58) promises that `getIdentityTokenBalances` is hopped off-main via `Task.detached`, but `extension SDK` in `PlatformQueryExtensions.swift` is annotated `@MainActor` at line 52, which transitively makes `getIdentityTokenBalances` (line 1344) `@MainActor`-isolated. Awaiting a `@MainActor` async function from a detached task hops back to the main actor before executing, so `dash_sdk_token_get_identity_balances` — a blocking `runtime.block_on(...)` on the Rust side — still runs on the iOS main thread. After every transfer/burn the UI can stall for the duration of the network round-trip, which is exactly the symptom this helper was added to avoid. Either declare `getIdentityTokenBalances` (or a thin nonisolated wrapper) outside MainActor isolation, or call `dash_sdk_token_get_identity_balances` directly inside the detached task with `processJSONResult` so only the SwiftData upsert returns to MainActor.

<sub>source: ['codex']</sub>

packages/swift-sdk/Sources/SwiftDashSDK/FFI/TokenBalanceRefresh.swift
```diff
@@ -0,0 +1,171 @@
...
+                row = existing
+            } else {
+                row = PersistentTokenBalance(
+                    tokenId: canonicalTokenId,
+                    identityId: identityId,
+                    balance: 0,
+                    network: network
+                )
+                context.insert(row)
+            }
```
thepastaclaw: **🟡 Suggestion: Upsert masks SwiftData fetch failures as missing rows and can duplicate**

Both the token lookup at line 129 and the balance lookup at line 139 use `try?`, which swallows genuine SwiftData fetch failures (corrupt store, schema mismatch, etc.) and treats them as "row not found". The fallback path then inserts a fresh `PersistentTokenBalance`. `PersistentTokenBalance` has no unique constraint on `(tokenId, identityId)` (see `PersistentTokenBalance.swift` — `tokenId`/`identityId` are plain `var`s, not `@Attribute(.unique)`), so the save can succeed and leave duplicate rows, which then surface twice in the `identity.tokenBalances` relationship the burn/transfer views read. At minimum, propagate fetch errors (let them surface to the caller's `try?`/log path in the view) instead of silently inserting; collapsing the lookup error and the genuine-miss case hides exactly the failure mode that this refresh exists to recover from.

<sub>source: ['codex']</sub>

packages/swift-sdk/Sources/SwiftDashSDK/FFI/TokenBalanceRefresh.swift
```diff
@@ -0,0 +1,171 @@
...
+    /// Callers pass plain values (not the `PersistentToken` model) so
+    /// no SwiftData model is read across the network `await` /
+    /// off-main: the caller derives these on the main actor.
+    ///
+    /// - Parameters:
+    ///   - contractId: The 32-byte data-contract id that owns the token.
+    ///   - tokenPosition: The token's position within the contract.
+    ///   - tokenRelationshipKey: The local `PersistentToken.id`
+    ///     uniqueness key (`contractId || u16_be(position)`), used only
+    ///     to relink the `token` relationship on upserted rows.
```
thepastaclaw: **💬 Nitpick: Docstring misstates PersistentToken.id encoding**

The doc comment describes `PersistentToken.id` (and the `tokenRelationshipKey` parameter) as `contractId || u16_be(position)`. The actual encoding in `PersistentToken.init` (`PersistentToken.swift:73-79`) is `contractId || withUnsafeBytes(of: position.bigEndian)` where `position` is `Int` — 8 bytes on 64-bit iOS, not 2. The code is correct (it treats `token.id` opaquely as the relationship key), but the comment is misleading as a source-of-truth reference. Either drop the layout description or correct it to `Int64 big-endian (8 bytes)`.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Small, focused MW-02 fix: a new @MainActor SDK helper fetches per-identity token balances and upserts PersistentTokenBalance rows after a successful transfer/burn. The SwiftData persist path is correct and recipient scoping (wallet-owned, same network) is sound. Three correctness-adjacent issues warrant attention before merge — none are consensus-relevant: the off-main claim in the new helper is not actually honored, IdentityDetailView's in-memory token list still shows stale balances on the MW-02 happy path, and the upsert masks SwiftData fetch errors as missing rows (no unique constraint on (tokenId, identityId)).

<details>

```

In `packages/swift-sdk/Sources/SwiftDashSDK/FFI/TokenBalanceRefresh.swift`:
- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/FFI/TokenBalanceRefresh.swift:87-102: Task.detached still runs the blocking FFI on the main actor
  The docstring (lines 55-58) promises that `getIdentityTokenBalances` is hopped off-main via `Task.detached`, but `extension SDK` in `PlatformQueryExtensions.swift` is annotated `@MainActor` at line 52, which transitively makes `getIdentityTokenBalances` (line 1344) `@MainActor`-isolated. Awaiting a `@MainActor` async function from a detached task hops back to the main actor before executing, so `dash_sdk_token_get_identity_balances` — a blocking `runtime.block_on(...)` on the Rust side — still runs on the iOS main thread. After every transfer/burn the UI can stall for the duration of the network round-trip, which is exactly the symptom this helper was added to avoid. Either declare `getIdentityTokenBalances` (or a thin nonisolated wrapper) outside MainActor isolation, or call `dash_sdk_token_get_identity_balances` directly inside the detached task with `processJSONResult` so only the SwiftData upsert returns to MainActor.
- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/FFI/TokenBalanceRefresh.swift:129-149: Upsert masks SwiftData fetch failures as missing rows and can duplicate
  Both the token lookup at line 129 and the balance lookup at line 139 use `try?`, which swallows genuine SwiftData fetch failures (corrupt store, schema mismatch, etc.) and treats them as "row not found". The fallback path then inserts a fresh `PersistentTokenBalance`. `PersistentTokenBalance` has no unique constraint on `(tokenId, identityId)` (see `PersistentTokenBalance.swift` — `tokenId`/`identityId` are plain `var`s, not `@Attribute(.unique)`), so the save can succeed and leave duplicate rows, which then surface twice in the `identity.tokenBalances` relationship the burn/transfer views read. At minimum, propagate fetch errors (let them surface to the caller's `try?`/log path in the view) instead of silently inserting; collapsing the lookup error and the genuine-miss case hides exactly the failure mode that this refresh exists to recover from.

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/IdentityDetailView.swift`:
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/IdentityDetailView.swift:344-360: Identity detail token rows remain stale after transfer/burn dismisses
  The Tokens section iterates an in-memory `@State` array `tokenBalances: [IdentityTokenEntry]` (declared at line 75) whose values are fetched once and not derived from `PersistentTokenBalance`. The `onAppear` reload at lines 519-520 is gated on `tokenBalances.isEmpty`, so when the user pops back from `TokenTransferActionView`/`TokenBurnActionView` (which now refreshes `PersistentTokenBalance`), this list — the primary MW-02 entry point — keeps showing the pre-transfer numbers, and a sender drained to zero can still appear non-zero until the manual refresh button is tapped. Either refresh `tokenBalances` after a successful action (e.g. unconditionally reload on `onAppear`, or pass a callback through `TokenActionPermissionsView`), or rebuild the row data from the SwiftData-backed `identity.tokenBalances` so the PR's new upsert is observed reactively.
```
</details>

packages/swift-sdk/Sources/SwiftDashSDK/FFI/TokenBalanceRefresh.swift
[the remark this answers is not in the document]
QuantumExplorer: Fixed in d24aa602a7 — the per-identity fetch is now wrapped in do/catch inside the detached task; successful fetches are persisted and the first error is rethrown only if *none* succeeded, so one identity (e.g. recipient) failing no longer drops the sender's update.

packages/swift-sdk/Sources/SwiftDashSDK/FFI/TokenBalanceRefresh.swift
```diff
@@ -0,0 +1,171 @@
...
+
+        for (identityId, balance) in freshBalances {
+            let descriptor = FetchDescriptor<PersistentTokenBalance>(
+                predicate: #Predicate {
+                    $0.tokenId == canonicalTokenId && $0.identityId == identityId
+                }
+            )
+
+            let row: PersistentTokenBalance
+            if let existing = try? context.fetch(descriptor).first {
```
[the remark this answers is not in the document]
QuantumExplorer: Fixed in d24aa602a7 — the three `context.fetch` calls now use `try ...first` instead of `try?`, so a genuine fetch error propagates (to the view's best-effort catch) instead of being misread as a miss and falling through to a duplicate insert.

packages/swift-sdk/Sources/SwiftDashSDK/FFI/TokenBalanceRefresh.swift
QuantumExplorer: Fixed in d24aa602a7 — added a `nonisolated getIdentityTokenBalancesOffMain` (mirroring the existing `nonisolated calculateTokenId`) and call it from the detached task, so the blocking `dash_sdk_token_get_identity_balances` now runs off the main actor instead of hopping back via the @MainActor extension.

packages/swift-sdk/Sources/SwiftDashSDK/FFI/TokenBalanceRefresh.swift
QuantumExplorer: Fixed in d24aa602a7 — `try?`→`try ...first` in persistTokenBalances; a real fetch error now surfaces instead of falling through to inserting a duplicate (no unique constraint on (tokenId, identityId)).

packages/swift-sdk/Sources/SwiftDashSDK/FFI/TokenBalanceRefresh.swift
QuantumExplorer: Fixed in d24aa602a7 — corrected the docstring; `token.id`/`tokenRelationshipKey` is now described as an opaque contractId+position composite (dropped the inaccurate `u16_be` byte-layout detail).

QuantumExplorer:
Pushed d24aa602a7 addressing the review feedback:

- **Off-main fetch (thepastaclaw):** added `nonisolated getIdentityTokenBalancesOffMain` and call it from the detached task, so the blocking FFI no longer hops back to the main actor.
- **Partial-batch resilience (CodeRabbit):** per-identity `do/catch`; the first error is rethrown only if no identity succeeded.
- **Propagate fetch errors (CodeRabbit + thepastaclaw):** `try?`→`try ...first` in the upsert, avoiding a duplicate-row insert on a masked error.
- **Docstring (thepastaclaw):** corrected the `token.id` encoding note.
- **`IdentityDetailView` stale tokens (thepastaclaw, out-of-diff):** the Tokens section reloaded a `@State` snapshot only when empty, so the primary identity-centric MW-02 entry point stayed stale after a transfer/burn. Removed the `isEmpty` guard so `onAppear` reloads unconditionally on return (kept the minimal fix rather than a full `@Query` rewrite, since `reloadTokenBalances()` also drives the token-sync registration pipeline).

Builds clean for the iOS simulator.

packages/swift-sdk/Sources/SwiftDashSDK/FFI/TokenBalanceRefresh.swift
```diff
@@ -0,0 +1,216 @@
...
+                // errors; `.first` still yields nil for a legit miss
+                // (identity not present locally).
+                if let parent = try context.fetch(identityDescriptor).first {
+                    row.identity = parent
+                }
+            }
+        }
+
+        try context.save()
+    }
```
thepastaclaw, line 155: **🟡 Suggestion: New SDK.refreshTokenBalances has no unit tests locking in its invariants**

SDK.refreshTokenBalances is a new public SDK entry point that encodes several subtle, easy-to-regress invariants: canonical-id derivation, per-identity partial-failure semantics (only rethrow when every identity failed), defaulting a missing balance entry to 0 so a drained sender row gets zeroed, propagating SwiftData fetch errors via try (not try?) so a fetch failure cannot fall through to a duplicate insert (PersistentTokenBalance has no unique constraint on (tokenId, identityId)), and re-linking the identity/token relationships on every upsert so rows inserted before their parents still get stitched in. None of these are exercised by tests, and the PR description's checklist flags the missing-tests box. A unit test using an in-memory ModelContainer and a stub FFI fetch would lock in: upsert of an existing row preserves the identity link, a missing balance entry zeroes the row, an all-fail batch rethrows the first error, and a partial-fail batch persists what succeeded without rethrowing.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

All four prior findings against 438d41ab are resolved in d24aa602: the off-main FFI hop now routes through the nonisolated getIdentityTokenBalancesOffMain (no @MainActor re-hop), IdentityDetailView reloads token rows unconditionally on appear, persistTokenBalances uses try (not try?) so genuine SwiftData fetch errors propagate instead of silently inserting duplicates, and the docstring no longer mis-describes the PersistentToken.id layout. The only remaining in-scope suggestion is the absence of unit tests for the new SDK.refreshTokenBalances entry point, which the PR description itself acknowledges. The reviewer-flagged JSON u64 precision loss and missing catch_unwind at the rs-sdk-ffi entry points are pre-existing FFI debt inherited but not introduced by this PR and are recorded as out-of-scope follow-ups.

🟡 1 suggestion(s)

<details>

```

In `packages/swift-sdk/Sources/SwiftDashSDK/FFI/TokenBalanceRefresh.swift`:
- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/FFI/TokenBalanceRefresh.swift:70-215: New SDK.refreshTokenBalances has no unit tests locking in its invariants
  SDK.refreshTokenBalances is a new public SDK entry point that encodes several subtle, easy-to-regress invariants: canonical-id derivation, per-identity partial-failure semantics (only rethrow when every identity failed), defaulting a missing balance entry to 0 so a drained sender row gets zeroed, propagating SwiftData fetch errors via try (not try?) so a fetch failure cannot fall through to a duplicate insert (PersistentTokenBalance has no unique constraint on (tokenId, identityId)), and re-linking the identity/token relationships on every upsert so rows inserted before their parents still get stitched in. None of these are exercised by tests, and the PR description's checklist flags the missing-tests box. A unit test using an in-memory ModelContainer and a stub FFI fetch would lock in: upsert of an existing row preserves the identity link, a missing balance entry zeroes the row, an all-fail batch rethrows the first error, and a partial-fail batch persists what succeeded without rethrowing.
```
</details>

packages/swift-sdk/Sources/SwiftDashSDK/FFI/TokenBalanceRefresh.swift
QuantumExplorer: Addressed via the rework in 1282a40e0c: `refreshTokenBalances` (and `getIdentityTokenBalancesOffMain`) are now **deleted** — the transfer/burn persist the proof-verified balances the state transition already returns (`VerifiedTokenIdentitiesBalances`/`VerifiedTokenBalance`), so there's no follow-up query to test. Added `ProvenTokenBalancePersistTests` (sender+recipient upsert, existing-row identity-link preserved, zero balance) plus 3 Rust unit tests for the `balances_json` serializer (u64-as-strings).

packages/rs-platform-wallet-ffi/src/tokens/burn.rs
```diff
@@ -76,6 +93,24 @@ pub unsafe extern "C" fn platform_wallet_token_burn(
...
+            let c_str =
+                unwrap_result_or_return!(single_token_balance_to_json_cstring(&owner, remaining));
+            *out_balances_json = c_str;
+        }
+        BurnResult::HistoricalDocument(_)
+        | BurnResult::GroupActionWithDocument(_, _)
+        | BurnResult::GroupActionWithBalance(_, _, _) => {
+            *out_balances_json = unwrap_result_or_return!(write_empty_balances_json());
+        }
+    }
```
thepastaclaw, line 122: **🟡 Suggestion: Group burn balance is dropped when the action closes**

`BurnResult::GroupActionWithBalance(GroupSumPower, GroupActionStatus, Option<TokenAmount>)` carries the burner's proof-verified post-burn balance whenever the group action is closed — the Drive verifier at `packages/rs-drive/src/verify/state_transition/verify_state_transition_was_executed_with_proof/v0/mod.rs:475-480` explicitly errors if `status == ActionClosed && balance.is_none()`, so a `Some(remaining)` here is exactly the value this PR is trying to surface. The new match arm collapses both `Some` and `None` into `write_empty_balances_json()`, so a co-signature that closes a group-authorized burn leaves the local `PersistentTokenBalance` stale even though the broadcast result already contains the balance. The `id` parameter passed into the FFI is the burn caller (the same identity the proof verifies the balance for), so it's the correct key to serialize against. Split the variant so the `Some` case emits the single-balance JSON like the standard `TokenBalance` arm does.

```suggestion
        BurnResult::GroupActionWithBalance(_, _, Some(remaining)) => {
            let c_str =
                unwrap_result_or_return!(single_token_balance_to_json_cstring(&id, remaining));
            *out_balances_json = c_str;
        }
        BurnResult::HistoricalDocument(_)
        | BurnResult::GroupActionWithDocument(_, _)
        | BurnResult::GroupActionWithBalance(_, _, None) => {
            *out_balances_json = unwrap_result_or_return!(write_empty_balances_json());
        }
```

<sub>source: ['codex']</sub>

thepastaclaw (commented):
## Code Review

Cumulative + latest-delta review of the burn/transfer balance persistence refactor. Prior-1 (untested refreshTokenBalances helper) is resolved — the helper was deleted and replaced by persistProvenTokenBalances/persistTokenBalances, now covered by ProvenTokenBalancePersistTests. One new in-scope suggestion remains: the new FFI BurnResult→JSON mapping discards the proof-verified owner balance for the GroupActionWithBalance(_, _, Some(_)) variant. Drive's proof verifier guarantees a Some balance whenever the group action closes (verify_state_transition_was_executed_with_proof/v0/mod.rs:475-480), so the FFI is silently dropping data the PR's own design relies on. Severity is suggestion, not blocking, because the standard (non-group) burn path — the main path this PR targets — is correct.

🟡 1 suggestion(s)

<details>

```

In `packages/rs-platform-wallet-ffi/src/tokens/burn.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/tokens/burn.rs:108-113: Group burn balance is dropped when the action closes
  `BurnResult::GroupActionWithBalance(GroupSumPower, GroupActionStatus, Option<TokenAmount>)` carries the burner's proof-verified post-burn balance whenever the group action is closed — the Drive verifier at `packages/rs-drive/src/verify/state_transition/verify_state_transition_was_executed_with_proof/v0/mod.rs:475-480` explicitly errors if `status == ActionClosed && balance.is_none()`, so a `Some(remaining)` here is exactly the value this PR is trying to surface. The new match arm collapses both `Some` and `None` into `write_empty_balances_json()`, so a co-signature that closes a group-authorized burn leaves the local `PersistentTokenBalance` stale even though the broadcast result already contains the balance. The `id` parameter passed into the FFI is the burn caller (the same identity the proof verifies the balance for), so it's the correct key to serialize against. Split the variant so the `Some` case emits the single-balance JSON like the standard `TokenBalance` arm does.
```
</details>

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/IdentityDetailView.swift
```diff
@@ -510,15 +510,19 @@ struct IdentityDetailView: View {
...
+                // balance and refresh the local PersistentTokenBalance
+                // rows, but this section reads a transient @State
+                // snapshot — gating the reload on `tokenBalances.isEmpty`
+                // (the old behavior) left a drained sender still showing
+                // its pre-transfer figure until a manual refresh
+                // (MW-02). An unconditional reload keeps the displayed
+                // balance current after returning here; it stays a
+                // single batched round-trip and shows the cached numbers
+                // while it refreshes.
+                reloadTokenBalances()
```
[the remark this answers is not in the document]
QuantumExplorer: Fixed in 68ee5cafbc — added an in-flight guard (`guard !isLoadingTokens else { return }`) at the top of `reloadTokenBalances()`. It's reliable on the main actor since everything up to `isLoadingTokens = true` runs synchronously, so a rapid second `onAppear` bails before the first call's Task yields.

QuantumExplorer:
Addressed the review nits in 68ee5cafbc:
- **In-flight guard** (IdentityDetailView) — added (replied on the thread).
- **Redundant `tokenContractId` capture** (TokenTransferActionView) — removed; the persist step now reuses the already-captured `contractId`, matching TokenBurnActionView.
- **Token-relationship test coverage** (ProvenTokenBalancePersistTests) — added a case asserting the `token` link is set on first persist and survives an upsert.

Skipped the debug-logging nit on `tokenReadingBalancesJSON` (the Rust side controls the JSON shape, so a malformed entry is a can't-happen-in-practice contract violation; silently skipping keeps the hot path clean — happy to add a `#if DEBUG` log if you'd prefer).

packages/rs-platform-wallet-ffi/src/tokens/burn.rs
QuantumExplorer: Fixed in cb628aae07 — split the arm so `GroupActionWithBalance(_, _, Some(remaining))` emits the single-balance JSON keyed by `id` (the burn caller the proof verifies against), matching the standard `TokenBalance` path. `None` (action still open) and the document variants stay empty. `cargo check -p platform-wallet-ffi` passes.

thepastaclaw, line 122: **🟡 Suggestion: Closed group burn drops the caller's proof-verified balance**

`BurnResult::GroupActionWithBalance(GroupSumPower, GroupActionStatus, Option<TokenAmount>)` (defined in `packages/rs-sdk/src/platform/tokens/transitions/burn.rs:30` and emitted from `VerifiedTokenGroupActionWithTokenBalance` at line 89-91) carries `Some(remaining)` precisely when a co-signature closes the group burn — Drive's verifier in `packages/rs-drive/src/verify/state_transition/verify_state_transition_was_executed_with_proof/v0/mod.rs:475-480` errors out if `status == ActionClosed && balance.is_none()`, so `Some` here is the exact value MW-02 was designed to persist. The current arm collapses every `GroupActionWithBalance(_, _, _)` into `write_empty_balances_json()`, so when a co-signature actually closes the burn the local `PersistentTokenBalance` stays stale until the next periodic sync — the same staleness this PR was filed to eliminate, just on the closed-group path. The PR description explicitly says this case is handled. The burning identity `id` (line 53) is the correct key, since the proof verifier returns the balance for the burn caller. Split the variant so the `Some` case emits the single-balance JSON like the standard `TokenBalance` arm; the `None` case (action still awaiting signatures) keeps writing `{}`.

```suggestion
    match burn_result {
        BurnResult::TokenBalance(owner, remaining) => {
            let c_str =
                unwrap_result_or_return!(single_token_balance_to_json_cstring(&owner, remaining));
            *out_balances_json = c_str;
        }
        BurnResult::GroupActionWithBalance(_, _, Some(remaining)) => {
            let c_str =
                unwrap_result_or_return!(single_token_balance_to_json_cstring(&id, remaining));
            *out_balances_json = c_str;
        }
        BurnResult::HistoricalDocument(_)
        | BurnResult::GroupActionWithDocument(_, _)
        | BurnResult::GroupActionWithBalance(_, _, None) => {
            *out_balances_json = unwrap_result_or_return!(write_empty_balances_json());
        }
    }
```

thepastaclaw (commented):
## Code Review

Cumulative review of MW-02 token balance refresh refactor at 68ee5caf. The new FFI out-param → SwiftData persist path is sound (null-on-error contract, u64-as-string to dodge JSON precision loss, idempotent upsert, in-flight guard, link-survival test). One prior finding is still valid and not addressed by the latest delta: `BurnResult::GroupActionWithBalance` collapses every group-burn outcome to empty JSON, dropping the proof-verified post-burn balance that the Rust SDK emits when the group action closes — contradicting the PR description's own claim that 'a closed group burn emits the caller's proof-verified balance'.

🟡 1 suggestion(s)

<details>

```

In `packages/rs-platform-wallet-ffi/src/tokens/burn.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/tokens/burn.rs:102-113: Closed group burn drops the caller's proof-verified balance
  `BurnResult::GroupActionWithBalance(GroupSumPower, GroupActionStatus, Option<TokenAmount>)` (defined in `packages/rs-sdk/src/platform/tokens/transitions/burn.rs:30` and emitted from `VerifiedTokenGroupActionWithTokenBalance` at line 89-91) carries `Some(remaining)` precisely when a co-signature closes the group burn — Drive's verifier in `packages/rs-drive/src/verify/state_transition/verify_state_transition_was_executed_with_proof/v0/mod.rs:475-480` errors out if `status == ActionClosed && balance.is_none()`, so `Some` here is the exact value MW-02 was designed to persist. The current arm collapses every `GroupActionWithBalance(_, _, _)` into `write_empty_balances_json()`, so when a co-signature actually closes the burn the local `PersistentTokenBalance` stays stale until the next periodic sync — the same staleness this PR was filed to eliminate, just on the closed-group path. The PR description explicitly says this case is handled. The burning identity `id` (line 53) is the correct key, since the proof verifier returns the balance for the burn caller. Split the variant so the `Some` case emits the single-balance JSON like the standard `TokenBalance` arm; the `None` case (action still awaiting signatures) keeps writing `{}`.
```
</details>

QuantumExplorer:
This finding was already addressed in `cb628aae07`, which post-dates the `68ee5caf` this review ran against — the burn arm is now split: `BurnResult::GroupActionWithBalance(_, _, Some(remaining))` emits the single-balance JSON keyed by the burn caller (`id`), and only `None` / the document variants write `{}`. The inline thread is resolved; a re-review of HEAD should pick it up. (`cargo check -p platform-wallet-ffi` clean.)

packages/rs-platform-wallet-ffi/src/tokens/burn.rs
QuantumExplorer: Already implemented in cb628aae07 — the current code is exactly this arm split: `GroupActionWithBalance(_, _, Some(remaining))` emits the single-balance JSON keyed by `id`, and `None` / the document variants write `{}`. (This review ran against 68ee5caf, one commit before the fix.) cargo check clean.

thepastaclaw (changes requested):
## Code Review

Prior closed-group burn balance finding is fixed in cb628aae. One new in-scope blocking issue: CoSignProposalView discards the [Data: UInt64] map now returned by ManagedPlatformWallet.tokenBurn, so the closed-group-burn balance the latest commit was specifically added to surface is never persisted on the co-sign path. The other new Codex finding (Int64 bit-pattern wrap on large balances) mirrors the existing periodic-sync persister and is a pre-existing schema constraint, not a regression caused by this PR.

🔴 1 blocking

<details>

```

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/TokenActions/CoSignProposalView.swift`:
- [BLOCKING] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/TokenActions/CoSignProposalView.swift:415-423: Co-sign burn path discards the proof-verified balance the latest commit added
  `ManagedPlatformWallet.tokenBurn` now returns `[Data: UInt64]`, and cb628aae specifically wired `BurnResult::GroupActionWithBalance(_, _, Some(remaining))` to emit the caller's proof-verified post-burn balance — the case that fires precisely when a co-signer's signature closes the group action. This `dispatch` branch (and its sibling group-burn entry in `TokenBurnActionView`, which does persist) is the call site where that newly-added balance is most likely to be non-empty, yet here the `[Data: UInt64]` result is dropped on the floor. The on-chain burn succeeds, but the local `PersistentTokenBalance` rows remain stale until the next periodic sync — exactly the UI-staleness this PR is meant to eliminate. Capture the returned map and hand it to `SDK.persistProvenTokenBalances` (same shape as `TokenBurnActionView.swift:248-294`) so the closing co-signer's view updates immediately. Mint / freeze / unfreeze etc. don't return balances, so no fix is needed for the other branches.
```
</details>

QuantumExplorer:
Addressed the blocking finding in `188c27d4ce`: **CoSignProposalView** now captures the `[Data: UInt64]` map `tokenBurn` returns on the co-sign path and persists it via `SDK.persistProvenTokenBalances` (keyed by the view's `token`, best-effort), mirroring `TokenBurnActionView`. This is exactly where a co-signature *closes* the group burn — the case `cb628aae07` wired to emit `GroupActionWithBalance(_, _, Some(remaining))` — so the closing co-signer's local row now updates immediately; an empty map (action still open) is a no-op. App builds clean.

The other two items are non-blocking, as your review noted: the `Int64` bit-pattern balance storage mirrors the existing periodic-sync persister (a pre-existing schema constraint, no `@Attribute(.unique)`) — not a regression introduced here; and the omitted finding isn't in this diff. Happy to spin either into a separate issue if you'd like them tracked.

packages/rs-platform-wallet-ffi/src/tokens/transfer.rs
```diff
@@ -25,8 +39,11 @@ pub unsafe extern "C" fn platform_wallet_token_transfer(
     public_note: *const c_char,
     _signing_key_id: u32,
     signer_handle: *mut SignerHandle,
+    out_balances_json: *mut *mut c_char,
 ) -> PlatformWalletFFIResult {
     check_ptr!(signer_handle);
+    check_ptr!(out_balances_json);
+    *out_balances_json = std::ptr::null_mut();
```
thepastaclaw, line 46: **🟡 Suggestion: Null out_balances_json before validating signer_handle**

The doc comment for this new out-parameter promises that on any error the slot is set to null first, but `check_ptr!(signer_handle)` runs before the slot is cleared. If a caller passes a non-null `out_balances_json` slot containing a stale pointer together with a null `signer_handle`, the function returns `ErrorPointerIsNull` without ever clearing the slot, breaking the documented ownership contract. The in-tree Swift wrapper pre-initializes the slot to nil so there is no current app-side bug, but other C/Swift consumers that follow the documented contract would observe a stale pointer. Move the null-write of `out_balances_json` ahead of the signer pointer check so the contract holds for every error path.

```suggestion
    check_ptr!(out_balances_json);
    *out_balances_json = std::ptr::null_mut();
    check_ptr!(signer_handle);
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/tokens/burn.rs
```diff
@@ -30,8 +44,11 @@ pub unsafe extern "C" fn platform_wallet_token_burn(
     group_info_action_is_proposer: bool,
     _signing_key_id: u32,
     signer_handle: *mut SignerHandle,
+    out_balances_json: *mut *mut c_char,
 ) -> PlatformWalletFFIResult {
     check_ptr!(signer_handle);
+    check_ptr!(out_balances_json);
+    *out_balances_json = std::ptr::null_mut();
```
thepastaclaw, line 51: **🟡 Suggestion: Null out_balances_json before validating signer_handle**

Same documented-contract gap as in `transfer.rs`: the doc comment states that on error the slot is set to null first, but `check_ptr!(signer_handle)` runs before `*out_balances_json = null_mut()`. A null signer with a non-null out slot returns early without clearing the slot, so a caller relying on the documented behavior could observe stale ownership state. Today's Swift wrapper avoids this by initializing the slot to nil, so this is defensive rather than an active app bug. Reorder so the out-slot is cleared first.

```suggestion
    check_ptr!(out_balances_json);
    *out_balances_json = std::ptr::null_mut();
    check_ptr!(signer_handle);
```

<sub>source: ['codex']</sub>

thepastaclaw (commented):
## Code Review

Cumulative PR surfaces proof-verified post-transfer/burn balances through the wallet FFI as JSON-string-encoded u64 maps, Swift decodes them to [Data: UInt64], and persistProvenTokenBalances upserts into SwiftData. The latest delta (188c27d4) addresses the prior blocking finding: CoSignProposalView now captures and persists the balance map on the group-burn close path. One low-priority defensive suggestion remains: the new out_balances_json slot should be nulled before any other early return to fully honor its documented null-first-on-error contract.

🟡 2 suggestion(s)

<details>

```

In `packages/rs-platform-wallet-ffi/src/tokens/transfer.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/tokens/transfer.rs:44-46: Null out_balances_json before validating signer_handle
  The doc comment for this new out-parameter promises that on any error the slot is set to null first, but `check_ptr!(signer_handle)` runs before the slot is cleared. If a caller passes a non-null `out_balances_json` slot containing a stale pointer together with a null `signer_handle`, the function returns `ErrorPointerIsNull` without ever clearing the slot, breaking the documented ownership contract. The in-tree Swift wrapper pre-initializes the slot to nil so there is no current app-side bug, but other C/Swift consumers that follow the documented contract would observe a stale pointer. Move the null-write of `out_balances_json` ahead of the signer pointer check so the contract holds for every error path.

In `packages/rs-platform-wallet-ffi/src/tokens/burn.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/tokens/burn.rs:49-51: Null out_balances_json before validating signer_handle
  Same documented-contract gap as in `transfer.rs`: the doc comment states that on error the slot is set to null first, but `check_ptr!(signer_handle)` runs before `*out_balances_json = null_mut()`. A null signer with a non-null out slot returns early without clearing the slot, so a caller relying on the documented behavior could observe stale ownership state. Today's Swift wrapper avoids this by initializing the slot to nil, so this is defensive rather than an active app bug. Reorder so the out-slot is cleared first.
```
</details>

PR #3937, opened 2026-06-18
fix(sdk): default to latest protocol version instead of pinning testnet to v1

shumkov:

Updating an existing data contract through `js-dash-sdk` on testnet fails with:

> Can't update Data Contract `<id>` config: config version 0 is not supported, minimum version is 1

`js-dash-sdk` hardcoded `networkToProtocolVersion: testnet → 1`, so on testnet it built wasm-dpp at protocol version **1**. When it fetched the contract via `dataContract.createFromBuffer(...)`, wasm-dpp deserialized at v1 and **downgraded the contract's V1 config (`sized_integer_types`) to V0**. The update path (`contracts/update.ts`) only clones + `incrementVersion()` (the *contract* version, not the *config* version), so the resulting `DataContractUpdateTransition` carried a V0 config. Since protocol v12 the network enforces `config.min_version = 1` and rejects V0.

The contract itself is fine (V1 on-chain) — only the SDK was misreading it. Every non-testnet network already fell back to `getLatestProtocolVersion()`; only testnet was pinned to this stale value.

## What was done?

- Removed the stale `networkToProtocolVersion` map (its only entry was `testnet → 1`) and the now-dead lookup in `Platform.initialize()`.
- `initialize()` now falls back to `getLatestProtocolVersion()` for **all** networks when `driveProtocolVersion` isn't supplied via client options.
- Replaced the obsolete `it.skip` tests in `Platform.spec.ts` with two active ones: the protocol version is taken from options when provided, and defaults to the latest protocol version on testnet (the regression test).

Callers that need to pin a specific protocol version can still pass `driveProtocolVersion` in client options.

## How Has This Been Tested?

`Platform.spec.ts` (compiled via `tsconfig.mocha.json`, run with mocha):

- **Before the fix** (against `v3.1-dev`), the new test fails: `AssertionError: expected 1 to equal 12`.
- **After the fix**, all three Platform specs pass.

```
  Dash - Platform
    ✔ should provide expected class
    ✔ should use the protocol version passed through options
    ✔ should default to the latest protocol version on testnet
  3 passing
```

## Breaking Changes

None. Default protocol version on testnet changes from `1` to the latest supported by the bundled DPP; an explicit `driveProtocolVersion` option still overrides it.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Platform SDK protocol version initialization enhanced to always default to the latest available protocol version when not explicitly specified by user configuration. This replaces the previous static network-to-version mapping with dynamic version detection, preventing potential protocol downgrades that could interrupt contract updates.

### DIFF
--- a/packages/js-dash-sdk/src/SDK/Client/Platform/Platform.spec.ts
+++ b/packages/js-dash-sdk/src/SDK/Client/Platform/Platform.spec.ts
@@ -1,5 +1,5 @@
 import { expect } from 'chai';
-// import { getLatestProtocolVersion } from '@dashevo/wasm-dpp';
+import { getLatestProtocolVersion } from '@dashevo/wasm-dpp';
 import { Platform } from './index';
 import 'mocha';
 import Client from '../Client';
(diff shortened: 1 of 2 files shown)


thepastaclaw:
✅ Review complete (commit 78190a2e)

thepastaclaw (approved):
## Code Review

Verified PR #3937 at head 78190a2e2e1a43865e72c2a31f894f6de12c82be against the pinned worktree. The change removes the stale testnet protocol-version pin, preserves explicit driveProtocolVersion precedence, and adds focused regression coverage; I found no in-scope correctness or consensus-safety issues. A targeted `yarn workspace dash test:unit` run did not execute due to a local Node 25 loader EBADF failure before tests started.

PR #3941, opened 2026-06-19
fix(dashmate): use active_dkgs for safe DKG stop

thepastaclaw:

`dashmate stop --safe` / `dashmate restart --safe` used `quorum dkginfo.next_dkg` as the only DKG safety signal. `next_dkg` is a countdown until the next potential DKG cycle; it is not proof that this node is not currently participating in DKG.

When the countdown wraps at the start of a new cycle, `next_dkg` can become large again while the node is still in an active DKG session. In that case dashmate could treat the node as safe to stop during active DKG participation, risking a failed DKG and PoSe penalty.

## What was done?

- Added `isMasternodeSafeToStopDuringDkg(...)` as the shared safety predicate.
- Kept the existing imminent-DKG guard based on `next_dkg <= MIN_BLOCKS_BEFORE_DKG` / `6` blocks.
- When `active_dkgs > 0`, inspect `quorum dkgstatus` session data plus the current block height instead of treating the aggregate `active_dkgs` counter as permanently unsafe.
- Compare each tracked session's `llmqType` and `status.quorumHeight` against Dash Core's `dkgMiningWindowStart` values, so stale sessions no longer block once their active DKG phase has passed.
- Fail safe when active sessions exist but session data is unknown, malformed, empty, or from an unknown LLMQ type.
- Updated both the immediate non-`--safe` check and the `--safe` wait path to use the shared predicate.
- Intentionally kept `--safe` fail-closed: it only proceeds when Core reports a safe state.
- Added unit coverage for:
  - imminent DKG staying unsafe even when `active_dkgs === 0`;
  - active platform sessions blocking during their first 10 blocks;
  - platform sessions becoming safe at offset `10` even if `active_dkgs` is still non-zero;
  - longer non-platform DKG windows such as `llmq_400_60` still blocking at offset `15`;
  - unknown, malformed, or empty active session data failing safe;
  - `waitForDKGWindowPass` waiting through an active session and resolving once the session reaches the safe window.

## How Has This Been Tested?

- Direct behavioral validation with `node --input-type=module` against the changed dashmate modules:
  - verified imminent DKG remains unsafe even without active sessions;
  - verified no active sessions above the imminent threshold are safe;
  - verified active `llmq_test_platform` session offset `5` is unsafe;
  - verified `llmq_test_platform` session offset `10` is safe;
  - verified active `llmq_400_60` session offset `15` is unsafe and offset `20` is safe;
  - verified unknown and empty active session data fail safe.
- `node --check` on all changed source/test files.
- `git diff --check upstream/v3.1-dev..HEAD`
- Pre-PR code review gate: `ship`.

Formal Mocha/eslint commands were attempted but are currently blocked in this environment before repo code executes by Node 25 + Yarn PnP loader failure:

```text
Error: EBADF: bad file descriptor, fstat
    at tryStatSync (node:fs:386:25)
    at readFileSync (node:fs:442:17)
    at getSourceSync (node:internal/modules/esm/load:37:14)
...
Node.js v25.9.0
```

Blocked commands:

```text
yarn workspace dashmate test:unit packages/dashmate/test/unit/core/quorum/isMasternodeSafeToStopDuringDkg.spec.js packages/dashmate/test/unit/core/quorum/waitForDKGWindowPass.spec.js
yarn workspace dashmate lint
```

## Breaking Changes

None.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Improved masternode stop safety during DKG by adding rule-based inspection of active DKG windows and blocking unsafe stops.
  * Updated the DKG window wait behavior to poll at regular intervals until the node is confirmed safe to stop, reducing false “deadline” failures.
* **Tests**
  * Added unit tests covering safety validation, window-boundary behavior, and fail-safe handling for malformed or unknown DKG status data.


thepastaclaw:
✅ Review complete (commit a14c214f)

thepastaclaw (commented):
## Code Review

Centralizes the DKG safe-stop predicate into a shared `isMasternodeSafeToStopDuringDkg` helper that requires both `active_dkgs === 0` and `next_dkg > MIN_BLOCKS_BEFORE_DKG`, fixing the prior bug where a wrapped `next_dkg` countdown was read as safe while the node was still mid-session. The `--safe` wait loop adopts the same predicate and gains both a block-height deadline and a no-progress wall-clock backstop. No in-scope defects identified; all agent findings are pre-existing fragilities or speculative tuning notes outside this PR's scope.

thepastaclaw (commented):
## Code Review

Current-head review for `a9cf279c9c` after scope reduction.

The PR now stays focused on the active-DKG regression: both the immediate stop check and the `--safe` wait loop use the same predicate, `active_dkgs === 0 && next_dkg > MIN_BLOCKS_BEFORE_DKG`. This preserves the imminent-DKG guard while preventing the countdown-wrap case (`next_dkg` back to a large value while `active_dkgs > 0`) from unblocking a safe stop.

The previous wall-clock/no-progress exit hatch has been removed from the diff and from the PR body.

Review result: no significant issues found.

Validation reviewed:
- `node --check` on the changed files
- direct helper smoke checks for safe/unsafe combinations
- `git diff --check upstream/v3.1-dev..HEAD`
- formal dashmate unit command still blocked locally by Node 25 + Yarn PnP `EBADF` before repo code executes

Recommendation: ship.

packages/dashmate/src/core/quorum/isMasternodeSafeToStopDuringDkg.js
```diff
@@ -0,0 +1,122 @@
...
+    if (windowLength === undefined
+      || typeof quorumHeight !== 'number'
+      || !Number.isFinite(quorumHeight)) {
+      return false;
+    }
+
+    const offset = currentHeight - quorumHeight;
+    if (offset < 0 || offset < windowLength) {
+      return false;
+    }
```
thepastaclaw, line 137: **💬 Nitpick: `offset < 0` is redundant given the minimum window length of 10**

Every entry in `DKG_MINING_WINDOW_START_BY_LLMQ_TYPE` is at least 10, so `offset < windowLength` already covers any negative offset. The explicit `offset < 0 ||` is dead code under the current table. It is harmless and reads as defensive documentation of the negative-offset fail-safe, so leaving it as-is is fine; if you want the intent to be clearer, a short comment communicates the fail-safe motivation better than the redundant predicate. Non-blocking.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Re-review of head SHA 81550f0c (incremental + cumulative). Prior review at a9cf279c was clean and the delta has not introduced new issues. Only one in-scope nitpick from claude-general was validated: the `offset < 0` clause at line 116 is mathematically redundant given the minimum window value of 10, but harmless and arguably documentary. No blocking or suggestion-level findings; codex-general, both security auditors, and CodeRabbit all reported clean.

💬 1 nitpick(s)

packages/dashmate/src/core/quorum/isMasternodeSafeToStopDuringDkg.js
thepastaclaw, line 137: **💬 Nitpick: Redundant `offset < 0` clause in active-window check**

Every value in `DKG_MINING_WINDOW_START_BY_LLMQ_TYPE` is a positive integer (>= 10), and unknown llmqTypes are already rejected above (line 109). A negative offset therefore always satisfies `offset < windowLength`, making `offset < 0` unreachable as a distinct branch — both clauses return the same `false`. The JSDoc at lines 64–66 keeps the conceptual distinction (negative offset = inconsistent state vs. past-window = stale), which is worth preserving in the comment, but the disjunction can be simplified to a single condition. Drop the redundant clause, or, if the distinction is worth surfacing for logs/debugging, lift `offset < 0` into its own branch with an explanatory comment. Either resolution is acceptable; this is not a defect.

```suggestion
    if (offset < windowLength) {
      return false;
    }
```

thepastaclaw (commented):
## Code Review

Cumulative PR is a sound, conservative safe-stop fix that composes the imminent-DKG guard with per-session active-window inspection and fails closed on malformed input. Latest delta is test-only Sinon formatting and introduces no behavioral change. One pre-existing nitpick about a redundant `offset < 0` clause persists at head; not blocking.

💬 1 nitpick(s)

thepastaclaw:
@coderabbitai review

thepastaclaw:
Addressed the malformed `dkgInfo` feedback in `a14c214f`:

- safe-stop helpers now fail closed for missing/non-object `dkgInfo`, missing/non-numeric/non-finite/negative `active_dkgs`, and missing/non-numeric/non-finite/negative `next_dkg`;
- added malformed-`dkgInfo` coverage for those cases;
- validation: `node --check` on changed files, `git diff --check`, and inline smoke harness passed.

I intentionally did **not** apply the kebab-case rename nitpicks: `packages/dashmate/src` currently uses camelCase for JS filenames package-wide (I counted 235 `.js` files and zero kebab-case), so renaming only these new files would diverge from the local convention rather than align with it.

Formal yarn/mocha/eslint commands are still blocked locally by the known Node 25 + Yarn PnP `EBADF: bad file descriptor, fstat` failure before repo code loads.

packages/dashmate/src/core/quorum/isMasternodeSafeToStopDuringDkg.js
```diff
@@ -0,0 +1,141 @@
...
+ * @param {{ active_dkgs: number, next_dkg: number }} dkgInfo
+ * @return {boolean}
+ */
+export function shouldInspectDkgStatusForSafeStop(dkgInfo) {
+  if (!hasValidDkgInfoShape(dkgInfo)) {
+    return false;
+  }
+
+  return dkgInfo.active_dkgs > 0 && dkgInfo.next_dkg > MIN_BLOCKS_BEFORE_DKG;
+}
```
thepastaclaw, line 51: **💬 Nitpick: JSDoc on shouldInspectDkgStatusForSafeStop omits malformed-input behavior**

The function now returns false for null/undefined/non-object dkgInfo and for non-finite active_dkgs/next_dkg via hasValidDkgInfoShape, but the JSDoc still types dkgInfo as `{ active_dkgs: number, next_dkg: number }` and only documents the happy-path return. Future callers (waitForDKGWindowPass, stopNodeTaskFactory) could read this as 'expects a well-formed dkgInfo' and add redundant upstream guards, defeating the centralization this PR just introduced. Tighten the JSDoc to state the predicate is also the malformed-dkgInfo guard.

<sub>source: ['claude']</sub>

packages/dashmate/src/core/quorum/isMasternodeSafeToStopDuringDkg.js
```diff
@@ -0,0 +1,141 @@
...
+    if (windowLength === undefined
+      || typeof quorumHeight !== 'number'
+      || !Number.isFinite(quorumHeight)) {
+      return false;
+    }
+
+    const offset = currentHeight - quorumHeight;
+    if (offset < 0 || offset < windowLength) {
+      return false;
+    }
```
thepastaclaw, line 137: **💬 Nitpick: Redundant negative-offset check**

Every value in DKG_MINING_WINDOW_START_BY_LLMQ_TYPE is positive (10/12/20/42), and unknown llmqTypes return false at the windowLength === undefined branch above. So any negative offset already satisfies `offset < windowLength`, making `offset < 0` strictly redundant. The doc comment at lines 79-81 explicitly calls out the negative branch as fail-safe, so this is style-only — keep it if you want the intent to read explicitly at the call site, otherwise collapse to `offset < windowLength`.

```suggestion
    if (offset < windowLength) {
      return false;
    }
```

<sub>source: ['codex']</sub>

thepastaclaw (commented):
## Code Review

Latest push (a14c214f) adds fail-closed handling for malformed dkginfo plus tests. Verified at head: two nitpicks remain — stale JSDoc on shouldInspectDkgStatusForSafeStop (Claude) and the redundant offset<0 clause carried over from the prior review (Codex). No blocking issues; the session-window math, llmqType mapping, and fail-closed paths are consistent between the immediate stop check and the --safe wait loop.

💬 2 nitpick(s)

PastaPastaPasta (approved):
Conceptually this makes sense to me.

The --safe flag prior was waiting until the DKG round actually started and then said "time to restart" but this is a problem. 

Now after this, it should make sure we aren't too close to next DKG, but if we are in an active DKG it also won't restart, so we avoid the problem of waiting into a DKG session.

I've asked pastaclaw to test this interactively. Not sure if they'll be able to. I cannot for a bit, but I think this change does make sense, and the prior behavior had a problematic bug we should fix.

I quickly reviewed the code. It seems proper. I didn't review tests in depth.

thepastaclaw:
Follow-up on the interactive-test request:

I don’t currently have a live dashmate/devnet group running on this host to exercise `dashmate stop --safe` end-to-end against real Core RPC, so I did not claim a full interactive stop test.

I did re-check the current head (`a14c214f`) locally with a direct Node smoke harness over the safe-stop predicate:

- no active DKG + next DKG far away => safe;
- next DKG imminent => unsafe;
- active `llmq_test_platform` session inside the mining window => unsafe;
- lingering `active_dkgs` counter after the session window closes => safe;
- malformed `dkgInfo` => fail closed;
- the `shouldInspectDkgStatusForSafeStop` gate returns true only for the active/far-from-next-DKG case.

That lines up with the intended behavior. CI’s dashmate unit/lint checks are green on the PR; local full yarn/mocha still hits the known Node 25 + Yarn PnP `EBADF` loader issue before repo code executes.

shumkov (approved):

PR #3943, opened 2026-06-20
fix(swift-sdk): parse getIdentitiesTokenBalances NSNumber result

QuantumExplorer:

`SDK.getIdentitiesTokenBalances(identityIds:tokenId:)` almost always returned an empty `[String: UInt64]`.

The result from the FFI is decoded with `JSONSerialization`, which boxes numeric JSON values as `NSNumber`. The method cast each value with `value as? UInt64`, which never succeeds for an `NSNumber`, so every entry was silently dropped and callers saw an empty dictionary regardless of the actual on-chain balances.

The sibling method `getIdentityTokenBalances(identityId:tokenIds:)` already handles this correctly via `NSNumber.uint64Value`.

## What was done?

In `packages/swift-sdk/Sources/SwiftDashSDK/FFI/PlatformQueryExtensions.swift`, fixed the parse loop in `getIdentitiesTokenBalances` to read `.uint64Value` off `NSNumber`, matching the sibling method:

```swift
for (key, value) in json {
    if let balance = value as? NSNumber {
        balances[key] = balance.uint64Value
    }
}
```

The Rust FFI (`packages/rs-sdk-ffi/src/token/queries/identities_balances.rs`) emits `null` for identities that have never held the token. `JSONSerialization` decodes that as `NSNull`, which fails the `as? NSNumber` cast and is therefore skipped — mirroring how the single-identity query omits never-held tokens (no balance row). This is safe: the cast simply doesn't match, so there is no crash.

## How Has This Been Tested?

- Reviewed the three callers (`QueryDetailView`, `DiagnosticsView`) — all consume the dictionary only for display and none depend on the previous (empty) behavior, so the now-populated result is strictly an improvement.
- A full iOS xcframework build was intentionally **not** run because a parallel `build_ios.sh` was in flight at the time. The change is a localized, type-only fix identical in shape to the already-compiling sibling method, and `import Foundation` (providing `NSNumber`/`NSNull`) is already present in the file.

## Breaking Changes

None. Public API signature is unchanged; only the parsing behavior is corrected.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of null values in token balance retrieval, ensuring missing balance data is properly managed without errors.

### DIFF
--- a/packages/swift-sdk/Sources/SwiftDashSDK/FFI/PlatformQueryExtensions.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/FFI/PlatformQueryExtensions.swift
@@ -1375,11 +1375,16 @@ extension SDK {
         let result = dash_sdk_identities_fetch_token_balances(handle, identityIdsStr, tokenId)
         let json = try processJSONResult(result)
 
-        // Convert the result to [String: UInt64]
+        // Convert the result to [String: UInt64].
+        // JSONSerialization decodes numbers as NSNumber, so read .uint64Value
+        // (matching getIdentityTokenBalances above). The Rust FFI emits `null`
+        // for identities that have never held the token; those decode to NSNull
+        // and are skipped, mirroring how the single-identity query omits
+        // never-held tokens.
         var balances: [String: UInt64] = [:]
         for (key, value) in json {
-            if let balance = value as? UInt64 {
-                balances[key] = balance
+            if let balance = value as? NSNumber {
+                balances[key] = balance.uint64Value
             }
         }
 


thepastaclaw:
✅ Review complete (commit 32e9398f)

thepastaclaw (approved):
## Code Review

Narrow, correct Swift SDK fix to `getIdentitiesTokenBalances` JSON decoding. The prior `as? UInt64` cast against `JSONSerialization`'s `NSNumber` values always failed, returning an empty dictionary; the patch mirrors the sibling `getIdentityTokenBalances` decoder (NSNumber → uint64Value) and continues to skip `NSNull` entries for never-held tokens. No FFI ABI, consensus, security, or ownership concerns; all six agents converged on no in-scope findings and CodeRabbit had none.

PR #3942, opened 2026-06-20
feat(swift-example-app): document sum/average aggregation view (DOC-13/14)

QuantumExplorer:
## Issue being fixed or feature implemented

The document SUM/AVERAGE aggregation FFI (`dash_sdk_document_sum` / `dash_sdk_document_average`) shipped in #3935, but SwiftExampleApp had no way to invoke it — so QA tests **DOC-13** (sum) and **DOC-14** (average) were stuck at 🔌 SDK-only (FFI exists, no UI to drive it), unlike the COUNT tests DOC-10/11/12 which got a view in #3926.

This adds the missing app surface so those two tests become 🧪 builder-only (drivable in the simulator).

## What was done?

- **Swift SDK wrappers** (`PlatformQueryExtensions.swift`): added `SDK.documentSum(...)` and `SDK.documentAverage(...)` as thin bridges over the merged FFI, plus the `DocumentSumResult` / `DocumentAverageEntry` / `DocumentAverageResult` result types. These mirror the existing `documentCount` / `DocumentCountResult` precedent exactly (contract-handle fetch + defer destroy, `withOptionalCString` marshalling, `limit == 0` guard, `SDKError.fromDashSDKError`). The only new wire detail is the required non-empty `sum_property` argument. Per the Swift SDK thin-bridge rule, the wrappers only marshal — the FFI returns the raw `(count, sum)` pair and the `sum / count` division for display is done in the view.
- **New view** `SumAverageDocumentsView.swift`: a read-only query view (nothing signed/broadcast) mirroring `CountDocumentsView`. It has an op selector (Sum / Average), a required numeric sum-property field, optional `where` / `group_by` JSON, and renders the total + per-group rows (and the computed average for the avg op) or the platform error.
- **Wiring** (`TransitionCategoryView.swift`): a `NavigationLink` next to the Count link under **Settings → Platform State Transitions → Document**.
- **TEST_PLAN.md**: DOC-13/14 flipped 🔌 → 🧪 pointing at the new view; the `getDocuments` catalog row and Appendix B SDK-only gaps list updated accordingly.

No Rust source was changed; the FFI was already merged in #3935.

### Accessibility identifiers (for idb-driven QA)
[...]

PR #3944, opened 2026-06-20
fix(swift-sdk): seed testnet at the per-network protocol-version floor instead of pinning PV11

QuantumExplorer:

`SDK.init(network:platformVersion:)` in `SwiftDashSDK` hardcoded `platformVersion = 11` for `.mainnet`/`.testnet` when the caller passed `0` (the default). That pinned the SDK via `SdkBuilder::with_version(11)`, which disables auto-detect, so the SDK kept speaking the **V0 `getDocuments` wire** even though testnet has rolled forward to **Platform v3.1 (protocol version 12)**.

The visible symptom: V1-only features — document **sum/average aggregation** (QA tests DOC-13/14, added in #3942) — were rejected client-side with:

> `SDK misconfigured: select=Sum requires Platform v3.1+ (V1 documents wire); pin/upgrade to a v3.1+ network`

The `11` pin was a deliberate workaround (with a `TODO(platform-version-bump)`) from when the public networks were on pre-v3.1 drive-abci. Previously a construction-time floor clamp in `rs-sdk` (`initial = self.version.protocol_version.max(min_protocol_version(network))`, "applies to pinned and auto-detect SDKs alike") silently raised the sub-floor pin up to the network floor (testnet 12), so the app actually spoke V1. #3900 ("per-network protocol-version floor + version_pinned unification") changed the floor into a **seed-only default** for the unpinned case (its doc now reads "Not a runtime clamp"), so the stale `11` pin now sticks → V0 → V1 features break.

## What was done?

In the `platformVersion == 0` (default) path, stop mapping to a Swift-side per-network constant and pass `0` straight through to `config.platform_version`. The FFI `apply_version(builder, 0)` returns the builder unpinned, so `rs-sdk` seeds at the per-network `min_protocol_version` floor — **mainnet 11, testnet/devnet/regtest 12** — with **auto-detect on**, and ratchets upward as the network reports newer versions. The network→version decision now lives entirely in `rs-sdk`.

A non-zero `platformVersion` still pins via `with_version`, unchanged.

Net effect:
- **testnet** → PV12 (V1 wire) — sum/avg and other V1 features work, matching the live network.
- **mainnet** → PV11 (V0 wire) — correct until mainnet upgrades, then auto-detect ratchets up on its own.

13 insertions / 28 deletions in one file (`SDK.swift`); the deleted block is the hardcoded `switch network` version map and its stale rationale comment.

## How Has This Been Tested?

Rebuilt the iOS xcframework + app (`./build_ios.sh --target sim --profile dev`) and re-ran the SwiftExampleApp **Sum / Average Documents** view on testnet against a summable fixture contract (`metric` doc type, `documentsSummable: "value"`, seeded values 10/20/40/50):

- **Before** the change: both ops failed with the `requires Platform v3.1+ (V1 documents wire)` error.
- **After**: **sum = 120**, **average = 30.0000** — proof-verified, matching the seeded data (DOC-13/DOC-14 green).

Regular identity/contract/document reads continued to work on testnet after the change (the app is now on the V1 wire the live network expects).

## Breaking Changes

None. No public API signature change — same initializer, different default-version resolution. Not consensus-affecting.

## Checklist:

### DIFF
--- a/packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
@@ -247,14 +247,13 @@ public final class SDK: @unchecked Sendable {
   /// This is suitable for mobile applications where proof verification would be resource-intensive.
   ///
   /// `platformVersion`:
-  /// - `0` (default) — pick a network-appropriate version: PV 11 for the
-  ///   public networks still on pre-v3.1 drive-abci (mainnet, testnet),
-  ///   PV 12 (latest) for the leading-edge networks (devnet, regtest).
-  ///   This avoids the V0/V1 `getDocuments` wire-format mismatch that
-  ///   would otherwise make Platform queries fail with
-  ///   `"decoding error: could not decode data contracts query"` on the
-  ///   public networks until they roll forward. Override by passing an
-  ///   explicit non-zero value.
+  /// - `0` (default) — let the Rust SDK seed at the per-network minimum
+  ///   protocol version (mainnet 11, testnet/devnet/regtest 12) with
+  ///   auto-detect on, ratcheting up as the network reports newer
+  ///   versions. The per-network floor encodes the V0/V1 `getDocuments`
+  ///   wire format (mainnet floor 11 = V0 until mainnet upgrades;
+  ///   testnet floor 12 = V1), so this picks the right wire without a
+  ///   Swift-side network→version map.
   /// - non-zero — pin the SDK to this exact `PlatformVersion`.
   public init(network: Network, platformVersion: UInt32 = 0) throws {
     var config = DashSDKConfig()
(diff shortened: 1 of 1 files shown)


thepastaclaw:
✅ Review complete (commit af9568a5)

thepastaclaw (approved):
## Code Review

Narrow Swift SDK fix that drops a stale Swift-side network→protocol-version map and forwards platformVersion=0 to the FFI so rs-sdk seeds at the per-network min_protocol_version floor (mainnet 11, testnet/devnet/regtest 12) with auto-detect enabled, restoring V1-wire features on testnet. Both reviewers independently verified the FFI/rs-sdk semantics (apply_version(_,0) leaves the builder unpinned; non-zero still pins via with_version), found no in-scope issues, and the change is not consensus-affecting. No CodeRabbit findings.

lklimek (approved):

PR #3946, opened 2026-06-21
fix(swift-sdk): keychain privk storage

ZocoLini:
How Private keys were been stored didn't work well with the legacy keyring the swift tests where using without signing the app, what I do with this PR is updated the store/load logic to use the public key as the keyring entry label, being able to query for a single priv key instead of requesting a bunch oof items to later iterate over them, what doesn't work well in macOS file based keyring  

## Checklist:

## Summary by CodeRabbit

* **Refactor**
  * Improved the speed and efficiency of private-key retrieval by switching to a direct lookup keyed by the corresponding public key.
  * Updated how private keys are stored to associate them with the public key label for faster matching.
  * Retrieval now more reliably returns results (or nil) based on that direct match, without scanning unrelated stored items.

### DIFF
--- a/packages/swift-sdk/Sources/SwiftDashSDK/Security/KeychainManager.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/Security/KeychainManager.swift
@@ -649,6 +649,7 @@ extension KeychainManager {
             kSecValueData as String: privateKey,
             kSecAttrAccessible as String: kSecAttrAccessibleWhenUnlockedThisDeviceOnly,
             kSecAttrSynchronizable as String: false,
+            kSecAttrLabel as String: metadata.publicKey.lowercased(),
         ]
 
         if let accessGroup = accessGroup {
@@ -710,8 +711,8 @@ extension KeychainManager {
         var query: [String: Any] = [
             kSecClass as String: kSecClassGenericPassword,
             kSecAttrService as String: serviceName,
-            kSecMatchLimit as String: kSecMatchLimitAll,
-            kSecReturnAttributes as String: true,
+            kSecAttrLabel as String: publicKeyHex.lowercased(),
+            kSecMatchLimit as String: kSecMatchLimitOne,
             kSecReturnData as String: true,
         ]
         if let accessGroup = accessGroup {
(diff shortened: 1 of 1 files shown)


thepastaclaw:
✅ Review complete (commit 1804f3f1)

packages/swift-sdk/Sources/SwiftDashSDK/Security/KeychainManager.swift
```diff
@@ -720,29 +721,8 @@ extension KeychainManager {
...
-            // Case-insensitive hex compare — both producers downcase
-            // their hex but be defensive against future writers.
-            if metadata.publicKey.caseInsensitiveCompare(publicKeyHex) == .orderedSame {
-                return item[kSecValueData as String] as? Data
-            }
-        }
-        return nil
+        guard status == errSecSuccess else { return nil }
+        return result as? Data
     }
```
thepastaclaw, line 726: **🟡 Suggestion: Legacy keychain rows without kSecAttrLabel become invisible to this fallback**

Before this PR, `storeIdentityPrivateKey` never set `kSecAttrLabel`, so any row already on a device from a prior build has no label attribute. The new query pins `kSecAttrLabel == publicKeyHex.lowercased()`, so `SecItemCopyMatching` returns `errSecItemNotFound` for those legacy rows and the signer trampoline treats the secret as missing until the persister happens to re-`storeIdentityPrivateKey` and overwrite the row with a labeled version. The primary signing path (`retrieveKeyData(identifier:)` via `PersistentPublicKey.privateKeyKeychainIdentifier`) still works, so this only affects the public-key fallback used mid-registration / before the SwiftData row is written — but for users upgrading mid-flow it silently breaks a previously working path. If pre-existing user data is possible, either (a) fall back to the prior metadata scan when the labeled query misses, (b) run a one-shot `SecItemUpdate` migration over `identity_privkey.*` rows to backfill `kSecAttrLabel`, or (c) document that callers must trigger a re-persist after upgrade. If the SDK has not shipped to real users yet, call that out explicitly so this is closed knowingly.

thepastaclaw (commented):
## Code Review

Small, focused fix that switches identity-private-key lookup from a scan-and-decode loop to a direct `kSecAttrLabel` query, addressing macOS file-keyring iteration quirks. One legitimate upgrade-compatibility concern (legacy unlabeled rows become invisible to the fallback path) and two stale docstrings/comments that still describe the removed scan behavior. Speculative defense-in-depth concerns from the security passes do not represent real issues and are dropped.

<details>

```

In `packages/swift-sdk/Sources/SwiftDashSDK/Security/KeychainManager.swift`:
- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/Security/KeychainManager.swift:710-726: Legacy keychain rows without kSecAttrLabel become invisible to this fallback
  Before this PR, `storeIdentityPrivateKey` never set `kSecAttrLabel`, so any row already on a device from a prior build has no label attribute. The new query pins `kSecAttrLabel == publicKeyHex.lowercased()`, so `SecItemCopyMatching` returns `errSecItemNotFound` for those legacy rows and the signer trampoline treats the secret as missing until the persister happens to re-`storeIdentityPrivateKey` and overwrite the row with a labeled version. The primary signing path (`retrieveKeyData(identifier:)` via `PersistentPublicKey.privateKeyKeychainIdentifier`) still works, so this only affects the public-key fallback used mid-registration / before the SwiftData row is written — but for users upgrading mid-flow it silently breaks a previously working path. If pre-existing user data is possible, either (a) fall back to the prior metadata scan when the labeled query misses, (b) run a one-shot `SecItemUpdate` migration over `identity_privkey.*` rows to backfill `kSecAttrLabel`, or (c) document that callers must trigger a re-persist after upgrade. If the SDK has not shipped to real users yet, call that out explicitly so this is closed knowingly.
```
</details>

llbartekll (approved):

PR #3951, opened 2026-06-22
fix(sdk): ban rate-limited node for Envoy-advertised reset window

Claudius-Maginificent:
## Why this PR exists
- **Problem**: `rs-dapi-client` treats a gRPC `ResourceExhausted` (per-IP rate-limit / backpressure) the same as a node-down failure — it applies the `60s × e^ban_count` *health* ban to the address. Banning a healthy-but-throttled node doesn't shed load, it relocates it onto the remaining nodes.
- **What breaks without it**: Under a sustained per-IP rate limit, one high-concurrency client cascades to total failure: node A returns `ResourceExhausted` → banned 60s → its traffic shifts to B/C → they cross the limit → banned → … → `NoAvailableAddressesToRetry`, with zero server faults.
- **Blocking relationship**: none. Independent fix off `v3.1-dev`.

> **Redesign note**: an earlier revision of this PR rotated to a different node on `ResourceExhausted` (plus exclusion/backoff bookkeeping). Review feedback flagged that as over-engineered. This revision replaces it with a single, server-driven ban whose **duration** is dictated by the rate limiter itself. The intermediate commit history reflects that evolution; the net diff is the redesign only.

## What was done?
A single rate-limit ban mechanism, driven by the duration Envoy already advertises:

- **`CanRetry::rate_limit_ban_duration(&self) -> Option<Duration>`** (default `None`), implemented for `tonic::Status` (`packages/rs-dapi-client/src/transport/grpc.rs`): returns `None` unless the code is `ResourceExhausted`; otherwise parses the `ratelimit-reset` response-metadata header (whole seconds), filters out `0`/non-numeric, and clamps to `[1s, 600s]`. Delegated unchanged through `TransportError → DapiClientError / ExecutionError`.
- **`update_address_ban_status` dispatch** (`dapi_client.rs`): `Some(period)` → `AddressList::ban_for(address, period)`; `None` → the existing `ban_with_reason` exponential health-ban ladder (the **fallback is the normal ladder**, not a hardcoded default).
- **`AddressList::ban_for` / `AddressStatus::ban_for`** (`address_list.rs`): advances `banned_until` to `now + period` **only when that extends the current window** (max-semantics — a short rate-limit reset can't shorten a longer active ban under out-of-order completions on the shared `AddressList`), updates `ban_reason` only on extension, and raises `ban_count` to a floor of `1` (so `is_banned()` stays consistent with `banned_until`). The rate-limit path is **flat** — it never inflates the exponential ladder. Reinstatement is the existing `banned_until`-expiry path.
- **dashmate** — `LIMIT_RESPONSE_HEADERS_ENABLED` on the Lyft RLS container (which makes Envoy emit `RateLimit-Reset`, surfaced to the client as `ratelimit-reset` gRPC metadata) is now driven by a first-class config option **`platform.gateway.rateLimiter.responseHeaders.enabled`** (default **on**, since the ban-for-duration feature depends on the header) instead of a hard-coded `=true`. Spans the JSON schema, the base-config default, a config migration keyed at the next release `4.0.0-rc.3` (back-fills existing deployments, default on, so an upgrade never silently disables it), env rendering in `docker-compose.rate_limiter.yml`, and `docs/config/gateway.md` (documents the privacy trade-off so a cautious operator can switch it off).
- **Gateway filter reorder for browser-client parity** (`packages/dashmate/templates/platform/gateway/envoy.yaml.dot`): reordered the Envoy HTTP filters to `cors → grpc_web → ratelimit → router` and added `ratelimit-reset`/`ratelimit-limit`/`ratelimit-remaining` to CORS `expose_headers`, so **browser (grpc-web / wasm-sdk) clients** can read the over-limit `RateLimit-Reset` and apply the same per-node backoff as the native client (the ban logic in `transport/grpc.rs` already compiles for wasm). The over-limit response is a *local reply* from the ratelimit filter, and an Envoy local reply only traverses encoder filters positioned **above** the filter that generated it ([envoyproxy/envoy#11776](https://github.com/envoyproxy/envoy/issues/11776)); placing `cors` + `grpc_web` above `ratelimit` is what gets that reply CORS-exposed and grpc-web-framed on encode. **The native path is unaffected** — security-reviewed against Envoy source: `grpc_web` latches `is_grpc_web_request_=false` and early-returns for `application/grpc`, and `cors` no-ops without an `Origin` (and only ever *appends* headers). The one behavioural delta is that `OPTIONS` preflights now short-circuit at `cors` before the limiter (negligible; arguably a fix).
  - **Trailers-only over-limit reply for grpc-web** (same file, HCM `local_reply_config`): the reorder alone isn't enough — Envoy's gRPC-detection matches `application/grpc` only, so `rate_limited_as_resource_exhausted` never tags the grpc-web path and the browser would get a bare `HTTP 429` (which tonic-web maps to `Unavailable`, not `ResourceExhausted`, so the ban never fires). A `local_reply_config` mapper — scoped to `status == 429` **and** request header `x-grpc-web` present, so **JSON-RPC and native gRPC are untouched** — rewrites that reply to `HTTP 200` + `grpc-status: 8`. Because the reply is headers-only (empty body), `grpc_web` passes it through without reframing, so `grpc-status` and `ratelimit-reset` stay **co-located in the HTTP response headers**; `tonic-web-wasm-client` (→ `tonic 0.14.6`) then builds a `ResourceExhausted` Status with `ratelimit-reset` in `status.metadata()` via `Status::from_header_map`, firing the same `ban_for` the native client uses. No client-side code change.

### Review fixes folded in
- **ban_for never shortens an active ban** (max-semantics above) — a short `ratelimit-reset` arriving after a longer ban no longer truncates the window; `ban_reason` is preserved unless the window extends. Pinned by `test_ban_for_never_shortens_active_ban`.
- **Rate-limit debug log ordering** (`dapi_client.rs`) — the `banning … for Ns (from RateLimit-Reset header)` debug line now fires **inside** the `if banned` branch, so it can no longer precede/contradict the `unable to ban … not in the list anymore` trace when `ban_for` returns `false`.

Net effect: a throttled node is banned for **exactly** the window the server says it needs — no more, no less — instead of a fixed exponential health ban; genuine node ill-health still bans exactly as before.

## How Has This Been Tested?
- `cargo test -p rs-dapi-client`: **121 pass** (106 unit + 7 `rate_limit_ban` integration + 3 failover + 5 doc-tests), 0 failures. `cargo clippy -p rs-dapi-client -p dash-sdk -- -D warnings` clean; `cargo fmt` clean.
- **End-to-end coverage**: `rate_limited_node_banned_for_advertised_window_via_execute` drives the real `DapiClient::execute()` loop — a `ResourceExhausted` carrying `ratelimit-reset: 300` is banned for a ~300 s window (deliberately ≠ the ~60 s ladder rung), proving `ban_for` fires through the full client path, not just via a hand-built `Status`. It shares a fake-transport harness (`tests/common/mod.rs`) with the unimplemented-failover suite (refactored onto it, no duplication).
- **dashmate toggle**: schema ↔ base-config default ↔ migration ↔ env-var rendering name-consistency statically verified; `convertObjectToEnvs` confirmed to derive `PLATFORM_GATEWAY_RATE_LIMITER_RESPONSE_HEADERS_ENABLED` from the config path. The dashmate mocha suite was **not** run in the isolated worktree (pre-existing `ERR_REQUIRE_CYCLE_MODULE` in that runner, unrelated to these changes); the config-migration spec passes in the main tree.
- Coverage: header → `ban_for`; clamp edges `1→1s` / `600→600s` / `601→600s`; `0` / garbage / empty / missing header → ladder fallback (assertions bound `banned_until` to the `≈60s` first ladder rung, so they **fail** if a bad header is wrongly routed to `ban_for`); full delegation chain; the `ban_for` ladder-floor side-effect (fresh node `ban_count 0→1`); and a banned address re-entering rotation after its window expires.
- **Operational dependency**: the per-duration ban only engages where the rate limiter advertises `RateLimit-Reset`; the dashmate change above enables it. Where the header is absent, behaviour falls back to the unchanged exponential ban ladder.
- **✅ Browser-parity is live-verified** against the local devnet gateway (`dashmate config render` + gateway restart, then `curl`/`grpcurl`). Confirmed on the running gateway:
  - **grpc-web over-limit** → `HTTP 200`, `content-type: application/grpc-web+proto`, `grpc-status: 8`, `grpc-message: rate limited`, `ratelimit-reset: <n>` — **`grpc-status` + `ratelimit-reset` co-located in HTTP headers, body 0 bytes** (`xxd` confirms no trailer frame), `access-control-allow-origin` + `access-control-expose-headers` listing both. This is exactly the trailers-only shape `tonic 0.14.6` `Status::from_header_map` turns into `ResourceExhausted` + `ratelimit-reset` metadata (verified that path short-circuits on the header `grpc-status` without reading the body).
  - **native grpcurl over-limit** (regression gate) → still `ResourceExhausted` + `ratelimit-reset` in trailers.
  - **under-limit, both protocols** → unchanged (`local_reply_config` only rewrites Envoy *local replies*; normal proxied responses are untouched).
  - Remaining gap: an actual in-browser wasm client executing the path was **not** run here (no browser harness; the devnet cert is SHA-1, which browsers reject) — but the wire contract that client depends on is now verified end-to-end against the real Envoy build.

## Breaking Changes
None. Adds a `CanRetry` method with a default implementation; the genuine-failure ban path is unchanged.

## Checklist:
- [ ] I have added "!" to the title and described breaking changes (n/a — no breaking changes)

## Attribution

## Summary by CodeRabbit

* **New Features**
  * Rate-limit responses now apply a flat temporary ban matching the server-provided reset window (when available), improving timing accuracy.
  * Added a Rate Limiter option to emit `RateLimit-*` headers (including `RateLimit-Reset`) for reset-window support.
* **Bug Fixes**
  * Follow-up bans no longer shorten an active ban window; ban reasons are only updated when the ban is extended.
  * Expired bans can re-enter rotation without losing ban history.


thepastaclaw:
✅ Review complete (commit 5d45556c)

packages/rs-dapi-client/src/lib.rs
```diff
@@ -95,6 +95,19 @@ pub trait CanRetry {
...
+    ///
+    /// Rate-limit errors are retryable (see [`CanRetry::can_retry`]) but,
+    /// unlike genuine node ill-health, the node must NOT be banned: it is
+    /// healthy, just throttled. Banning it would shift its load onto the
+    /// remaining nodes and cascade into `NoAvailableAddressesToRetry`.
+    /// Instead the retry should rotate to a *different* node and leave the
+    /// throttled one in the pool. See `update_address_ban_status`.
+    fn is_rate_limited(&self) -> bool {
+        false
+    }
```
thepastaclaw: **🟡 Suggestion: Pin `is_rate_limited ⇒ can_retry` with a test or debug_assert**

The fix is load-bearing on the implicit contract that any rate-limited error is also retryable: `update_address_ban_status` only consults `is_rate_limited()` from inside the `if error.can_retry()` branch. If a future change ever marks `ResourceExhausted` non-retryable (e.g. a tweak to `Status::can_retry`), rate-limit handling silently stops firing and the error falls through to the no-op tail — neither ban nor rotation happens. Lock the invariant with either a unit test asserting `is_rate_limited() ⇒ can_retry()` across the gRPC code matrix, or a `debug_assert!(!self.is_rate_limited() || self.can_retry())` on the hot path, so the two predicates can't drift unnoticed.

<sub>source: ['claude']</sub>

packages/rs-dapi-client/src/lib.rs
```diff
@@ -95,6 +95,19 @@ pub trait CanRetry {
...
 
+    /// Returns true if this error is a rate-limit / congestion signal
+    /// (gRPC `ResourceExhausted`).
+    ///
+    /// Rate-limit errors are retryable (see [`CanRetry::can_retry`]) but,
+    /// unlike genuine node ill-health, the node must NOT be banned: it is
+    /// healthy, just throttled. Banning it would shift its load onto the
+    /// remaining nodes and cascade into `NoAvailableAddressesToRetry`.
+    /// Instead the retry should rotate to a *different* node and leave the
+    /// throttled one in the pool. See `update_address_ban_status`.
```
thepastaclaw: **💬 Nitpick: Doc overstates the rotation guarantee**

The doc states the retry "should rotate to a *different* node and leave the throttled one in the pool." The executor's documented fallback in `dapi_client.rs:761` explicitly re-selects the throttled node when `get_live_address_excluding(&tried)` returns `None` (single-node pool, or every live node already in `tried`). Rotation is best-effort, not guaranteed. A one-line caveat keeps the trait contract honest, e.g. "... and prefer rotating to a *different* node when one is available."

<sub>source: ['claude']</sub>

packages/rs-dapi-client/src/dapi_client.rs
```diff
@@ -364,6 +436,114 @@ mod tests {
...
+                address: Some(addr.clone()),
+            });
+            update_address_ban_status(&address_list, &result, &make_applied_settings(true));
+            let after = chrono::Utc::now();
+
+            let info = address_list.ban_info();
+            let entry = info
+                .iter()
+                .find(|i| i.uri.contains("3000"))
+                .expect("address present in ban info");
```
thepastaclaw: **💬 Nitpick: Match the ban-info test entry by exact URI**

The invariant test locates the address under test with `i.uri.contains("3000")`. The current fixture only inserts one address, so this is correct today, but it ties the assertion to a magic substring inside `mock_address()`. A future fixture adding another URI containing `3000`, or a rename of `mock_address`, would silently match the wrong entry. Match the exact URI from `addr` so the test is tied to the value it actually inserted.

```suggestion
            let info = address_list.ban_info();
            let addr_uri = addr.to_string();
            let entry = info
                .iter()
                .find(|i| i.uri == addr_uri)
                .expect("address present in ban info");
```

<sub>source: ['codex']</sub>

thepastaclaw (commented):
## Code Review

Narrow, well-tested fix that adds an `is_rate_limited` axis to `CanRetry`, suppresses banning for gRPC `ResourceExhausted`, and rotates the next retry off the throttled node via `get_live_address_excluding` plus a per-execution `tried` set. The default trait impl preserves prior behavior and the genuine-failure ban ladder is pinned by tests. Only minor docs/test-robustness nits — nothing blocking.

🟡 1 suggestion(s) | 💬 4 nitpick(s)

<details>

```

In `packages/rs-dapi-client/src/lib.rs`:
- [SUGGESTION] packages/rs-dapi-client/src/lib.rs:107-109: Pin `is_rate_limited ⇒ can_retry` with a test or debug_assert
  The fix is load-bearing on the implicit contract that any rate-limited error is also retryable: `update_address_ban_status` only consults `is_rate_limited()` from inside the `if error.can_retry()` branch. If a future change ever marks `ResourceExhausted` non-retryable (e.g. a tweak to `Status::can_retry`), rate-limit handling silently stops firing and the error falls through to the no-op tail — neither ban nor rotation happens. Lock the invariant with either a unit test asserting `is_rate_limited() ⇒ can_retry()` across the gRPC code matrix, or a `debug_assert!(!self.is_rate_limited() || self.can_retry())` on the hot path, so the two predicates can't drift unnoticed.
```
</details>

packages/rs-dapi-client/src/dapi_client.rs
```diff
@@ -561,11 +741,25 @@ impl DapiRequestExecutor for DapiClient {
...
+                // the nodes already tried in this execution. If excluding the
+                // tried set empties the live pool (e.g. a single-node pool, or
+                // every live node already tried), fall back to the non-excluding
+                // selection so a healthy-but-throttled node can still be retried
+                // rather than surfacing a spurious NoAvailableAddresses.
+                let Some(address) = self
+                    .address_list
+                    .get_live_address_excluding(&tried)
+                    .or_else(|| self.address_list.get_live_address())
+                else {
```
thepastaclaw: **💬 Nitpick: Fallback re-picks the throttled node when exclusion empties the live pool**

When `get_live_address_excluding(&tried)` returns `None`, the `or_else(|| self.address_list.get_live_address())` fallback can hand back an address already present in `tried` — i.e. the just-rate-limited node — and `tried.push(address.clone())` then appends a duplicate. With a fixed 10ms `retry_delay`, the remaining retry budget collapses onto the throttled node within tens of ms, which adds no value against a per-minute Envoy rate-limit. The behavior is intentionally documented and bounded by `max_retries`, so it is non-blocking. Two cheap improvements worth considering: (a) when the exclusion empties the pool, surface the raw `ResourceExhausted` immediately rather than spinning on the same node; or (b) switch `tried` to a `HashSet<Address>` so the contract "already attempted" stays exact (also makes the intent self-documenting).

<sub>source: ['claude']</sub>

packages/rs-dapi-client/src/dapi_client.rs
```diff
@@ -364,6 +436,114 @@ mod tests {
...
+            let upper = (banned_until - after).num_milliseconds() as f64 / 1000.0;
+            assert!(
+                lower >= period - 0.05,
+                "ban #{expected_count} window lower bound {lower}s < expected {period}s",
+            );
+            assert!(
+                upper <= period + 0.05,
+                "ban #{expected_count} window upper bound {upper}s > expected {period}s",
+            );
+        }
```
thepastaclaw: **💬 Nitpick: Wall-clock tolerance on ban-window assertion is thin for CI**

The invariant test snapshots `chrono::Utc::now()` before/after `update_address_ban_status` and asserts the ban window is within ±50 ms of `base × e^(n-1)`. On a loaded shared CI runner a 50 ms scheduling stall between the `before` snapshot and the ban write is plausible and would flip the lower-bound check. The assertion itself is exact, but the wall-clock dependence is what flakes. Consider widening the tolerance (e.g. to a few hundred ms) or recomputing the period from a directly observable `banned_at` field if one is exposed. Not blocking — current headroom is empirically fine — but worth a follow-up if these ever flake.

<sub>source: ['claude']</sub>

Claudius-Maginificent:
## Review fixes (adversarial review pass)

Addressed reviewer findings (CodeRabbit / thepastaclaw / internal review). All **MEDIUM**; two LOW left open by design.

**M1 — `is_rate_limited ⇒ can_retry` invariant locked** (thepastaclaw)
- Exhaustive property test over 17 gRPC codes × 4 `CanRetry` impls (non-vacuous) + `debug_assert` at the rotation decision in the typed `execute()` loop.

**M2 — rate-limit retry hardening**
- Retry now rotates to a *different* node (3-tier selection; never re-picks the just-throttled node while any alternative exists) — closes a gap where small/single-node pools re-hit the throttled node.
- Flat 10 ms retry replaced with capped exponential backoff (10 ms → 500 ms) + full jitter on the rate-limit path only; genuine-failure path unchanged. Shift amount clamped to avoid pathological window collapse. Server `RetryInfo` honoring left as a documented TODO (needs grpc-status-details decoding).

**M3 — rs-sdk rate-limit classification propagated**
- `Error::is_rate_limited()` now delegates to the inner `DapiClientError`, future-proofing against a latent SDK-layer ban regression. Zero behavior change today (no consumer reads it outside `if can_retry()`).

**Left open (LOW, by design):** ban-evasion blind spot for a node that *always* returns `ResourceExhausted`; `.read().unwrap()` panic on a poisoned routing lock (TODO breadcrumb added).

Commits: `396865fc`, `b8a8c329`. Tests: clippy `-D warnings` clean; 110 lib + 7 integration + 5 doc-tests pass.

packages/rs-dapi-client/src/dapi_client.rs
Claudius-Maginificent: Addressed in 396865fc/b8a8c329: the rate-limit retry path now uses a 3-tier selection that rotates off the just-throttled node whenever any alternative live address exists, plus capped exponential backoff (10 ms → 500 ms) with full jitter. `tried` is intentionally kept as a `Vec` — it is bounded by `max_retries` and the adversarial review agreed a `HashSet` is not warranted here. Leaving this thread open for maintainer confirmation.

packages/rs-dapi-client/src/lib.rs
Claudius-Maginificent: Partially addressed: M2 makes the retry genuinely rotate to a different live node when one exists (396865fc). The doc wording at this site still does not spell out the single-node-pool fallback (where the sole live address is reused with backoff) — that doc caveat was intentionally not added in this pass. Flagging as a known doc nit; leaving open.

packages/rs-dapi-client/src/dapi_client.rs
Claudius-Maginificent: Left unchanged in this pass — the ±50 ms wall-clock tolerance is a known timing tradeoff in this test. Not modified by the rate-limit fix; leaving open for the maintainer's call.

packages/rs-dapi-client/src/dapi_client.rs
Claudius-Maginificent: Acknowledged — the `i.uri.contains("3000")` lookup is brittle. It is test-only and unaffected by the production rate-limit changes, so it was left as-is in this pass. Leaving open.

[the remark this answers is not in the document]
Claudius-Maginificent: Same brittle test-lookup as the parallel thread — acknowledged, left as-is in this pass (test-only, no production impact). Leaving open.

packages/rs-dapi-client/src/lib.rs
thepastaclaw: **💬 Nitpick: CanRetry::is_rate_limited doc asserts unconditional rotation**

The doc says `Instead the retry should rotate to a *different* node and leave the throttled one in the pool.` The executor at dapi_client.rs:980-990 has a three-tier fallback whose final tier (`get_live_address()`) intentionally returns the just-tried address when no alternative is live (single-node pool, or every live node already in `tried`). Rotation is best-effort, not guaranteed. Soften the contract so future maintainers don't read it as a hard invariant.

```suggestion
    /// Rate-limit errors are retryable (see [`CanRetry::can_retry`]) but,
    /// unlike genuine node ill-health, the node must NOT be banned: it is
    /// healthy, just throttled. Banning it would shift its load onto the
    /// remaining nodes and cascade into `NoAvailableAddressesToRetry`.
    /// Instead the retry path should prefer a different live node while leaving
    /// the throttled one in the pool; if no alternative exists, bounded retries
    /// may reuse it after backoff. See `update_address_ban_status`.
```

thepastaclaw (commented):
## Code Review

Cumulative review at b8a8c329. The latest delta hardens the rate-limit handling: capped exponential backoff with full jitter, three-tier address selection (excluding tried → just-tried → fallback), debug_assert! invariants on both retry arms, and a property test pinning is_rate_limited() ⇒ can_retry() across all CanRetry implementors. Prior findings #1 and #4 are resolved; #2, #3, #5 are still valid nitpicks. One new latest-delta finding: the new AddressStatus ban-ladder test inherits the same thin 50ms wall-clock tolerance. No blocking issues.

💬 3 nitpick(s)

packages/rs-dapi-client/src/dapi_client.rs
```diff
@@ -364,6 +651,114 @@ mod tests {
...
+                address: Some(addr.clone()),
+            });
+            update_address_ban_status(&address_list, &result, &make_applied_settings(true));
+            let after = chrono::Utc::now();
+
+            let info = address_list.ban_info();
+            let entry = info
+                .iter()
+                .find(|i| i.uri.contains("3000"))
+                .expect("address present in ban info");
```
thepastaclaw: **💬 Nitpick: Ban-info invariant test matches by URI substring instead of the inserted address**

`info.iter().find(|i| i.uri.contains("3000")).expect("address present in ban info")` ties the assertion to a magic substring inside `mock_address()`. The test owns the `addr` it inserted; matching by exact URI is both safer and self-documenting. A future fixture adding another URI that happens to contain `3000`, or a rename of `mock_address`, would silently match the wrong entry.

```suggestion
            let entry = info
                .iter()
                .find(|i| i.uri == addr.to_string())
                .expect("address present in ban info");
```

packages/rs-dapi-client/src/dapi_client.rs
```diff
@@ -688,19 +1125,33 @@ impl DapiRequestExecutor for DapiClient {
...
                     Err(error) => {
+                        // Invariant lock (load-bearing): a rate-limited error MUST
+                        // be retryable, otherwise the rotation below never fires
+                        // and the rotate-don't-ban behavior silently dies. See the
+                        // `CanRetry` docs and the `test_rate_limit_implies_retryable_invariant`
+                        // property test.
+                        debug_assert!(
+                            !error.is_rate_limited() || error.can_retry(),
+                            "is_rate_limited() must imply can_retry(); rate-limit rotation depends on it"
+                        );
```
thepastaclaw: **💬 Nitpick: debug_assert! placement is inconsistent between the two retry arms**

The transport-create-error arm (line 1049) places the `debug_assert!(!cloned_error.is_rate_limited() || cloned_error.can_retry(), ...)` *before* `update_address_ban_status`, while the transport-call-error arm (line 1133) places the equivalent assert *after* `update_address_ban_status`. Functionally equivalent (the assert reads nothing the ban-status update mutates), but the two arms otherwise mirror each other. Aligning placement removes a small future-maintainer puzzle about whether the ordering was intentional.

<sub>source: ['claude']</sub>

packages/rs-dapi-client/src/dapi_client.rs
thepastaclaw: Resolved in `396865f` — _Wall-clock tolerance on ban-window assertion is thin for CI_ no longer present.

packages/rs-dapi-client/src/dapi_client.rs
thepastaclaw: Resolved in `396865f` — _Fallback re-picks the throttled node when exclusion empties the live pool_ no longer present.

packages/rs-dapi-client/src/lib.rs
thepastaclaw: Resolved in this update — _Doc overstates the rotation guarantee_ no longer present.

packages/rs-dapi-client/src/dapi_client.rs
thepastaclaw: Resolved in `396865f` — _Match the ban-info test entry by exact URI_ no longer present.

thepastaclaw:
Opened #3956 against this branch to address the Copilot note by matching the ban-info entry on the exact URI instead of a substring search.

lklimek:
@thepastaclaw review most recent commit,

lklimek (dismissed):

packages/rs-dapi-client/src/address_list.rs
```diff
@@ -103,6 +103,22 @@ impl AddressStatus {
...
+    /// `max(ban_count, 1)` so that `is_banned()` and `ban_info()` correctly
+    /// report the node as banned.  Side-effect: a previously-clean node
+    /// (ban_count 0) enters the ladder at floor 1, meaning its *next* genuine
+    /// health failure via [`AddressStatus::ban_with_reason`] uses
+    /// `60 s × e¹ ≈ 163 s` rather than the first-rung `60 s × e⁰ = 60 s`.
+    /// The counter resets to 0 on [`AddressStatus::unban`].
+    pub fn ban_for(&mut self, period: Duration, reason: Option<String>) {
+        self.banned_until = Some(chrono::Utc::now() + period);
+        self.ban_count = self.ban_count.max(1);
+        self.ban_reason = reason;
```
thepastaclaw: **🔴 Blocking: Rate-limit ban can shorten an active longer ban**

`ban_for` always replaces `banned_until` with `now + period`. Because request completions can arrive out of order for the same shared `AddressList`, a later `ResourceExhausted` response with a short `ratelimit-reset` value can shorten a longer health-ban or longer prior rate-limit ban that was already applied. That reintroduces the address before the existing ban window expires, so genuine failure bans are no longer preserved under overlapping requests.

```suggestion
pub fn ban_for(&mut self, period: Duration, reason: Option<String>) {
    let advertised_until = chrono::Utc::now() + period;

    if self
        .banned_until
        .map(|current_until| current_until < advertised_until)
        .unwrap_or(true)
    {
        self.banned_until = Some(advertised_until);
        self.ban_reason = reason;
    }

    self.ban_count = self.ban_count.max(1);
}
```

<sub>source: ['codex']</sub>

packages/rs-dapi-client/src/dapi_client.rs
```diff
@@ -187,7 +205,25 @@ pub fn update_address_ban_status<R, E>(
...
+                            // as banned, but the exponential ladder is not inflated.
+                            Some(period) => {
+                                tracing::debug!(
+                                    ?address,
+                                    ban_secs = period.as_secs(),
+                                    "rate-limited (ResourceExhausted): banning {address} \
+                                     for {}s (from RateLimit-Reset header)",
+                                    period.as_secs()
+                                );
+                                address_list.ban_for(address, period, reason)
```
thepastaclaw: **🟡 Suggestion: Rate-limit ban log is emitted before the ban succeeds**

The `Some(period)` branch logs that it is banning the address before calling `address_list.ban_for`. If the address was removed by an SML update, `ban_for` returns `false` and the code then logs that it was unable to ban the same address. Move the rate-limit-specific debug log behind the `banned` check so logs describe the state transition that actually happened.

<sub>source: ['claude']</sub>

thepastaclaw (changes requested):
## Code Review

New latest-delta verification found one in-scope blocking state-management bug: the new exact-window rate-limit ban can overwrite a longer active ban. Two non-blocking latest-delta issues remain around WASM/browser header visibility and misleading logging, and the carried-forward wall-clock tolerance test flake is still present. No CodeRabbit inline findings or comment IDs were supplied.

<details>

```

In `packages/rs-dapi-client/src/address_list.rs`:
- [BLOCKING] packages/rs-dapi-client/src/address_list.rs:116-119: Rate-limit ban can shorten an active longer ban
  `ban_for` always replaces `banned_until` with `now + period`. Because request completions can arrive out of order for the same shared `AddressList`, a later `ResourceExhausted` response with a short `ratelimit-reset` value can shorten a longer health-ban or longer prior rate-limit ban that was already applied. That reintroduces the address before the existing ban window expires, so genuine failure bans are no longer preserved under overlapping requests.

In `packages/dashmate/templates/platform/gateway/envoy.yaml.dot`:
- [SUGGESTION] packages/dashmate/templates/platform/gateway/envoy.yaml.dot:201: Expose rate-limit reset headers to browser clients
  The PR enables the rate limiter to emit `RateLimit-Reset`, and the Rust client parses it from tonic metadata, but browser/WASM clients go through fetch/CORS. Since Envoy's CORS `expose_headers` list does not include `ratelimit-reset` or the other `ratelimit-*` headers, browser clients cannot read the advertised reset window and will fall back to the exponential health-ban path even when Envoy sent the exact duration.

In `packages/rs-dapi-client/src/dapi_client.rs`:
- [SUGGESTION] packages/rs-dapi-client/src/dapi_client.rs:213-221: Rate-limit ban log is emitted before the ban succeeds
  The `Some(period)` branch logs that it is banning the address before calling `address_list.ban_for`. If the address was removed by an SML update, `ban_for` returns `false` and the code then logs that it was unable to ban the same address. Move the rate-limit-specific debug log behind the `banned` check so logs describe the state transition that actually happened.
```
</details>

packages/dashmate/docker-compose.rate_limiter.yml
```diff
@@ -42,6 +42,10 @@ services:
       - GRPC_MAX_CONNECTION_AGE=1h
       - GRPC_MAX_CONNECTION_AGE_GRACE=10m
       - GRPC_PORT=8081
+      # Emit RateLimit-Limit / RateLimit-Remaining / RateLimit-Reset response
+      # headers so rs-dapi-client can read the exact reset window and ban the
+      # node for that duration instead of the exponential health-ban ladder.
+      - LIMIT_RESPONSE_HEADERS_ENABLED=true
```
shumkov: I feel like I suggested the wrong direction. I mean, the ban system is correct, but exposing headers that are disabled by default and enabling them via env is risky. Could you please investigate what consequences and why it's disabled by default?

thepastaclaw (commented):
## Code Review

Carried-forward prior findings reconcile cleanly against HEAD 2c76a14: `ban_for` still unconditionally overwrites `banned_until` (blocking regression introduced by this PR), Envoy's CORS `expose_headers` still omits the `ratelimit-*` family (suggestion — browser/WASM clients silently fall back to the ladder), and the rate-limit path still emits a success-shaped debug log before calling `ban_for` (nitpick). No new latest-delta findings beyond what the agents and prior review already flagged. Quality-agent suggestions about `ban_count` semantics, constant placement, and timing tolerances are out-of-scope architectural/test-stability commentary and were dropped.

🟡 1 suggestion(s)

<details>

```

In `packages/dashmate/templates/platform/gateway/envoy.yaml.dot`:
- [SUGGESTION] packages/dashmate/templates/platform/gateway/envoy.yaml.dot:201: Expose `ratelimit-*` headers via CORS so browser/WASM clients can read them
  This PR teaches `rs-dapi-client` to read Envoy's `ratelimit-reset` header and pin an exact ban window from it. The WASM transport (`tonic_web_wasm_client`, used by `wasm-sdk`) reaches the gateway through browser `fetch`, which only surfaces non-safelisted response headers that the server enumerates in `Access-Control-Expose-Headers`. The current `expose_headers` list (line 201) does not include `ratelimit-limit`, `ratelimit-remaining`, or `ratelimit-reset`, so browser clients receive `ResourceExhausted` but cannot see the reset window and fall back to the exponential health-ban ladder — defeating the PR's stated improvement for the WASM SDK path. Native Rust clients are unaffected; the fix is one line on the gateway template.
```
</details>

packages/rs-dapi-client/src/dapi_client.rs
thepastaclaw: Resolved in `8a41c92` — _Ban-info invariant test matches by URI substring instead of the inserted address_ no longer present.

packages/rs-dapi-client/src/address_list.rs
thepastaclaw: Resolved in this update — _Rate-limit ban can shorten an active longer ban_ no longer present.

packages/rs-dapi-client/src/lib.rs
thepastaclaw: Resolved in `8a41c92` — _CanRetry::is_rate_limited doc asserts unconditional rotation_ no longer present.

packages/rs-dapi-client/src/dapi_client.rs
thepastaclaw: Resolved in this update — _Rate-limit ban log is emitted before the ban succeeds_ no longer present.

packages/rs-dapi-client/src/dapi_client.rs
thepastaclaw: Resolved in `8a41c92` — _debug_assert! placement is inconsistent between the two retry arms_ no longer present.

thepastaclaw (changes requested):
## Code Review correction

The previous exact-SHA automated review at `2c76a14f` was posted as `COMMENT`, but the verifier output still contains a blocking carried-forward finding. Requesting changes so the PR state matches the verified result.

🔴 1 blocking | 🟡 1 suggestion | ⚪ 1 nitpick

**Blocking:** `packages/rs-dapi-client/src/address_list.rs:116-120` — `ban_for` still unconditionally writes `banned_until = Some(now + period)`, so a later short `ratelimit-reset` response can shorten an existing longer health/rate-limit ban and return the node to rotation early. Preserve the longer existing/advertised deadline.

**Suggestion:** `packages/dashmate/templates/platform/gateway/envoy.yaml.dot:201` — expose `ratelimit-limit`, `ratelimit-remaining`, and `ratelimit-reset` in CORS so browser/WASM clients can read the reset window instead of falling back to the exponential ladder.

**Nitpick:** `packages/rs-dapi-client/src/dapi_client.rs:213-225` — the rate-limit path logs a success-shaped ban message before `ban_for` returns, producing contradictory logs when the address has been removed.

Run artifacts: `/Users/claw/.openclaw/workspace/worktrees/review-platform-3951/.review/run-1782229049-2c76a14f`.

packages/rs-dapi-client/src/dapi_client.rs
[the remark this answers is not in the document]
Claudius-Maginificent: Obsolete after the redesign. This PR was reworked from the rotate-on-`ResourceExhausted` approach to a single Envoy-driven ban: the throttled node is banned for exactly the `ratelimit-reset` window it advertises. The `i.uri.contains("3000")` invariant test (`test_invariant_genuine_failure_still_bans_with_unchanged_ladder`) was removed in that rework and no longer exists on HEAD `2c76a14`. Resolving as no longer applicable.

packages/rs-dapi-client/src/lib.rs
Claudius-Maginificent: Obsolete after the redesign. This PR was reworked from the rotate-on-`ResourceExhausted` approach to a single Envoy-driven ban: the throttled node is banned for exactly the `ratelimit-reset` window it advertises. The trait doc that described best-effort rotation was removed in that rework and no longer exists on HEAD `2c76a14`. Resolving as no longer applicable.

packages/rs-dapi-client/src/dapi_client.rs
Claudius-Maginificent: Obsolete after the redesign. This PR was reworked from the rotate-on-`ResourceExhausted` approach to a single Envoy-driven ban: the throttled node is banned for exactly the `ratelimit-reset` window it advertises. The substring-matched ban-info invariant test was removed in that rework and no longer exists on HEAD `2c76a14`. Resolving as no longer applicable.

packages/rs-dapi-client/src/dapi_client.rs
Claudius-Maginificent: Obsolete after the redesign. This PR was reworked from the rotate-on-`ResourceExhausted` approach to a single Envoy-driven ban: the throttled node is banned for exactly the `ratelimit-reset` window it advertises. The `get_live_address_excluding`/`tried` rotation fallback was removed in that rework and no longer exists on HEAD `2c76a14`. Resolving as no longer applicable.

packages/rs-dapi-client/src/dapi_client.rs
Claudius-Maginificent: Obsolete after the redesign. This PR was reworked from the rotate-on-`ResourceExhausted` approach to a single Envoy-driven ban: the throttled node is banned for exactly the `ratelimit-reset` window it advertises. The ±50ms wall-clock ban-window invariant test was removed in that rework and no longer exists on HEAD `2c76a14`. Resolving as no longer applicable.

packages/rs-dapi-client/src/lib.rs
Claudius-Maginificent: Obsolete after the redesign. This PR was reworked from the rotate-on-`ResourceExhausted` approach to a single Envoy-driven ban: the throttled node is banned for exactly the `ratelimit-reset` window it advertises. The `CanRetry::is_rate_limited` doc asserting rotation was removed in that rework and no longer exists on HEAD `2c76a14`. Resolving as no longer applicable.

packages/rs-dapi-client/src/dapi_client.rs
Claudius-Maginificent: Obsolete after the redesign. This PR was reworked from the rotate-on-`ResourceExhausted` approach to a single Envoy-driven ban: the throttled node is banned for exactly the `ratelimit-reset` window it advertises. The substring-matched ban-info invariant test was removed in that rework and no longer exists on HEAD `2c76a14`. Resolving as no longer applicable.

packages/rs-dapi-client/src/dapi_client.rs
Claudius-Maginificent: Obsolete after the redesign. This PR was reworked from the rotate-on-`ResourceExhausted` approach to a single Envoy-driven ban: the throttled node is banned for exactly the `ratelimit-reset` window it advertises. The `debug_assert!(is_rate_limited ⇒ can_retry)` on both retry arms was removed in that rework and no longer exists on HEAD `2c76a14`. Resolving as no longer applicable.

packages/rs-dapi-client/src/address_list.rs
Claudius-Maginificent: Fixed in `47f4160f2a`. `AddressStatus::ban_for` now uses max-semantics: it advances `banned_until` to `now + period` **only when that is later** than the current `banned_until`, so a short `ratelimit-reset` response can no longer shorten a longer active ban (a genuine health ban or a longer prior rate-limit ban) under out-of-order completions on the shared `AddressList`. `ban_reason` is updated only when the window extends; `ban_count` is still raised to `max(1)` unconditionally. New integration test `test_ban_for_never_shortens_active_ban` pins it: LONG→SHORT keeps the long window, SHORT→LONG extends, `ban_count ≥ 1` in both. Thanks for the catch.

packages/rs-dapi-client/src/dapi_client.rs
Claudius-Maginificent: Fixed in `47f4160f2a`. The rate-limit-specific `banning {address} for Ns (from RateLimit-Reset header)` debug log now fires **inside** the `if banned` branch, so it can no longer precede — and contradict — the `unable to ban … not in the list anymore` trace emitted when `ban_for` returns `false` (address removed by an SML update). The exact message text and `ban_secs` field are unchanged.

packages/dashmate/docker-compose.rate_limiter.yml
Claudius-Maginificent: Investigated — thanks for flagging, the caution is fair. Summary:

**What the flag exposes:** `LIMIT_RESPONSE_HEADERS_ENABLED=true` makes the Lyft RLS populate `RateLimit-Limit` / `RateLimit-Remaining` / `RateLimit-Reset` on the `RateLimitResponse`. Per the RLS source ([`src/service/ratelimit.go`](https://github.com/envoyproxy/ratelimit/blob/main/src/service/ratelimit.go)) these are emitted on **every** response (OK *and* over-limit), not only on 429. (This is the RLS-side flag — distinct from Envoy's own `enable_x_ratelimit_headers`, which emits the differently-named `x-ratelimit-*` and is *not* what the client parses.)

**Why it's off by default upstream:** it's an opt-in, backward-compatible addition whose default header names track a **non-final Internet-Draft** ([draft-ietf-httpapi-ratelimit-headers](https://datatracker.ietf.org/doc/html/draft-ietf-httpapi-ratelimit-headers), still `-11`, not an RFC), and that draft's own Security Considerations caution against disclosing capacity info to untrusted parties — so a conservative "off" is the natural upstream default.

**Actual risk for Dash:** low. The headers are **advisory only** — verified they do *not* change enforcement (the over-limit decision is computed independently of header emission). What they reveal is the caller's *own* per-IP rate-limit policy (limit / remaining / reset), which is public-by-design and already discoverable by probing (send until 429, watch when it clears). No secrets, backend capacity, topology, or other-tenant data leak.

**Recommendation:** not a blocker. The cleanest resolution to exactly your concern is to gate this behind a documented **dashmate config toggle** (default **on**, since the ban-for-duration feature depends on the header) rather than a hard-coded env var — operators get a visible switch instead of a buried `=true`. One sharper note: `rs-dapi-client` only consumes `ratelimit-reset` on the over-limit (`ResourceExhausted`) path, so emitting the headers on *successful* responses is pure disclosure surface with zero benefit to our client; restricting emission to over-limit responses would be strictly better, though the Lyft flag is all-or-nothing so that needs Envoy-layer shaping (nice-to-have, not required).

**Implemented in this PR** (commit `c49adb26e2`): the hard-coded `LIMIT_RESPONSE_HEADERS_ENABLED=true` is now driven by a first-class dashmate config option — `platform.gateway.rateLimiter.responseHeaders.enabled` (default **on**, since the ban-for-duration feature depends on the header). A config migration keyed at `4.0.0` back-fills the field (default on) for existing deployments so an upgrade never silently disables it, and `packages/dashmate/docs/config/gateway.md` documents the privacy trade-off so a cautious operator can switch it off. You now get the visible, documented switch instead of a buried `=true`.

Claudius-Maginificent:
## 🔬 Verification evidence — server-advertised ban window

Documenting the real-network + client-side verification behind this PR. It is deliberately honest about one **environmental** limitation (a SHA-1 devnet TLS cert) that prevents a single-process live end-to-end run. Both ends of the chain are proven directly; the only inch between them is standard `tonic` gRPC trailer transport.

### 1. Server side — gateway emits `ratelimit-reset` on `ResourceExhausted` ✅ *(live, grpcurl)*

With `LIMIT_RESPONSE_HEADERS_ENABLED=true` (now the dashmate `platform.gateway.rateLimiter.responseHeaders.enabled` toggle this PR adds), the running devnet gateway advertises the reset window. Observed against the live gateway via `grpcurl`:

```
-- Request 6 --
ratelimit-limit: 5, ratelimit-remaining: 0, ratelimit-reset: 60, x-envoy-ratelimited: true
*** RESOURCE_EXHAUSTED on request 6 ***
```

Baseline confirms the header is genuinely gated by the flag — against a gateway **without** it, `x-envoy-ratelimited: true` is present but `ratelimit-reset` is **absent**. The toggle does exactly what it claims.

### 2. Client side — rs-dapi-client bans for exactly the advertised window ✅ *(mock H2C)*

A mock H2C gRPC server returns `ResourceExhausted` with `ratelimit-reset: 60` in the response metadata. rs-dapi-client parses it and routes to `ban_for`:

```
rate-limited (ResourceExhausted): banning http://127.0.0.1:.../ for 60s (from RateLimit-Reset header)
ban_info: banned=true … ban expires in 60.0s    ✅ ban_for PATH CONFIRMED
```

The clamp `[1s, 600s]`, the `0`/garbage/missing-header fallback to the exponential ladder, and the never-shorten max-semantics are pinned by the `rate_limit_ban` integration suite (6 tests).

### 3. Honest gap — no single-process live e2e, and why

A direct rs-dapi-client → devnet run is **not achievable in this environment**. The devnet gateway certificate is `sha1WithRSAEncryption`, and rustls (rs-dapi-client's TLS stack) hard-rejects SHA-1 signatures with no `danger_accept_invalid_certs` escape hatch for this case:

```
invalid peer certificate: UnsupportedSignatureAlgorithmContext {
  signature_algorithm_id: [6, 9, 42, 134, 72, 134, 247, 13, 1, 1, 5, 5, 0] }
```

OID `1.2.840.113549.1.1.5` = `sha1WithRSAEncryption`. `grpcurl` connects because Go's TLS accepts SHA-1; rustls does not. This is an inherited property of the devnet cert, **not** a defect in this PR.

### Net

| Link in the chain | Proven | How |
|---|---|---|
| Gateway emits `ratelimit-reset` on over-limit | ✅ live | `grpcurl` vs running gateway (flag on); absent with flag off |
| Client parses header → `ban_for(window)` | ✅ | mock H2C metadata → `ban expires in 60.0s` |
| Client clamps / falls back to ladder / never shortens | ✅ | `rate_limit_ban` integration tests |
| Metadata transport (server → client) | standard | `tonic::Status` trailers — no new code path |

Both ends are proven directly; the only unproven inch is the standard gRPC trailer hop between them, which `tonic` already handles. Confidence: high.

packages/dashmate/src/config/configJsonSchema.js
```diff
@@ -690,8 +690,22 @@ export default {
...
+                    enabled: {
+                      type: 'boolean',
+                    },
+                  },
+                  additionalProperties: false,
+                  required: ['enabled'],
+                },
               },
-              required: ['docker', 'enabled', 'unit', 'requestsPerUnit', 'blacklist', 'whitelist', 'metrics'],
+              required: ['docker', 'enabled', 'unit', 'requestsPerUnit', 'blacklist', 'whitelist', 'metrics', 'responseHeaders'],
```
thepastaclaw, line 708: **🟡 Suggestion: responseHeaders is schema-required but the 4.0.0 migration won't fire for existing 4.0.0-rc.2 configs at the current package version**

`configJsonSchema.js:708` makes `responseHeaders` a required field, and `getConfigFileMigrationsFactory.js:1523` backfills it via the `'4.0.0'` migration. `ConfigFileJsonRepository.read()` calls `migrateConfigFile(configData, configFormatVersion, packageVersion)` (ConfigFileJsonRepository.js:48-52), and `migrateConfigFile` returns the input unchanged when `fromVersion === toVersion` (migrateConfigFileFactory.js:12-14). `package.json` is still `4.0.0-rc.2` at this commit, so any operator already stamped at `4.0.0-rc.2` will hit `fromVersion === toVersion`, skip the new migration, and then fail AJV validation on the missing `responseHeaders` block. Older configFormatVersions migrate cleanly because `semver.gt('4.0.0', '<older>')` is true. The commit message (`fix(dashmate): key responseHeaders migration at 4.0.0 not released rc.2`) shows this is a deliberate release-coordination decision: the migration is expected to fire when the package bumps to `4.0.0` final. Please confirm the release flow will bump `package.json` to a version `> 4.0.0-rc.2` at the same time this PR ships; otherwise add a no-op `4.0.0-rc.3` migration, drop `responseHeaders` from the schema `required` array, or bump the package version in this PR.

<sub>source: ['codex']</sub>

thepastaclaw (commented):
## Code Review

Incremental review reconciling against 2c76a14. Prior blocking finding (ban_for shortening) and prior nit (debug log before ban) are both FIXED in c49adb26 — see address_list.rs:133-144 and dapi_client.rs:218-226 plus their regression tests. Prior CORS expose_headers concern remains valid: the PR's WASM/browser path depends on the gateway exposing ratelimit-* headers, but envoy.yaml.dot was not touched. One new in-scope suggestion: the migration is keyed at 4.0.0 while package.json is still 4.0.0-rc.2, so rc.2-stamped configs hit a self-validation gap until the next version bump — the author has acknowledged this in the commit message but the coordination should be confirmed by maintainers.

🟡 2 suggestion(s)

<details>

```

In `packages/dashmate/templates/platform/gateway/envoy.yaml.dot`:
- [SUGGESTION] packages/dashmate/templates/platform/gateway/envoy.yaml.dot:201: Expose ratelimit-* headers via CORS so browser/WASM clients can read the reset window
  The PR's whole value-add is letting rs-dapi-client read `ratelimit-reset` and pin an exact ban window from it. On the native gRPC transport this works because metadata is passed through transparently, but the WASM transport (`tonic-web-wasm-client`) goes through the browser Fetch API, which only surfaces response headers that the server lists in CORS `expose_headers`. The current list (`custom-header-1,grpc-status,grpc-message,code,drive-error-data-bin,dash-serialized-consensus-error-bin,stack-bin`) omits `ratelimit-limit`, `ratelimit-remaining`, and `ratelimit-reset`, so browser/WASM clients will silently fall back to the exponential health-ban ladder even when Envoy/RLS advertised the precise window enabled by the new dashmate `responseHeaders.enabled` toggle. This was flagged in the prior review and is not addressed in c49adb26.

In `packages/dashmate/src/config/configJsonSchema.js`:
- [SUGGESTION] packages/dashmate/src/config/configJsonSchema.js:708: responseHeaders is schema-required but the 4.0.0 migration won't fire for existing 4.0.0-rc.2 configs at the current package version
  `configJsonSchema.js:708` makes `responseHeaders` a required field, and `getConfigFileMigrationsFactory.js:1523` backfills it via the `'4.0.0'` migration. `ConfigFileJsonRepository.read()` calls `migrateConfigFile(configData, configFormatVersion, packageVersion)` (ConfigFileJsonRepository.js:48-52), and `migrateConfigFile` returns the input unchanged when `fromVersion === toVersion` (migrateConfigFileFactory.js:12-14). `package.json` is still `4.0.0-rc.2` at this commit, so any operator already stamped at `4.0.0-rc.2` will hit `fromVersion === toVersion`, skip the new migration, and then fail AJV validation on the missing `responseHeaders` block. Older configFormatVersions migrate cleanly because `semver.gt('4.0.0', '<older>')` is true. The commit message (`fix(dashmate): key responseHeaders migration at 4.0.0 not released rc.2`) shows this is a deliberate release-coordination decision: the migration is expected to fire when the package bumps to `4.0.0` final. Please confirm the release flow will bump `package.json` to a version `> 4.0.0-rc.2` at the same time this PR ships; otherwise add a no-op `4.0.0-rc.3` migration, drop `responseHeaders` from the schema `required` array, or bump the package version in this PR.
```
</details>

thepastaclaw (commented):
## Code Review

Incremental delta (458a28e2) is test-only — adds a shared fake-transport harness and an end-to-end DapiClient::execute() rate-limit ban test. Prior fixes (ban_for no-shorten max-semantics, debug-log scoped inside the ban branch) remain in place and are well-covered. Two prior findings carry forward unaddressed: (1) Envoy CORS expose_headers still does not surface `ratelimit-*`, so browser/WASM clients cannot observe the reset window and silently fall back to the exponential ladder; (2) dashmate schema requires `responseHeaders` while package.json is still `4.0.0-rc.2`, so existing rc.2 configs hit `migrateConfigFile`'s `fromVersion === toVersion` early-return and fail schema validation. New agent findings were nitpicks/style and pre-existing spirit; dropped per scope discipline.

🟡 1 suggestion(s)

<details>

```

In `packages/dashmate/templates/platform/gateway/envoy.yaml.dot`:
- [SUGGESTION] packages/dashmate/templates/platform/gateway/envoy.yaml.dot:201: CORS `expose_headers` does not surface `ratelimit-*`, so browser/WASM clients can never read the reset window
  This PR's whole value is letting rs-dapi-client read `ratelimit-reset` (now emitted because dashmate enables `responseHeaders`) and ban a node for the server-advertised window instead of climbing the exponential ladder. Native tonic transports pass metadata through transparently, but the WASM path uses `tonic-web-wasm-client`, which goes through the browser Fetch API. Browsers only expose non-safelisted response headers that are listed in CORS `expose_headers`. The current list at envoy.yaml.dot:201 omits `ratelimit-limit`, `ratelimit-remaining`, and `ratelimit-reset`, so browser/WASM SDK clients will silently fall back to the exponential health-ban ladder even though the headers are emitted. Add the three RateLimit headers to `expose_headers` so the new ban-window behavior actually applies on the WASM transport.
```
</details>

packages/dashmate/src/config/configJsonSchema.js
thepastaclaw: Resolved in this update — _responseHeaders is schema-required but the 4.0.0 migration won't fire for existing 4.0.0-rc.2 configs at the current package version_ no longer present.

thepastaclaw (changes requested):
## Code Review correction

The automated verifier output for `458a28e2` contains a blocking carried-forward finding, but the first exact-SHA post went out as `COMMENT` after the inline poster omitted the non-diff carried-forward item. Requesting changes so the PR state matches the verified result.

🔴 1 blocking | 🟡 1 suggestion

**Blocking:** `packages/dashmate/src/config/configJsonSchema.js:708` — `responseHeaders` is now schema-required, but the backfill is keyed to the `4.0.0` migration while `packages/dashmate/package.json` is still `4.0.0-rc.2`. Existing rc.2-stamped configs hit `migrateConfigFile`'s `fromVersion === toVersion` early return, skip the new migration, and then fail AJV validation for the missing `responseHeaders` block. Bump the package version in lockstep, add an rc migration, or relax the schema until the final-version migration can run.

**Suggestion:** `packages/dashmate/templates/platform/gateway/envoy.yaml.dot:201` — expose `ratelimit-limit`, `ratelimit-remaining`, and `ratelimit-reset` in CORS `expose_headers` so browser/WASM clients can read the reset window instead of falling back to the exponential ladder.

Run artifacts: `/Users/claw/.openclaw/workspace/worktrees/review-platform-3951/.review/run-1782294645-458a28e2`.

packages/dashmate/src/config/configJsonSchema.js
Claudius-Maginificent: Good catch — verified and addressed in `b23008e7`.

You're right that `ConfigFileJsonRepository.read()` passes the package version as `toVersion`, so with `package.json` at `4.0.0-rc.2` an operator stamped at `4.0.0-rc.2` hits the `fromVersion === toVersion` early-return, skips the migration, and then fails AJV on the required `responseHeaders`.

Fix: the migration is now keyed at **`4.0.0-rc.3`** — the next release version (your option *b*) rather than the far-future `4.0.0` final. The release that ships this PR bumps `package.json` to `4.0.0-rc.3` (your option *a*, confirmed), at which point `semver.gt('4.0.0-rc.3', '<= 4.0.0-rc.2')` is true and the backfill fires for every rc.2-and-earlier config. Older `configFormatVersion`s already migrate cleanly today.

This mirrors the established precedent: the `'3.1.0'` migration was added while `package.json` was `3.1.0-dev.1` — keyed at the immediate next version, fired by the release bump. The remaining `rc.2 == rc.2` window is the pre-release dev state only. Keying at `rc.3` (a unique version) also avoids a future `'4.0.0'` object-key collision had another change later landed its own `4.0.0` migration.

packages/dashmate/configs/getConfigFileMigrationsFactory.js
```diff
@@ -1520,6 +1520,25 @@ export default function getConfigFileMigrationsFactory(homeDir, defaultConfigs)
...
+            // an operator's current version never fires. Backfill runs once the
+            // package bumps to rc.3 (mirrors the 3.1.0 migration added at 3.1.0-dev.1).
+            if (options.platform?.gateway?.rateLimiter
+              && typeof options.platform.gateway.rateLimiter.responseHeaders === 'undefined') {
+              options.platform.gateway.rateLimiter.responseHeaders = base.get('platform.gateway.rateLimiter.responseHeaders');
+            }
+          });
+
+        return configFile;
+      },
```
thepastaclaw, line 1541: **🔴 Blocking: Existing 4.0.0-rc.2 configs break: required responseHeaders is queued behind a 4.0.0-rc.3 migration but package.json is still rc.2**

This commit re-keys the responseHeaders backfill from `4.0.0` to `4.0.0-rc.3`, but `packages/dashmate/package.json` still reports `4.0.0-rc.2`. `ConfigFileJsonRepository.read()` calls `migrateConfigFile(configFileData, configFileData.configFormatVersion, version)` with `version` read from package.json; `migrateConfigFileFactory` early-returns at `fromVersion === toVersion` (`packages/dashmate/src/config/configFile/migrateConfigFileFactory.js:12-14`). So for an operator already on rc.2, fromVersion === toVersion === `4.0.0-rc.2` and the new migration never runs.

Meanwhile this PR makes `responseHeaders` a required field of the rate-limiter schema (`packages/dashmate/src/config/configJsonSchema.js:705,708`), and the per-config schema is enforced by `new Config(name, opts, skipValidation)` inside `ConfigFileJsonRepository.read()`. The result: any existing rc.2 deployment that pulls this PR before package.json bumps to rc.3 fails AJV validation on read with a missing `responseHeaders` error, even though the migration that would have backfilled it is sitting one version ahead.

The comment correctly documents the design intent ("runs once the package bumps to rc.3"), so the change is internally consistent — but the coupling to the rc.3 version bump must land in the same PR. Either bump `packages/dashmate/package.json` to `4.0.0-rc.3` in this PR, or temporarily drop `responseHeaders` from the schema `required` list so existing rc.2 configs still validate until the migration can run.

thepastaclaw (changes requested):
## Code Review

Two prior findings remain unresolved at b23008e7: (1) the dashmate responseHeaders schema/migration coordination gap — the migration is now keyed at 4.0.0-rc.3 while package.json is still 4.0.0-rc.2, so existing rc.2 configs short-circuit the migration runner and then fail AJV validation against the now-required responseHeaders field; (2) the CORS expose_headers list in envoy.yaml.dot still omits ratelimit-*, blocking wasm/browser clients from reading the very header this PR teaches the client to consume. The two prior Rust findings (ban_for no-shorten guard, debug-log placement inside the success branch) are confirmed fixed in this revision. Two additional review-agent nits this round (expect! interpolation, dead MIN_RATE_LIMIT_BAN_SECS constant) were dropped as low-value.

<details>

```

In `packages/dashmate/configs/getConfigFileMigrationsFactory.js`:
- [BLOCKING] packages/dashmate/configs/getConfigFileMigrationsFactory.js:1523-1541: Existing 4.0.0-rc.2 configs break: required responseHeaders is queued behind a 4.0.0-rc.3 migration but package.json is still rc.2
  This commit re-keys the responseHeaders backfill from `4.0.0` to `4.0.0-rc.3`, but `packages/dashmate/package.json` still reports `4.0.0-rc.2`. `ConfigFileJsonRepository.read()` calls `migrateConfigFile(configFileData, configFileData.configFormatVersion, version)` with `version` read from package.json; `migrateConfigFileFactory` early-returns at `fromVersion === toVersion` (`packages/dashmate/src/config/configFile/migrateConfigFileFactory.js:12-14`). So for an operator already on rc.2, fromVersion === toVersion === `4.0.0-rc.2` and the new migration never runs.

  Meanwhile this PR makes `responseHeaders` a required field of the rate-limiter schema (`packages/dashmate/src/config/configJsonSchema.js:705,708`), and the per-config schema is enforced by `new Config(name, opts, skipValidation)` inside `ConfigFileJsonRepository.read()`. The result: any existing rc.2 deployment that pulls this PR before package.json bumps to rc.3 fails AJV validation on read with a missing `responseHeaders` error, even though the migration that would have backfilled it is sitting one version ahead.

  The comment correctly documents the design intent ("runs once the package bumps to rc.3"), so the change is internally consistent — but the coupling to the rc.3 version bump must land in the same PR. Either bump `packages/dashmate/package.json` to `4.0.0-rc.3` in this PR, or temporarily drop `responseHeaders` from the schema `required` list so existing rc.2 configs still validate until the migration can run.

In `packages/dashmate/templates/platform/gateway/envoy.yaml.dot`:
- [SUGGESTION] packages/dashmate/templates/platform/gateway/envoy.yaml.dot:201: CORS expose_headers omits ratelimit-*, so wasm/browser clients can't read the reset window this PR depends on
  This PR's value is letting clients read `ratelimit-reset` (now emitted because dashmate enables responseHeaders) and ban a node for the server-advertised window instead of climbing the exponential ladder. Native tonic transports surface response metadata transparently, but browser/wasm clients consuming the gateway via grpc-web can only read non-CORS-safelisted response headers that Envoy explicitly enumerates in `expose_headers`. The current list (`custom-header-1,grpc-status,grpc-message,code,drive-error-data-bin,dash-serialized-consensus-error-bin,stack-bin`) does not include `ratelimit-limit`, `ratelimit-remaining`, or `ratelimit-reset`, so a wasm-sdk consumer of this same gateway sees a `429`/`ResourceExhausted` with no readable reset hint — defeating the per-window ban path for that transport.

  This is in scope: dashmate ships the envoy template, dashmate now enables emission of these headers by default, and the entire chain (`configJsonSchema` → `docker-compose.rate_limiter.yml` → envoy → client) is being changed in this PR for the purpose of consuming exactly that header. Add the three `ratelimit-*` headers to `expose_headers`.
```
</details>

packages/dashmate/configs/getConfigFileMigrationsFactory.js
lklimek: intentional, leave it as it is now

shumkov (dismissed):

thepastaclaw (changes requested):
## Code Review

Cumulative review: rs-dapi-client rate-limit-driven flat-ban path is well-implemented and well-tested; the latest delta (gateway filter reorder, grpc-web trailers-only 429 rewrite, expose_headers, .env.*.bak gitignore) is narrow and live-verified. Prior CORS exposure gap is fixed at envoy.yaml.dot:259. One prior blocking issue remains: the responseHeaders schema requirement is shipped with a migration keyed at 4.0.0-rc.3 while package.json still reports 4.0.0-rc.2, so existing rc.2 configs will fail validation until the package version is bumped.

🚫 Requesting changes: 1 blocking carried-forward finding remains valid on `5d45556c`.

### Carried-forward finding still valid
- **Existing 4.0.0-rc.2 configs break: required responseHeaders is queued behind a 4.0.0-rc.3 migration, but package.json is still 4.0.0-rc.2** — `packages/dashmate/configs/getConfigFileMigrationsFactory.js:1523-1541`

The responseHeaders backfill is re-keyed to 4.0.0-rc.3, but `packages/dashmate/package.json:3` still reports `4.0.0-rc.2`. `ConfigFileJsonRepository.read()` calls `migrateConfigFile(configFileData, configFileData.configFormatVersion, version)` with `version` from package.json, and `migrateConfigFileFactory.js:12` early-returns when `fromVersion === toVersion`. For an operator on an existing rc.2 config, the rc.3 migration never runs — yet `configJsonSchema.js:708` already lists `responseHeaders` in the required array, so the config fails schema validation on load. The in-code comment at lines 1530-1533 documents the intent to wait for a future rc.3 bump, but the schema requirement is live now. Either bump dashmate to 4.0.0-rc.3 in this PR (so the migration runs and backfills `responseHeaders` for existing rc.2 configs), or drop `responseHeaders` from the `required` array until the rc.3 release cuts.

I am not re-opening the same inline thread here because the finding has already been raised on this PR; this review records exact-SHA coverage and keeps the blocking status attached to the current head.

### Prior findings reconciled
- FIXED: CORS expose_headers omits ratelimit-*, so wasm/browser clients can't read the reset window this PR depends on — Fixed at packages/dashmate/templates/platform/gateway/envoy.yaml.dot:259 — expose_headers now includes ratelimit-reset, ratelimit-limit, and ratelimit-remaining, allowing tonic-web-wasm-client to surface the reset metadata.

CodeRabbit inline findings: 0; no CodeRabbit reactions requested.

packages/dashmate/configs/getConfigFileMigrationsFactory.js
thepastaclaw, line 1541: **🟡 Suggestion: responseHeaders backfill keyed at 4.0.0-rc.3 leaves a validation gap until the version bump lands**

The schema marks `platform.gateway.rateLimiter.responseHeaders` as required (configJsonSchema.js:708) and the base config defaults `responseHeaders.enabled: true` (getBaseConfigFactory.js:244), but the backfill migration is keyed at `4.0.0-rc.3` while `packages/dashmate/package.json` is still `4.0.0-rc.2`. Per `migrateConfigFileFactory`, a migration only runs when `semver.gt(version, fromVersion)`, so an existing rc.2 deployment running this PR's dashmate (still reporting rc.2) skips the new migration and then fails schema validation on the missing required field. The added comment now explicitly documents the intent — mirror the 3.1.0/3.1.0-dev.1 pattern and let the backfill fire once dashmate bumps to rc.3 — which is fine as long as the release process couples this PR with the rc.3 version bump. If they could ship decoupled (e.g. a dashmate command run between merge and the rc.3 cut), either bump the package version in the same release cycle as this PR or make `responseHeaders` non-required in the schema with a defaulted reader. Carried forward from prior review at reduced severity now that the deferred-intent rationale is in-tree.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Cumulative review: prior-2 (CORS expose_headers) is FIXED at envoy.yaml.dot:259, which lists ratelimit-reset/limit/remaining so browser/wasm clients can read the ban window. Prior-1 (responseHeaders schema/migration vs package.json) is technically still open but now explicitly documented as intentionally deferred to the next version bump (mirroring the 3.1.0 pattern); carried forward as a suggestion. The new grpc-web local_reply mapper plus filter reorder are correctly scoped (gated by x-grpc-web + HTTP 429 + rateLimiter.enabled) and the trailers-only rewrite preserves grpc-status:8 and ratelimit-reset for tonic-web's Status::from_header_map. No new in-scope blockers; remaining agent items were nitpick-grade operational hardening.

🟡 1 suggestion(s)

<details>

```

In `packages/dashmate/configs/getConfigFileMigrationsFactory.js`:
- [SUGGESTION] packages/dashmate/configs/getConfigFileMigrationsFactory.js:1523-1541: responseHeaders backfill keyed at 4.0.0-rc.3 leaves a validation gap until the version bump lands
  The schema marks `platform.gateway.rateLimiter.responseHeaders` as required (configJsonSchema.js:708) and the base config defaults `responseHeaders.enabled: true` (getBaseConfigFactory.js:244), but the backfill migration is keyed at `4.0.0-rc.3` while `packages/dashmate/package.json` is still `4.0.0-rc.2`. Per `migrateConfigFileFactory`, a migration only runs when `semver.gt(version, fromVersion)`, so an existing rc.2 deployment running this PR's dashmate (still reporting rc.2) skips the new migration and then fails schema validation on the missing required field. The added comment now explicitly documents the intent — mirror the 3.1.0/3.1.0-dev.1 pattern and let the backfill fire once dashmate bumps to rc.3 — which is fine as long as the release process couples this PR with the rc.3 version bump. If they could ship decoupled (e.g. a dashmate command run between merge and the rc.3 cut), either bump the package version in the same release cycle as this PR or make `responseHeaders` non-required in the schema with a defaulted reader. Carried forward from prior review at reduced severity now that the deferred-intent rationale is in-tree.
```
</details>

Claudius-Maginificent:
**Live devnet verification**

Verified this change end-to-end against a running local devnet gateway already serving the committed Envoy config (filter reorder + `local_reply_config` + `expose_headers`). The `ratelimit-reset`/`-limit`/`-remaining` headers are exposed on both wires: native gRPC carries them in response metadata (trailers, with `ResourceExhausted`, when over-limit), and grpc-web gets them via CORS `expose_headers` plus the `local_reply_config` that rewrites the over-limit reply to `HTTP 200` + `grpc-status: 8` + `ratelimit-reset` co-located in the HTTP headers with a 0-byte body — exactly the trailers-only shape the wasm client parses. Running the rs-sdk test-vector generation through the modified Envoy showed no regression to normal proxied SDK traffic. A concurrent batch even captured `rs-dapi-client`'s own `ban_for` logic firing live — banning the node from the `ratelimit-reset` it read off the Envoy gRPC metadata — confirming the feature works with zero client-code changes (Envoy-only).

shumkov (approved):

PR #3828, opened 2026-06-09
refactor(platform-wallet)!: retire in-band pool snapshot; hardcode core UTXO account_index=0

Claudius-Maginificent:
## Why this PR exists

- **Problem**: On a genesis rescan, funded wallets entered an unterminated fatal-flush loop and their balance stayed stuck at zero (observed in production: 173× `flush failed fatally` across 7 wallets in one boot).
- **What breaks without it**: The SQLite persister resolved each core UTXO's owning account by looking its address up in a `core_derived_addresses` map. During a rescan, SPV matched historical UTXOs at registered addresses **before** the corresponding derive events landed → the lookup missed → the flush was classified fatal → the entire buffered changeset was dropped and re-emitted → infinite loop. Funded wallets showed 0 balance and never recovered.
- **Blocking relationship**: stacked on `feat/platform-wallet-rehydration`.

## What was done?

The core wallet uses only the **default funds account (index 0)** in the supported product (Platform, dash-evo-tool). Rather than resolve `address → account_index` at persist time (the race-prone lookup), the storage writer now records a constant `core_utxos.account_index = 0`. With no lookup there is no race and no fatal loop — and a changeset goes back to carrying **diffs, not a recurring full-pool snapshot**.

This retires the entire in-band pool-snapshot + address→account lookup machinery:
- **Removed**: the `core_derived_addresses` and `account_address_pools` tables; `ACCOUNT_INDEX_BY_ADDRESS_SQL`; `apply_pools`; the derived-address INSERT block in `core_state::apply()`; and the apply-time emitter-contract guard.
- **Kept**: the `account_address_pools` changeset *field* as the one-time **registration seed** for the iOS address registry (the only channel that can seed non-ECDSA pool addresses); and the `addresses_derived` delta, which still feeds the Swift `PersistentCoreAddress` rows **before** each UTXO changeset, so the cascade-delete chain stays intact.
- **Single-account safety (no fund loss)**: if a core UTXO ever belongs to a non-default account (multi-account / CoinJoin), `core_bridge` emits a `tracing::warn!` but **still persists** the UTXO under index 0 — it never skips or fails. Worst case is approximate per-account grouping (cosmetic, multi-account-only); the total balance is always correct.

**Also in this PR — sync-manager restart safety:**
[...]

PR #3953, opened 2026-06-22
feat(platform-wallet)!: layered opt-in secret protection (Tier-2 password envelope)

Claudius-Maginificent:
## Why this PR exists

- **Problem**: dash-evo-tool stores wallet secrets — multi-key/seed wallets and single private keys. By default they belong in the OS keychain (or, for the file backend, a passphrase-encrypted vault), but the user had no way to add an **extra password to individual critical objects**, and the SecretStore had a latent strip/downgrade weakness.
- **What breaks without it**: an attacker with write access to the secret backend replaces a password-protected secret with a well-formed *unprotected* blob carrying a chosen seed. If protection is inferred from the stored blob, the password prompt is bypassed and the attacker's seed is returned to the wallet → **funds redirection**.
- **Blocking relationship**: stacked on **#3828** (`fix/wallet-core-derived-rehydration`).

## What was done?

A **layered, opt-in** secret-protection scheme in `packages/rs-platform-wallet-storage`:

- **Tier-1 (baseline, always on)**: secrets live in the OS keychain (`SecretStore::Os`) or a passphrase-encrypted file vault (`EncryptedFileStore`); a real passphrase is required — blank is rejected (`BlankPassphrase`). `open_unprotected` is the explicit keyless door.
- **Tier-2 (opt-in, per object)**: a password envelope (**Argon2id** + **XChaCha20-Poly1305**) layered *above* the SecretStore, backend-independent. Applies to seed wallets and single private keys alike.
- **Envelope wire format** (`secrets/wire/`): bincode-encoded `Envelope { version, payload: Payload }` where `Payload` is `Unprotected(Vec<u8>) | Password { kdf, salt, nonce, ciphertext }`. The encoder/decoder share `WIRE_CONFIG = standard().with_big_endian().with_no_limit()` for AAD/encode; the decoder uses a derived `DECODE_CONFIG` adding `with_limit::<MAX_SECRET_LEN + MAX_ENVELOPE_OVERHEAD>()` so an attacker-controlled length prefix can't drive multi-GiB `Vec::with_capacity` before any tag check.
- **AAD bind** (three separate structs, all bincode-encoded with `-v2` domain tags for byte-disjoint cross-context defense): `Tier2Aad` (domain ‖ envelope_version ‖ scheme_discriminant ‖ kdf ‖ salt ‖ wallet_id ‖ label), `EntryAad`, `VerifyAad`. The KDF cost ceiling is enforced **before** derivation, and per-read also gated against `KdfParams::default_target()` on **both `m_kib` and `t` axes** so an inflated header bounded by the global `ARGON2_MAX_*` ceiling can't drive 16× memory or 5.3× CPU per protected read.
- **★ Strict fail-closed read (the keystone)**: `get_secret(service, label, Option<&SecretString>)`. A `Some(pw)` read of a scheme-0 / malformed blob returns `ExpectedProtectedButUnsealed` — it never returns the bytes. The "is this object protected?" expectation lives **only** in the caller's `Some`/`None` (its trusted DB), and is **never** inferred from the stored blob. This is what closes the strip/downgrade injection.
- **Symmetric blank-password guard**: `unwrap` rejects `Some(SecretString::empty())` with `BlankPassphrase` BEFORE any KDF/AEAD work — mirroring the existing wrap-side `pw.is_blank()` guard. Closes a backend-write attacker plant-under-blank-pw injection path.
- **Trailing-byte defense**: `bincode::decode_from_slice` returns `(value, consumed)`; `unwrap` asserts `consumed == blob.len()` and fails closed otherwise, so a truncation/extension probe surfaces as `Corruption`.
- **Write API**: `set_secret(.., password)` and `reprotect(current, new)` (add / change / remove a password, crash-safe same-slot rewrite). `set`/`get` are reimplemented as non-breaking `.., None` wrappers. **`delete` returns `Result<bool>`** (`true` = removed, `false` = was absent) and **`reprotect` returns `Err(NoEntry)`** on absent — surfaces upstream `keyring-core::Error::NoEntry` instead of swallowing it, and closes a downgrade-detection blind spot where a silent `reprotect` no-op left the caller's trusted protection-status record disagreeing with the backend.
- **Serde gating**: `SecretString` gains `Deserialize` + `JsonSchema` behind dedicated **default-OFF** features `secret-serde` / `secret-schemars` (no `Serialize`; schema leaks no value/min/max/pattern), plus an always-on `is_blank()`.
- No bespoke crypto — reuses the existing `crypto::*` primitives, `SecretBytes`/`SecretString`, the vault's atomic-write path. Wire format follows the project's standing `bincode = "=2.0.1"` + `standard().with_big_endian().with_no_limit()` convention (used by `rs-platform-serialization`, `rs-dpp`, `rs-drive-proof-verifier`, etc.).

## How Has This Been Tested?

- **`cargo test -p platform-wallet-storage --lib` → 215 passed / 0 failed / 0 ignored.**
- **`cargo test -p platform-wallet-storage --all-targets`** (`--lib` + 7 integration binaries + 3 doctests) all green.
- **`cargo check -p platform-wallet -p platform-wallet-ffi --all-targets`** clean — no caller-side API change despite the wire-format refactor.
- **★ Non-vacuous strip-injection regression**: `Some(pw)`+scheme-0 read fails closed AND the same blob would decode to the attacker seed under `None` — proving the strict rule (not malformation) blocks it. Symmetric tests on BOTH the file vault AND the OS-keychain arms.
- **Golden byte-vector tests** pin the exact wire output for scheme-0 and scheme-1 (deterministic salt/nonce via a `#[cfg(test)]` `seal_with_nonce` seam) — any future bincode-config drift trips them before it can corrupt stored data.
- **Property test** (`proptest = "1"`, already in the workspace) `prop_single_byte_flip_never_yields_plaintext` runs a single-byte flip across every field within its 60-second budget.
- **Adversarial fuzz** of `unwrap` — `decode_rejects_trailing_garbage`, `kdf_m_kib_ceiling_fires_before_derive`, `kdf_t_ceiling_fires_before_derive`, scheme-0/1 dispatch quadrants, AAD-bind one-byte flip per field.
- **AAD domain-separation tests** prove `Tier2Aad`, `EntryAad`, `VerifyAad` are pairwise byte-disjoint.
- **Independent QA** (5-specialist wave on the layered-protection feature itself): unanimous PASS, 0 findings above LOW; L-1 strip/downgrade, L-2 KDF-DoS, and L-3 relocation confirmed closed within the library's reach.
- **Independent QA on the bincode/AAD refactor** (`security-engineer-smythe` + `qa-engineer-marvin` + `project-reviewer-adams` parallel panel): 1 HIGH + 4 MED + 13 LOW — all MEDIUM+ items folded into commit `c4af266e4e`. Final verdict ACCEPT.
- `cargo fmt --all -- --check` clean; `cargo clippy -p platform-wallet-storage --all-targets -- -D warnings` zero warnings; `cargo audit` clean on the crypto stack (RUSTSEC-2025-0141 `bincode unmaintained` is an INFO-level project-wide notice, no exploit).

## Breaking Changes

- `open` / `rekey` now **reject a blank passphrase** (`BlankPassphrase`).
- `SecretStore::delete` returns `Result<bool, SecretStoreError>` instead of `Result<(), SecretStoreError>`. `.delete(...)?;` still compiles and silently discards the bool for callers that don't care.
- `SecretStore::reprotect` now returns `Err(SecretStoreError::NoEntry)` for an absent target instead of a silent `Ok(())` no-op.
- `SecretStoreError::NoEntry` promoted to a top-level variant (was nested under `OsKeyringErrorKind::NoEntry`).
- **Tier-2 envelope wire format changed** (bincode-encoded structs, all three AADs carry `-v2` domain tags). Any Tier-2 envelope written by an earlier commit of this PR no longer decodes — acceptable because this PR has not yet shipped.
- New public error variants: `ExpectedProtectedButUnsealed`, `NeedsPassword`, `WrongPassword`, `BlankPassphrase`, `UnsupportedEnvelopeVersion`, top-level `NoEntry`.
- `get_secret` is an **additive** password-aware read; existing `get`/`set` call sites are unchanged.

## Review feedback resolved this round

- **Thread #2 (BLOCKING)** — symmetric `is_blank()` guard on `unwrap` (now `wire/envelope.rs::unwrap_password_payload`).
- **Thread #3 (SUGGESTION)** — per-read KDF ceiling at `default_target()` on `m_kib` AND `t`.
- **Thread #4** — `bind_object_identity` helper, subsumed by the new bincode `Tier2Aad` / `EntryAad` / `VerifyAad` structs.
- **Thread #5** — clarified `SecretString::new` `deny(unsafe_code)` comment.
- **Thread #6** — `delete: Result<bool>` + `reprotect: Err(NoEntry)` on absent.
- **Smythe S3.1 (MED)** — t-axis added to per-read KDF ceiling.
- **Smythe S2.3 (LOW)** — `consumed == blob.len()` assert on bincode decode.
- **Marvin QA-008 (MED)** — `DECODE_CONFIG` asymmetric-config rustdoc justifies the security-positive deviation from the project's standard `with_no_limit()`.
- **Marvin QA-005 / Smythe S1.3 (LOW)** — `VERIFY_DOMAIN_V2` rustdoc corrected (TIER2 and VERIFY are both 18 bytes; disjointness is by content past the common prefix).
- **Adams PROJ-003 (HIGH)** — `SECRETS.md` rewritten for the new wire format.

## Checklist:
- [x] I have added "!" to the title and described breaking changes in the corresponding section


thepastaclaw:
✅ Review complete (commit 07579e3e)

packages/rs-platform-wallet-storage/src/secrets/envelope.rs
```diff
@@ -0,0 +1,807 @@
...
+        password,
+        plaintext,
+        KdfParams::default_target(),
+    )
+}
+
+/// [`wrap`] with explicit Argon2 `params` (tests use the floor params for
+/// speed; production uses [`KdfParams::default_target`]). `params` is
+/// ignored when `password` is `None`.
+pub(crate) fn wrap_with_params(
```
lklimek: why not just define some struct and  use bincode for envelope?

packages/rs-platform-wallet-storage/src/secrets/envelope.rs
```diff
@@ -0,0 +1,807 @@
...
+    label: &str,
+    password: &SecretString,
+    body: &[u8],
+) -> Result<SecretBytes, SecretStoreError> {
+    if body.len() < MIN_SCHEME1_BODY {
+        // The scheme byte says protected, but the body cannot hold a sealed
+        // payload — corrupt, not a strip.
+        return Err(SecretStoreError::Corruption);
+    }
+    let kdf = decode_kdf(&body[..KDF_FIELD_LEN]);
```
thepastaclaw: **🔴 Blocking: Reject blank object passwords on protected reads, mirroring the wrap-side invariant**

`wrap_with_params` rejects `pw.is_blank()` at line 151 before sealing, but `unwrap_scheme1` accepts `Some(blank)` and derives an Argon2 key from the empty string. Under this PR's threat model the backend bytes are attacker-writable; an attacker who plants a fresh scheme-1 envelope encrypted under the blank password — using the correct `(wallet_id, label)` AAD and a key derived from blank+salt — will produce a blob that decrypts successfully under `Some(SecretString::empty())`. Any caller that forwards an empty prompt value as `Some(pw)` (UI bug, defensive double-call, etc.) then receives attacker-controlled plaintext instead of the typed `BlankPassphrase` failure that the wrap path documents as the crate-wide invariant. Since `Option<&SecretString>` carries the protection assertion across this boundary, validate the inner value at the read boundary too. The fix is one line, and both Codex reviewers independently flagged it.

```suggestion
fn unwrap_scheme1(
    wallet_id: &WalletId,
    label: &str,
    password: &SecretString,
    body: &[u8],
) -> Result<SecretBytes, SecretStoreError> {
    if body.len() < MIN_SCHEME1_BODY {
        // The scheme byte says protected, but the body cannot hold a sealed
        // payload — corrupt, not a strip.
        return Err(SecretStoreError::Corruption);
    }
    // Mirror wrap's invariant: a blank object password is rejected on read
    // as well as enrol, so a backend-write attacker who plants a scheme-1
    // envelope sealed under the blank password cannot inject plaintext to
    // a caller that accidentally forwards Some(empty).
    if password.is_blank() {
        return Err(SecretStoreError::BlankPassphrase);
    }
    let kdf = decode_kdf(&body[..KDF_FIELD_LEN]);
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-storage/src/secrets/envelope.rs
```diff
@@ -0,0 +1,807 @@
...
+    kdf.enforce_bounds()?;
+
+    let mut salt = [0u8; SALT_LEN];
+    salt.copy_from_slice(&body[KDF_FIELD_LEN..KDF_FIELD_LEN + SALT_LEN]);
+    let mut nonce = [0u8; NONCE_LEN];
+    nonce.copy_from_slice(&body[KDF_FIELD_LEN + SALT_LEN..KDF_FIELD_LEN + SALT_LEN + NONCE_LEN]);
+    let ciphertext = &body[KDF_FIELD_LEN + SALT_LEN + NONCE_LEN..];
+
+    let aad = scheme1_aad(&kdf, &salt, wallet_id.as_bytes(), label);
+    let key = crypto::derive_key(password, &salt, kdf)?;
```
thepastaclaw: **🟡 Suggestion: Consider a per-read ceiling at `default_target()` to bound backend-write OOM/CPU attacks**

`unwrap_scheme1` reads KDF params from the attacker-controllable envelope and only gates them against the global `enforce_bounds` ceiling (`ARGON2_MAX_M_KIB = 1_048_576` KiB, `ARGON2_MAX_T = 16`). Because the public `wrap` always uses `KdfParams::default_target()` (64 MiB / t=3), a backend-write attacker can plant a scheme-1 envelope at the bounds maximum and force ~1 GiB allocation plus 16 Argon2 passes per read — substantial on resource-constrained hosts (iOS/embedded) and an availability attack on protected reads. The existing test `kdf_ceiling_enforced_before_derivation` pins ARGON2_MAX as valid, so this is a deliberate design choice rather than a defect; flagging as a suggestion in case the tighter read-time ceiling is desirable defense-in-depth. If the project intends to expose operator-tuned KDF params later, this should remain as-is and the residual should be documented next to the existing KDF-bounds discussion. Fix would be to add `kdf.m_kib > KdfParams::default_target().m_kib || kdf.t > KdfParams::default_target().t => return Err(SecretStoreError::KdfFailure)` immediately after `enforce_bounds()`.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):
## Code Review

Tier-2 per-object password envelope is carefully designed: AAD binds identity, KDF bounds enforced pre-derivation, fresh salt/nonce, strict fail-closed read with comprehensive File+Os tests. One blocking asymmetry: blank object passwords are rejected at wrap but not at unwrap, leaving an injection path under backend-write + Some(empty). One hardening suggestion on read-time KDF ceiling.

<details>

```

In `packages/rs-platform-wallet-storage/src/secrets/envelope.rs`:
- [BLOCKING] packages/rs-platform-wallet-storage/src/secrets/envelope.rs:256-267: Reject blank object passwords on protected reads, mirroring the wrap-side invariant
  `wrap_with_params` rejects `pw.is_blank()` at line 151 before sealing, but `unwrap_scheme1` accepts `Some(blank)` and derives an Argon2 key from the empty string. Under this PR's threat model the backend bytes are attacker-writable; an attacker who plants a fresh scheme-1 envelope encrypted under the blank password — using the correct `(wallet_id, label)` AAD and a key derived from blank+salt — will produce a blob that decrypts successfully under `Some(SecretString::empty())`. Any caller that forwards an empty prompt value as `Some(pw)` (UI bug, defensive double-call, etc.) then receives attacker-controlled plaintext instead of the typed `BlankPassphrase` failure that the wrap path documents as the crate-wide invariant. Since `Option<&SecretString>` carries the protection assertion across this boundary, validate the inner value at the read boundary too. The fix is one line, and both Codex reviewers independently flagged it.
- [SUGGESTION] packages/rs-platform-wallet-storage/src/secrets/envelope.rs:267-278: Consider a per-read ceiling at `default_target()` to bound backend-write OOM/CPU attacks
  `unwrap_scheme1` reads KDF params from the attacker-controllable envelope and only gates them against the global `enforce_bounds` ceiling (`ARGON2_MAX_M_KIB = 1_048_576` KiB, `ARGON2_MAX_T = 16`). Because the public `wrap` always uses `KdfParams::default_target()` (64 MiB / t=3), a backend-write attacker can plant a scheme-1 envelope at the bounds maximum and force ~1 GiB allocation plus 16 Argon2 passes per read — substantial on resource-constrained hosts (iOS/embedded) and an availability attack on protected reads. The existing test `kdf_ceiling_enforced_before_derivation` pins ARGON2_MAX as valid, so this is a deliberate design choice rather than a defect; flagging as a suggestion in case the tighter read-time ceiling is desirable defense-in-depth. If the project intends to expose operator-tuned KDF params later, this should remain as-is and the residual should be documented next to the existing KDF-bounds discussion. Fix would be to add `kdf.m_kib > KdfParams::default_target().m_kib || kdf.t > KdfParams::default_target().t => return Err(SecretStoreError::KdfFailure)` immediately after `enforce_bounds()`.
```
</details>

packages/rs-platform-wallet-storage/src/secrets/envelope.rs
```diff
@@ -0,0 +1,807 @@
...
+        // no plaintext is ever materialized (CWE-347).
+        Err(SecretStoreError::Decrypt) => Err(SecretStoreError::WrongPassword),
+        Err(e) => Err(e),
+    }
+}
+
+/// Build the scheme-1 AAD binding object identity + header,
+/// length-prefixed for the variable fields, mirroring
+/// [`format::aad`](super::file::format::aad)/`verify_aad`.
+fn scheme1_aad(
```
lklimek: ensure we don't already have that aad implemented in the secrets/file . If we do, prefer reuse over reimplementing

packages/rs-platform-wallet-storage/src/secrets/secret.rs
```diff
@@ -47,6 +59,8 @@ impl SecretString {
         let cap = source.len().max(DEFAULT_CAPACITY);
         let mut buf = String::with_capacity(cap);
         buf.push_str(&source);
+        // Do not remove: wipes the moved-in plaintext source before it drops
+        // (its freed buffer cannot be scanned in a test under deny(unsafe_code)).
```
lklimek: if thie deny(unsafe_code) is only an issue in test, you can annotate the test to explicitly allow unsafe for that one test. Don't allow unsafe on production code.

packages/rs-platform-wallet-storage/src/secrets/store.rs
```diff
@@ -85,14 +165,102 @@ impl SecretStore {
...
+    /// unwrap→rewrap→overwrite — the canonical re-protection flow.
+    ///
+    /// Reads the object under the `current` expectation (so a strip is
+    /// caught fail-closed before any rewrap), then re-writes it under
+    /// `new`:
+    /// - **add:** `current = None`, `new = Some(pw)`;
+    /// - **change:** `current = Some(old)`, `new = Some(pw_new)`;
+    /// - **remove:** `current = Some(old)`, `new = None`.
+    ///
+    /// An absent object is a no-op (`Ok(())`). The rewrite is the same-slot
```
lklimek: better error as not found

Claudius-Maginificent: Confirmed: `reprotect` (store.rs:251-262) currently returns `Ok(())` when the object is absent, and that's documented as 'An absent object is a no-op (`Ok(())`)' on line 238. The matching no-op behavior also exists for `delete` (line 266) which is idempotent.

Two paths forward:

**(a) Change to `Err(NotFound)`** — would surface programmer error at the boundary, matches what you're asking. Trade-off: a benign race where a parallel `delete` lands between the caller's `list`/check and `reprotect` becomes a hard error the caller must absorb, when today it silently no-ops. Currently absence and deletion-in-flight are indistinguishable (and there's no `NotFound` variant in `SecretStoreError`; `get_raw` returns `Ok(None)` for absence too).

**(b) Keep `Ok(())`, sharpen the doc** — the same-slot reprotect intent is 'enforce the protection state on this object'; if the object is gone, the protection state is vacuously enforced. Race-safe, idempotent, symmetric with `delete`.

I'd lean (b) for race safety + symmetry, but (a) is the stricter API. If you confirm (a), the follow-up will add `SecretStoreError::NotFound`, change `reprotect`'s absence branch to return it, and audit `delete`'s no-op for consistency (probably keep `delete` idempotent and only make `reprotect` strict — the caller of `reprotect` has typed 'here is a password I want enforced', so an absent target is meaningful in a way it isn't for `delete`). Which path do you want?

packages/rs-platform-wallet-storage/src/secrets/envelope.rs
Claudius-Maginificent: Checked `secrets/file/format.rs::aad` (file/format.rs:107) — it builds `format_version(u32 LE) ‖ wallet_id_len(u32 LE) ‖ wallet_id ‖ label_len(u32 LE) ‖ label`. Tier-2's `scheme1_aad` needs to bind a different field set: `TIER2_DOMAIN ‖ MAGIC ‖ ENVELOPE_VERSION ‖ SCHEME_PASSWORD ‖ encode_kdf ‖ salt_len ‖ salt ‖ wallet_id_len ‖ wallet_id ‖ label_len ‖ label`.

The two AADs share the `wallet_id`+`label` tail, but the prefixes are intentionally disjoint:
- The vault entry AAD uses `format_version` so a rolled-back `FORMAT_VERSION` fails the tag.
- The Tier-2 AAD uses `TIER2_DOMAIN` (`b"PWSEV-TIER2-AAD-v1"`) plus the envelope magic/version/scheme so a Tier-2 blob can never be confused with a vault entry AAD-bound value (defense against cross-context tag reuse), and the KDF params + salt go into AAD too so header tamper / KDF downgrade surfaces as `WrongPassword`.

Factoring out the shared `wallet_id_len ‖ wallet_id ‖ label_len ‖ label` tail into a `bind_object_identity(buf, wallet_id, label)` helper IS a clean reuse — it saves the duplicated `(len_le, bytes)` boilerplate without forcing the two AADs to share a prefix they must not share. Will land in the follow-up commit; rejecting the larger 'call `format::aad` directly' shape because domain-separation requires the prefix to differ.

packages/rs-platform-wallet-storage/src/secrets/envelope.rs
Claudius-Maginificent: Confirmed against current code. `wrap_with_params` rejects `pw.is_blank()` at envelope.rs:151, but `unwrap_scheme1` (256-286) goes straight from `body.len() < MIN_SCHEME1_BODY` to `decode_kdf` → `enforce_bounds` → `derive_key(password, …)`. No symmetric `password.is_blank()` guard exists, so the asymmetry you describe IS present: a backend-write attacker who plants a scheme-1 envelope sealed under the blank password with the correct AAD (`wallet_id`, `label`, `KdfParams::default_target()`, attacker-chosen salt) yields plaintext to a `Some(SecretString::empty())` read.

**Status: unresolved — actual gap, not addressed in HEAD.**

Proposed fix is exactly the one-line `if password.is_blank() { return Err(SecretStoreError::BlankPassphrase); }` immediately after the `MIN_SCHEME1_BODY` check, mirroring wrap's invariant. Will land in a follow-up commit with a regression test that wraps under blank, then asserts unwrap returns `BlankPassphrase` (not plaintext, not `WrongPassword`).

packages/rs-platform-wallet-storage/src/secrets/envelope.rs
Claudius-Maginificent: Verified the residual: `unwrap_scheme1` calls only `kdf.enforce_bounds()` (envelope.rs:269) before `derive_key`. With `ARGON2_MAX_M_KIB = 1_048_576` and `ARGON2_MAX_T = 16`, a backend-write attacker who plants an inflated scheme-1 envelope can drive ~1 GiB / 16-pass Argon2 per protected read on resource-constrained hosts. Public `wrap` paths all use `KdfParams::default_target()` (64 MiB / t=3), so a tighter read-time ceiling would not reject any value this crate's own writers produce.

**Taking the suggestion.** Adding a `default_target()` ceiling after `enforce_bounds()` is a strict tightening with no impact on legitimate envelopes and bounds the per-read DoS surface — defense in depth is cheap here. Will land in the same follow-up commit as the blank-pw fix, with a test that pins (a) the new error on inflated headers and (b) `kdf_ceiling_enforced_before_derivation` still passes for envelopes wrapped at the floor params.

Keeping the wider `ARGON2_MAX_*` as the absolute ceiling so a later operator-tunable wrap path remains representable; the per-read cap will be `KdfParams::default_target()` until/unless an operator-tuned wrap surfaces.

packages/rs-platform-wallet-storage/src/secrets/secret.rs
Claudius-Maginificent: Re-read the code carefully — production does NOT allow `unsafe`. `secret.rs` carries crate-wide `#![deny(unsafe_code)]` and there is no `#[allow(unsafe_code)]` anywhere in `SecretString::new` or its tests (confirmed via `grep -n unsafe packages/rs-platform-wallet-storage/src/secrets/secret.rs`).

The `// Do not remove` comment at secret.rs:62-63 is explaining *why* `source.zeroize()` runs unconditionally on the moved-in `String` source: the test `secret_string_new_zeroizes_string_source` at line 380 needs to PIN that the source-side wipe happens, but a direct scan of the freed buffer would be use-after-free — which would require `unsafe`, which the crate forbids. So the test instead asserts the primitive (`String::zeroize` empties a buffer) plus pins the call-site via this do-not-remove comment. The `source.zeroize()` IS the workaround that keeps the test sound under `deny(unsafe_code)`; it does not require, nor enable, any `unsafe` in production.

**Status: no code change needed — this is a comment misread.** If the comment wording is the concern (it's currently terse, and 'cannot be scanned under deny(unsafe_code)' reads as if the test is somehow constrained), I can reword to `(// a direct freed-buffer scan would require unsafe, which this crate forbids; the test at line 380 instead pins the String::zeroize primitive and this call site)` to make the rationale legible without re-reading the linked test. Let me know which you prefer.

packages/rs-platform-wallet-storage/src/secrets/envelope.rs
Claudius-Maginificent: Verified against current code (envelope.rs:122 `wrap_with_params`). The wire format is hand-rolled `magic ‖ version ‖ scheme ‖ kdf ‖ salt ‖ nonce ‖ ciphertext` rather than a `#[derive(bincode::Encode)]` struct on purpose:

1. **AAD parity with `secrets/file::format::aad`** — the on-disk vault binds AAD with explicit length-prefixed LE fields (`format::aad` at file/format.rs:107). Tier-2 must reproduce that exact byte layout in `scheme1_aad` so a relocated/confused blob fails the tag; a serde/bincode struct introduces an extra layer (field order, varint length encoding, optional framing) that we'd then have to match in AAD construction anyway.
2. **Forward compatibility** — the header is `magic ‖ version ‖ scheme` so an unknown version can fail closed at byte index 5 without parsing the rest. A bincode struct would need a sentinel pre-decode anyway, which is what `MAGIC` already is.
3. **Fixed bounded overhead** — `MAX_ENVELOPE_OVERHEAD = 128` is computed from the fixed-size fields, which feeds `MAX_PLAINTEXT_LEN`. Bincode-encoded `Vec<u8>` carries a runtime length prefix; the overhead would no longer be a compile-time constant.

Leaving as-is — `unwrap_scheme1` is ~25 lines and the byte layout is the security boundary.

packages/rs-platform-wallet-storage/src/secrets/envelope.rs
thepastaclaw: Still present at `1e70ebd`.

packages/rs-platform-wallet-storage/src/secrets/envelope.rs
thepastaclaw: Still present at `1e70ebd`.

thepastaclaw (changes requested):
## Code Review

Incremental review fb7953ea→1e70ebd0: the latest push is an unrelated upstream merge (rehydration/SDK/UI) that does not touch `packages/rs-platform-wallet-storage/src/secrets/envelope.rs`. No new in-scope defects were introduced. Both prior findings on the Tier-2 envelope read path are carried forward and verified STILL VALID against current HEAD: (1) `unwrap_scheme1` accepts `Some(blank)` even though `wrap_with_params` rejects it, leaving an asymmetry exploitable under this PR's backend-write threat model (blocking), and (2) `unwrap_scheme1` only enforces the wide global KDF ceiling, allowing a backend-write attacker to force ~1 GiB / t=16 Argon2 work per read (defense-in-depth suggestion).

No new latest-delta findings were verified. The two findings below are carried forward from the prior `fb7953ea` review after re-validation against `1e70ebd0`. The existing inline threads were updated as still present; this top-level review records the exact-SHA review action.

<details>

```

- [BLOCKING] packages/rs-platform-wallet-storage/src/secrets/envelope.rs:262-269: [Carried forward from prior review] Reject blank object passwords on protected reads, mirroring the wrap-side invariant
  STILL VALID at 1e70ebd0 — the latest push is a merge of unrelated rehydration work and does not touch `secrets/envelope.rs`. Verified against source: `wrap_with_params` rejects `pw.is_blank()` at envelope.rs:151 before sealing, but `unwrap_scheme1` (envelope.rs:256–286) takes `password: &SecretString` and proceeds straight from the body-length check (line 262) to `decode_kdf` / `enforce_bounds` / `crypto::derive_key(password, &salt, kdf)` at line 278 without validating `password.is_blank()`.
  
  Under this PR's stated backend-write threat model, an attacker who can write the stored bytes can plant a fresh scheme-1 envelope encrypted under the blank password — using the correct `(wallet_id, label)` AAD and a key derived from blank+salt — and that blob decrypts successfully when a caller forwards `Some(SecretString::empty())`. Any caller that accidentally forwards an empty prompt value as `Some(pw)` (UI bug, defensive double-call, blank field returned by an unfocused dialog) then receives attacker-controlled plaintext instead of the typed `BlankPassphrase` failure that the wrap path documents as the crate-wide invariant. Because `Option<&SecretString>` is what carries the protection assertion across this boundary, the inner value must also be validated at the read boundary. Six independent reviewers (claude+codex across general/security/rust-quality) converged on this in both the prior and current cycles.
- [SUGGESTION] packages/rs-platform-wallet-storage/src/secrets/envelope.rs:267-278: [Carried forward from prior review] Consider a per-read KDF ceiling at `default_target()` to bound backend-write OOM/CPU attacks
  STILL VALID at 1e70ebd0 — envelope.rs unchanged in this delta. `unwrap_scheme1` reads KDF params from the attacker-controllable envelope and only gates them against the global `enforce_bounds` ceiling at line 269 (`ARGON2_MAX_M_KIB = 1_048_576` KiB, `ARGON2_MAX_T = 16`). Because the public `wrap` always emits `KdfParams::default_target()` (64 MiB / t=3), a backend-write attacker can plant a scheme-1 envelope at the bounds maximum and force ~1 GiB allocation plus 16 Argon2 passes per protected read before AEAD tag verification fails — substantial on resource-constrained hosts (iOS/embedded) and an availability attack on every `get_secret(.., Some(pw))` call.
  
  The existing test `kdf_ceiling_enforced_before_derivation` pins ARGON2_MAX as the valid edge, so the current ceiling is a deliberate forward-compatibility choice rather than a defect — flagging as defense-in-depth. If operator-tuned KDF params remain out of scope, a minimal hardening would be `if kdf.m_kib > KdfParams::default_target().m_kib || kdf.t > KdfParams::default_target().t { return Err(SecretStoreError::KdfFailure); }` immediately after `enforce_bounds()`. If higher-cost envelopes are intended later, document the residual next to the existing KDF-bounds discussion instead.
```
</details>

packages/rs-platform-wallet-storage/SECRETS.md
```diff
@@ -72,6 +86,200 @@ filename); the parent directory is materialized on the first write.
...
+integrity-protected consumer metadata. Do not mistake the strict read for
+rollback protection.
+
+#### Add / change / remove an object password
+
+`reprotect(service, label, current, new)` does it in one same-slot
+unwrap→rewrap→overwrite: read under the `current` expectation (so a strip
+is caught before any rewrite), then write under `new` — `None`→`Some` adds,
+`Some`→`Some` changes, `Some`→`None` removes. An absent object is a no-op
+(`Ok(())`). The rewrite is a same-slot overwrite — atomic on the file arm,
```
thepastaclaw: **🟡 Suggestion: Update `reprotect` docs to match `NoEntry` semantics on absent objects**

The SECRETS.md paragraph still says an absent object is a no-op returning `Ok(())`, but the current `SecretStore::reprotect` implementation intentionally returns `Err(SecretStoreError::NoEntry)` when `get_secret` returns `None` (`secrets/store.rs:262-264`), and the test `reprotect_absent_returns_no_entry` (`store.rs:409-419`) pins that behavior. The rustdoc on `reprotect` (`store.rs:239-242`) is already correct. The PR description calls out this change as closing a downgrade-detection blind spot, so leaving the public design doc with the old no-op contract is misleading: a consumer reading SECRETS.md would treat an absent protected object as harmless instead of as a backend/protection-status disagreement that should not be silently dropped. This is in-scope because the semantic shift is introduced by this PR.

```suggestion
`reprotect(service, label, current, new)` does it in one same-slot
unwrap→rewrap→overwrite: read under the `current` expectation (so a strip
is caught before any rewrite), then write under `new` — `None`→`Some` adds,
`Some`→`Some` changes, `Some`→`None` removes. An absent object returns
`Err(SecretStoreError::NoEntry)` — `reprotect` is operational, so absence
means the caller's protection-status record disagrees with the backend and
must not be silently dropped. The rewrite is a same-slot overwrite — atomic on the file arm,
and on the OS arm inheriting the backend's single-item-replace contract —
so a crash between the read and the commit leaves the prior value intact
and readable under `current`. **After a successful call the consumer MUST
update its own protection-status record** (the protection expectation lives
there). There is **no password recovery** — losing an object password
bricks that object (an availability trade-off the UX must state plainly).
```

<sub>source: ['codex']</sub>

thepastaclaw (approved):
## Code Review

Both prior findings are FIXED at HEAD a29cbf8e: prior-1 (blank-password symmetry on read) is closed by the `password.is_blank()` guard at `wire/envelope.rs:270-272` before any KDF/AEAD work, with regression test `unwrap_password_payload_rejects_some_blank_password`; prior-2 (per-read KDF ceiling) is closed by the `default_target()` clamp on BOTH `m_kib` and `t` at `wire/envelope.rs:283-286`, with regression tests covering each axis. One new in-scope documentation finding (codex-general): `SECRETS.md` still describes the old `reprotect` no-op semantics for absent objects, but the implementation now returns `Err(NoEntry)` (pinned by `reprotect_absent_returns_no_entry`). No new blocking defects in the latest delta.

🟡 1 suggestion(s)

<details>

```

In `packages/rs-platform-wallet-storage/SECRETS.md`:
- [SUGGESTION] packages/rs-platform-wallet-storage/SECRETS.md:251-255: Update `reprotect` docs to match `NoEntry` semantics on absent objects
  The SECRETS.md paragraph still says an absent object is a no-op returning `Ok(())`, but the current `SecretStore::reprotect` implementation intentionally returns `Err(SecretStoreError::NoEntry)` when `get_secret` returns `None` (`secrets/store.rs:262-264`), and the test `reprotect_absent_returns_no_entry` (`store.rs:409-419`) pins that behavior. The rustdoc on `reprotect` (`store.rs:239-242`) is already correct. The PR description calls out this change as closing a downgrade-detection blind spot, so leaving the public design doc with the old no-op contract is misleading: a consumer reading SECRETS.md would treat an absent protected object as harmless instead of as a backend/protection-status disagreement that should not be silently dropped. This is in-scope because the semantic shift is introduced by this PR.
```
</details>

packages/rs-platform-wallet-storage/src/secrets/envelope.rs
thepastaclaw: Resolved in `121e8cd` — _Reject blank object passwords on protected reads, mirroring the wrap-side invariant_ no longer present.

packages/rs-platform-wallet-storage/src/secrets/envelope.rs
thepastaclaw: Resolved in `121e8cd` — _Consider a per-read ceiling at `default_target()` to bound backend-write OOM/CPU attacks_ no longer present.

packages/rs-platform-wallet-storage/src/secrets/envelope.rs
```diff
@@ -1,807 +1,22 @@
...
 //!
-//! Sits ABOVE [`SecretStore`](crate::secrets::SecretStore), over both the
-//! `File` vault and `Os` keyring arms: the backend stores opaque bytes,
-//! and a chosen critical object (a seed wallet, a single privkey) can be
-//! wrapped under an extra, user-supplied **object password** before it
-//! ever reaches the backend. Reading a protected object then needs BOTH
-//! backend access AND the password — the first control that survives a
-//! full backend compromise (the keychain scraped, the vault stolen and its
-//! passphrase cracked).
+//! Sits ABOVE [`SecretStore`](crate::secrets::SecretStore), over both
```
lklimek: I think this file can be removed, optionally we can add reexports to mod.rs

packages/rs-platform-wallet-storage/src/secrets/store.rs
Claudius-Maginificent: Done in `abe77810db` (merged via `a3ecc7208b`). Two related changes in `secrets/store.rs`:

1. **`reprotect` absence is now an error**: `secrets/store.rs:262-264` returns `Err(SecretStoreError::NoEntry)` when `get_secret` returns `None`. The doc on `reprotect` and the regression test `reprotect_absent_returns_no_entry` (`store.rs:409-419`) pin the new contract. Rationale: silent no-op on absent broke the security model — the caller's trusted protection-status record disagreeing with the backend should not be silently dropped.
2. **`delete` returns `Result<bool>` instead of `Result<()>`**: `Ok(true)` removed, `Ok(false)` was absent. Stops `map_spi` from swallowing upstream `keyring-core::Error::NoEntry`, and starts using the bool the file backend's inner `delete_bytes` was already returning. Test `delete_returns_false_on_absent_true_on_present` (renamed from `delete_is_idempotent`) pins all three transitions.
3. **`SecretStoreError::NoEntry` promoted to top-level** — was nested under `OsKeyringErrorKind::NoEntry`. Rewired `map_spi` and added SPI back-projection (`E::NoEntry → KeyringError::NoEntry`).

packages/rs-platform-wallet-storage/src/secrets/envelope.rs
Claudius-Maginificent: Done — the bincode refactor (commit chain `7099d6adf6..c4af266e4e`, merged via `a29cbf8eec`) subsumes the request. `packages/rs-platform-wallet-storage/src/secrets/wire/aad.rs` now holds three small structs sharing the same bincode encoding machinery:

```rust
#[derive(bincode::Encode)] pub(crate) struct Tier2Aad<'a>  { domain, envelope_version, scheme_discriminant, kdf, salt, wallet_id, label }
#[derive(bincode::Encode)] pub(crate) struct EntryAad<'a>  { domain, format_version, wallet_id, label }
#[derive(bincode::Encode)] pub(crate) struct VerifyAad     { domain, format_version, salt, kdf }
```

All three carry `-v2` domain tags (`PWSEV-TIER2-AAD-v2`, `PWSV-ENTRY-AAD-v2`, `PWSV-VERIFY-AAD-v2`) and share `wire::config::WIRE_CONFIG`. `secrets/file/format.rs::aad()` and `verify_aad()` are now thin wrappers that bincode-encode `EntryAad` / `VerifyAad` instead of hand-rolling the byte layout — the previously duplicated `(len_le, bytes)` pattern is gone. Pairwise byte-disjointness pinned by `tier2_and_verify_aad_byte_disjoint` / `tier2_and_entry_aad_byte_disjoint` / `entry_and_verify_aad_byte_disjoint`.

packages/rs-platform-wallet-storage/SECRETS.md
Claudius-Maginificent: Verified — you're right, that paragraph is stale. SECRETS.md line 254-255 still reads `An absent object is a no-op (\`Ok(())\`)` but `secrets/store.rs:262-264` now returns `Err(SecretStoreError::NoEntry)` (test `reprotect_absent_returns_no_entry` at `store.rs:409-419` pins it), and the PR body's Breaking Changes section calls this out. The rustdoc on `reprotect` is correct; only the public design doc lagged.

This is a docs-only landing — taking the proposed text verbatim in a small follow-up commit. Leaving this thread open until that commit lands so the resolve trail tracks the actual fix, not just the acknowledgement.

packages/rs-platform-wallet-storage/src/secrets/envelope.rs
Claudius-Maginificent: Closed in `121e8cd8bd` (merged via `11338b6b64`) plus the conflict-merge port-forward in `a29cbf8eec`. After the bincode refactor `unwrap_scheme1` is gone; the symmetric guard lives at `packages/rs-platform-wallet-storage/src/secrets/wire/envelope.rs::unwrap_password_payload` as the `(a0)` step BEFORE `KdfParams::try_from` and BEFORE `enforce_bounds`/`derive_key`:

```rust
if password.is_blank() {
    return Err(SecretStoreError::BlankPassphrase);
}
```

Regression test `unwrap_password_payload_rejects_some_blank_password` builds a valid scheme-1 envelope under `pw("good")` via `scheme1_blob`, then asserts `unwrap` with `Some(SecretString::empty())` returns `BlankPassphrase` — never `WrongPassword`, never `Decrypt`, never plaintext.

packages/rs-platform-wallet-storage/src/secrets/envelope.rs
Claudius-Maginificent: Closed by the refactor + a follow-up tightening. `wire/envelope.rs::unwrap_password_payload` now gates BOTH axes against `KdfParams::default_target()` BEFORE `derive_key` allocates:

```rust
let target = KdfParams::default_target();
if kdf.m_kib > target.m_kib || kdf.t > target.t {
    return Err(SecretStoreError::KdfFailure);
}
```

(The `m_kib`-only initial version was caught by smythe's S3.1 finding and the `t` axis was added in commit `c4af266e4e` — checking only memory would have left a ~5.3× CPU-DoS gap on attacker-planted headers between the global `ARGON2_MAX_T=16` and the shipped `default_target().t=3`.) `enforce_bounds()` still runs first as the wider hard ceiling.

Pinned by tests `per_read_default_target_ceiling_rejects_inflated_header` (m_kib axis, TC-024) and `kdf_t_ceiling_fires_before_derive` (t axis, sibling test). Both flip the axis above `default_target` but below the global `enforce_bounds` ceiling, so only the per-read gate can refuse them — non-vacuous.

packages/rs-platform-wallet-storage/src/secrets/secret.rs
Claudius-Maginificent: Done in `3e2fb63111` (merged via `cf7894e115`). Comment at `secret.rs:62-65` rewritten:

```rust
// Do not remove: wipes the moved-in plaintext source before it drops.
// A direct freed-buffer scan would require `unsafe`, which this crate
// forbids; the test `secret_string_new_zeroizes_string_source` instead
// pins the `String::zeroize` primitive and this call site.
source.zeroize();
```

Makes the rationale legible without a round-trip through the linked test, and explicitly names that production does NOT allow `unsafe` (the crate carries `#![deny(unsafe_code)]` and there is no `#[allow(unsafe_code)]` anywhere in `secret.rs` — verified).

packages/rs-platform-wallet-storage/src/secrets/envelope.rs
Claudius-Maginificent: Reasonable — the file is 22 lines, all docs + re-exports, with one caller (`secrets/store.rs:15`: `use super::envelope;` → `envelope::wrap`, `envelope::unwrap`). Two ways to land your suggestion:

**(a)** Move the re-exports into `secrets/mod.rs` and delete `envelope.rs` entirely. Updates `store.rs:15` to `use super::wire::envelope` (or matches the new re-export name in `mod.rs`). The Tier-2 conceptual doc comment that currently heads `envelope.rs` would move to either `secrets/mod.rs` or `secrets/wire/mod.rs` — losing the dedicated landing file is the trade-off.

**(b)** Leave the file as a 22-line shim. It costs almost nothing structurally and gives Tier-2 a dedicated documentation entry-point. The downside is that the existence of `secrets::envelope` AND `secrets::wire::envelope` is technically a small surprise for a new reader ("which one is the real impl?"), which the brief docstring already addresses but could still confuse on first scan.

I'd lean (a) — the doc value of a dedicated file is real but small, and `secrets/wire/envelope.rs` is already self-explanatory as the implementation home. Happy to land (a) as a small follow-up if you confirm. Leaving open for your call.

packages/rs-platform-wallet-storage/src/secrets/envelope.rs
Claudius-Maginificent: Done. Bincode wire-format refactor landed in `a29cbf8eec` (merge of `refactor/...-bincode-envelope-aad`). New module `packages/rs-platform-wallet-storage/src/secrets/wire/` with:
- `Envelope { version: u32, payload: Payload }` — payload is `enum Payload { Unprotected(Vec<u8>) | Password { kdf, salt, nonce, ciphertext } }` (`wire/envelope.rs`).
- Three bincode-encoded AAD structs (`Tier2Aad`, `EntryAad`, `VerifyAad`) in `wire/aad.rs`, all with `-v2` domain tags for byte-disjoint cross-context defense.
- `KdfParamsEncoded` derived struct in `wire/kdf.rs`, with `From`/`TryFrom` conversions against `KdfParams` so the bounds enforcement stays in one place.
- `WIRE_CONFIG = standard().with_big_endian().with_no_limit()` in `wire/config.rs` — matches the project convention (`rs-platform-serialization`, `rs-dpp`, etc.).
- `DECODE_CONFIG = WIRE_CONFIG.with_limit::<MAX_SECRET_LEN + MAX_ENVELOPE_OVERHEAD>()` for the decode-only path — defends against attacker-controlled length prefixes driving multi-GiB `Vec::with_capacity`.
- `bincode = "=2.0.1"` (matches the workspace pin).

packages/rs-platform-wallet-storage/SECRETS.md
Claudius-Maginificent: Fixed in `483e667bcf`. Applied your proposed text verbatim — SECRETS.md now describes `reprotect` as returning `Err(SecretStoreError::NoEntry)` on absent, matching the implementation at `secrets/store.rs:262-264` and the regression test `reprotect_absent_returns_no_entry`. Resolving.

packages/rs-platform-wallet-storage/src/secrets/envelope.rs
Claudius-Maginificent: Done. The 22-line `secrets/envelope.rs` shim is gone (`483e667bcf`); the import rewire (`use super::wire::envelope` in `secrets/store.rs` + the `MAX_PLAINTEXT_LEN` re-export moved into `secrets/mod.rs`) landed in `07579e3e98`. Call sites stay `envelope::wrap` / `envelope::unwrap` since the module path is brought into scope by the new `use`. No public API change; `cargo check -p platform-wallet -p platform-wallet-ffi` clean.

thepastaclaw (approved):
## Code Review

Incremental delta a29cbf8e→07579e3e is a small cleanup: SECRETS.md `reprotect` paragraph updated to document `Err(SecretStoreError::NoEntry)` for absent objects, and the thin `secrets/envelope.rs` shim removed with `secrets/mod.rs` and `secrets/store.rs` rewired to import from `wire::envelope` directly. Prior finding `prior-a29-1` is FIXED at HEAD. The one new agent-flagged blocking finding (codex-general: AAD format break) is a false positive — the entire `rs-platform-wallet-storage` crate does not exist on `origin/master`, so there are no existing v1 file vaults to strand; the PR's published 'unshipped' rationale applies to the whole crate, not only the Tier-2 envelope.

PR #3958, opened 2026-06-24
ci: add Platform testnet sync status reporting

infraclaw-dash:
## Summary

- add a persistent worker harness for latest-public-Core testnet Platform sync
- add systemd service/timer/install assets under `ops/latest-core-testnet-sync`
- add a lightweight GitHub status receiver for manual/repository-dispatch final status reporting
- document the last-completed-status contract and long-running worker model

## Notes

The long-running sync is intentionally not a GitHub-hosted scheduled workflow. The worker publishes only final completed commit statuses:

- `Sync Passed`
- `Build Failed`
- `Sync Failed`

No `running` status is published, so the previous completed result remains visible while a new sync is active.

## Validation

- `node --check .github/scripts/latest-core-testnet-sync/run.cjs`
- `bash -n ops/latest-core-testnet-sync/run-worker.sh`
- `bash -n ops/latest-core-testnet-sync/install-worker.sh`
- parsed `.github/workflows/latest-core-testnet-sync-status.yml` with PyYAML
- `systemd-analyze verify ops/latest-core-testnet-sync/latest-core-testnet-sync.service ops/latest-core-testnet-sync/latest-core-testnet-sync.timer`
- dry-run success path maps to `Sync Passed`
- dry-run build failure maps to `Build Failed`

## Summary by CodeRabbit

* **New Features**
  * Added a new status update workflow for Platform testnet sync results.
  * Exposed a new badge in the README to show the latest sync state.

* **Documentation**
  * Added documentation describing the supported sync result states, required fields, and how final outcomes are reported.
  * Clarified that detailed diagnostics are available via an external link, while the badge reflects only the final status.


thepastaclaw:
✅ Final review complete — no blockers (commit 3546112a)

.github/scripts/platform-testnet-sync/run.cjs
```diff
@@ -0,0 +1,307 @@
...
+    child.stdout.on('data', (chunk) => {
+      process.stdout.write(chunk);
+      output += chunk.toString();
+      appendLog(logFileName, chunk.toString());
+    });
+
+    child.stderr.on('data', (chunk) => {
+      process.stderr.write(chunk);
+      output += chunk.toString();
+      appendLog(logFileName, chunk.toString());
```
thepastaclaw: **🔴 Blocking: stderr is captured into the same buffer used to parse the Core version**

`runCommand` appends both stdout and stderr to the single `output` buffer (lines 86–96), and `resolveLatestCoreVersion` then calls `firstLine(result.output)` (line 130) to extract the version. The documented contract says `LATEST_CORE_TESTNET_CORE_VERSION_COMMAND` prints the version on stdout, but any diagnostic written to stderr before the tag (a `set -x` line, a warning from a git/curl helper, a deprecation notice) will appear first in the merged buffer and be silently adopted as `LATEST_CORE_VERSION`. The metadata is then written, the status is published, and the run reports success against a bogus Core version. The stderr stream should still be logged, but it must not be mixed into the captured stdout used for parsing.

<sub>source: ['codex']</sub>

.github/scripts/latest-core-testnet-sync/run.cjs
```diff
@@ -0,0 +1,307 @@
...
+      process.stdout.write(chunk);
+      output += chunk.toString();
+      appendLog(logFileName, chunk.toString());
+    });
+
+    child.stderr.on('data', (chunk) => {
+      process.stderr.write(chunk);
+      output += chunk.toString();
+      appendLog(logFileName, chunk.toString());
+    });
```
thepastaclaw: **🔴 Blocking: Phase output is retained in memory for the entire 16+ hour run**

`runCommand` appends every stdout and stderr chunk to the `output` string for every phase (lines 88, 94), including the documented long-running Core-sync and Platform-sync phases that run for many hours with verbose docker/cargo/dashmate output. Logs are already streamed to per-phase log files, but the Node process keeps a full duplicate in memory until the phase exits. A verbose build or sync can grow the heap to hundreds of MB or more, eventually OOM-killing the worker before `publishCommitStatus` runs — leaving the GitHub commit status stale and the systemd unit silently failed. Only the short `resolve-core-version` path actually consumes `result.output`; the long phases should stream without retaining the full log in memory.

<sub>source: ['codex']</sub>

.github/scripts/platform-testnet-sync/run.cjs
```diff
@@ -0,0 +1,307 @@
...
+    metadata.core_version = await resolveLatestCoreVersion(metadata);
+    writeMetadata(metadata);
+
+    const phaseEnv = {
+      ...process.env,
+      LATEST_CORE_VERSION: metadata.core_version,
+      PLATFORM_SHA: metadata.platform_sha,
+      LATEST_CORE_TESTNET_SYNC_RUN_DIR: runDir,
+    };
+
```
thepastaclaw: **🔴 Blocking: GITHUB_TOKEN is forwarded into every phase subprocess**

`phaseEnv` is built by spreading `process.env` (which is populated from `EnvironmentFile=/etc/latest-core-testnet-sync.env` and includes `GITHUB_TOKEN`/`GH_TOKEN`) and is then handed to `spawn('bash', ['-c', ...])` for the operator-defined Core-sync, Platform-build, and Platform-sync commands. `run-worker.sh` first does `git reset --hard origin/${PLATFORM_BRANCH}`, so those commands execute scripts from the freshly checked-out branch. Anyone able to land a commit on the tracked branch (or any compromised build dependency invoked during the phases) can read the token from the environment and forge `Latest public Core testnet sync` commit statuses on arbitrary commits — the exact signal this PR introduces for developer trust. The same exposure is compounded by `runCommand` persisting all stdout/stderr to `${runDir}/${name}.log`, so any `set -x` or `env` invocation also writes the token to disk under `/var/log/latest-core-testnet-sync/`. The token is only needed by `publishCommitStatus`, which reads from `process.env` directly; strip `GITHUB_TOKEN`/`GH_TOKEN` from `phaseEnv` before passing it to `spawn`.

```suggestion
    const phaseEnv = {
      ...process.env,
      LATEST_CORE_VERSION: metadata.core_version,
      PLATFORM_SHA: metadata.platform_sha,
      LATEST_CORE_TESTNET_SYNC_RUN_DIR: runDir,
    };
    delete phaseEnv.GITHUB_TOKEN;
    delete phaseEnv.GH_TOKEN;
```

.github/scripts/platform-testnet-sync/run.cjs
```diff
@@ -0,0 +1,307 @@
...
+    const child = spawn('bash', ['-c', `set -euo pipefail\n${command}`], {
+      cwd: repoRoot,
+      env,
+      stdio: ['ignore', 'pipe', 'pipe'],
+    });
+
+    const timeout = setTimeout(() => {
+      child.kill('SIGTERM');
+      reject(new Error(`${name} timed out after ${timeoutMinutes} minutes`));
+    }, timeoutMs);
```
thepastaclaw: **🟡 Suggestion: Timeout only signals the bash wrapper, not the phase's process tree**

`spawn('bash', ['-c', ...])` is launched without `detached: true`, and the timeout handler calls `child.kill('SIGTERM')` (line 80), which delivers the signal only to the bash wrapper. The phase commands fork docker/dashmate/cargo subprocesses that will become orphans when bash exits, continuing to hold ports, disk, CPU, and lock files. With a 1440-minute default timeout, a real timeout means the next nightly run collides with leftovers from the previous one. Detach the child into its own process group (`spawn(..., { detached: true })`) and signal the whole group on timeout (`process.kill(-child.pid, 'SIGTERM')`, escalating to `SIGKILL` after a grace period).

<sub>source: ['claude']</sub>

.github/scripts/latest-core-testnet-sync/run.cjs
```diff
@@ -0,0 +1,307 @@
...
+  const response = await fetch(`https://api.github.com/repos/${repository}/statuses/${metadata.target_sha}`, {
+    method: 'POST',
+    headers: {
+      Accept: 'application/vnd.github+json',
+      Authorization: `Bearer ${githubToken}`,
+      'Content-Type': 'application/json',
+      'User-Agent': 'dash-platform-latest-core-testnet-sync',
+    },
+    body: JSON.stringify(body),
+  });
```
thepastaclaw: **🟡 Suggestion: GitHub `fetch` calls have no timeout or AbortSignal**

Both `resolveLatestCoreVersion` (GET `/releases`, line 138) and `publishCommitStatus` (POST `/statuses/...`, line 210) call the global `fetch` with no `signal`. The systemd unit sets `TimeoutStartSec=0` and Node imposes no default request timeout, so a hung TCP connection to GitHub blocks the worker indefinitely. For the resolve call this delays the run start; for the publish call a completed sync sits silent forever, and the next nightly run is then blocked by the `flock` in `run-worker.sh` (which exits 0 without publishing anything). Wire `AbortSignal.timeout(...)` (or equivalent) into both fetches.

<sub>source: ['claude']</sub>

.github/scripts/platform-testnet-sync/run.cjs
```diff
@@ -0,0 +1,307 @@
...
+    metadata.completed_at = now();
+    writeMetadata(metadata);
+    await publishCommitStatus(finalStatus, metadata);
+  }
+
+  console.log(`${STATUS_CONTEXT}: ${STATUS_LABELS[finalStatus]}`);
+
+  if (finalStatus !== 'sync_passed') {
+    process.exitCode = 1;
+  }
```
thepastaclaw: **🟡 Suggestion: Publish failure in `finally` silently drops the run outcome**

If `publishCommitStatus` throws inside the `finally` block (transient 5xx, expired token, network blip), the rejection is swallowed by the outer `main().catch` which only sets `process.exitCode = 1`. The metadata file already records `status: 'sync_passed'`, but no GitHub commit status is created and there is no retry — the visible status panel goes stale until the next nightly run, which itself may fail to publish for the same reason. Wrap the publish call in retry-with-backoff and log a clear, distinguishable error when the final status could not be reported, so operators can act before the next run.

<sub>source: ['claude']</sub>

ops/latest-core-testnet-sync/run-worker.sh
```diff
@@ -0,0 +1,30 @@
...
+git checkout "${PLATFORM_BRANCH}"
+git reset --hard "origin/${PLATFORM_BRANCH}"
+
+RUN_ID="$(date -u +%Y%m%dT%H%M%SZ)"
+export LATEST_CORE_TESTNET_RUN_ID="${RUN_ID}"
+export LATEST_CORE_TESTNET_SYNC_RUN_DIR="${LATEST_CORE_TESTNET_LOG_DIR}/${RUN_ID}"
+export TARGET_SHA
+TARGET_SHA="$(git rev-parse HEAD)"
+
+node .github/scripts/latest-core-testnet-sync/run.cjs
```
thepastaclaw: **🟡 Suggestion: Per-run log/state directories accumulate without retention**

`run-worker.sh` creates `${LATEST_CORE_TESTNET_LOG_DIR}/${RUN_ID}` per nightly run, and `run.cjs` writes per-phase log files and metadata under it. Nothing prunes old run directories, so over a year of daily runs you accumulate 365+ directories of multi-MB logs under `/var/log/latest-core-testnet-sync/`. Either add a retention step here (e.g. `find ... -mtime +N -delete`) or document the operator's responsibility in `ops/latest-core-testnet-sync/README.md`.

<sub>source: ['claude']</sub>

.github/workflows/latest-core-testnet-sync-status.yml
```diff
@@ -0,0 +1,98 @@
...
+
+            await github.rest.repos.createCommitStatus({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              sha: targetSha,
+              state,
+              context: 'Latest public Core testnet sync',
+              description,
+              target_url: payload.target_url || undefined,
+            });
```
thepastaclaw: **🟡 Suggestion: Dispatch workflow accepts arbitrary `target_sha`/`target_url` without validation**

The reporter posts a commit status to whatever `target_sha` and `target_url` are supplied via `repository_dispatch` `client_payload` (or `workflow_dispatch` input). Because the job runs with `statuses: write`, any credential able to fire the `latest-core-testnet-sync-completed` dispatch event can publish a `success`/`failure` status under the `Latest public Core testnet sync` context against any commit SHA, with an arbitrary `target_url` that maintainers will click from the commit page. Worst case is a forged `Sync Passed` status linking to a phishing page. Mitigations: validate `target_sha` is a 40-hex commit that exists in the repo (e.g. `repos.getCommit`), restrict `target_url` to a known origin (e.g. `https://github.com/...`), and document who is expected to hold the dispatch credential in `docs/NIGHTLY_LATEST_CORE_TESTNET_SYNC.md`.

.github/scripts/latest-core-testnet-sync/run.cjs
```diff
@@ -0,0 +1,307 @@
...
+
+  if (!response.ok) {
+    const body = await response.text();
+    throw new Error(`Unable to publish commit status: HTTP ${response.status} ${body}`);
```
thepastaclaw: **💬 Nitpick: Inner `body` shadows outer request body in `publishCommitStatus`**

The outer `body` holds the JSON request payload (line 200); the `!response.ok` branch then redeclares `const body = await response.text()` (line 222) for the error response. It is harmless today because the outer `body` is not referenced after `JSON.stringify`, but it is an easy footgun the next time someone wants to log the request payload alongside a failure. Rename the inner one to `errorBody`.

<sub>source: ['claude']</sub>

thepastaclaw (changes requested):
## Code Review

PR adds a self-contained nightly Core-testnet sync worker (Node harness, systemd units, install script, dispatch workflow, docs). No consensus or cryptographic code is touched. Verification confirms three in-scope blocking issues — stderr contamination of version parsing, unbounded in-memory log accumulation across multi-hour phases, and GITHUB_TOKEN being forwarded into phase subprocesses that execute target-branch code — plus several reliability and security suggestions worth tightening before this runs unattended.

🔴 3 blocking | 🟡 5 suggestion(s) | 💬 1 nitpick(s)

<details>

```

In `.github/scripts/latest-core-testnet-sync/run.cjs`:
- [BLOCKING] .github/scripts/latest-core-testnet-sync/run.cjs:92-95: stderr is captured into the same buffer used to parse the Core version
  `runCommand` appends both stdout and stderr to the single `output` buffer (lines 86–96), and `resolveLatestCoreVersion` then calls `firstLine(result.output)` (line 130) to extract the version. The documented contract says `LATEST_CORE_TESTNET_CORE_VERSION_COMMAND` prints the version on stdout, but any diagnostic written to stderr before the tag (a `set -x` line, a warning from a git/curl helper, a deprecation notice) will appear first in the merged buffer and be silently adopted as `LATEST_CORE_VERSION`. The metadata is then written, the status is published, and the run reports success against a bogus Core version. The stderr stream should still be logged, but it must not be mixed into the captured stdout used for parsing.
- [BLOCKING] .github/scripts/latest-core-testnet-sync/run.cjs:84-96: Phase output is retained in memory for the entire 16+ hour run
  `runCommand` appends every stdout and stderr chunk to the `output` string for every phase (lines 88, 94), including the documented long-running Core-sync and Platform-sync phases that run for many hours with verbose docker/cargo/dashmate output. Logs are already streamed to per-phase log files, but the Node process keeps a full duplicate in memory until the phase exits. A verbose build or sync can grow the heap to hundreds of MB or more, eventually OOM-killing the worker before `publishCommitStatus` runs — leaving the GitHub commit status stale and the systemd unit silently failed. Only the short `resolve-core-version` path actually consumes `result.output`; the long phases should stream without retaining the full log in memory.
- [BLOCKING] .github/scripts/latest-core-testnet-sync/run.cjs:250-256: GITHUB_TOKEN is forwarded into every phase subprocess
  `phaseEnv` is built by spreading `process.env` (which is populated from `EnvironmentFile=/etc/latest-core-testnet-sync.env` and includes `GITHUB_TOKEN`/`GH_TOKEN`) and is then handed to `spawn('bash', ['-c', ...])` for the operator-defined Core-sync, Platform-build, and Platform-sync commands. `run-worker.sh` first does `git reset --hard origin/${PLATFORM_BRANCH}`, so those commands execute scripts from the freshly checked-out branch. Anyone able to land a commit on the tracked branch (or any compromised build dependency invoked during the phases) can read the token from the environment and forge `Latest public Core testnet sync` commit statuses on arbitrary commits — the exact signal this PR introduces for developer trust. The same exposure is compounded by `runCommand` persisting all stdout/stderr to `${runDir}/${name}.log`, so any `set -x` or `env` invocation also writes the token to disk under `/var/log/latest-core-testnet-sync/`. The token is only needed by `publishCommitStatus`, which reads from `process.env` directly; strip `GITHUB_TOKEN`/`GH_TOKEN` from `phaseEnv` before passing it to `spawn`.
- [SUGGESTION] .github/scripts/latest-core-testnet-sync/run.cjs:72-82: Timeout only signals the bash wrapper, not the phase's process tree
  `spawn('bash', ['-c', ...])` is launched without `detached: true`, and the timeout handler calls `child.kill('SIGTERM')` (line 80), which delivers the signal only to the bash wrapper. The phase commands fork docker/dashmate/cargo subprocesses that will become orphans when bash exits, continuing to hold ports, disk, CPU, and lock files. With a 1440-minute default timeout, a real timeout means the next nightly run collides with leftovers from the previous one. Detach the child into its own process group (`spawn(..., { detached: true })`) and signal the whole group on timeout (`process.kill(-child.pid, 'SIGTERM')`, escalating to `SIGKILL` after a grace period).
- [SUGGESTION] .github/scripts/latest-core-testnet-sync/run.cjs:138-219: GitHub `fetch` calls have no timeout or AbortSignal
  Both `resolveLatestCoreVersion` (GET `/releases`, line 138) and `publishCommitStatus` (POST `/statuses/...`, line 210) call the global `fetch` with no `signal`. The systemd unit sets `TimeoutStartSec=0` and Node imposes no default request timeout, so a hung TCP connection to GitHub blocks the worker indefinitely. For the resolve call this delays the run start; for the publish call a completed sync sits silent forever, and the next nightly run is then blocked by the `flock` in `run-worker.sh` (which exits 0 without publishing anything). Wire `AbortSignal.timeout(...)` (or equivalent) into both fetches.
- [SUGGESTION] .github/scripts/latest-core-testnet-sync/run.cjs:285-301: Publish failure in `finally` silently drops the run outcome
  If `publishCommitStatus` throws inside the `finally` block (transient 5xx, expired token, network blip), the rejection is swallowed by the outer `main().catch` which only sets `process.exitCode = 1`. The metadata file already records `status: 'sync_passed'`, but no GitHub commit status is created and there is no retry — the visible status panel goes stale until the next nightly run, which itself may fail to publish for the same reason. Wrap the publish call in retry-with-backoff and log a clear, distinguishable error when the final status could not be reported, so operators can act before the next run.

In `ops/latest-core-testnet-sync/run-worker.sh`:
- [SUGGESTION] ops/latest-core-testnet-sync/run-worker.sh:1-30: Per-run log/state directories accumulate without retention
  `run-worker.sh` creates `${LATEST_CORE_TESTNET_LOG_DIR}/${RUN_ID}` per nightly run, and `run.cjs` writes per-phase log files and metadata under it. Nothing prunes old run directories, so over a year of daily runs you accumulate 365+ directories of multi-MB logs under `/var/log/latest-core-testnet-sync/`. Either add a retention step here (e.g. `find ... -mtime +N -delete`) or document the operator's responsibility in `ops/latest-core-testnet-sync/README.md`.

In `.github/workflows/latest-core-testnet-sync-status.yml`:
- [SUGGESTION] .github/workflows/latest-core-testnet-sync-status.yml:70-88: Dispatch workflow accepts arbitrary `target_sha`/`target_url` without validation
  The reporter posts a commit status to whatever `target_sha` and `target_url` are supplied via `repository_dispatch` `client_payload` (or `workflow_dispatch` input). Because the job runs with `statuses: write`, any credential able to fire the `latest-core-testnet-sync-completed` dispatch event can publish a `success`/`failure` status under the `Latest public Core testnet sync` context against any commit SHA, with an arbitrary `target_url` that maintainers will click from the commit page. Worst case is a forged `Sync Passed` status linking to a phishing page. Mitigations: validate `target_sha` is a 40-hex commit that exists in the repo (e.g. `repos.getCommit`), restrict `target_url` to a known origin (e.g. `https://github.com/...`), and document who is expected to hold the dispatch credential in `docs/NIGHTLY_LATEST_CORE_TESTNET_SYNC.md`.
```
</details>

.github/scripts/platform-testnet-sync/run.cjs
```diff
@@ -0,0 +1,461 @@
...
+
+  writeMetadata(metadata);
+
+  let finalStatus = 'sync_passed';
+
+  try {
+    await prepareWorkspace(metadata, timeoutMinutes);
+
+    metadata.core_version = await resolveLatestCoreVersion(metadata);
+    writeMetadata(metadata);
```
thepastaclaw: **🟡 Suggestion: resolve-core-version failures are recorded as workspace-prepare failures**

`prepareWorkspace` sets `metadata.phase = 'workspace-prepare'` and never clears it on success. The next statement awaits `resolveLatestCoreVersion`, which can fail (bad `LATEST_CORE_TESTNET_CORE_VERSION_COMMAND`, GitHub releases fetch error, no public release, etc.). The catch at line 440 sets `metadata.failure_phase = metadata.phase || 'resolve-core-version'`, so any such failure is recorded as `failure_phase=workspace-prepare`. The `|| 'resolve-core-version'` fallback only fires when `prepareWorkspace` was skipped via `LATEST_CORE_TESTNET_SKIP_WORKSPACE_PREP=1`. Set the phase before resolving the Core version so the run-tracking contract points operators at the right phase.

```suggestion
    await prepareWorkspace(metadata, timeoutMinutes);

    metadata.phase = 'resolve-core-version';
    writeMetadata(metadata);
    metadata.core_version = await resolveLatestCoreVersion(metadata);
    writeMetadata(metadata);
```

<sub>source: ['codex']</sub>

.github/scripts/platform-testnet-sync/run.cjs
```diff
@@ -0,0 +1,461 @@
...
+    metadata.failure_phase = metadata.phase || 'resolve-core-version';
+    metadata.failure_summary = error.message;
+    finalStatus = statusForFailurePhase(metadata.failure_phase);
+    console.error(error.stack || error.message);
+  } finally {
+    metadata.status = finalStatus;
+    metadata.completed_at = now();
+    writeMetadata(metadata);
+    await publishCommitStatus(finalStatus, metadata);
+  }
```
thepastaclaw: **🟡 Suggestion: Metadata claims sync_passed even when status publish fails after retries (carried forward, partial fix)**

Prior finding #6 is only partially addressed. `publishCommitStatus` now retries up to `STATUS_PUBLISH_ATTEMPTS` times, which is a clear improvement, but if every retry fails the throw propagates to `main().catch`, leaving `run-metadata.json` on disk with `status: sync_passed` (or `build_failed`/`sync_failed`) while no GitHub status was published and the previous run's status remains visible. Operators that read `run-metadata.json` to decide whether the latest sync passed will see a green result while GitHub disagrees. Record the publish outcome in the metadata (e.g. `status_published_at`, `status_publish_error`) after the final `publishCommitStatus` call so the on-disk record and GitHub stay aligned.

<sub>source: ['claude']</sub>

.github/scripts/platform-testnet-sync/run.cjs
```diff
@@ -0,0 +1,461 @@
...
+    'workspace-prepare',
+    [
+      `git fetch --prune origin ${JSON.stringify(branch)}`,
+      `git checkout ${JSON.stringify(branch)}`,
+      `git reset --hard origin/${JSON.stringify(branch)}`,
+      'git clean -ffdx -e .latest-core-testnet-sync/',
+    ].join('\n'),
+    sanitizePhaseEnv(process.env),
+    timeoutMinutes,
+  );
```
thepastaclaw: **💬 Nitpick: JSON.stringify is not a bash-quoting primitive for PLATFORM_BRANCH**

`prepareWorkspace` builds a bash command by interpolating `JSON.stringify(branch)` into double-quoted positions. JSON escaping handles `"` and `\` but does not escape `$`, backticks, or `${}`, all of which remain active inside bash double quotes. A `PLATFORM_BRANCH` value such as `release-$(touch /tmp/pwn)` would execute command substitution as the `platform-sync` user with full write access to `PLATFORM_REPO_DIR` and `/var/log/latest-core-testnet-sync`. Today the value is sourced from a root-owned `/etc/latest-core-testnet-sync.env` at the same trust level as the operator-defined phase commands, so this is not exploitable as designed; however, the previous shell-driven version had the same exposure and the move into Node is a good place to harden it. Validate `branch` against a strict ref-name regex (e.g. `^[A-Za-z0-9._/-]+$`) or pass the branch via env and reference it as `"$BRANCH"` from the script body rather than splicing it.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

All nine prior findings are addressed in the latest delta: stderr/version corruption, unbounded output buffering, GITHUB_TOKEN propagation, process-group signaling, fetch timeouts, log retention, dispatch payload validation, and the shadowed body identifier are all fixed; the publish-failure-in-finally case is partially addressed via retries but the metadata divergence remains. The latest delta introduces two new low-severity issues: a phase-attribution bug for resolve-core-version failures, and a shell-interpolation hardening gap on PLATFORM_BRANCH. No blocking issues.

<details>

```

In `.github/scripts/latest-core-testnet-sync/run.cjs`:
- [SUGGESTION] .github/scripts/latest-core-testnet-sync/run.cjs:397-400: resolve-core-version failures are recorded as workspace-prepare failures
  `prepareWorkspace` sets `metadata.phase = 'workspace-prepare'` and never clears it on success. The next statement awaits `resolveLatestCoreVersion`, which can fail (bad `LATEST_CORE_TESTNET_CORE_VERSION_COMMAND`, GitHub releases fetch error, no public release, etc.). The catch at line 440 sets `metadata.failure_phase = metadata.phase || 'resolve-core-version'`, so any such failure is recorded as `failure_phase=workspace-prepare`. The `|| 'resolve-core-version'` fallback only fires when `prepareWorkspace` was skipped via `LATEST_CORE_TESTNET_SKIP_WORKSPACE_PREP=1`. Set the phase before resolving the Core version so the run-tracking contract points operators at the right phase.
- [SUGGESTION] .github/scripts/latest-core-testnet-sync/run.cjs:444-449: Metadata claims sync_passed even when status publish fails after retries (carried forward, partial fix)
  Prior finding #6 is only partially addressed. `publishCommitStatus` now retries up to `STATUS_PUBLISH_ATTEMPTS` times, which is a clear improvement, but if every retry fails the throw propagates to `main().catch`, leaving `run-metadata.json` on disk with `status: sync_passed` (or `build_failed`/`sync_failed`) while no GitHub status was published and the previous run's status remains visible. Operators that read `run-metadata.json` to decide whether the latest sync passed will see a green result while GitHub disagrees. Record the publish outcome in the metadata (e.g. `status_published_at`, `status_publish_error`) after the final `publishCommitStatus` call so the on-disk record and GitHub stay aligned.
```
</details>

.github/scripts/platform-testnet-sync/run.cjs
thepastaclaw: Resolved in `4069ab1` — _Publish failure in `finally` silently drops the run outcome_ no longer present.

.github/workflows/latest-core-testnet-sync-status.yml
thepastaclaw: Resolved in `4069ab1` — _Dispatch workflow accepts arbitrary `target_sha`/`target_url` without validation_ no longer present.

.github/scripts/latest-core-testnet-sync/run.cjs
thepastaclaw: Resolved in `4069ab1` — _Inner `body` shadows outer request body in `publishCommitStatus`_ no longer present.

.github/scripts/latest-core-testnet-sync/run.cjs
thepastaclaw: Resolved in `4069ab1` — _GitHub `fetch` calls have no timeout or AbortSignal_ no longer present.

ops/latest-core-testnet-sync/run-worker.sh
thepastaclaw: Resolved in `4069ab1` — _Per-run log/state directories accumulate without retention_ no longer present.

.github/scripts/platform-testnet-sync/run.cjs
thepastaclaw: Resolved in `4069ab1` — _stderr is captured into the same buffer used to parse the Core version_ no longer present.

.github/scripts/platform-testnet-sync/run.cjs
thepastaclaw: Resolved in `4069ab1` — _GITHUB_TOKEN is forwarded into every phase subprocess_ no longer present.

.github/scripts/platform-testnet-sync/run.cjs
thepastaclaw: Resolved in `4069ab1` — _Timeout only signals the bash wrapper, not the phase's process tree_ no longer present.

.github/scripts/latest-core-testnet-sync/run.cjs
thepastaclaw: Resolved in `4069ab1` — _Phase output is retained in memory for the entire 16+ hour run_ no longer present.

.github/scripts/platform-testnet-sync/run.cjs
```diff
@@ -0,0 +1,479 @@
...
+
+  const targetSha = resolveTargetSha();
+  metadata.target_sha = targetSha;
+  metadata.platform_sha = targetSha;
+  process.env.TARGET_SHA = targetSha;
+  writeMetadata(metadata);
```
thepastaclaw: **🟡 Suggestion: target_sha re-capture still prefers inherited env vars over the checked-out HEAD**

After `prepareWorkspace` resets the workspace to `origin/$PLATFORM_BRANCH`, it re-derives `targetSha` via `resolveTargetSha()`, which returns `process.env.TARGET_SHA || process.env.GITHUB_SHA || git rev-parse HEAD`. If either env var is set on the worker (systemd env file, prior run leakage into a long-lived shell, or future CI invocation), the script runs phase commands against the freshly reset branch tip but publishes the commit status against the inherited SHA — the worker's contract to report the actually-tested commit is silently broken. Worse, if `prepareWorkspace` fails before this line, the catch path publishes 'Sync Failed' against whatever stale SHA was captured at line 382. Resolve `target_sha` directly from `git rev-parse HEAD` after the reset, and only fall back to inherited env vars for the pre-prepare initial value.

```suggestion
  const targetSha = gitOutput(['rev-parse', 'HEAD']);
  metadata.target_sha = targetSha;
  metadata.platform_sha = targetSha;
  process.env.TARGET_SHA = targetSha;
  writeMetadata(metadata);
```

.github/workflows/platform-testnet-sync-status.yml
```diff
@@ -0,0 +1,131 @@
...
+        required: false
+        type: string
+      platform_sha:
+        description: Platform commit SHA built and synced
+        required: false
+        type: string
+      completed_at:
+        description: Completion timestamp, preferably ISO 8601 UTC
+        required: false
+        type: string
```
thepastaclaw, line 36: **🟡 Suggestion: workflow_dispatch lets any repo writer forge a sync status**

The receiver exposes both `repository_dispatch` (driven by the operator-only worker credential) and `workflow_dispatch` with free-form `status` and `target_sha` inputs. The job carries `statuses: write`, so anyone with repo write access can manually publish 'Sync Passed' against any existing SHA without the worker ever running. If maintainers, branch protection, or merge automation treat the `Latest public Core testnet sync` context as evidence the testnet sync actually ran, this becomes a trust-laundering primitive that bypasses the operator-credential boundary the rest of the pipeline carefully maintains. Either drop the `workflow_dispatch` trigger, or gate it behind an environment with required reviewers so the manual path requires an operator approval distinct from generic repo write.

<sub>source: ['codex']</sub>

thepastaclaw (commented):
## Code Review

Prior phase-tracking and publish-outcome metadata findings are resolved. One prior carry-forward (PLATFORM_BRANCH quoting) plus two new in-scope hardening items: target_sha resolution can still report a stale/inherited commit after workspace reset, and the new status-receiver workflow exposes a workflow_dispatch override that lets any repo writer forge a 'Sync Passed' status. All three are suggestions, no blockers.

🟡 2 suggestion(s)

<details>

```

In `.github/scripts/latest-core-testnet-sync/run.cjs`:
- [SUGGESTION] .github/scripts/latest-core-testnet-sync/run.cjs:233-237: target_sha re-capture still prefers inherited env vars over the checked-out HEAD
  After `prepareWorkspace` resets the workspace to `origin/$PLATFORM_BRANCH`, it re-derives `targetSha` via `resolveTargetSha()`, which returns `process.env.TARGET_SHA || process.env.GITHUB_SHA || git rev-parse HEAD`. If either env var is set on the worker (systemd env file, prior run leakage into a long-lived shell, or future CI invocation), the script runs phase commands against the freshly reset branch tip but publishes the commit status against the inherited SHA — the worker's contract to report the actually-tested commit is silently broken. Worse, if `prepareWorkspace` fails before this line, the catch path publishes 'Sync Failed' against whatever stale SHA was captured at line 382. Resolve `target_sha` directly from `git rev-parse HEAD` after the reset, and only fall back to inherited env vars for the pre-prepare initial value.

In `.github/workflows/latest-core-testnet-sync-status.yml`:
- [SUGGESTION] .github/workflows/latest-core-testnet-sync-status.yml:7-36: workflow_dispatch lets any repo writer forge a sync status
  The receiver exposes both `repository_dispatch` (driven by the operator-only worker credential) and `workflow_dispatch` with free-form `status` and `target_sha` inputs. The job carries `statuses: write`, so anyone with repo write access can manually publish 'Sync Passed' against any existing SHA without the worker ever running. If maintainers, branch protection, or merge automation treat the `Latest public Core testnet sync` context as evidence the testnet sync actually ran, this becomes a trust-laundering primitive that bypasses the operator-credential boundary the rest of the pipeline carefully maintains. Either drop the `workflow_dispatch` trigger, or gate it behind an environment with required reviewers so the manual path requires an operator approval distinct from generic repo write.
```
</details>

ops/latest-core-testnet-sync/README.md
```diff
@@ -0,0 +1,79 @@
+# Latest Public Core Testnet Sync Worker
```
lklimek: why it exists? who needs that? When I, as a user or developer, should use that tool?

thepastaclaw (changes requested):
## Code Review

Incremental review reconciling 3 prior findings against head f6783a6e. The JSON.stringify shell-quoting issue (prior-3) is now FIXED via env-var indirection plus validatePlatformBranch allow-list. The target_sha-from-env (prior-1) and workflow_dispatch trust-boundary (prior-2) issues are STILL VALID and carried forward. Two new in-scope issues were validated against the worker scripts: pre-lock log pruning can delete the active run directory, and install-worker.sh does not enforce root ownership on a pre-existing token-bearing env file.

🔴 2 blocking | 🟡 3 suggestion(s) | 💬 1 nitpick(s)

_6 additional finding(s) omitted (not in diff)._

<details>

```

In `.github/workflows/latest-core-testnet-sync-status.yml`:
- [BLOCKING] .github/workflows/latest-core-testnet-sync-status.yml:7-36: workflow_dispatch bypasses the operator-only worker trust boundary
  The workflow exposes workflow_dispatch with free-form status/target_sha inputs while the job carries statuses: write. Any user with repo write access can manually run it and publish 'Sync Passed' against any commit reachable in the repo, forging the 'Latest public Core testnet sync' context without the persistent worker ever running. The SHA format check and getCommit lookup only confirm the commit exists — they do not enforce that the worker produced the result. If this context is (or becomes) a required check, the manual path is a status-forgery vector. Remove the workflow_dispatch trigger or gate it behind a protected environment that requires operator approval.

In `ops/latest-core-testnet-sync/install-worker.sh`:
- [BLOCKING] ops/latest-core-testnet-sync/install-worker.sh:50-57: Existing /etc/latest-core-testnet-sync.env keeps prior ownership
  When /etc/latest-core-testnet-sync.env already exists, the installer only re-applies 0600 perms and does not chown to root:root. If the file was previously written by a non-root account (e.g. created manually before the installer ran), that account retains read/write access to GITHUB_TOKEN and can modify the phase commands EnvironmentFile-loaded by the service that later runs as root. The create path also relies on install's default ownership rather than explicitly forcing root:root. Force root:root ownership in both branches so the secret-bearing env file is never owned by a less-privileged account.

In `.github/scripts/latest-core-testnet-sync/run.cjs`:
- [SUGGESTION] .github/scripts/latest-core-testnet-sync/run.cjs:250-254: target_sha can still be inherited from env after the workspace reset
  After prepareWorkspace fetches/checks out/resets to origin/$PLATFORM_BRANCH, it re-derives the SHA via resolveTargetSha(), which returns process.env.TARGET_SHA || process.env.GITHUB_SHA || git rev-parse HEAD. If either env var is present in the worker environment (systemd EnvironmentFile, prior shell, CI re-invocation), metadata.target_sha/platform_sha and the published commit status will point at the inherited SHA even though the phases actually ran against the freshly reset branch tip. After the reset, take the SHA from git directly so metadata describes the commit that was actually tested. resolveTargetSha additionally never validates the value is a 40-char hex SHA, so a malformed env value silently propagates to the statuses API (which then 422s and burns the retry budget).
- [SUGGESTION] .github/scripts/latest-core-testnet-sync/run.cjs:374-389: publishCommitStatus retries on permanent HTTP failures
  publishCommitStatusOnce throws a generic `HTTP <status> <body>` error for every non-2xx, and publishCommitStatus retries up to STATUS_PUBLISH_ATTEMPTS times with exponential backoff for any error. 401 (bad token), 403 (missing scopes), 404 (repo/SHA missing), and 422 (malformed SHA) are deterministic failures that the retries cannot recover, so the worker just delays surfacing the real error. Inspect the response status and only retry transient classes (5xx, network errors, 429); fail fast on the rest.

In `ops/latest-core-testnet-sync/run-worker.sh`:
- [SUGGESTION] ops/latest-core-testnet-sync/run-worker.sh:13-24: Log pruning runs before the worker lock is acquired
  The destructive `find ... -mtime +RETENTION -exec rm -rf {} +` block executes before `flock -n 9`. A second timer invocation can therefore prune log directories while another sync is in flight, then exit when flock fails. If LATEST_CORE_TESTNET_LOG_RETENTION_DAYS is set low or a run exceeds the retention window, the active run's log directory (LATEST_CORE_TESTNET_SYNC_RUN_DIR under LATEST_CORE_TESTNET_LOG_DIR) can be removed mid-run, breaking the orchestrator's appendLog/writeMetadata calls. Move pruning inside the flock-held critical section so only the active worker performs cleanup.
```
</details>

.github/workflows/platform-testnet-sync-status.yml
thepastaclaw: Resolved in this update — _workflow_dispatch lets any repo writer forge a sync status_ no longer present.

.github/scripts/platform-testnet-sync/run.cjs
thepastaclaw: Resolved in `f6783a6` — _target_sha re-capture still prefers inherited env vars over the checked-out HEAD_ no longer present.

thepastaclaw (changes requested):
## Code Review

Cumulative incremental review for head 06c64059 against prior reviewed commit f6783a6e. The latest delta is a scope-narrowing rename (core-sync → core-readiness) and does not touch any consensus-critical code. All six prior findings remain STILL VALID at HEAD: two blocking trust-boundary/permission issues (workflow_dispatch status forgery, env file ownership), three suggestion-level operational correctness issues, and one nitpick. One additional new suggestion is added for the initial target_sha being resolved from pre-reset HEAD, which is adjacent to but distinct from prior-3.

🔴 2 blocking | 🟡 3 suggestion(s) | 💬 1 nitpick(s)

_6 additional finding(s) omitted (not in diff)._

<details>

```

In `.github/workflows/latest-core-testnet-sync-status.yml`:
- [BLOCKING] .github/workflows/latest-core-testnet-sync-status.yml:7-36: workflow_dispatch bypasses the operator-only worker trust boundary
  The workflow exposes a manual `workflow_dispatch` path that accepts caller-provided `status` and `target_sha` while the job carries `statuses: write`. Any repo user with workflow_dispatch permission can publish `Sync Passed` against any commit reachable in the repository without the persistent worker ever running. The 40-hex regex and `getCommit` lookup only prove the SHA exists; they do not prove a sync ran. If maintainers, branch protection, or downstream automation treat the `Latest public Core testnet sync` context as evidence of a real sync, this manual path forges that signal. Either drop `workflow_dispatch` entirely (the persistent worker uses `repository_dispatch`) or gate the job behind a protected environment requiring operator approval.

In `ops/latest-core-testnet-sync/install-worker.sh`:
- [BLOCKING] ops/latest-core-testnet-sync/install-worker.sh:50-57: Existing /etc/latest-core-testnet-sync.env keeps prior ownership
  When `/etc/latest-core-testnet-sync.env` already exists, the installer only re-applies mode `0600` and leaves the current owner intact. If the file was created or copied by a non-root account before this installer ran (or rotated in by hand), that account retains read access to `GITHUB_TOKEN` and write access to the phase commands the systemd service consumes. Because this file carries the statuses-write token and controls the commands the worker runs as root-installed unit, the installer should force `chown root:root` in both the create and existing-file paths.

In `.github/scripts/latest-core-testnet-sync/run.cjs`:
- [SUGGESTION] .github/scripts/latest-core-testnet-sync/run.cjs:142-254: target_sha can be reported for a commit that was never tested
  `resolveTargetSha()` (line 142) prefers `process.env.TARGET_SHA || process.env.GITHUB_SHA` over the working-tree HEAD, and it is called twice: once in `main()` at line 399 *before* the workspace is reset, and again inside `prepareWorkspace` at line 250 *after* the reset. Two failure modes follow: (1) if `TARGET_SHA`/`GITHUB_SHA` is present in the worker environment, both calls return the inherited SHA even though phases run against `origin/$PLATFORM_BRANCH`, so the published commit status points at a commit that was not actually tested; (2) if neither env var is set and `prepareWorkspace` fails (e.g. `git fetch` errors before the reset), the `finally` block publishes `Sync Failed` against the stale pre-fetch HEAD. Resolve the SHA strictly from `git rev-parse HEAD` after a successful reset, and use that single value for both metadata and the published status.
- [SUGGESTION] .github/scripts/latest-core-testnet-sync/run.cjs:374-389: publishCommitStatus retries permanent HTTP failures
  `publishCommitStatusOnce` throws a generic `Error` for every non-2xx response, and `publishCommitStatus` retries on any thrown error up to `STATUS_PUBLISH_ATTEMPTS`. Permanent failures such as 401, 403, 404, and 422 are deterministic and cannot recover, so the worker burns the entire retry budget (with exponential backoff up to 8 seconds on attempt 3) before surfacing what is really a misconfiguration. Inspect the response status and only retry on transient categories — network errors, 429, 502, 503, 504 — while failing fast on permanent HTTP errors.

In `ops/latest-core-testnet-sync/run-worker.sh`:
- [SUGGESTION] ops/latest-core-testnet-sync/run-worker.sh:13-24: Log pruning runs before the worker lock is acquired
  The destructive `find ... -mtime +RETENTION -exec rm -rf {} +` block runs before `flock -n 9` acquires `run.lock`. A second timer invocation triggered while another sync is in flight will execute the prune first, then exit because flock fails. If retention is low or a sync exceeds the retention window, that prune can remove the active run directory and break `appendLog`/`writeMetadata` mid-run. Acquire the flock first and perform pruning inside the single-worker critical section.
```
</details>

ops/platform-testnet-sync/platform-testnet-sync.service
```diff
@@ -0,0 +1,14 @@
...
+Description=Dash Platform testnet sync
+Wants=network-online.target
+After=network-online.target
+
+[Service]
+Type=oneshot
+User=platform-sync
+Group=platform-sync
+EnvironmentFile=/etc/platform-testnet-sync.env
+ExecStart=/bin/bash -lc 'cd "${PLATFORM_REPO_DIR:-/opt/dash-platform}" && exec ops/platform-testnet-sync/run-worker.sh'
```
thepastaclaw: **🔴 Blocking: Worker harness executes from the same mutable checkout it resets to the tested branch**

The unit loads `/etc/platform-testnet-sync.env` (containing `GITHUB_TOKEN` with `statuses: write`) and then runs `ops/platform-testnet-sync/run-worker.sh` and `.github/scripts/platform-testnet-sync/run.cjs` from `PLATFORM_REPO_DIR`. That directory is the same checkout that `prepareWorkspace` later resets to `origin/$PLATFORM_BRANCH` (run.cjs:240). Consequently the harness that receives the token is itself branch-controlled code: any commit that lands on `$PLATFORM_BRANCH` can rewrite `run-worker.sh` or `run.cjs` to read `process.env.GITHUB_TOKEN` before any scrubbing happens. The trusted credential-handling code must live outside the mutable checkout — install it as root-owned files under `/usr/local/lib/platform-testnet-sync/` (or similar) and let only the phase commands execute from the checked-out branch.

<sub>source: ['codex']</sub>

.github/scripts/platform-testnet-sync/run.cjs
```diff
@@ -0,0 +1,496 @@
...
+    writeMetadata(metadata);
+
+    const phaseEnv = {
+      ...process.env,
+      CORE_VERSION: metadata.core_version,
+      PLATFORM_SHA: metadata.platform_sha,
+      PLATFORM_TESTNET_SYNC_RUN_DIR: runDir,
+    };
+    delete phaseEnv.GITHUB_TOKEN;
+    delete phaseEnv.GH_TOKEN;
```
thepastaclaw: **🔴 Blocking: Phase commands can recover GITHUB_TOKEN from parent process /proc/$PPID/environ**

`sanitizePhaseEnv` and the explicit deletes at run.cjs:428-429 only strip `GITHUB_TOKEN`/`GH_TOKEN` from the child phase environment. The long-lived Node parent still holds those secrets in `process.env` because they were inherited from systemd's `EnvironmentFile` and are needed later for `publishCommitStatus`. On a standard Linux host, a same-UID child can read `/proc/$PPID/environ` (and scan other same-UID `/proc/*/environ`), so any code executed during `baseline-readiness`, `platform-build`, or `platform-sync` — including branch-controlled build scripts — can exfiltrate the token despite the child-env scrub. Move status publishing into an isolated helper (separate UID/container/process) that is invoked after phases finish, or hold the token in a file readable only after phases complete so it never lives in the credential-holding process's environment while untrusted code runs.

<sub>source: ['codex']</sub>

ops/platform-testnet-sync/install-worker.sh
```diff
@@ -0,0 +1,65 @@
...
+  /etc/systemd/system/platform-testnet-sync.timer
+
+if [[ ! -f /etc/platform-testnet-sync.env ]]; then
+  install -m 0600 \
+    "${SCRIPT_DIR}/platform-testnet-sync.env.example" \
+    /etc/platform-testnet-sync.env
+  echo "Created /etc/platform-testnet-sync.env; fill it before enabling the timer."
+else
+  chmod 0600 /etc/platform-testnet-sync.env
+fi
```
thepastaclaw: **🔴 Blocking: Pre-existing env file keeps prior ownership**

When `/etc/platform-testnet-sync.env` already exists, the installer only reapplies `chmod 0600` and leaves whatever ownership the file came with. If the file was created or copied by a non-root account before the installer ran, that account retains read access to `GITHUB_TOKEN` and write access to the phase commands that the systemd unit executes. Because the file carries the statuses-write token and controls the worker's behavior, the root-run installer should force `root:root` ownership in both the create and existing-file paths.

```suggestion
if [[ ! -f /etc/platform-testnet-sync.env ]]; then
  install -o root -g root -m 0600 \
    "${SCRIPT_DIR}/platform-testnet-sync.env.example" \
    /etc/platform-testnet-sync.env
  echo "Created /etc/platform-testnet-sync.env; fill it before enabling the timer."
else
  chown root:root /etc/platform-testnet-sync.env
  chmod 0600 /etc/platform-testnet-sync.env
fi
```

.github/scripts/platform-testnet-sync/run.cjs
```diff
@@ -0,0 +1,496 @@
...
+      PLATFORM_BRANCH: branch,
+    },
+    timeoutMinutes,
+  );
+
+  const targetSha = resolveTargetSha();
+  metadata.target_sha = targetSha;
+  metadata.platform_sha = targetSha;
+  process.env.TARGET_SHA = targetSha;
+  writeMetadata(metadata);
```
thepastaclaw: **🟡 Suggestion: target_sha can describe a commit that was never tested**

`resolveTargetSha` prefers `TARGET_SHA` and `GITHUB_SHA` over `git rev-parse HEAD`, and it runs at main() before workspace preparation (run.cjs:399) and again after the reset inside `prepareWorkspace` (run.cjs:250). If either env var is inherited by the worker, the initial metadata and any failure-path commit status published from the `finally` block point at that inherited SHA even though no phase has actually run against it. If `prepareWorkspace` fails before the reset and `TARGET_SHA` is unset, the published status describes the stale pre-fetch checkout HEAD. Resolve the tested SHA strictly from git after a successful reset, and avoid publishing a commit-specific status before a target commit has actually been selected.

ops/platform-testnet-sync/run-worker.sh
```diff
@@ -0,0 +1,32 @@
...
+  -maxdepth 1 \
+  -type d \
+  -mtime +"${PLATFORM_TESTNET_SYNC_LOG_RETENTION_DAYS}" \
+  -exec rm -rf {} +
+
+exec 9>"${PLATFORM_TESTNET_SYNC_STATE_DIR}/run.lock"
+if ! flock -n 9; then
+  echo "Platform testnet sync is already running"
+  exit 0
+fi
```
thepastaclaw: **🟡 Suggestion: Log pruning runs before the worker lock is acquired**

The destructive `find ... -mtime +RETENTION -exec rm -rf {} +` block at lines 13-18 runs before `flock -n 9` acquires `run.lock` at line 20. A second timer invocation while a sync is in flight prunes first and only then exits because the lock is held. If retention is configured low or a sync exceeds the retention window, the prune can remove the active run directory, breaking `appendLog`/`writeMetadata` in the running worker. Acquire the lock before pruning so only the active worker performs retention cleanup.

```suggestion
exec 9>"${PLATFORM_TESTNET_SYNC_STATE_DIR}/run.lock"
if ! flock -n 9; then
  echo "Platform testnet sync is already running"
  exit 0
fi

find "${PLATFORM_TESTNET_SYNC_LOG_DIR}" \
  -mindepth 1 \
  -maxdepth 1 \
  -type d \
  -mtime +"${PLATFORM_TESTNET_SYNC_LOG_RETENTION_DAYS}" \
  -exec rm -rf {} +
```

.github/scripts/platform-testnet-sync/run.cjs
```diff
@@ -0,0 +1,496 @@
...
+    } catch (error) {
+      lastError = error;
+      console.error(`Commit status publish attempt ${attempt} failed: ${error.message}`);
+      if (attempt < STATUS_PUBLISH_ATTEMPTS) {
+        await sleep(2 ** attempt * 1000);
+      }
+    }
+  }
+
+  throw lastError;
```
thepastaclaw: **🟡 Suggestion: Status publisher retries permanent HTTP failures**

`publishCommitStatusOnce` throws the same generic `Error` for every non-2xx response, and `publishCommitStatus` retries every thrown error up to `STATUS_PUBLISH_ATTEMPTS`. Permanent failures such as 401, 403, 404, and 422 cannot recover with retry, so the worker burns the retry budget (plus exponential backoff sleeps) before surfacing what is actually a token/repository/SHA misconfiguration. Distinguish transient failures (network errors, 429, 5xx) from deterministic HTTP errors and fail fast on the latter.

.github/scripts/platform-testnet-sync/run.cjs
```diff
@@ -0,0 +1,496 @@
...
+
+  const body = {
+    state: status === 'sync_passed' ? 'success' : 'failure',
+    context: STATUS_CONTEXT,
+    description: statusDescription(status, metadata),
+  };
+
+  if (targetUrl) {
+    body.target_url = targetUrl;
+  }
```
thepastaclaw: **🟡 Suggestion: Worker does not validate target_url symmetrically with the receiver workflow**

The status workflow rejects any `target_url` whose origin is not `https://github.com` (workflow lines 87-101), but the worker forwards `metadata.target_url` and `PLATFORM_TESTNET_SYNC_TARGET_URL` verbatim to the createCommitStatus API. The env file is root-owned so the immediate risk is bounded, but the two reporting paths drift: a manual workflow_dispatch with a non-github URL is refused, while a misconfigured or accidentally typo'd env value can publish a commit status whose link points anywhere (http://, third-party CDN, internal phishing). Apply the same `parsedUrl.origin === 'https://github.com'` (or an explicit allowlist) check in the worker before forwarding.

<sub>source: ['claude']</sub>

.github/scripts/platform-testnet-sync/run.cjs
```diff
@@ -0,0 +1,496 @@
...
+
+  const githubToken = process.env.GITHUB_TOKEN || process.env.GH_TOKEN;
+  if (!githubToken) {
+    throw new Error('GITHUB_TOKEN or GH_TOKEN is required to publish commit status');
+  }
+
+  const repository = process.env.GITHUB_REPOSITORY || 'dashpay/platform';
+  if (!repository) {
+    throw new Error('GITHUB_REPOSITORY is required to publish commit status');
+  }
```
thepastaclaw: **💬 Nitpick: Unreachable !repository guard after default fallback**

`const repository = process.env.GITHUB_REPOSITORY || 'dashpay/platform'` always produces a non-empty string, so the following `if (!repository)` throw is dead code. The fallback also silently routes status posts to `dashpay/platform` when `GITHUB_REPOSITORY` is unset, which can mask a worker configured for a fork or staging clone. Either require `GITHUB_REPOSITORY` and drop the default, or drop the dead guard — keeping both is misleading.

```suggestion
  const repository = process.env.GITHUB_REPOSITORY;
  if (!repository) {
    throw new Error('GITHUB_REPOSITORY is required to publish commit status');
  }
```

thepastaclaw (changes requested):
## Code Review

Cumulative review at head 07c7e221. The latest delta is a rename from latest-core-testnet-sync to platform-testnet-sync; no behavior change. All six prior findings remain present in the renamed files and are carried forward. Two additional blocking security findings surfaced this round: the worker harness itself lives in the mutable Platform checkout that the same worker resets to the tested branch (so anyone who can land code on PLATFORM_BRANCH can modify the script that holds GITHUB_TOKEN), and phase commands inherit access to the parent Node process environment via /proc/$PPID/environ, defeating the child-env GITHUB_TOKEN/GH_TOKEN scrub. One symmetry suggestion is added for target_url validation.

🔴 3 blocking | 🟡 4 suggestion(s) | 💬 1 nitpick(s)

<details>

```

In `ops/platform-testnet-sync/platform-testnet-sync.service`:
- [BLOCKING] ops/platform-testnet-sync/platform-testnet-sync.service:10-11: Worker harness executes from the same mutable checkout it resets to the tested branch
  The unit loads `/etc/platform-testnet-sync.env` (containing `GITHUB_TOKEN` with `statuses: write`) and then runs `ops/platform-testnet-sync/run-worker.sh` and `.github/scripts/platform-testnet-sync/run.cjs` from `PLATFORM_REPO_DIR`. That directory is the same checkout that `prepareWorkspace` later resets to `origin/$PLATFORM_BRANCH` (run.cjs:240). Consequently the harness that receives the token is itself branch-controlled code: any commit that lands on `$PLATFORM_BRANCH` can rewrite `run-worker.sh` or `run.cjs` to read `process.env.GITHUB_TOKEN` before any scrubbing happens. The trusted credential-handling code must live outside the mutable checkout — install it as root-owned files under `/usr/local/lib/platform-testnet-sync/` (or similar) and let only the phase commands execute from the checked-out branch.

In `.github/scripts/platform-testnet-sync/run.cjs`:
- [BLOCKING] .github/scripts/platform-testnet-sync/run.cjs:422-429: Phase commands can recover GITHUB_TOKEN from parent process /proc/$PPID/environ
  `sanitizePhaseEnv` and the explicit deletes at run.cjs:428-429 only strip `GITHUB_TOKEN`/`GH_TOKEN` from the child phase environment. The long-lived Node parent still holds those secrets in `process.env` because they were inherited from systemd's `EnvironmentFile` and are needed later for `publishCommitStatus`. On a standard Linux host, a same-UID child can read `/proc/$PPID/environ` (and scan other same-UID `/proc/*/environ`), so any code executed during `baseline-readiness`, `platform-build`, or `platform-sync` — including branch-controlled build scripts — can exfiltrate the token despite the child-env scrub. Move status publishing into an isolated helper (separate UID/container/process) that is invoked after phases finish, or hold the token in a file readable only after phases complete so it never lives in the credential-holding process's environment while untrusted code runs.
- [SUGGESTION] .github/scripts/platform-testnet-sync/run.cjs:142-254: target_sha can describe a commit that was never tested
  `resolveTargetSha` prefers `TARGET_SHA` and `GITHUB_SHA` over `git rev-parse HEAD`, and it runs at main() before workspace preparation (run.cjs:399) and again after the reset inside `prepareWorkspace` (run.cjs:250). If either env var is inherited by the worker, the initial metadata and any failure-path commit status published from the `finally` block point at that inherited SHA even though no phase has actually run against it. If `prepareWorkspace` fails before the reset and `TARGET_SHA` is unset, the published status describes the stale pre-fetch checkout HEAD. Resolve the tested SHA strictly from git after a successful reset, and avoid publishing a commit-specific status before a target commit has actually been selected.
- [SUGGESTION] .github/scripts/platform-testnet-sync/run.cjs:366-388: Status publisher retries permanent HTTP failures
  `publishCommitStatusOnce` throws the same generic `Error` for every non-2xx response, and `publishCommitStatus` retries every thrown error up to `STATUS_PUBLISH_ATTEMPTS`. Permanent failures such as 401, 403, 404, and 422 cannot recover with retry, so the worker burns the retry budget (plus exponential backoff sleeps) before surfacing what is actually a token/repository/SHA misconfiguration. Distinguish transient failures (network errors, 429, 5xx) from deterministic HTTP errors and fail fast on the latter.
- [SUGGESTION] .github/scripts/platform-testnet-sync/run.cjs:339-353: Worker does not validate target_url symmetrically with the receiver workflow
  The status workflow rejects any `target_url` whose origin is not `https://github.com` (workflow lines 87-101), but the worker forwards `metadata.target_url` and `PLATFORM_TESTNET_SYNC_TARGET_URL` verbatim to the createCommitStatus API. The env file is root-owned so the immediate risk is bounded, but the two reporting paths drift: a manual workflow_dispatch with a non-github URL is refused, while a misconfigured or accidentally typo'd env value can publish a commit status whose link points anywhere (http://, third-party CDN, internal phishing). Apply the same `parsedUrl.origin === 'https://github.com'` (or an explicit allowlist) check in the worker before forwarding.

In `ops/platform-testnet-sync/install-worker.sh`:
- [BLOCKING] ops/platform-testnet-sync/install-worker.sh:50-57: Pre-existing env file keeps prior ownership
  When `/etc/platform-testnet-sync.env` already exists, the installer only reapplies `chmod 0600` and leaves whatever ownership the file came with. If the file was created or copied by a non-root account before the installer ran, that account retains read access to `GITHUB_TOKEN` and write access to the phase commands that the systemd unit executes. Because the file carries the statuses-write token and controls the worker's behavior, the root-run installer should force `root:root` ownership in both the create and existing-file paths.

In `ops/platform-testnet-sync/run-worker.sh`:
- [SUGGESTION] ops/platform-testnet-sync/run-worker.sh:13-24: Log pruning runs before the worker lock is acquired
  The destructive `find ... -mtime +RETENTION -exec rm -rf {} +` block at lines 13-18 runs before `flock -n 9` acquires `run.lock` at line 20. A second timer invocation while a sync is in flight prunes first and only then exits because the lock is held. If retention is configured low or a sync exceeds the retention window, the prune can remove the active run directory, breaking `appendLog`/`writeMetadata` in the running worker. Acquire the lock before pruning so only the active worker performs retention cleanup.
```
</details>

.github/workflows/platform-testnet-sync-status.yml
```diff
@@ -0,0 +1,135 @@
...
+
+permissions:
+  contents: read
+  statuses: write
```
thepastaclaw, line 40: **🔴 Blocking: workflow_dispatch path can forge 'Platform testnet sync' commit statuses**

The job is granted statuses: write and accepts workflow_dispatch input directly for status, target_sha, target_url, core_version, platform_sha, and completed_at. The only checks performed are the enum match on status, a 40-hex regex on the SHA, repos.getCommit (which only proves the SHA exists in the repo), and an origin restriction on target_url. None of those prove the external sync worker actually ran against that commit. Any account with workflow-dispatch privilege on this repo (broader than the operator credential the design intends) can therefore publish state=success, context='Platform testnet sync' on an arbitrary commit. If branch protection, dashboards, or release automation treat this context as authoritative — and docs/PLATFORM_TESTNET_SYNC.md presents it that way — this is a status-forgery path that bypasses the worker boundary. Either drop the workflow_dispatch trigger (rely on repository_dispatch from the operator-held credential only), or gate the manual path behind a protected environment with a required reviewer and a restricted actor allowlist. Note: per the file-handling constraint, suggestion field is null; do not interpret that as agreement to leave the trigger in place.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):
## Code Review

Cumulative PR is now narrow: a status receiver workflow, README badge, and docs. The previously flagged worker harness (run.cjs, systemd unit, install/run shell, env example) was deleted in 3546112a, resolving 8 of 9 prior findings. One prior issue remains: the surviving workflow_dispatch entry on platform-testnet-sync-status.yml lets any actor with workflow-dispatch privilege forge a 'Sync Passed' commit status on an arbitrary in-repo SHA, since payload validation only proves the SHA exists. Two independent Codex passes re-flagged this; carrying forward as blocking.

🔴 1 blocking

<details>

```

In `.github/workflows/platform-testnet-sync-status.yml`:
- [BLOCKING] .github/workflows/platform-testnet-sync-status.yml:7-40: workflow_dispatch path can forge 'Platform testnet sync' commit statuses
  The job is granted statuses: write and accepts workflow_dispatch input directly for status, target_sha, target_url, core_version, platform_sha, and completed_at. The only checks performed are the enum match on status, a 40-hex regex on the SHA, repos.getCommit (which only proves the SHA exists in the repo), and an origin restriction on target_url. None of those prove the external sync worker actually ran against that commit. Any account with workflow-dispatch privilege on this repo (broader than the operator credential the design intends) can therefore publish state=success, context='Platform testnet sync' on an arbitrary commit. If branch protection, dashboards, or release automation treat this context as authoritative — and docs/PLATFORM_TESTNET_SYNC.md presents it that way — this is a status-forgery path that bypasses the worker boundary. Either drop the workflow_dispatch trigger (rely on repository_dispatch from the operator-held credential only), or gate the manual path behind a protected environment with a required reviewer and a restricted actor allowlist. Note: per the file-handling constraint, suggestion field is null; do not interpret that as agreement to leave the trigger in place.
```
</details>

ops/platform-testnet-sync/install-worker.sh
thepastaclaw: Resolved in `3546112` — _Pre-existing env file keeps prior ownership_ no longer present.

.github/scripts/platform-testnet-sync/run.cjs
thepastaclaw: Resolved in `3546112` — _Unreachable !repository guard after default fallback_ no longer present.

.github/scripts/platform-testnet-sync/run.cjs
thepastaclaw: Resolved in `3546112` — _target_sha can describe a commit that was never tested_ no longer present.

ops/platform-testnet-sync/platform-testnet-sync.service
thepastaclaw: Resolved in `3546112` — _Worker harness executes from the same mutable checkout it resets to the tested branch_ no longer present.

ops/platform-testnet-sync/run-worker.sh
thepastaclaw: Resolved in `3546112` — _Log pruning runs before the worker lock is acquired_ no longer present.

.github/scripts/platform-testnet-sync/run.cjs
thepastaclaw: Resolved in `3546112` — _Worker does not validate target_url symmetrically with the receiver workflow_ no longer present.

.github/scripts/platform-testnet-sync/run.cjs
thepastaclaw: Resolved in `3546112` — _Phase commands can recover GITHUB_TOKEN from parent process /proc/$PPID/environ_ no longer present.

.github/scripts/platform-testnet-sync/run.cjs
thepastaclaw: Resolved in `3546112` — _Status publisher retries permanent HTTP failures_ no longer present.

PR #3959, opened 2026-06-24
fix(swift-example-app): make Platform Sync "Clear" actually clear synced data

llbartekll:

In the SwiftExampleApp, the **Clear** button in the *Platform Sync Status* section (BLAST / DIP-17 address sync) did not remove any data — resyncing afterwards took ~2 seconds, so it looked like Clear did nothing.

Root cause: the button was wired to `PlatformBalanceSyncService.clearDisplay()`, which only zeroes the in-memory `@Published` UI vars. The synced data actually lives in three places, none of which Clear touched:

1. **SwiftData rows** — `PersistentPlatformAddress` (balances on durable derivation rows) and `PersistentPlatformAddressesSyncState` (the network-scoped incremental watermark).
2. **In-memory Rust watermark + balance seed** in `PlatformPaymentAddressProvider`. This is what makes the resync fast: the next `sync_now` resumes incrementally from the in-memory height instead of doing a full rescan. Because it lives in Rust process memory, a Swift-only delete cannot reset it.
3. **In-memory managed-account balance map** (key-wallet `ManagedPlatformAccount.address_balances`) — what `total_credits()` / `addresses_with_balances()` and the Transfer/Withdraw spend paths read.

The two sibling Clear buttons in the same view already do this correctly (Core Sync → `clearSpvStorage()`; Shielded Sync → `ShieldedService.clearLocalState(...)`). This PR mirrors that pattern for platform addresses.

## What was done?

A real clear wired to the button, bottom-up across Rust → FFI → Swift. The reset is **manager-level (all wallets on the active network)** to match the host-side clear, the background sync loop is **quiesced** first so no in-flight pass re-persists after the reset, and it is left stopped so data stays cleared until the user explicitly resyncs.

**Rust (`platform-wallet`)**
- `PlatformPaymentAddressProvider::reset_sync_state` — zero the incremental watermark and drop the cached `found`/`absent` seed so the next pass is a full rescan (zeroing the scalars alone is not enough; `found` doubles as the next-pass seed). The address bijection is intentionally preserved to avoid needless re-derivation.
- `PlatformAddressWallet::reset_sync_state` — the provider reset **plus** clears each managed account's balances via `all_platform_payment_managed_accounts_mut().clear_balances()`, so spend paths / `total_credits()` are immediately consistent.
- `PlatformWalletManager::reset_platform_address_sync_state` — quiesce the platform-address sync manager, snapshot the wallets under a short read lock, reset each.

**FFI (`rs-platform-wallet-ffi`)**
- `platform_wallet_manager_platform_address_sync_reset` — a thin pass-through mirroring `platform_wallet_manager_shielded_clear` (cbindgen auto-exports it; no manual header edit).

```c
struct PlatformWalletFFIResult platform_wallet_manager_platform_address_sync_reset(Handle handle);
```

**Swift SDK (`swift-sdk`)**
- `PlatformWalletManager.resetPlatformAddressSyncState()` — thin wrapper, run off the main actor.
- **Generation guard** mirroring the shielded path: the completion callback snapshots `platformAddressSyncGeneration` on the FFI thread before the main-actor hop; `handlePlatformAddressSyncCompleted` drops events whose snapshot predates a `stop`/`reset` bump. `resetPlatformAddressSyncState` and `stopPlatformAddressSync` bump the counter, and `resetPlatformAddressPublishedMirror()` nils the retained `lastPlatformAddressSyncEvent` / `platformAddressSyncIsSyncing` so a later `configure()` re-subscribe can't replay stale sync state.

**SwiftExampleApp**
- `PlatformBalanceSyncService.clearLocalState(modelContext:network:walletIdsOnNetwork:)` and the rewired Clear button. It:
  - calls the Rust reset first, **failing closed** (returns with `lastError` set on reset or persistence failure; `clearDisplay()` runs only on full success);
  - is **scoped to the active network** — `PersistentPlatformAddress` via the `walletIdsOnNetwork` pivot the view already computes, `PersistentPlatformAddressesSyncState` via `networkRaw` — so a testnet Clear can't wipe mainnet/devnet state;
  - **zeroes the volatile fields of `PersistentPlatformAddress` rows in place rather than deleting them.** Those rows carry durable derivation metadata (bech32m address, hash, public key, account/address index, derivation path) that no sync re-emits in-session, and the BLAST balance callback `persistAddressBalances` only *updates* existing rows (it skips missing ones by `addressHash`). Deleting them would leave the next Sync Now with nothing to upsert against, so the UI would stay empty until app restart. Only the network-keyed `PersistentPlatformAddressesSyncState` watermark row is deleted, which is what forces the full rescan.

## How Has This Been Tested?

- `cargo fmt --check` / `cargo clippy -p platform-wallet -p platform-wallet-ffi` clean.
- `cargo test -p platform-wallet --lib` → **203 passed**, incl. new `reset_sync_state_clears_watermark_and_seed`.
- New Swift tests, all passing on the iPhone 17 simulator:
  - `PlatformBalanceSyncServiceClearTests` — active-network rows zeroed in place with durable metadata preserved, watermark deleted, other networks untouched.
  - `PlatformAddressSyncGenerationTests` — stale completion dropped after a generation bump, current-generation completion published, straggler doesn't clobber, and the retained published mirror is dropped on reset.
- `packages/swift-sdk/build_ios.sh --target sim` → BUILD SUCCEEDED; the new FFI symbol is present in the regenerated header. Clean SwiftExampleApp simulator build → `** BUILD SUCCEEDED **`.

**Not yet runtime-verified:** no funded testnet platform account was available in this session, so the full UI flow (sync funded DIP-17 wallet → Clear → balances/height hit 0 → Sync Now performs a visibly slow full rescan that re-populates balances) was not driven end-to-end on a booted simulator.

## Breaking Changes

None. Additive FFI symbol + new Swift wrapper/method; no consensus or public-API breaking changes.

## Checklist:

### DIFF
--- a/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerAddressSync.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerAddressSync.swift
@@ -80,8 +80,13 @@ private func platformAddressSyncCompletedCallback(
         walletResults: results
     )
 
+    // Snapshot the generation now, on the FFI callback thread, BEFORE the
+    // main-actor hop, so a stop/reset that bumps the counter after this
+    // point invalidates the trailing event (mirrors the shielded path).
+    let generation = handler.manager?.platformAddressSyncGeneration.current() ?? 0
+
     Task { @MainActor [weak manager = handler.manager] in
-        manager?.handlePlatformAddressSyncCompleted(event)
+        manager?.handlePlatformAddressSyncCompleted(event, generation: generation)
     }
 }
 
@@ -100,7 +105,14 @@ private extension AddressSyncMetrics {
 }
 
 extension PlatformWalletManager {
-    func handlePlatformAddressSyncCompleted(_ event: PlatformAddressSyncEvent) {
+    func handlePlatformAddressSyncCompleted(_ event: PlatformAddressSyncEvent, generation: UInt64) {
+        // Drop a trailing event the Rust drain already dispatched but the
+        // main actor only delivers after a stop/reset bumped the counter —
+        // its snapshot predates the bump. Without this, a completion from a
+        // pass drained by `resetPlatformAddressSyncState` (Clear) repaints
+        // chain-tip height, last-sync time, and metrics over the freshly
+        // cleared UI. Mirrors the shielded guard.
+        guard generation == platformAddressSyncGeneration.current() else { return }
         lastPlatformAddressSyncEvent = event
     }
 
@@ -132,6 +144,10 @@ extension PlatformWalletManager {
         }
 
         try platform_wallet_manager_platform_address_sync_stop(handle).check()
+        // The Rust drain returned; bump the generation so any trailing
+        // completion the main actor delivers after this point is dropped
+        // (its snapshot predates this bump). Mirrors the shielded stop.
+        platformAddressSyncGeneration.bump()
     }
 
     public func isPlatformAddressSyncRunning() throws -> Bool {
@@ -241,5 +257,10 @@ extension PlatformWalletManager {
         try await Task.detached(priority: .userInitiated) {
             try platform_wallet_manager_platform_address_sync_reset(handle).check()
         }.value
+        // The Rust reset quiesced + drained the in-flight pass; bump the
+        // generation so a trailing completion captured before this point
+        // (and delivered onto the main actor after Clear) is dropped
+        // instead of repainting the just-cleared sync-status UI.
+        platformAddressSyncGeneration.bump()
     }
 }
--- a/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerAddressSync.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerAddressSync.swift
@@ -262,5 +262,10 @@ extension PlatformWalletManager {
         // (and delivered onto the main actor after Clear) is dropped
         // instead of repainting the just-cleared sync-status UI.
         platformAddressSyncGeneration.bump()
+        // Drop the retained published mirror too: the generation guard only
+        // blocks future stale callbacks, but a later `configure()` would
+        // replay the current `@Published` value to its fresh subscriber and
+        // repaint the cleared UI.
+        resetPlatformAddressPublishedMirror()
     }
 }
--- a/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerAddressSync.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerAddressSync.swift
@@ -219,4 +219,27 @@ extension PlatformWalletManager {
             try platform_wallet_manager_platform_address_sync_sync_now(handle).check()
         }.value
     }
+
+    /// Reset the platform-address (BLAST/DIP-17) incremental-sync
+    /// watermark and drop every cached balance across all registered
+    /// wallets, forcing a full rescan on the next sync. Backs the
+    /// SwiftExampleApp Platform Sync "Clear" button.
+    ///
+    /// Quiesces the background sync loop before resetting (so no
+    /// in-flight pass re-writes the watermark) and leaves it stopped —
+    /// callers re-arm via `startPlatformAddressSync` or one-shot
+    /// `syncPlatformAddressNow`. Runs off the main actor because the
+    /// quiesce drains any in-flight pass.
+    public func resetPlatformAddressSyncState() async throws {
+        guard isConfigured, handle != NULL_HANDLE else {
+            throw PlatformWalletError.invalidHandle(
+                "PlatformWalletManager not configured"
+            )
+        }
+
+        let handle = self.handle
+        try await Task.detached(priority: .userInitiated) {
+            try platform_wallet_manager_platform_address_sync_reset(handle).check()
+        }.value
+    }
 }
--- a/packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/PlatformBalanceSyncService.swift
+++ b/packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/PlatformBalanceSyncService.swift
@@ -195,18 +195,25 @@ class PlatformBalanceSyncService: ObservableObject {
     ///
     /// Plain [`clearDisplay`] only zeroes the in-memory `@Published`
     /// mirror, so the next sync resumed from the surviving watermark in
-    /// ~2s (the "Clear didn't work" symptom). This wipes the stores the
+    /// ~2s (the "Clear didn't work" symptom). This clears the stores the
     /// synced data actually lives in, mirroring
     /// `ShieldedService.clearLocalState`: Rust-side reset FIRST (so the
-    /// next sync can't re-persist stale rows), then the SwiftData wipe,
+    /// next sync can't re-persist stale rows), then the SwiftData clear,
     /// then the published-mirror reset.
     ///
+    /// The `PersistentPlatformAddress` rows are **zeroed in place, not
+    /// deleted** — they carry durable derivation metadata that no sync
+    /// re-emits in-session and that the balance callback only updates
+    /// (never creates), so deleting them would leave the UI empty until
+    /// app restart. Only the volatile balance/sync fields are cleared;
+    /// the network-keyed `PersistentPlatformAddressesSyncState` watermark
+    /// row is deleted to force a full rescan.
+    ///
     /// Scoped to the **active network**. The SwiftData store holds rows
-    /// for every network at once (the UI filters them by network), so a
-    /// blanket `delete(model:)` would also erase other networks' cached
-    /// platform state. `PersistentPlatformAddress` carries no network
-    /// column, so it's scoped via `walletIdsOnNetwork` — the same
-    /// wallet-id-per-network pivot the view uses;
+    /// for every network at once (the UI filters them by network), so
+    /// clearing must not touch other networks' state. `PersistentPlatformAddress`
+    /// carries no network column, so it's scoped via `walletIdsOnNetwork`
+    /// — the same wallet-id-per-network pivot the view uses;
     /// `PersistentPlatformAddressesSyncState` is network-keyed and is
     /// scoped by `networkRaw`. This matches the manager-level Rust reset,
     /// which only touches the active network's registered wallets.
@@ -237,15 +244,35 @@ class PlatformBalanceSyncService: ObservableObject {
             }
         }
 
-        // 2) Delete this network's platform-address rows: the cached
-        //    per-address balances (scoped via the network's wallet-id
-        //    set) and the network-keyed sync-state watermark.
+        // 2) Zero the volatile balance/sync fields of this network's
+        //    address rows IN PLACE — do NOT delete them. A
+        //    `PersistentPlatformAddress` row is durable derivation state
+        //    (bech32m address, 20-byte hash, 33-byte public key,
+        //    account/address index, derivation path), not just a balance
+        //    cache. The BLAST balance callback `persistAddressBalances`
+        //    only *updates* existing rows (it skips by `addressHash` when
+        //    the row is missing), and no sync re-emits the rows in-session:
+        //    Rust's `reset_sync_state` preserves the address bijection, so
+        //    there is no pool extension to re-fire the address-emit path.
+        //    Deleting would therefore leave the next "Sync Now" with
+        //    nothing to upsert against — balances (and the spend/signing
+        //    derivation lookups) would stay gone until an app restart
+        //    re-seeds the rows. Zeroing clears what the user sees while
+        //    keeping the metadata the rescan needs.
         do {
             let addresses = try modelContext.fetch(FetchDescriptor<PersistentPlatformAddress>())
             for row in addresses where walletIdsOnNetwork.contains(row.walletId) {
-                modelContext.delete(row)
+                row.balance = 0
+                row.nonce = 0
+                row.isUsed = false
+                row.firstSeenHeight = 0
+                row.lastSeenHeight = 0
+                row.lastUpdated = Date()
             }
 
+            // The sync-state watermark is pure volatile sync state (no
+            // durable metadata) and is what actually forces a full rescan,
+            // so delete it outright — the next sync re-creates it.
             let networkRaw = network.rawValue
             let syncStates = try modelContext.fetch(FetchDescriptor<PersistentPlatformAddressesSyncState>())
             for row in syncStates where row.networkRaw == networkRaw {
@@ -254,7 +281,7 @@ class PlatformBalanceSyncService: ObservableObject {
 
             try modelContext.save()
         } catch {
-            lastError = "Failed to wipe persisted platform-address state: \(error.localizedDescription)"
+            lastError = "Failed to clear persisted platform-address state: \(error.localizedDescription)"
             SDKLogger.error(lastError ?? "")
             return
         }
--- a/packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/PlatformBalanceSyncService.swift
+++ b/packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/PlatformBalanceSyncService.swift
@@ -8,6 +8,7 @@
 import Foundation
 import SwiftUI
 import Combine
+import SwiftData
 import SwiftDashSDK
 
 /// Observable service managing BLAST address balance sync UI state.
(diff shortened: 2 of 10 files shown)


thepastaclaw:
🕓 Ready for review — 7 ahead in queue (commit d1b02224)
_Queue position: 8/44_

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/PlatformBalanceSyncService.swift
```diff
@@ -189,6 +190,54 @@ class PlatformBalanceSyncService: ObservableObject {
...
+            try modelContext.delete(model: PersistentPlatformAddress.self)
+            try modelContext.delete(model: PersistentPlatformAddressesSyncState.self)
+            try modelContext.save()
+        } catch {
+            lastError = "Failed to wipe persisted platform-address state: \(error.localizedDescription)"
+            SDKLogger.error(lastError ?? "")
+        }
+
+        // 3) Zero the published display mirror.
+        clearDisplay()
```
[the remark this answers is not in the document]
llbartekll: Addressed in 8d3ca393e7. `clearLocalState` now fails closed: both the Rust reset and the SwiftData delete `return` early on failure with `lastError` set, and `clearDisplay()` runs only on full success — matching the ShieldedService precedent (which returns on delete failure). I made the reset path fail closed too, since the reset is load-bearing here (a surviving in-memory watermark reintroduces the fast-resync bug).

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/PlatformBalanceSyncService.swift
```diff
@@ -189,6 +190,54 @@ class PlatformBalanceSyncService: ObservableObject {
...
+            }
+        }
+
+        // 2) Delete every platform-address SwiftData row across all
+        //    wallets on this device: the cached per-address balances and
+        //    the network-scoped sync-state watermark.
+        do {
+            try modelContext.delete(model: PersistentPlatformAddress.self)
+            try modelContext.delete(model: PersistentPlatformAddressesSyncState.self)
+            try modelContext.save()
```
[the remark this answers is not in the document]
llbartekll: Addressed in 8d3ca393e7. The delete is now scoped to the active network: `PersistentPlatformAddress` via the `walletIdsOnNetwork` pivot the view already uses (the model has no network column), and `PersistentPlatformAddressesSyncState` via `networkRaw`. This matches the manager-level Rust reset (active-network wallets only) and stops a testnet Clear from wiping mainnet/devnet cached state. The unit test now seeds two networks and asserts the inactive one survives.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerAddressSync.swift
```diff
@@ -219,4 +235,32 @@ extension PlatformWalletManager {
...
+
+        let handle = self.handle
+        try await Task.detached(priority: .userInitiated) {
+            try platform_wallet_manager_platform_address_sync_reset(handle).check()
+        }.value
+        // The Rust reset quiesced + drained the in-flight pass; bump the
+        // generation so a trailing completion captured before this point
+        // (and delivered onto the main actor after Clear) is dropped
+        // instead of repainting the just-cleared sync-status UI.
+        platformAddressSyncGeneration.bump()
```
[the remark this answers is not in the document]
llbartekll: Fixed in 93f3bf0b9b. `resetPlatformAddressSyncState()` now calls `resetPlatformAddressPublishedMirror()` after the Rust drain, which nils `lastPlatformAddressSyncEvent` and sets `platformAddressSyncIsSyncing = false`. The helper lives in `PlatformWalletManager.swift` because `platformAddressSyncIsSyncing` is `private(set)`. Added `testResetPublishedMirrorDropsRetainedEvent` covering the replay-on-re-subscribe case.

Note: `clearShielded()` has the same latent pattern — it bumps `shieldedSyncGeneration` but does not nil `lastShieldedSyncEvent`, so a re-bind could replay it. Out of scope for this PR; flagging for a follow-up.

thepastaclaw (changes requested):
## Code Review

PR cleanly threads a real three-store wipe behind the Platform Sync 'Clear' button across Rust, FFI, Swift SDK, and the example app, and the lock ordering / quiesce-then-reset design is sound. However, the SwiftData wipe deletes durable address metadata (derivation path, public key, account/address index) that the BLAST balance callback hard-depends on (continues silently when the row is missing), and the address-emit path only re-fires on pool extension or wallet reload — so during a live session after Clear, balances cannot be re-persisted and spend/signing paths lose their derivation lookups until app restart. Two smaller issues: a stale completion-event race that the sibling shielded path guards against with a generation counter, and a SwiftData-wipe failure message that clearDisplay() immediately overwrites.

<details>

```

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/PlatformBalanceSyncService.swift`:
- [BLOCKING] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/PlatformBalanceSyncService.swift:228-231: Clear wipes durable address metadata that the balance callback and spend paths depend on
  `PersistentPlatformAddress` is not a balance cache — it also stores the bech32m address, 20-byte hash, account/address index, 33-byte public key, and derivation path (see PersistentPlatformAddress.swift:33-71). After this `delete(model: PersistentPlatformAddress.self)`:

  1. The Rust-side `addresses` bijection is intentionally preserved (see `PlatformPaymentAddressProvider::reset_sync_state` docstring), so the `on_persist_account_address_pools_fn` re-emit path does NOT fire on the next sync — it only fires on pool extension or wallet reload.
  2. The next full rescan emits balance changesets via `persistAddressBalances` (PlatformWalletPersistenceHandler.swift:88). That callback receives only `addressHash` and at line 97-99 hits `guard let existing = ... else { continue }` — silently dropping the update when the row is missing. The accompanying comment ("the next address-emit pass will bring the row back") is wrong for this scenario.
  3. Spend / signing paths that look up derivation metadata from `PersistentPlatformAddress` (account index, derivation path, public key) lose those lookups until app restart re-seeds the rows from Rust persistence.

  Fix: zero the balance / nonce / seen-height fields in-place rather than deleting the rows. The watermark wipe via `PersistentPlatformAddressesSyncState` is what actually forces a full rescan; the address rows themselves carry irreplaceable derivation metadata.
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/PlatformBalanceSyncService.swift:232-238: SwiftData-wipe failure message is immediately erased by clearDisplay()
  On a SwiftData wipe failure, the catch sets `lastError = "Failed to wipe persisted platform-address state: ..."`, then control falls through to `clearDisplay()` at line 238, which unconditionally sets `lastError = nil` (see clearDisplay() line 171). The user never sees the wipe failure, and the published mirror is reset to a fully-cleared state even though the on-disk rows still exist — misrepresenting state. `ShieldedService.clearLocalState` handles this by returning on wipe failure; mirror that here so the error is surfaced and the display isn't lied about.

In `packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerAddressSync.swift`:
- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerAddressSync.swift:83-86: Stale completion-event Task can repopulate display after Clear
  The Rust-side `quiesce()` waits for `platformAddressSyncCompletedCallback` (the C function) to return, but that callback only enqueues `Task { @MainActor ... handlePlatformAddressSyncCompleted(event) }` and returns immediately. If the callback fires near the tail of quiesce, the `@MainActor` Task can be enqueued after `clearLocalState` resumes from its await, then run after `clearDisplay()` completes — restoring `lastSyncHeight`, `chainTipHeight`, `lastSyncTime`, `syncCountSinceLaunch`, metrics counters, and triggering `refreshBalanceSnapshot()` (PlatformBalanceSyncService.swift:279-309). The user-visible impact is small (transient stale metrics until next sync), but it is the same race the sibling shielded path solves with a generation counter snapshotted in the FFI callback and bumped on clear/reset. Worth replicating here since the rest of the PR explicitly mirrors the shielded pattern.
```
</details>

---

_Note: inline comments could not be attached because the PR head moved after this exact-SHA review was claimed; preserving the review as a top-level exact-commit review._

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/PlatformBalanceSyncService.swift
```diff
@@ -189,6 +190,79 @@ class PlatformBalanceSyncService: ObservableObject {
...
+        }
+
+        // 2) Delete this network's platform-address rows: the cached
+        //    per-address balances (scoped via the network's wallet-id
+        //    set) and the network-keyed sync-state watermark.
+        do {
+            let addresses = try modelContext.fetch(FetchDescriptor<PersistentPlatformAddress>())
+            for row in addresses where walletIdsOnNetwork.contains(row.walletId) {
+                modelContext.delete(row)
+            }
```
thepastaclaw, line 271: **🔴 Blocking: Clear deletes durable platform-address metadata that the balance callback cannot rebuild**

`PersistentPlatformAddress` is not a balance cache — it stores the bech32m address, 20-byte hash, derivation path, account/address index, and 33-byte public key (see PersistentPlatformAddress.swift:33-71). On the Rust side, `PlatformPaymentAddressProvider::reset_sync_state` intentionally preserves the `addresses` bijection (provider.rs:491-494), and `AccountAddressPoolEntry`s are only pushed during initial `register_wallet` (wallet_lifecycle.rs:312) — no BLAST sync round, pool extension, or used-flip re-emits them. The balance callback `persistAddressBalances` keys by `addressHash` and explicitly `continue`s when the row is missing (PlatformWalletPersistenceHandler.swift:97-99). So after Clear deletes the rows, the next BLAST sync updates Rust-side balances but cannot re-create the SwiftData rows, and the Platform Sync UI stays at zero for the remainder of the live session. The fail-closed handling and active-network scoping added in this revision are correct but don't address this root cause. Zero the volatile fields in place instead, and still delete the network-scoped `PersistentPlatformAddressesSyncState` row to force a full rescan.

```suggestion
            let addresses = try modelContext.fetch(FetchDescriptor<PersistentPlatformAddress>())
            for row in addresses where walletIdsOnNetwork.contains(row.walletId) {
                row.balance = 0
                row.nonce = 0
                row.isUsed = false
                row.firstSeenHeight = 0
                row.lastSeenHeight = 0
                row.lastUpdated = Date()
            }
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):
## Code Review

Reconciled three prior findings against head 7ee50343: the fail-closed Clear path (prior-2) and the generation-guarded stale completion (prior-3) are FIXED; prior-1 is STILL VALID and remains the sole blocker. The Clear flow now scopes the SwiftData wipe to the active network, but it still deletes `PersistentPlatformAddress` rows — durable derivation/address metadata that Rust does not re-emit on subsequent syncs — so the next BLAST sync's balance callback (which only carries `addressHash`) has nothing to upsert against and the UI stays empty. Recommend zeroing balance/nonce/seen-height fields in place and only deleting the network-scoped sync-state row.

🔴 1 blocking

<details>

```

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/PlatformBalanceSyncService.swift`:
- [BLOCKING] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/PlatformBalanceSyncService.swift:244-247: Clear deletes durable platform-address metadata that the balance callback cannot rebuild
  `PersistentPlatformAddress` is not a balance cache — it stores the bech32m address, 20-byte hash, derivation path, account/address index, and 33-byte public key (see PersistentPlatformAddress.swift:33-71). On the Rust side, `PlatformPaymentAddressProvider::reset_sync_state` intentionally preserves the `addresses` bijection (provider.rs:491-494), and `AccountAddressPoolEntry`s are only pushed during initial `register_wallet` (wallet_lifecycle.rs:312) — no BLAST sync round, pool extension, or used-flip re-emits them. The balance callback `persistAddressBalances` keys by `addressHash` and explicitly `continue`s when the row is missing (PlatformWalletPersistenceHandler.swift:97-99). So after Clear deletes the rows, the next BLAST sync updates Rust-side balances but cannot re-create the SwiftData rows, and the Platform Sync UI stays at zero for the remainder of the live session. The fail-closed handling and active-network scoping added in this revision are correct but don't address this root cause. Zero the volatile fields in place instead, and still delete the network-scoped `PersistentPlatformAddressesSyncState` row to force a full rescan.
```
</details>

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/PlatformBalanceSyncService.swift
thepastaclaw, line 271: **🔴 Blocking: Clear deletes platform-address derivation metadata that no later sync can rebuild**

PersistentPlatformAddress is durable derivation state — bech32m address, 20-byte hash, 33-byte compressed public key, account/address index, derivation path (see PersistentPlatformAddress.swift:33-71) — not a balance cache. Deleting these rows on Clear creates an asymmetry that the rest of the pipeline cannot recover from in-session:

1. The Rust-side reset (provider.rs:495-506) intentionally preserves the `addresses` bijection and only clears `found`/`absent`/watermark scalars — its own doc comment notes this is to avoid 'needless re-derivation'.
2. `sync_balances` (sync.rs:110-165) emits only a `PlatformAddressChangeSet { addresses: <balance diff>, ... }`. The `on_persist_account_address_pools_fn` callback that creates `PersistentPlatformAddress` rows is fired only at wallet creation and gap-limit extension, not on every sync.
3. `persistAddressBalances` (PlatformWalletPersistenceHandler.swift:88-114) keys by `addressHash` and `guard let existing = ... else { continue }` skips when the row is missing.

Net effect: after Clear, the next 'Sync Now' produces balance updates against now-missing rows that silently no-op in SwiftData. The Platform Sync UI stays empty (and the spend/signing path loses its derivation-path source) until app restart triggers re-derivation through the wallet load path. This contradicts the file's own contract: 'data stays cleared until the user explicitly resyncs'.

Fix by zeroing only the volatile balance/sync fields in place for `walletIdsOnNetwork` and keeping the durable derivation row, while still deleting the network-scoped `PersistentPlatformAddressesSyncState` row to force a full rescan. The full rescan + preserved rows together give Clear the semantics the doc comment promises.

```suggestion
            let addresses = try modelContext.fetch(FetchDescriptor<PersistentPlatformAddress>())
            for row in addresses where walletIdsOnNetwork.contains(row.walletId) {
                row.balance = 0
                row.nonce = 0
                row.isUsed = false
                row.firstSeenHeight = 0
                row.lastSeenHeight = 0
                row.lastUpdated = Date()
            }
```

thepastaclaw (changes requested):
## Code Review

The latest delta correctly resolves the published-mirror replay issue flagged in the prior review (resetPlatformAddressPublishedMirror is now called after the Rust drain). However, the prior blocking finding about Clear deleting durable PersistentPlatformAddress derivation metadata remains valid in this head: persistAddressBalances keys by addressHash and skips missing rows, Rust's reset_sync_state preserves the addresses bijection without re-emitting account_address_pools, and sync_balances emits only a balance diff — so an explicit 'Sync Now' after Clear silently no-ops in SwiftData and the UI stays empty until app restart. Carrying that finding forward; no new in-scope issues.

🔴 1 blocking

<details>

```

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/PlatformBalanceSyncService.swift`:
- [BLOCKING] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/PlatformBalanceSyncService.swift:244-247: Clear deletes platform-address derivation metadata that no later sync can rebuild
  PersistentPlatformAddress is durable derivation state — bech32m address, 20-byte hash, 33-byte compressed public key, account/address index, derivation path (see PersistentPlatformAddress.swift:33-71) — not a balance cache. Deleting these rows on Clear creates an asymmetry that the rest of the pipeline cannot recover from in-session:

  1. The Rust-side reset (provider.rs:495-506) intentionally preserves the `addresses` bijection and only clears `found`/`absent`/watermark scalars — its own doc comment notes this is to avoid 'needless re-derivation'.
  2. `sync_balances` (sync.rs:110-165) emits only a `PlatformAddressChangeSet { addresses: <balance diff>, ... }`. The `on_persist_account_address_pools_fn` callback that creates `PersistentPlatformAddress` rows is fired only at wallet creation and gap-limit extension, not on every sync.
  3. `persistAddressBalances` (PlatformWalletPersistenceHandler.swift:88-114) keys by `addressHash` and `guard let existing = ... else { continue }` skips when the row is missing.

  Net effect: after Clear, the next 'Sync Now' produces balance updates against now-missing rows that silently no-op in SwiftData. The Platform Sync UI stays empty (and the spend/signing path loses its derivation-path source) until app restart triggers re-derivation through the wallet load path. This contradicts the file's own contract: 'data stays cleared until the user explicitly resyncs'.

  Fix by zeroing only the volatile balance/sync fields in place for `walletIdsOnNetwork` and keeping the durable derivation row, while still deleting the network-scoped `PersistentPlatformAddressesSyncState` row to force a full rescan. The full rescan + preserved rows together give Clear the semantics the doc comment promises.
```
</details>

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/PlatformBalanceSyncService.swift
thepastaclaw: Resolved in this update — _Clear deletes durable platform-address metadata that the balance callback cannot rebuild_ no longer present.

llbartekll: Fixed in d1b022249a. Confirmed the root cause against the code: `persistAddressBalances` only updates existing rows (skips by `addressHash` when missing, per its own doc at PlatformWalletPersistenceHandler.swift:78-99), and the Rust `reset_sync_state` preserves the address bijection so no sync re-emits the rows. `clearLocalState` now zeroes the volatile fields (balance, nonce, isUsed, firstSeenHeight, lastSeenHeight, lastUpdated) of the active-network rows IN PLACE — keeping the durable derivation metadata (address, hash, pubkey, indices, derivation path) — and still deletes the network-scoped `PersistentPlatformAddressesSyncState` watermark to force a full rescan. The full rescan then re-persists balances against the preserved rows. Test reworked to assert the row survives with volatile fields zeroed + durable metadata intact, watermark deleted, other networks untouched.

thepastaclaw (approved):
## Code Review

The prior blocking finding from 93f3bf0b is fully resolved in d1b02224: PlatformBalanceSyncService.clearLocalState now zero-resets the volatile fields of PersistentPlatformAddress rows in place (balance/nonce/isUsed/firstSeenHeight/lastSeenHeight/lastUpdated) instead of deleting them, while still removing the network-scoped PersistentPlatformAddressesSyncState watermark so the next Sync Now performs a full rescan. The updated test asserts zero-in-place semantics and active-network scoping. No new in-scope defects found across the Rust/FFI/Swift reset stack.

_Coverage note: Claude general and Claude FFI reviewer passes ran and were verified; Codex general and Codex FFI ACP passes failed with `Authentication required`, so no Codex findings were available for this exact-SHA review._

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/PlatformBalanceSyncService.swift
thepastaclaw: Resolved in `d1b0222` — _Clear deletes platform-address derivation metadata that no later sync can rebuild_ no longer present.

shumkov (approved):
Looks good to me 👍

thepastaclaw:
While processing the review-bot escalation, I rechecked CodeRabbit's latest current-head finding against `d1b0222`. The old escalated inline comments are addressed, but the later "Block manual sync while Clear is in progress" concern still looks valid: `clearLocalState()` awaits `resetPlatformAddressSyncState()` and then mutates SwiftData, while `performSync()` only gates on `isSyncing` and does not know Clear is in progress.

Suggested fix is small: add a dedicated clearing flag around `clearLocalState()` and have `performSync()` early-return while that flag is set. Given this PR is specifically about Clear semantics, I would address or explicitly accept that race before merge.

thepastaclaw:
Post-merge follow-up note: while handling the latest CodeRabbit notification, I verified the “Sync Now can start while Clear is in progress” race was still valid at merged head `d1b02224`. I prepared the minimal guard fix in `thepastaclaw/platform@16b2bc386f1763250e878871af315438d04cb003`, but I could not push to the in-repo PR branch and the PR merged before the fix could land. I opened tracker follow-up https://github.com/thepastaclaw/tracker/issues/1712 with the branch/validation details so it can be applied onto `v4.0-dev`.

PR #3969, opened 2026-06-29
fix(platform-wallet): reconcile platform-address balances after top-up-from-addresses

shumkov:

A user reported (2026-06-27, screenshots) that an identity **top-up failed** with:

> Failed to top up identity from addresses: Protocol error: Insufficient combined address balances: total available is less than required 1500000

…even though the wallet displayed **Platform Balance 1.19970434 DASH** and the top-up sheet showed **"Available: 1.19970434 DASH."** As a downstream effect, **DPNS registration also failed** (`Insufficient identity … balance 35781660 required 42539440`) because the identity could never be topped up — leaving the user stuck in a loop.

With `creditsPerDash = 100_000_000_000`, the displayed ~120 billion credits vs. the protocol's real `< 1,500,000` credits is an **~80,000× gap** between the wallet's local platform-address balance and on-chain reality — a phantom balance.

## What was done?

**Root cause:** `IdentityWallet::top_up_from_addresses` discarded the proof-attested `AddressInfos` the SDK returns — the new on-chain balance + bumped nonce of every platform address the top-up spent from. So the local platform-address balances stayed frozen at their pre-top-up values: the wallet kept showing the stale "Platform Balance", and the next top-up's input selection (driven by those stale balances) over-selected the now-drained addresses, which Drive rejected.

**The fix:**
- `IdentityWallet::top_up_from_addresses` now **returns** the proof-attested `(AddressInfos, Credits)` instead of discarding the infos.
- The caller routes those infos into **`PlatformAddressWallet::apply_top_up_reconciliation`**, which resolves each spent address through the address **provider's persisted `index ↔ address` bijection** (`per_wallet_state`). This covers addresses **restored from disk that are no longer in the live derived pool** — the exact cold/restored-wallet case the reported bug is about. It then applies the proof-attested post-spend balance in memory and **persists a `PlatformAddressChangeSet`**, so the displayed balance and the next input selection both reflect on-chain reality.
- Lock discipline: entries are resolved under the provider read lock (dropped), applied in memory under the wallet write lock (dropped), and **persisted with no locks held** — matching the transfer path.
- The shared entry-resolution is extracted into the unit-testable `build_top_up_balance_entries`.

This mirrors the sibling `fund_from_asset_lock` path, which already reconciles platform-address balances after spending (its comment documents this exact failure mode). Rust-only change in `rs-platform-wallet` + the top-up FFI.

## How Has This Been Tested?

- **`build_top_up_entries_resolves_restored_address_outside_live_pool`** — unit test pinning that a spent address present only in the persisted bijection (restored, not in the live pool) is resolved to the correct derivation index with the proof's post-spend balance/nonce. This is the load-bearing logic and the case the original bug hinges on.
- **`apply_top_up_reconciliation_persists_decremented_balance`** — recording-persister integration test pinning that the reconciliation **actually persists** a `PlatformAddressChangeSet` carrying the decremented entry (not only that the pure builder maps it). Verified red→green: with the persist call removed (the original discard bug) it fails; restored, it passes.
- **Full `rs-platform-wallet` test suite passes (204); `cargo fmt` clean; CI `Tests` green.**

**Not yet covered (network-gated):** the FFI boundary actually routing the returned `AddressInfos` into `apply_top_up_reconciliation`, and Drive accepting a *follow-up* top-up, require a live testnet e2e (top-up → second top-up succeeds). Best confirmed by the affected user re-testing on their already-funded device.

## Breaking Changes

None. `IdentityWallet::top_up_from_addresses` now returns `(AddressInfos, Credits)` instead of `Credits`; all in-tree callers are updated.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Top-ups now return detailed address information alongside the updated balance, enabling proper reconciliation after spending.
  * Platform wallet balances are now automatically reconciled after a top-up, including addresses restored from storage.

* **Bug Fixes**
  * Fixed cases where spent platform addresses could keep showing stale balances or next-address selection.
  * Improved persistence so balance updates are reflected consistently after reconciliation.


thepastaclaw:
✅ Review complete (commit e9aff46b)

packages/rs-platform-wallet/src/wallet/identity/network/top_up_from_addresses.rs
```diff
@@ -87,8 +90,167 @@ impl IdentityWallet {
...
+                            .iter()
+                            .find_map(|(&idx, ainfo)| {
+                                PlatformP2PKHAddress::from_address(&ainfo.address)
+                                    .ok()
+                                    .filter(|found| *found == p2pkh)
+                                    .map(|_| idx)
+                            })
+                    {
+                        owned.insert(p2pkh, index);
+                    }
```
thepastaclaw: **🔴 Blocking: Reconciliation ignores restored balances outside the live address pool**

The ownership lookup only scans `account.addresses.addresses`, so it only reconciles spent addresses that are currently present in the live derived-address pool. Restored platform-address state is broader than that: `initialize_from_persisted` rehydrates `account.address_balances` from persisted `found()` entries, and `PlatformPaymentAddressProvider::from_persisted` keeps the persisted `(address_index, address)` map even for addresses that are not currently in `managed.addresses.addresses`. If a top-up spends one of those restored cached rows, the SDK returns the proof-backed post-spend balance, but this loop leaves it out of `owned`; the in-memory balance and persisted row remain stale, which preserves the phantom balance this PR is intended to clear. Resolve spent wallet inputs from the persisted/indexed platform-address state as well as the live pool, or otherwise carry the stored derivation index for `address_balances` entries into this reconciliation.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/identity/network/top_up_from_addresses.rs
```diff
@@ -87,8 +90,167 @@ impl IdentityWallet {
...
+                    crate::wallet::platform_addresses::build_platform_address_persistence_entries(
+                        self.wallet_id,
+                        account_index,
+                        &owned,
+                        address_infos.iter().map(|(a, i)| (a, i.as_ref())),
+                    ),
+                );
+            }
+            if !addr_cs.addresses.is_empty() {
+                if let Err(e) = self.persister.store(addr_cs.into()) {
```
thepastaclaw: **🟡 Suggestion: Persist address reconciliation after releasing the wallet write lock**

This persistence call runs while the `wallet_manager` write guard acquired earlier in the block is still live. The FFI persistence backend executes its callback sequence inline, and the transfer path already avoids this pattern by building the changeset under the lock, dropping the guard, and then calling `store`. Keeping the write lock during persistence can unnecessarily block other wallet operations and creates avoidable re-entry or lock-order risk for persistence implementations. Build the identity/address changesets while the guard is held, drop the guard, then call the persister.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/identity/network/top_up_from_addresses.rs
```diff
@@ -87,8 +90,167 @@ impl IdentityWallet {
...
+        };
+        let address_infos: BTreeMap<PlatformAddress, Option<AddressInfo>> =
+            [(spent_addr, Some(post_spend))].into_iter().collect();
+
+        let entries = crate::wallet::platform_addresses::build_platform_address_persistence_entries(
+            wallet_id,
+            account_index,
+            &owned,
+            address_infos.iter().map(|(a, i)| (a, i.as_ref())),
+        );
```
thepastaclaw: **🟡 Suggestion: Regression test does not cover the top-up reconciliation path**

The regression was in `IdentityWallet::top_up_from_addresses` dropping or failing to persist the SDK-returned `address_infos`, but this test calls `build_platform_address_persistence_entries` directly. That proves the helper maps a provided owned address correctly, but it would still pass if the wallet method stopped invoking the helper, stopped storing the changeset, or failed the restored-address ownership lookup above. Add coverage around the wallet reconciliation path itself, or extract the reconciliation block into a unit-testable function that takes account state and returns the applied changeset.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):
## Code Review

The PR correctly starts using the SDK-returned proof-backed address info after identity top-up, but the new reconciliation still misses a restored-wallet state that this fix needs to cover. The main correctness issue is that ownership and derivation indexes are resolved only from the live derived address pool, while persisted platform-address balances can be restored separately.

<details>

```

In `packages/rs-platform-wallet/src/wallet/identity/network/top_up_from_addresses.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/wallet/identity/network/top_up_from_addresses.rs:124-137: Reconciliation ignores restored balances outside the live address pool
  The ownership lookup only scans `account.addresses.addresses`, so it only reconciles spent addresses that are currently present in the live derived-address pool. Restored platform-address state is broader than that: `initialize_from_persisted` rehydrates `account.address_balances` from persisted `found()` entries, and `PlatformPaymentAddressProvider::from_persisted` keeps the persisted `(address_index, address)` map even for addresses that are not currently in `managed.addresses.addresses`. If a top-up spends one of those restored cached rows, the SDK returns the proof-backed post-spend balance, but this loop leaves it out of `owned`; the in-memory balance and persisted row remain stale, which preserves the phantom balance this PR is intended to clear. Resolve spent wallet inputs from the persisted/indexed platform-address state as well as the live pool, or otherwise carry the stored derivation index for `address_balances` entries into this reconciliation.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/top_up_from_addresses.rs:163-164: Persist address reconciliation after releasing the wallet write lock
  This persistence call runs while the `wallet_manager` write guard acquired earlier in the block is still live. The FFI persistence backend executes its callback sequence inline, and the transfer path already avoids this pattern by building the changeset under the lock, dropping the guard, and then calling `store`. Keeping the write lock during persistence can unnecessarily block other wallet operations and creates avoidable re-entry or lock-order risk for persistence implementations. Build the identity/address changesets while the guard is held, drop the guard, then call the persister.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/top_up_from_addresses.rs:232-237: Regression test does not cover the top-up reconciliation path
  The regression was in `IdentityWallet::top_up_from_addresses` dropping or failing to persist the SDK-returned `address_infos`, but this test calls `build_platform_address_persistence_entries` directly. That proves the helper maps a provided owned address correctly, but it would still pass if the wallet method stopped invoking the helper, stopped storing the changeset, or failed the restored-address ownership lookup above. Add coverage around the wallet reconciliation path itself, or extract the reconciliation block into a unit-testable function that takes account state and returns the applied changeset.
```
</details>

packages/rs-platform-wallet-ffi/src/identity_top_up.rs
```diff
@@ -118,11 +118,21 @@ pub unsafe extern "C" fn platform_wallet_top_up_from_addresses_with_signer(
...
-                .await
+                .await?;
+            // Reconcile the spent platform-address balances from the proof so
+            // the wallet's displayed balance and next input selection reflect
+            // the spend. Resolves spent addresses via the address provider, so
+            // it covers addresses restored from disk that are no longer in a
+            // live derived pool.
+            platform_wallet
+                .apply_top_up_reconciliation(&address_infos)
+                .await;
```
thepastaclaw, line 134: **🟡 Suggestion: Top-up reconciliation integration remains untested**

The current regression coverage exercises `build_top_up_balance_entries`, but it still does not pin the production contract that fixes the reported stale-balance bug: `IdentityWallet::top_up_from_addresses` returns proof-backed `AddressInfos`, the caller routes those infos into `PlatformAddressWallet::apply_top_up_reconciliation`, and that path applies and persists the resulting `PlatformAddressChangeSet`. A future change could remove or skip lines 132-134, or regress `apply_top_up_reconciliation` so it no longer persists entries, while the helper-level test continues to pass. Add focused coverage for the FFI/top-up boundary or for `apply_top_up_reconciliation` with a recording persister so the actual reconciliation contract is tested, not only the pure entry builder.

<sub>source: ['codex']</sub>

thepastaclaw (approved):
## Code Review

Prior reconciliation against 1bd68c3fd4734d8131e9bbaee0ba2ca14bf5cca3: #1 is FIXED by resolving spent addresses through the provider's persisted index/address bijection, #2 is FIXED because address reconciliation persistence now runs after provider and wallet locks are dropped, and #3 is STILL VALID and carried forward. I found no new latest-delta findings beyond the carried-forward test coverage gap. CodeRabbit provided no actionable inline findings, so there are no reactions.

🟡 1 suggestion(s)

<details>

```

In `packages/rs-platform-wallet-ffi/src/identity_top_up.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/identity_top_up.rs:124-134: Top-up reconciliation integration remains untested
  The current regression coverage exercises `build_top_up_balance_entries`, but it still does not pin the production contract that fixes the reported stale-balance bug: `IdentityWallet::top_up_from_addresses` returns proof-backed `AddressInfos`, the caller routes those infos into `PlatformAddressWallet::apply_top_up_reconciliation`, and that path applies and persists the resulting `PlatformAddressChangeSet`. A future change could remove or skip lines 132-134, or regress `apply_top_up_reconciliation` so it no longer persists entries, while the helper-level test continues to pass. Add focused coverage for the FFI/top-up boundary or for `apply_top_up_reconciliation` with a recording persister so the actual reconciliation contract is tested, not only the pure entry builder.
```
</details>

packages/rs-platform-wallet/src/wallet/identity/network/top_up_from_addresses.rs
thepastaclaw: Resolved in `1bd68c3` — _Persist address reconciliation after releasing the wallet write lock_ no longer present.

packages/rs-platform-wallet/src/wallet/identity/network/top_up_from_addresses.rs
thepastaclaw: Resolved in `1bd68c3` — _Reconciliation ignores restored balances outside the live address pool_ no longer present.

packages/rs-platform-wallet/src/wallet/identity/network/top_up_from_addresses.rs
thepastaclaw: Resolved in `1bd68c3` — _Regression test does not cover the top-up reconciliation path_ no longer present.

packages/rs-platform-wallet/src/wallet/identity/network/top_up_from_addresses.rs
shumkov: Resolved in 1bd68c3fd4. The reconciliation now runs in `PlatformAddressWallet::apply_top_up_reconciliation`, which resolves each spent address through the provider's `per_wallet_state` — the full persisted `index ↔ address` bijection, **including addresses restored from disk that are no longer in a live derived pool**. So a top-up that spends a restored cached row is reconciled with its correct derivation index, balance, and nonce. Pinned by the unit test `build_top_up_entries_resolves_restored_address_outside_live_pool`.

packages/rs-platform-wallet/src/wallet/identity/network/top_up_from_addresses.rs
shumkov: Resolved in 1bd68c3fd4. `apply_top_up_reconciliation` resolves the entries under the provider **read** lock (then drops it), applies the in-memory balances under the wallet **write** lock (then drops it), and **persists with no locks held** — matching the transfer path's pattern.

packages/rs-platform-wallet/src/wallet/identity/network/top_up_from_addresses.rs
shumkov: Resolved in 1bd68c3fd4 via the second option suggested here: the reconciliation's address-resolution is extracted into the unit-testable `build_top_up_balance_entries(wallet_id, per_account, address_infos)`, covered by `build_top_up_entries_resolves_restored_address_outside_live_pool` — which also pins the restored-address path from the blocking finding above.

packages/rs-platform-wallet-ffi/src/identity_top_up.rs
shumkov: The load-bearing logic — resolving spent addresses (including restored ones outside the live pool) to decremented `PlatformAddressBalanceEntry`s — is now unit-tested via the extracted `build_top_up_balance_entries`. `apply_top_up_reconciliation` is a thin wrapper that applies those entries in memory and persists them (no locks held). A recording-persister test of that wrapper would need an xpub-backed provider + a platform-account-populated wallet-manager fixture (no helper exists for that today); given the helper coverage and that the FFI wiring + Drive acceptance are covered by the manual testnet e2e noted in the PR description, resolving — happy to add the wrapper-level integration test as a follow-up if preferred.

shumkov: Now actually addressed in c0d5ac78bf rather than argued away: added `apply_top_up_reconciliation_persists_decremented_balance` — a recording-persister integration test that injects a provider whose persisted bijection knows the spent address, runs `apply_top_up_reconciliation`, and asserts the resulting `PlatformAddressChangeSet` (proof's post-spend balance + bumped nonce) was **actually persisted**. Verified red→green: removing the persist call in `apply_top_up_reconciliation` (the original discard bug) makes it fail ("a persisted platform-address entry for the spent address"); restored, it passes. 204 lib tests green.

thepastaclaw (approved):
## Code Review

Carried-forward prior findings: none. The prior test-coverage finding is fixed by the new `apply_top_up_reconciliation_persists_decremented_balance` test, which covers the recording-persister contract that the previous review explicitly accepted as an equivalent fix path. New latest-delta findings: none; the remaining FFI/live top-up coverage gap is useful follow-up work but is not an in-scope PR finding after the reconciliation/persistence contract is pinned.

QuantumExplorer (approved):

packages/rs-platform-wallet/src/wallet/platform_addresses/provider.rs
```diff
@@ -811,6 +826,55 @@ impl AddressProvider for PlatformPaymentAddressProvider {
...
+        let funds = match maybe_info {
+            Some(ai) => AddressFunds {
+                balance: ai.balance,
+                nonce: ai.nonce,
+            },
+            None => AddressFunds {
+                balance: 0,
+                nonce: 0,
+            },
+        };
```
thepastaclaw, line 861: **🟡 Suggestion: `None` AddressInfo silently fabricates balance=0/nonce=0 instead of skipping+logging like the sibling reconciliation path**

`build_top_up_balance_entries` maps a `None` `AddressInfo` to `AddressFunds { balance: 0, nonce: 0 }` and unconditionally emits/persists an entry. The sibling `fund_from_asset_lock.rs::write_address_balances_changeset` (line 385) treats a `None` on the same `AddressInfos` proof shape as a protocol-contract violation, logs `tracing::error!`, and skips — precisely to avoid recording a fabricated 'credited 0' row.

For a just-spent input address, `None` should be effectively unreachable (Drive's `set_balance_to_address_operations_v0` never deletes the entry, and every spent address had a real prior nonce/balance verified by `ensure_address_balance`). If it ever fires, it indicates a proof/query anomaly, not a legitimate zero-state — and this code silently:

1. Overwrites the in-memory balance to 0 via `set_address_credit_balance(addr, 0, None)`.
2. Persists `PlatformAddressBalanceEntry { funds: { balance: 0, nonce: 0 } }`, which external persisters (e.g. evo-tool's `platform_address_balances` table, per `wallet/apply.rs:272-276`) treat as authoritative for both balance and nonce — regressing replay-protection state that other code (`sync.rs`'s `had_funds = f.nonce != 0` gate) actually reads.

This reintroduces a narrower version of the exact 'local diverges from on-chain silently' bug class this PR exists to fix. The PR description explicitly says it mirrors `fund_from_asset_lock`, so aligning the two paths is the right call. Not blocking because current code paths make the branch nearly unreachable, but cheap to close and it also has no test coverage today.

```suggestion
        let funds = match maybe_info {
            Some(ai) => AddressFunds {
                balance: ai.balance,
                nonce: ai.nonce,
            },
            None => {
                tracing::error!(
                    address = %p2pkh,
                    "Platform proof returned None AddressInfo for a spent address; skipping reconciliation entry to avoid persisting balance=0/nonce=0 over real state"
                );
                continue;
            }
        };
```

<sub>source: ['claude']</sub>

packages/rs-platform-wallet/src/wallet/platform_addresses/wallet.rs
```diff
@@ -157,6 +159,100 @@ impl PlatformAddressWallet {
...
+                    {
+                        account.set_address_credit_balance(
+                            entry.address,
+                            entry.funds.balance,
+                            None,
+                        );
+                    }
+                }
+            }
+        }
```
thepastaclaw, line 240: **🟡 Suggestion: Reconciliation updates ManagedPlatformAccount but leaves the live provider's `found` cache stale**

`apply_top_up_reconciliation` writes the proof-attested post-spend balance into `ManagedPlatformAccount.address_balances` via `set_address_credit_balance` and persists a changeset — this correctly fixes the two consumers the PR targets (displayed balance and `auto_select_inputs`, both of which read `ManagedPlatformAccount.address_balances`).

However, the parallel copy of this data in `PlatformPaymentAddressProvider.per_wallet[wallet_id][account_index].found` (a `BTreeMap<PlatformP2PKHAddress, AddressFunds>`) is never updated. That map is what `current_balances()` hands back to the SDK to seed the next incremental BLAST sync pass. The normal sync path keeps both maps in lockstep (`on_address_found` in provider.rs updates the scratch `found` map AND calls `account.set_address_credit_balance` in the same callback), but this new top-up path only does the second half.

Until the next full sync round rebuilds the provider's `found` entry (or the process restarts and reloads from the just-persisted changeset), the live in-memory provider keeps seeding the stale pre-spend balance for that address within the current session — the same two-copies-of-truth inconsistency the PR is trying to eliminate, one layer up.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet/src/wallet/identity/network/top_up_from_addresses.rs
```diff
@@ -26,6 +28,15 @@ impl IdentityWallet {
...
     /// looked up automatically.
     ///
+    /// Returns the proof-attested post-spend `AddressInfos` alongside the new
+    /// identity balance. The caller MUST reconcile the spent platform-address
+    /// balances from the `AddressInfos` via
+    /// [`PlatformAddressWallet::apply_top_up_reconciliation`] — this method
+    /// only owns the identity-side balance update, because resolving a spent
+    /// address back to its derivation index needs the address provider, which
+    /// lives on the platform-address wallet (and covers addresses restored
+    /// from disk that are no longer in a live derived pool).
```
thepastaclaw, line 38: **🟡 Suggestion: Two-call reconciliation contract enforced only by a doc comment, not the type system**

`top_up_from_addresses` now returns `(AddressInfos, Credits)` with a doc comment stating the caller "MUST" call `PlatformAddressWallet::apply_top_up_reconciliation`. This is exactly the same implicit side-channel shape whose neglect caused the bug this PR fixes (the SDK-level `AddressInfos` were previously discarded as `_address_infos` at the call site). Nothing prevents a future caller — or a mechanical refactor — from pattern-matching `let (_, new_balance) = ...` and compiling cleanly, silently reintroducing the phantom-balance regression.

Today there is exactly one in-tree caller (`rs-platform-wallet-ffi/src/identity_top_up.rs`) which does call reconciliation correctly, so the risk is low right now. But the more robust shape would be for `IdentityWallet::top_up_from_addresses` to take the `PlatformAddressWallet` (or the provider handle) and perform reconciliation internally so the two operations can't be pulled apart. Not blocking, but worth folding into a follow-up refactor.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet/src/wallet/platform_addresses/provider.rs
```diff
@@ -811,6 +826,55 @@ impl AddressProvider for PlatformPaymentAddressProvider {
...
+pub(crate) fn build_top_up_balance_entries(
+    wallet_id: WalletId,
+    per_account_addresses: &[(u32, &BiBTreeMap<AddressIndex, PlatformP2PKHAddress>)],
+    address_infos: &AddressInfos,
+) -> Vec<PlatformAddressBalanceEntry> {
+    let mut entries = Vec::new();
+    for (addr, maybe_info) in address_infos.iter() {
+        let PlatformAddress::P2pkh(hash) = addr else {
+            continue;
+        };
```
thepastaclaw, line 850: **💬 Nitpick: Non-P2PKH proof entries silently skipped without a diagnostic**

`let PlatformAddress::P2pkh(hash) = addr else { continue; }` drops any non-P2PKH variant with no log. Every other unresolved-address branch this PR introduces (missing provider state, empty entries with non-empty `address_infos`, persist failure) emits a `tracing::warn!/error!` precisely to surface phantom-balance conditions. If `PlatformAddress` ever gains a non-P2PKH variant, a silent drop here would hide a real reconciliation gap. A one-line `tracing::warn!` on the else-branch keeps the fix's diagnosability posture consistent. Not blocking — today only P2PKH platform addresses exist.

<sub>source: ['claude']</sub>

thepastaclaw (approved):
## Code Review

PR #3969 correctly routes SDK proof-attested AddressInfos into a new PlatformAddressWallet::apply_top_up_reconciliation that resolves spent addresses through the persisted index bijection (covering restored-from-disk addresses), updates in-memory ManagedPlatformAccount balances, and persists a PlatformAddressChangeSet. Lock discipline mirrors the sibling fund_from_asset_lock path, and there are load-bearing regression tests. No blockers; four suggestion/nitpick-level consistency and diagnosability gaps relative to the sibling pattern this PR claims to mirror. Prior PastaClaw review at c0d5ac78 had no findings — nothing carried forward.

Source: reviewers: `opus` (general), `claude-sonnet-5` (general), `gpt-5.5[high]` (general, failed), `opus` (security-auditor), `claude-sonnet-5` (security-auditor), `gpt-5.5[high]` (security-auditor, failed), `opus` (rust-quality), `claude-sonnet-5` (rust-quality), `gpt-5.5[high]` (rust-quality, failed), `opus` (ffi-engineer), `claude-sonnet-5` (ffi-engineer), `gpt-5.5[high]` (ffi-engineer, failed); verifier: `opus`; specialists: `security-auditor`, `rust-quality`, `ffi-engineer`

<details>

```

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/platform_addresses/provider.rs:852-861: `None` AddressInfo silently fabricates balance=0/nonce=0 instead of skipping+logging like the sibling reconciliation path
  `build_top_up_balance_entries` maps a `None` `AddressInfo` to `AddressFunds { balance: 0, nonce: 0 }` and unconditionally emits/persists an entry. The sibling `fund_from_asset_lock.rs::write_address_balances_changeset` (line 385) treats a `None` on the same `AddressInfos` proof shape as a protocol-contract violation, logs `tracing::error!`, and skips — precisely to avoid recording a fabricated 'credited 0' row.

  For a just-spent input address, `None` should be effectively unreachable (Drive's `set_balance_to_address_operations_v0` never deletes the entry, and every spent address had a real prior nonce/balance verified by `ensure_address_balance`). If it ever fires, it indicates a proof/query anomaly, not a legitimate zero-state — and this code silently:

  1. Overwrites the in-memory balance to 0 via `set_address_credit_balance(addr, 0, None)`.
  2. Persists `PlatformAddressBalanceEntry { funds: { balance: 0, nonce: 0 } }`, which external persisters (e.g. evo-tool's `platform_address_balances` table, per `wallet/apply.rs:272-276`) treat as authoritative for both balance and nonce — regressing replay-protection state that other code (`sync.rs`'s `had_funds = f.nonce != 0` gate) actually reads.

  This reintroduces a narrower version of the exact 'local diverges from on-chain silently' bug class this PR exists to fix. The PR description explicitly says it mirrors `fund_from_asset_lock`, so aligning the two paths is the right call. Not blocking because current code paths make the branch nearly unreachable, but cheap to close and it also has no test coverage today.

In `packages/rs-platform-wallet/src/wallet/platform_addresses/wallet.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/platform_addresses/wallet.rs:222-240: Reconciliation updates ManagedPlatformAccount but leaves the live provider's `found` cache stale
  `apply_top_up_reconciliation` writes the proof-attested post-spend balance into `ManagedPlatformAccount.address_balances` via `set_address_credit_balance` and persists a changeset — this correctly fixes the two consumers the PR targets (displayed balance and `auto_select_inputs`, both of which read `ManagedPlatformAccount.address_balances`).

  However, the parallel copy of this data in `PlatformPaymentAddressProvider.per_wallet[wallet_id][account_index].found` (a `BTreeMap<PlatformP2PKHAddress, AddressFunds>`) is never updated. That map is what `current_balances()` hands back to the SDK to seed the next incremental BLAST sync pass. The normal sync path keeps both maps in lockstep (`on_address_found` in provider.rs updates the scratch `found` map AND calls `account.set_address_credit_balance` in the same callback), but this new top-up path only does the second half.

  Until the next full sync round rebuilds the provider's `found` entry (or the process restarts and reloads from the just-persisted changeset), the live in-memory provider keeps seeding the stale pre-spend balance for that address within the current session — the same two-copies-of-truth inconsistency the PR is trying to eliminate, one layer up.

In `packages/rs-platform-wallet/src/wallet/identity/network/top_up_from_addresses.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/top_up_from_addresses.rs:31-38: Two-call reconciliation contract enforced only by a doc comment, not the type system
  `top_up_from_addresses` now returns `(AddressInfos, Credits)` with a doc comment stating the caller "MUST" call `PlatformAddressWallet::apply_top_up_reconciliation`. This is exactly the same implicit side-channel shape whose neglect caused the bug this PR fixes (the SDK-level `AddressInfos` were previously discarded as `_address_infos` at the call site). Nothing prevents a future caller — or a mechanical refactor — from pattern-matching `let (_, new_balance) = ...` and compiling cleanly, silently reintroducing the phantom-balance regression.

  Today there is exactly one in-tree caller (`rs-platform-wallet-ffi/src/identity_top_up.rs`) which does call reconciliation correctly, so the risk is low right now. But the more robust shape would be for `IdentityWallet::top_up_from_addresses` to take the `PlatformAddressWallet` (or the provider handle) and perform reconciliation internally so the two operations can't be pulled apart. Not blocking, but worth folding into a follow-up refactor.
```
</details>

PR #3968, opened 2026-06-29
feat(platform-wallet-storage): embeddable SQLite persistence backend with seedless rehydration

Claudius-Maginificent:
**TL;DR:** Adds a durable, embeddable SQLite storage backend for Dash Platform wallet state, so identities, contacts, keys, and balances survive an app restart instead of being lost or re-derived from scratch.

> **Scope note (2026-09-02):** this PR has been trimmed to the `rs-platform-wallet-storage` crate plus the minimal `platform-wallet` linkage changes it needs to build (feature-unification `shielded` fix, `rebuild_provider_key_account`, `insert_platform_node_pool_entry`, the `delete_wallet` trait default) and the config/CI it needs to pass. Everything else previously bundled here — typed persister errors + transient retry, FFI persister error codes, the FFI provider-rebuild dedup, the contact-account filter-scan fix, and assorted cleanup — moved to follow-up PRs: #4586 (typed persister errors + retry), #4587 (FFI persister codes + contact-account fix + cleanup, stacked on #4586, its last commit depends on this PR merging first), and #4585 (asset-lock proof size gate). See those PRs for that work; the "Outside the storage crate" narrative below is retained as history but no longer describes this PR's diff.

## User story
As a **Dash Platform wallet developer**, I want my wallet's Platform data (identities, contacts, keys, balances, core sync progress) to survive an app/node restart, so that users don't lose data or sit through a full re-sync every time the app starts.

### Base flow
A Dash Platform wallet app registers identities and contacts, tracks balances and asset locks, and derives Platform payment addresses as the user interacts with it.

### Actual behavior
`platform-wallet` defines the persistence trait (`PlatformWalletPersistence`) and the manager-side `load_from_persistor()` entry point, but ships no production storage backend. Restart the app and its Platform identities and contacts are gone until re-derived, the address-reuse guard resets, and core sync restarts from scratch — no on-disk durability, no backup/restore, no schema-migration path.

### Expected behavior
That state is durably persisted to a local SQLite database (one `.db` file can hold many wallets), with online backup/restore and automatic schema migration. Restarting the app restores everything **seedlessly** and signing works immediately post-load — and no private-key material is ever written to the database (signing material stays in the OS keyring or an encrypted vault).

## Detailed discussion

[...]

PR #3980, opened 2026-07-02
fix(platform-wallet): preserve persisted wallet state through seedless rehydration

QuantumExplorer:

Suggested fixes for the findings of a code review of #3692 (see the review comment there for the full findings list with verification details). The confirmed defects all share one root cause: the keyless projection in the FFI persister flattens state the store already holds exactly — per-account UTXO attribution, pool contents with per-index `used` flags, tx records — and the manager-side replay cannot reconstruct what the flattening discarded. Concretely:

1. **Address-reuse wedge** (`rehydrate.rs`): `mark_used` ran once before `maintain_gap_limit`, so a previously-used address past the discovery horizon but within gap-refill reach (e.g. used idx 45 with in-window used idx 20, gap 30) was derived *after* the mark pass and came back `used = false` — pool-visible as a fresh receive address, with the stale flag then persisted back over the store's `is_used = true`.
2. **Attribution loss**: every restored UTXO was routed to the first funds-bearing account, zeroing non-first-account (CoinJoin/BIP32/DashPay) balances and leaving their addresses unresolvable from the wrong xpub — although the FFI decode loop had already routed each UTXO to its exact account.
3. **Watch-set shrinkage**: only *used* addresses crossed the boundary, so derived-but-unused (handed-out) addresses and everything past the rebuilt depth fell out of the SPV watch set until a full rescan.
4. **Swift discarded the outcome**: a Rust-skipped corrupt wallet still went through the `get_wallet` loop, landing a spurious NotFound in the `@Published lastError`, and the user was never told a wallet was skipped.

## What was done?

- **Carry the snapshot instead of projecting it**: `ClientWalletStartState` gains `core_wallet_info: Option<Box<ManagedWalletInfo>>` — still keyless by type (no `Wallet`, no seed; `ManagedWalletInfo` holds balances/pools/UTXOs only). The FFI persister hands its fully-restored `wallet_info` across; the manager validates `wallet_id`/`network` against the row (mismatch ⇒ per-row corrupt skip) and consumes it directly. This fixes attribution, watch-set coverage, and used-flag fidelity in one move, restores the asset-lock funding-tx records to the in-memory map, deletes the projection block from the FFI crate (per `packages/swift-sdk/CLAUDE.md`, stitching belongs in the library), and removes the second eager gap-window derivation from the launch path. `core_state`/`used_core_addresses` remain as the documented fallback for persisters that can't build a snapshot (SQLite until #3968).
- **Mark↔refill fixpoint** in `extend_pools_for_restored_addresses` (still used by the fallback path): re-mark after every gap refill until nothing new resolves, so wedge-zone used addresses come back `used`.
- **Idempotency check hoisted** to the top of the load loop (cheap read-lock `get_wallet`), so a repeat `load_from_persistor` skips per-wallet reconstruction instead of discarding it at the `WalletExists` arm (which stays as a race backstop).
- **Swift consumes `LoadOutcomeFFI`**: skipped wallets are logged with a decoded reason, surfaced via a new `@Published lastLoadSkippedWallets: [SkippedWalletOnLoad]`, excluded from the `get_wallet` loop (no more `lastError` pollution), and the outcome is freed via `platform_wallet_load_outcome_free` (deferred ahead of the throwing `check()`).
- **Architecture record**: `packages/rs-platform-wallet/README.md` gains a normative "Persistence architecture" section codifying the agreed model (single writer, store-as-read-model, versioned schema contract, verbatim `load()`, hard persist errors, seedless load) — see the execution-plan comment on #3692.
- **Cleanups**: `RehydrateRowError` (variant-for-variant mirror of `CorruptKind` + identity `From`) deleted — `build_watch_only_wallet` returns `CorruptKind` directly; the FFI skipped-array allocation reuses the crate's `vec_to_ptr` helper.

## How Has This Been Tested?

- `cargo test -p platform-wallet` — 217 lib + 8 + 7 integration, all passing. New tests:
  - `rehydration_wedge_zone_used_address_marked_after_refill` (fails without the fixpoint fix),
  - `rt_snapshot_preserves_attribution_and_pools` (CoinJoin-account UTXO stays attributed; derived-but-unused deep address survives reload),
  - `rt_snapshot_wallet_id_mismatch_is_skipped` (snapshot/row mismatch ⇒ per-row skip, not a load).
- `cargo test -p platform-wallet-ffi --features shielded` — 111 + 26 + 5, all passing.
- `cargo clippy -p platform-wallet -p platform-wallet-ffi --all-targets -- -D warnings` clean; `cargo fmt --all --check` clean; `cargo check --all-targets` clean on the downstream `platform-wallet-storage` and `rs-unified-sdk-ffi` crates.
- Swift file changed in lockstep with the FFI signature; like the base PR, it still needs an iOS build check (the xcframework isn't built in this environment).

## Breaking Changes

None beyond the base PR's own unreleased shapes (`ClientWalletStartState` gains a field; `build_watch_only_wallet`'s error type changes — both introduced by #3692 and unreleased).

## Checklist:

### DIFF
--- a/packages/rs-platform-wallet/src/changeset/client_wallet_start_state.rs
+++ b/packages/rs-platform-wallet/src/changeset/client_wallet_start_state.rs
@@ -2,7 +2,8 @@
 //!
 //! **Keyless by type.** This carries everything needed to *reconstruct*
 //! a watch-only wallet — network, birth height, the account manifest,
-//! the rebuilt core-state projection, identities, filtered asset locks —
+//! the managed-state snapshot (or its keyless projection), identities,
+//! filtered asset locks —
 //! but **no** [`Wallet`](key_wallet::Wallet) and no seed. The persister
 //! can never mint a `Wallet`; the manager rebuilds a watch-only one via
 //! [`Wallet::new_watch_only`](key_wallet::wallet::Wallet::new_watch_only)
@@ -18,6 +19,7 @@ use crate::changeset::{
 };
 use crate::wallet::asset_lock::tracked::TrackedAssetLock;
 use dashcore::OutPoint;
+use key_wallet::wallet::managed_wallet_info::ManagedWalletInfo;
 use key_wallet::{Address, Network};
 
 /// Keyless per-wallet slice of the startup snapshot.
@@ -36,6 +38,23 @@ pub struct ClientWalletStartState {
     /// Keyless account manifest — the account-set oracle for building the
     /// watch-only wallet (one watch-only account per entry's xpub).
     pub account_manifest: Vec<AccountRegistrationEntry>,
+    /// Full keyless managed-wallet snapshot for persisters that can
+    /// reconstruct one — pools with exact derivation indices and `used`
+    /// flags, per-account UTXO and tx-record attribution, IS-lock set,
+    /// and sync metadata. [`ManagedWalletInfo`] carries **no key
+    /// material** (see its docs: balances, account metadata, UTXO set),
+    /// so the SECRETS.md boundary holds: still no `Wallet`, no seed.
+    ///
+    /// When `Some`, the manager consumes it directly (after validating
+    /// its `wallet_id`/`network` against the row) instead of minting a
+    /// `ManagedWalletInfo::from_wallet` skeleton and replaying the
+    /// projection below — preserving per-account attribution, the full
+    /// SPV watch set, and pool used-state verbatim, without re-deriving
+    /// anything. The FFI/iOS persister populates this. When `None` (the
+    /// native/SQLite persister until dashpay/platform#3968), the manager
+    /// falls back to the skeleton + [`core_state`](Self::core_state) /
+    /// [`used_core_addresses`](Self::used_core_addresses) replay.
+    pub core_wallet_info: Option<Box<ManagedWalletInfo>>,
     /// Keyless projection of the persisted core rows (UTXOs, tx
     /// records, IS-locks, sync watermarks, `last_applied_chain_lock`).
     /// The manager applies this onto a fresh
@@ -43,6 +62,9 @@ pub struct ClientWalletStartState {
     /// watch-only wallet. Populated by the persister's
     /// [`PlatformWalletPersistence::load`](crate::changeset::PlatformWalletPersistence::load)
     /// implementation reading the persisted core rows.
+    ///
+    /// Ignored when [`core_wallet_info`](Self::core_wallet_info) is
+    /// `Some` — the full snapshot supersedes the projection.
     pub core_state: CoreChangeSet,
     /// Lean snapshot of this wallet's
     /// [`IdentityManager`](crate::wallet::identity::IdentityManager).
--- a/packages/rs-platform-wallet/src/manager/load.rs
+++ b/packages/rs-platform-wallet/src/manager/load.rs
@@ -7,7 +7,7 @@ use key_wallet::wallet::managed_wallet_info::ManagedWalletInfo;
 
 use crate::changeset::{ClientStartState, ClientWalletStartState, PlatformWalletPersistence};
 use crate::error::PlatformWalletError;
-use crate::manager::load_outcome::{LoadOutcome, SkipReason};
+use crate::manager::load_outcome::{CorruptKind, LoadOutcome, SkipReason};
 use crate::wallet::core::WalletBalance;
 use crate::wallet::identity::IdentityManager;
 use crate::wallet::platform_wallet::{PlatformWalletInfo, WalletId};
@@ -21,8 +21,20 @@ impl<P: PlatformWalletPersistence + 'static> PlatformWalletManager<P> {
     /// keyless reconstruction snapshot; each wallet is rebuilt via
     /// [`Wallet::new_watch_only`](key_wallet::wallet::Wallet::new_watch_only)
     /// from its [`AccountRegistrationEntry`](crate::changeset::AccountRegistrationEntry)
-    /// manifest, the keyless core-state projection is applied, and the
-    /// result is registered into the manager.
+    /// manifest, the managed core state is restored, and the result is
+    /// registered into the manager.
+    ///
+    /// Core state comes in one of two shapes, per wallet:
+    /// - a full keyless snapshot
+    ///   ([`ClientWalletStartState::core_wallet_info`]) — consumed
+    ///   directly, preserving per-account UTXO/record attribution and
+    ///   exact pool contents (the FFI/iOS persister); or
+    /// - the keyless projection
+    ///   ([`core_state`](ClientWalletStartState::core_state) +
+    ///   [`used_core_addresses`](ClientWalletStartState::used_core_addresses)),
+    ///   replayed onto a fresh skeleton via
+    ///   [`apply_persisted_core_state`](super::rehydrate::apply_persisted_core_state)
+    ///   (persisters that cannot reconstruct the snapshot).
     ///
     /// The load path never touches the seed, so it performs no wrong-seed
     /// check. Signing happens later, on demand, via the configured
@@ -105,6 +117,7 @@ impl<P: PlatformWalletPersistence + 'static> PlatformWalletManager<P> {
                 network,
                 birth_height,
                 account_manifest,
+                core_wallet_info,
                 core_state,
                 identity_manager,
                 unused_asset_locks,
@@ -112,6 +125,20 @@ impl<P: PlatformWalletPersistence + 'static> PlatformWalletManager<P> {
                 identity_keys,
                 used_core_addresses,
             } = wallet_state;
+
+            // Idempotency, checked FIRST: a wallet already registered (a
+            // prior load pass, or a runtime create) is already-satisfied.
+            // Checking before any reconstruction work matters — the
+            // rebuild below derives eager gap windows (and possibly a
+            // deep discovery scan), all of which the `WalletExists` arm
+            // at insert time would only throw away.
+            {
+                let wm = self.wallet_manager.read().await;
+                if wm.get_wallet(&expected_wallet_id).is_some() {
+                    outcome.loaded.push(expected_wallet_id);
+                    continue 'load;
+                }
+            }
 
             // Build the watch-only wallet from the keyless manifest. A
             // structural decode failure skips this row (per-row
@@ -123,10 +150,8 @@ impl<P: PlatformWalletPersistence + 'static> PlatformWalletManager<P> {
                 &account_manifest,
             ) {
                 Ok(w) => w,
-                Err(row_err) => {
-                    let reason = SkipReason::CorruptPersistedRow {
-                        kind: row_err.into(),
-                    };
+                Err(kind) => {
+                    let reason = SkipReason::CorruptPersistedRow { kind };
                     outcome.skipped.push((expected_wallet_id, reason.clone()));
                     self.event_manager
                         .on_wallet_skipped_on_load(expected_wallet_id, &reason);
@@ -134,21 +159,65 @@ impl<P: PlatformWalletPersistence + 'static> PlatformWalletManager<P> {
                 }
             };
 
-            // Mint the managed-info skeleton from the watch-only wallet,
-            // then apply the keyless persisted core state (UTXOs, sync
-            // watermarks, per-account balances). A wallet with persisted
-            // UTXOs but no funds account hard-fails here rather than
-            // reconstructing a silent zero balance.
-            let mut wallet_info = ManagedWalletInfo::from_wallet(&wallet, birth_height);
-            if let Err(e) = super::rehydrate::apply_persisted_core_state(
-                &mut wallet_info,
-                &account_manifest,
-                &core_state,
-                &used_core_addresses,
-            ) {
-                load_error = Some(e);
-                break 'load;
-            }
+            let wallet_info = match core_wallet_info {
+                // Full keyless snapshot carried by the persister (the
+                // FFI/iOS path): consume it directly. This preserves
+                // per-account UTXO/record attribution, the exact pool
+                // contents (derived-but-unused addresses stay in the SPV
+                // watch set), and per-index used flags — none of which
+                // the projection replay below can reconstruct — and
+                // skips a second eager gap-window derivation.
+                Some(info) => {
+                    let mut info = *info;
+                    // The snapshot must describe this row's wallet; a
+                    // mismatch is a corrupt row, skipped like any other
+                    // structural failure.
+                    if info.wallet_id != expected_wallet_id || info.network != network {
+                        let reason = SkipReason::CorruptPersistedRow {
+                            kind: CorruptKind::DecodeError(format!(
+                                "managed-info snapshot (wallet {}, network {:?}) does not \
+                                 match its row (wallet {}, network {:?})",
+                                hex::encode(info.wallet_id),
+                                info.network,
+                                hex::encode(expected_wallet_id),
+                                network,
+                            )),
+                        };
+                        outcome.skipped.push((expected_wallet_id, reason.clone()));
+                        self.event_manager
+                            .on_wallet_skipped_on_load(expected_wallet_id, &reason);
+                        continue 'load;
+                    }
+                    // Recompute totals from the carried UTXO set so the
+                    // lock-free balance mirrored below can never drift
+                    // from it (no-silent-zero holds by recomputation).
+                    {
+                        use key_wallet::wallet::managed_wallet_info::wallet_info_interface::WalletInfoInterface;
+                        info.update_balance();
+                    }
+                    info
+                }
+                // No snapshot (native/SQLite persister until
+                // dashpay/platform#3968): mint the managed-info skeleton
+                // from the watch-only wallet, then replay the keyless
+                // projection (UTXOs, sync watermarks, used addresses). A
+                // wallet with persisted UTXOs but no funds account
+                // hard-fails here rather than reconstructing a silent
+                // zero balance.
+                None => {
+                    let mut wallet_info = ManagedWalletInfo::from_wallet(&wallet, birth_height);
+                    if let Err(e) = super::rehydrate::apply_persisted_core_state(
+                        &mut wallet_info,
+                        &account_manifest,
+                        &core_state,
+                        &used_core_addresses,
+                    ) {
+                        load_error = Some(e);
+                        break 'load;
+                    }
+                    wallet_info
+                }
+            };
 
             // Flatten the (account → outpoint → lock) map.
             let mut tracked_asset_locks = BTreeMap::new();
--- a/packages/rs-platform-wallet/tests/rehydration_load.rs
+++ b/packages/rs-platform-wallet/tests/rehydration_load.rs
@@ -15,6 +15,10 @@
 //!   fires on the registered handler, `load` returns `Ok`.
 //! - RT-Z: no key/seed material in any `LoadOutcome` / `SkipReason`
 //!   surface (the structural-only contract).
+//! - RT-Snapshot: a carried `core_wallet_info` snapshot is consumed
+//!   verbatim — per-account UTXO attribution and derived-but-unused
+//!   deep pool addresses survive the reload; a snapshot whose
+//!   `wallet_id` mismatches its row is skipped as corrupt.
 
 use std::sync::{Arc, Mutex};
 
@@ -69,6 +73,7 @@ impl PlatformWalletPersistence for FixedLoadPersister {
                             network: w.network,
                             birth_height: w.birth_height,
                             account_manifest: w.account_manifest.clone(),
+                            core_wallet_info: w.core_wallet_info.clone(),
                             core_state: w.core_state.clone(),
                             identity_manager: Default::default(),
                             unused_asset_locks: Default::default(),
@@ -129,6 +134,7 @@ fn slice(seed: [u8; 64]) -> (WalletId, ClientWalletStartState) {
             network: key_wallet::Network::Testnet,
             birth_height: 1,
             account_manifest: manifest,
+            core_wallet_info: None,
             core_state: CoreChangeSet::default(),
             identity_manager: Default::default(),
             unused_asset_locks: Default::default(),
@@ -280,6 +286,7 @@ async fn rt_corrupt_row_skipped_and_other_loads() {
         network: key_wallet::Network::Testnet,
         birth_height: 1,
         account_manifest: Vec::new(),
+        core_wallet_info: None,
         core_state: CoreChangeSet::default(),
         identity_manager: Default::default(),
         unused_asset_locks: Default::default(),
@@ -347,6 +354,7 @@ async fn rt_z_secret_hygiene_surfaces() {
         network: key_wallet::Network::Testnet,
         birth_height: 1,
         account_manifest: Vec::new(),
+        core_wallet_info: None,
         core_state: CoreChangeSet::default(),
         identity_manager: Default::default(),
         unused_asset_locks: Default::default(),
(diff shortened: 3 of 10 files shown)


thepastaclaw:
✅ Review complete (commit 51cd9c9e)

packages/rs-platform-wallet/src/changeset/client_wallet_start_state.rs
```diff
@@ -36,13 +38,33 @@ pub struct ClientWalletStartState {
...
+    pub core_wallet_info: Option<Box<ManagedWalletInfo>>,
     /// Keyless projection of the persisted core rows (UTXOs, tx
     /// records, IS-locks, sync watermarks, `last_applied_chain_lock`).
     /// The manager applies this onto a fresh
     /// `ManagedWalletInfo::from_wallet` skeleton built from the
     /// watch-only wallet. Populated by the persister's
     /// [`PlatformWalletPersistence::load`](crate::changeset::PlatformWalletPersistence::load)
     /// implementation reading the persisted core rows.
+    ///
+    /// Ignored when [`core_wallet_info`](Self::core_wallet_info) is
```
thepastaclaw, line 66: **🟡 Suggestion: Snapshot vs projection fields form an implicit-invariant union**

`ClientWalletStartState` now carries three overlapping sources of truth for core state: `core_wallet_info: Option<Box<ManagedWalletInfo>>`, `core_state: CoreChangeSet`, and `used_core_addresses: Vec<Address>`. The rule "if `core_wallet_info` is `Some`, ignore the other two" lives only in the load-loop match arm and a doc comment; a persister that populates both a snapshot and a projection would silently discard the projection with no compile-time signal. A tagged enum such as `enum CoreLoadState { Snapshot(Box<ManagedWalletInfo>), Projection { core_state: CoreChangeSet, used_core_addresses: Vec<Address> } }` would make the mutual exclusion type-level and remove the FFI persister's need to `Default::default()` the unused fields. Not a bug today but locks in the invariant the docs are pleading for, right before #3968 adds a second persister backend.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet/src/manager/load.rs
```diff
@@ -123,32 +150,74 @@ impl<P: PlatformWalletPersistence + 'static> PlatformWalletManager<P> {
...
+                        &account_manifest,
+                        &core_state,
+                        &used_core_addresses,
+                    ) {
+                        load_error = Some(e);
+                        break 'load;
+                    }
+                    wallet_info
+                }
+            };
```
thepastaclaw, line 220: **🟡 Suggestion: Snapshot path validates wallet_id/network but not manifest ↔ snapshot account consistency**

When `core_wallet_info: Some(info)` is consumed, the manager stitches a watch-only `Wallet` built from `account_manifest` together with `ManagedWalletInfo.accounts` from the snapshot, but only `wallet_id`/`network` are cross-checked. If a future persister populates the manifest and the snapshot from divergent sources (extra CoinJoin account only in one, mis-ordered types, etc.) `Wallet.accounts` and `wallet_info.accounts` disagree silently: one drives scanning/derivation while the other drives balances/UTXOs. The current FFI persister derives both from the same in-memory state so this isn't user-triggerable today, but `PlatformWalletPersistence` is a public trait and #3968 will add a SQLite backend. Even a cheap structural check (assert the account-type sets match) — or at minimum a doc line making the assumption explicit — would extend the existing wallet_id/network fail-closed pattern to the account collection.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet/src/manager/load.rs
```diff
@@ -123,32 +150,74 @@ impl<P: PlatformWalletPersistence + 'static> PlatformWalletManager<P> {
...
+                                info.network,
+                                hex::encode(expected_wallet_id),
+                                network,
+                            )),
+                        };
+                        outcome.skipped.push((expected_wallet_id, reason.clone()));
+                        self.event_manager
+                            .on_wallet_skipped_on_load(expected_wallet_id, &reason);
+                        continue 'load;
+                    }
```
thepastaclaw, line 190: **🟡 Suggestion: Snapshot wallet_id/network mismatch reuses CorruptKind::DecodeError, obscuring the failure mode**

A wallet_id/network mismatch on the persisted `core_wallet_info` snapshot is reported via `CorruptKind::DecodeError`, which the FFI maps to `LOAD_SKIP_REASON_DECODE_ERROR` (102) — the same code as a genuine bincode failure. These are semantically distinct: one means the persisted bytes are unreadable, the other means a persister returned a snapshot for the wrong row (a stronger data-integrity signal, e.g. persister row mixup). `CorruptKind` is `#[non_exhaustive]`, so a dedicated `SnapshotIdentityMismatch` variant (plus a new `LOAD_SKIP_REASON_*` code) is a low-cost, backward-compatible change that would make debug logs and Swift-side triage clearer.

<sub>source: ['claude']</sub>

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManager.swift
```diff
@@ -319,6 +326,29 @@ public class PlatformWalletManager: ObservableObject {
...
+        public let walletId: Data
+        public let reasonCode: UInt32
+
+        /// Human-readable rendering of `reasonCode`, mirroring the Rust
+        /// `LOAD_SKIP_REASON_*` constants.
+        public var reasonDescription: String {
+            switch reasonCode {
+            case 100: return "missing account manifest"
+            case 101: return "malformed account xpub"
+            case 102: return "decode error"
```
thepastaclaw, line 344: **💬 Nitpick: Swift reasonDescription hardcodes wire numbers instead of importing LOAD_SKIP_REASON_* constants**

Rust FFI exports `LOAD_SKIP_REASON_MISSING_MANIFEST` (100), `LOAD_SKIP_REASON_MALFORMED_XPUB` (101), `LOAD_SKIP_REASON_DECODE_ERROR` (102), `LOAD_SKIP_REASON_CORRUPT_OTHER` (199), `LOAD_SKIP_REASON_OTHER` (200) as `pub const u32` specifically so consumers can reference them by name. Swift's `SkippedWalletOnLoad.reasonDescription` re-encodes those numbers as bare `case 100:` literals, duplicating the wire contract. The Rust-side `load_skip_reason_wire_values_are_stable` test pins drift, but referencing the imported constants (`case UInt32(LOAD_SKIP_REASON_MISSING_MANIFEST):` …) would keep the two sides tied by symbol rather than convention.

<sub>source: ['claude']</sub>

thepastaclaw (approved):
## Code Review

PR #3980 cleanly addresses the four seedless-rehydration defects from #3692 (attribution loss, watch-set shrinkage, wedge-zone address reuse, Swift discarding LoadOutcome). Snapshot path is guarded by wallet_id/network fail-closed, mark↔refill fixpoint is correctly bounded on a finite restored-address set, FFI ownership/free contract is symmetric with a zero-init on early returns, and Swift now consumes and frees the outcome. No blockers; a handful of non-blocking suggestions and nitpicks — mostly about type-safety of the snapshot-vs-projection union, the DecodeError reuse for identity-mismatch, and Swift/Rust constant coupling.

🟡 5 suggestion(s) | 💬 3 nitpick(s)

<details>

```

In `packages/rs-platform-wallet/src/changeset/client_wallet_start_state.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/changeset/client_wallet_start_state.rs:41-66: Snapshot vs projection fields form an implicit-invariant union
  `ClientWalletStartState` now carries three overlapping sources of truth for core state: `core_wallet_info: Option<Box<ManagedWalletInfo>>`, `core_state: CoreChangeSet`, and `used_core_addresses: Vec<Address>`. The rule "if `core_wallet_info` is `Some`, ignore the other two" lives only in the load-loop match arm and a doc comment; a persister that populates both a snapshot and a projection would silently discard the projection with no compile-time signal. A tagged enum such as `enum CoreLoadState { Snapshot(Box<ManagedWalletInfo>), Projection { core_state: CoreChangeSet, used_core_addresses: Vec<Address> } }` would make the mutual exclusion type-level and remove the FFI persister's need to `Default::default()` the unused fields. Not a bug today but locks in the invariant the docs are pleading for, right before #3968 adds a second persister backend.

In `packages/rs-platform-wallet/src/manager/load.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/manager/load.rs:162-220: Snapshot path validates wallet_id/network but not manifest ↔ snapshot account consistency
  When `core_wallet_info: Some(info)` is consumed, the manager stitches a watch-only `Wallet` built from `account_manifest` together with `ManagedWalletInfo.accounts` from the snapshot, but only `wallet_id`/`network` are cross-checked. If a future persister populates the manifest and the snapshot from divergent sources (extra CoinJoin account only in one, mis-ordered types, etc.) `Wallet.accounts` and `wallet_info.accounts` disagree silently: one drives scanning/derivation while the other drives balances/UTXOs. The current FFI persister derives both from the same in-memory state so this isn't user-triggerable today, but `PlatformWalletPersistence` is a public trait and #3968 will add a SQLite backend. Even a cheap structural check (assert the account-type sets match) — or at minimum a doc line making the assumption explicit — would extend the existing wallet_id/network fail-closed pattern to the account collection.
- [SUGGESTION] packages/rs-platform-wallet/src/manager/load.rs:175-190: Snapshot wallet_id/network mismatch reuses CorruptKind::DecodeError, obscuring the failure mode
  A wallet_id/network mismatch on the persisted `core_wallet_info` snapshot is reported via `CorruptKind::DecodeError`, which the FFI maps to `LOAD_SKIP_REASON_DECODE_ERROR` (102) — the same code as a genuine bincode failure. These are semantically distinct: one means the persisted bytes are unreadable, the other means a persister returned a snapshot for the wrong row (a stronger data-integrity signal, e.g. persister row mixup). `CorruptKind` is `#[non_exhaustive]`, so a dedicated `SnapshotIdentityMismatch` variant (plus a new `LOAD_SKIP_REASON_*` code) is a low-cost, backward-compatible change that would make debug logs and Swift-side triage clearer.

In `packages/rs-platform-wallet/src/manager/rehydrate.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/manager/rehydrate.rs:462-478: New RehydrationPoolTypeMismatch runtime guard is not exercised by tests
  The `debug_assert!(pool.pool_type == probe.pool_type)` was promoted to a runtime fail-closed `Err(RehydrationPoolTypeMismatch { .. })` in release builds — good hardening — but no test verifies that a mis-ordered probe/pool pair actually produces this error. Because `probes` is built from `address_pools()` and the real pools come from `address_pools_mut()` on the same account, the mismatch is only reachable via divergence between the two `key_wallet` accessors. A small structural-invariant test (or a `debug_assert_eq!` alongside the `Err`) would catch a future upstream regression. Same applies to the pre-existing `RehydrationPoolMismatch` length check.

In `packages/rs-platform-wallet/tests/rehydration_load.rs`:
- [SUGGESTION] packages/rs-platform-wallet/tests/rehydration_load.rs:1: No test combines snapshot-mismatch skip with a co-occurring successful load
  `rt_snapshot_wallet_id_mismatch_is_skipped` verifies the mismatched row is skipped in isolation, and `rt_corrupt_row_skipped_and_other_loads` verifies a corrupt-manifest skip alongside a successful load in the same batch. There's no equivalent combined-batch test for the new snapshot-mismatch skip path (skipped wallet + a second healthy wallet loading successfully in the same `load_from_persistor` call). Extending the existing pattern would confirm the mismatch skip doesn't poison siblings in the batch.
```
</details>

packages/rs-platform-wallet/tests/rehydration_load.rs
```diff
@@ -369,3 +377,211 @@ async fn rt_z_secret_hygiene_surfaces() {
...
+    assert_eq!(*skipped_id, id_a);
+    assert!(matches!(
+        reason,
+        SkipReason::CorruptPersistedRow {
+            kind: CorruptKind::DecodeError(_)
+        }
+    ));
+    assert!(mgr.get_wallet(&id_a).await.is_none());
+    assert_eq!(h.skipped.lock().unwrap().len(), 1);
+}
```
thepastaclaw, line 587: **🟡 Suggestion: No test combines snapshot-mismatch skip with a co-occurring successful load**

STILL VALID (prior-5, code unchanged). `rt_snapshot_wallet_id_mismatch_is_skipped` verifies the mismatched row is skipped in isolation, and `rt_corrupt_row_skipped_and_other_loads` verifies a corrupt-manifest skip alongside a successful load in the same batch. There's no equivalent combined-batch test for the new snapshot-mismatch skip path (skipped wallet + a second healthy wallet loading successfully in the same `load_from_persistor` call). Extending the existing pattern would confirm the mismatch skip doesn't poison siblings in the batch.

<sub>source: ['opus-general', 'sonnet5-rust-quality', 'codex-general', 'codex-rust-quality']</sub>

thepastaclaw (approved):
## Code Review

Latest push (e66fbf55..51cd9c9e) is documentation-only — a 93-line 'Persistence architecture' section added to packages/rs-platform-wallet/README.md, with no changes to Rust or Swift source. All 8 prior findings remain valid at unchanged line ranges. One new suggestion from codex-general (projection-fallback path routes used addresses to only the first funds account) is included. No in-scope blockers.

🟡 8 suggestion(s) | 💬 5 nitpick(s)

<details>
<summary>Findings not posted inline (4)</summary>

- **[SUGGESTION] `packages/rs-platform-wallet/src/manager/rehydrate.rs:223-264`: Projection fallback only restores used flags on the first funds account** — `used_pool_addresses` is documented as a flattened snapshot across all accounts/pools, but both replay branches at lines 223-249 and 250-263 route the entire set through `wallet_info.accounts.all_funding_accounts_mut().into_iter().next()` and call `extend_pools_for_restored_addresses` on only tha...
- **[SUGGESTION] `packages/rs-platform-wallet/src/manager/rehydrate.rs:462-478`: New RehydrationPoolTypeMismatch runtime guard is not exercised by tests** — STILL VALID (prior-4, code unchanged). The `debug_assert!(pool.pool_type == probe.pool_type)` was promoted to a runtime fail-closed `Err(PlatformWalletError::RehydrationPoolTypeMismatch { .. })` in release builds (good hardening), but no test verifies that a mis-ordered probe/pool pair actually p...
- **[NITPICK] `packages/rs-platform-wallet-ffi/src/manager.rs:384-407`: Document that non-null out_outcome must be zero-initialised or previously freed** — STILL VALID (prior-6, code unchanged). `std::ptr::write(out_outcome, LoadOutcomeFFI { .., skipped: null_mut() })` unconditionally overwrites the destination without inspecting or freeing any prior `skipped` array — `ptr::write` does not run Drop. If a non-Swift caller reused the same `LoadOutcome...
- **[NITPICK] `packages/rs-platform-wallet-ffi/src/manager.rs:463-474`: platform_wallet_load_outcome_free leaves loaded_count non-zero after freeing** — STILL VALID (prior-8, code unchanged). The free function clears `skipped_count`/`skipped` for idempotence but leaves `loaded_count` untouched. The doc string says the free is idempotent and nulls the pointer, so callers who inspect the struct post-free see a stale `loaded_count`. No memory-safety...

</details>

<details>
<summary>Carried-forward findings already raised (4)</summary>

- **[SUGGESTION] (deduped existing open thread) `packages/rs-platform-wallet/src/changeset/client_wallet_start_state.rs:41-97`: Snapshot vs projection fields form an implicit-invariant union** — STILL VALID (prior-1, code unchanged). `ClientWalletStartState` carries three overlapping sources of truth for core state: `core_wallet_info: Option<Box<ManagedWalletInfo>>` (snapshot) and `core_state: CoreChangeSet` + `used_core_addresses: Vec<Address>` (projection). The 'if `core_wallet_info` i...
- **[SUGGESTION] (deduped existing open thread) `packages/rs-platform-wallet/src/manager/load.rs:162-220`: Snapshot path validates wallet_id/network but not manifest↔snapshot account consistency** — STILL VALID (prior-2, code unchanged). When `core_wallet_info: Some(info)` is consumed, the manager stitches a watch-only `Wallet` built from `account_manifest` together with `ManagedWalletInfo.accounts` from the snapshot, but only `wallet_id`/`network` are cross-checked. If a future persister po...
- **[SUGGESTION] (deduped existing open thread) `packages/rs-platform-wallet/src/manager/load.rs:175-190`: Snapshot wallet_id/network mismatch reuses CorruptKind::DecodeError, obscuring the failure mode** — STILL VALID (prior-3, code unchanged). A wallet_id/network mismatch on the persisted `core_wallet_info` snapshot is reported via `CorruptKind::DecodeError(format!(...))`, which the FFI maps to `LOAD_SKIP_REASON_DECODE_ERROR` (102) — the same code as a genuine bincode failure. These are semantical...
- **[NITPICK] (deduped existing open thread) `packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManager.swift:340-349`: Swift reasonDescription hardcodes wire numbers instead of importing LOAD_SKIP_REASON_* constants** — STILL VALID (prior-7, code unchanged). Rust exports `LOAD_SKIP_REASON_MISSING_MANIFEST` (100), `LOAD_SKIP_REASON_MALFORMED_XPUB` (101), `LOAD_SKIP_REASON_DECODE_ERROR` (102), `LOAD_SKIP_REASON_CORRUPT_OTHER` (199), `LOAD_SKIP_REASON_OTHER` (200) as `pub const u32` specifically so consumers can re...

</details>

<details>

```

In `packages/rs-platform-wallet/src/manager/rehydrate.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/manager/rehydrate.rs:223-264: Projection fallback only restores used flags on the first funds account
  `used_pool_addresses` is documented as a flattened snapshot across all accounts/pools, but both replay branches at lines 223-249 and 250-263 route the entire set through `wallet_info.accounts.all_funding_accounts_mut().into_iter().next()` and call `extend_pools_for_restored_addresses` on only that first funds account. A fallback persister that cannot provide `core_wallet_info` but does supply used addresses for a non-first account (e.g. CoinJoin, DashPay) will never probe those addresses against the matching account xpub, so they remain `used = false` and can be handed out again after seedless load — defeating the exact address-reuse guard this PR added via `used_pool_addresses`. Not user-triggerable today because the in-tree FFI persister always sets `core_wallet_info` and SQLite does not hydrate wallets yet (dashpay/platform#3968), but the public fallback contract is incorrect. Fix by grouping `addresses_to_mark` by their owning account (or dispatching per pool) before calling `extend_pools_for_restored_addresses`.
- [SUGGESTION] packages/rs-platform-wallet/src/manager/rehydrate.rs:462-478: New RehydrationPoolTypeMismatch runtime guard is not exercised by tests
  STILL VALID (prior-4, code unchanged). The `debug_assert!(pool.pool_type == probe.pool_type)` was promoted to a runtime fail-closed `Err(PlatformWalletError::RehydrationPoolTypeMismatch { .. })` in release builds (good hardening), but no test verifies that a mis-ordered probe/pool pair actually produces this error. Because `probes` is built from `address_pools()` and the real pools come from `address_pools_mut()` on the same account, the mismatch is only reachable via divergence between the two `key_wallet` accessors. A small structural-invariant test would catch a future upstream regression. Same applies to the pre-existing `RehydrationPoolMismatch` length check at lines 456-461.
- [SUGGESTION] packages/rs-platform-wallet/src/manager/rehydrate.rs:223-264: Projection fallback only restores used flags on the first funds account
  `used_pool_addresses` is documented as a flattened snapshot across all accounts/pools, but both replay branches at lines 223-249 and 250-263 route the entire set through `wallet_info.accounts.all_funding_accounts_mut().into_iter().next()` and call `extend_pools_for_restored_addresses` on only that first funds account. A fallback persister that cannot provide `core_wallet_info` but does supply used addresses for a non-first account (e.g. CoinJoin, DashPay) will never probe those addresses against the matching account xpub, so they remain `used = false` and can be handed out again after seedless load — defeating the exact address-reuse guard this PR added via `used_pool_addresses`. Not user-triggerable today because the in-tree FFI persister always sets `core_wallet_info` and SQLite does not hydrate wallets yet (dashpay/platform#3968), but the public fallback contract is incorrect. Fix by grouping `addresses_to_mark` by their owning account (or dispatching per pool) before calling `extend_pools_for_restored_addresses`.
- [SUGGESTION] packages/rs-platform-wallet/src/manager/rehydrate.rs:462-478: New RehydrationPoolTypeMismatch runtime guard is not exercised by tests
  STILL VALID (prior-4, code unchanged). The `debug_assert!(pool.pool_type == probe.pool_type)` was promoted to a runtime fail-closed `Err(PlatformWalletError::RehydrationPoolTypeMismatch { .. })` in release builds (good hardening), but no test verifies that a mis-ordered probe/pool pair actually produces this error. Because `probes` is built from `address_pools()` and the real pools come from `address_pools_mut()` on the same account, the mismatch is only reachable via divergence between the two `key_wallet` accessors. A small structural-invariant test would catch a future upstream regression. Same applies to the pre-existing `RehydrationPoolMismatch` length check at lines 456-461.

In `packages/rs-platform-wallet/tests/rehydration_load.rs`:
- [SUGGESTION] packages/rs-platform-wallet/tests/rehydration_load.rs:546-587: No test combines snapshot-mismatch skip with a co-occurring successful load
  STILL VALID (prior-5, code unchanged). `rt_snapshot_wallet_id_mismatch_is_skipped` verifies the mismatched row is skipped in isolation, and `rt_corrupt_row_skipped_and_other_loads` verifies a corrupt-manifest skip alongside a successful load in the same batch. There's no equivalent combined-batch test for the new snapshot-mismatch skip path (skipped wallet + a second healthy wallet loading successfully in the same `load_from_persistor` call). Extending the existing pattern would confirm the mismatch skip doesn't poison siblings in the batch.

In `packages/rs-platform-wallet/src/changeset/client_wallet_start_state.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/changeset/client_wallet_start_state.rs:41-97: Snapshot vs projection fields form an implicit-invariant union
  STILL VALID (prior-1, code unchanged). `ClientWalletStartState` carries three overlapping sources of truth for core state: `core_wallet_info: Option<Box<ManagedWalletInfo>>` (snapshot) and `core_state: CoreChangeSet` + `used_core_addresses: Vec<Address>` (projection). The 'if `core_wallet_info` is Some, ignore the other two' rule lives only in a match arm in `manager/load.rs` and doc comments — a persister that populates both would silently discard the projection with no compile-time signal. The new README section documents these projection fields as 'a transitional fallback … slated for removal once every in-tree persister produces snapshots', which strengthens the case for a tagged enum `enum CoreLoadState { Snapshot(Box<ManagedWalletInfo>), Projection { core_state, used_core_addresses } }`: it would (a) make mutual exclusion type-level, (b) remove the FFI persister's need to `Default::default()` unused fields, and (c) shrink the deletion when #3968 lands to removing one variant.

In `packages/rs-platform-wallet/src/manager/load.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/manager/load.rs:162-220: Snapshot path validates wallet_id/network but not manifest↔snapshot account consistency
  STILL VALID (prior-2, code unchanged). When `core_wallet_info: Some(info)` is consumed, the manager stitches a watch-only `Wallet` built from `account_manifest` together with `ManagedWalletInfo.accounts` from the snapshot, but only `wallet_id`/`network` are cross-checked. If a future persister populates the manifest and snapshot from divergent sources (extra CoinJoin account only in one, mis-ordered types, etc.), `Wallet.accounts` and `wallet_info.accounts` disagree silently: one drives scanning/derivation while the other drives balances/UTXOs. Not user-triggerable today since the FFI persister derives both from the same in-memory state, but `PlatformWalletPersistence` is a public trait and #3968 will add a SQLite backend. A cheap structural check comparing account-type sets — or at minimum an explicit doc line — would extend the wallet_id/network fail-closed pattern to the account collection.
- [SUGGESTION] packages/rs-platform-wallet/src/manager/load.rs:175-190: Snapshot wallet_id/network mismatch reuses CorruptKind::DecodeError, obscuring the failure mode
  STILL VALID (prior-3, code unchanged). A wallet_id/network mismatch on the persisted `core_wallet_info` snapshot is reported via `CorruptKind::DecodeError(format!(...))`, which the FFI maps to `LOAD_SKIP_REASON_DECODE_ERROR` (102) — the same code as a genuine bincode failure. These are semantically distinct: unreadable bytes vs a persister returning a snapshot for the wrong row (a stronger data-integrity signal, e.g. persister row mixup, especially relevant once SQLite lands per #3968). `CorruptKind` is `#[non_exhaustive]`, so a dedicated `SnapshotIdentityMismatch` variant plus a new `LOAD_SKIP_REASON_*` code is a low-cost backward-compatible improvement for debug logs and Swift-side triage.
```
</details>

packages/rs-platform-wallet/src/manager/load.rs
thepastaclaw: Still present at `51cd9c9`.

packages/rs-platform-wallet/src/changeset/client_wallet_start_state.rs
Claudius-Maginificent: Good idea in principle, but we're deliberately deferring the tagged `CoreLoadState::Snapshot/Projection` enum to the #3968 redirect. That PR deletes the projection fields (`core_state` / `used_core_addresses`) once every persister carries the full snapshot, which moots the mutual-exclusion the enum would enforce — encoding it now only to tear it down in #3968 would be churn. Tracking it there instead.

packages/rs-platform-wallet/src/manager/load.rs
Claudius-Maginificent: Fixed in #3984 (stacked on this branch), commit 9879a9edbc. The snapshot path now cross-checks the carried `ManagedWalletInfo`'s account set against the row's `account_manifest` (helper `snapshot_accounts_match_manifest`), and a mismatch skips the row through the existing skip machinery. Note the two enumerations differ structurally (manifest is ECDSA-only + PlatformPayment; snapshot carries BLS/EdDSA provider keys, no PlatformPayment), so the check compares only the families both can carry — otherwise legitimate snapshots would be rejected. Covered by `rt_snapshot_account_set_mismatch_is_skipped`.

packages/rs-platform-wallet/src/manager/load.rs
Claudius-Maginificent: Fixed in #3984 (stacked on this branch), commit 9879a9edbc. Added a dedicated `CorruptKind::SnapshotIdentityMismatch` variant; the `wallet_id`/`network` (and now account-set) mismatch maps to it instead of reusing `DecodeError`, so a wrong-row snapshot is no longer conflated with unreadable bytes. New FFI wire code `LOAD_SKIP_REASON_SNAPSHOT_IDENTITY_MISMATCH = 103` with the Swift `reasonDescription` mapping and both sides' tests updated.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManager.swift
Claudius-Maginificent: Addressed in #3984 (stacked on this branch), commit 9879a9edbc. The `LOAD_SKIP_REASON_*` constants aren't surfaced as named symbols in the generated C header (no header is checked in — cbindgen emits into the xcframework at build time), and Swift can't be compiled on this host to verify they'd be reachable, so the numeric `switch` stays with `case 103` added, plus a comment pointing at the wire-contract source of truth (`rs-platform-wallet-ffi/src/manager.rs`). If a follow-up confirms the constants are exported into the Swift-visible header, swapping the literals for the named constants is a clean mechanical change.

packages/rs-platform-wallet/tests/rehydration_load.rs
Claudius-Maginificent: Added in #3984 (stacked on this branch), commit 8999eeabc5: `rt_snapshot_mismatch_skip_coexists_with_healthy_load` exercises a `SnapshotIdentityMismatch` skip and a successful snapshot load in the same batch — healthy row loads, mismatched row skipped with the new variant, handler fires once, batch returns `Ok`. Mirrors the existing corrupt-manifest + healthy-load combined test.

PR #3981, opened 2026-07-02
feat(swift-sdk): transaction decoder over upstream key-wallet-ffi transaction_decode

llbartekll:

Host apps consuming the platform-wallet FFI (dashwallet-ios mid-migration off DashSync) need to match **arbitrary persisted transactions** against external addresses and exact amounts — e.g. the CrowdNode on-chain API protocol: *"did a transaction pay `apiOffset + code` duffs to the CrowdNode address?"* and *"did CrowdNode reply with amount X to my account address?"*.

The wallet layer persists every transaction's **raw bytes**, but only materializes TXO rows for the wallet's *own* outputs, and `AccountTransactionEntryFFI` carries no per-output data. There was **no way to evaluate a per-output `(address, amount)` predicate** from either the FFI or the Swift SDK — payments *to* external addresses (CrowdNode's) had no addressable representation at all. This blocked porting CrowdNode's on-chain response tracking (and several tx-history readers) off DashSync.

## What was done?

Adds `TransactionDecoder` to the Swift SDK, backed by the **upstream** decode FFI in rust-dashcore's key-wallet-ffi.

The decode entry point itself is pure Core-chain functionality (consensus decode, `Address::from_script`, no wallet handle or Platform state), so per review it was upstreamed rather than kept here: dashpay/rust-dashcore#869 → dashpay/rust-dashcore#870 (merged) add `transaction_decode` / `decoded_transaction_free` to key-wallet-ffi — one-shot struct-out decode into per-output `(address, value_duffs, script_pubkey)` and per-input previous outpoint + strictly-P2PKH-shaped, **unauthenticated** sender address (display/matching hint only; the attacker-resistant matching primitive is the per-output `(address, value_duffs)` pair). An earlier revision of this PR carried the same function in `rs-platform-wallet-ffi`; that copy is deleted.

This PR now contains:

- **rust-dashcore pin bump** to `d4692bb2` (exactly one commit ahead: the upstream decoder, additive only).
- **`swift-sdk` — `Sources/SwiftDashSDK/KeyWallet/TransactionDecoder.swift`:** `TransactionDecoder.decode(_ txData: Data, network: Network) throws -> DecodedTransaction` — thin wrapper over `transaction_decode` using the key-wallet-ffi `FFIError` + bool convention, typed `DecodedTransaction`/`Input`/`Output` structs, `defer`-freed FFI memory, `txidDisplayHex` / `prevTxidDisplayHex` display helpers. Throws `KeyWalletError.invalidInput` for malformed/empty bytes (including trailing garbage).
- **`TransactionDecoderTests`** — six pure-function tests against a byte-exact fixture shared with the Rust-side suite upstream (13 Rust unit tests live in key-wallet-ffi with the decoder).

## How Has This Been Tested?

- Upstream: full key-wallet-ffi suite + Address/Thread Sanitizer green on dashpay/rust-dashcore#870.
- Here: `cargo check -p platform-wallet-ffi` against the bumped pin; `swift test` — full package suite green (260 tests) including `TransactionDecoderTests` (6/6) and a Rust-side cross-check that the exact Swift fixture hex decodes through the FFI to the expected txid/addresses.
- The FFI surface was smoke-tested end-to-end on testnet by the consuming app (dashwallet-ios CrowdNode response tracking) against the pre-upstreaming build — the C struct layout is byte-identical; only the symbol names and error convention changed.

## Breaking Changes

None — purely additive (new Swift wrapper; upstream FFI symbols arrive via the pin bump).

## Checklist:

- [x] I have made corresponding changes to the documentation *(rustdoc lives upstream with the decoder; Swift wrapper fully doc-commented)*

- [x] I have assigned this pull request to a milestone *(v4.1.0)*

## Summary by CodeRabbit

- **Improvements**
  - Improved transaction decoding reliability for inputs, outputs, transaction IDs, and network-specific addresses.
  - Enhanced handling of malformed, incomplete, or empty transaction data with clearer decoding errors.
  - Improved support for identifying script-based outputs that do not contain standard addresses.

- **Quality**
  - Added broader validation coverage for transaction decoding across common and edge-case scenarios.


thepastaclaw:
✅ Review complete (commit edc0be7c)

packages/rs-platform-wallet/src/transaction_decode.rs
```diff
@@ -0,0 +1,229 @@
...
+            Err(_) => return None,
+        }
+    }
+    let candidate = last_push?;
+    if candidate.len() != 33 && candidate.len() != 65 {
+        return None;
+    }
+    let pubkey = PublicKey::from_slice(&candidate).ok()?;
+    Some(Address::p2pkh(&pubkey, network))
+}
```
thepastaclaw: **🟡 Suggestion: P2PKH sender heuristic doesn't verify scriptSig shape, only the last push's length**

`input_address_from_script_sig` iterates every instruction in the scriptSig and takes whichever data push occurred last, then treats it as a pubkey if it is 33 or 65 bytes. It never checks that the scriptSig actually has the P2PKH shape (exactly two pushes: signature then pubkey). For a P2SH spend the last push is the redeem script; if that redeem script happens to be 33 or 65 bytes and parses as a valid secp256k1 point (spender-controllable, only constrained by hashing to the P2SH address), the function silently returns a P2PKH address unrelated to the actual signer. The module doc claims `None` for non-P2PKH, but the implementation doesn't uphold that.

The simplest tightening is to require exactly two pushes (`sig`, `pubkey`) — anything else, including multisig P2SH redeem scripts, returns `None`. An additional guard that the first push looks like a DER-encoded ECDSA signature (starts with `0x30`, length ≤ 73) would eliminate essentially all remaining collision risk. This is not blocking for the stated CrowdNode use case (which matches on outputs, not input senders), but the failure mode — a wrong non-null address is worse than `None` — is worth closing since callers building sender-verification logic on top of `input.address` would be fooled.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet/src/transaction_decode.rs
```diff
@@ -0,0 +1,229 @@
...
+            value_duffs: txout.value,
+            script_pubkey: txout.script_pubkey.clone(),
+        })
+        .collect();
+
+    Ok(DecodedTransaction {
+        txid: tx.txid(),
+        inputs,
+        outputs,
+    })
```
thepastaclaw: **🟡 Suggestion: Unnecessary clone of every output's scriptPubkey**

`decode_transaction` iterates `tx.output` by reference and clones every `script_pubkey` to populate `DecodedOutput::script_pubkey`. Since `tx.txid()` only needs `&self` and doesn't care about the order relative to consuming other fields, the txid can be computed first, after which `tx.output` (and `tx.input`) can be consumed via `into_iter()` — eliminating the per-output `ScriptBuf` clone entirely. For transactions with large scripts (OP_RETURN payloads, multisig redeem scripts) this is a real, avoidable allocation on every decode call, and this function is positioned as the primitive for scanning arbitrary persisted transactions.

```suggestion
    let txid = tx.txid();

    let inputs = tx
        .input
        .into_iter()
        .map(|txin| {
            let address = if txin.previous_output.is_null() {
                None // coinbase
            } else {
                input_address_from_script_sig(&txin.script_sig, network)
            };
            DecodedInput {
                previous_output: txin.previous_output,
                address,
            }
        })
        .collect();

    let outputs = tx
        .output
        .into_iter()
        .map(|txout| DecodedOutput {
            address: Address::from_script(&txout.script_pubkey, network).ok(),
            value_duffs: txout.value,
            script_pubkey: txout.script_pubkey,
        })
        .collect();

    Ok(DecodedTransaction {
        txid,
        inputs,
        outputs,
    })
```

<sub>source: ['claude']</sub>

packages/rs-platform-wallet-ffi/src/tx_decode.rs
```diff
@@ -0,0 +1,288 @@
...
+    out_decoded: *mut *mut DecodedTransactionFFI,
+) -> PlatformWalletFFIResult {
+    check_ptr!(tx_bytes);
+    check_ptr!(out_decoded);
+    if tx_bytes_len == 0 {
+        return PlatformWalletFFIResult::err(
+            PlatformWalletFFIResultCode::ErrorInvalidParameter,
+            "tx_bytes_len is zero",
+        );
+    }
```
thepastaclaw: **🟡 Suggestion: `*out_decoded` is not zeroed on failure paths**

On every error return (`check_ptr!(tx_bytes)`, the `tx_bytes_len == 0` branch, `unwrap_result_or_return!(decode_transaction(...))`), the function returns a non-Success `PlatformWalletFFIResult` without assigning to `*out_decoded`. The Swift wrapper is safe because it initializes `outDecoded = nil` and checks the result code first, but any other C/Swift caller that reuses an already-populated out pointer — or reads it before checking the result code — will observe stale data and may double-free by re-calling `platform_wallet_decoded_transaction_free` on the previous value. Convention for out-pointers is to null on entry: after `check_ptr!(out_decoded)`, do `*out_decoded = std::ptr::null_mut();`. Small change, removes a foot-gun for anyone consuming this from C.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet/src/transaction_decode.rs
```diff
@@ -0,0 +1,229 @@
...
+        let (tx, _) = synthetic_tx(Network::Testnet);
+        let mut bytes = serialize(&tx);
+        bytes.extend_from_slice(&[0xDE, 0xAD]);
+        assert!(decode_transaction(&bytes, Network::Testnet).is_err());
+    }
+
+    #[test]
+    fn garbage_bytes_are_an_error() {
+        assert!(decode_transaction(&[0xFFu8; 16], Network::Testnet).is_err());
+    }
```
thepastaclaw: **🟡 Suggestion: No test exercises a non-P2PKH scriptSig (the documented `None` path)**

The only "no address" input test is `coinbase_input_has_no_address`. There is no test where the scriptSig's last push is not a valid P2PKH-shaped pubkey — for example a P2SH spend with a redeem-script push, a 3+-push scriptSig, or a 33/65-byte push that fails `PublicKey::from_slice`. These are exactly the ambiguous cases the length + parse checks are meant to filter out, and they'd pin the intended P2PKH-only contract so a future loosening of the heuristic (or the fix suggested above) has regression coverage.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet-ffi/src/tx_decode.rs
```diff
@@ -0,0 +1,288 @@
...
+pub struct DecodedTxInputFFI {
+    /// Previous output's txid in consensus (internal) byte order.
+    pub prev_txid: [u8; 32],
+    /// Previous output's index.
+    pub prev_vout: u32,
+    /// Sender address recovered from a P2PKH scriptSig, or null when the
+    /// input is coinbase / non-P2PKH / unparseable. Owned by the parent
+    /// structure — freed by `platform_wallet_decoded_transaction_free`.
+    pub address: *mut c_char,
+}
```
thepastaclaw: **🟡 Suggestion: Document that `DecodedTxInputFFI::address` is unauthenticated and must not be used for authorization**

The per-input `address` is derived from the last scriptSig push with no signature verification tying it to the UTXO being spent. A crafter can put any 33/65-byte value in that position, so the field is trivially spoofable. The pure module doc says "best-effort / sent from", but the FFI struct comment only says the value is "recovered from a P2PKH scriptSig", which a downstream consumer (dashwallet-ios, future CrowdNode integration) could reasonably read as "the sender is P" and use for routing/authorization decisions. Strengthen the comment on `DecodedTxInputFFI::address` (and the mirrored Swift `Input.address`) to say explicitly: derived from an unauthenticated script push, safe for display or matching hints only, not for authentication/authorization. The safe matching primitive for CrowdNode-style protocols is per-output `(address, amount)`, which is already exposed and not attacker-influenceable in the same way.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet/src/transaction_decode.rs
thepastaclaw: **💬 Nitpick: `last_push` reallocates on every push instead of holding a borrow**

The loop allocates a fresh `Vec<u8>` for every push instruction, discarding the previous allocation, only to keep the last one. Since `PushBytes` borrows from the underlying `ScriptBuf`, you can hold `&[u8]` for the loop duration and only inspect it at the end. Not a hot-path concern, but easy to fix and avoids `O(n_pushes)` allocations.

```suggestion
fn input_address_from_script_sig(script_sig: &ScriptBuf, network: Network) -> Option<Address> {
    let mut last_push: Option<&[u8]> = None;
    for instruction in script_sig.instructions() {
        match instruction {
            Ok(Instruction::PushBytes(bytes)) => last_push = Some(bytes.as_bytes()),
            Ok(_) => {}
            Err(_) => return None,
        }
    }
    let candidate = last_push?;
    if candidate.len() != 33 && candidate.len() != 65 {
        return None;
    }
    let pubkey = PublicKey::from_slice(candidate).ok()?;
    Some(Address::p2pkh(&pubkey, network))
}
```

<sub>source: ['claude']</sub>

packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/TransactionDecoder.swift
```diff
@@ -0,0 +1,113 @@
...
+    /// explorer-style display (see `txidDisplayHex`).
+    public let txid: Data
+    public let inputs: [Input]
+    public let outputs: [Output]
+
+    /// Explorer-style (reversed, hex) rendering of `txid`.
+    public var txidDisplayHex: String {
+        txid.reversed().map { String(format: "%02x", $0) }.joined()
+    }
+}
```
thepastaclaw, line 54: **💬 Nitpick: `Input.prevTxid` lacks a display-hex helper analogous to `txidDisplayHex`**

`txid` gets a `txidDisplayHex` computed property that reverses to explorer order, but `Input.prevTxid` (same consensus/internal byte-order convention per the FFI doc comment) has no parallel helper. Callers matching a stored transaction's inputs against an explorer-copied txid string will silently produce a byte-reversed mismatch. Given the CrowdNode use case involves matching known txids, exposing a symmetric `prevTxidDisplayHex` (or a shared helper) would prevent a common footgun. Pure ergonomics/consistency note.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Small, additive PR adding a pure transaction decoder in rs-platform-wallet plus a thin FFI/Swift wrapper. FFI memory management is correctly paired (Box::into_raw / Vec::from_raw_parts / CString round-trips), no consensus surfaces are touched, and tests cover the happy path plus trailing-garbage/coinbase/malformed cases. No blocking issues. Main suggestions cluster around the P2PKH sender-address heuristic (which infers a sender from the last scriptSig push without verifying P2PKH shape, so P2SH spends could occasionally be misattributed) and a couple of small performance/defensive tightenings.

🟡 5 suggestion(s) | 💬 2 nitpick(s)

Source: reviewers: general(opus, claude-sonnet-5, gpt-5.5[high] failed/unparseable), security-auditor(opus, claude-sonnet-5, gpt-5.5[high] failed/unparseable), rust-quality(opus, claude-sonnet-5, gpt-5.5[high] failed/unparseable), ffi-engineer(opus, claude-sonnet-5, gpt-5.5[high] failed/unparseable); verifier: opus.

<details>

```

In `packages/rs-platform-wallet/src/transaction_decode.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/transaction_decode.rs:96-111: P2PKH sender heuristic doesn't verify scriptSig shape, only the last push's length
  `input_address_from_script_sig` iterates every instruction in the scriptSig and takes whichever data push occurred last, then treats it as a pubkey if it is 33 or 65 bytes. It never checks that the scriptSig actually has the P2PKH shape (exactly two pushes: signature then pubkey). For a P2SH spend the last push is the redeem script; if that redeem script happens to be 33 or 65 bytes and parses as a valid secp256k1 point (spender-controllable, only constrained by hashing to the P2SH address), the function silently returns a P2PKH address unrelated to the actual signer. The module doc claims `None` for non-P2PKH, but the implementation doesn't uphold that.

  The simplest tightening is to require exactly two pushes (`sig`, `pubkey`) — anything else, including multisig P2SH redeem scripts, returns `None`. An additional guard that the first push looks like a DER-encoded ECDSA signature (starts with `0x30`, length ≤ 73) would eliminate essentially all remaining collision risk. This is not blocking for the stated CrowdNode use case (which matches on outputs, not input senders), but the failure mode — a wrong non-null address is worse than `None` — is worth closing since callers building sender-verification logic on top of `input.address` would be fooled.
- [SUGGESTION] packages/rs-platform-wallet/src/transaction_decode.rs:76-90: Unnecessary clone of every output's scriptPubkey
  `decode_transaction` iterates `tx.output` by reference and clones every `script_pubkey` to populate `DecodedOutput::script_pubkey`. Since `tx.txid()` only needs `&self` and doesn't care about the order relative to consuming other fields, the txid can be computed first, after which `tx.output` (and `tx.input`) can be consumed via `into_iter()` — eliminating the per-output `ScriptBuf` clone entirely. For transactions with large scripts (OP_RETURN payloads, multisig redeem scripts) this is a real, avoidable allocation on every decode call, and this function is positioned as the primitive for scanning arbitrary persisted transactions.
- [SUGGESTION] packages/rs-platform-wallet/src/transaction_decode.rs:113-228: No test exercises a non-P2PKH scriptSig (the documented `None` path)
  The only "no address" input test is `coinbase_input_has_no_address`. There is no test where the scriptSig's last push is not a valid P2PKH-shaped pubkey — for example a P2SH spend with a redeem-script push, a 3+-push scriptSig, or a 33/65-byte push that fails `PublicKey::from_slice`. These are exactly the ambiguous cases the length + parse checks are meant to filter out, and they'd pin the intended P2PKH-only contract so a future loosening of the heuristic (or the fix suggested above) has regression coverage.

In `packages/rs-platform-wallet-ffi/src/tx_decode.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/tx_decode.rs:87-100: `*out_decoded` is not zeroed on failure paths
  On every error return (`check_ptr!(tx_bytes)`, the `tx_bytes_len == 0` branch, `unwrap_result_or_return!(decode_transaction(...))`), the function returns a non-Success `PlatformWalletFFIResult` without assigning to `*out_decoded`. The Swift wrapper is safe because it initializes `outDecoded = nil` and checks the result code first, but any other C/Swift caller that reuses an already-populated out pointer — or reads it before checking the result code — will observe stale data and may double-free by re-calling `platform_wallet_decoded_transaction_free` on the previous value. Convention for out-pointers is to null on entry: after `check_ptr!(out_decoded)`, do `*out_decoded = std::ptr::null_mut();`. Small change, removes a foot-gun for anyone consuming this from C.
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/tx_decode.rs:30-34: Document that `DecodedTxInputFFI::address` is unauthenticated and must not be used for authorization
  The per-input `address` is derived from the last scriptSig push with no signature verification tying it to the UTXO being spent. A crafter can put any 33/65-byte value in that position, so the field is trivially spoofable. The pure module doc says "best-effort / sent from", but the FFI struct comment only says the value is "recovered from a P2PKH scriptSig", which a downstream consumer (dashwallet-ios, future CrowdNode integration) could reasonably read as "the sender is P" and use for routing/authorization decisions. Strengthen the comment on `DecodedTxInputFFI::address` (and the mirrored Swift `Input.address`) to say explicitly: derived from an unauthenticated script push, safe for display or matching hints only, not for authentication/authorization. The safe matching primitive for CrowdNode-style protocols is per-output `(address, amount)`, which is already exposed and not attacker-influenceable in the same way.
```
</details>

thepastaclaw:
Opened a focused helper PR for the CodeRabbit test feedback: https://github.com/dashpay/platform/pull/3993

It removes the tautological txid assertion and pins the expected `PlatformWalletError` cases for malformed and empty inputs. The two changed error-path tests pass individually; full `TransactionDecoderTests` still has unrelated decode happy-path failures with the freshly built FFI slice, noted in the helper PR body.

packages/rs-platform-wallet/src/transaction_decode.rs
```diff
@@ -0,0 +1,229 @@
...
+//!   of the scriptSig parsed as a public key — mirroring what wallet UIs
+//!   conventionally show as "sent from". `None` otherwise.
+
+use dashcore::blockdata::script::Instruction;
+use dashcore::consensus::deserialize;
+use dashcore::{Address, Network, OutPoint, PublicKey, ScriptBuf, Transaction, Txid};
+
+/// A consensus-decoded transaction with addresses rendered for a network.
+#[derive(Debug, Clone)]
+pub struct DecodedTransaction {
```
ZocoLini: Why do you need this step?? Isn't decoding bytes into a Transaction and then transforming it into the FFI structure more than enough?? I don't see a usage for DecodedTransaction/DecodedInput/DecodedOutput

packages/rs-platform-wallet/src/transaction_decode.rs
```diff
@@ -0,0 +1,229 @@
...
+    fn test_pubkey() -> PublicKey {
+        let secp = Secp256k1::new();
+        let sk = SecretKey::from_slice(&[0x42u8; 32]).expect("valid secret key");
+        PublicKey::new(sk.public_key(&secp))
+    }
+
+    /// One P2PKH input (spending 11..11:3), one P2PKH output of 151072 duffs,
+    /// one OP_RETURN output. The 151072 amount mirrors a CrowdNode signUp
+    /// signal so the fixture doubles as documentation.
+    fn synthetic_tx(network: Network) -> (Transaction, Address) {
```
ZocoLini: If you need Transaction for your tests there is a test_utils module in the dashcore crate, it may fit your needs, and if it doesn't let me now so I can consider updating the API

packages/rs-platform-wallet-ffi/src/tx_decode.rs
```diff
@@ -0,0 +1,288 @@
...
+use platform_wallet::decode_transaction;
+
+use crate::error::{PlatformWalletFFIResult, PlatformWalletFFIResultCode};
+use crate::types::{FFINetwork, Network};
+use crate::{check_ptr, unwrap_result_or_return};
+
+/// One decoded transaction input.
+#[repr(C)]
+pub struct DecodedTxInputFFI {
+    /// Previous output's txid in consensus (internal) byte order.
```
ZocoLini: Instead of creating new structs and mapping every field, have you consider an opaque pointer like:

```rust
#[repr(C)]
pub struct FFITransaction {
    inner: *mut c_void, // This is a pointer to the heap allocated transaction decoded
}
```

Then, instead of reading fields, we can call struct methods exposed though the FFI that are able to convert that inner pointer into the real transaction, and get the data you need. This may not be ideal, feel free to say no

ZocoLini: We can maybe update the places where the transaction bytes are being returned to return this FFI structure directly, avoiding having to decode it.

ZocoLini (changes requested):

packages/rs-platform-wallet/src/transaction_decode.rs
llbartekll: Agreed — done in 4eaa825736. `platform_wallet_decode_transaction` now deserializes straight into `dashcore::Transaction` and marshals that into the FFI structs; the `transaction_decode` module and its intermediate types are gone. The FFI struct layout and symbols are unchanged, so the generated header and the Swift wrapper API are unaffected.

packages/rs-platform-wallet/src/transaction_decode.rs
llbartekll: Adopted in 4eaa825736 — the tests now use `Address::dummy`, `Transaction::dummy`, and `Transaction::dummy_coinbase` (via a `dashcore = { workspace = true, features = ["test-utils"] }` dev-dependency). One gap: there's no builder for a P2PKH-shaped scriptSig (signature push + pubkey push) — `Transaction::dummy` fills `script_sig` with the lock script — so the fixture for the sender-address recovery path stays hand-built. If you'd like to add something like `Transaction::dummy_p2pkh_spend` to test_utils I'd happily consume it, but it's not blocking.

packages/rs-platform-wallet-ffi/src/tx_decode.rs
llbartekll: Good idea — that would save those consumers the decode round-trip. I'd keep it out of this PR though, since it changes the signatures of existing byte-returning APIs and their consumers. Follow-up material?

llbartekll:
Review-bot suggestions applied in 4eaa825736 alongside the restructure: the P2PKH sender heuristic now requires exactly two pushes (DER-shaped signature + 33/65-byte pubkey, anything else → null) so redeem scripts can't masquerade as sender addresses; `*out_decoded` is nulled on entry (pinned by tests); `input.address` is documented as unauthenticated on both the FFI struct and the Swift mirror; the per-output scriptPubkey clone and the per-push reallocation are gone (the decoded tx is consumed); non-P2PKH / redeem-script-collision / non-DER-first-push tests pin the `None` path; and `Input.prevTxidDisplayHex` was added. No order-assertion test for the latter — the fixture's prev txid `[0x11; 32]` is reversal-invariant, so the assertion would be vacuous.

packages/rs-platform-wallet-ffi/src/tx_decode.rs
```diff
@@ -0,0 +1,566 @@
...
+#[no_mangle]
+pub unsafe extern "C" fn platform_wallet_decode_transaction(
+    tx_bytes: *const u8,
+    tx_bytes_len: usize,
+    network: FFINetwork,
+    out_decoded: *mut *mut DecodedTransactionFFI,
+) -> PlatformWalletFFIResult {
+    check_ptr!(tx_bytes);
+    check_ptr!(out_decoded);
+    *out_decoded = std::ptr::null_mut();
```
thepastaclaw: **🟡 Suggestion: `*out_decoded` not nulled on the tx_bytes-null path, contradicting documented safety contract**

The rustdoc on `platform_wallet_decode_transaction` (lines 132-134) promises that `out_decoded` "is nulled on entry", so callers can trust the out-slot is cleared before any failure path. In the current ordering, `check_ptr!(tx_bytes)` (line 142) returns early via the macro before line 144 clears `*out_decoded`. A C/FFI caller that reuses an out-slot containing a previously-decoded (and possibly already-freed) `DecodedTransactionFFI*` and then invokes this function with `tx_bytes == NULL` receives `ErrorNullPointer` while the stale pointer remains visible. Contract-following cleanup code that calls `platform_wallet_decoded_transaction_free(*out)` on the assumption the slot was cleared then double-frees or dereferences freed memory. The Swift wrapper (`TransactionDecoder.decode`) zero-initializes its local `outDecoded`, so it is not exposed today; the risk lives on the public C ABI surface. The existing `rejects_null_and_empty_input` test never asserts the out-param state after the null-`tx_bytes` case, which is why the mismatch slipped through — extending the pre-poison pattern used by `trailing_garbage_is_a_deserialization_error` would pin the fix. The fix is one line: validate and clear `out_decoded` first, then validate `tx_bytes`.

```suggestion
    check_ptr!(out_decoded);
    *out_decoded = std::ptr::null_mut();
    check_ptr!(tx_bytes);
```

thepastaclaw (commented):
## Code Review

Cumulative review from `4eaa8257` to `225c1ae3`. The tx-decode FFI surface is unchanged since the prior review; the latest delta is a merge from `origin/v4.1-dev` containing unrelated work. The single carried-forward finding remains valid, and I found no new in-scope defects in the latest delta.

Source: reviewers = claude general `opus` (ok); codex general `gpt-5.5` (ok, parsed after rc=5); claude security-auditor `opus` (ok); codex security-auditor `gpt-5.5` (ok); claude ffi-engineer `opus` (ok); codex ffi-engineer `gpt-5.5` (ok); verifier = claude `opus` (ok); specialists = security-auditor, ffi-engineer; policy gate = review_policy.enforce_backport_prereq_policy.

🟡 1 suggestion(s)

**Prior Reconciliation**
- `prior-1` from `4eaa8257`: STILL VALID. At `225c1ae3`, `packages/rs-platform-wallet-ffi/src/tx_decode.rs:142-144` still checks `tx_bytes` before validating and clearing `out_decoded`, while the safety docs still promise `out_decoded` is nulled on entry.

**Carried-Forward Prior Findings**
- **[SUGGESTION]** `packages/rs-platform-wallet-ffi/src/tx_decode.rs:142-144`: `*out_decoded` not nulled on the `tx_bytes == NULL` path, contradicting the documented safety contract
  The rustdoc promises callers that the out-slot is cleared on entry, but `check_ptr!(tx_bytes)` returns before `*out_decoded = null_mut()` runs. A C/FFI caller that reuses an out-slot can therefore observe a stale pointer after `ErrorNullPointer`. The Swift wrapper initializes its local out pointer, so the current Swift path is not exposed; the risk is on the public C ABI surface. Validate and clear `out_decoded` before checking `tx_bytes`, and extend the pre-poison test pattern to cover the null-`tx_bytes` case.

**New Findings In Latest Delta**
- None. `git diff 4eaa8257..225c1ae3` is the upstream merge delta and does not touch the tx-decode FFI surface.

<details>

```

In `packages/rs-platform-wallet-ffi/src/tx_decode.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/tx_decode.rs:142-144: `*out_decoded` not nulled on the tx_bytes-null path, contradicting documented safety contract
  The rustdoc on `platform_wallet_decode_transaction` (lines 132-134) promises that `out_decoded` "is nulled on entry", so callers can trust the out-slot is cleared before any failure path. In the current ordering, `check_ptr!(tx_bytes)` returns early via the macro before line 144 clears `*out_decoded`. A C/FFI caller that reuses an out-slot containing a previously-decoded pointer and then invokes this function with `tx_bytes == NULL` receives `ErrorNullPointer` while the stale pointer remains visible. The fix is to validate and clear `out_decoded` first, then validate `tx_bytes`.
```
</details>

packages/rs-platform-wallet/src/transaction_decode.rs
thepastaclaw: Resolved in this update — _`last_push` reallocates on every push instead of holding a borrow_ no longer present.

packages/rs-platform-wallet/src/transaction_decode.rs
thepastaclaw: Resolved in this update — _No test exercises a non-P2PKH scriptSig (the documented `None` path)_ no longer present.

packages/rs-platform-wallet/src/transaction_decode.rs
thepastaclaw: Resolved in this update — _Unnecessary clone of every output's scriptPubkey_ no longer present.

packages/rs-platform-wallet/src/transaction_decode.rs
thepastaclaw: Resolved in this update — _P2PKH sender heuristic doesn't verify scriptSig shape, only the last push's length_ no longer present.

packages/rs-platform-wallet-ffi/src/tx_decode.rs
thepastaclaw: Correction: this broad out-param thread is not fully resolved in this update. The issue has been narrowed: most failure paths now clear `*out_decoded`, but the `tx_bytes == NULL` path still returns before the documented "nulled on entry" invariant is applied.

I carried the narrowed finding forward in the exact-SHA review for `225c1ae3`: https://github.com/dashpay/platform/pull/3981#pullrequestreview-4634224367

packages/rs-platform-wallet-ffi/src/tx_decode.rs
thepastaclaw: Resolved in this update — _Document that `DecodedTxInputFFI::address` is unauthenticated and must not be used for authorization_ no longer present.

packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/TransactionDecoder.swift
thepastaclaw: Resolved in this update — _`Input.prevTxid` lacks a display-hex helper analogous to `txidDisplayHex`_ no longer present.

packages/rs-platform-wallet-ffi/src/tx_decode.rs
llbartekll: Fixed in ffa8c9434e — `out_decoded` is now validated and nulled before `check_ptr!(tx_bytes)`, so the documented "nulled on entry" contract holds on every failure path. `rejects_null_and_empty_input` now pre-poisons the out param (same pattern as `trailing_garbage_is_a_deserialization_error`) and asserts it comes back null for both the null-`tx_bytes` and zero-length cases.

packages/rs-platform-wallet-ffi/src/tx_decode.rs
llbartekll: The remaining null-`tx_bytes` gap is closed in ffa8c9434e — the out-param is now nulled before any input validation, and the pre-poison test covers that path too.

thepastaclaw (approved):
## Code Review

Cumulative review from `225c1ae3` to `ffa8c943`. The latest delta directly fixes the prior FFI out-param contract issue: `platform_wallet_decode_transaction` now validates and clears `out_decoded` before checking `tx_bytes`, and `rejects_null_and_empty_input` pre-poisons the out slot and asserts it is nulled on both null-`tx_bytes` and zero-length failure paths. No carried-forward prior findings remain, and I found no new in-scope defects in the latest delta or current PR head.

Source: reviewers = claude general `opus` (ok); codex general `gpt-5.5` (ok, parsed after rc=5); claude security-auditor `opus` (ok); codex security-auditor `gpt-5.5` (ok); claude ffi-engineer `opus` (ok); codex ffi-engineer `gpt-5.5` (ok); verifier = claude `opus` (ok); specialists = security-auditor, ffi-engineer; policy gate = review_policy.enforce_backport_prereq_policy.

Prior Reconciliation:
- `prior-1` from `225c1ae3`: FIXED. At `ffa8c943`, `packages/rs-platform-wallet-ffi/src/tx_decode.rs:142-144` now runs `check_ptr!(out_decoded); *out_decoded = std::ptr::null_mut(); check_ptr!(tx_bytes);`, so the documented "nulled on entry" contract holds even when `tx_bytes == NULL`. The regression test now pre-poisons the out pointer and asserts it is nulled for the null and empty-input failure paths.

Carried-Forward Prior Findings:
- None.

New Findings In Latest Delta:
- None. `git diff 225c1ae3..ffa8c943` only reorders the FFI pointer checks and strengthens the null/empty-input test around the prior finding.

packages/rs-platform-wallet-ffi/src/tx_decode.rs
thepastaclaw: Resolved in `ffa8c94` — _`*out_decoded` not nulled on the tx_bytes-null path, contradicting documented safety contract_ no longer present.

packages/rs-platform-wallet-ffi/src/tx_decode.rs
ZocoLini: Let me study the impact of changing the return value to the structure in this PR (Our only ffi consumer is you, I don't see how that is an issue). I will also check on merging the DecodeTransactionFFI with the new FFICoreTransaction, is redundant to have both in the codebase

thepastaclaw (approved):
## Code Review

**Prior Reconciliation**: None. **Carried-Forward Prior Findings**: None (prior review at ffa8c943 had zero active findings). **New Findings In Latest Delta**: None — the only change since ffa8c943 is a test-fixture hex correction in TransactionDecoderTests.swift, which all six reviewer lanes and my own independent verification confirm produces a correctly-parsing 37-byte transaction matching the asserted address and txid. **Additional Cumulative Findings**: None — tx_decode.rs (packages/rs-platform-wallet-ffi/src/tx_decode.rs) is a clean, self-contained, stateless FFI entry point with consistent alloc/free pairing, null-safe out-parameter handling, a conservative P2PKH scriptSig heuristic, and thorough test coverage (13 tests). TransactionDecoder.swift is pure marshalling with no business logic, consistent with the swift-sdk architectural rule. All six independent reviewer lanes (3 Claude, 3 Codex) converged on zero findings; I re-read every changed file at the current head and confirm this assessment.

Source (experiment sonnet-primary-opus-quarter-sample-20260710, cohort sonnet_primary, bucket 3): reviewers codex/general=gpt-5.6-sol(completed_parseable_output_with_nonzero_wrapper_exit); sonnet5/general=claude-sonnet-5(completed); codex/security-auditor=gpt-5.6-sol(completed); sonnet5/security-auditor=claude-sonnet-5(completed_parseable_output_with_nonzero_wrapper_exit); codex/ffi-engineer=gpt-5.6-sol(completed); sonnet5/ffi-engineer=claude-sonnet-5(completed); verifier=verifier-sonnet5-3981-1783813537=claude-sonnet-5; orchestrator=openai/gpt-5.6-sol reasoning=high (orchestration-only, not a reviewer/verifier).

thepastaclaw (approved):
## Code Review

**Prior Reconciliation**: None (prior review at 1c180926 had zero active findings). **Carried-Forward Prior Findings**: None (prior review at 1c180926 had zero active findings). **New Findings In Latest Delta**: None — I cloned upstream rust-dashcore at the pinned rev d4692bb2 and diffed key-wallet-ffi/src/tx_decode.rs field-by-field against the deleted rs-platform-wallet-ffi copy: the `transaction_decode(tx_bytes, tx_bytes_len, network, out_decoded, error) -> bool` signature, `DecodedTransactionFFI`/`DecodedTxInputFFI`/`DecodedTxOutputFFI` layouts, `vec_to_ptr`/`decoded_transaction_free` ownership, and null-check/error-set macros are unchanged in shape — only the symbol source moved upstream. The Swift wrapper's `KeyWalletError(ffiError:)` copies the message string before the deferred `error_message_free` runs, so there's no use-after-free. **Additional Cumulative Findings**: None — I confirmed `rs-platform-wallet-ffi/src/tx_decode.rs` did not exist at the merge-base (807ba0e0), so the cumulative diff correctly nets out to just the new Swift decoder, its tests, and the Cargo pin bump; both agent lanes' 'no findings' conclusions are corroborated against ground truth.

Source (experiment sonnet-primary-opus-quarter-sample-20260710, cohort sonnet_primary, bucket 1): reviewers codex/general=gpt-5.6-sol(completed_parseable_output_with_nonzero_wrapper_exit); sonnet5/general=claude-sonnet-5(completed_parseable_output_with_nonzero_wrapper_exit); codex/ffi-engineer=gpt-5.6-sol(completed_parseable_output_with_nonzero_wrapper_exit); sonnet5/ffi-engineer=claude-sonnet-5(completed); verifier=verifier-sonnet5-3981-1783844991=claude-sonnet-5; orchestrator=openai/gpt-5.6-sol reasoning=high (orchestration-only, not a reviewer/verifier).

QuantumExplorer (approved):

PR #3976, opened 2026-07-01
chore: bump rust-dashcore to afcff156, export xpub via ExtendedPubKeySigner

Claudius-Maginificent:
## Why this PR exists
- **Problem**: rust-dashcore's `key-wallet` gained a way to derive a BIP-32 extended public key (xpub) from a signer without ever exposing the private key, and the workspace needed to track rust-dashcore's `dev` HEAD.
- **What breaks without it**: without the bump, the workspace is pinned behind rust-dashcore `dev`; without the xpub export, `MnemonicResolverCoreSigner` callers needing offline non-hardened descendant derivation (e.g. DashPay contact-payment addresses) have no safe path except deriving the full `ExtendedPrivKey` and extracting the public half by hand, risking unwiped private-key material on the stack.
- **Blocking relationship**: targets `v4.1-dev`, the current base for wallet work.

## What was done
- Bumped all 8 rust-dashcore workspace deps (dashcore, dash-spv, key-wallet, key-wallet-ffi, key-wallet-manager, dash-network, dash-network-seeds, dashcore-rpc) to `dev` HEAD `afcff1566c81abb0a9cf64b99b0ee675169fddfc`, regenerating `Cargo.lock` so every rust-dashcore-sourced crate (including transitive `dashcore_hashes`, `dashcore-private`, `dashcore-rpc-json`, `git-state`) resolves to the new rev.
- Upstream refactored the signing API between the previous pin and this one: `extended_public_key` moved out of the base `Signer` trait into a new opt-in `ExtendedPubKeySigner: Signer` trait. Implemented `ExtendedPubKeySigner::extended_public_key` on `MnemonicResolverCoreSigner` accordingly — full BIP-32 derivation via a shared `resolve_and_derive(extract: impl FnOnce(&ExtendedPrivKey) -> T)` helper reused by both `derive_priv` and `extended_public_key`, so the extended key never crosses a function boundary (only the extracted `T` — a scalar or `ExtendedPubKey` — does).
- Dropped the 3 `rs-dpp` test-signer stubs (`FixedKeySigner` in `state_transition/mod.rs` and its two `signing_tests.rs` call sites) added in earlier iterations: they only existed to satisfy the previously *required* `Signer::extended_public_key` method, which is no longer required now that it's opt-in — the stubs no longer compile and nothing needs them as `ExtendedPubKeySigner`.
[...]

PR #3985, opened 2026-07-02
fix(platform-wallet): release UTXO reservation when broadcast fails

Claudius-Maginificent:

`TransactionBuilder::build_signed` reserves the selected UTXOs in the funding account's `ReservationSet` and leaves the reservation held on success, expecting the broadcast to follow. When the broadcast **failed**, nothing released the reservation — the inputs stayed locked until the 24-block (~1h) TTL backstop reclaimed them. A user whose send failed and who immediately retried hit a spurious "insufficient funds".

This affected every build-then-broadcast path in the crate: `CoreWallet::send_to_addresses`, `send_dashpay_payment`, and `create_funded_asset_lock_proof` (identity funding).

Not a funds-loss bug (self-heals via TTL), but a real UX rough edge, reachable in practice via dash-evo-tool's `WalletBackend::send_payment` ("Send Dash" screen) and the SwiftExampleApp send flows.

Depends on the reservation system introduced by the rust-dashcore bump to `afcff156` in #3976 (merged).

## What was done?

- **Typed broadcast errors.** `TransactionBroadcaster::broadcast` now returns `BroadcastError` with two variants: `Rejected` (the transaction provably never reached the network — safe to release its inputs and retry) and `MaybeSent` (outcome unknown — the reservation must be kept for the TTL backstop, since releasing and retrying could double-spend). The classification is made **inside each broadcaster implementation**, where the transport knowledge lives, rather than by matching catch-all error variants at the wallet layer:
  - `DapiBroadcaster` classifies every failure `MaybeSent`: its internal multi-node retries (`RequestSettings::default()`) mean the surfaced error is only the last attempt's — an earlier attempt may have delivered the transaction (e.g. a node accepts the tx but its gRPC response times out).
  - `SpvRuntime` classifies "client not started" and dash-spv's zero-peer `NetworkError::NotConnected` as `Rejected` — both fire before any bytes leave the process — and everything else `MaybeSent` (a later failure may follow a partial peer send).
  - `MaybeSent` surfaces to callers as the new `PlatformWalletError::TransactionBroadcastUnconfirmed`, whose message tells the user not to immediately re-spend.
- **One shared release seam.** The release logic lives in `wallet::reservations::broadcast_releasing_on_rejection`, used by all three build-then-broadcast paths (`send_to_addresses`, `send_dashpay_payment`, `create_funded_asset_lock_proof`), so a rejected broadcast releases the reserved inputs immediately on every path instead of only core sends. The asset-lock **recovery** rebroadcast intentionally does not release — it re-drives a previously built transaction and is not paired with a fresh reserve. The release path is read-lock only (`release_reservation` takes `&self`), so it doesn't serialize concurrent sends, and the wallet-manager lock is never held across the broadcast `await`.
- Account lookup uses key-wallet's existing `bip44_managed_account_at_index` / `bip32_managed_account_at_index` accessors instead of hand-rolled map dispatch; a missing wallet/account on the release path is `tracing::warn!`-logged (with `wallet_id`) without altering the returned error.

## How Has This Been Tested?

- Table-driven `send_to_addresses_releases_reservation_on_broadcast_failure` covers both `BIP44Account` and `BIP32Account`: funds a wallet, fails the first broadcast via a `Rejected`-returning mock broadcaster, asserts an immediate retry succeeds for each account type.
- `send_to_addresses_keeps_reservation_on_ambiguous_broadcast_failure` (both account types) proves the conservative side: a `MaybeSent` broadcaster error keeps the reservation, and an immediate retry fails at UTXO selection rather than silently double-spending.
- `cargo fmt --all`
- `cargo clippy -p platform-wallet --all-targets -- -D warnings` — clean
- `cargo test -p platform-wallet` — 204 passed, 0 failed
- `cargo check --all-features --tests` for `platform-wallet`, plus `platform-wallet-ffi`, `platform-wallet-storage`, and `rs-unified-sdk-ffi` (all dependents of the changed trait)

## Breaking Changes

None (not consensus-affecting). `TransactionBroadcaster::broadcast`'s error type changed from `PlatformWalletError` to `BroadcastError` — all in-workspace implementations and callers are updated in this PR; a `From<BroadcastError> for PlatformWalletError` impl preserves the existing `?` ergonomics.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved broadcast outcome handling: definitive rejections now trigger reservation cleanup to enable safe retries.
  * Ambiguous outcomes keep reservations and return a “broadcast unconfirmed” error to prevent unsafe auto-retries.
  * Updated payment and asset-lock broadcast flows to follow the reservation-aware behavior.
* **New Features**
  * Added a dedicated “broadcast unconfirmed” typed error/result code across platform interfaces (including FFI/Swift).
* **Tests**
  * Added retry-behavior coverage for both BIP44 and BIP32 scenarios (reject-first vs maybe-sent).
* **Documentation**
  * Clarified when retries are safe versus when the outcome may have reached the network.


thepastaclaw:
✅ Review complete (commit 1a9b6c0e)

lklimek (dismissed):

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
```diff
@@ -142,7 +165,185 @@ impl<B: TransactionBroadcaster + ?Sized> CoreWallet<B> {
...
+                        "could not release UTXO reservation after failed broadcast: \
+                         funds account not found"
+                    ),
+                },
+                None => tracing::warn!(
+                    "could not release UTXO reservation after failed broadcast: wallet not found"
+                ),
+            }
+            return Err(e);
+        }
```
thepastaclaw: **🟡 Suggestion: Failure-path release takes an exclusive write lock it doesn't need**

`ManagedCoreFundsAccount::release_reservation` takes `&self` (the `ReservationSet` uses an internal `Arc<Mutex<_>>`), so no mutable borrow of the account is required. The current code takes `self.wallet_manager.write().await` and looks the account up via `get_wallet_and_info_mut` + `get_mut` purely to satisfy `funds_account_mut`'s `&mut` signature. `WalletManager` stores all wallets in shared maps, so this write lock serializes against every other in-flight `send_to_addresses` / balance-read across every wallet managed by this instance for the duration of the cleanup — on the exact path meant to make retries fast. Switch this branch to `wallet_manager.read().await` + immutable `get_wallet_and_info` + `BTreeMap::get`, and either change `funds_account_mut` to return `&ManagedCoreFundsAccount` or introduce an immutable variant. Not a bug today (the section is short), but it's an avoidable point of contention on the exact code path meant to unblock retries.

<sub>source: ['claude']</sub>

thepastaclaw: **🟡 Suggestion: Releasing on every broadcaster Err assumes the tx never reached the network**

Any `Err` from `broadcast_transaction` is treated as "tx definitely did not reach the network", which triggers an immediate release so a retry can reselect the same UTXOs. In practice a broadcaster can fail *after* the network has accepted the transaction — e.g. a dropped SPV/gRPC connection after send, a timeout waiting for ack, or a node error that still relayed the tx. In those cases the release + retry produces a second signed spend over the same inputs, turning a transient network hiccup into a self-inflicted double-spend attempt (the second broadcast usually then fails on the node, but the UX is confusing). This is defensible as a UX call — the pre-PR TTL-backstop behavior had the opposite failure mode — but it should be a deliberate contract. Either narrow the release to error variants that are unambiguously pre-send (transport/serialization/local rejection), or add a comment on `TransactionBroadcaster::broadcast` documenting that `Err` MUST mean "the network did not accept this transaction" so other broadcaster impls are held to the same contract.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
```diff
@@ -142,7 +165,185 @@ impl<B: TransactionBroadcaster + ?Sized> CoreWallet<B> {
...
+        // this fails with a build error instead.
+        let second = core
+            .send_to_addresses(StandardAccountType::BIP44Account, 0, outputs, &signer)
+            .await;
+        assert!(
+            second.is_ok(),
+            "retry after a failed broadcast should succeed once the reservation \
+             is released, got {second:?}"
+        );
+    }
```
thepastaclaw, line 265: **💬 Nitpick: Regression test only covers the BIP44 arm of `funds_account_mut`**

The test exercises `StandardAccountType::BIP44Account` only. The helper and the release path also have a `BIP32Account` arm that the fix depends on to resolve the same account the builder reserved on. The two arms are structurally identical today, but a table-driven parameterization over `account_type` (or a second minimal case) would prevent silent divergence if a new `StandardAccountType` variant is added without updating the helper.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
```diff
@@ -142,7 +165,185 @@ impl<B: TransactionBroadcaster + ?Sized> CoreWallet<B> {
...
+                    None => tracing::warn!(
+                        ?account_type,
+                        account_index,
+                        "could not release UTXO reservation after failed broadcast: \
+                         funds account not found"
+                    ),
+                },
+                None => tracing::warn!(
+                    "could not release UTXO reservation after failed broadcast: wallet not found"
+                ),
```
thepastaclaw: **💬 Nitpick: Warn logs on release-path fallback lack `wallet_id` for correlation**

If the wallet or funds account has been removed between build and the failed broadcast, the release path logs `tracing::warn!` and returns the original broadcast error. The reservation is dropped along with the account, so state is fine, but neither warn arm includes `wallet_id`, which makes it hard to correlate an operator-visible warning with a specific send. Add `wallet_id=%self.wallet_id` to both arms, and for the outer arm also `?account_type, account_index`, so orphaned-reservation warnings are greppable.

<sub>source: ['claude']</sub>

thepastaclaw (approved):
## Code Review

Small, well-scoped fix that releases the transaction builder's UTXO reservation when broadcast fails, preventing a spurious ~1h stranding of coins after a failed send. No blocking issues; a few suggestions and nitpicks around release-path locking, error semantics, and test coverage.

Source: reviewers opus general/rust-quality; claude-sonnet-5 general/rust-quality (ACP wrappers terminated after parseable output); gpt-5.5[high] general/rust-quality failed at initialize; verifier opus.

<details>

```

In `packages/rs-platform-wallet/src/wallet/core/broadcast.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/core/broadcast.rs:168-189: Failure-path release takes an exclusive write lock it doesn't need
  `ManagedCoreFundsAccount::release_reservation` takes `&self` (the `ReservationSet` uses an internal `Arc<Mutex<_>>`), so no mutable borrow of the account is required. The current code takes `self.wallet_manager.write().await` and looks the account up via `get_wallet_and_info_mut` + `get_mut` purely to satisfy `funds_account_mut`'s `&mut` signature. `WalletManager` stores all wallets in shared maps, so this write lock serializes against every other in-flight `send_to_addresses` / balance-read across every wallet managed by this instance for the duration of the cleanup — on the exact path meant to make retries fast. Switch this branch to `wallet_manager.read().await` + immutable `get_wallet_and_info` + `BTreeMap::get`, and either change `funds_account_mut` to return `&ManagedCoreFundsAccount` or introduce an immutable variant. Not a bug today (the section is short), but it's an avoidable point of contention on the exact code path meant to unblock retries.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/core/broadcast.rs:168-189: Releasing on every broadcaster Err assumes the tx never reached the network
  Any `Err` from `broadcast_transaction` is treated as "tx definitely did not reach the network", which triggers an immediate release so a retry can reselect the same UTXOs. In practice a broadcaster can fail *after* the network has accepted the transaction — e.g. a dropped SPV/gRPC connection after send, a timeout waiting for ack, or a node error that still relayed the tx. In those cases the release + retry produces a second signed spend over the same inputs, turning a transient network hiccup into a self-inflicted double-spend attempt (the second broadcast usually then fails on the node, but the UX is confusing). This is defensible as a UX call — the pre-PR TTL-backstop behavior had the opposite failure mode — but it should be a deliberate contract. Either narrow the release to error variants that are unambiguously pre-send (transport/serialization/local rejection), or add a comment on `TransactionBroadcaster::broadcast` documenting that `Err` MUST mean "the network did not accept this transaction" so other broadcaster impls are held to the same contract.
- [NITPICK] packages/rs-platform-wallet/src/wallet/core/broadcast.rs:287-348: Regression test only covers the BIP44 arm of `funds_account_mut`
  The test exercises `StandardAccountType::BIP44Account` only. The helper and the release path also have a `BIP32Account` arm that the fix depends on to resolve the same account the builder reserved on. The two arms are structurally identical today, but a table-driven parameterization over `account_type` (or a second minimal case) would prevent silent divergence if a new `StandardAccountType` variant is added without updating the helper.
- [NITPICK] packages/rs-platform-wallet/src/wallet/core/broadcast.rs:177-186: Warn logs on release-path fallback lack `wallet_id` for correlation
  If the wallet or funds account has been removed between build and the failed broadcast, the release path logs `tracing::warn!` and returns the original broadcast error. The reservation is dropped along with the account, so state is fine, but neither warn arm includes `wallet_id`, which makes it hard to correlate an operator-visible warning with a specific send. Add `wallet_id=%self.wallet_id` to both arms, and for the outer arm also `?account_type, account_index`, so orphaned-reservation warnings are greppable.

```
</details>

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
Claudius-Maginificent: Agreed — verified `release_reservation` takes `&self` (rust-dashcore `key-wallet/src/managed_account/managed_core_funds_account.rs:129`) and `WalletManager::get_wallet_and_info` (immutable) already exists (`key-wallet-manager/src/accessors.rs:29`). Switching this branch to `read().await` + an immutable `funds_account` lookup is a valid follow-up to avoid contending the write lock on the cleanup path. Not fixed in this PR yet — tracked as a follow-up.

Claudius-Maginificent: Fair point — `broadcast_transaction` currently has no documented contract on what `Err` means (just delegates to `self.broadcaster.broadcast`). Agree this is a real gap: a post-send transport failure would trigger release+retry against an already-accepted tx. Leaving the current UX trade-off as-is for this PR, but the `TransactionBroadcaster::broadcast` contract should get a doc comment stating `Err` MUST mean "not accepted by the network" — tracked as a follow-up, not fixed here.

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
Claudius-Maginificent: Confirmed — the only regression test (`send_to_addresses_releases_reservation_on_broadcast_failure`) covers `BIP44Account` only; the `BIP32Account` arm in `funds_account_mut` is untested. Valid gap, not fixed in this PR yet — tracked as a follow-up.

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
Claudius-Maginificent: Confirmed — both `tracing::warn!` calls on the release-failure fallback paths omit `wallet_id` (and the outer one omits `account_type`/`account_index` too). Valid correlation gap, not fixed in this PR yet — tracked as a follow-up.

thepastaclaw (commented):
## Code Review

Incremental review of eb54b370 vs prior f9c49ed1. All four prior findings are FIXED at HEAD: release path now takes a read lock via immutable `funds_account` helper; release is gated by `broadcast_failure_is_pre_send` with the documented `TransactionBroadcaster::broadcast` `Err`-contract; regression tests iterate BIP44/BIP32 via `funded_core_wallet` and add an ambiguous-failure keep-reservation counterpart; both warn arms include `wallet_id`. One new in-scope suggestion: the documented `Err` = 'not accepted by the network' contract that `broadcast_failure_is_pre_send` relies on is not actually upheld by `DapiBroadcaster` (the default non-SPV path), because it flattens all gRPC errors — including `DeadlineExceeded`/`Unavailable` after Core may have accepted the tx — into `TransactionBroadcast`, which the predicate treats as safe to release. Verified against dash-spv afcff15: `SpvError` genuinely does mean pre-send (peer success counter). No blocking issues.

🟡 1 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet/src/broadcaster.rs:41-67: `DapiBroadcaster` does not uphold the new `Err` = not-accepted contract, narrowing but not closing the double-spend-on-retry window
  The new `broadcast_failure_is_pre_send` (broadcast.rs:44-49) classifies `PlatformWalletError::TransactionBroadcast(_)` as unambiguously pre-send and safe to release-and-retry, backed only by the doc comment on `TransactionBroadcaster::broadcast` (broadcaster.rs:23-24). `SpvBroadcaster` genuinely honors that contract — dash-spv (pinned rev `afcff15`) counts per-peer send results and only surfaces `Err` when zero peers received the tx. `DapiBroadcaster::broadcast` (broadcaster.rs:41-67) does not: it wraps *any* `DapiClientError` returned by `self.sdk.execute(...).into_inner()` into `TransactionBroadcast(String)`, with no discrimination by cause. The underlying `DapiClientError` includes `Transport(TransportError::Grpc(tonic::Status))` covering codes like `DeadlineExceeded` and `Unavailable` — the classic 'client gave up waiting for an ack, but the server-side call may have already put the tx in Core's mempool' case. A user hitting a gRPC deadline/connection drop after DAPI already relayed the tx will see `TransactionBroadcast`, get classified as pre-send, have the reservation released, and immediately retry — signing a second tx over the same inputs while the first is propagating. That is exactly the release-and-retry double-spend the PR's granular condition is meant to prevent, narrowed but not closed on the default broadcast path. Options: (a) in `DapiBroadcaster::broadcast`, inspect the `DapiClientError` and only return `TransactionBroadcast` for definite pre-send rejections (e.g. `Server(_)` decode/reject responses), routing ambiguous transport failures into a distinct 'possibly accepted' variant that `broadcast_failure_is_pre_send` treats as keep-reservation; or (b) drop `TransactionBroadcast` from the pre-send allowlist and rely on the TTL backstop for DAPI failures until (a) lands. Contract-by-doc-comment alone doesn't make it safe.
```
</details>

Source: reviewers: `opus` (general), `claude-sonnet-5` (general), `gpt-5.5[high]` (general, failed_or_unparseable), `opus` (rust-quality), `claude-sonnet-5` (rust-quality), `gpt-5.5[high]` (rust-quality, failed_or_unparseable), `opus` (ffi-engineer), `claude-sonnet-5` (ffi-engineer), `gpt-5.5[high]` (ffi-engineer, failed_or_unparseable); verifier: `opus`; specialists: `rust-quality`, `ffi-engineer`

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
thepastaclaw: Resolved in `eb54b37` — _Releasing on every broadcaster Err assumes the tx never reached the network_ no longer present.

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
thepastaclaw: Resolved in `eb54b37` — _Warn logs on release-path fallback lack `wallet_id` for correlation_ no longer present.

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
thepastaclaw: Resolved in `eb54b37` — _Regression test only covers the BIP44 arm of `funds_account_mut`_ no longer present.

packages/rs-platform-wallet/src/wallet/core/broadcast.rs
thepastaclaw: Resolved in `eb54b37` — _Failure-path release takes an exclusive write lock it doesn't need_ no longer present.

thepastaclaw (commented):
## Code Review

HEAD c81f9d2c is a no-op merge of v4.1-dev into the fix branch — `git diff eb54b37001..c81f9d2c81 -- packages/rs-platform-wallet/` is empty. No new delta findings. The one prior in-scope finding (DapiBroadcaster does not uphold the `Err`=not-accepted contract that `broadcast_failure_is_pre_send` depends on) remains STILL_VALID at HEAD and is carried forward as a suggestion. All 6 agents converged on this single finding.

🟡 1 suggestion(s)


<details>

```

- [SUGGESTION] packages/rs-platform-wallet/src/broadcaster.rs:41-67: `DapiBroadcaster` does not uphold the new `Err` = not-accepted contract, narrowing but not closing the double-spend-on-retry window
  The new `broadcast_failure_is_pre_send` (packages/rs-platform-wallet/src/wallet/core/broadcast.rs:44-49) classifies `PlatformWalletError::TransactionBroadcast(_)` as unambiguously pre-send and therefore safe to release the UTXO reservation and let the caller retry. That predicate is backed only by the doc comment added on `TransactionBroadcaster::broadcast` (broadcaster.rs:23-24): "Err must mean the transaction was not accepted by the network".

  `SpvBroadcaster` genuinely upholds that contract — dash-spv (pinned rev afcff156) counts per-peer send results and only surfaces `Err` when zero peers received the tx. `DapiBroadcaster::broadcast` does not: at broadcaster.rs:56-63 it wraps *any* error returned by `self.sdk.execute(request, RequestSettings::default()).await.into_inner()` into `PlatformWalletError::TransactionBroadcast(String)` with no discrimination by cause. The underlying `DapiClientError` includes `Transport(TransportError::Grpc(tonic::Status))`, which covers codes such as `DeadlineExceeded`, `Unavailable`, `Cancelled`, and mid-response resets — the classic "client gave up waiting for an ack, but the server-side call may have already put the tx in Core's mempool" case.

  A user hitting a gRPC deadline or connection drop after DAPI already relayed the tx will see `TransactionBroadcast`, be classified pre-send by the predicate, have the reservation released, and immediately retry — signing a second tx over the same inputs while the first may already be propagating. That is exactly the release-and-retry double-spend the granular condition is meant to prevent, narrowed but not closed on the default (non-SPV) broadcast path. On the security side, the input-conflict property means at most one tx confirms (no funds theft), but a hostile or degraded DAPI relay can deliberately drop the response after forwarding to Core to coerce the wallet into duplicate broadcasts (mempool spam using the user's key, retry-policy fingerprinting).

  The contract is also enforced only by convention (a doc comment), not the type system, so any future `TransactionBroadcaster` impl that raises `TransactionBroadcast` for an ambiguous submission will silently violate it.

  Options: (a) in `DapiBroadcaster::broadcast`, match on `DapiClientError` and only return `TransactionBroadcast` for definite pre-send rejections (e.g. `Server(_)` decode/reject responses, connection-refused before the request was written), routing ambiguous transport failures (`Transport(Grpc(_))` with `DeadlineExceeded`/`Unavailable`/`Cancelled`/`Unknown`, or any post-write-side error) into a distinct "possibly accepted" `PlatformWalletError` variant that `broadcast_failure_is_pre_send` treats as keep-reservation; or (b) drop `TransactionBroadcast` from the pre-send allowlist and rely on the TTL backstop for DAPI failures until (a) lands — SPV keeps its fast-path.
```
</details>

Source: reviewers: `opus` (general), `claude-sonnet-5` (general), `gpt-5.5[high]` (general, failed_or_unparseable), `opus` (security-auditor), `claude-sonnet-5` (security-auditor), `gpt-5.5[high]` (security-auditor, failed_or_unparseable), `opus` (rust-quality), `claude-sonnet-5` (rust-quality), `gpt-5.5[high]` (rust-quality, failed_or_unparseable); verifier: `opus`; specialists: `security-auditor`, `rust-quality`

thepastaclaw:
Handled the CodeRabbit follow-up in helper PR #3996: https://github.com/dashpay/platform/pull/3996

Current status: #3996 is docs-only per maintainer request. It clarifies the broadcaster error contract in `packages/rs-platform-wallet/src/broadcaster.rs`; the wallet broadcast code/test changes were removed, so this helper no longer extends the retry test coverage.

Latest checks are green: PR title, action-semantic-pull-request, and CodeRabbit manual review. Local validation before the docs-only push included `git diff --check`.

thepastaclaw (commented):
## Code Review

Prior finding prior-1 is resolved at HEAD 314540296d: the doc-comment-only pre-send contract in `DapiBroadcaster` is replaced by a typed `BroadcastError { Rejected, MaybeSent }` enum, `DapiBroadcaster` conservatively classifies every DAPI failure as `MaybeSent`, and the release logic flows through a shared `broadcast_releasing_on_rejection` seam consumed by all three build-then-broadcast callers. Two new latest-delta findings surface: (1) the new `PlatformWalletError::TransactionBroadcastUnconfirmed` variant flattens to `ErrorUnknown` at the FFI boundary because the analogous arms added for `ShieldedBroadcastUnconfirmed`/`ShieldedSpendUnconfirmed` were not added for the new variant; (2) on `BroadcastError::Rejected` the asset-lock path releases the funding UTXO reservation while leaving a persisted `Built` tracked lock that `resume_asset_lock` will re-broadcast if later invoked with that outpoint. Both are suggestions, not blockers — the DAPI path structurally never returns `Rejected` today, so the second is currently only reachable via the SPV path.

🟡 2 suggestion(s)

_The FFI-boundary suggestion is included body-only in the details below because `packages/rs-platform-wallet-ffi/src/error.rs` is outside the PR diff._

Source: reviewers opus/opus, sonnet5/claude-sonnet-5, codex/gpt-5.5 for general/security-auditor/rust-quality/ffi-engineer; verifier opus/opus.

<details>

```

- [SUGGESTION] packages/rs-platform-wallet-ffi/src/error.rs:231-240: New `TransactionBroadcastUnconfirmed` variant flattens to `ErrorUnknown` at the FFI boundary, erasing the PR's retry-safety signal
  This PR's entire point is a typed distinction between broadcast outcomes that are safe to retry immediately (`BroadcastError::Rejected` → `PlatformWalletError::TransactionBroadcast`) and outcomes that must NOT be retried because the transaction may already be propagating (`BroadcastError::MaybeSent` → the new `PlatformWalletError::TransactionBroadcastUnconfirmed`, per broadcaster.rs:45-56 and error.rs:60-77). The distinction is enforced by the type system in Rust, but is lost crossing the C ABI.

  `From<PlatformWalletError> for PlatformWalletFFIResult` at rs-platform-wallet-ffi/src/error.rs:205-244 has dedicated arms for the structurally identical shielded siblings — `ShieldedBroadcastFailed`, `ShieldedBroadcastUnconfirmed`, `ShieldedSpendUnconfirmed` — precisely because the in-code comment (lines 221-230) documents that flattening these to `ErrorUnknown` would 'defeat the slot-holding contract.' The new `TransactionBroadcastUnconfirmed` variant carries the same slot-holding contract (its own doc at error.rs:63-77 names `ShieldedSpendUnconfirmed` as its direct sibling) but has no arm here, so every ambiguous broadcast on the Core-wallet, DashPay, and asset-lock send paths crosses the boundary as generic code 99. Since every DAPI broadcast failure now maps to `MaybeSent` unconditionally, this is the primary error class the DAPI-backed send paths (`core_wallet_send_to_addresses`, `platform_wallet_send_dashpay_payment`, `asset_lock_manager_create_funded_proof`) will emit on failure.

  Swift's `PlatformWalletError.init(result:)` is a pure code switch with no message-string fallback, and `SendViewModel` already special-cases `shieldedSpendUnconfirmed` as non-retryable — that special-casing cannot be replicated for the Core-wallet path until the FFI code exists. The safety property is still preserved by the TTL-backstopped reservation, but the UX degrades from a specific 'tx may already be on network — do not retry' to a generic error, followed by a misleading `NoSpendableInputs` on retry.

  Add a dedicated `ErrorTransactionBroadcastUnconfirmed` variant to `PlatformWalletFFIResultCode`, add the matching arm here mirroring the shielded siblings, and add the corresponding case to Swift's `PlatformWalletError`. The PR diff is currently confined to `packages/rs-platform-wallet/`, but this variant is only useful once foreign callers can act on it.

In `packages/rs-platform-wallet/src/wallet/asset_lock/build.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/asset_lock/build.rs:329-361: Rejected asset-lock broadcast releases the funding reservation while leaving a resumable `Built` lock behind
  `create_funded_asset_lock_proof` tracks and persists the transaction as `AssetLockStatus::Built` (lines 329-341) before calling `broadcast_releasing_on_rejection`. On `BroadcastError::Rejected`, that helper releases the BIP44 funding reservation and returns via `?` without advancing or removing the tracked lock, leaving the persisted `Built` row live in `tracked_asset_locks`. The inline comment at lines 348-352 documents this as intentional ('the tracked lock stays `Built`, so a later resume can still re-drive the same transaction while its inputs remain unspent') — but with the reservation now released, the 'inputs remain unspent' precondition can no longer be relied on. If the caller retries `create_funded_asset_lock_proof` successfully with a fresh transaction spending the same UTXOs, `resume_asset_lock(&old_out_point)` (recovery.rs:251-261, driven by `FundingResolution` orchestration or a `list_tracked_locks`-based recovery UI) will re-broadcast the stale `Built` transaction whose inputs have already been consumed.

  The worst-case fallout is not a double-spend — the second broadcast will fail mempool acceptance because the inputs are already spent — but it does re-broadcast a superseded transaction, produce confusing failures, and leave clutter in the tracked-lock persistence. This is currently only reachable via the SPV path (`DapiBroadcaster` never returns `Rejected`; only `SpvRuntime::broadcast_transaction`'s zero-peer / client-not-started arms do).

  On `BroadcastError::Rejected`, either remove/tombstone the tracked lock (so it isn't offered up by `list_tracked_locks` for resume), or keep the reservation held for the `Built` row and only release it once the lock is definitively abandoned. Either shape restores the invariant the doc comment relies on.
```
</details>

packages/rs-platform-wallet/src/wallet/asset_lock/build.rs
```diff
@@ -345,8 +345,20 @@ impl<B: TransactionBroadcaster + ?Sized> AssetLockManager<B> {
...
+        //    the same transaction while its inputs remain unspent.
+        crate::wallet::reservations::broadcast_releasing_on_rejection(
+            self.broadcaster.as_ref(),
+            &self.wallet_manager,
+            &self.wallet_id,
+            key_wallet::account::account_type::StandardAccountType::BIP44Account,
+            account_index,
+            &tx,
+        )
+        .await?;
```
thepastaclaw: **🟡 Suggestion: Rejected asset-lock broadcast releases the funding reservation while leaving a resumable `Built` lock behind**

`create_funded_asset_lock_proof` tracks and persists the transaction as `AssetLockStatus::Built` (lines 329-341) before calling `broadcast_releasing_on_rejection`. On `BroadcastError::Rejected`, that helper releases the BIP44 funding reservation and returns via `?` without advancing or removing the tracked lock, leaving the persisted `Built` row live in `tracked_asset_locks`. The inline comment at lines 348-352 documents this as intentional ('the tracked lock stays `Built`, so a later resume can still re-drive the same transaction while its inputs remain unspent') — but with the reservation now released, the 'inputs remain unspent' precondition can no longer be relied on. If the caller retries `create_funded_asset_lock_proof` successfully with a fresh transaction spending the same UTXOs, `resume_asset_lock(&old_out_point)` (recovery.rs:251-261, driven by `FundingResolution` orchestration or a `list_tracked_locks`-based recovery UI) will re-broadcast the stale `Built` transaction whose inputs have already been consumed.

The worst-case fallout is not a double-spend — the second broadcast will fail mempool acceptance because the inputs are already spent — but it does re-broadcast a superseded transaction, produce confusing failures, and leave clutter in the tracked-lock persistence. This is currently only reachable via the SPV path (`DapiBroadcaster` never returns `Rejected`; only `SpvRuntime::broadcast_transaction`'s zero-peer / client-not-started arms do).

On `BroadcastError::Rejected`, either remove/tombstone the tracked lock (so it isn't offered up by `list_tracked_locks` for resume), or keep the reservation held for the `Built` row and only release it once the lock is definitively abandoned. Either shape restores the invariant the doc comment relies on.

<sub>source: ['codex']</sub>

QuantumExplorer: Fixed in 12ab17f373 — on `BroadcastError::Rejected` the `Built` lock is now untracked (in-memory + persisted row via the changeset's `removed` set, applied by `apply.rs`), restoring the invariant: a resumable row only exists while its reservation is held. `MaybeSent` keeps both.

packages/rs-platform-wallet/src/broadcaster.rs
```diff
@@ -20,6 +20,8 @@ use crate::spv::SpvRuntime;
 /// Implementations may use DAPI (gRPC), SPV (P2P peers), or Core RPC.
 #[async_trait]
 pub trait TransactionBroadcaster: Send + Sync {
+    /// Contract: `Err` must mean the transaction was **not** accepted by the
+    /// network, so callers may safely release its reserved inputs and retry.
```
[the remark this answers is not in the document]
QuantumExplorer: Addressed in 3145402 — the doc-comment contract was replaced by a typed `BroadcastError { Rejected, MaybeSent }` returned by `TransactionBroadcaster::broadcast`, so implementations must classify explicitly and callers can no longer collapse post-submit uncertainty into a safe-retry error.

thepastaclaw (commented):
## Code Review

Both prior findings (prior-1 FFI flattening of TransactionBroadcastUnconfirmed and prior-2 rejected asset-lock leaving a resumable Built row) are fully resolved at head 12ab17f3. Carried-forward findings: none. New in-scope findings from the latest delta: two suggestion-level items — (a) the new asset-lock untrack-on-Rejected branch in create_funded_asset_lock_proof has no direct regression test, unlike the core-send sibling that got table-driven tests; (b) a narrow window between broadcast_releasing_on_rejection releasing the UTXO reservation and untrack_asset_lock removing the Built row lets a concurrent resume_asset_lock re-broadcast the just-rejected transaction using inputs now eligible for re-selection. Neither rises to blocking.

🟡 2 suggestion(s)

<details>

```

In `packages/rs-platform-wallet/src/wallet/asset_lock/build.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/asset_lock/build.rs:356-371: New untrack-on-Rejected branch in create_funded_asset_lock_proof has no direct regression test
  The PR added table-driven tests (`send_to_addresses_releases_reservation_on_broadcast_failure` / `..._keeps_reservation_on_ambiguous_broadcast_failure`) for the parallel core-send path, but the newly added asset-lock branch here — matching `BroadcastError::Rejected` and calling `untrack_asset_lock`, while `MaybeSent` deliberately leaves the reservation and Built row intact — has no equivalent coverage. `build.rs` has no `#[cfg(test)]` module, and no test elsewhere in the crate exercises `AssetLockManager` together with `BroadcastError::Rejected`. Since this is the exact invariant that closes prior-2 (a `resume_asset_lock` on a rejected outpoint must not re-broadcast a stale tx whose inputs may already be re-spent), a future refactor could silently regress the branch or the `matches!` guard without breaking any existing test. A direct test that asserts a `Rejected` outcome (i) queues a changeset with the outpoint in `removed`, (ii) empties the in-memory `tracked_asset_locks` entry, and that a `MaybeSent` outcome does neither, would pin the invariant.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/asset_lock/build.rs:356-371: Reservation release and Built-row untrack are not atomic, leaving a window where a concurrent resume_asset_lock can re-broadcast the rejected tx
  `broadcast_releasing_on_rejection` releases the funding UTXO reservation and returns `Err(BroadcastError::Rejected)`; only afterward does this call site acquire the write lock via `untrack_asset_lock` to remove the `Built` row. Between those two lock acquisitions, the `Built` entry for `out_point` is still visible via `list_tracked_locks()` / `tracked_asset_locks.values()`. `resume_asset_lock` (sync/recovery.rs:209-261) for `Built` status calls `self.broadcaster.broadcast(&tx).await?` directly — it does not go through `broadcast_releasing_on_rejection` and does not re-check reservation state. If a host recovery/resume flow interleaves precisely with the rejection cleanup, it can re-drive the just-rejected transaction using inputs whose reservation is already released and therefore concurrently eligible for re-selection by a new build. `untrack_asset_lock` also has no `AssetLockStatus::Built` guard, so if `resume_asset_lock` advanced the row to `Broadcast` in that window, the untrack still removes it — clobbering genuine progress. Blast radius is bounded: Core's mempool resolves double-spend at the network layer, so this is not a fund-loss bug, but it can produce a confusing state where the wallet believes a new build succeeded while the confirmed tx is actually the resumed one. Consider combining the reservation release and row untrack under a single write-lock section, or gating `untrack_asset_lock` on the status still being `Built`.
```
</details>

Source: reviewers: `opus` (general), `claude-sonnet-5` (general), `gpt-5.5` (general), `opus` (security-auditor), `claude-sonnet-5` (security-auditor), `gpt-5.5` (security-auditor), `opus` (rust-quality), `claude-sonnet-5` (rust-quality), `gpt-5.5` (rust-quality), `opus` (ffi-engineer), `claude-sonnet-5` (ffi-engineer), `gpt-5.5` (ffi-engineer); verifier: `opus`; specialists: `security-auditor`, `rust-quality`, `ffi-engineer`

packages/rs-platform-wallet/src/wallet/asset_lock/build.rs
```diff
@@ -345,8 +345,30 @@ impl<B: TransactionBroadcaster + ?Sized> AssetLockManager<B> {
...
+            &tx,
+        )
+        .await
+        {
+            if matches!(e, crate::broadcaster::BroadcastError::Rejected { .. }) {
+                let cs_untrack = self.untrack_asset_lock(&out_point).await;
+                self.queue_asset_lock_changeset(cs_untrack);
+            }
+            return Err(e.into());
+        }
```
thepastaclaw: **🟡 Suggestion: New untrack-on-Rejected branch in create_funded_asset_lock_proof has no direct regression test**

The PR added table-driven tests (`send_to_addresses_releases_reservation_on_broadcast_failure` / `..._keeps_reservation_on_ambiguous_broadcast_failure`) for the parallel core-send path, but the newly added asset-lock branch here — matching `BroadcastError::Rejected` and calling `untrack_asset_lock`, while `MaybeSent` deliberately leaves the reservation and Built row intact — has no equivalent coverage. `build.rs` has no `#[cfg(test)]` module, and no test elsewhere in the crate exercises `AssetLockManager` together with `BroadcastError::Rejected`. Since this is the exact invariant that closes prior-2 (a `resume_asset_lock` on a rejected outpoint must not re-broadcast a stale tx whose inputs may already be re-spent), a future refactor could silently regress the branch or the `matches!` guard without breaking any existing test. A direct test that asserts a `Rejected` outcome (i) queues a changeset with the outpoint in `removed`, (ii) empties the in-memory `tracked_asset_locks` entry, and that a `MaybeSent` outcome does neither, would pin the invariant.

thepastaclaw: **🟡 Suggestion: Reservation release and Built-row untrack are not atomic, leaving a window where a concurrent resume_asset_lock can re-broadcast the rejected tx**

`broadcast_releasing_on_rejection` releases the funding UTXO reservation and returns `Err(BroadcastError::Rejected)`; only afterward does this call site acquire the write lock via `untrack_asset_lock` to remove the `Built` row. Between those two lock acquisitions, the `Built` entry for `out_point` is still visible via `list_tracked_locks()` / `tracked_asset_locks.values()`. `resume_asset_lock` (sync/recovery.rs:209-261) for `Built` status calls `self.broadcaster.broadcast(&tx).await?` directly — it does not go through `broadcast_releasing_on_rejection` and does not re-check reservation state. If a host recovery/resume flow interleaves precisely with the rejection cleanup, it can re-drive the just-rejected transaction using inputs whose reservation is already released and therefore concurrently eligible for re-selection by a new build. `untrack_asset_lock` also has no `AssetLockStatus::Built` guard, so if `resume_asset_lock` advanced the row to `Broadcast` in that window, the untrack still removes it — clobbering genuine progress. Blast radius is bounded: Core's mempool resolves double-spend at the network layer, so this is not a fund-loss bug, but it can produce a confusing state where the wallet believes a new build succeeded while the confirmed tx is actually the resumed one. Consider combining the reservation release and row untrack under a single write-lock section, or gating `untrack_asset_lock` on the status still being `Built`.

<sub>source: ['claude']</sub>

QuantumExplorer: Fixed in 25ea9d2123 — added `rejected_asset_lock_broadcast_untracks_row_and_releases_reservation` (asserts the outpoint lands in the changeset's `removed` set, `tracked_asset_locks` empties, and a fresh build reselects the inputs) and `ambiguous_asset_lock_broadcast_keeps_reservation_and_built_row` (Built row kept, no deletion queued, rebuild fails at input selection), sharing the funded-wallet fixture via a new `test_support` module.

QuantumExplorer: Fixed in 25ea9d2123 — the cleanup now untracks the `Built` row *before* releasing the reservation (while the reservation is held the inputs can't be re-selected; once the row is gone `resume_asset_lock` can't re-drive it), and `untrack_asset_lock` is guarded on the row still being `Built` so a concurrent resume that advanced it to `Broadcast` is kept rather than clobbered.

packages/rs-platform-wallet/src/wallet/asset_lock/build.rs
```diff
@@ -345,8 +345,30 @@ impl<B: TransactionBroadcaster + ?Sized> AssetLockManager<B> {
...
+                    &self.wallet_manager,
+                    &self.wallet_id,
+                    key_wallet::account::account_type::StandardAccountType::BIP44Account,
+                    account_index,
+                    &tx,
+                )
+                .await;
+            }
+            return Err(e.into());
+        }
```
thepastaclaw, line 381: **🔴 Blocking: Release the funding reservation only when untrack_asset_lock actually removed the Built row**

The reorder at build.rs:357-371 correctly untracks the `Built` asset-lock row before releasing the funding reservation, and `untrack_asset_lock` (sync/tracking.rs:46-64) is now status-guarded so a concurrent `resume_asset_lock` that advanced the row to `Broadcast` in the window between this call's rejected broadcast and its cleanup is preserved (logged, not clobbered) — returning an empty changeset in that case.

But `release_reservation_after_rejected_broadcast` on line 361 runs unconditionally inside the `Rejected` branch, regardless of whether `untrack_asset_lock` actually removed the entry. When the guard fires, the row is left tracked as `Broadcast` (positive evidence the transaction reached the network via the concurrent resume), yet the funding reservation is still released, making its inputs immediately selectable by a fresh build. This is strictly worse than the `MaybeSent` case this PR is designed to protect (there the outcome is merely ambiguous; here the status transition is direct evidence of a successful send), and it reopens a variant of the exact double-spend race the commit message ("close the rejected-asset-lock race") claims to eliminate.

The invariant the doc comment at build.rs:348-356 claims — "at no point is the row resumable while its inputs are re-spendable" — is violated in this interleaving: after cleanup the row is resumable/tracked *and* the inputs are re-spendable. `reservations.rs:70-75` explicitly states the ordering precondition, and that precondition is only satisfied when the untrack actually succeeded. Gate the release on `cs_untrack.removed.contains(&out_point)`.

```suggestion
        if let Err(e) = self.broadcaster.broadcast(&tx).await {
            if matches!(e, crate::broadcaster::BroadcastError::Rejected { .. }) {
                let cs_untrack = self.untrack_asset_lock(&out_point).await;
                let removed_built_row = cs_untrack.removed.contains(&out_point);
                self.queue_asset_lock_changeset(cs_untrack);
                if removed_built_row {
                    crate::wallet::reservations::release_reservation_after_rejected_broadcast(
                        &self.wallet_manager,
                        &self.wallet_id,
                        key_wallet::account::account_type::StandardAccountType::BIP44Account,
                        account_index,
                        &tx,
                    )
                    .await;
                }
            }
            return Err(e.into());
        }
```

thepastaclaw (changes requested):
## Code Review

Both prior findings from 12ab17f3 are FIXED at head 25ea9d21: dedicated Rejected/MaybeSent regression tests were added to build.rs, and the cleanup order was inverted (untrack before release) with `untrack_asset_lock` newly guarded on `status == Built`. However, the same reorder introduced one new in-scope logic gap flagged by 5 of 12 reviewers: the reservation release in `create_funded_asset_lock_proof` runs unconditionally even when the untrack guard fires (i.e. a concurrent `resume_asset_lock` already advanced the row past `Built`), freeing UTXOs of a transaction the wallet is still tracking as `Broadcast`. That is a variant of the exact race this commit's title claims to close, and the fix is a one-line conditional on `cs_untrack.removed.contains(&out_point)`.

🔴 1 blocking


Source: reviewers: `opus` (general), `claude-sonnet-5` (general), `gpt-5.5` (general), `opus` (security-auditor), `claude-sonnet-5` (security-auditor), `gpt-5.5` (security-auditor), `opus` (rust-quality), `claude-sonnet-5` (rust-quality), `gpt-5.5` (rust-quality), `opus` (ffi-engineer), `claude-sonnet-5` (ffi-engineer), `gpt-5.5` (ffi-engineer); verifier: `opus`; specialists: `security-auditor`, `rust-quality`, `ffi-engineer`.

<details>

```

In `packages/rs-platform-wallet/src/wallet/asset_lock/build.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/wallet/asset_lock/build.rs:357-371: Release the funding reservation only when untrack_asset_lock actually removed the Built row
  The reorder at build.rs:357-371 correctly untracks the `Built` asset-lock row before releasing the funding reservation, and `untrack_asset_lock` (sync/tracking.rs:46-64) is now status-guarded so a concurrent `resume_asset_lock` that advanced the row to `Broadcast` in the window between this call's rejected broadcast and its cleanup is preserved (logged, not clobbered) — returning an empty changeset in that case.

  But `release_reservation_after_rejected_broadcast` on line 361 runs unconditionally inside the `Rejected` branch, regardless of whether `untrack_asset_lock` actually removed the entry. When the guard fires, the row is left tracked as `Broadcast` (positive evidence the transaction reached the network via the concurrent resume), yet the funding reservation is still released, making its inputs immediately selectable by a fresh build. This is strictly worse than the `MaybeSent` case this PR is designed to protect (there the outcome is merely ambiguous; here the status transition is direct evidence of a successful send), and it reopens a variant of the exact double-spend race the commit message ("close the rejected-asset-lock race") claims to eliminate.

  The invariant the doc comment at build.rs:348-356 claims — "at no point is the row resumable while its inputs are re-spendable" — is violated in this interleaving: after cleanup the row is resumable/tracked *and* the inputs are re-spendable. `reservations.rs:70-75` explicitly states the ordering precondition, and that precondition is only satisfied when the untrack actually succeeded. Gate the release on `cs_untrack.removed.contains(&out_point)`.
```
</details>

packages/rs-platform-wallet/src/wallet/asset_lock/build.rs
QuantumExplorer: Fixed in 5df222b15d — the release is now gated on `cs_untrack.removed.contains(&out_point)` as suggested, so when the Built guard fires the reservation stays held exactly like a `MaybeSent` outcome. Pinned with `rejected_broadcast_racing_concurrent_resume_keeps_row_and_reservation`, which advances the row to `Broadcast` mid-broadcast, returns `Rejected`, and asserts the row survives, no deletion is queued, and a rebuild still fails at input selection.

thepastaclaw (approved):
## Code Review

Incremental review from 25ea9d21 to 5df222b1. The sole prior blocking finding (prior-1) is verified FIXED: `create_funded_asset_lock_proof` now captures `removed_built_row = cs_untrack.removed.contains(&out_point)` before queuing the changeset and gates the reservation release on that flag, exactly matching the suggested fix. `untrack_asset_lock` only populates `cs.removed` when the row was actually removed (guarded on `status == Built`), so the flag is a correct proxy for the race outcome. A new regression test `rejected_broadcast_racing_concurrent_resume_keeps_row_and_reservation` pins the concurrent-resume interleaving. No new in-scope findings in the latest delta or cumulative PR.

Source: reviewers: `opus` (general), `claude-sonnet-5` (general), `gpt-5.5` (general), `opus` (security-auditor), `claude-sonnet-5` (security-auditor), `gpt-5.5` (security-auditor), `opus` (rust-quality), `claude-sonnet-5` (rust-quality), `gpt-5.5` (rust-quality), `opus` (ffi-engineer), `claude-sonnet-5` (ffi-engineer), `gpt-5.5` (ffi-engineer); verifier: `opus`; specialists: `security-auditor`, `rust-quality`, `ffi-engineer`.

thepastaclaw (commented):
## Code Review

Carried-forward prior findings: none — the prior review at 5df222b15df8d20779fd58a58de697c8548fac08 had zero unresolved findings and nothing in the current head aef46b99 regresses it. New findings in the latest delta: one suggestion, independently raised by two sonnet5 reviewers (security + rust-quality), that the safety-critical SPV Rejected/MaybeSent classification in SpvRuntime::broadcast_transaction has no direct unit test coverage against dash-spv's error semantics. Verified in-tree: the core reservation-release seam (BroadcastError split, broadcast_releasing_on_rejection, asset-lock untrack-before-release with the removed_built_row guard) is correct, read-lock discipline is sound, and the FFI code renumbering to 20 (to accommodate ErrorShieldedNoRecordedAnchor=19 from the v4.0-dev merge) is mirrored consistently in the Swift PlatformWalletResult. The DapiBroadcaster blanket-MaybeSent behavior flagged by opus-rust-quality is an author-acknowledged, code-commented tradeoff (safe, conservative, documented as a follow-up), not a defect — dropped.

🟡 1 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet/src/spv/runtime.rs:104-128: SpvRuntime::broadcast_transaction's Rejected/MaybeSent classification is untested against dash-spv error semantics
  This method is the concrete classification point that decides whether an SPV broadcast failure is safe to treat as `Rejected` (release the UTXO reservation and let a later resume re-spend the inputs) versus `MaybeSent` (must keep the reservation to prevent double-spend). The safety of releasing on `SpvError::Network(NetworkError::NotConnected)` depends entirely on an external dash-spv invariant: that `NotConnected` is only raised when zero peers are connected *before* any send attempt, never after the client has already handed the tx to at least one peer. dash-spv is pinned by git rev and its error contract is not covered by any test in this crate — `mod tests` is absent from `spv/runtime.rs`, and every test in `wallet/core/broadcast.rs` and `wallet/asset_lock/build.rs` uses mock broadcasters that never construct a real `SpvRuntime`. A future dash-spv change that raises `NotConnected` post-partial-send would compile cleanly and silently reintroduce the double-spend window this PR is designed to prevent, with no test signal. At minimum, add a targeted unit test that stubs the internal SPV client and pins both match arms (unstarted client → `Rejected`, `NetworkError::NotConnected` → `Rejected`, any other `SpvError` → `MaybeSent`), so a dash-spv semantic change is caught at the crate boundary rather than the trust boundary.
```
</details>

Source: reviewers: `opus` (general), `claude-sonnet-5` (general), `gpt-5.5` (general), `opus` (security-auditor), `claude-sonnet-5` (security-auditor), `gpt-5.5` (security-auditor), `opus` (rust-quality), `claude-sonnet-5` (rust-quality), `gpt-5.5` (rust-quality), `opus` (ffi-engineer), `claude-sonnet-5` (ffi-engineer, failed_or_unparseable), `gpt-5.5` (ffi-engineer); verifier: `opus`; specialists: `security-auditor`, `rust-quality`, `ffi-engineer`

packages/rs-platform-wallet/src/spv/runtime.rs
```diff
@@ -92,19 +93,36 @@ impl SpvRuntime {
...
+                SpvError::Network(NetworkError::NotConnected) => BroadcastError::Rejected {
+                    reason: "SPV broadcast failed: no connected peers".to_string(),
+                },
+                other => BroadcastError::MaybeSent {
+                    reason: format!("SPV broadcast failed: {}", other),
+                },
+            })?;
 
         Ok(())
     }
```
thepastaclaw, line 144: **🟡 Suggestion: SpvRuntime::broadcast_transaction's Rejected/MaybeSent classification is untested against dash-spv error semantics**

This method is the concrete classification point that decides whether an SPV broadcast failure is safe to treat as `Rejected` (release the UTXO reservation and let a later resume re-spend the inputs) versus `MaybeSent` (must keep the reservation to prevent double-spend). The safety of releasing on `SpvError::Network(NetworkError::NotConnected)` depends entirely on an external dash-spv invariant: that `NotConnected` is only raised when zero peers are connected *before* any send attempt, never after the client has already handed the tx to at least one peer. dash-spv is pinned by git rev and its error contract is not covered by any test in this crate — `mod tests` is absent from `spv/runtime.rs`, and every test in `wallet/core/broadcast.rs` and `wallet/asset_lock/build.rs` uses mock broadcasters that never construct a real `SpvRuntime`. A future dash-spv change that raises `NotConnected` post-partial-send would compile cleanly and silently reintroduce the double-spend window this PR is designed to prevent, with no test signal. At minimum, add a targeted unit test that stubs the internal SPV client and pins both match arms (unstarted client → `Rejected`, `NetworkError::NotConnected` → `Rejected`, any other `SpvError` → `MaybeSent`), so a dash-spv semantic change is caught at the crate boundary rather than the trust boundary.

<sub>source: ['sonnet5', 'sonnet5']</sub>

QuantumExplorer: Fixed in 79caf9b101 — extracted the mapping into `classify_spv_broadcast_error` and pinned all three arms with unit tests in `spv/runtime.rs`: unstarted client → `Rejected`, `NetworkError::NotConnected` → `Rejected`, and a table of other `SpvError`s (Timeout, PeerDisconnected, ConnectionFailed, Config) → `MaybeSent`, with a doc note that the NotConnected pin must be revisited if dash-spv ever raises it post-partial-send.

thepastaclaw (approved):
## Code Review

Prior finding on untested SPV Rejected/MaybeSent classification is FIXED at 79caf9b1 by the extracted `classify_spv_broadcast_error` and its three-arm test module (unstarted-client, `NotConnected`→Rejected, all other `SpvError`→MaybeSent). Reservation-release seam, DAPI conservative MaybeSent, asset-lock untrack-before-release ordering, and FFI code-20 mapping (with Swift mirror) all verify clean. No in-scope blocking or suggestion findings remain.

Source: reviewers: `opus` (general), `claude-sonnet-5` (general), `gpt-5.5` (general), `opus` (security-auditor), `claude-sonnet-5` (security-auditor, failed_or_unparseable), `gpt-5.5` (security-auditor), `opus` (rust-quality), `claude-sonnet-5` (rust-quality), `gpt-5.5` (rust-quality), `opus` (ffi-engineer), `claude-sonnet-5` (ffi-engineer), `gpt-5.5` (ffi-engineer); verifier: `opus`; specialists: `security-auditor`, `rust-quality`, `ffi-engineer`.

packages/rs-platform-wallet/src/wallet/asset_lock/build.rs
```diff
@@ -345,8 +345,40 @@ impl<B: TransactionBroadcaster + ?Sized> AssetLockManager<B> {
...
+                // the untrack guard fired instead — a concurrent
+                // `resume_asset_lock` advanced the row past `Built`, positive
+                // evidence the transaction reached the network after all —
+                // the inputs must stay reserved exactly like a `MaybeSent`
+                // outcome, or the still-tracked row would be resumable while
+                // its inputs are re-spendable.
+                let removed_built_row = cs_untrack.removed.contains(&out_point);
+                self.queue_asset_lock_changeset(cs_untrack);
+                if removed_built_row {
+                    crate::wallet::reservations::release_reservation_after_rejected_broadcast(
```
thepastaclaw, line 370: **🔴 Blocking: Rejected cleanup can release inputs while a resume is broadcasting the same asset-lock transaction**

The release gate only checks whether `untrack_asset_lock` removed a row that was still `Built`. That misses the interleave where `resume_asset_lock` has already read the `Built` row, cloned the transaction under the read lock, dropped the lock, and is awaiting `self.broadcaster.broadcast(&tx)` before it can advance the row to `Broadcast`. In that window, the original create path can receive `BroadcastError::Rejected`, remove the still-`Built` row here, and release the funding reservation while the resume path can still hand the same transaction to the network. That makes the inputs selectable for a retry while the original transaction may already be propagating, reopening the double-spend window this reservation contract is meant to close. The existing race test only covers the case where the row is advanced before cleanup, not the read-before-broadcast case.

thepastaclaw (changes requested):
## Code Review

Carried-forward prior findings: none; the prior SPV Rejected/MaybeSent classification issue remains resolved by the classifier tests in `packages/rs-platform-wallet/src/spv/runtime.rs`. New current-head findings: one blocking in-scope asset-lock race remains in the PR diff against the updated base. The wasm protocol API change is real in the checked-out tree, but it comes from the merged base parent rather than this PR, so it is dropped from PR review output.

Source: reviewers `gpt-5.5` for general, security-auditor, rust-quality, and ffi-engineer; attempted `opus` and `claude-sonnet-5` reviewer lanes failed with usage exhaustion; verifier `gpt-5.5` recovered after `opus` verifier usage exhaustion.

🔴 1 blocking

<details>

```

In `packages/rs-platform-wallet/src/wallet/asset_lock/build.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/wallet/asset_lock/build.rs:367-370: Rejected cleanup can release inputs while a resume is broadcasting the same asset-lock transaction
  The release gate only checks whether `untrack_asset_lock` removed a row that was still `Built`. That misses the interleave where `resume_asset_lock` has already read the `Built` row, cloned the transaction under the read lock, dropped the lock, and is awaiting `self.broadcaster.broadcast(&tx)` before it can advance the row to `Broadcast`. In that window, the original create path can receive `BroadcastError::Rejected`, remove the still-`Built` row here, and release the funding reservation while the resume path can still hand the same transaction to the network. That makes the inputs selectable for a retry while the original transaction may already be propagating, reopening the double-spend window this reservation contract is meant to close. The existing race test only covers the case where the row is advanced before cleanup, not the read-before-broadcast case.
```
</details>

PR #3987, opened 2026-07-02
refactor(platform-wallet): consolidate address-balance reconciliation into one guarded seam

QuantumExplorer:

Follow-up to the review of #3969. `rs-platform-wallet` carried four hand-rolled copies of the platform-address balance reconciliation (translate proof-attested `AddressInfos` → apply to `ManagedPlatformAccount` → persist a `PlatformAddressChangeSet`), each with its own gaps:

- `transfer.rs` and `withdrawal.rs` resolved derivation indexes only through the live derived pool — and withdrawal's `unwrap_or(0)` fallback could **corrupt index 0** for restored addresses no longer in a live pool.
- `register_from_addresses.rs` discarded `_address_infos` entirely (its own TODO asked for reconciliation), and `transfer_to_addresses.rs` left wallet-owned recipient balances stale until the next BLAST sync.
- The top-up path required the FFI to make two calls with a "caller MUST reconcile" doc-contract.
- None of the copies coordinated with the 15s background sync, so a sync pass racing a state transition could overwrite fresh proof-attested balances with stale scanned ones — and the provider's committed `found` map (the sync-diff baseline and SDK sync seed) diverged from what reconciliation applied.

## What was done?

- **One shared seam**: `PlatformAddressWallet::reconcile_address_infos` now owns translate → freshness-guard → apply → persist. `transfer`, `withdraw`, `fund_from_asset_lock`, and all identity address flows route through it. `build_top_up_balance_entries` / `build_transfer_persistence_entries` are generalized into `build_address_balance_entries` over an index-resolver: resolution goes through the provider's persisted index↔address bijection first (restored-wallet coverage), falling back to the live pools for addresses derived since the last sync (fresh change addresses); pool-resolved pairs are merged into the bijection. The withdrawal index-0 corruption is gone.
- **Previously discarded proofs wired in**: `IdentityWallet::register_from_addresses` and `transfer_credits_to_addresses_with_external_signer` now return their proof-attested `AddressInfos`; external recipients are skipped by the seam.
- **Composite orchestration on `PlatformWallet`** (`top_up_from_addresses`, `register_from_addresses`, `transfer_credits_to_addresses_with_external_signer`): the struct that owns both `.identity()` and `.platform()` composes the identity op with the reconciliation, so the FFI makes one call and the "caller MUST reconcile" contract disappears. FFI signatures are unchanged (internals only — no header regeneration needed).
- **Freshness guard against the background sync**: the seam commits through `PlatformPaymentAddressProvider::commit_reconciliation` while holding the provider write lock across both the provider commit and the account-balance write (lock order provider → wallet-manager, matching the sync callbacks). Entries whose nonce is below the committed `found` seed's are dropped (a sync or later transition already committed fresher state); identical entries are dropped as no-ops; proof-attested removals (zero funds) drop the address from the seed, mirroring absent handling. The committed `found` map is updated so the sync-diff baseline and the SDK sync seed agree with what was applied. `sync_balances` now persists its diff **before** releasing the provider lock, so a fresher reconciliation row can never be overwritten on disk by an older sync diff.

Behavioral notes: withdrawal now persists its changeset internally (matching transfer — the FFI still also returns it; host-side application is an idempotent upsert), and transfer reconciles owned addresses across **all** the wallet's platform payment accounts rather than only the input account.

## How Has This Been Tested?

- `cargo test -p platform-wallet --lib` — 209 passed, including new tests: freshness-guard stale-nonce drop, unchanged no-op drop, fresh-entry commit to `found`, zero-funds removal from the seed, live-pool fallback extending the bijection, and the retargeted persist-contract integration test now also asserting the committed seed carries the reconciled funds.
- `cargo check -p platform-wallet-ffi`, `cargo check -p platform-wallet --all-targets` and `--all-features --tests`, plus reverse deps `platform-wallet-storage --all-targets` and `rs-unified-sdk-ffi` — all clean.
- `cargo clippy -p platform-wallet --lib --tests` and `-p platform-wallet-ffi` — no warnings; `cargo fmt --all` applied.

## Breaking Changes

None on consensus or FFI surfaces. Within the `platform-wallet` crate (pre-release, internal): `IdentityWallet::register_from_addresses` / `transfer_credits_to_addresses_with_external_signer` now also return `AddressInfos`, and `PlatformAddressWallet::apply_top_up_reconciliation` is replaced by `reconcile_address_infos`.

## Checklist:

### DIFF
--- a/packages/rs-platform-wallet/src/wallet/platform_addresses/provider.rs
+++ b/packages/rs-platform-wallet/src/wallet/platform_addresses/provider.rs
@@ -826,21 +826,26 @@ impl AddressProvider for PlatformPaymentAddressProvider {
     }
 }
 
-/// Resolve each spent platform address in a top-up's proof-attested
-/// `address_infos` to its `(account_index, address_index)` using the
-/// per-account index bijections, and pair it with the proof's post-spend
-/// balance + nonce. Resolving against the bijections — rather than the live
-/// derived address pool — means addresses restored from disk (present in the
-/// persisted index map but not in `ManagedPlatformAccount::addresses`) are
-/// still reconciled; without this, a top-up that spends a restored cached row
-/// would leave its stale balance behind, preserving the phantom balance.
+/// Translate a state transition's proof-attested `address_infos` into
+/// persistence-changeset entries, resolving each address's
+/// `(account_index, address_index)` through `resolve_index`.
 ///
-/// Addresses the proof returns that the wallet doesn't own (no bijection
-/// entry) are skipped. Pure and lock-free so the reconciliation is
-/// unit-testable.
-pub(crate) fn build_top_up_balance_entries(
+/// Non-P2PKH addresses and addresses the resolver doesn't recognise
+/// (external recipients, or addresses the wallet doesn't own) are skipped.
+/// Missing per-address info (`None`) maps to zero balance / zero nonce —
+/// the on-chain post-transition state for an address removed from state
+/// (e.g. a fully consumed input). Pure and lock-free so every caller's
+/// translation is unit-testable.
+///
+/// Callers supply the resolver so every reconciliation path can resolve
+/// through the provider's persisted `index <-> address` bijection —
+/// covering addresses restored from disk that are no longer in a live
+/// derived pool — with the live pool as fallback for addresses derived
+/// since the last sync (see
+/// [`PlatformPaymentAddressProvider::commit_reconciliation`]).
+pub(crate) fn build_address_balance_entries(
     wallet_id: WalletId,
-    per_account_addresses: &[(u32, &BiBTreeMap<AddressIndex, PlatformP2PKHAddress>)],
+    resolve_index: impl Fn(&PlatformP2PKHAddress) -> Option<(u32, AddressIndex)>,
     address_infos: &AddressInfos,
 ) -> Vec<PlatformAddressBalanceEntry> {
     let mut entries = Vec::new();
@@ -849,6 +854,9 @@ pub(crate) fn build_top_up_balance_entries(
             continue;
         };
         let p2pkh = PlatformP2PKHAddress::new(*hash);
+        let Some((account_index, address_index)) = resolve_index(&p2pkh) else {
+            continue;
+        };
         let funds = match maybe_info {
             Some(ai) => AddressFunds {
                 balance: ai.balance,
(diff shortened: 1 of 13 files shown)


thepastaclaw:
✅ Review complete (commit 1d48d6fb)

packages/rs-platform-wallet/src/wallet/platform_addresses/wallet.rs
```diff
@@ -159,98 +164,182 @@ impl PlatformAddressWallet {
...
             tracing::error!(
+                context,
                 error = %e,
-                "Failed to persist top-up platform-address reconciliation; \
+                "Failed to persist platform-address reconciliation; \
                  in-memory balances are updated but durable rows stay stale \
                  until the next platform-address sync"
             );
         }
+        cs
```
thepastaclaw: **🟡 Suggestion: reconcile_address_infos persists after releasing the provider lock, breaking the invariant sync.rs was just changed to establish**

`reconcile_address_infos` drops the provider write guard at line 325 and only then calls `self.persister.store(cs.clone().into())` at line 333. The companion change in `sync.rs` (lines 155-168) moves its `persister.store` call *inside* the same provider write lock with an explicit comment documenting why: `reconcile_address_infos` "serializes on this lock and persists its proof-attested entries after acquiring it, so keeping this store inside the critical section guarantees any reconciliation that observed this pass's committed state also lands on disk after it — releasing first would let a fresher reconciliation row be overwritten by this (older) diff."

The reconcile side of that invariant is not upheld. Because `WalletPersister::store` is a synchronous call into an `Arc<dyn PlatformWalletPersistence>` reachable from any Tokio worker, once `drop(guard)` runs a background sync (or a second concurrent reconciliation) can fully acquire the provider lock, commit a fresher `found` seed, and persist its own diff *before* this reconciliation's delayed `store()` finally fires — at which point the older reconciliation row overwrites the fresher sync row on disk. In-memory state stays correct because both writers apply the newer value under the lock, and the next successful sync heals the disk row; but the durable state is silently stale during that window, which is exactly the property this PR advertises as fixed.

Moving the `self.persister.store(...)` call to just before `drop(guard)` (mirroring `sync.rs`) closes the race with no extra locking cost.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet/src/wallet/platform_addresses/provider.rs
```diff
@@ -859,20 +867,145 @@ pub(crate) fn build_top_up_balance_entries(
...
+                             maps to a different address — state drift; \
+                             leaving the bijection untouched"
+                        );
+                    } else {
+                        state.addresses.insert(entry.address_index, entry.address);
+                    }
+                }
             }
+            outcome.entries.push(entry);
         }
```
thepastaclaw, line 1027: **🟡 Suggestion: commit_reconciliation pushes entries into the outcome even when the account isn't tracked in per_wallet**

`outcome.entries.push(entry)` at line 1005 runs unconditionally, outside the `if let Some(state) = account_state { ... }` block (lines 979-1004). `account_state` is `wallet_state.get_mut(&entry.account_index)`; it is guaranteed `Some` only when the resolver picked `entry.account_index` from iterating `wallet_state` itself, not when it picked from the live-pool fallback (`pool_indexes.get(p2pkh).copied()` at line 950).

`pool_indexes` is built in `reconcile_address_infos` by scanning `info.core_wallet.all_platform_payment_managed_accounts()` — every key-class-0 platform-payment account on the live wallet — not filtered to accounts already tracked in the provider's `per_wallet` state. If a platform-payment account exists in the live wallet but is not yet tracked by the provider (e.g. added after the provider snapshot was built), an address in that account resolves via the pool fallback to an `account_index` for which `wallet_state.get_mut(...)` returns `None`. In that case: `state.found.insert/remove` never runs, the bijection merge never runs, but the entry is still pushed into `outcome.entries`, so the caller applies it to the live `ManagedPlatformAccount` and persists a `PlatformAddressChangeSet` for it — with none of the `tracing::warn!`/`tracing::debug!` diagnostics the rest of the function uses for skipped entries.

This contradicts `ReconciliationOutcome::entries`' own doc comment ("Already committed to the provider's `found` map / bijection") and undermines the PR's stated goal that "the committed found map is updated so the sync-diff baseline and the SDK sync seed agree with what was applied." Either move the `outcome.entries.push(entry)` inside the `if let Some(state) = account_state` block (dropping the entry with a log if the account isn't tracked) or update the resolver to skip pool-only addresses whose account isn't in `wallet_state`.

<sub>source: ['claude']</sub>

thepastaclaw (approved):
## Code Review

PR consolidates four hand-rolled address-balance reconciliation copies into a single guarded seam. The core design is sound and well-tested. Three concrete suggestions surface: the reconcile seam persists outside the provider lock while the sibling sync path was just changed to persist inside it (asymmetric with the invariant this PR documents), `commit_reconciliation` unconditionally returns entries the outcome doc says are committed even when the target account isn't tracked in `per_wallet`, and `reset_sync_state` inverts the crate's provider→wallet_manager lock order.

Source: reviewers: opus and claude-sonnet-5 for general, security-auditor, rust-quality, and ffi-engineer; gpt-5.5[high] Codex lanes failed at ACP initialize and are recorded as failed; verifier: opus.

<details>
<summary>Findings not posted inline (1)</summary>

- **[NITPICK] `packages/rs-platform-wallet/src/wallet/platform_addresses/wallet.rs:433-446`: reset_sync_state inverts the crate's documented provider→wallet_manager lock order** — `reset_sync_state` takes `wallet_manager.write()` first (lines 434-441), releases it, then takes `provider.write()` (lines 442-445). Every other path in this module — `sync_balances`, the `AddressProvider` callbacks, and `reconcile_address_infos` — takes `provider.write()` first and holds it acro...

</details>

<details>

```

In `packages/rs-platform-wallet/src/wallet/platform_addresses/wallet.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/platform_addresses/wallet.rs:325-342: reconcile_address_infos persists after releasing the provider lock, breaking the invariant sync.rs was just changed to establish
  `reconcile_address_infos` drops the provider write guard at line 325 and only then calls `self.persister.store(cs.clone().into())` at line 333. The companion change in `sync.rs` (lines 155-168) moves its `persister.store` call *inside* the same provider write lock with an explicit comment documenting why: `reconcile_address_infos` "serializes on this lock and persists its proof-attested entries after acquiring it, so keeping this store inside the critical section guarantees any reconciliation that observed this pass's committed state also lands on disk after it — releasing first would let a fresher reconciliation row be overwritten by this (older) diff."

  The reconcile side of that invariant is not upheld. Because `WalletPersister::store` is a synchronous call into an `Arc<dyn PlatformWalletPersistence>` reachable from any Tokio worker, once `drop(guard)` runs a background sync (or a second concurrent reconciliation) can fully acquire the provider lock, commit a fresher `found` seed, and persist its own diff *before* this reconciliation's delayed `store()` finally fires — at which point the older reconciliation row overwrites the fresher sync row on disk. In-memory state stays correct because both writers apply the newer value under the lock, and the next successful sync heals the disk row; but the durable state is silently stale during that window, which is exactly the property this PR advertises as fixed.

  Moving the `self.persister.store(...)` call to just before `drop(guard)` (mirroring `sync.rs`) closes the race with no extra locking cost.

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/platform_addresses/provider.rs:956-1006: commit_reconciliation pushes entries into the outcome even when the account isn't tracked in per_wallet
  `outcome.entries.push(entry)` at line 1005 runs unconditionally, outside the `if let Some(state) = account_state { ... }` block (lines 979-1004). `account_state` is `wallet_state.get_mut(&entry.account_index)`; it is guaranteed `Some` only when the resolver picked `entry.account_index` from iterating `wallet_state` itself, not when it picked from the live-pool fallback (`pool_indexes.get(p2pkh).copied()` at line 950).

  `pool_indexes` is built in `reconcile_address_infos` by scanning `info.core_wallet.all_platform_payment_managed_accounts()` — every key-class-0 platform-payment account on the live wallet — not filtered to accounts already tracked in the provider's `per_wallet` state. If a platform-payment account exists in the live wallet but is not yet tracked by the provider (e.g. added after the provider snapshot was built), an address in that account resolves via the pool fallback to an `account_index` for which `wallet_state.get_mut(...)` returns `None`. In that case: `state.found.insert/remove` never runs, the bijection merge never runs, but the entry is still pushed into `outcome.entries`, so the caller applies it to the live `ManagedPlatformAccount` and persists a `PlatformAddressChangeSet` for it — with none of the `tracing::warn!`/`tracing::debug!` diagnostics the rest of the function uses for skipped entries.

  This contradicts `ReconciliationOutcome::entries`' own doc comment ("Already committed to the provider's `found` map / bijection") and undermines the PR's stated goal that "the committed found map is updated so the sync-diff baseline and the SDK sync seed agree with what was applied." Either move the `outcome.entries.push(entry)` inside the `if let Some(state) = account_state` block (dropping the entry with a log if the account isn't tracked) or update the resolver to skip pool-only addresses whose account isn't in `wallet_state`.
```
</details>

packages/rs-platform-wallet/src/wallet/platform_addresses/wallet.rs
QuantumExplorer: Fixed in 1d48d6fb52: `reconcile_address_infos` now persists before `drop(guard)`, mirroring `sync_balances`, so both writers' stores are totally ordered by the provider lock.

packages/rs-platform-wallet/src/wallet/platform_addresses/provider.rs
QuantumExplorer: Fixed in 1d48d6fb52: the pool fallback now filters to accounts present in `wallet_state`, and `outcome.entries.push` only runs after the found/bijection commit (defensive skip+warn otherwise) — plus a regression test pinning that an untracked account's pool address stays unresolved.

thepastaclaw (approved):
## Code Review

Delta cleanly resolves both prior suggestion-level findings: reconcile_address_infos now persists before releasing the provider write lock (mirroring sync.rs, restoring the total-order invariant), and commit_reconciliation's pool fallback is restricted to accounts the provider tracks, with a defensive continue+warn inside the commit loop and a new regression test. Only the nitpick-level reset_sync_state lock-inversion is carried forward (prior-3, code unchanged in this delta). No new defects introduced.

Source: reviewers: `opus` (general), `claude-sonnet-5` (general), `gpt-5.5[high]` (general, failed), `opus` (security-auditor), `claude-sonnet-5` (security-auditor), `gpt-5.5[high]` (security-auditor, failed), `opus` (rust-quality), `claude-sonnet-5` (rust-quality, failed), `gpt-5.5[high]` (rust-quality, failed), `opus` (ffi-engineer), `claude-sonnet-5` (ffi-engineer), `gpt-5.5[high]` (ffi-engineer, failed); verifier: `opus`; specialists: `security-auditor`, `rust-quality`, `ffi-engineer`

💬 1 nitpick(s)

<details>
<summary>Findings not posted inline (1)</summary>

- **[NITPICK] `packages/rs-platform-wallet/src/wallet/platform_addresses/wallet.rs:440-453`: reset_sync_state inverts the crate's provider→wallet_manager lock acquisition order** — Carried forward from the prior review (prior-3, unchanged in this delta). `reset_sync_state` acquires `wallet_manager.write()` first (lines 441-448), releases it, then acquires `provider.write()` (lines 449-452). Every other seam in this module — `sync_balances`, the `AddressProvider` callbacks,...

</details>

PR #3988, opened 2026-07-02
feat(platform-wallet): actively re-drive unconfirmed shielded spends

QuantumExplorer:
> Rebased onto `v4.0-dev` after #3982 squash-merged. The prune backstop this PR falls back to after exhausted attempts is #3982's release pass.

A shielded spend whose broadcast is accepted but whose result wait fails ambiguously (`ShieldedSpendUnconfirmed` — the TestFlight "may have gone through, do not retry" case) previously sat frozen until its anchor aged out of Platform's ~1000-block retention window (#3982's prune backstop), and after an app restart its persisted activity row could never resolve at all.

This implements the active-resolution design: **on timeout, trigger a shielded sync to see if it landed; if it didn't, re-broadcast; repeat up to 3 times; only then fall back to the prune backstop.**

## What was done?

- **Persist the transition on the ambiguous outcome.** All four spend flows (unshield / transfer / withdraw / identity-create-from-pool) serialize the signed state transition and arm a `PendingRedrive` — new `shielded_pending_spends` SQLite table carrying `(activity_id, anchor, nullifiers, st_bytes, attempts)`. On store open the records rehydrate both the redrive and the note reservations, so an unconfirmed spend keeps its notes reserved **across restarts** and its activity row can still flip — closing the restart gap noted in #3982's review.
- **Immediate first re-check.** The FFI spend wrappers fire a fire-and-forget forced shielded sync when the ambiguous outcome surfaces, so the first nullifier re-check happens in seconds instead of at the next background tick.
- **Sync-time re-drive pass** (runs after the scan's spent-note reconcile, before the prune-backstop release, judged against the same pre-scan recorded-anchor set): a landing was already detected by the reconcile (the `mark_spent` hook drops the record); otherwise, while the anchor is still recorded and `attempts < 3`, re-broadcast the **byte-identical** transition — fund-safe, identical nullifiers cannot double-spend — and classify:
  - accepted / inconclusive → count the attempt, let the next scan detect the landing;
  - `NullifierAlreadySpent` → the **original** executed; success signal, touch nothing (this check deliberately runs before the generic consensus-rejection arm);
  - any other consensus verdict → provably dead **now**: release the notes and flip the activity row to Failed, hours before the prune would;
  - after 3 silent attempts → the #3982 prune backstop owns it, unchanged.
- Store-level consistency: resolving any of a redrive's nullifiers (`mark_spent` / `clear_pending`) drops the whole record durably — a transition lands or dies atomically for all its nullifiers.

## How Has This Been Tested?

- `cargo test -p platform-wallet --lib --features shielded` → **320 passed** (316 base + 4 new).
- New coverage: SQLite roundtrip + rehydration (record, attempt counter, and reservations survive reopen; resolution deletions are durable); redrive decision paths against a mock SDK with **no broadcast expectation** — proving pruned-anchor and exhausted records are left untouched and a corrupt record is dropped, all without touching the network; `NullifierAlreadySpent` classification incl. the ordering constraint vs `carries_consensus_rejection`.
- `cargo fmt --all -- --check` and `cargo clippy -p platform-wallet --features shielded` clean; `platform-wallet-ffi` and default-features checks pass.

**Not covered (network-gated):** a live ambiguous timeout → re-broadcast → landing on testnet; best confirmed on a device once this ships in a build.

## Breaking Changes

None externally. `ShieldedStore` gains four required methods (`arm_redrive` / `pending_redrives` / `bump_redrive_attempts` / `clear_redrive`); both in-tree implementations updated. New SQLite table is additive (pre-release; no migration concerns).

## Checklist:

### DIFF
--- a/packages/rs-platform-wallet-ffi/src/shielded_send.rs
+++ b/packages/rs-platform-wallet-ffi/src/shielded_send.rs
@@ -314,7 +314,7 @@ pub unsafe extern "C" fn platform_wallet_manager_shielded_transfer(
         let r = wallet
             .shielded_transfer_to(&coordinator, account, &recipient, amount, memo, &prover)
             .await;
-        poke_sync_on_unconfirmed(&r, &coordinator);
+        poke_sync_on_unconfirmed(&r, handle);
         r
     });
     map_spend_result(result, "shielded transfer")
@@ -368,7 +368,7 @@ pub unsafe extern "C" fn platform_wallet_manager_shielded_unshield(
         let r = wallet
             .shielded_unshield_to(&coordinator, account, &to_addr_str, amount, &prover)
             .await;
-        poke_sync_on_unconfirmed(&r, &coordinator);
+        poke_sync_on_unconfirmed(&r, handle);
         r
     });
     map_spend_result(result, "shielded unshield")
@@ -426,7 +426,7 @@ pub unsafe extern "C" fn platform_wallet_manager_shielded_withdraw(
                 &prover,
             )
             .await;
-        poke_sync_on_unconfirmed(&r, &coordinator);
+        poke_sync_on_unconfirmed(&r, handle);
         r
     });
     map_spend_result(result, "shielded withdraw")
@@ -435,13 +435,23 @@ pub unsafe extern "C" fn platform_wallet_manager_shielded_withdraw(
 /// On the AMBIGUOUS outcome (broadcast accepted, result unconfirmed),
 /// kick an immediate forced shielded sync so the first re-drive check —
 /// nullifier re-check, then re-broadcast of the persisted transition —
-/// happens now instead of at the next background tick. Fire-and-forget:
-/// the spend's own result is already decided, and the sync pass owns
-/// resolution from here (`redrive_pending_spends` + the prune backstop).
-fn poke_sync_on_unconfirmed<T>(
-    result: &Result<T, PlatformWalletError>,
-    coordinator: &std::sync::Arc<platform_wallet::wallet::shielded::NetworkShieldedCoordinator>,
-) {
+/// happens now instead of at the next background tick.
+///
+/// Routed through the manager's [`ShieldedSyncManager::sync_now`] so the
+/// pass respects the same `is_syncing` CAS + `quiescing` drain barrier as
+/// the periodic loop and the host's Sync Now button — a raw
+/// `coordinator.sync(...)` here would race both. `force = true` bypasses
+/// only the caught-up cooldown, never the serialization gate. If a pass
+/// is already in flight the poke no-ops (empty summary) and the next
+/// tick's pass picks the redrive up — that pass's pre-scan snapshot may
+/// predate this arm, which is fine.
+///
+/// Fire-and-forget: the spend's own result is already decided and the
+/// sync pass owns resolution from here (`redrive_pending_spends` + the
+/// prune backstop); the pass outcome is logged, not surfaced.
+///
+/// [`ShieldedSyncManager::sync_now`]: platform_wallet::manager::shielded_sync::ShieldedSyncManager::sync_now
+fn poke_sync_on_unconfirmed<T>(result: &Result<T, PlatformWalletError>, handle: Handle) {
     let ambiguous = matches!(
         result,
         Err(PlatformWalletError::ShieldedSpendUnconfirmed { .. })
@@ -450,9 +460,23 @@ fn poke_sync_on_unconfirmed<T>(
     if !ambiguous {
         return;
     }
-    let coordinator = std::sync::Arc::clone(coordinator);
+    let Some(sync_manager) =
+        PLATFORM_WALLET_MANAGER_STORAGE.with_item(handle, |manager| manager.shielded_sync_arc())
+    else {
+        return;
+    };
     runtime().spawn(async move {
-        let _ = coordinator.sync(true).await;
+        let summary = sync_manager.sync_now(true).await;
+        if summary.sync_unix_seconds == 0 {
+            tracing::debug!(
+                "post-unconfirmed shielded sync poke skipped (a pass was already in flight                  or shielded is unconfigured); the next pass owns the re-drive"
+            );
+        } else {
+            tracing::debug!(
+                wallets = summary.wallet_results.len(),
+                "post-unconfirmed shielded sync pass completed"
+            );
+        }
     });
 }
 
@@ -642,7 +666,7 @@ pub unsafe extern "C" fn platform_wallet_manager_shielded_identity_create_from_p
                 &prover,
             )
             .await;
-        poke_sync_on_unconfirmed(&r, &coordinator);
+        poke_sync_on_unconfirmed(&r, handle);
         r
     });
 
--- a/packages/rs-platform-wallet-ffi/src/shielded_send.rs
+++ b/packages/rs-platform-wallet-ffi/src/shielded_send.rs
@@ -311,9 +311,11 @@ pub unsafe extern "C" fn platform_wallet_manager_shielded_transfer(
     // calling thread.
     let result = block_on_worker(async move {
         let prover = CachedOrchardProver::new();
-        wallet
+        let r = wallet
             .shielded_transfer_to(&coordinator, account, &recipient, amount, memo, &prover)
-            .await
+            .await;
+        poke_sync_on_unconfirmed(&r, &coordinator);
+        r
     });
     map_spend_result(result, "shielded transfer")
 }
@@ -363,9 +365,11 @@ pub unsafe extern "C" fn platform_wallet_manager_shielded_unshield(
 
     let result = block_on_worker(async move {
         let prover = CachedOrchardProver::new();
-        wallet
+        let r = wallet
             .shielded_unshield_to(&coordinator, account, &to_addr_str, amount, &prover)
-            .await
+            .await;
+        poke_sync_on_unconfirmed(&r, &coordinator);
+        r
     });
     map_spend_result(result, "shielded unshield")
 }
@@ -412,7 +416,7 @@ pub unsafe extern "C" fn platform_wallet_manager_shielded_withdraw(
 
     let result = block_on_worker(async move {
         let prover = CachedOrchardProver::new();
-        wallet
+        let r = wallet
             .shielded_withdraw_to(
                 &coordinator,
                 account,
@@ -421,9 +425,35 @@ pub unsafe extern "C" fn platform_wallet_manager_shielded_withdraw(
                 core_fee_per_byte,
                 &prover,
             )
-            .await
+            .await;
+        poke_sync_on_unconfirmed(&r, &coordinator);
+        r
     });
     map_spend_result(result, "shielded withdraw")
+}
+
+/// On the AMBIGUOUS outcome (broadcast accepted, result unconfirmed),
+/// kick an immediate forced shielded sync so the first re-drive check —
+/// nullifier re-check, then re-broadcast of the persisted transition —
+/// happens now instead of at the next background tick. Fire-and-forget:
+/// the spend's own result is already decided, and the sync pass owns
+/// resolution from here (`redrive_pending_spends` + the prune backstop).
+fn poke_sync_on_unconfirmed<T>(
+    result: &Result<T, PlatformWalletError>,
+    coordinator: &std::sync::Arc<platform_wallet::wallet::shielded::NetworkShieldedCoordinator>,
+) {
+    let ambiguous = matches!(
+        result,
+        Err(PlatformWalletError::ShieldedSpendUnconfirmed { .. })
+            | Err(PlatformWalletError::ShieldedBroadcastUnconfirmed { .. })
+    );
+    if !ambiguous {
+        return;
+    }
+    let coordinator = std::sync::Arc::clone(coordinator);
+    runtime().spawn(async move {
+        let _ = coordinator.sync(true).await;
+    });
 }
 
 /// Map a shielded operation outcome (shield / unshield / transfer /
@@ -600,7 +630,7 @@ pub unsafe extern "C" fn platform_wallet_manager_shielded_identity_create_from_p
         // `Signer<IdentityPublicKey>`.
         let identity_signer: &VTableSigner = &*(signer_identity_addr as *const VTableSigner);
         let prover = CachedOrchardProver::new();
-        wallet
+        let r = wallet
             .shielded_identity_create_from_pool(
                 &coordinator,
                 account,
@@ -611,7 +641,9 @@ pub unsafe extern "C" fn platform_wallet_manager_shielded_identity_create_from_p
                 identity_signer,
                 &prover,
             )
-            .await
+            .await;
+        poke_sync_on_unconfirmed(&r, &coordinator);
+        r
     });
 
     match result {
--- a/packages/rs-platform-wallet/src/wallet/shielded/coordinator.rs
+++ b/packages/rs-platform-wallet/src/wallet/shielded/coordinator.rs
@@ -655,20 +655,28 @@ impl NetworkShieldedCoordinator {
         // Runs before the balance read so freed notes are reflected in
         // this pass's balances.
         if let Some((snapshot, recorded)) = stranded_release {
-            {
+            // Snapshot the per-wallet persisters BEFORE the loop and drop
+            // the read guard: `redrive_pending_spends` performs network
+            // broadcasts, and holding the persisters lock across those
+            // awaits would block wallet register/unregister for the
+            // duration of the round trips.
+            let subwallet_persisters: Vec<(SubwalletId, Option<WalletPersister>)> = {
                 let persisters = self.persisters.read().await;
-                for (id, _) in &subwallets {
-                    let persister = persisters.get(&id.wallet_id).cloned();
-                    super::operations::redrive_pending_spends(
-                        &self.sdk,
-                        &self.store,
-                        persister.as_ref(),
-                        id.wallet_id,
-                        *id,
-                        &recorded,
-                    )
-                    .await;
-                }
+                subwallets
+                    .iter()
+                    .map(|(id, _)| (*id, persisters.get(&id.wallet_id).cloned()))
+                    .collect()
+            };
+            for (id, persister) in &subwallet_persisters {
+                super::operations::redrive_pending_spends(
+                    &self.sdk,
+                    &self.store,
+                    persister.as_ref(),
+                    id.wallet_id,
+                    *id,
+                    &recorded,
+                )
+                .await;
             }
             self.release_stranded_spends(snapshot, &recorded).await;
         }
--- a/packages/rs-platform-wallet/src/wallet/shielded/coordinator.rs
+++ b/packages/rs-platform-wallet/src/wallet/shielded/coordinator.rs
@@ -638,13 +638,38 @@ impl NetworkShieldedCoordinator {
         // `notes` result and `notes.changeset` the receipts do.
         let newly_spent_per_sub = notes.per_subwallet_newly_spent.clone();
 
-        // Residual-spend reconcile: `sync_notes_across` above marked every
-        // landed spend (clearing its reservation). Now release any still-
-        // pending pre-scan reservation whose recorded anchor Platform had
-        // already pruned before the scan — a spend broadcast-accepted but
-        // never landed, otherwise stranded for the session. Runs before the
-        // balance read so freed notes are reflected in this pass's balances.
+        // Residual-spend resolution: `sync_notes_across` above marked every
+        // landed spend (clearing its reservation and dropping its redrive
+        // record via the store hook). Two passes over what's left, both
+        // judged against the PRE-scan recorded-anchor set:
+        //
+        // 1. Re-drive — for each armed unconfirmed spend whose anchor is
+        //    still recorded, re-broadcast the stored byte-identical
+        //    transition (bounded by MAX_REDRIVE_ATTEMPTS) to actively
+        //    resolve the ambiguity instead of waiting out the retention
+        //    window.
+        // 2. Prune backstop — release any still-pending pre-scan
+        //    reservation whose anchor was already pruned (the spend can
+        //    never execute).
+        //
+        // Runs before the balance read so freed notes are reflected in
+        // this pass's balances.
         if let Some((snapshot, recorded)) = stranded_release {
+            {
+                let persisters = self.persisters.read().await;
+                for (id, _) in &subwallets {
+                    let persister = persisters.get(&id.wallet_id).cloned();
+                    super::operations::redrive_pending_spends(
+                        &self.sdk,
+                        &self.store,
+                        persister.as_ref(),
+                        id.wallet_id,
+                        *id,
+                        &recorded,
+                    )
+                    .await;
+                }
+            }
             self.release_stranded_spends(snapshot, &recorded).await;
         }
 
--- a/packages/rs-platform-wallet/src/wallet/shielded/file_store.rs
+++ b/packages/rs-platform-wallet/src/wallet/shielded/file_store.rs
@@ -20,8 +20,8 @@ use std::sync::Mutex;
 use grovedb_commitment_tree::{ClientPersistentCommitmentTree, Position, Retention};
 
 use super::store::{
-    ShieldedNote, ShieldedOutgoingNote, ShieldedStore, StalePendingSpend, SubwalletId,
-    SubwalletState,
+    PendingRedrive, ShieldedNote, ShieldedOutgoingNote, ShieldedStore, StalePendingSpend,
+    SubwalletId, SubwalletState,
 };
 use crate::wallet::platform_wallet::WalletId;
 
@@ -66,6 +66,13 @@ pub struct FileBackedShieldedStore {
     /// Per-subwallet notes + sync state, keyed by `(wallet_id,
     /// account_index)`. Lazily populated on first use of an id.
     subwallets: BTreeMap<SubwalletId, SubwalletState>,
+    /// Second connection on the same SQLite file, owning the
+    /// `shielded_pending_spends` table (armed [`PendingRedrive`]
+    /// records). Separate from `tree` because the commitment-tree
+    /// wrapper takes its `Connection` by value; WAL mode makes the
+    /// two-connection setup safe. `Mutex` for the same `Sync`-shim
+    /// reason as `tree`.
+    pending_conn: Mutex<rusqlite::Connection>,
 }
 
 impl FileBackedShieldedStore {
@@ -95,12 +102,94 @@ impl FileBackedShieldedStore {
         let conn = Self::open_tuned_connection(&path)?;
         let tree = ClientPersistentCommitmentTree::open(conn, max_checkpoints)
             .map_err(|e| FileShieldedStoreError(format!("open commitment tree: {e}")))?;
-        Ok(Self {
+        let pending_conn = Self::open_tuned_connection(&path)?;
+        pending_conn
+            .execute(
+                "CREATE TABLE IF NOT EXISTS shielded_pending_spends (
+                    wallet_id     BLOB    NOT NULL,
+                    account_index INTEGER NOT NULL,
+                    activity_id   BLOB    NOT NULL,
+                    anchor        BLOB    NOT NULL,
+                    nullifiers    BLOB    NOT NULL,
+                    st_bytes      BLOB    NOT NULL,
+                    attempts      INTEGER NOT NULL DEFAULT 0,
+                    PRIMARY KEY (wallet_id, account_index, activity_id)
+                )",
+                [],
+            )
+            .map_err(|e| FileShieldedStoreError(format!("create pending_spends table: {e}")))?;
+        let mut store = Self {
             tree: Mutex::new(tree),
             path,
             max_checkpoints,
             subwallets: BTreeMap::new(),
-        })
+            pending_conn: Mutex::new(pending_conn),
+        };
+        store.rehydrate_pending_spends()?;
+        Ok(store)
+    }
+
+    /// Reload every persisted [`PendingRedrive`] into the in-memory
+    /// per-subwallet state, re-arming both the redrive record and the
+    /// note reservations its nullifiers carry — an unconfirmed
+    /// broadcast therefore keeps its notes reserved (and its re-drive
+    /// alive) across restarts. Corrupt rows are dropped with a warning
+    /// rather than failing the open.
+    fn rehydrate_pending_spends(&mut self) -> Result<(), FileShieldedStoreError> {
+        let conn = self.pending_conn.lock().expect("pending_conn mutex");
+        let mut stmt = conn
+            .prepare(
+                "SELECT wallet_id, account_index, activity_id, anchor, nullifiers, st_bytes, \
+                 attempts FROM shielded_pending_spends",
+            )
+            .map_err(|e| FileShieldedStoreError(format!("prepare rehydrate: {e}")))?;
+        let rows = stmt
+            .query_map([], |row| {
+                Ok((
+                    row.get::<_, Vec<u8>>(0)?,
+                    row.get::<_, u32>(1)?,
+                    row.get::<_, Vec<u8>>(2)?,
+                    row.get::<_, Vec<u8>>(3)?,
+                    row.get::<_, Vec<u8>>(4)?,
+                    row.get::<_, Vec<u8>>(5)?,
+                    row.get::<_, u32>(6)?,
+                ))
+            })
+            .map_err(|e| FileShieldedStoreError(format!("query rehydrate: {e}")))?;
+        for row in rows {
+            let (wallet_id, account_index, activity_id, anchor, nullifiers, st_bytes, attempts) =
+                row.map_err(|e| FileShieldedStoreError(format!("read rehydrate row: {e}")))?;
+            let (Ok(wallet_id), Ok(activity_id), Ok(anchor)) = (
+                <[u8; 32]>::try_from(wallet_id.as_slice()),
+                <[u8; 32]>::try_from(activity_id.as_slice()),
+                <[u8; 32]>::try_from(anchor.as_slice()),
+            ) else {
+                tracing::warn!("dropping corrupt shielded_pending_spends row (bad key widths)");
+                continue;
+            };
+            if nullifiers.is_empty() || nullifiers.len() % 32 != 0 {
+                tracing::warn!("dropping corrupt shielded_pending_spends row (bad nullifiers)");
+                continue;
+            }
+            let nullifiers: Vec<[u8; 32]> = nullifiers
+                .chunks_exact(32)
+                .map(|c| <[u8; 32]>::try_from(c).expect("chunks_exact(32)"))
+                .collect();
+            let id = SubwalletId::new(wallet_id, account_index);
+            let sw = self.subwallets.entry(id).or_default();
+            for n in &nullifiers {
+                sw.mark_pending(n);
+                sw.set_pending_spend(n, anchor, activity_id);
+            }
+            sw.arm_redrive(PendingRedrive {
+                activity_id,
+                anchor,
+                nullifiers,
+                st_bytes,
+                attempts,
+            });
+        }
+        Ok(())
     }
 
     /// Open a `rusqlite::Connection` on `path` with the same WAL /
@@ -129,6 +218,47 @@ impl FileBackedShieldedStore {
         }
         Ok(conn)
     }
+
+    /// Mirror to SQLite the redrive deletions [`SubwalletState`] performs
+    /// in memory when a nullifier resolves (`mark_spent` /
+    /// `clear_pending`): delete every persisted row for `id` whose
+    /// nullifier blob contains `nullifier`.
+    fn delete_redrive_rows_containing(
+        &self,
+        id: SubwalletId,
+        nullifier: &[u8; 32],
+    ) -> Result<(), FileShieldedStoreError> {
+        let conn = self.pending_conn.lock().expect("pending_conn mutex");
+        let mut stmt = conn
+            .prepare(
+                "SELECT activity_id, nullifiers FROM shielded_pending_spends \
+                 WHERE wallet_id = ?1 AND account_index = ?2",
+            )
+            .map_err(|e| FileShieldedStoreError(format!("prepare redrive lookup: {e}")))?;
+        let rows: Vec<(Vec<u8>, Vec<u8>)> = stmt
+            .query_map(
+                rusqlite::params![id.wallet_id.as_slice(), id.account_index],
+                |row| Ok((row.get::<_, Vec<u8>>(0)?, row.get::<_, Vec<u8>>(1)?)),
+            )
+            .map_err(|e| FileShieldedStoreError(format!("query redrive lookup: {e}")))?
+            .collect::<Result<_, _>>()
+            .map_err(|e| FileShieldedStoreError(format!("read redrive lookup: {e}")))?;
+        drop(stmt);
+        for (activity_id, nullifiers) in rows {
+            if nullifiers
+                .chunks_exact(32)
+                .any(|c| c == nullifier.as_slice())
+            {
+                conn.execute(
+                    "DELETE FROM shielded_pending_spends \
+                     WHERE wallet_id = ?1 AND account_index = ?2 AND activity_id = ?3",
+                    rusqlite::params![id.wallet_id.as_slice(), id.account_index, activity_id],
+                )
+                .map_err(|e| FileShieldedStoreError(format!("delete redrive row: {e}")))?;
+            }
+        }
+        Ok(())
+    }
 }
 
 impl ShieldedStore for FileBackedShieldedStore {
@@ -156,11 +286,16 @@ impl ShieldedStore for FileBackedShieldedStore {
     }
 
     fn mark_spent(&mut self, id: SubwalletId, nullifier: &[u8; 32]) -> Result<bool, Self::Error> {
-        Ok(self
-            .subwallets
-            .get_mut(&id)
-            .map(|sw| sw.mark_spent(nullifier))
-            .unwrap_or(false))
+        let Some(sw) = self.subwallets.get_mut(&id) else {
+            return Ok(false);
+        };
+        let marked = sw.mark_spent(nullifier);
+        if marked {
+            // `SubwalletState::mark_spent` dropped any redrive carrying
+            // this nullifier from memory; mirror the deletions.
+            self.delete_redrive_rows_containing(id, nullifier)?;
+        }
+        Ok(marked)
     }
 
     fn mark_pending(&mut self, id: SubwalletId, nullifier: &[u8; 32]) -> Result<bool, Self::Error> {
@@ -176,11 +311,14 @@ impl ShieldedStore for FileBackedShieldedStore {
         id: SubwalletId,
         nullifier: &[u8; 32],
     ) -> Result<bool, Self::Error> {
-        Ok(self
-            .subwallets
-            .get_mut(&id)
-            .map(|sw| sw.clear_pending(nullifier))
-            .unwrap_or(false))
+        let Some(sw) = self.subwallets.get_mut(&id) else {
+            return Ok(false);
+        };
+        let removed = sw.clear_pending(nullifier);
+        if removed {
+            self.delete_redrive_rows_containing(id, nullifier)?;
+        }
+        Ok(removed)
     }
 
     fn set_pending_spend(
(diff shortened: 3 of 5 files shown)


thepastaclaw:
✅ Review complete (commit 997d9e77)

packages/rs-platform-wallet-ffi/src/shielded_send.rs
```diff
@@ -421,11 +425,37 @@ pub unsafe extern "C" fn platform_wallet_manager_shielded_withdraw(
...
+            | Err(PlatformWalletError::ShieldedBroadcastUnconfirmed { .. })
+    );
+    if !ambiguous {
+        return;
+    }
+    let coordinator = std::sync::Arc::clone(coordinator);
+    runtime().spawn(async move {
+        let _ = coordinator.sync(true).await;
+    });
+}
```
thepastaclaw: **🔴 Blocking: poke_sync_on_unconfirmed bypasses the manager's is_syncing/quiescing gate**

`poke_sync_on_unconfirmed` spawns `coordinator.sync(true).await` directly against the same `Arc<NetworkShieldedCoordinator>` used by the periodic loop and `sync_wallet`/`sync_now`. Every other in-tree caller routes through `ShieldedSyncManager::sync_now` (packages/rs-platform-wallet/src/manager/shielded_sync.rs:336-395), which CAS-guards on `is_syncing` and checks `quiescing` before invoking `coordinator.sync()`. `NetworkShieldedCoordinator::sync()` itself (coordinator.rs:557) has no internal reentrancy guard — the `is_syncing` atomic is the only serializer — and `force=true` also skips the cooldown check.

Two concrete consequences: (1) a background pass already in flight when a spend goes ambiguous will run a second, fully concurrent `sync()` on the same store; both take the same pre-scan `stale_pending_spends`+`recorded` snapshot and both call `redrive_pending_spends` on the same subwallet, letting each rebroadcast the stored transition and independently bump `attempts`, so one ambiguous event can burn 2 of the 3 redrive attempts in a single race. (2) It breaks the documented invariant of `WalletManager::quiesce()` (shielded_sync.rs:299-323): because this spawn never sets `is_syncing`, the `while self.is_syncing.load(...) {}` drain loop will not wait for it, so a Clear/unregister/rebind that follows `quiesce()` can still race a resolving sync pass triggered by this poke. Fund-safety is preserved (identical nullifiers can't double-spend and the duplicate hits `NullifierAlreadySpent`), but the retry-budget intent and the teardown barrier are not. Route this through the manager's gated `sync_now`, or add an in-flight mutex/atomic inside `NetworkShieldedCoordinator::sync` now that it has two independent callers.

<sub>source: ['opus', 'sonnet5']</sub>

packages/rs-platform-wallet/src/wallet/shielded/file_store.rs
```diff
@@ -204,6 +342,87 @@ impl ShieldedStore for FileBackedShieldedStore {
...
+                rusqlite::params![
+                    id.wallet_id.as_slice(),
+                    id.account_index,
+                    activity_id.as_slice(),
+                    attempts,
+                ],
+            )
+            .map_err(|e| FileShieldedStoreError(format!("bump redrive attempts: {e}")))?;
+        }
+        Ok(attempts)
```
thepastaclaw, line 436: **🟡 Suggestion: bump_redrive_attempts mutates memory before persisting, so a SQLite failure diverges the two views**

`FileBackedShieldedStore::bump_redrive_attempts` calls `sw.bump_redrive_attempts(activity_id)` — which unconditionally increments in-memory `SubwalletState::redrives[activity_id].attempts` — before issuing the `UPDATE shielded_pending_spends SET attempts = ?4`. If the UPDATE fails the function returns `Err`, but the in-memory counter has already advanced. On the next reopen `rehydrate_pending_spends` rebuilds from the stale SQLite value, so the visible attempt count silently rewinds. The caller in `redrive_pending_spends` (operations.rs:2156-2160 and 2203-2207) additionally does `.unwrap_or(0)`, folding the persistence failure into an `attempts=0` log line and continuing. `arm_redrive` already orders SQLite before memory — mirror that here (read/UPDATE against SQLite first, then mirror the resulting count to memory) so the two views cannot diverge across restarts.

<sub>source: ['opus']</sub>

packages/rs-platform-wallet/src/wallet/shielded/file_store.rs
```diff
@@ -129,6 +218,47 @@ impl FileBackedShieldedStore {
         }
         Ok(conn)
     }
```
thepastaclaw, line 226: **🟡 Suggestion: Second SQLite connection on the same file lacks a busy_timeout**

The PR opens a second `rusqlite::Connection` (`pending_conn`) via `open_tuned_connection` against the same file as `tree`. Under WAL, readers do not block writers, but two writer connections still serialize on the write lock and a concurrent writer receives `SQLITE_BUSY` immediately unless `busy_timeout` (or `PRAGMA busy_timeout=...`) is configured. `open_tuned_connection` sets `journal_mode=WAL`, `synchronous=NORMAL`, `temp_store=MEMORY`, but installs no busy handler. On iOS under concurrent activity (a `mark_spent`/tree write while `arm_redrive` or `bump_redrive_attempts` writes to `pending_conn`) an unlucky ordering surfaces as `FileShieldedStoreError("persist redrive: database is locked")`, either propagating up as a hard `arm_redrive` failure (silently disabling redrive persistence for that spend so only the prune backstop rescues it) or aborting the operation entirely. Set a modest `busy_timeout` (a few seconds) on both connections so SQLite retries transparently.

<sub>source: ['opus']</sub>

packages/rs-platform-wallet/src/wallet/shielded/operations.rs
```diff
@@ -1936,6 +1981,241 @@ async fn arm_pending_release<S: ShieldedStore>(
...
+                    max = MAX_REDRIVE_ATTEMPTS,
+                    "redrive: re-broadcast accepted; the next scan detects the landing"
+                );
+            }
+            Err(e) if is_nullifier_already_spent(&e) => {
+                info!(
+                    "redrive: transition already executed on-chain; the next scan confirms \
+                     the notes spent"
+                );
+            }
```
thepastaclaw: **🟡 Suggestion: NullifierAlreadySpent arm does not bound retries or clear the record**

The `Err(e) if is_nullifier_already_spent(&e)` arm in `redrive_pending_spends` only logs — it does not `bump_redrive_attempts` and does not `clear_redrive`. The comment ("the next scan confirms") assumes the following scan will land the nullifier via `apply_scanned_nullifier_spends` → `mark_spent`, which drops the record. That is the common path, but the arm is otherwise unbounded: if the scan repeatedly fails to cover the block that landed the nullifier (transport hiccup during a chunk, a subwallet whose store lost the note, delivery lag past cooldowns), every subsequent sync will re-broadcast the same transition and receive `NullifierAlreadySpent` again indefinitely. Since this arm is already a definitive success signal (`NullifierAlreadySpent` proves the transition executed), the record could safely be cleared here — leaving the reservation in place for the next scan's `mark_spent` to release. At minimum, count the attempt against `MAX_REDRIVE_ATTEMPTS` so the retry loop is bounded even in the pathological scan-gap case.

<sub>source: ['opus']</sub>

packages/rs-platform-wallet/src/wallet/shielded/file_store.rs
```diff
@@ -176,11 +311,14 @@ impl ShieldedStore for FileBackedShieldedStore {
...
-            .unwrap_or(false))
+        let Some(sw) = self.subwallets.get_mut(&id) else {
+            return Ok(false);
+        };
+        let removed = sw.clear_pending(nullifier);
+        if removed {
+            self.delete_redrive_rows_containing(id, nullifier)?;
+        }
+        Ok(removed)
     }
```
thepastaclaw, line 346: **🟡 Suggestion: delete_redrive_rows_containing error can abort mark_spent/clear_pending after the in-memory drop already happened**

Both `mark_spent` (line 296) and `clear_pending` (line 319) call `self.delete_redrive_rows_containing(id, nullifier)?` after the in-memory `SubwalletState::mark_spent`/`clear_pending` mutation already ran and already dropped the in-memory `redrives` entry. If the SQL delete fails (WAL contention between the two connections on the same file, disk error), the function returns `Err` even though the in-memory state has already changed — the persisted `shielded_pending_spends` row is left behind, desynced from memory.

Concretely: `mark_spent` is called from `apply_scanned_nullifier_spends` as `store.mark_spent(*id, nf)?`, and this `?` now aborts detection of all remaining scanned nullifier spends across every subwallet for the pass — a new failure mode, since pre-PR `mark_spent` only touched the in-memory map and could not fail in practice. On `clear_pending` SQL failure the in-memory reservation is already freed, but the stale SQLite redrive row survives; on restart `rehydrate_pending_spends` resurrects it and re-reserves the (legitimately released) nullifier, locking the note back up until a later pass notices. Perform the SQL delete before mutating memory, or catch the persistence error and log without propagating so memory and SQLite cannot desync mid-call.

<sub>source: ['sonnet5']</sub>

packages/rs-platform-wallet/src/wallet/shielded/coordinator.rs
```diff
@@ -638,13 +638,38 @@ impl NetworkShieldedCoordinator {
...
+                        &self.store,
+                        persister.as_ref(),
+                        id.wallet_id,
+                        *id,
+                        &recorded,
+                    )
+                    .await;
+                }
+            }
             self.release_stranded_spends(snapshot, &recorded).await;
```
thepastaclaw, line 681: **🟡 Suggestion: persisters read lock held across the entire network-calling redrive loop**

The new redrive block acquires `self.persisters.read().await` and holds it for the full `for (id, _) in &subwallets` loop, and each iteration awaits `redrive_pending_spends`, which performs up to `MAX_REDRIVE_ATTEMPTS` real `st.broadcast(sdk, None).await` round-trips per subwallet. A `tokio::sync::RwLock` read guard blocks any writer for its full lifetime, so `register_wallet`/`unregister_wallet`/`clear` (all of which take `persisters.write().await`) will stall for the duration of this pass any time a host tries to add/remove a wallet while a redrive-heavy sync is running. The pre-existing `release_stranded_spends` uses the same locking pattern but only does local store work; extending that pattern across network I/O changes the assumption. Snapshot the needed `WalletPersister` clones under the lock, drop the guard, then run the network-calling loop.

```suggestion
            {
                let redrive_targets: Vec<(SubwalletId, Option<crate::wallet::persister::WalletPersister>)> = {
                    let persisters = self.persisters.read().await;
                    subwallets
                        .iter()
                        .map(|(id, _)| (*id, persisters.get(&id.wallet_id).cloned()))
                        .collect()
                };
                for (id, persister) in redrive_targets {
                    super::operations::redrive_pending_spends(
                        &self.sdk,
                        &self.store,
                        persister.as_ref(),
                        id.wallet_id,
                        id,
                        &recorded,
                    )
                    .await;
                }
            }
```

<sub>source: ['sonnet5']</sub>

packages/rs-platform-wallet/src/wallet/shielded/operations.rs
```diff
@@ -1936,6 +1981,241 @@ async fn arm_pending_release<S: ShieldedStore>(
...
+            Err(e) => {
+                // `AlreadyExists` (still in a mempool after a lost-ACK
+                // retry) or transport noise: inconclusive; counts toward
+                // the cap.
+                let attempts = store
+                    .write()
+                    .await
+                    .bump_redrive_attempts(id, &redrive.activity_id)
+                    .unwrap_or(0);
```
thepastaclaw: **🟡 Suggestion: bump_redrive_attempts persistence errors silently discarded at both call sites**

Both `bump_redrive_attempts` invocations in `redrive_pending_spends` (the `Ok(())` arm at 2156-2160 and the inconclusive-error arm at 2203-2207) do `.unwrap_or(0)` on the Result, swallowing any store-persistence error entirely — no `warn!`/`debug!` is logged. Every other store-call failure in the same function (`pending_redrives`, `clear_redrive`, `clear_pending`) logs a `warn!` on error; these two are the odd ones out. Combined with the memory-before-SQLite ordering in the file store, this converts a SQLite failure into a silent stale-count on restart with no diagnostic trail. Log the error the same way the sibling arms do.

```suggestion
                let attempts = match store
                    .write()
                    .await
                    .bump_redrive_attempts(id, &redrive.activity_id)
                {
                    Ok(a) => a,
                    Err(e) => {
                        warn!(error = %e, "redrive: bump_redrive_attempts failed");
                        0
                    }
                };
```

<sub>source: ['sonnet5']</sub>

packages/rs-platform-wallet-ffi/src/shielded_send.rs
```diff
@@ -421,11 +425,37 @@ pub unsafe extern "C" fn platform_wallet_manager_shielded_withdraw(
...
+        Err(PlatformWalletError::ShieldedSpendUnconfirmed { .. })
+            | Err(PlatformWalletError::ShieldedBroadcastUnconfirmed { .. })
+    );
+    if !ambiguous {
+        return;
+    }
+    let coordinator = std::sync::Arc::clone(coordinator);
+    runtime().spawn(async move {
+        let _ = coordinator.sync(true).await;
+    });
```
thepastaclaw, line 480: **💬 Nitpick: poke_sync_on_unconfirmed silently discards the forced-sync result**

`let _ = coordinator.sync(true).await;` throws away both errors and any panic captured in the (immediately dropped) `JoinHandle`. Discarding the typed spend result is fine — the ambiguity is already decided — but discarding the log makes the very case this PR is trying to accelerate (the first re-check after an ambiguous outcome) invisible when it fails to run. A `warn!` on `Err` keeps the fire-and-forget shape while making a stuck redrive diagnosable.

<sub>source: ['opus']</sub>

packages/rs-platform-wallet/src/wallet/shielded/operations.rs
```diff
@@ -1936,6 +1981,241 @@ async fn arm_pending_release<S: ShieldedStore>(
...
+                debug!(
+                    attempts,
+                    max = MAX_REDRIVE_ATTEMPTS,
+                    error = %e,
+                    "redrive: re-broadcast inconclusive"
+                );
+            }
+        }
+    }
```
thepastaclaw, line 2272: **💬 Nitpick: redrive_pending_spends' actual broadcast-response classification is untested**

The only test that exercises `redrive_pending_spends` (`redrive_skips_pruned_and_exhausted_and_drops_garbage`, line 2499) deliberately never reaches `st.broadcast(sdk, None).await` — its three records are pruned-anchor, exhausted-attempts, and corrupt-bytes, all of which short-circuit before broadcast, and the test explicitly asserts "no attempt counters were bumped — nothing touched the network." `is_nullifier_already_spent` and `carries_consensus_rejection` are unit-tested in isolation, but the integration inside `redrive_pending_spends` — the `match st.broadcast(...).await` arms that bump attempts, release reservations via `clear_pending`, and flip the activity row to `Failed` on a definitive rejection — has no coverage. Given this path releases fund reservations and flips activity status on network response classification, an integration test against a stub `broadcast()` for each of accepted / already-spent / definitive-rejection / inconclusive would meaningfully de-risk it.

<sub>source: ['sonnet5']</sub>

thepastaclaw (changes requested):
## Code Review

Sound active-redrive lifecycle with fund-safe byte-identical rebroadcast, correct `NullifierAlreadySpent`-first classification ordering, and coherent SQLite/rehydrate persistence. The main defect is the new FFI-triggered `poke_sync_on_unconfirmed` bypasses the `ShieldedSyncManager` `is_syncing`/`quiescing` gate that the rest of the codebase relies on as a hard invariant, allowing concurrent sync passes to race the periodic loop and undermine `quiesce()`'s drain barrier. Secondary concerns cluster around SQLite consistency in `file_store.rs` (memory-before-persist ordering in `bump_redrive_attempts`, missing `busy_timeout` on the second writer connection, `?`-abort desync in `mark_spent`/`clear_pending`), plus an unbounded `NullifierAlreadySpent` success arm and a network-blocking `persisters` read-lock scope.

Source: reviewers: `opus` (general), `claude-sonnet-5` (general), `gpt-5.5[high]` (general, failed_or_unparseable), `opus` (rust-quality), `claude-sonnet-5` (rust-quality), `gpt-5.5[high]` (rust-quality, failed_or_unparseable), `opus` (ffi-engineer), `claude-sonnet-5` (ffi-engineer), `gpt-5.5[high]` (ffi-engineer, failed_or_unparseable); verifier: `opus`; specialists: `rust-quality`, `ffi-engineer`

🔴 1 blocking | 🟡 6 suggestion(s) | 💬 2 nitpick(s)

<details>

```

- [BLOCKING] packages/rs-platform-wallet-ffi/src/shielded_send.rs:441-457: poke_sync_on_unconfirmed bypasses the manager's is_syncing/quiescing gate
  `poke_sync_on_unconfirmed` spawns `coordinator.sync(true).await` directly against the same `Arc<NetworkShieldedCoordinator>` used by the periodic loop and `sync_wallet`/`sync_now`. Every other in-tree caller routes through `ShieldedSyncManager::sync_now` (packages/rs-platform-wallet/src/manager/shielded_sync.rs:336-395), which CAS-guards on `is_syncing` and checks `quiescing` before invoking `coordinator.sync()`. `NetworkShieldedCoordinator::sync()` itself (coordinator.rs:557) has no internal reentrancy guard — the `is_syncing` atomic is the only serializer — and `force=true` also skips the cooldown check.

  Two concrete consequences: (1) a background pass already in flight when a spend goes ambiguous will run a second, fully concurrent `sync()` on the same store; both take the same pre-scan `stale_pending_spends`+`recorded` snapshot and both call `redrive_pending_spends` on the same subwallet, letting each rebroadcast the stored transition and independently bump `attempts`, so one ambiguous event can burn 2 of the 3 redrive attempts in a single race. (2) It breaks the documented invariant of `WalletManager::quiesce()` (shielded_sync.rs:299-323): because this spawn never sets `is_syncing`, the `while self.is_syncing.load(...) {}` drain loop will not wait for it, so a Clear/unregister/rebind that follows `quiesce()` can still race a resolving sync pass triggered by this poke. Fund-safety is preserved (identical nullifiers can't double-spend and the duplicate hits `NullifierAlreadySpent`), but the retry-budget intent and the teardown barrier are not. Route this through the manager's gated `sync_now`, or add an in-flight mutex/atomic inside `NetworkShieldedCoordinator::sync` now that it has two independent callers.

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/file_store.rs:377-401: bump_redrive_attempts mutates memory before persisting, so a SQLite failure diverges the two views
  `FileBackedShieldedStore::bump_redrive_attempts` calls `sw.bump_redrive_attempts(activity_id)` — which unconditionally increments in-memory `SubwalletState::redrives[activity_id].attempts` — before issuing the `UPDATE shielded_pending_spends SET attempts = ?4`. If the UPDATE fails the function returns `Err`, but the in-memory counter has already advanced. On the next reopen `rehydrate_pending_spends` rebuilds from the stale SQLite value, so the visible attempt count silently rewinds. The caller in `redrive_pending_spends` (operations.rs:2156-2160 and 2203-2207) additionally does `.unwrap_or(0)`, folding the persistence failure into an `attempts=0` log line and continuing. `arm_redrive` already orders SQLite before memory — mirror that here (read/UPDATE against SQLite first, then mirror the resulting count to memory) so the two views cannot diverge across restarts.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/file_store.rs:105-220: Second SQLite connection on the same file lacks a busy_timeout
  The PR opens a second `rusqlite::Connection` (`pending_conn`) via `open_tuned_connection` against the same file as `tree`. Under WAL, readers do not block writers, but two writer connections still serialize on the write lock and a concurrent writer receives `SQLITE_BUSY` immediately unless `busy_timeout` (or `PRAGMA busy_timeout=...`) is configured. `open_tuned_connection` sets `journal_mode=WAL`, `synchronous=NORMAL`, `temp_store=MEMORY`, but installs no busy handler. On iOS under concurrent activity (a `mark_spent`/tree write while `arm_redrive` or `bump_redrive_attempts` writes to `pending_conn`) an unlucky ordering surfaces as `FileShieldedStoreError("persist redrive: database is locked")`, either propagating up as a hard `arm_redrive` failure (silently disabling redrive persistence for that spend so only the prune backstop rescues it) or aborting the operation entirely. Set a modest `busy_timeout` (a few seconds) on both connections so SQLite retries transparently.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/file_store.rs:288-322: delete_redrive_rows_containing error can abort mark_spent/clear_pending after the in-memory drop already happened
  Both `mark_spent` (line 296) and `clear_pending` (line 319) call `self.delete_redrive_rows_containing(id, nullifier)?` after the in-memory `SubwalletState::mark_spent`/`clear_pending` mutation already ran and already dropped the in-memory `redrives` entry. If the SQL delete fails (WAL contention between the two connections on the same file, disk error), the function returns `Err` even though the in-memory state has already changed — the persisted `shielded_pending_spends` row is left behind, desynced from memory.

  Concretely: `mark_spent` is called from `apply_scanned_nullifier_spends` as `store.mark_spent(*id, nf)?`, and this `?` now aborts detection of all remaining scanned nullifier spends across every subwallet for the pass — a new failure mode, since pre-PR `mark_spent` only touched the in-memory map and could not fail in practice. On `clear_pending` SQL failure the in-memory reservation is already freed, but the stale SQLite redrive row survives; on restart `rehydrate_pending_spends` resurrects it and re-reserves the (legitimately released) nullifier, locking the note back up until a later pass notices. Perform the SQL delete before mutating memory, or catch the persistence error and log without propagating so memory and SQLite cannot desync mid-call.

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/operations.rs:2167-2172: NullifierAlreadySpent arm does not bound retries or clear the record
  The `Err(e) if is_nullifier_already_spent(&e)` arm in `redrive_pending_spends` only logs — it does not `bump_redrive_attempts` and does not `clear_redrive`. The comment ("the next scan confirms") assumes the following scan will land the nullifier via `apply_scanned_nullifier_spends` → `mark_spent`, which drops the record. That is the common path, but the arm is otherwise unbounded: if the scan repeatedly fails to cover the block that landed the nullifier (transport hiccup during a chunk, a subwallet whose store lost the note, delivery lag past cooldowns), every subsequent sync will re-broadcast the same transition and receive `NullifierAlreadySpent` again indefinitely. Since this arm is already a definitive success signal (`NullifierAlreadySpent` proves the transition executed), the record could safely be cleared here — leaving the reservation in place for the next scan's `mark_spent` to release. At minimum, count the attempt against `MAX_REDRIVE_ATTEMPTS` so the retry loop is bounded even in the pathological scan-gap case.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/operations.rs:2156-2207: bump_redrive_attempts persistence errors silently discarded at both call sites
  Both `bump_redrive_attempts` invocations in `redrive_pending_spends` (the `Ok(())` arm at 2156-2160 and the inconclusive-error arm at 2203-2207) do `.unwrap_or(0)` on the Result, swallowing any store-persistence error entirely — no `warn!`/`debug!` is logged. Every other store-call failure in the same function (`pending_redrives`, `clear_redrive`, `clear_pending`) logs a `warn!` on error; these two are the odd ones out. Combined with the memory-before-SQLite ordering in the file store, this converts a SQLite failure into a silent stale-count on restart with no diagnostic trail. Log the error the same way the sibling arms do.

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/coordinator.rs:657-673: persisters read lock held across the entire network-calling redrive loop
  The new redrive block acquires `self.persisters.read().await` and holds it for the full `for (id, _) in &subwallets` loop, and each iteration awaits `redrive_pending_spends`, which performs up to `MAX_REDRIVE_ATTEMPTS` real `st.broadcast(sdk, None).await` round-trips per subwallet. A `tokio::sync::RwLock` read guard blocks any writer for its full lifetime, so `register_wallet`/`unregister_wallet`/`clear` (all of which take `persisters.write().await`) will stall for the duration of this pass any time a host tries to add/remove a wallet while a redrive-heavy sync is running. The pre-existing `release_stranded_spends` uses the same locking pattern but only does local store work; extending that pattern across network I/O changes the assumption. Snapshot the needed `WalletPersister` clones under the lock, drop the guard, then run the network-calling loop.
```
</details>

thepastaclaw (changes requested):
## Code Review

Pure rebase onto v4.0-dev — the five shielded files under review (shielded_send.rs, coordinator.rs, file_store.rs, operations.rs, store.rs) are byte-identical to prior reviewed SHA da6655ba (git diff confirms zero changes). None of the 9 prior findings have been addressed; all are carried forward as STILL VALID. The blocking concern remains that `poke_sync_on_unconfirmed` bypasses the `ShieldedSyncManager` is_syncing/quiescing gate that every other in-tree sync trigger relies on.

Source: reviewers: `opus` (general), `claude-sonnet-5` (general), `gpt-5.5[high]` (general, failed_or_unparseable), `opus` (rust-quality), `claude-sonnet-5` (rust-quality), `gpt-5.5[high]` (rust-quality, failed_or_unparseable), `opus` (ffi-engineer), `claude-sonnet-5` (ffi-engineer), `gpt-5.5[high]` (ffi-engineer, failed_or_unparseable); verifier: `opus`; specialists: `rust-quality`, `ffi-engineer`

🔴 1 blocking | 🟡 6 suggestion(s) | 💬 2 nitpick(s)

<details>
<summary>Carried-forward findings already raised (9)</summary>

- **[BLOCKING] (deduped existing open thread) `packages/rs-platform-wallet-ffi/src/shielded_send.rs:441-457`: poke_sync_on_unconfirmed bypasses the manager's is_syncing/quiescing gate** — Carried forward from prior review (prior-1); verified byte-identical at head 5a9c704358. `poke_sync_on_unconfirmed` spawns `coordinator.sync(true).await` directly against the shared `Arc<NetworkShieldedCoordinator>`, bypassing `ShieldedSyncManager::sync_now` (packages/rs-platform-wallet/src/manag...
- **[SUGGESTION] (deduped existing open thread) `packages/rs-platform-wallet/src/wallet/shielded/file_store.rs:377-402`: bump_redrive_attempts mutates memory before persisting, so a SQLite failure diverges the two views** — Carried forward from prior review (prior-2); code unchanged at head 5a9c704358. `FileBackedShieldedStore::bump_redrive_attempts` calls `sw.bump_redrive_attempts(activity_id)` — which unconditionally increments the in-memory `SubwalletState::redrives[activity_id].attempts` — before issuing `UPDATE...
- **[SUGGESTION] (deduped existing open thread) `packages/rs-platform-wallet/src/wallet/shielded/file_store.rs:205-220`: Second SQLite connection on the same file lacks a busy_timeout** — Carried forward from prior review (prior-3); code unchanged at head. The PR opens a second `rusqlite::Connection` (`pending_conn`) via `open_tuned_connection` against the same file as `tree`. `open_tuned_connection` sets `journal_mode=WAL`, `synchronous=NORMAL`, `temp_store=MEMORY`, but installs...
- **[SUGGESTION] (deduped existing open thread) `packages/rs-platform-wallet/src/wallet/shielded/operations.rs:2167-2172`: NullifierAlreadySpent arm does not bound retries or clear the record** — Carried forward from prior review (prior-4); code unchanged at head. The `Err(e) if is_nullifier_already_spent(&e)` arm in `redrive_pending_spends` only logs — it does not `bump_redrive_attempts` and does not `clear_redrive`. The comment ("the next scan confirms") assumes the following scan will...
- **[SUGGESTION] (deduped existing open thread) `packages/rs-platform-wallet/src/wallet/shielded/file_store.rs:288-322`: delete_redrive_rows_containing error can abort mark_spent/clear_pending after the in-memory drop already happened** — Carried forward from prior review (prior-5); code unchanged at head. Both `mark_spent` (line 296) and `clear_pending` (line 319) call `self.delete_redrive_rows_containing(id, nullifier)?` after the in-memory `SubwalletState::mark_spent`/`clear_pending` mutation already ran and already dropped the...
- **[SUGGESTION] (deduped existing open thread) `packages/rs-platform-wallet/src/wallet/shielded/coordinator.rs:657-673`: persisters read lock held across the entire network-calling redrive loop** — Carried forward from prior review (prior-6); code unchanged at head. The new redrive block acquires `self.persisters.read().await` and holds it for the full `for (id, _) in &subwallets` loop, and each iteration awaits `redrive_pending_spends`, which performs up to `MAX_REDRIVE_ATTEMPTS` real `st....
- **[SUGGESTION] (deduped existing open thread) `packages/rs-platform-wallet/src/wallet/shielded/operations.rs:2156-2207`: bump_redrive_attempts persistence errors silently discarded at both call sites** — Carried forward from prior review (prior-7); code unchanged at head. Both `bump_redrive_attempts` invocations in `redrive_pending_spends` (the `Ok(())` arm at 2156-2160 and the inconclusive-error arm at 2203-2207) do `.unwrap_or(0)` on the Result, swallowing any store-persistence error entirely w...
- **[NITPICK] (deduped existing open thread) `packages/rs-platform-wallet-ffi/src/shielded_send.rs:454-456`: poke_sync_on_unconfirmed silently discards the forced-sync result** — Carried forward from prior review (prior-8); code unchanged at head. `let _ = coordinator.sync(true).await;` throws away both errors and any panic captured in the (immediately dropped) `JoinHandle`. Discarding the typed spend result is fine — the ambiguity is already decided — but discarding the...
- **[NITPICK] (deduped existing open thread) `packages/rs-platform-wallet/src/wallet/shielded/operations.rs:2114-2216`: redrive_pending_spends' actual broadcast-response classification is untested** — Carried forward from prior review (prior-9); code unchanged at head. The only test exercising `redrive_pending_spends` (`redrive_skips_pruned_and_exhausted_and_drops_garbage`) deliberately never reaches `st.broadcast(sdk, None).await` — its three records are pruned-anchor, exhausted-attempts, and...

</details>

<details>

```

- [BLOCKING] packages/rs-platform-wallet-ffi/src/shielded_send.rs:441-457: poke_sync_on_unconfirmed bypasses the manager's is_syncing/quiescing gate
  Carried forward from prior review (prior-1); verified byte-identical at head 5a9c704358. `poke_sync_on_unconfirmed` spawns `coordinator.sync(true).await` directly against the shared `Arc<NetworkShieldedCoordinator>`, bypassing `ShieldedSyncManager::sync_now` (packages/rs-platform-wallet/src/manager/shielded_sync.rs:336-395) — the only place that CAS-guards on `is_syncing` and checks `quiescing`. `NetworkShieldedCoordinator::sync()` has no internal reentrancy guard, and `force=true` also skips the cooldown check.

  Two consequences:
  1. A background pass already in flight when a spend goes ambiguous will run a second, fully concurrent `sync()` against the same store; both take the same pre-scan `stale_pending_spends`+`recorded` snapshot and both call `redrive_pending_spends` on the same subwallet. One ambiguous event can burn 2 of the 3 redrive attempts in a single race, and independently bump `attempts` twice.
  2. It breaks `WalletManager::quiesce()`'s drain invariant (shielded_sync.rs:299-323): since this spawned task never sets `is_syncing`, the `while self.is_syncing.load(...) {}` drain loop will not wait for it. A Clear/unregister/rebind following `quiesce()` can still race a resolving sync pass triggered by this poke — holding coordinator persisters/store RwLocks after the host was told teardown was safe.

  Also worth noting: because `on_shielded_sync_completed` is dispatched by the manager's `sync_now` (shielded_sync.rs:393), not by `coordinator.sync`, this manually-triggered pass never emits the completion event that hosts (Swift `handleShieldedSyncCompleted`) rely on to reflect the resolved re-check.

  Fix: route the FFI poke through a `manager.poke_sync_now(true)` helper that runs `sync_now(true)` under the same CAS + quiescing check, or add an in-flight guard inside `NetworkShieldedCoordinator::sync` now that it has two independent callers. Fund-safety is preserved (duplicate nullifiers hit `NullifierAlreadySpent`) but the retry-budget intent and the teardown barrier are not.

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/file_store.rs:377-402: bump_redrive_attempts mutates memory before persisting, so a SQLite failure diverges the two views
  Carried forward from prior review (prior-2); code unchanged at head 5a9c704358. `FileBackedShieldedStore::bump_redrive_attempts` calls `sw.bump_redrive_attempts(activity_id)` — which unconditionally increments the in-memory `SubwalletState::redrives[activity_id].attempts` — before issuing `UPDATE shielded_pending_spends SET attempts = ?4`. If the UPDATE fails the function returns `Err`, but the in-memory counter has already advanced. On next reopen `rehydrate_pending_spends` rebuilds from the stale SQLite value, silently rewinding the visible attempt count.

  Compounded by the caller in `redrive_pending_spends` (operations.rs:2156-2160 and 2203-2207) doing `.unwrap_or(0)`, which folds the persistence failure into an `attempts=0` log line and continues. `arm_redrive` already orders SQLite before memory — mirror that here (UPDATE against SQLite first, then mirror the resulting count to memory) so the two views cannot diverge across restarts.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/file_store.rs:205-220: Second SQLite connection on the same file lacks a busy_timeout
  Carried forward from prior review (prior-3); code unchanged at head. The PR opens a second `rusqlite::Connection` (`pending_conn`) via `open_tuned_connection` against the same file as `tree`. `open_tuned_connection` sets `journal_mode=WAL`, `synchronous=NORMAL`, `temp_store=MEMORY`, but installs no `busy_timeout`. Under WAL, readers do not block writers, but two writer connections still serialize on the write lock and a concurrent writer receives `SQLITE_BUSY` immediately without a busy handler.

  On iOS under concurrent activity — a `mark_spent`/tree write racing an `arm_redrive` or `bump_redrive_attempts` write on `pending_conn` — this surfaces as `FileShieldedStoreError("persist redrive: database is locked")`, either propagating up as a hard `arm_redrive` failure (silently disabling redrive persistence for that spend so only the prune backstop rescues it) or aborting the operation entirely. Set a modest `busy_timeout` (a few seconds) on both connections so SQLite retries transparently.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/file_store.rs:288-322: delete_redrive_rows_containing error can abort mark_spent/clear_pending after the in-memory drop already happened
  Carried forward from prior review (prior-5); code unchanged at head. Both `mark_spent` (line 296) and `clear_pending` (line 319) call `self.delete_redrive_rows_containing(id, nullifier)?` after the in-memory `SubwalletState::mark_spent`/`clear_pending` mutation already ran and already dropped the in-memory `redrives` entry. If the SQL delete fails (WAL contention between the two connections on the same file, disk error), the function returns `Err` even though the in-memory state has already changed — the persisted `shielded_pending_spends` row is left behind, desynced from memory.

  `mark_spent` is called from `apply_scanned_nullifier_spends` as `store.mark_spent(*id, nf)?`, so this `?` now aborts detection of all remaining scanned nullifier spends across every subwallet for the pass — a new failure mode, since pre-PR `mark_spent` only touched the in-memory map and could not fail. On `clear_pending` SQL failure the in-memory reservation is already freed but the stale SQLite redrive row survives; on restart `rehydrate_pending_spends` resurrects it and re-reserves the (legitimately released) nullifier, locking the note back up until a later pass notices. Perform the SQL delete before mutating memory, or catch the persistence error and log without propagating.

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/operations.rs:2167-2172: NullifierAlreadySpent arm does not bound retries or clear the record
  Carried forward from prior review (prior-4); code unchanged at head. The `Err(e) if is_nullifier_already_spent(&e)` arm in `redrive_pending_spends` only logs — it does not `bump_redrive_attempts` and does not `clear_redrive`. The comment ("the next scan confirms") assumes the following scan will land the nullifier via `apply_scanned_nullifier_spends` → `mark_spent`, which drops the record.

  That is the common path but the arm is otherwise unbounded: if the scan repeatedly fails to cover the block that landed the nullifier (transport hiccup during a chunk, a subwallet whose store lost the note, delivery lag past cooldowns), every subsequent sync re-broadcasts and receives `NullifierAlreadySpent` again indefinitely. Since this arm is already a definitive success signal (`NullifierAlreadySpent` proves the transition executed), clear the record here — leaving the reservation in place for the next scan's `mark_spent` to release. At minimum, count the attempt against `MAX_REDRIVE_ATTEMPTS` so the retry loop is bounded even in the pathological scan-gap case.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/operations.rs:2156-2207: bump_redrive_attempts persistence errors silently discarded at both call sites
  Carried forward from prior review (prior-7); code unchanged at head. Both `bump_redrive_attempts` invocations in `redrive_pending_spends` (the `Ok(())` arm at 2156-2160 and the inconclusive-error arm at 2203-2207) do `.unwrap_or(0)` on the Result, swallowing any store-persistence error entirely with no `warn!`/`debug!`. Every other store-call failure in the same function (`pending_redrives`, `clear_redrive`, `clear_pending`) logs a `warn!` on error; these two are the odd ones out. Combined with the memory-before-SQLite ordering in the file store (prior-2), this converts a SQLite failure into a silent stale-count on restart with no diagnostic trail.

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/coordinator.rs:657-673: persisters read lock held across the entire network-calling redrive loop
  Carried forward from prior review (prior-6); code unchanged at head. The new redrive block acquires `self.persisters.read().await` and holds it for the full `for (id, _) in &subwallets` loop, and each iteration awaits `redrive_pending_spends`, which performs up to `MAX_REDRIVE_ATTEMPTS` real `st.broadcast(sdk, None).await` round-trips per subwallet. A `tokio::sync::RwLock` read guard blocks any writer for its full lifetime, so `register_wallet`/`unregister_wallet`/`clear` (all needing `persisters.write().await`) stall for the duration of this pass any time a host adds/removes a wallet while a redrive-heavy sync is running.

  The pre-existing `release_stranded_spends` uses the same locking pattern but only does local store work; extending that pattern across network I/O changes the assumption. Snapshot the needed `WalletPersister` clones under the lock, drop the guard, then run the network-calling loop.
```
</details>

packages/rs-platform-wallet-ffi/src/shielded_send.rs
QuantumExplorer: Fixed in 997d9e77c9: the poke now routes through `ShieldedSyncManager::sync_now(true)` (resolved via `shielded_sync_arc()` from the manager handle), so it sits behind the same `is_syncing` CAS + `quiescing` drain barrier as the periodic loop — `force` bypasses only the caught-up cooldown, never serialization. An in-flight pass makes the poke a no-op; the next tick's pass owns the re-drive.

packages/rs-platform-wallet/src/wallet/shielded/file_store.rs
QuantumExplorer: Fixed in 997d9e77c9: `bump_redrive_attempts` now persists the incremented counter to SQLite FIRST and mutates memory only on success (via a new read-only `redrive_attempts` accessor), so a SQLite failure leaves both views at the previous count.

packages/rs-platform-wallet/src/wallet/shielded/file_store.rs
QuantumExplorer: Fixed in 997d9e77c9: `open_tuned_connection` now sets `busy_timeout(5s)`, covering both writer connections on the file (the commitment tree's and `pending_conn`).

packages/rs-platform-wallet/src/wallet/shielded/operations.rs
QuantumExplorer: Fixed in 997d9e77c9: the `AlreadyExecuted` arm now consumes an attempt via the shared bump helper, so a lagging scan can no longer induce unbounded per-pass re-broadcasts — the cap parks the record for the scan/prune passes to settle.

packages/rs-platform-wallet/src/wallet/shielded/file_store.rs
QuantumExplorer: Fixed in 997d9e77c9: `mark_spent` / `clear_pending` now log-and-continue when the SQLite deletion mirror fails — the in-memory transition already happened, so the call keeps its trait behavior; a surviving stale row rehydrates on the next open and self-heals via the reconcile/prune passes (noted in the log line).

packages/rs-platform-wallet/src/wallet/shielded/coordinator.rs
QuantumExplorer: Fixed in 997d9e77c9: the per-wallet persisters are snapshotted into a `Vec<(SubwalletId, Option<WalletPersister>)>` under the read guard, which is dropped before the network-calling redrive loop runs.

packages/rs-platform-wallet/src/wallet/shielded/operations.rs
QuantumExplorer: Fixed in 997d9e77c9: both call sites (now three, incl. the AlreadyExecuted arm) route through `bump_redrive_attempts_logged`, which warns on a persistence failure — with the persist-first store ordering an `Err` means the attempt slot wasn't consumed and retries next pass, and the log line makes that visible.

packages/rs-platform-wallet-ffi/src/shielded_send.rs
QuantumExplorer: Fixed in 997d9e77c9: the spawned poke now logs the pass outcome — a debug line distinguishing a completed pass (with wallet count) from a skipped one (pass already in flight / shielded unconfigured).

packages/rs-platform-wallet/src/wallet/shielded/operations.rs
QuantumExplorer: Addressed in 997d9e77c9 by restructuring: the broadcast-outcome decision is now the pure `classify_redrive_broadcast` (Accepted / AlreadyExecuted / DefinitiveRejection / Inconclusive), and `redrive_broadcast_classification_matrix` pins all four arms including the ordering constraint that `NullifierAlreadySpent` — itself a consensus error — must classify as success before the generic rejection arm.

packages/rs-platform-wallet-ffi/src/shielded_send.rs
```diff
@@ -421,11 +425,61 @@ pub unsafe extern "C" fn platform_wallet_manager_shielded_withdraw(
...
+        if summary.sync_unix_seconds == 0 {
+            tracing::debug!(
+                "post-unconfirmed shielded sync poke skipped (a pass was already in flight                  or shielded is unconfigured); the next pass owns the re-drive"
+            );
+        } else {
+            tracing::debug!(
+                wallets = summary.wallet_results.len(),
+                "post-unconfirmed shielded sync pass completed"
+            );
+        }
```
thepastaclaw, line 479: **💬 Nitpick: Stray run of spaces in the poke-skipped debug log, and the message slightly mischaracterizes the `sync_unix_seconds == 0` branch**

Two small issues in the new `poke_sync_on_unconfirmed` debug log:

1. Line 472 has ~18 literal spaces baked into the string between "in flight" and "or shielded is unconfigured" — an editor-unwrap artifact that `rustfmt` won't catch inside a string literal. It will appear verbatim in operator logs and become a grep landmine.

2. The message says `sync_unix_seconds == 0` covers both "a pass was already in flight" and "shielded is unconfigured". Looking at `ShieldedSyncManager::sync_now` (packages/rs-platform-wallet/src/manager/shielded_sync.rs:336-398): only the CAS-lost and `quiescing` short-circuits leave `sync_unix_seconds == 0`. The no-coordinator path returns a default summary and then falls through to the `summary.sync_unix_seconds = now` stamp + completion-event dispatch, so "unconfigured" actually lands in the `"...pass completed"` branch (with `wallets = 0`). The unconfigured case is also practically unreachable here since the spend that produced the ambiguous result required a configured coordinator. Purely cosmetic — no runtime impact.

```suggestion
            tracing::debug!(
                "post-unconfirmed shielded sync poke skipped (a pass was already in flight or shielded is unconfigured); the next pass owns the re-drive"
            );
```

<sub>source: ['claude']</sub>

thepastaclaw (approved):
## Code Review

Delta commit 997d9e77 ("address review findings on the re-drive pass") resolves all 9 prior findings from 5a9c7043: the blocking `poke_sync_on_unconfirmed` bypass now routes through `ShieldedSyncManager::sync_now` (respecting the `is_syncing` CAS + `quiescing` drain barrier), `bump_redrive_attempts` is now persist-first, `busy_timeout(5s)` is installed on the shared SQLite writer, the `AlreadyExecuted` arm now consumes an attempt (bounding retries), `mark_spent`/`clear_pending` log-and-continue on SQL delete failures, the persisters read guard is snapshotted before the network-calling redrive loop, a new `bump_redrive_attempts_logged` helper logs persistence errors, the poke's summary is inspected/logged, and a new `redrive_broadcast_classification_matrix` test pins the `AlreadyExecuted`-before-`DefinitiveRejection` arm order. The only new observation is a cosmetic whitespace/log-wording nitpick in the poke's own debug line.

Source: reviewers: `opus` (general), `claude-sonnet-5` (general), `gpt-5.5[high]` (general, failed_or_unparseable), `opus` (rust-quality), `claude-sonnet-5` (rust-quality), `gpt-5.5[high]` (rust-quality, failed_or_unparseable), `opus` (ffi-engineer), `claude-sonnet-5` (ffi-engineer), `gpt-5.5[high]` (ffi-engineer, failed_or_unparseable); verifier: `opus`; specialists: `rust-quality`, `ffi-engineer`

💬 1 nitpick(s)

PR #3994, opened 2026-07-03
fix(platform-wallet): size withdrawal plan from on-chain balances

QuantumExplorer:

AUTO platform-address withdrawal (ADDR-04) could pass the UI preflight (`can_withdraw = true`, a plausible `net_withdrawable`) and then **fail every time at spend time** with:

```
AddressNotEnoughFundsError: Insufficient address balance for P2PKH(<hash>): has 100000000, requires at least 200000000
```

Repro (testnet): fund platform account #0 via asset-lock top-up (ADDR-03), transfer 0.001 DASH to a second own-wallet address on the same account (ADDR-02), then withdraw the account to Core.

**Root cause:** `PlatformAddressWallet::plan_withdrawal` sized each input's withdraw amount from the wallet's **cached** `address_credit_balance`, whereas the spend path re-fetches the **on-chain** balance in `fetch_inputs_with_nonce` and hard-checks it with `ensure_address_balance`. When an address's cached balance exceeded its on-chain balance (observed: a transfer recipient cached at 2× its real 100,000,000-credit balance), the preflight approved a plan requesting more than the address holds, and the spend rejected it. The planner's fee math (`reserve_withdrawal_fee_on_largest_input`) was correct; only the *balance source* diverged from the spend, so the "single source of truth" guarantee didn't actually hold.

Note: the earlier report's "input at exactly `min_input_amount`" framing was imprecise — `min_input_amount` is 100,000 credits (1 DASH = 1e11 credits), so 100,000,000 is the transfer amount and the `200000000` in the error is exactly **2× the input's on-chain balance**.

## What was done?

- **`plan_withdrawal`** now sizes every per-input amount from **authoritative on-chain balances** via `AddressInfo::fetch_many` — the same query the spend re-runs in `fetch_inputs_with_nonce`. The account's derived-address pool supplies only the candidate address *set*; balances come from the chain. Addresses the proof reports absent/`None` map to 0 and are dropped as dust. Preflight and the AUTO spend can no longer over-request relative to reality even when the cache is stale or doubled. The `reserve_withdrawal_fee_on_largest_input` fee-reservation math is unchanged.
- **`platform_address_wallet_preflight_withdrawal`** (FFI) now clones the wallet out and polls on the 8 MB `block_on_worker` instead of `runtime().block_on`: the preflight now verifies a fetch proof, and GroveDB proof verification recurses past the ~512 KB iOS dispatch/Swift-concurrency thread stacks (would `EXC_BAD_ACCESS`) — the same reason the withdraw path already uses the worker.

## How Has This Been Tested?

- Added 3 unit tests in `platform-wallet` (pure, no network):
  - `plan_min_input_amount_input_is_spendable_at_full_balance` — a multi-input plan with one input at **exactly `min_input_amount`** beside a larger fee source: the min input is withdrawn at full balance, the fee is reserved on the larger input, and every per-input amount ≤ its balance.
  - `plan_does_not_over_request_small_recipient_input` — the ADDR-04 shape: a 100,000,000-credit recipient input is planned at exactly its balance, explicitly **not** 200,000,000.
  - `plan_every_input_is_spendable_and_only_fee_source_is_reduced` — general spendability invariant.
- `cargo test -p platform-wallet` — all pass.
- `cargo check -p platform-wallet-ffi --all-targets` — clean.
- `cargo fmt --all -- --check` — clean.

## Breaking Changes

None. This is a client-side wallet planning change; no consensus rules, serialization, or FFI signatures change. (`plan_withdrawal`/`preflight_withdrawal` now issue one proof query instead of being purely in-memory.)

## Follow-up (out of scope)

The fix corrects the withdrawal regardless of *why* the cache was wrong. The underlying **doubled-cache write** was not pinned down: every balance write in this repo is an idempotent absolute `set_address_credit_balance` (the accumulating `add_` variant is called nowhere), so the double is likely a sync/reconcile race or a SwiftData hydration issue — tracked as a separate task. That doubled cache would also mis-display balances elsewhere and should be fixed at its source.

## Checklist:

### DIFF
--- a/packages/rs-platform-wallet-ffi/src/platform_addresses/withdrawal.rs
+++ b/packages/rs-platform-wallet-ffi/src/platform_addresses/withdrawal.rs
@@ -11,7 +11,7 @@ use rs_sdk_ffi::{SignerHandle, VTableSigner};
 use std::os::raw::c_char;
 use std::str::FromStr;
 
-use super::{parse_input_selection, runtime};
+use super::parse_input_selection;
 use crate::runtime::block_on_worker;
 
 /// Withdraw platform credits to a Core L1 address.
@@ -202,12 +202,16 @@ pub unsafe extern "C" fn platform_address_wallet_withdraw_to_address(
 ///
 /// Runs the same Rust planning phase the real withdraw path executes
 /// (`PlatformAddressWallet::preflight_withdrawal` →
-/// `plan_withdrawal`/`reserve_withdrawal_fee_on_largest_input`): it drops
+/// `plan_withdrawal`/`reserve_withdrawal_fee_on_largest_input`): it reads the
+/// account's authoritative on-chain balances (one `AddressInfo::fetch_many`
+/// proof query — the SAME balances the spend re-fetches and hard-checks), drops
 /// sub-`min_input_amount` dust, estimates the transition fee from the selected
 /// input count (NOT from any destination script — no Core address is needed or
 /// touched), reserves that fee on the largest-balance input, and verifies the
-/// net clears `system_limits.min_withdrawal_amount`. Gating a UI submit button
-/// on the result keeps it in lockstep with what the spend path will accept.
+/// net clears `system_limits.min_withdrawal_amount`. Because the plan is sized
+/// from on-chain balances rather than the wallet cache, gating a UI submit
+/// button on the result keeps it in lockstep with what the spend path accepts
+/// even when the cached balance is stale or doubled.
 ///
 /// On success `out` is written with `can_withdraw = true` and the net /
 /// estimated-fee figures, and the call returns [`PlatformWalletFFIResult::ok`].
@@ -239,11 +243,17 @@ pub unsafe extern "C" fn platform_address_wallet_preflight_withdrawal(
 ) -> PlatformWalletFFIResult {
     check_ptr!(out);
 
-    let option = PLATFORM_ADDRESS_WALLET_STORAGE.with_item(handle, |wallet| {
-        runtime().block_on(wallet.preflight_withdrawal(account_index))
-    });
-    // `None` → invalid handle (mapped to NotFound by the blanket Option impl).
-    let result = unwrap_option_or_return!(option);
+    // Clone the wallet out of handle storage so the read lock is released
+    // before the fetch, then poll on a worker thread (8 MB stack). The
+    // preflight now issues an `AddressInfo::fetch_many` proof query to read
+    // authoritative on-chain balances (so the gate can't approve what the
+    // spend rejects), and GroveDB proof verification recurses past the
+    // ~512 KB stacks of iOS dispatch / Swift-concurrency threads (see
+    // runtime.rs) — polling it on the calling thread would crash with
+    // EXC_BAD_ACCESS, the same reason the withdraw path uses the worker.
+    let option = PLATFORM_ADDRESS_WALLET_STORAGE.with_item(handle, |wallet| wallet.clone());
+    let wallet = unwrap_option_or_return!(option);
+    let result = block_on_worker(async move { wallet.preflight_withdrawal(account_index).await });
 
     match result {
         Ok(plan) => {
--- a/packages/rs-platform-wallet/src/wallet/platform_addresses/withdrawal.rs
+++ b/packages/rs-platform-wallet/src/wallet/platform_addresses/withdrawal.rs
@@ -1,4 +1,4 @@
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet};
 
 use dpp::address_funds::{AddressFundsFeeStrategy, AddressFundsFeeStrategyStep, PlatformAddress};
 use dpp::fee::Credits;
@@ -13,6 +13,8 @@ use super::InputSelection;
 use crate::wallet::PlatformAddressWallet;
 use crate::{PlatformAddressChangeSet, PlatformWalletError};
 use dash_sdk::platform::transition::address_credit_withdrawal::WithdrawAddressFunds;
+use dash_sdk::platform::FetchMany;
+use dash_sdk::query_types::AddressInfo;
 
 /// The fully-planned shape of an AUTO withdrawal, computed by
 /// [`PlatformAddressWallet::plan_withdrawal`] without any signing, broadcast,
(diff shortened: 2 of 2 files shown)


thepastaclaw:
✅ Review complete (commit a397cbad)

thepastaclaw (changes requested):
## Code Review

The Rust-side fix — sizing plan_withdrawal from AddressInfo::fetch_many instead of the cached balance — is correct, well-motivated, and the FFI worker-thread dispatch mirrors the withdraw path's established pattern. However, moving the preflight from a pure in-memory computation to a synchronous network + proof-verification call surfaces two boundary regressions that were not updated in lockstep: (a) the Swift wrapper's docstring and its main-actor call sites still assume the pre-change 'no network, safe on main actor' contract, so the withdraw sheet now blocks the SwiftUI main thread on a DAPI round-trip on every account/fee-rate change; and (b) the FFI error path folds the newly-reachable dash_sdk::Error path into the 'structural failure' bucket, so a transient network blip is now indistinguishable from a bad handle to Swift and silently disables submit.

Source: reviewers opus/opus, sonnet5/claude-sonnet-5, codex/gpt-5.5[high] (failed: codex/gpt-5.5[high]:general, codex/gpt-5.5[high]:security-auditor, codex/gpt-5.5[high]:rust-quality, codex/gpt-5.5[high]:ffi-engineer) for general + security-auditor + rust-quality + ffi-engineer; verifier opus/opus.

_3 finding(s) were posted in the review body only because their anchors are outside the PR diff._

<details>

```

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/WithdrawPlatformAddressView.swift`:
- [BLOCKING] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/WithdrawPlatformAddressView.swift:164-184: preflightWithdrawal now blocks the main actor on a proof-verified network round-trip
  This PR changes plan_withdrawal (rs-platform-wallet/src/wallet/platform_addresses/withdrawal.rs:347) to issue an AddressInfo::fetch_many proof query, and correspondingly routes the FFI through block_on_worker (rs-platform-wallet-ffi/src/platform_addresses/withdrawal.rs:256). block_on_worker only moves polling onto a bigger-stack worker — the calling thread still parks on rt.block_on (runtime.rs:52-61). On the Swift side, ManagedPlatformAddressWallet.preflightWithdrawal (ManagedPlatformAddressWallet.swift:338-341) remains a synchronous `throws -> WithdrawalPreflight`, and its docstring (lines 326-327) still asserts it is 'a pure in-memory computation over cached balances, so it's fast and safe to call on the main actor.' WithdrawPlatformAddressView invokes it directly (not wrapped in `Task {}`) from `.onAppear` (line 170), `.onChange(of: sourceAccountIndex)` (lines 179-181) and `.onChange(of: coreFeePerByte)` (lines 182-184). The adjacent comment at lines 175-178 also explicitly (and now incorrectly) states 'The preflight is a local in-memory computation (no network), so this stays cheap.' Consequence: every time the user changes the source account or bumps the fee rate slider on the withdrawal sheet, the SwiftUI main thread now blocks on a full DAPI round-trip + GroveDB proof verification. This will cause visible hitches and risks main-thread-watchdog termination on slow networks. Either move the wrapper to async and dispatch the call sites off the main actor (Task { ... }), or provide a background-fetch entry point that caches the on-chain balances and lets the preflight remain sync. At minimum, update the Swift wrapper's docstring and the view's `.onChange` comment to reflect the new contract so callers stop propagating the wrong pattern.

In `packages/rs-platform-wallet-ffi/src/platform_addresses/withdrawal.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/platform_addresses/withdrawal.rs:231-297: Transient SDK/proof-fetch failures are now reported as structural FFI errors
  plan_withdrawal now propagates network/proof failures from AddressInfo::fetch_many via PlatformWalletError::Sdk (packages/rs-platform-wallet/src/error.rs:77 `#[from] dash_sdk::Error`). The FFI dispatch below only special-cases `OnlyDustInputs | AddressOperation` as the normal 'can't fund' outcome; every other variant — including the newly-reachable Sdk variant — falls through to `Err(other) => other.into()` and is mapped by the blanket From<PlatformWalletError> to a generic error code. The docstring at lines 231-233 still promises callers that 'Only a structural failure — a bad/destroyed handle, or a missing account at `account_index` (WalletNotFound / AddressSync) — is reported as an FFI error code with `out` left untouched.' That's no longer true. On the Swift side, `recomputePreflight` catches the throw with `preflight = nil` under the rationale 'Structural failure: leave submit disabled.' A dropped DAPI connection or a transient proof-verification failure now looks structurally identical to a bad handle and silently disables submit with no user-visible reason and no retry signal, where previously this path never depended on the network at all. Decide the policy explicitly: either fold Sdk into the `can_withdraw = false` arm (with the SDK error's Display in `reason`, so the UI can distinguish 'can't verify balances right now' from a genuine can't-fund), or update the docstring and Swift-side handling so a network error is not silently treated as fatal. Leaving both untouched drifts the preflight from a preflight into a coin-flip on network conditions.

In `packages/rs-platform-wallet/src/wallet/platform_addresses/withdrawal.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/platform_addresses/withdrawal.rs:292-363: plan_withdrawal's cache-vs-chain sourcing has no direct regression test
  The behavior this PR actually changes — plan_withdrawal reading balances from AddressInfo::fetch_many rather than account.address_credit_balance — has no direct test. All three new tests exercise reserve_withdrawal_fee_on_largest_input and select_withdrawable_inputs with hand-built BTreeMaps; the fetch-vs-cache seam that produced the ADDR-04 regression is only validated by the doc comments. A future refactor could silently reintroduce the cache read (or add an `.unwrap_or_else(|| cached)` fallback on the fetch) without a single test failing. packages/rs-sdk/src/mock already supports FetchMany mocking. A small mock-SDK test that (a) seeds the account cache with a doubled/stale balance for one address, (b) returns the true (smaller) on-chain balance from the mocked fetch_many, and (c) asserts the resulting plan sizes the input at the on-chain value would pin the fix at the layer it actually lives in. This is worth adding in this PR precisely because the fix is at that seam.
```
</details>

QuantumExplorer:
This merged before the CHANGES_REQUESTED review landed. All three findings are addressed in the follow-up **#3995** (branch `claude/withdrawal-preflight-async-a0276e`, on top of the squash-merged `891bd687`):

- 🔴 **(a) BLOCKING** — `preflightWithdrawal` is now `async` and driven off the main actor via a cancel-and-replace `Task`; `canSubmit` stays closed with a "Checking…" row while a preflight is in flight; the now-false "in-memory / safe on main actor" docstrings are corrected.
- 🟡 **(b)** — `PlatformWalletError::Sdk` (transient fetch/proof failures) is folded into the `can_withdraw = false` + reason arm so it's retryable, not a silent structural error; `WalletNotFound`/`AddressSync` stay structural.
- 🟡 **(c)** — added a mock-SDK seam test that seeds a doubled cache, mocks `fetch_many` to the true on-chain balance, and asserts `plan_withdrawal` sizes from chain (not cache).

`cargo test -p platform-wallet` (incl. the seam test), `cargo check -p platform-wallet-ffi --all-targets`, and `cargo fmt --all --check` are all clean.

PR #3997, opened 2026-07-04
fix: address verified findings from the dashpay m1 deep review

QuantumExplorer:

Fixes the verified findings from the deep review of #3841 ([correctness findings](https://github.com/dashpay/platform/pull/3841#issuecomment-4880070637), [design/conventions findings](https://github.com/dashpay/platform/pull/3841#issuecomment-4880070701)). Every finding below was adversarially verified against the branch head (`e20d730`) before being fixed; two review candidates were refuted during verification and are not addressed (F10 optimistic-clear, F13 regtest-HRP).

## What was done?

**Correctness (finding ids from the review):**
- **F1** — sent-side rotation supersede: `add_sent_contact_request` advances an established contact's `outgoing_request` on a rotation re-send instead of no-opping (metadata preserved), and the sent-doc ingest collapses newest-per-recipient — rotation no longer bricks after one use, and restore-from-seed tracks the freshest reference.
- **F2** — rotation teardown survives restart: stale external accounts are detected and re-torn-down/re-enqueued at load/sweep time (`external_account_reference` self-heal marker) instead of relying on the in-memory pending queue a restart wiped.
- **F3** — `backfillCore` no longer commits the shared context mid changeset round; the backfill defers until the round closes, preserving `store()` atomicity.
- **F4** — `ContactChangeSet::merge` reconciles insert-vs-tombstone last-write-wins per key for all three pairs (`ignored`/`unignored`, `sent`/`removed_sent`, `incoming`/`removed_incoming`); the fixed INSERT-before-DELETE apply order previously made a merged both-sets changeset always resolve to the tombstone.
- **F5** — Accept's adopt branch establishes the contact before the final lookup instead of returning `ContactRequestNotFound` after account registrations already ran.
- **F6** — the pending-crypto drain removes queue entries only when they still hold the drained payload, so a rotation upserted mid-drain isn't silently dropped.
- **F7** — cryptographically-failed autoAcceptProofs are marked per (sender, proof) and not re-enqueued every sweep, so attacker-controlled garbage can no longer pin the unlock indicator; transient failures stay retryable.
- **F8** — `TransactionListView` scopes the DashPay payment/counterparty join to the displayed wallet's identities (multi-identity stores no longer mislabel direction).
- **F9/C3** — the Swift signer routes derive-sign-destroy by attempting the resolver and falling back on its unsupported-key-type error instead of mirroring the Rust key-type set; `canSign` matches what sign can do.
- **F11** — `accepted_accounts` round-trips on the FFI/SwiftData path (additive `ContactRequestFFI` fields + Swift persist/restore), matching the SQLite backend.
- **F12** — `payment_channel_broken` clears on successful external-account registration.
- **F14** — key-material hygiene: temporary signing scalar wiped in `platform_wallet_pubkey_hash_from_private_key`; ECDH shared secrets stay in `Zeroizing` end-to-end through `ContactCryptoProvider`.
- **F15** — send-success toast shows the txid in display byte order (matches history rows and explorers).
- **F16** — hidden contacts are recoverable via a new `HiddenContactsView` (mirrors the ignored-senders screen); unhide preserves alias/note.

**Design / conventions:**
- **D2** — `put_to_platform` verifies a caller-supplied entropy actually derives the document id before broadcasting, converting the consensus-only `InvalidDocumentTransitionIdError` failure mode into a local error for every future two-phase flow.
- **D4** — one shared key-purpose predicate (exported from `dash-sdk`) owns the DECRYPTION-or-ENCRYPTION cohort policy across the SDK check, wallet validator, and selector.
- **D8** — dedups: crate `now_ms()` util, shared profile query, shared payment-recording helper.
- **D7b** — `sync_profiles` batches identities through the existing `In`-chunk query and persists only on change.
- **D1 (minimal)** — the stale-loop generation guard is ported to `platform_address_sync` and `identity_sync` (the unconditional cancel-slot clear was a latent stop()+start() clobber); full shared-runner extraction deferred.
- **D3** — the unused `DashPaySdkWriter` trait seam is removed rather than exercised: `sdk_writer` becomes a concrete `Arc<Sdk>`-backed `SdkWriter` (still erasing the 7-type-param SDK generics behind two concrete methods); SPEC.md's as-built note records the revision.
- **D6** — the four copy-pasted `cachedProfile` helpers collapse into one shared lookup; `ContactsView` derives its row list once per body evaluation; the FFI profile accessor clones one field instead of the whole `ManagedIdentity`.
- **C1** — the FFI sign-time pubkey-binding decision moved into a shared platform-wallet helper aligned with rs-dpp's discovery-time validation.
- Also fixes a pre-existing `--all-features` compile break in `platform-wallet-storage` (dead `private_key` initializers left by the scalar-removal refactor).

**Deferred (tracked in the review comments, deliberately not in this PR):** full shared sync-runner extraction (D1), CompactSize codec swap to `dashcore::consensus::encode::VarInt` (D5 — cross-client wire format, wants dedicated interop testing), persistence-seam re-architecture (D9), D7 sub-items a/c/d, and the FFI ignored-rows projection ordering (defense-in-depth once F4's merge fix landed).

## How Has This Been Tested?

- `cargo test -p platform-wallet -p platform-wallet-storage -p platform-wallet-ffi --all-features` — green (platform-wallet lib now 422+ tests; every behavioral fix has a test that fails against the unfixed code)
- `cargo test -p dash-sdk --features mocks,offline-testing --lib` — 152 passed
- `cargo check --all-features --all-targets` on all five touched crates + `cargo clippy` — clean; `cargo fmt --all` applied
- `build_ios.sh --target sim` — `** BUILD SUCCEEDED **` (Rust xcframework with regenerated header, SwiftDashSDK, SwiftExampleApp)
- SwiftDashSDK package tests: 238 passed, 0 failures (`xcodebuild test -scheme SwiftDashSDK`, iPhone 16 sim), excluding `IdentityResolverSignIntegrationTests` — its 2 tests fail with `keychainError(-34018)` (missing keychain entitlement in the headless package-test runner); that file and `WalletStorage.swift` are untouched by this PR (added on the base branch in `fa1098c`), so the failure is environmental and pre-existing

## Breaking Changes

None. FFI struct additions are additive (new fields at the end of `ContactRequestFFI`); the `ContactCryptoProvider::ecdh_shared_secret` return type widened to `Zeroizing<[u8;32]>` within the workspace (no external consumers).

## Checklist:


thepastaclaw:
✅ Review complete (commit 9156d539)

thepastaclaw (commented):
## Code Review

Well-scoped PR fixing 20+ prior deep-review findings across DashPay wallet correctness (Rust core, FFI, Swift). Every core fix I traced is logically sound and backed by targeted tests. No blocking correctness or consensus issues survive scoping. Several convergent suggestions worth addressing before merge: two persist-order patterns in the new rotation-supersede code that leave memory ahead of disk on persister failure, a wall-clock-millisecond freshness token in the deferred-crypto retain that is not guaranteed-unique, and an accept-adopt path that stamps the external-account marker after a no-op registration. The FFI ABI-break finding is a false positive for this monorepo (compile-time layout guard forces both sides to rebuild together; layout guard was correctly updated).

🟡 4 suggestion(s) | 💬 5 nitpick(s)

Source: reviewers: `opus` (general), `claude-sonnet-5` (general), `gpt-5.5` (general), `opus` (security-auditor), `claude-sonnet-5` (security-auditor), `gpt-5.5` (security-auditor), `opus` (rust-quality), `claude-sonnet-5` (rust-quality), `gpt-5.5` (rust-quality), `opus` (ffi-engineer), `claude-sonnet-5` (ffi-engineer), `gpt-5.5` (ffi-engineer); verifier: `opus`; specialists: `security-auditor`, `rust-quality`, `ffi-engineer`.

<details>

```

In `packages/rs-platform-wallet/src/wallet/identity/state/managed_identity/contact_requests.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/state/managed_identity/contact_requests.rs:91-105: Sent-side rotation supersede mutates memory before the fallible persist
  The new rotation-supersede branch in `add_sent_contact_request` assigns `contact.outgoing_request = request` (line 95) BEFORE calling `persister.store(cs.into())?` (line 104). If the store fails, the method returns `Err`, but `self.established_contacts` has already advanced to the new outgoing reference. The sync caller intentionally leaves the sent cursor unadvanced on error so the next sweep retries — but the retry now sees the same reference in memory and hits the no-op guard on line 85 (`existing.outgoing_request.account_reference == request.account_reference`), returning `Ok(())` without re-persisting. The rotation is silently lost from disk for the process lifetime; only a restart (which re-fetches from platform) can heal it.

  This is the mirror-image of the persist-before-commit pattern the sibling `set_contact_metadata` fix uses in this same PR. Build a cloned updated contact, persist with `?`, then commit to `established_contacts` only after the store succeeds.

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs:2445-2467: note_external_account_registered mutates memory before persist — persist failure suppresses in-process retries
  `note_external_account_registered` sets `contact.external_account_reference = Some(current_reference)` and `contact.payment_channel_broken = false` on the in-memory contact BEFORE `self.persister.store(cs.into())`. If the persist fails, memory has the fresh marker + cleared broken flag but disk still holds the stale marker / broken flag, and the early-return guard on line 2447 (`if already_current && !contact.payment_channel_broken { return; }`) will suppress every subsequent invocation for the same contact within the process. The sweep's `external_account_needs_rebuild` predicate also compares against the in-memory marker (line 930), so it will see memory == current_reference and stop enqueueing rebuilds — nothing in-process retries the persist. Self-healing only happens across a restart when the marker reloads as `None`/stale from SQLite.

  This is the same ordering the newly-fixed `set_contact_metadata` inverted. Build the snapshot, persist first, then commit to memory only on `Ok` so a persist failure leaves memory stale and the sweep keeps triggering rebuilds until a persist succeeds.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs:897-911: retain_drained_by_token uses a wall-clock-millisecond freshness token
  `retain_drained_by_token` treats `(key, enqueued_at_ms)` as a unique snapshot token, but `enqueued_at_ms` is a millisecond `SystemTime` timestamp and is documented on `PendingContactCrypto` as observability/ordering only. In principle, a concurrent `upsert_pending_contact_crypto` refreshing a payload could produce the same millisecond as the drained snapshot's entry (either through same-ms upsert or a clock rollback), making the token equal; the retain would then remove the fresh payload and persist-clear it, re-introducing the class of silent queue loss this helper is meant to prevent.

  In practice the drain does network I/O so a same-ms collision is unlikely, but the token is not guaranteed unique by construction. Prefer a monotonic per-entry generation counter (e.g., a `u64` bumped on every upsert), or compare the full snapshotted `PendingContactCrypto` value, so freshness is not tied to wall-clock precision or monotonicity.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs:2838-2852: note_external_account_registered stamps marker after register_external_contact_account no-ops on an existing (possibly stale) account
  `accept_register_external_validated` calls `register_external_contact_account` and unconditionally calls `note_external_account_registered` on `Ok(())`. But `register_external_contact_account` (contacts.rs:400-421) returns `Ok(())` without decrypting or replacing anything when `dashpay_external_accounts.contains_key(&key)` — the account key is `(index, user_id, friend_id)`, independent of `account_reference`.

  Race scenario the F2 self-heal is designed to catch: cold restart with `external_account_reference = None` and a persisted, stale `dashpay_external_accounts` row built from the OLD xpub. The rotation self-heal in the sync sweep tears down + rebuilds. If the user accepts a rotated incoming request before the sweep's teardown runs, the accept path's register no-ops (row already there) and this code stamps the marker as current with `contact.incoming_request.account_reference`. Now `external_account_needs_rebuild` returns false and the sweep will never rebuild — `send_payment` derives addresses from the old xpub the contact no longer watches.

  Either surface an `already_existed` result from `register_external_contact_account` so this path only stamps when a rebuild actually happened, or have the accept path force teardown before registering.
```
</details>

packages/rs-platform-wallet/src/wallet/identity/state/managed_identity/contact_requests.rs
```diff
@@ -64,10 +77,31 @@ impl ManagedIdentity {
...
+            let mut cs = ContactChangeSet::default();
+            cs.established.insert(
+                SentContactRequestKey {
+                    owner_id,
+                    recipient_id,
+                },
+                contact.clone(),
+            );
+            persister.store(cs.into())?;
             return Ok(());
```
thepastaclaw: **🟡 Suggestion: Sent-side rotation supersede mutates memory before the fallible persist**

The new rotation-supersede branch in `add_sent_contact_request` assigns `contact.outgoing_request = request` (line 95) BEFORE calling `persister.store(cs.into())?` (line 104). If the store fails, the method returns `Err`, but `self.established_contacts` has already advanced to the new outgoing reference. The sync caller intentionally leaves the sent cursor unadvanced on error so the next sweep retries — but the retry now sees the same reference in memory and hits the no-op guard on line 85 (`existing.outgoing_request.account_reference == request.account_reference`), returning `Ok(())` without re-persisting. The rotation is silently lost from disk for the process lifetime; only a restart (which re-fetches from platform) can heal it.

This is the mirror-image of the persist-before-commit pattern the sibling `set_contact_metadata` fix uses in this same PR. Build a cloned updated contact, persist with `?`, then commit to `established_contacts` only after the store succeeds.

```suggestion
            // Different outgoing reference → a rotation re-send. Advance
            // `outgoing_request` in place so the next rotation reads the
            // fresh version, preserving all user metadata.
            let mut updated = existing.clone();
            updated.outgoing_request = request;
            let mut cs = ContactChangeSet::default();
            cs.established.insert(
                SentContactRequestKey {
                    owner_id,
                    recipient_id,
                },
                updated.clone(),
            );
            persister.store(cs.into())?;
            self.established_contacts.insert(recipient_id, updated);
            return Ok(());
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs
```diff
@@ -2253,6 +2411,62 @@ impl<B: TransactionBroadcaster + ?Sized> IdentityWallet<B> {
...
+                recipient_id: *contact_id,
+            },
+            snapshot,
+        );
+        if let Err(e) = self.persister.store(cs.into()) {
+            tracing::error!(
+                "Failed to persist external-account-registered changeset: {}",
+                e
+            );
+        }
```
thepastaclaw, line 2479: **🟡 Suggestion: note_external_account_registered mutates memory before persist — persist failure suppresses in-process retries**

`note_external_account_registered` sets `contact.external_account_reference = Some(current_reference)` and `contact.payment_channel_broken = false` on the in-memory contact BEFORE `self.persister.store(cs.into())`. If the persist fails, memory has the fresh marker + cleared broken flag but disk still holds the stale marker / broken flag, and the early-return guard on line 2447 (`if already_current && !contact.payment_channel_broken { return; }`) will suppress every subsequent invocation for the same contact within the process. The sweep's `external_account_needs_rebuild` predicate also compares against the in-memory marker (line 930), so it will see memory == current_reference and stop enqueueing rebuilds — nothing in-process retries the persist. Self-healing only happens across a restart when the marker reloads as `None`/stale from SQLite.

This is the same ordering the newly-fixed `set_contact_metadata` inverted. Build the snapshot, persist first, then commit to memory only on `Ok` so a persist failure leaves memory stale and the sweep keeps triggering rebuilds until a persist succeeds.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs
```diff
@@ -851,6 +851,85 @@ fn newest_received_per_sender(
...
+    queue.retain(|e| {
+        let stale_match = cleared_with_token
+            .iter()
+            .any(|(k, tok)| *k == e.key() && *tok == e.enqueued_at_ms);
+        if stale_match {
+            removed.push(e.key());
+        }
+        !stale_match
+    });
+    removed
```
thepastaclaw: **🟡 Suggestion: retain_drained_by_token uses a wall-clock-millisecond freshness token**

`retain_drained_by_token` treats `(key, enqueued_at_ms)` as a unique snapshot token, but `enqueued_at_ms` is a millisecond `SystemTime` timestamp and is documented on `PendingContactCrypto` as observability/ordering only. In principle, a concurrent `upsert_pending_contact_crypto` refreshing a payload could produce the same millisecond as the drained snapshot's entry (either through same-ms upsert or a clock rollback), making the token equal; the retain would then remove the fresh payload and persist-clear it, re-introducing the class of silent queue loss this helper is meant to prevent.

In practice the drain does network I/O so a same-ms collision is unlikely, but the token is not guaranteed unique by construction. Prefer a monotonic per-entry generation counter (e.g., a `u64` bumped on every upsert), or compare the full snapshotted `PendingContactCrypto` value, so freshness is not tied to wall-clock precision or monotonicity.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs
```diff
@@ -2597,14 +2839,20 @@ impl<B: TransactionBroadcaster + ?Sized> IdentityWallet<B> {
...
 
-        // The external account is built; surface the contact's account label
-        // from the same ECDH shared key (best-effort, cosmetic — a label
-        // failure never fails the accept).
+        // The external account is built from the current incoming reference:
+        // stamp the rotation self-heal marker and clear any stale
+        // broken-channel flag (a user re-accept of a healed channel must stop
+        // reporting broken — the sweep never re-registers broken contacts).
+        self.note_external_account_registered(our_identity_id, contact_id)
+            .await;
```
thepastaclaw: **🟡 Suggestion: note_external_account_registered stamps marker after register_external_contact_account no-ops on an existing (possibly stale) account**

`accept_register_external_validated` calls `register_external_contact_account` and unconditionally calls `note_external_account_registered` on `Ok(())`. But `register_external_contact_account` (contacts.rs:400-421) returns `Ok(())` without decrypting or replacing anything when `dashpay_external_accounts.contains_key(&key)` — the account key is `(index, user_id, friend_id)`, independent of `account_reference`.

Race scenario the F2 self-heal is designed to catch: cold restart with `external_account_reference = None` and a persisted, stale `dashpay_external_accounts` row built from the OLD xpub. The rotation self-heal in the sync sweep tears down + rebuilds. If the user accepts a rotated incoming request before the sweep's teardown runs, the accept path's register no-ops (row already there) and this code stamps the marker as current with `contact.incoming_request.account_reference`. Now `external_account_needs_rebuild` returns false and the sweep will never rebuild — `send_payment` derives addresses from the old xpub the contact no longer watches.

Either surface an `already_existed` result from `register_external_contact_account` so this path only stamps when a rebuild actually happened, or have the accept path force teardown before registering.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/identity/state/managed_identity/mod.rs
```diff
@@ -84,6 +84,28 @@ pub struct ManagedIdentity {
...
+    /// failures (invalid signature / expired / malformed / bad index) are
+    /// recorded here; transient failures (provider unavailable, reciprocal-send
+    /// failure) stay retryable and are never marked.
+    ///
+    /// In-memory only (never persisted): a relaunch clears it, so a bad proof is
+    /// retried at most once per launch. The request stays manually acceptable
+    /// meanwhile, and the sender never establishes off a bogus proof — so
+    /// retrying once per launch is harmless and avoids a persisted tombstone for
+    /// attacker-controlled input.
+    pub auto_accept_verify_failed: std::collections::BTreeSet<[u8; 32]>,
```
thepastaclaw, line 110: **💬 Nitpick: auto_accept_verify_failed grows without eviction for the lifetime of the process**

`auto_accept_verify_failed: BTreeSet<[u8; 32]>` is keyed by `SHA256(sender_id ‖ proof)` — one entry per distinct proof blob. It is only ever inserted into, never evicted, and cleared only by process restart. A sender willing to spend the platform-credit cost of publishing many distinct malformed `contactRequest` documents (varying the trailing bytes or `account_reference` to make each proof distinct) can grow this set for the life of the process. The exploit is attacker-funded (each doc-create costs credits) and the design comment at contact_requests.rs:207-210 documents the in-memory/per-launch tradeoff, so this is a hardening gap rather than a functional bug. Consider capping the set (LRU/max-size, or per-sender scoping with a small cap) — three reviewers converged on this as a defense-in-depth item.

<sub>source: ['claude']</sub>

packages/swift-sdk/Sources/SwiftDashSDK/FFI/KeychainSigner.swift
```diff
@@ -333,6 +333,24 @@ public final class KeychainSigner: Signer, @unchecked Sendable {
...
+    /// resolver's supported set here to stay consistent with sign-time routing:
+    /// a breadcrumb-only row (no stored scalar) is only reported signable when
+    /// its type is one the resolver handles.
+    ///
+    /// The set mirrors `dash_sdk_sign_with_mnemonic_resolver_and_path`
+    /// (`ECDSA_SECP256K1 = 0`, `ECDSA_HASH160 = 2`). This is the one remaining
+    /// Swift-side mirror of a Rust decision; a lightweight FFI predicate
+    /// (`dash_sdk_resolver_supports_key_type`) would let this drop too.
+    static func resolverCanDeriveSign(keyType: UInt8) -> Bool {
+        keyType == 0 || keyType == 2
```
thepastaclaw: **💬 Nitpick: canSign preflight re-implements Rust's resolver key-type policy in Swift**

`resolverCanDeriveSign(keyType:)` hardcodes `keyType == 0 || keyType == 2` as a Swift-side copy of the resolver's supported-key-type set. The sign trampoline was correctly reworked in this PR to route on Rust's `UNSUPPORTED_KEY_TYPE` tag instead of mirroring the set, but `canSign` still needs a synchronous answer. Today the two sides agree, so there is no live bug, but per `packages/swift-sdk/CLAUDE.md` ("No re-implementing protocol constants as Swift mirrors") this is exactly the pattern the project has decided to eliminate. If Rust's resolver-supported set is ever extended, `canSign` would silently under-report signability until Swift is updated. Consider adding a small `dash_sdk_resolver_supports_key_type(u8) -> bool` FFI predicate and dropping the Swift constant.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet/src/wallet/identity/network/sdk_writer.rs
```diff
@@ -1,34 +1,20 @@
...
-//! Keeping the trait to two methods is deliberate: this is a test
-//! seam, not a refactor. Everything the SDK call needs that
-//! `IdentityWallet` can compute up front travels in by value so the
-//! trait stays `dyn`-compatible.
+//! `IdentityWallet` keeps doing all of the derivation (key-index
+//! resolution, ECDH-key derivation, xpub derivation, avatar hashing,
+//! document construction); this helper receives the already-derived
+//! primitives and performs the final SDK call.
 
 use std::sync::Arc;
```
thepastaclaw, line 20: **💬 Nitpick: PR description and SPEC still claim DashPaySdkWriter is an injectable test seam, but the trait was removed**

The PR description says D3 `DashPaySdkWriter` "gets a real test injection point + a recording-stub test; docs now match reality." The diff removes the `DashPaySdkWriter` trait and `#[async_trait]` boxing entirely, converting `sdk_writer` into a concrete `Arc<SdkWriter>` field. No recording-stub test is added — no `DashPaySdkWriter` remains in the tree. `docs/dashpay/SPEC.md:586-591` still lists `DashPaySdkWriter` as DONE (2026-06-10) with a description of the trait it now contradicts. The simplification (drop the unused seam) is fine — but the PR description and the SPEC should be reconciled with the code so a future reader isn't looking for a trait that no longer exists.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet/src/wallet/identity/state/managed_identity/contact_requests.rs
QuantumExplorer: Fixed in 9156d53942: the supersede branch now builds the updated clone, persists with `?`, and commits to `established_contacts` only after the store succeeds — same order as `set_contact_metadata`. Pinned by `rotation_supersede_failed_persist_leaves_memory_on_old_reference` (failed persist leaves the old reference; the retry takes the supersede path and lands).

packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs
QuantumExplorer: Fixed in 9156d53942: `note_external_account_registered` now builds the updated snapshot, persists first, and commits to memory only on `Ok` — a persist failure leaves the stale marker/broken flag in place so the guard and `external_account_needs_rebuild` keep retriggering until a store succeeds.

packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs
QuantumExplorer: Fixed in 9156d53942: replaced the `(key, enqueued_at_ms)` token with full-value snapshot equality (`retain_drained_by_snapshot`; `PendingContactCrypto` already derives `Eq`). Freshness no longer depends on wall-clock precision or monotonicity; a same-ms refresh that actually changed the payload can never alias. An identical-payload refresh compares equal and is removed, which is fine — the drain just processed exactly those bytes.

packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs
QuantumExplorer: Fixed in 9156d53942 via the first suggested option: `register_external_contact_account` now returns `ExternalAccountRegistration::{Built, AlreadyExisted}`, and both the accept path and the deferred-crypto drain call `note_external_account_registered` only on `Built`. An `AlreadyExisted` no-op leaves the marker unstamped, so `external_account_needs_rebuild` still flags the stale row and the sweep's teardown + rebuild (whose build path stamps + clears broken) proceeds.

packages/rs-platform-wallet/src/wallet/identity/state/managed_identity/mod.rs
QuantumExplorer: Fixed in 9156d53942: capped at `AUTO_ACCEPT_VERIFY_FAILED_CAP = 1024` entries (32 KiB of keys) with arbitrary eviction (`pop_first`) over cap. Eviction is safe per the existing per-launch-retry design — an evicted proof merely re-verifies (and re-fails) on the next sweep.

packages/swift-sdk/Sources/SwiftDashSDK/FFI/KeychainSigner.swift
QuantumExplorer: Fixed in 9156d53942: added `dash_sdk_resolver_supports_key_type(u8) -> bool` to `sign_with_mnemonic_resolver.rs`, with the supported set factored into one internal helper both the predicate and the sign path use. `resolverCanDeriveSign` now delegates to the FFI predicate; an exhaustive all-`u8` test pins that the predicate agrees with the sign path's `UNSUPPORTED_KEY_TYPE` rejection.

packages/rs-platform-wallet/src/wallet/identity/network/sdk_writer.rs
QuantumExplorer: Fixed in 9156d53942: SPEC.md's as-built note now records the revision (trait shipped 2026-06-10, later removed as an unused seam; `sdk_writer` is a concrete `Arc<Sdk>`-backed `SdkWriter` that still erases the 7-type-param generics), and the PR description's D3 bullet has been updated to match.

packages/rs-platform-wallet/src/manager/identity_sync.rs
```diff
@@ -424,18 +422,54 @@ where
...
+    /// pass. When loop A finally exits, an unconditional `*guard = None`
+    /// would null **loop B's** live token, leaving loop B uncancellable —
+    /// a later shutdown `stop()`/`quiesce()` silently no-ops while loop B
+    /// keeps calling `persister.store(...)` through a freed FFI context.
+    fn clear_cancel_if_current(&self, my_generation: u64) {
+        let mut guard = self.background_cancel.lock().expect("bg_cancel poisoned");
+        if self.loop_generation.load(Ordering::Acquire) == my_generation {
+            *guard = None;
+        }
+    }
```
thepastaclaw: **💬 Nitpick: install_cancel/clear_cancel_if_current now duplicated across three sync managers**

Carried forward from prior-6. The generation-guarded install_cancel/clear_cancel_if_current pair is duplicated verbatim in three places — identity_sync.rs:441-471, dashpay_sync.rs:253-283, and platform_address_sync.rs:246-276 (one more copy than at the previous review, since the D1-minimal port added it to platform_address_sync.rs and dashpay_sync.rs). The guard's own doc comment ties the invariant to FFI safety ("loop A keeps calling persister.store(...) through a freed FFI context"), so any future edit to one copy that isn't mirrored to the other two silently re-opens the use-after-free-adjacent hazard on shutdown in whichever copy lags behind. The PR description explicitly defers the full shared-runner extraction (D1), so this is not blocking, but the duplication surface has grown rather than shrunk. Extract a shared LoopCancelGuard helper in a follow-up.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Second-round push (9156d539) surgically resolves 7 of the 9 prior findings with targeted code changes and regression tests: persist-before-commit ordering for both the sent-side rotation supersede (prior-1) and note_external_account_registered (prior-2); the wall-clock-ms freshness token replaced with full-value snapshot equality via retain_drained_by_snapshot (prior-3); an ExternalAccountRegistration::{Built, AlreadyExisted} split gating the self-heal marker (prior-4); an AUTO_ACCEPT_VERIFY_FAILED_CAP=1024 bound with pop_first() eviction (prior-5); a new dash_sdk_resolver_supports_key_type FFI predicate with an exhaustive 0..=255 agreement test replacing the Swift-side key-type mirror (prior-7); and a revised SPEC.md as-built note reconciling the removed DashPaySdkWriter (prior-9). No new latest-delta findings. Two carried-forward nitpicks remain: install_cancel/clear_cancel_if_current is now duplicated across THREE sync managers (identity_sync, dashpay_sync, platform_address_sync) — the PR body explicitly defers the shared-runner extraction — and platform_wallet_sync_dashpay_profiles's out_synced_count parameter/doc still labels a persist-on-change count as "synced". Both were nitpicks previously and remain nitpicks.

💬 2 nitpick(s)

Source: reviewers: `opus` (general), `claude-sonnet-5` (general), `gpt-5.5` (general, failed), `opus` (security-auditor), `claude-sonnet-5` (security-auditor), `gpt-5.5` (security-auditor, failed), `opus` (rust-quality), `claude-sonnet-5` (rust-quality), `gpt-5.5` (rust-quality, failed), `opus` (ffi-engineer), `claude-sonnet-5` (ffi-engineer), `gpt-5.5` (ffi-engineer, failed); verifier: `opus`; specialists: `security-auditor`, `rust-quality`, `ffi-engineer`.

<details>
<summary>Carried-forward prior findings</summary>

New latest-delta findings: none.

- [NITPICK] `packages/rs-platform-wallet/src/manager/identity_sync.rs:441-471`: install_cancel/clear_cancel_if_current now duplicated across three sync managers
  Carried forward from prior-6. The generation-guarded install_cancel/clear_cancel_if_current pair is duplicated verbatim in three places — identity_sync.rs:441-471, dashpay_sync.rs:253-283, and platform_address_sync.rs:246-276 (one more copy than at the previous review, since the D1-minimal port added it to platform_address_sync.rs and dashpay_sync.rs). The guard's own doc comment ties the invariant to FFI safety ("loop A keeps calling persister.store(...) through a freed FFI context"), so any future edit to one copy that isn't mirrored to the other two silently re-opens the use-after-free-adjacent hazard on shutdown in whichever copy lags behind. The PR description explicitly defers the full shared-runner extraction (D1), so this is not blocking, but the duplication surface has grown rather than shrunk. Extract a shared LoopCancelGuard helper in a follow-up.

- [NITPICK] `packages/rs-platform-wallet-ffi/src/dashpay_profile.rs:299-316`: platform_wallet_sync_dashpay_profiles out_synced_count still labels a persist-on-change count as "synced"
  Carried forward from prior-8; dashpay_profile.rs was not touched by this delta. sync_profiles() now returns the count of identities whose cached profile actually changed and was persisted (chunked persist-on-change semantics), but the FFI entry point still names the out-parameter out_synced_count and its docstring at lines 299-305 still says "refresh the local cache" with no note about the count meaning. The Swift wrapper ManagedPlatformWallet.syncDashPayProfiles() -> UInt32 inherits the ambiguous name. No in-tree caller depends on the exact value today, but the public FFI contract's meaning changed silently — a future Swift caller wiring a "synced N profiles" UI off this count will silently under-report when nothing changed. Rename to out_changed_count or add a docstring line clarifying the semantics.

</details>

packages/rs-platform-wallet/src/wallet/identity/state/managed_identity/contact_requests.rs
thepastaclaw: Resolved in `9156d53` — _Sent-side rotation supersede mutates memory before the fallible persist_ no longer present.

packages/swift-sdk/Sources/SwiftDashSDK/FFI/KeychainSigner.swift
thepastaclaw: Resolved in `9156d53` — _canSign preflight re-implements Rust's resolver key-type policy in Swift_ no longer present.

packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs
thepastaclaw: Resolved in `9156d53` — _retain_drained_by_token uses a wall-clock-millisecond freshness token_ no longer present.

packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs
thepastaclaw: Resolved in `9156d53` — _note_external_account_registered mutates memory before persist — persist failure suppresses in-process retries_ no longer present.

packages/rs-platform-wallet/src/wallet/identity/state/managed_identity/mod.rs
thepastaclaw: Resolved in `9156d53` — _auto_accept_verify_failed grows without eviction for the lifetime of the process_ no longer present.

packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs
thepastaclaw: Resolved in `9156d53` — _note_external_account_registered stamps marker after register_external_contact_account no-ops on an existing (possibly stale) account_ no longer present.

packages/rs-platform-wallet/src/wallet/identity/network/sdk_writer.rs
thepastaclaw: Resolved in this update — _PR description and SPEC still claim DashPaySdkWriter is an injectable test seam, but the trait was removed_ no longer present.

packages/rs-platform-wallet/src/manager/identity_sync.rs
QuantumExplorer: Acknowledged — this is the D1-minimal trade-off the description defers: the port copied the guard rather than extracting the shared runner. Tracking the `LoopCancelGuard` extraction (one generation-guarded cancel-slot helper shared by identity_sync / dashpay_sync / platform_address_sync) as a follow-up to keep this PR review-scoped.

PR #4002, opened 2026-07-04
fix(kotlin-sdk): address Android JNI review issues

QuantumExplorer:
## Summary
- split Android cargo feature flags into distinct argv entries so cargo-ndk passes shielded builds correctly
- wrap daemon-attached JNI callbacks in local frames, including per-entry frames for large persistence/event batches
- expand Kotlin SDK workflow path filters to include Rust workspace files and dependent crates

## Verification
- bash -n packages/kotlin-sdk/build_android.sh
- confirmed feature argv expands as separate --features / shielded arguments
- cargo check -p rs-unified-sdk-jni --no-default-features
- cargo check -p rs-unified-sdk-jni --features shielded --no-default-features
- git diff --check

Gradle/Kotlin verification was not run locally because this machine has no Java runtime on PATH.

### DIFF
--- a/.github/workflows/kotlin-sdk-build.yml
+++ b/.github/workflows/kotlin-sdk-build.yml
@@ -6,9 +6,16 @@ on:
   pull_request:
     paths:
       - 'packages/kotlin-sdk/**'
+      - 'Cargo.toml'
+      - 'Cargo.lock'
+      - 'packages/rs-*/**'
       - 'packages/rs-unified-sdk-jni/**'
       - 'packages/rs-sdk-ffi/**'
       - 'packages/rs-platform-wallet-ffi/**'
+      - 'packages/dapi-grpc/**'
+      - 'packages/data-contracts/**'
+      - 'packages/*-contract/**'
+      - 'packages/simple-signer/**'
       - '.github/workflows/kotlin-sdk-build.yml'
 
 concurrency:
--- a/packages/rs-unified-sdk-jni/src/persistence.rs
+++ b/packages/rs-unified-sdk-jni/src/persistence.rs
@@ -238,7 +238,7 @@ where
         let ctx = &*(context as *const KotlinPersistenceCtx);
         let bridge = ctx.bridge.as_obj();
         let env: &mut JNIEnv = &mut env;
-        match f(env, bridge) {
+        match env.with_local_frame(64, |env| f(env, bridge)) {
             Ok(code) => code,
             Err(_) => {
                 let _ = env.exception_clear();
@@ -264,7 +264,7 @@ where
         let ctx = &*(context as *const KotlinPersistenceCtx);
         let bridge = ctx.bridge.as_obj();
         let env: &mut JNIEnv = &mut env;
-        match f(env, bridge) {
+        match env.with_local_frame(64, |env| f(env, bridge)) {
             Ok(v) => Some(v),
             Err(_) => {
                 let _ = env.exception_clear();
@@ -381,10 +381,10 @@ unsafe extern "C" fn tramp_persist_address_balances(
         let wid = id32(env, wallet_id)?;
         let slice = slice_or_empty(entries, count);
         for e in slice {
-            let PlatformAddressFFI { address_type, hash } = e.address;
-            let hash_arr = env.byte_array_from_slice(&hash)?;
-            let code = env
-                .call_method(
+            let code = env.with_local_frame(16, |env| {
+                let PlatformAddressFFI { address_type, hash } = e.address;
+                let hash_arr = env.byte_array_from_slice(&hash)?;
+                env.call_method(
                     bridge,
                     "onPersistAddressBalance",
                     "([BB[BJIII)I",
@@ -398,7 +398,8 @@ unsafe extern "C" fn tramp_persist_address_balances(
                         JValue::Int(e.address_index as i32),
                     ],
                 )?
-                .i()?;
+                .i()
+            })?;
             if code != 0 {
                 return Ok(code);
             }
@@ -471,7 +472,9 @@ unsafe extern "C" fn tramp_persist_account_registrations(
     with_bridge(context, |env, bridge| {
         let wid = id32(env, wallet_id)?;
         for s in slice_or_empty(specs, count) {
-            let code = call_persist_account_registration(env, bridge, &wid, s)?;
+            let code = env.with_local_frame(16, |env| {
+                call_persist_account_registration(env, bridge, &wid, s)
+            })?;
             if code != 0 {
                 return Ok(code);
             }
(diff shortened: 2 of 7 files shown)


thepastaclaw:
✅ Review complete (commit c3734134)

packages/rs-unified-sdk-jni/src/signer.rs
```diff
@@ -103,9 +108,11 @@ unsafe extern "C" fn sign_async_trampoline(
         );
         if dispatched.is_err() {
             let _ = env.exception_clear();
+            let _ = env.pop_local_frame(&JObject::null());
             // Reclaim the token we just leaked and fail the request.
             return Err(Box::from_raw(token as *mut PendingSign));
         }
+        let _ = env.pop_local_frame(&JObject::null());
         Ok(())
```
thepastaclaw, line 116: **🟡 Suggestion: Use `with_local_frame` in `sign_async_trampoline` for unwind-safe frame cleanup**

Every other trampoline in this PR (and `can_sign_trampoline` right below in the same file at line 145) uses `env.with_local_frame(cap, |env| { ... })`, which guarantees the frame is popped on every exit path — including a Rust unwind, since `with_local_frame` pops before resuming. `sign_async_trampoline` alone hand-brackets `push_local_frame(16)` with three separate `pop_local_frame` calls (lines 93, 111, 115). The outer `catch_unwind` at line 69 will swallow any panic between the push and the final pop, and the pushed JNI local frame stays on the daemon-attached Tokio worker for the lifetime of the thread. Because that thread is reused for every state-transition signature, repeated failures accumulate leaked frames on a live worker.

The reason this isn't already using `with_local_frame` is that `Box::into_raw(pending)` must survive across the JNI dispatch so it can be reclaimed on failure. That's expressible in the closure form by returning `Result<(), Box<PendingSign>>` from the closure — the raw pointer itself is unaffected by the frame pop. See Codex's suggested rewrite below.

```suggestion
        let token = Box::into_raw(pending) as jlong;
        let dispatched = env.with_local_frame(16, |env| -> Result<(), jni::errors::Error> {
            let jpubkey = env.byte_array_from_slice(pubkey)?;
            let jdata = env.byte_array_from_slice(payload)?;
            env.call_method(
                ctx.bridge.as_obj(),
                "signAsync",
                "([BI[BJ)V",
                &[
                    (&jpubkey).into(),
                    (key_type as jint).into(),
                    (&jdata).into(),
                    token.into(),
                ],
            )?;
            Ok(())
        });
        if dispatched.is_err() {
            let _ = env.exception_clear();
            // Reclaim the token we just leaked and fail the request.
            return Err(Box::from_raw(token as *mut PendingSign));
        }
```

packages/rs-unified-sdk-jni/src/persistence.rs
```diff
@@ -552,7 +555,8 @@ unsafe extern "C" fn tramp_persist_account_address_pools(
                             (&path).into(),
                         ],
                     )?
-                    .i()?;
+                    .i()
+                })?;
```
thepastaclaw, line 559: **💬 Nitpick: `user`/`friend` pool identity arrays reallocated per address**

Before the inner per-address `with_local_frame(32, ...)` was added, `spec.user_identity_id` and `spec.friend_identity_id` were pool-scoped and could be built once per pool and reused across every address in the pool. With the inner frame at line 528, both 32-byte arrays are rebuilt for every address entry — an extra pair of `byte_array_from_slice` upcalls and JVM allocations per address. The correctness win (bounded local-ref usage) is worth it, but if pools with many addresses are common the shared identity arrays could be hoisted to a pool-level frame while the per-address frame handles only the entry-specific refs. Not blocking; the current shape is simpler.

<sub>source: ['claude']</sub>

.github/workflows/kotlin-sdk-build.yml
```diff
@@ -6,9 +6,16 @@ on:
   pull_request:
     paths:
       - 'packages/kotlin-sdk/**'
+      - 'Cargo.toml'
+      - 'Cargo.lock'
+      - 'packages/rs-*/**'
       - 'packages/rs-unified-sdk-jni/**'
       - 'packages/rs-sdk-ffi/**'
       - 'packages/rs-platform-wallet-ffi/**'
```
thepastaclaw, line 14: **💬 Nitpick: `packages/rs-*/**` already subsumes the three specific `rs-*` paths**

The umbrella glob at line 11 (`packages/rs-*/**`) already matches `packages/rs-unified-sdk-jni/**`, `packages/rs-sdk-ffi/**`, and `packages/rs-platform-wallet-ffi/**`. GitHub Actions unions the patterns so this is harmless, but the redundancy suggests the three specific entries are load-bearing when they aren't. Drop lines 12–14.

```suggestion
      - 'packages/rs-*/**'
```

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

PR is a well-scoped defensive fix that wraps JNI callbacks in `with_local_frame` to prevent local-ref table overflow on Android's daemon-attached signer/persistence threads, plus a real bash argv-splitting fix in `build_android.sh` and expanded CI trigger paths. Three reviewers converged on the same issue: `sign_async_trampoline` uses manual `push_local_frame`/`pop_local_frame` instead of the RAII closure form used everywhere else in the PR — currently correct but not unwind-safe under the outer `catch_unwind`. Two minor nits on redundant per-address allocations and workflow trigger paths.

Source: reviewers `opus` for general, `gpt-5.5` for general, `opus` for rust-quality, `gpt-5.5` for rust-quality, `opus` for ffi-engineer, `gpt-5.5` for ffi-engineer; verifier `opus`.

<details>

```

In `packages/rs-unified-sdk-jni/src/signer.rs`:
- [SUGGESTION] packages/rs-unified-sdk-jni/src/signer.rs:84-116: Use `with_local_frame` in `sign_async_trampoline` for unwind-safe frame cleanup
  Every other trampoline in this PR (and `can_sign_trampoline` right below in the same file at line 145) uses `env.with_local_frame(cap, |env| { ... })`, which guarantees the frame is popped on every exit path — including a Rust unwind, since `with_local_frame` pops before resuming. `sign_async_trampoline` alone hand-brackets `push_local_frame(16)` with three separate `pop_local_frame` calls (lines 93, 111, 115). The outer `catch_unwind` at line 69 will swallow any panic between the push and the final pop, and the pushed JNI local frame stays on the daemon-attached Tokio worker for the lifetime of the thread. Because that thread is reused for every state-transition signature, repeated failures accumulate leaked frames on a live worker.

  The reason this isn't already using `with_local_frame` is that `Box::into_raw(pending)` must survive across the JNI dispatch so it can be reclaimed on failure. That's expressible in the closure form by returning `Result<(), Box<PendingSign>>` from the closure — the raw pointer itself is unaffected by the frame pop. See Codex's suggested rewrite below.
```
</details>

PR #4004, opened 2026-07-05
fix(platform-wallet): reflect asset-lock top-up balance once, not doubled

QuantumExplorer:

QA **ADDR-09**: after a successful "Top Up from Core" (Core L1 asset lock → Platform credits on a DIP-17 platform-payment address), the SwiftExampleApp Platform Balance showed **~2×** the topped-up amount until a manual Sync-tab **Clear** + **Sync Now** corrected it. The on-chain balance was always correct — this was a client-side balance-sync bug in `platform-wallet`.

## What was done?

Root cause: `AddressFundingFromAssetLockTransition` records the recipient credit on-chain as `AddBalanceToAddress`, i.e. a **delta** (`AddToCredits`) in Drive's recent-address-balance-changes tree. The fund path reconciled the proof-attested **absolute** balance `X` into the provider's committed `found` seed but left the **incremental sync watermark stale**. The next *incremental* BLAST pass seeded `result.found` from `current_balances()` (already `X`) and then re-applied the recent `AddToCredits(X)` delta on top → `X + X = 2X`. A **full** scan (what the manual Clear forces) seeds absolutely from the tree, which is why Clear + resync fixed it.

Fix:
- New `PlatformPaymentAddressProvider::invalidate_sync_watermark()` — zeroes the incremental watermark only (`sync_height`/`sync_timestamp`/`last_known_recent_block`), **keeping** the just-reconciled `found` seed so display and `auto_select_inputs` budgeting keep `X` visible. With `sync_timestamp == 0`, `last_sync_timestamp()` returns `None`, so the next pass takes the **full-scan** branch and reconciles to `X`. This is the automated equivalent of the manual **Clear + Sync Now**.
- Call it from `fund_from_asset_lock` right after `reconcile_address_infos` — covers both the fresh and resume top-up paths (they share one Rust fn).
- Added an `ADDR-09` regression-guard row to `SwiftExampleApp/TEST_PLAN.md`.

Durability is intentionally **in-memory only**: the persisted watermark can't be reset to 0 through the normal changeset (`PlatformAddressChangeSet::merge` combines `sync_height` with `.max()`, and the persister only fires `on_persist_sync_state_fn` when a component is `> 0`). The next in-session BLAST pass (~15s) full-scans, reconciles to `X`, and persists a correct **forward** watermark, so durable state self-corrects within ~15s. The only residual gap — an app kill inside that window — is documented in-code and accepted rather than over-engineering against the merge/gate.

## How Has This Been Tested?

- New provider unit test `invalidate_sync_watermark_forces_full_scan_keeps_seed` — **passes** (asserts all three watermark scalars zero, `last_sync_timestamp() == None` (the full-scan trigger), and the `found` seed survives with balance `X`).
- Full `platform_addresses` module suite green (78 tests); `cargo fmt --all --check` and `cargo clippy -p platform-wallet --all-targets` clean.
- Simulator xcframework (`build_ios.sh --target sim`) + SwiftExampleApp build green; app launches with the fix and renders the Platform Balance single-counted.
- A rigorous behavioral A/B top-up repro needs a **Core-funded** wallet (the QA sim wallet had 0 Core balance); that end-to-end pass is the ADDR-09 QA step with a funded wallet. The fix's behavioral correctness reduces to "invalidating the watermark triggers the full-scan reconcile," which the unit test proves and which is the exact known-good manual workaround.

## Breaking Changes

None. Client-side balance-sync fix only; no consensus or protocol change.

## Related / follow-up

`transfer.rs` and `withdrawal.rs` route through the same `reconcile_address_infos` seam and also emit `AddBalanceToAddress` deltas. Same-wallet is self-consistent (paired `SetBalanceToAddress` on the source), but **cross-wallet transfers** are a plausible candidate for the same double-count — out of scope here, flagged for a separate investigation.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a platform address balance issue where certain asset-lock top-ups could be counted twice during sync.
  * Next sync now performs a full reconciliation when needed, preventing incorrect incremental updates.

* **Tests**
  * Added coverage to verify sync invalidation triggers a full scan while preserving the already reconciled balance state.

* **Documentation**
  * Updated the QA test plan with a new regression case and clear pass/fail criteria for the affected balance flow.


thepastaclaw:
✅ Review complete (commit 180da720)

QuantumExplorer (commented):
Approved

packages/rs-platform-wallet/src/wallet/platform_addresses/fund_from_asset_lock.rs
```diff
@@ -293,6 +293,51 @@ impl PlatformAddressWallet {
...
+        // double-count could briefly reappear until the next full rescan
+        // (or a manual Clear). That narrow window is accepted here rather
+        // than over-engineering a durable invalidation against the
+        // `.max()` merge / `> 0` gate.
+        {
+            let mut guard = self.provider.write().await;
+            if let Some(provider) = guard.as_mut() {
+                provider.invalidate_sync_watermark();
+            }
+        }
```
thepastaclaw, line 339: **🟡 Suggestion: Watermark invalidation is not atomic with `reconcile_address_infos`**

`reconcile_address_infos` at line 292-294 takes `self.provider.write()` internally, commits the reconciled `found = X` seed, and drops that guard before returning. The block at 334-339 then re-acquires `self.provider.write()` to call `invalidate_sync_watermark`. Meanwhile `sync_balances` in `sync.rs:115` also acquires `self.provider.write()` for the entire sync round — including the SDK network call at lines 133-136 — and calls `provider.last_sync_timestamp()` before seeding `before` from `current_balances(&*provider)`. tokio's `RwLock` is fair, so a background BLAST sync waiting on the lock during reconciliation can be granted the lock between the two acquisitions here. In that interleaving it will observe (a) the freshly committed `found = X` and (b) the still-non-zero watermark, take the incremental branch in `sync_address_balances`, re-apply the recent `AddToCredits(X)` delta, and persist `X + X = 2X` — the exact reading this PR claims to prevent. The subsequent invalidation still causes a full-scan reconcile within ~15s, so worst-case is a brief reappearance of the pre-PR bug rather than a regression, but the fix is not race-free the way the comment implies. A race-free variant would fold the watermark zeroing into the same critical section as the reconciliation commit — for example a fund-specific reconcile helper, or an `invalidate_watermark_after: bool` flag on `reconcile_address_infos` so transfer/withdrawal don't trigger it. Worth doing, since the whole point of this PR is to eliminate the ability to observe the double-count from a fresh top-up.

packages/rs-platform-wallet/src/wallet/platform_addresses/provider.rs
```diff
@@ -520,6 +520,46 @@ impl PlatformPaymentAddressProvider {
...
+    /// visible during the ~15s until the reconciling scan completes,
+    /// instead of the momentary zero `reset_sync_state` would show.
+    ///
+    /// This is the in-memory equivalent of the manual Sync-tab
+    /// "Clear" + "Sync Now" that also fixes the double-count.
+    pub(crate) fn invalidate_sync_watermark(&mut self) {
+        self.sync_height = 0;
+        self.sync_timestamp = 0;
+        self.last_known_recent_block = 0;
+    }
```
thepastaclaw, line 561: **💬 Nitpick: `reset_sync_state` and `invalidate_sync_watermark` duplicate the three-scalar zeroing**

Both `reset_sync_state` (line 510) and the new `invalidate_sync_watermark` (line 557) zero the same three watermark fields (`sync_height`, `sync_timestamp`, `last_known_recent_block`). The two operations are semantically nested — Clear = invalidate watermark + drop `found`/`absent` + drop `per_wallet_in_sync` — so `reset_sync_state` could compose on top of `invalidate_sync_watermark` to keep the watermark reset single-sourced. If the watermark ever grows a fourth scalar, forgetting it in either function would silently reintroduce an ADDR-09-like bug or regress Clear. Not a blocker.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Focused ADDR-09 fix: after `reconcile_address_infos` commits the proof-attested absolute balance, `fund_from_asset_lock` invalidates the incremental sync watermark so the next BLAST pass full-scans instead of re-applying the recent delta on top of the seeded balance. The mechanism is correct, well-documented, and covered by a targeted unit test. Reviewers converged on one substantive concern: the reconciliation commit and the watermark invalidation happen in two separate provider write-lock critical sections, leaving a narrow window in which a queued background `sync_balances` can grab the lock between them, take the incremental branch, and briefly reproduce the 2X reading — self-healing on the next full scan within ~15s.

<details>

```

In `packages/rs-platform-wallet/src/wallet/platform_addresses/fund_from_asset_lock.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/platform_addresses/fund_from_asset_lock.rs:292-339: Watermark invalidation is not atomic with `reconcile_address_infos`
  `reconcile_address_infos` at line 292-294 takes `self.provider.write()` internally, commits the reconciled `found = X` seed, and drops that guard before returning. The block at 334-339 then re-acquires `self.provider.write()` to call `invalidate_sync_watermark`. Meanwhile `sync_balances` in `sync.rs:115` also acquires `self.provider.write()` for the entire sync round — including the SDK network call at lines 133-136 — and calls `provider.last_sync_timestamp()` before seeding `before` from `current_balances(&*provider)`. tokio's `RwLock` is fair, so a background BLAST sync waiting on the lock during reconciliation can be granted the lock between the two acquisitions here. In that interleaving it will observe (a) the freshly committed `found = X` and (b) the still-non-zero watermark, take the incremental branch in `sync_address_balances`, re-apply the recent `AddToCredits(X)` delta, and persist `X + X = 2X` — the exact reading this PR claims to prevent. The subsequent invalidation still causes a full-scan reconcile within ~15s, so worst-case is a brief reappearance of the pre-PR bug rather than a regression, but the fix is not race-free the way the comment implies. A race-free variant would fold the watermark zeroing into the same critical section as the reconciliation commit — for example a fund-specific reconcile helper, or an `invalidate_watermark_after: bool` flag on `reconcile_address_infos` so transfer/withdrawal don't trigger it. Worth doing, since the whole point of this PR is to eliminate the ability to observe the double-count from a fresh top-up.
```
</details>

Source: reviewers `opus` (claude-general), `gpt-5.5` (codex-general), `opus` (claude-security-auditor), `gpt-5.5` (codex-security-auditor), `opus` (claude-rust-quality), `gpt-5.5` (codex-rust-quality), `opus` (claude-ffi-engineer), `gpt-5.5` (codex-ffi-engineer); verifier `opus`.

PR #4005, opened 2026-07-05
fix(platform-wallet): gate ADDR-09 watermark invalidation inside the reconcile seam

QuantumExplorer:

Follow-up to the ADDR-09 asset-lock top-up double-credit fix (#4004), generalizing it to **every** flow that credits a wallet-owned address.

Any transition that credits an address does so via an on-chain `AddBalanceToAddress` op — a DELTA (`AddToCredits`) in Drive's recent-address-balance-changes tree — while inputs are recorded as absolute `SetBalanceToAddress` ops. When `reconcile_address_infos` commits the proof-attested ABSOLUTE post-balance into the provider's `found` seed without advancing the incremental sync watermark, the next incremental BLAST pass re-applies the recent delta on top of the already-absolute seed → `X + delta` — the ADDR-09 double-count.

#4004 patched this caller-side for the asset-lock top-up only. An adversarial review of the first cut of this PR (which patched only `transfer`, also caller-side) confirmed the same live bug in more flows and a race window in the caller-side pattern itself:

- **`transfer`** (this PR's original target) — a same-wallet output (change address, or a transfer to one of the wallet's own addresses) double-counts. CONFIRMED.
- **`transfer_credits_to_addresses`** — recipients are delta-credited, and the primary use of this flow is consolidating identity credits into the wallet's **own** DIP-17 addresses. CONFIRMED, previously unfixed.
- **`register_from_addresses`** — the optional refund-style output is delta-credited; reachable whenever it lands on an owned address. CONFIRMED mechanism.
- **Race window** — invalidating *after* `reconcile_address_infos` returns (the #4004 caller-side pattern) leaves a gap between the seam releasing the provider lock and the invalidation re-acquiring it, in which a ~15s-cadence background sync could run incrementally with the stale watermark and transiently commit 2×.

Verified NOT affected: **cross-wallet transfers** (each wallet's provider tracks only its own wallet via `from_wallets([self.wallet_id])`; the recipient wallet's own sync applies the delta exactly once), **withdrawals** and **`top_up_from_addresses`** (input-only drains — absolute ops, idempotent under re-apply; no change output today).

## What was done?

**Moved the gate into the seam** instead of patching callers one by one:

- `PlatformAddressWallet::reconcile_address_infos` now takes a `credited_outputs: &BTreeSet<PlatformP2PKHAddress>` parameter naming the addresses the transition credited via deltas (built with the new `credited_outputs_set` helper). When a **committed** entry matches, the seam calls `invalidate_sync_watermark()` **inside its provider-lock critical section** — no sync pass can interleave between the seed commit and the invalidation, closing the race window — forcing the next BLAST pass to full-scan-reconcile.
- The gate keys on entries actually committed, not merely requested: an entry skipped as stale/unchanged means a sync already applied that credit (and advanced the watermark past it), so the fast incremental cadence is preserved.
- All six callers audited and updated: `transfer` (outputs), `fund_from_asset_lock` (recipients; replaces #4004's caller-side unconditional invalidation), `transfer_credits_to_addresses` (recipients), `register_from_addresses` (optional output), `top_up_from_addresses` + `withdraw` (empty set, with comments directing any future change output through the gate).
- Mandatory parameter (not an `Option`) so a future caller cannot silently forget the ADDR-09 analysis — every call site must state its credited-output set.
- The ADDR-09 mechanism + durability notes are consolidated onto `PlatformPaymentAddressProvider::invalidate_sync_watermark`, whose single call site is now the seam gate.

## How Has This Been Tested?

- `cargo test -p platform-wallet --lib` — 242 passed, 0 failed.
- New seam-level tests (full wallet + capturing-persister fixture): gate fires on a committed credited output and keeps the reconciled seed; watermark preserved for input-only reconciliation; watermark preserved when the credited entry was skipped as already applied. #4004's provider-level `invalidate_sync_watermark_forces_full_scan_keeps_seed` still covers the invalidation effect itself.
- `cargo clippy -p platform-wallet --all-targets` — clean; `cargo check -p platform-wallet-ffi` — compiles (no FFI surface change).
- `cargo fmt --all`.

## Breaking Changes

None on-chain / consensus. `reconcile_address_infos` (a `pub` Rust API on `PlatformAddressWallet`, not exposed over FFI) gains a required parameter; all in-repo callers updated.

## Checklist:

### DIFF
--- a/packages/rs-platform-wallet/src/wallet/platform_addresses/transfer.rs
+++ b/packages/rs-platform-wallet/src/wallet/platform_addresses/transfer.rs
@@ -212,6 +212,14 @@ impl PlatformAddressWallet {
         // gate and the spend path diverge on a non-latest-pinned SDK.
         let version = platform_version.unwrap_or_else(|| self.sdk.version());
 
+        // Capture the credited output addresses BEFORE `outputs` is moved
+        // into the SDK call below. Transfer outputs are recorded on-chain
+        // as `AddBalanceToAddress` DELTAS (a same-wallet output — e.g. a
+        // change address — is the ADDR-09 double-count shape); the seam's
+        // watermark gate needs to know them. See
+        // `reconcile_address_infos` for the full mechanism.
+        let credited_outputs = super::credited_outputs_set(outputs.keys());
+
         let address_infos = match input_selection {
             InputSelection::Explicit(inputs) => {
                 if inputs.is_empty() {
@@ -265,9 +273,13 @@ impl PlatformAddressWallet {
         // `transfer_address_funds` returns address info for the full
         // `inputs ∪ outputs` set, including external recipients the wallet
         // does not own — the shared seam filters those out, applies the
-        // proof-attested balances, updates the sync seed, and persists.
+        // proof-attested balances, updates the sync seed, persists, and
+        // (when a wallet-owned output in `credited_outputs` was committed)
+        // invalidates the incremental sync watermark so the next BLAST
+        // pass full-scan-reconciles instead of re-applying the on-chain
+        // credit delta on top of the absolute seed (ADDR-09).
         Ok(self
-            .reconcile_address_infos(&address_infos, "address transfer")
+            .reconcile_address_infos(&address_infos, &credited_outputs, "address transfer")
             .await)
     }
 
--- a/packages/rs-platform-wallet/src/wallet/platform_addresses/fund_from_asset_lock.rs
+++ b/packages/rs-platform-wallet/src/wallet/platform_addresses/fund_from_asset_lock.rs
@@ -289,54 +289,22 @@ impl PlatformAddressWallet {
         // Persistence errors are logged inside the seam rather than
         // propagated: Platform already accepted the transition, and a
         // persistence hiccup shouldn't mask that.
-        let cs = self
-            .reconcile_address_infos(&address_infos, "fund from asset lock")
-            .await;
-
-        // ADDR-09: force the next BLAST sync to full-scan-reconcile
-        // instead of applying an incremental delta.
         //
-        // `reconcile_address_infos` above set the provider's committed
-        // `found` seed to the proof-attested ABSOLUTE balance `X`, but the
-        // top-up is recorded on-chain as a DELTA (`AddBalanceToAddress` →
-        // `AddToCredits`) in Drive's recent-address-balance-changes tree,
-        // and we did NOT advance the incremental watermark. An incremental
-        // next pass would seed `result.found` from `current_balances()`
-        // (already `X`) and then re-apply that recent `AddToCredits(X)`
-        // delta from the stale watermark, landing at `X + X = 2X` — the
-        // ADDR-09 double-count. Zeroing the in-memory watermark makes
-        // `last_sync_timestamp()` return `None`, so the next pass full-scans
-        // (absolute seed from the tree, catch-up from the fresh checkpoint)
-        // and reconciles to the correct `X`. This is the automated
+        // ADDR-09: every recipient of an asset-lock top-up is credited via
+        // an on-chain `AddBalanceToAddress` DELTA, so the whole recipient
+        // set goes into the seam's `credited_outputs` gate. Committing
+        // their proof-attested ABSOLUTE balances while the incremental
+        // watermark stayed stale would let the next incremental BLAST pass
+        // re-apply the delta on top → `X + X = 2X`, the ADDR-09
+        // double-count. The seam invalidates the watermark inside its own
+        // critical section (see `reconcile_address_infos` and
+        // `PlatformPaymentAddressProvider::invalidate_sync_watermark`),
+        // forcing the next pass to full-scan-reconcile — the automated
         // equivalent of the manual Sync-tab "Clear" + "Sync Now".
-        //
-        // Unlike transfer/withdrawal (which also route through the seam),
-        // an asset-lock top-up credits its recipient with a pure additive
-        // delta and no offsetting input for that same address, so updating
-        // the `found` seed alone does not neutralize the re-applied delta —
-        // hence this path-specific watermark invalidation.
-        //
-        // DURABILITY: in-memory only. The persisted sync watermark cannot
-        // be reset to 0 through the normal changeset —
-        // `PlatformAddressChangeSet::merge` combines `sync_height` with
-        // `.max()` and the persister only fires `on_persist_sync_state_fn`
-        // when a component is `> 0` — so a durable zero would have to fight
-        // both the merge and the `> 0` gate. Instead we rely on the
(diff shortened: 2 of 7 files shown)


thepastaclaw:
✅ Review complete (commit 93517330)

packages/rs-platform-wallet/src/wallet/platform_addresses/transfer.rs
```diff
@@ -212,6 +212,29 @@ impl PlatformAddressWallet {
...
+        // address, or a transfer to one of the wallet's own addresses) is
+        // owned by the source wallet. See the watermark-invalidation block
+        // after the reconcile below.
+        let output_addresses: BTreeSet<PlatformP2PKHAddress> = outputs
+            .keys()
+            .filter_map(|addr| match addr {
+                PlatformAddress::P2pkh(hash) => Some(PlatformP2PKHAddress::new(*hash)),
+                _ => None,
+            })
+            .collect();
```
thepastaclaw: **💬 Nitpick: P2sh output addresses are silently dropped from the invalidation gate**

`output_addresses` is built via `filter_map` that only keeps `PlatformAddress::P2pkh` and drops `PlatformAddress::P2sh`. This is correct today because `PlatformAddressBalanceEntry.address` is a `PlatformP2PKHAddress` and wallets do not own P2sh outputs, so the invariant holds — but the coupling is enforced silently by the filter rather than by a type check, `debug_assert!`, or doc comment. If wallet-owned P2sh support is ever added, an owned P2sh change output would reintroduce the same double-count with no signal. Consider a `debug_assert!(matches!(addr, PlatformAddress::P2pkh(_)))` in the collector or a comment on the collector documenting the coupling to `PlatformAddressBalanceEntry`'s P2PKH-only shape.

<sub>source: ['claude']</sub>

thepastaclaw (changes requested):
## Code Review

The address-funds transfer path is patched correctly, but the sibling identity-to-address transfer wrapper at platform_wallet.rs:342-344 still has the same optimistic-absolute-seed vs incremental-delta double-count — `IdentityCreditTransferToAddressesTransitionAction` emits `AddBalanceToAddress` for every recipient, so a recipient owned by the calling wallet reproduces the exact bug this PR set out to fix. The reconcile/invalidate atomicity race is real but inherited unchanged from the merged #4004 asset-lock pattern (which documents acceptance of the same transient window), so it does not block this PR.

1 blocking | 1 nitpick(s)

Source: reviewers: reviewer-claude-general-4005-1783237422 (opus), reviewer-codex-general-4005-1783237422 (gpt-5.5), reviewer-claude-security-auditor-4005-1783237422 (opus), reviewer-codex-security-auditor-4005-1783237422 (gpt-5.5), reviewer-claude-rust-quality-4005-1783237422 (opus), reviewer-codex-rust-quality-4005-1783237422 (gpt-5.5); verifier: verifier-4005-1783237422 (opus).

<details>

```

- [BLOCKING] packages/rs-platform-wallet/src/wallet/platform_wallet.rs:342-344: Same double-count on identity-to-address transfer to an owned recipient
  `transfer_credits_to_addresses_with_external_signer` calls `reconcile_address_infos` without invalidating the sync watermark, so it hits the exact ADDR-09-sibling bug this PR set out to close. `IdentityCreditTransferToAddressesTransitionAction::into_high_level_drive_operations` (packages/rs-drive/src/state_transition_action/action_convert_to_operations/identity/identity_credit_transfer_to_addresses_transition.rs:42-49) emits `AddressFundsOperationType::AddBalanceToAddress` — a DELTA in Drive's recent-address-balance-changes tree — for every entry of `recipient_addresses`. When the caller passes one of this wallet's own addresses as a recipient (a straightforward, public-API-reachable scenario: identity → own platform address), the seam writes the proof-attested absolute post-balance into the provider's `found` seed but leaves the incremental watermark stale. The next incremental BLAST pass re-applies that `AddToCredits` delta on top of the already-absolute seed → `X + delta` over-count, identical to the mechanism transfer.rs now guards against. The narrower `platform_addresses/transfer.rs` guard does not cover this wrapper, so the PR's stated scope (invalidate on same-wallet transfer credit) is incomplete. Apply the same reconcile-credited-owned-output gate + `invalidate_sync_watermark()` here (or fold the decision into `reconcile_address_infos`), keyed on the recipient set overlapping with wallet-owned addresses in the returned `AddressInfos`.

In `packages/rs-platform-wallet/src/wallet/platform_addresses/transfer.rs`:
- [NITPICK] packages/rs-platform-wallet/src/wallet/platform_addresses/transfer.rs:230-236: P2sh output addresses are silently dropped from the invalidation gate
  `output_addresses` is built via `filter_map` that only keeps `PlatformAddress::P2pkh` and drops `PlatformAddress::P2sh`. This is correct today because `PlatformAddressBalanceEntry.address` is a `PlatformP2PKHAddress` and wallets do not own P2sh outputs, so the invariant holds — but the coupling is enforced silently by the filter rather than by a type check, `debug_assert!`, or doc comment. If wallet-owned P2sh support is ever added, an owned P2sh change output would reintroduce the same double-count with no signal. Consider a `debug_assert!(matches!(addr, PlatformAddress::P2pkh(_)))` in the collector or a comment on the collector documenting the coupling to `PlatformAddressBalanceEntry`'s P2PKH-only shape.
```
</details>

thepastaclaw (approved):
## Code Review

Both prior findings resolved. The blocking identity-to-address double-count is fixed by threading credited_outputs from recipient_addresses.keys() into reconcile_address_infos (platform_wallet.rs:344-370), and the P2SH-drop nitpick is addressed by the credited_outputs_set helper's docstring documenting the P2PKH-only wallet invariant (platform_addresses/mod.rs:44-63). Bonus: the invalidation gate now lives inside reconcile_address_infos under the provider write lock (wallet.rs:369-384), closing the reconcile/invalidate atomicity race that was previously out-of-scope. No new in-scope defects in the latest delta.

Prior finding reconciliation:
- FIXED: `transfer_credits_to_addresses_with_external_signer` now passes recipient credited outputs into `reconcile_address_infos`, so owned identity-to-address recipients invalidate the watermark inside the reconcile critical section.
- OUTDATED: the P2SH-filter nitpick is covered by the shared `credited_outputs_set` helper's doc comment documenting the P2PKH-only wallet-owned address invariant.

Source: reviewers claude general `opus`, codex general `gpt-5.5`, claude security-auditor `opus`, codex security-auditor `gpt-5.5`, claude rust-quality `opus`, codex rust-quality `gpt-5.5`; verifier claude `opus`.

packages/rs-platform-wallet/src/wallet/platform_addresses/transfer.rs
thepastaclaw: Resolved in this update — _P2sh output addresses are silently dropped from the invalidation gate_ no longer present.

PR #3991, opened 2026-07-03
fix(platform-wallet): address sync review feedback for #3990

thepastaclaw:
Follow-up fixes for CodeRabbit feedback on #3990.\n\nThis branch addresses the platform wallet sync/reconciliation, shielded pending spend cleanup, Swift clear/sync race, and best-effort identity persistence review comments.\n\nValidation:\n- cargo fmt --manifest-path packages/rs-platform-wallet/Cargo.toml --check\n- cargo check -p platform-wallet --features shielded\n\nNote: I could not push directly to dashpay/platform:chore/sync-v4.1-dev-with-v4.0-dev (403/permission denied), so this PR targets that branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved wallet reconciliation so balance updates are only finalized when local changes are safely saved.
  * Prevented stale or duplicate shielded spend records from reappearing after wallet resets or clears.
  * Made balance syncing more resilient during clear operations, avoiding old data from repopulating the UI.
  * Tightened handling of address and spend reconciliation conflicts to reduce inconsistent wallet states.

* **New Features**
  * Added safer clearing behavior in the example app, with the Clear action disabled while cleanup is in progress.

PR #3990, opened 2026-07-03
chore: sync v4.1-dev with v4.0-dev

QuantumExplorer:
## Issue being fixed or feature implemented

Routine sync merge (same pattern as #3978): brings the 4.0.x release line's fixes up to the v4.1 development branch.

## What was done?

Merged `v4.0-dev` into `v4.1-dev` — a clean merge, no conflicts. Carries:

- #3959 — Platform Sync "Clear" actually clears synced data
- #3969 — platform-address balance reconciliation after top-up (phantom-balance fix)
- #3977 — shielded spends build against a Platform-recorded anchor
- #3982 — stranded shielded-spend reservation auto-release
- #3987 — unified platform-address reconciliation seam (`reconcile_address_infos`)
- #3988 — active re-drive of unconfirmed shielded spends (persist + rebroadcast ×3)
- #3979 — wasm-sdk protocol upgrade state fix

## How Has This Been Tested?

`cargo test -p platform-wallet --lib --features shielded` → 326 passed on the merged tree; CI runs the full suite.

## Breaking Changes

None (merge only).

## Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added or updated relevant unit/integration/functional/e2e tests
- [ ] I have added "!" to the title and described breaking changes in the corresponding section if my code contains any
- [ ] I have made corresponding changes to the documentation if needed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a new local reset action for platform address sync, clearing cached state and forcing a full rescan on next sync.
  * Improved wallet flows so identity top-ups, registrations, transfers, withdrawals, and shielded sends now reconcile balances more consistently after proofs.

* **Bug Fixes**
  * Added a clearer retryable error when a shielded spend lacks a recorded anchor.
  * Improved shielded spend handling so unconfirmed or pruned cases are retried or released more reliably.
  * Fixed stale sync completions so cleared or stopped sync states are no longer repainted in the UI.

PR #4008, opened 2026-07-05
fix(platform-wallet): data-integrity follow-ups from the #3990 sync review

QuantumExplorer:

Addresses the CodeRabbit **data-integrity** findings raised on the v4.0-dev → v4.1-dev sync (#3990). Those findings are against content that is now on `v4.1-dev`; they were never fixed — [#3991](https://github.com/dashpay/platform/pull/3991) prepared fixes but was orphaned/closed when its base (the sync branch) merged, and its diff no longer applies because the reconcile seam grew a `credited_outputs` argument in the ADDR-09 work (#4004/#4005) since. Each fix here is **re-derived against current `v4.1-dev`** and independently verified, not cherry-picked.

## What was done?

**Shielded store (durable redrive lifecycle — gaps from #3988):**
- `purge_wallet` / `purge_all_subwallets` now also delete the durable `shielded_pending_spends` rows. They were only clearing in-memory state, so a Clear / unregister left stale redrive rows that **rehydrate ghost reservations on the next open**. `reset_commitment_tree` deliberately does *not* touch them — a redrive is broadcast state, not tree state.
- `SubwalletState::mark_spent` resolves the reservation + redrive whenever the nullifier is **known**, not only on the first `unspent→spent` transition — so a note restored *already spent* still clears its rehydrated redrive. It returns `MarkSpentOutcome { newly_spent, dropped_redrives }` so the file store mirrors the exact SQLite deletion on the already-spent path too (targeted by activity id, so the common no-redrive path issues zero SQLite work).

**Address reconciliation seam:**
- `reconcile_address_infos` gains a `_with_persistence` variant that reports whether the changeset was durably stored. `fund_from_asset_lock` gates `consume_asset_lock` on it: an irreversible `Consumed` lock is never paired with balance rows that failed to persist (which would under-budget the next spend after a restart). The **public signature is unchanged** (thin discarding wrapper), so the other seven callers don't churn.
- `commit_reconciliation` index conflict: a **credit** whose derivation index already maps to a different address is dropped outright (committing it would evict the pairing via `BiBTreeMap::insert`, orphaning another address's `found`, or persist a seed `current_balances` can't round-trip); a **removal** still zeroes `found` and is emitted (so the durable zero-out lands and a stale balance can't resurrect) but skips the bijection merge.
- `reset_sync_state` holds the provider write lock across **both** the provider reset and the managed-account balance clear (wallet-manager write nested inside, matching the seam's provider→wallet-manager lock order), so the reset is atomic against a concurrent sync / reconciliation rather than leaving a window between two separate lock scopes.

**Identity registration:**
- `register_from_addresses` treats the post-acceptance local `add_identity` as best-effort (warn, don't propagate) — mirroring the transfer / top-up flows — so a local persistence failure no longer suppresses the `(identity, address_infos)` return the composite needs to reconcile spent funding addresses.

**SwiftExampleApp Clear races:**
- `PlatformBalanceSyncService` gains a `balanceSnapshotGeneration` token: a `refreshBalanceSnapshot` that was in flight when a Clear runs is dropped instead of republishing pre-clear balances over the cleared UI. It also gains an `isClearing` flag, and `CoreContentView` gates both Clear and Sync Now on it (Clear is fire-and-forget, so it could otherwise interleave with the Rust reset + SwiftData wipe).

## How Has This Been Tested?

- `cargo test -p platform-wallet --lib` → **244 passed** (default features); `--features shielded` → **362 passed**.
- **5 new regression tests**: `purge_clears_durable_redrive_rows_but_tree_reset_does_not`, `mark_spent_on_restored_spent_note_clears_durable_redrive`, `commit_reconciliation_index_conflict_still_emits_removal`, `commit_reconciliation_index_conflict_drops_credit`.
- `cargo fmt --all -- --check` clean; `cargo clippy -p platform-wallet --features shielded` clean (only pre-existing workspace-profile notes).
- The two Swift changes are **app-level** (no FFI signature change) and are compiled by the repo's **Swift SDK build + tests** CI check.

**Not covered (deferred / reasoned):** the lock-ordering (`reset_sync_state`) and best-effort (`register_from_addresses`) fixes are correctness-by-construction and covered by the existing integration suites rather than dedicated unit tests; the asset-lock consume gate's persist-failure branch needs an erroring-persister harness and is covered by reasoning + the plain-path integration test.

## Breaking Changes

None. `MarkSpentOutcome` is a `pub(super)` internal type; the public `reconcile_address_infos` signature is unchanged.

## Notes for reviewers

Original findings by CodeRabbit on #3990; fix approach cross-referenced with @thepastaclaw's #3991. I **dropped** the anchor-probe "skip-vs-break" finding — an earlier adversarial verification on #3977 refuted the harmful variant (deeper probes fail identically; the "fix" would make every routine mid-block spend probe all 100 depths) — and with it the collateral `ErrorShieldedMerkleWitnessUnavailable = 20` FFI code, since not adding an ABI code we don't need is strictly better.

## Checklist:

### DIFF
--- a/packages/rs-platform-wallet/src/wallet/platform_addresses/fund_from_asset_lock.rs
+++ b/packages/rs-platform-wallet/src/wallet/platform_addresses/fund_from_asset_lock.rs
@@ -302,11 +302,40 @@ impl PlatformAddressWallet {
         // forcing the next pass to full-scan-reconcile — the automated
         // equivalent of the manual Sync-tab "Clear" + "Sync Now".
         let credited_outputs = super::credited_outputs_set(addresses.keys());
-        let cs = self
-            .reconcile_address_infos(&address_infos, &credited_outputs, "fund from asset lock")
+        // Use the persistence-reporting variant: marking the lock
+        // `Consumed` below is irreversible, so it MUST be gated on the
+        // reconciled balances actually reaching disk. `persisted` is
+        // false ONLY when the in-memory balances were updated but the
+        // durable write failed — exactly the case where a Consumed lock
+        // would pair with stale rows and under-budget the next spend
+        // after a restart.
+        let (cs, persisted) = self
+            .reconcile_address_infos_with_persistence(
+                &address_infos,
+                &credited_outputs,
+                "fund from asset lock",
+            )
             .await;
 
         if let Some(out_point) = tracked_out_point {
+            if !persisted {
+                // The proof-attested balances were applied in memory but
+                // did not reach disk. Leave the lock non-Consumed: it
+                // stays in the Resumable Funding list, and a user Resume
+                // gets Platform's deterministic "lock already consumed"
+                // rejection — the same benign recovery path as a failed
+                // consume below — while the next platform-address sync
+                // repairs the stale rows. Consuming here would strand the
+                // lock as Consumed over durable balances that under-report
+                // the credit.
+                tracing::error!(
+                    outpoint = %out_point,
+                    "skipping consume_asset_lock: the reconciled balance changeset \
+                     was not durably stored; the lock stays non-Consumed (Resumable) \
+                     rather than pairing a Consumed lock with stale balance rows on disk"
+                );
+                return Ok(cs);
+            }
             // Platform DID accept the top-up — propagating an Err
             // here would misreport the protocol outcome, since the
             // caller's recipient(s) already have credits attested
--- a/packages/rs-platform-wallet/src/wallet/shielded/file_store.rs
+++ b/packages/rs-platform-wallet/src/wallet/shielded/file_store.rs
@@ -225,10 +225,34 @@ impl FileBackedShieldedStore {
         Ok(conn)
     }
 
+    /// Delete the single persisted redrive row for `id` keyed by
+    /// `activity_id`. Used to mirror the exact in-memory drops
+    /// [`SubwalletState::mark_spent`] reports, avoiding the
+    /// scan-every-row cost of [`Self::delete_redrive_rows_containing`]
+    /// on the common path where the resolved note had no armed redrive.
+    fn delete_redrive_row(
+        &self,
+        id: SubwalletId,
+        activity_id: &[u8; 32],
+    ) -> Result<(), FileShieldedStoreError> {
+        let conn = self.pending_conn.lock().expect("pending_conn mutex");
+        conn.execute(
+            "DELETE FROM shielded_pending_spends \
+             WHERE wallet_id = ?1 AND account_index = ?2 AND activity_id = ?3",
+            rusqlite::params![
+                id.wallet_id.as_slice(),
+                id.account_index,
+                activity_id.as_slice(),
+            ],
+        )
+        .map_err(|e| FileShieldedStoreError(format!("delete redrive row by activity: {e}")))?;
+        Ok(())
+    }
+
     /// Mirror to SQLite the redrive deletions [`SubwalletState`] performs
-    /// in memory when a nullifier resolves (`mark_spent` /
-    /// `clear_pending`): delete every persisted row for `id` whose
-    /// nullifier blob contains `nullifier`.
+    /// in memory when a nullifier resolves via `clear_pending`: delete
+    /// every persisted row for `id` whose nullifier blob contains
+    /// `nullifier`.
     fn delete_redrive_rows_containing(
         &self,
         id: SubwalletId,
@@ -295,16 +319,19 @@ impl ShieldedStore for FileBackedShieldedStore {
         let Some(sw) = self.subwallets.get_mut(&id) else {
             return Ok(false);
         };
-        let marked = sw.mark_spent(nullifier);
-        if marked {
-            // `SubwalletState::mark_spent` dropped any redrive carrying
-            // this nullifier from memory; mirror the deletions. The
-            // in-memory transition already happened, so a SQLite failure
-            // must not abort the call — log it and keep the trait
-            // behavior consistent. A surviving stale row rehydrates a
-            // reservation on the next open, which the reconcile / prune
-            // passes then clear.
-            if let Err(e) = self.delete_redrive_rows_containing(id, nullifier) {
+        let outcome = sw.mark_spent(nullifier);
+        // Mirror the durable deletion whenever the in-memory drop
+        // happened — keyed on the returned activity ids, NOT on
+        // `newly_spent`. A note restored already-spent still resolves a
+        // rehydrated redrive here (`newly_spent == false`), and leaving
+        // the SQLite row would resurrect the reservation on the next
+        // open. Targeting the exact activity ids means the common case
+        // (no armed redrive for this note) issues zero SQLite work. The
+        // in-memory transition already happened, so a SQLite failure
+        // only warns — a surviving row rehydrates and self-heals via the
+        // reconcile / prune passes.
+        for activity_id in &outcome.dropped_redrives {
+            if let Err(e) = self.delete_redrive_row(id, activity_id) {
                 tracing::warn!(
                     error = %e,
                     "redrive row deletion failed after mark_spent; a stale row may \
@@ -312,7 +339,7 @@ impl ShieldedStore for FileBackedShieldedStore {
                 );
             }
         }
-        Ok(marked)
+        Ok(outcome.newly_spent)
     }
 
     fn mark_pending(&mut self, id: SubwalletId, nullifier: &[u8; 32]) -> Result<bool, Self::Error> {
@@ -613,11 +640,27 @@ impl ShieldedStore for FileBackedShieldedStore {
         // in-memory only (`subwallets`); the commitment tree in
         // SQLite is chain-wide and intentionally left intact.
         self.subwallets.retain(|id, _| id.wallet_id != wallet_id);
+        // The redrive table IS durable (unlike the rest of subwallet
+        // state), so purging the in-memory map alone would leave this
+        // wallet's rows to rehydrate stale reservations / rebroadcast
+        // state on the next open. Delete them here too, scoped by
+        // wallet_id.
+        let conn = self.pending_conn.lock().expect("pending_conn mutex");
+        conn.execute(
+            "DELETE FROM shielded_pending_spends WHERE wallet_id = ?1",
+            rusqlite::params![wallet_id.as_slice()],
+        )
+        .map_err(|e| FileShieldedStoreError(format!("purge pending spends for wallet: {e}")))?;
         Ok(())
     }
 
     fn purge_all_subwallets(&mut self) -> Result<(), Self::Error> {
         self.subwallets.clear();
+        // Durable redrive rows for every wallet go with the in-memory
+        // purge (see `purge_wallet`).
+        let conn = self.pending_conn.lock().expect("pending_conn mutex");
+        conn.execute("DELETE FROM shielded_pending_spends", [])
+            .map_err(|e| FileShieldedStoreError(format!("purge all pending spends: {e}")))?;
         Ok(())
     }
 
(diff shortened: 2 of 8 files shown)


thepastaclaw:
✅ Review complete (commit 153e51c6)

packages/rs-platform-wallet/src/wallet/shielded/file_store.rs
```diff
@@ -613,11 +640,27 @@ impl ShieldedStore for FileBackedShieldedStore {
...
 
     fn purge_all_subwallets(&mut self) -> Result<(), Self::Error> {
         self.subwallets.clear();
+        // Durable redrive rows for every wallet go with the in-memory
+        // purge (see `purge_wallet`).
+        let conn = self.pending_conn.lock().expect("pending_conn mutex");
+        conn.execute("DELETE FROM shielded_pending_spends", [])
+            .map_err(|e| FileShieldedStoreError(format!("purge all pending spends: {e}")))?;
         Ok(())
     }
```
thepastaclaw, line 661: **🟡 Suggestion: Durable redrive DELETE runs after the in-memory purge, so a SQL failure leaves memory and disk inconsistent**

`purge_wallet` and `purge_all_subwallets` clear `self.subwallets` before running `DELETE FROM shielded_pending_spends`. If the SQL DELETE fails, both functions return `Err`, but the in-memory purge has already committed — the exact memory/disk divergence this PR is fixing on the `mark_spent` path. Two consistent options: (a) run the SQL DELETE first, then purge memory, so an `Err` return leaves both stores unchanged (fail-atomic); or (b) match the `mark_spent` / `clear_pending` error posture in this file — `warn!` the SQL failure and return `Ok(())` because the in-memory purge is already authoritative. Either is preferable to the current shape, where callers see `Err` and cannot tell whether the memory purge landed.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet/src/wallet/platform_addresses/fund_from_asset_lock.rs
```diff
@@ -302,11 +302,40 @@ impl PlatformAddressWallet {
...
+                // lock as Consumed over durable balances that under-report
+                // the credit.
+                tracing::error!(
+                    outpoint = %out_point,
+                    "skipping consume_asset_lock: the reconciled balance changeset \
+                     was not durably stored; the lock stays non-Consumed (Resumable) \
+                     rather than pairing a Consumed lock with stale balance rows on disk"
+                );
+                return Ok(cs);
+            }
```
thepastaclaw, line 338: **🟡 Suggestion: `fund_from_asset_lock` returns `Ok(cs)` identically on the durable-success and skip-consume paths**

When `persisted == false`, the function returns `Ok(cs)` — the same shape as the fully-successful path — after logging at `error!` and skipping `consume_asset_lock`. The rationale (Platform will deterministically reject a Resume with 'lock already consumed') is correct, but the return signal collapses two very different outcomes: 'top-up landed and is durable' vs. 'top-up applied in memory only, lock stays Resumable, balance may resurface stale after restart'. Downstream UI aggregation and telemetry can't distinguish the two. A distinct variant or a boolean on the returned changeset would let callers surface the discrepancy rather than treat it as full success. Non-blocking — current behavior is correct, just opaque.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

PR delivers narrow data-integrity follow-ups (durable redrive lifecycle, persistence-gated asset-lock consume, address-reconciliation index-conflict handling, Swift Clear/Sync race gating). Verification against 7ce9faeb confirms invariants match the extensive rationale comments; no blocking issues. Two suggestion-level polish items on the shielded-store purge ordering and the fund_from_asset_lock non-persisted return signal are worth flagging.

🟡 2 suggestion(s)


Source: reviewers claude-general/security-auditor/rust-quality/ffi-engineer (opus); codex-general/security-auditor/rust-quality/ffi-engineer (gpt-5.5[high], failed/unparseable); verifier claude (opus).

<details>

```

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/file_store.rs:638-665: Durable redrive DELETE runs after the in-memory purge, so a SQL failure leaves memory and disk inconsistent
  `purge_wallet` and `purge_all_subwallets` clear `self.subwallets` before running `DELETE FROM shielded_pending_spends`. If the SQL DELETE fails, both functions return `Err`, but the in-memory purge has already committed — the exact memory/disk divergence this PR is fixing on the `mark_spent` path. Two consistent options: (a) run the SQL DELETE first, then purge memory, so an `Err` return leaves both stores unchanged (fail-atomic); or (b) match the `mark_spent` / `clear_pending` error posture in this file — `warn!` the SQL failure and return `Ok(())` because the in-memory purge is already authoritative. Either is preferable to the current shape, where callers see `Err` and cannot tell whether the memory purge landed.

In `packages/rs-platform-wallet/src/wallet/platform_addresses/fund_from_asset_lock.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/platform_addresses/fund_from_asset_lock.rs:320-338: `fund_from_asset_lock` returns `Ok(cs)` identically on the durable-success and skip-consume paths
  When `persisted == false`, the function returns `Ok(cs)` — the same shape as the fully-successful path — after logging at `error!` and skipping `consume_asset_lock`. The rationale (Platform will deterministically reject a Resume with 'lock already consumed') is correct, but the return signal collapses two very different outcomes: 'top-up landed and is durable' vs. 'top-up applied in memory only, lock stays Resumable, balance may resurface stale after restart'. Downstream UI aggregation and telemetry can't distinguish the two. A distinct variant or a boolean on the returned changeset would let callers surface the discrepancy rather than treat it as full success. Non-blocking — current behavior is correct, just opaque.
```
</details>

packages/rs-platform-wallet/src/wallet/shielded/file_store.rs
thepastaclaw, line 661: **🟡 Suggestion: purge_wallet / purge_all_subwallets mutate memory before the durable DELETE, so a SQL failure returns Err with memory already purged**

`purge_wallet` clears `self.subwallets` at line 642 before executing `DELETE FROM shielded_pending_spends WHERE wallet_id = ?1` at line 649; `purge_all_subwallets` does the same at line 658 before the unscoped DELETE at line 662. If the SQL execute fails, both functions map the error and return `Err(...)` — but the in-memory purge has already committed. Callers cannot distinguish 'nothing happened' from 'memory purged, disk unchanged', which is exactly the memory/disk divergence this PR is otherwise fixing on the mark_spent path. Two consistent fixes: (a) run the SQL DELETE first, then purge memory only on success (fail-atomic); or (b) match the log-don't-abort posture used earlier in this same file for `mark_spent` and `clear_pending` — `warn!` the SQL failure and return `Ok(())` because the in-memory purge is already authoritative and the durable rows will be reconciled on the next open.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Latest push (7ce9faeb → abf952ea) is a pure cargo fmt on two test assertions in platform_addresses/provider.rs — no behavioral change. Two prior suggestion-level findings from the shielded-store purge and fund_from_asset_lock persistence-gated paths still apply at the current head. No new findings from the delta; no blocking issues.

🟡 2 carried-forward suggestion(s) | 0 new latest-delta findings

**Prior finding reconciliation**
- STILL VALID: packages/rs-platform-wallet/src/wallet/shielded/file_store.rs:638-665 — purge_wallet / purge_all_subwallets mutate memory before the durable DELETE, so a SQL failure returns Err with memory already purged
- STILL VALID: packages/rs-platform-wallet/src/wallet/platform_addresses/fund_from_asset_lock.rs:320-338 — fund_from_asset_lock returns Ok(cs) identically on the durable-success and skip-consume paths

**Latest delta**
`7ce9faeb..abf952ea` only changes `packages/rs-platform-wallet/src/wallet/platform_addresses/provider.rs` formatting/test assertion layout. No new latest-delta findings.

Source: reviewers claude-general/security-auditor/rust-quality/ffi-engineer (opus, ok); codex-general/security-auditor/rust-quality/ffi-engineer (gpt-5.5[high], failed/unparseable: ACP advertises gpt-5.5 not gpt-5.5[high]); verifier claude (opus).

<details>

```

Carried-forward prior findings still valid at abf952ea:

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/file_store.rs:638-665: purge_wallet / purge_all_subwallets mutate memory before the durable DELETE, so a SQL failure returns Err with memory already purged
  `purge_wallet` clears `self.subwallets` at line 642 before executing `DELETE FROM shielded_pending_spends WHERE wallet_id = ?1` at line 649; `purge_all_subwallets` does the same at line 658 before the unscoped DELETE at line 662. If the SQL execute fails, both functions map the error and return `Err(...)` — but the in-memory purge has already committed. Callers cannot distinguish 'nothing happened' from 'memory purged, disk unchanged', which is exactly the memory/disk divergence this PR is otherwise fixing on the mark_spent path. Two consistent fixes: (a) run the SQL DELETE first, then purge memory only on success (fail-atomic); or (b) match the log-don't-abort posture used earlier in this same file for `mark_spent` and `clear_pending` — `warn!` the SQL failure and return `Ok(())` because the in-memory purge is already authoritative and the durable rows will be reconciled on the next open.

In `packages/rs-platform-wallet/src/wallet/platform_addresses/fund_from_asset_lock.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/platform_addresses/fund_from_asset_lock.rs:320-338: fund_from_asset_lock returns Ok(cs) identically on the durable-success and skip-consume paths
  When `persisted == false`, the function returns `Ok(cs)` at line 337 after logging at `error!` and skipping `consume_asset_lock` — the same return shape as the fully-successful path further down. The persistence-gate itself is the correct invariant (a Consumed lock over non-durable balances would under-budget after restart), and Platform will deterministically reject a Resume with 'lock already consumed', but the collapsed return signal makes 'top-up landed and is durable' indistinguishable from 'top-up applied in memory only, lock stays Resumable, balance may resurface stale after restart'. Downstream UI aggregation and telemetry cannot detect the degraded outcome. A distinct variant on the returned changeset — or a `persisted: bool` companion mirroring the `_with_persistence` reconciliation seam this already calls — would let callers surface the discrepancy rather than treat it as full success. Non-blocking; behavior is correct, just opaque.
```
</details>

packages/rs-platform-wallet/src/wallet/shielded/file_store.rs
thepastaclaw: Correction: still valid at `abf952ea` — _purge_wallet / purge_all_subwallets mutate memory before the durable DELETE, so a SQL failure returns Err with memory already purged_ remains present. The latest push was formatting-only in `platform_addresses/provider.rs` and did not change this `file_store.rs` ordering.

<sub>Corrected after verifier reconciliation; the previous auto-resolved wording was wrong.</sub>

packages/rs-platform-wallet/src/wallet/platform_addresses/fund_from_asset_lock.rs
thepastaclaw: Correction: still valid at `abf952ea` — _fund_from_asset_lock returns Ok(cs) identically on the durable-success and skip-consume paths_ remains present. The latest push was formatting-only in `platform_addresses/provider.rs` and did not change this `fund_from_asset_lock.rs` return signal.

<sub>Corrected after verifier reconciliation; the previous auto-resolved wording was wrong.</sub>

packages/rs-platform-wallet/src/wallet/shielded/file_store.rs
QuantumExplorer: Reordered in 153e51c603: the SQLite DELETE now runs before the in-memory purge in both `purge_wallet` and `purge_all_subwallets`, so an `Err` return means neither store was touched (fail-atomic).

packages/rs-platform-wallet/src/wallet/platform_addresses/fund_from_asset_lock.rs
QuantumExplorer: Not changed: by this point the ST is accepted and `validate_address_infos_complete` has proven every recipient's credit is attested on-chain, so 'Consumed' is truthful on both paths; the in-memory state is correct and the only divergence is the durable rows, which the next sync pass re-persists. Distinguishing the outcomes on the changeset would push a transient persistence detail into the FFI surface — the `error!` log is the intended observability. Happy to revisit if telemetry needs it.

packages/rs-platform-wallet/src/wallet/shielded/file_store.rs
QuantumExplorer: Fixed in 153e51c603 with option (a): the SQLite DELETE now runs before the in-memory purge in both functions, so an `Err` leaves both stores untouched (fail-atomic).

QuantumExplorer:
## Self-review results (8 finder angles × adversarial verification)

Ran a full multi-angle review of this PR. **Fixed in `153e51c603`** (plus `abf952eace` fmt):
- **purge ordering** — SQLite DELETE now runs before the in-memory purge (fail-atomic); also flagged by thepastaclaw.
- **`performSync()` gate** — the Clear/Sync mutual exclusion now lives in the service (`guard !isClearing`), not just on CoreContentView's buttons; pull-to-refresh (Wallets/Identities) and the post-submit resyncs (Transfer/Withdraw) called `performSync` directly and could start a sync mid-Clear.
- **wallet-switch generation gap** — `reset()`/`configure()` now bump `balanceSnapshotGeneration`; an in-flight snapshot strongly captures the old wallet handle and would otherwise publish the old wallet's balances over the new configuration.
- **stale doc comment** — `reset_sync_state`'s outer doc still claimed the two locks are taken sequentially; rewritten to describe the nested provider → wallet-manager order.
- **SQL dedup** — `clear_redrive` now delegates its durable delete to `delete_redrive_row`.

**Deferred (tracked separately, both pre-existing):**
1. Index-conflicting *removals* are still emitted carrying the conflicting `address_index`; the Swift persister writes it onto the zero-balance row, so two durable rows can claim one index and the restore-time bimap insert may evict the protected pairing. Needs a persister-contract fix (sentinel or skip-index-on-zero-balance) — deeper than this PR.
2. `registration.rs` (`register_identity_with_funding`) still propagates post-acceptance `add_identity` failure with `?` and skips `consume_asset_lock`, the exact hazard class this PR fixes in `register_from_addresses.rs`.

**Verified-refuted** (for the record): the vacuous `persisted=true` early returns are safe (credits are proof-attested on-chain; consuming is truthful), best-effort `add_identity` cannot cause index reuse (Swift derives the next index and the discovery rescan self-heals), and the rehydrated-ghost-redrive scenario self-heals because the shielded watermark is in-memory (every reopen rescans from 0 and re-encounters the nullifier). The index-conflict credit drop self-heals within one sync round via `prepare_for_sync`'s bijection rebuild, so the drop-don't-apply choice stands.

thepastaclaw (commented):
## Code Review

Latest push (`abf952ea` → `153e51c6`) fixes one of the two carried-forward findings and leaves the other still valid. `purge_wallet` / `purge_all_subwallets` now delete durable redrive rows before mutating memory, so the prior fail-atomicity concern is fixed. `fund_from_asset_lock` still returns the same `Ok(cs)` for durable success and the persistence-failure skip-consume path, so that suggestion carries forward. No new latest-delta findings and no blockers.

🟡 1 carried-forward suggestion(s) | 0 new latest-delta findings

**Prior finding reconciliation**
- FIXED: `packages/rs-platform-wallet/src/wallet/shielded/file_store.rs:638-665` — purge ordering now runs SQL before memory in both purge paths.
- STILL VALID: `packages/rs-platform-wallet/src/wallet/platform_addresses/fund_from_asset_lock.rs:320-338` — `fund_from_asset_lock` still returns `Ok(cs)` identically on durable-success and skip-consume paths.

**Latest delta**
`abf952ea..153e51c6` touches `wallet.rs`, `shielded/file_store.rs`, and `PlatformBalanceSyncService.swift`. It fixes the shielded purge ordering, updates the reset-sync lock-order doc, and adds Swift-side snapshot/clear-sync hardening. No new latest-delta findings.

Source: reviewers claude-general (opus, ok); codex-general (gpt-5.5, ok); claude-security-auditor (opus, ok); codex-security-auditor (gpt-5.5, ok); claude-rust-quality (opus, ok); codex-rust-quality (gpt-5.5, ok); claude-ffi-engineer (opus, ok); codex-ffi-engineer (gpt-5.5, ok); verifier claude (opus).

<details>

```

Carried-forward prior findings still valid at 153e51c6:

In `packages/rs-platform-wallet/src/wallet/platform_addresses/fund_from_asset_lock.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/platform_addresses/fund_from_asset_lock.rs:320-338: fund_from_asset_lock returns Ok(cs) identically on the durable-success and skip-consume paths
  Carried forward from prior-2; unchanged at 153e51c6 (git diff of this file across abf952ea..HEAD is empty). When `persisted == false`, the function logs at `error!`, skips `consume_asset_lock`, and returns `Ok(cs)` at line 337 — the same return shape as the fully-successful path further down. The persistence-gate itself is the correct invariant (a Consumed lock over non-durable balances would under-budget after restart, and Platform will deterministically reject a Resume with 'lock already consumed'), but the collapsed return signal makes 'top-up landed and is durable' indistinguishable from 'top-up applied in memory only, lock stays Resumable, balance may resurface stale after restart'. The signal loss propagates through the C FFI (`rs-platform-wallet-ffi/src/platform_addresses/fund_from_asset_lock.rs:116-119` converts to `PlatformWalletFFIResult::ok()`) and through Swift (`ManagedPlatformAddressWallet.swift:614-617` decodes as a successful `[UpdatedBalance]`), so downstream UI aggregation and telemetry cannot detect the degraded outcome. A distinct variant on the returned changeset — or a `persisted: bool` companion mirroring the `_with_persistence` reconciliation seam this already calls — would let callers surface the discrepancy rather than treat it as full success. Consistent with the fail-atomic reordering just applied to `purge_wallet` / `purge_all_subwallets`, this is the last place in the data-integrity series where the caller can't distinguish partial-success from full-success. Non-blocking; behavior is correct, just opaque.
```
</details>

PR #4007, opened 2026-07-05
fix(swift-example-app): resume orphaned Broadcast asset-lock top-ups

QuantumExplorer:
Follow-up to #4006 (merged). That PR made the asset-lock ChainLock wait unbounded so a top-up can't falsely "fail"; this one gives an *interrupted* top-up a way out of the resulting Broadcast state.

## Issue being fixed or feature implemented

Defect 2 from the ADDR-09 top-up QA. A platform-address top-up that broadcast its asset lock but was interrupted before finality — app killed mid-wait, or a prior attempt failed — leaves a `PersistentAssetLock` at `statusRaw 1` (Broadcast). The "Pending Platform Top Ups" row gated its Resume affordance on `canFundIdentity` (`statusRaw ∈ {2,3}`), so a Broadcast row rendered only a **non-interactive spinner** — no way to resume, cancel, or retry. Result: a permanent "Waiting for InstantSend / ChainLock…" dead end with committed on-chain funds stuck behind it.

## What was done?

`resume_asset_lock` already supports resuming a Broadcast lock (it re-enters the finality wait — now unbounded, per #4006), so the fix is UI-only.

- **Offer Resume on a Broadcast (`statusRaw 1`) orphan row** in `PendingPlatformFundFromAssetLocksList`. Tapping it routes through the existing `resumeFundFromAssetLock` → `platform-wallet` path.
- **Guard against a duplicate-consume race.** A fresh-build fund sits at Broadcast for the entire (unbounded) finality wait, so its lock co-renders as an orphan row. The controller doesn't expose the outpoint it's driving, so we can't distinguish that in-flight lock from a genuinely-orphaned one; resuming the in-flight one would spawn a second consuming flow on the same outpoint. So the Broadcast Resume is gated on **"no funding in flight on this wallet"** — a safe superset, since active controllers are transient (a truly orphaned lock always becomes resumable once they clear). Proof-ready (`statusRaw 2/3`) rows keep their existing unconditional Resume (their co-render window is brief).
- **Status-aware resume footer** in `FundFromAssetLockPlatformAddressView`: it no longer claims a Broadcast lock "reached a usable proof state" — it says resuming will wait for finality, then credit the address.

No Rust and no FFI changes; this is presentation logic only (the resume itself routes through `platform-wallet`), consistent with the Swift-SDK architectural rules.

## How Has This Been Tested?

[...]

PR #4009, opened 2026-07-05
fix(platform-wallet): re-broadcast a Broadcast-status lock on resume

QuantumExplorer:
Follow-up to the ADDR-09 finality work (#4006 / #4007). Surfaced by the review of #4007: resuming an orphaned `Broadcast`-status lock could hang forever.

`resume_asset_lock`'s `Broadcast` arm only waited for a proof — it never re-broadcast the funding tx (only the `Built` arm does). If the tx had been evicted from every mempool (Core's default `-mempoolexpiry` is two weeks) or the original broadcast reached no peers (an SPV connectivity gap — a real condition on the testnet QA setup), no InstantSend/ChainLock proof can ever arrive. And since the user-facing funding flows now wait for the ChainLock **indefinitely** (#4006), the resume hangs forever: the controller stays `.inFlight`, Cancel is disabled, the network toggle is disabled app-wide, and no in-app recovery exists short of relaunch. #4007's new "Resume on a Broadcast orphan" button makes this reachable from the UI.

## What was done?

Add a defensive re-broadcast to the `Broadcast` arm of `resume_asset_lock` before the wait.

Unlike the `Built` arm — whose tx was never broadcast, so a broadcast error is genuine and propagates — a `Broadcast`-status tx may still be in a mempool or already mined, where the network reports "already known" / "already in block chain". The broadcaster can't distinguish that from a real rejection (`DapiBroadcaster` classifies every failure as `MaybeSent`), so the re-broadcast is **best-effort**: log the error and proceed to `wait_for_proof` regardless, rather than failing a resume on a tx that is actually fine. If it was mined, the wait resolves immediately from the persisted record.

This path is shared by the identity, platform-address, and shielded resume flows plus the app-launch catch-up sweep; re-broadcasting an already-known tx is idempotent, so the extra call is harmless where the tx is still live.

## How Has This Been Tested?

- `cargo check -p platform-wallet --features shielded --all-targets` — clean.
- `cargo test -p platform-wallet --features shielded asset_lock` — 19 asset-lock unit tests pass.
- `cargo fmt --all` applied.

Not yet exercised end-to-end (would require an evicted/undelivered asset-lock tx on testnet).

## Breaking Changes

None. No consensus or ABI change.

## Checklist:

### DIFF
--- a/packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs
+++ b/packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs
@@ -265,7 +265,36 @@ impl<B: TransactionBroadcaster + ?Sized> AssetLockManager<B> {
                     .await?
             }
             AssetLockStatus::Broadcast => {
-                // Already broadcast — just wait for proof.
+                // Defensive re-broadcast, then wait for proof. A lock can
+                // sit at `Broadcast` across app restarts long enough for
+                // its funding tx to be evicted from every mempool (Core's
+                // default `-mempoolexpiry` is two weeks), or the original
+                // broadcast may have reached no peers at all (SPV
+                // connectivity gap). Once no node holds the tx, no IS/CL
+                // proof can ever arrive, and the wait — now unbounded for
+                // the user-facing funding flows — would hang forever. A
+                // re-broadcast revives an evicted/undelivered tx.
+                //
+                // Best-effort: unlike the `Built` arm, this tx was already
+                // broadcast once (that's what `Broadcast` means), so it may
+                // still be in a mempool or already mined — in which case the
+                // network reports "already known" / "already in block
+                // chain". The broadcaster can't distinguish that from a real
+                // rejection (DAPI classifies every failure as `MaybeSent`),
+                // so we log and proceed to `wait_for_proof` regardless
+                // rather than failing the resume on a tx that is actually
+                // fine. If the tx really was mined, `wait_for_proof`
+                // resolves immediately from the SPV/persisted record.
+                if let Err(e) = self.broadcaster.broadcast(&tx).await {
+                    tracing::debug!(
+                        outpoint = %out_point,
+                        error = %e,
+                        "resume_asset_lock: defensive re-broadcast of a \
+                         Broadcast-status lock returned an error (likely \
+                         already in a mempool or mined); proceeding to wait \
+                         for proof"
+                    );
+                }
                 let proof = self.wait_for_proof(out_point, timeout).await?;
                 self.validate_or_upgrade_proof(proof, account_index, out_point)
                     .await?


thepastaclaw:
✅ Review complete (commit 35a8885c)

packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs
```diff
@@ -265,7 +265,36 @@ impl<B: TransactionBroadcaster + ?Sized> AssetLockManager<B> {
...
+                if let Err(e) = self.broadcaster.broadcast(&tx).await {
+                    tracing::debug!(
+                        outpoint = %out_point,
+                        error = %e,
+                        "resume_asset_lock: defensive re-broadcast of a \
+                         Broadcast-status lock returned an error (likely \
+                         already in a mempool or mined); proceeding to wait \
+                         for proof"
+                    );
+                }
```
thepastaclaw, line 297: **💬 Nitpick: `tracing::debug!` may hide genuine broadcaster failures on resume**

Swallowing the broadcaster error is correct (DAPI can't distinguish 'already known' / 'already in block chain' from a real rejection because everything uncertain is `MaybeSent`). But at `debug!` level, the pathological case this PR is trying to defend against — the broadcaster reaching zero peers — leaves no trace at default log levels, after which the unbounded `wait_for_proof` from #4006 blocks silently. Promoting to `warn!` (or `info!` with a clear 'best-effort re-broadcast reported error' phrasing) would keep the healthy case quiet (the line only fires on `Err`) while making the failure diagnosable from a normal log capture. Volume isn't a concern since this only runs on resume, not steady-state.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Small, well-scoped defensive fix that adds a best-effort re-broadcast to `resume_asset_lock`'s `Broadcast` arm before waiting for a proof. Implementation is correct — the swallowed broadcaster error is documented and matches DAPI's `MaybeSent` semantics. All four reviewers converged on the same gap: the new arm has no unit test, even though the crate's generic `TransactionBroadcaster` and existing test doubles (`AlwaysMaybeSentBroadcaster`, `RejectFirstBroadcaster`) make this cheap to pin down.

Source: reviewers=claude-general(opus), codex-general(gpt-5.5), claude-security-auditor(opus), codex-security-auditor(gpt-5.5), claude-rust-quality(opus), codex-rust-quality(gpt-5.5); verifier=claude(opus)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs:267-301: New `Broadcast`-arm re-broadcast has no unit test
  The functional fix in this PR is the side effect added to `resume_asset_lock`'s `AssetLockStatus::Broadcast` arm: it must (a) invoke `self.broadcaster.broadcast(&tx).await` before entering `wait_for_proof`, and (b) continue to `wait_for_proof` even if that re-broadcast returns `Err`. Neither invariant is exercised by the existing asset-lock tests, which cover the initial-broadcast rejection/ambiguity paths but not resume from an already-`Broadcast` row.

  The crate is set up to test this cheaply — `AssetLockManager` is generic over `B: TransactionBroadcaster + ?Sized`, and the existing test doubles (`AlwaysMaybeSentBroadcaster` in `test_support.rs`, plus the `RejectFirstBroadcaster`-style pattern used in `core/broadcast.rs`) can assert both invariants with a counter and an `Err`-returning broadcaster. Without such a test, a future refactor could silently demote the `if let Err` back to `?` (reintroducing the hang) or drop the re-broadcast call entirely, and CI would still pass. Given the failure mode this PR is defending against is 'resume hangs forever with no in-app recovery', this contract is worth pinning down.
```
</details>

PR #4011, opened 2026-07-05
fix(platform-wallet): make post-acceptance identity bookkeeping best-effort

QuantumExplorer:

PR #4008 made post-acceptance `add_identity` best-effort (warn, not `?`) in the address-funded `register_from_addresses` flow. The sibling asset-lock flows in `identity/network/registration.rs` still propagated post-acceptance failures with `?`:

- **`register_identity_with_funding`** — Step 4's `get_wallet_info_mut(...).ok_or_else(...)?` and `info.identity_manager.add_identity(...)?` both ran *after* Platform accepted the `IdentityCreate`.
- **`top_up_identity_with_funding`** — Step 4's `get_wallet_info_mut(...).ok_or_else(...)?` ran *after* Platform accepted the top-up.

In both, an early `Err` there had two bad consequences:

1. The app reported **failure for an identity/credits that already exist on chain**.
2. The early return **skipped Step 5's `consume_asset_lock`**, leaving the spent lock in the Resumable Funding list. A user Resume then hits Platform's deterministic *"lock already consumed"* rejection — a dead-end for a lock that did its job.

## What was done?

Converted both post-acceptance steps to the same warn-and-continue posture as `register_from_addresses` Step 3 (#4008):

- `get_wallet_info_mut` returning `None` → `tracing::warn!` and skip local persistence instead of returning `WalletNotFound`.
- `add_identity` returning `Err` (register flow) → `tracing::warn!` and skip the key-breadcrumb stamping (an `IdentityAlreadyExists` may point at an out-of-wallet entry; stamping `wallet_id` without moving buckets would contradict the location index).
- The flow now always falls through to `consume_asset_lock`, so the spent lock is cleaned up. Stale local state self-heals on the next identity re-sync.

Method doc-comment step lists updated to note the best-effort posture. Comment style mirrors `register_from_addresses.rs` Step 3.

## How Has This Been Tested?

- `cargo fmt --all`
- `cargo test -p platform-wallet` — green (8 passed, 0 failed).

The unit-test module in `registration.rs` covers only the IS→CL timeout discriminator; exercising the post-acceptance path would require a live SDK submission mock the module doesn't set up, so — as with #4008's treatment of `register_from_addresses` — no unit test was added for this branch.

## Breaking Changes

None. Local-only behavior change (error → warn); no consensus, serialization, or API surface change.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Identity registration and balance top-ups now complete successfully even if local bookkeeping or wallet-state saving fails after platform acceptance.
  * These non-critical failures are now handled as warnings, helping avoid unnecessary error returns.
  * Improved handling when wallet information is unavailable during post-acceptance updates.

### DIFF
--- a/packages/rs-platform-wallet/src/wallet/identity/network/registration.rs
+++ b/packages/rs-platform-wallet/src/wallet/identity/network/registration.rs
@@ -96,6 +96,8 @@ impl IdentityWallet {
     ///    (`InvalidInstantAssetLockProofSignatureError`).
     /// 4. On success, add the confirmed identity to the local
     ///    `IdentityManager` and record each key's derivation breadcrumb.
+    ///    Best-effort: Platform has already accepted, so a local
+    ///    bookkeeping failure is logged, not propagated.
     /// 5. Remove the tracked asset lock (if any) — the credit output
     ///    has been consumed, so the entry is no longer needed.
     ///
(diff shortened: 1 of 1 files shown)


thepastaclaw:
✅ Review complete (commit b3e43959)

packages/rs-platform-wallet/src/wallet/identity/network/registration.rs
```diff
@@ -262,40 +264,71 @@ impl IdentityWallet {
...
+                        // `wallet_id` on one without moving buckets would
+                        // contradict the manager's location index.
+                        tracing::warn!(
+                            error = %e,
+                            identity_id = %identity.id(),
+                            "register_identity_with_funding: identity registered on \
+                             Platform but local add_identity failed; continuing so \
+                             the spent asset lock is still consumed"
+                        );
+                    }
```
thepastaclaw, line 324: **🟡 Suggestion: Step 4 Err arm swallows every add_identity failure, not just IdentityAlreadyExists**

The catch-all `Err(e)` arm at line 312 warns and continues for every error from `add_identity`, but the inline justification comment only reasons about `IdentityAlreadyExists`. Two consequences worth acknowledging in the comment or the match shape:

1. `add_identity` returns `IdentityAlreadyExists` whether the pre-existing entry is in the wallet-owned bucket *or* the out-of-wallet bucket (verified: `lifecycle.rs:45` guards on `self.identity(...).is_some()`, and `accessors.rs:21-32` scans both buckets). For the out-of-wallet duplicate case, this PR now returns `Ok(identity)` while `location_index` still reports `OutOfWallet`, so a follow-up `top_up_identity_with_funding` on the same id will fail with `IdentityIndexNotSet` (`identity_index` returns `None` for the OOW discriminant at `lifecycle.rs:144-149`). The comment claims this self-heals on re-sync, but the current refresh/discovery paths mutate the existing managed identity rather than moving buckets, so it does not.
2. For non-`IdentityAlreadyExists` variants (e.g. an internal persister propagation error surfaced by a future signature change), silently swallowing is not obviously the desired posture — the location-index reasoning does not apply.

Either discriminate on the error variant (promote/replace on `IdentityAlreadyExists` with an OutOfWallet location; propagate — or at least error-level log — the rest) or expand the comment so future maintainers know the catch-all is deliberate for reasons beyond the location-index invariant. Not blocking because the register-funded fresh-identity path where a same-id OOW entry already exists is a narrow scenario, and this PR is a strict improvement over the previous behavior that stranded the asset lock.

thepastaclaw (commented):
## Code Review

Well-scoped follow-up to #4008 that converts two post-acceptance `?` sites in the asset-lock-funded register/top-up flows into warn-and-continue, so `consume_asset_lock` runs and the spent lock is not stranded. No consensus, serialization, proof, or FFI surface is touched. Verified against b3e43959: `IdentityManager::add_identity` does raise `IdentityAlreadyExists` for both buckets, but the OOW-collision path is speculative for freshly-registered identities and the PR is a strict improvement over the prior stranded-lock behavior. Two observability suggestions worth surfacing; no in-scope blockers.

Source: reviewers claude/opus (general), codex/gpt-5.5 (general), claude/opus (rust-quality), codex/gpt-5.5 (rust-quality); verifier claude/opus; specialists: rust-quality.

<details>

```

In `packages/rs-platform-wallet/src/wallet/identity/network/registration.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/registration.rs:312-324: Step 4 Err arm swallows every add_identity failure, not just IdentityAlreadyExists
  The catch-all `Err(e)` arm at line 312 warns and continues for every error from `add_identity`, but the inline justification comment only reasons about `IdentityAlreadyExists`. Two consequences worth acknowledging in the comment or the match shape:

  1. `add_identity` returns `IdentityAlreadyExists` whether the pre-existing entry is in the wallet-owned bucket *or* the out-of-wallet bucket (verified: `lifecycle.rs:45` guards on `self.identity(...).is_some()`, and `accessors.rs:21-32` scans both buckets). For the out-of-wallet duplicate case, this PR now returns `Ok(identity)` while `location_index` still reports `OutOfWallet`, so a follow-up `top_up_identity_with_funding` on the same id will fail with `IdentityIndexNotSet` (`identity_index` returns `None` for the OOW discriminant at `lifecycle.rs:144-149`). The comment claims this self-heals on re-sync, but the current refresh/discovery paths mutate the existing managed identity rather than moving buckets, so it does not.
  2. For non-`IdentityAlreadyExists` variants (e.g. an internal persister propagation error surfaced by a future signature change), silently swallowing is not obviously the desired posture — the location-index reasoning does not apply.

  Either discriminate on the error variant (promote/replace on `IdentityAlreadyExists` with an OutOfWallet location; propagate — or at least error-level log — the rest) or expand the comment so future maintainers know the catch-all is deliberate for reasons beyond the location-index invariant. Not blocking because the register-funded fresh-identity path where a same-id OOW entry already exists is a narrow scenario, and this PR is a strict improvement over the previous behavior that stranded the asset lock.

- [NITPICK] packages/rs-platform-wallet/src/wallet/identity/network/registration.rs:505-527: Top-up Step 4 silently no-ops when managed identity is missing under the write lock
  When `wm.get_wallet_info_mut(&self.wallet_id)` returns `Some(info)` but `info.identity_manager.managed_identity_mut(identity_id)` returns `None` (line 507), the `if let Some(managed)` block exits without logging. That's the case where the identity was present at Step 1 under the read lock but disappeared before Step 4 acquired the write lock — Platform accepted the top-up and the new balance is dropped with no operator signal. The outer `None` arm at line 518 warns; the inner `None` case should too, matching the register-flow style ("top-up accepted on Platform but managed identity vanished locally; skipping balance persistence"). Small, but the PR's stated posture is that every post-acceptance skip is logged.

```
</details>

PR #4013, opened 2026-07-05
fix(platform-wallet): index-conflicting removal no longer orphans a restored address balance

QuantumExplorer:

A reconcile **removal** (a fully consumed input) can resolve to a pool
`address_index` that collides with a *different, funded* address's true
derivation index. `commit_reconciliation` correctly leaves the in-memory
`(index → address)` bijection untouched in that case, but it still emits
the zero-funds removal (so a stale balance can't resurrect after a
restart). That emitted entry carried the conflicting `address_index` all
the way to durable storage:

- The FFI hands the entry to Swift's
  `PlatformWalletPersistenceHandler.persistAddressBalances`, which
  **unconditionally overwrote** the row's `addressIndex` with the
  conflicting value — even though that field is authoritative derivation
  metadata owned by the address-emit path, and this callback is
  documented to refresh only the volatile balance/nonce/`isUsed` fields.
- Result: two durable `PersistentPlatformAddress` rows end up claiming
  the same `(accountIndex, addressIndex)`.
- On the next app start, the restore rebuild
  (`PerAccountPlatformAddressState::insert_persisted_entry`) does a
  `BiBTreeMap` insert per row. A colliding insert evicts the prior
  pairing, so — depending on SwiftData fetch order — the **funded**
  address's `(index → address)` pairing can be evicted, orphaning its
  balance from `current_balances` (the address silently loses its
  restored credits).

This was surfaced by the #3990 sync review and is closed here at the
persister seam.

## What was done?

The persister contract has two sides; both are hardened.

**Write side — root cause (`swift-sdk`).**
`persistAddressBalances` no longer writes `accountIndex` / `addressIndex`.
A row's derivation index is fixed the moment its address is derived and
is owned by the address-emit path (`persistPlatformPaymentAddresses`);
the balance-update callback now touches balance / nonce / `isUsed` only —
exactly what its own doc comment already promised. A conflicting removal
therefore zeroes the removed address's balance without ever rebinding its
index, so the durable store stays a bijection.

**Read side — defense in depth (`platform-wallet`).**
`PerAccountPlatformAddressState::insert_persisted_entry` now guards the
bijection. A zero-balance/zero-nonce row is *indistinguishable* from a
legitimate freshly-derived, never-funded address, so it still seeds
`found` exactly as before; but it extends the bijection via
`insert_no_overwrite`, so it can never evict an incumbent `(index →
address)` pairing, while a funded row keeps overwrite semantics and wins
its slot. For the common case of unique indices this is byte-identical to
the previous unconditional insert — it only changes behaviour on an index
collision, keeping the funded pairing regardless of restore order. This
protects any already-persisted store and — because both the SwiftData/FFI
loader and the SQLite loader (`platform_addrs::build_per_account`) route
through this one method — covers both backends.

No FFI struct / ABI change; the entry still carries the indices, the
persister just stops trusting the removal's copy of them.

## How Has This Been Tested?

- **Rust** (`cargo test -p platform-wallet`, `-p platform-wallet-storage`):
  - `insert_persisted_entry_removal_remnant_never_orphans_funded_pairing`
    round-trips the conflict through restore in **both** fetch orders and
    asserts that the intersection of `found` and the bijection (i.e. what
    `current_balances` yields) is exactly the funded balance — the
    remnant's inert zero row is never paired.
  - `insert_persisted_entry_unfunded_derived_address_restores_normally`
    pins that a real never-funded (`{0, 0}`) address still restores into
    both `found` and the bijection, so the guard can't swallow it.
  - All existing `commit_reconciliation` / `platform_addresses` tests and
    the storage crate's SQLite `load_state` reconstruction tests stay
    green.
- **Swift** (`swift test`, `SwiftDashSDKTests`): new
  `AddressBalancePersistTests` round-trips through persist + reload —
  seeds a funded row at index 5 and a row at index 2, applies a removal
  for the index-2 address carrying the conflicting index 5, and asserts
  the removed row keeps index 2 (zeroed) while the funded row keeps index
  5, so no two durable rows share an index.
- `cargo fmt --all`; `cargo check -p platform-wallet --all-targets`.

## Breaking Changes

None. No consensus-affecting changes; no FFI ABI change; pre-release, so
no SwiftData migration concern.

## Checklist:

### DIFF
--- a/packages/rs-platform-wallet/src/wallet/platform_addresses/provider.rs
+++ b/packages/rs-platform-wallet/src/wallet/platform_addresses/provider.rs
@@ -109,14 +109,45 @@ impl PerAccountPlatformAddressState {
     }
 
     /// Seed one persisted address/funds entry into the account state.
+    ///
+    /// Guards the `index <-> address` bijection against an
+    /// index-conflicting *removal remnant*.
+    /// [`commit_reconciliation`](PlatformPaymentAddressProvider::commit_reconciliation)
+    /// can zero an address whose pool-resolved `address_index` equals a
+    /// *funded* address's true index; it leaves the in-memory bijection
+    /// untouched but still emits the zero, so the durable store can hold
+    /// two rows claiming one index. On restore those rows seed the account
+    /// one at a time, and a plain [`BiBTreeMap::insert`] drops conflicting
+    /// pairs — so, depending on fetch order, the zeroed row could evict the
+    /// funded `(index -> address)` pairing and orphan its balance from
+    /// [`current_balances`](AddressProvider::current_balances).
+    ///
+    /// A zero-balance/zero-nonce row can't be told apart from a legitimate
+    /// freshly-derived, never-funded address by its funds alone (both are
+    /// `{0, 0}`), and the latter must still restore into `found` and the
+    /// bijection. So `found` is seeded for *every* row exactly as before;
+    /// the guard is only on the bijection, and only bites on a collision:
+    /// a zeroed row inserts via `insert_no_overwrite` (it can never evict
+    /// an incumbent pairing), while a funded row keeps overwrite semantics
+    /// (it wins its slot, displacing a stale zero remnant that grabbed it
+    /// first — that remnant's dangling `found` entry is inert, since
+    /// `current_balances` only yields addresses still paired in the
+    /// bijection). For the common case of unique indices this is identical
+    /// to the previous unconditional insert.
     pub fn insert_persisted_entry(
         &mut self,
         address_index: AddressIndex,
         address: PlatformP2PKHAddress,
         funds: AddressFunds,
     ) {
-        self.addresses.insert(address_index, address);
         self.found.insert(address, funds);
+        if funds.balance == 0 && funds.nonce == 0 {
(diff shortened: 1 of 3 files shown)


thepastaclaw:
✅ Review complete (commit 542490da)

thepastaclaw (approved):
## Code Review

Prior reviewed SHA a125539e had zero findings, so no findings were carried forward. Both agents (Claude and Codex, all sub-reviewers) reported zero findings on both the latest delta and cumulative head. Verified against worktree at 542490da: the fix hardens the platform-wallet restore path via a bijection guard in `insert_persisted_entry` (uses `insert_no_overwrite` only for zero-balance/zero-nonce rows) and drops non-authoritative derivation index writes on the Swift `persistAddressBalances` callback. No in-scope defects identified.

Source: reviewers claude/opus and codex/gpt-5.5 for general + security-auditor + ffi-engineer; verifier claude/opus.

Prior reconciliation: the previous automated review at `a125539e` found no issues, so there are no carried-forward prior findings to keep or resolve.

Latest delta: reviewed with `git diff a125539e..542490da` plus cumulative current-head context.

PR #4017, opened 2026-07-06
fix(kotlin-sdk): accept Base58 identity ids on the Load Identity screen

QuantumExplorer:

The KotlinExampleApp **Load Identity** screen advertises `Identity ID (hex or Base58)` but rejected Base58 input. Pasting a Base58 identity id (e.g. `5DbLwAxGBzUzo81VewMUwn4b5P4bpv9FNFybi25XB5Bk`) and tapping **Load Identity** surfaced the misleading error *"Could not decode identity id — paste the 64-char hex form."* The hex form of the same id loaded fine.

Root cause: any non-hex input fell back to `decodeIdFromJson(json)`, which read the fetched identity's `"id"` field and only accepted it if it was a 64-char hex string. But `identities.fetch(id)` returns the identity's `"id"` in Base58 (~44 chars), so the `length == 64` check always failed → `null` → the error — even though `fetch` itself succeeded.

## What was done?

Decode the **input string** directly via the app's existing `Base58.decodeIdentifier` helper, which accepts either 64-char hex or Base58 and enforces the 32-byte identifier width. This replaces the fragile round-trip through the fetched JSON.

- Removed the now-dead `decodeIdFromJson` and `isHexDigit` helpers and the unused `hexToBytes` import.
- Reworded the failure message to *"Could not decode identity id — enter a 64-char hex or Base58 identity id."*
- The `fetch(id)` call and balance parse are unchanged (`fetch` already accepts Base58).

UI/example-app only — no SDK/FFI changes.

## How Has This Been Tested?

- Reproduced `Base58.decode`'s algorithm and confirmed `5DbLwAxGBzUzo81VewMUwn4b5P4bpv9FNFybi25XB5Bk` decodes to exactly **32 bytes** = `3ea88cc9622ec22d94db9e2e7a823be13c263d0328146ba80afec33b34ba81af`, identical to the hex path. Both encodings therefore persist the same `identityId`, and `decodeIdentifier`'s 32-byte guard still fails a mistyped id fast.
- Verified the change is self-consistent: no dangling references to the removed helpers, and the retained `kotlinx.serialization` imports are still used by the balance parse.

_Not yet run on an emulator against Testnet — the Android build environment (native `.so` / APFS build volume / emulator) was not provisioned in this session._

## Breaking Changes

None.

## Checklist:

### DIFF
--- a/packages/kotlin-sdk/KotlinExampleApp/app/src/main/java/org/dashfoundation/example/ui/identity/LoadIdentityScreen.kt
+++ b/packages/kotlin-sdk/KotlinExampleApp/app/src/main/java/org/dashfoundation/example/ui/identity/LoadIdentityScreen.kt
@@ -45,7 +45,7 @@ import org.dashfoundation.example.navigation.QrScanner
 import org.dashfoundation.example.ui.components.ErrorAlertDialog
 import org.dashfoundation.example.ui.components.FormSection
 import org.dashfoundation.example.ui.components.SubmitButton
-import org.dashfoundation.example.util.hexToBytes
+import org.dashfoundation.example.util.Base58
 
 /**
  * Load an existing identity by id — port of `LoadIdentityView.swift`. Paste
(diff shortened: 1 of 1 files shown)


thepastaclaw:
✅ Review complete (commit c2d6b041)

thepastaclaw (approved):
## Code Review

The PR is scoped to the Kotlin example app and correctly replaces the failed JSON fallback with direct identifier normalization through the existing Base58 helper. The only validated in-scope gap is missing regression coverage for the advertised Base58 load path; no FFI, consensus, dependency-direction, or CodeRabbit issues apply.

Source: reviewers claude/opus and codex/gpt-5.5 for general + ffi-engineer; verifier codex/gpt-5.5. Note: the original posted verifier was claude/opus; a later full-coverage Claude verifier rerun hit ACP auth 401.

Full-coverage correction: the initial posted approval had two Codex lanes fail because `gpt-5.5[high]` is not advertised by ACP Codex here. I reran those lanes with `gpt-5.5`; the only added item is the non-blocking regression-test suggestion below.

🟡 1 suggestion(s)

<details>

```

In `packages/kotlin-sdk/KotlinExampleApp/app/src/main/java/org/dashfoundation/example/ui/identity/LoadIdentityScreen.kt`:
- [SUGGESTION] packages/kotlin-sdk/KotlinExampleApp/app/src/main/java/org/dashfoundation/example/ui/identity/LoadIdentityScreen.kt:154-156: Add regression coverage for Base58 identity IDs
  This line is now the only normalization step before persisting the loaded identity's primary key, and the PR specifically fixes Base58 input being rejected after the network fetch succeeds. I verified there is no KotlinExampleApp unit or instrumentation test covering `Base58.decodeIdentifier`, and the existing SDK encoding tests cover a different helper. Add a small pure unit test proving a 64-character hex identifier and its Base58 form decode to the same 32-byte value so this user-visible path does not silently regress again.
```
</details>

PR #4018, opened 2026-07-06
fix(platform-wallet): dashpay M1 review findings — rotation, ignore durability, size cap, key selection

QuantumExplorer:

Stacked on #3841 (base: `feat/dashpay-m1-sync-correctness`). Fixes the six medium-severity-and-above findings from the multi-agent review of the DashPay M1 platform-wallet changes ([review summary](https://github.com/dashpay/platform/pull/3841#issuecomment-4888724468) covers the earlier FFI batch; this PR addresses the wallet-core findings).

## What was done?

1. **Pending-branch rotation supersede** — `add_sent_contact_request`'s pending guard no-oped on `contains_key` regardless of `accountReference`, freezing the tracked reference at the first send; the next rotation re-derived an already-broadcast reference and collided forever on the contract's `($ownerId, toUserId, accountReference)` unique index. The pending branch now supersedes on a reference change, mirroring the established branch (persist-before-commit ordering included).

2. **Rotation/registration TOCTOU** — the deferred-crypto drain registers the external account and stamps the rotation self-heal marker under separate guards; a rotation sweep landing in between made the stamp mark an account built from the rotated-away xpub as current, permanently silencing `external_account_needs_rebuild` (sends would derive addresses the contact no longer watches). `note_external_account_registered` now takes the ciphertext the account was built from and skips the stamp on mismatch; the sweep's teardown+rebuild then picks up the fresh request. Applies to both the drain and the accept path.

3. **`ignore_sender` tombstone over-reach** — it emitted `removed_incoming` unconditionally, and the SQLite writer's DELETE had no state filter, so ignoring a sender with no pending incoming entry (an established contact, or one that raced auto-establish) destroyed the established pair-row including the user's alias/note/hidden/accepted-accounts. Emission is now guarded on an actual removal (same `removed.is_some()` discipline as the sibling removers), and both pending-tombstone DELETEs are state-filtered as defense in depth.

4. **Un-ignore durability (SQLite backend)** — restore read the identity `entry_blob`'s `ignored_senders` snapshot, which goes stale on un-ignore (only the table row is deleted; no fresh identity flush) and is UNION-merged across buffered snapshots — a restart resurrected the ignore, stickily. The loader now restores from the transactionally-maintained `ignored_senders` table via a new `load_ignored_senders` reader; the table is authoritative in both directions. (Latent until `Wallet::from_persisted` rehydration is wired; the FFI/Swift backend was already correct.)

5. **contactInfo `privateData` size cap** — nothing capped alias/note, so an over-2048-byte encrypted payload failed broadcast against the contract's `maxItems` permanently AFTER step 1 already durably persisted the local metadata (no retry queue) — permanent local/chain divergence. New `encode_private_data_bounded` enforces the schema cap (max plaintext 2031 bytes = 2048 − IV − min PKCS7 pad) and the publish path runs it as step 0, before any local persist, reusing the encoded bytes at the encrypt step. Rejects rather than truncates (alias/note are user-visible verbatim).

6. **Own-ECDH-key selection unified** — the send path selected the sender's first ENCRYPTION key without filtering disabled keys, while the contactInfo path filtered them: after a disable-and-replace key rotation, every new outgoing contact request hard-failed pre-send validation ("Sender key N is disabled") despite an enabled replacement, and the two surfaces disagreed about the ECDH root. Both paths now share `select_own_encryption_key` (enabled-only, mirroring `select_recipient_key_index`).

## How Has This Been Tested?

TDD: every fix carries a test that was red against the unfixed code:

- `add_sent_contact_request_rotation_supersedes_pending_reference` (state)
- `stamp_skips_when_registration_raced_a_rotation` (network, mock-SDK harness)
- `ignore_sender_without_pending_incoming_emits_no_tombstone` (state)
- `pending_tombstones_do_not_destroy_established_rows` (storage)
- `tc_p4_020_unignore_survives_restart_table_is_authoritative` (storage, both directions)
- `private_data_over_cap_is_rejected_at_cap_is_accepted` (codec, incl. exact-boundary)
- `own_key_selector_skips_disabled_first_encryption_key` + `own_key_selector_errors_when_only_encryption_key_is_disabled`

Suites: `platform-wallet` 388 lib + 9 integration green; `platform-wallet-storage` full suite green; `cargo clippy --all-targets` clean on both; `cargo check -p platform-wallet-ffi` green.

## Breaking Changes

None. `note_external_account_registered` gained a parameter but is `pub(crate)`; `encode_private_data` is unchanged (the bounded variant is additive); the SQLite schema is unchanged (reader added, DELETEs narrowed).

## Checklist:

### DIFF
--- a/packages/rs-platform-wallet-storage/src/sqlite/schema/contacts.rs
+++ b/packages/rs-platform-wallet-storage/src/sqlite/schema/contacts.rs
@@ -16,7 +16,10 @@ use platform_wallet::wallet::platform_wallet::WalletId;
 use crate::sqlite::error::WalletStorageError;
 use crate::sqlite::schema::blob;
 
-#[cfg(feature = "__test-helpers")]
+// `any(test, …)`, not `__test-helpers`-only: `load_ignored_senders` (and its
+// `decode_pair_key` helper) are gated the same way — they're reached by
+// `identities::load_state`, whose gate includes plain `test`.
+#[cfg(any(test, feature = "__test-helpers"))]
 use dpp::prelude::Identifier;
 #[cfg(feature = "__test-helpers")]
 use platform_wallet::changeset::{
@@ -24,9 +27,9 @@ use platform_wallet::changeset::{
 };
 #[cfg(feature = "__test-helpers")]
 use platform_wallet::wallet::identity::{ContactRequest, EstablishedContact};
-#[cfg(feature = "__test-helpers")]
+#[cfg(any(test, feature = "__test-helpers"))]
 use rusqlite::Connection;
-#[cfg(feature = "__test-helpers")]
+#[cfg(any(test, feature = "__test-helpers"))]
 use std::collections::BTreeMap;
 
 /// Single source of truth for the `contacts.state` TEXT-column domain.
@@ -393,7 +396,9 @@ fn decode_request(
     }
 }
 
-#[cfg(feature = "__test-helpers")]
+// Widened to `any(test, …)` alongside `load_ignored_senders`, its only
+// `test`-arm caller (the `__test-helpers`-gated readers use it too).
+#[cfg(any(test, feature = "__test-helpers"))]
 fn decode_pair_key(a: &[u8], b: &[u8]) -> Result<(Identifier, Identifier), WalletStorageError> {
     let a32 = <[u8; 32]>::try_from(a)
         .map_err(|_| WalletStorageError::blob_decode("contacts.id column is not 32 bytes"))?;
--- a/packages/rs-platform-wallet-storage/src/sqlite/schema/contacts.rs
+++ b/packages/rs-platform-wallet-storage/src/sqlite/schema/contacts.rs
@@ -140,14 +140,24 @@ pub fn apply(
         }
     }
     if !cs.removed_sent.is_empty() {
+        // Pending-sent tombstone. State-filtered: the contacts table is one
+        // row per pair, so an unfiltered DELETE would also destroy an
+        // `established` row — both request blobs plus the user's
+        // alias/note/hidden/accepted-accounts — if an emitter ever produced
+        // a pending tombstone for an established pair. The changeset
+        // contract says `removed_sent` means "the pending sent entry was
+        // removed", so the writer only ever deletes a pending-sent row.
+        let sent = contact_state_db_label(ContactState::Sent);
         let mut stmt = tx.prepare_cached(
-            "DELETE FROM contacts WHERE wallet_id = ?1 AND owner_id = ?2 AND contact_id = ?3",
+            "DELETE FROM contacts \
+             WHERE wallet_id = ?1 AND owner_id = ?2 AND contact_id = ?3 AND state = ?4",
         )?;
         for key in &cs.removed_sent {
             stmt.execute(params![
                 wallet_id.as_slice(),
                 key.owner_id.as_slice(),
                 key.recipient_id.as_slice(),
+                sent,
             ])?;
         }
     }
@@ -180,14 +190,20 @@ pub fn apply(
         }
     }
     if !cs.removed_incoming.is_empty() {
+        // Pending-received tombstone. State-filtered for the same reason
+        // as `removed_sent` above: never let a pending tombstone destroy
+        // an `established` pair-row.
+        let received = contact_state_db_label(ContactState::Received);
         let mut stmt = tx.prepare_cached(
-            "DELETE FROM contacts WHERE wallet_id = ?1 AND owner_id = ?2 AND contact_id = ?3",
+            "DELETE FROM contacts \
+             WHERE wallet_id = ?1 AND owner_id = ?2 AND contact_id = ?3 AND state = ?4",
         )?;
         for key in &cs.removed_incoming {
             stmt.execute(params![
                 wallet_id.as_slice(),
                 key.owner_id.as_slice(),
                 key.sender_id.as_slice(),
+                received,
             ])?;
         }
     }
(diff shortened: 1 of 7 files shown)


thepastaclaw:
✅ Review complete (commit ffdf73af)

packages/rs-platform-wallet-storage/src/sqlite/schema/contacts.rs
```diff
@@ -397,6 +413,36 @@ pub fn load_state_for_test(
...
+    wallet_id: &WalletId,
+) -> Result<BTreeMap<Identifier, std::collections::BTreeSet<Identifier>>, WalletStorageError> {
+    let mut stmt =
+        conn.prepare("SELECT owner_id, sender_id FROM ignored_senders WHERE wallet_id = ?1")?;
+    let mut map: BTreeMap<Identifier, std::collections::BTreeSet<Identifier>> = BTreeMap::new();
+    let mut rows = stmt.query(params![wallet_id.as_slice()])?;
+    while let Some(row) = rows.next()? {
+        let owner: Vec<u8> = row.get(0)?;
+        let sender: Vec<u8> = row.get(1)?;
+        let (owner, sender) = decode_pair_key(&owner, &sender)?;
```
thepastaclaw, line 445: **🔴 Blocking: Align test-only helper cfg gates**

`load_ignored_senders` is compiled whenever `cfg(test)` is set, but the names it uses are still compiled only with the `__test-helpers` feature: `Identifier`, `Connection`, `BTreeMap`, and `decode_pair_key`. In a unit-test build of this crate that enables `sqlite` without `__test-helpers`, Rust keeps this function but removes those dependencies before name resolution, so the test target cannot compile. Gate this function only on `__test-helpers`, or widen the dependent imports and helper to the same `#[cfg(any(test, feature = "__test-helpers"))]` condition.

<sub>source: ['codex-general', 'codex-rust-quality']</sub>

thepastaclaw (changes requested):
## Code Review

Source: reviewers = claude ffi-engineer `opus` (failed); claude general `opus` (failed); claude rust-quality `opus` (failed); claude security-auditor `opus` (failed); codex ffi-engineer `gpt-5.5`; codex general `gpt-5.5`; codex rust-quality `gpt-5.5`; codex security-auditor `gpt-5.5`; verifier = codex `gpt-5.5`; specialists = security-auditor, rust-quality, ffi-engineer.

Verified PR head 29c7188a0b196d7f564f697295325e54d4fd40f1 against the reported findings. The Rust cfg-gating issue is real and in scope because the PR adds a `cfg(test)` helper that depends on items still gated only by `__test-helpers`; the Swift embedded-NUL concern is real behavior but pre-existing and outside this PR's diff/required fix surface.

🔴 1 blocking

<details>

```

In `packages/rs-platform-wallet-storage/src/sqlite/schema/contacts.rs`:
- [BLOCKING] packages/rs-platform-wallet-storage/src/sqlite/schema/contacts.rs:428-440: Align test-only helper cfg gates
  `load_ignored_senders` is compiled whenever `cfg(test)` is set, but the names it uses are still compiled only with the `__test-helpers` feature: `Identifier`, `Connection`, `BTreeMap`, and `decode_pair_key`. In a unit-test build of this crate that enables `sqlite` without `__test-helpers`, Rust keeps this function but removes those dependencies before name resolution, so the test target cannot compile. Gate this function only on `__test-helpers`, or widen the dependent imports and helper to the same `#[cfg(any(test, feature = "__test-helpers"))]` condition.
```
</details>

QuantumExplorer:
Fixed in ffdf73affe.

**The gate mismatch is real** and I aligned it: `load_ignored_senders` was gated `#[cfg(any(test, feature = "__test-helpers"))]` (matching its caller `identities::load_state`), but the names it uses — `Identifier`, `Connection`, `BTreeMap`, and `decode_pair_key` — were gated `#[cfg(feature = "__test-helpers")]` only. I widened those four to the same `any(test, …)` condition (your option 2), so the function and everything it references now compile in exactly the same configs.

Why not the other direction (narrow the function to `__test-helpers`-only): its caller `identities::load_state` is `any(test, …)`, so narrowing the callee would just move the same mismatch up to the caller.

**One note on reachability:** this crate's dev-dependency self-reference pins `__test-helpers` on for every test build:
```toml
platform-wallet-storage = { path = ".", default-features = false, features = ["sqlite", "cli", "secrets", "kv", "__test-helpers"] }
```
so `cfg(test) ⟹ __test-helpers` holds and no real `cargo` invocation hits the broken combo — it was a latent trap, not a live break. Aligned anyway since it's a footgun the moment that dev-dep line changes or the pattern is copied.

Verified: `cargo clippy -p platform-wallet-storage --tests` clean (no unused imports in any arm), lib-only `--no-default-features --features sqlite` (production-shaped, widened items cfg'd out) compiles, and the full test suite is green (129 lib + all integration incl. the new `tc_p4_020_unignore_survives_restart…` and `pending_tombstones_do_not_destroy_established_rows`).

thepastaclaw (approved):
## Code Review

Source: reviewers = claude general `opus`; codex general `gpt-5.5`; claude security-auditor `opus`; codex security-auditor `gpt-5.5`; claude rust-quality `opus`; codex rust-quality `gpt-5.5`; claude ffi-engineer `opus`; codex ffi-engineer `gpt-5.5`; verifier = claude `opus`; specialists = security-auditor, rust-quality, ffi-engineer.

Prior reconciliation: `prior-1` [blocking] `packages/rs-platform-wallet-storage/src/sqlite/schema/contacts.rs` "Align test-only helper cfg gates" is FIXED at `ffdf73af`. The `Identifier`, `Connection`, `BTreeMap` imports and `decode_pair_key` helper now use `#[cfg(any(test, feature = "__test-helpers"))]`, matching `load_ignored_senders` and its caller `identities::load_state`.

Carried-forward prior findings: none.

New findings in latest delta (`29c7188a..ffdf73af`): none.

Cumulative current-head findings: none. All reviewer lanes converged on the prior blocker being fixed, and no new in-scope issues were verified in the latest delta or cumulative PR. Approving.

packages/rs-platform-wallet-storage/src/sqlite/schema/contacts.rs
thepastaclaw: Resolved in `ffdf73a` — _Align test-only helper cfg gates_ no longer present.

PR #4019, opened 2026-07-06
fix(platform-wallet): height-pin address balances so delta replay cannot double-count (ADDR-09)

QuantumExplorer:
ADDR-09 behavioral QA (2026-07-06) showed the top-up double-count is **not** fixed by the watermark approach (#4004/#4005): after "Top Up from Core", the topped-up platform address showed exactly **2× its proof-verified on-chain balance**, durable across automatic BLAST syncs, manual Clear+resync, and app restart. Ground truth: on-chain (via `getWithProof`) held the single credit; the client double-counted.

Root cause: two sources of truth — proof-attested **absolutes** (ST results, trunk/branch scan elements) and the recent/compacted **delta** stream — were reconciled through *global height cursors*. Any path that threaded the wrong height re-applied a delta that was already baked into an absolute. #4004/#4005 forced a full rescan via watermark invalidation, but the rescan itself replays the recent window and re-added the same delta.

## What was done?
Every absolute balance now carries a **height pin** — `AddressFunds::as_of_height`, the Platform block height the balance is current *as of*:

- **Delta gating (rs-sdk `address_sync`)**: a delta at block `B` applies only when `B > pin`; applying advances the pin. Recent loop gates per entry; compacted loop gates **per operation** (block-aware ops carry `(height, credits)`, so a pin inside a compacted range drops exactly the already-included ops).
- **Scan pinning**: trunk/branch absolutes pinned at the scan `checkpoint_height`; absence-zeroing entries too.
- **Proof heights surfaced**: `BroadcastStateTransition` gains `*_with_metadata` variants exposing the quorum-authenticated `metadata.height` (previously verified then discarded); all address/identity transition helpers return it alongside `AddressInfos`.
- **Reconcile authority (rs-platform-wallet)**: `reconcile_address_infos(infos, as_of_height, ctx)` pins ST-attested absolutes; freshness = pin ordering (nonce breaks same-block ties only). A later pin wins **even when it revises the balance downward** — protecting against stale-node scans and self-healing rows poisoned by the old bug. The now-redundant `invalidate_sync_watermark` / `credited_outputs` machinery is deleted.
- **Persistence round-trip**: sqlite `V002` migration adds `platform_addresses.as_of_height`; `AddressBalanceEntryFFI.as_of_height`; Swift stores it in `PersistentPlatformAddress.lastSeenHeight` (previously a dead column). Pin `0` = "unknown provenance" (legacy rows): deltas apply as before and any pinned absolute supersedes them.

## How was it tested?
- ~835 unit tests green across `dash-sdk`, `platform-wallet`, `platform-wallet-storage`, `platform-wallet-ffi`, `rs-sdk-ffi`, including new `commit_reconciliation_later_pin_revises_balance_downward` (the healing property) and `commit_reconciliation_drops_stale_height`.
- **On-device (testnet sim)**: fresh 0.05 DASH top-up reflected exactly once and stable through multiple automatic BLAST syncs, manual Clear + full rescan, and app restart; all four wallet addresses matched proof-verified on-chain balances exactly.
- Live race win: the post-Clear rescan hit a node whose snapshot (height 379728) was *older* than the funding proof (379731) — pin ordering preserved the newer balance instead of letting the stale scan clobber it.

## Breaking changes
None consensus-wise. SDK API: address/identity transition helpers now return the proof height (tuple extension); `reconcile_address_infos` takes `as_of_height` instead of `credited_outputs`.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Wallet and balance-sync state now includes an “as-of” block height pin to improve cross-session reconciliation and restoration.
  * Several wallet actions now surface the confirmed proof block height alongside their usual results.

* **Bug Fixes**
  * Prevented balance double-counting by making reconciliation height-pinned and dropping stale balance/absence updates.
  * Improved persistence and reload behavior so restored balances retain the correct as-of height watermark, keeping devices in sync.


thepastaclaw:
✅ Review complete (commit 4dd20980)

thepastaclaw (commented):
## Code Review

Solid, well-scoped fix for ADDR-09. The per-address height pin is threaded end-to-end (SDK → provider → SQLite → Swift) with strong test coverage and clear rationale. One public-API serde gap should be closed so legacy blobs still deserialize; other agent findings are either false positives (the raw FFI debug path documents dropping the height on purpose) or speculative nits not caused by this PR.

Source: reviewers claude/opus + codex/gpt-5.5 for general, security-auditor, rust-quality, and ffi-engineer; verifier claude/opus; policy gate review_policy.enforce_backport_prereq_policy.

🟡 1 suggestion(s)

<details>

```

In `packages/rs-sdk/src/platform/address_sync/types.rs`:
- [SUGGESTION] packages/rs-sdk/src/platform/address_sync/types.rs:78-80: `AddressFunds::as_of_height` breaks serde compatibility for legacy blobs
  `AddressFunds` derives `Serialize`/`Deserialize` behind the `serde` feature, and this PR adds `as_of_height` without `#[serde(default)]`. Any external consumer using `dash-sdk/serde` to persist `AddressFunds` directly will now fail to deserialize pre-PR blobs, even though the field's own doc-comment explicitly promises legacy-safe semantics (`0` = "unknown provenance"). The in-tree `changeset/serde_adapters.rs` correctly uses `#[serde(default)]` on its `Wire` mirror, but the public type does not — so anything not routed through that adapter breaks. Add `#[serde(default)]` to align the derive with the documented "pin = 0 for legacy" contract.
```
</details>

packages/rs-sdk/src/platform/address_sync/types.rs
```diff
@@ -55,6 +55,29 @@ pub struct AddressFunds {
...
+    ///   absolute); applying it advances the pin to `B`.
+    ///
+    /// Freshness between two absolutes is decided by comparing pins — a
+    /// later pin is authoritative *even when it revises the balance
+    /// downward* (nonces only advance on outgoing ops, so they cannot
+    /// order receive-only state).
+    ///
+    /// `0` means "unknown provenance" (legacy rows persisted before the
+    /// pin existed): every delta applies, matching pre-pin behavior.
+    pub as_of_height: u64,
```
thepastaclaw, line 80: **🟡 Suggestion: `AddressFunds::as_of_height` breaks serde compatibility for legacy blobs**

`AddressFunds` derives `Serialize`/`Deserialize` behind the `serde` feature, and this PR adds `as_of_height` without `#[serde(default)]`. Any external consumer using `dash-sdk/serde` to persist `AddressFunds` directly will now fail to deserialize pre-PR blobs, even though the field's own doc-comment explicitly promises legacy-safe semantics (`0` = "unknown provenance"). The in-tree `changeset/serde_adapters.rs` correctly uses `#[serde(default)]` on its `Wire` mirror, but the public type does not — so anything not routed through that adapter breaks. Add `#[serde(default)]` to align the derive with the documented "pin = 0 for legacy" contract.

```suggestion
    /// `0` means "unknown provenance" (legacy rows persisted before the
    /// pin existed): every delta applies, matching pre-pin behavior.
    #[cfg_attr(feature = "serde", serde(default))]
    pub as_of_height: u64,
```

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Source: reviewers = claude ffi-engineer `opus`; claude general `opus`; claude rust-quality `opus`; claude security-auditor `opus`; codex ffi-engineer `gpt-5.5`; codex general `gpt-5.5`; codex rust-quality `gpt-5.5`; codex security-auditor `gpt-5.5`; verifier = claude `opus`; specialists = security-auditor, rust-quality, ffi-engineer; policy gate = review_policy.enforce_backport_prereq_policy.

Prior reconciliation: prior-1 STILL VALID at packages/rs-sdk/src/platform/address_sync/types.rs:80 — file unchanged in latest delta. Carried-forward: `AddressFunds::as_of_height` still lacks `#[serde(default)]`. New in latest delta (wasm-sdk destructuring of proof height): nothing to flag — the wasm surface is render-only. Cumulative: one moderate-confidence single-source suggestion from codex-security-auditor about the sync-path emit lacking pin-freshness enforcement, which would let a stale full-scan clobber a fresher commit_reconciliation row.

🟡 1 suggestion(s)

<details>

```

In `packages/rs-platform-wallet/src/wallet/platform_addresses/sync.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/platform_addresses/sync.rs:49-61: Sync-path diff emits without checking pin freshness against `before`
  `compute_address_balance_diff` only skips a `found` entry when `before.get(&tag) == Some(&funds)` — an exact `AddressFunds` equality check. This means a `found` entry whose `as_of_height` is *below* the snapshot's pin still emits (the pin field differs, so the funds compare unequal) and gets persisted. The freshness guard the PR introduces in `commit_reconciliation` (provider.rs, per its doc block) is therefore not mirrored here. If the address trunk/branch scan returns a stale-but-valid proof — the SDK path disables metadata freshness checks and the recent catch-up query is best-effort — the resulting older-pinned funds will clobber a fresher `commit_reconciliation` row on disk. The lock-order argument in the comment at lines 169–174 protects against interleaved writes, not against stale inputs to the sync itself. Applying the same pin-freshness rule (`funds.as_of_height >= before.get(&tag).map(|f| f.as_of_height).unwrap_or(0)`) before emitting closes the gap and keeps the mitigation symmetrical across both writers. Single-source finding (codex-security-auditor only, moderate confidence); defense-in-depth rather than a proven exploit path.
```
</details>

thepastaclaw (commented):
## Code Review

Incremental + cumulative review at `cfa400c2`. Both prior findings from `b8eedda4` remain STILL VALID and are carried forward. The latest delta rewrites the SDK address-sync apply path; I found two additional suggestion-level issues there/at the adjacent standalone FFI surface, but no blockers.

Source: reviewers = claude general `opus` (ok); codex general `gpt-5.5` (ok); claude security-auditor `opus` (ok); codex security-auditor `gpt-5.5` (ok); claude rust-quality `opus` (ok); codex rust-quality `gpt-5.5` (ok); claude ffi-engineer `opus` (ok); codex ffi-engineer `gpt-5.5` (ok); verifier = claude `opus` (ok); specialists = security-auditor, rust-quality, ffi-engineer; policy gate = review_policy.enforce_backport_prereq_policy.

🟡 4 suggestion(s)

**Prior Reconciliation**
- `prior-1` STILL VALID: Verified at packages/rs-sdk/src/platform/address_sync/types.rs:52,80 — struct still derives Serialize/Deserialize behind the serde feature and as_of_height still lacks #[serde(default)].
- `prior-2` STILL VALID: Verified at packages/rs-platform-wallet/src/wallet/platform_addresses/sync.rs:49-96 — the found loop skips only when `before.get(&tag) == Some(&funds)` (exact equality) and the absent-zeroing loop only checks that before had non-default funds; neither branch compares the emitted pin to the snapshot pin.

**Carried-Forward Prior Findings**
- **[SUGGESTION]** `packages/rs-sdk/src/platform/address_sync/types.rs:78-80`: AddressFunds::as_of_height breaks serde compatibility for legacy blobs
  AddressFunds derives Serialize/Deserialize behind the `serde` feature (line 52) but the newly added `as_of_height` field has no `#[serde(default)]`. The doc-comment on lines 78–79 explicitly promises legacy-safe semantics (pin `0` = "unknown provenance", every delta applies), but any external consumer of `dash-sdk/serde` that persists AddressFunds directly (JS/Swift bindings via serde_json, downstream Rust caches, on-disk snapshots written pre-PR) will fail to deserialize with a `missing field "as_of_height"` error rather than loading with the documented pin-0 legacy contract. The in-tree `changeset/serde_adapters.rs::Wire` mirror already uses `#[serde(default)]`, so the wallet-changeset path is fine — but the public SDK type is not routed through that adapter. Adding `#[serde(default)]` to the field aligns the derive with the documented contract.
- **[SUGGESTION]** `packages/rs-platform-wallet/src/wallet/platform_addresses/sync.rs:48-96`: compute_address_balance_diff can persist stale-pinned scan results over fresher rows
  The found loop at lines 49-61 skips only when `before.get(&tag) == Some(&funds)` — a byte-for-byte AddressFunds equality that includes `as_of_height`. A found entry whose `as_of_height` is *older* than the snapshot pin still compares unequal and is emitted, so a stale-but-valid trunk/branch scan proof can overwrite a fresher `commit_reconciliation` row on disk. The absent-zeroing loop at lines 65-96 has the same shape — it emits a zero pinned at `absent_as_of_height` without checking that the scan checkpoint is newer than the snapshot pin. `commit_reconciliation` in provider.rs already applies the height-freshness rule (`existing.as_of_height > entry.funds.as_of_height ⇒ stale`); mirroring that rule here keeps both writers into the same rows consistent. The lock ordering at sync.rs:169-174 serializes concurrent writers but does not filter stale inputs to this writer. Skip the emit when `funds.as_of_height < before.get(&tag).map(|f| f.as_of_height).unwrap_or(0)` on the found branch, and when `absent_as_of_height < …` on the absent branch. Defense-in-depth against DAPI node lag / eclipse routing — the PR description itself calls out that a scan at height 379728 can arrive alongside a funding proof pinned at 379731 in the wild.

**New Findings In Latest Delta / Current Head**
- **[SUGGESTION]** `packages/rs-sdk/src/platform/address_sync/mod.rs:958-984`: pending_unknown accumulates foreign-wallet changes without a hard cap on shared chains
  The new unknown-address path clones every unmatched recent/compacted change into `pending_unknown` and the in-code comment explicitly leaves the buffer unbounded, relying on a one-shot warn at PENDING_UNKNOWN_WARN_THRESHOLD (1000). On a shared/busy chain the incoming stream is *global*: most changes seen during pagination belong to other wallets and remain buffered until the end-of-pass refresh, where they get discarded at the `unknown_keys` intersection. The comment correctly notes that a bounded-memory fix would need to re-query, not just stash keys, so replacing the design isn't trivial — but for mobile/embedded consumers, letting one wallet sync retain unbounded foreign owned-changes across a long catch-up window is a real memory-pressure risk (`OwnedBalanceOp::Compacted` carries a Vec of arbitrary length). Consider either (a) tightening to a hard byte-count cap that triggers a mid-pass drain + refresh, or (b) draining pending_unknown after each pagination page rather than at end-of-pass so foreign changes are discarded promptly. The current tradeoff is defensible for well-behaved networks, so this is a hardening suggestion rather than a blocker — but the warn hook alone doesn't prevent the exhaustion it detects.
- **[SUGGESTION]** `packages/rs-sdk-ffi/src/address_sync/types.rs:93-108`: Standalone Swift batch sync (DashSDKFoundAddress) drops the new as_of_height pin
  `dash_sdk_sync_addresses_batch_with_result` returns `DashSDKFoundAddress` over the C boundary and Swift feeds those values back as `knownBalances` on later calls. The C struct exposes only index/key/nonce/balance — the per-address `as_of_height` that this PR introduces on AddressFunds is not carried across. When the Rust side rebuilds AddressFunds from FFI inputs it fills `as_of_height: 0`, which per the field's own doc-comment means "unknown provenance ⇒ every delta applies". Standalone Swift consumers of this batch API therefore get pre-PR delta-replay semantics across persistence/restart, even though the platform-wallet FFI path (AddressBalanceEntryFFI) does round-trip the pin. Add a `as_of_height: u64` field to DashSDKFoundAddress (mirror it on the Swift Codable) so ADDR-09 replay protection extends to that surface too.

<details>

```

In `packages/rs-sdk/src/platform/address_sync/types.rs`:
- [SUGGESTION] packages/rs-sdk/src/platform/address_sync/types.rs:78-80: AddressFunds::as_of_height breaks serde compatibility for legacy blobs
  AddressFunds derives Serialize/Deserialize behind the `serde` feature (line 52) but the newly added `as_of_height` field has no `#[serde(default)]`. The doc-comment on lines 78–79 explicitly promises legacy-safe semantics (pin `0` = "unknown provenance", every delta applies), but any external consumer of `dash-sdk/serde` that persists AddressFunds directly (JS/Swift bindings via serde_json, downstream Rust caches, on-disk snapshots written pre-PR) will fail to deserialize with a `missing field "as_of_height"` error rather than loading with the documented pin-0 legacy contract. The in-tree `changeset/serde_adapters.rs::Wire` mirror already uses `#[serde(default)]`, so the wallet-changeset path is fine — but the public SDK type is not routed through that adapter. Adding `#[serde(default)]` to the field aligns the derive with the documented contract.

In `packages/rs-platform-wallet/src/wallet/platform_addresses/sync.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/platform_addresses/sync.rs:48-96: compute_address_balance_diff can persist stale-pinned scan results over fresher rows
  The found loop at lines 49-61 skips only when `before.get(&tag) == Some(&funds)` — a byte-for-byte AddressFunds equality that includes `as_of_height`. A found entry whose `as_of_height` is *older* than the snapshot pin still compares unequal and is emitted, so a stale-but-valid trunk/branch scan proof can overwrite a fresher `commit_reconciliation` row on disk. The absent-zeroing loop at lines 65-96 has the same shape — it emits a zero pinned at `absent_as_of_height` without checking that the scan checkpoint is newer than the snapshot pin. `commit_reconciliation` in provider.rs already applies the height-freshness rule (`existing.as_of_height > entry.funds.as_of_height ⇒ stale`); mirroring that rule here keeps both writers into the same rows consistent. The lock ordering at sync.rs:169-174 serializes concurrent writers but does not filter stale inputs to this writer. Skip the emit when `funds.as_of_height < before.get(&tag).map(|f| f.as_of_height).unwrap_or(0)` on the found branch, and when `absent_as_of_height < …` on the absent branch. Defense-in-depth against DAPI node lag / eclipse routing — the PR description itself calls out that a scan at height 379728 can arrive alongside a funding proof pinned at 379731 in the wild.

In `packages/rs-sdk/src/platform/address_sync/mod.rs`:
- [SUGGESTION] packages/rs-sdk/src/platform/address_sync/mod.rs:958-984: pending_unknown accumulates foreign-wallet changes without a hard cap on shared chains
  The new unknown-address path clones every unmatched recent/compacted change into `pending_unknown` and the in-code comment explicitly leaves the buffer unbounded, relying on a one-shot warn at PENDING_UNKNOWN_WARN_THRESHOLD (1000). On a shared/busy chain the incoming stream is *global*: most changes seen during pagination belong to other wallets and remain buffered until the end-of-pass refresh, where they get discarded at the `unknown_keys` intersection. The comment correctly notes that a bounded-memory fix would need to re-query, not just stash keys, so replacing the design isn't trivial — but for mobile/embedded consumers, letting one wallet sync retain unbounded foreign owned-changes across a long catch-up window is a real memory-pressure risk (`OwnedBalanceOp::Compacted` carries a Vec of arbitrary length). Consider either (a) tightening to a hard byte-count cap that triggers a mid-pass drain + refresh, or (b) draining pending_unknown after each pagination page rather than at end-of-pass so foreign changes are discarded promptly. The current tradeoff is defensible for well-behaved networks, so this is a hardening suggestion rather than a blocker — but the warn hook alone doesn't prevent the exhaustion it detects.

In `packages/rs-sdk-ffi/src/address_sync/types.rs`:
- [SUGGESTION] packages/rs-sdk-ffi/src/address_sync/types.rs:93-108: Standalone Swift batch sync (DashSDKFoundAddress) drops the new as_of_height pin
  `dash_sdk_sync_addresses_batch_with_result` returns `DashSDKFoundAddress` over the C boundary and Swift feeds those values back as `knownBalances` on later calls. The C struct exposes only index/key/nonce/balance — the per-address `as_of_height` that this PR introduces on AddressFunds is not carried across. When the Rust side rebuilds AddressFunds from FFI inputs it fills `as_of_height: 0`, which per the field's own doc-comment means "unknown provenance ⇒ every delta applies". Standalone Swift consumers of this batch API therefore get pre-PR delta-replay semantics across persistence/restart, even though the platform-wallet FFI path (AddressBalanceEntryFFI) does round-trip the pin. Add a `as_of_height: u64` field to DashSDKFoundAddress (mirror it on the Swift Codable) so ADDR-09 replay protection extends to that surface too.
```
</details>

packages/rs-platform-wallet/src/wallet/platform_addresses/sync.rs
```diff
@@ -82,10 +83,14 @@ pub(crate) fn compute_address_balance_diff(
...
+            // absence was proven by the tree scan, so the zero is pinned
+            // at the scan's checkpoint height (`absent_as_of_height`) —
+            // a later re-credit carries a higher pin and wins.
             funds: AddressFunds {
                 balance: 0,
                 nonce: 0,
+                as_of_height: absent_as_of_height,
             },
         });
     }
```
thepastaclaw, line 111: **🟡 Suggestion: compute_address_balance_diff can persist stale-pinned scan results over fresher rows**

The found loop at lines 49-61 skips only when `before.get(&tag) == Some(&funds)` — a byte-for-byte AddressFunds equality that includes `as_of_height`. A found entry whose `as_of_height` is *older* than the snapshot pin still compares unequal and is emitted, so a stale-but-valid trunk/branch scan proof can overwrite a fresher `commit_reconciliation` row on disk. The absent-zeroing loop at lines 65-96 has the same shape — it emits a zero pinned at `absent_as_of_height` without checking that the scan checkpoint is newer than the snapshot pin. `commit_reconciliation` in provider.rs already applies the height-freshness rule (`existing.as_of_height > entry.funds.as_of_height ⇒ stale`); mirroring that rule here keeps both writers into the same rows consistent. The lock ordering at sync.rs:169-174 serializes concurrent writers but does not filter stale inputs to this writer. Skip the emit when `funds.as_of_height < before.get(&tag).map(|f| f.as_of_height).unwrap_or(0)` on the found branch, and when `absent_as_of_height < …` on the absent branch. Defense-in-depth against DAPI node lag / eclipse routing — the PR description itself calls out that a scan at height 379728 can arrive alongside a funding proof pinned at 379731 in the wild.

```suggestion
    // 1. Found-and-changed.
    for (&(tag, p2pkh), &funds) in found {
        if let Some(existing) = before.get(&tag) {
            // Skip when unchanged, or when the newly scanned pin is
            // older than the already committed pin — a stale-but-valid
            // proof must not clobber a fresher row.
            if existing == &funds || existing.as_of_height > funds.as_of_height {
                continue;
            }
        }
        let (wallet_id, account_index, address_index) = tag;
        entries.push(PlatformAddressBalanceEntry {
            wallet_id,
            account_index,
            address_index,
            address: p2pkh,
            funds,
        });
    }

    // 2. Absent-but-previously-funded — zero out cached balances that the
    //    new chain no longer knows about.
    for &(tag, p2pkh) in absent {
        // Absent at the checkpoint but re-found during incremental
        // catch-up — the find is newer (tip), so never zero it.
        if found.contains_key(&(tag, p2pkh)) {
            continue;
        }
        let Some(existing) = before.get(&tag) else {
            continue;
        };
        // Only emit when we actually had non-default cached funds for the
        // address. An address that was already empty (or never cached)
        // doesn't need a zeroing write.
        if existing.balance == 0 && existing.nonce == 0 {
            continue;
        }
        // A stale absence proof must not clobber a row pinned by a newer
        // proof or transition reconciliation.
        if existing.as_of_height > absent_as_of_height {
            continue;
        }
        let (wallet_id, account_index, address_index) = tag;
        entries.push(PlatformAddressBalanceEntry {
            wallet_id,
            account_index,
            address_index,
            address: p2pkh,
            // Absent in state ⇒ no balance and no nonce. Reset both. The
            // absence was proven by the tree scan, so the zero is pinned
            // at the scan's checkpoint height (`absent_as_of_height`) —
            // a later re-credit carries a higher pin and wins.
            funds: AddressFunds {
                balance: 0,
                nonce: 0,
                as_of_height: absent_as_of_height,
            },
        });
    }
```

<sub>source: ['claude', 'codex', 'prior']</sub>

packages/rs-sdk/src/platform/address_sync/types.rs
thepastaclaw: Resolved in this update — _`AddressFunds::as_of_height` breaks serde compatibility for legacy blobs_ no longer present.

packages/rs-platform-wallet/src/wallet/platform_addresses/sync.rs
QuantumExplorer: Good catch — taken in 78a16ceab4, with one extension: the suggestion as written guards only the persist diff, which would leave `sync_finished`'s scratch merge (`found.extend(...)`) regressing the **in-memory** seed while disk holds the fresher row — a disk-vs-memory divergence in exactly the protected scenario. So the height-freshness rule now lands **symmetrically in both writers**:

- `compute_address_balance_diff`: your two skips (older-pinned found entry; absence zeroing whose scan checkpoint is older than the committed pin), essentially as suggested.
- `sync_finished`: per-entry scratch merge keeping the committed entry when its pin is fresher.

Absent-removal stays unguarded in memory (absence carries no per-entry height at the provider layer); the diff guard protects the durable row, and the next pass reconstructs the in-memory entry from a full replay (base 0, pin 0 → every delta applies).

Worth noting the blast radius even unguarded was transient: the regressed row keeps its old pin, so the next pass on a healthy node re-applies the missing delta and self-heals within one cycle — but no reason to allow the window at all. Three new tests cover the stale-pin found skip, the stale-checkpoint absence skip, and the merge freshness rule (plus newer-pin paths still landing).

thepastaclaw (commented):
## Code Review

The delta from 78a16cea..a93f7caf is documentation-only in Rust (doc-comment clarifications on top_up_address / transfer_to_addresses and one intra-doc link fix on address_sync/types.rs) plus a Swift persist-test update to the 7-tuple shape asserting the pin round-trips into lastSeenHeight. No new Rust logic or FFI signatures changed. Prior-1 (AddressFunds::as_of_height lacks #[serde(default)]) and prior-3 (pending_unknown unbounded, intentionally deferred by author) remain STILL VALID and are carried forward. No new blockers.

Source: reviewers = claude general `opus` (ok); codex general `gpt-5.5` (ok); claude security-auditor `opus` (ok); codex security-auditor `gpt-5.5` (ok); claude rust-quality `opus` (ok); codex rust-quality `gpt-5.5` (ok); claude ffi-engineer `opus` (ok); codex ffi-engineer `gpt-5.5` (ok); verifier = claude `opus` (ok); specialists = security-auditor, rust-quality, ffi-engineer; policy gate = review_policy.enforce_backport_prereq_policy.

🟡 2 suggestion(s)

**Prior Reconciliation**
- `prior-1` STILL VALID: verified at `packages/rs-sdk/src/platform/address_sync/types.rs:52,80`; `AddressFunds` still derives serde when the feature is enabled, and `as_of_height` still lacks `#[serde(default)]`.
- `prior-3` STILL VALID: verified at `packages/rs-sdk/src/platform/address_sync/mod.rs:958-984`; `pending_unknown` still appends unmatched global changes into an intentionally unbounded buffer with only the one-shot threshold warning.

**Carried-Forward Prior Findings**
- **[SUGGESTION]** `packages/rs-sdk/src/platform/address_sync/types.rs:78-80`: AddressFunds::as_of_height breaks serde compatibility for legacy blobs
  AddressFunds derives Serialize/Deserialize behind the `serde` feature (line 52), but the new `as_of_height` field on line 80 has no `#[serde(default)]`. The field's own doc-comment on lines 78–79 explicitly promises legacy-safe semantics — pin `0` = 'unknown provenance', every delta applies, matching pre-pin behavior. Without the serde default, any external `dash-sdk/serde` consumer decoding a pre-PR persisted blob (JS/Swift wrappers, downstream Rust caches, on-disk snapshots) fails with `missing field "as_of_height"` instead of loading the documented pin-0 fallback. The in-tree wallet-changeset path is safe because `changeset/serde_adapters.rs::Wire` uses `#[serde(default)]`, but the public SDK type is not routed through that adapter. Adding the attribute aligns the derive with the documented contract. Carried forward from prior review; the latest delta touched this file only for an intra-doc link and did not address this.
- **[SUGGESTION]** `packages/rs-sdk/src/platform/address_sync/mod.rs:958-984`: pending_unknown accumulates foreign-wallet changes without a hard cap
  The unknown-address branch clones every unmatched recent/compacted balance change into `pending_unknown` and the in-line comment (964–976) intentionally leaves the buffer unbounded, relying on a one-shot warn at PENDING_UNKNOWN_WARN_THRESHOLD (1000). The recent/compacted streams are global to the chain, so on a busy shared chain most entries belong to other wallets and remain buffered until the end-of-pass refresh intersects them with `unknown_keys`. `OwnedBalanceOp::Compacted` carries a `Vec` of arbitrary length, so for mobile/embedded consumers on a long catch-up window this is a real memory-pressure surface; a stale/eclipsing DAPI route could amplify it. The author's comment correctly notes that a bounded-memory fix must re-query rather than merely stash keys, so this remains a hardening suggestion — but the warn hook only observes the condition, it does not prevent exhaustion. Consider a hard byte-count cap that triggers a mid-pass drain + refresh, or draining `pending_unknown` after each pagination page rather than end-of-pass.

**New Findings In Latest Delta**
- None. The latest delta from `78a16cea` to `a93f7caf` only updates Rust docs/intra-doc links and Swift persistence tests for the height-pin tuple.

<details>

```

In `packages/rs-sdk/src/platform/address_sync/types.rs`:
- [SUGGESTION] packages/rs-sdk/src/platform/address_sync/types.rs:78-80: AddressFunds::as_of_height breaks serde compatibility for legacy blobs
  AddressFunds derives Serialize/Deserialize behind the `serde` feature (line 52), but the new `as_of_height` field on line 80 has no `#[serde(default)]`. The field's own doc-comment on lines 78–79 explicitly promises legacy-safe semantics — pin `0` = 'unknown provenance', every delta applies, matching pre-pin behavior. Without the serde default, any external `dash-sdk/serde` consumer decoding a pre-PR persisted blob (JS/Swift wrappers, downstream Rust caches, on-disk snapshots) fails with `missing field "as_of_height"` instead of loading the documented pin-0 fallback. The in-tree wallet-changeset path is safe because `changeset/serde_adapters.rs::Wire` uses `#[serde(default)]`, but the public SDK type is not routed through that adapter. Adding the attribute aligns the derive with the documented contract. Carried forward from prior review; the latest delta touched this file only for an intra-doc link and did not address this.

In `packages/rs-sdk/src/platform/address_sync/mod.rs`:
- [SUGGESTION] packages/rs-sdk/src/platform/address_sync/mod.rs:958-984: pending_unknown accumulates foreign-wallet changes without a hard cap
  The unknown-address branch clones every unmatched recent/compacted balance change into `pending_unknown` and the in-line comment (964–976) intentionally leaves the buffer unbounded, relying on a one-shot warn at PENDING_UNKNOWN_WARN_THRESHOLD (1000). The recent/compacted streams are global to the chain, so on a busy shared chain most entries belong to other wallets and remain buffered until the end-of-pass refresh intersects them with `unknown_keys`. `OwnedBalanceOp::Compacted` carries a `Vec` of arbitrary length, so for mobile/embedded consumers on a long catch-up window this is a real memory-pressure surface; a stale/eclipsing DAPI route could amplify it. The author's comment correctly notes that a bounded-memory fix must re-query rather than merely stash keys, so this remains a hardening suggestion — but the warn hook only observes the condition, it does not prevent exhaustion. Consider a hard byte-count cap that triggers a mid-pass drain + refresh, or draining `pending_unknown` after each pagination page rather than end-of-pass.

```
</details>

packages/rs-platform-wallet/src/wallet/platform_addresses/sync.rs
thepastaclaw: Resolved in this update — _compute_address_balance_diff can persist stale-pinned scan results over fresher rows_ no longer present.

thepastaclaw (commented):
## Code Review

Incremental + cumulative review at `4dd20980`. The latest delta `a93f7caf..4dd20980` is a broad `v4.1-dev` merge and does not change the ADDR-09 height-pin files. Both prior findings remain STILL VALID and are carried forward; the verifier also kept one additional cumulative current-head suggestion around stale absence clearing in the in-memory provider state.

Source: reviewers = claude general `opus` (ok); codex general `gpt-5.5` (ok); claude security-auditor `opus` (ok); codex security-auditor `gpt-5.5` (ok); claude rust-quality `opus` (ok); codex rust-quality `gpt-5.5` (ok); claude ffi-engineer `opus` (ok); codex ffi-engineer `gpt-5.5` (ok); verifier = claude `opus` (ok); specialists = security-auditor, rust-quality, ffi-engineer; policy gate = review_policy.enforce_backport_prereq_policy.

3 suggestion(s)

**Prior Reconciliation**
- `prior-1` STILL VALID: `packages/rs-sdk/src/platform/address_sync/types.rs:52` still derives serde behind the `serde` feature, and `as_of_height` at line 80 still lacks `#[serde(default)]`; the latest delta did not modify this file.
- `prior-2` STILL VALID: `packages/rs-sdk/src/platform/address_sync/mod.rs:958-984` still appends unmatched global changes into `pending_unknown` with only the one-shot warning threshold and no hard cap; the latest delta did not modify this file.

**Carried-Forward Prior Findings**
- **[SUGGESTION]** `packages/rs-sdk/src/platform/address_sync/types.rs:78-80`: AddressFunds::as_of_height breaks serde compatibility for legacy blobs
  `AddressFunds` derives `Serialize`/`Deserialize` behind the `serde` feature, but the new `as_of_height: u64` field has no `#[serde(default)]`. The field docs define pin `0` as the legacy/unknown-provenance fallback, so external `dash-sdk/serde` consumers decoding pre-PR blobs can fail with `missing field "as_of_height"` instead of loading the documented pin-0 behavior. The in-tree wallet changeset adapter is safe, but the public SDK type is not routed through that adapter.
- **[SUGGESTION]** `packages/rs-sdk/src/platform/address_sync/mod.rs:958-984`: pending_unknown accumulates foreign-wallet changes without a hard cap
  The unknown-address branch clones every unmatched recent/compacted balance change into `pending_unknown` until the end-of-pass refresh. Recent/compacted streams are global to the chain, so busy shared-chain catch-up can buffer mostly foreign-wallet changes; `OwnedBalanceOp::Compacted` can also carry an arbitrary-length `Vec`. The existing warning observes the condition but does not bound memory. A byte-count cap with mid-pass drain/refresh, or per-page draining, would close the hardening gap.

**New Findings In Latest Delta**
- None. The latest delta is a merge from `v4.1-dev`; it does not touch the two carried-forward ADDR-09 files.

**Additional Cumulative Current-Head Finding**
- **[SUGGESTION]** `packages/rs-platform-wallet/src/wallet/platform_addresses/provider.rs:840-847`: Stale absence pass still clears in-memory found entries even when durable rows are pin-fresher
  In `sync_finished`, the `found` merge enforces height-pin freshness, but the following absent-removal is not pin-guarded: every address in `account_scratch.absent` is removed from `account_state.found`, and `on_address_absent` has already zeroed the managed-account balance cache. In the stale-node race this PR targets, disk keeps the fresher pinned row, but the active in-memory session can lose the funded address until reload or a later repair pass. The code comment documents this tradeoff, so this remains a suggestion rather than a blocker; consider skipping removal when the committed pin is fresher than the pass, mirroring the found-merge rule.

<details>

```

In `packages/rs-sdk/src/platform/address_sync/types.rs`:
- [SUGGESTION] packages/rs-sdk/src/platform/address_sync/types.rs:78-80: AddressFunds::as_of_height breaks serde compatibility for legacy blobs
  `AddressFunds` derives `Serialize`/`Deserialize` behind the `serde` feature (types.rs:52), but the new `as_of_height: u64` field at line 80 has no `#[serde(default)]`. The field's own doc-comment on lines 78-79 documents pin `0` as "unknown provenance" -- the legacy value for rows persisted before the pin existed. Without the serde default, any external `dash-sdk/serde` consumer decoding a pre-PR persisted blob fails instead of loading at the documented pin-0 fallback.

In `packages/rs-sdk/src/platform/address_sync/mod.rs`:
- [SUGGESTION] packages/rs-sdk/src/platform/address_sync/mod.rs:958-984: pending_unknown accumulates foreign-wallet changes without a hard cap
  The unknown-address branch clones every unmatched recent/compacted balance change into `pending_unknown` until end-of-pass refresh. The recent/compacted streams are global to the chain, so on a busy shared chain most entries belong to foreign wallets and remain buffered indefinitely. The one-shot warning observes but does not bound memory.

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/platform_addresses/provider.rs:840-847: Stale absence pass still clears in-memory found entries even when durable rows are pin-fresher
  The found merge enforces height-pin freshness, but the absent-removal path is not pin-guarded. A stale-node absence can remove the in-memory found entry and zero the managed-account balance cache even though durable storage keeps the fresher pinned row.
```
</details>

packages/rs-platform-wallet/src/wallet/platform_addresses/provider.rs
```diff
@@ -883,7 +821,27 @@ impl AddressProvider for PlatformPaymentAddressProvider {
...
+                    }
+                }
+                // Absence carries no per-entry height, so a stale pass's
+                // removal is NOT pin-guarded here; the persist diff skips
+                // the durable zero for fresher-pinned rows, and the next
+                // pass reconstructs the in-memory entry from a full
+                // replay (base 0, pin 0 → every delta applies).
                 for absent_addr in &account_scratch.absent {
                     account_state.found.remove(absent_addr);
                 }
```
thepastaclaw, line 847: **🟡 Suggestion: Stale absence pass still clears in-memory found entries even when durable rows are pin-fresher**

In `sync_finished`, the `found` merge at lines 832–838 correctly enforces height-pin freshness (the fresher committed entry wins). The following absent-removal at lines 845–847 is NOT pin-guarded: every address in `account_scratch.absent` is removed from `account_state.found`, and `on_address_absent` (line 788) has already zeroed the managed-account balance cache. In the stale-node race this PR is designed to handle, disk keeps the fresher pinned row (the persist diff at `compute_address_balance_diff` guards it), but the in-memory session loses the funded address. Spending/selection reads `current_balances()` from that same in-memory `found` map, so a valid funded address can disappear until reload or a later repair pass. The author's comment at lines 840–844 documents this as intentional — noting the next pass will reconstruct from full replay — so this is a hardening suggestion rather than a blocker; consider skipping the removal for entries whose committed pin is fresher than the pass, mirroring the found-merge rule.

<sub>source: ['codex']</sub>

PR #4023, opened 2026-07-06
test(dapi): make transactionsFilter bloom-filter test deterministic

QuantumExplorer:

The `testTransactionAgainstFilter` unit test **"should not add outpoint to the filter if BLOOM_UPDATE_NONE flag is set in the filter"** is flaky. It intermittently fails with:

```
AssertionError: expected true to be false
```

This flaked on CI run [28774171971](https://github.com/dashpay/platform/actions/runs/28774171971) (job 85315592482) for an unrelated PR (#3841) that touches no dapi code.

**Root cause:** the affected tests build a fresh `BloomFilter.create(1, 0.0001)`, insert a **randomly generated** address/pubkey (`new PrivateKey().toAddress()`), and then assert that unrelated data does **not** match. But `testTransactionAgainstFilter` also runs the transaction hash (`checkHash`) and the input outpoint `txid:vout` (`checkInputs`) through `filter.contains()` — neither of which was inserted. A bloom filter has a non-zero false-positive rate, so with random inputs those negative checks occasionally match by chance and the expected `false` flips to `true`.

The same latent flakiness affects every test in the file that generates random keys and asserts `.to.be.false()` (the "no such output", "input doesn't contain desired data", `BLOOM_UPDATE_P2PUBKEY_ONLY` pub-key-hash, and multisig `BLOOM_UPDATE_NONE` cases).

## What was done?

Replaced the random `new PrivateKey()` inputs with a fixed pool of deterministic test-vector private keys (`0x01…0x0a`), mirroring the idiom already used by the neighboring *"should pass the same test vector as dashcore implementation does"* test. Every address, transaction hash, and outpoint is now fixed, so each assertion is fully reproducible. A comment documents why the keys must stay fixed.

No production code changed — this is a test-only fix.

## How Has This Been Tested?

Because the repo's `node_modules` weren't installed locally, I ran the **actual post-edit spec file** through the same tool versions the package pins (mocha 11, chai 4, dirty-chai 2, `@dashevo/dashcore-lib` 0.22.0) in an isolated sandbox against an exact copy of the lib under test:

| Version | Runs | Failures |
| --- | --- | --- |
| Original (random keys) | 200 | **2** (~1%) — flake reproduced |
| Fixed (this PR) | **400** | **0** |

All 12 tests in the file pass. Reviewers can confirm in an installed tree with:

```bash
yarn workspace @dashevo/dapi test:unit
```

## Breaking Changes

None.

## Checklist:

## Summary by CodeRabbit

* **Tests**
  * Updated transaction filter tests to use fixed, reproducible key vectors instead of randomly generated keys.
  * Reduced occasional flaky failures by making address and public-key test inputs deterministic.
  * Kept all changes confined to test coverage; no production behavior was changed.

### DIFF
--- a/packages/dapi/test/unit/transactionsFilter/testTransactionAgainstFilter.spec.js
+++ b/packages/dapi/test/unit/transactionsFilter/testTransactionAgainstFilter.spec.js
@@ -15,10 +15,32 @@ chai.use(dirtyChai);
 
 const testTransactionAgainstFilter = require('../../../lib/transactionsFilter/testTransactionAgainstFilter');
 
+// Deterministic test-vector private keys.
+//
+// These tests build a fresh, single-element BloomFilter and then assert that
+// unrelated data (an outpoint, a different address, a transaction hash) does
+// NOT match. A bloom filter has a non-zero false-positive rate, so with
+// randomly generated keys those negative assertions occasionally match by
+// chance and the test flakes (observed rate ~0.3% for the BLOOM_UPDATE_NONE
+// case). Fixed keys make every assertion reproducible. Do not replace these
+// with `new PrivateKey()`.
+const testPrivateKeys = [
+  '0000000000000000000000000000000000000000000000000000000000000001',
+  '0000000000000000000000000000000000000000000000000000000000000002',
+  '0000000000000000000000000000000000000000000000000000000000000003',
+  '0000000000000000000000000000000000000000000000000000000000000004',
+  '0000000000000000000000000000000000000000000000000000000000000005',
+  '0000000000000000000000000000000000000000000000000000000000000006',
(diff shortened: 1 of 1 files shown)


thepastaclaw:
✅ Review complete (commit 9988b1c2)

thepastaclaw (approved):
## Code Review

Source: reviewers `opus` (claude-general) and `gpt-5.5` (codex-general); verifier `opus` (claude); specialists: none.

Both agents independently reviewed this test-only PR that replaces random private keys with deterministic secp256k1 test vectors in the DAPI transactionsFilter unit tests. Neither found any in-scope defects, and the change is well-scoped to eliminating bloom-filter false-positive flakiness. Approving.

PR #4032, opened 2026-07-06
test(platform-wallet): pin Orchard key derivation to official ZIP-32 vectors

QuantumExplorer:

The shielded wallet's Orchard key derivation (`OrchardKeySet::from_seed`) is only as trustworthy as the `grovedb_commitment_tree` → `orchard` dependency chain beneath it. A silent change in that chain — a grovedb or orchard bump — could alter FVK / IVK / OVK / address derivation and leave existing wallets unable to detect their own notes, with **no compile error** to catch it. There was no test pinning derivation to the canonical Zcash/ZIP-32 vectors.

This came out of an audit of our key handling in the wake of the June 2026 Zcash Orchard incident; derivation was verified correct, and this PR locks that in so it stays correct.

## What was done?

Added regression coverage plus a spec-required guard in `packages/rs-platform-wallet/src/wallet/shielded/keys.rs`:

- **`key_pipeline_matches_official_orchard_test_vector`** — runs Zcash test vector 0 (`orchard_key_components.py`) through the exact `grovedb_commitment_tree` re-exports that `from_seed` uses, asserting the `dk‖ivk` IVK encoding, the OVK, and the default address all match the official bytes.
- **`from_seed_is_deterministic_and_domain_separated`** — same seed is deterministic; mainnet (coin type 5) and testnet (1) derive different keys; devnet shares the testnet coin type; accounts are isolated; ZIP-32 bounds (u31 account index, 32..=252-byte seed) are enforced.
- **`from_seed_known_answer_pin`** — pins the full derivation output for a fixed seed (`0x42`×64, testnet, account 0) so any future drift is caught explicitly. Must not be updated without a migration story.
- **ZIP-32 seed-length guard** in `from_seed` (32..=252 bytes). The spec requires it but neither the orchard fork nor the `zip32` crate enforces it. Unreachable for 64-byte BIP-39 callers, but the bound is a MUST and the guard makes the doc comment accurate.

No behavioral change for real callers; no consensus impact.

## How Has This Been Tested?

`cargo test -p platform-wallet --features shielded shielded::keys::tests` — all 3 pass on the `v4.1-dev` base. `cargo fmt --check` clean. The vector test additionally passes against the orchard fork's own official `keys`/`zip32` suites at the pinned `dashified-0.14.0` commit.

## Breaking Changes

None.

## Checklist

## For repository code-owners and reviewers only

### DIFF
--- a/packages/rs-platform-wallet/src/wallet/shielded/keys.rs
+++ b/packages/rs-platform-wallet/src/wallet/shielded/keys.rs
@@ -54,17 +54,27 @@ impl OrchardKeySet {
     /// Derive the full Orchard key set from a wallet seed.
     ///
     /// The `seed` should be the BIP-39 seed bytes (typically 64 bytes).
-    /// `SpendingKey::from_zip32_seed` accepts seeds of 32-252 bytes.
+    /// ZIP-32 requires master seeds of 32-252 bytes; the underlying
+    /// `SpendingKey::from_zip32_seed` does not enforce that bound
+    /// itself, so it is checked here.
     ///
     /// # Errors
     ///
-    /// Returns an error if the seed is invalid or the ZIP-32 derivation
-    /// fails (e.g. the derived key is the zero scalar).
+    /// Returns an error if the seed length is out of range or the
+    /// ZIP-32 derivation fails (e.g. the derived key is the zero
+    /// scalar).
     pub fn from_seed(
         seed: &[u8],
         network: Network,
         account: u32,
     ) -> Result<Self, PlatformWalletError> {
+        if seed.len() < 32 || seed.len() > 252 {
+            return Err(PlatformWalletError::ShieldedKeyDerivation(format!(
+                "seed must be 32..=252 bytes per ZIP-32, got {}",
+                seed.len()
+            )));
+        }
+
         let coin_type = match network {
             Network::Mainnet => DASH_COIN_TYPE_MAINNET,
             _ => DASH_COIN_TYPE_TESTNET,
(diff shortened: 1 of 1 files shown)


thepastaclaw:
✅ Review complete (commit 8231ea37)

packages/rs-platform-wallet/src/wallet/shielded/keys.rs
```diff
@@ -157,3 +167,153 @@ pub struct AccountViewingKeys {
...
+        let ovk = fvk.to_ovk(Scope::External);
+        assert_eq!(ovk.as_ref(), &TV_OVK, "outgoing viewing key mismatch");
+
+        // Raw address encoding is d ‖ pk_d; the "default" address in
+        // the vectors is diversifier index 0, which is what
+        // `OrchardKeySet::from_seed` exposes as `default_address`.
+        let raw = fvk.address_at(0u32, Scope::External).to_raw_address_bytes();
+        assert_eq!(&raw[..11], &TV_DEFAULT_D, "default diversifier mismatch");
+        assert_eq!(&raw[11..], &TV_DEFAULT_PK_D, "default pk_d mismatch");
+    }
```
thepastaclaw, line 233: **🟡 Suggestion: "Official ZIP-32 vector" test only exercises the SK→FVK pipeline, not seed→SK**

TV_SK is sourced from `orchard_key_components.py` (which pins SK→FVK/IVK/OVK/address), and this test constructs the SpendingKey directly via `SpendingKey::from_bytes(TV_SK)`. It never calls `SpendingKey::from_zip32_seed` or `OrchardKeySet::from_seed`, so the actual ZIP-32 `m/32'/coin_type'/account'` seed derivation is only guarded by the self-generated `from_seed_known_answer_pin` at line 300 — not by an official upstream vector. The commit message and module comment both advertise "official ZIP-32 vectors", which is stronger than what this test actually verifies. Consider adding a second test that starts from an `orchard_zip32.py` seed/path vector and compares the derived SK bytes (or the resulting FVK) so the seed→SK step is also pinned to upstream.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/shielded/keys.rs
```diff
@@ -157,3 +167,153 @@ pub struct AccountViewingKeys {
...
+        // ZIP-32 hardened child bounds: account indices are u31.
+        assert!(
+            OrchardKeySet::from_seed(&seed, Network::Testnet, 1 << 31).is_err(),
+            "account index ≥ 2^31 must be rejected"
+        );
+        // ZIP-32 master seed must be 32..=252 bytes.
+        assert!(
+            OrchardKeySet::from_seed(&[0u8; 16], Network::Testnet, 0).is_err(),
+            "16-byte seed must be rejected"
+        );
```
thepastaclaw, line 289: **💬 Nitpick: Seed upper-bound (>252 bytes) branch of the guard is untested**

The new guard at line 71 is two-sided (`seed.len() < 32 || seed.len() > 252`), but only the lower branch is exercised (`[0u8; 16]`). A future off-by-one on the upper branch (e.g. someone flipping `> 252` to `>= 252` or dropping it entirely) would not be caught. A single additional assertion with a 253-byte seed would close the guard's coverage symmetrically.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Source: reviewers = claude general `opus`; claude security-auditor `opus`; codex general `gpt-5.5` (parsed after rc=5); codex security-auditor `gpt-5.5` (parsed after rc=5); verifier = claude `opus`; specialists = security-auditor; policy gate = review_policy.enforce_backport_prereq_policy.

Well-scoped defensive PR adding a ZIP-32 seed-length guard and regression tests for Orchard key derivation. Verified against the file: the seed-length guard is spec-correct (32..=252), the SK→FVK/IVK/OVK/address pipeline test matches upstream orchard_key_components.py vector 0, and the known-answer pin locks the full from_seed path. Only test-coverage suggestions, no blocking issues.

<details>

```

In `packages/rs-platform-wallet/src/wallet/shielded/keys.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/keys.rs:213-233: "Official ZIP-32 vector" test only exercises the SK→FVK pipeline, not seed→SK
  TV_SK is sourced from `orchard_key_components.py` (which pins SK→FVK/IVK/OVK/address), and this test constructs the SpendingKey directly via `SpendingKey::from_bytes(TV_SK)`. It never calls `SpendingKey::from_zip32_seed` or `OrchardKeySet::from_seed`, so the actual ZIP-32 `m/32'/coin_type'/account'` seed derivation is only guarded by the self-generated `from_seed_known_answer_pin` at line 300 — not by an official upstream vector. The commit message and module comment both advertise "official ZIP-32 vectors", which is stronger than what this test actually verifies. Consider adding a second test that starts from an `orchard_zip32.py` seed/path vector and compares the derived SK bytes (or the resulting FVK) so the seed→SK step is also pinned to upstream.
```
</details>

QuantumExplorer (commented):
Reviewed

PR #4037, opened 2026-07-07
fix(swift-sdk): updated the swift integration tests to use the new tx broadcast flow

ZocoLini:
## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added or updated relevant unit/integration/functional/e2e tests
- [ ] I have added "!" to the title and described breaking changes in the corresponding section if my code contains any
- [ ] I have made corresponding changes to the documentation if needed

**For repository code-owners and collaborators only**
- [ ] I have assigned this pull request to a milestone

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved coverage for wallet-to-wallet sends, self-sends, and restart scenarios to better verify transaction handling and balance updates.

* **Tests**
  * Updated integration tests to use a simpler send flow for single-recipient transfers.
  * Added support for building and broadcasting a single-output transaction in test wallet utilities.

PR #3970, opened 2026-06-29
refactor(platform-wallet): expose the new core TransactionBuilder API

ZocoLini:
- Core transaction builder exposed over FFI
  - set_gap_limit for any account type (focus: CoinJoin)
  - Swift wrappers: CoreTransactionBuilder (fluent API), CoreTransaction, ManagedCoreWallet.broadcastTransaction/setGapLimit, and migration of SendViewModel (core→core) to the new build→broadcast flow
 
## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added or updated relevant unit/integration/functional/e2e tests
- [x] I have added "!" to the title and described breaking changes in the corresponding section if my code contains any
- [x] I have made corresponding changes to the documentation if needed

**For repository code-owners and collaborators only**
- [ ] I have assigned this pull request to a milestone

## Summary by CodeRabbit

* **New Features**
  * Added a fluent builder for creating and signing core wallet transactions prior to broadcasting.
  * Added support for increasing an account’s address gap limit.
  * Broadcasting now operates on the signed transaction object to preserve build-time account context.
* **Bug Fixes**
  * Updated core-to-core send flow to use the new build-then-broadcast flow for single and multi-recipient batches.
  * Improved reservation handling so failed broadcasts can be retried more reliably with clearer keep/release behavior.

PR #3853, opened 2026-06-11
fix(swift-sdk): drop legacy headers pre-processing in build_ios.sh

ZocoLini:
The build_ios.sh script was modifying the headers in place for compatibility after the rewrite I made months ago, with this PR I remove that logic to use the generated headers the way they were generated

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added or updated relevant unit/integration/functional/e2e tests
- [ ] I have added "!" to the title and described breaking changes in the corresponding section if my code contains any
- [x] I have made corresponding changes to the documentation if needed

**For repository code-owners and collaborators only**
- [ ] I have assigned this pull request to a milestone

## Summary by CodeRabbit

* **Refactor**
  * Internal improvements to pointer handling across the Swift SDK for enhanced compatibility with FFI (Foreign Function Interface) layer.
  * Optimized code structure in SDK components without affecting visible functionality.

PR #4038, opened 2026-07-07
feat(swift-example-app): bind every wallet's shielded sub-wallet for multi-wallet flows

QuantumExplorer:

The iOS SwiftExampleApp only bound the **lexicographically-smallest** wallet's shielded (Orchard) sub-wallet to the Rust sync coordinator. With multiple wallets loaded, every non-first wallet's shielded pool was invisible: no receive address, no balance, no note sync. This blocked the multi-wallet shielded QA items **SH-14** (A→B shielded transfer between two on-device wallets), **SH-15** (unshield A→B platform address), and **SH-16** (shielded withdraw A→B core address).

This is **app-side only** — the Rust `NetworkShieldedCoordinator` was already fully multi-wallet (one shared per-network commitment tree; each sync pass trial-decrypts every block against the union of all registered viewing keys and routes hits to each wallet's own persister). The single-bind limit was purely a UI-service choice.

## What was done?

Design: **"single UI mirror + multi-engine-bind"**.

- `ShieldedService` still mirrors exactly **one** wallet (`firstWallet`) for the global Sync-status surface via `bind(...)`.
- New `ShieldedService.bindEngine(walletManager:walletId:network:resolver:accounts:)` does **engine-only registration** (`configureShielded` idempotent + `bindShielded`), no mirror repoint, best-effort (catches + logs, never throws).
- `rebindWalletScopedServices()` now engine-binds every **other** loaded wallet through a pure, unit-testable seam `engineBindOtherWallets(allWalletIds:mirrorWalletId:bindEngine:)`.
- Per-wallet surfaces read the **engine / SwiftData**, not the singleton mirror:
  - **Receive→Shielded address** → `walletManager.shieldedDefaultAddress(walletId:account:0)` + `DashAddress.encodeOrchard` (`ReceiveAddressView`, `AccountListView`, `SendViewModel` self-shield guard).
  - **Per-wallet shielded balance** → sum of unspent `PersistentShieldedNote` scoped by `walletId` (`BalanceCardView`, `SendTransactionView`, `CreateIdentityView`).
  - `CreateIdentityView.shieldedOptionAvailable(for:)` relaxed from "mirror-bound" to "engine-bound + own pool > 0 + fallback address".
  - The global Sync tab (`CoreContentView`) stays mirror-backed (unchanged).

**Eager vs. lazy binding:** chose **eager** (bind all wallets at rebind time). Each wallet's mnemonic is stored `kSecAttrAccessibleWhenUnlockedThisDeviceOnly` (device-unlock only; `.biometryCurrentSet` guards only the separate opt-in biometric seed item), so N reads at startup are fast keychain reads with **no biometric/passcode prompts**. This matches the Android eager-bind pattern to be ported next, and makes cross-wallet flows work with no wallet-swap/rebind.

**Clear (SH-12) scope:** `clearShielded` drops all wallets from the coordinator (global wipe, unchanged). Only the mirror re-binds on "Sync Now"; the other wallets re-bind on the next `rebindWalletScopedServices` fire. This is deliberate — an immediate re-bind + sync restart would repopulate before the tester can observe SH-12's reset. Documented inline.

The change is portable: the same shape will be ported to Android (`KotlinExampleApp` `AppContainer.rebindWalletScopedServices` + `ShieldedService` in `packages/kotlin-sdk`).

## How Has This Been Tested?

- **Build:** `build_ios.sh --target sim` (features = `shielded`) then the app compiles clean — **zero errors, zero Swift 6 strict-concurrency warnings**.
- **Unit tests:** all 4 `ShieldedEngineBindPlanTests` pass — binds every non-mirror wallet, one failing bind doesn't block the rest, single-wallet no-op, mirror skipped by identity regardless of position.
- **On-simulator (iPhone 17, testnet, two fresh wallets A `1211CF…` / B `357386…`):**
  1. Each wallet's Receive→Shielded shows its **own distinct** `tdash1…` Orchard address (A `…w5af7z…`, B `…qjeqvfhl…`). Under the old code both would have shown wallet A's mirror address — `shieldedDefaultAddress(walletId:)` returns non-nil only for a wallet actually registered in the coordinator.
  2. `ZPERSISTENTSHIELDEDSYNCSTATE` has **one row per wallet** — a single sync pass processed both registrations (the `walletCount == 2` behavior). No crashes.
- **Not run:** the full SH-14 money movement (A's pool → B's address) requires funded pools and a multi-hour cold shielded sync on testnet; the structural engine-bind (the actual code change) is what's proven here. TEST_PLAN SH-14/15/16 notes updated to reflect automatic binding of both wallets.

## Breaking Changes

None.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Improved multi-wallet shielded support, including per-wallet receive addresses, balances, and send funding from the correct wallet.
  * “Sync Now” now refreshes shielded data across all loaded wallets more reliably.

* **Bug Fixes**
  * Fixed shielded balance displays so they reflect each wallet’s own funds instead of a single mirrored wallet.
  * Recipient validation now checks against the correct shielded address for the selected wallet.
  * Clear-and-resync flows now restore shielded wallet data more consistently after state resets.


thepastaclaw:
✅ Review complete (commit f9465093)

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/SwiftExampleAppApp.swift
```diff
@@ -263,6 +263,32 @@ struct SwiftExampleAppApp: App {
...
+                allWalletIds: walletManager.wallets.keys,
+                mirrorWalletId: wallet.walletId
+            ) { otherWalletId in
+                shieldedService.bindEngine(
+                    walletManager: walletManager,
+                    walletId: otherWalletId,
+                    network: platformState.currentNetwork,
+                    resolver: shieldedResolver
+                )
+            }
```
thepastaclaw: **🔴 Blocking: Register all shielded wallets before starting the sync loop**

`startShieldedSync()` starts a Rust background thread whose first iteration calls `sync_now(false)` before sleeping. The coordinator snapshots the currently registered shielded accounts at the start of that pass, and a no-activity pass sets the coordinator-wide caught-up cooldown. Since this PR registers the non-mirror wallets only after starting the loop, a cold-start first pass can run with just `firstWallet`, emit results only for that wallet, and leave the newly added wallets waiting for a later tick/manual sync. Move the `engineBindOtherWallets` block before `startShieldedSync()` so the first shared pass sees the complete loaded wallet set.

```suggestion
            // Engine-bind every OTHER loaded wallet into the shared
            // network-scoped shielded coordinator. `firstWallet` above
            // already drives the UI mirror AND its own engine
            // registration via `bind(...)`; this loop registers the
            // remaining wallets so a single shielded sync pass
            // trial-decrypts against the union of every wallet's viewing
            // keys (SH-14/15/16 cross-wallet flows). Each bind is
            // best-effort + independent — one wallet's missing mnemonic
            // must not block the others. Reading each mnemonic is a
            // device-unlock-only keychain read (no biometric prompt), so
            // eager binding at startup is safe. The iteration seam is a
            // pure free function (`engineBindOtherWallets`) so its
            // "visit every non-mirror wallet" contract can be
            // unit-tested without a configured manager.
            engineBindOtherWallets(
                allWalletIds: walletManager.wallets.keys,
                mirrorWalletId: wallet.walletId
            ) { otherWalletId in
                shieldedService.bindEngine(
                    walletManager: walletManager,
                    walletId: otherWalletId,
                    network: platformState.currentNetwork,
                    resolver: shieldedResolver
                )
            }
            if try !walletManager.isShieldedSyncRunning() {
                try walletManager.startShieldedSync()
            }

            // DashPay contact-request + profile sweep (background
            // loop). Wallet-driven — every registered wallet is swept
            // each pass — so manager scope is the right place to start
            // it, same as the address / shielded loops above.
            // Idempotent: starting while running is a no-op.
            if try !walletManager.isDashPaySyncRunning() {
                try walletManager.startDashPaySync()
            }
```

<sub>source: ['codex-general', 'codex-ffi-engineer', 'codex-verifier']</sub>

thepastaclaw (changes requested):
## Code Review

Source: reviewers codex-general/codex-ffi-engineer completed; claude-general/claude-ffi-engineer failed due Claude extra-usage quota; verifier recovered with codex after Claude verifier quota failure.

The reported ordering bug is valid. This PR adds automatic engine binding for every non-mirror wallet, but the app starts the shielded sync loop before those wallets are registered, so the loop's immediate first pass can snapshot only the mirror wallet and delay the rest behind the next cadence/cooldown window.

🔴 1 blocking

<details>

```

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/SwiftExampleAppApp.swift`:
- [BLOCKING] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/SwiftExampleAppApp.swift:254-291: Register all shielded wallets before starting the sync loop
  `startShieldedSync()` starts a Rust background thread whose first iteration calls `sync_now(false)` before sleeping. The coordinator snapshots the currently registered shielded accounts at the start of that pass, and a no-activity pass sets the coordinator-wide caught-up cooldown. Since this PR registers the non-mirror wallets only after starting the loop, a cold-start first pass can run with just `firstWallet`, emit results only for that wallet, and leave the newly added wallets waiting for a later tick/manual sync. Move the `engineBindOtherWallets` block before `startShieldedSync()` so the first shared pass sees the complete loaded wallet set.
```
</details>

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/CreateIdentityView.swift
```diff
@@ -1740,20 +1741,51 @@ struct CreateIdentityView: View {
...
+    /// rather than the single-mirror `shieldedService.shieldedBalance`.
+    /// Correct for ANY loaded wallet, not just the mirror's
+    /// `firstWallet`.
+    private func shieldedPoolBalance(for walletId: Data) -> UInt64 {
+        let descriptor = FetchDescriptor<PersistentShieldedNote>(
+            predicate: #Predicate { $0.walletId == walletId && $0.isSpent == false }
+        )
+        let notes = (try? modelContext.fetch(descriptor)) ?? []
+        return notes.reduce(0) { $0 + $1.value }
+    }
```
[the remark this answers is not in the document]
QuantumExplorer: Fixed in cce68d0110. `CreateIdentityView` now holds `@Query(filter: PersistentShieldedNote.unspentPredicate) private var unspentShieldedNotes` (all-wallets predicate — the funding picker spans multiple wallets, so a single-walletId query could not serve it) and `shieldedPoolBalance(for:)` filters that live array by walletId in memory. Note rows arriving while the sheet is open now re-render the funding option, denomination list, and footer balance; the per-call-site `FetchDescriptor` round-trips are gone too. The predicate itself moved to `PersistentShieldedNote.unspentPredicate` so every balance surface shares one definition of "unspent".

QuantumExplorer:
**Review-fix round (cce68d0110)** — applied five self-review findings plus the CodeRabbit comment:

1. **Post-Clear dark window (confirmed on-simulator):** "Sync Now" now unconditionally re-registers every loaded wallet. `clearShielded` drops all coordinator registrations, and a detail-view `switchTo` between Clear and Sync Now re-binds only the viewed wallet — the recovery branch alone left the others unscanned until app restart. Verified: Clear → visit wallet A → Sync Now restores both wallets' `ZPERSISTENTSHIELDEDSYNCSTATE` rows.
2. **Rebind hardening:** the engine-bind loop moved ahead of the fallible shielded/DashPay start calls in `rebindWalletScopedServices()`.
3. **Dedup:** `PlatformWalletManager.shieldedDisplayAddress(walletId:account:network:)` replaces the 3× resolve+encode pattern; `PersistentShieldedNote.unspentPredicate` replaces the 4× unspent-note filter (+ a unit test).
4. **Reactive pool balance** in `CreateIdentityView` via `@Query` (CodeRabbit thread).
5. **Deliberately NO "already bound" fast path in `bindEngine`:** the only cheap probe (`shieldedDefaultAddress`) reflects the wallet-level sub-wallet binding, which *survives* `clearShielded` — an earlier draft of this round used it and on-simulator testing showed it silently left post-Clear wallets unregistered. Documented inline.

**CI failure root cause (`wallet_manager_import_wallet_from_bytes` signature mismatch):** not a code drift — our pinned rust-dashcore rev (`647fa982`) has the 5-arg signature and builds clean from scratch. The FFI crates' `build.rs` write cbindgen headers to the shared `target/<triple>/<profile>/include/` path but only re-execute on fingerprint change; on a runner that preserves `target/` across branches, a branch building a newer rust-dashcore (master now has `birth_height` in the import FFI, #693) leaves its header behind, and a later run of this branch pairs old Swift with the new header. `build_ios.sh` now `cargo clean`s the four bundled-header crates per target before building, forcing cbindgen to regenerate against this checkout's pins. Verified locally: the clean pass removed the stale artifacts and the full sim build + tests are green.

QuantumExplorer:
**CI update:** the header fix in cce68d0110 worked — the `wallet_manager_import_wallet_from_bytes` phantom mismatch is gone (the run's log shows the header-crate clean executing for all three targets). The run then surfaced a **second, pre-existing breakage at the v4.1-dev base**: `SwiftDashSDKIntegrationTests` still called `ManagedCoreWallet.sendToAddresses`, which #3970 removed in favor of `CoreTransactionBuilder` — #3712's tests and #3970's refactor never met in CI until this PR ran the Swift job on their combination.

Fixed in 50a53a7e21: `TestWalletWrapper.sendToAddresses` reproduces the removed flow (build + sign + broadcast → serialized signed tx) on the canonical builder path, with the three call sites migrated. No other removed members are referenced by the tests.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/ShieldedService.swift
```diff
@@ -428,6 +501,28 @@ class ShieldedService: ObservableObject {
             guard isBound else { return }
```
thepastaclaw: **🔴 Blocking: Do not let mirror rebind failure skip other wallets after Clear**

After `clearShielded()` drops every coordinator registration, `manualSync()` is supposed to re-register every loaded wallet on the next Sync Now tap. The new non-mirror `engineBindOtherWallets` pass is below this `guard isBound else { return }`, so if the mirror wallet's `bind(...)` fails because its mnemonic is missing, declined, or corrupt, the method returns before attempting to bind any other loaded wallet and before calling `syncShieldedNow()`. That contradicts the best-effort multi-wallet recovery contract documented in this file: one wallet's bind failure must not prevent other wallets from being re-registered and synced. Run the non-mirror bind pass independently of the mirror bind result, then skip the sync only if no wallet could be registered.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):
## Code Review

Carried-forward prior findings: the previous startup ordering blocker is fixed; current head binds non-mirror shielded wallets at `SwiftExampleAppApp.swift` lines 275-285 before `startShieldedSync()` at lines 287-288. New findings in the latest delta: the post-Clear `manualSync()` recovery path still exits before the new non-mirror rebind-all pass when the mirror wallet rebind fails, so one mirror-wallet credential failure can prevent otherwise recoverable wallets from syncing.

🔴 1 blocking

<details>

```

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/ShieldedService.swift`:
- [BLOCKING] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/ShieldedService.swift:497-501: Do not let mirror rebind failure skip other wallets after Clear
  After `clearShielded()` drops every coordinator registration, `manualSync()` is supposed to re-register every loaded wallet on the next Sync Now tap. The new non-mirror `engineBindOtherWallets` pass is below this `guard isBound else { return }`, so if the mirror wallet's `bind(...)` fails because its mnemonic is missing, declined, or corrupt, the method returns before attempting to bind any other loaded wallet and before calling `syncShieldedNow()`. That contradicts the best-effort multi-wallet recovery contract documented in this file: one wallet's bind failure must not prevent other wallets from being re-registered and synced. Run the non-mirror bind pass independently of the mirror bind result, then skip the sync only if no wallet could be registered.
```
</details>

Source: reviewers: `opus` (general, failed_or_unparseable), `claude-sonnet-5` (general, failed_or_unparseable), `gpt-5.5` (general), `opus` (ffi-engineer, failed_or_unparseable), `claude-sonnet-5` (ffi-engineer, failed_or_unparseable), `gpt-5.5` (ffi-engineer); verifier: `gpt-5.5`; specialists: `ffi-engineer`

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/SwiftExampleAppApp.swift
thepastaclaw: Resolved in `cce68d0` — _Register all shielded wallets before starting the sync loop_ no longer present.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/ShieldedService.swift
QuantumExplorer: Confirmed real and fixed in 7b01abafff. The mirror bind in `manualSync` is now best-effort like everything else: on failure it no longer bails (its `lastError` stays visible), the engine-bind pass always runs so other wallets with intact mnemonics re-register, and `bindEngine` now reports success so the pass can tell whether ANYTHING registered — the sync is skipped only in the nothing-registered case, preserving the original "do not chain a sync that will fail the same way" intent per-wallet. Regression-checked on-simulator: Clear → visit wallet A → Sync Now still restores both wallets' sync-state rows.

thepastaclaw (changes requested):
## Code Review

Source: reviewers claude-general (opus, failed: extra-usage quota), codex-general (gpt-5.5); verifier codex (gpt-5.5, recovered from opus verifier failure); specialists: none.

Carried-forward prior findings: the prior `cce68d01` blocker is STILL VALID at `50a53a7e`. `manualSync()` still returns after a failed mirror rebind before running the non-mirror `engineBindOtherWallets` pass, so the post-Clear recovery path can skip otherwise recoverable wallets.

New findings in latest delta: none. The latest delta only migrates integration tests to the `TestWalletWrapper.sendToAddresses` helper. The earlier startup-ordering issue from `ac684c9d` remains fixed.

🔴 1 blocking

<details>

```

- [BLOCKING] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/ShieldedService.swift:497-501: Do not let mirror rebind failure skip other wallets after Clear
  After `clearShielded()` drops every wallet registration from the Rust coordinator, `manualSync()` is documented as the recovery path that re-registers every loaded wallet on the next Sync Now tap. The mirror-wallet recovery branch still exits at `guard isBound else { return }` when `bind(...)` fails, so the non-mirror `engineBindOtherWallets` pass below never runs. That means one missing, declined, or corrupt mirror-wallet mnemonic can keep otherwise valid loaded wallets unregistered after Clear and prevent `syncShieldedNow()` from recovering them. Run the non-mirror bind pass independently of the mirror bind result, and only skip the sync when no wallet could be registered.
```
</details>

thepastaclaw (commented):
## Code Review

Source: reviewers claude-general (opus, failed: extra-usage quota), codex-general (gpt-5.5), claude-ffi-engineer (opus, failed: extra-usage quota), codex-ffi-engineer (gpt-5.5); verifier codex (gpt-5.5, recovered from opus verifier quota failure); specialists: ffi-engineer.

Carried-forward prior findings: prior-1 is FIXED at `7b01abaffff35c155ef69bab8f99e9032f5b1b59`. `manualSync()` no longer returns after a failed mirror bind before running the non-mirror engine-bind pass; it now runs `engineBindOtherWallets(...)`, tracks whether any wallet registered, and skips the sync only when no wallet could be registered.

New findings in latest delta: one suggestion remains in the same recovery path. When only non-mirror wallets bind after a mirror bind failure, the service clears the mirror bind error and opens manual sync timing even though its event handler will drop completion events while `isBound` is false.

1 suggestion

<details>

```

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/ShieldedService.swift`:
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/ShieldedService.swift:545-555: Only start mirror timing when the mirror wallet is bound
  When the mirror `bind(...)` fails but at least one non-mirror `bindEngine(...)` succeeds, `anyWalletRegistered` is true while `isBound` remains false. This path then clears the mirror bind error and opens the manual sync timing bracket, but `handleShieldedSyncEvent` drops the completion at `guard isBound else { return }`, so `endSyncTiming()` is never reached for this pass. Keep syncing the successfully registered non-mirror wallets, but only clear the mirror error and start mirror timing when the mirror wallet is actually bound; otherwise close any stale timing state after the awaited sync returns.
```
</details>

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/ShieldedService.swift
thepastaclaw: Resolved in `7b01aba` — _Do not let mirror rebind failure skip other wallets after Clear_ no longer present.

thepastaclaw (commented):
## Code Review

Source: reviewers: `opus` (general, failed: extra-usage quota), `gpt-5.5` (general), `opus` (ffi-engineer, failed: extra-usage quota), `gpt-5.5` (ffi-engineer); verifier: `gpt-5.5` (recovered from Opus verifier quota failure); specialists: `ffi-engineer`.

Prior findings reconciliation: prior-1 from `7b01abaf` is **STILL VALID** as the same mirror-timing root cause, now through the manager-wide sync-state subscription. The latest delta fixes the direct `manualSync()` timing start, but the subscription can still open mirror timing while the mirror wallet is unbound.

Carried-forward prior findings: 1 suggestion, re-anchored to `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/ShieldedService.swift:316-317`.

New findings in latest delta: none.

🟡 1 suggestion

<details>

```

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/ShieldedService.swift`:
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/ShieldedService.swift:316-317: Gate sync-state timing on the mirror being bound
  The latest change gates the `manualSync()` fast-path timing start on `isBound`, but this subscription still starts the mirror timing bracket on any manager-wide sync start. That remains reachable after a mirror bind failure because `bind(...)` still installs the subscription, other wallets can be engine-bound, and the shared shielded loop can run for those wallets. Since `handleShieldedSyncEvent` returns immediately while `isBound` is false, the unbound mirror never closes the timing bracket for that background pass. Keep the manager-wide spinner behavior if desired, but only start mirror timing when the mirrored wallet is actually bound.
```
</details>

packages/swift-sdk/SwiftTests/SwiftDashSDKIntegrationTests/Support/TestWallet.swift
```diff
@@ -19,6 +19,35 @@ final class TestWalletWrapper {
...
     }
 
+    /// Build, sign, and broadcast a BIP44 payment from this wallet,
+    /// returning the serialized signed transaction — the flow the
+    /// removed `ManagedCoreWallet.sendToAddresses` performed before
+    /// #3970 moved core sends onto `CoreTransactionBuilder`. Kept as a
+    /// test-support shorthand so the integration tests read as
+    /// "send X to Y" rather than five builder steps.
+    func sendToAddresses(
+        recipients: [(address: String, amountDuffs: UInt64)],
```
ZocoLini: This is not needed, tell AI to rebase and remove all modifications to the Swift Integration tests

packages/swift-sdk/build_ios.sh
```diff
@@ -201,10 +201,36 @@ if [ "$PROFILE" = "dev-ios" ]; then
...
+# against the OTHER rev's header (seen as phantom FFI signature
+# mismatches in `WalletManager.swift` on CI). Cleaning the header
+# crates for the triple being built forces their build scripts to
+# re-run so headers always match this checkout's pins. Costs a
+# rebuild of the FFI crates only; the heavy dependency graph stays
+# cached. NOTE: `cargo clean -p` must be scoped with --target and
+# --profile — unscoped it only touches the host-default dirs and
+# leaves the cross-compiled artifacts (and the stale header) intact.
+clean_header_crates() {
+  local triple="$1"
```
ZocoLini: This shouldn't be needed, headers are always regenerated if you change the source code, if not, the issue is in the build.rs, for example, my PR https://github.com/dashpay/platform/pull/3853 modifies the headers output by removing the legacy header pre-processing we had, not changing the source code, thats why I have to force the regeneration temporally

ZocoLini (changes requested):
I don't see the build.rs you told me about but anyway, I left a couple of comments in the build_ios.sh and in the Integrations test files

packages/swift-sdk/SwiftTests/SwiftDashSDKIntegrationTests/Support/TestWallet.swift
QuantumExplorer: Done — rebased onto v4.1-dev (now includes your #4037) and dropped our integration-test commit entirely; `SwiftTests/` is pristine upstream in this PR. We had patched the same breakage independently before #4037 landed (the branch was the first to run the Swift CI on the #3712+#3970 combination).

packages/swift-sdk/build_ios.sh
QuantumExplorer: Removed in f9465093e8 — agreed this belongs at the build.rs level and #3853 owns that space. For that PR, the exact mechanism we hit (empirically, on mac-runner-1): the FFI crates' build.rs write cbindgen headers to the SHARED `target/<triple>/<profile>/include/<crate>/` path, and cargo caches artifacts for multiple git-dep revs side by side — so when branch A builds rust-dashcore rev X, its build.rs overwrites the shared header; when branch B (rev Y, fingerprint-fresh from an earlier build) builds next, cargo skips its build.rs entirely and the Swift compile pairs rev Y's code with rev X's header. That was the phantom `wallet_manager_import_wallet_from_bytes` 6-arg mismatch this PR's first CI run hit (master's birth_height signature from a different branch's build, while this branch pins 647fa982). "Headers regenerate when source changes" holds per-rev, but the shared output path + per-rev fingerprints breaks it across branch switches. A durable fix would be emitting to OUT_DIR (fingerprint-scoped) and collecting from there, or an unconditional cargo:rerun trigger for the header write. Until #3853 lands, runs on a shared runner can still hit this — worth keeping in mind if a red CI shows an FFI signature that matches a different rust-dashcore rev than Cargo.lock pins.

ZocoLini (approved):

thepastaclaw (commented):
## Code Review

Source: reviewers claude-general (opus, failed: extra-usage quota), codex-general (gpt-5.5), claude-ffi-engineer (opus, failed: extra-usage quota), codex-ffi-engineer (gpt-5.5); verifier codex (gpt-5.5, recovered from opus verifier failure); specialists: ffi-engineer.

Carried-forward prior findings: prior-1 is still valid; the manager-wide shielded sync-state subscription still opens the UI mirror timing bracket even when the mirrored wallet is not bound. New findings in latest delta: the latest commit also removed the `isBound` guard in `manualSync()`, so non-mirror-only syncs can clear the mirror bind error and leave stale timing state. No CodeRabbit inline findings were provided to validate.

🟡 2 suggestion(s)

<details>

```

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/ShieldedService.swift`:
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/ShieldedService.swift:316-322: Carried-forward prior: Gate sync timing on the mirror being bound
  The `$shieldedSyncIsSyncing` publisher is manager-wide, but this service mirrors one wallet. `bind(...)` installs this subscription even after `bindShielded` fails and leaves `isBound == false`; another wallet can still be engine-bound and drive the shared shielded sync loop. In that case this subscriber opens `currentSyncStartedAt`, while `handleShieldedSyncEvent` immediately returns at `guard isBound else { return }`, so the timing bracket is never closed for a sync pass that did not belong to the mirrored wallet.
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/ShieldedService.swift:545-561: New latest-delta: Do not mutate mirror timing for non-mirror syncs
  The latest delta removed the `isBound` guard around `manualSync()`'s mirror-only state updates. If the mirror bind fails but another loaded wallet binds successfully, `anyWalletRegistered` is true while this mirror remains unbound; `manualSync()` now clears the bind error and opens timing before running a sync whose completion event is dropped by `handleShieldedSyncEvent`. Keep the mirror error and timing changes scoped to a bound mirror, and close any unbound timing opened by the manager-wide publisher.
```
</details>

PR #4045, opened 2026-07-08
feat(kotlin-sdk): add maven-publish for the release AAR

bfoss765:
Adds `maven-publish` to the Kotlin SDK library so consumers can depend on a Maven coordinate instead of a GitHub-release AAR.

- Coordinates: `org.dashfoundation:dash-sdk-android` (version via `-PsdkVersion=x.y.z`, defaults to `0.1.0-SNAPSHOT`)
- Publishes the release AAR + sources jar + POM (Room/DataStore/Biometric `api` deps carried in the POM)
- `./gradlew :sdk:publishToMavenLocal` verified end-to-end: with `../build_android.sh` run first, the published AAR packages `libdash_sdk_jni.so` for arm64-v8a and x86_64, and the Dash Android wallet consumes it from mavenLocal alongside dashj with no class conflicts (see dashpay/dash-wallet#1507)
- A remote `repositories {}` block is intentionally left out pending a hosting decision (Maven Central vs GitHub Packages)

Context: Phase 0 of the Android wallet's dashj → Kotlin SDK migration.

## Summary by CodeRabbit

* **Chores**
  * Added publishing support for the Kotlin SDK, including versioning and package metadata.
  * Enabled generation of a sources archive and prepared release artifacts for distribution.

### DIFF
--- a/packages/kotlin-sdk/PUBLISHING.md
+++ b/packages/kotlin-sdk/PUBLISHING.md
@@ -1,0 +1,84 @@
+# Publishing `org.dashj:dash-sdk-android` to Maven Central
+
+The SDK publishes under the existing **`org.dashj`** namespace (same as
+`dashj-core`), so no new Central namespace verification is needed. This is a
+Maven-coordinates decision only — the Kotlin source packages remain
+`org.dashfoundation.dashsdk`. Deployment uses **JReleaser**, mirroring
+[dashj/core/build.gradle](https://github.com/dashpay/dashj/blob/master/core/build.gradle):
+release versions go to Maven Central via the Central Portal; `-SNAPSHOT`
+versions go to the Sonatype snapshots repository.
+
+## Prerequisites
+
+- **Sonatype Central Portal token** for the `org.dashj` namespace (held by
+  HashEngineering — the same account that publishes `dashj-core`).
+- **GPG signing key** (the dashj release key works), with the public key on the
+  keyservers.
+- **Native toolchain**: Android NDK r28+ (`28.1.13356709`), `cargo-ndk`, rustup
+  targets `aarch64-linux-android` and `x86_64-linux-android`, JDK 17.
+
+## Credentials — environment only, never committed
+
+Gradle signing (signs artifacts during the staging publish; the classic
+`signing.keyId` / `signing.password` / `signing.secretKeyRingFile` entries in
+`~/.gradle/gradle.properties`, as used for dashj-core, also work):
+
+```bash
+export ORG_GRADLE_PROJECT_signingKey="$(cat private-key.asc)"   # ASCII-armored
+export ORG_GRADLE_PROJECT_signingPassword="…"
+```
+
+JReleaser deploy (read only when a `jreleaser*` task runs):
+
+```bash
+export JRELEASER_MAVENCENTRAL_SONATYPE_USERNAME="…"   # Central Portal token name
+export JRELEASER_MAVENCENTRAL_SONATYPE_PASSWORD="…"   # Central Portal token value
+export JRELEASER_NEXUS2_SNAPSHOTS_USERNAME="…"        # same token (snapshots repo)
+export JRELEASER_NEXUS2_SNAPSHOTS_PASSWORD="…"
+export JRELEASER_GPG_PUBLIC_KEY="$(cat public-key.asc)"
+export JRELEASER_GPG_SECRET_KEY="$(cat private-key.asc)"
+export JRELEASER_GPG_PASSPHRASE="…"
+```
+
+## Release steps (from `packages/kotlin-sdk/`)
+
+```bash
+# 1. Build the native JNI library (arm64-v8a + x86_64) and verify its exports.
+./build_android.sh --verify
+
+# 2. Stage the signed release AAR + sources/javadoc jars into sdk/build/staging-deploy.
+./gradlew :sdk:publishReleasePublicationToStagingRepository -PsdkVersion=0.1.0
+
+# 3. Upload the staged artifacts. Release versions -> Maven Central;
+#    -SNAPSHOT versions -> Sonatype snapshots.
+./gradlew :sdk:jreleaserDeploy -PsdkVersion=0.1.0
+```
+
+Then approve/verify the deployment at <https://central.sonatype.com> and
+confirm the artifact appears on search.maven.org (sync takes ~30 min).
+
+**Versioning:** the default is `0.1.0-SNAPSHOT`; pass the same
+`-PsdkVersion=x.y.z` to steps 2 and 3. The first release should be a real
+version (e.g. `0.1.0`) — the Android wallet's CI must depend on a pinned
+release, never a snapshot.
+
+## The JNI guard
+
+`sdk/src/main/jniLibs/` is gitignored, so a clean checkout has no
+`libdash_sdk_jni.so` — and an AAR published without it installs fine but dies
+at runtime with `UnsatisfiedLinkError`. The `verifyJniLibsForRemotePublish`
+task therefore **hard-fails** every staging/remote/jreleaser publish task when
+the `.so` is missing for `arm64-v8a` or `x86_64`. There is no override for
+remote publishes; `-PallowMissingJni` only downgrades the check for
+`publishToMavenLocal` (metadata-only dry runs into `~/.m2`). If the guard
+fires, run step 1.
+
+## Consuming the artifact
+
+```kotlin
+repositories { mavenCentral() }
+dependencies { implementation("org.dashj:dash-sdk-android:0.1.0") }
+```
+
+Requires `minSdk 29`. The AAR bundles the native library for `arm64-v8a` and
+`x86_64`; `armeabi-v7a` is intentionally unsupported.
--- a/packages/kotlin-sdk/PUBLISHING.md
+++ b/packages/kotlin-sdk/PUBLISHING.md
@@ -48,6 +48,9 @@ export JRELEASER_GPG_PASSPHRASE="…"
 
 # 2. Stage the signed release AAR + sources/javadoc jars into sdk/build/staging-deploy.
 ./gradlew :sdk:publishReleasePublicationToStagingRepository -PsdkVersion=0.1.0
+# (the staging dir sdk/build/staging-deploy is wiped automatically before every
+#  staging publish — cleanStagingDeploy — so stale versions from earlier runs
+#  can never reach JReleaser)
 
 # 3. Upload the staged artifacts. Release versions -> Maven Central;
 #    -SNAPSHOT versions -> Sonatype snapshots.
--- a/packages/kotlin-sdk/sdk/build.gradle.kts
+++ b/packages/kotlin-sdk/sdk/build.gradle.kts
@@ -1,5 +1,7 @@
+import java.util.concurrent.Callable
 import org.gradle.api.publish.maven.tasks.PublishToMavenLocal
 import org.gradle.api.publish.maven.tasks.PublishToMavenRepository
+import org.jreleaser.model.Active
 
 plugins {
     alias(libs.plugins.android.library)
@@ -7,14 +9,20 @@ plugins {
     alias(libs.plugins.kotlin.serialization)
     alias(libs.plugins.ksp)
     `maven-publish`
+    id("signing")
+    alias(libs.plugins.jreleaser)
 }
 
-// NOTE (Maven Central, future): the `org.dashfoundation` groupId can't be
-// published to Maven Central yet — that requires owning the matching domain,
-// and dashfoundation.org is held by a third party (HashEngineering, PR #4045).
-// This coordinate is for `publishToMavenLocal` / an internal repo for now; a
-// Maven Central move is a separate groupId/hosting decision.
-group = "org.dashfoundation"
+// Published Maven coordinate group. Per HashEngineering's call on PR #4045:
+// publish under the existing `org.dashj` group — Dash already owns dashj.org and
+// ships its legacy Core/Platform SDKs there, so this needs no separate domain
+// verification for Maven Central (`org.dashfoundation` would have, since
+// dashfoundation.org is held by a third party). The Kotlin source packages stay
+// `org.dashfoundation.dashsdk`; this is a Maven-coordinates decision only.
+// `maven-publish` handles local / internal-repo publishing; jreleaser (bottom of
+// file) handles Maven Central releases and Sonatype SNAPSHOTs, mirroring
+// dashj/core/build.gradle.
+group = "org.dashj"
 version = project.findProperty("sdkVersion")?.toString() ?: "0.1.0-SNAPSHOT"
 
 android {
@@ -108,14 +116,25 @@ dependencies {
 
 // Publishes the release AAR (incl. libdash_sdk_jni.so if ../build_android.sh ran first) with a
 // POM carrying the Room/DataStore/Biometric api dependencies.
-// Local: ./gradlew :sdk:publishToMavenLocal   Coordinates: org.dashfoundation:dash-sdk-android:<version>
+// Local:  ./gradlew :sdk:publishToMavenLocal   Coordinates: org.dashj:dash-sdk-android:<version>
+// Remote: `:sdk:publishReleasePublicationToStagingRepository` stages the signed
+//         artifacts into build/staging-deploy, then `:sdk:jreleaserDeploy` uploads
+//         them to Maven Central (release) / Sonatype snapshots (see jreleaser block).
 // Override the version with -PsdkVersion=x.y.z (defaults to 0.1.0-SNAPSHOT).
 afterEvaluate {
     publishing {
+        repositories {
+            // Local staging dir that jreleaser reads from and uploads (below); it is
+            // not itself a network repo, so staging never needs credentials.
+            maven {
+                name = "staging"
+                url = uri(layout.buildDirectory.dir("staging-deploy"))
+            }
+        }
         publications {
             create<MavenPublication>("release") {
                 from(components["release"])
-                groupId = "org.dashfoundation"
+                groupId = "org.dashj"
                 artifactId = "dash-sdk-android"
                 pom {
                     name.set("Dash Platform Kotlin SDK")
@@ -131,9 +150,7 @@ afterEvaluate {
                         }
                     }
                     // scm + developers: required by Maven Central (and most
-                    // remote repos) for POM validation. Pointed at the GitHub
-                    // repo rather than a vanity domain (see the group-id note
-                    // above re: dashfoundation.org ownership).
+                    // remote repos) for POM validation. Pointed at the GitHub repo.
                     scm {
                         url.set("https://github.com/dashpay/platform")
                         connection.set("scm:git:git://github.com/dashpay/platform.git")
@@ -151,6 +168,18 @@ afterEvaluate {
                 }
             }
         }
+    }
+
+    // Sign published artifacts, but only for a real remote publish. `publishToMavenLocal`
+    // and a bare `assemble` never put a PublishToMavenRepository task in the graph, so a
+    // local build requires no GPG key. Keys come from the standard signing properties/env
+    // (signing.keyId + signing.password + signing.secretKeyRingFile, or the in-memory
+    // ORG_GRADLE_PROJECT_signingKey / signingPassword pair) when a remote publish runs.
+    signing {
+        setRequired(Callable {
+            gradle.taskGraph.allTasks.any { it is PublishToMavenRepository }
+        })
+        sign(publishing.publications["release"])
     }
 }
 
@@ -183,3 +212,63 @@ val verifyJniLibsPresent = tasks.register("verifyJniLibsPresent") {
 
 tasks.withType<PublishToMavenRepository>().configureEach { dependsOn(verifyJniLibsPresent) }
 tasks.withType<PublishToMavenLocal>().configureEach { dependsOn(verifyJniLibsPresent) }
+
+// Maven Central / Sonatype deployment via jreleaser, mirroring
+// dashj/core/build.gradle (https://github.com/dashpay/dashj/blob/master/core/build.gradle#L139).
+// Flow: `maven-publish` stages the signed artifacts into build/staging-deploy (the
+// `staging` repository configured above), then `./gradlew :sdk:jreleaserDeploy`
+// uploads them — release versions to Maven Central, -SNAPSHOT versions to the
+// Sonatype snapshots repo. jreleaser reads its GPG key and Sonatype credentials
+// from env/props (JRELEASER_GPG_*, JRELEASER_MAVENCENTRAL_*/JRELEASER_NEXUS2_*)
+// at deploy time only, so `publishToMavenLocal` and any build that never invokes a
+// jreleaser task need no signing key or account.
+jreleaser {
+    project {
+        name.set("dash-sdk-android")
+        description.set(
+            "Kotlin SDK for Dash Core (L1 SPV) and Dash Platform " +
+                "(identities, DPNS, DashPay, shielded balances)"
+        )
+        links {
+            homepage.set("https://github.com/dashpay/platform")
+        }
+        authors.set(listOf("Dash Platform Contributors"))
+        license.set("MIT")
+        gitRootSearch.set(true)
+    }
+
+    signing {
+        active.set(Active.ALWAYS)
+        armored.set(true)
+    }
+
+    deploy {
+        maven {
+            // Tagged release versions -> Maven Central via the Sonatype portal.
+            mavenCentral {
+                create("sonatype") {
+                    active.set(Active.RELEASE)
+                    url.set("https://central.sonatype.com/api/v1/publisher")
+                    stagingRepository(
+                        layout.buildDirectory.dir("staging-deploy").get().asFile.path
+                    )
+                }
+            }
+            // -SNAPSHOT versions -> Sonatype snapshots repository.
+            nexus2 {
+                create("snapshots") {
+                    active.set(Active.SNAPSHOT)
+                    url.set("https://central.sonatype.com/repository/maven-snapshots/")
+                    snapshotUrl.set("https://central.sonatype.com/repository/maven-snapshots/")
+                    applyMavenCentralRules.set(true)
+                    snapshotSupported.set(true)
+                    closeRepository.set(true)
+                    releaseRepository.set(true)
+                    stagingRepository(
+                        layout.buildDirectory.dir("staging-deploy").get().asFile.path
+                    )
+                }
+            }
+        }
+    }
+}
--- a/packages/kotlin-sdk/sdk/build.gradle.kts
+++ b/packages/kotlin-sdk/sdk/build.gradle.kts
@@ -137,6 +137,19 @@ afterEvaluate {
                 url = uri(layout.buildDirectory.dir("staging-deploy"))
             }
         }
+
+        // The staging dir is reused across runs and maven-publish never removes
+        // other versions' directories — a release deploy after an earlier
+        // snapshot/release staging run would hand JReleaser stale coordinates.
+        // Wipe it before every staging publish so each deploy starts clean.
+        val cleanStagingDeploy = tasks.register<Delete>("cleanStagingDeploy") {
+            delete(layout.buildDirectory.dir("staging-deploy"))
+        }
+        tasks.withType<PublishToMavenRepository>().configureEach {
+            if (name.endsWith("ToStagingRepository")) {
+                dependsOn(cleanStagingDeploy)
+            }
+        }
         publications {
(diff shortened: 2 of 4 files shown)


thepastaclaw:
🔍 Review in progress — actively reviewing now (commit ffeeb829)

thepastaclaw:
@coderabbitai review

packages/kotlin-sdk/sdk/build.gradle.kts
```diff
@@ -83,3 +93,33 @@ dependencies {
...
+
+// Publishes the release AAR (incl. libdash_sdk_jni.so if ../build_android.sh ran first) with a
+// POM carrying the Room/DataStore/Biometric api dependencies.
+// Local: ./gradlew :sdk:publishToMavenLocal   Coordinates: org.dashfoundation:dash-sdk-android:<version>
+// Override the version with -PsdkVersion=x.y.z (defaults to 0.1.0-SNAPSHOT).
+afterEvaluate {
+    publishing {
+        publications {
+            create<MavenPublication>("release") {
+                from(components["release"])
```
thepastaclaw, line 155: **🔴 Blocking: Publishing can create an AAR without the JNI library**

The release publication is created directly from the Android release component, but there is no Gradle task dependency or validation that builds `sdk/src/main/jniLibs/<abi>/libdash_sdk_jni.so` first. That directory is gitignored and absent in a clean checkout; the repo docs and build script show it is produced only by `packages/kotlin-sdk/build_android.sh`. Because `NativeLoader.ensureLoaded()` calls `System.loadLibrary("dash_sdk_jni")`, publishing from a clean tree can produce a Maven coordinate that installs successfully but fails at runtime with `UnsatisfiedLinkError`. The publish path should either depend on the native build or fail fast when the expected ABI libraries are missing.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):
## Code Review

The PR adds Maven publication for the Kotlin SDK release AAR, but the resulting coordinate is not reliably consumable from a clean checkout. The publication can succeed without packaging the required JNI library, and the published compile API omits coroutines even though public SDK types expose Flow/StateFlow.

Source: reviewers: opus/claude general failed (quota), gpt-5.5/codex general completed; verifier: gpt-5.5/codex; specialists: none.

🔴 2 blocking

<details>

```

In `packages/kotlin-sdk/sdk/build.gradle.kts`:
- [BLOCKING] packages/kotlin-sdk/sdk/build.gradle.kts:104-105: Publishing can create an AAR without the JNI library
  The release publication is created directly from the Android release component, but there is no Gradle task dependency or validation that builds `sdk/src/main/jniLibs/<abi>/libdash_sdk_jni.so` first. That directory is gitignored and absent in a clean checkout; the repo docs and build script show it is produced only by `packages/kotlin-sdk/build_android.sh`. Because `NativeLoader.ensureLoaded()` calls `System.loadLibrary("dash_sdk_jni")`, publishing from a clean tree can produce a Maven coordinate that installs successfully but fails at runtime with `UnsatisfiedLinkError`. The publish path should either depend on the native build or fail fast when the expected ABI libraries are missing.
- [BLOCKING] packages/kotlin-sdk/sdk/build.gradle.kts:72: Coroutines core is missing from the published compile API
  `kotlinx-coroutines-core` is declared as an `implementation` dependency, so the published component does not expose it on consumers' compile classpaths. The SDK's public API returns coroutines types such as `Flow`, `StateFlow`, and `SharedFlow` from classes including `DataManager`, `ShieldedService`, `WalletManagerStore`, and `PlatformWalletManager`. A consumer depending only on `org.dashfoundation:dash-sdk-android` can therefore fail to compile unless it independently declares coroutines. Since this PR makes the SDK consumable through a Maven coordinate, public API dependencies need to be published as `api` dependencies.
```
</details>

packages/kotlin-sdk/sdk/build.gradle.kts
HashEngineering: indeed, we will need to integrate building of the JNI libs into the gradle build system.  It can be done by running the build_android.sh script from the grade script or using CMake.  There could be other better ways.

packages/kotlin-sdk/sdk/build.gradle.kts
```diff
@@ -3,8 +3,12 @@ plugins {
     alias(libs.plugins.kotlin.android)
     alias(libs.plugins.kotlin.serialization)
     alias(libs.plugins.ksp)
+    `maven-publish`
 }
 
+group = "org.dashfoundation"
```
HashEngineering: one issue with this name is that http://dashfoundation.org is owned by someone else.  To get our artifacts published on Maven Central, we need to own the domain that goes with this group id.

We currently own dashj.org and are publishing there all of our legacy platform and core related SDK libraries.  This may not be ideal for this new SDK.

HashEngineering: This is good for now, but still requires a local build process to put the Kotlin SDK libraries into the local Maven Repository.  It does not allow for publishing to Maven Central, where everyone who wants to build an app that consumes the Kotlin SDK can access it without cloning this repo and building the Kotlin SDK.

Recently, on Android, we have had two outside contributors.

HashEngineering (commented):
comments left regarding future changes to support publishing to Maven Central.

thepastaclaw (commented):
## Code Review

## Carried-Forward Prior Findings
Both prior blocking findings are FIXED at head 3d05a4f6: coroutines-core is now published as an `api` dependency (build.gradle.kts:84), and a `verifyJniLibsPresent` task now gates both `PublishToMavenLocal` and `PublishToMavenRepository` tasks (build.gradle.kts:166-185), failing a clean-tree publish unless `-PallowMissingJni` is passed. Neither is carried forward as an open finding.

## New Findings In Latest Delta
The latest delta (b27971bc..3d05a4f6) only touched the coroutines-core line and added SCM/developer POM metadata plus the JNI guard task — no new issues found there. Codex additionally raised a cumulative (pre-existing, untouched-by-delta) concern that `kotlinx-serialization-json` remains `implementation` rather than `api`. Verified against source: the only public `@Serializable` type is `Sdk.MasternodeEntry`, but it is never returned by any public function (only consumed by `internal fun parseActiveMasternodes` and a `private` envelope class) — the actually-public `ActiveMasternode` return type isn't `@Serializable`. This is a real but narrow edge case (a consumer would only hit it by explicitly calling `Sdk.MasternodeEntry.serializer()`), unlike the coroutines-core issue which affected the SDK's primary, pervasively-used reactive surface (Room `Flow`, `StateFlow` properties across many public classes). Downgraded from blocking to suggestion accordingly; an equally valid fix is marking `MasternodeEntry` `internal` since it looks like a wire-format DTO that leaked to public visibility rather than an intentional serialization API.

Source (experiment sonnet-primary-opus-quarter-sample-20260710, cohort sonnet_primary, bucket 1): reviewers codex/general=gpt-5.6-sol(completed); sonnet5/general=claude-sonnet-5(completed); verifier=verifier-sonnet5-4045-1783737274=claude-sonnet-5; orchestrator=openai/gpt-5.6-sol reasoning=high (orchestration-only, not a reviewer/verifier).

🟡 1 suggestion(s)

<details>

```

In `packages/kotlin-sdk/sdk/build.gradle.kts`:
- [SUGGESTION] packages/kotlin-sdk/sdk/build.gradle.kts:86: kotlinx-serialization-json is runtime-only despite a public @Serializable type
  `kotlinx-serialization-json` is declared with `implementation`, so it won't be on the compile classpath of consumers pulling the new Maven coordinate. `Sdk.MasternodeEntry` (Sdk.kt:127-133) is a public, default-visibility data class annotated `@Serializable`, so the serialization compiler plugin generates a public `serializer(): KSerializer<MasternodeEntry>` on it — a consumer calling that would fail to compile without kotlinx-serialization-core on their classpath. That said, this is narrower than the coroutines-core issue: `MasternodeEntry` is never returned by any public function — only `internal fun parseActiveMasternodes` and a `private` envelope class touch it, and the actually-public `discoverActiveMasternodes()` return type (`ActiveMasternode`) isn't `@Serializable`. A typical consumer is very unlikely to hit this. The cleaner fix may be marking `MasternodeEntry` `internal` (it reads like a wire-format DTO that leaked to public visibility) rather than promoting the dependency to `api`.
```
</details>

packages/kotlin-sdk/sdk/build.gradle.kts
```diff
@@ -59,7 +77,11 @@ ksp {
...
 
 dependencies {
-    implementation(libs.kotlinx.coroutines.core)
+    // `api`, not `implementation`: the public SDK surface returns coroutines
+    // types (Room DAO `Flow<…>`, `PlatformWalletManager.pendingIdentityKeys:
+    // StateFlow<…>`, etc.), so a consumer of the published coordinate needs
+    // coroutines-core on its own compile classpath.
+    api(libs.kotlinx.coroutines.core)
     implementation(libs.kotlinx.coroutines.android)
     implementation(libs.kotlinx.serialization.json)
```
thepastaclaw, line 100: **🟡 Suggestion: kotlinx-serialization-json is runtime-only despite a public @Serializable type**

`kotlinx-serialization-json` is declared with `implementation`, so it won't be on the compile classpath of consumers pulling the new Maven coordinate. `Sdk.MasternodeEntry` (Sdk.kt:127-133) is a public, default-visibility data class annotated `@Serializable`, so the serialization compiler plugin generates a public `serializer(): KSerializer<MasternodeEntry>` on it — a consumer calling that would fail to compile without kotlinx-serialization-core on their classpath. That said, this is narrower than the coroutines-core issue: `MasternodeEntry` is never returned by any public function — only `internal fun parseActiveMasternodes` and a `private` envelope class touch it, and the actually-public `discoverActiveMasternodes()` return type (`ActiveMasternode`) isn't `@Serializable`. A typical consumer is very unlikely to hit this. The cleaner fix may be marking `MasternodeEntry` `internal` (it reads like a wire-format DTO that leaked to public visibility) rather than promoting the dependency to `api`.

```suggestion
    api(libs.kotlinx.serialization.json)
```

<sub>source: ['codex']</sub>

packages/kotlin-sdk/sdk/build.gradle.kts
thepastaclaw: Resolved in `3d05a4f` — _Publishing can create an AAR without the JNI library_ no longer present.

bfoss765:
Fixed. Publishing now fails fast via `verifyJniLibsPresent` when `libdash_sdk_jni.so` is missing for arm64-v8a/x86_64, with `-PallowMissingJni` to override for a metadata-only run. `kotlinx-coroutines-core` moved to `api` (public API returns `Flow`/`StateFlow`; POM now shows compile scope). Added `scm`/`developers` to the POM pointed at github.com/dashpay/platform rather than dashfoundation.org — and a note that the `org.dashfoundation` groupId can't go to Maven Central until we own the matching domain (thanks @HashEngineering). Wiring the native build into gradle is left for the follow-up you mentioned.

bfoss765:
@HashEngineering — proposal for the publishing coordinates, to resolve the Maven Central namespace question: use **`org.dash`** (verified via the dash.org domain, which the org controls) rather than `org.dashfoundation`. E.g. `org.dash.platform:dash-sdk-android`. Central's namespace verification for a domain groupId is a one-time DNS TXT record on dash.org (whoever controls that DNS adds it to the Sonatype/Central account used for publishing). If you'd rather avoid new namespace setup entirely, the already-verified `org.dashj` namespace works too — but `org.dash` matches where the project is headed. If there's agreement I'll push the groupId change to this PR.

HashEngineering:
in our group discussion we decided to continue using `org.dashj` which won't require any domain verification.

the `jreleaser` plugin can be used to publish to maven central.  Consult the file here https://github.com/dashpay/dashj/blob/master/core/build.gradle#L139 for a reference to support both Maven Central for releases and Sonatype for SNAPSHOTS.

We will still need to use the 'maven-publish' plugin to support publishing other the local repo on the dev's system.

packages/kotlin-sdk/sdk/build.gradle.kts
```diff
@@ -35,6 +55,12 @@ android {
     // libdash_sdk_jni.so is produced by ../build_android.sh (cargo-ndk) into
     // src/main/jniLibs/<abi>/ — AGP packages it from there. NDK r28 emits
     // 16 KB-aligned ELF segments by default (Android 15+ requirement).
+    publishing {
+        singleVariant("release") {
+            withSourcesJar()
+        }
+    }
```
thepastaclaw, line 68: **🔴 Blocking: Release publication is missing a javadoc jar — real Maven Central deploys will be rejected**

This delta turns on a real Maven Central / Sonatype deploy path (`jreleaser { deploy { maven { mavenCentral { ... } nexus2 { ... } } } }`, lines 245-273), meaning `:sdk:jreleaserDeploy` is now expected to actually push release and snapshot artifacts to Central. `android.publishing.singleVariant("release")` (lines 58-62) only calls `withSourcesJar()` — it never calls `withJavadocJar()`. Central's Publisher Portal requires a javadoc jar alongside the sources jar for every non-POM component (https://central.sonatype.org/publish/requirements/), and the snapshot deployer even sets `applyMavenCentralRules(true)` (line 263), which makes JReleaser enforce that rule before upload. The gap in `singleVariant` predates this commit, but it was inert until this delta made the deploy path real — `publishToMavenLocal` never exercises it, so this wasn't caught by local verification. Without this, the first real release or snapshot deploy fails validation.

```suggestion
    publishing {
        singleVariant("release") {
            withSourcesJar()
            withJavadocJar()
        }
    }
```

<sub>source: ['sonnet5', 'codex']</sub>

packages/kotlin-sdk/sdk/build.gradle.kts
```diff
@@ -83,3 +113,162 @@ dependencies {
...
+    // and a bare `assemble` never put a PublishToMavenRepository task in the graph, so a
+    // local build requires no GPG key. Keys come from the standard signing properties/env
+    // (signing.keyId + signing.password + signing.secretKeyRingFile, or the in-memory
+    // ORG_GRADLE_PROJECT_signingKey / signingPassword pair) when a remote publish runs.
+    signing {
+        setRequired(Callable {
+            gradle.taskGraph.allTasks.any { it is PublishToMavenRepository }
+        })
+        sign(publishing.publications["release"])
+    }
```
thepastaclaw, line 215: **🔴 Blocking: Documented in-memory signing env vars are never wired to a signatory**

The comment directly above this block (lines 173-177) says signing keys can come from either the legacy `signing.keyId` / `signing.password` / `signing.secretKeyRingFile` properties or the in-memory `ORG_GRADLE_PROJECT_signingKey` / `signingPassword` pair. The latter environment variables only populate Gradle project properties named `signingKey` and `signingPassword`; the signing plugin does not consume them automatically, and this block never calls `useInMemoryPgpKeys`. Therefore a CI release following the documented in-memory-key path fails in `signReleasePublication` before staging anything. Explicitly read those properties and pass them to `useInMemoryPgpKeys(signingKey, signingPassword)`; the separate legacy keyring-file path can continue to use Gradle's standard signing properties.

```suggestion
    signing {
        val signingKey = project.findProperty("signingKey") as String?
        val signingPassword = project.findProperty("signingPassword") as String?
        if (signingKey != null) {
            useInMemoryPgpKeys(signingKey, signingPassword)
        }
        setRequired(Callable {
            gradle.taskGraph.allTasks.any { it is PublishToMavenRepository }
        })
        sign(publishing.publications["release"])
    }
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):
## Code Review

## Prior Finding Reconciliation
- **STILL VALID:** `kotlinx-serialization-json` remains `implementation` while `Sdk.MasternodeEntry` is still a public `@Serializable` type.

## Carried-Forward Prior Findings
- The serialization compile-classpath suggestion remains open at the current head and is carried forward below.

## New Findings In Latest Delta
- Two blockers in the new remote-publishing path: the release publication stages no required Javadoc JAR, and the documented in-memory Gradle signing properties are not wired through `useInMemoryPgpKeys`.

## Additional Cumulative Findings
None.

Source: Sol reviewer `gpt-5.6-sol` (general, completed); Sonnet reviewer `claude-sonnet-5` (general, completed); verifier `claude-sonnet-5` (primary, completed). Orchestrator: `openai/gpt-5.6-sol`, reasoning=high, orchestration-only.

🔴 2 blocking

<details>

```

In `packages/kotlin-sdk/sdk/build.gradle.kts`:
- [BLOCKING] packages/kotlin-sdk/sdk/build.gradle.kts:58-62: Release publication is missing a javadoc jar — real Maven Central deploys will be rejected
  This delta turns on a real Maven Central / Sonatype deploy path (`jreleaser { deploy { maven { mavenCentral { ... } nexus2 { ... } } } }`, lines 245-273), meaning `:sdk:jreleaserDeploy` is now expected to actually push release and snapshot artifacts to Central. `android.publishing.singleVariant("release")` (lines 58-62) only calls `withSourcesJar()` — it never calls `withJavadocJar()`. Central's Publisher Portal requires a javadoc jar alongside the sources jar for every non-POM component (https://central.sonatype.org/publish/requirements/), and the snapshot deployer even sets `applyMavenCentralRules(true)` (line 263), which makes JReleaser enforce that rule before upload. The gap in `singleVariant` predates this commit, but it was inert until this delta made the deploy path real — `publishToMavenLocal` never exercises it, so this wasn't caught by local verification. Without this, the first real release or snapshot deploy fails validation.
- [BLOCKING] packages/kotlin-sdk/sdk/build.gradle.kts:178-183: Documented in-memory signing env vars are never wired to a signatory
  The comment directly above this block (lines 173-177) says signing keys can come from either the legacy `signing.keyId` / `signing.password` / `signing.secretKeyRingFile` properties or the in-memory `ORG_GRADLE_PROJECT_signingKey` / `signingPassword` pair. The latter environment variables only populate Gradle project properties named `signingKey` and `signingPassword`; the signing plugin does not consume them automatically, and this block never calls `useInMemoryPgpKeys`. Therefore a CI release following the documented in-memory-key path fails in `signReleasePublication` before staging anything. Explicitly read those properties and pass them to `useInMemoryPgpKeys(signingKey, signingPassword)`; the separate legacy keyring-file path can continue to use Gradle's standard signing properties.
```
</details>

bfoss765:
Done in `b0b8d61960`. Switched the published coordinate to the **org.dashj** group per your call — since Dash already owns dashj.org, no separate domain verification is needed for Central. This is coordinates-only: the artifact id stays `dash-sdk-android` and the Kotlin source packages stay `org.dashfoundation.dashsdk`.

Added a **jreleaser** deploy block mirroring dashj/core/build.gradle — `maven-publish` stages signed artifacts into `build/staging-deploy`, then `jreleaserDeploy` pushes release versions to Maven Central and `-SNAPSHOT`s to the Sonatype snapshots repo; signing/Sonatype creds are read from jreleaser env/props at deploy time only. `publishToMavenLocal` still works with no GPG key (gradle signing is required only when a remote publish task is in the graph). One tweak vs. dashj: the mavenCentral deployer is `active=RELEASE` (not `ALWAYS`) so it doesn't collide with the snapshot deployer.

Validated: `publishToMavenLocal` lands the artifacts under `org/dashj/dash-sdk-android/` with the right POM. Note: the wallet's dependency coordinates change to `org.dashj:dash-sdk-android` on its next AAR rebuild.

packages/kotlin-sdk/sdk/build.gradle.kts
```diff
@@ -35,6 +55,18 @@ android {
...
     // 16 KB-aligned ELF segments by default (Android 15+ requirement).
+    publishing {
+        singleVariant("release") {
+            withSourcesJar()
+            // Maven Central's Publisher Portal rejects any non-POM component that
+            // lacks a javadoc jar, and the snapshot deployer sets
+            // applyMavenCentralRules(true) which enforces it before upload
+            // (dashpay/platform#4045). AGP emits a (Kotlin-appropriate) javadoc jar
+            // here — no Dokka dependency required.
+            withJavadocJar()
```
thepastaclaw, line 66: **💬 Nitpick: Generated javadoc jar will be effectively empty — comment overstates what AGP produces**

The module (packages/kotlin-sdk/sdk/src/main) contains zero .java source files — confirmed by search, it is pure Kotlin. AGP’s `withJavadocJar()` wires up the standard Javadoc Gradle task, which only processes Java sources; it does not understand Kotlin and will not document any of the SDK’s public Kotlin API. The comment at lines 64-65 (“AGP emits a (Kotlin-appropriate) javadoc jar here — no Dokka dependency required”) overstates this: the produced jar satisfies Central’s structural requirement (a `-javadoc.jar` artifact must exist) but contains no meaningful API documentation. This does not block a real Central upload because only artifact presence is validated, not content — flagging as a comment-accuracy nitpick in case real API docs via Dokka are wanted later.

<sub>source: [sonnet5-general]</sub>

thepastaclaw (commented):
## Code Review

## Carried-Forward Prior Findings
None — all three prior findings (missing javadoc jar, unwired in-memory signing keys, leaked public MasternodeEntry serializer) are verified FIXED at fcaa91fa.

## New Findings In Latest Delta
One nitpick: the module has zero .java sources, so AGP's generated javadoc jar via withJavadocJar() will satisfy Central's structural requirement but contain no real API documentation, despite the added comment calling it "Kotlin-appropriate." This doesn't block a Central upload since only artifact presence is validated.

Source (experiment sonnet-primary-opus-quarter-sample-20260710, cohort sonnet_primary, bucket 3): reviewers codex/general=gpt-5.6-sol(completed); sonnet5/general=claude-sonnet-5(completed); verifier=verifier-sonnet5-4045-1783826342=claude-sonnet-5; orchestrator=openai/gpt-5.6-sol reasoning=high (orchestration-only, not a reviewer/verifier).

💬 1 nitpick(s)

packages/kotlin-sdk/sdk/build.gradle.kts
thepastaclaw: Resolved in `fcaa91f` — _Release publication is missing a javadoc jar — real Maven Central deploys will be rejected_ no longer present.

packages/kotlin-sdk/sdk/build.gradle.kts
thepastaclaw: Resolved in `fcaa91f` — _kotlinx-serialization-json is runtime-only despite a public @Serializable type_ no longer present.

packages/kotlin-sdk/sdk/build.gradle.kts
thepastaclaw: Resolved in `fcaa91f` — _Documented in-memory signing env vars are never wired to a signatory_ no longer present.

bfoss765:
Round-2 blockers addressed (verified against the current branch; validated with a local publish under a throwaway version):

- **Javadoc jar (3acc85ab85d4):** the release publication now calls `withJavadocJar()`; the published `-javadoc.jar` is real Dokka-generated Kotlin API docs (~1.3 MB via AGP 8.9.2's Dokka-backed javadoc task), alongside `-sources.jar` — which also closes the "effectively empty" nitpick f9d3bbb5f45d.
- **Nativeless-AAR-to-remote (8b899bc2e5e8):** split the guard — `verifyJniLibsForRemotePublish` always hard-fails and is wired to **every** `PublishToMavenRepository` (including the `file://build/staging-deploy` staging repo) and to `jreleaser{Deploy,Upload,Release}`; `-PallowMissingJni` is honored only by `verifyJniLibsForLocalPublish` on `publishToMavenLocal`. Proven end-to-end: `verifyJniLibsForRemotePublish -PallowMissingJni` fails; the local variant warns and passes. No path can push a nativeless AAR to a remote/staging repo.
- **Serialization scope (96b7b2a236ff):** `Sdk.MasternodeEntry` is now `internal`, so its generated `serializer()` isn't public API and `implementation` scope is correct.

bfoss765:
Hosting is now settled: **Maven Central under `org.dashj`**, exactly as decided above — same namespace as `dashj-core`, so no domain verification needed. The deploy config was already in place (jreleaser mirroring dashj/core/build.gradle: releases → Central Portal, `-SNAPSHOT` → Sonatype snapshots); commit `a2e94fc8c7` adds **`packages/kotlin-sdk/PUBLISHING.md`**, a one-page runbook covering the env-var-only credential wiring (`JRELEASER_MAVENCENTRAL_SONATYPE_*`, `JRELEASER_NEXUS2_SNAPSHOTS_*`, `JRELEASER_GPG_*`, plus the Gradle signing key for the staging publish — nothing committed), the release commands, and what the JNI guard refuses.

Post-merge, the first release is:

1. `./build_android.sh --verify` — builds `libdash_sdk_jni.so` (arm64-v8a + x86_64); the publish guard hard-fails without it.
2. `./gradlew :sdk:publishReleasePublicationToStagingRepository -PsdkVersion=0.1.0` — stages the signed AAR + sources/javadoc jars.
3. `./gradlew :sdk:jreleaserDeploy -PsdkVersion=0.1.0` — uploads to Maven Central; approve at central.sonatype.com.
4. Android wallet CI then pins `org.dashj:dash-sdk-android:0.1.0` (a real release version, not a snapshot).

@HashEngineering since the `org.dashj` Central Portal token and GPG key are yours, could you run this first publish after merge? PUBLISHING.md has the exact env var names — happy to adjust anything that doesn't match your dashj-core release setup.

packages/kotlin-sdk/PUBLISHING.md
```diff
@@ -0,0 +1,84 @@
...
+```bash
+# 1. Build the native JNI library (arm64-v8a + x86_64) and verify its exports.
+./build_android.sh --verify
+
+# 2. Stage the signed release AAR + sources/javadoc jars into sdk/build/staging-deploy.
+./gradlew :sdk:publishReleasePublicationToStagingRepository -PsdkVersion=0.1.0
+
+# 3. Upload the staged artifacts. Release versions -> Maven Central;
+#    -SNAPSHOT versions -> Sonatype snapshots.
+./gradlew :sdk:jreleaserDeploy -PsdkVersion=0.1.0
```
thepastaclaw, line 57: **🔴 Blocking: Clean the staging repository before assembling a release**

The runbook publishes each requested version into the fixed `sdk/build/staging-deploy` Maven repository and then gives that entire directory to both configured JReleaser deployers. Maven publication adds or replaces files for the current coordinates but does not remove other version directories, and JReleaser explicitly recommends deploying from a clean state. Reusing this sequence after an earlier snapshot or release staging run can therefore bundle stale versions with the intended release, causing duplicate-version or validation failures and potentially uploading unintended artifacts. Run `:sdk:clean` before staging; it removes the configured build directory while preserving the JNI libraries under `src/main/jniLibs`.

```suggestion
# 2. Clear previously staged versions, then stage the signed release AAR + sources/javadoc jars.
./gradlew :sdk:clean
./gradlew :sdk:publishReleasePublicationToStagingRepository -PsdkVersion=0.1.0

# 3. Upload the staged artifacts. Release versions -> Maven Central;
#    -SNAPSHOT versions -> Sonatype snapshots.
./gradlew :sdk:jreleaserDeploy -PsdkVersion=0.1.0
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

Carried-forward prior findings: none; the prior Javadoc nitpick is outdated because the configured AGP 8.9 API explicitly generates documentation from Java and Kotlin sources. New latest-delta findings: the release runbook has one blocking issue because it deploys a persistent staging repository without cleaning stale versions first. No additional cumulative findings were identified.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (failed_before_session_creation), `gpt-5.6-sol` — general (completed)

🔴 1 blocking

<details>

```

In `packages/kotlin-sdk/PUBLISHING.md`:
- [BLOCKING] packages/kotlin-sdk/PUBLISHING.md:49-54: Clean the staging repository before assembling a release
  The runbook publishes each requested version into the fixed `sdk/build/staging-deploy` Maven repository and then gives that entire directory to both configured JReleaser deployers. Maven publication adds or replaces files for the current coordinates but does not remove other version directories, and JReleaser explicitly recommends deploying from a clean state. Reusing this sequence after an earlier snapshot or release staging run can therefore bundle stale versions with the intended release, causing duplicate-version or validation failures and potentially uploading unintended artifacts. Run `:sdk:clean` before staging; it removes the configured build directory while preserving the JNI libraries under `src/main/jniLibs`.
```
</details>

packages/kotlin-sdk/PUBLISHING.md
bfoss765: Fixed in the latest push: a `cleanStagingDeploy` Delete task now wipes `sdk/build/staging-deploy` before **every** `*ToStagingRepository` publish (wired via `tasks.withType<PublishToMavenRepository>`), so JReleaser always deploys from a clean staging state regardless of prior snapshot/release runs. Verified in the task graph with `--dry-run`; PUBLISHING.md updated to note the automatic wipe.

PR #4048, opened 2026-07-09
fix(swift-example-app): let pre-programmed distribution recipients open token Claim

QuantumExplorer:

A pre-programmed-distribution recipient could never open the token Claim action row in SwiftExampleApp. Two compounding bugs (iOS mirror of the KotlinExampleApp `TokenActionResolver.resolveClaim` fix):

1. `DataContractParser` parsed `preProgrammedDistribution` expecting a `distributionSchedule` array, but real contract JSON is `{"$formatVersion":"0","distributions":{"<timestampMs>":{"<recipientBase58>":amount}}}` — so the schedule was always empty and recipient info was lost.
2. `resolveClaim` only allowed the designated `newTokensDestinationIdentity`; pre-programmed recipients fell through to a denial.

## What was done?

- `Sources/SwiftDashSDK/Core/Utils/DataContractParser.swift`: parse the real rs-dpp `distributions` map into `DistributionEvent` entries (`triggerTime` from the ms timestamp key, `recipient` from the base58 key, `amount` stringified losslessly through `UInt64` so values above `Int64.max` aren't truncated). The legacy `distributionSchedule` array shape is kept as a fallback.
- `SwiftExampleApp/Views/TokenActionPermissionsView.swift` (`resolveClaim`): return `.allowed` when the identity appears as a recipient in any pre-programmed distribution event. The check runs before the minting-side `mintingAllowChoosingDestination`/designated-recipient guards, which would otherwise deny legitimate recipients. Maturity and already-claimed checks are left to on-chain enforcement, matching Android.
- `SwiftTests/SwiftDashSDKTests/SDKMethodTests.swift`: marked `testDirectMethodCall` `@MainActor` to fix a pre-existing Swift 6 `sending-risks-data-race` compile error that blocked building the hermetic test target.
- `TokenClaimActionView` needed no change: it auto-selects `.preProgrammed` when it's the only schedule and already passes the correct FFI discriminant (0) to `platform_wallet_token_claim`.

Note: tokens persisted before this fix keep an empty schedule; re-fetching the contract re-materializes the recipient list.

## How Has This Been Tested?

- New hermetic unit tests in `SwiftTests/SwiftDashSDKTests/DataContractParserPreProgrammedTests.swift` (5 tests, in-memory SwiftData container): real `distributions` shape mirroring live testnet contract `CNpuW3ZjmTrfxKm2VyEJYSBm2jFn7Va2YCmRyVP1ZPvC` (token `androidqadist`, recipient `29n6ZVEWvTVM7BaQCs7Ubsi4KdnqkPBTQ8QxEyyXciAc`), string and >`Int64.max` amounts, multi-timestamp × multi-recipient flattening, legacy array fallback, and absent pre-programmed config.
- `swift test --filter DataContractParserPreProgrammedTests` — all 5 pass.
- Full `./build_ios.sh` (xcframework + SwiftExampleApp simulator build) succeeds.

## Breaking Changes

None.

## Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved claim eligibility handling so accounts listed as recipients in pre-programmed distribution schedules are recognized correctly.
  - Fixed parsing of pre-programmed distribution data from newer contract formats, while keeping support for the older format.
  - Preserved exact distribution amounts, including very large values, and handled timestamp-based schedules more reliably.

- **Tests**
  - Added coverage for new and legacy pre-programmed distribution shapes, amount formatting, and empty/missing distribution cases.

### DIFF
--- a/packages/swift-sdk/Sources/SwiftDashSDK/Core/Utils/DataContractParser.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/Core/Utils/DataContractParser.swift
@@ -407,6 +407,39 @@ public struct DataContractParser {
         return "0"
     }
 
+    /// Render a pre-programmed distribution amount to a canonical
+    /// decimal string. JSONSerialization decodes numeric JSON values as
+    /// `NSNumber` and string values as `String`; token amounts can
+    /// exceed `Int.max`, so read through `UInt64`/`NSNumber` rather than
+    /// `Int` to avoid truncating large balances. Returns `nil` for
+    /// unparseable values so the caller can skip the malformed entry.
+    private static func stringifyDistributionAmount(_ value: Any) -> String? {
+        if let string = value as? String {
+            return string
+        }
+        // `NSNumber` covers Int/UInt/Double bridged from JSON. Prefer a
+        // lossless UInt64 read; fall back to the number's own string.
+        if let number = value as? NSNumber {
+            if let unsigned = UInt64(exactly: number) {
+                return String(unsigned)
+            }
+            if let signed = Int64(exactly: number) {
+                return String(signed)
+            }
+            return number.stringValue
+        }
+        if let unsigned = value as? UInt64 {
+            return String(unsigned)
+        }
+        if let signed = value as? Int64 {
+            return String(signed)
+        }
+        if let intValue = value as? Int {
+            return String(intValue)
+        }
+        return nil
+    }
+
     private static func parseTokenConfiguration(token: PersistentToken, from tokenDict: [String: Any]) {
         // Basic properties
         let maxSupplyStr = extractTokenSupply(from: tokenDict, key: "maxSupply")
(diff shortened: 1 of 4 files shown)


thepastaclaw:
✅ Review complete (commit 77c33709)

thepastaclaw (changes requested):
## Code Review

Source: reviewers: claude `opus` general (failed: quota), codex `gpt-5.5` general (completed; ACP exited non-zero after emitting parseable JSON). Verifier: codex `gpt-5.5`.

The parser and claim resolver changes are directionally correct for newly ingested contracts, but the normal download path still returns existing persisted contracts before running the updated parser. That leaves contracts saved with the old parser stuck with an empty pre-programmed schedule, so the PR does not fix the affected Claim eligibility path for existing local data.

🔴 1 blocking


<details>

```

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Utils/ContractDownloader.swift`:
- [BLOCKING] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Utils/ContractDownloader.swift:254-266: Existing saved contracts are not repaired by the parser fix
  Contracts persisted before this parser change already have `preProgrammedDistribution` saved with an empty `distributionSchedule`, because the old parser did not understand the real `distributions` map shape. When the user re-adds or re-fetches the same contract, this branch returns the existing row after only touching `lastAccessedAt` and linking the owner, so `DataContractParser.parseDataContract` never runs against the freshly fetched JSON. The new `resolveClaim` logic depends on `token.preProgrammedDistribution?.distributionSchedule` containing the recipient, so those already-saved contracts still deny the Claim row. Update the existing contract/token rows from the fetched JSON here, or add an explicit migration/refresh path that rebuilds the saved token schedule.
```
</details>

PR #4050, opened 2026-07-09
feat(platform-wallet): expose classified connected SPV peers

QuantumExplorer:

SPV status screens (dashwallet-ios' SwiftDashSDK SPV Status, SwiftExampleApp) can show sync progress but not *who* the client is connected to. dash-spv pushes connected-peer addresses through `NetworkEvent::PeersUpdated` but offers no query for them, and nothing classifies a peer as an evonode / masternode / regular node.

## What was done?

- **rs-platform-wallet**: new `spv/peers.rs`. A `PeerTracker` event handler rides the SPV client's handler list and mirrors the latest `PeersUpdated` snapshot (cleared on stop and when the run loop exits). `SpvRuntime::connected_peers()` classifies each address against the masternode list engine's latest list: `HighPerformance` entry → Evonode, `Regular` → Masternode, absent → Normal, masternode list not yet synced → Unknown. Matching is by `ip:port` first, then bare IP.
- **rs-platform-wallet-ffi**: `platform_wallet_manager_spv_connected_peers` returns a heap-owned `[FFISpvPeerInfo]` (`address` C string + `SPV_PEER_NODE_TYPE_*` constant) with a paired `platform_wallet_manager_spv_connected_peers_free`, following the ban-list array idiom.
- **swift-sdk**: `PlatformSpvPeerInfo` / `PlatformSpvPeerNodeType` types, a one-shot `connectedSpvPeers()` query, and a `@Published spvPeers` mirror fed by the existing 1 Hz progress poll (equality-gated like its siblings).

Consumed by the dashwallet-ios SPV Status screen (companion PR on `swift-sdk-integration`).

## How Has This Been Tested?

- Unit tests in `spv/peers.rs`: classification (masternode / evonode / normal), unknown-when-list-absent, IP-only fallback match, and `PeersUpdated` snapshot replace + clear semantics — all pass.
- `cargo check` / `cargo clippy --all-targets` / per-crate `--all-features --tests` clean for `platform-wallet` and `platform-wallet-ffi`; `cargo fmt --all`.
- `build_ios.sh --target sim` rebuilt the simulator xcframework and SwiftExampleApp builds against the regenerated header.
- End-to-end verification (live peers + badges on-screen) done via the dashwallet-ios companion branch.

## Breaking Changes

None — additive FFI surface only.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added visibility into currently connected SPV peers.
  * Introduced peer classification labels for normal, masternode, evonode, and unknown peers.
  * Added a new Swift API and published state so apps can observe the live SPV peer list.
* **Bug Fixes**
  * Peer updates now refresh only when the list changes, helping keep UI state in sync.
  * Empty or unavailable peer data is handled cleanly without breaking the app.

### DIFF
--- a/packages/rs-platform-wallet/src/spv/peers.rs
+++ b/packages/rs-platform-wallet/src/spv/peers.rs
@@ -8,7 +8,7 @@
 //! address into Evonode / Masternode / Normal for display.
 
 use std::collections::HashMap;
-use std::net::{IpAddr, SocketAddr};
+use std::net::SocketAddr;
 use std::sync::Mutex;
 
 use dash_spv::network::NetworkEvent;
@@ -77,10 +77,12 @@ impl EventHandler for PeerTracker {
 
 /// Classify peer addresses against a masternode list.
 ///
-/// A peer is matched by full `ip:port` first, then by bare IP — the list
-/// advertises the P2P endpoint we dialed, but an IP match alone is already
-/// conclusive since masternodes don't share hosts with unrelated peers.
-/// `None` (masternode list not yet synced) classifies every peer
+/// A peer is matched by exact `ip:port` only — the list advertises the
+/// P2P endpoint we dial, so a listed masternode we're connected to
+/// matches exactly. A bare-IP fallback would misclassify setups where
+/// one host runs both a masternode and an unrelated peer on another
+/// port (localhost regtest/devnet stacks in particular). `None`
+/// (masternode list not yet synced) classifies every peer
 /// [`SpvPeerNodeType::Unknown`].
 pub(crate) fn classify_peers(
     addresses: &[SocketAddr],
@@ -97,7 +99,6 @@ pub(crate) fn classify_peers(
     };
 
     let mut by_socket: HashMap<SocketAddr, SpvPeerNodeType> = HashMap::new();
-    let mut by_ip: HashMap<IpAddr, SpvPeerNodeType> = HashMap::new();
     for qualified in list.masternodes.values() {
         let entry = &qualified.masternode_list_entry;
         let Some(service) = entry.service_address.primary_service_address() else {
@@ -108,7 +109,6 @@ pub(crate) fn classify_peers(
             EntryMasternodeType::HighPerformance { .. } => SpvPeerNodeType::Evonode,
         };
         by_socket.insert(service, node_type);
-        by_ip.insert(service.ip(), node_type);
     }
 
     addresses
@@ -117,7 +117,6 @@ pub(crate) fn classify_peers(
             address,
             node_type: by_socket
                 .get(&address)
-                .or_else(|| by_ip.get(&address.ip()))
                 .copied()
                 .unwrap_or(SpvPeerNodeType::Normal),
         })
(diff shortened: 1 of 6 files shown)


thepastaclaw:
✅ Review complete (commit e1eada04)

packages/rs-platform-wallet/src/spv/peers.rs
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    addresses
+        .iter()
+        .map(|&address| SpvPeerInfo {
+            address,
+            node_type: by_socket
+                .get(&address)
+                .or_else(|| by_ip.get(&address.ip()))
+                .copied()
```
thepastaclaw, line 120: **🔴 Blocking: Avoid classifying peers by bare IP on shared localhost setups**

The classifier falls back from an exact socket match to a bare-IP match, so any connected peer on the same IP as a masternode is reported as that masternode type even when its port is not in the masternode list. This is reachable in the existing Swift integration setup, which connects to a normal Core peer at `127.0.0.1:<coreP2P>` plus masternode peers on `127.0.0.1`; the Core peer misses `by_socket` and then matches `by_ip`, causing the new connected-peer API to report the normal peer as a masternode/evonode. Use exact `ip:port` matching for shared/local addresses, or otherwise disable the bare-IP fallback when the IP can host unrelated local peers.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/spv/runtime.rs
```diff
@@ -254,6 +259,33 @@ impl SpvRuntime {
...
+            .as_ref()
+            .and_then(|client| client.masternode_list_engine().ok());
+        drop(client_guard);
+
+        match engine {
+            Some(engine) => {
+                let engine_guard = engine.read().await;
+                classify_peers(&addresses, engine_guard.latest_masternode_list())
+            }
+            None => classify_peers(&addresses, None),
```
thepastaclaw, line 291: **🟡 Suggestion: Check the live client before copying the peer snapshot**

`connected_peers()` copies the tracker snapshot before checking whether `self.client` is still present. During a concurrent `stop()`, `stop()` removes the client under the write lock and clears the tracker afterward, so this method can copy old addresses, observe no client, and still return those stale peers classified as `Unknown`. Take the client read lock first and return an empty list when the client has already been stopped, matching the Rust doc comment and FFI contract.

```suggestion
        let client_guard = self.client.read().await;
        let Some(client) = client_guard.as_ref() else {
            return Vec::new();
        };

        let addresses = self.peer_tracker.snapshot();
        if addresses.is_empty() {
            return Vec::new();
        }

        let engine = client.masternode_list_engine().ok();
        drop(client_guard);

        match engine {
            Some(engine) => {
                let engine_guard = engine.read().await;
                classify_peers(&addresses, engine_guard.latest_masternode_list())
            }
            None => classify_peers(&addresses, None),
        }
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):
## Code Review

Both agent-reported issues are present in the checked-out PR. The bare-IP fallback can misclassify supported localhost/regtest peer configurations and should be fixed before merge; the concurrent stop race is also real but limited to a stale snapshot edge case.

Source: reviewers: codex-general gpt-5.5, codex-rust-quality gpt-5.5, codex-ffi-engineer gpt-5.5 completed; claude/opus general, rust-quality, and ffi-engineer failed due extra-usage quota; verifier: codex gpt-5.5 after claude/opus verifier quota failure.

<details>

```

In `packages/rs-platform-wallet/src/spv/peers.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/spv/peers.rs:118-121: Avoid classifying peers by bare IP on shared localhost setups
  The classifier falls back from an exact socket match to a bare-IP match, so any connected peer on the same IP as a masternode is reported as that masternode type even when its port is not in the masternode list. This is reachable in the existing Swift integration setup, which connects to a normal Core peer at `127.0.0.1:<coreP2P>` plus masternode peers on `127.0.0.1`; the Core peer misses `by_socket` and then matches `by_ip`, causing the new connected-peer API to report the normal peer as a masternode/evonode. Use exact `ip:port` matching for shared/local addresses, or otherwise disable the bare-IP fallback when the IP can host unrelated local peers.

- [SUGGESTION] packages/rs-platform-wallet/src/spv/runtime.rs:269-285: Check the live client before copying the peer snapshot
  `connected_peers()` copies the tracker snapshot before checking whether `self.client` is still present. During a concurrent `stop()`, `stop()` removes the client under the write lock and clears the tracker afterward, so this method can copy old addresses, observe no client, and still return those stale peers classified as `Unknown`. Take the client read lock first and return an empty list when the client has already been stopped, matching the Rust doc comment and FFI contract.
```
</details>

packages/rs-platform-wallet/src/spv/peers.rs
QuantumExplorer: Fixed in e1eada04d9 — dropped the bare-IP fallback entirely; classification now matches exact ip:port only, and the test pins that a same-IP/different-port peer classifies Normal.

packages/rs-platform-wallet/src/spv/runtime.rs
QuantumExplorer: Fixed in e1eada04d9 — connected_peers() now takes the client read lock first and returns empty when the client is stopped, before copying the tracker snapshot.

QuantumExplorer:
Reviewed

Issue #4069, opened 2026-07-09
kotlin-sdk/platform-wallet: duplicated unspent TXO rows after SPV rescan following unclean shutdown (inflated balance)


bfoss765:
**Observed live** (Android wallet integration, testnet, Galaxy S22, kotlin-sdk 4.0.0 @ feat/kotlin-sdk-and-example-app, rust-dashcore 647fa982):

A wallet whose SDK view initially matched dashj **exactly** (154,427,919 duffs both sides, verified by a continuous parity harness) became inflated after several unclean app terminations and subsequent SPV restarts/rescans:

- SDK unspent TXO rows: **271**, summing **413,124,109 duffs**
- dashj spend candidates (source of truth): **128** outputs, **154,427,919 duffs**
- Delta: +258,696,190 duffs (~2.1× row duplication)
- Distinct txids over the SDK's TXO rows stayed constant (370) throughout — the inflation is duplicated/unreconciled **rows**, not newly discovered transactions.

Reproduction context: repeated process kills without `stopSpv` (no teardown hook), with the SPV header store observed regressing between runs (resumed at a lower height than previously reached), followed by rescans over the already-persisted TXO set. Suspect: the rescan re-emits TXO persistence events that append rather than reconcile against existing rows (dedup key not enforced end-to-end through the changeset → Room bridge), and/or spent-markers lost when the header store regresses.

Requests:
1. Idempotent TXO persistence on rescan (outpoint-unique upsert semantics through the persistence bridge).
2. Crash-safe SPV storage (the header-store regression after SIGKILL suggests non-atomic writes).
3. A wallet-level `rescan`/reset API — integrators currently have to delete Room TXO/tx rows directly and clear SPV storage to recover (works, but the live Rust in-memory view stays stale until process restart).

The integrating app now detects this state (balance parity vs dashj) and auto-recovers by clearing + rescanning, but the corruption is SDK-side. cc @quantumexplorer for review.

Context: dashpay/dash-wallet#1507 Phase 5a shadow-sync harness. 🤖 Filed with [Claude Code](https://claude.com/claude-code)

bfoss765:
[...]

PR #4071, opened 2026-07-09
fix(platform-wallet): freeze sync watermark on persistence fault — TXO loss/duplication (#4069)

bfoss765:
Fixes the root cause behind #4069 (all three reported signatures).

## Root cause
`key-wallet-manager` publishes wallet persistence events on a **bounded tokio broadcast ring (capacity 1000) with fire-and-forget sends**. During historical SPV catch-up the producer outruns the JNI→Room consumer, `recv()` returns `Lagged(n)`, and the dropped events — exactly the TXO/record/spent-marker payloads — are silently lost. `SyncHeightAdvanced` is the only event whose height reaches the persisted watermark (`WalletEntity.syncedHeight`), so the watermark advances past blocks whose rows never hit storage:

- **store-nothing**: fresh scan floods the ring → all record events dropped → zero rows persisted, watermark = tip → wallet rehydrates empty-and-"fully scanned" (funds invisible, unrecoverable without wallet re-creation)
- **inflated balance**: dropped spend events → consumed TXOs rehydrate as spendable
- **watermark vs rows**: nothing gates the watermark on record persistence across changesets (within one changeset the Kotlin handler is already atomic — verified)

Also verified NOT the cause: outpoint encoding is consistently wire-order end-to-end, and Room uses a true `@Upsert` on the outpoint PK.

## Fix (workspace-local, `rs-platform-wallet/src/changeset/core_bridge.rs`)
Once a persistence fault occurs in a session — broadcast `Lagged` **or** a rejected `store()` — `freeze_synced_height_if_faulted` strips `synced_height` from every subsequent changeset (record/UTXO deltas still persist), pinning the durable watermark at the last fully-committed height. On next launch the wallet restores the lower watermark, the SPV scan resumes over the gap, and the idempotent outpoint-keyed upserts restore missing rows / re-mark lost spends — self-healing instead of silent corruption. Fault logs raised to `error!`.

The bounded capacity itself lives in the pinned `key-wallet-manager` crate; bumping it and/or moving persistence to a backpressured channel is the recommended upstream follow-up (noted in code comments).

## Tests
- Rust: 2 new unit tests over the freeze helper; `platform-wallet` suite 400 passed / 0 failed.
- Kotlin: new regression test pinning that a rolled-back changeset does not advance `WalletEntity.syncedHeight`; `:sdk:testDebugUnitTest` green.
- On-device validation on the Android wallet's parity harness (dashpay/dash-wallet#1507, Galaxy S22/testnet) to follow — the harness that produced the three reproductions in #4069.

cc @quantumexplorer

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented synchronization progress from advancing when wallet persistence encounters an error or misses events.
  * Ensured failed or rolled-back changesets do not permanently update the stored synchronization height.
  * Preserved other wallet progress information so affected data can be reprocessed safely on the next scan.

### DIFF
--- a/packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandlerTest.kt
+++ b/packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandlerTest.kt
@@ -150,6 +150,57 @@ class PlatformWalletPersistenceHandlerTest {
         assertEquals(123_456, db.walletDao().getByWalletId(walletId)!!.syncedHeight)
     }
 
+    /**
+     * dashpay/platform#4069 (Kotlin half of signature C): the
+     * `syncedHeight` watermark is written by [onWalletChangesetHeader]
+     * into the SAME buffered transaction as the TXO/tx rows of its
+     * changeset. A round that rolls back (`success = false`) must
+     * therefore NOT advance the persisted watermark — otherwise the
+     * durable watermark could outrun the rows it implies, exactly the
+     * "empty-and-scanned after restart" corruption in #4069. Pins the
+     * rollback path for the core header specifically (the sibling
+     * `changesetRollbackDiscardsBufferedWrites` only covers the
+     * platform sync-state write).
+     */
+    @Test
+    fun walletChangesetHeaderDoesNotAdvanceSyncedHeightOnRollback() = runTest {
+        handler.onPersistWalletMetadata(walletId, testnet, groupId, 0)
+        // Establish a committed baseline watermark.
+        handler.onChangesetBegin(walletId)
+        handler.onWalletChangesetHeader(
+            walletId = walletId,
+            hasSyncedHeight = true,
+            syncedHeight = 1_000,
+            hasBalance = false,
+            confirmedDelta = 0,
+            unconfirmedDelta = 0,
+            immatureDelta = 0,
+            lockedDelta = 0,
+            lastAppliedChainLockBytes = ByteArray(0),
+        )
+        handler.onChangesetEnd(walletId, success = true)
+        assertEquals(1_000, db.walletDao().getByWalletId(walletId)!!.syncedHeight)
+
+        // A later round tries to advance the watermark but rolls back.
+        handler.onChangesetBegin(walletId)
+        handler.onWalletChangesetHeader(
+            walletId = walletId,
+            hasSyncedHeight = true,
+            syncedHeight = 2_000,
+            hasBalance = false,
+            confirmedDelta = 0,
+            unconfirmedDelta = 0,
+            immatureDelta = 0,
+            lockedDelta = 0,
+            lastAppliedChainLockBytes = ByteArray(0),
+        )
+        handler.onChangesetEnd(walletId, success = false)
+
+        // Watermark stays at the last committed value — never the
+        // rolled-back 2_000.
+        assertEquals(1_000, db.walletDao().getByWalletId(walletId)!!.syncedHeight)
+    }
+
     @Test
     fun walletChangesetAddsAccountUtxoAndTransaction() = runTest {
         handler.onPersistWalletMetadata(walletId, testnet, groupId, 0)
--- a/packages/rs-platform-wallet-ffi/src/persistence.rs
+++ b/packages/rs-platform-wallet-ffi/src/persistence.rs
@@ -15,7 +15,7 @@ use key_wallet::wallet::managed_wallet_info::wallet_info_interface::WalletInfoIn
 use key_wallet::wallet::managed_wallet_info::ManagedWalletInfo;
 use key_wallet::wallet::Wallet;
 use key_wallet::AddressInfo;
-use parking_lot::RwLock;
+use parking_lot::{Mutex, RwLock};
 use std::str::FromStr;
 
 use crate::types::{FFINetwork, Network};
@@ -610,10 +610,62 @@ impl Default for PersistenceCallbacks {
     }
 }
 
+/// Defensive state machine for the begin→end FFI callback round, guarded
+/// by [`FFIPersister::round_lock`]. `in_round` is set when a round opens
+/// and cleared once it closes, so a nested begin (or an `end` with no
+/// matching `begin`) is detectable and rejected — as an error, never a
+/// panic — instead of silently corrupting the client's single in-flight
+/// transaction state.
+#[derive(Default)]
+struct RoundGuardState {
+    in_round: bool,
+}
+
+impl RoundGuardState {
+    /// Open a round. Rejects (does not panic) if one is already open —
+    /// a nested begin, or an unclean round left open by a prior call
+    /// that unwound between its begin and end.
+    fn begin_round(&mut self) -> Result<(), PersistenceError> {
+        if self.in_round {
+            return Err(PersistenceError::backend(
+                "FFIPersister: changeset round already open (nested begin); \
+                 refusing to start a new round",
+            ));
+        }
+        self.in_round = true;
+        Ok(())
+    }
+
+    /// Close the current round. Rejects (does not panic) if no round is
+    /// open — an unmatched end.
+    fn end_round(&mut self) -> Result<(), PersistenceError> {
+        if !self.in_round {
+            return Err(PersistenceError::backend(
+                "FFIPersister: changeset round is not open (unmatched end)",
+            ));
+        }
+        self.in_round = false;
+        Ok(())
+    }
+}
+
 /// In-memory persister that accumulates changesets and notifies via callbacks.
 pub struct FFIPersister {
     callbacks: PersistenceCallbacks,
     pending: RwLock<BTreeMap<WalletId, PlatformWalletChangeSet>>,
+    /// Serializes the ENTIRE begin→per-kind→end callback round of
+    /// [`Self::store`]. Every round producer (the core-changeset bridge,
+    /// platform-address sync, shielded sync, spawned DashPay tasks) shares
+    /// one `Arc<FFIPersister>` and calls `store()` concurrently; the host
+    /// client keeps a single in-flight-round transaction state (Kotlin: one
+    /// per-wallet buffer; Swift: one global `inChangeset` flag), so two
+    /// overlapping rounds would let one round's writes land in — or roll
+    /// back with — the other round's transaction. That drops core TXO /
+    /// spent-marker rows while both `store()` calls still return `Ok`,
+    /// bypassing the durable-watermark fault latch and recreating
+    /// dashpay/platform#4069. Holding this lock for the whole round makes
+    /// each round atomic with respect to every other round.
+    round_lock: Mutex<RoundGuardState>,
 }
 
 impl FFIPersister {
@@ -621,6 +673,7 @@ impl FFIPersister {
         Self {
             callbacks,
             pending: RwLock::new(BTreeMap::new()),
+            round_lock: Mutex::new(RoundGuardState::default()),
         }
     }
 }
@@ -637,6 +690,23 @@ impl PlatformWalletPersistence for FFIPersister {
         wallet_id: WalletId,
         changeset: PlatformWalletChangeSet,
     ) -> Result<(), PersistenceError> {
+        // Serialize the ENTIRE begin→per-kind→end round against every
+        // other round producer (see `round_lock`'s field doc and
+        // dashpay/platform#4069). The lock is a synchronous
+        // `parking_lot::Mutex`, NOT a `tokio::sync::Mutex`: `store()`
+        // is a synchronous trait method invoked directly (blocking) from
+        // both async tasks and blocking FFI entry points, so an async
+        // mutex cannot be `.await`ed here and `blocking_lock()` panics
+        // inside a runtime. Serialization — not async yielding — is the
+        // requirement; callers already block for the round's duration, so
+        // the sync mutex only adds waiting under genuine round contention.
+        let mut round = self.round_lock.lock();
+
+        // Open the round on the Rust side (rejects a nested begin / an
+        // unclean round left open by a prior unwind — error, never
+        // panic). Matched 1:1 with the `round.end_round()` below.
+        round.begin_round()?;
+
         // Bracket the whole per-kind callback sequence with a
         // begin/end pair so clients (Swift, etc.) can treat the
         // round as a single atomic transaction: begin opens a
@@ -648,7 +718,19 @@ impl PlatformWalletPersistence for FFIPersister {
         if let Some(cb) = self.callbacks.on_changeset_begin_fn {
             let result = unsafe { cb(self.callbacks.context, wallet_id.as_ptr()) };
             if result != 0 {
-                eprintln!("Changeset-begin callback returned error code {}", result);
+                // A nonzero begin means the client could NOT open its
+                // transaction. Proceeding would run every per-kind
+                // callback against no batch and then fire an unmatched
+                // `end`. Treat it as fatal: close the Rust-side round
+                // (so `in_round` doesn't wedge) and fail now, before any
+                // per-kind write. (Unlike the previous advisory-log
+                // behavior, the round is aborted so no state advances
+                // against an unopened batch.)
+                let _ = round.end_round();
+                return Err(PersistenceError::backend(format!(
+                    "changeset-begin callback returned error code {result}; \
+                     round aborted before any write"
+                )));
             }
         }
         let mut round_success = true;
@@ -1525,6 +1607,15 @@ impl PlatformWalletPersistence for FFIPersister {
                 round_success = false;
             }
         }
+
+        // Close the round: its `end` callback has fired (committing or
+        // rolling back the client transaction), or there was no end
+        // callback wired. Clear the state-machine flag now — BEFORE any
+        // early return below — so a rejected round doesn't wedge the
+        // persister into permanent "round already open" rejection on the
+        // next `store()`. (`end_round` only errors on an unmatched end,
+        // which cannot happen here since `begin_round` succeeded above.)
+        round.end_round()?;
 
         if !round_success {
             return Err(PersistenceError::backend(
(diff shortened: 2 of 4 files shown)


thepastaclaw:
✅ Review complete (commit 4313d843)

packages/rs-platform-wallet/src/changeset/core_bridge.rs
```diff
@@ -78,22 +123,34 @@ where
...
+                                // outrun them; the next scan re-emits and the
+                                // idempotent upserts recover the state.
+                                persistence_faulted = true;
+                                tracing::error!(
                                     wallet_id = %hex::encode(wallet_id),
                                     error = %e,
-                                    "Persister rejected core changeset; state will be re-emitted on next sync round"
+                                    "Persister rejected core changeset; freezing sync watermark so the next scan re-persists the missing rows (dashpay/platform#4069)"
                                 );
                             }
```
thepastaclaw: **🟡 Suggestion: Task-global fault flag freezes every wallet's watermark on a single wallet's fault**

`persistence_faulted` is a single `bool` in the shared adapter task, and every event for every wallet drains from the same broadcast subscriber. When wallet A's `store()` fails, wallet B's `synced_height` also stops advancing for the rest of the session even though wallet B's rows are still landing atomically. This is safe (worst case is a redundant rescan on next launch), which is why it's not blocking — but a `HashMap<WalletId, bool>` on the adapter would confine the freeze to the wallet that actually lost rows. A `Lagged` event has no wallet id, so that path would still have to fault every known wallet — the win is limited to the `store()`-rejection path. Worth considering if multi-wallet users are common on the JNI target where the throughput mismatch that motivated this fix is severe.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Source: reviewers: claude `opus` general (ok), codex `gpt-5.5` general (failed), claude `opus` security-auditor (ok), codex `gpt-5.5` security-auditor (failed), claude `opus` ffi-engineer (ok), codex `gpt-5.5` ffi-engineer (failed). Verifier: claude `opus`.

Focused, well-documented fix for a persistence-vs-watermark race in the wallet event adapter. Freezing durable `synced_height` after any broadcast lag or `store()` rejection is correctly narrow — only `synced_height` is stripped, records/UTXOs/last_processed_height stay intact. Two non-blocking suggestions: the fault flag is task-global (freezes all wallets when one faults) and the async fault-then-freeze loop lacks integration coverage.

🟡 2 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet/src/changeset/core_bridge.rs:114-155: Task-global fault flag freezes every wallet's watermark on a single wallet's fault
  `persistence_faulted` is a single `bool` in the shared adapter task, and every event for every wallet drains from the same broadcast subscriber. When wallet A's `store()` fails, wallet B's `synced_height` also stops advancing for the rest of the session even though wallet B's rows are still landing atomically. This is safe (worst case is a redundant rescan on next launch), which is why it's not blocking — but a `HashMap<WalletId, bool>` on the adapter would confine the freeze to the wallet that actually lost rows. A `Lagged` event has no wallet id, so that path would still have to fault every known wallet — the win is limited to the `store()`-rejection path. Worth considering if multi-wallet users are common on the JNI target where the throughput mismatch that motivated this fix is severe.
- [SUGGESTION] packages/rs-platform-wallet/src/changeset/core_bridge.rs:116-178: No integration coverage for the async fault-then-freeze loop
  The new Rust tests exercise `freeze_synced_height_if_faulted` in isolation and the Kotlin test pins the atomic-rollback invariant. Nothing exercises the loop itself: that a `RecvError::Lagged` actually flips `persistence_faulted`, and that a subsequent `SyncHeightAdvanced` is stripped and dropped via `is_empty_no_records`. A small test with a mock `PlatformWalletPersistence` that (a) returns `Err` on the first `store()` and asserts a following watermark-only event is not delivered, and (b) drives a real `broadcast::Sender` past its capacity to synthesize `Lagged` and asserts the same, would lock in the wiring — not just the helper. The loop is exactly where the earlier `warn!`-and-continue behavior lived and where a future refactor is most likely to silently reintroduce the bug.
```
</details>

bfoss765:
**On-device validation: PASSED** (Galaxy S22, testnet, Android wallet parity harness from dashpay/dash-wallet#1507).

With this fix in the native library: full wallet re-creation → complete chain scan → discovered TXOs **persisted correctly** → balance parity vs dashj confirmed → **verification re-passed after app restart** (the exact scenario that previously left the wallet empty-and-scanned or inflated — see #4069's three reproductions). Followed by the first successful end-to-end L1→shielded transfer through the SDK pipeline (asset lock → IS lock → Halo2 proof → Type 18 → credits minted).

The integrating app's evidence gate now opens and stays open across restarts. Recommend prioritizing review — this unblocks SDK-consuming wallets generally, not just this integration.

thepastaclaw:
@coderabbitai review

QuantumExplorer (changes requested):
Codex review of c5a0358074: request changes.

The watermark freeze is directionally correct, but it relies on FFIPersister.store returning Ok only after one isolated atomic callback round. That invariant is currently false. FFIPersister has no lock across begin, per-kind callbacks, and end, while core persistence, platform-address sync, shielded sync, and spawned DashPay work share it concurrently. Kotlin keeps one unsynchronized buffer per wallet and Swift keeps one global inChangeset transaction state. Interleaved rounds can therefore drop or roll back core TXO or spent-marker writes while both stores return Ok; the new fault latch remains false and a later SyncHeightAdvanced can still commit. This recreates #4069.

Before merge, serialize the entire FFI callback round with one global lock, treat a nonzero begin as fatal, reject nested begins and unmatched ends, and add a forced concurrent-store regression. Also subscribe to WalletEvents before publishing the manager: subscription currently occurs inside the spawned task, so pre-poll events are lost without Lagged. Finally, the capacity-1000 fire-and-forget channel remains lossy; the patch is fail-closed after detected lag but does not guarantee the claimed one-restart self-healing under repeatable overload. Use a backpressured or atomic persistence path, or narrow the claim and surface a hard sync fault.

The current tests do not exercise concurrent FFI rounds, actual Lagged handling, a rejected store followed by a watermark, startup timing, or two-launch recovery. This is client wallet persistence only and has no consensus impact.

packages/rs-platform-wallet/src/changeset/core_bridge.rs
```diff
@@ -78,22 +123,34 @@ where
...
+                                // empty BlockProcessed, a watermark-only event
+                                // stripped by the fault guard above, etc. —
+                                // nothing to persist. Skip the round-trip.
                                 continue;
                             }
                             let cs = PlatformWalletChangeSet {
                                 core: Some(core),
                                 ..PlatformWalletChangeSet::default()
                             };
                             if let Err(e) = persister.store(wallet_id, cs) {
```
QuantumExplorer: P1: an Ok result here is not currently evidence that this core changeset committed atomically. FFIPersister permits overlapping begin-to-end callback rounds from core, platform-address, shielded, and DashPay writers. Kotlin can replace the open wallet buffer and Swift can commit or roll back another round, while both Rust calls return Ok. The fault flag then remains false and the later SyncHeightAdvanced is persisted past missing TXO or spend rows. Please serialize the complete FFI callback round globally and add a forced-interleaving regression.

packages/rs-platform-wallet/src/changeset/core_bridge.rs
```diff
@@ -68,6 +68,51 @@ where
...
+        // Symptoms (all one root cause): rows dropped entirely while the
+        // watermark advances (fresh scan persists nothing yet reports
+        // "scanned"); spent-markers lost so consumed outputs rehydrate
+        // as spendable (inflated balance); and — because the wallet's
+        // own `synced_height` is what gates a rescan — the wallet
+        // believes it is fully scanned and never re-matches, so the
+        // corruption is unrecoverable without deleting + recreating the
+        // wallet.
+        //
+        // Fix: once persistence has faulted this session (a broadcast
```
QuantumExplorer: This prevents a detected lag from advancing the durable cursor, but it does not guarantee the claimed self-healing on the next launch. The capacity-1000 producer remains fire-and-forget and ordinary transaction records are not restored into the in-memory map, so the same historical replay can exceed consumer throughput and lag again on every restart. Please use a lossless or backpressured persistence path, stop and rewind with throttling, or narrow this to a fail-closed mitigation and surface a hard sync fault.

packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandlerTest.kt
```diff
@@ -150,6 +150,57 @@ class PlatformWalletPersistenceHandlerTest {
...
     }
 
+    /**
+     * dashpay/platform#4069 (Kotlin half of signature C): the
+     * `syncedHeight` watermark is written by [onWalletChangesetHeader]
+     * into the SAME buffered transaction as the TXO/tx rows of its
+     * changeset. A round that rolls back (`success = false`) must
+     * therefore NOT advance the persisted watermark — otherwise the
+     * durable watermark could outrun the rows it implies, exactly the
+     * "empty-and-scanned after restart" corruption in #4069. Pins the
```
QuantumExplorer, line 160: This test validates the pre-existing single-threaded rollback behavior; no production Kotlin code changes in this PR, so the same test would pass on the base commit. It does not exercise RecvError::Lagged, a rejected store followed by a watermark, the stateful fault latch, or overlapping JNI store rounds. Please add adapter-level lag and store-error tests plus a concurrent FFI round test.

bfoss765:
@QuantumExplorer thanks for the review — all findings addressed. Pushed as `0556e58` (FFI persister lock) and `4313d84` (adapter follow-ups) on `fix/kotlin-sdk-txo-persistence`.

### 1. Serialize the entire FFI callback round under one lock
The P1 was correct: `FFIPersister::store` had no lock across begin → per-kind → end, so overlapping rounds from the core bridge / platform-address / shielded / DashPay writers could roll back or clobber each other's client transaction while both `store()` calls returned `Ok`, bypassing the fault latch.

- New per-persister lock held for the whole round: `round_lock: Mutex<RoundGuardState>` — `persistence.rs:668`, acquired at the top of `store()` and held through `end`.
- Nonzero `begin` is now fatal — the round aborts before any per-kind write instead of logging and continuing (`persistence.rs:730`).
- Defensive state machine `RoundGuardState::begin_round` / `end_round` (`persistence.rs:628`, `:641`; used at `:708`, `:1618`) rejects a nested begin / unmatched end as an **error, never a panic**, and never wedges the persister after a rejected round.

**One deliberate deviation from the ask:** the lock is a synchronous `parking_lot::Mutex`, not an async mutex. `store()` is a synchronous trait method invoked *blocking* from both async tasks and blocking FFI entry points, so a `tokio::sync::Mutex` can't be `.await`ed there and `blocking_lock()` panics inside a runtime. Serialization — not async yielding — is the requirement, and callers already block for the round's duration, so the sync mutex only adds waiting under genuine contention. Rationale is in the code comment at the lock site; happy to revisit if you'd prefer the trait go async.

Regression test: `concurrent_store_rounds_are_serialized` (`persistence.rs:5243`) drives two threads through the same persister with a probe begin/end callback that latches any interleave; it fails without the lock.

### 2. Subscribe to WalletEvents before publishing the manager
The receiver is now created synchronously in `PlatformWalletManager::new` *before* the manager is wrapped in the shared `Arc<RwLock>` and handed to producers (`manager/mod.rs:131`), then moved into the task. `spawn_wallet_event_adapter` takes the `Receiver` as a parameter instead of subscribing inside the task (`core_bridge.rs:111`). Test `events_emitted_before_task_poll_are_received` (`core_bridge.rs:870`) pins that an event emitted before the task polls is still delivered.

### 3. Lossy channel — narrowed claim + hard sync-fault signal
- (a) The guard doc is rewritten to the narrower claim: a restart resumes from the last durable watermark, but the capacity-1000 fire-and-forget producer means repeated overload keeps freezing. It's a fail-closed safety mitigation (the watermark never outruns its rows → funds never silently lost or inflated), **not** a guaranteed one-restart self-heal (`run_wallet_event_adapter` doc, `core_bridge.rs:197`).
- (b) Hard fault signal: the adapter raises an `Arc<AtomicBool>` the host polls via `PlatformWalletManager::sync_fault_detected` (`manager/mod.rs:236`) the first time it freezes a watermark, plus an error-level log. No `pendingIdentityKeys` FFI precedent exists on this branch, so this is the Rust-side flag + accessor + log per the fallback; wiring it through a dedicated FFI symbol can follow once a host wants to render it.

### 4. Per-wallet fault scoping (CodeRabbit)
`AdapterFaultState` (`core_bridge.rs:63`) replaces the task-global bool: `store()`-rejection faults only the named wallet (`fault_wallet`, `:78`); `Lagged` has no wallet id so it faults everyone via a global latch that also covers wallet ids first seen later (`fault_all`, `:85`); `is_faulted` (`:72`) is `global || per_wallet[id]`.

### 5. Tests the review demanded
- (a) concurrent-store serialization — `persistence.rs:5243`
- (b) real `RecvError::Lagged` through the loop → fault → subsequent watermark stripped — `lagged_broadcast_freezes_and_strips_subsequent_watermark` (`core_bridge.rs:726`)
- (c) rejected `store()` then watermark-only → not delivered — `rejected_store_freezes_wallet_and_strips_watermark` (`core_bridge.rs:781`)
- (d) record-bearing changesets still persist after the guard activates — `record_bearing_changesets_persist_after_guard_activates` (`core_bridge.rs:829`)
- (e) subscribe-before-publish — `events_emitted_before_task_poll_are_received` (`core_bridge.rs:870`)
- (f) nested-begin / unmatched-end rejection — `round_guard_state_machine_rejects_nesting_and_unmatched_end` and `nonzero_begin_aborts_the_round` (`persistence.rs:5341`, `:5303`)

`cargo test -p platform-wallet` and `-p platform-wallet-ffi` both green; `cargo check -p platform-wallet-ffi --features shielded` clean. Ready for re-review.

packages/rs-platform-wallet-ffi/src/persistence.rs
```diff
@@ -637,6 +690,23 @@ impl PlatformWalletPersistence for FFIPersister {
...
+        // mutex cannot be `.await`ed here and `blocking_lock()` panics
+        // inside a runtime. Serialization — not async yielding — is the
+        // requirement; callers already block for the round's duration, so
+        // the sync mutex only adds waiting under genuine round contention.
+        let mut round = self.round_lock.lock();
+
+        // Open the round on the Rust side (rejects a nested begin / an
+        // unclean round left open by a prior unwind — error, never
+        // panic). Matched 1:1 with the `round.end_round()` below.
+        round.begin_round()?;
```
thepastaclaw, line 708: **🟡 Suggestion: A panic between begin_round and end_round permanently wedges the shared FFIPersister for the process**

The new round_lock is a `parking_lot::Mutex` (persistence.rs:18, 668), which — unlike `std::sync::Mutex` — does not poison on unwind. `store()` acquires it (703), calls `begin_round()` which sets `in_round=true` (708), runs the ~900-line per-kind FFI callback sequence, and only clears `in_round` via `end_round()` at line 1618. If any code in that window unwinds (an allocation failure, a `CString::new` on interior-NUL data, or a future panic in per-kind marshalling), the guard's Drop releases the mutex but leaves `in_round=true`. Every subsequent `store()` on this `Arc<FFIPersister>` — shared by the core-changeset bridge, platform-address sync, shielded sync, and DashPay tasks — then hits `begin_round()`, sees `in_round==true`, and is rejected as 'round already open' for the rest of the process; the persister never self-heals, so a single transient unwind degrades into a whole-session outage requiring a wallet restart.

Worth noting: because round_lock fully serializes rounds, only one thread is ever inside the begin→end window at a time, so `begin_round()` can realistically only observe `in_round==true` as the residue of a prior unwind — the 'concurrent nested begin' the guard nominally defends against cannot occur under the lock (a same-thread reentrant host callback would deadlock on the non-reentrant mutex before reaching `begin_round`). In other words the guard's only reachable trigger is the stale-unwind case, which it then refuses to recover from.

This is fail-closed rather than corrupting — a rejected store propagates up, the core_bridge adapter faults the wallet and freezes its watermark, so the worst case is a rescan on next launch, not lost or inflated funds — which is why it is a suggestion. It is also a genuine regression relative to the pre-PR code, which had no round lock and therefore could not wedge. The agents verified there are no `?`/`unwrap`/`expect` early returns in the window today (all failure paths set `round_success=false` and fall through to `end_round`), so this is primarily defense against a future edit or a hard panic. A small RAII drop-guard that resets `in_round` on Drop (or a `begin_round` that treats an already-open state as a recoverable stale round — logging at error! and resetting, since a prior unwind means that host transaction is already gone) would still serialize rounds and reject genuine reentrancy while letting the persister recover on the next round after a transient unwind.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Code Review

Source: reviewers: claude `opus` general (ok), codex `gpt-5.6-sol` general (failed: revoked refresh token), claude `opus` security-auditor (ok), codex `gpt-5.6-sol` security-auditor (failed: revoked refresh token), claude `opus` rust-quality (ok), codex `gpt-5.6-sol` rust-quality (failed: revoked refresh token), claude `opus` ffi-engineer (ok), codex `gpt-5.6-sol` ffi-engineer (failed: revoked refresh token). Verifier: claude `opus`.

Prior-finding reconciliation (`c5a03580` -> `4313d843`):
- FIXED: the task-global store-fault latch is now scoped per wallet, while wallet-id-less broadcast lag remains global.
- FIXED: four async adapter-loop tests now cover real lag, rejected store, post-fault record persistence, and subscribe-before-publish delivery.

Carried-forward prior findings: none.

New findings in the latest delta: one non-blocking suggestion. A panic after `begin_round()` can leave the non-poisoning FFI round state latched open, causing every later store in the process to fail closed until restart. The cumulative PR otherwise closes the startup-subscription and concurrent FFI-round persistence gaps without introducing a blocking defect.

Validation: `git diff --check c5a03580..4313d843`; 7/7 focused `platform-wallet` core-bridge tests; 3/3 focused `platform-wallet-ffi` round-state/concurrency tests.

🟡 1 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet-ffi/src/persistence.rs:703-708: A panic between begin_round and end_round permanently wedges the shared FFIPersister for the process
  The new round_lock is a `parking_lot::Mutex` (persistence.rs:18, 668), which — unlike `std::sync::Mutex` — does not poison on unwind. `store()` acquires it (703), calls `begin_round()` which sets `in_round=true` (708), runs the ~900-line per-kind FFI callback sequence, and only clears `in_round` via `end_round()` at line 1618. If any code in that window unwinds (an allocation failure, a `CString::new` on interior-NUL data, or a future panic in per-kind marshalling), the guard's Drop releases the mutex but leaves `in_round=true`. Every subsequent `store()` on this `Arc<FFIPersister>` — shared by the core-changeset bridge, platform-address sync, shielded sync, and DashPay tasks — then hits `begin_round()`, sees `in_round==true`, and is rejected as 'round already open' for the rest of the process; the persister never self-heals, so a single transient unwind degrades into a whole-session outage requiring a wallet restart.

  Worth noting: because round_lock fully serializes rounds, only one thread is ever inside the begin→end window at a time, so `begin_round()` can realistically only observe `in_round==true` as the residue of a prior unwind — the 'concurrent nested begin' the guard nominally defends against cannot occur under the lock (a same-thread reentrant host callback would deadlock on the non-reentrant mutex before reaching `begin_round`). In other words the guard's only reachable trigger is the stale-unwind case, which it then refuses to recover from.

  This is fail-closed rather than corrupting — a rejected store propagates up, the core_bridge adapter faults the wallet and freezes its watermark, so the worst case is a rescan on next launch, not lost or inflated funds — which is why it is a suggestion. It is also a genuine regression relative to the pre-PR code, which had no round lock and therefore could not wedge. The agents verified there are no `?`/`unwrap`/`expect` early returns in the window today (all failure paths set `round_success=false` and fall through to `end_round`), so this is primarily defense against a future edit or a hard panic. A small RAII drop-guard that resets `in_round` on Drop (or a `begin_round` that treats an already-open state as a recoverable stale round — logging at error! and resetting, since a prior unwind means that host transaction is already gone) would still serialize rounds and reject genuine reentrancy while letting the persister recover on the next round after a transient unwind.
```
</details>

packages/rs-platform-wallet/src/changeset/core_bridge.rs
thepastaclaw: ✅ Fixed in `4313d843`: `AdapterFaultState` now scopes rejected `store()` rounds to the affected wallet while keeping wallet-id-less broadcast lag global. `fault_state_scopes_store_rejection_per_wallet_and_lag_globally` covers the split.

Issue #4089, opened 2026-07-10
[kotlin-sdk] Split build/broadcast with reservation release for BIP70-style deferred submission


bfoss765:
**Priority: P0 — hard blocker for the BIP70/BIP270 (CTX/DashSpend) send cutover; structurally impossible on `sendToAddresses` (those protocols sign, POST to a merchant server, and broadcast only on ack). Implementation will be driven by @bfoss765 — this issue tracks design agreement.**

**Ask:** expose the existing internal build/broadcast split with an explicit reservation lifecycle:
```kotlin
data class SignedCoreTransaction(val txidHex: String, val rawTxBytes: ByteArray, val feeDuffs: Long, val reservationToken: Long)
suspend fun ManagedPlatformWallet.buildSignedPayment(recipients, network, coreSignerHandle, accountType, accountIndex, options): SignedCoreTransaction
suspend fun ManagedPlatformWallet.broadcastSigned(token: Long): String
suspend fun ManagedPlatformWallet.releaseReservation(token: Long)  // payment abandoned / server nack
```
The internal `new → setFunding → buildSigned` already produces exactly this (UTXOs reserved at `buildSigned`); the manager remains the sole driver so the `CoreTransactionBuilder` KDoc's safety argument is preserved.

Context: dashpay/dash-wallet#1507 Phase 5c GAP-4. cc @quantumexplorer

bfoss765:
Closing as shipped. The asked-for surface exists on `v4.2-dev` and is what the Android BIP70 flow runs in production QA builds today:

- **Build-only + explicit reservation lifecycle**: `CoreWallet` finalize produces a `SignedCoreTransaction` (txid, raw bytes, fee, reservation token + height) without touching the network.
- **Deferred broadcast**: `core_wallet_signed_payment_broadcast` submits a previously-finalized payment via the `SignedPaymentRegistry`, with generation-liveness and reservation-age guards (#4185's successor work, plus the age-guard hardening in #4309).
- **Release path**: abandon/release is owner-token-guarded, so a merchant-nack (the BIP70 "signed, POSTed, not acked" case) frees the inputs for an immediate rebuild without being able to free a newer build's reservation.

Field evidence: the wallet's scanned-invoice flow (`PaymentProtocolFragment` → `l1DeferredBuild` → merchant ACK → `l1DeferredBroadcast`) has been running against this surface on the 11.10.x QA line, including a live nack/reservation-release exercise via the local BIP70 test server.

[...]

PR #4090, opened 2026-07-10
feat(kotlin-sdk): split build/broadcast with reservation release for BIP70-style deferred submission

bfoss765:
Implements #4089 (design agreed there). Context: dashpay/dash-wallet#1507 Phase 5c GAP-4.

## Problem

BIP70/BIP270 (CTX/DashSpend) sends **sign, POST the raw bytes to a merchant server, and broadcast only on the server's ack** — a flow that is structurally impossible on the one-shot `ManagedPlatformWallet.sendToAddresses` (which builds, signs, and broadcasts in one uninterrupted step). This exposes the existing internal build/broadcast split with an explicit reservation lifecycle, while keeping `CoreTransactionBuilder` internal so the manager remains the sole driver of the `setFunding`/`buildSigned` selection race (its KDoc safety argument is preserved).

## API (purely ADDITIVE — no existing signature changes)

Kotlin, on `ManagedPlatformWallet`:

```kotlin
class SignedCoreTransaction(val txidHex, val rawTxBytes, val feeDuffs, val reservationToken: Long)

suspend fun buildSignedPayment(recipients, network, coreSignerHandle, accountType, accountIndex): SignedCoreTransaction
suspend fun broadcastSigned(token: Long): String   // merchant server acked
suspend fun releaseReservation(token: Long)         // abandoned / server nacked
```

`buildSignedPayment` runs `new → addOutput* → setFunding → buildSigned` under the **same per-wallet `coreSendMutex`** as `sendToAddresses`, closing the double-selection window; `buildSigned` reserves the UTXOs, so `broadcastSigned` / `releaseReservation` operate on the token **later, without the mutex**.

## Token / reservation design

- A new generic `SignedPaymentRegistry` (in `rs-platform-wallet`) owns the built tx + its held reservation between build and submission, keyed by an opaque `ReservationToken`. The FFI layer instantiates **one process-global registry pinned to `SpvBroadcaster`**.
- `broadcast` removes the entry **before** sending, so a repeated/concurrent broadcast of the same token gets a typed `StaleToken` error — never a double-broadcast. On failure it reuses the existing release-on-rejection / keep-on-ambiguous policy.
- Each token is **bound to its originating wallet instance** (`Arc::ptr_eq` on the shared `WalletManager`), so a re-created wallet is rejected rather than spending against stale state.
- `release` is **idempotent** (unknown / already-consumed token → silent no-op) and needs no handle.
- **Process death:** the registry and the underlying `ReservationSet` are both in-memory, so an app crash between build and broadcast drops the entry **and** the reservation together — nothing leaks across a restart. This matches dashj's memory-only in-flight reservations; there is no on-disk reservation persistence to follow.

### Surface additions

- FFI (`platform-wallet-ffi`): `core_wallet_transaction_get_bytes`, `core_wallet_signed_payment_register`, `core_wallet_signed_payment_broadcast`, `core_wallet_signed_payment_release`; new `ErrorStaleReservationToken` (22).
- JNI (`rs-unified-sdk-jni`): `coreTransactionGetBytes`, `coreWalletRegisterSignedPayment`, `coreWalletBroadcastSignedPayment`, `coreWalletReleaseSignedPayment`.
- Kotlin: `SignedCoreTransaction`, the three methods above, and `DashSdkError.PlatformWallet.StaleReservationToken`.

The immediate `sendToAddresses` / `coreWalletBroadcastTransaction` path is unchanged.

## Tests

Rust (`SignedPaymentRegistry`, 10 tests, all green):
- build → broadcast happy path (broadcast bytes byte-identical to the registered tx; returned txid matches)
- build → release → funds spendable again (a rebuild reselects the released UTXO; both BIP44 and BIP32 arms)
- double-broadcast → `StaleToken` (network hit exactly once)
- double-release → idempotent; broadcast-after-release → `StaleToken`; unknown token → `StaleToken`
- token from a different wallet instance → `WalletMismatch` (nothing sent)
- ambiguous broadcast keeps the reservation and still consumes the token
- concurrent broadcasts serialize to exactly one send; concurrent registers yield distinct tokens

Full `platform-wallet` / `platform-wallet-ffi` / `rs-unified-sdk-jni` suites pass; clippy green. Kotlin: added `DashSdkErrorTest` coverage for the code-22 mapping; `:sdk:compileReleaseKotlin` green. The Kotlin/JNI native path is exercised only by instrumented tests, so per convention Rust coverage + compile stands in there. The native library was **not** rebuilt.

cc @quantumexplorer

### DIFF
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedCoreWallet.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedCoreWallet.kt
@@ -55,6 +55,45 @@ class ManagedCoreWallet internal constructor(handle: Long) : AutoCloseable {
         )
     }
 
+    /**
+     * Register a built+signed [tx] for deferred (BIP70/BIP270) submission,
+     * holding its UTXO reservation, and return the resulting
+     * [ManagedPlatformWallet.SignedCoreTransaction]. Does NOT consume [tx] — the
+     * caller still closes it. Reads the raw bytes off [tx] and decodes the
+     * register BLOB (`token, feeDuffs, txid`).
+     */
+    internal fun registerSignedPayment(
+        tx: CoreTransaction,
+    ): ManagedPlatformWallet.SignedCoreTransaction {
+        val rawTxBytes = WalletManagerNative.coreTransactionGetBytes(tx.handle)
+        val blob = WalletManagerNative.coreWalletRegisterSignedPayment(
+            handle,
+            tx.handle,
+            tx.accountType.ffiValue,
+            tx.accountIndex,
+        )
+        val buffer = java.nio.ByteBuffer.wrap(blob) // big-endian by default
+        val token = buffer.long
+        val feeDuffs = buffer.long
+        val txidLen = buffer.int
+        val txidBytes = ByteArray(txidLen)
+        buffer.get(txidBytes)
+        return ManagedPlatformWallet.SignedCoreTransaction(
+            txidHex = String(txidBytes, Charsets.UTF_8),
+            rawTxBytes = rawTxBytes,
+            feeDuffs = feeDuffs,
+            reservationToken = token,
+        )
+    }
+
+    /**
+     * Broadcast the deferred payment behind [token] and return its txid. A
+     * stale / already-broadcast / wrong-wallet token surfaces as
+     * [org.dashfoundation.dashsdk.errors.DashSdkError.PlatformWallet.StaleReservationToken].
+     */
+    internal fun broadcastSignedPayment(token: Long): String =
+        WalletManagerNative.coreWalletBroadcastSignedPayment(handle, token)
+
     override fun close() {
         cleanable.clean()
     }
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedCoreWallet.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedCoreWallet.kt
@@ -71,21 +71,7 @@ class ManagedCoreWallet internal constructor(handle: Long) : AutoCloseable {
             tx.accountType.ffiValue,
             tx.accountIndex,
         )
-        val buffer = java.nio.ByteBuffer.wrap(blob) // big-endian by default
-        val token = buffer.long
-        val feeDuffs = buffer.long
-        val txidLen = buffer.int
-        val txidBytes = ByteArray(txidLen)
-        buffer.get(txidBytes)
-        val txBytesLen = buffer.int
-        val rawTxBytes = ByteArray(txBytesLen)
-        buffer.get(rawTxBytes)
-        return ManagedPlatformWallet.SignedCoreTransaction(
-            txidHex = String(txidBytes, Charsets.UTF_8),
-            rawTxBytes = rawTxBytes,
-            feeDuffs = feeDuffs,
-            reservationToken = token,
-        )
+        return ManagedPlatformWallet.SignedCoreTransaction.fromRegisterBlob(blob)
     }
 
     /**
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedCoreWallet.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedCoreWallet.kt
@@ -59,13 +59,12 @@ class ManagedCoreWallet internal constructor(handle: Long) : AutoCloseable {
      * Register a built+signed [tx] for deferred (BIP70/BIP270) submission,
      * holding its UTXO reservation, and return the resulting
      * [ManagedPlatformWallet.SignedCoreTransaction]. Does NOT consume [tx] — the
-     * caller still closes it. Reads the raw bytes off [tx] and decodes the
-     * register BLOB (`token, feeDuffs, txid`).
+     * caller still closes it. Decodes the single register BLOB
+     * (`token, feeDuffs, txid, rawTxBytes`) — one native round trip.
      */
     internal fun registerSignedPayment(
         tx: CoreTransaction,
     ): ManagedPlatformWallet.SignedCoreTransaction {
-        val rawTxBytes = WalletManagerNative.coreTransactionGetBytes(tx.handle)
         val blob = WalletManagerNative.coreWalletRegisterSignedPayment(
             handle,
             tx.handle,
@@ -78,6 +77,9 @@ class ManagedCoreWallet internal constructor(handle: Long) : AutoCloseable {
         val txidLen = buffer.int
         val txidBytes = ByteArray(txidLen)
         buffer.get(txidBytes)
+        val txBytesLen = buffer.int
+        val rawTxBytes = ByteArray(txBytesLen)
+        buffer.get(rawTxBytes)
         return ManagedPlatformWallet.SignedCoreTransaction(
             txidHex = String(txidBytes, Charsets.UTF_8),
             rawTxBytes = rawTxBytes,
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
@@ -173,6 +173,143 @@ class ManagedPlatformWallet internal constructor(
     }
 
     /**
+     * A built, signed Core transaction whose funding UTXOs are reserved,
+     * awaiting a deferred [broadcastSigned] or [releaseReservation] — the
+     * split-out result of [buildSignedPayment] for BIP70/BIP270 (CTX/DashSpend)
+     * flows that must sign now, POST the raw bytes to a merchant server, and
+     * broadcast only on the server's ack.
+     *
+     * @property txidHex the transaction id (lowercase hex) the broadcast will
+     *   return — computed from the signed bytes Rust-side so it matches exactly.
+     * @property rawTxBytes the consensus-serialized signed transaction, to hand
+     *   to the merchant server.
+     * @property feeDuffs the fee the build charged, in duffs.
+     * @property reservationToken the opaque token for [broadcastSigned] /
+     *   [releaseReservation]. Valid only for this wallet instance and only until
+     *   consumed by one of those calls.
+     */
+    class SignedCoreTransaction internal constructor(
+        val txidHex: String,
+        val rawTxBytes: ByteArray,
+        val feeDuffs: Long,
+        val reservationToken: Long,
+    ) {
+        override fun equals(other: Any?): Boolean =
+            other is SignedCoreTransaction &&
+                txidHex == other.txidHex &&
+                rawTxBytes.contentEquals(other.rawTxBytes) &&
+                feeDuffs == other.feeDuffs &&
+                reservationToken == other.reservationToken
+
+        override fun hashCode(): Int {
+            var result = txidHex.hashCode()
+            result = 31 * result + rawTxBytes.contentHashCode()
+            result = 31 * result + feeDuffs.hashCode()
+            result = 31 * result + reservationToken.hashCode()
+            return result
+        }
+
+        override fun toString(): String =
+            "SignedCoreTransaction(txidHex=$txidHex, feeDuffs=$feeDuffs, " +
+                "reservationToken=$reservationToken, rawTxBytes=${rawTxBytes.size} bytes)"
+    }
+
+    /**
+     * Build and sign a Core payment to [recipients] WITHOUT broadcasting,
+     * reserving the funding UTXOs and returning a [SignedCoreTransaction] whose
+     * [SignedCoreTransaction.reservationToken] later drives [broadcastSigned]
+     * (server acked) or [releaseReservation] (abandoned / server nacked).
+     *
+     * The BIP70/BIP270 counterpart to [sendToAddresses]: those protocols sign,
+     * POST the raw bytes to a merchant server, and broadcast only on ack, which
+     * a single build-sign-broadcast call cannot express. The `new → addOutput* →
+     * setFunding → buildSigned` build runs under the same per-wallet
+     * [coreSendMutex] as [sendToAddresses] (closing the setFunding/buildSigned
+     * selection race); [buildSigned] reserves the selected UTXOs, so once this
+     * returns the reservation holds the inputs and [broadcastSigned] /
+     * [releaseReservation] operate on the token later WITHOUT the mutex.
+     *
+     * Process-death note: the reservation is in-memory. An app crash between
+     * this call and [broadcastSigned] drops the reservation on restart (the
+     * UTXOs become spendable again) — the same property dashj has.
+     *
+     * @param network the wallet network — see [sendToAddresses].
+     * @param coreSignerHandle the manager's `MnemonicResolverHandle` — see
+     *   [sendToAddresses]. No private key crosses the boundary.
+     */
+    suspend fun buildSignedPayment(
+        recipients: List<Pair<String, Long>>,
+        network: org.dashfoundation.dashsdk.Network,
+        coreSignerHandle: Long,
+        accountType: AccountType = AccountType.BIP44,
+        accountIndex: Int = 0,
+    ): SignedCoreTransaction = withContext(Dispatchers.IO) {
+        require(accountIndex >= 0) { "accountIndex must be non-negative, got $accountIndex" }
+        require(recipients.isNotEmpty()) { "recipients must not be empty" }
+        require(recipients.all { it.second > 0 }) {
+            "every recipient amount must be positive"
+        }
+        val builderAccountType = when (accountType) {
+            AccountType.BIP44 -> CoreTransactionBuilder.AccountType.BIP44
+            AccountType.BIP32 -> CoreTransactionBuilder.AccountType.BIP32
+        }
+        coreSendMutex.withLock {
+            mapNativeErrors {
+                coreWallet().use { core ->
+                    val builder = CoreTransactionBuilder(network)
+                    // `buildSigned` consumes the builder; `use` still safely
+                    // destroys it on the pre-build failure paths.
+                    val signedTx = builder.use {
+                        for ((address, amount) in recipients) {
+                            it.addOutput(address, amount)
+                        }
+                        it.setFunding(this@ManagedPlatformWallet, builderAccountType, accountIndex)
+                        it.buildSigned(
+                            this@ManagedPlatformWallet,
+                            builderAccountType,
+                            accountIndex,
+                            coreSignerHandle,
+                        )
+                    }
+                    // Register the signed tx (holding its reservation) before the
+                    // native transaction is freed; `use` frees it afterward.
+                    signedTx.use { tx -> core.registerSignedPayment(tx) }
+                }
+            }
+        }
+    }
+
+    /**
+     * Broadcast the deferred payment behind [token] (from [buildSignedPayment])
+     * and return its broadcast txid — the "merchant server acked" arm. Consumes
+     * the token: a second [broadcastSigned] with the same token, or one for a
+     * re-created wallet, throws
+     * [org.dashfoundation.dashsdk.errors.DashSdkError.PlatformWallet.StaleReservationToken]
+     * rather than double-broadcasting. Operates on the token WITHOUT the
+     * [coreSendMutex] (the inputs are already reserved).
+     */
+    suspend fun broadcastSigned(token: Long): String = withContext(Dispatchers.IO) {
+        mapNativeErrors {
+            coreWallet().use { core -> core.broadcastSignedPayment(token) }
+        }
+    }
+
+    /**
+     * Release the funding reservation behind [token] (from [buildSignedPayment])
+     * — the "payment abandoned / merchant server nacked" arm — returning the
+     * reserved UTXOs to spendable. Idempotent: releasing an unknown /
+     * already-broadcast / already-released token is a silent no-op, so it is
+     * always safe to call defensively.
+     */
+    suspend fun releaseReservation(token: Long) {
+        withContext(Dispatchers.IO) {
+            mapNativeErrors {
+                WalletManagerNative.coreWalletReleaseSignedPayment(token)
+            }
+        }
+    }
+
+    /**
      * The wallet's Platform-payment addresses that currently hold credits,
      * each as a [FundingInput] whose `credits` is the full cached balance —
      * the funding-input candidates for an identity top-up. Port of the
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
@@ -212,6 +212,32 @@ class ManagedPlatformWallet internal constructor(
         override fun toString(): String =
             "SignedCoreTransaction(txidHex=$txidHex, feeDuffs=$feeDuffs, " +
                 "reservationToken=$reservationToken, rawTxBytes=${rawTxBytes.size} bytes)"
+
+        internal companion object {
+            /**
+             * Decode the big-endian native BLOB the deferred build/register FFI
+             * returns: `u64 token, u64 feeDuffs, u32 txidLen, txid utf8,
+             * u32 txBytesLen, txBytes`. Shared by the atomic
+             * finalize-and-register path and the deprecated register path.
+             */
+            internal fun fromRegisterBlob(blob: ByteArray): SignedCoreTransaction {
+                val buffer = java.nio.ByteBuffer.wrap(blob) // big-endian by default
+                val token = buffer.long
+                val feeDuffs = buffer.long
+                val txidLen = buffer.int
+                val txidBytes = ByteArray(txidLen)
+                buffer.get(txidBytes)
+                val txBytesLen = buffer.int
+                val rawTxBytes = ByteArray(txBytesLen)
+                buffer.get(rawTxBytes)
+                return SignedCoreTransaction(
+                    txidHex = String(txidBytes, Charsets.UTF_8),
+                    rawTxBytes = rawTxBytes,
+                    feeDuffs = feeDuffs,
+                    reservationToken = token,
+                )
+            }
+        }
     }
 
     /**
@@ -255,25 +281,24 @@ class ManagedPlatformWallet internal constructor(
             AccountType.BIP32 -> CoreTransactionBuilder.AccountType.BIP32
         }
         mapNativeErrors {
-            coreWallet().use { core ->
-                val builder = CoreTransactionBuilder(network)
-                // `buildSigned` consumes the builder; `use` still safely
-                // destroys it on the pre-build failure paths.
-                val signedTx = builder.use {
-                    for ((address, amount) in recipients) {
-                        it.addOutput(address, amount)
-                    }
-                    it.setFunding(this@ManagedPlatformWallet, builderAccountType, accountIndex)
-                    it.buildSigned(
-                        this@ManagedPlatformWallet,
-                        builderAccountType,
-                        accountIndex,
-                        coreSignerHandle,
-                    )
+            // One atomic native operation: select + reserve + sign + register.
+            // `finalizeSignedPayment` consumes the builder on every path, so
+            // `use` only needs to destroy it on the pre-finalize failure paths
+            // (adding outputs). Selection and reservation commit as a single unit
+            // under the wallet-manager lock, so a concurrent deferred build — or a
+            // deferred build racing an immediate send — can no longer double-
+            // select the same input, restoring the atomicity the removed Kotlin
+            // per-wallet send mutex used to provide.
+            CoreTransactionBuilder(network).use { builder ->
+                for ((address, amount) in recipients) {
+                    builder.addOutput(address, amount)
                 }
-                // Register the signed tx (holding its reservation) before the
-                // native transaction is freed; `use` frees it afterward.
-                signedTx.use { tx -> core.registerSignedPayment(tx) }
+                builder.finalizeSignedPayment(
+                    this@ManagedPlatformWallet,
+                    builderAccountType,
+                    accountIndex,
+                    coreSignerHandle,
+                )
             }
         }
     }
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
@@ -223,11 +223,12 @@ class ManagedPlatformWallet internal constructor(
      * The BIP70/BIP270 counterpart to [sendToAddresses]: those protocols sign,
      * POST the raw bytes to a merchant server, and broadcast only on ack, which
      * a single build-sign-broadcast call cannot express. The `new → addOutput* →
-     * setFunding → buildSigned` build runs under the same per-wallet
-     * [coreSendMutex] as [sendToAddresses] (closing the setFunding/buildSigned
-     * selection race); [buildSigned] reserves the selected UTXOs, so once this
-     * returns the reservation holds the inputs and [broadcastSigned] /
-     * [releaseReservation] operate on the token later WITHOUT the mutex.
+     * setFunding → buildSigned` build runs under the same per-wallet teardown
+     * gate ([gate]) as [sendToAddresses]; [buildSigned] atomically reserves the
+     * selected UTXOs in the Rust reservation layer (which closes the
+     * setFunding/buildSigned selection race), so once this returns the
+     * reservation holds the inputs and [broadcastSigned] / [releaseReservation]
+     * operate on the token later.
      *
      * Process-death note: the reservation is in-memory. An app crash between
      * this call and [broadcastSigned] drops the reservation on restart (the
@@ -243,7 +244,7 @@ class ManagedPlatformWallet internal constructor(
         coreSignerHandle: Long,
         accountType: AccountType = AccountType.BIP44,
         accountIndex: Int = 0,
-    ): SignedCoreTransaction = withContext(Dispatchers.IO) {
+    ): SignedCoreTransaction = gate.op {
         require(accountIndex >= 0) { "accountIndex must be non-negative, got $accountIndex" }
         require(recipients.isNotEmpty()) { "recipients must not be empty" }
         require(recipients.all { it.second > 0 }) {
@@ -253,28 +254,26 @@ class ManagedPlatformWallet internal constructor(
             AccountType.BIP44 -> CoreTransactionBuilder.AccountType.BIP44
             AccountType.BIP32 -> CoreTransactionBuilder.AccountType.BIP32
         }
-        coreSendMutex.withLock {
-            mapNativeErrors {
-                coreWallet().use { core ->
-                    val builder = CoreTransactionBuilder(network)
-                    // `buildSigned` consumes the builder; `use` still safely
-                    // destroys it on the pre-build failure paths.
-                    val signedTx = builder.use {
-                        for ((address, amount) in recipients) {
-                            it.addOutput(address, amount)
-                        }
-                        it.setFunding(this@ManagedPlatformWallet, builderAccountType, accountIndex)
-                        it.buildSigned(
-                            this@ManagedPlatformWallet,
-                            builderAccountType,
-                            accountIndex,
-                            coreSignerHandle,
-                        )
+        mapNativeErrors {
+            coreWallet().use { core ->
+                val builder = CoreTransactionBuilder(network)
+                // `buildSigned` consumes the builder; `use` still safely
+                // destroys it on the pre-build failure paths.
+                val signedTx = builder.use {
+                    for ((address, amount) in recipients) {
+                        it.addOutput(address, amount)
                     }
-                    // Register the signed tx (holding its reservation) before the
-                    // native transaction is freed; `use` frees it afterward.
-                    signedTx.use { tx -> core.registerSignedPayment(tx) }
+                    it.setFunding(this@ManagedPlatformWallet, builderAccountType, accountIndex)
+                    it.buildSigned(
+                        this@ManagedPlatformWallet,
+                        builderAccountType,
+                        accountIndex,
+                        coreSignerHandle,
+                    )
                 }
+                // Register the signed tx (holding its reservation) before the
+                // native transaction is freed; `use` frees it afterward.
+                signedTx.use { tx -> core.registerSignedPayment(tx) }
             }
         }
     }
@@ -285,8 +284,8 @@ class ManagedPlatformWallet internal constructor(
      * the token: a second [broadcastSigned] with the same token, or one for a
      * re-created wallet, throws
      * [org.dashfoundation.dashsdk.errors.DashSdkError.PlatformWallet.StaleReservationToken]
-     * rather than double-broadcasting. Operates on the token WITHOUT the
-     * [coreSendMutex] (the inputs are already reserved).
+     * rather than double-broadcasting. Operates on the token directly (the
+     * inputs are already reserved).
      */
     suspend fun broadcastSigned(token: Long): String = withContext(Dispatchers.IO) {
         mapNativeErrors {
--- a/packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/errors/DashSdkErrorTest.kt
+++ b/packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/errors/DashSdkErrorTest.kt
@@ -103,6 +103,16 @@ class DashSdkErrorTest {
         )
         // The message must warn against retrying (distinct from the anchor case).
         assertTrue(broadcastUnconfirmed.message!!.contains("do NOT retry"))
+
+        // Deferred build/broadcast: a stale/consumed/wrong-wallet reservation
+        // token → typed StaleReservationToken, not retryable.
+        val staleToken = DashSdkError.fromNative(DashSDKException(offset + 22, "stale token 7"))
+        assertTrue(staleToken is DashSdkError.PlatformWallet.StaleReservationToken)
+        assertFalse(
+            "StaleReservationToken must NOT be retryable (rebuild the payment)",
+            staleToken.isRetryable,
+        )
+        assertEquals("stale token 7", staleToken.message)
     }
 
     @Test
--- a/packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/errors/DashSdkErrorTest.kt
+++ b/packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/errors/DashSdkErrorTest.kt
@@ -106,7 +106,7 @@ class DashSdkErrorTest {
 
         // Deferred build/broadcast: a stale/consumed/wrong-wallet reservation
         // token → typed StaleReservationToken, not retryable.
-        val staleToken = DashSdkError.fromNative(DashSDKException(offset + 22, "stale token 7"))
+        val staleToken = DashSdkError.fromNative(DashSDKException(offset + 26, "stale token 7"))
         assertTrue(staleToken is DashSdkError.PlatformWallet.StaleReservationToken)
         assertFalse(
             "StaleReservationToken must NOT be retryable (rebuild the payment)",
--- a/packages/rs-platform-wallet-ffi/src/core_wallet/signed_payment.rs
+++ b/packages/rs-platform-wallet-ffi/src/core_wallet/signed_payment.rs
@@ -30,36 +30,8 @@ use std::os::raw::c_char;
 /// [`ReservationToken`]. In-memory only: an app crash between build and
 /// broadcast drops the registry entry and the underlying UTXO reservation
 /// together, so nothing leaks across a restart.
-static SIGNED_PAYMENT_REGISTRY: Lazy<SignedPaymentRegistry<SpvBroadcaster>> =
+pub(crate) static SIGNED_PAYMENT_REGISTRY: Lazy<SignedPaymentRegistry<SpvBroadcaster>> =
     Lazy::new(SignedPaymentRegistry::new);
-
-/// Borrow the consensus-serialized bytes of a transaction built by
-/// `core_wallet_tx_builder_build_signed`, for the caller to copy into
-/// `SignedCoreTransaction.rawTxBytes`.
-///
-/// The written pointer borrows the `FFICoreTransaction`'s own buffer — it is
-/// valid only until the transaction is freed with
-/// `core_wallet_transaction_free`, so the caller must copy the bytes out
-/// immediately and must not retain the pointer.
-///
-/// # Safety
-/// `tx` must be a valid, non-freed `FFICoreTransaction` pointer;
-/// `out_ptr`/`out_len` must be writable.
-#[no_mangle]
-pub unsafe extern "C" fn core_wallet_transaction_get_bytes(
-    tx: *const FFICoreTransaction,
-    out_ptr: *mut *const u8,
-    out_len: *mut usize,
-) -> PlatformWalletFFIResult {
-    check_ptr!(tx);
-    check_ptr!(out_ptr);
-    check_ptr!(out_len);
-
-    let bytes = (*tx).bytes();
-    *out_ptr = bytes.as_ptr();
-    *out_len = bytes.len();
-    PlatformWalletFFIResult::ok()
-}
 
 /// Register a built, signed transaction for deferred submission and return a
 /// reservation token.
@@ -73,13 +45,20 @@ pub unsafe extern "C" fn core_wallet_transaction_get_bytes(
 ///
 /// `account_type`/`account_index` identify the funding account handed to
 /// `set_funding`, so the reservation can be released on rejection/abandonment.
-/// Writes `out_token`, `out_fee` (the build's fee in duffs), and `out_txid` (a
+/// Writes `out_token`, `out_fee` (the build's fee in duffs), `out_txid` (a
 /// heap-allocated lowercase-hex C string the caller frees with
-/// `core_wallet_free_address`).
+/// `core_wallet_free_address`), and `out_bytes_ptr`/`out_bytes_len` (the
+/// consensus-serialized transaction bytes, returned in the same call so the
+/// caller needs no second native round trip).
+///
+/// The `out_bytes_ptr` buffer borrows the `FFICoreTransaction`'s own storage —
+/// it is valid only until `tx` is freed with `core_wallet_transaction_free`, so
+/// the caller must copy the bytes out immediately and must not retain the
+/// pointer.
 ///
 /// # Safety
 /// `tx` must be a valid, non-freed `FFICoreTransaction`; `core_handle` a valid
-/// core-wallet handle; the three out-pointers must be writable.
+/// core-wallet handle; all out-pointers must be writable.
 #[no_mangle]
 pub unsafe extern "C" fn core_wallet_signed_payment_register(
     core_handle: Handle,
@@ -89,15 +68,20 @@ pub unsafe extern "C" fn core_wallet_signed_payment_register(
     out_token: *mut u64,
     out_fee: *mut u64,
     out_txid: *mut *mut c_char,
+    out_bytes_ptr: *mut *const u8,
+    out_bytes_len: *mut usize,
 ) -> PlatformWalletFFIResult {
     check_ptr!(tx);
     check_ptr!(out_token);
     check_ptr!(out_fee);
     check_ptr!(out_txid);
+    check_ptr!(out_bytes_ptr);
+    check_ptr!(out_bytes_len);
 
     let core = unwrap_option_or_return!(CORE_WALLET_STORAGE.with_item(core_handle, |w| w.clone()));
 
-    let transaction: dashcore::Transaction = match dashcore::consensus::deserialize((*tx).bytes()) {
+    let bytes = (*tx).bytes();
+    let transaction: dashcore::Transaction = match dashcore::consensus::deserialize(bytes) {
         Ok(t) => t,
         Err(e) => {
             return PlatformWalletFFIResult::err(
@@ -109,14 +93,10 @@ pub unsafe extern "C" fn core_wallet_signed_payment_register(
     let txid = transaction.txid();
     let fee = (*tx).fee();
 
-    let token = SIGNED_PAYMENT_REGISTRY.register(
-        core,
-        transaction,
-        account_type.as_standard_account_type(),
-        account_index,
-    );
-
-    // txid hex never contains a NUL, but handle the impossible case anyway.
+    // Do all fallible/pure marshalling BEFORE the registry insert — that insert
+    // mints a token and holds the funding reservation, so a later failure would
+    // orphan the reservation with no token to release it. txid hex never
+    // contains a NUL, but handle the impossible case anyway.
     let c_txid = match CString::new(txid.to_string()) {
         Ok(s) => s,
         Err(_) => {
@@ -127,9 +107,20 @@ pub unsafe extern "C" fn core_wallet_signed_payment_register(
         }
     };
 
+    let token = runtime().block_on(SIGNED_PAYMENT_REGISTRY.register(
+        core,
+        transaction,
+        account_type.as_standard_account_type(),
+        account_index,
+    ));
+
     *out_token = token;
     *out_fee = fee;
     *out_txid = c_txid.into_raw();
+    // Borrowed view into the still-live `tx` buffer; the caller copies it out
+    // before freeing `tx` (mirrors the retired `core_wallet_transaction_get_bytes`).
+    *out_bytes_ptr = bytes.as_ptr();
+    *out_bytes_len = bytes.len();
     PlatformWalletFFIResult::ok()
 }
 
@@ -156,7 +147,8 @@ pub unsafe extern "C" fn core_wallet_signed_payment_broadcast(
 
     let core = unwrap_option_or_return!(CORE_WALLET_STORAGE.with_item(core_handle, |w| w.clone()));
 
-    let result = runtime().block_on(SIGNED_PAYMENT_REGISTRY.broadcast(token as ReservationToken, &core));
+    let result =
+        runtime().block_on(SIGNED_PAYMENT_REGISTRY.broadcast(token as ReservationToken, &core));
 
     match result {
         Ok(txid) => {
@@ -172,12 +164,14 @@ pub unsafe extern "C" fn core_wallet_signed_payment_broadcast(
             *out_txid = c_txid.into_raw();
             PlatformWalletFFIResult::ok()
         }
-        Err(e @ (SignedPaymentError::StaleToken(_) | SignedPaymentError::WalletMismatch(_))) => {
-            PlatformWalletFFIResult::err(
-                PlatformWalletFFIResultCode::ErrorStaleReservationToken,
-                e.to_string(),
-            )
-        }
+        Err(
+            e @ (SignedPaymentError::StaleToken(_)
+            | SignedPaymentError::WalletMismatch(_)
+            | SignedPaymentError::StaleReservationToken(_)),
+        ) => PlatformWalletFFIResult::err(
+            PlatformWalletFFIResultCode::ErrorStaleReservationToken,
+            e.to_string(),
+        ),
         // Preserve the typed underlying wallet error (keeps the ambiguous
         // "may already be on the network" retry semantics intact).
         Err(SignedPaymentError::Broadcast(e)) => PlatformWalletFFIResult::from(e),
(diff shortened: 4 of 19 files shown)


thepastaclaw:
⛔ Blockers found — Sonnet deferred (commit 84ef2678)
_Canonical validated blockers: 3_

packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs
```diff
@@ -0,0 +1,688 @@
...
+            // Unknown / already consumed — idempotent no-op.
+            return;
+        };
+        if let Some(account_type) = entry.account_type {
+            entry
+                .core
+                .release_transaction_reservation(account_type, entry.account_index, &entry.tx)
+                .await;
+        }
+    }
```
thepastaclaw: **🔴 Blocking: Registry tokens never expire, but the UTXO reservation they depend on does — release-by-outpoint can corrupt a newer, unrelated reservation**

Verified against key-wallet's `ReservationSet` (`managed_account/reservation.rs`): `RESERVATION_TTL_BLOCKS = 24` (~1hr), and any later `reserve()`/`reserved()` call sweeps entries older than that, silently returning the outpoint to the selectable pool. `release()` then removes an outpoint unconditionally (`reserved.remove(outpoint)`) with no generation/ownership check.

`SignedPaymentRegistry` entries carry no age/height and are never swept — a token stays broadcastable/releasable indefinitely until an explicit `broadcast`/`release` call. This PR's entire purpose is to widen the build→submit gap for BIP70/BIP270 flows where the merchant ack is asynchronous and can legitimately take longer than an hour (slow, unresponsive, or simply a user who steps away).

Concrete failure sequence: (1) build+register payment A, reserving UTXO X at height H; merchant never acks. (2) At height ≥ H+24, X's reservation is swept on the next unrelated `reserve`/`reserved` call. (3) A completely different payment B is built, reselecting and re-reserving X at the new height. (4) The merchant for A finally acks; `broadcastSigned(tokenA)` runs. If the network rejects A's now-conflicting tx, `broadcast_releasing_on_rejection` releases X by outpoint — which removes B's *active* reservation, letting a third build reselect X while B's transaction may still be in flight, risking a real double-spend. If A's broadcast instead succeeds first, B's later build/broadcast fails unpredictably. Either way the outcome is driven entirely by third-party (merchant) timing outside the wallet's control.

The registry needs either a bounded token lifetime enforced client-side (reject broadcast/require re-registration once the reservation could plausibly have been swept) or a generation-aware release that only frees an outpoint if it still belongs to the same token's reservation.

packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs
```diff
@@ -0,0 +1,688 @@
...
+                .lock()
+                .expect("signed-payment registry mutex poisoned");
+            entries.remove(&token)
+        }
+        .ok_or(SignedPaymentError::StaleToken(token))?;
+
+        if !Arc::ptr_eq(&entry.core.wallet_manager, &current.wallet_manager) {
+            // The token belongs to another wallet instance; it has been removed,
+            // so it can never be replayed here.
+            return Err(SignedPaymentError::WalletMismatch(token));
```
thepastaclaw: **🟡 Suggestion: Wallet-instance guard only detects a different PlatformWalletManager, not a different wallet within the same manager**

Verified: `PlatformWalletManager` (`manager/mod.rs:39-47`) is documented as a "Multi-wallet coordinator" holding exactly one `wallet_manager: Arc<RwLock<WalletManager<PlatformWalletInfo>>>` shared by every registered wallet (`wallets: Arc<RwLock<BTreeMap<WalletId, Arc<PlatformWallet>>>>`), and `PlatformWallet::new`/`platform_wallet_get_core` construct every `CoreWallet` from that same cloned Arc. So `Arc::ptr_eq(&entry.core.wallet_manager, &current.wallet_manager)` is true for *any two wallets in the same manager* — it only catches recreation of the whole manager (e.g. the documented network-switch destroy+recreate), not per-wallet removal/recreation or cross-wallet token reuse within one live manager.

Impact is bounded, not a fund-safety bypass: the actual broadcast/release always operates through `entry.core` (the wallet captured at registration), which still resolves the correct `wallet_id`/account internally — a caller can't spend wallet B's funds with wallet A's token. But the doc comment's claim that a token is "bound to the exact wallet instance it was minted against" is only true across manager instances, not across wallets sharing one manager, and the existing test (`broadcast_rejects_a_different_wallet_instance`) only covers the cross-manager case with two entirely separate `WalletManager`s, missing the realistic same-manager, different-`wallet_id` scenario this multi-wallet coordinator is built for. Tighten the check to also compare `entry.core.wallet_id()` against `current.wallet_id()`.

```suggestion
if !Arc::ptr_eq(&entry.core.wallet_manager, &current.wallet_manager)
    || entry.core.wallet_id() != current.wallet_id()
{
    return Err(SignedPaymentError::WalletMismatch(token));
}
```

<sub>source: ['codex']</sub>

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedCoreWallet.kt
```diff
@@ -39,6 +39,45 @@ class ManagedCoreWallet internal constructor(handle: Long) : AutoCloseable {
...
+        val txidLen = buffer.int
+        val txidBytes = ByteArray(txidLen)
+        buffer.get(txidBytes)
+        return ManagedPlatformWallet.SignedCoreTransaction(
+            txidHex = String(txidBytes, Charsets.UTF_8),
+            rawTxBytes = rawTxBytes,
+            feeDuffs = feeDuffs,
+            reservationToken = token,
+        )
+    }
```
thepastaclaw, line 75: **🟡 Suggestion: registerSignedPayment stitches two native round-trips in Kotlin, contrary to this repo's kotlin-sdk rule**

`registerSignedPayment` calls `coreTransactionGetBytes(tx.handle)` and then `coreWalletRegisterSignedPayment(...)`, stitching the two results into one `SignedCoreTransaction` with hand-rolled `ByteBuffer` decoding. `packages/kotlin-sdk/CLAUDE.md` states explicitly: "No JNI functions that stitch together existing Rust calls — add the composite to Rust instead." Confirmed `core_wallet_transaction_get_bytes`/`coreTransactionGetBytes` has exactly one Kotlin call site — this one — so there's no other caller relying on it staying separate.

`core_wallet_signed_payment_register` (`rs-platform-wallet-ffi/src/core_wallet/signed_payment.rs:84`) already has `(*tx).bytes()` in scope (it uses it to deserialize the transaction at line 100); it could copy those bytes into an additional out-parameter and return everything in one native round trip, removing `coreTransactionGetBytes` from this path entirely.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet-ffi/src/core_wallet/signed_payment.rs
```diff
@@ -0,0 +1,199 @@
...
+                "txid string contained an interior NUL".to_string(),
+            );
+        }
+    };
+
+    *out_token = token;
+    *out_fee = fee;
+    *out_txid = c_txid.into_raw();
+    PlatformWalletFFIResult::ok()
+}
```
thepastaclaw, line 125: **🟡 Suggestion: Reservation-holding register() can orphan its UTXO reservation on failure paths after the side effect commits**

`core_wallet_signed_payment_register` calls `SIGNED_PAYMENT_REGISTRY.register(core, transaction, ...)` at line 112 — an irreversible side effect that mints a token and keeps `build_signed`'s UTXO reservation held — before the fallible `CString::new(txid.to_string())` at line 120. If that (admittedly near-impossible, since txid hex can't contain a NUL) conversion fails, the function returns an error without ever writing `out_token`, orphaning the reservation with no token to release it.

The same class of bug reaches further down the JNI boundary: in `rs-unified-sdk-jni/src/wallet_manager.rs` (`coreWalletRegisterSignedPayment`, lines ~1053-1090), the native call already committed the registration by the time `env.byte_array_from_slice(&blob)` runs at line 1087. That call's failure path is `.unwrap_or(ptr::null_mut())` — no exception thrown, no token returned to Kotlin, same orphaned-reservation outcome (this time reachable on JVM allocation pressure rather than an impossible NUL byte).

Both paths leave the reservation to be reclaimed only by the 24-block TTL backstop. Reorder the fallible/pure steps before the registry insert on the Rust side, and release the token on the JNI marshalling-failure path.

```suggestion
match env.byte_array_from_slice(&blob) {
    Ok(array) => array.into_raw(),
    Err(_) => {
        let _ = unsafe {
            platform_wallet_ffi::core_wallet_signed_payment_release(token)
        };
        ptr::null_mut()
    }
}
```

packages/rs-platform-wallet-ffi/src/core_wallet/signed_payment.rs
```diff
@@ -0,0 +1,199 @@
...
+    out_token: *mut u64,
+    out_fee: *mut u64,
+    out_txid: *mut *mut c_char,
+) -> PlatformWalletFFIResult {
+    check_ptr!(tx);
+    check_ptr!(out_token);
+    check_ptr!(out_fee);
+    check_ptr!(out_txid);
+
+    let core = unwrap_option_or_return!(CORE_WALLET_STORAGE.with_item(core_handle, |w| w.clone()));
```
thepastaclaw, line 81: **🟡 Suggestion: Registry captures its own CoreWallet Arc clone, bypassing core_wallet_destroy/platform_wallet_destroy teardown**

`register()` clones the `CoreWallet<B>` from `CORE_WALLET_STORAGE` (line 98) and stores it inside `RegisteredPayment` in the process-global `SIGNED_PAYMENT_REGISTRY` — independent of the handle-storage lifecycle. `core_wallet_destroy`/`platform_wallet_destroy` only remove the handle-storage entry, dropping *that* Arc reference; they have no visibility into the registry.

If a caller registers a token and never calls `broadcastSigned`/`releaseReservation` on it (app killed mid-flow, or — per `packages/kotlin-sdk/CLAUDE.md`'s own documented pattern, "Network switch = destroy + new instance" — a network switch destroys the `PlatformWalletManager` while a token is outstanding), the registry entry keeps the entire `WalletManager` (accounts, keys, sync state) alive for the rest of the process. Not memory-unsafe, but it silently defeats the destroy-wallet invariant the SDK's own docs rely on, with no error surfaced to the caller.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs
thepastaclaw: **💬 Nitpick: std::Mutex poisoning permanently disables the process-global deferred-payment surface**

`register`/`broadcast`/`release` all call `.expect("signed-payment registry mutex poisoned")` on the `entries` lock (lines 140, 178, 217). Since `SIGNED_PAYMENT_REGISTRY` is a single process-global, any panic anywhere while this lock is held poisons it for the entire process — every subsequent call then panics via `.expect()`, permanently disabling BIP70 deferred payments (for every wallet) until the app restarts. Availability hardening rather than an exploitable defect, but worth a `poisoned.into_inner()` recovery given the registry's process-wide blast radius — the sibling `ReservationSet::lock()` in key-wallet already does exactly this.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs
```diff
@@ -0,0 +1,688 @@
...
+use crate::broadcaster::TransactionBroadcaster;
+use crate::wallet::core::CoreWallet;
+use crate::PlatformWalletError;
+
+/// Opaque handle to a registered, signed-but-unsent payment. Minted by
+/// [`SignedPaymentRegistry::register`]; consumed by
+/// [`SignedPaymentRegistry::broadcast`] or
+/// [`SignedPaymentRegistry::release`]. Values are unique for the process
+/// lifetime and never reused, so a stale token can always be recognised.
+pub type ReservationToken = u64;
```
thepastaclaw, line 69: **💬 Nitpick: ReservationToken is documented as opaque but is only a bare u64 alias**

`pub type ReservationToken = u64` gives Rust callers no domain separation for a capability-like, process-scoped handle whose uniqueness/provenance matters. A `#[repr(transparent)]` newtype would preserve the C ABI while catching accidental misuse of unrelated numeric IDs at compile time. Low priority — style/hygiene only.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):
## Code Review

This PR adds a well-tested SignedPaymentRegistry enabling BIP70-style deferred build→broadcast/release for Core sends, with clean Rust idioms (no unwrap, correct pre-await mutex release, double-broadcast prevention). Verification against the actual code confirms one real blocking defect: registry tokens never expire even though the underlying key-wallet UTXO reservation they depend on auto-expires after 24 blocks (~1hr) and is released by raw outpoint with no ownership check, so a delayed merchant ack/nack on a long-outstanding token can silently corrupt or double-spend a completely unrelated, later-built transaction that reused the same now-expired UTXO. Several other issues (wallet-instance binding gaps, an architecture-rule violation on the Kotlin JNI boundary, and orphan-on-failure edge cases) are real but lower-impact.

Source (experiment sonnet-primary-opus-quarter-sample-20260710, cohort sonnet_opus_sample, bucket 0): reviewers codex/general=gpt-5.6-sol(completed); sonnet5/general=claude-sonnet-5(completed); opus/general=claude-opus-4-8(completed); codex/security-auditor=gpt-5.6-sol(completed); sonnet5/security-auditor=claude-sonnet-5(completed); opus/security-auditor=claude-opus-4-8(completed); codex/rust-quality=gpt-5.6-sol(completed); sonnet5/rust-quality=claude-sonnet-5(completed); opus/rust-quality=claude-opus-4-8(completed); codex/ffi-engineer=gpt-5.6-sol(completed); sonnet5/ffi-engineer=claude-sonnet-5(completed); opus/ffi-engineer=claude-opus-4-8(completed); verifier=verifier-sonnet5-4090-1783717967=claude-sonnet-5; orchestrator=openai/gpt-5.6-sol reasoning=high (orchestration-only, not a reviewer/verifier).

🔴 1 blocking | 🟡 4 suggestion(s) | 💬 2 nitpick(s)

<details>

```

In `packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs:130-230: Registry tokens never expire, but the UTXO reservation they depend on does — release-by-outpoint can corrupt a newer, unrelated reservation
  Verified against key-wallet's `ReservationSet` (`managed_account/reservation.rs`): `RESERVATION_TTL_BLOCKS = 24` (~1hr), and any later `reserve()`/`reserved()` call sweeps entries older than that, silently returning the outpoint to the selectable pool. `release()` then removes an outpoint unconditionally (`reserved.remove(outpoint)`) with no generation/ownership check.

  `SignedPaymentRegistry` entries carry no age/height and are never swept — a token stays broadcastable/releasable indefinitely until an explicit `broadcast`/`release` call. This PR's entire purpose is to widen the build→submit gap for BIP70/BIP270 flows where the merchant ack is asynchronous and can legitimately take longer than an hour (slow, unresponsive, or simply a user who steps away).

  Concrete failure sequence: (1) build+register payment A, reserving UTXO X at height H; merchant never acks. (2) At height ≥ H+24, X's reservation is swept on the next unrelated `reserve`/`reserved` call. (3) A completely different payment B is built, reselecting and re-reserving X at the new height. (4) The merchant for A finally acks; `broadcastSigned(tokenA)` runs. If the network rejects A's now-conflicting tx, `broadcast_releasing_on_rejection` releases X by outpoint — which removes B's *active* reservation, letting a third build reselect X while B's transaction may still be in flight, risking a real double-spend. If A's broadcast instead succeeds first, B's later build/broadcast fails unpredictably. Either way the outcome is driven entirely by third-party (merchant) timing outside the wallet's control.

  The registry needs either a bounded token lifetime enforced client-side (reject broadcast/require re-registration once the reservation could plausibly have been swept) or a generation-aware release that only frees an outpoint if it still belongs to the same token's reservation.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs:183-186: Wallet-instance guard only detects a different PlatformWalletManager, not a different wallet within the same manager
  Verified: `PlatformWalletManager` (`manager/mod.rs:39-47`) is documented as a "Multi-wallet coordinator" holding exactly one `wallet_manager: Arc<RwLock<WalletManager<PlatformWalletInfo>>>` shared by every registered wallet (`wallets: Arc<RwLock<BTreeMap<WalletId, Arc<PlatformWallet>>>>`), and `PlatformWallet::new`/`platform_wallet_get_core` construct every `CoreWallet` from that same cloned Arc. So `Arc::ptr_eq(&entry.core.wallet_manager, &current.wallet_manager)` is true for *any two wallets in the same manager* — it only catches recreation of the whole manager (e.g. the documented network-switch destroy+recreate), not per-wallet removal/recreation or cross-wallet token reuse within one live manager.

  Impact is bounded, not a fund-safety bypass: the actual broadcast/release always operates through `entry.core` (the wallet captured at registration), which still resolves the correct `wallet_id`/account internally — a caller can't spend wallet B's funds with wallet A's token. But the doc comment's claim that a token is "bound to the exact wallet instance it was minted against" is only true across manager instances, not across wallets sharing one manager, and the existing test (`broadcast_rejects_a_different_wallet_instance`) only covers the cross-manager case with two entirely separate `WalletManager`s, missing the realistic same-manager, different-`wallet_id` scenario this multi-wallet coordinator is built for. Tighten the check to also compare `entry.core.wallet_id()` against `current.wallet_id()`.

In `packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedCoreWallet.kt`:
- [SUGGESTION] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedCoreWallet.kt:49-71: registerSignedPayment stitches two native round-trips in Kotlin, contrary to this repo's kotlin-sdk rule
  `registerSignedPayment` calls `coreTransactionGetBytes(tx.handle)` and then `coreWalletRegisterSignedPayment(...)`, stitching the two results into one `SignedCoreTransaction` with hand-rolled `ByteBuffer` decoding. `packages/kotlin-sdk/CLAUDE.md` states explicitly: "No JNI functions that stitch together existing Rust calls — add the composite to Rust instead." Confirmed `core_wallet_transaction_get_bytes`/`coreTransactionGetBytes` has exactly one Kotlin call site — this one — so there's no other caller relying on it staying separate.

  `core_wallet_signed_payment_register` (`rs-platform-wallet-ffi/src/core_wallet/signed_payment.rs:84`) already has `(*tx).bytes()` in scope (it uses it to deserialize the transaction at line 100); it could copy those bytes into an additional out-parameter and return everything in one native round trip, removing `coreTransactionGetBytes` from this path entirely.

In `packages/rs-platform-wallet-ffi/src/core_wallet/signed_payment.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/core_wallet/signed_payment.rs:84-134: Reservation-holding register() can orphan its UTXO reservation on failure paths after the side effect commits
  `core_wallet_signed_payment_register` calls `SIGNED_PAYMENT_REGISTRY.register(core, transaction, ...)` at line 112 — an irreversible side effect that mints a token and keeps `build_signed`'s UTXO reservation held — before the fallible `CString::new(txid.to_string())` at line 120. If that (admittedly near-impossible, since txid hex can't contain a NUL) conversion fails, the function returns an error without ever writing `out_token`, orphaning the reservation with no token to release it.

  The same class of bug reaches further down the JNI boundary: in `rs-unified-sdk-jni/src/wallet_manager.rs` (`coreWalletRegisterSignedPayment`, lines ~1053-1090), the native call already committed the registration by the time `env.byte_array_from_slice(&blob)` runs at line 1087. That call's failure path is `.unwrap_or(ptr::null_mut())` — no exception thrown, no token returned to Kotlin, same orphaned-reservation outcome (this time reachable on JVM allocation pressure rather than an impossible NUL byte).

  Both paths leave the reservation to be reclaimed only by the 24-block TTL backstop. Reorder the fallible/pure steps before the registry insert on the Rust side, and release the token on the JNI marshalling-failure path.
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/core_wallet/signed_payment.rs:98: Registry captures its own CoreWallet Arc clone, bypassing core_wallet_destroy/platform_wallet_destroy teardown
  `register()` clones the `CoreWallet<B>` from `CORE_WALLET_STORAGE` (line 98) and stores it inside `RegisteredPayment` in the process-global `SIGNED_PAYMENT_REGISTRY` — independent of the handle-storage lifecycle. `core_wallet_destroy`/`platform_wallet_destroy` only remove the handle-storage entry, dropping *that* Arc reference; they have no visibility into the registry.

  If a caller registers a token and never calls `broadcastSigned`/`releaseReservation` on it (app killed mid-flow, or — per `packages/kotlin-sdk/CLAUDE.md`'s own documented pattern, "Network switch = destroy + new instance" — a network switch destroys the `PlatformWalletManager` while a token is outstanding), the registry entry keeps the entire `WalletManager` (accounts, keys, sync state) alive for the rest of the process. Not memory-unsafe, but it silently defeats the destroy-wallet invariant the SDK's own docs rely on, with no error surfaced to the caller.
```
</details>

packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs
```diff
@@ -0,0 +1,1089 @@
...
+            RegisteredPayment {
+                core,
+                tx,
+                account_type,
+                account_index,
+                registered_height,
+            },
+        );
+        token
+    }
```
thepastaclaw, line 227: **🔴 Blocking: Token expiry guard measures a different height clock than the reservation it bounds (prior-1, still valid)**

`register()` records `core.synced_height()` as `registered_height`, and `broadcast`/`release` compare it against a later `synced_height()` read using `RESERVATION_MAX_AGE_BLOCKS = 20`. But the underlying `ReservationSet` TTL that this guard is meant to stay under is stamped by `core_wallet_tx_builder_build_signed` (packages/rs-platform-wallet-ffi/src/core_wallet/transaction_builder.rs:550-557) using `last_processed_height()`, not `synced_height()`. `payments.rs:96-99` documents these as intentionally decoupled clocks — `synced_height` is the filter-scan checkpoint and can regress during a DashPay rescan while `last_processed_height` remains monotonic. Whenever `synced_height` lags `last_processed_height` at registration time (or is later rolled back by a rescan), the 20-block guard can fail to trip before key-wallet's real 24-block `RESERVATION_TTL_BLOCKS` sweep (packages/rs-platform-wallet's pinned key-wallet dependency, `managed_account/reservation.rs:32,59-61`) has already reclaimed and possibly re-reserved the outpoint for an unrelated build. A broadcast or release on the stale token then acts on state it no longer owns — `ReservationSet::release`/`reserve` have no per-outpoint ownership check, so this can silently release or corrupt a newer build's reservation, the exact corruption class the guard was added to close. The new tests do not catch this because their `build_signed_tx` helper stamps and advances only `synced_height`, unlike the production FFI path. Track and pass through the exact height `build_signed` used (e.g. via `FFICoreTransaction`) instead of independently re-reading `synced_height` at registration.

<sub>source: ['codex-general', 'codex-security-auditor', 'codex-rust-quality', 'sonnet5-security-auditor']</sub>

packages/rs-platform-wallet-ffi/src/wallet.rs
```diff
@@ -198,6 +198,17 @@ pub unsafe extern "C" fn platform_wallet_manager_free_account_balances(
...
+    // deferred flow builds/registers on one short-lived core handle and
+    // broadcasts on another, so sweeping on core-handle destroy would drop
+    // tokens between register and broadcast.
+    PLATFORM_WALLET_STORAGE.with_item(handle, |wallet| {
+        crate::core_wallet::signed_payment::SIGNED_PAYMENT_REGISTRY
+            .remove_entries_for_wallet(wallet.core());
+    });
     PLATFORM_WALLET_STORAGE.remove(handle);
     PlatformWalletFFIResult::ok()
 }
```
thepastaclaw, line 425: **🔴 Blocking: platform_wallet_destroy's registry sweep is keyed on wallet identity, not handle identity — destroying one alias invalidates a sibling wrapper's deferred payment**

`platform_wallet_destroy` now sweeps every registry entry whose `Arc::ptr_eq(wallet_manager)` and `wallet_id` match the destroyed handle's wallet (via `remove_entries_for_wallet`, packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs:329-337). But a `PlatformWallet` handle is not a 1:1 proxy for the logical wallet's lifetime: `platform_wallet_manager_get_wallet` (packages/rs-platform-wallet-ffi/src/manager.rs:322-349) mints a fresh, independent handle on every call, cloning the same underlying `wallet_manager` Arc and `wallet_id`. The Kotlin SDK explicitly documents and relies on this — `ManagedPlatformWallet.kt:38-40` notes `loadPersistedWallets` can hand out a new wrapper for a wallet id an earlier live wrapper still holds. If wrapper B registers a deferred payment and wrapper A (a sibling alias of the same wallet) is then closed via `close()`/the `Cleaner` backstop, A's `platform_wallet_destroy` call sweeps B's token too, since the match is on wallet identity, not on which handle actually produced the token. B's later `broadcastSigned`/`releaseReservation` then wrongly fails with `StaleToken`, and because this sweep deliberately does not release the reservation, the funding UTXO stays locked until the 20-block TTL backstop. A narrower variant of the same root cause is a race between an in-flight `core_wallet_signed_payment_register` on a short-lived core handle and a concurrent `platform_wallet_destroy` on another alias — if the register lands after the sweep, that token is never covered by any future sweep, pinning the `WalletManager` alive indefinitely instead of just until TTL. Scope cleanup to the wallet actually being removed from the manager (or track live handle count and sweep only when the last alias is destroyed) rather than any single handle's teardown.

<sub>source: ['codex-general', 'codex-rust-quality', 'codex-ffi-engineer', 'sonnet5-ffi-engineer', 'sonnet5-security-auditor']</sub>

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
```diff
@@ -191,6 +191,143 @@ class ManagedPlatformWallet internal constructor(
...
+                            coreSignerHandle,
+                        )
+                    }
+                    // Register the signed tx (holding its reservation) before the
+                    // native transaction is freed; `use` frees it afterward.
+                    signedTx.use { tx -> core.registerSignedPayment(tx) }
+                }
+            }
+        }
+    }
```
thepastaclaw: **🟡 Suggestion: Coroutine cancellation can discard a minted reservation token after buildSignedPayment's native call has already committed it**

`core.registerSignedPayment(tx)` (line 295) mints a process-global registry token and starts holding the UTXO reservation, but this happens inside the dispatcher-changing `withContext(Dispatchers.IO)` wrapping the whole function (line 265). Kotlin's prompt-cancellation semantics mean a `Job` cancellation racing the dispatcher hop back to the caller can cause `withContext` to throw `CancellationException` even though the enclosed block already completed successfully and committed the native registration. In that case `buildSignedPayment` never delivers the `SignedCoreTransaction` (and its `reservationToken`) to the caller, so nothing can `broadcastSigned` or `releaseReservation` it — the registry entry sits until the `RESERVATION_MAX_AGE_BLOCKS` TTL backstop reclaims it. This is a narrow, timing-dependent window (must race cancellation against the specific dispatcher-resume point after a successful native call) and is bounded by the existing TTL guard, so it's an availability/leak concern rather than a correctness or fund-loss issue — but it is new exposure from this PR's side-effecting native call inside a cancellable coroutine scope.

thepastaclaw (changes requested):
## Code Review

Follow-up review of PR #4090 at head a428f54021d6cbd6fae13dd1cb681f1bbcfce2f9 (prior head ee2135bf4665d0a58c6c57833122a8868721eda5). All seven prior findings were re-verified against current source, including the exact pinned key-wallet dependency (rev 1860089eddea27f66e5b3e637d18a73ae138478c) fetched from the cargo git checkout cache to resolve a factual disagreement among reviewer lanes. Carried-forward prior findings: prior-1 (blocking) is STILL VALID — first-principles inspection confirms `core_wallet_tx_builder_build_signed` (packages/rs-platform-wallet-ffi/src/core_wallet/transaction_builder.rs:550) stamps the underlying `ReservationSet` TTL using `last_processed_height()`, while `SignedPaymentRegistry::register` (packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs:210) records the wallet's `synced_height()` — a field independently documented (packages/rs-platform-wallet/src/wallet/identity/network/payments.rs:96-99) as decoupled from and capable of regressing relative to the monotonic `last_processed_height`. The new registry test helper (`build_signed_tx` in the registry's own test module) happens to use `synced_height()` for both stamping and comparison, masking this divergence in tests while the production FFI path does not share that property. Three lanes (sonnet5-general, sonnet5-rust-quality, sonnet5-ffi-engineer) incorrectly reconciled this as FIXED by conflating the test helper's clock with the production FFI clock; that reconciliation does not hold under direct source inspection and is overridden here. prior-7 (nitpick) is STILL VALID — `ReservationToken` remains a bare `pub type ReservationToken = u64` alias. Prior-2 (wallet-instance guard now also compares `wallet_id()`), prior-3 (Kotlin double round-trip eliminated — `core_wallet_transaction_get_bytes`/`coreTransactionGetBytes` no longer exist, tx bytes ride back in the single register BLOB), prior-4 (fallible `CString::new` now runs before the registry insert; the JNI `byte_array_from_slice` failure path now releases the token), prior-5 (registry-held `CoreWallet` Arc now swept via `remove_entries_for_wallet` on `platform_wallet_destroy`), and prior-6 (mutex poisoning now recovered via a `lock()` helper) are all FIXED and verified directly against current code. New findings in the latest delta: the `remove_entries_for_wallet` sweep added to fix prior-5 matches on `Arc::ptr_eq(wallet_manager) && wallet_id` rather than on the specific destroyed handle, so it does not distinguish between multiple live `PlatformWallet` handles that alias the same logical wallet — a pattern the Kotlin SDK's own doc comments (ManagedPlatformWallet.kt:38-40) explicitly anticipate (`loadPersistedWallets` can hand out a fresh wrapper for a wallet id an earlier wrapper still holds). Destroying/GC'ing one such alias silently invalidates a deferred-payment token registered through a sibling, wrongly rejecting a legitimate broadcast/release with StaleToken and pinning the UTXO reservation until the 20-block TTL backstop. A second, narrower manifestation of the same root cause is a race between an in-flight `core_wallet_signed_payment_register` on one short-lived core handle and a concurrent `platform_wallet_destroy` on another alias of the same wallet. Separately, `buildSignedPayment` in Kotlin commits the native registry token inside `withContext(Dispatchers.IO)`; coroutine cancellation racing the dispatcher hop back to the caller can discard the returned `SignedCoreTransaction` after the token was already minted, orphaning it until the TTL backstop reclaims it. Review action is REQUEST_CHANGES due to the carried-forward blocking clock-mismatch and the new alias-teardown regression, both introduced/exposed by this PR's own reservation-hardening code.

Source (experiment sonnet-primary-opus-quarter-sample-20260710, cohort sonnet_primary, bucket 1): reviewers codex/general=gpt-5.6-sol(completed); sonnet5/general=claude-sonnet-5(completed); codex/security-auditor=gpt-5.6-sol(completed); sonnet5/security-auditor=claude-sonnet-5(completed); codex/rust-quality=gpt-5.6-sol(completed); sonnet5/rust-quality=claude-sonnet-5(completed); codex/ffi-engineer=gpt-5.6-sol(completed); sonnet5/ffi-engineer=claude-sonnet-5(completed); verifier=verifier-sonnet5-4090-1783735156=claude-sonnet-5; orchestrator=openai/gpt-5.6-sol reasoning=high (orchestration-only, not a reviewer/verifier).

<details>

```

In `packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs:203-223: Token expiry guard measures a different height clock than the reservation it bounds (prior-1, still valid)
  `register()` records `core.synced_height()` as `registered_height`, and `broadcast`/`release` compare it against a later `synced_height()` read using `RESERVATION_MAX_AGE_BLOCKS = 20`. But the underlying `ReservationSet` TTL that this guard is meant to stay under is stamped by `core_wallet_tx_builder_build_signed` (packages/rs-platform-wallet-ffi/src/core_wallet/transaction_builder.rs:550-557) using `last_processed_height()`, not `synced_height()`. `payments.rs:96-99` documents these as intentionally decoupled clocks — `synced_height` is the filter-scan checkpoint and can regress during a DashPay rescan while `last_processed_height` remains monotonic. Whenever `synced_height` lags `last_processed_height` at registration time (or is later rolled back by a rescan), the 20-block guard can fail to trip before key-wallet's real 24-block `RESERVATION_TTL_BLOCKS` sweep (packages/rs-platform-wallet's pinned key-wallet dependency, `managed_account/reservation.rs:32,59-61`) has already reclaimed and possibly re-reserved the outpoint for an unrelated build. A broadcast or release on the stale token then acts on state it no longer owns — `ReservationSet::release`/`reserve` have no per-outpoint ownership check, so this can silently release or corrupt a newer build's reservation, the exact corruption class the guard was added to close. The new tests do not catch this because their `build_signed_tx` helper stamps and advances only `synced_height`, unlike the production FFI path. Track and pass through the exact height `build_signed` used (e.g. via `FFICoreTransaction`) instead of independently re-reading `synced_height` at registration.

In `packages/rs-platform-wallet-ffi/src/wallet.rs`:
- [BLOCKING] packages/rs-platform-wallet-ffi/src/wallet.rs:200-214: platform_wallet_destroy's registry sweep is keyed on wallet identity, not handle identity — destroying one alias invalidates a sibling wrapper's deferred payment
  `platform_wallet_destroy` now sweeps every registry entry whose `Arc::ptr_eq(wallet_manager)` and `wallet_id` match the destroyed handle's wallet (via `remove_entries_for_wallet`, packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs:329-337). But a `PlatformWallet` handle is not a 1:1 proxy for the logical wallet's lifetime: `platform_wallet_manager_get_wallet` (packages/rs-platform-wallet-ffi/src/manager.rs:322-349) mints a fresh, independent handle on every call, cloning the same underlying `wallet_manager` Arc and `wallet_id`. The Kotlin SDK explicitly documents and relies on this — `ManagedPlatformWallet.kt:38-40` notes `loadPersistedWallets` can hand out a new wrapper for a wallet id an earlier live wrapper still holds. If wrapper B registers a deferred payment and wrapper A (a sibling alias of the same wallet) is then closed via `close()`/the `Cleaner` backstop, A's `platform_wallet_destroy` call sweeps B's token too, since the match is on wallet identity, not on which handle actually produced the token. B's later `broadcastSigned`/`releaseReservation` then wrongly fails with `StaleToken`, and because this sweep deliberately does not release the reservation, the funding UTXO stays locked until the 20-block TTL backstop. A narrower variant of the same root cause is a race between an in-flight `core_wallet_signed_payment_register` on a short-lived core handle and a concurrent `platform_wallet_destroy` on another alias — if the register lands after the sweep, that token is never covered by any future sweep, pinning the `WalletManager` alive indefinitely instead of just until TTL. Scope cleanup to the wallet actually being removed from the manager (or track live handle count and sweep only when the last alias is destroyed) rather than any single handle's teardown.

- [SUGGESTION] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt:265-299: Coroutine cancellation can discard a minted reservation token after buildSignedPayment's native call has already committed it
  `core.registerSignedPayment(tx)` (line 295) mints a process-global registry token and starts holding the UTXO reservation, but this happens inside the dispatcher-changing `withContext(Dispatchers.IO)` wrapping the whole function (line 265). Kotlin's prompt-cancellation semantics mean a `Job` cancellation racing the dispatcher hop back to the caller can cause `withContext` to throw `CancellationException` even though the enclosed block already completed successfully and committed the native registration. In that case `buildSignedPayment` never delivers the `SignedCoreTransaction` (and its `reservationToken`) to the caller, so nothing can `broadcastSigned` or `releaseReservation` it — the registry entry sits until the `RESERVATION_MAX_AGE_BLOCKS` TTL backstop reclaims it. This is a narrow, timing-dependent window (must race cancellation against the specific dispatcher-resume point after a successful native call) and is bounded by the existing TTL guard, so it's an availability/leak concern rather than a correctness or fund-loss issue — but it is new exposure from this PR's side-effecting native call inside a cancellable coroutine scope.
```
</details>

packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs
thepastaclaw: Resolved in `a428f54` — _Wallet-instance guard only detects a different PlatformWalletManager, not a different wallet within the same manager_ no longer present.

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedCoreWallet.kt
thepastaclaw: Resolved in `a428f54` — _registerSignedPayment stitches two native round-trips in Kotlin, contrary to this repo's kotlin-sdk rule_ no longer present.

packages/rs-platform-wallet-ffi/src/core_wallet/signed_payment.rs
thepastaclaw: Resolved in `a428f54` — _Registry captures its own CoreWallet Arc clone, bypassing core_wallet_destroy/platform_wallet_destroy teardown_ no longer present.

packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs
thepastaclaw: Resolved in `a428f54` — _std::Mutex poisoning permanently disables the process-global deferred-payment surface_ no longer present.

packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs
thepastaclaw: Still valid at `a428f54` — `ReservationToken` remains a bare `pub type ReservationToken = u64` alias. This nitpick was carried forward in the current review and intentionally not duplicated as a new inline comment.

<sub>The earlier automated resolution reply was corrected because title-based reconciliation failed to match the reworded carried-forward finding.</sub>

packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs
thepastaclaw: Still valid at `a428f54` — the new 20-block registry guard reads `synced_height()`, while the production reservation is stamped with `last_processed_height()`, so the clocks can diverge and the original reservation-ownership risk remains. The current-head blocking thread is https://github.com/dashpay/platform/pull/4090#discussion_r3562924527.

<sub>The earlier automated resolution reply was corrected because title-based reconciliation failed to match the reworded carried-forward finding.</sub>

packages/rs-platform-wallet-ffi/src/core_wallet/signed_payment.rs
thepastaclaw: Resolved in `a428f54` — _Reservation-holding register() can orphan its UTXO reservation on failure paths after the side effect commits_ no longer present.

bfoss765:
Thanks — all findings were real. Bounded token lifetime added: `register` captures the wallet's synced height and `broadcast`/`release` now refuse a token older than `RESERVATION_MAX_AGE_BLOCKS` (20, below key-wallet's 24-block TTL) with a typed `StaleReservationToken`, deliberately *without* releasing (a post-sweep release could free a newer build's reservation). The pinned key-wallet exposes no per-outpoint generation check, so generation-aware release isn't possible without touching that crate — the bounded lifetime is the guard, documented in-code. Also: `WalletMismatch` now compares `wallet_id`; `register` returns the tx bytes in one native round trip (dead `coreTransactionGetBytes` removed); fallible marshalling reordered before the registry insert with token-release on JNI marshal failure; and a registry sweep on `platform_wallet_destroy` (not `core_wallet_destroy` — the deferred flow registers and broadcasts on different transient core handles, so sweeping on core-handle destroy would drop live tokens). Mutex now recovers from poisoning. Added expiry, same-manager `wallet_id`, and sweep tests. Left the `ReservationToken` newtype nitpick as-is (C ABI).

thepastaclaw:
Thanks — this summary matches the changes already present at `a428f540` and the five findings marked fixed in the current review. The current review still has two blocking findings on that same head: the guard records `synced_height()` while production stamps the underlying reservation with `last_processed_height()`, and `platform_wallet_destroy` sweeps tokens for sibling handles that alias the same logical wallet. The coroutine-cancellation window is also still open as a non-blocking suggestion. Please see the current-head threads linked from the `a428f540` review; I’ll re-review after the next push.

packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs
```diff
@@ -0,0 +1,1089 @@
...
+    /// lifetime against key-wallet's reservation TTL (see
+    /// [`RESERVATION_MAX_AGE_BLOCKS`]).
+    pub async fn register(
+        &self,
+        core: CoreWallet<B>,
+        tx: Transaction,
+        account_type: Option<StandardAccountType>,
+        account_index: u32,
+    ) -> ReservationToken {
+        let registered_height = core.synced_height().await;
```
thepastaclaw: **🔴 Blocking: Token age uses a different height clock than the reserved inputs**

`register` records `core.synced_height()`, while the production `core_wallet_tx_builder_build_signed` path stamps the underlying key-wallet reservation using `last_processed_height`. These clocks are intentionally independent: the DashPay rescan code documents that `synced_height` can regress while `last_processed_height` remains monotonic. The registry's 20-block guard can therefore continue accepting an old token after key-wallet's 24-block reservation TTL has swept and allowed the outpoint to be reserved by a newer transaction. A later release or definitively rejected broadcast for the old token removes reservations by outpoint without an ownership generation and can free the newer transaction's reservation. Pass the exact height used during transaction finalization into the registry instead of independently reading `synced_height`.

<sub>source: ['codex-general', 'codex-security-auditor', 'codex-rust-quality', 'codex-ffi-engineer']</sub>

packages/rs-platform-wallet-ffi/src/wallet.rs
```diff
@@ -390,6 +390,17 @@ pub unsafe extern "C" fn platform_wallet_manager_masternode_withdraw(
...
+    // state) alive for the rest of the process via the `CoreWallet` clone each
+    // token captured. Hooked here rather than into `core_wallet_destroy`: the
+    // deferred flow builds/registers on one short-lived core handle and
+    // broadcasts on another, so sweeping on core-handle destroy would drop
+    // tokens between register and broadcast.
+    PLATFORM_WALLET_STORAGE.with_item(handle, |wallet| {
+        crate::core_wallet::signed_payment::SIGNED_PAYMENT_REGISTRY
+            .remove_entries_for_wallet(wallet.core());
+    });
     PLATFORM_WALLET_STORAGE.remove(handle);
```
thepastaclaw: **🔴 Blocking: Destroying one alias handle invalidates another wrapper's payment token**

`platform_wallet_destroy` removes every registry entry matching the underlying wallet-manager pointer and wallet ID, although each `platform_wallet_manager_get_wallet` call creates an independent handle for the same logical wallet. The Kotlin `loadPersistedWallets` path can consequently publish a new wrapper while callers still retain an older wrapper. If wrapper B registers a deferred payment and wrapper A is then closed or cleaned, destroying A consumes B's live token. B's later broadcast fails as stale, and the sweep intentionally leaves the UTXO reserved until its TTL. A handle destructor should only drop that handle; wallet-wide registry cleanup must occur when the logical wallet is removed or when its final alias is destroyed.

<sub>source: ['codex-general', 'codex-security-auditor', 'codex-rust-quality', 'codex-ffi-engineer']</sub>

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
```diff
@@ -172,6 +172,142 @@ class ManagedPlatformWallet internal constructor(
...
+                    it.buildSigned(
+                        this@ManagedPlatformWallet,
+                        builderAccountType,
+                        accountIndex,
+                        coreSignerHandle,
+                    )
+                }
+                // Register the signed tx (holding its reservation) before the
+                // native transaction is freed; `use` frees it afterward.
+                signedTx.use { tx -> core.registerSignedPayment(tx) }
```
thepastaclaw: **🔴 Blocking: Deferred builds lost the atomic selection-and-reservation boundary**

`buildSignedPayment` still calls the deprecated `setFunding` and `buildSigned` methods as two native operations, but the latest delta removed the shared per-wallet `coreSendMutex` and replaced it with `gate.op`. `TeardownGate` only counts active operations for safe teardown; it releases its mutex before running the block and does not serialize sends. Key-wallet explicitly requires `set_funding` through `assemble_unsigned` to run under one uninterrupted wallet lock because selection observes reservations before the chosen inputs are reserved. Concurrent deferred builds can therefore select the same UTXO, and an atomic immediate send can also reserve an input after the deferred path selected it but before `buildSigned` reserves it. Both paths can return signed transactions spending the same input. Use an atomic finalize-and-register native operation for deferred payments, or restore shared per-wallet serialization around the entire funding, signing, and registration sequence.

<sub>source: ['codex-general', 'codex-security-auditor', 'codex-rust-quality', 'codex-ffi-engineer']</sub>

packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/errors/DashSdkErrorTest.kt
```diff
@@ -103,6 +103,16 @@ class DashSdkErrorTest {
         )
         // The message must warn against retrying (distinct from the anchor case).
         assertTrue(broadcastUnconfirmed.message!!.contains("do NOT retry"))
+
+        // Deferred build/broadcast: a stale/consumed/wrong-wallet reservation
+        // token → typed StaleReservationToken, not retryable.
+        val staleToken = DashSdkError.fromNative(DashSDKException(offset + 22, "stale token 7"))
```
thepastaclaw: **🔴 Blocking: The stale-token mapping test still uses native code 22**

The latest delta assigns code 22 to `ErrorCoreInsufficientFunds` and moves `ErrorStaleReservationToken` to code 26. The production Kotlin mapping correctly reflects those assignments, but this test constructs code 22 and asserts that it produces `StaleReservationToken`. It deterministically receives `CoreInsufficientFunds`, so the test fails and no longer verifies the code-26 mapping.

```suggestion
val staleToken = DashSdkError.fromNative(DashSDKException(offset + 26, "stale token 7"))
```

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
```diff
@@ -172,6 +172,142 @@ class ManagedPlatformWallet internal constructor(
...
+            }
+        }
```
thepastaclaw: **🟡 Suggestion: Coroutine cancellation can orphan a committed reservation token**

`core.registerSignedPayment(tx)` commits the process-global registry entry inside `gate.op`, whose implementation executes the block through `withContext(Dispatchers.IO)`. Cancellation can win during the dispatcher handoff after the native call has returned but before the completed `SignedCoreTransaction` is delivered to the caller. The caller then never receives the token required to broadcast or release the reservation, while `TeardownGate`'s `NonCancellable` cleanup only decrements its active-operation counter. Track whether token ownership was delivered and release a minted token from `NonCancellable` cleanup when delivery is cancelled.

<sub>source: ['codex-general', 'codex-security-auditor', 'codex-rust-quality', 'codex-ffi-engineer']</sub>

packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs
```diff
@@ -0,0 +1,1089 @@
...
+use crate::broadcaster::TransactionBroadcaster;
+use crate::wallet::core::CoreWallet;
+use crate::PlatformWalletError;
+
+/// Opaque handle to a registered, signed-but-unsent payment. Minted by
+/// [`SignedPaymentRegistry::register`]; consumed by
+/// [`SignedPaymentRegistry::broadcast`] or
+/// [`SignedPaymentRegistry::release`]. Values are unique for the process
+/// lifetime and never reused, so a stale token can always be recognised.
+pub type ReservationToken = u64;
```
thepastaclaw, line 69: **💬 Nitpick: The opaque reservation capability remains an untyped u64 alias**

`ReservationToken` has lifecycle and provenance semantics distinct from other numeric identifiers, but a type alias provides no Rust-side domain separation. Rust callers can accidentally interchange unrelated `u64` values with reservation tokens without a type error. A `#[repr(transparent)]` newtype would preserve the C-compatible representation while enforcing the distinction within Rust.

<sub>source: ['codex-general', 'codex-security-auditor', 'codex-rust-quality']</sub>

thepastaclaw (changes requested):

Carried-forward prior findings: all four priors remain valid at bbcd88e51f4b8a2f9ad00eb127ccfa96e017f8cd—two blocking reservation-lifecycle defects, one coroutine-cancellation cleanup gap, and one token type-safety nitpick. New findings in the latest delta: removing the shared send mutex leaves deferred builds on the explicitly non-atomic funding/signing sequence, and the stale-token mapping test still uses code 22 after the mapping moved to code 26. The PR has four confirmed blockers and is not ready to merge.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — security-auditor (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — ffi-engineer (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — security-auditor (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — ffi-engineer (failed), `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — security-auditor (completed), `gpt-5.6-sol` — rust-quality (completed), `gpt-5.6-sol` — ffi-engineer (completed)

🔴 4 blocking | 🟡 1 suggestion(s) | 💬 1 nitpick(s)

<details>

```

In `packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs:203-210: Token age uses a different height clock than the reserved inputs
  `register` records `core.synced_height()`, while the production `core_wallet_tx_builder_build_signed` path stamps the underlying key-wallet reservation using `last_processed_height`. These clocks are intentionally independent: the DashPay rescan code documents that `synced_height` can regress while `last_processed_height` remains monotonic. The registry's 20-block guard can therefore continue accepting an old token after key-wallet's 24-block reservation TTL has swept and allowed the outpoint to be reserved by a newer transaction. A later release or definitively rejected broadcast for the old token removes reservations by outpoint without an ownership generation and can free the newer transaction's reservation. Pass the exact height used during transaction finalization into the registry instead of independently reading `synced_height`.

In `packages/rs-platform-wallet-ffi/src/wallet.rs`:
- [BLOCKING] packages/rs-platform-wallet-ffi/src/wallet.rs:400-404: Destroying one alias handle invalidates another wrapper's payment token
  `platform_wallet_destroy` removes every registry entry matching the underlying wallet-manager pointer and wallet ID, although each `platform_wallet_manager_get_wallet` call creates an independent handle for the same logical wallet. The Kotlin `loadPersistedWallets` path can consequently publish a new wrapper while callers still retain an older wrapper. If wrapper B registers a deferred payment and wrapper A is then closed or cleaned, destroying A consumes B's live token. B's later broadcast fails as stale, and the sweep intentionally leaves the UTXO reserved until its TTL. A handle destructor should only drop that handle; wallet-wide registry cleanup must occur when the logical wallet is removed or when its final alias is destroyed.

- [BLOCKING] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt:262-276: Deferred builds lost the atomic selection-and-reservation boundary
  `buildSignedPayment` still calls the deprecated `setFunding` and `buildSigned` methods as two native operations, but the latest delta removed the shared per-wallet `coreSendMutex` and replaced it with `gate.op`. `TeardownGate` only counts active operations for safe teardown; it releases its mutex before running the block and does not serialize sends. Key-wallet explicitly requires `set_funding` through `assemble_unsigned` to run under one uninterrupted wallet lock because selection observes reservations before the chosen inputs are reserved. Concurrent deferred builds can therefore select the same UTXO, and an atomic immediate send can also reserve an input after the deferred path selected it but before `buildSigned` reserves it. Both paths can return signed transactions spending the same input. Use an atomic finalize-and-register native operation for deferred payments, or restore shared per-wallet serialization around the entire funding, signing, and registration sequence.
- [SUGGESTION] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt:274-278: Coroutine cancellation can orphan a committed reservation token
  `core.registerSignedPayment(tx)` commits the process-global registry entry inside `gate.op`, whose implementation executes the block through `withContext(Dispatchers.IO)`. Cancellation can win during the dispatcher handoff after the native call has returned but before the completed `SignedCoreTransaction` is delivered to the caller. The caller then never receives the token required to broadcast or release the reservation, while `TeardownGate`'s `NonCancellable` cleanup only decrements its active-operation counter. Track whether token ownership was delivered and release a minted token from `NonCancellable` cleanup when delivery is cancelled.

In `packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/errors/DashSdkErrorTest.kt`:
- [BLOCKING] packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/errors/DashSdkErrorTest.kt:109: The stale-token mapping test still uses native code 22
  The latest delta assigns code 22 to `ErrorCoreInsufficientFunds` and moves `ErrorStaleReservationToken` to code 26. The production Kotlin mapping correctly reflects those assignments, but this test constructs code 22 and asserts that it produces `StaleReservationToken`. It deterministically receives `CoreInsufficientFunds`, so the test fails and no longer verifies the code-26 mapping.
```
</details>

packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs
```diff
@@ -0,0 +1,1100 @@
...
+        self.lock().insert(
+            token,
+            RegisteredPayment {
+                core,
+                tx,
+                account_type,
+                account_index,
+                registered_height,
+            },
+        );
```
thepastaclaw, line 225: **🔴 Blocking: Registration loses the reservation's exact height**

`finalize_transaction` stamps the selected inputs using `last_processed_height` before releasing the wallet lock and awaiting the external signer, but `register` independently reads the height after signing finishes. If signing spans at least five blocks, the registry's 20-block guard can still accept the token after key-wallet's 24-block TTL has swept and re-reserved its outpoint. A later release or definitive broadcast rejection can then remove the newer reservation because release operates unconditionally by outpoint. Store the height used by `set_current_height` in `SignedCoreTransaction` and transfer that exact value into the registry.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/core_wallet/signed_payment.rs
```diff
@@ -0,0 +1,193 @@
...
+            );
+        }
+    };
+
+    let token = runtime().block_on(SIGNED_PAYMENT_REGISTRY.register(
+        core,
+        transaction,
+        account_type.as_standard_account_type(),
+        account_index,
+    ));
```
thepastaclaw, line 115: **🔴 Blocking: Registration does not preserve unique reservation ownership**

`FFICoreTransaction` contains only serialized bytes and a fee, while this function independently accepts the wallet and funding account and does not consume the transaction. A valid transaction can therefore be registered repeatedly or registered against an unrelated wallet/account. Duplicate tokens defeat reservation-level single-use semantics: one token can release the outpoint, a new payment can reserve it, and another old token can later release that newer reservation. Registering wallet A's transaction against wallet B instead reconciles B and leaves A's actual reservation held until TTL. Consume a single opaque finalized-payment handle containing its originating wallet, account, reservation height, and transaction, or remove the ownership-less legacy registration path.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/wallet.rs
```diff
@@ -390,6 +390,107 @@ pub unsafe extern "C" fn platform_wallet_manager_masternode_withdraw(
...
+    let wallet_id = core.wallet_id();
+    let manager = wallet.wallet_manager();
+    let sibling_alias_alive = PLATFORM_WALLET_STORAGE.any(|other| {
+        other.wallet_id() == wallet_id
+            && std::sync::Arc::ptr_eq(other.wallet_manager(), manager)
+    });
+    if !sibling_alias_alive {
+        crate::core_wallet::signed_payment::SIGNED_PAYMENT_REGISTRY
+            .remove_entries_for_wallet(core);
+    }
```
thepastaclaw, line 423: **🔴 Blocking: Token lifecycle is tied to transient wallet aliases**

`platform_wallet_destroy` deletes every token when no sibling wrapper is currently visible, even though the manager still owns the logical wallet and can immediately return another handle for it. Closing the last current wrapper therefore makes a retained payment token stale while intentionally leaving its UTXO reserved until TTL. The `remove`, sibling scan, alias insertion, and registry sweep also use separate locks, so a concurrent lookup or registration can race the sweep. Conversely, an alias retained across logical wallet removal and recreation still matches `(WalletManager Arc, wallet_id)`, allowing an old token to operate on the replacement generation. Perform cleanup on logical wallet removal or manager shutdown and coordinate it with a stable per-wallet generation or atomic lifecycle reference count.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

Carried-forward prior findings: prior-5 remains a cancellation-safety suggestion and prior-6 remains a token-type nitpick, while prior-1 through prior-4 are fixed as originally reported. New findings in the latest delta: three blocking root causes remain—the token lifetime starts after signing instead of at reservation, registration can mint duplicate or misattributed capabilities, and wallet cleanup and identity are tied to transient wrapper aliases instead of the logical wallet generation—so the PR is not ready to merge.

### Review provenance

🔴 3 blocking

<details>

```

In `packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs:207-225: Registration loses the reservation's exact height
  `finalize_transaction` stamps the selected inputs using `last_processed_height` before releasing the wallet lock and awaiting the external signer, but `register` independently reads the height after signing finishes. If signing spans at least five blocks, the registry's 20-block guard can still accept the token after key-wallet's 24-block TTL has swept and re-reserved its outpoint. A later release or definitive broadcast rejection can then remove the newer reservation because release operates unconditionally by outpoint. Store the height used by `set_current_height` in `SignedCoreTransaction` and transfer that exact value into the registry.

In `packages/rs-platform-wallet-ffi/src/core_wallet/signed_payment.rs`:
- [BLOCKING] packages/rs-platform-wallet-ffi/src/core_wallet/signed_payment.rs:81-115: Registration does not preserve unique reservation ownership
  `FFICoreTransaction` contains only serialized bytes and a fee, while this function independently accepts the wallet and funding account and does not consume the transaction. A valid transaction can therefore be registered repeatedly or registered against an unrelated wallet/account. Duplicate tokens defeat reservation-level single-use semantics: one token can release the outpoint, a new payment can reserve it, and another old token can later release that newer reservation. Registering wallet A's transaction against wallet B instead reconciles B and leaves A's actual reservation held until TTL. Consume a single opaque finalized-payment handle containing its originating wallet, account, reservation height, and transaction, or remove the ownership-less legacy registration path.

In `packages/rs-platform-wallet-ffi/src/wallet.rs`:
- [BLOCKING] packages/rs-platform-wallet-ffi/src/wallet.rs:392-423: Token lifecycle is tied to transient wallet aliases
  `platform_wallet_destroy` deletes every token when no sibling wrapper is currently visible, even though the manager still owns the logical wallet and can immediately return another handle for it. Closing the last current wrapper therefore makes a retained payment token stale while intentionally leaving its UTXO reserved until TTL. The `remove`, sibling scan, alias insertion, and registry sweep also use separate locks, so a concurrent lookup or registration can race the sweep. Conversely, an alias retained across logical wallet removal and recreation still matches `(WalletManager Arc, wallet_id)`, allowing an old token to operate on the replacement generation. Perform cleanup on logical wallet removal or manager shutdown and coordinate it with a stable per-wallet generation or atomic lifecycle reference count.
```
</details>

packages/rs-platform-wallet-ffi/src/wallet.rs
thepastaclaw: Resolved in this update — _platform_wallet_destroy's registry sweep is keyed on wallet identity, not handle identity — destroying one alias invalidates a sibling wrapper's deferred payment_ no longer present.

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
thepastaclaw: Resolved in this update — _Coroutine cancellation can discard a minted reservation token after buildSignedPayment's native call has already committed it_ no longer present.

packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs
thepastaclaw: Resolved in this update — _Token expiry guard measures a different height clock than the reservation it bounds (prior-1, still valid)_ no longer present.

PR #4095, opened 2026-07-11
feat(platform-wallet): return the exact network fee from send_payment

QuantumExplorer:

DashPay payment callers (dashwallet-ios in particular) need to show the exact network fee of the broadcast transaction. The transaction builder already computes this fee for every DashPay payment, but `send_payment` discarded it, so wallets could only re-estimate it after the fact.

This has been carried as a local patch on the dashwallet-ios platform pin branch and is flagged in that repo's DASHSYNC_MIGRATION.md as needing upstreaming — landing it here shrinks the next pin bump.

## What was done?

Thread the fee the builder already computes through the full stack:

- `PlatformWallet::send_payment` (rust): returns `(txid, entry, fee)` instead of dropping the fee.
- `platform_wallet_send_dashpay_payment` (FFI): gains a nullable `out_fee_duffs` out-param.
- `ManagedPlatformWallet.sendDashPayPayment` (Swift): returns `(txid: Data, feeDuffs: UInt64)`.
- `SendDashPayPaymentSheet` (SwiftExampleApp): adjusted to the tuple return.

## How Has This Been Tested?

- `cargo check -p platform-wallet -p platform-wallet-ffi --all-targets` passes.
- `cargo fmt --all --check` clean.
- SwiftExampleApp builds for the iOS simulator against a freshly rebuilt `DashSDKFFI.xcframework` containing the updated header.
- The same change has been exercised end-to-end on the dashwallet-ios pin branch, where the returned fee is displayed after a DashPay payment broadcast.

## Breaking Changes

None for consensus or the platform protocol. API-wise, `send_payment`'s return type and the FFI/Swift signatures change; the FFI out-param is nullable so existing C callers can pass NULL.

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - DashPay payment results now include the exact network fee in duffs, along with the transaction ID.
  - Swift SDK integrations can access the fee returned for each completed DashPay payment.
- **Updates**
  - Existing payment flows continue to display success and handle errors as before.
  - Applications using the updated payment API should accommodate the additional fee value in the result.

### DIFF
--- a/packages/rs-platform-wallet/src/wallet/identity/network/payments.rs
+++ b/packages/rs-platform-wallet/src/wallet/identity/network/payments.rs
@@ -552,6 +552,7 @@ impl<B: TransactionBroadcaster + ?Sized> DashPayView<'_, B> {
         (
             dashcore::Txid,
             crate::wallet::identity::types::dashpay::payment::PaymentEntry,
+            u64, // network fee in duffs, from the broadcast transaction
         ),
         PlatformWalletError,
     >
@@ -575,7 +576,7 @@ impl<B: TransactionBroadcaster + ?Sized> DashPayView<'_, B> {
         // re-acquires that (non-reentrant) lock internally.
         self.drain_pending_contact_crypto(provider).await;
 
-        let (payment_address, tx) = {
+        let (payment_address, tx, fee) = {
             let mut wm = self.wallet_manager.write().await;
 
             // Resolve the external account's xpub so we can derive addresses.
@@ -664,14 +665,14 @@ impl<B: TransactionBroadcaster + ?Sized> DashPayView<'_, B> {
             // `impl<S: Signer> TransactionSigner for S`) rather than the
             // resident `wallet`, so funding-input signatures are produced
             // from Keychain-derived keys without a resident seed.
-            let (tx, _fee) = builder
+            let (tx, fee) = builder
                 .build_signed(signer, |addr| {
                     managed_account.address_derivation_path(&addr)
                 })
                 .await
                 .map_err(|e| PlatformWalletError::TransactionBuild(e.to_string()))?;
 
-            (payment_address, tx)
+            (payment_address, tx, fee)
         };
 
         // --- 3. Broadcast the transaction, releasing the build's UTXO
@@ -726,7 +727,7 @@ impl<B: TransactionBroadcaster + ?Sized> DashPayView<'_, B> {
                 })?;
         }
 
-        Ok((txid, entry))
+        Ok((txid, entry, fee))
     }
 }
 
--- a/packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DashPay/SendDashPayPaymentSheet.swift
+++ b/packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DashPay/SendDashPayPaymentSheet.swift
@@ -364,7 +364,7 @@ struct SendDashPayPaymentSheet: View {
                 // useful to pass. The Rust-side
                 // `PaymentEntry.memo` slot stays available for
                 // future local-note wiring.
-                let txid = try await wallet.sendDashPayPayment(
+                let (txid, _) = try await wallet.sendDashPayPayment(
                     fromIdentityId: senderIdentity.identityId,
                     toContactIdentityId: contact.identityId,
                     amountDuffs: duffs,
--- a/packages/rs-platform-wallet-ffi/src/dashpay.rs
+++ b/packages/rs-platform-wallet-ffi/src/dashpay.rs
@@ -543,6 +543,7 @@ pub unsafe extern "C" fn platform_wallet_send_dashpay_payment(
     memo: *const c_char,
     core_signer_handle: *mut MnemonicResolverHandle,
     out_txid: *mut [u8; 32],
+    out_fee_duffs: *mut u64,
 ) -> PlatformWalletFFIResult {
     check_ptr!(core_signer_handle);
     check_ptr!(out_txid);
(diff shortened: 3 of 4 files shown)


thepastaclaw:
✅ Review complete (commit 5f4bb14b)

packages/rs-platform-wallet/src/wallet/identity/network/payments.rs
```diff
@@ -664,14 +665,14 @@ impl<B: TransactionBroadcaster + ?Sized> DashPayView<'_, B> {
...
-            let (tx, _fee) = builder
+            let (tx, fee) = builder
                 .build_signed(signer, |addr| {
                     managed_account.address_derivation_path(&addr)
                 })
                 .await
                 .map_err(|e| PlatformWalletError::TransactionBuild(e.to_string()))?;
 
-            (payment_address, tx)
+            (payment_address, tx, fee)
```
thepastaclaw, line 675: **🔴 Blocking: Piped-through fee is a size-based recomputation, not the actual inputs-minus-outputs fee**

I traced this through to `TransactionBuilder::build_signed` in the pinned `key-wallet` crate (rev `1ee1c94`, `key-wallet/src/wallet/managed_wallet_info/transaction_builder.rs`). The value now surfaced by this PR is `fee_rate.calculate_fee(encoded_size(&signed_tx))` — a fee-rate × final-serialized-size recomputation done *after* signing.

That is provably not the same quantity as the fee the broadcast transaction actually pays. During `assemble_unsigned`, the change output (if any) is sized as `total_input - total_output - selection.estimated_fee`, so the true on-chain fee (inputs minus outputs) always equals `selection.estimated_fee` — a value computed *before* signing from an estimated per-input size (148 bytes) and the coin-selector's dust-folding rules, not from the final signed size. The two numbers only match by coincidence:
- Dust case: when `change_amount` falls below the dust threshold, `accumulate_coins_with_size` folds the dust into `estimated_fee` (`total_value - target_amount`) so the true paid fee is inflated by up to 546 duffs. `build_signed`'s returned value is oblivious to this — it just does size × rate — so it undercounts the actual fee by exactly the folded-in dust amount whenever no change output is created.
- General case: the 148-byte/input estimate used for coin selection and the real encoded size of a signed P2PKH input (variable-length DER signatures) aren't guaranteed to match, so the two fee numbers can diverge even when a change output exists.

This PR's whole purpose is to let wallets show the *exact* fee instead of re-estimating after the fact — but the plumbed-through value inherits an estimate that's decoupled from what the change output / dust folding actually paid. Since `send_payment` previously discarded this value entirely (`_fee`), this inaccuracy had no consumer before; this PR is what turns it into an externally-relied-upon contract, which is why it's in scope here even though the root computation lives in the external `key-wallet` crate.

<sub>source: ['codex']</sub>

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Views/DashPay/SendDashPayPaymentSheet.swift
```diff
@@ -364,7 +364,7 @@ struct SendDashPayPaymentSheet: View {
                 // useful to pass. The Rust-side
                 // `PaymentEntry.memo` slot stays available for
                 // future local-note wiring.
-                let txid = try await wallet.sendDashPayPayment(
+                let (txid, _) = try await wallet.sendDashPayPayment(
```
thepastaclaw, line 367: **💬 Nitpick: In-repo reference implementation discards the newly-exposed fee**

This PR's stated motivation is that wallets could previously only re-estimate the fee after the fact. `SendDashPayPaymentSheet` is the in-tree DashPay send flow and the natural place to demonstrate the new capability, but it discards `feeDuffs` (`let (txid, _) = try await wallet.sendDashPayPayment(...)`) instead of surfacing it in the confirmation UI. Not a defect — the real consumer is dashwallet-ios — but the example app doesn't showcase the feature this PR adds.

<sub>source: ['claude']</sub>

thepastaclaw (changes requested):
## Code Review

Small plumbing change that threads an already-computed fee value from TransactionBuilder::build_signed through send_payment -> FFI out-param -> Swift tuple. The Swift/FFI marshalling is correct and safe (nullable out-param, exhaustive tuple matching, proper lifetime handling of the resolver signer). However, tracing the fee value back to its source in the vendored key-wallet crate (TransactionBuilder::build_signed, pinned rev 1ee1c94) shows it is a size*rate recomputation from the final signed tx, not the actual inputs-minus-outputs amount baked into the transaction during coin selection/change calculation — these two values provably diverge whenever a dust remainder is folded into the fee (no change output created) and can diverge more generally from encoded-size estimation differences. This directly undermines the PR's stated goal of exposing the "exact network fee."

Source (experiment sonnet-primary-opus-quarter-sample-20260710, cohort sonnet_primary, bucket 2): reviewers codex/general=gpt-5.6-sol(completed_parseable_output_with_nonzero_wrapper_exit); sonnet5/general=claude-sonnet-5(completed); verifier=verifier-sonnet5-4095-1783808164=claude-sonnet-5; orchestrator=openai/gpt-5.6-sol reasoning=high (orchestration-only, not a reviewer/verifier).

🔴 1 blocking | 💬 1 nitpick(s)

<details>

```

- [BLOCKING] packages/rs-platform-wallet/src/wallet/identity/network/payments.rs:668-675: Piped-through fee is a size-based recomputation, not the actual inputs-minus-outputs fee
  I traced this through to `TransactionBuilder::build_signed` in the pinned `key-wallet` crate (rev `1ee1c94`, `key-wallet/src/wallet/managed_wallet_info/transaction_builder.rs`). The value now surfaced by this PR is `fee_rate.calculate_fee(encoded_size(&signed_tx))` — a fee-rate × final-serialized-size recomputation done *after* signing.

  That is provably not the same quantity as the fee the broadcast transaction actually pays. During `assemble_unsigned`, the change output (if any) is sized as `total_input - total_output - selection.estimated_fee`, so the true on-chain fee (inputs minus outputs) always equals `selection.estimated_fee` — a value computed *before* signing from an estimated per-input size (148 bytes) and the coin-selector's dust-folding rules, not from the final signed size. The two numbers only match by coincidence:
  - Dust case: when `change_amount` falls below the dust threshold, `accumulate_coins_with_size` folds the dust into `estimated_fee` (`total_value - target_amount`) so the true paid fee is inflated by up to 546 duffs. `build_signed`'s returned value is oblivious to this — it just does size × rate — so it undercounts the actual fee by exactly the folded-in dust amount whenever no change output is created.
  - General case: the 148-byte/input estimate used for coin selection and the real encoded size of a signed P2PKH input (variable-length DER signatures) aren't guaranteed to match, so the two fee numbers can diverge even when a change output exists.

  This PR's whole purpose is to let wallets show the *exact* fee instead of re-estimating after the fact — but the plumbed-through value inherits an estimate that's decoupled from what the change output / dust folding actually paid. Since `send_payment` previously discarded this value entirely (`_fee`), this inaccuracy had no consumer before; this PR is what turns it into an externally-relied-upon contract, which is why it's in scope here even though the root computation lives in the external `key-wallet` crate.
```
</details>

PR #4098, opened 2026-07-12
refactor(swift-sdk): promote the testnet faucet client into the SDK package

QuantumExplorer:
## Issue being fixed or feature implemented

The testnet faucet client (faucet.thepasta.org HTTP client + cap.js proof-of-work solver) lived inside SwiftExampleApp, so other SDK consumers (dashwallet-ios QA builds in particular) couldn't reuse it. dashwallet-ios has been carrying this promotion on its platform pin branch; upstreaming it shrinks the pin.

## What was done?

Moved `SwiftExampleApp/Core/Services/TestnetFaucetService.swift` → `Sources/SwiftDashSDK/Utils/TestnetFaucet.swift` (95% rename), renaming the type to `TestnetFaucet` and adjusting access levels for the package boundary. Pure code motion — no behavior change; testnet-only tooling, never invoked on mainnet paths.

## How Has This Been Tested?

- SwiftExampleApp builds for the iOS simulator on this branch (both the SDK package and the app's existing faucet call sites compile).
- The promoted client is exercised regularly from dashwallet-ios QA builds on the pin branch (funding testnet wallets via the in-app faucet button).

## Breaking Changes

None for the SDK (additive API). Within SwiftExampleApp the type moved/renamed, which is app-internal.

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added or updated relevant unit/integration/functional/e2e tests
- [ ] I have added "!" to the title and described breaking changes in the corresponding section if my code contains any
- [ ] I have made corresponding changes to the documentation if needed

**For repository code-owners and collaborators only**
- [ ] I have assigned this pull request to a milestone

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4049, opened 2026-07-09
fix(platform-wallet): report the exact send_payment fee (inputs − outputs)

QuantumExplorer:
Two commits used by dashwallet-ios (the DashPay contacts arc, dashwallet-ios #787), cherry-picked clean onto `v4.1-dev`:

1. **Fee threading** — `send_payment` returns the exact network fee of the broadcast transaction; `platform_wallet_send_dashpay_payment` gains a nullable `out_fee_duffs` out-param; Swift `sendDashPayPayment` returns `(txid: Data, feeDuffs: UInt64)`. Lets wallet UIs show the real fee on the payment success screen instead of an estimate.

2. **Testnet faucet client promoted into the SDK package** — `TestnetFaucet` + the cap.js proof-of-work solver move from SwiftExampleApp into `Sources/SwiftDashSDK/Utils` so dashwallet-ios can drive the faucet in-app without carrying a drifting copy (public API unchanged; the example app picks it up through its existing import). Includes the example-app call-site fix for the new `sendDashPayPayment` tuple return.

Verified: `cargo check -p platform-wallet -p platform-wallet-ffi` clean on this branch; both SwiftExampleApp and dashwallet-ios build against these commits (on the pre-rebase base); the fee display and the faucet request were exercised live on testnet (faucet txid `395eeb08…`).

Not included (owned elsewhere): the tx-decode commit (PR #3981) and the `RawKeySigner` extraction (`47a83cf6d0`, Row #27's branch).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - DashPay payments now return the exact network transaction fee alongside the transaction ID.
  - Swift SDK users can access the payment fee in duffs after sending a payment.

- **Bug Fixes**
  - Improved fee reporting to reflect the actual broadcast transaction fee.

PR #4100, opened 2026-07-12
fix(swift-sdk): cap the faucet captcha's aggregate proof-of-work

QuantumExplorer:
Hardening for the `TestnetFaucet` client that merged in #4098: its per-field challenge bounds (`c ≤ 256`, `d ≤ 6`) still admit an aggregate of ~4.3B expected SHA-256 attempts, so a misbehaving or compromised same-domain faucet endpoint can pin the device's cores across 256 parallel solver tasks until the 30-second timeout on every faucet tap. Add an aggregate bound (`c × 16^d ≤ 64M`, ~10× the live faucet's actual soft challenge of c=100, d=4 ≈ 6.6M) so oversized challenges throw and route to the web-faucet fallback instead of burning CPU/battery.

Split out of #4049 when that PR was narrowed to the `send_payment` fee fix — this commit was originally review hardening there (`ad29deae05`) and was deliberately parked for a standalone PR rather than dropped. One hunk, comment-documented; Swift parse-checked.

Note: the faucet client's placement in the SDK package is itself under review (it may move back app-side); if #4098 is reverted, this rides along with the file.

### DIFF
--- a/packages/swift-sdk/Sources/SwiftDashSDK/Utils/TestnetFaucet.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/Utils/TestnetFaucet.swift
@@ -273,7 +273,9 @@ public struct TestnetFaucet {
         // few seconds of parallel hashing.
         let expectedWork = Double(c) * pow(16, Double(d))
         guard expectedWork <= 64_000_000 else {
-            throw FaucetError.message("Captcha challenge too expensive (c=\(c), d=\(d))")
+            throw FaucetError.message(
+                "Captcha challenge too expensive (c=\(c), d=\(d), "
+                    + "expectedWork=\(Int(expectedWork)) > 64000000)")
         }
 
         // Solve all c sub-challenges off the main actor, parallelized, under
--- a/packages/swift-sdk/Sources/SwiftDashSDK/Utils/TestnetFaucet.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/Utils/TestnetFaucet.swift
@@ -265,6 +265,17 @@ public struct TestnetFaucet {
             throw FaucetError.message("Unsupported captcha challenge (c=\(c), s=\(s), d=\(d))")
         }
 
+        // The per-field bounds alone still admit c=256, d=6 ≈ 4.3B expected
+        // attempts — enough to pin the CPU until the 30 s timeout on every
+        // tap. Cap the AGGREGATE expected work (c × 16^d). The live faucet's
+        // soft challenge is c=100, d=4 ≈ 6.6M attempts; 64M gives ~10×
+        // headroom for server-side tuning while keeping the worst case to a
+        // few seconds of parallel hashing.
+        let expectedWork = Double(c) * pow(16, Double(d))
+        guard expectedWork <= 64_000_000 else {
+            throw FaucetError.message("Captcha challenge too expensive (c=\(c), d=\(d))")
+        }
+
         // Solve all c sub-challenges off the main actor, parallelized, under
         // an overall timeout so even an in-bounds-but-slow challenge can't
         // hang the UI indefinitely (the timeout cancels the solve tasks).


thepastaclaw:
✅ Review complete (commit 7ab9aa6b)

thepastaclaw (commented):
## Code Review

This is a small, well-scoped hardening fix that adds an aggregate proof-of-work bound (c × 16^d ≤ 64M) after the existing per-field bounds check in TestnetFaucet.swift's solveCaptcha. Both agents verified the arithmetic against the stated live-faucet baseline (~6.6M) and worst-case (~4.3B) and confirmed the guard is correctly placed before any solver tasks spawn. Only one low-severity documentation nitpick was found; no correctness, security, or consensus-relevant issues.

Source (experiment sonnet-primary-opus-quarter-sample-20260710, cohort sonnet_primary, bucket 2, trigger new_pr, specialists none): reviewers codex/general=gpt-5.6-sol(completed); sonnet5/general=claude-sonnet-5(completed); verifier=verifier-sonnet5-4100-1783853670=claude-sonnet-5; orchestrator=openai/gpt-5.6-sol reasoning=high (orchestration-only, not a reviewer/verifier).

💬 1 nitpick(s)

packages/swift-sdk/Sources/SwiftDashSDK/Utils/TestnetFaucet.swift
```diff
@@ -265,6 +265,17 @@ public struct TestnetFaucet {
...
 
+        // The per-field bounds alone still admit c=256, d=6 ≈ 4.3B expected
+        // attempts — enough to pin the CPU until the 30 s timeout on every
+        // tap. Cap the AGGREGATE expected work (c × 16^d). The live faucet's
+        // soft challenge is c=100, d=4 ≈ 6.6M attempts; 64M gives ~10×
+        // headroom for server-side tuning while keeping the worst case to a
+        // few seconds of parallel hashing.
+        let expectedWork = Double(c) * pow(16, Double(d))
+        guard expectedWork <= 64_000_000 else {
+            throw FaucetError.message("Captcha challenge too expensive (c=\(c), d=\(d))")
```
thepastaclaw: **💬 Nitpick: Error message omits the computed aggregate work value**

The thrown message reports the raw `c` and `d` inputs but not the derived `expectedWork` or the 64M threshold it was compared against. This mirrors the existing per-field guard's style (which also only echoes raw inputs), so it's consistent with the surrounding code rather than a regression — but since this is a numeric threshold check specifically, including the computed value (e.g. `expectedWork=\(Int(expectedWork))`) would make it faster to triage if a faucet endpoint starts drifting toward the cap in production logs.

<sub>source: ['sonnet5-general']</sub>

QuantumExplorer: Fixed in 7ab9aa6b9a — the guard's error now includes the computed value and the threshold: `expectedWork=<n> > 64000000` alongside the raw c/d, so logs show how close an endpoint is drifting toward the cap.

thepastaclaw (approved):
## Code Review

This new_push follow-up (b0f1ef11 → 7ab9aa6b) makes exactly one change: the aggregate PoW-cap error message now reports the computed expectedWork value and the 64,000,000 threshold alongside c and d, directly resolving the sole prior nitpick. The guard's math and placement are unchanged and remain correct (c × 16^d ≤ 64M, checked before any solver tasks spawn). No new issues in this delta or across the cumulative PR.

Source (experiment sonnet-primary-opus-quarter-sample-20260710, cohort sonnet_primary, bucket 2, trigger new_push, specialists none): reviewers codex/general=gpt-5.6-sol(completed); sonnet5/general=claude-sonnet-5(completed); verifier=verifier-sonnet5-4100-1783854540=claude-sonnet-5; orchestrator=openai/gpt-5.6-sol reasoning=high (orchestration-only, not a reviewer/verifier).

PR #4101, opened 2026-07-12
fix(drive-abci): bound value decoding in contested-resource queries

QuantumExplorer:

The contested-resource voting queries decoded caller-supplied serialized index and cursor values without complete per-value or aggregate allocation bounds. Malformed input could trigger unreasonable allocation and affect node availability.

## What was done?

- Added a shared bounded decoder with a 4 KiB encoded-value ceiling and a separate 64 KiB decoded-memory budget.
- Added pre-decode validation of aggregate encoded bytes and cursor count against the contested index arity in all three query handlers.
- Rejected trailing bytes so cursor encodings are canonical.
- Removed complete attacker-controlled payloads from validation error strings while preserving field and position context.
- Preserved trusted response encoding behavior.

Affected inputs:
- getContestedResources: start_index_values, end_index_values, and start_at_value_info.start_value
- getContestedResourceVoteState: index_values
- getContestedResourceVotersForIdentity: index_values

## How Has This Been Tested?

- cargo test -p drive-abci --locked: 2,553 library tests, 90 strategy tests, and 12 doctests passed; zero failures.
- Focused decoder and aggregate-budget tests cover declared-length allocation, per-value encoded limits, aggregate size, cursor count, trailing bytes, valid scalar values, and valid nested values.
- Real DPNS handler tests cover malformed and aggregate-over-limit input through all three affected public query methods.
- cargo clippy -p drive-abci --all-targets --locked -- -D warnings passed.
- cargo fmt --all -- --check passed.

## Breaking Changes

None. Valid contested-index values remain well below the encoded ceiling. Existing field and position error context is preserved, while rejected payload contents are no longer echoed.

## Checklist:


thepastaclaw:
✅ Review complete (commit a6a61c89)

packages/rs-drive-abci/src/query/voting/contested_resources/v0/mod.rs
```diff
@@ -81,25 +81,16 @@ impl<C> Platform<C> {
             ))));
         }
 
-        let bincode_config = bincode::config::standard()
-            .with_big_endian()
-            .with_no_limit();
-
         let start_index_values = match start_index_values
```
QuantumExplorer, line 100: The 64 KiB budget resets for every repeated value, but this handler validates neither component count nor aggregate encoded/decoded size before collecting the results; the voters-for-identity handler has the same gap. Since rs-dapi accepts 64 MiB requests, many compact nested Values can each consume nearly the full decode budget and multiply into multi-gigabyte allocations before downstream arity checks. Please validate counts against the contested index and cap aggregate bytes before decoding.

packages/rs-drive-abci/src/query/voting/contested_resources/v0/mod.rs
```diff
@@ -81,25 +81,16 @@ impl<C> Platform<C> {
...
-                    .map_err(|_| {
-                        QueryError::InvalidArgument(format!(
-                            "could not convert {:?} to a value in the start index values at position {}",
-                            serialized_value, pos
-                        ))
-                    })?
-                    .0)
+                super::super::decode_serialized_index_value(serialized_value.as_slice(), || {
+                    format!(
+                        "could not convert {:?} to a value in the start index values at position {}",
```
QuantumExplorer: Every decode rejection formats the complete attacker-controlled byte vector. In particular, the helper rejects a value above 64 KiB and then invokes this closure; a near-64 MiB permitted gRPC field can therefore expand into a several-hundred-megabyte error string. Please report the field position and bounded length or prefix instead of rendering the entire payload. The other two handlers repeat this pattern.

packages/rs-drive-abci/src/query/voting/contested_resources/v0/mod.rs
QuantumExplorer: Fixed in a6a61c8931. All three handlers now cap cursor count against the contested index arity and enforce a 4 KiB aggregate encoded-input budget before decoding. Added helper and real-handler regressions.

packages/rs-drive-abci/src/query/voting/contested_resources/v0/mod.rs
QuantumExplorer: Fixed in a6a61c8931. Rejection messages no longer render attacker-controlled cursor contents; they retain bounded field and position context. Added an oversized-handler regression that asserts the fixed error text.

packages/rs-drive-abci/src/query/voting/decode_index_value.rs
```diff
@@ -0,0 +1,189 @@
...
+        // Reject trailing bytes so each encoded cursor value is canonical.
+        if consumed != serialized_value.len() {
+            return None;
+        }
+
+        Some(value)
+    }
+
+    try_decode(serialized_value).ok_or_else(|| QueryError::InvalidArgument(invalid_argument()))
+}
```
thepastaclaw, line 96: **🔴 Blocking: [Carried forward: prior-recursion-depth — STILL VALID] Byte/allocation ceilings still don't bound recursion depth**

This PR tightens the encoded-value ceiling from 64 KiB to 4 KiB and adds an aggregate request cap, but neither limit bounds recursion depth. platform_value::Value still decodes recursively through Array and Map, and bincode 2.0.1 has no depth guard. The canonical worst case—2,047 nested Array(len=1) levels terminated by Null—is exactly 4,095 bytes and passes every byte, allocation, aggregate, and trailing-byte check. Independent reviewer probes reproduced a hard stack-overflow abort when decoding this payload on an 8 MiB dev-profile thread matching Drive-ABCI's configured worker stack; the same payload survives with release optimization. Both results matter because the release workflow publishes dashpay/drive:<version>-debug with cargo_profile=dev, while the root Dockerfile and dashmate local-build path also default to dev. Relying on optimizer-dependent frame size leaves unauthenticated query callers able to abort supported node artifacts and could regress with toolchain, target, or Value-layout changes. Add an explicit nesting-depth guard or iterative decoder, matching the value_from_proto_at_depth pattern already used for hostile nested values.

<sub>source: ['codex-general', 'sonnet5-general', 'codex-rust-quality', 'sonnet5-rust-quality', 'sonnet5-security-auditor']</sub>

packages/rs-drive-abci/src/query/voting/decode_index_value.rs
```diff
@@ -0,0 +1,189 @@
...
+        count = count.saturating_add(1);
+        total_bytes = total_bytes.saturating_add(serialized_value.len());
+
+        if count > max_values || total_bytes > MAX_ENCODED_INDEX_VALUE_BYTES {
+            return Err(QueryError::InvalidArgument(invalid_argument()));
+        }
+    }
+
+    Ok(())
+}
```
thepastaclaw, line 55: **🟡 Suggestion: Separate the per-value and aggregate encoded-byte budgets**

MAX_ENCODED_INDEX_VALUE_BYTES is documented and checked as a per-value ceiling, but validate_serialized_index_values reuses it for the total bytes across the entire request. The two invariants happen to share a value today, but tuning one silently changes the other. Introduce a separately named aggregate budget so each safety boundary is independently documented and auditable.

<sub>source: ['sonnet5-rust-quality']</sub>

packages/rs-drive-abci/src/query/voting/contested_resources/v0/mod.rs
```diff
@@ -81,25 +81,32 @@ impl<C> Platform<C> {
...
+                .chain(
+                    start_at_value_info
+                        .iter()
+                        .map(|info| info.start_value.as_slice()),
+                ),
+            index.properties.len(),
+            || "serialized index values exceed the contested index query limits".to_string(),
+        ) {
+            return Ok(QueryValidationResult::new_with_error(error));
+        }
```
thepastaclaw, line 98: **💬 Nitpick: Pin the aggregate arity relationship with a boundary test**

The max_values bound is correct because the downstream query enforces start_index_values.len() + end_index_values.len() <= properties.len() - 1, leaving one slot for start_at_value_info. Add a boundary test with both ranges near full arity plus start_at so future changes to either module cannot silently drift out of sync.

<sub>source: ['sonnet5-rust-quality']</sub>

thepastaclaw (changes requested):
## Code Review

Incremental and cumulative verification after cabb5f131a confirms that the aggregate count/byte cap is correctly applied and all 51 targeted voting tests, formatting, and clippy checks pass.

Carried-forward prior findings: prior-recursion-depth is STILL VALID—the exact 4,095-byte/2,047-level canonical payload aborts an 8 MiB dev/debug-profile thread while surviving under release optimization; because the repository publishes a supported dev-profile Drive debug image and local builds default to dev, the availability defect remains in scope.

New findings in the latest delta: one architecture suggestion and one boundary-test nitpick.

Source: reviewers gpt-5.6-sol (general, security-auditor, rust-quality) and claude-sonnet-5 (general, security-auditor, rust-quality); final verifier claude-sonnet-5. Orchestrator openai/gpt-5.6-sol (reasoning=high, orchestration-only).

<details>

```

In `packages/rs-drive-abci/src/query/voting/decode_index_value.rs`:
- [BLOCKING] packages/rs-drive-abci/src/query/voting/decode_index_value.rs:20-96: [Carried forward: prior-recursion-depth — STILL VALID] Byte/allocation ceilings still don't bound recursion depth
  This PR tightens the encoded-value ceiling from 64 KiB to 4 KiB and adds an aggregate request cap, but neither limit bounds recursion depth. platform_value::Value still decodes recursively through Array and Map, and bincode 2.0.1 has no depth guard. The canonical worst case—2,047 nested Array(len=1) levels terminated by Null—is exactly 4,095 bytes and passes every byte, allocation, aggregate, and trailing-byte check. Independent reviewer probes reproduced a hard stack-overflow abort when decoding this payload on an 8 MiB dev-profile thread matching Drive-ABCI's configured worker stack; the same payload survives with release optimization. Both results matter because the release workflow publishes dashpay/drive:<version>-debug with cargo_profile=dev, while the root Dockerfile and dashmate local-build path also default to dev. Relying on optimizer-dependent frame size leaves unauthenticated query callers able to abort supported node artifacts and could regress with toolchain, target, or Value-layout changes. Add an explicit nesting-depth guard or iterative decoder, matching the value_from_proto_at_depth pattern already used for hostile nested values.
- [SUGGESTION] packages/rs-drive-abci/src/query/voting/decode_index_value.rs:20-55: Separate the per-value and aggregate encoded-byte budgets
  MAX_ENCODED_INDEX_VALUE_BYTES is documented and checked as a per-value ceiling, but validate_serialized_index_values reuses it for the total bytes across the entire request. The two invariants happen to share a value today, but tuning one silently changes the other. Introduce a separately named aggregate budget so each safety boundary is independently documented and auditable.
```
</details>

QuantumExplorer (commented):
Human Reviewed

packages/rs-drive-abci/src/query/voting/contested_resources/v0/mod.rs
QuantumExplorer: Addressed in follow-up PR #4104. The new boundary test exercises a DPNS contested-index request whose range cursor plus start-at cursor exactly consume the index arity and verifies that the query remains accepted.

packages/rs-drive-abci/src/query/voting/decode_index_value.rs
QuantumExplorer: Agreed and addressed in follow-up PR #4104. The aggregate request ceiling now has its own MAX_AGGREGATE_ENCODED_INDEX_VALUES_BYTES constant, independent of the per-value limit.

packages/rs-drive-abci/src/query/voting/decode_index_value.rs
QuantumExplorer: Agreed and addressed in follow-up PR #4104. It adds an explicit Array/Map nesting limit of 64 and a regression using the exact 4,095-byte, 2,047-level payload on an 8 MiB thread; the payload now returns InvalidArgument without overflowing the stack.

Issue #4092, opened 2026-07-11
Wire iOS managed identity top-up from asset lock + converge the 3 top-up paths (v4.1-dev)


shumkov:
## Summary

Converge identity **top-up from an asset lock** onto a single managed path used by both iOS and Android, and wire it on iOS (currently unimplemented). Land it as a standalone PR against **`v4.1-dev`** so the shared, iOS-affecting change is reviewed on its own rather than buried inside the large Kotlin-SDK PR (#3999).

## Background — three fragmented paths today

There are currently **three** ways to top up an identity from a Core asset lock, at different layers, and none is shared:

| Path | Layer | Asset-lock source | Origin | Wired to |
|---|---|---|---|---|
| `dash_sdk_identity_topup_with_instant_lock` (+`_and_wait`) | DPP-SDK (`rs-sdk-ffi/src/identity/topup.rs`) | caller supplies a pre-built **InstantSend** proof (`instantLock`+`tx`+`outputIndex`) | original iOS epic #2756 | **nothing** — unwired on iOS |
| `platform_wallet_top_up_identity_with_funding_signer` | platform-wallet (managed) | builds a new asset lock `FromWalletBalance`, **IS→CL fallback** | PR #3999 (commit `7a1d04792f`) | **Android only** |
| `platform_wallet_topup_identity_with_existing_asset_lock_signer` | platform-wallet (managed) | consumes an already-tracked lock `FromExistingAssetLock` | `feat/dip15-dashpay-invitations` (#4041) | Swift reclaim |

The DPP-SDK primitive is **IS-only** (registration has both `..._with_instant_lock` **and** `..._with_chain_lock`; top-up only ever got the IS variant) and pushes all asset-lock creation / funding / proof acquisition onto the caller. The **managed** library orchestrator is the correct "from asset lock" abstraction.

## The good news — most of the work already exists upstream

- The managed orchestrator **`IdentityWallet::top_up_identity_with_funding`** is **already in `v4.1-dev`** (`packages/rs-platform-wallet/src/wallet/identity/network/registration.rs:388`). It does the full lifecycle: build/resolve the asset lock, **IS→CL fallback**, retries, persist balance.
- iOS **already uses the registration twin** `platform_wallet_register_identity_with_funding_signer` (`packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/ManagedPlatformWallet.swift:3598`), so the managed-funding-signer pattern is proven on iOS.
[...]

PR #3999, opened 2026-07-04
feat(sdk): add Kotlin SDK and KotlinExampleApp (Android port of SwiftExampleApp)

bezibalazs:
## Issue being fixed or feature implemented

The Kotlin/Android SDK has been planned but never started (`docs/SDK_ARCHITECTURE.md`, `book/src/sdk-support.md` list it as "coming"). This PR delivers it, together with **KotlinExampleApp** — a one-for-one Android port of SwiftExampleApp — so Android has the same reference integration iOS has.

## What was done?

Three new components (~50k insertions, mirroring the `packages/swift-sdk` layering):

- **`packages/rs-unified-sdk-jni`** — Rust JNI cdylib (110 exports, arm64-v8a + x86_64, 16KB-aligned for Android 15). Calls the `extern "C"` entry points of `rs-sdk-ffi`/`platform-wallet-ffi`/`key-wallet-ffi` as **rlib dependencies** (no C glue, no cbindgen headers on Android). Panics are caught at every export; Rust→Kotlin callbacks (32-slot persistence vtable, async signer, mnemonic resolver, sync events) attach Tokio threads as JVM daemons and copy payloads before return.
- **`packages/kotlin-sdk/sdk`** — the Kotlin SDK (`org.dashfoundation.dashsdk`): 28 Room entities transcribed 1:1 from the SwiftData models, Keystore-wrapped secret storage (`org.dashfoundation.wallet.*` aliases), network-locked `PlatformWalletManager`/`WalletManagerStore`, sync services, per the persist/load/bridge doctrine (`kotlin-sdk/CLAUDE.md`, ported from `swift-sdk/CLAUDE.md`).
- **`packages/kotlin-sdk/KotlinExampleApp`** — single-activity Compose app: 5 tabs, wallet create/seed-backup/send/receive with QR, identity registration coordinators, DPNS, contracts/documents/storage explorer, the TokenActionScaffold with 12 token actions, transitions catalog, asset-lock + shielded funding, DashPay, diagnostics. **`packages/kotlin-sdk/PARITY.md` tracks all 90 Swift views: 75 ported / 8 partial / 7 deferred — every partial/deferred row names the exact missing FFI export.**

Cross-cutting changes reviewers should look at:

[...]

PR #4106, opened 2026-07-13
refactor(sdk): dedup shared wallet code + remove dead FFI/JNI chains

shumkov:

Cleanup from the duplication/slop review of the Kotlin-SDK PR (#3999), stacked on `feat/kotlin-sdk-and-example-app`. Three kinds of change, all behavior-preserving. Net **−651 lines**.

### Dedup
Extracted the byte-identical candidate-address-SET union block (wallet-manager read lock → resolve payment account → union the derived pool with the hydrated `address_balances` map) from `transfer.rs::auto_select_inputs` and `withdrawal.rs::plan_withdrawal` into `PlatformAddressWallet::candidate_address_set`. Both are **live spend paths** and the block was copy-pasted (`8bd3a858db` retrofitted transfer onto withdrawal's shape) — a divergence risk. Behavior unchanged: each caller keeps its own empty-handling (transfer tolerates an empty set; withdrawal errors early), and balances are still read fresh from chain via `AddressInfo::fetch_many` at each site.

### Dead-code removal
Bindings declared but never called — verified **zero callers in-repo and org-wide** (`gh search code --owner dashpay`); the Kotlin bindings are all `internal` (not public API). Each removes both the Kotlin `external fun` and its `Java_...` JNI trampoline.

- **Handle-based identity-derive chain** `platform_wallet_derive_identity_private_key_at_slot` (~465 LOC). It deadlocks in its only intended context — the identity-key persist callback fires while platform-wallet holds the wallet-manager **write** lock, and this export re-takes a **read** lock in its capability check — so every real flow uses the resolver-keyed sibling `dash_sdk_derive_identity_key_at_slot_with_resolver` (the same one iOS uses). Removed the Rust export + `IdentityPrivateKeyFFI`, its `lib.rs` module/re-export, the JNI trampoline + import (kept `net_from_ord`), and the Kotlin `deriveIdentityPrivateKey` binding.
- **`createCallbacks` / `destroyCallbacks`** — an alternate persistence-vtable entry point superseded by the live `WalletManagerNative.nativeCreate → build_vtable` path; the whole `PersistenceNative` object was dead, so the file is removed (kept `build_vtable` / `KotlinPersistenceCtx`, which the live path uses).
- **`managedIdentityDashPayProfile`**, **`resolverSupportsKeyType`**, **`walletGetId`**.

Also drops the now-unused `guard` / `JClass` / `jlong` imports and refreshes the persistence module docs (incl. a stale "wiring lands in a later milestone" note).

### Minor cleanups
- Dropped two no-op `#[allow(clippy::too_many_arguments)]` on the top-up exports (6 args each, below clippy's threshold).
- Replaced a misleading `Identifier::from_bytes(&[u8;32]).unwrap_or_default()` with the infallible `Identifier::from([u8;32])` — the input is a fixed 32-byte array, so the length check never fails and the zero-default was unreachable (it read as if a malformed id silently became all-zeros).
- Removed a stale `ADDR-02` test-case reference from a `transfer.rs` comment.

## How Has This Been Tested?
- `cargo check -p platform-wallet -p platform-wallet-ffi -p rs-unified-sdk-jni` — clean (warnings-as-errors).
- `cargo fmt --check` — clean.
- The `candidate_address_set` regression tests pass through the new helper: `plan_withdrawal_sizes_inputs_from_chain_not_doubled_cache`, `plan_withdrawal_finds_candidates_from_balance_map_when_pool_empty`.
- Every removed symbol was verified to have zero callers before deletion; the Kotlin edits are exercised by the Kotlin-SDK CI build.

## Breaking Changes
None. All removed symbols were uncalled (internal bindings, no public API, no org-wide reference); the dedup is behavior-preserving.

## Checklist:
- [x] I have added or updated relevant unit/integration/functional/e2e tests (existing regression tests cover the dedup; dead-code removal needs none)
- [x] I have added "!" to the title and described breaking changes in the corresponding section if my code contains any (n/a — none)

---

### Deferred (from the same #3999 review, with reasons)
- **B2** (register-contract-for-proof-verification dup) — cosmetic; deduping adds ~as much code as it removes in a core file.
- **B3** (top-up account persistence round) — a forced abstraction across two load-bearing persistence paths; ~4 shared lines, not worth the risk.
- **B4 / B5** (FFI capability-check ×6, build-preview-row ×5) — real duplication, but ~⅔ of the copies are in pre-existing files untouched by #3999 → belongs in a standalone v4.1-dev hygiene PR, not stacked here.
- **D4** (`release_row` / `free_rows`) — on inspection these are named, documented, unit-tested single-zeroization-site helpers, not vestigial; left as-is.
- Identity-top-up-from-asset-lock path convergence is tracked in #4092.


thepastaclaw:
🕓 Ready for review — 11 ahead in queue (commit 011c866b)
_Queue position: 12/12_

thepastaclaw (approved):
## Code Review

Verified this is a byte-for-byte behavior-preserving extraction: the shared candidate_address_set helper in wallet.rs is identical to the two previously duplicated inline blocks in transfer.rs and withdrawal.rs, including error variants, address union/conversion, and lock release timing before the network fetch. cargo check passes cleanly and no in-scope issues survive review.

Source: reviewers gpt-5.6-sol (Sol reviewer), claude-sonnet-5 (Sonnet reviewer), and claude-opus-4-8 (Opus reviewer), general lane; verifier claude-sonnet-5. Orchestrator openai/gpt-5.6-sol (orchestration-only).

PR #4107, opened 2026-07-13
fix(platform-wallet): persist address usage discovered during SPV block processing

QuantumExplorer:

On a mainnet wallet whose seed owns masternode provider keys, SPV block processing correctly detects a ProRegTx payload match on the Provider Owner Keys address (verified with txid `5076f9ed731bcb7b5ddf47e21bf1262d22d99665ca89bb6329e6e706d605da85` at height 1030673) and `key-wallet` marks the address used in memory — but nothing ever reaches the persistence layer. The `WalletEvent` bus carries neither the `mark_address_used` results nor the pools' `highest_used` bumps, so:

- every mirrored store keeps `is_used = false` on matched addresses (provider keys and regular BIP44 alike), and
- `WalletChangeSetFFI::from_changeset` hardcoded `has_external_highest_used` / `has_internal_highest_used` to `false`, so `PersistentAccount.externalHighestUsed` stays `-1` on every account in every store.

## What was done?

- **`CoreChangeSet` gains two deltas** (`rs-platform-wallet`): `addresses_marked_used` (the authoritative post-mark `AddressInfo` per involved address, as `DerivedAddressInfo`) and `account_highest_used` (per-account external/internal pool watermarks via a new `HighestUsedIndexes` type). Merge dedups marked-used entries on `(account_type, pool_type, index)` — same discipline as `addresses_derived` — and folds watermarks monotonic-max per slot; both fields gate `is_empty` so a usage-only event still persists.
- **Event-bridge population** (`core_bridge.rs`): `collect_usage_deltas` re-runs the same matcher the wallet used during processing — the read-only `ManagedAccountCollection::check_transaction` — against each record's transaction after processing, then resolves every involved address back to its owning pool. That recovers the pool type (which the match doesn't carry) plus the authoritative `AddressInfo` and `highest_used`, covering both TXO matches (BIP44) and special-tx payload matches (provider owner/voting keys). Single-pool accounts (provider keys, identity funding) surface their watermark in the external slot, matching the account row's two persisted fields.
- **FFI emit** (`rs-platform-wallet-ffi`): marked-used entries are bucketed per `(account_type, pool_type)` and fanned out through the existing `on_persist_account_address_pools_fn` pipeline, deliberately **after** the derived-address emit — the Swift persister overwrites `isUsed` with the latest emit, so derive-then-use within one round lands `true`. `from_changeset` now fills `external/internal_highest_used` + `has_*` flags from the changeset; values are authoritative pool state (not batch maxima) and the flags stay `false` on no-usage batches, so stored watermarks can never regress.

No Swift changes are required — `persistAccountAddresses` and the account-changeset handler already consume both surfaces.

## How Has This Been Tested?

- New unit tests: merge dedup of marked-used entries, monotonic-max watermark merge (lower value / `None` never regresses a slot), usage-only changesets are non-empty (`rs-platform-wallet`), and `(account, pool)` grouping incl. a `ProviderOwnerKeys`/Absent-pool case with `used == true` preserved (`rs-platform-wallet-ffi`).
- `cargo test -p platform-wallet --lib` (415 passed) and `cargo test -p platform-wallet-ffi --lib` (141 passed).
- `cargo clippy -p platform-wallet -p platform-wallet-ffi --all-targets` clean; `cargo check --all-features --tests` clean on both crates; `cargo fmt --all`.
- On-device verification pending: mainnet wallet import + "Rescan Everything" should show Provider Owner Keys address #0 (`XkxD3qhMb7uaQg7rvm9ZNCzyYhZ7tLPQc2`) as used with Highest Used 0.

## Breaking Changes

None. New changeset fields are additive (`serde(skip)` breadcrumbs, `Default`-constructed), and the FFI struct layout is unchanged — only previously-hardcoded fields are now populated.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added tracking for addresses marked as used and the highest-used address index for each account.
  * Preserved address usage status and derivation progress when wallet changes are saved.
* **Bug Fixes**
  * Improved wallet synchronization so address pools reflect authoritative usage information.
  * Prevented address usage and highest-used indexes from being lost or regressing when multiple updates are combined.

### DIFF
--- a/packages/rs-platform-wallet-ffi/src/core_wallet_types.rs
+++ b/packages/rs-platform-wallet-ffi/src/core_wallet_types.rs
@@ -305,6 +305,20 @@ impl WalletChangeSetFFI {
             }
         }
 
+        // Watermark-only accounts still need an `AccountChangeSetFFI`
+        // row. A batch can carry a highest-used advance for an account
+        // with no record of its own this round — e.g. upstream's
+        // `confirm_transaction` returning `None` for an idempotent
+        // re-confirmation while `mark_address_used` still ran, or the
+        // bridge's re-check resolving a sibling account of the same
+        // category. Without an empty bucket the watermark would be
+        // silently dropped below.
+        for account_type in cs.account_highest_used.keys() {
+            if !by_account.iter().any(|(at, _)| at == account_type) {
+                by_account.push((*account_type, Vec::new()));
+            }
+        }
+
         let mut ffi_accounts = Vec::with_capacity(by_account.len());
         for (account_type, recs) in by_account {
             let type_name = CString::new(format!("{:?}", account_type))
@@ -1091,4 +1105,55 @@ pub unsafe fn free_wallet_changeset_ffi(cs: &WalletChangeSetFFI) {
         cs.accounts_count,
         cs.accounts_count,
     ));
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use key_wallet::account::{AccountType, StandardAccountType};
+    use platform_wallet::changeset::{CoreChangeSet, HighestUsedIndexes};
+
+    /// A batch can carry a highest-used advance for an account with no
+    /// record of its own in the same `store()` round (idempotent
+    /// re-confirmation, sibling-account resolution). `from_changeset`
+    /// must still surface it as an (otherwise empty) account bucket —
+    /// regression for the shape where the watermark was silently
+    /// dropped because buckets were built from `cs.records` alone.
+    #[test]
+    fn watermark_only_account_survives_from_changeset() {
+        let account = AccountType::Standard {
+            index: 0,
+            standard_account_type: StandardAccountType::BIP44Account,
+        };
+        let mut cs = CoreChangeSet::default();
+        cs.account_highest_used.insert(
+            account,
+            HighestUsedIndexes {
+                external: Some(7),
+                internal: Some(3),
+            },
+        );
+
+        let ffi = WalletChangeSetFFI::from_changeset(&cs);
+        assert_eq!(ffi.accounts_count, 1, "watermark-only account must emit");
+        let bucket = unsafe { &*ffi.accounts };
+        assert!(bucket.has_external_highest_used);
+        assert_eq!(bucket.external_highest_used, 7);
+        assert!(bucket.has_internal_highest_used);
+        assert_eq!(bucket.internal_highest_used, 3);
+        assert_eq!(bucket.transactions_count, 0);
+        assert_eq!(bucket.utxos_added_count, 0);
+        unsafe { free_wallet_changeset_ffi(&ffi) };
+    }
+
+    /// The `has_*` flags stay false (values -1) for accounts with
+    /// records but no watermark update this batch, so the Swift
+    /// persister never regresses a stored value on a no-usage round.
+    #[test]
+    fn accounts_without_watermarks_emit_unset_flags() {
+        let cs = CoreChangeSet::default();
+        let ffi = WalletChangeSetFFI::from_changeset(&cs);
+        assert_eq!(ffi.accounts_count, 0);
+        unsafe { free_wallet_changeset_ffi(&ffi) };
+    }
 }
--- a/packages/rs-platform-wallet-ffi/src/core_wallet_types.rs
+++ b/packages/rs-platform-wallet-ffi/src/core_wallet_types.rs
@@ -338,6 +338,16 @@ impl WalletChangeSetFFI {
             // sync-path emit for the same account collapse onto a
             // single SwiftData row.
             let tags = account_type_to_tags(&account_type);
+            // Post-batch highest-used watermarks, captured by the
+            // event bridge from the authoritative in-memory pools for
+            // every account this batch marked an address used on
+            // (`CoreChangeSet::account_highest_used`). `has_* = false`
+            // means "no update this batch" — the Swift persister only
+            // overwrites its row when the flag is set, so batches
+            // without usage never regress a stored watermark.
+            let highest = cs.account_highest_used.get(&account_type);
+            let external_highest_used = highest.and_then(|h| h.external);
+            let internal_highest_used = highest.and_then(|h| h.internal);
             ffi_accounts.push(AccountChangeSetFFI {
                 account_type_name: type_name.into_raw(),
                 account_index,
@@ -357,15 +367,10 @@ impl WalletChangeSetFFI {
                 utxos_instant_locked_count: 0,
                 transactions: vec_to_ptr(transactions),
                 transactions_count,
-                // Highest-used pool indices were a feature of the
-                // deleted upstream changeset's per-account bucket.
-                // The new event-bus model doesn't surface them; the
-                // persister can derive them from monitored addresses
-                // if needed.
-                external_highest_used: -1,
-                has_external_highest_used: false,
-                internal_highest_used: -1,
-                has_internal_highest_used: false,
+                external_highest_used: external_highest_used.map_or(-1, |v| v as i32),
+                has_external_highest_used: external_highest_used.is_some(),
+                internal_highest_used: internal_highest_used.map_or(-1, |v| v as i32),
+                has_internal_highest_used: internal_highest_used.is_some(),
             });
         }
 
--- a/packages/rs-platform-wallet/src/changeset/core_bridge.rs
+++ b/packages/rs-platform-wallet/src/changeset/core_bridge.rs
@@ -30,7 +30,7 @@ use std::sync::Arc;
 use dashcore::blockdata::transaction::{txout::TxOut, OutPoint};
 use dashcore::ScriptBuf;
 use key_wallet::account::AccountType;
-use key_wallet::managed_account::address_pool::AddressPoolType;
+use key_wallet::managed_account::address_pool::{AddressPool, AddressPoolType};
 use key_wallet::managed_account::transaction_record::{OutputRole, TransactionRecord};
 use key_wallet::transaction_checking::transaction_router::AccountTypeToCheck;
 use key_wallet::transaction_checking::{DerivedAddressInfo, TransactionContext};
@@ -263,18 +263,9 @@ async fn build_core_changeset(
 /// reaches the persister and every mirrored store keeps
 /// `is_used = false` / `highest_used = None` forever.
 ///
-/// For each record this re-runs the same **read-only** matcher the
-/// wallet used during processing
-/// ([`ManagedAccountCollection::check_transaction`]), scoped to the
-/// record's own account type, then resolves every involved address
-/// back to its owning pool for the authoritative post-mark
-/// [`AddressInfo`](key_wallet::managed_account::address_pool::AddressInfo)
-/// and the pool's `highest_used` watermark. `used` is forced `true`
-/// on the emitted entry — involvement in a recorded transaction is
-/// the definition of "used", independent of snapshot timing.
-///
 /// Returns empty deltas when the wallet is unknown (raced a removal)
-/// — the next sync round re-emits.
+/// — the next sync round re-emits. See
+/// [`collect_usage_deltas_from_accounts`] for the derivation itself.
 async fn collect_usage_deltas(
     wallet_manager: &Arc<RwLock<WalletManager<PlatformWalletInfo>>>,
     wallet_id: &WalletId,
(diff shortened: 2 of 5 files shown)


thepastaclaw:
✅ Review complete (commit 03dfc489)

packages/rs-platform-wallet/src/changeset/core_bridge.rs
```diff
@@ -228,6 +252,130 @@ async fn build_core_changeset(
...
+        };
+        let Ok(type_to_check) = AccountTypeToCheck::try_from(account.managed_account_type()) else {
+            continue;
+        };
+        if checked.contains(&(record.txid, type_to_check)) {
+            continue;
+        }
+        checked.push((record.txid, type_to_check));
+
+        let result = accounts.check_transaction(&record.transaction, &[type_to_check]);
```
thepastaclaw: **🔴 Blocking: Post-event replay can never observe spent-input addresses — UTXOs are already removed by the time this code runs**

`collect_usage_deltas` re-runs `accounts.check_transaction(&record.transaction, &[type_to_check])` from inside the wallet-event subscriber, which only fires *after* `record_transaction`/`confirm_transaction` has already mutated the account (see `wallet_checker.rs:150-179` in the pinned key-wallet rev 234e3c4: `record_transaction`/`confirm_transaction` runs first, `mark_address_used` second, using the `account_match` computed *before* either mutation). Input matching inside `check_transaction_for_match` depends entirely on the live UTXO set: `managed_core_funds_account.rs:671` does `if let Some(utxo) = self.utxos.get(&input.previous_output)`. But `record_transaction`'s `update_utxos` removes spent outpoints from `self.utxos` before returning (`managed_core_funds_account.rs:263-279`), and that removal always precedes the `WalletEvent` this bridge reacts to. So for any transaction that spends a wallet-owned input, this bridge's replay will never find that input's UTXO and will silently omit that address from both `marked_used` and `highest_used` — even though the original in-wallet processing correctly called `mark_address_used` on it via the pre-mutation `account_match`.

This isn't just a theoretical edge case: the PR's own motivation is that real mainnet wallets currently have addresses persisted with `is_used=false` due to the exact bug this PR fixes. When one of those already-broken addresses is later spent as an input, this bridge's replay-based recovery cannot repair it, because by construction it only ever sees post-mutation state. The fix needs to source spent-input addresses from `TransactionRecord.input_details` (which key-wallet already populates *before* `update_utxos` runs, per the comment at `managed_core_funds_account.rs:393`) rather than relying solely on re-running `check_transaction` after the fact.

packages/rs-platform-wallet-ffi/src/core_wallet_types.rs
```diff
@@ -338,6 +338,16 @@ impl WalletChangeSetFFI {
...
+            // Post-batch highest-used watermarks, captured by the
+            // event bridge from the authoritative in-memory pools for
+            // every account this batch marked an address used on
+            // (`CoreChangeSet::account_highest_used`). `has_* = false`
+            // means "no update this batch" — the Swift persister only
+            // overwrites its row when the flag is set, so batches
+            // without usage never regress a stored watermark.
+            let highest = cs.account_highest_used.get(&account_type);
+            let external_highest_used = highest.and_then(|h| h.external);
+            let internal_highest_used = highest.and_then(|h| h.internal);
```
thepastaclaw, line 364: **🔴 Blocking: `from_changeset` drops `account_highest_used` watermarks for any account not also present in `cs.records`**

`ffi_accounts` (and therefore every `has_external_highest_used`/`has_internal_highest_used` flag) is built exclusively by iterating `by_account`, which is populated only from `cs.records` (lines 297-306). `cs.account_highest_used` is looked up per-bucket at line 348 — but only for buckets that already exist, i.e. only for account types that have a record in `cs.records` for this specific `store()` call. `collect_usage_deltas` (core_bridge.rs:308-374) does not guarantee that alignment: it resolves highest-used watermarks for *any* account whose pool owns a matched address, using a `type_to_check` scoped re-check that (for indexed categories like `StandardBIP44`) sweeps every account index in that category, not just the index of the record(s) passed in. Combined with key-wallet's `confirm_transaction` returning `None` (no record emitted) for already-finalized/idempotent re-confirmations while still unconditionally calling `mark_address_used` (`wallet_checker.rs:158-179`), a `BlockProcessed` batch can produce an `account_highest_used` entry for an account whose own record isn't in `inserted`/`updated`/`matured` this round — and that entry is then silently dropped by `from_changeset`.

This is not just a hypothetical: `changeset.rs`'s `marked_used_and_highest_used_mark_core_changeset_non_empty` test constructs exactly this shape — a `CoreChangeSet` with `account_highest_used` populated and `records` empty — and asserts the changeset must be treated as non-empty/persistable. But no test exercises `from_changeset` with that same shape, and reading the conversion confirms it produces `accounts_count == 0`, so the watermark never reaches Swift. Note this specifically affects the `external_highest_used`/`internal_highest_used` fields (gap-limit bookkeeping); the `is_used` flag itself is safe since `addresses_marked_used` is fanned out through a separate, unfiltered callback path (`persistence.rs:848-869`).

```suggestion
let mut by_account: Vec<(
    AccountType,
    Vec<&key_wallet::managed_account::transaction_record::TransactionRecord>,
)> = Vec::new();
for rec in &cs.records {
    if let Some(bucket) = by_account
        .iter_mut()
        .find(|(at, _)| at == &rec.account_type)
    {
        bucket.1.push(rec);
    } else {
        by_account.push((rec.account_type, vec![rec]));
    }
}
// Watermark-only accounts (e.g. a re-check that resolved a highest-used
// advance for an account whose own record wasn't re-emitted this round)
// still need an `AccountChangeSetFFI` row, or `account_highest_used` is
// silently dropped below.
for account_type in cs.account_highest_used.keys() {
    if !by_account.iter().any(|(at, _)| at == account_type) {
        by_account.push((*account_type, Vec::new()));
    }
}
```

<sub>source: ['codex-general', 'sonnet5-general']</sub>

packages/rs-platform-wallet/src/changeset/core_bridge.rs
```diff
@@ -228,6 +252,130 @@ async fn build_core_changeset(
...
+                        }
+                        slot.merge_max(snapshot);
+                    }
+                }
+            }
+        }
+    }
+
+    (marked_used, highest_used)
+}
```
thepastaclaw, line 437: **🟡 Suggestion: `collect_usage_deltas` has no direct test coverage**

This is the function that actually fixes the reported bug — re-running `ManagedAccountCollection::check_transaction`, resolving an involved address back to its owning `(account_type, pool_type)` by scanning every account's pools, and snapshotting `highest_used`. There is no `#[cfg(test)]` module in `core_bridge.rs` at all, and no other file references `collect_usage_deltas`. Every existing test in this PR exercises pure data-structure logic downstream of it (`CoreChangeSet::merge` in `changeset.rs`, `group_marked_used_into_pool_entries` in `persistence.rs`), none of which would catch a regression in the matching/resolution logic itself. A regression test using a `ManagedAccountCollection` fixture covering a ProReg-payload owner-key match and a standard BIP44 output match would directly protect the scenario described in the PR (mainnet txid `5076f9ed...`) rather than only its downstream plumbing.

<sub>source: ['sonnet5-general']</sub>

packages/rs-platform-wallet/src/changeset/core_bridge.rs
```diff
@@ -228,6 +252,130 @@ async fn build_core_changeset(
...
+                                    (Some(a), Some(b)) => Some(a.max(b)),
+                                    (a, b) => a.or(b),
+                                };
+                            }
+                        }
+                        slot.merge_max(snapshot);
+                    }
+                }
+            }
+        }
```
thepastaclaw: **💬 Nitpick: `address_pools()` allocates a `Vec` twice per matched address per account**

`owner.managed_account_type().address_pools()` allocates a fresh `Vec<&AddressPool>` on every call (confirmed: `managed_account_type.rs:215` constructs a new `vec![...]` each invocation) — it's called once at line 339 and again at line 359 for every `(involved address, owner account)` pair. For wallets with more than a handful of accounts this is O(accounts × pools) redundant allocations per involved address per record. Hoisting the call into a single local before the two inner loops would avoid the duplicate allocation. Unlikely to matter at typical wallet account counts, hence a nitpick rather than a blocker.

<sub>source: ['sonnet5-general']</sub>

thepastaclaw (changes requested):
## Code Review

This PR fixes a real, previously-reported bug where SPV-discovered address usage never reached SwiftData because WalletEvent carries no mark-used/highest-used deltas. The re-derivation approach (replaying ManagedAccountCollection::check_transaction after the fact) is fundamentally sound for output-side and payload-based matches, and I verified against the pinned key-wallet rev (234e3c4) that the addresses_marked_used path (which drives the actual is_used flag) is wired correctly and independently of the buggy bucketing described below. However, two verified gaps remain in scope: replaying check_transaction after record_transaction has already removed spent UTXOs from the live account state means input-side usage (an address later spent as a transaction input) is never captured, and WalletChangeSetFFI::from_changeset buckets accounts strictly by cs.records, silently dropping account_highest_used watermarks for any account whose own record isn't in the current batch — a shape the PR's own test suite already constructs and asserts is meant to be persisted.

Source: reviewers gpt-5.6-sol (Sol reviewer) and claude-sonnet-5 (Sonnet reviewer), general lane; verifier claude-sonnet-5. Orchestrator openai/gpt-5.6-sol (orchestration-only).

🔴 2 blocking | 🟡 1 suggestion(s) | 💬 1 nitpick(s)

<details>

```

- [BLOCKING] packages/rs-platform-wallet/src/changeset/core_bridge.rs:308-328: Post-event replay can never observe spent-input addresses — UTXOs are already removed by the time this code runs
  `collect_usage_deltas` re-runs `accounts.check_transaction(&record.transaction, &[type_to_check])` from inside the wallet-event subscriber, which only fires *after* `record_transaction`/`confirm_transaction` has already mutated the account (see `wallet_checker.rs:150-179` in the pinned key-wallet rev 234e3c4: `record_transaction`/`confirm_transaction` runs first, `mark_address_used` second, using the `account_match` computed *before* either mutation). Input matching inside `check_transaction_for_match` depends entirely on the live UTXO set: `managed_core_funds_account.rs:671` does `if let Some(utxo) = self.utxos.get(&input.previous_output)`. But `record_transaction`'s `update_utxos` removes spent outpoints from `self.utxos` before returning (`managed_core_funds_account.rs:263-279`), and that removal always precedes the `WalletEvent` this bridge reacts to. So for any transaction that spends a wallet-owned input, this bridge's replay will never find that input's UTXO and will silently omit that address from both `marked_used` and `highest_used` — even though the original in-wallet processing correctly called `mark_address_used` on it via the pre-mutation `account_match`.

  This isn't just a theoretical edge case: the PR's own motivation is that real mainnet wallets currently have addresses persisted with `is_used=false` due to the exact bug this PR fixes. When one of those already-broken addresses is later spent as an input, this bridge's replay-based recovery cannot repair it, because by construction it only ever sees post-mutation state. The fix needs to source spent-input addresses from `TransactionRecord.input_details` (which key-wallet already populates *before* `update_utxos` runs, per the comment at `managed_core_funds_account.rs:393`) rather than relying solely on re-running `check_transaction` after the fact.
- [SUGGESTION] packages/rs-platform-wallet/src/changeset/core_bridge.rs:278-377: `collect_usage_deltas` has no direct test coverage
  This is the function that actually fixes the reported bug — re-running `ManagedAccountCollection::check_transaction`, resolving an involved address back to its owning `(account_type, pool_type)` by scanning every account's pools, and snapshotting `highest_used`. There is no `#[cfg(test)]` module in `core_bridge.rs` at all, and no other file references `collect_usage_deltas`. Every existing test in this PR exercises pure data-structure logic downstream of it (`CoreChangeSet::merge` in `changeset.rs`, `group_marked_used_into_pool_entries` in `persistence.rs`), none of which would catch a regression in the matching/resolution logic itself. A regression test using a `ManagedAccountCollection` fixture covering a ProReg-payload owner-key match and a standard BIP44 output match would directly protect the scenario described in the PR (mainnet txid `5076f9ed...`) rather than only its downstream plumbing.

- [BLOCKING] packages/rs-platform-wallet-ffi/src/core_wallet_types.rs:293-350: `from_changeset` drops `account_highest_used` watermarks for any account not also present in `cs.records`
  `ffi_accounts` (and therefore every `has_external_highest_used`/`has_internal_highest_used` flag) is built exclusively by iterating `by_account`, which is populated only from `cs.records` (lines 297-306). `cs.account_highest_used` is looked up per-bucket at line 348 — but only for buckets that already exist, i.e. only for account types that have a record in `cs.records` for this specific `store()` call. `collect_usage_deltas` (core_bridge.rs:308-374) does not guarantee that alignment: it resolves highest-used watermarks for *any* account whose pool owns a matched address, using a `type_to_check` scoped re-check that (for indexed categories like `StandardBIP44`) sweeps every account index in that category, not just the index of the record(s) passed in. Combined with key-wallet's `confirm_transaction` returning `None` (no record emitted) for already-finalized/idempotent re-confirmations while still unconditionally calling `mark_address_used` (`wallet_checker.rs:158-179`), a `BlockProcessed` batch can produce an `account_highest_used` entry for an account whose own record isn't in `inserted`/`updated`/`matured` this round — and that entry is then silently dropped by `from_changeset`.

  This is not just a hypothetical: `changeset.rs`'s `marked_used_and_highest_used_mark_core_changeset_non_empty` test constructs exactly this shape — a `CoreChangeSet` with `account_highest_used` populated and `records` empty — and asserts the changeset must be treated as non-empty/persistable. But no test exercises `from_changeset` with that same shape, and reading the conversion confirms it produces `accounts_count == 0`, so the watermark never reaches Swift. Note this specifically affects the `external_highest_used`/`internal_highest_used` fields (gap-limit bookkeeping); the `is_used` flag itself is safe since `addresses_marked_used` is fanned out through a separate, unfiltered callback path (`persistence.rs:848-869`).
```
</details>

packages/rs-platform-wallet/src/changeset/core_bridge.rs
QuantumExplorer: Fixed in 03dfc4896a: the record's own `input_details` (and Received/Change `output_details`) are now a candidate source alongside the replayed match, so spent-input addresses survive the pre-event UTXO removal; covered by the new `spent_input_address_is_captured_after_utxo_removal` test which spends a funded UTXO and derives deltas from the spend record only.

packages/rs-platform-wallet-ffi/src/core_wallet_types.rs
QuantumExplorer: Fixed in 03dfc4896a: `from_changeset` now emits an (otherwise empty) `AccountChangeSetFFI` bucket for every account in `cs.account_highest_used` that has no record this round, per the suggested shape; regression test `watermark_only_account_survives_from_changeset` covers the records-empty case.

packages/rs-platform-wallet/src/changeset/core_bridge.rs
QuantumExplorer: Added in 03dfc4896a: the derivation is now a sync `collect_usage_deltas_from_accounts` with a `usage_delta_tests` module driving a real `ManagedWalletInfo` through `check_core_transaction` — a BIP44 receive asserts the External #0 used flip + watermark advance, and a spend asserts input-side capture after UTXO removal.

packages/rs-platform-wallet/src/changeset/core_bridge.rs
QuantumExplorer: Fixed in 03dfc4896a: `address_pools()` is now called once per account per batch (hoisted `pools_by_account`), and highest-used watermarks are snapshotted once per touched account after the address walk (also resolves the CodeRabbit note).

thepastaclaw (approved):
## Code Review

This commit fixes both prior blocking data-loss paths. Spent-input addresses are now sourced from TransactionRecord.input_details, which key-wallet populates before update_utxos removes the spent outpoints, and watermark-only accounts now emit an AccountChangeSetFFI row. Direct regression tests cover both fixes, address-pool snapshots eliminate the prior repeated allocation, and the full platform-wallet and platform-wallet-ffi test suites, clippy, and formatting checks are clean. No new issue was found in the latest delta or cumulative diff.

Source: Sol reviewer `gpt-5.6-sol` (general), Sonnet reviewer `claude-sonnet-5` (general), sampled Opus reviewer `claude-opus-4-8` (general); final verifier `claude-sonnet-5`. The Sol orchestrator `openai/gpt-5.6-sol` was orchestration-only and is not reviewer or verifier evidence.

Prior reconciliation: prior-1 FIXED; prior-2 FIXED; prior-3 FIXED; prior-4 FIXED.

Carried-forward prior findings: None.

New findings in the latest delta: None.

PR #4115, opened 2026-07-13
fix(dpp): bound document value validation depth

QuantumExplorer:
## Summary

- enforce a protocol-v13 document value depth limit of 256 during state-transition decoding, before field-size scanning, JSON conversion, or schema validation
- decode `Value` iteratively with a version-scoped container-depth ceiling so nesting depth can never consume attacker-controlled stack
- rewrite `Value::has_data_larger_than` as an iterative traversal that borrows (never clones) the reported field
- align the decode ceiling and the validation-side depth checks so every depth violation follows the same consensus path
- add depth-boundary, traversal, version-boundary, regression, and valid-document coverage

## Root cause

Document create and replace validation passed decoded, attacker-controlled properties to `has_data_larger_than` before JSON/schema validation. `Value` derived a recursive `Decode`, so deeply nested payloads could overflow the stack during `StateTransition::deserialize_from_bytes` before any validation ran, and the oversized-array error path repeatedly cloned nested subtrees while unwinding.

## Impact

On protocol v13, `Value` decoding rejects more than 256 nested containers per value, so an over-depth document value is an undecodable transition — the same deterministic, unpaid rejection as any other malformed encoding, applied identically at every depth. Because each document property value receives the same depth budget in the decoder and in validation (the plain `BTreeMap` data wrapper is not counted, matching the existing contract-schema rule whose root map is a real `Value` container), no decodable payload can violate the depth rule: equivalent violations can no longer split between paid and unpaid fee/nonce semantics. The validation-side depth checks in `validate_document_properties` and the batch transformer remain as defense in depth for values that do not arrive via the wire decoder. Protocol versions before v13 retain their prior consensus rule, while the shared field-size helper and the `Value` decoder are iterative for all callers.

## Validation

- pre-fix regression reproduced the vulnerable ordering: depth 257 returned `DocumentFieldMaxSizeExceededError`
- boundary test proves a value at the decoder ceiling round-trips through versioned deserialization and passes the validation-side depth check
- `cargo test -p platform-value --all-features` — all passed
- `cargo test -p platform-version -q` — passed
- `cargo test -p dpp --all-features` — 3,870 passed, 6 ignored
- `cargo check -p drive-abci` — passed
- `cargo clippy -p platform-value --all-targets -- -D warnings` — passed
- `cargo clippy -p dpp --all-features --lib -- -D warnings` — passed
- `cargo clippy -p dpp --all-features --tests -- -D warnings -A clippy::needless_borrows_for_generic_args` — passed
- `cargo fmt --all -- --check` and `git diff --check` — passed

The unmodified strict DPP test-target Clippy command is currently blocked by an unrelated pre-existing `needless_borrows_for_generic_args` warning in `data_contract/factory/v0/mod.rs`.

### DIFF
--- a/packages/rs-dpp/src/data_contract/methods/validate_document/v0/mod.rs
+++ b/packages/rs-dpp/src/data_contract/methods/validate_document/v0/mod.rs
@@ -5,6 +5,7 @@ use crate::data_contract::document_type::DocumentType;
 use crate::consensus::basic::document::{
     DocumentFieldMaxSizeExceededError, InvalidDocumentTypeError,
 };
+use crate::consensus::basic::value_error::ValueError;
 use crate::consensus::basic::BasicError;
 use crate::consensus::ConsensusError;
 use crate::data_contract::schema::DataContractSchemaMethodsV0;
@@ -45,6 +46,18 @@ impl DataContract {
                 InvalidDocumentTypeError::new(name.to_owned(), self.id()).into(),
             ));
         };
+
+        if let Some(max_depth) = platform_version.system_limits.max_document_value_depth {
+            if let Some(actual_depth) = value.first_depth_exceeding(max_depth as usize) {
+                return Ok(SimpleConsensusValidationResult::new_with_error(
+                    ConsensusError::BasicError(BasicError::ValueError(
+                        ValueError::new_from_string(format!(
+                            "document value depth {actual_depth} exceeds system maximum {max_depth}"
+                        )),
+                    )),
+                ));
+            }
+        }
 
         let validator = document_type.json_schema_validator_ref().deref();
 
@@ -106,5 +119,113 @@ impl DataContract {
     ) -> Result<SimpleConsensusValidationResult, ProtocolError> {
         // Validate user defined properties
         self.validate_document_properties_v0(name, document.properties().into(), platform_version)
+    }
+}
+
+#[cfg(all(test, feature = "fixtures-and-mocks"))]
+mod tests {
+    use super::DataContractDocumentValidationMethodsV0;
+    use crate::consensus::basic::value_error::ValueError;
+    use crate::consensus::basic::BasicError;
+    use crate::consensus::ConsensusError;
+    use crate::data_contract::created_data_contract::CreatedDataContract;
+    use crate::tests::fixtures::get_data_contract_fixture;
+    use platform_value::Value;
+    use platform_version::version::PlatformVersion;
+
+    fn data_contract() -> CreatedDataContract {
+        let platform_version = PlatformVersion::latest();
+        get_data_contract_fixture(None, 0, platform_version.protocol_version)
+    }
+
+    fn nested_document_value(container_count: usize, leaf: Value) -> Value {
+        let nested = (0..container_count).fold(leaf, |value, depth| {
+            if depth % 2 == 0 {
+                Value::Array(vec![value])
+            } else {
+                Value::Map(vec![(Value::Text("nested".to_owned()), value)])
+            }
+        });
+
+        Value::Map(vec![(Value::Text("name".to_owned()), nested)])
+    }
+
+    #[test]
+    fn should_reject_excessive_document_value_depth_before_field_size_validation() {
+        let platform_version = PlatformVersion::latest();
+        let data_contract = data_contract().data_contract_owned();
+        let max_depth = platform_version
+            .system_limits
+            .max_document_value_depth
+            .expect("latest protocol should enforce document value depth");
+        let value = nested_document_value(
+            max_depth as usize,
+            Value::Text(
+                "x".repeat(platform_version.system_limits.max_field_value_size as usize + 1),
+            ),
+        );
+
+        let result = data_contract
+            .validate_document_properties("noTimeDocument", value, platform_version)
+            .expect("validation should return a consensus result");
+
+        let Some(ConsensusError::BasicError(BasicError::ValueError(ValueError { .. }))) =
+            result.first_error()
+        else {
+            panic!("expected document value depth error, got {result:?}");
+        };
+        assert_eq!(
+            result.first_error().expect("expected an error").to_string(),
+            format!(
+                "document value depth {} exceeds system maximum {max_depth}",
+                max_depth + 1
+            )
+        );
+    }
+
+    #[test]
+    fn should_allow_document_value_depth_at_the_limit() {
+        let platform_version = PlatformVersion::latest();
+        let data_contract = data_contract().data_contract_owned();
+        let max_depth = platform_version
+            .system_limits
+            .max_document_value_depth
+            .expect("latest protocol should enforce document value depth");
+        let value = nested_document_value(
+            max_depth as usize - 1,
+            Value::Text(
+                "x".repeat(platform_version.system_limits.max_field_value_size as usize + 1),
+            ),
+        );
+
+        let result = data_contract
+            .validate_document_properties("noTimeDocument", value, platform_version)
+            .expect("validation should return a consensus result");
+
+        assert!(matches!(
+            result.first_error(),
+            Some(ConsensusError::BasicError(
+                BasicError::DocumentFieldMaxSizeExceededError(_)
+            ))
+        ));
+    }
+
+    #[test]
+    fn should_preserve_valid_document_properties() {
+        let platform_version = PlatformVersion::latest();
+        let data_contract = data_contract().data_contract_owned();
+        let value = Value::Map(vec![(
+            Value::Text("name".to_owned()),
+            Value::Text("Alice".to_owned()),
+        )]);
+
+        let result = data_contract
+            .validate_document_properties("noTimeDocument", value, platform_version)
+            .expect("validation should return a consensus result");
+
+        assert!(
+            result.is_valid(),
+            "expected valid properties, got {result:?}"
+        );
     }
 }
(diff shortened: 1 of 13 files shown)


thepastaclaw:
⛔ Blockers found — Sonnet deferred (commit aa3edcf2)

packages/rs-dpp/src/data_contract/methods/validate_document/v0/mod.rs
```diff
@@ -46,6 +47,18 @@ impl DataContract {
...
+        if let Some(max_depth) = platform_version.system_limits.max_document_value_depth {
+            if let Some(actual_depth) = value.first_depth_exceeding(max_depth as usize) {
+                return Ok(SimpleConsensusValidationResult::new_with_error(
+                    ConsensusError::BasicError(BasicError::ValueError(
+                        ValueError::new_from_string(format!(
+                            "document value depth {actual_depth} exceeds system maximum {max_depth}"
+                        )),
+                    )),
+                ));
+            }
```
thepastaclaw, line 70: **🔴 Blocking: Enforce the depth limit during decoding**

This validation runs after attacker-controlled document values have already undergone recursive processing. An actual document-batch state transition containing 4,000 single-element arrays encoded to 8,149 bytes, below the 20,480-byte transition limit, and aborted with a stack overflow inside `StateTransition::deserialize_from_bytes`. `Value` derives recursive `Decode`, so the transition never reaches this deterministic `ValueError`. Values that survive decoding are also recursively cloned while create/replace actions are built and again when `self.data().into()` prepares validation. Enforce the depth budget in a custom or iterative decoder, then check borrowed transition data before action transformation performs recursive clones.

<sub>source: ['codex']</sub>

packages/rs-platform-value/src/lib.rs
```diff
@@ -1318,178 +1318,91 @@ impl Value {
...
+                    for (key, value) in map.iter().rev() {
+                        pending.push((value, reported_value.or(Some(key))));
                     }
+                    continue;
                 }
-                None
+            };
+
+            if let Some(actual_size) = actual_size {
+                return Some((reported_value.cloned(), actual_size));
```
thepastaclaw: **🟡 Suggestion: Avoid cloning a recursive root-array result**

For a root array, `reported_value` is its first child. If an oversized scalar occurs beneath that child, `reported_value.cloned()` recursively clones the remaining subtree and defeats the helper's stack-independent traversal. An actual 8,000-level `Value::Array` ending in oversized text aborted here, while the new deep test uses `Null` and never enters this result path. The current DPP production caller wraps document properties in a root map, preserving a shallow field key, and v12 depth-checks first, so this is not a second blocking document-validation exploit. It still violates the changed public helper's stated behavior; represent the reported field without cloning a recursive `Value` and add a deeply nested oversized root-array test.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

At exact HEAD 22a91fc003e52101651b8569d47e5e7587d855e4, the new depth check remains unreachable for some under-limit attacker payloads because recursive state-transition decoding can abort first; this is a blocking denial of service. The field-size helper also retains a direct root-array clone failure, but the current DPP caller reports a shallow map key and applies the v12 depth guard first, so that issue is nonblocking. Targeted tests passed, and isolated probes against the actual crates reproduced both aborts.

### Review provenance

<details>

```

In `packages/rs-dpp/src/data_contract/methods/validate_document/v0/mod.rs`:
- [BLOCKING] packages/rs-dpp/src/data_contract/methods/validate_document/v0/mod.rs:50-59: Enforce the depth limit during decoding
  This validation runs after attacker-controlled document values have already undergone recursive processing. An actual document-batch state transition containing 4,000 single-element arrays encoded to 8,149 bytes, below the 20,480-byte transition limit, and aborted with a stack overflow inside `StateTransition::deserialize_from_bytes`. `Value` derives recursive `Decode`, so the transition never reaches this deterministic `ValueError`. Values that survive decoding are also recursively cloned while create/replace actions are built and again when `self.data().into()` prepares validation. Enforce the depth budget in a custom or iterative decoder, then check borrowed transition data before action transformation performs recursive clones.

In `packages/rs-platform-value/src/lib.rs`:
- [SUGGESTION] packages/rs-platform-value/src/lib.rs:1401: Avoid cloning a recursive root-array result
  For a root array, `reported_value` is its first child. If an oversized scalar occurs beneath that child, `reported_value.cloned()` recursively clones the remaining subtree and defeats the helper's stack-independent traversal. An actual 8,000-level `Value::Array` ending in oversized text aborted here, while the new deep test uses `Null` and never enters this result path. The current DPP production caller wraps document properties in a root map, preserving a shallow field key, and v12 depth-checks first, so this is not a second blocking document-validation exploit. It still violates the changed public helper's stated behavior; represent the reported field without cloning a recursive `Value` and add a deeply nested oversized root-array test.
```
</details>

packages/rs-dpp/src/state_transition/mod.rs
```diff
@@ -797,7 +797,14 @@ impl StateTransition {
...
     ) -> Result<Self, ProtocolError> {
-        let state_transition = StateTransition::deserialize_from_bytes(bytes)?;
+        let max_value_depth = platform_version
+            .system_limits
+            .max_document_value_depth
+            .map(usize::from);
+        let state_transition =
+            platform_value::with_value_decode_depth_limit(max_value_depth, || {
+                StateTransition::deserialize_from_bytes(bytes)
+            })?;
```
thepastaclaw, line 807: **🔴 Blocking: Keep all over-depth documents on the same consensus path**

The decoder applies `max_document_value_depth` to each `Value`, while document validation includes the enclosing `BTreeMap<String, Value>` as depth one. A property containing 256 nested arrays therefore has document depth 257 but Value depth 256: it decodes, reaches `first_data_depth_exceeding`, returns `BasicError::ValueError`, and on protocol v12 produces a `BumpIdentityDataContractNonce` action and paid error. With 257 nested arrays, the decoder instead raises `DecodeError::OtherString`; deserialization maps that to `PlatformDeserializationError`, raw-transition decoding maps it to `SerializedObjectParsingError`, and processing returns an unpaid error without the nonce action. `prepare_proposal` consequently retains the first transition but removes the second. This contradicts the PR's stated deterministic `ValueError` behavior and lets equivalent violations select different fee and nonce semantics. Preserve the iterative safety boundary, but coordinate it with document-depth validation through a typed depth failure or a separate defensive ceiling so every protocol depth violation follows the intended consensus path.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

At exact head aa3edcf25f1dbc2fe4d46059beb95f9f10bb7af2, both prior findings are fixed: decoding is iterative and version-scoped, and oversized-field reporting no longer clones recursive values. One blocking consensus-path inconsistency remains: over-depth documents are charged and nonce-bumped only when their nested Value depth remains within the decoder ceiling; deeper instances become unpaid invalid encodings.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — security-auditor (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — security-auditor (completed), `gpt-5.6-sol` — rust-quality (completed)

🔴 1 blocking

<details>

```

In `packages/rs-dpp/src/state_transition/mod.rs`:
- [BLOCKING] packages/rs-dpp/src/state_transition/mod.rs:800-807: Keep all over-depth documents on the same consensus path
  The decoder applies `max_document_value_depth` to each `Value`, while document validation includes the enclosing `BTreeMap<String, Value>` as depth one. A property containing 256 nested arrays therefore has document depth 257 but Value depth 256: it decodes, reaches `first_data_depth_exceeding`, returns `BasicError::ValueError`, and on protocol v12 produces a `BumpIdentityDataContractNonce` action and paid error. With 257 nested arrays, the decoder instead raises `DecodeError::OtherString`; deserialization maps that to `PlatformDeserializationError`, raw-transition decoding maps it to `SerializedObjectParsingError`, and processing returns an unpaid error without the nonce action. `prepare_proposal` consequently retains the first transition but removes the second. This contradicts the PR's stated deterministic `ValueError` behavior and lets equivalent violations select different fee and nonce semantics. Preserve the iterative safety boundary, but coordinate it with document-depth validation through a typed depth failure or a separate defensive ceiling so every protocol depth violation follows the intended consensus path.
```
</details>

QuantumExplorer:
Addressed the remaining blocker in be4294984d. The split came from an off-by-one between the two depth counts: the decoder bounds each decoded `Value` at 256 containers, while the document checks counted the enclosing `BTreeMap` data wrapper — which is not a decoded `Value` — as depth one. That left exactly one band (value depth 256) that decoded yet failed validation on the paid path, while anything deeper failed decoding unpaid.

Both DPP depth checks now give each property value the full 256-container budget, identical to the decoder's per-value ceiling (and consistent with the contract-schema rule, whose root map *is* a real `Value` container and already aligned). No decodable payload can violate the depth rule, so every depth violation follows one consensus path: an undecodable transition, rejected the same unpaid way as any other malformed encoding at every depth. The `validate_document_properties` and transformer checks stay as defense in depth for non-wire inputs. A new boundary test proves a value at the ceiling round-trips through versioned deserialization and passes the validation-side check.

QuantumExplorer:
Rebased onto current v4.1-dev and moved the activation gate: protocol v12 is already active on live networks, so the depth limit now lives in a new `SYSTEM_LIMITS_V3` wired into protocol v13, with v12 restored to `None`. Applying it at v12 would have retroactively changed an active version's consensus behavior (historical replay and mixed-version validation both diverge).

QuantumExplorer (commented):
Reviewed

PR #4112, opened 2026-07-13
feat(platform-wallet): surface ProRegTx masternode details in iOS via rust-dashcore bump

QuantumExplorer:
## Issue being fixed or feature implemented

dashpay/rust-dashcore#876 landed typed DIP-2/DIP-3 special-transaction payloads (ProRegTx service IP, collateral, key hashes) plus provider-record retention, but nothing in the platform stack consumed them: a masternode registration in the wallet showed up as just "Provider Registration" with no details. This PR bumps the pin to pick that work up and surfaces the payload end-to-end in the iOS app.

## What was done?

**Dependency bump** (`Cargo.toml`/`Cargo.lock`): rust-dashcore `234e3c43` → `337f6ccc`, picking up #876 (typed special-tx payloads + provider retention), #867 (AssetLock routing covers CoinJoin/DashPay accounts), #868 (CoinJoin gap limit 30 → 100).

**Rust — `rs-platform-wallet-ffi/src/core_wallet_types.rs`** (all DIP-3 decoding stays in Rust):
- `TransactionRecordFFI` gains optional provider fields: `provider_service_address` (C string "ip:port"), `provider_pro_tx_hash`, collateral txid+vout, owner/voting key hashes — inline arrays with `has_*` flags following the existing `block_hash`/`has_fee` pattern, freed in `free_wallet_changeset_ffi`.
- New `provider_payload_fields()` seam decodes `dashcore::TransactionPayload::{ProviderRegistration, ProviderUpdateService}` where `tx_record_to_ffi` builds the record (ProRegTx: service/collateral/key hashes; ProUpServTx: service + proTxHash).

**Swift SDK** (pure persist/bridge per architecture rules): `PersistentTransaction` gains matching optional columns + hex display helpers; `PlatformWalletPersistenceHandler.upsertTransaction` marshals the FFI fields gated on the `has_*` flags.

**SwiftExampleApp**: `TransactionDetailView` renders a "Masternode" section (service address, proTxHash, collateral, key hashes with tap-to-copy rows) for provider registration / update-service transactions.

Follow-ups deliberately out of scope: payout address (needs network-dependent script→address encoding), operator BLS key, and proTxHash follow in the compact-filter watch set (rust-dashcore#861 remainder — without it, ProUpServTx updates for owned masternodes don't match during sync).

## How Has This Been Tested?

[...]

PR #4116, opened 2026-07-13
feat(platform-wallet): wallet masternode list with DML-derived status in iOS

QuantumExplorer:

A wallet holding provider keys had no way to see its masternodes: the Provider Owner Keys account screen showed a jargon "Absent Addresses" header, address rows said only "#0 · used" with no hint of what used them, and there was no masternode list anywhere. Builds on #4112's payload surfacing and rust-dashcore #876's provider-record retention.

## What was done?

**Rust — masternode aggregation + DML status (`rs-platform-wallet`, `rs-platform-wallet-ffi`)**
- `aggregate_masternodes` groups the wallet's retained provider special txs (ProRegTx/ProUpServTx/ProUpRegTx/ProUpRevTx) by proTxHash: latest-update-wins service address, evonode flag from ProRegTx `masternode_type`, owner/voting hashes + collateral, revocation data, stable registration-order sort, and per-type numbering (`type_index`) so evonodes and masternodes number independently.
- Status is validated against the SPV deterministic masternode list via a new `SpvRuntime::masternode_validity_snapshot_blocking`: **Active** (in list, valid), **Inactive** (in list, PoSe-banned), **Retired** (absent — revoked/spent/expired), **Unknown** (DML not synced yet; the persist layer keeps the prior value instead of clobbering).
- New FFI quartet `platform_wallet_manager_list_masternodes`/`free` (mirrors `get_account_balances`), with Rust-side base58 owner/voting address encoding so Swift never hashes keys.
- **Restart durability**: wallet restore now stages provider special txs (`ProviderSpecialTxRestoreEntryFFI`, mirroring the asset-lock staging) onto every provider-key account, where #876's retention keeps them resident — the masternode list survives app restarts without a rescan.

**Swift SDK** (pure persist/bridge): new `PersistentMasternode` model (`(walletId, proTxHash)` unique) + `PlatformWalletManager.masternodes(for:)` wrapper.

**SwiftExampleApp**
- Identities tab: segmented **Identities / Masternodes** toggle, shown only when the network has masternodes. The list shows "Evonode N" / "Masternode N" rows with type badge, service IP, short proTxHash, and colored status — plus a **"Show retired" toggle (default off)** with a hidden-count hint.
- Provider-key account address rows now read "used once on Evonode 2 · <ip>" (or the IP alone when unmatched), joined by Rust-encoded base58 addresses.
- `MasternodeSync` upserts/deletes rows from the aggregation; an empty aggregation never prunes (indistinguishable from "not rehydrated yet").
- The key-pool "Absent" / "Absent (Hardened)" labels are renamed **"Additional"** across all three views (they're the on-demand pools where provider keys live, not missing addresses).

**Masternode detail page** (second commit block): tapping a row opens a detail view with full key material (owner/voting addresses+hashes, operator BLS key, evonode platform node id, payout address), key ownership resolved via the persisted-address join ("ProviderOwnerKeys #14" / "not in this wallet"), Core payout history (TXOs paying the payout address), and for evonodes the claimable balance — the credit balance of the Platform identity whose id is the display-order proTxHash — with a Refresh affordance. Also fixes a pre-existing `SDK.swift` `getBalance` bug (the FFI returns a decimal string; the wrapper misread it as binary u64).

**Claim flow (owner key, signer gated)**: a Claim button (enabled when claimable > 0 and the owner key is in this wallet) opens a confirmation sheet — amount with ~0.005 DASH fee headroom and live credits echo, the registered payout address as the immutable destination, and the signing key ("ProviderOwnerKeys #14"). The `platform_wallet_manager_masternode_withdraw` FFI enforces the owner-key contract and carries a unit-tested guard selecting the identity's matching OWNER `ECDSA_HASH160` key. The final sign+broadcast is deliberately **gated behind a distinct error** ("pending verified signer") until the owner-key signer's DPP signature encoding is verified against a testnet masternode — the whole button→sheet→FFI→guard chain is exercised without risking a malformed money transition.

## How Has This Been Tested?

- `cargo test -p platform-wallet-ffi` (152 tests) green, including new suites: aggregation (registration / update-only / grouping / revocation via a synthetic ProUpRevTx), DML status matrix (all four `ListMembership` outcomes), and per-type numbering (evonode + regular each getting index 1). `cargo clippy` clean.
- `build_ios.sh --target sim` (warnings-as-errors) green.
- Live on-simulator with a mainnet wallet holding 21 provider transactions: masternode rows materialize on a fresh launch **without a rescan** (restore staging verified), the toggle appears, the list shows Evonode 1-6 Active (matching the real DML) with the 15 retired masternodes hidden by default, and a terminate/relaunch cycle keeps all 21 rows with statuses intact.

## Breaking Changes

None.

## Checklist:

### DIFF
--- a/packages/rs-platform-wallet-ffi/src/core_wallet_types.rs
+++ b/packages/rs-platform-wallet-ffi/src/core_wallet_types.rs
@@ -1052,6 +1052,19 @@ pub(crate) struct MasternodeAggregate {
     pub voting_key_hash: Option<[u8; 20]>,
     /// Height that set `voting_key_hash` (drives latest-wins).
     voting_height: u32,
+    /// Operator BLS public key (48 bytes) — follows the latest ProRegTx /
+    /// ProUpReg.
+    pub operator_public_key: Option<[u8; 48]>,
+    operator_height: u32,
+    /// Platform node id (hash160, 20 bytes) for evonodes — follows the
+    /// latest ProRegTx / ProUpServ.
+    pub platform_node_id: Option<[u8; 20]>,
+    platform_node_height: u32,
+    /// Payout script (raw bytes) — follows the latest ProRegTx / ProUpReg
+    /// (owner payout). Encoded to a base58 address by `masternode_entry_ffi`
+    /// where the network is available.
+    pub payout_script: Option<Vec<u8>>,
+    payout_height: u32,
     /// Collateral outpoint (`txid` wire bytes, `vout`) from the ProRegTx.
     pub collateral: Option<([u8; 32], u32)>,
     /// A ProUpRevTx was seen ⇒ the masternode was revoked ("previously
@@ -1152,17 +1165,49 @@ where
                     agg.voting_key_hash = Some(provider_hash_to_20(p.voting_key_hash.as_ref()));
                     agg.voting_height = height;
                 }
+                if agg.operator_public_key.is_none() || height >= agg.operator_height {
+                    let bls: &[u8; 48] = p.operator_public_key.as_ref();
+                    agg.operator_public_key = Some(*bls);
+                    agg.operator_height = height;
+                }
+                if agg.platform_node_id.is_none() || height >= agg.platform_node_height {
+                    // Evonode-only; `None` on a regular masternode.
+                    if let Some(node_id) = p.platform_node_id.as_ref() {
+                        agg.platform_node_id = Some(provider_hash_to_20(node_id.as_ref()));
+                        agg.platform_node_height = height;
+                    }
+                }
+                if agg.payout_script.is_none() || height >= agg.payout_height {
+                    agg.payout_script = Some(p.script_payout.as_bytes().to_vec());
+                    agg.payout_height = height;
+                }
             }
             Some(TransactionPayload::ProviderUpdateServicePayloadType(p)) => {
                 if agg.service_address.is_none() || height >= agg.service_height {
                     agg.service_address = Some(provider_ip_port(p.ip_address, p.port));
                     agg.service_height = height;
                 }
+                // ProUpServ's `platform_node_id` is a raw `Option<[u8; 20]>`.
+                if let Some(node_id) = p.platform_node_id {
+                    if agg.platform_node_id.is_none() || height >= agg.platform_node_height {
+                        agg.platform_node_id = Some(node_id);
+                        agg.platform_node_height = height;
+                    }
+                }
             }
             Some(TransactionPayload::ProviderUpdateRegistrarPayloadType(p)) => {
                 if agg.voting_key_hash.is_none() || height >= agg.voting_height {
                     agg.voting_key_hash = Some(provider_hash_to_20(p.voting_key_hash.as_ref()));
                     agg.voting_height = height;
+                }
+                if agg.operator_public_key.is_none() || height >= agg.operator_height {
+                    let bls: &[u8; 48] = p.operator_public_key.as_ref();
+                    agg.operator_public_key = Some(*bls);
+                    agg.operator_height = height;
+                }
+                if agg.payout_script.is_none() || height >= agg.payout_height {
+                    agg.payout_script = Some(p.script_payout.as_bytes().to_vec());
+                    agg.payout_height = height;
                 }
             }
             Some(TransactionPayload::ProviderUpdateRevocationPayloadType(p)) => {
@@ -1258,6 +1303,26 @@ pub struct MasternodeEntryFFI {
     /// hashes keys.
     pub owner_address: *mut c_char,
     pub voting_address: *mut c_char,
+    /// Operator BLS public key (48 bytes), gated by `has_operator_key`.
+    pub operator_public_key: [u8; 48],
+    pub has_operator_key: bool,
+    /// Platform node id (hash160, 20 bytes) — evonodes only — gated by
+    /// `has_platform_node_id`.
+    pub platform_node_id: [u8; 20],
+    pub has_platform_node_id: bool,
+    /// Base58 payout address for the network, or null (non-standard
+    /// payout script, or none seen).
+    pub payout_address: *mut c_char,
+    // --- Base58 P2PKH addresses of the operator / platform keys ---
+    // Owner / voting addresses are `owner_address` / `voting_address`
+    // above. These two complete the set so the app can join ALL FOUR key
+    // kinds against persisted `PersistentCoreAddress` rows (address ⇒
+    // account type + index) — the durable ownership source. Pure encoding,
+    // no in-wallet lookup on the Rust side.
+    /// Base58 P2PKH pseudo-address of `hash160(operator BLS key)`, or null.
+    pub operator_pseudo_address: *mut c_char,
+    /// Base58 P2PKH address of the platform node id (evonode), or null.
+    pub platform_node_address: *mut c_char,
 }
 
 /// Encode a hash160 as a network-specific base58 P2PKH address string
@@ -1276,15 +1341,44 @@ fn masternode_p2pkh_cstring(hash: [u8; 20], network: dashcore::Network) -> *mut 
         .unwrap_or(std::ptr::null_mut())
 }
 
+/// Encode a payout `script` to its base58 address for `network`, or null
+/// when the script is non-standard (not addressable).
+fn masternode_payout_cstring(script_bytes: &[u8], network: dashcore::Network) -> *mut c_char {
+    let script = dashcore::ScriptBuf::from_bytes(script_bytes.to_vec());
+    match dashcore::Address::from_script(&script, network) {
+        Ok(address) => std::ffi::CString::new(address.to_string())
+            .map(std::ffi::CString::into_raw)
+            .unwrap_or(std::ptr::null_mut()),
+        Err(_) => std::ptr::null_mut(),
+    }
+}
+
 /// Flatten one aggregate into its C-ABI entry, encoding the owner /
-/// voting addresses for `network`. `order_index` is the caller's stable
-/// position in the sorted aggregate list.
+/// voting / payout / operator / platform-node base58 addresses for
+/// `network`. `order_index` is the caller's stable position in the sorted
+/// aggregate list. Key-ownership is resolved app-side by joining these
+/// address strings against persisted `PersistentCoreAddress` rows.
 pub(crate) fn masternode_entry_ffi(
     mn: &MasternodeAggregate,
     order_index: u32,
     network: dashcore::Network,
 ) -> MasternodeEntryFFI {
+    use dashcore::hashes::{hash160, Hash};
     use std::ffi::CString;
+
+    // Operator pseudo-address = P2PKH of hash160(BLS key); platform-node
+    // address = P2PKH of the 20-byte node id.
+    let operator_pseudo_address = mn
+        .operator_public_key
+        .map(|k| {
+            let h: [u8; 20] = hash160::Hash::hash(&k).to_byte_array();
+            masternode_p2pkh_cstring(h, network)
+        })
+        .unwrap_or(std::ptr::null_mut());
+    let platform_node_address = mn
+        .platform_node_id
+        .map(|h| masternode_p2pkh_cstring(h, network))
+        .unwrap_or(std::ptr::null_mut());
 
     let service_address = match &mn.service_address {
         Some(s) => CString::new(s.clone())
@@ -1303,6 +1397,11 @@ pub(crate) fn masternode_entry_ffi(
     let voting_address = mn
         .voting_key_hash
         .map(|h| masternode_p2pkh_cstring(h, network))
+        .unwrap_or(std::ptr::null_mut());
+    let payout_address = mn
+        .payout_script
+        .as_deref()
+        .map(|s| masternode_payout_cstring(s, network))
         .unwrap_or(std::ptr::null_mut());
 
     MasternodeEntryFFI {
@@ -1326,6 +1425,13 @@ pub(crate) fn masternode_entry_ffi(
         service_address,
         owner_address,
         voting_address,
+        operator_public_key: mn.operator_public_key.unwrap_or([0u8; 48]),
+        has_operator_key: mn.operator_public_key.is_some(),
+        platform_node_id: mn.platform_node_id.unwrap_or([0u8; 20]),
+        has_platform_node_id: mn.platform_node_id.is_some(),
+        payout_address,
+        operator_pseudo_address,
+        platform_node_address,
     }
 }
 
@@ -1766,6 +1872,21 @@ mod tests {
         assert!(mn.owner_key_hash.is_some());
         assert!(mn.voting_key_hash.is_some());
         assert!(mn.collateral.is_some());
+        // #4116 key-ownership extraction: operator BLS key + payout script
+        // are lifted; the legacy (v1) fixture is a regular MN so it has no
+        // platform node id.
+        assert!(
+            mn.operator_public_key.is_some(),
+            "ProRegTx carries a 48-byte operator BLS key"
+        );
+        assert!(
+            mn.payout_script.as_ref().map_or(false, |s| !s.is_empty()),
+            "ProRegTx carries a payout script"
+        );
+        assert!(
+            mn.platform_node_id.is_none(),
+            "legacy regular-MN fixture has no platform node id"
+        );
         assert_eq!(mn.tx_count, 1);
     }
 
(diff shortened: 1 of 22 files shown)


thepastaclaw:
⛔ Blockers found — Sonnet deferred (commit d93dda19)
_Canonical validated blockers: 8_

QuantumExplorer:
Reviewed

packages/rs-platform-wallet/src/manager/accessors.rs
```diff
@@ -783,6 +809,121 @@ impl<P: PlatformWalletPersistence + 'static> PlatformWalletManager<P> {
...
+                    None,
+                    false,
+                ) {
+                    Ok(key) => {
+                        if let Some(node_id) = key.node_id {
+                            platform_index.insert(node_id, index);
+                        }
+                    }
+                    // No platform account, or watch-only (needs the seed). Stop.
+                    Err(_) => break,
```
thepastaclaw, line 919: **🔴 Blocking: Platform-node ownership can never resolve for managed wallets**

This passes no resolved master while deriving hardened Ed25519 platform-node keys and stops on the first error. `register_wallet` unconditionally downgrades the wallet to external-signable before constructing its `PlatformWallet`, and restored wallets are also external-signable, so neither path has the resident private root required by `derive_provider_key_at_index`. Registration persists a pre-derived platform-node key batch, but this accessor does not consume it or invoke the mnemonic resolver; consequently `platform_index` remains empty and every platform-node key is reported as outside the wallet.

<sub>source: ['codex']</sub>

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/MasternodeSync.swift
```diff
@@ -0,0 +1,150 @@
...
+        guard let address, !address.isEmpty else { return (false, 0, 0) }
+        let descriptor = FetchDescriptor<PersistentCoreAddress>(
+            predicate: #Predicate { $0.address == address }
+        )
+        guard let row = try? modelContext.fetch(descriptor).first,
+              let account = row.account
+        else {
+            return (false, 0, 0)
+        }
+        return (true, UInt8(truncatingIfNeeded: account.accountType), row.addressIndex)
```
thepastaclaw, line 148: **🔴 Blocking: Key ownership lookup is not scoped to the masternode wallet**

The helper fetches the globally matching address without checking `row.account.wallet.walletId`. Provider roles can be split across wallets, allowing the same ProRegTx aggregate to appear for wallet A through its owner key and wallet B through its voting key. Wallet B then resolves wallet A's owner row and records its account type and index as B's ownership; the detail view trusts that flag to enable Claim and passes wallet B together with wallet A's index. The current Rust owner-key guard fails closed, but the displayed ownership and claim metadata are still wrong. Pass the aggregate's wallet ID into this helper and require the fetched account to belong to it.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/spv/runtime.rs
```diff
@@ -292,6 +292,45 @@ impl SpvRuntime {
...
+        let mut map = std::collections::HashMap::with_capacity(list.masternodes.len());
+        for qualified in list.masternodes.values() {
+            let entry = &qualified.masternode_list_entry;
+            // `pro_reg_tx_hash` is internal order (the DML map itself keys
+            // by the reversed/display form, so read it off the entry).
+            let mut pro_tx = [0u8; 32];
+            pro_tx.copy_from_slice(entry.pro_reg_tx_hash.as_ref());
+            map.insert(pro_tx, entry.is_valid);
+        }
+        Some(map)
```
thepastaclaw, line 331: **🔴 Blocking: An intermediate DML snapshot is treated as fully synchronized**

`latest_masternode_list()` only returns the last list stored in the engine; it does not indicate synchronization readiness. The pinned dash-spv implementation restores that list during startup while explicitly setting masternode progress to `WaitingForConnections`, so this accessor can expose a stale checkpoint or intermediate list despite promising `None` until synchronization completes. The FFI treats every missing proTxHash in that snapshot as absent and therefore Retired, producing incorrect persisted status for nodes registered or retired after the snapshot. Gate the result on masternode sync state and list height relative to the current header tip.

<sub>source: ['codex']</sub>

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Views/IdentitiesContentView.swift
```diff
@@ -199,6 +233,78 @@ struct IdentitiesContentView: View {
...
+        // Refresh the persisted masternode aggregation from Rust when the
+        // tab appears. Pure persist step — Rust owns the aggregation
+        // (`MasternodeSync.refresh`).
+        .onAppear {
+            MasternodeSync.refresh(
+                walletManager: walletManager,
+                walletIds: walletIdsOnNetwork,
+                modelContext: modelContext
+            )
+        }
```
thepastaclaw, line 245: **🔴 Blocking: Masternodes are never refreshed when Core or DML synchronization completes**

The only repository call to `MasternodeSync.refresh` is this `.onAppear` handler. Pull-to-refresh runs only Platform balance synchronization, and the view has no completion observer for provider-record restoration, Core rescan, or DML progress. If the tab appears before those operations finish, its masternode rows remain empty, Unknown, or stale after synchronization because ordinary progress updates do not retrigger `.onAppear`; the user must leave and re-enter the screen. Refresh on the relevant sync-completion changes and include masternodes in the explicit refresh action.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/core_wallet_types.rs
```diff
@@ -928,48 +955,520 @@ struct ProviderPayloadFields {
...
+        ha.cmp(&hb).then_with(|| a.pro_tx_hash.cmp(&b.pro_tx_hash))
+    });
+
+    // Resolve authoritative status against the DML and assign per-type
+    // numbering (separate Evonode / Masternode sequences), both in the
+    // stable registration order established above.
+    let mut evonode_n: u32 = 0;
+    let mut masternode_n: u32 = 0;
+    for agg in result.iter_mut() {
+        agg.status = MasternodeStatus::from_membership(list_lookup(&agg.pro_tx_hash));
```
thepastaclaw, line 1248: **🔴 Blocking: Unconfirmed registrations are deterministically labeled Retired**

The accessor includes unconfirmed provider records with height zero, but this loop assigns status solely from current DML membership. A newly broadcast ProRegTx cannot yet appear in the confirmed DML, so an available and fully current list still returns `Absent`, which maps directly to Retired. The Swift view hides Retired rows by default, making a pending registration disappear immediately. Treat a registration with `registration_height == 0` as Unknown or Pending until it confirms instead of applying the retirement rule.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/core_wallet_types.rs
```diff
@@ -928,48 +955,520 @@ struct ProviderPayloadFields {
...
+                    agg.service_address = Some(provider_ip_port(p.ip_address, p.port));
+                    agg.service_height = height;
+                }
+                // ProUpServ's `platform_node_id` is a raw `Option<[u8; 20]>`.
+                if let Some(node_id) = p.platform_node_id {
+                    if agg.platform_node_id.is_none() || height >= agg.platform_node_height {
+                        agg.platform_node_id = Some(node_id);
+                        agg.platform_node_height = height;
+                    }
+                }
```
thepastaclaw, line 1196: **🔴 Blocking: Update-only evonodes are classified as regular masternodes**

Update-only aggregation is a supported case when the wallet retained a ProUpServTx without its ProRegTx. Such an aggregate starts with `is_evonode == false`, and this branch updates its service and platform-node fields while ignoring `ProviderUpdateServicePayload.mn_type`. BasicBLS ProUpServ payloads explicitly identify HighPerformance nodes, so an update-only evonode carries sufficient type information but is rendered and numbered as a regular masternode. Set the aggregate's type from `mn_type` when processing the update.

<sub>source: ['codex']</sub>

packages/swift-sdk/Sources/SwiftDashSDK/Persistence/Models/PersistentMasternode.swift
```diff
@@ -0,0 +1,277 @@
...
+/// is recovered by joining `walletId` to the parent
+/// `PersistentWallet.networkRaw`, the same pivot `PersistentPlatformAddress`
+/// / shielded rows use. `(walletId, proTxHash)` is unique so a masternode
+/// tracked under one wallet collapses onto a single row across re-syncs.
+@Model
+public final class PersistentMasternode {
+    #Unique<PersistentMasternode>([\.walletId, \.proTxHash])
+
+    /// Owning wallet id (32 bytes). Network is derived via the wallet.
+    public var walletId: Data
```
thepastaclaw, line 24: **🔴 Blocking: Wallet deletion leaves persistent masternode metadata behind**

This model stores only a raw wallet ID and has no relationship to `PersistentWallet`, so deleting the wallet cannot cascade these rows. `PlatformWalletPersistenceHandler.deleteWalletData` explicitly removes other raw-wallet-ID tables but never fetches or deletes `PersistentMasternode`. Deleting and reimporting the same deterministic wallet ID can therefore resurrect stale ownership and status metadata, especially because `MasternodeSync` intentionally does not prune persisted rows when the fresh aggregation is empty. Delete these rows explicitly during the wallet wipe or add an appropriate cascade relationship.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/core_wallet_types.rs
```diff
@@ -903,6 +903,33 @@ fn transaction_type_to_u8(
...
+/// Rebuild an `"ip:port"` string from a ProUpServTx-style little-endian
+/// IPv6-mapped `u128` address + `port`, collapsing IPv4-mapped addresses
+/// to V4 so a normal masternode renders as `"1.2.3.4:port"`.
+fn provider_ip_port(ip_address: u128, port: u16) -> String {
+    let v6 = std::net::Ipv6Addr::from(ip_address.to_le_bytes());
+    let ip = v6
+        .to_ipv4_mapped()
+        .map(std::net::IpAddr::V4)
+        .unwrap_or(std::net::IpAddr::V6(v6));
+    format!("{}:{}", ip, port)
```
thepastaclaw, line 930: **🔴 Blocking: IPv6 service endpoints are emitted without port brackets**

Formatting an IPv6 address and port independently emits an ambiguous value such as `2001:db8::1:9999` instead of `[2001:db8::1]:9999`. ProReg endpoints already use `SocketAddr::to_string()`, so the same native IPv6 endpoint also changes representation when a ProUpServTx supersedes it. Constructing a `SocketAddr` preserves valid and consistent IPv4 and IPv6 syntax.

```suggestion
std::net::SocketAddr::new(ip, port).to_string()
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

Verified exact HEAD d93dda196397b37d79df75700253d9217f1dc415 against the full ad657c81e4b3b345576efb860274f2e24ab50ed3..d93dda196397b37d79df75700253d9217f1dc415 range. All eight Codex findings reproduce in the new masternode list, detail, ownership, and claim flows; they cause incorrect ownership, stale or invalid status, misclassification, orphaned persistence, and malformed IPv6 endpoints. These in-scope defects should be fixed before merge.

### Review provenance

🔴 8 blocking

<details>

```

- [BLOCKING] packages/rs-platform-wallet/src/manager/accessors.rs:907-919: Platform-node ownership can never resolve for managed wallets
  This passes no resolved master while deriving hardened Ed25519 platform-node keys and stops on the first error. `register_wallet` unconditionally downgrades the wallet to external-signable before constructing its `PlatformWallet`, and restored wallets are also external-signable, so neither path has the resident private root required by `derive_provider_key_at_index`. Registration persists a pre-derived platform-node key batch, but this accessor does not consume it or invoke the mnemonic resolver; consequently `platform_index` remains empty and every platform-node key is reported as outside the wallet.

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/MasternodeSync.swift`:
- [BLOCKING] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/MasternodeSync.swift:135-148: Key ownership lookup is not scoped to the masternode wallet
  The helper fetches the globally matching address without checking `row.account.wallet.walletId`. Provider roles can be split across wallets, allowing the same ProRegTx aggregate to appear for wallet A through its owner key and wallet B through its voting key. Wallet B then resolves wallet A's owner row and records its account type and index as B's ownership; the detail view trusts that flag to enable Claim and passes wallet B together with wallet A's index. The current Rust owner-key guard fails closed, but the displayed ownership and claim metadata are still wrong. Pass the aggregate's wallet ID into this helper and require the fetched account to belong to it.

- [BLOCKING] packages/rs-platform-wallet/src/spv/runtime.rs:319-331: An intermediate DML snapshot is treated as fully synchronized
  `latest_masternode_list()` only returns the last list stored in the engine; it does not indicate synchronization readiness. The pinned dash-spv implementation restores that list during startup while explicitly setting masternode progress to `WaitingForConnections`, so this accessor can expose a stale checkpoint or intermediate list despite promising `None` until synchronization completes. The FFI treats every missing proTxHash in that snapshot as absent and therefore Retired, producing incorrect persisted status for nodes registered or retired after the snapshot. Gate the result on masternode sync state and list height relative to the current header tip.

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Views/IdentitiesContentView.swift`:
- [BLOCKING] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Views/IdentitiesContentView.swift:233-245: Masternodes are never refreshed when Core or DML synchronization completes
  The only repository call to `MasternodeSync.refresh` is this `.onAppear` handler. Pull-to-refresh runs only Platform balance synchronization, and the view has no completion observer for provider-record restoration, Core rescan, or DML progress. If the tab appears before those operations finish, its masternode rows remain empty, Unknown, or stale after synchronization because ordinary progress updates do not retrigger `.onAppear`; the user must leave and re-enter the screen. Refresh on the relevant sync-completion changes and include masternodes in the explicit refresh action.

- [BLOCKING] packages/rs-platform-wallet-ffi/src/core_wallet_types.rs:1247-1248: Unconfirmed registrations are deterministically labeled Retired
  The accessor includes unconfirmed provider records with height zero, but this loop assigns status solely from current DML membership. A newly broadcast ProRegTx cannot yet appear in the confirmed DML, so an available and fully current list still returns `Absent`, which maps directly to Retired. The Swift view hides Retired rows by default, making a pending registration disappear immediately. Treat a registration with `registration_height == 0` as Unknown or Pending until it confirms instead of applying the retirement rule.
- [BLOCKING] packages/rs-platform-wallet-ffi/src/core_wallet_types.rs:1185-1196: Update-only evonodes are classified as regular masternodes
  Update-only aggregation is a supported case when the wallet retained a ProUpServTx without its ProRegTx. Such an aggregate starts with `is_evonode == false`, and this branch updates its service and platform-node fields while ignoring `ProviderUpdateServicePayload.mn_type`. BasicBLS ProUpServ payloads explicitly identify HighPerformance nodes, so an update-only evonode carries sufficient type information but is rendered and numbered as a regular masternode. Set the aggregate's type from `mn_type` when processing the update.
- [BLOCKING] packages/rs-platform-wallet-ffi/src/core_wallet_types.rs:924-930: IPv6 service endpoints are emitted without port brackets
  Formatting an IPv6 address and port independently emits an ambiguous value such as `2001:db8::1:9999` instead of `[2001:db8::1]:9999`. ProReg endpoints already use `SocketAddr::to_string()`, so the same native IPv6 endpoint also changes representation when a ProUpServTx supersedes it. Constructing a `SocketAddr` preserves valid and consistent IPv4 and IPv6 syntax.

In `packages/swift-sdk/Sources/SwiftDashSDK/Persistence/Models/PersistentMasternode.swift`:
- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/Persistence/Models/PersistentMasternode.swift:20-24: Wallet deletion leaves persistent masternode metadata behind
  This model stores only a raw wallet ID and has no relationship to `PersistentWallet`, so deleting the wallet cannot cascade these rows. `PlatformWalletPersistenceHandler.deleteWalletData` explicitly removes other raw-wallet-ID tables but never fetches or deletes `PersistentMasternode`. Deleting and reimporting the same deterministic wallet ID can therefore resurrect stale ownership and status metadata, especially because `MasternodeSync` intentionally does not prune persisted rows when the fresh aggregation is empty. Delete these rows explicitly during the wallet wipe or add an appropriate cascade relationship.
```
</details>

PR #884, opened 2023-04-04
fix: llmqType must be equal to one of the allowed values

strophy:
## Issue being fixed or feature implemented

Dashmate integration on testnet was failing due to inability to select llmqType 6, which is required on testnet.
```
fatal: [hp-masternode-3]: FAILED! => changed=false 
  cmd:
  - yarn
  - dashmate
  - restart
  delta: '0:00:11.555743'
  end: '2023-04-04 13:05:12.249023'
  msg: non-zero return code
  rc: 1
  start: '2023-04-04 13:05:00.693280'
  stderr: |2-
        InvalidConfigFileFormatError: Invalid
        '/home/dashmate/.dashmate/config.json' config format:
        config/platform/drive/abci/validatorSet/llmqType must be equal to one of
        the allowed values
  stderr_lines: <omitted>
  stdout: ''
  stdout_lines: <omitted>
```

## What was done?

Update enum `llmqType`

## How Has This Been Tested?

In testing

## Breaking Changes

None

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added or updated relevant unit/integration/functional/e2e tests
- [x] I have made corresponding changes to the documentation

**For repository code-owners and collaborators only**
- [x] I have assigned this pull request to a milestone

PR #4120, opened 2026-07-13
feat(platform-wallet): fixed provider key derivation via rust-dashcore bump + legacy BLS display

QuantumExplorer:

key-wallet derived wrong masternode operator BLS keys and platform node ids (dashpay/rust-dashcore#878) — they never matched DashSync/on-chain registrations, so #4116's operator/platform ownership matching could never succeed. The fix merged upstream as dashpay/rust-dashcore#879; this picks it up and adds the dual BLS serialization display users expect from dashwallet-ios.

## What was done?

- **Pin bump** rust-dashcore `337f6ccc` → `56a40788`: #879 (BLS operator / Ed25519 platform-node derivation per DashSync/dashbls — drops the spurious `0x00` in hardened BLS children, uses legacy G1 serialization in the HD chain, feeds the bip39 seed directly instead of the secp256k1 hybrid with double path application), #877 (mainnet sync floors at the HD/BIP39 activation height), #874 (masternode seed refresh).
- **Dual BLS serialization display**: `ProviderDerivedKey` (rs-platform-wallet) now carries `legacy_public_key_bytes` for BLS operator keys (via #879's `to_bytes_legacy()` / `public_key_bytes_legacy()` — Rust-side only, no Swift crypto), threaded through `ProviderKeyAtIndexFFI` and the Swift DTO; the Provider Operator Keys account screen renders both "BLS Public Key" (modern, `82…`) and "BLS Public Key (Legacy)" (`02…`) as copyable rows. Ed25519 platform-node keys are unaffected (no legacy row).

Note: wallets created before this bump persisted provider account roots derived under the broken scheme — a wallet delete + re-import is needed for the corrected keys (the old keys never corresponded to anything on-chain). Follow-up (deliberately excluded): the masternode detail page's raw on-chain operator key row could also show both forms, but that requires a version-aware re-encode (ProRegTx v1 payloads carry legacy encoding, v2+ modern) threaded through the aggregation.

**Follow-up rounds on this PR:**
- Deleted platform-wallet's duplicated derivation and, after dashpay/rust-dashcore#881 landed (from our #880), shrank `provider_key_at_index.rs` to a pure adapter over the gate-free upstream API (`operator_public_key_at` / `operator_private_key_at` / `platform_node_key_at`) — platform holds zero derivation knowledge now. An always-on consistency check guards the private-reveal (the upstream public and private paths derive from independent sources).
- Provider key accounts now route through the address-pool restore, so their persisted address rows rehydrate the in-memory pools at load (previously dropped by the `_ => None` arm).
- Dropped the provider-scheme versioning (pre-release; stale dev wallets delete+re-import).
- User-verified end-to-end: same fresh mnemonic imported in dashwallet-ios (DashSync) and this app on mainnet produces identical operator keypairs (modern + legacy + private).

## How Has This Been Tested?

- `cargo check --workspace --all-targets` and `cargo test -p platform-wallet-ffi` (152 tests) green against the new pin; clippy clean.
- `build_ios.sh --target sim` (warnings-as-errors) green, cbindgen headers regenerated with the new FFI field.
- Runtime verification of the corrected derivation requires a wallet re-import (persisted provider account roots predate the fix); the derive-and-compare ownership matching from #4116 is already wired to consume the corrected keys.

## Breaking Changes

None.

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Added legacy BLS operator key support to provider key results, including legacy public-key hex exposure in the Swift SDK and account details UI.
  - Provider key display now supports rendering derived public keys in hex and base64, with an optional “Legacy” row when available.
- **Bug Fixes**
  - Improved restoration and merging of persisted provider key material and core address pools to prevent provider-key matching regressions.
  - Fixed masternode ownership/aggregation consistency during delete/reimport and refresh cycles; ensured clean empty outputs for masternode list calls.
- **Documentation**
  - Updated platform-node id descriptions to use the Tenderdash SHA256[..20] convention, including related persistence notes.

### DIFF
--- a/packages/rs-platform-wallet-ffi/src/persistence.rs
+++ b/packages/rs-platform-wallet-ffi/src/persistence.rs
@@ -2756,9 +2756,28 @@ unsafe fn address_info_from_ffi(
 /// the same index, and the reverse-lookup maps + `highest_*`
 /// watermarks are extended to cover indices past that default
 /// gap window.
+///
+/// **Typed-key preservation:** a persisted row can lose its typed public
+/// key on the FFI round-trip — the [`CoreAddressEntryFFI`] key field is a
+/// bare 33-byte slot that only fits a compressed ECDSA key and carries no
+/// key-type discriminator, so BLS (48B) operator and Ed25519 (32B)
+/// platform-node entries come back with `public_key: None`. Overwriting a
+/// prederived typed entry (which `from_wallet` filled with
+/// `PublicKeyType::BLS`/`EdDSA`) with such a `None` row would break
+/// key-wallet's provider-key matching. So this MERGES: when the incoming
+/// row has no typed key but the existing entry at that index does, the
+/// existing typed key is kept. (Funds/ECDSA rows always carry their key,
+/// so the merge is a no-op for them.)
 fn restore_address_pool(pool: &mut AddressPool, infos: Vec<AddressInfo>) {
-    for info in infos {
+    for mut info in infos {
         let idx = info.index;
+        if info.public_key.is_none() {
+            if let Some(existing) = pool.addresses.get(&idx) {
+                if existing.public_key.is_some() {
+                    info.public_key = existing.public_key.clone();
+                }
+            }
+        }
         pool.address_index.insert(info.address.clone(), idx);
         pool.script_pubkey_index
             .insert(info.script_pubkey.clone(), idx);
@@ -2980,6 +2999,79 @@ unsafe fn restore_core_address_pools(
         routed: pools_routed,
         dropped: pools_dropped,
     })
+}
+
+/// Rehydrate the persisted platform-node (Ed25519) key `batch` into the
+/// managed `ProviderPlatformKeys` pool.
+///
+/// SLIP-10 Ed25519 is hardened-only, so `ManagedWalletInfo::from_wallet`
+/// can't re-derive these keys from the account xpub on an external-signable
+/// restore — the batch (persisted at registration, supplied by the host on
+/// load) is the only seedless source. Each entry is inserted as a typed
+/// [`PublicKeyType::EdDSA`] pool address, keyed by its hardened index, with
+/// the platform node id recomputed as `SHA256(pubkey)[..20]` (rust-dashcore
+/// #884) — **never** trusting any persisted hash160-era id. This lets the
+/// masternode-ownership scan match a ProRegTx `platform_node_id` to the
+/// wallet's own derivation index on a seedless wallet.
+fn restore_platform_node_pool(
+    wallet_info: &mut ManagedWalletInfo,
+    batch: &[(u32, [u8; 32])],
+    network: Network,
+) -> Result<(), PersistenceError> {
+    use dashcore::hashes::Hash;
+    use key_wallet::managed_account::managed_account_trait::ManagedAccountTrait;
+    if batch.is_empty() {
+        return Ok(());
+    }
+    let Some(account) = wallet_info.accounts.provider_platform_keys.as_mut() else {
+        return Ok(());
+    };
+    let account_path = AccountType::ProviderPlatformKeys
+        .derivation_path(network)
+        .map_err(|e| PersistenceError::backend(format!("platform-node account path: {:?}", e)))?;
+    let base_children: Vec<key_wallet::bip32::ChildNumber> = account_path.as_ref().to_vec();
+
+    let mut infos: Vec<AddressInfo> = Vec::with_capacity(batch.len());
+    for (index, pubkey) in batch {
+        // Recompute the node id from the pubkey (SHA256[..20], #884).
+        let node_id = dashcore::PlatformNodeId::from_ed25519_public_key(pubkey).to_byte_array();
+        let payload =
+            dashcore::address::Payload::PubkeyHash(dashcore::PubkeyHash::from_byte_array(node_id));
+        let address = dashcore::Address::new(network, payload);
+        let script_pubkey = address.script_pubkey();
+        let child = key_wallet::bip32::ChildNumber::from_hardened_idx(*index).map_err(|e| {
+            PersistenceError::backend(format!("platform-node child index {}: {:?}", index, e))
+        })?;
+        let mut children = base_children.clone();
+        children.push(child);
+        let path = DerivationPath::from(children);
+        infos.push(AddressInfo {
+            address,
+            script_pubkey,
+            public_key: Some(PublicKeyType::EdDSA(pubkey.to_vec())),
+            index: *index,
+            path,
+            used: false,
+            generated_at: 0,
+            used_at: None,
+            tx_count: 0,
+            total_received: 0,
+            total_sent: 0,
+            balance: 0,
+            label: None,
+            metadata: std::collections::BTreeMap::new(),
+        });
+    }
+    // The platform-node pool is `AbsentHardened` (SLIP-10 hardened-only).
+    if let Some(pool) = account
+        .managed_account_type_mut()
+        .address_pools_mut()
+        .into_iter()
+        .find(|p| p.pool_type == AddressPoolType::AbsentHardened)
+    {
+        restore_address_pool(pool, infos);
+    }
+    Ok(())
 }
 
 /// Bucket a slice of upstream-emitted `DerivedAddress` entries into the
@@ -3205,6 +3297,13 @@ fn build_wallet_start_state(
 
     // Build the per-account collection from the typed spec array.
     let mut accounts = AccountCollection::new();
+    // The persisted platform-node (Ed25519) key batch, captured from the
+    // `ProviderPlatformKeys` spec so it can be rehydrated into the managed
+    // pool once `ManagedWalletInfo::from_wallet` has created it. SLIP-10 is
+    // hardened-only, so from_wallet can't re-derive these seedlessly on the
+    // external-signable restore — the batch is the only source. Stored as
+    // (index, pubkey); node ids are recomputed SHA256[..20] at insert.
+    let mut platform_node_batch: Vec<(u32, [u8; 32])> = Vec::new();
     let specs: &[AccountSpecFFI] = if entry.accounts.is_null() || entry.accounts_count == 0 {
         &[]
     } else {
@@ -3305,6 +3404,21 @@ fn build_wallet_start_state(
                         e
                     ))
                 })?;
+                // Capture the persisted platform-node key batch (if the host
+                // supplied it on this load) for rehydration into the managed
+                // pool below. Only the pubkeys are trusted; node ids are
+                // recomputed under the #884 convention at insert time.
+                if !spec.derived_platform_node_keys.is_null()
+                    && spec.derived_platform_node_keys_count > 0
+                {
+                    let rows = unsafe {
+                        slice::from_raw_parts(
+                            spec.derived_platform_node_keys,
+                            spec.derived_platform_node_keys_count,
+                        )
+                    };
+                    platform_node_batch = rows.iter().map(|r| (r.index, r.public_key)).collect();
+                }
                 continue;
             }
             _ => {}
@@ -3414,6 +3528,14 @@ fn build_wallet_start_state(
             restore_core_address_pools(&mut wallet_info, pool_entries, network, &entry.wallet_id)?;
         }
     }
+
+    // Rehydrate the platform-node (Ed25519) key batch into its managed pool.
+    // Unlike the ECDSA/BLS pools this can't come from `core_address_pools`
+    // (SLIP-10 is hardened-only and the 33-byte pool ABI can't carry a typed
+    // 32-byte Ed25519 key), so it flows through the dedicated batch captured
+    // from the `ProviderPlatformKeys` spec above. Without it a seedless
+    // restore has no platform-node keys to match masternode ownership.
+    restore_platform_node_pool(&mut wallet_info, &platform_node_batch, network)?;
 
     // Persisted unspent UTXOs → funds-bearing accounts. Keys-only and
     // PlatformPayment variants are skipped: the former never carry
@@ -5239,6 +5361,97 @@ mod tests {
 
         drop(addr_c);
         drop(path_c);
+    }
+
+    /// Build a minimal `AddressInfo` for merge tests. The address/script/
+    /// path content is irrelevant — `restore_address_pool`'s merge only
+    /// inspects `public_key` — so a P2PKH keyed off `index` suffices.
+    fn merge_test_address_info(index: u32, public_key: Option<PublicKeyType>) -> AddressInfo {
+        use dashcore::hashes::Hash;
+        let mut h = [0u8; 20];
+        h[0] = index as u8;
+        let payload =
+            dashcore::address::Payload::PubkeyHash(dashcore::PubkeyHash::from_byte_array(h));
+        let address = dashcore::Address::new(Network::Testnet, payload);
+        let script_pubkey = address.script_pubkey();
+        AddressInfo {
+            address,
+            script_pubkey,
+            public_key,
+            index,
+            path: DerivationPath::from(Vec::<key_wallet::bip32::ChildNumber>::new()),
+            used: false,
+            generated_at: 0,
+            used_at: None,
+            tx_count: 0,
+            total_received: 0,
+            total_sent: 0,
+            balance: 0,
+            label: None,
+            metadata: std::collections::BTreeMap::new(),
+        }
+    }
+
+    /// Blocker #1: a prederived operator entry carrying a typed BLS (48B)
+    /// public key must NOT be clobbered by a persisted row that lost its
+    /// typed key on the 33-byte-slot FFI round-trip (`public_key: None`).
+    /// Pins the merge in `restore_address_pool` so key-wallet's provider-key
+    /// matching keeps a usable typed key after restore.
+    #[test]
+    fn restore_address_pool_merge_preserves_typed_bls_operator_key() {
+        use key_wallet::managed_account::managed_account_trait::ManagedAccountTrait;
+
+        // Any managed keys account gives a real `AddressPool` to exercise
+        // the (pool-type-agnostic) merge against.
+        let mut wallet_info = test_managed_wallet_info_with_provider_owner();
+        let owner = wallet_info
+            .accounts
+            .provider_owner_keys
+            .as_mut()
+            .expect("managed provider-owner account must exist");
+        let pool = owner
+            .managed_account_type_mut()
+            .address_pools_mut()
+            .into_iter()
+            .next()
+            .expect("the account must have at least one pool");
+
+        let bls = vec![0xABu8; 48];
+        // Seed a prederived typed BLS entry at index 0.
+        restore_address_pool(
+            pool,
+            vec![merge_test_address_info(
+                0,
+                Some(PublicKeyType::BLS(bls.clone())),
+            )],
+        );
+        // A persisted row that lost its typed key (None) at the SAME index
+        // must keep the existing BLS key, not overwrite it with None.
+        restore_address_pool(pool, vec![merge_test_address_info(0, None)]);
+        match pool
+            .addresses
+            .get(&0)
+            .expect("entry at index 0")
+            .public_key
+            .as_ref()
+        {
+            Some(PublicKeyType::BLS(bytes)) => {
+                assert_eq!(bytes, &bls, "the prederived BLS key must be preserved")
+            }
+            other => panic!("expected a preserved BLS key at index 0, got {:?}", other),
+        }
+
+        // Control: a `None` row at a fresh index with no existing entry
+        // stays `None` (merge only fills from an existing typed entry).
+        restore_address_pool(pool, vec![merge_test_address_info(7, None)]);
+        assert!(
+            pool.addresses
+                .get(&7)
+                .expect("entry at index 7")
+                .public_key
+                .is_none(),
+            "a None row with no prederived entry must remain None"
+        );
     }
 
     /// `account_xpub` must survive the persist→restore byte round-trip — it is
--- a/packages/rs-platform-wallet-ffi/src/persistence.rs
+++ b/packages/rs-platform-wallet-ffi/src/persistence.rs
@@ -3226,48 +3226,102 @@ fn build_wallet_start_state(
                     continue;
                 }
             };
-            let funds = match account_type {
+            // Resolve the persisted pool's target account to its
+            // `&mut ManagedAccountType` (where the address pools live).
+            // Funds-bearing accounts (`ManagedCoreFundsAccount`) live in the
+            // keyed maps; the four masternode key-material provider accounts
+            // (`ManagedCoreKeysAccount`) live in dedicated `Option` fields.
+            // Both implement `ManagedAccountTrait`, so `managed_account_type_mut()`
+            // unifies them to a single `&mut ManagedAccountType` and the
+            // pool-population below is identical. (A single match — rather
+            // than a funds-match-then-provider-fallback — is required: the
+            // borrow checker won't let a `None` fallback re-borrow
+            // `wallet_info.accounts` after the funds `get_mut`.)
+            let managed_type = match account_type {
                 AccountType::Standard {
                     index,
                     standard_account_type: StandardAccountType::BIP44Account,
-                } => wallet_info.accounts.standard_bip44_accounts.get_mut(&index),
+                } => wallet_info
+                    .accounts
+                    .standard_bip44_accounts
+                    .get_mut(&index)
+                    .map(|a| a.managed_account_type_mut()),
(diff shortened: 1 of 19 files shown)


thepastaclaw:
⛔ Blockers found — Sonnet deferred (commit 4ac9108f)

thepastaclaw (changes requested):

All three Codex findings are confirmed. The rust-dashcore #879 bump changes provider-account derivation to use the BIP39 seed directly, but Platform still duplicates the obsolete secp-hybrid derivation, restores incompatible persisted provider xpubs without an enforced migration, and indexes only the modern serialization when matching historical legacy BLS operator keys.

Source: Codex reviewer `gpt-5.6-sol` (general); verifier `gpt-5.6-sol`.

### Review provenance

🔴 3 blocking

<details>

```

In `packages/rs-platform-wallet/src/wallet/provider_key_at_index.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/wallet/provider_key_at_index.rs:268-296: Provider derivation still uses the obsolete secp-hybrid seed
  The pinned rust-dashcore #879 change makes provider accounts derive from the BIP39 seed directly, but this path still derives a secp256k1 child at the provider account path and passes its 32-byte scalar to the BLS or Ed25519 master. The registration helper repeats that obsolete derivation at lines 96-113, and the FFI resolver reduces the mnemonic to an ExtendedPrivKey before reaching this method. Fresh accounts created by Wallet::from_mnemonic therefore use different account keys from private operator reveals, platform-node derivations, and the platform-node batch persisted during registration. Debug builds can hit the existing account-xpub assertions; release builds can silently return or persist a different valid key. Preserve the raw BIP39 seed through the resolver and share the same provider derivation routine used by upstream account creation, with cross-path tests against Wallet::from_mnemonic.

- [BLOCKING] packages/rs-platform-wallet/src/manager/accessors.rs:897-900: Legacy BLS operator keys are omitted from ownership matching
  ProviderDerivedKey now returns modern and legacy serializations of the same BLS point, but the ownership map records only public_key_bytes. aggregate_masternodes preserves the exact 48 bytes carried by ProRegTx or ProUpRegTx, and masternode_entry_ffi performs an exact lookup against this map. Wallet-owned masternodes whose payload contains the legacy serialization therefore remain operator_in_wallet=false. Index both serializations under the same derivation index.

- [BLOCKING] packages/rs-platform-wallet-ffi/src/persistence.rs:3042-3065: Pre-fix provider xpubs are restored without migration or rejection
  The restore path decodes and reinstalls persisted BLS and Ed25519 provider xpubs without a derivation-version marker or compatibility check. Consequently, upgraded wallets continue using pre-#879 provider account points while fresh imports use the corrected direct-seed scheme. The commit message notes that existing wallets require deletion and re-import, but the implementation neither enforces nor communicates that requirement at load time, so stale provider keys remain silently usable. Rederive or migrate these accounts through the mnemonic resolver, version the persisted representation, or reject incompatible rows with an explicit re-import error.
```
</details>

QuantumExplorer:
Both CodeRabbit nitpicks addressed in 8b68f98c06:

- Extracted `platform_node_xpriv_at` — the seed → SLIP-10 master → DIP-3 account path → hardened child sequence now exists once, used by both `derive_platform_node_public_keys` and the per-index reveal branch.
- Promoted the operator seed↔xpub cross-check from a debug assertion to an always-on guard: a mismatch now returns a `KeyDerivation` error instead of handing out a key (fails closed on exactly the divergence this PR fixed). If the cross-check itself cannot run, it logs and proceeds rather than failing a working path — `derive_pub_legacy` is gate-free and pure, so that branch is effectively unreachable for valid accounts.

`cargo test -p platform-wallet` green (420 tests, golden vectors unchanged).

QuantumExplorer:
Both CodeRabbit test nitpicks addressed in the latest commit:

- `provider_owner_address_pool_round_trips_used_and_highest_index` — pins the new provider restore arms: a persisted `ProviderOwnerKeys` pool row (used index 50, beyond the gap window) round-trips into the managed pool ("routed: 1, dropped: 0"; a regression to the funds-only match fails the assert). Implemented via an extracted `restore_core_address_pools` helper per the module's extract-and-test convention.
- `operator_cross_check_rejects_mismatched_seed` — pins the guard's negative path: mnemonic A's xpub + mnemonic B's seed → `Err(KeyDerivation)`; the matching control still yields the golden secret.

All 430 tests green.

packages/rs-platform-wallet-ffi/src/persistence.rs
```diff
@@ -2771,6 +2771,217 @@ fn restore_address_pool(pool: &mut AddressPool, infos: Vec<AddressInfo>) {
...
+            AccountType::ProviderOperatorKeys => wallet_info
+                .accounts
+                .provider_operator_keys
+                .as_mut()
+                .map(|a| a.managed_account_type_mut()),
+            AccountType::ProviderPlatformKeys => wallet_info
+                .accounts
+                .provider_platform_keys
+                .as_mut()
+                .map(|a| a.managed_account_type_mut()),
```
thepastaclaw, line 2941: **🔴 Blocking: Preserve typed provider keys when restoring address pools**

The new provider-account arms restore BLS operator and Ed25519 platform-node pools through `CoreAddressEntryFFI`, whose 33-byte public-key field can represent only compressed ECDSA keys. `build_core_address_entry_ffi` consequently emits `has_public_key = false` for BLS and EdDSA entries, and `address_info_from_ffi` restores them as `public_key: None`. `restore_address_pool` then overwrites any prederived entry at that index. The pinned key-wallet requires `PublicKeyType::BLS` or `PublicKeyType::EdDSA` when matching and consuming these keys, so restored provider entries fail after relaunch. Extend the ABI with a key-type discriminator and enough storage for 48-byte BLS keys, or preserve/rederive the correctly typed key during restore. The new round-trip test covers only the ECDSA-based provider-owner account and does not exercise either affected type.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The raw-seed provider derivation and legacy BLS ownership fixes are correct, and the five targeted provider-key tests pass. The current head still has five blocking lifecycle and persistence defects affecting restored provider keys, masternode ownership, claims, and wallet deletion, plus one C-ABI output-contract issue.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — security-auditor (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — ffi-engineer (failed), `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — security-auditor (completed), `gpt-5.6-sol` — rust-quality (completed), `gpt-5.6-sol` — ffi-engineer (completed)

🔴 5 blocking | 🟡 1 suggestion(s)

_5 additional finding(s) omitted (not in diff)._

<details>

```

- [BLOCKING] packages/rs-platform-wallet-ffi/src/persistence.rs:2913-2922: Preserve typed provider keys when restoring address pools
  The new provider-account arms restore BLS operator and Ed25519 platform-node pools through `CoreAddressEntryFFI`, whose 33-byte public-key field can represent only compressed ECDSA keys. `build_core_address_entry_ffi` consequently emits `has_public_key = false` for BLS and EdDSA entries, and `address_info_from_ffi` restores them as `public_key: None`. `restore_address_pool` then overwrites any prederived entry at that index. The pinned key-wallet requires `PublicKeyType::BLS` or `PublicKeyType::EdDSA` when matching and consuming these keys, so restored provider entries fail after relaunch. Extend the ABI with a key-type discriminator and enough storage for 48-byte BLS keys, or preserve/rederive the correctly typed key during restore. The new round-trip test covers only the ECDSA-based provider-owner account and does not exercise either affected type.

- [BLOCKING] packages/rs-platform-wallet/src/manager/accessors.rs:922-927: Build platform-node ownership from the persisted key batch
  This loop calls platform-node derivation without a resolved seed. Registration persists a 20-key platform-node batch and then downgrades the wallet to external-signable, while restored wallets are also seedless. The first derivation therefore fails and leaves `platform_index` empty during the normal lifecycle. Swift subsequently overwrites `PersistentMasternode.platformInWallet` with this false result instead of matching the durable `derivedPlatformNodeKeys` batch, so wallet-owned platform node IDs are never recognized. Build the ownership index from the persisted batch or provide the same resolver-backed seed path used by explicit provider-key derivation.
- [BLOCKING] packages/rs-platform-wallet/src/manager/accessors.rs:888-890: Scan operator ownership through the pool's high-water mark
  The ownership index treats 20 as a permanent maximum, although BLS operator children are non-hardened and can be derived beyond the initial window from the account xpub. This PR also restores provider pools and their beyond-gap indices. Any operator key at index 20 or later is omitted from `operator_index`, so an otherwise valid wallet-owned masternode is reported as external. Derive through the restored/generated pool's highest index or build the index directly from its typed entries.

In `packages/rs-platform-wallet-ffi/src/wallet.rs`:
- [BLOCKING] packages/rs-platform-wallet-ffi/src/wallet.rs:374-382: Keep Claim unavailable while withdrawal is a stub
  The Swift masternode view presents an enabled Claim action for eligible balances and sends confirmed claims through this function, but this implementation unconditionally returns `ErrorWalletOperation` without fetching, signing, or broadcasting anything. Every user-confirmed claim is therefore guaranteed to fail. Hide or disable the action until the verified signer path is implemented, or complete the withdrawal orchestration before exposing it.
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/wallet.rs:220-231: Initialize masternode outputs before fallible lookup
  The public C function documents null and zero outputs when the wallet is not found, but invalid-handle and unknown-wallet paths return through `unwrap_option_or_return!` before writing either output. Swift initializes its locals, but other C callers can retain stale pointer and count values despite the documented contract. Establish the empty result immediately after validating the pointers.

- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift:3703-3731: Delete persisted masternodes with their wallet
  `PersistentMasternode` stores only a raw `walletId` and has no relationship through which deleting `PersistentWallet` could cascade. This deletion path explicitly purges other raw-ID tables but never fetches or deletes masternode rows, and `MasternodeSync` intentionally declines to prune when a fresh aggregate is empty. Deleting and reimporting the deterministic wallet ID can therefore resurrect stale masternode and ownership state indefinitely. Purge `PersistentMasternode` rows for the deleted wallet ID alongside the other raw-ID records.
```
</details>

packages/rs-platform-wallet-ffi/src/persistence.rs
QuantumExplorer: Fixed in c2797b7a9b — the provider address-pool restore now MERGES: `restore_address_pool` keeps an existing prederived typed key when the incoming `CoreAddressEntryFFI` row carries none (the 33-byte slot only fits compressed ECDSA, so BLS/EdDSA rows come back `public_key: None`), so typed provider-key matching survives relaunch. Chose the merge over an ABI extension since it is zero-ABI-change and correct for the operator pool (`from_wallet` prederives it from the xpub post-#882). Added `restore_address_pool_merge_preserves_typed_bls_operator_key` covering the BLS case. Separately, platform-node keys — which the pool restore genuinely cannot carry (SLIP-10 hardened-only, seedless lifecycle) — are threaded through the persisted EdDSA batch and reinserted as typed entries with node ids recomputed per #884.

QuantumExplorer:
Addressed the remaining review blockers (the body-only findings) in c2797b7a9b, alongside the pin bump to rust-dashcore 2045cd29 (#882/#884/#885):

- **Platform-node ownership (accessors seedless):** built from the persisted EdDSA batch (threaded restore→FFI→pool as typed entries), with node ids recomputed under #884 (SHA256[0..20]) — never trusting hash160-era persisted ids. MasternodeSync no longer downgrades a true `platformInWallet` on a transient absent signal.
- **Operator ownership beyond index 20:** the scan bound is now the operator pool's high-water mark (+1, floored at 20), so #882 gap-extension and restored beyond-window keys match.
- **Claim button:** gated off with a "not yet available (pending verified signer)" note while `masternode_withdraw` remains a stub — sheet plumbing retained for a one-line re-enable.
- **Wallet deletion:** now purges `PersistentMasternode` rows for the deleted walletId, so delete+reimport can't resurrect stale masternode/ownership state.
- **`list_masternodes` C-ABI:** writes empty `*out_entries=null; *out_count=0` immediately after pointer validation, before the fallible lookup.

Verified: `cargo test -p platform-wallet -p platform-wallet-ffi` green (incl. the #884 node-id golden `302f2615…`, the typed-key merge test, and the owner-pool round-trip), full `build_ios.sh --target sim` green.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -3767,6 +3767,20 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
...
+                // rule), so without an explicit purge a delete-then-reimport
+                // of the same wallet resurrects stale masternode rows
+                // indefinitely.
+                let masternodeDescriptor = FetchDescriptor<PersistentMasternode>(
+                    predicate: #Predicate<PersistentMasternode> { $0.walletId == walletId }
+                )
+                for row in try backgroundContext.fetch(masternodeDescriptor) {
+                    backgroundContext.delete(row)
+                }
+
```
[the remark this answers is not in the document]
QuantumExplorer: Not applied — the finding's premise doesn't hold: `walletId` is network-scoped by construction. key-wallet's `compute_wallet_id_from_root_extended_pub_key` appends a domain tag + network discriminant byte to the digest preimage (key-wallet/src/wallet/mod.rs:121-133, "a `Some(other_network)` id for the same key must **not** expect equality — those are intentionally different digests"), so the same mnemonic yields different walletIds per network and the ungated purge cannot touch a sibling network's rows. The cross-network tie is the separate network-independent `walletGroupId`. (The sibling `isLastNetworkRow` gating this finding cites predates the network-scoped id and is now vestigial — a cleanup candidate, but out of scope here.)

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/MasternodeSync.swift
```diff
@@ -82,15 +82,27 @@ enum MasternodeSync {
...
+                // yet". So never DOWNGRADE an already-established platform
+                // ownership to false on an existing row — mirror the
+                // no-prune-on-empty rule. It still upgrades to true and sets
+                // fresh true values (and new rows default to false, so they
+                // take the fresh value regardless).
+                if mn.platformInWallet || !row.platformInWallet {
+                    row.platformInWallet = mn.platformInWallet
+                    row.platformAccountType = mn.platformAccountType
+                    row.platformKeyIndex = mn.platformKeyIndex
+                }
```
thepastaclaw: **🔴 Blocking: Clear platform ownership after an on-chain key rotation**

Once an existing row has `platformInWallet == true`, this condition rejects every later definitive `false`. A valid ProUpServ transaction can rotate the evonode's platform node id from a wallet-owned Ed25519 key to an external key; Rust then reports the new node id and `platformInWallet == false`, but Swift updates `platformNodeId` while retaining the old ownership flag and derivation index. The UI therefore claims ownership of a key the wallet does not hold. Preserve transient restore behavior by exposing a checked/unknown state over the FFI, or distinguish an unchanged-key transient miss from a changed-key definitive non-match before retaining the prior value. This behavior entered the cumulative PR range in c2797b7a9b352319e374f0382770c81aac44d67f.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

All six indexed prior findings are fixed at the exact head. Two in-scope blocking defects remain: same-block provider updates are aggregated in txid order instead of block order, and persisted platform ownership cannot be cleared after an on-chain key rotation. Static formatting and diff checks passed; targeted Rust tests could not start because the pinned rust-dashcore revision was unavailable offline.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (completed)

🔴 2 blocking

<details>

```

- [BLOCKING] packages/rs-platform-wallet/src/manager/accessors.rs:853-866: Preserve block order when aggregating provider updates
  The BTreeMap deduplicates provider records by txid and `into_values()` subsequently feeds them to `aggregate_masternodes` in txid order. That aggregator resolves mutable fields with `height >= previous_height`, so equal-height updates overwrite one another according to txid ordering. Dash Core's `RebuildListFromBlock` applies provider transactions sequentially in `block.vtx` order and does not prohibit multiple updates for the same proTxHash in one block. Consequently, two valid same-block ProUpServ or ProUpReg transactions can leave this wallet showing an earlier service address, operator key, voting key, payout script, or platform node id as current. Retain transaction position alongside height and compare `(height, position)` when selecting the latest update. This behavior entered the cumulative PR range in ccaca2818d5ebf11a96a309662e7abe8a8ba873b.

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/MasternodeSync.swift`:
- [BLOCKING] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/MasternodeSync.swift:101-105: Clear platform ownership after an on-chain key rotation
  Once an existing row has `platformInWallet == true`, this condition rejects every later definitive `false`. A valid ProUpServ transaction can rotate the evonode's platform node id from a wallet-owned Ed25519 key to an external key; Rust then reports the new node id and `platformInWallet == false`, but Swift updates `platformNodeId` while retaining the old ownership flag and derivation index. The UI therefore claims ownership of a key the wallet does not hold. Preserve transient restore behavior by exposing a checked/unknown state over the FFI, or distinguish an unchanged-key transient miss from a changed-key definitive non-match before retaining the prior value. This behavior entered the cumulative PR range in c2797b7a9b352319e374f0382770c81aac44d67f.
```
</details>

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Services/MasternodeSync.swift
QuantumExplorer: Fixed in cc4cbcb69c. The guard is now a tri-state driven by a new `platform_ownership_checked` flag from Rust (set when the wallet's derived platform-node index actually had entries to compare against): when checked, `platformInWallet` is written verbatim — true OR false — so an on-chain rotation to an external key clears stale ownership; when unchecked (the seedless-restore transient this guard existed for), the prior value is retained. Belt-and-suspenders: a changed `platformNodeId` also overwrites regardless of checked-ness, since it's definitive on-chain evidence of a rotation.

QuantumExplorer:
Both blockers addressed in cc4cbcb69c:

**Same-block provider-update ordering** (`accessors.rs` aggregation): `aggregate_masternodes` now processes transactions in ascending `(height, in-block position)` — matching how Core's `RebuildListFromBlock` applies same-block provider updates in `block.vtx` order — instead of inheriting the `BTreeMap<Txid>` dedup's txid order. `ProviderMasternodeTxs` threads the position through, and `same_block_updates_resolve_by_position_not_txid` feeds the same same-height pair in both orders and asserts the later-positioned update wins each time.

One honest caveat: upstream retained records don't expose an in-block index — `BlockInfo` is only `{height, block_hash, timestamp}`, and the index is discarded during block processing (`key-wallet-manager/src/process_block.rs:60` iterates `block.txdata` without enumerating). Filed as [rust-dashcore#890](https://github.com/dashpay/rust-dashcore/issues/890); until it lands, the live caller passes `position = 0` (commented in code) and same-block ties fall back to feed order. The aggregation mechanism and test make the fix drop-in once the field exists — it rides the same future pin bump as [rust-dashcore#887](https://github.com/dashpay/rust-dashcore/issues/887) (node-id wire byte order, also deliberately not worked around downstream).

**Platform-ownership rotation clearing** (`MasternodeSync.swift`): replaced the height-only no-downgrade guard with a tri-state — see the inline thread reply for details.

Verification: `cargo test -p platform-wallet -p platform-wallet-ffi` all green (421 + 155 + 26, 0 failed); `cargo check --all-targets` + per-crate `--all-features --tests` clean; `build_ios.sh --target sim` + SwiftExampleApp build succeeded; app installed and launched on simulator.

QuantumExplorer:
rust-dashcore#887 is now fixed upstream ([rust-dashcore#889](https://github.com/dashpay/rust-dashcore/pull/889)) and pinned here in 4ac9108fcd (rev 7221d866): `PlatformNodeId::consensus_decode` normalizes wire bytes to canonical Tenderdash order at the consensus boundary, so payload-sourced node ids display forward and match the derived platform-node ownership index directly — no platform-side reversal ever needed. Rust suites green on the new pin (421 + 155 + 26, 0 failed); iOS sim build + app launch verified. Remaining known gap on this PR is [rust-dashcore#890](https://github.com/dashpay/rust-dashcore/issues/890) (in-block position for same-block provider updates), which is mechanism-ready downstream.

packages/rs-platform-wallet/src/manager/accessors.rs
```diff
@@ -855,39 +867,94 @@ impl<P: PlatformWalletPersistence + 'static> PlatformWalletManager<P> {
...
+                    // (txid-ordered, not processing order), so no real position is
+                    // recoverable here. Pass 0 for now; a correct fix needs an
+                    // upstream in-block tx index on `BlockInfo`, set during block
+                    // processing (rust-dashcore#890). Until then two same-block
+                    // updates to one field fall back to feed order.
+                    let position = 0u32;
                     by_txid
                         .entry(record.txid)
-                        .or_insert_with(|| (height, record.transaction.clone()));
+                        .or_insert_with(|| (height, position, record.transaction.clone()));
```
thepastaclaw, line 883: **🔴 Blocking: Preserve block order when aggregating provider updates**

Provider records are deduplicated in a txid-keyed BTreeMap and every live record receives `position = 0`. The stable `(height, position)` aggregation sort therefore leaves same-block ties in txid order rather than Dash Core's `block.vtx` order. Multiple same-block ProReg, ProUpServ, or ProUpReg transactions can consequently leave an earlier service address, operator key, voting key, payout script, or platform node ID as current. Thread the actual in-block transaction index from block processing into the retained transaction context.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The latest delta fixes the platform-node byte-order blocker by pinning rust-dashcore 7221d866, whose consensus boundary converts wire-order IDs to canonical Tenderdash order. One carried-forward blocker remains: production records still lack in-block positions, so same-block provider updates resolve in txid order instead of block order. No new latest-delta finding was found; formatting, diff checks, 155 FFI tests, 421 wallet tests, and 12 Drive decoder tests passed.

### Review provenance

🔴 1 blocking

<details>

```

- [BLOCKING] packages/rs-platform-wallet/src/manager/accessors.rs:870-883: Preserve block order when aggregating provider updates
  Provider records are deduplicated in a txid-keyed BTreeMap and every live record receives `position = 0`. The stable `(height, position)` aggregation sort therefore leaves same-block ties in txid order rather than Dash Core's `block.vtx` order. Multiple same-block ProReg, ProUpServ, or ProUpReg transactions can consequently leave an earlier service address, operator key, voting key, payout script, or platform node ID as current. Thread the actual in-block transaction index from block processing into the retained transaction context.
```
</details>

packages/rs-platform-wallet/src/manager/accessors.rs
QuantumExplorer: Agreed — this cannot be fixed platform-side: the in-block index does not exist on retained records (`BlockInfo` is height/hash/timestamp only; the index is discarded at `key-wallet-manager/src/process_block.rs:60`). Filed [rust-dashcore#890](https://github.com/dashpay/rust-dashcore/issues/890) and authored the fix in [rust-dashcore#891](https://github.com/dashpay/rust-dashcore/pull/891): `BlockInfo` gains an optional `position` stamped from `block.txdata` enumeration during block processing (chainlock promotion already copies the whole `BlockInfo`, so it survives context transitions). Once #891 merges I'll pin-bump this PR and replace the `position = 0` placeholder with `block_info.position()` — the aggregation ordering and its test in this PR are already position-driven. Leaving this thread open until that lands.

QuantumExplorer: Fixed in 5802824073: rust-dashcore#891 merged (pinned 38c689d9) — block processing now stamps each retained record's `BlockInfo` with its `block.vtx` index, the live aggregation feed reads `block_info.position()` instead of the previous 0 placeholder, and the position round-trips persistence (`TransactionRecordFFI`/`PersistentTransaction`/`ProviderSpecialTxRestoreEntryFFI`) so restored wallets keep Core's same-block apply order; pre-field rows restore as `None` and are covered by a round-trip test.

QuantumExplorer:
Reviewed

PR #4126, opened 2026-07-14
feat(platform-wallet): persist Orchard viewing keys for seedless shielded bind

QuantumExplorer:

At every app launch, `rebindWalletScopedServices` shielded-bound every loaded wallet through the mnemonic resolver: the mnemonic was pulled from the Keychain, the seed derived, and a resident Orchard `SpendAuthorizingKey` stayed in memory for the whole session — re-firing on network switch and Sync Now. The mnemonic should not be touched at launch for shielded sync, and no spend authority should be resident between spends.

## What was done?

**rs-platform-wallet**
- Split the bind into a viewing bind vs on-demand spend authorization. `PlatformWallet.shielded_keys` now holds `AccountViewingKeys` (FVK / IVK / OVK / default address) — everything trial-decryption and sync need, nothing that can spend.
- The first seed-backed `bind_shielded` persists each account's raw 96-byte FVK through the persister (`ShieldedChangeSet.viewing_keys`). `AccountViewingKeys::from_fvk_bytes` reconstructs the whole viewing set from those bytes (IVK / OVK / default address are pure functions of the FVK).
- New `bind_shielded_from_persisted` rebinds from those rows with no seed parameter at all; a corrupt row is an error, a missing row means "fall back to the seed path".
- The four spend entry points (`transfer` / `unshield` / `withdraw` / `identity_create_from_pool`) take a `seed` and re-derive the full `OrchardKeySet` per operation via `derive_spend_keyset`, which also guards derived-FVK == bound-FVK so a wrong seed fails before the ~30 s Halo 2 proof. The keyset is dropped (evicted) when the call returns. `shield` and fund-from-asset-lock need only viewing-grade material and take none.

**rs-platform-wallet-ffi**
- `platform_wallet_manager_bind_shielded` keeps its ABI but tries the persisted-keys path first — the resolver fires only when a requested account has no persisted row (first bind after create/import). Launch, network switch, and Sync Now all ride this and never resolve the mnemonic once rows exist.
- New `ShieldedViewingKeyFFI` / `ShieldedViewingKeyRestoreFFI` + `on_persist_shielded_viewing_keys_fn` / `on_load_shielded_viewing_keys_fn`(+free) callback pair, modeled on the platform-node pubkey batch pattern.
- The four spend FFIs gained a `MnemonicResolverHandle` parameter (reusing `resolve_seed_from_resolver`; seed lives in `Zeroizing` buffers), mirroring the transparent `MnemonicResolverCoreSigner` flow.

**swift-sdk / SwiftExampleApp**
- `PersistentShieldedViewingKey` SwiftData model (viewing-grade key material stored like the other derived-key batches, not Keychain) + persistence-handler arms; wallet deletion and shielded Clear purge the rows (Clear so a corrupt row can't outlive it — the next bind re-resolves once and re-persists).
- Spend wrappers/call sites pass a per-operation `MnemonicResolver`; a secret-free `NSLog` audit line fires on every resolver invocation so the no-mnemonic-at-launch property is checkable from the unified log.

## How Has This Been Tested?

- `cargo test -p platform-wallet -p platform-wallet-ffi --all-features` — 753 tests pass, including new ones: FVK byte round-trip reproduces identical FVK/IVK/OVK/default address; bind→persist→simulated-restart→`bind_shielded_from_persisted` yields identical addresses and a working coordinator balance read with no seed anywhere (the API takes none); missing rows report `false` with no state change; corrupt rows error.
- On the iPhone 16 Pro simulator (testnet, 5 wallets): first launch after the change fired the resolver once per wallet (legacy-bind fallback) and persisted one 96-byte FVK row per account; after restart the shielded rebind fired **zero** resolver calls (per-wallet firings went 2 → 1, the remaining one being the separate seed-verify startup path) while the shielded balance (0.09987149 DASH), 1,315 synced notes, and activity all displayed and background/manual syncs completed; Sync Now fired zero resolver calls; a 0.001 DASH unshield fired the resolver exactly once and executed on-chain (platform balance +0.001, pool −0.001 − 0.00168934 fee). A full store-wipe recovery also exercised the first-bind fallback end-to-end (rows re-persisted).

## Breaking Changes

None on-chain / consensus. The four shielded spend FFI signatures gained a resolver parameter (pre-release SDK surface, Swift call sites updated in this PR).

## Checklist:

### DIFF
--- a/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerShieldedSync.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerShieldedSync.swift
@@ -154,21 +154,26 @@ extension PlatformWalletManager {
         currentShieldedTreeTotal = total
     }
 
-    /// Derive Orchard keys for `walletId` from the host-side mnemonic
-    /// resolver, open or create the per-network commitment tree at
-    /// `dbPath`, and bind the resulting multi-account shielded
-    /// sub-wallet to the `PlatformWallet`.
+    /// Bind `walletId`'s multi-account shielded sub-wallet to the
+    /// `PlatformWallet` — from viewing keys the persister already
+    /// holds when possible, deriving from the mnemonic only when it
+    /// doesn't.
     ///
-    /// `accounts` is the list of ZIP-32 account indices to derive.
+    /// `accounts` is the list of ZIP-32 account indices to bind.
     /// Pass `[0]` for the single-account default; pass
     /// `[0, 1, …]` to bind multiple accounts up front. Each entry
     /// produces an independent FVK / IVK / OVK / default address;
     /// notes are scoped per-`(walletId, accountIndex)` inside the
     /// store. Must be non-empty and at most 64 entries.
     ///
-    /// The resolver is fired exactly once. The mnemonic and the
-    /// derived seed live in zeroized buffers on the Rust side and
-    /// are scrubbed before this call returns.
+    /// The resolver does NOT fire on the common path: when every
+    /// requested account has a persisted viewing key (written by the
+    /// first seed-backed bind into `PersistentShieldedViewingKey`
+    /// rows), Rust rebinds from those rows and the mnemonic is never
+    /// read. Only when a row is missing (first bind after create /
+    /// import) does the resolver fire — exactly once, with the
+    /// mnemonic and derived seed in zeroized Rust-side buffers,
+    /// scrubbed before this call returns.
     ///
     /// Idempotent: calling again replaces the previously-bound
     /// shielded wallet.
@@ -492,6 +497,12 @@ extension PlatformWalletManager {
     /// Rust rejects it. The 36-byte on-chain encoding is done on the
     /// Rust side.
     ///
+    /// `resolver` supplies the Orchard spend authority for this one
+    /// operation: Rust resolves the mnemonic, derives the spend key,
+    /// signs, and scrubs everything before returning — no spend key
+    /// stays resident between spends (the launch-time bind is
+    /// viewing-grade only).
+    ///
     /// Throws `PlatformWalletError.shieldedSpendUnconfirmed` when the
     /// broadcast was accepted but its execution result couldn't be
     /// confirmed — the spend may already be on chain, so the caller
@@ -499,6 +510,7 @@ extension PlatformWalletManager {
     /// next shielded sync reconciles them).
     public func shieldedTransfer(
         walletId: Data,
+        resolver: MnemonicResolver,
         account: UInt32 = 0,
         recipientRaw43: Data,
         amount: UInt64,
@@ -519,9 +531,19 @@ extension PlatformWalletManager {
                 "recipient must be exactly 43 raw Orchard bytes"
             )
         }
+        guard let resolverHandle = resolver.handle else {
+            throw PlatformWalletError.invalidParameter(
+                "MnemonicResolver has no handle"
+            )
+        }
 
         let handle = self.handle
         try await Task.detached(priority: .userInitiated) {
+            // Keepalive — the trampoline ctx pointer inside the
+            // resolver dangles unless the Swift owner outlives this
+            // detached work.
+            _ = resolver
+
             try walletId.withUnsafeBytes { widRaw in
                 guard let widPtr = widRaw.baseAddress?.assumingMemoryBound(to: UInt8.self)
                 else {
@@ -540,7 +562,8 @@ extension PlatformWalletManager {
                     // 32-byte limit and does the 36-byte encoding.
                     let send: (UnsafePointer<CChar>?) throws -> Void = { memoCStr in
                         try platform_wallet_manager_shielded_transfer(
-                            handle, widPtr, account, recipientPtr, amount, memoCStr
+                            handle, widPtr, resolverHandle, account, recipientPtr, amount,
+                            memoCStr
                         ).check()
                     }
                     if let memo, !memo.isEmpty {
@@ -621,6 +644,9 @@ extension PlatformWalletManager {
     /// parses and network-checks the address; hosts don't have to
     /// hand-roll the bincode storage variant tag.
     ///
+    /// `resolver` supplies the per-operation Orchard spend authority
+    /// (see [`shieldedTransfer`]).
+    ///
     /// Throws `PlatformWalletError.shieldedSpendUnconfirmed` when the
     /// broadcast was accepted but its execution result couldn't be
     /// confirmed — the spend may already be on chain, so the caller
@@ -628,6 +654,7 @@ extension PlatformWalletManager {
     /// next shielded sync reconciles them).
     public func shieldedUnshield(
         walletId: Data,
+        resolver: MnemonicResolver,
         account: UInt32 = 0,
         toPlatformAddress: String,
         amount: UInt64
@@ -647,9 +674,17 @@ extension PlatformWalletManager {
                 "toPlatformAddress is empty"
             )
         }
+        guard let resolverHandle = resolver.handle else {
+            throw PlatformWalletError.invalidParameter(
+                "MnemonicResolver has no handle"
+            )
+        }
 
         let handle = self.handle
         try await Task.detached(priority: .userInitiated) {
+            // Keepalive — same rationale as `shieldedTransfer`.
+            _ = resolver
+
             try walletId.withUnsafeBytes { widRaw in
                 guard let widPtr = widRaw.baseAddress?.assumingMemoryBound(to: UInt8.self)
                 else {
@@ -657,7 +692,7 @@ extension PlatformWalletManager {
                 }
                 try toPlatformAddress.withCString { addrCStr in
                     try platform_wallet_manager_shielded_unshield(
-                        handle, widPtr, account, addrCStr, amount
+                        handle, widPtr, resolverHandle, account, addrCStr, amount
                     ).check()
                 }
             }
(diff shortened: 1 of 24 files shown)


thepastaclaw:
⛔ Blockers found — Sonnet deferred (commit aca1d3c2)

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerShieldedSync.swift
```diff
@@ -519,9 +531,19 @@ extension PlatformWalletManager {
...
+                "MnemonicResolver has no handle"
+            )
+        }
 
         let handle = self.handle
         try await Task.detached(priority: .userInitiated) {
+            // Keepalive — the trampoline ctx pointer inside the
+            // resolver dangles unless the Swift owner outlives this
+            // detached work.
+            _ = resolver
```
thepastaclaw: **🔴 Blocking: Keep the mnemonic resolver alive through each spend FFI call**

MnemonicResolver passes Rust an Unmanaged.passUnretained(self) callback context and destroys its native handle in deinit. Here, `_ = resolver` is the last semantic use before the FFI operation, so optimized ARC may release the captured owner before Rust calls resolve_seed_from_resolver, invalidating both resolverHandle and its callback context. This is concrete for the temporary resolvers constructed by SendViewModel and is repeated in shieldedUnshield at lines 684-686 and shieldedWithdraw at lines 740-742. The identity-create sibling already uses withExtendedLifetime because a bare `_ =` is not a guaranteed keepalive. Wrap each complete marshalling and FFI operation in withExtendedLifetime(resolver).

<sub>source: ['codex']</sub>

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -3494,6 +3536,80 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
...
+                capacity: rows.count
+            )
+            allocation.entries = buf
+            allocation.entriesCount = rows.count
+            // Same `written`-counter pattern as `loadShieldedSyncStates`:
+            // skip malformed rows without leaving holes in the
+            // contiguous prefix Rust will read.
+            var written = 0
+            for row in rows {
+                guard row.walletId.count == 32, row.fvkBytes.count == 96 else { continue }
```
thepastaclaw: **🔴 Blocking: Surface malformed persisted viewing keys instead of treating them as absent**

The loader silently omits any persisted row whose wallet ID or FVK has an invalid length and reports success with only the remaining rows. Rust therefore sees a requested account as missing, bind_shielded_from_persisted returns Ok(false), and the FFI bind path resolves the mnemonic and re-persists the key. This violates the documented fail-closed behavior for present-but-malformed viewing keys and bypasses the Rust corruption test because that test supplies the malformed row directly. Validate rows before any filtering or marshalling step can discard malformed data and before allocating the FFI buffer; set resultErrored and return on corruption. The previously proposed narrow replacement is unsafe because it returns after allocation without releasing the buffer, so no exact-range suggestion is supplied.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

Both blocking candidates are verified at the exact head. Three Swift spend wrappers can release their unretained mnemonic resolver before Rust invokes it, and the Swift persistence loader converts malformed viewing-key rows into apparent misses, triggering mnemonic fallback instead of surfacing corruption.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (completed)

🔴 2 blocking

<details>

```

In `packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerShieldedSync.swift`:
- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerShieldedSync.swift:541-545: Keep the mnemonic resolver alive through each spend FFI call
  MnemonicResolver passes Rust an Unmanaged.passUnretained(self) callback context and destroys its native handle in deinit. Here, `_ = resolver` is the last semantic use before the FFI operation, so optimized ARC may release the captured owner before Rust calls resolve_seed_from_resolver, invalidating both resolverHandle and its callback context. This is concrete for the temporary resolvers constructed by SendViewModel and is repeated in shieldedUnshield at lines 684-686 and shieldedWithdraw at lines 740-742. The identity-create sibling already uses withExtendedLifetime because a bare `_ =` is not a guaranteed keepalive. Wrap each complete marshalling and FFI operation in withExtendedLifetime(resolver).

- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift:3576-3581: Surface malformed persisted viewing keys instead of treating them as absent
  The loader silently omits any persisted row whose wallet ID or FVK has an invalid length and reports success with only the remaining rows. Rust therefore sees a requested account as missing, bind_shielded_from_persisted returns Ok(false), and the FFI bind path resolves the mnemonic and re-persists the key. This violates the documented fail-closed behavior for present-but-malformed viewing keys and bypasses the Rust corruption test because that test supplies the malformed row directly. Validate rows before any filtering or marshalling step can discard malformed data and before allocating the FFI buffer; set resultErrored and return on corruption. The previously proposed narrow replacement is unsafe because it returns after allocation without releasing the buffer, so no exact-range suggestion is supplied.
```
</details>

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerShieldedSync.swift
QuantumExplorer: Fixed in aca1d3c2d1 — all three spend wrappers (transfer/unshield/withdraw) now wrap the full marshalling + FFI call in `withExtendedLifetime(resolver)`, matching the identity-create sibling.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
QuantumExplorer: Fixed in aca1d3c2d1 — `loadShieldedViewingKeys` now validates every row up front and fails the load (`resultErrored = true`, before any buffer allocation) on a present-but-malformed walletId/FVK length, so corruption surfaces as a bind error instead of a silent mnemonic fallback.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -3494,6 +3536,94 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
...
+                $0.walletId.count != 32 || $0.fvkBytes.count != 96
+            }) {
+                SDKLogger.error(
+                    "loadShieldedViewingKeys: corrupt row "
+                        + "(walletId \(bad.walletId.count)B, fvk \(bad.fvkBytes.count)B) — "
+                        + "failing the load rather than masking it as a missing key"
+                )
+                resultErrored = true
+                return
+            }
```
thepastaclaw, line 3587: **🔴 Blocking: Validate malformed wallet IDs before network filtering**

`PersistentShieldedViewingKey.walletId` is unconstrained `Data`. A corrupted non-32-byte wallet ID cannot match any normal ID returned by `inNetworkWalletIds()`, so lines3564-3565 discard the row before the length check runs. If no rows remain, the callback reports a successful empty load; Rust then treats the requested key as missing and invokes the mnemonic fallback, violating this PR's documented requirement that persisted corruption fail closed. Validate wallet-ID lengths before network filtering, then validate FVK lengths after filtering so a corrupt FVK belonging to another network does not block this manager.

```suggestion
if let bad = rows.first(where: { $0.walletId.count !=32 }) {
 SDKLogger.error(
 "loadShieldedViewingKeys: corrupt wallet id "
 + "(\(bad.walletId.count)B) — failing the load rather than masking it"
 )
 resultErrored = true
 return
}
if let inNetworkIds = self.inNetworkWalletIds() {
 rows = rows.filter { inNetworkIds.contains($0.walletId) }
}
if rows.isEmpty {
 return
}
if let bad = rows.first(where: { $0.fvkBytes.count !=96 }) {
 SDKLogger.error(
 "loadShieldedViewingKeys: corrupt FVK "
 + "(\(bad.fvkBytes.count)B) — failing the load rather than masking it"
 )
 resultErrored = true
 return
}
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The resolver-lifetime blocker is fixed at exact head aca1d3c2d166742ee8b557296f3ea8e329abfcea. One blocking persistence issue remains: malformed wallet IDs are filtered out before validation, converting corruption into a successful missing-key result and mnemonic fallback. This is a preliminary Codex-only verification; Sonnet was not run.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (completed)

🔴 1 blocking

<details>

```

- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift:3564-3587: Validate malformed wallet IDs before network filtering
  `PersistentShieldedViewingKey.walletId` is unconstrained `Data`. A corrupted non-32-byte wallet ID cannot match any normal ID returned by `inNetworkWalletIds()`, so lines3564-3565 discard the row before the length check runs. If no rows remain, the callback reports a successful empty load; Rust then treats the requested key as missing and invokes the mnemonic fallback, violating this PR's documented requirement that persisted corruption fail closed. Validate wallet-ID lengths before network filtering, then validate FVK lengths after filtering so a corrupt FVK belonging to another network does not block this manager.
```
</details>

PR #4127, opened 2026-07-14
feat(platform-wallet): persist typed BLS/EdDSA provider keys as core address rows

QuantumExplorer:
## Issue being fixed or feature implemented

`CoreAddressEntryFFI` carried a fixed 33-byte ECDSA-only `public_key` slot, so non-secp256k1 provider keys were dropped on the SwiftData persist round-trip:
- **Provider Operator Keys (BLS, 48 bytes)** — address rows persisted with an empty `publicKey`, kept alive in-memory only by a merge workaround in `restore_address_pool`.
- **Provider Platform Node Keys (Ed25519, 32 bytes)** — persisted **zero** address rows; the keys traveled through a dedicated account-level `derivedPlatformNodeKeys` batch (`ProviderPlatformNodeKeyFFI` on `AccountSpecFFI`) with its own restore path (`restore_platform_node_pool`).

## What was done?

**Rust / FFI**
- Widened `CoreAddressEntryFFI` to a typed key: `public_key: [u8; 48]` + `public_key_len` (0 = none / 33 ECDSA / 48 BLS / 32 EdDSA) + `key_type_tag` (new `KeyTypeTagFFI` enum). `build_core_address_entry_ffi` marshals all three `PublicKeyType` variants with per-curve length validation; `address_info_from_ffi` rebuilds the typed key, forgiving on invalid tag/len pairs.
- `register_wallet` now populates the managed platform-node pool from the pre-derived batch (new `populate_platform_node_pool` in `rs-platform-wallet`) **before** the address-pool snapshot, so EdDSA keys persist as ordinary typed address rows and the in-memory pool matches what a restore reconstructs.
- Retired the batch plumbing: `ProviderKeyAccountEntry.derived_platform_node_keys`, `ProviderPlatformNodeKeyFFI`, the `AccountSpecFFI` ptr/len pair, and `restore_platform_node_pool` are gone (`AccountAddressPoolFFI` layout guard updated 136 → 120).
- Removed the typed-key merge workaround in `restore_address_pool` — the row itself is now authoritative for its typed key.
- Added a thin FFI bridge `platform_wallet_platform_node_id_from_ed25519_pubkey` (wraps `dashcore::PlatformNodeId::from_ed25519_public_key`) so the host can render node ids from persisted pubkeys.

**Swift**
[...]

PR #893, opened 2023-04-07
fix(dashmate): api binds to all interfaces

pshenmic:
## Issue being fixed or feature implemented

Bind dashmate to localhost for security reason.

## What was done?

* bind dashmate rpc api to 127.0.0.1
* add curl to dashmate docker image (used in smoke tests)

## How Has This Been Tested?

Locally

## Breaking Changes

No

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added or updated relevant unit/integration/functional/e2e tests
- [ ] I have made corresponding changes to the documentation

**For repository code-owners and collaborators only**
- [ ] I have assigned this pull request to a milestone

PR #4129, opened 2026-07-14
fix(swift-sdk): stored default on keyType so pre-column stores can migrate

QuantumExplorer:

#4127 added `PersistentCoreAddress.keyType: UInt8` with only an init-parameter default, which SwiftData lightweight migration never consults. Opening any pre-#4127 store fails migration ("Validation error missing attribute values on mandatory destination attribute: PersistentCoreAddress.keyType"), `DashModelContainer.create()` throws, and `SwiftExampleAppApp` hits `fatalError` — the app crash-loops on every device that has existing data.

## What was done?

Stored default `= 0` (ECDSA) on the property, matching the established convention for post-hoc columns (`isImported`, `statusRaw`, `blockPosition`). Defaulted rows read as ECDSA until the next Rust address-pool persist pulse re-tags typed entries, which happens on every load — so provider BLS/EdDSA rows self-heal immediately.

## How Has This Been Tested?

Reproduced the crash on the booted simulator: pre-#4127 store + v4.1-dev tip → container load error + crash to home screen. With this fix the same store migrates, the container loads, and the app launches.

## Breaking Changes

None.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility when opening existing local data stores after an update.
  * Prevented launch crashes by providing a default address key type for legacy records.
  * Existing records are interpreted consistently until their key type is refreshed.

### DIFF
--- a/packages/swift-sdk/Sources/SwiftDashSDK/Persistence/Models/PersistentCoreAddress.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/Persistence/Models/PersistentCoreAddress.swift
@@ -23,8 +23,15 @@ public final class PersistentCoreAddress {
     public var publicKey: Data
     /// `KeyTypeTagFFI` raw value identifying the curve of `publicKey`:
     /// 0 ECDSA / 1 BLS / 2 EdDSA. Meaningful only when `publicKey` is
-    /// non-empty.
-    public var keyType: UInt8
+    /// non-empty. The stored default (NOT just the init-parameter
+    /// default, which SwiftData migration never consults) keeps
+    /// pre-column stores openable: without it, lightweight migration
+    /// fails with "missing attribute values on mandatory destination
+    /// attribute" and the container refuses to load — a launch crash on
+    /// every device that has existing rows. Defaulted rows read as
+    /// ECDSA until the next Rust address-pool persist pulse re-tags
+    /// typed entries, which it does on every load.
+    public var keyType: UInt8 = 0
     /// `AddressPoolTypeTagFFI` raw value — 0 External, 1 Internal,
     /// 2 Absent, 3 AbsentHardened.
     public var poolTypeTag: UInt8


thepastaclaw:
⛔ Blockers found — Sonnet deferred (commit 6177500d)

packages/swift-sdk/Sources/SwiftDashSDK/Persistence/Models/PersistentCoreAddress.swift
```diff
@@ -23,8 +23,15 @@ public final class PersistentCoreAddress {
...
-    public var keyType: UInt8
+    /// non-empty. The stored default (NOT just the init-parameter
+    /// default, which SwiftData migration never consults) keeps
+    /// pre-column stores openable: without it, lightweight migration
+    /// fails with "missing attribute values on mandatory destination
+    /// attribute" and the container refuses to load — a launch crash on
+    /// every device that has existing rows. Defaulted rows read as
+    /// ECDSA until the next Rust address-pool persist pulse re-tags
+    /// typed entries, which it does on every load.
+    public var keyType: UInt8 = 0
```
thepastaclaw, line 34: **🔴 Blocking: Legacy provider keys are not re-tagged on load**

Pre-#4127 address rows serialized only ECDSA keys, so BLS and Ed25519 rows have an empty `publicKey`; the Ed25519 batch was stored separately in the since-removed `PersistentAccount.derivedPlatformNodeKeys`. After this default allows the store to open, `buildCoreAddressPoolBuffer` forwards those empty keys with length zero, Rust decodes them as `public_key: None`, and `restore_address_pool` overwrites the managed-pool entry. `FFIPersister::load` only reads the snapshot and never invokes the persistence callback, so the claimed load-time re-tagging does not occur. BLS ownership matching can still derive keys from its xpub, but its typed persisted entries remain empty; hardened Ed25519 children cannot be derived from the public xpub, leaving platform-node ownership matching and persisted key display unavailable. Add an explicit legacy conversion that preserves pre-derived typed entries and migrates the old Ed25519 batch into typed core-address rows, with an on-disk pre-#4127 migration test.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

Exact HEAD 6177500da9a7e170b4aa67bcbd263eeccb0e7ec8 and the supplied full range were verified. The stored default makes legacy stores openable, but it does not perform the claimed provider-key migration, leaving legacy Ed25519 platform-node ownership data unavailable after restoration.

### Review provenance

🔴 1 blocking

<details>

```

In `packages/swift-sdk/Sources/SwiftDashSDK/Persistence/Models/PersistentCoreAddress.swift`:
- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/Persistence/Models/PersistentCoreAddress.swift:34: Legacy provider keys are not re-tagged on load
  Pre-#4127 address rows serialized only ECDSA keys, so BLS and Ed25519 rows have an empty `publicKey`; the Ed25519 batch was stored separately in the since-removed `PersistentAccount.derivedPlatformNodeKeys`. After this default allows the store to open, `buildCoreAddressPoolBuffer` forwards those empty keys with length zero, Rust decodes them as `public_key: None`, and `restore_address_pool` overwrites the managed-pool entry. `FFIPersister::load` only reads the snapshot and never invokes the persistence callback, so the claimed load-time re-tagging does not occur. BLS ownership matching can still derive keys from its xpub, but its typed persisted entries remain empty; hardened Ed25519 children cannot be derived from the public xpub, leaving platform-node ownership matching and persisted key display unavailable. Add an explicit legacy conversion that preserves pre-derived typed entries and migrates the old Ed25519 batch into typed core-address rows, with an on-disk pre-#4127 migration test.
```
</details>

QuantumExplorer:
Scope note: this branch now carries a second commit pinning rust-dashcore to the head of [rust-dashcore#893](https://github.com/dashpay/rust-dashcore/pull/893) — the fix for the OTHER startup crash on v4.1-dev ([rust-dashcore#892](https://github.com/dashpay/rust-dashcore/issues/892), a dash-spv debug SIGABRT crash-loop when a wallet's scan floor sits below the stored filter range). Verified live on the previously crash-looping simulator: the new discard-and-redownload path fired and the app runs cleanly. The pin will be re-pointed to the #893 merge commit once it lands; together with the keyType migration default, this makes v4.1-dev launchable again on devices with existing data.

QuantumExplorer:
rust-dashcore#893 merged — pin re-pointed to the merge commit (19690d31) in 3ad874049a. This PR is now merge-ready: it carries the two fixes that make v4.1-dev launchable on devices with existing data (keyType migration default + the sparse-filter crash-loop fix), both verified live on-device.

QuantumExplorer:
Follow-up for the [post-merge blocking finding](https://github.com/dashpay/platform/pull/4129#pullrequestreview-4697392092): **[#4132](https://github.com/dashpay/platform/pull/4132)**. Confirmed the finding — the plain upsert did strip prederived typed BLS keys for legacy key-less rows, and the "re-tags on every load" claim in this PR's comment was wrong (load only reads the snapshot). #4132 restores typed-key preservation for legacy rows (no-op for post-#4127 rows), corrects the doc, and also lands the rust-dashcore#893 crash-fix pin that was pushed to this branch after the merge. Legacy Ed25519 platform-node rows cannot be recovered at load (hardened-only; the removed batch column's data was already dropped by the schema migration) — per the pre-release convention they re-derive on delete+re-import.

packages/swift-sdk/Sources/SwiftDashSDK/Persistence/Models/PersistentCoreAddress.swift
QuantumExplorer: Addressed in #4132 (merged): `restore_address_pool` now keeps the pre-derived typed key when a legacy key-less row arrives, so in-memory BLS operator-ownership matching survives pre-#4127 stores, with a regression test covering the legacy-row overwrite. The Ed25519 half of the ask is not implementable as stated: the removed `derivedPlatformNodeKeys` batch column's data is dropped by the SwiftData schema migration at first open, so there is nothing on disk left to convert — legacy platform-node rows follow the pre-release delete+re-import convention (hardened-only, no public re-derivation), now documented accurately on `keyType` instead of the previous over-claim about load-time re-tagging.

PR #4132, opened 2026-07-15
fix(platform-wallet): keep prederived typed keys for legacy key-less rows + land the #893 crash-fix pin

QuantumExplorer:

Two follow-ups that missed the #4129 merge window:

1. **thepastaclaw's post-merge blocking finding on #4129** ([review](https://github.com/dashpay/platform/pull/4129#pullrequestreview-4697392092)): the keyType stored default makes legacy stores openable, but legacy rows (persisted before #4127's typed-key column) decode with `public_key: None` and `restore_address_pool`'s plain upsert then STRIPS the typed BLS key the gap-limit prederivation placed at the same index — silently breaking in-memory operator-ownership matching for pre-#4127 stores. The claimed "re-tags on every load" in the #4129 comment was also wrong (load only reads the snapshot; re-tag happens on the next persist pulse).

2. **The rust-dashcore #893 pin never landed on v4.1-dev**: #4129 was merged minutes after opening, and the crash-fix pin commits were pushed to its branch afterwards. v4.1-dev still pins 38c689d9, leaving the dash-spv sparse-filter startup crash-loop (rust-dashcore#892) unfixed on mainline.

## What was done?

- `restore_address_pool`: when an incoming row carries no key but the prederived entry at that index does, keep the existing typed key (no-op for post-#4127 rows, which always carry their key). Doc comment spells out why, and why legacy Ed25519 rows cannot be recovered this way (hardened-only; the old account-level batch's data was already dropped by the schema migration) — per the pre-release convention those stores re-derive on delete+re-import.
- New test `legacy_keyless_row_keeps_prederived_typed_key` pins the preservation and the inverse (a legacy row with no prederived counterpart stays key-less).
- Corrected the `keyType` doc comment to reflect actual re-tag timing.
- Bumped rust-dashcore to 19690d31 (the merged rust-dashcore#893, fixes rust-dashcore#892) — verified live earlier on the previously crash-looping simulator state (one-time discard + re-download, clean restarts since).

## How Has This Been Tested?

`cargo test -p platform-wallet -p platform-wallet-ffi` — 468 + FFI suites, 0 failures, including the new legacy-preservation test. `cargo check --all-targets` clean. The pinned rust-dashcore rev is content-identical to the PR-head pin (828f707d) already exercised on both simulators.

## Breaking Changes

None.

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1206,7 +1206,7 @@ version = "3.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "faf9468729b8cbcea668e36183cb69d317348c2e08e994829fb56ebfdfbaac34"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


thepastaclaw:
🔍 Review in progress — actively reviewing now (commit b1866bf6)

Issue #817, opened 2023-03-11
bug(dashmate): dashmate fails to setup regular testnet node with funding private key


pshenmic:
## Expected Behavior

Setup operation completes successfully when funding key is provided, f.e `dashmate setup -p <funding_private_key>`

## Current Behavior

It fails to setup because it tries to run masternode in wallet mode.

✔ Set configuration preset
✔ Set node type
  › Selected masternode type
✔ Configure external IP address
  › x.x.x.x is set
✔ Set masternode operator private key
  › BLS public key: 8753c2a96e1a653ef4b9fd3a5599b3699aa85bc8fd551381cfa6adcb959573e759724d6e458fd5dcc0e4ccfd64e18dd4
    BLS private key: 644c673bf2c9ea3f9f1b8a42ad039ca10196c25c6766237ac089ac91ae2522bf
❯ Register masternode
  ✖ You cannot run masternode with wallet mode on
  ◼ Import funding private key
  ◼ Sync Core with network
  ◼ Check funding address balance
  ◼ Create a new collateral address
  ◼ Create a new owner addresses
  ◼ Create a new reward addresses
  ◼ Send 0.1 dash from funding address to reward address
  ◼ Wait for balance to confirm
  ◼ Send 1000 dash from funding address to collateral address
  ◼ Wait for 1 confirmation
  ◼ Broadcast masternode registration transaction
  ◼ Wait for 1 confirmation
  ◼ Stop Core

## Possible Solution

This is related to `startCoreFactory` refactor I believe, where we moved command line arguments to the config. As a solution, we can empty masternode operator private key in the config during `registerMasternodeTask`

## Your Environment

* Version used: v0.23.2
* Environment name and version (e.g. Chrome 39, node.js 5.4): MacOS Ventura (arm64)

pshenmic:
This mode no more being supported

PR #4139, opened 2026-07-16
feat(swift-sdk): partial-amount platform address withdrawal wrapper

QuantumExplorer:
## Summary

Adds a `withdraw(accountIndex:coreAddress:inputs:signer:)` overload to `ManagedPlatformAddressWallet` using `INPUT_SELECTION_TYPE_EXPLICIT`: each `WithdrawalInput` names an address and the exact credits to withdraw, so the payout equals the requested amount (the transition fee is deducted from the fee-source input's remaining balance, per the default `DeductFromInput(0)` strategy).

Complements the existing AUTO overload, which always withdraws the full non-dust balance. Swift-only — the FFI entry point already ships in `DashSDKFFI`.

## Why now

dashwallet-ios has required this API since its partial Platform → Transparent transfer landed (dashwallet-ios #817, and the Send redesign #819 builds on it) — building the app's `swift-sdk-integration` branch against `v4.1-dev` without this commit fails with missing-member errors on `withdraw(…inputs:)` / `WithdrawalInput`. This upstreams the local commit so plain `v4.1-dev` builds the app again.

## Test plan

- Swift-only wrapper over the shipped FFI; exercised end-to-end by dashwallet-ios partial Platform → Transparent withdrawals on testnet (payout equals the requested amount, fee deducted from the source address's remaining balance).

## Summary by CodeRabbit

- **New Features**
  - Added explicit-input withdrawals for managed platform address wallets.
  - Added support for specifying withdrawal inputs, destination address, and fee rate.
  - Added validation for destination addresses and withdrawal input details.
  - Existing automatic input selection withdrawals remain available.

### DIFF
--- a/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/ManagedPlatformAddressWallet.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/ManagedPlatformAddressWallet.swift
@@ -497,6 +497,130 @@ public final class ManagedPlatformAddressWallet: @unchecked Sendable {
         }.value
     }
 
+    /// Explicit withdrawal input for the partial-amount `withdraw` overload:
+    /// withdraw exactly `credits` from this address. The transition fee is
+    /// deducted by the chain from the address's REMAINING on-chain balance
+    /// (`balance − credits`), so the fee-source input (index 0 under the
+    /// default `DeductFromInput(0)` strategy) must keep at least the fee
+    /// after the withdrawal — an exact-balance input is rejected with
+    /// `fee_fully_covered = false`.
+    public struct WithdrawalInput: Sendable {
+        /// Must be `0` (P2PKH) — same Rust `TryFrom<PlatformAddressFFI>`
+        /// constraint as transfer inputs.
+        public let addressType: UInt8
+        /// 20-byte address hash.
+        public let hash: Data
+        /// Credits to withdraw from this address (its payout contribution).
+        public let credits: UInt64
+
+        public init(addressType: UInt8, hash: Data, credits: UInt64) {
+            self.addressType = addressType
+            self.hash = hash
+            self.credits = credits
+        }
+    }
+
+    /// Partial-amount withdrawal: pay out exactly `Σ inputs.credits` to
+    /// `coreAddress`, drawing the listed amounts from the listed addresses
+    /// (`INPUT_SELECTION_TYPE_EXPLICIT`). Unlike the AUTO overload above,
+    /// the payout equals the requested amount — the transition fee comes out
+    /// of the fee-source input's remaining balance, not the payout. The Rust
+    /// side re-fetches and hard-checks every input's on-chain balance before
+    /// signing, so a stale caller-side balance fails with a typed error
+    /// rather than a wrong-amount withdrawal.
+    ///
+    /// Same `coreAddress` parsing/network rules and signer requirements as
+    /// the AUTO overload.
+    @discardableResult
+    public func withdraw(
+        accountIndex: UInt32,
+        coreAddress: String,
+        inputs: [WithdrawalInput],
+        coreFeePerByte: UInt32 = 1,
+        signer: KeychainSigner
+    ) async throws -> [UpdatedBalance] {
+        let trimmed = coreAddress.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard !trimmed.isEmpty else {
+            throw PlatformWalletError.invalidParameter("coreAddress is empty")
+        }
+        // Same embedded-NUL guard as the AUTO overload: Rust's
+        // `CStr::from_ptr(...).to_str()` stops at the first NUL, so a
+        // `valid\0suffix` value would silently withdraw to `valid`.
+        guard !trimmed.utf8.contains(0) else {
+            throw PlatformWalletError.invalidParameter(
+                "coreAddress contains an embedded NUL byte"
+            )
+        }
+        guard !inputs.isEmpty else {
+            throw PlatformWalletError.invalidParameter("inputs is empty")
+        }
+        for input in inputs {
+            guard input.addressType == 0 else {
+                throw PlatformWalletError.invalidParameter(
+                    "only P2PKH (addressType 0) withdrawal inputs are supported"
+                )
+            }
+            guard input.hash.count == 20 else {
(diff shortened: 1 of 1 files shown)


thepastaclaw:
⛔ Blockers found — Sonnet deferred (commit 5043b867)

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/ManagedPlatformAddressWallet.swift
```diff
@@ -497,6 +497,130 @@ public final class ManagedPlatformAddressWallet: @unchecked Sendable {
...
+                            UInt(inputBp.count),
+                            nil,
+                            0,
+                            addrCStr,
+                            feePerByte,
+                            nil,
+                            0,
+                            signerHandle,
+                            &changeset
+                        )
```
thepastaclaw, line 613: **🔴 Blocking: Explicit withdrawal inputs are not bound to the managed wallet or account**

This managed-wallet overload accepts arbitrary address hashes and forwards them as `InputSelection::Explicit` without checking that they belong to `self` or `accountIndex`. The Rust explicit branch does not use `wallet_id` or `account_index` for authorization; the SDK only fetches the supplied addresses' balances and then asks the signer to sign every map key. `KeychainSigner` resolves a platform key by `addressHash` across the shared `ModelContainer`, takes the matched row's `walletId` and derivation path, and loads that wallet's mnemonic. Because one manager/container holds multiple wallets, code using wallet A's handle can supply a persisted wallet B hash and have the normal signer authorize a withdrawal to an arbitrary Core address. Post-broadcast reconciliation merely skips the foreign address and cannot prevent the spend. Validate every explicit input against this wallet and account before signing, preferably in the Rust wallet/FFI layer, and bind signer resolution to the expected wallet ID.

<sub>source: ['codex']</sub>

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/ManagedPlatformAddressWallet.swift
```diff
@@ -497,6 +497,130 @@ public final class ManagedPlatformAddressWallet: @unchecked Sendable {
...
+                            accountIndex,
+                            INPUT_SELECTION_TYPE_EXPLICIT,
+                            inputBp.baseAddress,
+                            UInt(inputBp.count),
+                            nil,
+                            0,
+                            addrCStr,
+                            feePerByte,
+                            nil,
+                            0,
```
thepastaclaw, line 610: **🔴 Blocking: Default fee strategy targets the canonical first address, not inputs[0]**

The API tells callers to leave fee headroom on the first Swift input, but passing `nil, 0` becomes `DeductFromInput(0)` only after `parse_explicit_inputs` has inserted the array into a `BTreeMap<PlatformAddress, Credits>`. DPP resolves index 0 against that map's derived `Ord` order, so it means the lexicographically smallest address hash rather than `inputs[0]`. If a later, smaller-sorting input is withdrawn in full, fee validation fails because that unintended target has no remainder; if it has headroom, the fee is charged to the wrong address. This is the same insertion-order-versus-canonical-index class previously fixed for transfer outputs. Compute the canonical rank of the caller's intended fee source and pass an explicit `FeeStrategyStepFFI`, with a regression test whose Swift order differs from hash order.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The new partial-withdrawal API has two in-scope blockers. Caller-supplied explicit inputs are not authorized against the managed wallet/account before the shared signer resolves keys, and the documented first-input fee source is reinterpreted using Rust's canonical address order.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (failed), `default` — general (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — security-auditor (failed), `default` — security-auditor (failed), `gpt-5.6-sol` — security-auditor (failed), `gpt-5.6-sol` — security-auditor (completed), `gpt-5.6-sol` — ffi-engineer (failed), `default` — ffi-engineer (failed), `gpt-5.6-sol` — ffi-engineer (failed), `gpt-5.6-sol` — ffi-engineer (completed)

🔴 2 blocking

<details>

```

In `packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/ManagedPlatformAddressWallet.swift`:
- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/ManagedPlatformAddressWallet.swift:583-613: Explicit withdrawal inputs are not bound to the managed wallet or account
  This managed-wallet overload accepts arbitrary address hashes and forwards them as `InputSelection::Explicit` without checking that they belong to `self` or `accountIndex`. The Rust explicit branch does not use `wallet_id` or `account_index` for authorization; the SDK only fetches the supplied addresses' balances and then asks the signer to sign every map key. `KeychainSigner` resolves a platform key by `addressHash` across the shared `ModelContainer`, takes the matched row's `walletId` and derivation path, and loads that wallet's mnemonic. Because one manager/container holds multiple wallets, code using wallet A's handle can supply a persisted wallet B hash and have the normal signer authorize a withdrawal to an arbitrary Core address. Post-broadcast reconciliation merely skips the foreign address and cannot prevent the spend. Validate every explicit input against this wallet and account before signing, preferably in the Rust wallet/FFI layer, and bind signer resolution to the expected wallet ID.
- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/ManagedPlatformAddressWallet.swift:593-610: Default fee strategy targets the canonical first address, not inputs[0]
  The API tells callers to leave fee headroom on the first Swift input, but passing `nil, 0` becomes `DeductFromInput(0)` only after `parse_explicit_inputs` has inserted the array into a `BTreeMap<PlatformAddress, Credits>`. DPP resolves index 0 against that map's derived `Ord` order, so it means the lexicographically smallest address hash rather than `inputs[0]`. If a later, smaller-sorting input is withdrawn in full, fee validation fails because that unintended target has no remainder; if it has headroom, the fee is charged to the wrong address. This is the same insertion-order-versus-canonical-index class previously fixed for transfer outputs. Compute the canonical rank of the caller's intended fee source and pass an explicit `FeeStrategyStepFFI`, with a regression test whose Swift order differs from hash order.
```
</details>

PR #4147, opened 2026-07-20
fix: prevent secret tracing in dependency build stages

QuantumExplorer:
## What

- disable shell command tracing in the Drive and rs-dapi dependency stages that consume an optional GitHub credential
- require a non-empty mounted credential before configuring Git authentication
- remove the temporary Git configuration through an exit trap on both successful and failed builds

## Why

The affected stages expanded a mounted credential while command tracing was enabled. Raw build output could therefore include credential material even though BuildKit kept the mounted file out of image layers.

This covers PLATFORM-DS-CAND-376.

## Checks

- Dockerfile diff and whitespace validation
- synthetic credential output test: marker absent from stdout and stderr
- failure-path cleanup test: temporary Git configuration removed

A Docker build parser check was attempted, but the local Docker daemon is not running.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Docker build reliability by enforcing stricter shell error handling.
  * Updated credential cleanup to run automatically when build steps exit.
  * Prevented empty secret files from triggering Git URL rewriting.
  * Simplified build-cache statistics handling while maintaining cleanup safeguards.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -464,7 +464,7 @@ RUN --mount=type=cache,sharing=shared,id=cargo_registry_index,target=${CARGO_HOM
     --mount=type=secret,id=GITHUB_TOKEN \
     trap 'rm -f "${HOME}/.gitconfig"' EXIT; \
     if [ -s /run/secrets/GITHUB_TOKEN ]; then \
-    git config --global url."https://$(cat /run/secrets/GITHUB_TOKEN)@github.com/".insteadOf "https://github.com/"; \
+    git config --global url."https://x-access-token:$(cat /run/secrets/GITHUB_TOKEN)@github.com/".insteadOf "https://github.com/"; \
     fi && \
     source /root/env && \
     export FEATURES_FLAG=""; \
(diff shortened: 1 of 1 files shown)


thepastaclaw:
🕓 Ready for review — 37 ahead in queue (commit 475f09f4)
_Queue position: 38/72 · 2 reviews active_
_ETA: start ~04:34 UTC · complete ~05:03 UTC (median 28m across 30 recent reviews; 2 slots)_
_Queued 12h 45m ago · Last checked: 2026-07-22 19:40 UTC_

infraclaw-dash (changes requested):
Requesting changes because the credential can still be emitted to build logs on a Git failure.

Lines 467 and 889, plus the existing occurrence at line 645, embed GITHUB_TOKEN as the URL username. Git includes URL usernames in stderr. I reproduced this failure output with a synthetic token:

fatal: could not read Password for https://SYNTHETIC_TOKEN_DO_NOT_USE@github.com

Please use a non-secret username and put the token in the password position, for example https://x-access-token:TOKEN@github.com/, or use a credential helper. Git redacts URL passwords in errors. Please update all three occurrences.

The EXIT trap and removal of xtrace otherwise look correct. The Docker build checks were skipped, so this Dockerfile change has not been exercised by CI.

infraclaw-dash:
I implemented the requested credential redaction fix in #4179. It updates all three URL rewrites and passes the synthetic failure-output check. This account cannot push directly to the PR branch, so #4179 targets codex/fix-docker-build-secret-logging and can be merged into this PR.

PR #4148, opened 2026-07-20
fix(drive): order ascending cursor range bounds

QuantumExplorer:
## What

- construct ascending less-than cursor ranges with the cursor as the lower bound
- preserve startAt versus startAfter inclusivity for both less-than operators
- add a regression matrix covering all four affected range forms

## Why

The previous endpoint order produced reversed GroveDB ranges for ascending queries whose cursor was below the predicate bound. The proof and no-proof paths share this construction, so correcting it restores the query semantics at their common boundary.

This covers PLATFORM-DS-CAND-388.

## Checks

- cargo test -p drive --features server query::conditions::tests (205 passed)
- cargo clippy -p drive --features server --all-targets -- -D warnings
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/packages/rs-drive/src/query/conditions.rs
+++ b/packages/rs-drive/src/query/conditions.rs
@@ -948,9 +948,9 @@ impl<'a> WhereClause {
                         if left_to_right {
                             if starts_at_key < key {
                                 if included {
-                                    query.insert_range(key..starts_at_key);
+                                    query.insert_range(starts_at_key..key);
                                 } else {
-                                    query.insert_range_after_to(key..starts_at_key);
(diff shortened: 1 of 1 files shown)


thepastaclaw:
🔍 Review in progress — actively reviewing now (commit 9f62a35a)
_Stage: Sonnet review + final verification_
_ETA: complete ~15:15 UTC (median 36m across 30 recent reviews)_
_Running 1h 2m · Last checked: 2026-07-22 15:10 UTC_

QuantumExplorer:
reviewed

PR #4156, opened 2026-07-20
fix(drive): preserve typed proof query fields

QuantumExplorer:
## Summary

- preserve checked pagination fields when converting identity-vote proof requests
- decode WASM identity-vote queries into bounded Rust types instead of narrowing generic numeric values
- validate JavaScript key-purpose values before converting them into proof-query discriminants

## Compatibility

This changes only client-side request/proof-query reconstruction. It does not change consensus or require a protocol activation. Malformed or out-of-range request fields now fail before proof verification.

## Validation

- `cargo test -p drive-proof-verifier --lib` (242 passed)
- `cargo test -p wasm-drive-verify --lib` (11 passed)
- strict Clippy for both crates
- native and `wasm32-unknown-unknown` builds
- formatting and diff checks

## Summary by CodeRabbit

* **Bug Fixes**
  * Pagination offsets and limits are now preserved when converting vote queries.
  * Invalid or out-of-range pagination values are rejected instead of being truncated.
  * Identity key purposes now reject fractional, negative, non-finite, and out-of-range values.
  * Vote query decoding now correctly validates malformed optional fields and invalid identifiers.
  * Missing optional query fields now receive the appropriate defaults.

### DIFF
--- a/packages/rs-drive-proof-verifier/src/from_request.rs
+++ b/packages/rs-drive-proof-verifier/src/from_request.rs
@@ -188,6 +188,22 @@ impl TryFromRequest<GetContestedResourceIdentityVotesRequest>
                     .map(|id| (id, v.start_poll_identifier_included))
             })
             .transpose()?;
+        let offset =
+            value
+                .offset
+                .map(u16::try_from)
+                .transpose()
+                .map_err(|_| Error::RequestError {
+                    error: "offset out of bounds".to_string(),
+                })?;
+        let limit =
+            value
+                .limit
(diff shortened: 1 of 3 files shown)


thepastaclaw:
🕓 Ready for review — 17 ahead in queue (commit 61e96e49)
_Queue position: 18/27_

thepastaclaw:
@coderabbitai review

QuantumExplorer (commented):
Reviewed

PR #4161, opened 2026-07-20
fix(dapi): bound public request work

QuantumExplorer:

Applies service-owned work and retention budgets to public DAPI request paths. Covers internal review items 005, 006, 022, 023, 097, 098, 099, 100, 114, 190, 191, 192, and 401.

## What was done?

- Gave the Core gRPC service a separate inbound message ceiling and enforced transaction policy before forwarding.
- Validated BIP37 limits before construction and moved accepted Bloom bytes without intermediate payload clones.
- Added process-wide admission pools for transaction, header, masternode, and state-transition-wait operations.
- Resolved and capped historical replay ranges before creating subscriptions or workers.
- Rolled back combined-stream setup on planning failures and made receiver closure visible while workers are gated.
- Added absolute handoff deadlines plus item and byte budgets for staged live events.
- Bounded mempool snapshot entries, retained bytes, and delivered-identifier state.
- Added cancellation checks between backend calls and regression coverage for narrow input and accounting boundaries.

## How Has This Been Tested?

- `cargo test -p rs-dapi --lib` — 290 tests passed.
- Focused transaction, Bloom, selector-shape, and retained-byte boundary tests passed.
- `cargo clippy -p rs-dapi --all-targets --no-deps -- -D warnings` passed.
- Workspace formatting and diff checks passed.

The dependency-inclusive strict Clippy command remains blocked by unrelated pre-existing warnings in `rs-dpp`; the affected `rs-dapi` targets are clean.

## Breaking Changes

Requests outside the public service's documented transaction, replay, streaming, and BIP37 work budgets now fail with an explicit resource status.

## Checklist:

- [x] I have performed a self-review of my own code.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.
- [x] I have run formatting and strict package lint checks.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved transaction validation with clearer errors for empty, oversized, or overweight broadcasts.
  - Added protection against excessive concurrent requests and streams.
  - Improved stream cancellation when clients disconnect.

- **Improvements**
  - Added bounded buffering and deduplication for transaction and block-header streams.
  - Enforced limits on pending streamed data to prevent resource exhaustion.
  - Improved historical-to-live stream handoff reliability.
  - Added validation for bloom filters and request selectors.
  - Reduced the maximum accepted Core request size to 512 KiB.


thepastaclaw:
🕓 Ready for review — 66 ahead in queue (commit 25510816)
_Queue position: 67/72 · 2 reviews active_
_ETA: start ~11:31 UTC · complete ~12:00 UTC (median 28m across 30 recent reviews; 2 slots)_
_Queued 2h 57m ago · Last checked: 2026-07-22 19:40 UTC_

packages/rs-dapi/src/services/platform_service/wait_for_state_transition_result.rs
```diff
@@ -38,6 +38,16 @@ impl PlatformServiceImpl {
...
+            .try_acquire_owned()
+            .map_err(|_| {
+                DapiError::ResourceExhausted(
+                    "too many pending state transition result waits".to_string(),
+                )
+            })?;
+
         // Convert hash to commonly used representations
         let hash_hex = hex::encode(&state_transition_hash).to_uppercase();
         let hash_base64 = base64::prelude::BASE64_STANDARD.encode(&state_transition_hash);
```
thepastaclaw: **🔴 Blocking: Validate the state transition hash before admitting the wait**

A state-transition transaction ID is exactly 32 bytes, but this path rejects only an empty value. The public Platform service accepts messages up to 64 MiB, and each admitted request then retains the original bytes while creating an approximately 2x hexadecimal string, an approximately 4/3x base64 string, and another hexadecimal clone in the subscription filter. A malformed request can therefore consume hundreds of MiB, while the new semaphore still permits enough concurrent waits to exhaust process memory. Enforce the fixed-size invariant before acquiring a permit or performing either encoding.

```suggestion
        let state_transition_hash = v0.state_transition_hash;
        if state_transition_hash.is_empty() {
            return Err(DapiError::InvalidArgument(
                "state transition hash is not specified".to_string(),
            ));
        }
        if state_transition_hash.len() != 32 {
            return Err(DapiError::InvalidArgument(
                "state transition hash must be exactly 32 bytes".to_string(),
            ));
        }
```

<sub>source: ['codex']</sub>

packages/rs-dapi/src/services/streaming_service/block_header_stream.rs
```diff
@@ -259,11 +297,26 @@ impl StreamingServiceImpl {
                         && !Self::flush_pending(&subscriber_id, &tx, &delivered_hashes, &mut pending).await {
                             break;
                         }
+                    if !gated {
+                        pending_bytes = 0;
+                        delivered_hashes.lock().await.clear();
+                    }
```
thepastaclaw, line 425: **🔴 Blocking: Do not clear historical hashes as soon as the delivery gate opens**

The gate branch is biased ahead of `block_handle.recv()`, so it can run while duplicate block notifications are still queued in the subscription's 256-item channel. `flush_pending` handles only events already moved into the local vector; immediately clearing `delivered_hashes` means a queued notification for a block included in the replay is subsequently forwarded as a new header. Even without this explicit clear, the capacity branch in `forward_event` clears the full historical set when the first non-matching live block arrives because a replay may record up to 10,000 hashes while the threshold is 2,048. Retain a bounded recent overlap window until the subscription backlog has been drained instead of clearing all replay hashes at gate opening or on the first new block.

<sub>source: ['codex']</sub>

packages/rs-dapi/src/services/streaming_service/transaction_stream.rs
```diff
@@ -251,6 +365,17 @@ impl StreamingServiceImpl {
...
                             if gated {
+                                let event_bytes = event.retained_bytes();
+                                let next_bytes = pending_bytes.saturating_add(event_bytes);
+                                if pending.len() >= MAX_TRANSACTION_PENDING_EVENTS
+                                    || next_bytes > MAX_TRANSACTION_PENDING_BYTES
+                                {
+                                    let _ = tx.try_send(Err(Status::resource_exhausted(
+                                        "live events exceeded the history handoff budget",
+                                    )));
+                                    break;
```
thepastaclaw, line 447: **🔴 Blocking: Do not silently discard the handoff-budget error when the response queue is full**

Both transaction subscription branches discard the result of `try_send` when the staged-event budget is exceeded. A transaction replay can produce up to 4,000 response items while the response channel holds 512, so `TrySendError::Full` is expected for a slow client. The live worker then exits, but the history and mempool workers retain sender clones and continue; after buffered successful responses are consumed, the stream can close without ever delivering `ResourceExhausted`, appearing as a successful but truncated stream. Coordinate cancellation of all workers and use a delivery path that preserves the terminal status once channel capacity becomes available.

<sub>source: ['codex']</sub>

packages/rs-dapi/src/services/streaming_service/transaction_stream.rs
```diff
@@ -142,11 +179,58 @@ impl TransactionStreamState {
 
     async fn mark_instant_lock_delivered(&self, txid: &[u8]) -> bool {
         let mut guard = self.delivered_instant_locks.lock().await;
+        if guard.len() >= MAX_DELIVERED_IDS_PER_STREAM {
+            guard.clear();
+        }
         guard.insert(txid.to_vec())
```
thepastaclaw: **🟡 Suggestion: Capacity rollover reports an existing transaction as newly delivered**

The delivery helpers clear their sets before checking membership. At exactly `MAX_DELIVERED_IDS_PER_STREAM`, receiving any currently tracked transaction clears the set and makes `mark_transaction_delivered` return `true`, so the duplicate is forwarded despite the method's documented contract. The block and instant-lock helpers use the same ordering. Check membership before performing capacity rollover, and apply the same ordering consistently to all three sets; a bounded FIFO or LRU window would preserve stronger deduplication than clearing the entire set.

```suggestion
        let mut guard = self.delivered_txs.lock().await;
        if guard.contains(txid) {
            return false;
        }
        if guard.len() >= MAX_DELIVERED_IDS_PER_STREAM {
            guard.clear();
        }
        guard.insert(txid.to_vec())
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The PR adds useful request bounds, replay caps, and admission controls, but four in-scope blockers remain in the newly bounded paths. Malformed state-transition hashes can still consume hundreds of MiB per admitted wait, block-header handoff can emit duplicate headers, transaction budget failures can disappear behind a full response queue, and combined streams release admission permits before their replay workers finish; the transaction deduplication cap also has a boundary-ordering bug.

### Review provenance

🔴 4 blocking | 🟡 1 suggestion(s)

<details>

```

In `packages/rs-dapi/src/services/platform_service/wait_for_state_transition_result.rs`:
- [BLOCKING] packages/rs-dapi/src/services/platform_service/wait_for_state_transition_result.rs:33-53: Validate the state transition hash before admitting the wait
  A state-transition transaction ID is exactly 32 bytes, but this path rejects only an empty value. The public Platform service accepts messages up to 64 MiB, and each admitted request then retains the original bytes while creating an approximately 2x hexadecimal string, an approximately 4/3x base64 string, and another hexadecimal clone in the subscription filter. A malformed request can therefore consume hundreds of MiB, while the new semaphore still permits enough concurrent waits to exhaust process memory. Enforce the fixed-size invariant before acquiring a permit or performing either encoding.

In `packages/rs-dapi/src/services/streaming_service/block_header_stream.rs`:
- [BLOCKING] packages/rs-dapi/src/services/streaming_service/block_header_stream.rs:291-303: Do not clear historical hashes as soon as the delivery gate opens
  The gate branch is biased ahead of `block_handle.recv()`, so it can run while duplicate block notifications are still queued in the subscription's 256-item channel. `flush_pending` handles only events already moved into the local vector; immediately clearing `delivered_hashes` means a queued notification for a block included in the replay is subsequently forwarded as a new header. Even without this explicit clear, the capacity branch in `forward_event` clears the full historical set when the first non-matching live block arrives because a replay may record up to 10,000 hashes while the threshold is 2,048. Retain a bounded recent overlap window until the subscription backlog has been drained instead of clearing all replay hashes at gate opening or on the first new block.

In `packages/rs-dapi/src/services/streaming_service/transaction_stream.rs`:
- [BLOCKING] packages/rs-dapi/src/services/streaming_service/transaction_stream.rs:334-376: Do not silently discard the handoff-budget error when the response queue is full
  Both transaction subscription branches discard the result of `try_send` when the staged-event budget is exceeded. A transaction replay can produce up to 4,000 response items while the response channel holds 512, so `TrySendError::Full` is expected for a slow client. The live worker then exits, but the history and mempool workers retain sender clones and continue; after buffered successful responses are consumed, the stream can close without ever delivering `ResourceExhausted`, appearing as a successful but truncated stream. Coordinate cancellation of all workers and use a delivery path that preserves the terminal status once channel capacity becomes available.
- [BLOCKING] packages/rs-dapi/src/services/streaming_service/transaction_stream.rs:776-810: Keep the stream permit while replay and mempool workers are still running
  In combined mode, the admission permit is moved only into the live transaction worker, while the historical and mempool workers are spawned separately without it. When the response receiver closes, `transaction_worker` takes its `tx.closed()` branch and releases the permit even if a child is still in a Core RPC call. The history loop checks closure only before starting a block, so a disconnect during `get_block_hash` can still proceed through both full-block requests; tasks waiting for the Core RPC semaphore can also accumulate. Repeated connect/disconnect cycles can therefore recycle the 64 stream permits while substantially more replay work remains active. The combined-mode coordinator must retain the permit until every child is canceled or completed and cancel its `JoinSet` when the stream closes. The block-header combined path has the same split ownership between its live worker and independently spawned historical worker.
- [SUGGESTION] packages/rs-dapi/src/services/streaming_service/transaction_stream.rs:130-185: Capacity rollover reports an existing transaction as newly delivered
  The delivery helpers clear their sets before checking membership. At exactly `MAX_DELIVERED_IDS_PER_STREAM`, receiving any currently tracked transaction clears the set and makes `mark_transaction_delivered` return `true`, so the duplicate is forwarded despite the method's documented contract. The block and instant-lock helpers use the same ordering. Check membership before performing capacity rollover, and apply the same ordering consistently to all three sets; a bounded FIFO or LRU window would preserve stronger deduplication than clearing the entire set.
```
</details>

packages/rs-dapi/src/services/platform_service/wait_for_state_transition_result.rs
QuantumExplorer: Addressed in 5a21049571. State transition hashes must be exactly 32 bytes before the wait permit is acquired or any encoding occurs; the unit test covers empty, short, exact, and oversized inputs.

packages/rs-dapi/src/services/streaming_service/block_header_stream.rs
QuantumExplorer: Addressed in 5a21049571. Historical hashes survive gate opening in a bounded FIFO dedupe window, and duplicate membership is checked before oldest-entry eviction. The focused overlap-window test passes.

packages/rs-dapi/src/services/streaming_service/transaction_stream.rs
QuantumExplorer: Addressed in 5a21049571. Handoff overflow first cancels all producers and then awaits the terminal RESOURCE_EXHAUSTED send, so a full response queue cannot discard it. A full-queue regression test verifies delivery.

packages/rs-dapi/src/services/streaming_service/transaction_stream.rs
QuantumExplorer: Addressed in 5a21049571. Transaction, block, and instant-lock delivery IDs now use bounded FIFO eviction with membership checked before rollover; focused dedupe and capacity tests pass.

QuantumExplorer:
Addressed the remaining blocker in 25510816b2. Both combined-mode paths (transactions and block headers) now spawn a coordinator that owns the stream admission permit for the life of the stream: on disconnect it cancels the stream state, shuts down and awaits the replay/mempool JoinSet, and stops the live worker before the permit is released. Focused regression tests verify the permit is only released after the replay workers are actually gone, and that the gate still opens with the permit held after replay completes.

Also aligned the history bounds with shipped clients: dropped MAX_TRANSACTION_HISTORY_BLOCKS (2,000) and MAX_HISTORICAL_HEADERS_PER_STREAM (10,000), since wallet-lib replays its whole range in a single stream and BlockHeadersProvider targets 50,000 headers per stream — the caps turned legitimate initial syncs into RESOURCE_EXHAUSTED retry loops. Replay work remains bounded without them: both fetchers stream in per-block/per-batch chunks with awaited sends into bounded channels, the Core RPC semaphore bounds work in flight, and the admission permits (now correctly held for the full stream lifetime) bound concurrency. The replay→live dedupe window keeps its previous size as a standalone constant because live events can only overlap the tail of a replay.

Also merged v4.1-dev, reconciling with the payload-validation and metric-label changes that landed there.

PR #4167, opened 2026-07-20
fix(sdk): enforce response freshness anchors

QuantumExplorer:
## Summary

- enable a signed-metadata time freshness anchor by default for network SDK builders, with an explicit trusted-height alternative
- validate time before advancing a monotonic, concurrency-aware height high-water mark
- reject proof-enabled network configurations that disable every independent freshness anchor
- parse every current-quorum hash through one checked 32-byte conversion, including nested validator sets

This addresses security review findings DS-CAND-232, 249, and 267.

## Compatibility

Network builders now default to a 31-minute signed-time window. Deployments with intentionally unsynchronized clocks or longer network pauses should configure an appropriate time tolerance or seed a trusted initial Platform height. Mock builders retain their previous no-time-check default.

## Validation

- cargo check -p dash-sdk --lib
- cargo check -p drive-proof-verifier --lib
- cargo test -p dash-sdk --lib sdk::test:: (46 passed)
- cargo test -p drive-proof-verifier --lib unproved::tests:: (7 passed)
- cargo clippy -p dash-sdk -p drive-proof-verifier --lib --no-deps -- -D warnings
- cargo fmt --all
- git diff --check

### DIFF
--- a/packages/rs-sdk/src/sdk.rs
+++ b/packages/rs-sdk/src/sdk.rs
@@ -276,7 +276,14 @@ impl Sdk {
     fn freshness_criteria(&self, method_name: &str) -> (Option<u64>, Option<u64>) {
         match method_name {
             "get_addresses_trunk_state" | "get_addresses_branch_state" => (
-                None,
+                // Address synchronization checkpoints can lag the latest
+                // Platform height. Prefer their signed time when available,
+                // but retain the independently trusted height floor for the
+                // explicitly supported height-only configuration.
+                self.metadata_time_tolerance_ms
+                    .is_none()
+                    .then_some(self.metadata_height_tolerance)
+                    .flatten(),
                 self.metadata_time_tolerance_ms
                     .map(|configured| configured.min(DEFAULT_METADATA_TIME_TOLERANCE_MS)),
             ),
@@ -1090,8 +1097,10 @@ impl SdkBuilder {
     /// This method will return an error if the Sdk cannot be created.
     pub fn build(self) -> Result<Sdk, Error> {
         let is_network_sdk = self.addresses.is_some();
-        let has_height_anchor =
-            self.trusted_initial_height.is_some() && self.metadata_height_tolerance.is_some();
+        let has_height_anchor = self
+            .trusted_initial_height
+            .zip(self.metadata_height_tolerance)
+            .is_some_and(|(height, tolerance)| height > tolerance);
         if is_network_sdk
             && self.proofs
             && self.metadata_time_tolerance_ms.is_none()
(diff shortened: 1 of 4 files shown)


thepastaclaw:
🕓 Ready for review — 24 ahead in queue (commit 4585c42e)
_Queue position: 25/25_

packages/rs-sdk/src/sdk.rs
```diff
@@ -278,7 +278,7 @@ impl Sdk {
             "get_addresses_trunk_state" | "get_addresses_branch_state" => (
                 None,
                 self.metadata_time_tolerance_ms
-                    .and(Some(ADDRESS_STATE_TIME_TOLERANCE_MS)),
+                    .map(|configured| configured.min(DEFAULT_METADATA_TIME_TOLERANCE_MS)),
             ),
```
thepastaclaw, line 289: **🔴 Blocking: Height-only mode leaves address trunk proofs without a freshness anchor**

The builder now accepts a proof-enabled network SDK with time validation disabled when a trusted initial height and height tolerance are configured, but `freshness_criteria` unconditionally disables height validation for `get_addresses_trunk_state`. This produces `(None, None)` for the signed trunk response in that supported configuration. The trunk proof and metadata describe the node's latest stored checkpoint, and the address scan uses its returned height and root to pin all subsequent branch requests. A stale node can therefore replay an old, correctly signed checkpoint without comparing it to the caller's trusted height. The incremental catch-up does not make this safe because its first query treats failures as non-fatal and can return the accepted stale trunk state. Apply a trusted-height check suitable for checkpoint responses or reject time-disabled configurations for operations that require address trunk synchronization.

<sub>source: ['codex']</sub>

packages/rs-sdk/src/sdk.rs
```diff
@@ -1090,6 +1089,20 @@ impl SdkBuilder {
     ///
     /// This method will return an error if the Sdk cannot be created.
     pub fn build(self) -> Result<Sdk, Error> {
+        let is_network_sdk = self.addresses.is_some();
+        let has_height_anchor =
+            self.trusted_initial_height.is_some() && self.metadata_height_tolerance.is_some();
```
thepastaclaw: **🔴 Blocking: Builder accepts trusted heights that impose no freshness floor**

`has_height_anchor` checks only that the height and tolerance options are present. A configuration such as `with_trusted_initial_height(0)` or any trusted height less than or equal to its tolerance therefore passes construction with time validation disabled. In `verify_metadata_height`, the stale-response comparison is skipped whenever `expected_height <= tolerance`, so every possible response height is accepted and the first historical response establishes the effective watermark. Require the trusted height to exceed the configured rollback tolerance and add builder tests for zero and `height <= tolerance`.

```suggestion
let has_height_anchor = self
    .trusted_initial_height
    .zip(self.metadata_height_tolerance)
    .is_some_and(|(height, tolerance)| height > tolerance);
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The PR improves the default freshness policy and correctly prevents stale-time responses from mutating the height watermark, but the new trusted-height alternative does not provide the promised independent anchor in every accepted configuration. Address trunk responses bypass height validation entirely when time validation is disabled, and trusted heights at or below their rollback tolerance impose no effective lower bound.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — security-auditor (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — security-auditor (completed), `gpt-5.6-sol` — rust-quality (completed)

🔴 2 blocking

<details>

```

- [BLOCKING] packages/rs-sdk/src/sdk.rs:278-282: Height-only mode leaves address trunk proofs without a freshness anchor
  The builder now accepts a proof-enabled network SDK with time validation disabled when a trusted initial height and height tolerance are configured, but `freshness_criteria` unconditionally disables height validation for `get_addresses_trunk_state`. This produces `(None, None)` for the signed trunk response in that supported configuration. The trunk proof and metadata describe the node's latest stored checkpoint, and the address scan uses its returned height and root to pin all subsequent branch requests. A stale node can therefore replay an old, correctly signed checkpoint without comparing it to the caller's trusted height. The incremental catch-up does not make this safe because its first query treats failures as non-fatal and can return the accepted stale trunk state. Apply a trusted-height check suitable for checkpoint responses or reject time-disabled configurations for operations that require address trunk synchronization.
- [BLOCKING] packages/rs-sdk/src/sdk.rs:1093-1094: Builder accepts trusted heights that impose no freshness floor
  `has_height_anchor` checks only that the height and tolerance options are present. A configuration such as `with_trusted_initial_height(0)` or any trusted height less than or equal to its tolerance therefore passes construction with time validation disabled. In `verify_metadata_height`, the stale-response comparison is skipped whenever `expected_height <= tolerance`, so every possible response height is accepted and the first historical response establishes the effective watermark. Require the trusted height to exceed the configured rollback tolerance and add builder tests for zero and `height <= tolerance`.
```
</details>

QuantumExplorer:
Reviewed

packages/rs-sdk/src/sdk.rs
QuantumExplorer: Addressed in 547389f028 before merge. Height-only address trunk and branch checkpoint responses now retain the configured height tolerance, so they are checked against the independently trusted high-water mark; the height_only_address_checkpoint_uses_trusted_height_floor test verifies a stale checkpoint is rejected without lowering that floor.

packages/rs-sdk/src/sdk.rs
QuantumExplorer: Addressed in 547389f028 before merge. Builder validation now accepts a height-only proof configuration only when trusted_initial_height > metadata_height_tolerance, and the proof_enabled_network_builder_rejects_ineffective_height_anchor matrix covers zero and height-at-or-below-tolerance anchors.

PR #4169, opened 2026-07-20
fix(drive): bound shielded anchor query work

QuantumExplorer:
## Summary

- add a policy-derived storage-work safety envelope to the unpaginated V0 anchor query
- preflight the retained set with a bounded read before full proof generation
- preserve complete, wire-compatible responses within the active retention policy
- fail closed with a retryable `RESOURCE_EXHAUSTED` if retained state exceeds the envelope instead of silently truncating
- expose `max_retained_shielded_anchors()` on the versioned validation constants as the single statement of the retention policy's worst-case retained count, so the query bound stays coupled to the pruning algorithm rather than re-deriving it from raw constants

Covers security review finding DS-CAND-387.

## Compatibility

No protobuf, proof-query, or consensus format changes. Normal retained state remains below the derived envelope; a configuration or state that grows beyond it receives a retryable query error.

## Validation

- `cargo check -p drive-abci -p platform-version`
- targeted shielded-anchor tests (7 passing: exact-boundary, over-boundary, small-state round-trip for plain + proved, empty-state, and a cross-version invariant asserting the envelope covers the retention policy for every entry in `PLATFORM_VERSIONS`)
- `cargo clippy -p drive-abci -p platform-version --lib --no-deps -- -D warnings`
- `cargo fmt --all -- --check`


thepastaclaw:
🔍 Review in progress — actively reviewing now (commit 58862cf5)
_ETA: complete ~19:58 UTC (median 28m across 30 recent reviews)_
_Running 20m · Last checked: 2026-07-22 19:50 UTC_

packages/rs-drive-abci/src/query/shielded/anchors/v0/mod.rs
```diff
@@ -12,17 +13,52 @@ use dpp::check_validation_result_with_data;
...
+        )?;
+        let bounded_key_elements = bounded_results.to_key_elements();
+        if bounded_key_elements.len() > usize::from(MAX_V0_SHIELDED_ANCHORS) {
+            return Ok(QueryValidationResult::new_with_error(QueryError::Query(
+                QuerySyntaxError::InvalidLimit(format!(
+                    "retained shielded anchor set exceeds the V0 safety limit of {}",
+                    MAX_V0_SHIELDED_ANCHORS
+                )),
+            )));
+        }
```
thepastaclaw: **🟡 Suggestion: Server-side retained-state overflow is reported as an invalid client limit**

`GetShieldedAnchorsRequestV0` has only a `prove` field — no limit at all — so when the retained anchor set exceeds `MAX_V0_SHIELDED_ANCHORS`, this is purely a server-state condition the caller cannot have caused or can correct by retrying differently. Yet the handler reports it via `QuerySyntaxError::InvalidLimit`, and `query_error_into_status` (packages/rs-drive-abci/src/query/service.rs:988-999) collapses the entire `QuerySyntaxError` family — including `InvalidLimit` — unconditionally to gRPC `Status::invalid_argument`, with no variant-specific handling. Contrast this with the genuine uses of `InvalidLimit` elsewhere (e.g. `shielded/nullifiers/v0/mod.rs:31-40` and `token_queries/token_status/v0/mod.rs:28-36`), which check a client-supplied list length against a maximum — there, `InvalidLimit`/`INVALID_ARGUMENT` accurately describes a bad request. Here it misleads clients into treating a transient service-capacity condition as a permanently malformed request, discouraging retry-on-another-node or backoff-and-retry logic that would actually resolve it. Use a distinct server-state/capacity error and map it to a retryable status such as `RESOURCE_EXHAUSTED` or `UNAVAILABLE` instead.

packages/rs-drive-abci/src/query/shielded/anchors/v0/mod.rs
```diff
@@ -12,17 +13,52 @@ use dpp::check_validation_result_with_data;
...
 use drive::drive::shielded::paths::shielded_credit_pool_anchors_path_vec;
+use drive::error::query::QuerySyntaxError;
 use drive::grovedb::query_result_type::QueryResultType;
 use drive::grovedb::{PathQuery, Query, SizedQuery};
 use drive::util::grove_operations::GroveDBToUse;
 
+/// V0 has no pagination fields, so it may only serve a complete result while
+/// the retained set stays within this fixed storage-work budget. The active
+/// retention policy peaks at roughly 1,100 entries immediately before prune.
+const MAX_V0_SHIELDED_ANCHORS: u16 = 1_200;
```
thepastaclaw: **🟡 Suggestion: Safety envelope is a hardcoded constant instead of being derived from the versioned retention parameters it depends on**

`MAX_V0_SHIELDED_ANCHORS = 1_200` is justified purely by a code comment computing it from `platform_version.drive_abci.validation_and_processing.event_constants.shielded_anchor_retention_blocks` (1000) + `.shielded_anchor_pruning_interval` (100), read at runtime in `prune_shielded_pool_anchors_v0` (packages/rs-drive-abci/src/execution/platform_events/block_processing_end_events/prune_shielded_pool_anchors/v0/mod.rs:25-26). Those two fields are versioned specifically so retention behavior can change across protocol versions without call sites silently drifting out of sync — but this query module hardcodes their sum as a plain `const` instead of reading `platform_version` at call time. The sibling nullifiers query in the same module already establishes the correct pattern (`shielded/nullifiers/v0/mod.rs:31`: `platform_version.drive_abci.query.max_returned_elements`, read live rather than hardcoded). Currently there's no live drift — every defined `PlatformVersion` (v1 through v9) uses the same 1000/100 values — but if a future protocol version tunes retention without a matching update here, this V0 query would silently fail closed on legitimately in-policy state, with no compile-time or version-dispatch signal tying the two together. Derive the bound from `platform_version` at call time (retention_blocks + pruning_interval + margin), or add a test asserting the invariant holds for every supported `PlatformVersion` so a future version bump that violates it fails CI instead of drifting.

<sub>source: ['claude']</sub>

packages/rs-drive-abci/src/query/shielded/anchors/v0/mod.rs
```diff
@@ -74,3 +101,87 @@ impl<C> Platform<C> {
...
+
+            assert!(matches!(
+                result.errors.as_slice(),
+                [QueryError::Query(QuerySyntaxError::InvalidLimit(message))]
+                    if message.contains("V0 safety limit")
+            ));
+            assert!(result.data.is_none());
+        }
+    }
+}
```
thepastaclaw, line 331: **💬 Nitpick: No test covers the ordinary small-state path for query_shielded_anchors_v0**

This is the first `#[cfg(test)]` module ever added for `query_shielded_anchors_v0`. The single test exercises only the at-limit and one-over-limit boundary cases; there is still no test asserting the common-case behavior the PR reshuffled — e.g. an empty pool returning an empty `Anchors` list, or a small handful of anchors round-tripping correctly (matching keys, expected order) for both `prove: false` and `prove: true`. Since the response now reads from `bounded_key_elements` sourced from the bounded query instead of the original unbounded `results`, a regression in that plumbing for the ordinary case wouldn't be caught by the boundary-only test. Confirmed no other test (rs-sdk, dapi, platform-test-suite) currently exercises this endpoint's data path.

<sub>source: ['claude']</sub>

thepastaclaw (commented):

This PR adds a bounded preflight read (limit 1,201) ahead of the unpaginated V0 shielded-anchors query and correctly fails closed instead of running unbounded proof/scan work once the retained set exceeds the 1,200-entry envelope; the boundary logic is off-by-one correct and the new tests exercise both the at-limit and one-over-limit cases for prove=true/false. Two in-scope suggestions remain: the overflow condition is server-side retained state, not a client-supplied limit, yet it is reported via QuerySyntaxError::InvalidLimit which unconditionally maps to gRPC INVALID_ARGUMENT; and the 1,200-entry safety envelope is a hardcoded constant that duplicates rather than derives from the versioned shielded_anchor_retention_blocks/pruning_interval parameters it was calibrated against. A claimed TOCTOU gap between the preflight read and the full proof read was investigated and refuted: drive-abci's handle_blocking_query wraps the entire query call in a block-height-guard check that detects any commit landing mid-execution and discards/retries the whole result, so an inconsistent unbounded proof is never returned to a client.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — security-auditor (failed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (failed), `gpt-5.6-sol` — security-auditor (completed), `gpt-5.6-sol` — rust-quality (failed), `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — rust-quality (completed)
- Sonnet reviewers: `claude-sonnet-5` — general (failed), `claude-sonnet-5` — security-auditor (failed), `claude-sonnet-5` — rust-quality (failed), `claude-sonnet-5` — general (completed), `claude-sonnet-5` — security-auditor (completed), `claude-sonnet-5` — rust-quality (completed)

<details>

```

In `packages/rs-drive-abci/src/query/shielded/anchors/v0/mod.rs`:
- [SUGGESTION] packages/rs-drive-abci/src/query/shielded/anchors/v0/mod.rs:53-60: Server-side retained-state overflow is reported as an invalid client limit
  `GetShieldedAnchorsRequestV0` has only a `prove` field — no limit at all — so when the retained anchor set exceeds `MAX_V0_SHIELDED_ANCHORS`, this is purely a server-state condition the caller cannot have caused or can correct by retrying differently. Yet the handler reports it via `QuerySyntaxError::InvalidLimit`, and `query_error_into_status` (packages/rs-drive-abci/src/query/service.rs:988-999) collapses the entire `QuerySyntaxError` family — including `InvalidLimit` — unconditionally to gRPC `Status::invalid_argument`, with no variant-specific handling. Contrast this with the genuine uses of `InvalidLimit` elsewhere (e.g. `shielded/nullifiers/v0/mod.rs:31-40` and `token_queries/token_status/v0/mod.rs:28-36`), which check a client-supplied list length against a maximum — there, `InvalidLimit`/`INVALID_ARGUMENT` accurately describes a bad request. Here it misleads clients into treating a transient service-capacity condition as a permanently malformed request, discouraging retry-on-another-node or backoff-and-retry logic that would actually resolve it. Use a distinct server-state/capacity error and map it to a retryable status such as `RESOURCE_EXHAUSTED` or `UNAVAILABLE` instead.
- [SUGGESTION] packages/rs-drive-abci/src/query/shielded/anchors/v0/mod.rs:21-24: Safety envelope is a hardcoded constant instead of being derived from the versioned retention parameters it depends on
  `MAX_V0_SHIELDED_ANCHORS = 1_200` is justified purely by a code comment computing it from `platform_version.drive_abci.validation_and_processing.event_constants.shielded_anchor_retention_blocks` (1000) + `.shielded_anchor_pruning_interval` (100), read at runtime in `prune_shielded_pool_anchors_v0` (packages/rs-drive-abci/src/execution/platform_events/block_processing_end_events/prune_shielded_pool_anchors/v0/mod.rs:25-26). Those two fields are versioned specifically so retention behavior can change across protocol versions without call sites silently drifting out of sync — but this query module hardcodes their sum as a plain `const` instead of reading `platform_version` at call time. The sibling nullifiers query in the same module already establishes the correct pattern (`shielded/nullifiers/v0/mod.rs:31`: `platform_version.drive_abci.query.max_returned_elements`, read live rather than hardcoded). Currently there's no live drift — every defined `PlatformVersion` (v1 through v9) uses the same 1000/100 values — but if a future protocol version tunes retention without a matching update here, this V0 query would silently fail closed on legitimately in-policy state, with no compile-time or version-dispatch signal tying the two together. Derive the bound from `platform_version` at call time (retention_blocks + pruning_interval + margin), or add a test asserting the invariant holds for every supported `PlatformVersion` so a future version bump that violates it fails CI instead of drifting.
```
</details>

packages/rs-drive-abci/src/query/shielded/anchors/v0/mod.rs
QuantumExplorer: Addressed in bfac4f29f0. Retained-state overflow now uses QueryError::ResourceExhausted, which maps to gRPC RESOURCE_EXHAUSTED, with resource_exhausted_query_error_maps_to_retryable_grpc_status covering the mapping.

packages/rs-drive-abci/src/query/shielded/anchors/v0/mod.rs
QuantumExplorer: Addressed in bfac4f29f0. The V0 envelope is derived at runtime from the active versioned retention and pruning parameters plus bounded headroom; boundary tests cover the derived limit and overflow.

packages/rs-drive-abci/src/query/shielded/anchors/v0/mod.rs
QuantumExplorer: Addressed in bfac4f29f0. small_anchor_set_round_trips_for_plain_and_proved_queries covers ordinary small-state ordering for both the plain response and verified proof path.

PR #4176, opened 2026-07-21
fix(platform-wallet): persist the DashPay send used-flip after releasing the wallet-manager lock

shumkov:

An external review of the Kotlin-SDK PR's `rs-platform-wallet` / FFI changes flagged that `send_payment`'s used-flip store invokes the host persistence callback (Room/SwiftData FFI) **while holding the `wallet_manager` write lock**. I verified the finding by tracing it from source:

- `WalletPersister::store` → `FFIPersister::store` is **synchronous** and fires the host C-ABI callback inline; the trait doc notes `SqliteWalletPersister` even flushes on every `store`.
- The wallet is guarded by a single `tokio::sync::RwLock` (~100 read + ~104 write consumers). While the write guard is held across a slow host write, **every** accessor — balance queries, signing, other sends, background sync — is blocked, and the tokio worker thread is blocked too (the call is not `.await`ed).
- The lock is **non-reentrant**: a host store that called back into any manager API deadlocks. Both shipping hosts are safe today (Swift: synchronous SwiftData write, no manager reference; Kotlin: synchronous Room write, and its one native call — identity-key derive — is deliberately lock-free *because* a handle-keyed derive would re-`blocking_read` this same lock and deadlock). But the contract was **undocumented on the trait/vtable** — it existed only as an incidental inline comment — so a future change could reintroduce the deadlock with no signal.

## What was done?

**Persist-after-drop for the used-flip (the one clean site).** `send_payment` now captures the pool snapshot as an owned changeset under the guard, **drops the guard, then stores before the broadcast**. The durability-before-broadcast invariant is fully preserved (a relaunch must never re-hand a broadcast-to address). A funding-build failure now returns *before* the persist, leaving the consumed address only in memory — safe to re-hand, since it was never broadcast on-chain (and strictly less address-waste than before).

**Documented no-reentrancy contract.** Hoisted the previously-informal rule onto `PlatformWalletPersistence::store` and the FFI `PersistenceCallbacks` vtable: a `store` callback MUST NOT call back into any wallet-manager API (deadlocks the non-reentrant lock), and is latency-sensitive because it may run under the write guard.

**Left as-is (deliberately):** `ensure_identity_topup_account`'s registration persist and `record_dashpay_payment`'s Sent-entry persist both couple their rollback to the write-guarded `&mut` borrow, so moving them out of the lock cleanly would need a re-acquire + a TOCTOU window — more risk than benefit. `ensure_identity_topup_account` also already does a larger under-lock host round-trip (the Keychain xpub derivation), so isolating only its persister buys little.

## How Has This Been Tested?

- New regression test `send_payment_persists_external_pool_used_flip_after_releasing_lock` uses a probe persister that, from inside `store`, checks the shared wallet-manager lock state (`try_read` succeeds iff no writer holds it). Demonstrated red → green:
  - **✖ before fix** (store under the write guard): `observed: Some(false)` — assertion fails.
  - **✔ after fix** (store after the guard is released): `observed: Some(true)` — passes.
- `cargo test -p platform-wallet --lib`: **490 passed, 0 failed**.
- `cargo clippy -p platform-wallet -p platform-wallet-ffi --all-targets`: clean on the changed code (the 2 remaining warnings are pre-existing, in untouched test code).
- `cargo fmt --check` on both crates: clean.

## Breaking Changes

None. Public API unchanged; the persist-ordering change is internal and preserves the durability-before-broadcast guarantee.

## Checklist:

### DIFF
--- a/packages/rs-platform-wallet-ffi/src/persistence.rs
+++ b/packages/rs-platform-wallet-ffi/src/persistence.rs
@@ -113,6 +113,22 @@ pub const PLATFORM_WALLET_PERSISTENCE_CAPABILITY_WALLET_RESTORE: u64 = 1 << 7;
 /// General-purpose notifications (`on_store_fn`, `on_flush_fn`) plus
 /// typed callbacks that send incremental data across FFI for the caller
 /// to persist in their preferred storage backend.
+///
+/// # Reentrancy contract (required of every callback)
+///
+/// A persistence callback runs **synchronously on the calling native thread**,
+/// which is frequently a wallet-manager task **holding the manager's write
+/// lock**. That lock is non-reentrant. Therefore a callback MUST NOT call back
+/// into any wallet-manager / platform-wallet FFI (`dash_sdk_*` /
+/// `platform_wallet_*` and friends), directly or indirectly — doing so
+/// deadlocks. Write the changeset to your storage backend and return; do not
+/// query live wallet state to enrich a row, and do not synchronously drive an
+/// observer/UI reaction that re-enters native on the same thread. (A UI layer
+/// reacting to the committed rows on a *later* turn — SwiftUI `@Query`, a Room
+/// `Flow` collector on an async dispatcher — is fine: it runs after the
+/// callback returns and the lock is released.) Keep the work bounded; the call
+/// blocks every other wallet accessor while it runs. Mirrors the Rust-side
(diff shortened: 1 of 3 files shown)


thepastaclaw:
🕓 Ready for review — 29 ahead in queue (commit 32024cae)
_Queue position: 30/30_

PR #4171, opened 2026-07-20
feat(platform)!: document history system contract with per-doctype opt-in

QuantumExplorer:
## Issue being fixed or feature implemented

PR #4145 enabled DPNS username transfers and sales at protocol version 13, but they leave no queryable trace: the domain document is updated in place, overwriting the previous owner, the sale price, and when it happened. More generally, no document type on Platform has any transfer/sale/listing history — even though tokens already solved this exact problem with the TokenHistory system contract and per-token `keepsHistory` rules.

This PR ports that architecture to documents: a document-history system contract that any document type can subscribe to, with DPNS `domain` as the first subscriber.

## What was done?

**New `document-history` system contract** (`packages/document-history-contract`, `SystemDataContract::DocumentHistory = 8`) with three document types — `transfer`, `purchase` (records seller and the explicit price paid) and `priceUpdate`. All are `creationRestrictionMode: 2` (only the protocol can write them), immutable, and non-deletable. Indices support per-document provenance (`dataContractId, documentId, $createdAt`), per-contract feeds, per-identity history (buyer/seller/sender/recipient) and price analytics. `documentTypeName` is recorded as a data field but deliberately not indexed: document type names may be 64 chars while indexed strings cap at 63.

**Per-doctype subscription flags** in DPP — `keepsTransferHistory`, `keepsPurchaseHistory`, `keepsPricingHistory` — mirroring `TokenKeepsHistoryRules`. Parsed from the document type schema (default false), whitelisted in the v1 document meta-schema only (they are v13-introduced, so the v11→v12 migration allowlist is untouched, same as `documentsCountable`), and immutable on contract update like `documentsKeepHistory`.

**History writes in ops conversion** — the v1 transfer/purchase arms and a new v1 update-price arm check the source doctype's flags and emit a `DocumentOperationType::DocumentHistory` op. At apply time drive builds the history document via the new `DocumentEvent` (`rs-dpp`), with a deterministic ID derived from the source document, acting identity, and identity contract nonce — every validator produces identical history documents. The acting party (sender / buyer / seller respectively) owns and pays for the record.

**DPNS v2** subscribes `domain` to all three flags (`SYSTEM_DATA_CONTRACT_VERSIONS_V2`, consumed by `PLATFORM_V13`), so as of PV13: username transfers record sender→recipient, listings record the asking price, and sales record buyer, seller, and price paid — all queryable through the standard document query path with proofs.

[...]

PR #4179, opened 2026-07-21
fix: redact GitHub token from git errors

infraclaw-dash:
## What

- use x-access-token as the non-secret URL username and place GITHUB_TOKEN in the password position
- update all three credential URL rewrites in the Dockerfile
- prevent Git authentication errors from printing the credential

## Verification

- synthetic invalid credential was absent from Git stderr
- all three rewrites use the password position
- git diff --check passed

This targets the head branch of dashpay/platform#4147 because infraclaw-dash does not have direct push permission to dashpay/platform.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -464,7 +464,7 @@ RUN --mount=type=cache,sharing=shared,id=cargo_registry_index,target=${CARGO_HOM
     --mount=type=secret,id=GITHUB_TOKEN \
     trap 'rm -f "${HOME}/.gitconfig"' EXIT; \
     if [ -s /run/secrets/GITHUB_TOKEN ]; then \
-    git config --global url."https://$(cat /run/secrets/GITHUB_TOKEN)@github.com/".insteadOf "https://github.com/"; \
+    git config --global url."https://x-access-token:$(cat /run/secrets/GITHUB_TOKEN)@github.com/".insteadOf "https://github.com/"; \
     fi && \
     source /root/env && \
     export FEATURES_FLAG=""; \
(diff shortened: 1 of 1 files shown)


thepastaclaw:
✅ Final review complete — no blockers (commit 417f7b09)
_Last checked: 2026-07-22 19:02 UTC_

shumkov:
@thepastaclaw @infraclaw-dash  are you sure docker image is building well?

QuantumExplorer:
Superseded by 475f09f4db, which landed directly on the base branch (codex/fix-docker-build-secret-logging / #4147): it applies the same x-access-token: username form at all three insteadOf sites, and additionally brings the build-js stage up to the same standard (-s guard for empty secret files + EXIT-trap cleanup). The conflict here was this branch's context still expecting the pre-fix lines. Nothing from this PR is lost — thanks for the catch on git's username-vs-password redaction behavior.

thepastaclaw:
I checked the actual workflow results: neither this PR nor the superseding #4147 exercised the Dockerfile. In both runs, `Build Docker images` was skipped because the test workflow only builds images on a platform version change, the nightly schedule, or manual dispatch. So there is no CI evidence from these PR runs that the images build. The redaction change did land in #4147 as 475f09f4db; the next nightly/manual workflow run is the first full image-build validation.

thepastaclaw:
@shumkov The nightly run has now exercised the Docker builds, and no—the images are not building successfully yet.

All three image jobs (Drive, RS-DAPI, and Dashmate helper) failed in `build-planner` before reaching the GitHub credential/dependency-fetch stages. `cargo chef prepare` cannot load the newly added `document-history-contract` workspace member because `packages/document-history-contract/Cargo.toml` is absent from the Docker build context. The Dockerfile's explicit package copy lists were not updated when #4171 added that crate.

So the credential-redaction change from #4147 is not implicated in this failure, but it also still has not been exercised end-to-end by CI because the build stops earlier.

Nightly run: https://github.com/dashpay/platform/actions/runs/29965837295
Representative failed job: https://github.com/dashpay/platform/actions/runs/29965837295/job/89077700709

PR #4183, opened 2026-07-21
feat(kotlin-sdk)!: keystore rework — policy-alias split, layered key recovery, durable repair, structured signer errors (stacked on #4191)

bfoss765:
**Stacked on #4191** (`port/v4.1/unmanaged-identity-reads`) — review only the
commits above `ce3a17d71f`. Once #4191 merges,
`git rebase --onto origin/v4.1-dev ce3a17d71f` and force-push with
`--force-with-lease`.

This replaces the old single-squash `port/v4.1/android-integration`
(preserved at `backup/port-4060-pre-rework`). That squash predated #4172 and
silently rewrote the exact regions #4172 later hardened; this branch is a
rebuild on the current base with each review finding resolved explicitly.
The unmanaged-identity reads and the text-classified SigningKeyUnavailable
from the old squash shipped separately as #4191 and are NOT duplicated here.

**Scope — Keystore only, one logical commit per concern:**
1. **Alias split + policy** — `KeySecurityPolicy` (AUTH_GATED default /
   DEVICE_BOUND opt-in), two dedicated RSA aliases, legacy `KEYS_ALIAS`
   read-only; #4172's fingerprint capture + invalidation recovery survive
   alias-parameterized (finding 1).
2. **Lockless degradation, honest** — AUTH_GATED writes on a lockless device
   redirect to the DEVICE_BOUND alias (never a silently gate-less key under
   the auth-gated alias); `effectiveKeySecurityPolicy()` +
   `requireAuthGated` strict mode; per-blob alias tag routes reads;
   tightened KeyMint failure classifier (finding 4).
3. **Layered retrieve ladder** — legacy shape-dispatch FIRST (the fp gate
   never applies to legacy blobs, finding 2), fingerprint fast path with
   KPIE rethrow-after-generation-checked-cleanup (brick-loop fix, finding 1),
   mismatch/missing fp routes into former-RSA recovery, migration forward.
   Cheap `canSignWith` never decrypts; the real-decrypt probe moved to
   `probeIdentityKeyRecoverability` (finding 3).
4. **Pending keys** — queryable `pendingIdentityKeys` StateFlow,
   round-transactional staging.
5. **Durable repair signal** — Room 7→8 derivation-breadcrumb columns;
   restart reconstruction incl. blobs stranded by keypair replacement
   (finding 5).
6. **Forced, verified repair** — `replacePrivateKey` + `force` derive path +
   real-decrypt verification before `markIdentityKeyRepaired` (finding 6).
7. **⚠ BREAKING: 98 → `PlatformWallet.NotFound`** (Swift parity; release
   note in the commit).
8. **Structured signer discriminator** — typed `error_code` on the sign
   completion ABI, `ErrorSigningKeyUnavailable = 31` (26-30 reserved for
   #4185/#4184), Kotlin/Swift typed mappings; MESSAGE_MARKER sniff retained as a
   deprecated fallback for the #4191 merge-order transition (mixed
   old-native/new-Kotlin artifacts are unsupported — the completion JNI
   arity changed) (finding 7). Known residual: the
   Rust-internal segment rides a stable machine prefix through
   `ProtocolError::Generic` — a typed rs-dpp variant was deliberately
   avoided (serialization blast radius).
9. **Instrumented tests** adapted to the policy alias (finding 8; the five
   InvalidKeyException failures were the pre-#4172 emulator-relock flake —
   inherited workflow guard covers them).
10. **Docs** — parity manifest/spec entries (alias split N/A for iOS, 98 now
   convergent, pending-repair gap recorded for Swift, code-31 capability),
   migration ladder v8 note.

CI expectations: emulator job runs `:sdk:connectedDebugAndroidTest` (watch
the `deviceLocked=0` guard); swift job rebuilds the xcframework so the
5-arg completion ABI and code 31 compile against fresh headers.

---

*Round-3 addendum: verification passes found and fixed three repair-routing defects (health-probe fingerprint disproof for replaced aliases, typed first-attempt invalidation errors, durable repair seeding for legacy and former-RSA keys — commits `6d1fa5be83`, `f6fe865e56`, `7ee6e4b7f9`, `ce3490a3a0`). Unit suite: 249 passing; instrumented suite runs on this repo's emulator job.*

## Summary by CodeRabbit

- **New Features**
  - Added selectable identity-key security policies (auth-gated vs device-bound) and stricter policy enforcement on unsupported devices.
  - Introduced durable “pending identity keys” tracking with a restart-surviving repair eligibility probe.
  - Added typed “signing key unavailable” reporting for clearer, non-ambiguous signer failures.
- **Bug Fixes**
  - Correctly distinguishes missing wallets from unmanaged-identity situations.
  - DashPay now treats “managed identity not found” as an absence (instead of surfacing an error).
- **Security**
  - Improved identity-key generation/recovery and alias handling to avoid unsafe fallback behavior on lock-screen constrained devices.

### DIFF
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandler.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandler.kt
@@ -5,7 +5,11 @@ import androidx.room.withTransaction
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.ExecutorCoroutineDispatcher
 import kotlinx.coroutines.asCoroutineDispatcher
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
@@ -167,6 +171,25 @@ class PlatformWalletPersistenceHandler(
      */
     private class ChangesetBuffer {
         val ops: MutableList<suspend (DashDatabase) -> Unit> = mutableListOf()
+
+        /**
+         * [pendingIdentityKeys] map deltas staged by
+         * [onPersistIdentityKeyUpsert] during this round. The matching
+         * `PublicKeyEntity` row is only BUFFERED until [onChangesetEnd], so
+         * the pending-state change it describes is not true until that row
+         * commits: publishing a record early would flag a key whose
+         * watch-only row may be discarded by rollback, and publishing a
+         * clear early would drop the repair signal for an old watch-only
+         * row whose successful re-derive then rolls back (alias cleanup
+         * deletes the newly stored scalar). Applied in order, in ONE atomic
+         * [MutableStateFlow.update], only after the Room transaction
+         * commits; discarded with the buffer on rollback/abort so an
+         * aborted round leaves the pre-round map untouched
+         * (dashpay/platform#4060, finding de3cf44a71fc).
+         */
+        val pendingKeyDeltas:
+            MutableList<(Map<String, PendingIdentityKey>) -> Map<String, PendingIdentityKey>> =
+                mutableListOf()
     }
 
     /** Open rounds keyed by walletId hex (a round is per-walletId). */
@@ -296,6 +319,52 @@ class PlatformWalletPersistenceHandler(
         callbackExclusion.withLock { block() }
 
     /**
+     * An identity key whose private-half derivation/storage failed — the
+     * key was persisted **watch-only** and cannot sign until re-derived
+     * (e.g. via `PlatformWalletManager.repairIdentityKey`).
+     */
+    data class PendingIdentityKey(
+        /** Hex of the wallet the key belongs to. */
+        val walletIdHex: String,
+        /** Base58 of the owning identity id. */
+        val identityIdBase58: String,
+        /** On-identity key id. */
+        val keyId: Int,
+        /** Lowercase hex of the compressed public key (the storage key). */
+        val publicKeyHex: String,
+        /** Derivation breadcrumb: identity index. */
+        val identityIndex: Int,
+        /** Derivation breadcrumb: key index. */
+        val keyIndex: Int,
+        /** Human-readable failure reason (exception message or contract miss). */
+        val reason: String,
+        /** Epoch millis of the (latest) failure. */
+        val failedAtMs: Long,
+    )
+
+    private val _pendingIdentityKeys =
+        MutableStateFlow<Map<String, PendingIdentityKey>>(emptyMap())
+
+    /**
+     * Queryable "keys pending" state: identity keys whose private half
+     * could not be derived/stored by [onPersistIdentityKeyUpsert] (keyed by
+     * public-key hex). Such keys are persisted watch-only — signing with
+     * them fails — so hosts should watch this flow and surface a repair
+     * path. An entry clears automatically when a later persist round (or an
+     * explicit re-derive that replays the upsert) stores the key.
+     *
+     * Transactional with the round it belongs to: while a store round is
+     * open, record/clear mutations are staged in the round's
+     * [ChangesetBuffer] and published only after the Room transaction
+     * commits — a rolled-back or aborted round leaves this map exactly as
+     * it was before the round (see [ChangesetBuffer.pendingKeyDeltas]).
+     * Standalone (non-bracketed) upserts and [markIdentityKeyRepaired]
+     * publish immediately.
+     */
+    val pendingIdentityKeys: StateFlow<Map<String, PendingIdentityKey>> =
+        _pendingIdentityKeys.asStateFlow()
+
+    /**
      * Stage a write. If a round is open for [walletId] the op is buffered
      * for the round's single transaction; otherwise it runs immediately
      * in its own transaction (the standalone-callback path).
@@ -319,7 +388,8 @@ class PlatformWalletPersistenceHandler(
         // A pending leftover here means the previous round never reached
         // its end callback (abandoned mid-round) — its rows never
         // committed, so its aliases are orphans; scrub them like a
-        // rolled-back round. Then retry any earlier failed cleanup.
+        // rolled-back round (its staged pending-key deltas vanish with the
+        // replaced buffer). Then retry any earlier failed cleanup.
         scrubPendingAliases(key)
         retryOrphanedAliases(key)
         buffers[key] = ChangesetBuffer()
@@ -334,7 +404,9 @@ class PlatformWalletPersistenceHandler(
             // `backgroundContext.rollback()` — and delete the aliases the
             // deriver already wrote for this round: their rows will never
             // commit, so leaving them would strand undiscoverable
-            // identity-key ciphertext in the DataStore forever.
+            // identity-key ciphertext in the DataStore forever. The round's
+            // staged pending-key deltas are discarded with the buffer, so
+            // the pre-round [pendingIdentityKeys] map survives untouched.
             scrubPendingAliases(key)
             return@guarded 0
         }
@@ -346,11 +418,14 @@ class PlatformWalletPersistenceHandler(
                     }
                 }
             }
-            // Rows committed — the aliases are discoverable the normal way.
+            // Rows committed — the aliases are discoverable the normal way,
+            // and the round's pending-key state changes are now true.
             pendingRoundAliases.remove(key)
+            publishPendingKeyDeltas(buffer)
         } catch (t: Throwable) {
             // Commit failed: the staged rows never landed, so the round's
-            // aliases are orphans exactly like the !success branch.
+            // aliases are orphans exactly like the !success branch — and its
+            // pending-key deltas are equally void (discarded with the buffer).
             scrubPendingAliases(key)
             throw t
         }
@@ -1034,17 +1109,61 @@ class PlatformWalletPersistenceHandler(
         // wallet-deletion sweep still reaches it through the committed row).
         val deriveResult: DerivedKeyStoreResult? =
             if (derivationIndicesIsSome && deriver != null && !readOnly && walletStillPersisted) {
-                runCatching {
+                val keyOwnerWalletId = if (walletIdIsSome) keyWalletId else walletId
+                val outcome = runCatching {
                     deriver.deriveAndStore(
-                        walletId = if (walletIdIsSome) keyWalletId else walletId,
+                        walletId = keyOwnerWalletId,
                         publicKeyData = publicKeyData,
                         identityIndex = identityIndex,
                         keyIndex = keyIndex,
                     )
-                }.getOrElse { t ->
-                    Log.w(TAG, "identity private-key derive/store failed; key stays watch-only", t)
-                    null
                 }
+                val id = outcome.getOrNull()
+                if (id != null) {
+                    // Stored — clear any earlier failure for this pubkey.
+                    // Staged with the round (when one is open): if this
+                    // round rolls back, alias cleanup deletes the newly
+                    // stored scalar, so the old watch-only row must keep
+                    // its repair signal (finding de3cf44a71fc).
+                    stagePendingKeyDelta(roundKey, clearPendingKeyDelta(publicKeyData.toHex()))
+                } else {
+                    // NOT silent (dashpay/platform#4053): the key is being
+                    // persisted watch-only, so every signature with it will
+                    // fail until it is re-derived. Log loudly and record a
+                    // queryable pending entry (see [pendingIdentityKeys]).
+                    val reason = outcome.exceptionOrNull()?.let { t ->
+                        t.message ?: t.javaClass.simpleName
+                    } ?: "deriver returned no storage identifier"
+                    Log.e(
+                        TAG,
+                        "identity private-key derive/store FAILED — key " +
+                            "${publicKeyData.toHex()} (identity ${identityId.toBase58String()}, " +
+                            "keyId $keyId, slot $identityIndex/$keyIndex) is persisted " +
+                            "WATCH-ONLY and cannot sign until re-derived " +
+                            "(see PlatformWalletPersistenceHandler.pendingIdentityKeys): $reason",
+                        outcome.exceptionOrNull(),
+                    )
+                    // Staged with the round (when one is open): the row is
+                    // being persisted watch-only INSIDE the round's buffer,
+                    // so if the round aborts that row never commits and the
+                    // pending entry would be a phantom (finding de3cf44a71fc).
+                    stagePendingKeyDelta(
+                        roundKey,
+                        recordPendingKeyDelta(
+                            PendingIdentityKey(
+                                walletIdHex = keyOwnerWalletId.toHex(),
+                                identityIdBase58 = identityId.toBase58String(),
+                                keyId = keyId,
+                                publicKeyHex = publicKeyData.toHex(),
+                                identityIndex = identityIndex,
+                                keyIndex = keyIndex,
+                                reason = reason,
+                                failedAtMs = System.currentTimeMillis(),
+                            ),
+                        ),
+                    )
+                }
+                id
             } else {
                 null
             }
@@ -2428,6 +2547,78 @@ class PlatformWalletPersistenceHandler(
                 }
             }
         }
+    }
+
+    // ── Pending identity-key bookkeeping (#4053) ──────────────────────
+
+    // Every publish to the map goes through `MutableStateFlow.update`
+    // (atomic compare-and-set) rather than a plain read-modify-write on
+    // `.value`: the persistence callback publishes on the Rust caller
+    // thread while `markIdentityKeyRepaired` can clear from an arbitrary
+    // host thread (via PlatformWalletManager.repairIdentityKey). A
+    // non-atomic read-then-write could interleave and drop one of the two
+    // mutations — losing a record leaves a watch-only key with no queryable
+    // pending state, losing a clear leaves a repaired key stale.
+    //
+    // The mutations themselves are expressed as pure map deltas so the
+    // upsert callback can STAGE them with the round's [ChangesetBuffer]
+    // instead of publishing mid-round (finding de3cf44a71fc): the pending
+    // state a delta describes only becomes true when the round's Room
+    // transaction commits, and an aborted round must leave no trace.
+
+    private fun recordPendingKeyDelta(
+        entry: PendingIdentityKey,
+    ): (Map<String, PendingIdentityKey>) -> Map<String, PendingIdentityKey> =
+        { it + (entry.publicKeyHex to entry) }
+
+    private fun clearPendingKeyDelta(
+        publicKeyHex: String,
+    ): (Map<String, PendingIdentityKey>) -> Map<String, PendingIdentityKey> =
+        { if (publicKeyHex in it) it - publicKeyHex else it }
+
+    /**
+     * Stage [delta] with the wallet's open round (published atomically by
+     * [publishPendingKeyDeltas] after the round's transaction commits,
+     * discarded on rollback/abort), or publish immediately when no round is
+     * open — the standalone-callback path, whose Room write also commits
+     * immediately. Caller must hold [callbackExclusion] (every persist
+     * callback does), which also guards [buffers].
+     */
+    private fun stagePendingKeyDelta(
+        walletIdHex: String,
+        delta: (Map<String, PendingIdentityKey>) -> Map<String, PendingIdentityKey>,
+    ) {
+        val buffer = buffers[walletIdHex]
+        if (buffer != null) {
+            buffer.pendingKeyDeltas.add(delta)
+        } else {
+            _pendingIdentityKeys.update(delta)
+        }
+    }
+
+    /** Publish a committed round's staged deltas in ONE atomic map update. */
+    private fun publishPendingKeyDeltas(buffer: ChangesetBuffer) {
+        if (buffer.pendingKeyDeltas.isEmpty()) return
+        _pendingIdentityKeys.update { map ->
+            buffer.pendingKeyDeltas.fold(map) { acc, delta -> delta(acc) }
+        }
+    }
+
+    /**
+     * Drop [publicKeyHex] from [pendingIdentityKeys] after a successful
+     * out-of-band repair.
+     *
+     * [onPersistIdentityKeyUpsert] is the only *persist-callback* path that
+     * clears a pending entry, but [org.dashfoundation.dashsdk.wallet.PlatformWalletManager.repairIdentityKey]
+     * re-derives and stores the private key directly through the deriver,
+     * bypassing that callback — so it must call this on success or a repaired
+     * key would linger in [pendingIdentityKeys] until an unrelated re-persist
+     * happens to fire for the same key. Idempotent: clearing an absent key is a
+     * no-op. Publishes immediately (never staged with a round): the repair's
+     * scalar store already happened out-of-band, not inside any changeset.
+     */
+    internal fun markIdentityKeyRepaired(publicKeyHex: String) {
+        _pendingIdentityKeys.update(clearPendingKeyDelta(publicKeyHex))
     }
 
     // ── Error / threading guards ──────────────────────────────────────
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandler.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandler.kt
@@ -13,6 +13,7 @@ import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
+import org.dashfoundation.dashsdk.errors.DashSdkError
 import org.dashfoundation.dashsdk.ffi.AccountSpecData
 import org.dashfoundation.dashsdk.ffi.ContactProfileRestoreData
 import org.dashfoundation.dashsdk.ffi.ContactRequestRestoreData
@@ -2628,6 +2629,98 @@ class PlatformWalletPersistenceHandler(
      */
     internal fun markIdentityKeyRepaired(publicKeyHex: String) {
         _pendingIdentityKeys.update(clearPendingKeyDelta(publicKeyHex))
+    }
+
+    /**
+     * Re-derive, verify, and durably repair the identity key identified by
+     * [publicKeyData] — the orchestration behind
+     * `PlatformWalletManager.repairIdentityKey`, hoisted here so it is
+     * unit-testable (the manager cannot be constructed on the JVM) and so it
+     * shares this handler's authoritative [pendingIdentityKeys] state.
+     *
+     * ## Derivation source (dashpay/platform#4060 blocker 1)
+     *
+     * The derivation indices are read from the PERSISTED `public_keys` row's
+     * derivation breadcrumbs ([PublicKeyEntity.derivationIdentityIndex] /
+     * [PublicKeyEntity.derivationKeyIndex]) — NEVER from a caller-supplied
+     * key id. A caller-supplied index (e.g. the DPP key id) can derive a
+     * DIFFERENT valid scalar that round-trips through encrypt/decrypt fine;
+     * the deriver's [PrivateKeyDeriver.deriveAndStore] `force` path then
+     * proves the derived PUBLIC key equals [publicKeyData] BEFORE persisting
+     * and throws [org.dashfoundation.dashsdk.security.IdentityKeyDerivationMismatchException]
+     * on mismatch (nothing persisted, pending state untouched). A row with no
+     * breadcrumbs cannot be safely repaired, so the repair fails without
+     * clearing pending.
+     *
+     * @param verifyRecoverable the real-decrypt probe
+     *   (`WalletStorage.probeIdentityKeyRecoverability`) proving the just-written
+     *   blob actually opens; injected by the manager (this handler holds no
+     *   `WalletStorage`).
+     * @return the recorded storage identifier, or null when the deriver
+     *   declined to store (pending left intact). Throws (pending left intact)
+     *   on a derivation/verification failure.
+     */
+    internal suspend fun repairIdentityKeyDurably(
+        walletId: ByteArray,
+        publicKeyData: ByteArray,
+        verifyRecoverable: suspend (pubkeyHex: String) -> Boolean,
+    ): String? {
+        val pubkeyHex = publicKeyData.toHex()
+        val deriver = privateKeyDeriver
+            ?: throw DashSdkError.PlatformWallet.SigningKeyUnavailable(
+                "identity-key repair for $pubkeyHex has no private-key deriver wired; " +
+                    "the key remains unusable and pending state is left intact",
+            )
+
+        // BLOCKER 1: read the derivation indices from the persisted row, never
+        // from the caller. A row lacking breadcrumbs cannot be safely repaired
+        // (we would have to guess the slot), so fail WITHOUT clearing pending.
+        val breadcrumbs = database.publicKeyDao().getByPublicKeyData(publicKeyData)
+            .firstNotNullOfOrNull { row ->
+                val identityIndex = row.derivationIdentityIndex
+                val keyIndex = row.derivationKeyIndex
+                if (identityIndex != null && keyIndex != null) identityIndex to keyIndex else null
+            } ?: throw DashSdkError.PlatformWallet.SigningKeyUnavailable(
+                "cannot repair identity key $pubkeyHex: no derivation breadcrumbs are " +
+                    "persisted for it (derivationIdentityIndex/derivationKeyIndex are " +
+                    "null) — the correct slot is unknown; pending state left intact",
+            )
+        val (identityIndex, keyIndex) = breadcrumbs
+
+        // force = true routes through WalletStorage.replacePrivateKey and, in
+        // the production deriver, derives the KEYPAIR and verifies the derived
+        // public key equals publicKeyData BEFORE any store — a mismatch throws
+        // IdentityKeyDerivationMismatchException here, so nothing below runs
+        // and pending is never cleared (BLOCKER 1).
+        val storageIdentifier = deriver.deriveAndStore(
+            walletId = walletId,
+            publicKeyData = publicKeyData,
+            identityIndex = identityIndex,
+            keyIndex = keyIndex,
+            force = true,
+        )?.identifier ?: return null
+
+        // Independent confirmation the stored blob actually decrypts.
+        if (!verifyRecoverable(pubkeyHex)) {
+            throw DashSdkError.PlatformWallet.SigningKeyUnavailable(
+                "identity-key repair stored a blob that does not decrypt for pubkey " +
+                    "$pubkeyHex (slot $identityIndex/$keyIndex) — the key remains " +
+                    "unusable; pending state left intact",
+            )
+        }
+
+        // Record the identifier on the Room rows so the durable pending-repair
+        // reconstruction does not resurrect this key on the next launch.
+        runCatching {
+            database.publicKeyDao().getByPublicKeyData(publicKeyData).forEach { row ->
+                if (row.privateKeyKeychainIdentifier != storageIdentifier) {
+                    database.publicKeyDao()
+                        .update(row.copy(privateKeyKeychainIdentifier = storageIdentifier))
+                }
+            }
+        }
+        markIdentityKeyRepaired(pubkeyHex)
+        return storageIdentifier
     }
 
     /**
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandler.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandler.kt
@@ -1228,9 +1228,20 @@ class PlatformWalletPersistenceHandler(
         identityId: ByteArray,
         keyId: Int,
     ): Int = guarded {
+        val identityBase58 = identityId.toBase58String()
         stage(walletId) { db ->
-            db.publicKeyDao().deleteByIdentityAndKeyId(identityId.toBase58String(), keyId)
+            db.publicKeyDao().deleteByIdentityAndKeyId(identityBase58, keyId)
         }
+        // The row is gone, so a pending-repair entry for it is now a phantom:
+        // the key can never be re-derived/repaired into an identity that no
+        // longer carries it (dashpay/platform#4183 review). Drop it from
+        // [pendingIdentityKeys]. Staged with the round (mirroring the upsert
+        // path): published only if the deletion commits and discarded on
+        // rollback, so a rolled-back removal keeps the pre-round map intact.
+        stagePendingKeyDelta(
+            walletId.toHex(),
+            clearPendingKeyByIdentityKeyDelta(identityBase58, keyId),
+        )
         0
     }
 
@@ -2588,6 +2599,28 @@ class PlatformWalletPersistenceHandler(
         { if (publicKeyHex in it) it - publicKeyHex else it }
 
     /**
+     * Delta that drops any pending entry for the identity key
+     * ([identityIdBase58], [keyId]) — the shape [onPersistIdentityKeyRemoval]
+     * has (the map is keyed by public-key hex, which a removal callback does
+     * not carry, so it matches on the entry's identity + keyId instead).
+     * No-op (returns the same map instance) when nothing matches, mirroring
+     * [clearPendingKeyDelta] so an unrelated removal publishes no map update.
+     */
+    private fun clearPendingKeyByIdentityKeyDelta(
+        identityIdBase58: String,
+        keyId: Int,
+    ): (Map<String, PendingIdentityKey>) -> Map<String, PendingIdentityKey> =
+        { map ->
+            if (map.values.any { it.identityIdBase58 == identityIdBase58 && it.keyId == keyId }) {
+                map.filterValues {
+                    !(it.identityIdBase58 == identityIdBase58 && it.keyId == keyId)
+                }
+            } else {
+                map
+            }
+        }
+
+    /**
      * Stage [delta] with the wallet's open round (published atomically by
      * [publishPendingKeyDeltas] after the round's transaction commits,
      * discarded on rollback/abort), or publish immediately when no round is
@@ -2834,7 +2867,19 @@ class PlatformWalletPersistenceHandler(
             val walletId = identity.walletId ?: continue
             val pubkeyHex = row.publicKeyData.toHex()
             val usable = row.privateKeyKeychainIdentifier != null &&
-                runCatching { isPrivateKeyDecryptable(pubkeyHex) }.getOrDefault(false)
+                try {
+                    isPrivateKeyDecryptable(pubkeyHex)
+                } catch (cancellation: kotlin.coroutines.cancellation.CancellationException) {
+                    // The probe is suspend; runCatching turned its cancellation
+                    // into `false`, marking the row pending and returning
+                    // normally so callers (loadPersistedWallets, invalidation
+                    // bookkeeping) never observed the cancellation. Rethrow to
+                    // preserve structured concurrency; only genuine probe
+                    // failures become an unusable result (dashpay/platform#4183).
+                    throw cancellation
+                } catch (_: Throwable) {
+                    false
+                }
             if (usable) continue
             entries += PendingIdentityKey(
                 walletIdHex = walletId.toHex(),
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandler.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandler.kt
@@ -2652,18 +2652,41 @@ class PlatformWalletPersistenceHandler(
      * breadcrumbs cannot be safely repaired, so the repair fails without
      * clearing pending.
      *
+     * ## Durability (dashpay/platform#4060 blocker 3)
+     *
+     * The durable Room write (recording the storage identifier so the restart
+     * reconstruction does not resurrect the key) fails CLOSED: if it throws,
+     * the pending state is NOT cleared and the failure propagates, so the live
+     * session and a subsequent restart agree the repair is still pending. A
+     * swallowed durable-write failure that still cleared live pending state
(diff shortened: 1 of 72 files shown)


thepastaclaw:
🕓 Ready for review — 11 ahead in queue (commit 6e076c08)
_Queue position: 12/15 · 2 reviews active_
_ETA: start ~06:49 UTC · complete ~07:06 UTC (median 17m across 30 recent reviews; 2 slots)_
_Queued 1h 6m ago · Last checked: 2026-08-04 05:10 UTC_

bfoss765:
@coderabbitai full review

shumkov:
This needs reconciliation with #4172 before it can merge — the branch is `CONFLICTING`, and the conflicts are semantic, not textual. Concretely:

- #4172's `KeyPermanentlyInvalidatedException` recovery (generation-checked alias deletion) must survive: this PR's `decrypt()` rewrite lacks it, and with the alias split, repair would re-encrypt under the invalidated pair's still-present cert → permanent brick loop. (Your `generateWithLockScreenDegradation` actually complements it — both need to survive.)
- #4172's fingerprint gate in `retrievePrivateKey` and this PR's legacy-blob recovery are mutually exclusive as written (legacy blobs by definition fail the fingerprint check). Needs a layered resolution: legacy shape-detect → recovery first; fingerprint fast-path for current blobs.
- Auto-merge trap: v4.1-dev's `canSignWith` is contractually cheap/non-mutating on the Rust callback thread; this PR redefines `isPrivateKeyDecryptable` into a real decrypt probe (RSA op + `runBlocking`). Keep the cheap check on `canSignWith`; move the probing version to a key-health method.
- Instrumented CI is red: `storeIfAbsentRederivesWhenTheKeysAliasKeypairWasReplaced` still deletes legacy `KEYS_ALIAS` but writes moved to `KEYS_ALIAS_AUTH_GATED`; five more failures need triage.
- Sweep incomplete: `ManagedPlatformWallet.kt:549` and `IdentityRegistration.kt:93` keep the dead `== 0L` check — one unmanaged id now throws through the whole listing.

Suggestion: split the unmanaged-identity-reads + typed `SigningKeyUnavailable` changes (both good, conflict-free, wanted) into their own PR so they aren't blocked on the Keystore rework. Also: the 98→`PlatformWallet.NotFound` remap is host-breaking and needs a release-note flag; parity docs (#4172's manifest/spec) need updating for the new deliberate divergences.

shumkov:
Additional findings after checking the existing threads and the current #4172 base:

1. **`AUTH_GATED` silently loses authentication on lockless devices.** `KeystoreManager` drops both lock-bound parameters but continues to use the `AUTH_GATED` alias/policy. The fallback classifier also accepts any nested `KeyStoreException` after the generic "key generation failed" marker. Either fail with a typed policy error or explicitly select `DEVICE_BOUND`; do not describe the resulting key as auth-gated.
2. **The repair signal is not durable.** `pendingIdentityKeys` starts as an empty in-memory `MutableStateFlow` and is not reconstructed from Room. A normal restart loses the key slot needed for repair while the durable identity remains watch-only.
3. **Repair can report success without replacing the bad blob.** `storeIfAbsent` treats RSA shape plus alias fingerprint as usable without performing a real decrypt. `repairIdentityKey` can then skip derivation and still call `markIdentityKeyRepaired`. Repair needs a forced atomic replacement path, followed by a real-decrypt check before clearing pending state.
4. **`SigningKeyUnavailable` is still text-classified.** `KeystoreSigner` emits a marker and `DashSdkError` uses `message.contains`. Please carry a structured signer/native discriminator through Rust/C/JNI and map it in Kotlin and Swift.

I am not repeating the existing rebase/invalidation-recovery and cancellation comments; those remain valid. The earlier `canSignWith` decrypt concern is resolved at this head, and the legacy-before-fingerprint recovery is already present, so both should simply be preserved during conflict resolution.

bfoss765:
Took your suggestion: the conflict-free slices are split out as #4191 — unmanaged-identity reads (including both `== 0L` sites you caught, so one unmanaged id no longer fails a whole listing) plus the typed `SigningKeyUnavailable`, applied clean on current v4.1-dev with the Keystore rework and the host-breaking 98-remap deliberately excluded. This PR then becomes the Keystore rework only, properly reconciled with #4172's invalidation recovery and fingerprint gate; your four second-pass findings (lockless-device policy, repair durability, forced re-encrypt, structured discriminator) will be addressed there, with the 98→NotFound remap release-note-flagged and the instrumented failures triaged as part of the reconciliation.

shumkov:
Consolidated re-review of the rework (two independent passes). First the good news: all seven previously-blocking items are genuinely resolved — invalidated-key cleanup survives with generation-checked deletion, the retrieve ladder layers legacy shape-dispatch before the fingerprint gate, `canSignWith` is back to presence/fingerprint-only, the instrumented suite is retargeted and green, the #4191 stacking is clean, lockless devices get an honest `effectiveKeySecurityPolicy()`, and the typed code-31 signer error is a real upgrade.

However, the rework surfaces new blocking findings:

1. **Repair can persist the wrong private key.** The repair path derives from the caller-supplied indices, and the example app passes `keyIndex = key.keyId` (the DPP key id) — even though `PublicKeyEntity` carries `derivationIdentityIndex`/`derivationKeyIndex` breadcrumbs for exactly this purpose. `repairIdentityKey`'s verification is `probeIdentityKeyRecoverability`, which proves the stored blob *decrypts* — not that the derived key *matches* `publicKeyData`. A wrong index derives a different valid scalar that round-trips fine, clears pending state, and persists an unusable key. Fix: repair from the persisted derivation indices and verify derived-public-key equality against `publicKeyData` before persistence and state-clear. (`WalletKeyHealthSheet.kt:173-178`, `PlatformWalletManager.kt:482-540`, `IdentityKeyPrivateKeyDeriver.kt:48-100`)

2. **KeyMint failure classifier is too broad.** `isNoSecureLockScreenKeyGenFailure` path A classifies any `KeyStoreException` whose message contains `generate_key` (`KeystoreManager.kt:874-910`) — a transient KeyMint generation failure on a device *with* a lock screen silently and permanently downgrades AUTH_GATED to DEVICE_BOUND. Narrow to the authoritative lock-screen signal (numeric code / "lock screen" text) and drop the bare `generate_key` match.

3. **Durable repair state fails open.** In `repairIdentityKey`, the Room `privateKeyKeychainIdentifier` update is wrapped in `runCatching` (swallowed on failure) while `markIdentityKeyRepaired` clears live pending state regardless — a failed durable write resurrects the repair after restart while the session believes it's done. Make the durable update fail closed.

4. The existing unresolved thread on `CancellationException` being caught and suppressed in the WalletStorage migration remains valid.

5. Minor: DEVICE_BOUND documentation promises hardware-backed storage while the implementation permits the software AndroidKeyStore fallback — align the docs.

bfoss765:
All five new findings addressed (255 unit tests, 0 failures):

1. **Repair-key correctness** — repair now derives from the persisted `derivationIdentityIndex`/`derivationKeyIndex` breadcrumbs (never caller-supplied indices) and verifies the derived public key equals `publicKeyData` *before* any store or state-clear; a mismatch throws `IdentityKeyDerivationMismatchException` and leaves pending intact. The example app no longer passes `key.keyId` as the index.
2. **Classifier** — dropped the bare `generate_key` substring match; lockless classification now rests only on the authoritative lock-screen signal (text or numeric code 4/10309), so a transient KeyMint failure on a lock-screen device no longer permanently downgrades AUTH_GATED→DEVICE_BOUND.
3. **Durable repair fails closed** — the `privateKeyKeychainIdentifier` write propagates on failure and `markIdentityKeyRepaired` runs only after it commits, so a failed write leaves pending intact and self-heals on restart.
4. **Cancellation** — the WalletStorage migration rethrows `CancellationException`; only genuine rewrite failures stay best-effort.
5. **DEVICE_BOUND docs** — corrected to document the StrongBox→TEE→software-AndroidKeyStore fallback rather than promising hardware backing.

Two honest residuals, both fail-closed (neither can persist a wrong key): the real deriver's FFI pubkey-equality is exercised on-device rather than in a JVM unit test (AndroidKeyStore crypto can't run on the JVM) — it's in our device test plan; and on API 29–32 the numeric error code isn't available, so an exotic "reports-secure-but-rejects" OEM quirk on old Android wouldn't be text-classified, though genuine lockless devices are still caught by the primary `isDeviceSecure()` probe.

shumkov:
Round-3 verification (two independent passes, reconciled): all five previous findings are genuinely fixed — breadcrumb-driven repair with the index parameters removed from the API entirely, fail-closed durable writes, derived-pubkey verification before persistence, narrowed classifier, cancellation rethrow, honest DEVICE_BOUND docs. Two new findings from the deeper pass:

1. **P1 — repair permanently rejects valid `ECDSA_HASH160` keys.** `derivedPublicKeyMatches` is a raw `contentEquals` (`IdentityKeyPrivateKeyDeriver.kt:176`), comparing the 33-byte derived compressed pubkey against the stored `publicKeyData` — which for `ECDSA_HASH160` (and `EDDSA_25519_HASH160`) rows is the 20-byte hash, so the comparison always fails and that supported key type can never be repaired. Make the ownership check key-type-aware (ideally in Rust, matching DPP's comparison semantics) and add a real HASH160 repair test.
2. **P1 — `KeyStoreException` numeric code 4 is `ERROR_INTERNAL_SYSTEM_ERROR`, not a lock-screen signal** (code 3 is the documented no-LSKF/auth-bound-key condition). As classified, a transient internal Keystore error during generation still silently downgrades AUTH_GATED → DEVICE_BOUND. Remove 4 from the downgrade classifier (handle OEM-internal signatures separately) and add a transient-code-4 negative test.
3. P2 — the sign-time legacy invalidation path still converts durable-bookkeeping failure/cancellation into success via nested `runCatching` (`PlatformWalletManager.kt:385`): keep the typed code-31 outcome, but preserve a retryable repair signal instead of clearing it.

Nit: several new comments carry fix-round narration (issue/finding/reviewer provenance) — keep the invariant text, move the provenance to the PR description.

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/security/KeystoreManager.kt
```diff
@@ -42,38 +46,208 @@ import javax.crypto.spec.PSource
...
+     * at most once per call and only when the requested spec actually
+     * carried the lock-bound parameters (the auth-gated alias always does).
      */
-    fun encrypt(plaintext: ByteArray, alias: String = MASTER_ALIAS): EncryptedBlob {
-        if (alias == KEYS_ALIAS) {
-            return encryptForKeysAlias(plaintext).blob
+    private fun resolveIdentityKeysWriteAlias(): String {
+        if (keySecurityPolicy == KeySecurityPolicy.DEVICE_BOUND) return KEYS_ALIAS_DEVICE_BOUND
+        if (hasIdentityKeysKey(KEYS_ALIAS_AUTH_GATED)) return KEYS_ALIAS_AUTH_GATED
+        if (!deviceSecureProbe()) {
```
thepastaclaw, line 183: **🔴 Blocking: Lock-screen removal still reuses the auth-gated alias**

The auth-gated alias presence check runs before `deviceSecureProbe()`. Android can retain the private-key handle and certificate after an authentication-required key has been permanently invalidated by credential or lock-screen changes; this class's decrypt path explicitly handles that retained-handle behavior. Because encryption uses only the retained public key, new private scalars can be encrypted successfully under an alias whose private half will always throw `KeyPermanentlyInvalidatedException`. This also bypasses `requireAuthGated`: a strict manager on a now-lockless device reuses the alias instead of throwing, while `effectiveKeySecurityPolicy()` reports `AUTH_GATED` for the same unusable state. Probe the current lock-screen state before accepting an existing auth-gated alias, and apply the same ordering to effective-policy reporting.

<sub>source: ['codex']</sub>

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandler.kt
```diff
@@ -2425,6 +2555,301 @@ class PlatformWalletPersistenceHandler(
...
+            val identityIndex = row.derivationIdentityIndex ?: continue
+            val keyIndex = row.derivationKeyIndex ?: continue
+            val identityIdData = row.identityIdData ?: continue
+            val identity = database.identityDao().getByIdentityId(identityIdData) ?: continue
+            val networkRaw = network?.ffiValue
+            if (networkRaw != null && identity.networkRaw != networkRaw) continue
+            val walletId = identity.walletId ?: continue
+            val pubkeyHex = row.publicKeyData.toHex()
+            val usable = row.privateKeyKeychainIdentifier != null &&
+                runCatching { isPrivateKeyDecryptable(pubkeyHex) }.getOrDefault(false)
```
thepastaclaw: **🔴 Blocking: Pending-key reconstruction swallows coroutine cancellation**

`runCatching` catches `CancellationException` thrown by the suspend `isPrivateKeyDecryptable` callback and converts it to `false`. Reconstruction can then mark the row pending and return normally, preventing callers such as `loadPersistedWallets()` and the invalidation-bookkeeping path from observing cancellation at this boundary. Preserve structured concurrency by rethrowing cancellation and converting only genuine probe failures to an unusable result.

```suggestion
val usable = row.privateKeyKeychainIdentifier != null &&
    try {
        isPrivateKeyDecryptable(pubkeyHex)
    } catch (cancellation: kotlin.coroutines.cancellation.CancellationException) {
        throw cancellation
    } catch (_: Throwable) {
        false
    }
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/error.rs
```diff
@@ -336,6 +359,19 @@ impl From<PlatformWalletError> for PlatformWalletFFIResult {
...
+            // wallet operation that propagates the SDK error via `?`). The
+            // typed discriminator rides the stable machine prefix in the
+            // rendered message — restore it here too, but ONLY on the
+            // catch-all: the dedicated retry-semantics codes above are never
+            // overridden (dashpay/platform#4060 finding 7).
+            _ if error
+                .to_string()
+                .contains(rs_sdk_ffi::DASH_SDK_SIGNER_ERR_KEY_UNAVAILABLE_PREFIX) =>
+            {
+                PlatformWalletFFIResultCode::ErrorSigningKeyUnavailable
```
thepastaclaw: **🟡 Suggestion: Generic signer messages can be reclassified as typed code 31**

Both this catch-all and the direct `From<ProtocolError>` conversion at lines 467-472 search a fully rendered error with `contains(DASH_SDK_SIGNER_ERR_KEY_UNAVAILABLE_PREFIX)`. A foreign signer can return generic error code 0 with that token anywhere in its human-readable message, after which the platform-wallet ABI reports `ErrorSigningKeyUnavailable` (31) and hosts route an operational failure into key repair. Inspect the nested `ProtocolError::Generic` payload structurally and require the reserved marker at the payload start. The signer completion boundary should also escape or otherwise disambiguate generic-code messages that begin with the reserved marker, so only a typed `SigningKeyUnavailable` completion can create that representation.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The keystore alias split and durable repair plumbing are substantial, but three in-scope correctness blockers remain. Lock-screen removal can route new writes to an invalidated auth-gated key, cancellation is swallowed inside pending-key reconstruction, and deletion paths leave stale pending-key state; the signer error conversion also allows generic messages to be misclassified as typed key-unavailable failures.

### Review provenance

<details>

```

In `packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/security/KeystoreManager.kt`:
- [BLOCKING] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/security/KeystoreManager.kt:166-169: Lock-screen removal still reuses the auth-gated alias
  The auth-gated alias presence check runs before `deviceSecureProbe()`. Android can retain the private-key handle and certificate after an authentication-required key has been permanently invalidated by credential or lock-screen changes; this class's decrypt path explicitly handles that retained-handle behavior. Because encryption uses only the retained public key, new private scalars can be encrypted successfully under an alias whose private half will always throw `KeyPermanentlyInvalidatedException`. This also bypasses `requireAuthGated`: a strict manager on a now-lockless device reuses the alias instead of throwing, while `effectiveKeySecurityPolicy()` reports `AUTH_GATED` for the same unusable state. Probe the current lock-screen state before accepting an existing auth-gated alias, and apply the same ordering to effective-policy reporting.

- [BLOCKING] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandler.kt:2831-2832: Pending-key reconstruction swallows coroutine cancellation
  `runCatching` catches `CancellationException` thrown by the suspend `isPrivateKeyDecryptable` callback and converts it to `false`. Reconstruction can then mark the row pending and return normally, preventing callers such as `loadPersistedWallets()` and the invalidation-bookkeeping path from observing cancellation at this boundary. Preserve structured concurrency by rethrowing cancellation and converting only genuine probe failures to an unusable result.
- [BLOCKING] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandler.kt:1221-1229: Deleted identity keys remain in pendingIdentityKeys
  `onPersistIdentityKeyRemoval` deletes the Room row without removing its entry from `_pendingIdentityKeys`. Identity removal and `deleteWalletDataLocked` have the same omission: Room cascades remove durable rows, but they cannot mutate the process-local `StateFlow`. After a failed derivation followed by key, identity, or wallet deletion, hosts continue receiving a repairable-key signal for a row that no longer exists, and repair fails because no derivation breadcrumbs remain. Stage key and identity pruning with the same changeset transaction so rollback preserves valid signals, and prune wallet-scoped entries only after the wallet-delete transaction succeeds.

- [SUGGESTION] packages/rs-platform-wallet-ffi/src/error.rs:369-373: Generic signer messages can be reclassified as typed code 31
  Both this catch-all and the direct `From<ProtocolError>` conversion at lines 467-472 search a fully rendered error with `contains(DASH_SDK_SIGNER_ERR_KEY_UNAVAILABLE_PREFIX)`. A foreign signer can return generic error code 0 with that token anywhere in its human-readable message, after which the platform-wallet ABI reports `ErrorSigningKeyUnavailable` (31) and hosts route an operational failure into key repair. Inspect the nested `ProtocolError::Generic` payload structurally and require the reserved marker at the payload start. The signer completion boundary should also escape or otherwise disambiguate generic-code messages that begin with the reserved marker, so only a typed `SigningKeyUnavailable` completion can create that representation.
```
</details>

shumkov:
Round-4 verification (two independent passes, reconciled): the substance of all three fixes is verified — breadcrumb-driven repair with byte-verify backed by genuine external test vectors (ISO RIPEMD-160 + the canonical sipa pubkey), code-4 removal, and durable-write-before-state-clear ordering. Remaining:

1. **P1 — EDDSA_25519_HASH160 repair hashes the wrong curve.** The type-4 branch hashes the derived **secp256k1** pubkey (`IdentityKeyPrivateKeyDeriver.kt:193`), but DPP type 4 hashes the **Ed25519 verifying key** (`key_type.rs:246-251`) while the Rust derive always returns a compressed secp key (`derive_identity_key_at_slot.rs:69-79`) — so EDDSA repair can never succeed, and the test reuses one secp vector for both types, masking it. Recommended: move KeyType parsing + canonical public-data derivation + equality into one resolver-keyed **Rust** operation (this also removes the need for the hand-rolled Kotlin RIPEMD-160 — well-tested, but `platform_wallet_hash160` already exists and lacks only a JNI export), and add a real Ed25519 vector.
2. **P1 — the classifier checks the wrong channel.** `getNumericErrorCode()` cannot return 10309 — that's an internal KeyMint code that appears in the *message*; the public codes are 3 (keystore uninitialized / no LSKF) and 4 (internal system error). As shipped, the numeric path is dead code. Keep 4 negative, accept numeric **3**, recognize a bounded `10309` token only inside the matched key-generation message subtree, and restore the observed getter=4/message=10309 fixture.
3. The cancellation fix is one layer short: `KeystoreSigner.kt:136`'s `runCatching` around `onSigningKeyInvalidated` swallows the very `CancellationException` the manager now rethrows.
4. **Restack warning:** this branch now conflicts with #4191's current head in four files (`DashSdkError.kt`, `KeystoreSigner.kt`, `Dashpay.kt`, `DashSdkErrorTest.kt`) — both branches carry divergent copies of the typed-signing-error work. #4191 is merge-ready; plan a manual restack when it lands.
5. Standing nit: comment narration wasn't removed — the new commits add more.

bfoss765:
@thepastaclaw addressed all three blockers in `0d28d5ea` (`:sdk:testDebugUnitTest` = 265 tests, 0 failures):

- **Lock-screen alias reuse** — moved the `!deviceSecureProbe()` check *before* accepting an existing `KEYS_ALIAS_AUTH_GATED` alias, in both `resolveIdentityKeysWriteAlias()` and `effectiveKeySecurityPolicy()`, and corrected the now-false doc invariant. On a now-lockless device the auth-gated private half is treated as invalidated (public half retained); the write path degrades to DEVICE_BOUND / throws under `requireAuthGated`, and effective-policy reports honestly. Updated `KeySecurityPolicyTest`.
- **Pending-key reconstruction swallowing cancellation** (`PlatformWalletPersistenceHandler.kt:2831`) — replaced `runCatching{}.getOrDefault(false)` with try / rethrow-`CancellationException` / catch-`Throwable`→false.
- **Deleted identity keys lingering in `pendingIdentityKeys`** (`onPersistIdentityKeyRemoval`) — now stages a pending-clear delta (keyed by identityId+keyId, since the removal callback carries no pubkey hex), published on commit / discarded on rollback, mirroring the upsert path. Added commit-clears / rollback-preserves tests.

Also fixed the `KeystoreSigner` cancellation swallow at its two sites (inner callback + outer catch) so cancellation actually propagates.

**Separately flagging a real correctness bug I found while in here (not one of your three):** type-4 **EDDSA** identity-key *repair* can never succeed — `derive_at_slot_inner` derives an ECDSA/secp256k1 key, so `derivedPublicKeyMatches` HASH160s a secp256k1 pubkey and compares it against the stored HASH160 of the **Ed25519** pubkey → guaranteed mismatch. The correct EdDSA derivation reuses the same 32-byte scalar as an `ed25519_dalek` seed (per `derive_identity_key_bytes`), but the Kotlin comparison only receives the secp256k1 pubkey and has no Ed25519 path. Fixing it is a cross-language change (either the FFI returns the type-appropriate pubkey, or the Kotlin repair path computes the Ed25519 pubkey from the scalar) — I left it out of this commit rather than guess at the seed convention on a keystore path. Happy to take it as a follow-up.

The numeric lock-screen classifier (10309) I also looked at but left untouched: the existing `KeystoreKeyGenPolicyTest` cases assert `getNumericErrorCode()==10309` classifies, which contradicts the "dead branch" concern — I didn't want to change security-sensitive downgrade logic without pinning the exact mechanism.

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/security/KeystoreSigner.kt
```diff
@@ -79,10 +95,23 @@ class KeystoreSigner(
                     return@launch
                 }
                 signWithStoredKey(pubkeyBytes, data, completionToken)
+            } catch (cancellation: kotlin.coroutines.cancellation.CancellationException) {
+                // NEVER swallow structured-concurrency cancellation (a teardown
+                // of the signer's IO scope): rethrow so the coroutine unwinds
+                // instead of masking the cancellation as a native completion
+                // (dashpay/platform#4183 review). The generic Exception handler
+                // below would otherwise catch it and call completeSign.
+                throw cancellation
```
thepastaclaw, line 124: **🔴 Blocking: Signer cancellation abandons the native completion token**

After `signAsync` accepts a completion token, this cancellation path exits without invoking `SignerNative.completeSign`, despite `NativeSignerBridge` requiring exactly one completion per token. The Kotlin `scope.launch` job is independent of the Rust request, so rethrowing cancellation does not cancel the native receiver; Rust instead waits for the five-minute `SIGN_ASYNC_COMPLETION_TIMEOUT`, while the JNI `PendingSign` token remains allocated. The invalidation-bookkeeping callback can also route cancellation here before the already-known key-unavailable result is completed. Complete the token exactly once before unwinding—using the typed key-unavailable result when a `KeyPermanentlyInvalidatedException` is already known and a generic cancellation result elsewhere—or add an explicit native cancellation mechanism.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The latest patch fixes the lock-screen alias ordering and pending-key reconstruction cancellation, and key-specific removals now clear their pending state transactionally. Two blocking issues remain: identity and wallet deletion retain stale repair entries, and signer cancellation abandons the required native completion token; generic signer text can also still be misclassified as typed code 31.

### Review provenance

🔴 2 blocking

<details>

```

- [BLOCKING] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandler.kt:1039-1041: Identity and wallet deletion still leave pending repair entries
  Key-specific removal now stages a matching pending-state clear, but `onPersistIdentityRemoval` still stages only the Room identity deletion. Room's cascade removes the public-key rows without changing the process-local `_pendingIdentityKeys` StateFlow. `deleteWalletDataLocked` has the same omission: it deletes the wallet's identities and commits the wallet transaction without pruning entries whose `walletIdHex` belongs to that wallet. After either successful deletion, hosts continue receiving repair targets whose rows and derivation breadcrumbs no longer exist, so repair fails until the handler is recreated. Stage an identity-scoped clear in the identity-removal changeset, and prune wallet-scoped entries only after the wallet deletion transaction commits so rollback preserves valid signals.

In `packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/security/KeystoreSigner.kt`:
- [BLOCKING] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/security/KeystoreSigner.kt:98-104: Signer cancellation abandons the native completion token
  After `signAsync` accepts a completion token, this cancellation path exits without invoking `SignerNative.completeSign`, despite `NativeSignerBridge` requiring exactly one completion per token. The Kotlin `scope.launch` job is independent of the Rust request, so rethrowing cancellation does not cancel the native receiver; Rust instead waits for the five-minute `SIGN_ASYNC_COMPLETION_TIMEOUT`, while the JNI `PendingSign` token remains allocated. The invalidation-bookkeeping callback can also route cancellation here before the already-known key-unavailable result is completed. Complete the token exactly once before unwinding—using the typed key-unavailable result when a `KeyPermanentlyInvalidatedException` is already known and a generic cancellation result elsewhere—or add an explicit native cancellation mechanism.
```
</details>

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/security/KeystoreManager.kt
thepastaclaw: Resolved in `0d28d5e` — _Lock-screen removal still reuses the auth-gated alias_ no longer present.

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandler.kt
thepastaclaw: Resolved in `0d28d5e` — _Pending-key reconstruction swallows coroutine cancellation_ no longer present.

bfoss765:
@thepastaclaw fixed the signer-cancellation token leak in `d0048ef4`: added a per-invocation idempotent completer (AtomicBoolean CAS guarding `SignerNative.completeSign`, since the JNI does `Box::from_raw` and a second completion is a double-free). The cancellation catch now completes the token before rethrowing — typed `SIGNER_ERROR_CODE_KEY_UNAVAILABLE` when a `KeyPermanentlyInvalidatedException` is already known, generic `SIGNER_ERROR_CODE_GENERIC` otherwise — so the native PendingSign is released immediately instead of waiting out the 5-min timeout. All `completeSign` sites in `signWithStoredKey`/`signPlatformAddressOnDemand` route through the guard (exactly-once). CancellationException is rethrown. Working the second blocker (deletion retaining stale repair entries) + the code-31 suggestion next.

bfoss765:
@thepastaclaw addressed the remaining two in `297d56aa`:
- **Deletion retaining stale repair entries:** `onPersistIdentityRemoval` now stages an identity-scoped clear of `_pendingIdentityKeys` (published on commit / discarded on rollback, matching the `onPersistIdentityKeyRemoval` changeset model), and `deleteWalletDataLocked` prunes wallet-scoped entries **after** the delete transaction commits so a rollback preserves valid signals. Tests cover identity-clear, rollback-preserve, and wallet-clear.
- **code-31 (suggestion):** replaced the `contains(PREFIX)` checks with structural `ProtocolError::Generic(s) if s.starts_with(PREFIX)` in both `From` impls, and the signer completion boundary now escapes a generic message that begins with the reserved marker (`generic_signer_error: …`), so only a typed `SigningKeyUnavailable` completion can produce the marker-at-start form. Tests cover the substring-not-misclassified cases.
All green (69 persistence tests, error:: 18, signer:: 27).

packages/rs-platform-wallet-ffi/src/error.rs
```diff
@@ -336,6 +359,24 @@ impl From<PlatformWalletError> for PlatformWalletFFIResult {
...
+            // rendered error: a foreign signer can emit a generic (code-0)
+            // failure whose human-readable text merely mentions the token, or
+            // it can hide inside another variant's Display, and neither must
+            // be routed into key repair (dashpay/platform#4183 review).
+            PlatformWalletError::Sdk(dash_sdk::Error::Protocol(dpp::ProtocolError::Generic(s)))
+                if s.starts_with(rs_sdk_ffi::DASH_SDK_SIGNER_ERR_KEY_UNAVAILABLE_PREFIX) =>
+            {
+                PlatformWalletFFIResultCode::ErrorSigningKeyUnavailable
+            }
             _ => PlatformWalletFFIResultCode::ErrorUnknown,
```
thepastaclaw, line 426: **🔴 Blocking: Preserve typed code 31 through stringifying operation wrappers**

This narrowed arm recognizes the typed signer discriminator only while the error remains exactly `PlatformWalletError::Sdk(dash_sdk::Error::Protocol(ProtocolError::Generic(...)))`. Several public signing paths discard that structure before `unwrap_result_or_return!` converts the wallet error: token transfer wraps every SDK failure in `TokenError` at `packages/rs-platform-wallet/src/wallet/identity/network/tokens/transfer.rs:84-85`, DPNS registration wraps it in `InvalidIdentityData` at `wallet/identity/network/dpns.rs:236-241`, and document replacement does likewise at `wallet/identity/network/document.rs:528-534`. A genuine `SigningKeyUnavailable` completion still leaves the reserved prefix inside those strings, but the resulting variant now reaches the `_` arm and returns `ErrorUnknown`; the previous rendered-message check returned code 31. Swift maps that result to `.unknown` rather than `.signingKeyUnavailable`, so its key-repair routing is lost. Kotlin currently retains a deprecated marker-based fallback, but the native result still violates the structured code-31 contract. Preserve the typed SDK failure through these wrappers or promote it to a dedicated `PlatformWalletError` variant before stringification; do not restore unrestricted rendered-string searching.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

All three carried-forward findings are fixed at head 297d56aa99e4b8127b1b25769f3747f4c12c7101: deletion now transactionally clears pending repair state, cancellation completes the native token exactly once, and generic signer messages cannot impersonate typed code 31. The latest classifier narrowing nevertheless introduces a blocking regression because several public signing operations stringify the typed SDK error before FFI conversion, causing Swift to receive ErrorUnknown instead of ErrorSigningKeyUnavailable. Targeted Rust tests for the direct typed and generic-message classification paths pass, but they do not cover these stringifying operation wrappers.

### Review provenance

🔴 1 blocking

<details>

```

- [BLOCKING] packages/rs-platform-wallet-ffi/src/error.rs:375-380: Preserve typed code 31 through stringifying operation wrappers
  This narrowed arm recognizes the typed signer discriminator only while the error remains exactly `PlatformWalletError::Sdk(dash_sdk::Error::Protocol(ProtocolError::Generic(...)))`. Several public signing paths discard that structure before `unwrap_result_or_return!` converts the wallet error: token transfer wraps every SDK failure in `TokenError` at `packages/rs-platform-wallet/src/wallet/identity/network/tokens/transfer.rs:84-85`, DPNS registration wraps it in `InvalidIdentityData` at `wallet/identity/network/dpns.rs:236-241`, and document replacement does likewise at `wallet/identity/network/document.rs:528-534`. A genuine `SigningKeyUnavailable` completion still leaves the reserved prefix inside those strings, but the resulting variant now reaches the `_` arm and returns `ErrorUnknown`; the previous rendered-message check returned code 31. Swift maps that result to `.unknown` rather than `.signingKeyUnavailable`, so its key-repair routing is lost. Kotlin currently retains a deprecated marker-based fallback, but the native result still violates the structured code-31 contract. Preserve the typed SDK failure through these wrappers or promote it to a dedicated `PlatformWalletError` variant before stringification; do not restore unrestricted rendered-string searching.
```
</details>

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/security/KeystoreSigner.kt
thepastaclaw: Resolved in `d0048ef` — _Signer cancellation abandons the native completion token_ no longer present.

packages/rs-platform-wallet-ffi/src/error.rs
thepastaclaw: Resolved in `297d56a` — _Generic signer messages can be reclassified as typed code 31_ no longer present.

packages/rs-platform-wallet-ffi/src/error.rs
bfoss765: Fixed in 14efed0dd17957aff61da8f3e2ba76171633858c.

The three named signing paths no longer flatten a structured `SigningKeyUnavailable` completion before the FFI can classify it. New helper `platform_wallet::error::preserve_signer_key_unavailable_or(error, wrap)` keeps the failure verbatim under `PlatformWalletError::Sdk` — the one shape this arm already maps to code 31 — when the nested `ProtocolError::Generic` payload *starts with* the reserved marker, and otherwise hands the error to the caller's stringifying wrapper unchanged. The check is structural and position-0 only (no substring search), so a foreign signer that merely mentions the marker is still not routed into key repair.

Rerouted through the helper:
- `wallet/identity/network/tokens/transfer.rs` (`TokenError`)
- `wallet/identity/network/dpns.rs` `register_dpns_name` (`InvalidIdentityData`)
- `wallet/identity/network/document.rs` `document_replace` (`InvalidIdentityData`)

`platform-wallet` mirrors the marker locally (it deliberately does not depend on the FFI crate); a `const _: () = assert!(...)` in `platform-wallet-ffi` pins the mirror byte-identical to `rs_sdk_ffi::DASH_SDK_SIGNER_ERR_KEY_UNAVAILABLE_PREFIX`, so any drift is a build failure rather than a silent code-31 regression. No new `PlatformWalletError` variant or rendered-string search was introduced.

`cargo check -p platform-wallet-ffi` passes; added unit tests in `platform-wallet/src/error.rs` cover preserve / wrap-unrelated / mid-message-marker-not-preserved.

Scope note: only the three call sites cited in this thread were rerouted. Sibling wrappers (other token operations, document create/delete) still stringify their SDK error and would need the same treatment to be fully covered — happy to extend if you want that in this PR.

bfoss765: Coverage extended to the sibling wrappers in 2cd948331bb75968c0a3052bc029a8862fda9186 — the earlier three plus these complete the coverage of every signer-bearing stringifying wrapper.

Rerouted through `preserve_signer_key_unavailable_or(error, wrap)` exactly as the original three (helper checks the reserved marker at position 0 of the nested `ProtocolError::Generic`, keeps the failure verbatim under `Sdk`, otherwise hands the error to the unchanged wrapper):

- **Token operations** (`wallet/identity/network/tokens/`, `TokenError`): `burn`, `mint`, `freeze`, `unfreeze`, `claim`, `destroy_frozen_funds`, `pause`, `resume`, `purchase`, `set_price`, `update_config`
- **Document operations** (`wallet/identity/network/document.rs`, `InvalidIdentityData`): `create` (`put_to_platform_and_wait_for_response`), `delete`, `transfer`, `set_price`, `purchase`
- **Credit operations** (`InvalidIdentityData`): `withdrawal.rs` (`withdraw`), `transfer.rs` (`transfer_credits`), `transfer_to_addresses.rs` (`transfer_credits_to_addresses`)
- **`register_from_addresses.rs`** (`put_with_address_funding_fetching_nonces`): nested inside the existing `promote_address_nonce_error(...).unwrap_or_else(...)` wrapper so the marker is checked first and the address-nonce promotion is preserved unchanged.

Deliberately **not** touched, with reasons:

- Paths that already preserve the SDK error via `map_err(PlatformWalletError::Sdk)` — the shared `sdk_writer` (`put_document` / `send_contact_request`, which backs profile/contacts/contact_info/contact_requests), `invitation.rs`, and the identity `registration.rs` SDK calls — already reach code 31; no stringifier to reroute.
- `top_up_from_addresses.rs`: signs with the asset-lock `ExtendedPubKeySigner`, not the identity `Signer<IdentityPublicKey>` that stamps the reserved marker, so no code-31 completion flows there.
- Read-only fetches (`dpns.rs` resolve/sync/search, token `fetch_data_contract` helper) and post-broadcast local `add_key`/`add_keys` persistence — no signer invocation.

One structurally-distinct residual for a follow-up (not a combined-SDK-call `dash_sdk::Error`): `update.rs::update_identity_with_external_signer` splits build/broadcast, so the signer failure surfaces at `IdentityUpdateTransition::try_from_identity_with_signer` as a **bare `ProtocolError`** (not `dash_sdk::Error`) before being stringified into `InvalidIdentityData`. Rerouting it through this helper would require wrapping in `dash_sdk::Error::Protocol(e)`, which — because `Error::Protocol` renders as `"Protocol error: {0}"` — would prepend that text to unrelated protocol-error messages. I left it out of this pass rather than alter unrelated message text; it can be covered cleanly with a `ProtocolError`-typed sibling helper if you'd like it in this PR. It's the one realistic key-update path where the master key could be genuinely unavailable, so I'd lean toward doing it.

`cargo test -p platform-wallet --lib error::` (12 passed) and `cargo check -p platform-wallet-ffi` both clean at the new head.

bfoss765:
**On-device verification (keystore / screen-lock scenarios).** The keystore keygen + signer behavior was exercised on two physical devices (Samsung S21 + S22, testnet) across the screen-lock matrix — device screen-lock enabled and disabled, and the lock screen toggled mid-session — with **no user-facing failures**. Detailed device logs are omitted here to avoid re-running the security-sensitive scenarios (they alter on-device keystore state); they can be captured on request if required for merge.

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/security/KeystoreManager.kt
bfoss765: Addressed in `0d28d5e`; re-verified today at the rebased head `8387858016891417110e5ccca02fc9575e0fa6e5`.

`resolveIdentityKeysWriteAlias()` now runs `deviceSecureProbe()` **before** the auth-gated alias-presence check, so a retained-but-permanently-invalidated alias on a now-lockless device can no longer be reused for a write:

```kotlin
if (keySecurityPolicy == KeySecurityPolicy.DEVICE_BOUND) return KEYS_ALIAS_DEVICE_BOUND
if (!deviceSecureProbe()) {
    return degradeAuthGatedWrite("no secure lock screen (KeyguardManager.isDeviceSecure=false)", cause = null)
}
if (hasIdentityKeysKey(KEYS_ALIAS_AUTH_GATED)) return KEYS_ALIAS_AUTH_GATED
```

`degradeAuthGatedWrite` throws `KeySecurityPolicyUnavailableException` under `requireAuthGated`, so the strict-manager bypass you described is closed too — a strict manager on a lockless device now throws instead of silently reusing the alias.

`effectiveKeySecurityPolicy()` uses the same ordering (`!deviceSecureProbe() -> DEVICE_BOUND` before the alias check), so the report and the write path cannot disagree.

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/security/KeystoreSigner.kt
bfoss765: Addressed in `d0048ef`; re-verified today at the rebased head `8387858016891417110e5ccca02fc9575e0fa6e5`.

`signAsync` now routes every completion through an idempotent guard, so the token is completed exactly once no matter which path unwinds:

```kotlin
val completed = java.util.concurrent.atomic.AtomicBoolean(false)
val complete: (ByteArray?, Int, String?) -> Unit = { sig, code, msg ->
    if (completed.compareAndSet(false, true)) {
        SignerNative.completeSign(completionToken, sig, code, msg)
    }
}
```

Both cancellation routes you identified are covered:

- The outer `catch (cancellation: CancellationException)` completes with the generic cancellation result and then rethrows, so the JNI `PendingSign` token is released immediately instead of waiting out `SIGN_ASYNC_COMPLETION_TIMEOUT`, and structured-concurrency cancellation still propagates.
- The invalidation-bookkeeping route in `signWithStoredKey` catches cancellation from the suspend `onSigningKeyInvalidated` callback and completes with the **typed** `SIGNER_ERROR_CODE_KEY_UNAVAILABLE` result before rethrowing, since the key-unavailable outcome is already known at that point. The outer catch's completion is then a no-op via the guard — no double `Box::from_raw`.

packages/rs-platform-wallet-ffi/src/error.rs
bfoss765: Re-verified at the rebased head `8387858016891417110e5ccca02fc9575e0fa6e5` (the branch was rebased onto `v4.2-dev` at `5d68612a45` today; this content is unchanged from `2cd948331b`).

`preserve_signer_key_unavailable_or` is now applied across 18 files under `packages/rs-platform-wallet/src/wallet/identity/network/`, including the three you named — `tokens/transfer.rs`, `dpns.rs`, and `document.rs` — plus the sibling token, document, and credit operations.

The classifier arm in `packages/rs-platform-wallet-ffi/src/error.rs` is structural and position-0 only, not a rendered-string search:

```rust
PlatformWalletError::Sdk(dash_sdk::Error::Protocol(dpp::ProtocolError::Generic(s)))
    if s.starts_with(rs_sdk_ffi::DASH_SDK_SIGNER_ERR_KEY_UNAVAILABLE_PREFIX) =>
{
    PlatformWalletFFIResultCode::ErrorSigningKeyUnavailable
}
```

`cargo check -p platform-wallet-ffi -p platform-wallet -p rs-sdk-ffi -p rs-unified-sdk-jni` is clean at the rebased head.

The one residual documented above is unchanged and still open for a follow-up: `update.rs::update_identity_with_external_signer` surfaces the signer failure as a bare `ProtocolError` rather than a `dash_sdk::Error`, so it needs a `ProtocolError`-typed sibling helper rather than this one. Happy to fold that in here if you'd prefer it in this PR.

packages/rs-platform-wallet-ffi/src/error.rs
bfoss765: Addressed in `297d56a`; re-verified today at the rebased head `8387858016891417110e5ccca02fc9575e0fa6e5`.

Both classification sites now inspect the nested `ProtocolError::Generic` payload structurally and require the reserved marker at position 0, so `contains(...)` over a fully rendered error is gone:

```rust
PlatformWalletError::Sdk(dash_sdk::Error::Protocol(dpp::ProtocolError::Generic(s)))
    if s.starts_with(rs_sdk_ffi::DASH_SDK_SIGNER_ERR_KEY_UNAVAILABLE_PREFIX) =>
```

A foreign signer returning generic code 0 whose human-readable message merely mentions the token — or where the token is buried inside another variant's `Display` — no longer reaches `ErrorSigningKeyUnavailable` (31), so it cannot be routed into key repair. Unit tests in `platform-wallet/src/error.rs` cover preserve / wrap-unrelated / mid-message-marker-not-preserved.

bfoss765:
### Rebased onto `v4.2-dev` at `5d68612a45` — now MERGEABLE

Head is now `8387858016891417110e5ccca02fc9575e0fa6e5`. The branch was CONFLICTING against #4268 (registry-owned coordinator lifecycle), which rewrote the FFI enum, `manager.rs`, `persistence.rs`, and their Kotlin/Swift mirrors.

**Conflicts and how they were resolved** — all in commit `22fd3e5507`, in the two error-code mirrors:

- `packages/rs-platform-wallet-ffi/src/error.rs` — kept #4268's `ErrorShutdownIncomplete = 27` and its `for_platform_wallet_error` arm alongside this branch's `ErrorSigningKeyUnavailable = 31` and its classifier arm.
- `packages/swift-sdk/.../PlatformWalletResult.swift` — same, across the raw-value enum, `init(ffi:)`, the `PlatformWalletError` cases, `errorDescription`, and `init(result:)`.

**`ErrorSigningKeyUnavailable = 31` did not move**, per `ERROR_CODE_REGISTRY.md` — 31 is still this PR's, and existing codes are never renumbered.

**One deliberate content change.** The in-enum comment claimed 27–28 were reserved for the deferred-payment reservation trio and that #4185 must renumber off 26. Both statements are stale: 27 is now merged ABI, and the trio moved to 34–36. It now records that 28 and 30 are reserved, that 29 is #4184's, and points at the registry. Comment-only — no discriminant changed.

The two Kotlin files #4268 also touched — `PlatformWalletPersistenceHandler.kt` and `PlatformWalletManager.kt` — merged without conflict, and #4268's ownership/`release_fn` documentation is intact in the rebased tree (spot-checked alongside `WalletManagerNative.kt`).

**Verification at the rebased head:**

- `cargo check -p platform-wallet-ffi -p platform-wallet -p rs-sdk-ffi -p rs-unified-sdk-jni` — clean.
- The set of files this branch changes relative to its base is the same set as before the rebase.

All four outstanding blocking findings were re-verified against the rebased code and answered on their threads.

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandler.kt
bfoss765: Addressed in `0d28d5e`; re-verified today at the rebased head `8387858016891417110e5ccca02fc9575e0fa6e5`.

`PlatformWalletPersistenceHandler.kt` no longer routes the suspend probe through bare `runCatching`. Cancellation is rethrown and only genuine probe failures become an unusable result:

```kotlin
try {
    isPrivateKeyDecryptable(pubkeyHex)
} catch (cancellation: kotlin.coroutines.cancellation.CancellationException) {
    // The probe is suspend; runCatching turned its cancellation into `false`…
    throw cancellation
} catch (_: Throwable) {
    false
}
```

So `loadPersistedWallets()` and the invalidation-bookkeeping path observe cancellation at this boundary instead of seeing a row silently marked pending.

Worth noting for this file specifically: #4268 also modified `PlatformWalletPersistenceHandler.kt` (the owned-callback-context lifecycle docs). Today's rebase onto `v4.2-dev` at `5d68612a45` merged the two without conflict, and #4268's ownership/`release_fn` documentation is intact in the rebased tree.

QuantumExplorer (approved):

PR #4091, opened 2026-07-10
feat(kotlin-sdk): wire-compatible encrypted txMetadata document create + decrypt-on-fetch

bfoss765:
Closes #4086 (encrypted-document create/update) and #4087 (decrypt-on-fetch) for the wallet `txMetadata` contract in the Kotlin SDK. Context: dash-wallet migration plan Phase 3h step 3 (dashpay/dash-wallet#1507).

The single hard requirement is **wire-compatibility with the legacy `org.dashj.platform` stack** (`BlockchainIdentity.publishTxMetaData` / `getTxMetaData`): the Android wallet already publishes per-identity encrypted `TxMetadata` documents through the legacy library, so documents written by the legacy stack must decrypt with this implementation and vice versa — otherwise migrated users lose their memos / tax categories / exchange-rate records / gift cards.

## Recovered legacy scheme (byte-level)

Recovered by decompiling the legacy jars (`dash-sdk-kotlin` 4.0.0-RC2-SNAPSHOT — `BlockchainIdentity`, `TxMetadataDocument`, `TxMetadata`) and `org.bitcoinj.crypto.KeyCrypterAESCBC` (`dashj-core` 22.0.3), cross-checked against the dash-wallet `PlatformSyncService` fetch/publish loop.

**Key derivation — NOT ECDH, NOT HKDF, no hashing.** The AES-256 key is literally the raw 32-byte secp256k1 private scalar of a hardened HD child. `KeyCrypterAESCBC.deriveKey(ECKey)` is `new KeyParameter(ecKey.getPrivKeyBytes())`. The EC key is derived at:

```
<BLOCKCHAIN_IDENTITY accountPath> / keyIndex' / 32769' / encryptionKeyIndex'
```

- All three trailing components hardened; the fixed middle child is `ChildNumber(32769, hardened)` (`0x8001`, from `TxMetadataDocument`'s static init).
- On create, `keyIndex` = the identity's `getFirstPublicKey(ENCRYPTION, MEDIUM).getId()` (fallback `AUTHENTICATION, HIGH`); `encryptionKeyIndex` = the caller-supplied per-document index. `decryptTxMetadata` reads both back from the document and re-derives symmetrically. (The naming is counter-intuitive but internally consistent: document field `keyIndex` holds the encryption **key id**, field `encryptionKeyIndex` holds the app's index.)
[...]

PR #4186, opened 2026-07-21
feat(kotlin-sdk): wire-compatible encrypted txMetadata document create + decrypt-on-fetch

bfoss765:
Wire-compatible encrypted `txMetadata` document create + decrypt-on-fetch: keys derived through the mnemonic-resolver path, wire version enforcement (0|1) in Rust with rejection of version bytes the legacy stack cannot parse, dashj-generated wire-compat pin vectors (including a nonzero-identity vector and a real-DerivationPathFactory provenance vector), a shared fetched-DataContract Arc, dual-emit encrypted-document breadcrumbs, and TeardownGate-gated create/fetch on the Kotlin side.

Re-opens #4091 which was auto-closed when the #3999 base branch was deleted; rebased onto `v4.1-dev`. All thirteen original commits replayed cleanly — no hunks needed to be dropped as already-absorbed (`Cargo.lock` additions merged without conflict and are stable under `cargo check`).

Verified: `cargo test -p platform-wallet -p platform-wallet-ffi -p rs-unified-sdk-jni` all pass (630 / 241 / 10, including the legacy wire-compat suites); `:sdk:assembleRelease` + 177 sdk unit tests pass.

---

- **Size precheck**: typed `TxMetadataPayloadTooLarge { len, max }` raised before any key derivation or network work, at both `prepare_encrypted_txmetadata_properties` and inside `seal_tx_metadata`. `MAX_TX_METADATA_PLAINTEXT_LEN` is derived from the field limit in code and a new test anchors the 4096 constant to the wallet-utils contract schema's `maxItems` so a contract-side change can't drift silently. One arithmetic correction from implementing this: the 4063/4064 boundary from review is confirmed, but a 4063-byte plaintext frames to a **4081**-byte blob, not 4096 (PKCS7 only adds a full pad block on block-aligned input; 4064 → 4097, and 4082..4096 is unreachable). The boundary test asserts the real envelope and round-trips it.
- **Plaintext lifetime**: the payload copy is `Zeroizing`, owned by the prepare closure, and explicitly dropped the moment encrypted properties exist — before the broadcast `await`, on success and error paths alike.
- **Ignored fetch test**: doc now states it's a manual/testnet-gated `--ignored` check rather than claiming CI coverage.
- **Rebase**: replayed onto current `v4.1-dev` (post-#4172/#4187); range-diff shows all feature commits identical (`=` patch-ids).
- **Follow-ups (not in this PR, per review discussion)**: moving `encryptionKeyIndex` allocation into Rust wallet state; an independent legacy-blob decrypt check from a real legacy install.

### Provenance notes (tracking tokens moved out of code comments)

Internal review-tracking tokens `9c0ce58c3bb7`, `79595960d201`, `dd246b5e17d0`, `4c0754158cc6`, `989be307db0f` were removed from 8 comment sites (seal/version-guard docs and wire-compat vector docs in `tx_metadata.rs`, `encrypted_document.rs`, `transactions.rs`, and the legacy wire-compat harness README/Java header). They referenced findings resolved in earlier review rounds of the original PR line; the rationale text at each site is unchanged.

---
### Provenance (tracker refs moved from code comments per review)
`rs-unified-sdk-jni/src/transactions.rs` carried two `dashpay/platform#4091` review-tracking tokens (at the seal-version-guard and the external-signable-derive comments); removed from the comments with the rationale text kept.


thepastaclaw:
🔍 Review in progress — actively reviewing now (commit 1fcdfd6b)
_ETA: complete ~10:35 UTC (median 20m across 30 recent reviews)_
_Running 4m · Last checked: 2026-08-04 10:20 UTC_

shumkov:
Reviewed with focus on the crypto and wire-compat claims — this is done right: all wire-defining logic (derivation, AES-CBC envelope, keyIndex selection, fetch semantics) lives in shared Rust behind C-ABI exports iOS already links, keys are `Zeroizing` with RAII master wipe, and the three-layer vector evidence (dashj end-to-end, NIST CBC, resident-vs-resolver pin) is unusually strong. Asks:

- The legacy vectors are self-generated via the checked-in dashj scratch programs. Since wire-compat is the core claim, one independent check — decrypting a blob produced by a real legacy dash-wallet install — would close the loop.
- `seal_tx_metadata`: no client-side size check; payloads >~4078 B derive+seal then die at broadcast with an opaque DPP error. A `17 + padded_len <= 4096` precheck would fail fast.
- `tests/txmetadata_fetch.rs` doc says the regression is "caught in CI" but the test is `#[ignore = "hits testnet"]` and nothing runs `--ignored` — wire it into a scheduled job or soften the comment.
- Branch is 8 behind base (merges clean, but #4167/#4173 are semantically nearby) — please rebase before merge. Also a dozen "finding …" tracker refs in comments belong in the PR description.
- Follow-up to file: Swift wrappers for the two new FFI exports (mechanical — symbols already land in `DashSDKFFI.h`), and pinning the app-level protobuf payload schema cross-platform.

bfoss765:
@coderabbitai full review

shumkov:
Additional findings after checking the existing txMetadata comments:

1. **`encryptionKeyIndex` allocation policy remains in the Kotlin host.** The public API tells callers to supply the legacy `1 + countAllRequests()` counter. Concurrent callers/devices can select the same index, and this conflicts with the Kotlin SDK's host-thin rule against key-index policy loops. Rust wallet/document state should allocate or atomically reserve the index; hosts should provide only the opaque payload.
2. **The native plaintext copy remains live across the network await.** `platform_wallet_create_encrypted_document_with_signer` copies into `payload_vec`, prepares ciphertext, drops the master key, then leaves `payload_vec` in the outer scope while `block_on_worker` broadcasts. Wrap it in `Zeroizing<Vec<u8>>` and explicitly drop it immediately after encrypted properties are prepared.

The existing size-limit comment is correct, with one boundary correction: the 4096-byte field limit and `1-byte version + 16-byte IV + PKCS7 ciphertext` envelope make **4063 bytes** the largest plaintext. A 4064-byte plaintext produces 4097 bytes. Please reject above 4063 before derivation/network work and test 4063/4064.

bfoss765:
All asks are done and pushed: typed size precheck before any derivation/network work, `Zeroizing` plaintext dropped before the broadcast await, the ignored-test comment softened, tracker refs moved to the description, and the branch rebased onto current v4.1-dev (range-diff clean over #4172/#4187). One arithmetic correction from implementing the precheck: your 4063/4064 boundary is confirmed correct, but the max blob is **4081** bytes, not 4096 — PKCS7 only adds a full pad block on block-aligned input, so 4082..4096 is unreachable; the boundary test pins the real envelope and round-trips it. We also anchored the 4096 constant to the contract schema's `maxItems` with a test so a contract-side change can't drift silently.

Your two second-pass items are implemented as stacked follow-up PRs rather than deferred: (1) `encryptionKeyIndex` allocation moved into Rust — hosts pass nothing, Rust seeds `1 + count` from Platform state and serializes in-process allocation behind a per-wallet mutex (cross-device collisions remain possible but are provably non-lossy: per-document CSPRNG IVs under the deterministic key make CBC sound, and readers derive from each document's own stored index); (2) Swift wrappers for both new FFI exports, verified against a rebuilt framework slice. For the independent legacy check, we're producing a blob from a real legacy (11.9) install on a throwaway testnet wallet and will pin the decryption as a fixture.

bfoss765:
The independent legacy check promised above is done. A stock dash-wallet 11.9 testnet install (DPNS name `yabba2`, a designated throwaway wallet) published an encrypted txMetadata document through the legacy dashj path; we fetched it from testnet with the production query and the new Rust crypto decrypted the real legacy blob byte-for-byte — a protobuf `TxMetadataBatch` with two transactions (send + receive), memos, and USD exchange rates.

Pinned on this branch as `legacy_install_yabba2_wire_compat_vector` — a network-free fixture hard-coding the recovery phrase, the captured blob, and the expected plaintext (the phrase is deliberately public; the wallet is burned) — plus a testnet-gated `#[ignore]` capture helper, in commit 61072a94. This closes the loop on the wire-compat claim with a vector this repo did not generate.

bfoss765:
@coderabbitai full review

shumkov:
Consolidated re-verification (two independent passes). All round-1 asks are done and done well: the on-chain `yabba2` vector is a genuine independent check (production query path, publicly re-runnable capture helper — materially stronger than the earlier self-generated vectors), the size precheck landed at both layers with the corrected 4081 arithmetic (anchored to the contract schema), the ignored-test comment is now honest, tracker refs are gone, and the wire-compat/size-cap/schema tests pass. One remaining change plus two pre-merge notes:

1. **The JNI plaintext copy outlives the fix.** `payload_bytes` in `rs-unified-sdk-jni/src/transactions.rs:822-848` is a plain `Vec<u8>` held across the entire synchronous FFI call — including the network broadcast wait — and dropped without zeroization. The inner FFI copy got `Zeroizing` + early drop; give the JNI-owned copy the same treatment so the plaintext-lifetime discipline actually holds end-to-end.

2. Base has advanced ~14 commits since the rebase; #4160 (FFI lifetime invariants) touches adjacent `document.rs` ground — worth a rebase-check before landing.

3. Please cite #4194/#4195 by number in the body (they're referenced only as "stacked follow-ups"), noting #4194 currently wraps the explicit-index export — see the sequencing comment there.

shumkov:
Addendum/correction: two tracker references survive in `packages/rs-unified-sdk-jni/src/transactions.rs:811,896` (the earlier sweep covered the other crates) — same move-to-PR-description treatment, non-blocking.

bfoss765:
Addressed:

- **JNI plaintext lifetime** — the JNI-owned `payload_bytes` copy in `transactions.rs` is now `Zeroizing` and explicitly dropped the instant the FFI call returns, the earliest point reachable from JNI (the broadcast runs inside that synchronous call), so the discipline holds end-to-end. One honest boundary: the source JVM `byte[]` the caller passes still lives on the JVM heap until GC — Rust can't scrub that; host-side scrubbing would be separate Kotlin work.
- **Rebase** — onto v4.1-dev tip (267e1cb33a). #4160 touches different files (`rs-sdk-ffi/document/queries/info.rs`, `tx_decode.rs`), so there's no `document.rs` overlap — clean rebase, yabba2 fixture intact.
- **Tracker refs** — the two remaining #4091 tokens in `transactions.rs` are removed (rationale kept), moved to the description.
- Stacked follow-ups are #4194 / #4195.

platform-wallet 505/505, ffi 204/204, both legacy wire-compat vectors pass.

shumkov:
Round-3 verification (two independent passes, reconciled): the JNI create-path fix is correct — `payload_bytes` is `Zeroizing`, dropped immediately after the synchronous FFI call, with unwind coverage and no leaking early-returns. The #4160 rebase concern is satisfied (its merge commit is an ancestor of this head). Rust suites ran green in CI at this head. **Ready**, with one scope decision and body cleanup:

1. **P2 (scope decision) — the fetch/decrypt direction never scrubs plaintext, and the gap is in shared layers iOS inherits.** `DecryptedEncryptedDocument.payload` is a plain `Vec<u8>` (rs-platform-wallet), the FFI base64-encodes it into a plain JSON `CString` freed unzeroized by `platform_wallet_string_free` (rs-platform-wallet-ffi — the same export and free that #4194's Swift `fetchEncryptedDocuments` uses), and each host then copies into an unscrubbable `String`. The scrub discipline is currently create-only. Because the unscrubbed layers are shared Rust/FFI, one shared-layer fix (`Zeroizing` payload + a zeroizing free or a binary out-buffer instead of base64-in-JSON) covers Android and iOS at once — recommend a tracked follow-up scoped as shared-layer work rather than a per-SDK patch, with the terminal host-`String` residual documented symmetrically on both platforms.
2. The PR body still doesn't reference #4194/#4195 by number (only "stacked follow-ups" in comments) — please add.
3. Numerous residual `dashpay/platform#4091` comment references remain in this PR's new code (tx_metadata.rs, encrypted_document.rs, ffi document.rs) — the earlier sweep removed the hex tokens but not the issue-number refs; same move-to-description treatment.

shumkov:
Full-suite evidence via a same-repo dispatch run on the stack head ([29994291221](https://github.com/dashpay/platform/actions/runs/29994291221), stack + v4.1-dev merge): **the entire Rust workspace suite passed** — wire-compat vectors, size-cap, and allocator tests all executed in CI at last.

One ask: `cargo fmt --all` — this PR's new files (`tx_metadata.rs`, `crypto/mod.rs`, `lib.rs`) have real `cargo fmt --check` violations that fork CI never runs; merging unformatted reddens v4.1-dev. (All other failures in that run are pre-existing base issues — the Docker build context and `AppState.swift:185` — being handled separately.)

packages/rs-platform-wallet-ffi/src/document.rs
```diff
@@ -155,6 +270,263 @@ fn confirmed_document_to_json(document: &Document) -> Result<String, PlatformWal
...
+                        &contract_id_for_async,
+                        &document_type_str,
+                        &properties_json,
+                        signer,
+                    )
+                    .await?;
+                let json_string = confirmed_document_to_json(&confirmed)?;
+                Ok::<_, PlatformWalletError>((confirmed.id(), json_string))
+            });
+        result.map_err(PlatformWalletFFIResult::from)
```
thepastaclaw, line 404: **🟡 Suggestion: Contain panics inside the encrypted C exports**

Both PR-added encrypted operations call `block_on_worker` from plain, non-unwind `extern "C"` exports: create here and fetch at lines 477-504. `block_on_worker` can panic when the Tokio runtime cannot be built and converts a panicked worker task into another panic through the `.expect(...)` calls at `runtime.rs:38` and `runtime.rs:60`. The Android profiles intentionally retain `panic=unwind` so the outer JNI `support::guard` can translate Rust panics into Java exceptions, but a panic attempting to leave one of these nested `extern "C"` functions aborts at that boundary before the outer guard can catch it. Refactor runtime initialization and task joining to return ordinary errors, or catch and translate panics inside each C export before they reach the ABI boundary.

packages/rs-platform-wallet-ffi/src/document.rs
```diff
@@ -155,6 +270,263 @@ fn confirmed_document_to_json(document: &Document) -> Result<String, PlatformWal
...
+    // zero-length payload. It is wrapped in `Zeroizing` so the native plaintext
+    // copy is scrubbed on drop, and it is dropped explicitly the instant the
+    // encrypted properties are prepared (below) — the plaintext must NOT linger
+    // in scope across the broadcast `.await` (dashpay/platform#4091).
+    let payload_vec: Zeroizing<Vec<u8>> = Zeroizing::new(if payload_len == 0 {
+        Vec::new()
+    } else {
+        check_ptr!(payload);
+        slice::from_raw_parts(payload, payload_len).to_vec()
+    });
```
thepastaclaw, line 341: **🟡 Suggestion: Reject oversized payloads before copying or resolving the mnemonic**

The first payload-size guard currently runs inside `prepare_encrypted_txmetadata_properties`, after this export has copied all `payload_len` bytes and, for an external-signable wallet, synchronously invoked the host mnemonic resolver and constructed a master xprv at lines 357-359. An oversized request can therefore perform avoidable allocation and secret derivation, and may return a resolver error instead of the documented `TxMetadataPayloadTooLarge` error. Run the shared guard before this copy and resolver dispatch. The JNI bridge should likewise inspect the Java array length before `convert_byte_array` so it does not create its own oversized plaintext copy.

```suggestion
    unwrap_result_or_return!(
        platform_wallet::wallet::identity::crypto::tx_metadata::ensure_tx_metadata_payload_fits(
            payload_len,
        )
    );

    // Copy the payload into an owned buffer. Null is allowed only for a
    // zero-length payload. It is wrapped in `Zeroizing` so the native plaintext
    // copy is scrubbed on drop, and it is dropped explicitly the instant the
    // encrypted properties are prepared below.
    let payload_vec: Zeroizing<Vec<u8>> = Zeroizing::new(if payload_len == 0 {
        Vec::new()
    } else {
        check_ptr!(payload);
        slice::from_raw_parts(payload, payload_len).to_vec()
    });
```

packages/rs-platform-wallet-ffi/src/document.rs
```diff
@@ -570,4 +942,51 @@ mod tests {
...
+    /// An external-signable / watch-only wallet with a NULL resolver handle is
+    /// the "resolver required" error path (the on-device shape that must not
+    /// silently derive the wrong key).
+    #[test]
+    fn external_signable_wallet_null_resolver_is_resolver_required() {
+        assert_eq!(
+            decide_key_source(false, true),
+            KeySourceDecision::ResolverRequired,
+            "external-signable / watch-only wallet + null resolver must error, not derive"
+        );
```
thepastaclaw, line 990: **🟡 Suggestion: Test resolver dispatch through the encrypted-document FFI path**

These tests cover only the extracted `decide_key_source(bool, bool)` truth table. No test constructs an external-signable `PlatformWallet`, installs a live `MnemonicResolverHandle` callback, verifies the wallet ID and callback buffer contract, and drives either encrypted-document export through `tx_metadata_key_master_for_wallet`. That callback path is the Android/iOS wallet shape, so callback routing, ownership, or master-derivation regressions can pass the existing truth-table and direct-crypto tests. Add an FFI integration test using a real resolver vtable and external-signable wallet.

packages/rs-platform-wallet/tests/txmetadata_fetch.rs
```diff
@@ -0,0 +1,266 @@
...
+    let sdk = SdkBuilder::new_testnet()
+        .with_context_provider(provider)
+        .build()
+        .expect("build testnet sdk");
+    Arc::new(sdk)
+}
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+#[ignore = "hits testnet"]
+async fn fetch_returns_both_legacy_txmetadata_documents() {
```
thepastaclaw, line 59: **🟡 Suggestion: Add normal-CI coverage for fetch, pagination, and decrypt orchestration**

The only test invoking the production encrypted-document query remains `#[ignore = "hits testnet"]`, and its module documentation confirms that default CI does not run it and that it does not exercise decryption. It also returns only two documents, so it does not cross the 100-document page boundary. The independent legacy-install fixture verifies standalone key derivation and decryption, but no normal-CI test combines `StartAfter` pagination, field extraction, per-document derivation/decryption, and malformed-document skipping through the production orchestration. Extract the page-processing/query source behind a test seam or add an equivalent network-free integration harness.

packages/rs-platform-wallet/src/wallet/identity/network/encrypted_document.rs
```diff
@@ -0,0 +1,640 @@
...
+        contract_id: &Identifier,
+        document_type_name: &str,
+        since_ms: u64,
+        key_source: TxMetadataKeySource<'_>,
+    ) -> Result<Vec<DecryptedEncryptedDocument>, PlatformWalletError> {
+        use dash_sdk::platform::{ContextProvider, Fetch};
+
+        // On-device diagnostic breadcrumbs, dual-emitted at warn level (see
+        // [`breadcrumb`]): this call sits under an active `sdkFetched=0`
+        // investigation — every stage must be provably visible in `adb logcat`.
```
thepastaclaw, line 393: **💬 Nitpick: Update the obsolete WARN-level breadcrumb comment**

This comment still describes the informational breadcrumb as WARN-level output for an active `sdkFetched=0` investigation. The called `breadcrumb` function emits at DEBUG, its documentation states that the root cause is fixed, and only `breadcrumb_error` emits at WARN. Update the call-site description to match the current observability behavior.

```suggestion
        // Informational stage breadcrumbs are dual-emitted at DEBUG through
        // `breadcrumb`; genuine failures and skipped documents use
        // `breadcrumb_error` at WARN.
```

thepastaclaw (commented):

Full-current-PR verification at head 6f7abbadc184cdbcae4e8e179454d72a56596282 found no blocking or consensus-critical defect. Five non-blocking findings remain in the PR-added encrypted-document paths: C-ABI panic containment, oversized-payload validation ordering, two production-path coverage gaps, and one obsolete observability comment; the latest push is formatting-only and resolves none of them.

### Review provenance
- Sonnet reviewers: `claude-sonnet-5` — general (completed), `claude-sonnet-5` — security-auditor (completed), `claude-sonnet-5` — rust-quality (completed), `claude-sonnet-5` — ffi-engineer (completed)

🟡 4 suggestion(s) | 💬 1 nitpick(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet-ffi/src/document.rs:385-404: Contain panics inside the encrypted C exports
  Both PR-added encrypted operations call `block_on_worker` from plain, non-unwind `extern "C"` exports: create here and fetch at lines 477-504. `block_on_worker` can panic when the Tokio runtime cannot be built and converts a panicked worker task into another panic through the `.expect(...)` calls at `runtime.rs:38` and `runtime.rs:60`. The Android profiles intentionally retain `panic=unwind` so the outer JNI `support::guard` can translate Rust panics into Java exceptions, but a panic attempting to leave one of these nested `extern "C"` functions aborts at that boundary before the outer guard can catch it. Refactor runtime initialization and task joining to return ordinary errors, or catch and translate panics inside each C export before they reach the ABI boundary.
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/document.rs:331-341: Reject oversized payloads before copying or resolving the mnemonic
  The first payload-size guard currently runs inside `prepare_encrypted_txmetadata_properties`, after this export has copied all `payload_len` bytes and, for an external-signable wallet, synchronously invoked the host mnemonic resolver and constructed a master xprv at lines 357-359. An oversized request can therefore perform avoidable allocation and secret derivation, and may return a resolver error instead of the documented `TxMetadataPayloadTooLarge` error. Run the shared guard before this copy and resolver dispatch. The JNI bridge should likewise inspect the Java array length before `convert_byte_array` so it does not create its own oversized plaintext copy.
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/document.rs:946-990: Test resolver dispatch through the encrypted-document FFI path
  These tests cover only the extracted `decide_key_source(bool, bool)` truth table. No test constructs an external-signable `PlatformWallet`, installs a live `MnemonicResolverHandle` callback, verifies the wallet ID and callback buffer contract, and drives either encrypted-document export through `tx_metadata_key_master_for_wallet`. That callback path is the Android/iOS wallet shape, so callback routing, ownership, or master-derivation regressions can pass the existing truth-table and direct-crypto tests. Add an FFI integration test using a real resolver vtable and external-signable wallet.

In `packages/rs-platform-wallet/tests/txmetadata_fetch.rs`:
- [SUGGESTION] packages/rs-platform-wallet/tests/txmetadata_fetch.rs:57-59: Add normal-CI coverage for fetch, pagination, and decrypt orchestration
  The only test invoking the production encrypted-document query remains `#[ignore = "hits testnet"]`, and its module documentation confirms that default CI does not run it and that it does not exercise decryption. It also returns only two documents, so it does not cross the 100-document page boundary. The independent legacy-install fixture verifies standalone key derivation and decryption, but no normal-CI test combines `StartAfter` pagination, field extraction, per-document derivation/decryption, and malformed-document skipping through the production orchestration. Extract the page-processing/query source behind a test seam or add an equivalent network-free integration harness.
```
</details>

bfoss765:
@shumkov thepastaclaw's final review found no blocking or consensus-critical defects (only non-blocking notes) and the full Rust suite passed — looks merge-ready. One heads-up: the PastaClaw CI check still reads fail despite the clean final review; it likely just needs a re-trigger. Thanks!

shumkov:
Hi @bfoss765 — heads up before anything changes on my side, and thanks for this work.

I'm going to land this feature as four same-repository branches under `dashpay/platform` rather than merging these three PRs directly. This is about where the branches live, not about the quality of the code.

Two reasons:

1. **Writable branches.** There is a set of fixes that need to go into these layers — payload-size ordering, FFI panic containment, out-pointer contracts, allocator scoping and concurrency, log redaction, and test coverage. They belong in the layer that introduced the code, and I need branches I can commit to in order to put them there.
2. **Isolated, complete per-layer CI.** The Rust workspace and wallet jobs were skipped on all three of these PRs, so no layer here has ever had an isolated Rust run. (Swift varied: it was skipped on #4186, failed on #4195, and passed on #4194.) The only passing Rust workspace evidence anywhere in this chain is the cumulative run on #4243, which tells us the union of all four layers builds but nothing about any layer on its own. Same-repository branches get each layer its own complete run before it merges.

What that means concretely:

- Your commits are replayed with `git cherry-pick -x`, so **you remain the author of every one of them** and each carries a link back to its original SHA.
- Each successor PR body will credit and link this PR, and name the archived source head.
- Your branch is not modified in any way — I'm not pushing to it or rewriting it.
- I'll start building the replacements now so the work isn't blocked, but **I will not close this PR as superseded until the replacement is verified and you've acknowledged here.**

If you'd rather take a different route, say so and I'll stop and we'll work it out first.

bfoss765:
Acknowledged — go ahead. Same-repo branches are the right call, and this covers #4194 and #4195 too.

Your two reasons are the same ones we've hit from the other side. The fork skip isn't just inconvenient for us: the `fmt` gate lives in `Tests (macOS)`, which is gated to same-repo heads, so it has never run on any of these — which is how the violations on #4195 and #4247 reached a human instead of CI. Getting these onto same-repo branches fixes that for this stack.

Two things, both small:

- Keep authorship attribution on the commits where the code carries over.
- Ping us when the replacement branches are up — we'll review them against the originals before you close these, so the verification step has our eyes on it too.

Anything you want from us to make the port easier — rebases, splitting a layer differently — say the word.

thepastaclaw (commented):

Final verification at head 951260520c1f14fd6db45e368a30440aa949018e found no blocking or consensus-critical defect, and the latest delta's zeroization and compatibility-test changes introduce no new finding. All five prior non-blocking findings remain reproducible: two FFI robustness issues, two production-path coverage gaps, and one obsolete observability comment.

### Review provenance
- Sonnet reviewers: `claude-sonnet-5` — general (completed), `claude-sonnet-5` — security-auditor (failed), `claude-sonnet-5` — rust-quality (failed), `claude-sonnet-5` — ffi-engineer (completed), `claude-sonnet-5` — rust-quality (completed), `claude-sonnet-5` — security-auditor (completed)


_5 carried-forward finding(s) already raised on this PR; not re-posting as new inline comments._

bfoss765:
Superseded by #4277, which re-lands this PR's encrypted txMetadata document support as a verified superset (core tx_metadata.rs byte-identical; #4277 adds the index redesign on top). The two follow-ups remain live and will be rebased onto #4277's design: #4195 (Rust-side encryptionKeyIndex allocation) and #4194 (Swift wrappers — #4277 ships no Swift surface).

PR #4182, opened 2026-07-21
feat(kotlin-sdk): add maven-publish for the release AAR

bfoss765:
Adds `maven-publish` for the kotlin-sdk release AAR under the `org.dashj` group, with a JReleaser Maven Central/Sonatype deploy pipeline, a javadoc jar for Central, a JNI-presence guard that hard-fails remote publishes (and wipes `staging-deploy` before every staging publish), and a `PUBLISHING.md` runbook.

Re-opens #4045 which was auto-closed when the #3999 base branch was deleted; rebased onto `v4.1-dev`. All six original commits replayed cleanly — no hunks needed to be dropped as already-absorbed.

Verified: `:sdk:assembleRelease` builds the AAR, JReleaser/publish task graph configures, and the 175 sdk unit tests pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Android SDK can now be published to Maven Central and Sonatype snapshots.
  - Published artifacts now include source and Javadoc documentation.
  - Expanded release workflow support via automated deployment.

- **Bug Fixes**
  - Added checks to block publishing when required native JNI libraries are missing from the staged artifacts.

- **Documentation**
  - Added a publishing guide covering prerequisites, verification, signing, and how to consume the SDK from Maven Central.

---

### Provenance notes (moved out of code comments per review)

| Decision in code | Origin |
|---|---|
| Publish under the existing `org.dashj` group (`build.gradle.kts`, coordinate block) | HashEngineering's namespace call on PR #4045 |
| Javadoc jar required — Central Portal rejects non-POM components without one (`applyMavenCentralRules(true)`) | dashpay/platform#4045 |
| In-memory signing-key wiring so CI release via `ORG_GRADLE_PROJECT_signingKey` doesn't fail in `signReleasePublication` | dashpay/platform#4045 |
| `-PallowMissingJni` escape closed for remote/staging publishes | dashpay/platform#4045, finding 8b899bc2e5e8 |
| `MasternodeEntry` kept `internal` so its generated `serializer()` stays out of the public ABI (`Sdk.kt`) | dashpay/platform#4045, finding 96b7b2a236ff |

### Review-response commits
- `d517ba20bc` — POM license URL → `LICENSE.md`; tracking refs moved from comments to this table
[...]

PR #4187, opened 2026-07-21
feat(kotlin-sdk): transaction decoder JNI binding over key-wallet-ffi transaction_decode

bfoss765:
Adds the Android/Kotlin counterpart of the Swift `TransactionDecoder` (#3981): a single JNI export (`TxDecodeNative.decodeTransaction`) over key-wallet-ffi's `transaction_decode` — already present in `v4.1-dev`'s rust-dashcore pin via upstream rust-dashcore#870, so no vendored Rust changes are needed. The JNI layer marshals the decode result into a packed big-endian BLOB parsed by the new `org.dashfoundation.dashsdk.keywallet.TransactionDecoder` into `DecodedTransaction` (txid + per-input prev outpoints / unauthenticated P2PKH sender hints + per-output address/value/script).

The JNI-free `decode_to_blob` core is host-tested (7 Rust tests), including a fixture blob pinned byte-for-byte against Kotlin's `TransactionDecoderTest` (4 tests) as a cross-language layout pin.

Verified: `cargo test -p rs-unified-sdk-jni` (17 tests) and `:sdk:assembleRelease` + sdk unit tests all pass against the pin's `transaction_decode` signature (no API drift).

### DIFF
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/ffi/TxDecodeNative.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/ffi/TxDecodeNative.kt
@@ -1,0 +1,20 @@
+package org.dashfoundation.dashsdk.ffi
+
+/**
+ * Raw JNI surface for consensus transaction decoding — mirrors
+ * `rs-unified-sdk-jni/src/tx_decode.rs`.
+ */
+internal object TxDecodeNative {
+
+    /**
+     * Consensus-decode raw transaction bytes (key-wallet-ffi
+     * `transaction_decode`) into the packed big-endian BLOB documented in
+     * `tx_decode.rs` / parsed by
+     * [org.dashfoundation.dashsdk.keywallet.TransactionDecoder.parseBlob].
+     * [networkOrd] is [org.dashfoundation.dashsdk.Network.ffiValue].
+     *
+     * @throws DashSDKException on malformed bytes (including trailing
+     *   garbage after a valid transaction)
+     */
+    external fun decodeTransaction(txBytes: ByteArray, networkOrd: Int): ByteArray
+}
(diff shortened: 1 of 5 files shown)


thepastaclaw:
✅ Final review complete — no blockers (commit 40aa94fc)

shumkov (approved):

shumkov:
Reviewed against v4.1-dev with focus on FFI fidelity and Swift parity. This is a faithful counterpart of the Swift `TransactionDecoder` (#3981): same key-wallet-ffi `transaction_decode` at the same rust-dashcore pin, exact field parity, and the cross-pinned fixture blob (Rust + Kotlin) is a nice touch. Mergeable as-is. Minor, non-blocking:

- `rs-unified-sdk-jni/src/tx_decode.rs:110-116` — script encoding is half-defensive: length is always written but bytes only if ptr non-null; make the null branch write length 0 so the blob can't shift.
- `net_from_ord` is now the third private copy (also `transactions.rs`, `identity.rs`) — worth hoisting into `support.rs`.
- Follow-up idea (upstream, not this PR): `transaction_decode` exposes no version/locktime/special-tx payloads, so neither platform can render ProRegTx/AssetLock details — a key-wallet-ffi enhancement both SDKs would pick up together.

shumkov:
Post-merge follow-up only; I found no blocker in the decoder architecture.

Kotlin exposes `prevVout` as signed `Int`, so the unsigned coinbase sentinel `0xFFFFFFFF` appears as `-1`, while the Swift wrapper uses `UInt32`. Consider exposing `UInt`/`Long`, or explicitly document that Kotlin preserves the raw u32 bits in a signed value. The current test intentionally pins `-1`, so this is a host-parity/API clarity issue rather than a decoding bug.

bfoss765:
@coderabbitai full review

bfoss765:
Thanks for the review and approval. All three notes are implemented in a small follow-up PR (opening shortly): the script-encoder null branch now writes length 0 (plus the same guard one level up for the input/output arrays, with lying-fixture tests), `net_from_ord` is hoisted into `support.rs` (we found and removed a fourth copy in persistence.rs), and `prevVout` now documents the raw-u32-bits contract with a `prevVoutUnsigned` convenience accessor — the pinned `-1` test is untouched.

Separately: could you hit "Re-run failed jobs" on the x86_64 emulator check on #4182 and #4183? As you noted, the failure predates the diffs — it's the known emulator-boot flake, and we don't have rerun rights.

PR #4192, opened 2026-07-21
chore(kotlin-sdk): tx-decode follow-up — blob hardening, net_from_ord hoist, prevVout docs (#4187 review)

bfoss765:
Small follow-up promised on #4187, addressing @shumkov's three post-approval review notes on the Kotlin tx-decode JNI binding:

1. **Blob hardening** (`rs-unified-sdk-jni/src/tx_decode.rs`) — the output encoder wrote `script_pubkey_len` unconditionally but the script bytes only when the pointer was non-null, so a defensively-null pointer with a nonzero length field could shift every subsequent field in the blob. Length and bytes are now written together: a null (or empty) script encodes as length 0, never a nonzero length without its bytes. Pinned by a new unit test with a hand-built null-script fixture whose length field deliberately lies. The cross-pinned Kotlin/Rust fixture blob is unchanged.
2. **`net_from_ord` deduplication** — tx_decode.rs had added the third private copy (after transactions.rs and identity.rs). There is now a single shared `support::net_from_ord`; `rs_sdk_ffi::FFINetwork` and `platform_wallet_ffi::FFINetwork` are both re-exports of `dash_network::ffi::FFINetwork`, so the one helper serves every module. A fourth identical copy in persistence.rs was consolidated too. The exhaustive ordinal test moved to support.rs.
3. **`prevVout` signedness** — documented rather than retyped: `DecodedTransaction.Input.prevVout` stays a signed `Int` carrying the raw u32 bits (the coinbase sentinel `0xFFFFFFFF` reads as `-1`; Swift's `UInt32` shows `4294967295`), with KDoc spelling that out and a new `prevVoutUnsigned: Long` convenience view for the numeric u32 / Swift-parity rendering. The existing pinned `-1` test is untouched; assertions for the unsigned view were added alongside it.

Verified: `cargo test -p rs-unified-sdk-jni` (37 pass), `cargo clippy -p rs-unified-sdk-jni --all-targets` clean, and kotlin-sdk `:sdk:compileDebugKotlin` + `:sdk:testDebugUnitTest` green (`TransactionDecoderTest` 4/4). Each commit builds and tests green standalone.

## Summary by CodeRabbit

* **New Features**
  * Added an unsigned view of transaction input output indexes, supporting the full 32-bit range and clearly representing coinbase inputs.

* **Bug Fixes**
  * Improved transaction encoding when input or output data is unavailable or contains inconsistent lengths, preventing malformed data from causing decoding issues.
  * Standardized network handling across SDK operations, including safe fallback behavior for unknown network values.

* **Documentation**
  * Clarified how signed and unsigned transaction output indexes are represented.

### DIFF
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/keywallet/TransactionDecoder.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/keywallet/TransactionDecoder.kt
@@ -33,7 +33,16 @@ class DecodedTransaction(
          * reverse for explorer-style display (see [prevTxidDisplayHex]).
          */
         val prevTxid: ByteArray,
-        /** Previous output's index. */
+        /**
+         * Previous output's index — the raw u32 bits carried in a signed
+         * [Int]. Kotlin has no unsigned field type in this API surface, so
+         * the wire value is preserved bit-for-bit rather than range-checked:
+         * indices above `Int.MAX_VALUE` appear negative, and in particular
+         * the coinbase sentinel `0xFFFFFFFF` appears as `-1` (the Swift
+         * wrapper's `UInt32` shows the same value as `4294967295`). Compare
+         * against literal bit patterns, or use [prevVoutUnsigned] for the
+         * numeric u32 value.
+         */
         val prevVout: Int,
         /**
          * Sender address recovered from a P2PKH-shaped scriptSig; null for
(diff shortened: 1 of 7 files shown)


thepastaclaw:
🕓 Ready for review — 30 ahead in queue (commit d6d03130)
_Queue position: 31/72 · 2 reviews active_
_ETA: start ~02:55 UTC · complete ~03:23 UTC (median 28m across 30 recent reviews; 2 slots)_
_Queued 22h 42m ago · Last checked: 2026-07-22 19:40 UTC_

shumkov:
Reviewed (two independent passes) — all three #4187 follow-up items are done completely and correctly: the null-pointer/length-lie hazards are guarded at both the script and array levels with tests that hand-build lying FFI structs, the `net_from_ord` hoist actually removed a *fourth* private copy (`persistence.rs`) beyond the three flagged, with ordinal semantics preserved and test-pinned, and `prevVoutUnsigned` documents and widens the signed-bits contract properly. The fixture cross-pin is untouched, which is correct — real decode output can never produce the null-with-nonzero-length state. Nothing to add; ready.

shumkov (approved):

PR #4194, opened 2026-07-21
feat(swift-sdk): add Swift wrappers for the encrypted-txMetadata FFI exports

bfoss765:
Swift wrappers for the encrypted-txMetadata FFI surface.

## Rebased onto v4.2-dev — no longer stacked on #4186/#4195

**What changed vs. the pre-#4277 version of this PR:**

- **Base:** was stacked on #4186 → #4195. #4186 has been **superseded and closed** — QuantumExplorer's #4277 (`feat(platform-wallet): add encrypted txMetadata document support`, merge `6704a41a85`) re-landed the whole encrypted-txMetadata feature as a superset. This branch is now **a single commit directly on `v4.2-dev`**, with no borrowed commits.
- **No longer stacked on #4195.** #4195's auto-index feature landed *inside* #4277 (see below), so the auto-index export this PR binds already exists at `v4.2-dev` head. There is nothing left for this PR to stack on.
- **Diff is now exactly 2 new Swift files** (+307). Previously the branch carried the entire #4186 stack until it merged.

## What this adds

#4277 shipped the Rust core, the C ABI, the JNI bridge and the Kotlin host — but **zero Swift**. This is the iOS half, so both hosts sit on the same surface.

- `ManagedPlatformWallet.createEncryptedDocument(...)` → `platform_wallet_create_encrypted_document_with_signer_auto_index`
- `ManagedPlatformWallet.fetchEncryptedDocuments(...)` → `platform_wallet_fetch_encrypted_documents`

Both follow the existing wrapper patterns exactly: nested `withExtendedLifetime` pinning for signer/resolver across the full FFI call, byte buffers via `withUnsafeBufferPointer`, payload borrowed in place with no added Swift-side copy (the Rust-owned `Zeroizing` buffer does the scrubbing; callers should scrub their own), and the returned JSON C-string freed exactly once via `defer` registered after the result check.

`createEncryptedDocument` takes **no** `encryptionKeyIndex` parameter — it binds the auto-index export, matching Kotlin's `createEncryptedDocument`, whose `encryptionKeyIndex` defaults to `null`. The explicit-index export stays in the C ABI for migration/compat tests and is deliberately not surfaced in Swift.

Argument order, types, and the pre-dispatch `version ∈ {0, 1}` guard match the Kotlin `DocumentTransactions` counterparts; Kotlin's runtime `require`s are enforced structurally by stronger Swift types (`Identifier`, `UInt32`, `UInt64`).

[...]

PR #4195, opened 2026-07-21
[SUPERSEDED by #4277] feat(platform-wallet): allocate txMetadata encryptionKeyIndex in Rust, not the host

bfoss765:
> ## ⚠️ FULLY SUPERSEDED BY #4277 — recommend closing without merge
>
> Every layer of this PR landed in #4277 (`6704a41a85`). There is **no residual content**. Nothing has been force-pushed; the branch is left at its pre-#4277 head for the record. Details below.

## Why this is closed out

This PR moved `encryptionKeyIndex` selection off the Kotlin host and into Rust, so hosts pass only the opaque payload and can't reuse an index. QuantumExplorer's #4277 re-landed the encrypted-txMetadata feature as a superset **and folded this follow-up in**, so the guarantee ships at `v4.2-dev` today.

I verified this mechanically rather than by eye: cherry-picking this branch's three commits onto current `v4.2-dev` conflicts in all six touched files, and **every conflict on the host-facing layers is doc wording only** ("allocate" vs "generate"). The code is identical.

| Layer | This PR | On `v4.2-dev` after #4277 |
|---|---|---|
| Rust core | `prepare_encrypted_txmetadata_properties_auto_index` | ✅ present |
| C ABI | `platform_wallet_create_encrypted_document_with_signer_auto_index` (ABI-additive sibling, `Option<u32>` shared impl) | ✅ present, same signature |
| JNI | `encryptionKeyIndex == -1` sentinel, `< -1` rejected | ✅ present |
| Kotlin | `createEncryptedDocument(encryptionKeyIndex: Int? = null)` | ✅ present |
| Kotlin tests | `rejectsExplicitNegativeIndex`, `noIndexPathPassesValidation` | ✅ present, verbatim |
| Size-check-before-allocate | `reserve_next_index_checked` | ✅ **stronger** — #4277 runs `ensure_tx_metadata_payload_fits` before the native plaintext copy, not just before index selection |

## The one thing #4277 changed: the allocation mechanism

This PR allocated `1 + <count of the identity's txMetadata documents on Platform>` (matching dash-wallet's retired `1 + countAllRequests()`), serialized behind a per-wallet `EncryptionKeyIndexAllocator` mutex.

**#4277 instead generates the index with the OS CSPRNG** — `generate_encryption_key_index()` draws a non-zero 31-bit value, bounded by `MAX_ENCRYPTION_KEY_INDEX = 0x7fff_ffff` so it is always a valid BIP-32 hardened child. No Platform count, no allocator state, no extra network round trip.

I re-examined the trade-off and **#4277's choice is the better one**; I am not re-proposing the count-based allocator:

- `encryptionKeyIndex` is a per-document derivation **input** stored on the document, not a protocol sequence number. Readers derive each document's key from that document's own `{keyIndex, encryptionKeyIndex}` — nobody enumerates indices.
- The `txMetadata` document type in `wallet-utils-contract` indexes only `($ownerId, $updatedAt)` and has **no unique index** on `encryptionKeyIndex`, so a repeat is not a rejected state transition.
- Every document carries a fresh random IV and the cipher is AES-256-CBC, so a shared key across two documents is not a nonce-reuse hazard. A repeated index is non-lossy: both documents decrypt independently.
- The count-based scheme was never cross-device-safe either (two devices could seed from the same base before either's write was visible) — #4277 simply stops implying an atomicity that a client-side count can't deliver.
- Legacy `1 + countAllRequests()` parity buys nothing on the read path, since the legacy reader also uses each document's stored field.
- `MAX_ENCRYPTION_KEY_INDEX` is a hardening bound this PR did **not** have. The explicit-index path is safe in both designs — an out-of-domain index fails cleanly via `ChildNumber::from_hardened_idx`, no panic.

## Knock-on

#4194 (Swift wrappers) was stacked on this branch for the auto-index export. That export now exists on `v4.2-dev`, so **#4194 has been re-based directly onto `v4.2-dev`** and is a single self-contained commit. It is the only part of this follow-up line still outstanding — #4277 shipped no Swift at all.

No error codes were ever added by this PR, so the registry (#4261) is unaffected either way.


thepastaclaw:
🕓 Ready for review — 8 ahead in queue (commit 7d20a638)
_Queue position: 9/10 · 2 reviews active_
_ETA: start ~14:04 UTC · complete ~14:26 UTC (median 21m across 30 recent reviews; 2 slots)_
_Queued 11m ago · Last checked: 2026-08-04 12:20 UTC_

shumkov:
Reviewed (two independent passes). The allocation design is sound and legacy-compatible: `1 + Platform-count` reproduces dash-wallet's formula while fixing the restore-from-seed collision, the per-wallet mutex correctly serializes the seed fetch (race-tested with a genuinely discriminating test), offline fails loud, and the cross-device index-reuse analysis (CBC + fresh IV, readers derive from each document's stored index) is honest. Two items:

1. **Validate payload size before allocating the index.** The auto-index path performs the network count/allocation before the deterministic 4063-byte payload validation, so an oversized payload — which must fail — still consumes an index (leaving a gap). Run the size check first; it needs no network.

2. **Coordinate with #4194**: that PR currently wraps only the explicit-index export this PR deprecates (no `*_auto_index` wrapper in Swift) — see the comment there; the two should land in an agreed order so iOS doesn't re-diverge on the exact policy this PR unifies.

Minor: `encryptionKeyIndex` moved to the last positional slot in Kotlin — source-breaking for positional callers, worth a changelog line when the #4186 stack ships.

bfoss765:
Fixed: the size gate now runs *before* index allocation on the auto-index path — an oversized payload returns `TxMetadataPayloadTooLarge` without ever seeding or advancing the allocator's high-water (no consumed index, no gap), covered by a new test that asserts the owner is absent from the allocator map and the next well-sized reservation still seeds at 1. Added the release note that `encryptionKeyIndex`'s move to the last positional Kotlin slot is source-breaking for positional callers.

Sequencing per your decision: **this PR (#4195) lands first**; #4194 has been reworked onto the new auto-index export.

shumkov:
Round-3 verification (two independent passes, reconciled): the fix is verified — the payload size gate runs strictly before the allocator mutex and before any Platform seed polling on the auto-index path (and before derivation on the explicit path), all three gates share the single computed constant via one function (defense-in-depth, not drift), and the new `oversized_payload_does_not_advance_highwater` test genuinely proves the seed future is never polled and the allocator stays unseeded.

Code-ready. Two asks before merge:
- A clean Swift CI rerun — the current red is the mac-runner keychain regressing again (`keychainError(-61)` in unrelated `IdentityResolverSignIntegrationTests`), not this PR.
- Body refresh: this body doesn't mention the resolved #4194 sequencing (re-stack), and both stacked bodies still describe the old ordering. Nit: fix-round narration in a few new comments — provenance to the description.

shumkov:
Covered by the same-repo stack dispatch run ([29994291221](https://github.com/dashpay/platform/actions/runs/29994291221)) — full Rust workspace suite green, including the oversized-allocator test. No further asks beyond the earlier body-refresh notes; merge order #4195 → #4194 stands.

thepastaclaw (commented):

This PR correctly moves txMetadata encryptionKeyIndex allocation from the Kotlin host into Rust: the tokio::sync::Mutex-based allocator is properly serialized across awaits, the size-gate-before-allocation fix is sound and unit-tested, and the duplicate-index-is-safe cryptographic reasoning holds up. Remaining issues are non-blocking: the FFI boundary copies the plaintext payload before the deterministic size check runs, the authoritative Platform-count seed path lacks automated test coverage, two locks (the FFI handle-storage RwLock and the allocator's own tokio Mutex) are now held across longer network round trips, and there is a duplicated contract-fetch block plus a couple of nitpicks.

_Corrected in place after the final gate was rerun against the checkpointed Codex canonical output; three Codex-only findings already closed by the precheck verifier were retracted._

### Review provenance
- Sonnet reviewers: `claude-sonnet-5` — general (failed), `claude-sonnet-5` — security-auditor (failed), `claude-sonnet-5` — rust-quality (completed), `claude-sonnet-5` — ffi-engineer (completed), `claude-sonnet-5` — general (failed), `claude-sonnet-5` — security-auditor (failed), `claude-sonnet-5` — general (completed), `claude-sonnet-5` — security-auditor (failed), `claude-sonnet-5` — security-auditor (failed), `claude-sonnet-5` — security-auditor (completed)

🟡 7 suggestion(s) | 💬 2 nitpick(s)

### Nitpick details

- **Comment edit merged two unrelated sentences onto one run-on line** — `packages/rs-unified-sdk-jni/src/transactions.rs:928-931`
  Removing the '(dashpay/platform#4091)' reference collapsed two separate comment thoughts onto a single line: '...so these no longer need to be loud. Android visibility: `JNI_OnLoad` installs `android_logger` at'. `cargo fmt` does not reflow comments, so this run-on will persist as-is; split it back across lines for readability.
- **Explicit-index create path bypasses the allocator entirely, so the no-collision guarantee only holds when a process never mixes both entry points** — `packages/rs-platform-wallet/src/wallet/identity/network/encrypted_document.rs:401-429`
  `allocate_encryption_key_index`/`reserve_next_index_checked` only runs when the host omits the index. The explicit-index export (`platform_wallet_create_encrypted_document_with_signer`) never touches `enc_key_index_allocator`, so a caller mixing the explicit-index (migration/tests) and auto-index paths for the SAME owner in the SAME process can still race: an in-flight explicit create at index N doesn't update the in-process high-water, and a concurrent auto-index call whose Platform-count seed hasn't yet observed N's unconfirmed broadcast can allocate N again. Per the PR's own design (duplicate indices are safe since each document stores and decrypts from its own indices), this isn't exploitable for data loss or key compromise, but the advertised guarantee ('two concurrent creates through the SAME wallet process can never pick the same index') only holds when all callers in that process exclusively use the auto-index path — worth a one-line doc note on the explicit-index export making that scope limit explicit.

<details>

```

- [SUGGESTION] packages/rs-platform-wallet-ffi/src/document.rs:435-445: Oversized payload is copied into an owned buffer before the deterministic size gate rejects it
  create_encrypted_document_impl unconditionally copies `payload_len` bytes via `slice::from_raw_parts(payload, payload_len).to_vec()` (wrapped in `Zeroizing`) before any size validation runs. The network-free `ensure_tx_metadata_payload_fits` check only executes later — inside `allocate_encryption_key_index`/`reserve_next_index_checked` on the auto-index path, or inside `prepare_encrypted_txmetadata_properties` on the explicit-index path — both of which run strictly after this copy. A caller passing an oversized `payload_len` (host bug, fuzzed input, or a misbehaving JNI caller) forces a full allocation and memcopy proportional to the invalid input before the typed `TxMetadataPayloadTooLarge` error is returned; an allocation failure for a very large length aborts the process rather than surfacing the typed error. The length is known before any bytes are touched, so hoist the pure length check ahead of the copy — and mirror it in the JNI wrapper's `convert_byte_array` call so a Java caller doesn't pay for the doomed copy on its side either.
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/document.rs:454-538: Auto-index path performs a second network round trip (plus a reentrant host callback) inside the shared handle-storage read lock
  The whole `create_encrypted_document_impl` body runs inside `PLATFORM_WALLET_STORAGE.with_item(wallet_handle, move |wallet| {...})`, which holds a `parking_lot::RwLock` read guard over the entire global `HashMap<Handle, T>` (see `HandleStorage::with_item` in handle.rs) for the whole closure. Before this PR that closure did one blocking network round trip (the broadcast). Now, on the host-omitted-index path, it does a second one first — `allocate_encryption_key_index`, which itself may paginate through every existing txMetadata document for the owner on the first call — followed by a synchronous reentrant call into the host's mnemonic resolver for external-signable wallets. Any thread needing the write half of this lock for the same handle (`with_item_mut`/`remove`, e.g. during wallet teardown) now blocks for roughly double the previous worst case, plus however long the resolver callback (Keychain/biometric prompt) takes. This extends a pre-existing accepted pattern rather than introducing a new bug class, but since `Arc<PlatformWallet>` is already cheaply cloneable, cloning it and dropping the guard before any network/resolver work would remove the exposure rather than lengthen it.

In `packages/rs-platform-wallet/src/wallet/identity/network/encrypted_document.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/encrypted_document.rs:329-425: The authoritative Platform-count seed path has no automated test coverage
  `allocator_tests` thoroughly pins `next_encryption_key_index_from_count`, `reserve_next_index`, and `reserve_next_index_checked` by injecting the seed as a bare future — none of them call `count_owned_txmetadata_documents` or `allocate_encryption_key_index`, so the pagination loop, contract-fetch/register step, contract-not-found error path, raw-entry counting (including un-materialized `None` entries), and the `u32::try_from` conversion are never exercised by `cargo test`. Per the PR description this path is only covered by a manual/testnet-gated integration test. Add a mock-SDK test that drives the real first-allocation path end to end (including a multi-page and a `None`-entry case) and asserts the resulting index equals `raw_count + 1`, plus that a second allocation for the same owner reuses the cached high-water without a second query. This is genuinely new logic introduced by this PR and is exactly the kind of regression CI won't currently catch.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/encrypted_document.rs:81-97: Wallet-wide allocator mutex held across an un-timeboxed network round trip
  `reserve_next_index` acquires `allocator.lock().await` and holds it across the entire first-time seed future — which, via `allocate_encryption_key_index` → `count_owned_txmetadata_documents`, does a `DataContract::fetch` plus a paginated `Document::fetch_many` against Platform. `enc_key_index_allocator` is a single mutex shared across every owner identity in the wallet process, and rs-sdk's default `RequestSettings` leaves `timeout`/`connect_timeout` unset. A slow or unresponsive DAPI node can therefore stall the seed indefinitely, and because the lock is held for the duration, every OTHER identity's auto-index encrypted-document create in the same wallet process blocks behind it too — a single bad node turns into a wallet-wide freeze of the encrypted-document-create feature rather than a failure scoped to the identity that triggered the fetch. Scoping the lock per-owner (e.g. a `HashMap<Identifier, Mutex<u32>>` or an entry-level lock) or bounding the seed future with a timeout would contain the blast radius to the one owner actually waiting on Platform.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/encrypted_document.rs:329-354: count_owned_txmetadata_documents duplicates the fetch+register-contract block from fetch_encrypted_documents
  `count_owned_txmetadata_documents` (lines 337-349: `DataContract::fetch` → `Arc::new` → `context_provider().register_data_contract`) is nearly a line-for-line copy of the same three steps in `fetch_encrypted_documents` (lines 604-626) further down in the same `impl` block. This PR adds a third instance, in spirit, of a pattern already duplicated once — and the two existing copies already drift slightly (different 'contract not found' message text, and only one emits a `breadcrumb_error` on fetch failure). Extract a small helper, e.g. `async fn fetch_and_register_contract(&self, contract_id: &Identifier) -> Result<Arc<DataContract>, PlatformWalletError>`, to stop the duplication and the wording drift.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/encrypted_document.rs:43-44: Allocator high-water map is keyed only by owner identity, not by (owner, contract, document type)
  `EncryptionKeyIndexAllocator` is `HashMap<Identifier, u32>`, keyed solely by `owner_identity_id`. `allocate_encryption_key_index` takes `contract_id` and `document_type_name` as first-class parameters and seeds the high-water from a count scoped to that specific contract/document-type pair — but once seeded, the stored high-water is shared across ANY subsequent call for that owner regardless of `contract_id`/`document_type_name`. Today there is exactly one wallet-contract document type (`txMetadata`), so this is dormant, but the API is deliberately generalized (both the FFI exports and the Kotlin surface accept an arbitrary `document_type_name`). If this is ever reused for a second encrypted-document type on the same identity, the second type's allocation would silently continue from the first type's high-water instead of reseeding from `1 + count` for its own contract/type — producing an `encryptionKeyIndex` that no longer matches the documented `1 + count` semantics for that type.

In `packages/rs-unified-sdk-jni/src/transactions.rs`:
- [SUGGESTION] packages/rs-unified-sdk-jni/src/transactions.rs:809-818: Zeroization effort covers only the native-side plaintext copies, not the originating JVM byte[]
  `documentCreateEncrypted` wraps its native copy of the txMetadata plaintext in `Zeroizing` and drops it as soon as the FFI call returns, and `create_encrypted_document_impl` does the same for its own copy — both explicitly documented as guarding against the plaintext lingering across the broadcast await. But the actual origin of that plaintext is the Kotlin-side `payload: ByteArray` (`DocumentTransactions.kt createEncryptedDocument`), which is passed straight through to `TransactionsNative.documentCreateEncrypted`. `env.convert_byte_array(&payload)` copies out of the JVM array into the Rust `Vec`, but the JVM array itself — and any Kotlin-side protobuf-serialization buffers that produced it — remain resident in the JVM heap indefinitely, subject to ordinary GC rather than scrubbing, with no wipe performed by or requested of the Kotlin caller after the native call returns. Given how much care this PR takes to scrub the two Rust-side copies, the asymmetry is worth a one-line doc note on `createEncryptedDocument` so future auditors don't assume the whole cross-language plaintext lifecycle is scrubbed.
```
</details>

packages/rs-unified-sdk-jni/src/transactions.rs
```diff
@@ -750,6 +750,264 @@ pub extern "system" fn Java_org_dashfoundation_dashsdk_ffi_TransactionsNative_do
...
+        // explicitly there rather than left to linger (unscrubbed) to end of
+        // scope. This is the only plaintext copy in this function.
+        let payload_bytes = match env.convert_byte_array(&payload) {
+            Ok(b) => zeroize::Zeroizing::new(b),
+            Err(_) => {
+                let _ = env.exception_clear();
+                throw_sdk_exception(env, 1, "payload byte[] was null/invalid");
+                return ptr::null_mut();
+            }
+        };
```
thepastaclaw, line 839: **🟡 Suggestion: Oversized txMetadata payloads are copied twice across the JNI/FFI boundary before the size limit is enforced**

`documentCreateEncrypted` calls `env.convert_byte_array(&payload)` to copy the full JVM `byte[]` into a native `Vec` before any size check runs. The downstream FFI call (`rs-platform-wallet-ffi/src/document.rs` ~line 440-445, `create_encrypted_document_impl`) copies it a second time via `slice::from_raw_parts(...).to_vec()`. Only after both copies exist does `ensure_tx_metadata_payload_fits` (invoked from `allocate_encryption_key_index`/`prepare_encrypted_txmetadata_properties`) reject payloads over the 4063-byte limit. A caller supplying a large valid buffer (a compromised host component, or a direct C/JNI caller bypassing normal app logic) forces two full native allocations before receiving the typed `TxMetadataPayloadTooLarge`/`ErrorInvalidParameter` error, risking an OOM abort of the host process instead of a clean typed failure. Check the Java array length via `env.get_array_length` (already used elsewhere in this file, e.g. line 195) before `convert_byte_array`, and check `payload_len` against the same limit at the C FFI entry point before constructing the slice, so non-JNI callers get the same protection.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/document.rs
```diff
@@ -155,6 +266,402 @@ fn confirmed_document_to_json(document: &Document) -> Result<String, PlatformWal
...
+                        &contract_id_for_async,
+                        &document_type_str,
+                        &properties_json,
+                        signer,
+                    )
+                    .await?;
+                let json_string = confirmed_document_to_json(&confirmed)?;
+                Ok::<_, PlatformWalletError>((confirmed.id(), json_string))
+            });
+        result.map_err(PlatformWalletFFIResult::from)
```
thepastaclaw, line 541: **🟡 Suggestion: Auto-index create path extends how long the global wallet-handle-storage lock is held, now across a network round trip and a reentrant host callback**

The entire create body runs inside the closure passed to `PLATFORM_WALLET_STORAGE.with_item(wallet_handle, ...)`, which holds a `parking_lot::RwLock` read guard over the process-wide `HashMap<Handle, Arc<PlatformWallet>>` shared by every wallet handle. Before this PR, that closure performed one network round trip (the broadcast). This PR adds two more phases inside the held guard on the `None`-index path: `allocate_encryption_key_index` (a full network fetch — contract lookup plus a possibly-paginated document count on the first call per owner) and, for external-signable/watch-only wallets, a synchronous reentrant call into the host's Kotlin mnemonic resolver via `tx_metadata_key_master_for_wallet` (which may block on Keychain/Keystore or a biometric prompt). Any thread needing the write half of this lock for ANY handle (wallet dispose/create via `with_item_mut`/`insert`/`remove`) now blocks for the duration of all three phases plus the broadcast. Since `Arc<PlatformWallet>` is cheap to clone, the guard could be dropped immediately after cloning it, before any network or resolver work, removing the exposure. This extends a pre-existing accepted pattern (also present unmodified in `identity_discovery.rs`/`identity_key_preview.rs`) rather than introducing it, but it measurably lengthens the window during which unrelated wallet-lifecycle calls can stall (a potential ANR on Android).

<sub>source: ['claude']</sub>

packages/rs-platform-wallet/src/wallet/identity/network/encrypted_document.rs
```diff
@@ -0,0 +1,999 @@
...
+                let index = next_encryption_key_index_from_count(count);
+                breadcrumb(&format!(
+                    "allocate_encryption_key_index: seeded owner={owner_identity_id} \
+                     existing_count={count} next_index={index}"
+                ));
+                Ok(index)
+            },
+        )
+        .await
+    }
```
thepastaclaw, line 428: **🟡 Suggestion: The authoritative Platform-derived allocator seed path has no automated regression coverage**

`allocator_tests` (module starting ~line 842) thoroughly pins `next_encryption_key_index_from_count`, `reserve_next_index`, and `reserve_next_index_checked` — but every one of those tests injects the seed as a bare `async { Ok(N) }` future, never calling `allocate_encryption_key_index` or `count_owned_txmetadata_documents` (lines 329-354, 397-424), the code that actually fetches/registers the contract, runs the paginated owner-scoped query, and turns the raw result into `1 + count`. Per the PR description, this path is only exercised by a manual/testnet-gated integration test that is `#[ignore]`d in CI and doesn't invoke the allocator. A regression in the query construction, pagination, contract registration, or the count→index wiring (e.g. an off-by-one, a missed page, or miscounting un-materialized entries) would ship undetected. A mock-SDK test that returns a known owner-scoped count (including a multi-page case and a `None` document entry) and asserts the resulting allocated index would close this gap without a live network.

packages/rs-platform-wallet/src/wallet/identity/network/encrypted_document.rs
```diff
@@ -0,0 +1,999 @@
...
+                ))
+            })?;
+        let contract = Arc::new(contract);
+        if let Some(provider) = self.sdk.context_provider() {
+            provider.register_data_contract(Arc::clone(&contract));
+        }
+        let raw =
+            query_owned_encrypted_documents(&self.sdk, contract, owner_identity_id, document_type_name, 0)
+                .await?;
+        Ok(u32::try_from(raw.len()).unwrap_or(u32::MAX))
```
thepastaclaw, line 356: **Retracted after final-gate correction.** The Codex precheck verifier had already refuted this finding: the deployed wallet contract's `$ownerId,$updatedAt` index is neither `countable` nor `rangeCountable`, while `DocumentCount::fetch` requires a matching countable/range-countable index. Those flags are immutable after contract creation, so the suggested aggregate query would fail rather than improve this path.

packages/rs-platform-wallet/src/wallet/identity/network/encrypted_document.rs
```diff
@@ -0,0 +1,999 @@
...
+    pub key_index: u32,
+    /// The document's `encryptionKeyIndex` field (the app's per-document index).
+    pub encryption_key_index: u32,
+    /// The blob's leading version byte (0 = CBOR, 1 = protobuf).
+    pub version: u8,
+    /// $updatedAt in epoch-millis, if the document carries it. The app tracks
+    /// this as its since-timestamp high-water mark for the next fetch.
+    pub updated_at_ms: Option<u64>,
+    /// The decrypted, opaque payload bytes.
+    pub payload: Vec<u8>,
```
thepastaclaw, line 261: **Retracted after final-gate correction.** The precheck verifier correctly classified this as pre-existing, out-of-scope decrypt-side memory-hygiene debt. The result types, AES helper behavior, and fetch-side serialization copies predate this allocator PR; fixing the full ownership/zeroization pipeline belongs in a separate redesign.

packages/rs-platform-wallet/src/wallet/identity/network/encrypted_document.rs
```diff
@@ -0,0 +1,999 @@
...
+            })?;
+        let contract = Arc::new(contract);
+        if let Some(provider) = self.sdk.context_provider() {
+            provider.register_data_contract(Arc::clone(&contract));
+        }
+        let raw =
+            query_owned_encrypted_documents(&self.sdk, contract, owner_identity_id, document_type_name, 0)
+                .await?;
+        Ok(u32::try_from(raw.len()).unwrap_or(u32::MAX))
+    }
```
thepastaclaw, line 357: **🟡 Suggestion: count_owned_txmetadata_documents duplicates the fetch-and-register-contract block already present in fetch_encrypted_documents**

Lines 337-349 (`DataContract::fetch` → `Arc::new` → `context_provider().register_data_contract`) are nearly a line-for-line copy of the same three steps in `fetch_encrypted_documents` at lines 604-626 further down in the same `impl` block. The two copies have already drifted: only the `fetch_encrypted_documents` copy emits a `breadcrumb_error` on fetch failure and includes the identifying context in its 'not found' message, while the new copy in `count_owned_txmetadata_documents` has a slightly different message text and no breadcrumb. Extracting a shared `async fn fetch_and_register_contract(&self, contract_id: &Identifier) -> Result<Arc<DataContract>, PlatformWalletError>` would remove the duplication this PR adds and stop the error-message/breadcrumb wording from drifting further between the two call sites.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet/src/wallet/identity/network/encrypted_document.rs
```diff
@@ -0,0 +1,999 @@
...
+
+use super::*;
+
+/// In-process high-water map for txMetadata `encryptionKeyIndex` allocation,
+/// keyed by owner identity id → the NEXT index to hand out for that identity.
+/// Wrapped in an `Arc<tokio::sync::Mutex<..>>` so it is shared across every
+/// clone of [`IdentityWallet`] and serializes concurrent allocations (see
+/// [`reserve_next_index`] / [`IdentityWallet::allocate_encryption_key_index`]).
+pub(crate) type EncryptionKeyIndexAllocator =
+    Arc<tokio::sync::Mutex<std::collections::HashMap<Identifier, u32>>>;
```
thepastaclaw, line 44: **🟡 Suggestion: Allocator high-water map is keyed only by owner identity, not by (owner, contract, document type)**

`EncryptionKeyIndexAllocator` is `HashMap<Identifier, u32>`, keyed solely by `owner_identity_id`. `allocate_encryption_key_index` takes `contract_id` and `document_type_name` as first-class parameters and seeds the high-water from a count that IS scoped to a specific contract/document-type pair — but once seeded, the stored high-water is reused for ANY subsequent call for that owner regardless of `contract_id`/`document_type_name`. Today there is exactly one wallet-contract document type (`txMetadata`), so this is dormant. But both the FFI exports and the Kotlin surface already accept an arbitrary `document_type_name`, generalizing the API; if a second encrypted-document type is ever added for the same identity, its allocation would silently continue from the first type's high-water instead of reseeding from `1 + count` for its own contract/type, producing an index that no longer matches the documented `1 + count` semantics for that type.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet/src/wallet/identity/network/encrypted_document.rs
```diff
@@ -0,0 +1,999 @@
...
+{
+    // Hold the guard across the (first-time only) seed await: this is exactly
+    // what serializes racing allocators — a second caller that finds the map
+    // empty blocks here until the first has seeded and inserted its `next + 1`.
+    let mut guard = allocator.lock().await;
+    let next = match guard.get(owner).copied() {
+        Some(n) => n,
+        None => seed.await?,
+    };
+    guard.insert(*owner, next.saturating_add(1));
```
thepastaclaw, line 97: **🟡 Suggestion: Wallet-wide allocator mutex is held across an un-timeboxed network round trip, letting one slow node stall every identity's encrypted-document creates**

`reserve_next_index` acquires `allocator.lock().await` and holds it across the entire Platform fetch used to seed the high-water on a caller's first allocation (`seed.await?` at line 96). `enc_key_index_allocator` is a single `Arc<tokio::sync::Mutex<..>>` shared across every identity in the wallet process (cloned via `Arc::clone` in `identity_handle.rs`), and `rs-sdk`'s default `RequestSettings` sets `timeout: None` / `connect_timeout: None` (`packages/rs-sdk/src/sdk.rs:80-82`), so nothing at the SDK layer bounds this await. A DAPI node that accepts the connection but never completes the response can stall the seed indefinitely, and because the mutex is held for the duration, EVERY OTHER identity's auto-index create in the same wallet process blocks behind it too — a single unresponsive node turns into a wallet-wide freeze of the encrypted-document-create feature, not just a failure for the identity that triggered the fetch. Scoping the lock per-owner, or wrapping the seed future with a bounded timeout, would contain the blast radius to the affected identity.

<sub>source: ['claude']</sub>

packages/rs-platform-wallet/src/wallet/identity/network/encrypted_document.rs
```diff
@@ -0,0 +1,999 @@
...
+    /// [`Self::resolve_encryption_context`], resolving
+    /// `(identity, identity_index, wallet)` without crossing an `.await`. MUST
+    /// be called from a sync context — never inside an async task (`blocking_read`
+    /// panics there). Used by [`Self::prepare_encrypted_txmetadata_properties`]
+    /// so the master xprv can be wiped BEFORE any network round-trip.
+    fn resolve_encryption_context_blocking(
+        &self,
+        owner_identity_id: &Identifier,
+    ) -> Result<(dpp::identity::Identity, u32, key_wallet::wallet::Wallet), PlatformWalletError> {
+        let wm = self.wallet_manager.blocking_read();
```
thepastaclaw, line 473: **Retracted after final-gate correction.** This synchronous preparation API and its `blocking_read` implementation predate the allocator change, its rustdoc explicitly requires a synchronous context, and the production FFI call site satisfies that precondition before entering the broadcast future. It is not an in-scope defect introduced by this PR.

packages/rs-unified-sdk-jni/src/transactions.rs
```diff
@@ -750,6 +750,264 @@ pub extern "system" fn Java_org_dashfoundation_dashsdk_ffi_TransactionsNative_do
...
+        }
+        let auto_index = encryption_key_index == -1;
+        // Only 0 (CBOR) and 1 (protobuf) are wire-decodable by the legacy dashj
+        // decryptTxMetadata; anything else seals a document the legacy stack
+        // can't read. Fail fast here with the correct bound instead of the stale
+        // 0..=255 range. The Rust core `seal_tx_metadata` enforces the same
+        // invariant as the last line of defense.
+        if !(0..=1).contains(&version) {
+            throw_sdk_exception(
+                env,
```
thepastaclaw, line 818: **🟡 Suggestion: Plaintext-zeroization effort covers only the native-side copies, not the originating JVM byte[]**

`documentCreateEncrypted` wraps its native copy of the txMetadata plaintext in `Zeroizing` and drops it as soon as the FFI call returns, and `create_encrypted_document_impl` does the same for its own copy — both explicitly guard against the plaintext lingering across the broadcast await. But the actual origin of that plaintext is the Kotlin-side `payload: ByteArray`, which is passed straight through from `DocumentTransactions.createEncryptedDocument` to the native call. `env.convert_byte_array` copies OUT of the JVM array into the Rust `Vec`, but the JVM array itself — and any Kotlin-side protobuf-serialization buffers that produced it — remain resident in the JVM heap subject to ordinary GC, not scrubbing, with no wipe requested of the Kotlin caller after the native call returns. Given how carefully the two Rust-side copies are scrubbed, the JVM-side copy is the one that actually lives longest and is entirely outside Rust's control for a heap-dump/memory-scraping threat model. If that boundary is intentionally out of scope, a one-line doc note on `createEncryptedDocument` would prevent future auditors from assuming the whole cross-language plaintext lifecycle is scrubbed.

<sub>source: ['claude']</sub>

shumkov:
Re-reviewed at `4f2eb06d`. The allocator design re-verifies clean: the size gate genuinely precedes allocation (`encrypted_document.rs:125` before `:126`), the allocator mutex is held across the seed await so concurrent allocations can't collide (`concurrent_allocations_never_collide` is a real discriminating test — it yields inside the seed to widen the window), and `oversized_payload_does_not_advance_highwater` proves the doomed request leaves no gap. The documented failure-gap trade-off at `:78-82` is accurate.

**One must-fix: `cargo fmt --check --all` fails on this PR's own new code.** 7 violations — the `query_owned_encrypted_documents(...)` call in `count_owned_txmetadata_documents`, and 6 sites in `allocator_tests`. Worth knowing *why* CI didn't catch it: the fmt gate lives in the `Tests (macOS)` job, which is gated to same-repo heads, so it has never run on this PR. The full-suite evidence run we did earlier was on a mirror branch that carried an extra unpushed rustfmt commit — that's what made it green. (Five further violations in `rs-unified-sdk-jni` are pre-existing base drift, not yours; a rebase erases them.)

**Merge order is now enforced by topology:** #4186 → #4195 → #4194. #4194 already sits on this head and consumes `auto_index`, so the earlier sequencing conflict is resolved in code. Note `v4.2-dev` squash-merges, so once #4186 lands this branch needs `git rebase --onto v4.2-dev 983154de80` rather than a plain rebase — a plain one will replay 20 now-duplicated commits.

Informational, not a gate: `page.keys().last()` takes the max *Identifier* from a `BTreeMap` while the query orders by `$updatedAt`, so `StartAfter` resumes from a position that needn't match result order. Pre-existing pattern, but this PR makes the row count authoritative for index allocation, which raises the stakes above 100 documents.

PR body still says "Stacked on #4186 — do not merge before it" and describes only the pre-fix state; worth refreshing before merge.

shumkov:
Hi @bfoss765 — heads up before anything changes on my side, and thanks for this work.

I'm going to land this feature as four same-repository branches under `dashpay/platform` rather than merging these three PRs directly. This is about where the branches live, not about the quality of the code.

Two reasons:

1. **Writable branches.** There is a set of fixes that need to go into these layers — payload-size ordering, FFI panic containment, out-pointer contracts, allocator scoping and concurrency, log redaction, and test coverage. They belong in the layer that introduced the code, and I need branches I can commit to in order to put them there.
2. **Isolated, complete per-layer CI.** The Rust workspace and wallet jobs were skipped on all three of these PRs, so no layer here has ever had an isolated Rust run. (Swift varied: it was skipped on #4186, failed on #4195, and passed on #4194.) The only passing Rust workspace evidence anywhere in this chain is the cumulative run on #4243, which tells us the union of all four layers builds but nothing about any layer on its own. Same-repository branches get each layer its own complete run before it merges.

What that means concretely:

- Your commits are replayed with `git cherry-pick -x`, so **you remain the author of every one of them** and each carries a link back to its original SHA.
- Each successor PR body will credit and link this PR, and name the archived source head.
- Your branch is not modified in any way — I'm not pushing to it or rewriting it.
- I'll start building the replacements now so the work isn't blocked, but **I will not close this PR as superseded until the replacement is verified and you've acknowledged here.**

If you'd rather take a different route, say so and I'll stop and we'll work it out first.

bfoss765:
Superseded by #4277 along with #4186 — verified layer-by-layer: the auto-index Rust helper, FFI export, JNI sentinel handling, Kotlin API, and both validation tests all exist at v4.2-dev head, byte-identical or stronger (payload size check now runs before the native plaintext copy). #4277's CSPRNG index generation (bounded to a valid BIP-32 hardened child) supersedes this PR's count-based allocator by design: the field has no unique index in the contract, a fresh per-document IV makes shared keys non-hazardous, and the count scheme was not cross-device-safe. Evidence table in the PR description. The Swift surface remains in #4194, now rebased directly onto v4.2-dev.

PR #4149, opened 2026-07-20
fix(drive): enforce aggregate query limits

QuantumExplorer:
## What

- normalize caller limits against Drive defaults and maximums
- carry the effective bound into storage query construction
- reject any internal distinct execution path that reaches storage without a bound
- add default, explicit, clamp, and disabled-limit regressions

## Why

One aggregate query family did not preserve its request bound at the storage boundary. This change centralizes the effective limit and fails closed if a future caller omits it.

This covers PLATFORM-DS-CAND-386.

## Checks

- cargo test -p drive --features server query::drive_document_sum_query (29 passed)
- cargo clippy -p drive --features server --all-targets -- -D warnings
- cargo fmt --all -- --check
- git diff --check

## Summary by CodeRabbit

- **New Features**
  - Added explicit range-sum modes for aggregate results and bounded distinct results.
  - Distinct range-sum queries without proofs now apply default limits, honor requested limits, and cap results at the configured maximum.

- **Bug Fixes**
  - Improved handling of invalid or zero effective query limits with clear errors.
  - Increased reliability of SDK build and emulator test workflows through improved cleanup and retry handling.

PR #4197, opened 2026-07-22
fix(drive): reject zero effective distinct count limit

QuantumExplorer:
## What

- reject a zero effective limit on the no-proof distinct COUNT walk before it reaches storage, mirroring the sum-side `effective_no_proof_distinct_limit` policy (with a regression test)
- return `CorruptedCodeExecution` instead of `unreachable!` ahead of the distinct SUM storage walk
- strengthen the distinct-SUM limit regression so the max-clamp case actually truncates at storage (four matching values against `max_query_limit: 3`)

## Why

Review follow-ups to #4149, which merged while they were being prepared:

- A `max_query_limit: 0` config (or explicit zero limit) previously reached grovedb with a zero bound on the count path, which returns nothing and masquerades as an empty result set. The sum side already failed closed; the count side now matches.
- The `unreachable!` was provably dead, but a future early-exit in the aggregate branch would have turned it into a query-path panic rather than an error.
- The prior clamp test case matched only three distinct values, so it passed identically with or without the storage bound.

## Checks

- cargo test -p drive --features server drive_document_sum_query (30 passed)
- cargo test -p drive --features server drive_document_count_query (49 passed)
- cargo clippy -p drive --features server --all-targets
- cargo check -p drive --no-default-features --features verify
- cargo fmt --all -- --check

### DIFF
--- a/packages/rs-drive/src/query/drive_document_count_query/drive_dispatcher.rs
+++ b/packages/rs-drive/src/query/drive_document_count_query/drive_dispatcher.rs
@@ -496,10 +496,21 @@ impl Drive {
                 let effective_limit = if request.mode.is_aggregate() {
                     super::MAX_LIMIT_AS_FAILSAFE
                 } else {
-                    request
+                    let effective_limit = request
                         .limit
                         .unwrap_or(request.drive_config.default_query_limit as u32)
-                        .min(request.drive_config.max_query_limit as u32)
+                        .min(request.drive_config.max_query_limit as u32);
+                    // Fail closed instead of walking storage with a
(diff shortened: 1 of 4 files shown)


thepastaclaw:
🕓 Ready for review — 36 ahead in queue (commit 0b19b3f9)
_Queue position: 37/72 · 2 reviews active_
_ETA: start ~04:21 UTC · complete ~04:50 UTC (median 28m across 30 recent reviews; 2 slots)_
_Queued 12h 55m ago · Last checked: 2026-07-22 19:40 UTC_

PR #4181, opened 2026-07-21
fix(wallet): report authoritative L1 broadcast outcomes

llbartekll:
## Summary

- use DAPI/Core as the L1 transaction acceptance authority
- expose accepted, rejected, and unknown outcomes through Rust, FFI, and Swift
- relay through SPV and apply locally only after Core acceptance
- reconcile UTXO reservations according to definitive versus ambiguous failures
- adapt both legacy CoreTransaction and current atomic FinalizedCoreTransaction flows
- prevent the Swift Example App from reporting success without Core acceptance

## ABI note

The current v4.1-dev branch already assigns code 22 to CoreInsufficientFunds and codes 23 through 25 to asset-lock errors. TransactionBroadcastRejected therefore uses the next free stable code, 26. TransactionBroadcastUnconfirmed remains code 20.

## Tests

- platform-wallet broadcaster tests: 9 passed
- SPV runtime tests: 3 passed
- UTXO reservation tests: 2 passed
- platform-wallet-ffi broadcast tests: 5 passed
- rejected error-code mapping test: passed
- cargo fmt --all -- --check

## Summary by CodeRabbit

* **New Features**
  * Added authoritative transaction broadcast outcomes: accepted, rejected, or unknown.
  * Added dedicated handling and messaging for transactions definitively rejected by Core.
  * Added outcome-based broadcast APIs while retaining deprecated compatibility methods.
  * Improved send status messaging to distinguish successful, rejected, and uncertain broadcasts.

* **Bug Fixes**
  * Transactions now include their ID for rejected or uncertain network outcomes.
  * Uncertain broadcasts are reconciled before reporting status, reducing incorrect success or rejection results.
  * SPV relay failures no longer override confirmed Core acceptance.

PR #4198, opened 2026-07-22
fix(wallet): trustless SPV broadcast acceptance with authoritative outcomes

QuantumExplorer:
> [!NOTE]
> Supersedes #4181. Built on dashpay/rust-dashcore#913 (merged); rust-dashcore is pinned to its merge commit `70d4bf8e`.

## Problem

The wallet reported a broadcast as successful once bytes were written to an SPV peer socket. No network signal was consulted, so an invalid or dropped transaction still surfaced as "Payment sent", and UTXO reservation handling had only a coarse rejected/maybe-sent split derived from transport errors.

## Approach

Network acceptance, not a socket write, decides the outcome. The authority depends on how the wallet runs:

- **SPV wallets (production): pure P2P** — built on dashpay/rust-dashcore#913. The wallet broadcasts exclusively through dash-spv: the transaction is sent to a subset of SPV peers while withheld from the rest, and a withheld peer announcing the txid back via `inv`, an InstantSend lock, or a confirmation proves the network accepted it → `Accepted`. Trustless — no DAPI involvement in the L1 send path, and no separate relay step (dash-spv injects the transaction into its local mempool pipeline as part of the broadcast). No signal within 30s → `MaybeSent` (dash-spv keeps rebroadcasting; a later echo/IS-lock/confirmation or the reservation-TTL backstop reconciles). Failures that provably precede any send (client not started, zero connected peers — pinned against dash-spv semantics by boundary tests) → definitive `Rejected`.
- **Wallets without an SPV runtime: DAPI/Core** — unchanged from `v4.1-dev`: submission via DAPI's gRPC endpoint with every failure conservatively classified `MaybeSent`. (This path has no production call sites today, so this PR deliberately leaves it untouched.)

Outcome semantics through the stack:

- `Rejected` releases the transaction's UTXO reservation for immediate retry; `MaybeSent` keeps it (double-spend guard) — unchanged reconcile point in `wallet/reservations.rs`. Failures that provably never left the process (SPV client down, zero connected peers) release rather than lock the reservation.
- FFI: dedicated rejection code `26` (`ErrorTransactionBroadcastRejected`); the txid out-param is populated on **all** network outcomes (accepted, rejected, unknown), null only on operational errors.
- Swift: `CoreTransactionBroadcastOutcome` (accepted/rejected/unknown) returned by `broadcastTransactionWithOutcome`; the old throwing `broadcastTransaction` remains as a deprecated wrapper. `SendViewModel` applies the outcome as a pure state transition — success is only shown on acceptance, and unknown outcomes warn against retrying.

## Testing

- `cargo test -p platform-wallet`: **493 passed, 0 failed** — the SPV broadcaster maps every dash-spv verdict (accepted / uncertain / never-sent / transport error) with the reservation released only on never-sent; boundary pins catch dash-spv semantic drift on the never-sent classification.
- `cargo test -p platform-wallet-ffi`: **236 passed, 0 failed** — outcome classification carries the txid on all network outcomes; code-26 mapping round-trips.
- Swift-side view-model and error-mapping tests carried over from #4181; the full SwiftExampleApp suite (156 tests) passes on an iOS 18.6 simulator against the real FFI binary.
- Upstream, rust-dashcore#913's two-node dashd integration test exercises the echo mechanism end-to-end against dashd 23.1.0 (accepted in ~119ms via withheld-peer echo; double-spends resolve uncertain since dashd 23 sends no BIP61 rejects).

## Summary by CodeRabbit

* **New Features**
  * Added clear transaction broadcast outcomes: accepted, rejected, or unknown.
  * Added typed errors for transactions definitively rejected by Core.
  * Swift wallet APIs now expose broadcast outcomes and transaction IDs.
  * Send flow messaging now distinguishes successful, rejected, and uncertain broadcasts.

* **Bug Fixes**
  * Improved handling of ambiguous broadcasts to avoid unsafe retries.
  * Corrected transaction ID reporting for rejected and uncertain outcomes.
  * Enhanced fallback behavior when Core or SPV connectivity is unavailable.

* **Tests**
  * Expanded coverage for broadcast classification, error mapping, and user-facing send results.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1660,7 +1660,7 @@ dependencies = [
 [[package]]
 name = "dash-network"
 version = "0.45.0"
-source = "git+https://github.com/dashpay/rust-dashcore?rev=58f207e55301792f91ca318c54738541d65e1dd3#58f207e55301792f91ca318c54738541d65e1dd3"
+source = "git+https://github.com/dashpay/rust-dashcore?rev=c4625d18c2c0c42b3b8cb92894a9d9a65cd700e7#c4625d18c2c0c42b3b8cb92894a9d9a65cd700e7"
 dependencies = [
  "bincode",
  "bincode_derive",
@@ -1671,7 +1671,7 @@ dependencies = [
 [[package]]
 name = "dash-network-seeds"
 version = "0.45.0"
-source = "git+https://github.com/dashpay/rust-dashcore?rev=58f207e55301792f91ca318c54738541d65e1dd3#58f207e55301792f91ca318c54738541d65e1dd3"
+source = "git+https://github.com/dashpay/rust-dashcore?rev=c4625d18c2c0c42b3b8cb92894a9d9a65cd700e7#c4625d18c2c0c42b3b8cb92894a9d9a65cd700e7"
 dependencies = [
  "dash-network",
 ]
@@ -1748,7 +1748,7 @@ dependencies = [
 [[package]]
 name = "dash-spv"
 version = "0.45.0"
-source = "git+https://github.com/dashpay/rust-dashcore?rev=58f207e55301792f91ca318c54738541d65e1dd3#58f207e55301792f91ca318c54738541d65e1dd3"
+source = "git+https://github.com/dashpay/rust-dashcore?rev=c4625d18c2c0c42b3b8cb92894a9d9a65cd700e7#c4625d18c2c0c42b3b8cb92894a9d9a65cd700e7"
 dependencies = [
  "async-trait",
  "chrono",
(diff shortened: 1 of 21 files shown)


llbartekll:
Small heads-up: while testing #4181, I found that `DapiClientError::NoAvailableAddresses` was classified as `unknown`, which kept the UTXO reservation and caused the next retry to fail with an artificial `Insufficient funds`. Worth checking whether the same can still happen here when DAPI has no addresses and the SPV fallback doesn’t produce an acceptance signal.

thepastaclaw:
🕓 Ready for review — 58 ahead in queue (commit c82fe1d8)
_Queue position: 59/72 · 2 reviews active_
_ETA: start ~09:36 UTC · complete ~10:05 UTC (median 28m across 30 recent reviews; 2 slots)_
_Queued 3h 46m ago · Last checked: 2026-07-22 19:40 UTC_

Issue #4178, opened 2026-07-21
platform-wallet: confirmed UTXO height never mirrored back from core_transactions — asset-lock funding fails with "No UTXOs available" after restart


Claudius-Maginificent:
**TL;DR:** A UTXO that arrives unconfirmed and later confirms on-chain gets stuck marked "unconfirmed" forever in the wallet's local storage, permanently blocking Platform identity funding from that UTXO even though the funds are fully confirmed and spendable everywhere else.

## User story
As a **Dash Platform wallet user**, I want a UTXO that has genuinely confirmed on-chain to remain usable for funding my identity, so that my confirmed balance doesn't silently fail when I try to spend it on identity creation or top-up.
As a **wallet integrator (e.g. dash-evo-tool)**, I want `platform-wallet-storage` to persist UTXO confirmation state correctly across restarts, so that my app doesn't need special-case workarounds to detect and recover stranded funds.

### Base flow
A wallet receives funds to one of its addresses. It detects the transaction immediately while still in the mempool, and later observes it confirm in a block — all while the wallet process keeps running.

### Actual behavior
The wallet's local database never updates that UTXO's confirmation status past "unconfirmed," even though the transaction is fully confirmed on-chain and in the wallet's own transaction history. On the next wallet restart, the UTXO reloads as non-final.

Ordinary sends still work fine (they don't check finality), and the wallet's visible balance is correct — but funding a Platform identity from that UTXO fails outright with "No UTXOs available for selection," because identity funding requires finalized inputs. The funds are effectively stranded for identity-funding purposes until the user happens to spend the UTXO some other way (which incidentally resets it).

### Expected behavior
Once a UTXO's funding transaction confirms on-chain, the wallet's persisted state should reflect that confirmation, so the UTXO remains usable — including by finality-gated operations like identity funding — across any number of restarts.

### Root Cause

Confirmed against the pinned revs (`platform-wallet`/`platform-wallet-storage` @ `d18020f526e2a8eb1d1e868b436a7a9735795abb`; `dash-spv`/`key-wallet` @ `be6e776d69af9f31ec622898176e4b33bdc969d3`):

[...]

PR #4199, opened 2026-07-22
fix(platform): reconcile confirmed UTXO height after wallet restart

Claudius-Maginificent:
**TL;DR:** Fixes identity/asset-lock funding failing with "No UTXOs available" after a wallet restart, by making UTXO confirmation height single-sourced instead of duplicated and prone to going stale.

## User story
As a **Dash Platform wallet user**, I want a Core-funded UTXO that confirmed while the wallet was running to still be recognized as confirmed after I restart the app, so that topping up an identity or funding an asset lock doesn't spuriously fail.

### Base flow
A wallet detects an incoming transaction while it's still in the mempool, tracks it, and later sees that same transaction confirm in a block — all while the app keeps running. Separately, a UTXO can also arrive already confirmed with no transaction record of its own (a legal `apply()` input, e.g. during rehydration).

### Actual behavior
On the very next app restart, a UTXO that confirmed live — or one that was inserted already confirmed but recordless — reloads from disk as if it were still unconfirmed, even though it's genuinely confirmed on-chain, because its confirmation status was stored in more than one place that didn't reliably stay in sync. Anything requiring a final (confirmed or instant-locked) UTXO, such as an identity top-up or an asset-lock funding, then fails with "No UTXOs available" for funds that are, in fact, spendable. See dashpay/platform#4178.

### Expected behavior
A UTXO's confirmation status survives a restart intact, sourced from exactly one place, so it can never drift out of sync with reality again — regardless of whether it arrived with a full transaction record or not.

### What was done

Stacked on #3968 (`feat/platform-wallet-storage-rehydration`) — this closes a gap in that PR's own rehydration path.

`core_transactions.height` is the single authority for a UTXO's confirmation height. `core_utxos.height` is removed entirely (migration `V009`) so the fact can no longer be stored twice. To cover every legal input shape — including a UTXO that arrives with no matching transaction record — `core_transactions.record_blob` is nullable, so the table can also hold height-only rows: `apply()` upserts one per `new_utxos` entry (`Some(height)` iff confirmed, `NULL` otherwise), and that upsert only ever touches rows where `record_blob IS NULL` — a real transaction record always wins and is never overwritten by a height-only placeholder. `load_state()` / `get_tx_record()` read height from every row but only surface a `TransactionRecord` for blob-bearing rows, and cross-check each decoded record's txid/height against its own typed columns (a new `WalletStorageError::CoreTransactionEntryMismatch` on disagreement) rather than trusting the blob unconditionally.

`rs-platform-wallet`'s `core_bridge.rs` intentionally still routes confirmation-only updates through `records` rather than `new_utxos` (by design — a confirmation doesn't change UTXO topology); that file is untouched and out of scope here.

Migration `V009` backfills any pre-existing recordless confirmed UTXO's height into a height-only `core_transactions` row before dropping `core_utxos.height`, so upgrading an already-populated store can't silently lose a confirmation. Height-only writes are monotonic-max (an out-of-order or unconfirmed `new_utxos` update can never regress an already-known placeholder height; a real transaction record always overrides a placeholder regardless of height, and a suppressed placeholder write is logged naming the txid). A typed-column/blob disagreement on `core_transactions` is treated as reconstructible: the blob is authoritative, the disagreement is logged, and the typed columns are best-effort repaired, rather than aborting the whole wallet-store load. `rs-platform-wallet`'s poll loops (`asset_lock::sync::proof`, `identity::network::payments::reconcile_sent_payments`) that consume this storage layer now distinguish transient from permanent persistence errors — a permanent error is surfaced instead of retried forever.

A schema guard test (`tc031_confirmation_height_is_single_sourced_in_core_transactions`) asserts `core_utxos.height` is gone, `core_transactions.height` exists, and `core_transactions.record_blob` is nullable, so this bug class can't silently resurface.

Related: #4178.

### Testing
- `cargo test -p platform-wallet-storage --all-features --locked` — 667 passed / 0 failed, including `rt2_nonzero_balance_survives_reopen`, `load_state_restores_confirmed_recordless_utxo_height`, `transaction_record_always_overrides_height_only_placeholder`, `load_state_reconciles_utxo_height_from_confirmed_transaction_record`, `load_state_defaults_utxo_without_transaction_record_to_unconfirmed`, `height_only_placeholder_does_not_regress`, `v009_backfills_recordless_confirmed_utxo_height`, typed txid/height mismatch soft-repair tests, and the schema guard.
- `cargo clippy -p platform-wallet-storage --all-features --locked -- --no-deps -D warnings` — clean.
- `cargo fmt -p platform-wallet-storage -- --check` — clean.
- `cargo test -p platform-wallet --all-features --locked` — 647 passed / 0 failed, including `record_or_persister_or_log_retries_transient_backend_errors`, `record_or_persister_or_log_surfaces_permanent_backend_errors`, `reconcile_sent_payments_confirms_from_persisted_record`.
- `cargo clippy -p platform-wallet --all-features --locked -- --no-deps -D warnings` — clean.
- `cargo fmt -p platform-wallet -- --check` — clean.

### Breaking changes
None outside this unreleased, unmerged branch — `core_utxos.height` never shipped in any release. The schema change lands via an additive migration (`V009`) with a golden schema-freeze fingerprint bump, consistent with this crate's existing migration discipline.

### Checklist

### Prior work
- #3968 — this PR is stacked directly on top of it (base branch `feat/platform-wallet-storage-rehydration`); fixes a gap discovered in that PR's own rehydration path.
- #4178 — the issue this PR addresses (storage-layer portion only).

### Attribution


lklimek (approved):

thepastaclaw:
🕓 Ready for review — 5 ahead in queue (commit 9c27e3d4)
_Queue position: 6/11 · 2 reviews active_
_ETA: start ~14:42 UTC · complete ~15:05 UTC (median 23m across 30 recent reviews; 2 slots)_
_Queued 2h 52m ago · Last checked: 2026-07-24 13:50 UTC_

packages/rs-platform-wallet-storage/src/sqlite/schema/core_state.rs
```diff
@@ -314,27 +335,23 @@ pub fn load_state(
...
+            let confirmed = height.map(|height| height > 0).unwrap_or(false);
             let utxo = Utxo {
                 outpoint,
                 txout: dashcore::TxOut {
                     value,
                     script_pubkey: script,
                 },
                 address,
-                height: height_u32,
+                height: height.unwrap_or(0),
```
thepastaclaw, line 413: **🔴 Blocking: Recordless confirmed UTXOs lose confirmation across restart**

The single-source path fixes stale heights when `core_transactions` contains the matching record, but `apply()` still accepts a `CoreChangeSet` containing only `new_utxos`. The new writer no longer persists `utxo.height`, and these lines have no fallback when the transaction map lacks the txid, so a confirmed height-100 UTXO reloads with height 0 and `is_confirmed = false`. V009 also drops the previous fallback height from existing recordless rows. The unchanged `rt2_nonzero_balance_survives_reopen` integration test constructs exactly this legal input and, with `rehydration-apply` enabled by `--all-features`, expects the confirmed balance to remain 1,234,500; this implementation produces zero confirmed balance. This is functional, not merely cosmetic: key-wallet's transaction builder removes such UTXOs when `require_final_inputs` is enabled. Preserve confirmation in an authoritative normalized structure, retain a safe fallback, or reject and migrate/backfill confirmed UTXO-only state before removing the column.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-storage/src/sqlite/schema/core_state.rs
```diff
@@ -292,8 +292,29 @@ pub fn load_state(
...
+            blob::check_fixed_width(row.get::<_, i64>(0)?, 32, "core_transactions.txid")?;
+            let txid_bytes: Vec<u8> = row.get(1)?;
+            let txid = dashcore::Txid::from_slice(&txid_bytes)?;
+            let height = row
+                .get::<_, Option<i64>>(2)?
+                .map(|height| {
+                    crate::sqlite::util::safe_cast::i64_to_u32("core_transactions.height", height)
+                })
+                .transpose()?;
+            transaction_heights.insert(txid, height);
```
thepastaclaw: **🟡 Suggestion: Validate authoritative transaction metadata against record_blob**

The confirmation map is built from the typed `txid` and `height` columns, while the later transaction-record query decodes only `record_blob` and never compares the decoded record's txid or block height with those columns. A validly encoded but semantically inconsistent row can therefore mark a UTXO for the typed txid as confirmed while returning a transaction record for another txid or context. Since this PR makes these columns authoritative for UTXO confirmation and neighboring schema readers fail hard on typed-column/blob disagreement, decode and cross-check the corresponding `TransactionRecord` before accepting its typed confirmation metadata.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The transaction-backed reconciliation correctly fixes the stale mempool-to-confirmed path when a matching transaction record exists. However, the new schema and loader silently downgrade confirmed UTXO-only changesets and migrated recordless UTXOs to unconfirmed, contradicting an existing all-features restart test and excluding those funds from final-input transaction building. The new authoritative transaction columns should also be cross-checked against their serialized records.

### Review provenance

<details>

```

In `packages/rs-platform-wallet-storage/src/sqlite/schema/core_state.rs`:
- [BLOCKING] packages/rs-platform-wallet-storage/src/sqlite/schema/core_state.rs:340-354: Recordless confirmed UTXOs lose confirmation across restart
  The single-source path fixes stale heights when `core_transactions` contains the matching record, but `apply()` still accepts a `CoreChangeSet` containing only `new_utxos`. The new writer no longer persists `utxo.height`, and these lines have no fallback when the transaction map lacks the txid, so a confirmed height-100 UTXO reloads with height 0 and `is_confirmed = false`. V009 also drops the previous fallback height from existing recordless rows. The unchanged `rt2_nonzero_balance_survives_reopen` integration test constructs exactly this legal input and, with `rehydration-apply` enabled by `--all-features`, expects the confirmed balance to remain 1,234,500; this implementation produces zero confirmed balance. This is functional, not merely cosmetic: key-wallet's transaction builder removes such UTXOs when `require_final_inputs` is enabled. Preserve confirmation in an authoritative normalized structure, retain a safe fallback, or reject and migrate/backfill confirmed UTXO-only state before removing the column.
- [SUGGESTION] packages/rs-platform-wallet-storage/src/sqlite/schema/core_state.rs:297-315: Validate authoritative transaction metadata against record_blob
  The confirmation map is built from the typed `txid` and `height` columns, while the later transaction-record query decodes only `record_blob` and never compares the decoded record's txid or block height with those columns. A validly encoded but semantically inconsistent row can therefore mark a UTXO for the typed txid as confirmed while returning a transaction record for another txid or context. Since this PR makes these columns authoritative for UTXO confirmation and neighboring schema readers fail hard on typed-column/blob disagreement, decode and cross-check the corresponding `TransactionRecord` before accepting its typed confirmation metadata.
```
</details>

packages/rs-platform-wallet-storage/src/sqlite/schema/core_state.rs
Claudius-Maginificent: Verified against current HEAD (9000d38a0b, same commit reviewed) — still present, not a false positive.

Ran `rt2_nonzero_balance_survives_reopen` with `--all-features`: it FAILS with `left: 0, right: 1234500` — exactly the silent-zero-confirmed-balance scenario described. `core_utxos.height` was dropped by migration `V009__drop_core_utxos_height.rs`; `load_state`'s height lookup (`core_state.rs` ~L340-354) still sources height only from `transaction_heights` (built from `core_transactions`), with no fallback for a recordless confirmed UTXO. Remains unresolved — blocking.

packages/rs-platform-wallet-storage/src/sqlite/schema/core_state.rs
Claudius-Maginificent: Verified against current HEAD (9000d38a0b, same commit reviewed) — still present.

`load_state` builds `transaction_heights` from the typed `txid`/`height` columns of `core_transactions` (`core_state.rs` ~L297-317) and separately decodes `record_blob` into `TransactionRecord` (~L365-379) without cross-checking the two against each other. No change since review. Suggestion still applies — non-blocking.

packages/rs-platform-wallet-storage/src/sqlite/schema/core_state.rs
Claudius-Maginificent: Fixed in commit 3543168300 (via Codex Sol, coordinator-verified).

`core_transactions.record_blob` is now nullable, so the table can hold height-only rows. `apply()` upserts a height-only row for every `new_utxos` entry (`Some(height)` iff confirmed, `NULL` otherwise), and that upsert only ever touches rows where `record_blob IS NULL` — a real transaction record always wins and is never overwritten by a height-only placeholder. `load_state()`/`get_tx_record()` read height from every row but only emit a `TransactionRecord` for blob-bearing rows.

Verified: `rt2_nonzero_balance_survives_reopen` now passes (was failing before), plus new regression/precedence tests (`load_state_restores_confirmed_recordless_utxo_height`, `transaction_record_always_overrides_height_only_placeholder`). Full crate suite green: 295 passed / 0 failed, clippy `-D warnings` clean, fmt clean.

packages/rs-platform-wallet-storage/src/sqlite/schema/core_state.rs
Claudius-Maginificent: Fixed in commit 3543168300 (via Codex Sol, coordinator-verified).

`load_state()`/`get_tx_record()` now decode each blob-bearing `core_transactions` row and cross-check the decoded `TransactionRecord`'s txid and block-height against the row's typed `txid`/`height` columns via a new `ensure_transaction_record_matches_columns` helper, returning a new typed error (`WalletStorageError::CoreTransactionEntryMismatch`) on disagreement instead of trusting the blob unconditionally. Covered by new corruption tests for txid/height mismatch.

PR #4201, opened 2026-07-22
fix(drive): accept MMR-aligned shielded note proof ranges

llbartekll:
## Summary
- validate shielded encrypted-note proof resume offsets against the on-chain MMR chunk size instead of the multi-chunk request limit
- add a proof-based regression test for start index 2048 with an 8192-element request cap

## Context
Wallet sync rewinds its encrypted-note watermark to a 2048-note MMR chunk boundary, while one request may fetch four chunks. The verifier previously treated the 8192-element request limit as the alignment unit and rejected the valid 2048 resume point, leaving shielded sync and balance recovery stuck.

## Validation
- cargo fmt --package drive -- --check
- cargo test -p drive accepts_mmr_aligned_start_inside_multi_chunk_query_boundary --lib
- ./packages/swift-sdk/build_ios.sh --target ios --target sim
- xcodebuild dashwallet-dashpay Debug for arm64 iOS Simulator

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shielded notes verification when resuming from multi-chunk queries.
  * Resume points are now validated against the correct MMR chunk boundaries (not element-count alignment).
  * Added safety to prevent index overflow during range calculations, returning a clear corruption error when invalid.
* **Tests**
  * Added coverage for aligned resume points inside multi-chunk requests.
  * Added coverage for rejecting chunk-aligned start points near the maximum index with the expected error.

### DIFF
--- a/packages/rs-drive/src/verify/shielded/verify_shielded_encrypted_notes/v0/mod.rs
+++ b/packages/rs-drive/src/verify/shielded/verify_shielded_encrypted_notes/v0/mod.rs
@@ -1,5 +1,7 @@
 use super::VerifiedShieldedEncryptedNote;
-use crate::drive::shielded::paths::{shielded_credit_pool_path_vec, SHIELDED_NOTES_KEY};
+use crate::drive::shielded::paths::{
+    shielded_credit_pool_path_vec, SHIELDED_NOTES_CHUNK_POWER, SHIELDED_NOTES_KEY,
+};
 use crate::drive::Drive;
 use crate::error::drive::DriveError;
 use crate::error::proof::ProofError;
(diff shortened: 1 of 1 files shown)


thepastaclaw:
🕓 Ready for review — next in queue (commit 8e680ab6)
_Queue position: 1/23 · 3 reviews active_
_ETA: start ~13:15 UTC · complete ~13:35 UTC (median 20m across 30 recent reviews; 3 slots)_
_Queued 23h 42m ago · Last checked: 2026-07-23 13:10 UTC_

QuantumExplorer (approved):

PR #4165, opened 2026-07-20
fix(drive): bind and bound proof decoding

QuantumExplorer:
## Summary

- authenticate the compacted-balance predecessor before deriving the forward proof query, and require both proof halves to commit to the same root
- apply bounded, exact-consumption decoding to proof-derived vote references, compacted rows, recent per-block balance rows, and data contracts
- fail closed on the legacy adaptive single-proof payload

This addresses security review findings DS-CAND-008, 009, 037, 104, and 127 through 130.

## Compatibility

The compacted-balance proof payload is now a two-proof envelope. Producers and verifiers should be deployed together; this does not change consensus state or protobuf schema.

## Validation

- cargo check -p drive --lib
- cargo check -p drive --no-default-features --features verify
- cargo test -p drive --lib bounded_decode
- cargo test -p drive --lib verify::address_funds (24 passed, including the non-containing-predecessor fallback, mixed-root envelope rejection, and hostile/trailing recent-row decode cases)
- cargo clippy -p drive --lib --no-deps -- -D warnings
- cargo fmt --all
- git diff --check

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compacted address-balance proof verification, including queries beginning within ranges and requests beyond the final range.
  - Rejects mismatched, malformed, oversized, or trailing proof data.
  - Strengthened validation of contract, voting, and recent balance proof data to prevent invalid or unsafe decoding.

- **Reliability**
  - Android emulator authentication tests now complete PIN entry automatically.
  - Extended macOS test timeouts and made cleanup more resilient.
  - Hardened iOS build cleanup against unsafe empty paths.

PR #3792, opened 2026-06-04
fix(drive): authenticate boundary in compacted absence proofs

QuantumExplorer:
> [!CAUTION]
> ## ⚠️ DO NOT MERGE — confirmed liveness bug (found in adversarial review)
>
> This fix correctly closes the absence-proof **soundness** hole, but the implementation has a **critical liveness regression**: the chained verification uses a **descending** boundary query (greatest key `<= bound`) and an **ascending** forward query (paginated `range_from`), but a single GroveDB proof is **one-directional**. The prover emits an ascending merged proof (`prove_query_many`), so the descending boundary verification fails.
>
> **Symptom:** when **≥2 compacted ranges** sort at/below the query height — the *normal* sync case (the query handler hardcodes `limit=25` and the SDK paginates in a loop) — the honest prove→verify roundtrip fails:
> ```
> GroveDB(InvalidProof(..., "Cannot verify upper bound of queried range"))
> ```
> The original tests only exercised the **single-range** and **empty** cases (one key ≤ bound), which is why this slipped through. Two `#[ignore]`d regression tests (`multiple_ranges_below_query_height_verify`, `containing_range_with_two_lower_ranges_verifies`) now reproduce it, and the verifier carries a `# KNOWN LIVENESS BUG` note.
>
> **Planned fix (deferred):** re-key compacted entries by `(end_block, start_block)` instead of `(start_block, end_block)`. Then "ranges overlapping `[H, ∞)`" is a single **ascending** `range_from((H, 0)..)` — no boundary query, no direction conflict — and the containing range becomes an ordinary in-range result that a malicious node cannot omit, so the soundness hole is closed **structurally** as a bonus. Both the nullifier and address-balance paths need this. (Alternative considered: a two-proof response — rejected as a larger wire/API change.)
>
> The same direction bug affects the address-balance compacted path in this PR.

---
## Issue being fixed or feature implemented

**Critical proof-soundness hole (double-spend / balance-change concealment for light clients).**

The shielded-pool *compacted nullifier changes* verifier — and its twin, the *compacted address-balance changes* verifier — derived the forward-scan lower bound (`start_key`) from **raw, not-yet-authenticated proof bytes** (`extract_kv_entries_from_merk_proof`), then verified `range_from(start_key..)`.

[...]

PR #4208, opened 2026-07-23
test(drive): cover multi-range and paginated compacted balance proofs

QuantumExplorer:

The compacted address-balance proof was reworked in #4165 into a two-proof envelope (authenticated descending predecessor + separately verified forward scan, root-bound to each other). That shipped without regression coverage for the exact scenario that sank the earlier chained-single-proof attempt in #3792: **two or more compacted ranges sorting at/below the queried height**, which is the normal paginated-sync shape — the gRPC handler passes an explicit limit and the SDK pages in a loop. All existing verify tests pass `limit: None` and set up at most one compacted range below the bound.

## What was done?

Test-only. Added three honest-roundtrip regression tests to `verify_compacted_address_balance_changes/v0`:

- `multiple_ranges_below_query_height_verify` — three ranges `(1,64) (65,128) (129,192)`, query from 150: verification must succeed (the case the chained approach could not satisfy) and return only the containing range with its stored changes.
- `containing_range_with_lower_and_higher_ranges` — a fourth range `(193,256)` above the bound: the forward scan starts at the authenticated containing range and continues through later ranges in order, while wholly-lower ranges stay excluded.
- `caller_limit_paginates_across_ranges` — three `limit: Some(1)` pages resuming past the last returned range, ending on an empty page that exercises the non-containing-predecessor fallback key under a limit.

Also ported the `store_compacted_entry` test helper (direct grovedb insert, bypassing compaction) from the #3792 branch so tests can build exact tree shapes.

## How Has This Been Tested?

`cargo test -p drive --features fixtures-and-mocks --lib verify_compacted` — 10 passed, 0 failed (3 new). `cargo clippy -p drive --features fixtures-and-mocks --lib --tests` clean; `cargo fmt --all` applied.

## Breaking Changes

None — test-only.

## Checklist:

## Summary by CodeRabbit

* **Tests**
  * Added regression coverage for balance-change queries spanning multiple ranges.
  * Verified correct handling of ranges around a queried height, including predecessor fallback behavior.
  * Added paging coverage to ensure limits truncate results and subsequent pages resume correctly.


QuantumExplorer (commented):
reviewed

PR #4221, opened 2026-07-24
feat(sdk): prefer DAPI peers supporting the client's latest protocol version

QuantumExplorer:

When a client (rs-sdk / platform wallet / mobile bindings) connects to DAPI, peer selection is uniformly random over all live addresses. Nodes running older software may not support the client's newest protocol version, which can degrade behavior for version-sensitive flows. On startup the client should make a best effort to talk to peers whose software already supports (i.e. desires/votes for) the client's latest protocol version.

## What was done?

- **rs-dapi-client**: `AddressList` now tracks, per address, the highest protocol version the node reports supporting, plus an optional preferred-version target shared across clones. `get_live_address` picks randomly among live peers whose reported version is at least the preferred one, and falls back to any live peer when none match — no peer is ever excluded by version alone, and version knowledge survives unban (health and version are orthogonal).
- **rs-sdk**: new `Sdk::prefer_peers_with_latest_protocol_version()` concurrently issues an unproven `getStatus` to every live peer (4 s per-node timeout, no retries, no banning), records each node's `version.protocol.drive.latest`, and sets the preference target to `PlatformVersion::latest().protocol_version` (the client's newest supported version). Probe failures leave the peer in rotation, unranked.
- The probe is wired into `Sdk::refresh_protocol_version()` — the documented app-start / network-switch hook — so rs-sdk-ffi (Swift), the JNI bindings (Kotlin), and any caller of refresh get the behavior with no binding changes. It runs even for pinned or proofs-disabled SDKs (peer preference is orthogonal to version pinning); the proven version-ratchet path is unchanged and the unproven probe never feeds it.

## How Has This Been Tested?

- New unit tests in `rs-dapi-client` covering: preference targeting (`>=` semantics, newer peers count as preferred), best-effort fallback when no peer matches, ban filtering taking precedence over preference, version knowledge surviving unban, and preference propagation across `AddressList` clones.
- `cargo test -p rs-dapi-client` and `cargo test -p dash-sdk --lib` pass; `cargo clippy --all-targets` clean on both crates; `rs-sdk-ffi` and `wasm-sdk` (wasm32-unknown-unknown) still compile.

## Breaking Changes

None — all API additions are backward compatible, and with no preference set peer selection behaves exactly as before.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Improved peer selection by preferring live nodes that support the latest protocol version.
  * Added protocol-version preference and per-peer capability tracking, preserved across unbans.
  * Added an SDK method to proactively prefer compatible peers.
* **Bug Fixes**
  * Banned peers are excluded, with fallback to other live peers when needed.
  * Peer sampling now returns bounded, distinct, non-banned addresses.
* **Documentation**
  * Clarified bounded peer probing, fallback behavior, ignored probe failures, and blocking requirements during protocol refresh.

### DIFF
--- a/packages/rs-dapi-client/src/address_list.rs
+++ b/packages/rs-dapi-client/src/address_list.rs
@@ -3,7 +3,7 @@
 use crate::address_ban_info::AddressBanInfo;
 use crate::Uri;
 use chrono::Utc;
-use rand::{rngs::SmallRng, seq::IteratorRandom, SeedableRng};
+use rand::{rngs::SmallRng, seq::IteratorRandom, Rng, SeedableRng};
 use std::collections::hash_map::Entry;
 use std::collections::HashMap;
 use std::hash::{Hash, Hasher};
@@ -14,6 +14,14 @@ use std::sync::{Arc, RwLock};
 use std::time::Duration;
 
 const DEFAULT_BASE_BAN_PERIOD: Duration = Duration::from_secs(60);
+
+/// Share of selections (in percent) that consult the preferred-protocol tier
+/// first in [`AddressList::get_live_address`]. The remainder select uniformly
+/// from all live addresses even when preferred peers exist, so a peer's
+/// unproved, self-reported protocol version can bias routing but never
+/// exclusively capture it: every live peer keeps a nonzero selection
+/// probability.
+const PREFERRED_PROTOCOL_SELECTION_PERCENT: u32 = 90;
 
 /// DAPI address.
 #[derive(Debug, Clone, Eq)]
@@ -214,8 +222,10 @@ impl AddressList {
     /// Set (or clear with `None`) the protocol version peers should support for
     /// [`AddressList::get_live_address`] to prefer them.
     ///
-    /// The preference is best-effort: selection falls back to any live address
-    /// when no live address is known to support `version` (see
+    /// The preference is best-effort and non-exclusive: selection is only
+    /// weighted towards supporting peers
+    /// ([`PREFERRED_PROTOCOL_SELECTION_PERCENT`]%), and falls back to any
+    /// live address when no live address is known to support `version` (see
     /// [`AddressList::get_live_address`]). Protocol versions start at 1, so `0`
     /// is rejected as if `None` was passed.
     pub fn set_preferred_protocol_version(&self, version: Option<u32>) {
@@ -358,12 +368,15 @@ impl AddressList {
     ///
     /// When a preferred protocol version is set (see
     /// [`AddressList::set_preferred_protocol_version`]), selection is
-    /// best-effort biased towards it: a live address whose reported supported
-    /// protocol version (see [`AddressList::set_supported_protocol_version`])
-    /// is at least the preferred one is chosen first, and only when no such
-    /// address exists does selection fall back to a random live address —
-    /// including nodes whose version is unknown or lower. Nodes are never
-    /// excluded by version alone.
+    /// best-effort *weighted* towards it: for
+    /// [`PREFERRED_PROTOCOL_SELECTION_PERCENT`]% of selections, a live
+    /// address whose reported supported protocol version (see
+    /// [`AddressList::set_supported_protocol_version`]) is at least the
+    /// preferred one is chosen first, falling back to a random live address
+    /// when no such address exists. The remaining selections go uniformly to
+    /// any live address — including nodes whose version is unknown or lower —
+    /// so the unproved, self-reported version signal can never exclusively
+    /// capture routing. Nodes are never excluded by version alone.
     pub fn get_live_address(&self) -> Option<Address> {
         // TODO(low): module-wide `.read()/.write().unwrap()` panics on a
         // poisoned lock; adopt poison-tolerant locking consistently (SEC-003).
@@ -380,20 +393,22 @@ impl AddressList {
         };
 
         if let Some(preferred) = self.preferred_protocol_version() {
-            let preferred_address = guard
-                .iter()
-                .filter(|(_, status)| {
-                    is_live(status)
-                        && status
-                            .supported_protocol_version
-                            .map(|version| version >= preferred)
-                            .unwrap_or(false)
-                })
-                .choose(&mut rng)
-                .map(|(addr, _)| addr.clone());
+            if rng.gen_range(0..100) < PREFERRED_PROTOCOL_SELECTION_PERCENT {
+                let preferred_address = guard
+                    .iter()
+                    .filter(|(_, status)| {
+                        is_live(status)
+                            && status
+                                .supported_protocol_version
+                                .map(|version| version >= preferred)
+                                .unwrap_or(false)
+                    })
+                    .choose(&mut rng)
+                    .map(|(addr, _)| addr.clone());
 
-            if preferred_address.is_some() {
-                return preferred_address;
+                if preferred_address.is_some() {
+                    return preferred_address;
+                }
             }
         }
 
@@ -1139,11 +1154,13 @@ mod tests {
         assert_eq!(list.supported_protocol_version(&addr), None);
     }
 
-    /// With a preference set, only live addresses whose reported supported
-    /// version is at least the preferred one are selected — nodes with an
-    /// unknown or lower version are skipped while a matching node exists.
+    /// With a preference set, selection is *weighted* towards live addresses
+    /// whose reported supported version is at least the preferred one — but
+    /// never exclusive: nodes with an unknown or lower version must keep a
+    /// nonzero selection probability so an unproved self-reported version
+    /// cannot fully capture routing.
     #[test]
-    fn test_get_live_address_prefers_supporting_peers() {
+    fn test_get_live_address_prefers_supporting_peers_without_capturing_routing() {
         let mut list = AddressList::new();
         let matching: Address = "http://127.0.0.1:3000".parse().unwrap();
         let older: Address = "http://127.0.0.1:3001".parse().unwrap();
@@ -1157,10 +1174,25 @@ mod tests {
         list.set_supported_protocol_version(&older, Some(12));
         list.set_preferred_protocol_version(Some(13));
 
-        // Random selection: sample repeatedly, must always land on `matching`.
-        for _ in 0..50 {
-            assert_eq!(list.get_live_address(), Some(matching.clone()));
+        // Expected share for `matching`: 90% (weighted tier) + 10%×⅓ ≈ 93%.
+        // Bounds are generous (>5σ) so the test cannot realistically flake.
+        let draws = 2000;
+        let mut matching_hits = 0;
+        for _ in 0..draws {
+            match list.get_live_address() {
+                Some(addr) if addr == matching => matching_hits += 1,
+                Some(_) => {}
+                None => panic!("a live address must always be selectable"),
+            }
         }
+        assert!(
+            matching_hits > draws * 8 / 10,
+            "preferred peer must receive a strong majority of selections, got {matching_hits}/{draws}"
+        );
+        assert!(
+            matching_hits < draws,
+            "non-preferred peers must remain reachable — preference must not be exclusive"
+        );
     }
 
     /// A peer reporting a version *newer* than the preferred one also counts
@@ -1178,9 +1210,19 @@ mod tests {
         list.set_supported_protocol_version(&older, Some(12));
         list.set_preferred_protocol_version(Some(13));
 
-        for _ in 0..50 {
-            assert_eq!(list.get_live_address(), Some(newer.clone()));
-        }
+        // Expected share for `newer`: 90% + 10%×½ = 95%; generous bounds.
+        let draws = 2000;
+        let newer_hits = (0..draws)
+            .filter(|_| list.get_live_address() == Some(newer.clone()))
+            .count();
+        assert!(
+            newer_hits > draws * 8 / 10,
+            "newer-than-preferred peer must count as preferred, got {newer_hits}/{draws}"
+        );
+        assert!(
+            newer_hits < draws,
+            "the older peer must remain reachable — preference must not be exclusive"
+        );
     }
 
     /// Best-effort fallback: when no live peer is known to support the
@@ -1257,10 +1299,20 @@ mod tests {
         clone.set_supported_protocol_version(&older, Some(12));
         clone.set_preferred_protocol_version(Some(13));
 
+        // State written through the clone is visible through the original …
         assert_eq!(list.preferred_protocol_version(), Some(13));
-        for _ in 0..50 {
-            assert_eq!(list.get_live_address(), Some(matching.clone()));
-        }
+        assert_eq!(list.supported_protocol_version(&matching), Some(13));
+        assert_eq!(list.supported_protocol_version(&older), Some(12));
+
+        // … and drives (weighted) selection through the original handle.
+        let draws = 2000;
+        let matching_hits = (0..draws)
+            .filter(|_| list.get_live_address() == Some(matching.clone()))
+            .count();
+        assert!(
+            matching_hits > draws * 8 / 10,
+            "preference set via a clone must weight selection on the original, got {matching_hits}/{draws}"
+        );
     }
 
     /// Invariant 1 at the ladder source: the exponential ban window is
--- a/packages/rs-dapi-client/src/address_list.rs
+++ b/packages/rs-dapi-client/src/address_list.rs
@@ -440,6 +440,32 @@ impl AddressList {
             .collect()
     }
 
+    /// Randomly sample up to `count` distinct not-banned addresses.
+    ///
+    /// Uses the same liveness filter as [`Self::get_live_addresses`]. Intended
+    /// to bound best-effort maintenance fan-outs (e.g. protocol-version
+    /// probes) so they never contact every known node at once.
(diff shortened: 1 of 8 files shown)


thepastaclaw:
✅ Final review complete — no blockers (commit 3709c6c5)

QuantumExplorer:
Re CodeRabbit's nitpick about overlapping the peer probe with the proven ratchet query: declining — the sequencing is intentional. The probe sets the peer preference *before* the proven `fetch_current`, so the ratchet query itself is routed to a peer already known to support our latest protocol version (a stale peer answering the startup refresh is exactly what this feature mitigates). Overlapping the two would race the ratchet against unranked random peer selection. Since the probes fan out concurrently, the added startup latency is a single status round-trip (bounded by the 4 s probe timeout), which is acceptable for an app-start hook. Pushed 46da2945fb documenting this in the code so the ordering reads as deliberate.

QuantumExplorer (commented):
Reviewed at `5c5cb98c`. I found two mobile/runtime blockers and one binding-documentation regression; details are inline.

Validation completed:
- `cargo test -p rs-dapi-client` — 115 unit tests plus integration/doc tests passed
- `cargo test -p dash-sdk --lib` — 174 tests passed
- `git diff --check` — clean

I did not find FFI ABI, pointer-lifetime, CString ownership, or Swift memory-management regressions.

packages/rs-sdk/src/sdk.rs
```diff
@@ -403,6 +424,96 @@ impl Sdk {
...
+                self,
+                EvoNode::new(address.clone()),
+                settings,
+            )
+            .await;
+            (address, result)
+        });
+
+        let mut preferred_count = 0;
+        for (address, result) in futures::future::join_all(probes).await {
```
QuantumExplorer: [P1] Bound the startup probe fan-out

The default mainnet builder adds every Evo seed, and the seed revision pinned by this PR contains 340 mainnet Evo entries. `join_all` drives all 340 futures simultaneously, while each `EvoNode` request creates a fresh one-slot `ConnectionPool` and gRPC client. A normal app start/network switch can therefore attempt roughly 340 concurrent TLS connections, risking file-descriptor/memory exhaustion and a substantial traffic storm on mobile. Please cap/sample the peers and use bounded concurrency (with an overall time budget) rather than polling the entire list at once.

packages/rs-sdk/src/sdk.rs
```diff
@@ -374,19 +382,32 @@ impl Sdk {
...
+    /// relies on a proven query, so with proofs off there is no trusted source
+    /// to ratchet from. The peer probe still runs.
     ///
     /// Returns the SDK's protocol version number after the (possible) ratchet.
     ///
     /// [`SdkBuilder::with_version`]: SdkBuilder::with_version
     pub async fn refresh_protocol_version(&self) -> Result<u32, Error> {
+        // Best-effort: bias peer selection towards nodes that already support
+        // this client's newest protocol version. Never fails, never ratchets.
+        self.prefer_peers_with_latest_protocol_version().await;
```
QuantumExplorer, line 436: [P1] Keep this probe off Android's main dispatcher

This await is exposed through the synchronous FFI `block_on`, which joins the worker before returning. The Kotlin `suspend` wrapper calls JNI directly, and `KotlinExampleApp` launches it from a `Dispatchers.Main.immediate` scope. A slow/dead peer now holds the UI through the full 4-second probe before the proven query even starts; together they can cross the ANR window. Please move this native call to `Dispatchers.IO`/an async native API, or otherwise guarantee the refresh is not invoked on the caller's main thread.

packages/rs-sdk-ffi/src/protocol_version/queries/refresh.rs
```diff
@@ -17,9 +17,15 @@ use std::ffi::CString;
...
+/// (unproven `getStatus`, concurrent, short timeout) and biases subsequent peer
+/// selection towards nodes whose software already supports this client's latest
+/// protocol version. Probe failures are ignored and no peer is ever excluded —
+/// when no peer matches, selection falls back to any live peer.
+///
 /// For an SDK pinned to a fixed protocol version (version updating disabled)
-/// this is a no-op: no network request is made and the pinned version is
-/// returned unchanged.
+/// the proven version query is skipped and the pinned version is returned
+/// unchanged; the best-effort peer probe above still runs.
```
QuantumExplorer, line 33: [P2] Update the binding-facing API contracts too

The Rust FFI docs now correctly say a pinned refresh still performs network probes, but `packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift:506-509` still promises that pinned refresh is a no-op and makes no network request. The Kotlin/JNI public docs likewise describe only the proven ratchet and omit the unconditional probe/blocking behavior. Callers may rely on the existing no-I/O contract, so please update the public Swift/Kotlin/JNI documentation as part of this behavior change.

packages/rs-sdk/src/sdk.rs
QuantumExplorer: Bounded in f3a2e28726: the probe now samples at most 16 live peers (new `AddressList::sample_live_addresses`), runs at most 8 connections concurrently (`buffer_unordered`), and is cut off by a 6 s overall wall-clock budget (`take_until` on `transport::sleep`) on top of the 4 s per-node timeout — results that arrived in time are kept, everything else stays unranked and reachable via the fallback.

packages/rs-sdk/src/sdk.rs
QuantumExplorer: Verified the Kotlin path: `sdk.system.refreshProtocolVersion()` runs through `queryGate.op` → `TeardownGate.withOp` → `withContext(Dispatchers.IO) { JNI }`, so the blocking native call is already marshalled onto an IO worker regardless of the launching scope (KotlinExampleApp's Main.immediate scope only hosts the suspend frame) — no main-thread hold/ANR path. The real cost was worst-case duration holding an IO worker and the query gate, which the new 6 s overall probe budget (f3a2e28726) bounds; docs in PlatformQueries/QueriesNative/JNI and Swift now state the blocking + IO-marshalling contract explicitly.

packages/rs-sdk-ffi/src/protocol_version/queries/refresh.rs
QuantumExplorer: Updated in f3a2e28726: Swift `SDK.refreshProtocolVersion` no longer claims pinned refresh is a no-op — it documents the bounded peer probe, that network I/O happens even when pinned, and that the call blocks (invoke off the main thread); Kotlin `PlatformQueries.refreshProtocolVersion`, `QueriesNative`, and the JNI export docs now state the probe, pinned-mode I/O, and the Dispatchers.IO marshalling.

packages/swift-sdk/Sources/SwiftDashSDK/SDK.swift
```diff
@@ -503,10 +503,21 @@ public final class SDK: @unchecked Sendable {
...
-  /// fixed protocol version (version updating disabled) this is a no-op: no
-  /// network request is made and the pinned version is returned. Bridges
-  /// `dash_sdk_refresh_protocol_version`.
+  /// fixed protocol version (version updating disabled) the proven version
+  /// query is skipped and the pinned version is returned unchanged; the
+  /// best-effort peer probe above still runs, so this call performs network
+  /// I/O even when pinned. Bridges `dash_sdk_refresh_protocol_version`.
+  ///
+  /// This call blocks the calling thread until the probe and the proven query
+  /// complete — invoke it from a background queue/task, never the main thread.
```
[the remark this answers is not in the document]
QuantumExplorer: Applied the suggested wording in 1e9df58f10 — the call now documents blocking for the probe and, when applicable (unpinned), the proven query.

packages/rs-sdk/src/sdk.rs
```diff
@@ -403,6 +444,111 @@ impl Sdk {
...
+        // overall deadline. `take_until` keeps results that completed before
+        // the budget elapsed and drops the rest (those nodes stay unranked).
+        let mut probes = std::pin::pin!(futures::stream::iter(addresses.into_iter().map(
+            |address| async move {
+                let result = EvoNodeStatus::fetch_unproved_with_settings(
+                    self,
+                    EvoNode::new(address.clone()),
+                    settings,
+                )
+                .await;
```
thepastaclaw: **🔴 Blocking: Probe responses can unban an unrelated peer**

The targeted `EvoNode` request still runs through the ordinary `DapiClient` executor. That executor selects a live address B and later records B in `ExecutionResponse.address`, while `EvoNode::execute_transport` ignores the supplied client and creates a separate connection to the requested probe address A. A successful response from A therefore invokes `update_address_ban_status` for B in both the DAPI-client and SDK retry layers. `ban_failed_address: false` suppresses failure-driven bans but does not suppress success-driven unbanning, so the probe can clear B's expired ban ladder or lift a ban placed concurrently while A was being queried. Execute targeted probes through a path that reports A as the response address, or disable health-state mutation for these maintenance requests.

<sub>source: ['codex']</sub>

packages/rs-sdk/src/sdk.rs
```diff
@@ -403,6 +444,111 @@ impl Sdk {
...
+                        target: "dash_sdk::protocol_version",
+                        %address,
+                        %error,
+                        "peer protocol-version probe failed; leaving peer unranked"
+                    );
+                    None
+                }
+            };
+
+            address_list.set_supported_protocol_version(&address, supported);
```
thepastaclaw, line 608: **🔴 Blocking: Clear stale versions for probes dropped by the total deadline**

Only futures yielded before `take_until` reach `set_supported_protocol_version`. If a peer already has `Some(target)` from an earlier refresh and its new probe is dropped at the six-second deadline, the old value remains and the peer is still treated as preferred even though the current pass learned nothing. This is especially reachable because the second concurrency batch can start after the four-second per-node timeouts and then has less than two seconds before the total deadline. Clear sampled addresses before launching the pass or explicitly clear every unfinished sampled address after the stream ends.

<sub>source: ['codex']</sub>

packages/rs-dapi-client/src/address_list.rs
```diff
@@ -306,14 +372,34 @@ impl AddressList {
...
+                            .supported_protocol_version
+                            .map(|version| version >= preferred)
+                            .unwrap_or(false)
+                })
+                .choose(&mut rng)
+                .map(|(addr, _)| addr.clone());
+
+            if preferred_address.is_some() {
+                return preferred_address;
+            }
```
thepastaclaw, line 412: **🔴 Blocking: Do not let an unproved capability report become an exclusive routing gate**

The version stored for each peer comes directly from its unproved, self-asserted `getStatus` response, but this branch turns that assertion into an exclusive eligibility filter whenever at least one matching peer exists. A malicious sampled peer can claim `version >= preferred` and become the only selected endpoint when no other sampled peer reports the target. The ranking has no expiry and survives unban, while non-retryable gRPC responses such as `InvalidArgument` and `PermissionDenied` do not ban the peer; the attacker can therefore remain selected and persistently block proven refreshes, queries, and state-transition submissions while observing the client's DAPI traffic. Use this untrusted signal as a non-exclusive weight, preserving a nonzero selection probability for other live peers, or require independently authenticated capability information before filtering exclusively.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The prior fan-out and binding-documentation issues are fixed, and the Android call path was already marshalled through Dispatchers.IO. Three blocking issues remain: targeted probes can mutate an unrelated peer's ban state, deadline-dropped probes preserve stale rankings, and an unproved self-reported capability can become an exclusive and persistent routing gate.

### Review provenance

🔴 3 blocking

<details>

```

- [BLOCKING] packages/rs-sdk/src/sdk.rs:501-506: Probe responses can unban an unrelated peer
  The targeted `EvoNode` request still runs through the ordinary `DapiClient` executor. That executor selects a live address B and later records B in `ExecutionResponse.address`, while `EvoNode::execute_transport` ignores the supplied client and creates a separate connection to the requested probe address A. A successful response from A therefore invokes `update_address_ban_status` for B in both the DAPI-client and SDK retry layers. `ban_failed_address: false` suppresses failure-driven bans but does not suppress success-driven unbanning, so the probe can clear B's expired ban ladder or lift a ban placed concurrently while A was being queried. Execute targeted probes through a path that reports A as the response address, or disable health-state mutation for these maintenance requests.
- [BLOCKING] packages/rs-sdk/src/sdk.rs:496-535: Clear stale versions for probes dropped by the total deadline
  Only futures yielded before `take_until` reach `set_supported_protocol_version`. If a peer already has `Some(target)` from an earlier refresh and its new probe is dropped at the six-second deadline, the old value remains and the peer is still treated as preferred even though the current pass learned nothing. This is especially reachable because the second concurrency batch can start after the four-second per-node timeouts and then has less than two seconds before the total deadline. Clear sampled addresses before launching the pass or explicitly clear every unfinished sampled address after the stream ends.

In `packages/rs-dapi-client/src/address_list.rs`:
- [BLOCKING] packages/rs-dapi-client/src/address_list.rs:382-397: Do not let an unproved capability report become an exclusive routing gate
  The version stored for each peer comes directly from its unproved, self-asserted `getStatus` response, but this branch turns that assertion into an exclusive eligibility filter whenever at least one matching peer exists. A malicious sampled peer can claim `version >= preferred` and become the only selected endpoint when no other sampled peer reports the target. The ranking has no expiry and survives unban, while non-retryable gRPC responses such as `InvalidArgument` and `PermissionDenied` do not ban the peer; the attacker can therefore remain selected and persistently block proven refreshes, queries, and state-transition submissions while observing the client's DAPI traffic. Use this untrusted signal as a non-exclusive weight, preserving a nonzero selection probability for other live peers, or require independently authenticated capability information before filtering exclusively.
```
</details>

packages/rs-sdk/src/sdk.rs
QuantumExplorer: Fixed in 6dbe886b6a — confirmed real (a peer with an expired ban window but nonzero ban_count is selectable, and a probe success attributed to it resets its ladder). The probe now bypasses the executor entirely: it opens its own transport client to the target address and parses via `FromUnproved`, so no request-routing or ban-state machinery runs at all.

packages/rs-sdk/src/sdk.rs
QuantumExplorer: Fixed in 6dbe886b6a — the pass now clears the recorded supported version of every sampled peer before launching the probes, so a deadline-dropped or failed probe consistently leaves its peer unranked; only fresh results from the current pass set rankings.

packages/rs-dapi-client/src/address_list.rs
QuantumExplorer: Fixed in 6dbe886b6a using the suggested non-exclusive weight: 90% of selections consult the preferred tier, the remaining 10% select uniformly over all live peers, so every live peer keeps a nonzero selection probability and an unproved self-report can bias but never capture routing. Combined with the per-pass ranking reset (rankings now live only until the next refresh pass), a false claim cannot persist as the sole endpoint. Selection tests updated to statistical bounds asserting both the majority weighting and the reachability of non-preferred peers.

thepastaclaw (approved):

At exact head 6dbe886b6a4ad4d207917a896515f3ac3583e7c0, combined Codex checkpoint and successful Sonnet general, security, and FFI evidence found no remaining in-scope issues. Source inspection confirms that all three prior blockers are fixed: probes bypass health-mutating DAPI execution, sampled rankings are cleared before probing, and protocol preference is weighted rather than exclusive.

### Review provenance
- Sonnet reviewers: `claude-sonnet-5` — general (completed), `claude-sonnet-5` — security-auditor (failed), `claude-sonnet-5` — ffi-engineer (completed), `claude-sonnet-5` — security-auditor (completed)

QuantumExplorer:
Applied CodeRabbit's cancellation nitpick in 3709c6c5ac: the probe's `take_until` now stops on whichever comes first — the 6 s budget or `Sdk::cancelled()` — so a `shutdown()` issued during `refresh_protocol_version` is no longer delayed by the remaining probe budget. Verified: dash-sdk unit tests (181) pass and wasm32-unknown-unknown still builds.

thepastaclaw (approved):

The only substantive change since the last clean review (6dbe886b) is commit 3709c6c5, a 14-line diff to packages/rs-sdk/src/sdk.rs that adds a futures::future::select cancellation arm to the peer protocol-version probe's take_until, so Sdk::shutdown() now stops the probe fan-out immediately instead of waiting out the full PEER_VERSION_PROBE_TOTAL_BUDGET. I re-verified this directly: cargo check and cargo clippy --lib -- -D warnings both pass cleanly for dash-sdk, the cancel_token/cancelled()/shutdown() wiring is correct and consistent with its documented contract, and the change touches no consensus-critical, proof, or FFI-exposed surface. No agent raised any in-scope blocking or suggestion findings, and none were fabricated here — the PR remains clean.

### Review provenance
- Sonnet reviewers: `claude-sonnet-5` — general (completed), `claude-sonnet-5` — security-auditor (failed), `claude-sonnet-5` — ffi-engineer (completed), `claude-sonnet-5` — security-auditor (failed), `claude-sonnet-5` — security-auditor (completed)

QuantumExplorer:
Not actually sure we should do this.

PR #4222, opened 2026-07-24
fix(platform): harden v12 to v13 protocol upgrades

shumkov:

Three defects on the v12-to-v13 upgrade path. The first was observed on testnet; the other two were found reviewing it.

**1. Consensus divergence at the v13 activation, decided by cache history.** `transition_to_version_13` rewrites the persisted DPNS definition with a direct `apply_contract`. That reaches state outside the drive operation batch and so carries none of its cache invalidation — nothing evicts the superseded definition from `DriveCache::data_contracts`, which is the cache the batch document transformer resolves contracts through. A validator that had read DPNS once kept serializing domain documents against its cached pre-migration copy; a validator that had restarted read the migrated one. The same transition serialized to different bytes, giving a different Merkle root and a different app hash. The split followed restart/cache history, not node role.

**2. Speculative system-contract loading leaked across protocol versions.** `reload_system_contracts` kept one "current" materialization per contract in an `ArcSwap`. It runs on the first block of a protocol change — speculatively, for a candidate block that may be rejected — and that swap has no part in the grovedb rollback which follows. There is no correct moment to move such a pointer. Consensus requests are processed sequentially, but that is not enough to save it: a rejected v13 candidate leaves the pointer on v13 while the next sequential block executes at v12. And the two versions are genuinely live at once, because `check_tx` runs on its own ABCI connection and thread pool and reaches the same cache through `validate_identity_public_keys_contract_bounds` -> `find_by_id`, validating against the last committed version while the upgrade block executes at the candidate one — and the upgrade block itself must read the candidate's materialization, since `add_history_operations` runs exactly there. Advancing on reload is wrong after a rejection; advancing on commit is wrong during the upgrade block.

**3. Genesis on a fresh node registered protocol-version-1 contracts** regardless of the version the chain was born at. `Drive::open` seeded from `PlatformVersion::first()`, and the production entry point `Platform::open` passes no initial protocol version while a fresh node has nothing committed to reload from. Mainnet and testnet were born at protocol version 1, where the materializations coincide, so neither is affected.

The repository also lacked a reusable, release-tag-pinned workflow for reviewing and rehearsing protocol upgrades end to end.

## What was done?

**Ordinary data contract cache (defect 1).** Added `Drive::refresh_data_contract_cache_from_state`, called at the end of `perform_events_on_first_block_of_protocol_change_v0` for every system contract a migration can rewrite. It reads the contract as state now holds it **bypassing both caches**, evicts the superseded copy from both tiers, and seeds the result into the **block** cache — which transactional reads consult first, which is promoted to global on commit, and which is dropped if the candidate never commits.

Eviction alone is not sufficient, and this is the subtlety worth reviewing closely. A transactional lookup falls back to the global cache on a block-cache miss (`cache/data_contract.rs:47`); read-only queries read committed state with no transaction and populate the global cache (`query/document_query/v0/mod.rs:173`); and the batch document transformer passes `add_to_cache_if_pulled = false` (`batch/transformer/v0/mod.rs:420`), so it never populates the block cache and re-runs that lookup on every read. A query landing between the eviction and the transformer's read puts the pre-change definition straight back, and the block serializes against it — a divergence decided by query timing rather than by state.

That in turn required an **ordering change**: `clear_drive_block_cache` moved out of `run_block_proposal_v0` and up into `run_block_proposal`, above the protocol-change branch, so the order is now **clear -> migrate and seed -> execute**. Previously the clear ran *after* the migration and would wipe the seed before any state transition read it. The only other behavioural difference is that a proposal rejected by the app-version or height pre-checks now also clears the block cache, which is scratch state cleared at the start of every proposal regardless.

**System contract cache (defects 2 and 3).** Replaced the per-contract `ArcSwap` "current" plus its reload with a single memo of the pure function `load_system_data_contract(variant, platform_version)`: one `BTreeMap<(ProtocolVersion, SystemDataContract), Arc<DataContract>>` published copy-on-write through one `ArcSwap`, filled lazily. Reads are lock-free; the map only changes when a protocol version is first read from.

Every read is now pinned to the `PlatformVersion` the caller is already executing under. That is the only correct notion of "current" — it is a local, it rolls back with the block that proposed it, and it needs no commit hook. Keying entries by protocol version makes a candidate's materializations unreachable from a committed-version read, so a rejected candidate is inert rather than something requiring an undo. `reload_system_contracts` and `load_genesis_system_contracts` are deleted; genesis now materializes at the version it is actually starting at.

Retention is exact in steady state and bounded unconditionally: `drop_versions_below(committed)` runs when a block commits, releasing the outgoing version at the moment it becomes garbage, with a hard cap of two versions as a backstop. A chain replayed from genesis therefore holds one set of materializations between upgrades and two only while a switch block executes.

The DocumentHistory pin to protocol version 13 is dropped. It existed only because the eager reload materialized every contract at every version including below its activation; it was also wrong on a direct 12-to-14 jump, where `transition_to_version_13` persists the v14 materialization while a v13-pinned cache would serve the v13 one. Protocol version 13 survives as DocumentHistory's activation version only.

**Review workflow.** Added a shared protocol-upgrade review skill, checklist, and comparison helper. Release certification requires immutable tags; mutable development refs are explicitly pre-release evidence only.

The review also confirmed that PoSe-banned validators may retain an earlier protocol vote. That behavior is intentionally unchanged here: v12 and v13 both dispatch the existing v0 methods, so changing persisted vote writes during this rolling release could make mixed binaries compute different roots. It needs a future protocol-versioned method change.

## How Has This Been Tested?

All passing on `5a2e2f7880`:

- `cargo test -p drive --lib` — 3246 passed
- `cargo test -p drive-abci --lib` — 2627 passed. Two `data_contract_create::tests::keywords` failures are pre-existing and reproduce on the unmodified base commit: `DataContract`'s manual `Serialize` reads a process-global `PlatformVersion::get_current()` that parallel tests overwrite.
- `cargo test -p drive cache::system_contracts --lib` — 5 passed
- `cargo test -p drive-abci execution::platform_events::protocol_upgrade --lib` — 43 passed
- `cargo test -p drive-abci --test strategy_tests run_chain_v12_to_v13_locks_in_before_activation` — 1 passed
- `cargo test -p drive-abci --lib batch::tests::document::dpns::dpns_username_transfer_tests` — 11 passed
- `cargo test -p drive-abci --lib run_block_proposal::tests` — 3 passed
- `cargo clippy -p drive -p drive-abci --all-targets` clean, `cargo fmt --all --check` clean, `git diff --check` clean

Each cache regression was verified red before the fix and green after, by disabling the specific mechanism it pins:

| test | red before |
|---|---|
| `warm_dpns_contract_cache_must_not_survive_the_committed_v13_activation` | warm node resolved DPNS version 1 (`keeps_transfer_history = false`), cold node version 2 |
| `concurrent_committed_query_must_not_divert_the_activation_block_from_migrated_dpns` | with eviction only and no block-cache seed, the activation block resolved DPNS version 1 after a committed-state query repopulated the global cache |
| `warm_dpns_contract_cache_survives_a_rejected_v13_activation_and_its_retry` | after the retry committed, the warm node still resolved DPNS version 1 |
| `reloading_v13_must_not_make_dpns_v2_visible_to_v12_reads` | against the old API, `reload_system_contracts(v13)` then `load_dpns()` returned the v13 contract while the committed version was v12 |
| `should_seed_migrated_dpns_into_the_block_cache_through_run_block_proposal` | with the clear moved back into `run_block_proposal_v0`, the activation left an empty block cache |
| `should_clear_the_block_cache_before_rejecting_a_wrong_height_proposal` | with the clear moved back below validation, a rejected proposal left uncleared scratch behind |

### Known gaps

- **The live three-validator rehearsal predates this commit.** It rolled published v4.0.0 images to v4.1.0-beta.2 one validator at a time without stopping Core, locked at height 82796 (`current=12,next=13`), activated at 82912, and reached 82949 with every node at `current=13,next=13`, `catchingUp=false`, app hash `c6845c8ed89d015f954d05ffbe01f9d7529834f4384415bbc81409cafcddb674`. Because this commit changes block-execution ordering, that run no longer certifies the current head and should be repeated.
- A delayed-writer window remains open and is not specific to migrations: a query thread that reads a contract from committed state, is descheduled, and performs its cache insert after a block that rewrote that contract has committed can clobber the new definition in the global cache. Closing it needs a generation token bumped on `merge_and_clear_block_cache`; contract version does not work as a validity token, because `strip_unknown_document_schema_properties` rewrites contract bytes without incrementing it.
- Three live gates remain skipped: two-identity DPNS fixture with post-upgrade transfer, validator restart during the lock-in epoch, and a rejected populated activation candidate through a full public proposal round. Their deterministic contract, restart, cache, rollback, and retry equivalents pass.

## Breaking Changes

No wire-format, state-format, or consensus-format change.

`drive`'s public Rust API changed: `SystemDataContracts::reload_system_contracts` and `load_genesis_system_contracts` are removed; every `load_*` accessor now takes `&PlatformVersion` and returns `Result<Arc<DataContract>>`; `find_by_id` takes `&PlatformVersion` and returns `Result<Option<Arc<DataContract>>>` so that a materialization failure surfaces instead of being reinterpreted as contract-not-found. All in-tree callers are updated.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added support for protocol-version-aware system contract handling during upgrades.
  * Improved contract availability and cache refresh behavior across activation, restart, retry, and rollback scenarios.
  * Added automated protocol-upgrade review and rehearsal guidance.

* **Bug Fixes**
  * Corrected contract and document-history behavior during protocol transitions.
  * Prevented stale or speculative data from affecting committed state.

* **Tests**
  * Expanded coverage for upgrade lock-in, activation, persistence, retries, restarts, and post-upgrade functionality.


thepastaclaw:
⛔ Blockers found — Sonnet deferred (commit 2dd27f82)

thepastaclaw (commented):

This PR fixes a real bug where find_by_id() explicit protocol-version reads could observe a speculative reload_system_contracts() materialization for a different protocol version; the versioned BTreeMap cache and new regression tests (upgrade_fork_tests.rs, dpns.rs) correctly demonstrate and close that gap. Two moderate-confidence, non-blocking correctness gaps remain in the new find_by_id lazy-fill path: it doesn't replicate the DocumentHistory activation-version override used by reload_system_contracts, and it silently swallows load_system_data_contract errors as a plain not-found. Neither is exploitable at the current max protocol version (13), so this is a COMMENT rather than a blocker.

### Review provenance

🟡 2 suggestion(s)

<details>

```

In `packages/rs-drive/src/cache/system_contracts.rs`:
- [SUGGESTION] packages/rs-drive/src/cache/system_contracts.rs:260-294: find_by_id lazy-fill doesn't honor the DocumentHistory activation-version override
  reload_system_contracts (lines ~124-135) special-cases SystemDataContract::DocumentHistory: no matter which platform_version the reload is invoked with, the contract is always materialized using PlatformVersion::get(DOCUMENT_HISTORY_ACTIVATION_PROTOCOL_VERSION) (pinned to 13), because its schema uses index-aggregation keywords (`averageable`, `rangeCountable`, ...) that only that exact meta-schema recognizes.

  The new lazy-fill path in find_by_id doesn't apply this override: when a protocol version isn't yet present in contracts_by_protocol_version, line 288 materializes the missing entry with load_system_data_contract(system_contract, platform_version), where platform_version is derived directly from the requested protocol_version (line 260), not from DOCUMENT_HISTORY_ACTIVATION_PROTOCOL_VERSION.

  This is unobservable today because DOCUMENT_HISTORY_ACTIVATION_PROTOCOL_VERSION == 13 == LATEST_PLATFORM_VERSION, so the requested version and the forced override version always coincide. Once protocol v14+ ships (and per the doc comment this contract's schema is not expected to change again), any node that reaches the lazy-fill path for DocumentHistory at v14+ (e.g. a node that only reloaded its current version and is later asked for an older explicit version) will materialize the contract using the requested version's meta-schema instead of the pinned v13 one the normal reload path forces. Two nodes populating the same cache key through the two different code paths (reload vs. lazy-fill) could then diverge on what that key holds, which is exactly the footgun this PR's caching mechanism is meant to eliminate.
- [SUGGESTION] packages/rs-drive/src/cache/system_contracts.rs:288: find_by_id silently treats a load_system_data_contract error as "not a system contract"
  Before this PR, find_by_id could not fail: it returned Arc::clone(&active.load()) for any id matching a known system contract's static id. With the new lazy-fill path, once the id is confirmed to match a known SystemDataContract variant and the activation-version check has passed, a failure from load_system_data_contract(...) (e.g. a bug in a future protocol version's contract definition, or a PlatformVersionError) is swallowed via `.ok()?` and the function returns None.

  The caller, get_system_or_user_contract_with_fee_v0, treats None as "this id is not a cached system contract" and falls through to the normal fee-charged GroveDB contract fetch, which will most likely also fail to find a system contract id, producing a confusing NotFound instead of surfacing the real underlying load error. Distinguish "id doesn't match any system contract" from "id matches but materialization failed" (e.g. log the swallowed error, or return a Result) so a genuine internal defect isn't silently reinterpreted as an ordinary contract-not-found case.
```
</details>

packages/rs-drive/src/cache/system_contracts.rs
```diff
@@ -217,26 +257,154 @@ impl SystemDataContracts {
...
+            return Some(contract);
+        }
+
+        let contract = Arc::new(load_system_data_contract(system_contract, platform_version).ok()?);
+        let mut contracts_by_protocol_version = active.contracts_by_protocol_version.write();
+        Some(Arc::clone(
+            contracts_by_protocol_version
+                .entry(protocol_version)
+                .or_insert(contract),
+        ))
```
thepastaclaw: **🟡 Suggestion: find_by_id lazy-fill doesn't honor the DocumentHistory activation-version override**

reload_system_contracts (lines ~124-135) special-cases SystemDataContract::DocumentHistory: no matter which platform_version the reload is invoked with, the contract is always materialized using PlatformVersion::get(DOCUMENT_HISTORY_ACTIVATION_PROTOCOL_VERSION) (pinned to 13), because its schema uses index-aggregation keywords (`averageable`, `rangeCountable`, ...) that only that exact meta-schema recognizes.

The new lazy-fill path in find_by_id doesn't apply this override: when a protocol version isn't yet present in contracts_by_protocol_version, line 288 materializes the missing entry with load_system_data_contract(system_contract, platform_version), where platform_version is derived directly from the requested protocol_version (line 260), not from DOCUMENT_HISTORY_ACTIVATION_PROTOCOL_VERSION.

This is unobservable today because DOCUMENT_HISTORY_ACTIVATION_PROTOCOL_VERSION == 13 == LATEST_PLATFORM_VERSION, so the requested version and the forced override version always coincide. Once protocol v14+ ships (and per the doc comment this contract's schema is not expected to change again), any node that reaches the lazy-fill path for DocumentHistory at v14+ (e.g. a node that only reloaded its current version and is later asked for an older explicit version) will materialize the contract using the requested version's meta-schema instead of the pinned v13 one the normal reload path forces. Two nodes populating the same cache key through the two different code paths (reload vs. lazy-fill) could then diverge on what that key holds, which is exactly the footgun this PR's caching mechanism is meant to eliminate.

```suggestion
        let load_platform_version = if system_contract == SystemDataContract::DocumentHistory {
            PlatformVersion::get(DOCUMENT_HISTORY_ACTIVATION_PROTOCOL_VERSION)
                .map_err(ProtocolError::from)
                .ok()?
        } else {
            platform_version
        };
        let contract = Arc::new(load_system_data_contract(system_contract, load_platform_version).ok()?);
```

<sub>source: ['claude']</sub>

packages/rs-drive/src/cache/system_contracts.rs
```diff
@@ -217,26 +257,154 @@ impl SystemDataContracts {
...
         };
         if active.active_since_protocol_version > protocol_version {
             return None;
         }
-        Some(Arc::clone(&active.load()))
+        if let Some(contract) = active.load_for_protocol_version(protocol_version) {
+            return Some(contract);
+        }
+
+        let contract = Arc::new(load_system_data_contract(system_contract, platform_version).ok()?);
```
thepastaclaw: **🟡 Suggestion: find_by_id silently treats a load_system_data_contract error as "not a system contract"**

Before this PR, find_by_id could not fail: it returned Arc::clone(&active.load()) for any id matching a known system contract's static id. With the new lazy-fill path, once the id is confirmed to match a known SystemDataContract variant and the activation-version check has passed, a failure from load_system_data_contract(...) (e.g. a bug in a future protocol version's contract definition, or a PlatformVersionError) is swallowed via `.ok()?` and the function returns None.

The caller, get_system_or_user_contract_with_fee_v0, treats None as "this id is not a cached system contract" and falls through to the normal fee-charged GroveDB contract fetch, which will most likely also fail to find a system contract id, producing a confusing NotFound instead of surfacing the real underlying load error. Distinguish "id doesn't match any system contract" from "id matches but materialization failed" (e.g. log the swallowed error, or return a Result) so a genuine internal defect isn't silently reinterpreted as an ordinary contract-not-found case.

```suggestion
        let contract = match load_system_data_contract(system_contract, platform_version) {
            Ok(contract) => Arc::new(contract),
            Err(e) => {
                tracing::error!(
                    ?system_contract,
                    protocol_version,
                    error = ?e,
                    "failed to lazily materialize system contract for protocol version"
                );
                return None;
            }
        };
```

<sub>source: ['claude']</sub>

packages/rs-drive/src/cache/system_contracts.rs
shumkov: The divergence you describe — two code paths populating the same cache key differently — is now structurally impossible. `reload_system_contracts` is gone, and there is a single materialization path (`SystemDataContracts::load`), lazily filled and keyed by `(protocol version, variant)`.

I deliberately did **not** carry the v13 pin into it, for two reasons.

The pin only existed because the eager reload materialized every contract at every version, including below its activation: `reload_system_contracts(v12)` would otherwise try to build DocumentHistory under a meta-schema that rejects its index-aggregation keywords. With lazy materialization the contract is only ever built when someone asks for it at a version where it is active, so the reason for the pin is gone.

It was also wrong in the direction you are worried about, just later. `transition_to_version_13` persists DocumentHistory materialized at the **activating** platform version, and a node can jump 12 → 14 directly (the guard is `previous < 13 && new >= 13`). In that case state holds the v14 materialization while a v13-pinned cache would serve the v13 one — the cache disagreeing with state, which is the more dangerous of the two failures. Materializing at the version being executed at keeps them in agreement.

The invariant this rests on is now stated in the code and in `docs/SYSTEM_CONTRACT_CACHE_SPEC.md`: if a system contract's materialization changes at protocol version N, the first block of that change must rewrite the persisted contract at N — which is what `transition_to_version_13` does for both DPNS and DocumentHistory.

Protocol version 13 survives only as DocumentHistory's activation version, in `CACHED_SYSTEM_CONTRACTS`, where it gates `find_by_id` into returning `None` below v13 (pinned by `document_history_cache_respects_its_activation_version`).

packages/rs-drive/src/cache/system_contracts.rs
shumkov: Addressed, and taken further than the suggestion. `find_by_id` now returns `Result<Option<Arc<DataContract>>, Error>` and propagates a materialization failure rather than logging it and returning `None`:

```rust
if active_since_protocol_version > platform_version.protocol_version {
    return Ok(None);
}

self.load(system_contract, platform_version).map(Some)
```

There is no `.ok()?` left anywhere on this path — the `PlatformVersion::get` call that needed one is gone too, since the method now takes `&PlatformVersion` directly instead of a bare protocol version.

So `None` now means only what you wanted it to mean: the id doesn't match a contract this cache materializes, or the contract isn't active at that protocol version yet. A genuine defect surfaces as an error out of `get_system_or_user_contract_with_fee_v0` instead of being reinterpreted as an ordinary contract-not-found.

packages/rs-drive/src/drive/contract/refresh_cache/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+
+        // Drop the superseded copy from both caches before re-seeding.
+        self.cache.data_contracts.remove(contract_id);
+
+        // A contract that is not in state has nothing to cache — leaving both caches empty is
+        // then the correct outcome, and the next reader will fetch and find it absent.
+        if let Some(fetch_info) = maybe_fetch_info {
+            self.cache
+                .data_contracts
+                .insert(fetch_info, transaction.is_some());
```
thepastaclaw, line 60: **🔴 Blocking: A delayed committed query can restore stale DPNS after cache promotion**

Removing the old global entry and seeding the migrated definition into the block cache does not prevent an already-running query from publishing an older result afterward. A query can miss the global cache, fetch committed v12 DPNS before the upgrade transaction commits, and be descheduled before its insertion. The upgrade then removes the old entry, seeds v13 DPNS into the block cache, and promotes that entry to the global cache during finalization; when the query resumes, `DataContractCache::insert` unconditionally overwrites the promoted entry with v12 DPNS. The next block clears the v13 block copy, allowing transactional contract resolution to fall back to stale v12 DPNS and serialize documents differently from a cold validator. Prevent stale publications with an atomic generation/version-aware insertion rule or equivalent synchronization covering the fetch-to-insert interval; the current tests only exercise queries that finish before promotion.

<sub>source: ['codex']</sub>

packages/rs-drive-abci/src/execution/platform_events/protocol_upgrade/perform_events_on_first_block_of_protocol_change/v0/mod.rs
```diff
@@ -118,6 +132,30 @@ impl<C> Platform<C> {
...
+        //
+        // This runs after `clear_drive_block_cache`, which the block start already performed —
+        // seeding before that would be wiped before anything read it.
+        for system_contract in SYSTEM_CONTRACTS_REWRITABLE_BY_A_PROTOCOL_CHANGE {
+            self.drive.refresh_data_contract_cache_from_state(
+                system_contract.id().to_buffer(),
+                Some(transaction),
+                platform_version,
+            )?;
+        }
```
thepastaclaw: **🔴 Blocking: Root-affecting upgrade cache behavior changes without a feature-version bump**

The new refresh changes `perform_events_on_first_block_of_protocol_change_v0`, while the companion change moves block-cache clearing out of `run_block_proposal_v0` and ahead of protocol-change events in the unversioned wrapper. These changes can determine which DPNS schema serializes a transition and therefore affect the app hash, yet protocol v13 still selects `run_block_proposal: 0` and `perform_events_on_first_block_of_protocol_change: Some(0)`. The same feature-version identifiers now describe observably different consensus implementations, and historical v0 behavior is no longer preserved. Keep the old v0 paths byte-for-byte, introduce separately versioned implementations for the revised lifecycle, and activate them only in the protocol version intended to carry this fix—v13 if it is still unreleased on the target network, otherwise a later protocol version.

<sub>source: ['codex']</sub>

packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/batch/tests/document/dpns.rs
```diff
@@ -1441,6 +1452,713 @@ mod dpns_username_transfer_tests {
...
+    /// had warmed that cache before the activation would keep serializing domain documents
+    /// against the pre-upgrade definition while a validator that restarted reads the new one
+    /// from the state. The two would write different bytes for the same transition and diverge
+    /// on the app hash.
+    ///
+    /// Two platforms are driven into byte-identical populated v12 state, the activation is
+    /// committed on both, and the only difference between them is that one had DPNS warmed into
+    /// its contract cache beforehand.
+    /// The block info the protocol version 13 activation is run at in the cache-coherence
+    /// tests below.
```
thepastaclaw: **💬 Nitpick: Warm-cache scenario documentation is attached to the block-info helper**

The paragraph describing the warm-versus-cold DPNS divergence is currently rustdoc for `v13_activation_block_info`, although it documents `warm_dpns_contract_cache_must_not_survive_the_committed_v13_activation`. Leave only the block-info description above the helper and move the scenario paragraph immediately above that test.

thepastaclaw (changes requested):

The protocol-keyed system-contract cache now isolates speculative materializations correctly, propagates lazy-load errors, and remains consistent with persisted state for direct 12-to-14-or-later upgrades; the prior error finding is fixed and the DocumentHistory pin finding is outdated. Two blocking issues remain: a delayed committed-state query can overwrite the promoted DPNS contract with its stale pre-upgrade copy, and the root-affecting cache lifecycle changes were added to existing v0 consensus paths without a feature-version bump. A minor test documentation comment is also attached to the wrong helper.

### Review provenance

🔴 2 blocking | 💬 1 nitpick(s)

<details>

```

In `packages/rs-drive/src/drive/contract/refresh_cache/mod.rs`:
- [BLOCKING] packages/rs-drive/src/drive/contract/refresh_cache/mod.rs:52-60: A delayed committed query can restore stale DPNS after cache promotion
  Removing the old global entry and seeding the migrated definition into the block cache does not prevent an already-running query from publishing an older result afterward. A query can miss the global cache, fetch committed v12 DPNS before the upgrade transaction commits, and be descheduled before its insertion. The upgrade then removes the old entry, seeds v13 DPNS into the block cache, and promotes that entry to the global cache during finalization; when the query resumes, `DataContractCache::insert` unconditionally overwrites the promoted entry with v12 DPNS. The next block clears the v13 block copy, allowing transactional contract resolution to fall back to stale v12 DPNS and serialize documents differently from a cold validator. Prevent stale publications with an atomic generation/version-aware insertion rule or equivalent synchronization covering the fetch-to-insert interval; the current tests only exercise queries that finish before promotion.

In `packages/rs-drive-abci/src/execution/platform_events/protocol_upgrade/perform_events_on_first_block_of_protocol_change/v0/mod.rs`:
- [BLOCKING] packages/rs-drive-abci/src/execution/platform_events/protocol_upgrade/perform_events_on_first_block_of_protocol_change/v0/mod.rs:135-157: Root-affecting upgrade cache behavior changes without a feature-version bump
  The new refresh changes `perform_events_on_first_block_of_protocol_change_v0`, while the companion change moves block-cache clearing out of `run_block_proposal_v0` and ahead of protocol-change events in the unversioned wrapper. These changes can determine which DPNS schema serializes a transition and therefore affect the app hash, yet protocol v13 still selects `run_block_proposal: 0` and `perform_events_on_first_block_of_protocol_change: Some(0)`. The same feature-version identifiers now describe observably different consensus implementations, and historical v0 behavior is no longer preserved. Keep the old v0 paths byte-for-byte, introduce separately versioned implementations for the revised lifecycle, and activate them only in the protocol version intended to carry this fix—v13 if it is still unreleased on the target network, otherwise a later protocol version.
```
</details>

PR #4220, opened 2026-07-23
docs(kotlin-sdk): maven-central environment infra runbook (infra half of #4193)

bfoss765:
## What this is

The **infrastructure/admin half** of enabling Maven Central publishing for `org.dashj:dash-sdk-android`. It adds a step-by-step runbook (`packages/kotlin-sdk/MAVEN_CENTRAL_INFRA_SETUP.md`) for creating the `maven-central` GitHub Environment and scoping the publishing secrets to it.

The **code half** is #4193, which restructured `.github/workflows/kotlin-sdk-release.yml` so every publishing secret is referenced only inside the `environment: maven-central` job. That boundary is only airtight once the secrets are actually scoped to this environment and this repository's access to the repo/org copies is removed — which is what this runbook covers.

> This PR is documentation only; it changes no code. It exists to (a) give the infra team a single, reviewable, linkable set of instructions and (b) keep the runbook versioned next to [`PUBLISHING.md`](https://github.com/dashpay/platform/blob/v4.1-dev/packages/kotlin-sdk/PUBLISHING.md).

## What the infra team needs to do (summary — full detail in the added file)

All in **`dashpay/platform` → Settings → Environments** (needs org-owner / repo-admin):

1. **Create an Environment named exactly `maven-central`** — must match the `environment:` value in the workflow verbatim.
2. **Add Required reviewers** (the approval gate), and ideally a **tag policy `kotlin-sdk-v*`**.
3. **Add these five secrets as *environment* secrets on `maven-central`, then delete the repo-scoped copies and revoke `dashpay/platform`'s access to any org-scoped copies of the same names** (delete an org secret outright only if no other repository uses it; otherwise just exclude `dashpay/platform` from its access policy):
   - `JRELEASER_MAVENCENTRAL_SONATYPE_USERNAME`
   - `JRELEASER_MAVENCENTRAL_SONATYPE_PASSWORD`
   - `JRELEASER_GPG_SECRET_KEY`
   - `JRELEASER_GPG_PASSPHRASE`
   - `JRELEASER_GPG_PUBLIC_KEY`
4. **Verify** a `kotlin-sdk-v*` tag run pauses at the deploy job for approval and publishes only after approval.

**Removing this repository's access in step 3 is the crux** — as long as a repo-scoped copy of these names exists, *or* an org-scoped copy remains available to `dashpay/platform`, an unprotected job can still read the credentials and the security finding behind #4193 is not closed.

### Transitional note
[...]

PR #4193, opened 2026-07-21
ci(kotlin-sdk): couple Maven Central deploy to the kotlin-sdk-v* release tag (stacked on #4182)

bfoss765:
**Stacked on #4182 (Maven publishing) — merge after it.** The branch includes #4182's commits until it lands; the last commits are this PR's content.

Follow-up to #4182, addressing @shumkov's review point: nothing tied `-PsdkVersion` on the manual Maven publish to the `kotlin-sdk-*` tag that drives the GitHub-release AAR, so the two channels could ship different commits under one version.

This folds `jreleaserDeploy` into `kotlin-sdk-release.yml`:

- **Version comes only from the tag.** `kotlin-sdk-vX.Y.Z` → `-PsdkVersion=X.Y.Z`; the checkout is already pinned to the same tag, so the Maven Central artifact and the GitHub-release AAR are always the same commit under the same version. There is no version input anywhere in the workflow. Malformed `kotlin-sdk-v*` tags and tag-derived `-SNAPSHOT` versions fail the run (early, before the native build); non-versioned `kotlin-sdk-*` tags keep today's AAR-only behavior.
- **Secrets-gated deploy.** The Central Portal token for `org.dashj` is currently personal, so the deploy steps skip with a notice when the secrets (`JRELEASER_MAVENCENTRAL_SONATYPE_*`, `JRELEASER_GPG_*`) aren't configured — the tag still produces the GitHub-release AAR, and the documented manual runbook (now explicitly "from the tag's commit with the tag's version") remains the fallback. A partially configured secret set fails hard instead of guessing.
- **Existing guards run in CI.** `cleanStagingDeploy`, `verifyJniLibsForRemotePublish` and `verifyStagedAarForRemotePublish` are wired via task dependencies and appear in the `--dry-run` graph for both the staging and deploy invocations used by the workflow.
- **`workflow_dispatch` stays, tag-derived.** It only accepts an existing tag name (re-run/emergency path); the version is derived from that tag exactly as on tag push.

`PUBLISHING.md` documents the CI path as preferred and demotes the manual runbook to a fallback.

Verified: workflow parses (`yaml.safe_load`); tag→version derivation exercised against valid/invalid/snapshot/injection-shaped tags; both Gradle invocations dry-run with the guards in the task graph (JDK 17).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
[...]

PR #4228, opened 2026-07-24
ci: release Kotlin and Swift SDKs with the platform release

shumkov:

The monorepo versions and releases everything under a single `vX.Y.Z` tag, but the Kotlin SDK (`kotlin-sdk-*` tags) and Swift FFI (`ffi-*` tags) had independent release workflows in their own tag namespaces. Neither was ever used — no such tag was ever pushed and nothing is published to Maven Central — while the actual SDK releases should carry the platform version like every other artifact.

## What was done?

Publishing the platform GitHub release for `vX.Y.Z[-pre.N]` now also:

- builds and attaches **`dash-sdk-android-<version>.aar`** to that same release and publishes **`org.dashj:dash-sdk-android:<version>`** to Maven Central,
- builds and attaches **`DashSDKFFI-<version>.xcframework.zip`** + its SwiftPM checksum,

via two new reusable workflows (`release-kotlin-sdk.yml`, `release-swift-sdk.yml`) called from `release.yml`, following the existing `release-docker-image.yml` pattern. The standalone `kotlin-sdk-release.yml` / `swift-sdk-release.yml` and their tag triggers are removed.

Details:

- **Maven Central publishing** absorbs the hardened two-job flow from #4193 (bfoss765), rewritten for the `workflow_call` context: unprotected build job → deploy job whose five `JRELEASER_*` secrets exist **only** as environment secrets on the `maven-central` environment (deployment tag policy `v*`, already configured). Guards kept: tag force-move check, never-overwrite-attached-asset, all-or-nothing secrets gate (skip-with-notice when unconfigured, hard-fail when partial), deploy checkout pinned to the built SHA. New: skip-with-notice when the version is already on Maven Central, so re-runs are idempotent.
- **Emergency re-run path**: each reusable workflow keeps `workflow_dispatch` for re-releasing just one SDK for an existing platform release — must be dispatched at the tag ref (`gh workflow run release-kotlin-sdk.yml --ref vX.Y.Z -f tag=vX.Y.Z`); refuses branches, nonexistent tags, and tags without a published release (it can never create one). Per-tag concurrency groups prevent same-tag races.
- **Release hygiene**: SDK jobs attach assets passing only `tag_name` + `files` — the platform release's notes, title and prerelease flag are never touched. Caller jobs elevate `permissions: contents: write` job-level only; no `secrets: inherit` (the SDK builds need no repo/org secrets; environment secrets resolve via the deploy job's `environment:` declaration). `softprops/action-gh-release` is pinned to a commit SHA (v2.6.2).
- **Docs**: `PUBLISHING.md` CI-path section (incl. #4220's environment/secrets guidance adapted to platform tags), release skill runbook, and the design spec (`docs/RELEASE_WORKFLOWS_CONSOLIDATION_SPEC.md`).

Supersedes #4193 and #4220 (workflow code absorbed with credit; the standalone tag namespace is retired).

## How Has This Been Tested?

- `actionlint` clean on both new workflow files (remaining `release.yml` findings are pre-existing, untouched sections).
- Spec + implementation each reviewed by independent review passes (GHA semantics, publishing-credential security, release-ops); all must-fix findings addressed — see the spec's revision notes.
- Runtime verification plan (post-merge): dispatch each reusable workflow at the existing `v4.1.0-rc.1` tag ref — assets attach to that release without modifying its notes; Maven deploy skips-with-notice until the environment secrets are installed, publishes once they are. The next published prerelease exercises the `release: published` path end-to-end.

## Breaking Changes

None for consumers (no `ffi-*`/`kotlin-sdk-*` tag was ever released). Operationally: SDK releases now happen only via the platform release; `release.yml`'s manual dispatch path intentionally skips the SDK jobs (dispatch runs at a branch ref, which the `maven-central` tag policy rejects — use the per-SDK dispatch instead).

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added automated Kotlin Android SDK AAR and Swift SDK XCFramework publishing to platform releases.
  * Kotlin SDK artifacts can now be published to Maven Central.
  * Added validation for release tags, commits, existing assets, and artifact provenance.
  * Added support for rerunning SDK release workflows for an existing platform tag.

* **Documentation**
  * Expanded release and Kotlin SDK publishing guidance, including verification, reruns, Maven Central deployment, and manual fallback procedures.


thepastaclaw:
✅ Final review complete — no blockers (commit 8303ae8f)

.github/workflows/release-kotlin-sdk.yml
```diff
@@ -0,0 +1,387 @@
...
+          fi
+          if [ "$OBJ_SHA" != "$BUILT_SHA" ]; then
+            echo "::error::Tag ${TAG} now points at ${OBJ_SHA} but this run built ${BUILT_SHA} — the tag was force-moved. Refusing to attach/deploy."
+            exit 1
+          fi
+          if gh api "repos/${REPO}/releases/tags/${TAG}" --jq '.assets[].name' | grep -Fxq "$ASSET"; then
+            echo "asset_exists=true" >> "$GITHUB_OUTPUT"
+            echo "::notice::Release ${TAG} already has ${ASSET} from an earlier run — leaving it untouched. Delete the asset deliberately to re-release."
+          else
+            echo "asset_exists=false" >> "$GITHUB_OUTPUT"
```
thepastaclaw: **🔴 Blocking: An existing AAR is not tied to the build Maven Central will publish**

This guard identifies a previous AAR only by filename, but the workflow always uploads the current build's JNI libraries and the deploy job checks out the current build SHA to assemble the Maven artifacts. If an earlier run attached an AAR from commit A, the tag is later force-moved, and a recovery run builds commit B, this step leaves A's GitHub asset untouched while Maven Central receives B under the same platform version. The current tag guard proves only that the new run built the tag's current commit; it provides no provenance for the existing asset. Persist and validate a commit or artifact digest alongside the release asset, or make deployment consume and verify the exact build represented by the attached AAR.

<sub>source: ['codex']</sub>

.github/workflows/release-kotlin-sdk.yml
```diff
@@ -0,0 +1,387 @@
...
+    runs-on: ubuntu-24.04
+    timeout-minutes: 180
+    permissions:
+      contents: write # attach the AAR to the platform release
+    # Serialize same-tag runs (e.g. an emergency dispatch racing the
+    # release-triggered run) so the asset-exists guard cannot be bypassed by
+    # two concurrent builds. Never cancel a release build in progress.
+    concurrency:
+      group: release-kotlin-sdk-build-${{ inputs.tag }}
+      cancel-in-progress: false
```
thepastaclaw, line 49: **🔴 Blocking: Raw tag aliases bypass the per-release concurrency locks**

Both SDK workflows accept `vX.Y.Z` and `refs/tags/vX.Y.Z`, but their job-level concurrency keys use the unnormalized input. Those spellings therefore create separate locks for the same release in the Kotlin build, Kotlin deploy, and Swift build jobs. Concurrent runs can both pass the asset-exists guards; the pinned `softprops/action-gh-release` action defaults `overwrite_files` to true and explicitly resolves same-name upload races by deleting and re-uploading the competing asset. On Swift, that can leave the zip and checksum sourced from different runs, while parallel Kotlin deploy jobs can both attempt the same immutable Maven version. Derive the locks from one canonical identity such as the run's validated tag ref, enforce that a dispatch input matches that ref, and disable release-asset overwrites as a final defense.

docs/RELEASE_WORKFLOWS_CONSOLIDATION_SPEC.md
```diff
@@ -0,0 +1,273 @@
...
+  pre-build) — inherited limitation, unchanged.
+
+## Test / verification plan
+
+1. `actionlint` on all changed workflow files.
+2. Dry-run against the existing `v4.1.0-rc.1` tag via each reusable
+   workflow's dispatch (started **at the tag ref**): AAR + xcframework
+   attach to that release; release body/name/prerelease flag unchanged;
+   Maven deploy pauses at the gate (skip-with-notice if secrets not yet
+   installed; full publish once they are and a reviewer approves).
```
thepastaclaw: **🔴 Blocking: The documented v4.1.0-rc.1 dispatch cannot load these workflows**

The `v4.1.0-rc.1` tag resolves to `08152ea51e`, and that tree contains neither `release-kotlin-sdk.yml` nor `release-swift-sdk.yml`; both files are introduced by this PR. A `workflow_dispatch --ref` run uses the workflow at the selected branch or tag, so the proposed post-merge dispatches at `v4.1.0-rc.1` cannot start. This also makes the advertised emergency path unavailable for every platform tag predating the consolidation. Provide a secure rerun path whose workflow definition exists at the dispatch ref, or explicitly restrict reruns to tags containing these workflows and replace the rc.1 verification plan with an executable test.

<sub>source: ['codex']</sub>

.claude/skills/release/SKILL.md
```diff
@@ -94,7 +94,7 @@ gh release create v<version> --target <merge-commit-sha> \
 
 Changelog link for the notes: `https://github.com/dashpay/platform/blob/v<version>/CHANGELOG.md` (or the compare URL `.../compare/v<prev>...v<version>`).
 
-4. Publishing fires `release.yml` — verify the release build succeeds and the dashmate / Docker assets attach to the release.
+4. Publishing fires `release.yml` — verify the release build succeeds and the dashmate / Docker assets attach to the release. The Kotlin and Swift SDK jobs run in the same workflow but take longer (the Kotlin native build can run ~3 h) — the release page fills in as they finish. The Maven Central deploy runs automatically after the Kotlin build (no approval gate); note that PR CI does **not** exercise the release-profile arm64 SDK builds, so an SDK build break surfaces here first — for a stable release, confirm the SDK jobs were green on the preceding rc. To re-run just one SDK release: `gh workflow run release-kotlin-sdk.yml --ref v<version> -f tag=v<version>` (same for `release-swift-sdk.yml`) — must be dispatched **at the tag ref**.
```
thepastaclaw: **🟡 Suggestion: Require release-owner confirmation before SDK reruns**

The newly documented commands can attach public release assets and, for Kotlin, automatically publish an irrevocable Maven Central version because the environment has no approval gate. Unlike the initial release command, this rerun guidance does not require human confirmation immediately before execution. Require explicit confirmation from the release owner before dispatching either SDK workflow, while retaining the existing tag-ref commands.

thepastaclaw (changes requested):

The consolidation generally preserves credential isolation and validates release tags correctly, but exact-head verification confirms three blocking release-integrity and operability gaps. Existing GitHub assets are not tied to the Maven build, accepted tag aliases bypass all same-release concurrency locks, and the documented v4.1.0-rc.1 verification dispatch cannot load workflows absent from that tag; the SDK rerun guidance should also require explicit release-owner confirmation.

### Review provenance

<details>

```

In `.github/workflows/release-kotlin-sdk.yml`:
- [BLOCKING] .github/workflows/release-kotlin-sdk.yml:211-215: An existing AAR is not tied to the build Maven Central will publish
  This guard identifies a previous AAR only by filename, but the workflow always uploads the current build's JNI libraries and the deploy job checks out the current build SHA to assemble the Maven artifacts. If an earlier run attached an AAR from commit A, the tag is later force-moved, and a recovery run builds commit B, this step leaves A's GitHub asset untouched while Maven Central receives B under the same platform version. The current tag guard proves only that the new run built the tag's current commit; it provides no provenance for the existing asset. Persist and validate a commit or artifact digest alongside the release asset, or make deployment consume and verify the exact build represented by the attached AAR.
- [BLOCKING] .github/workflows/release-kotlin-sdk.yml:45-47: Raw tag aliases bypass the per-release concurrency locks
  Both SDK workflows accept `vX.Y.Z` and `refs/tags/vX.Y.Z`, but their job-level concurrency keys use the unnormalized input. Those spellings therefore create separate locks for the same release in the Kotlin build, Kotlin deploy, and Swift build jobs. Concurrent runs can both pass the asset-exists guards; the pinned `softprops/action-gh-release` action defaults `overwrite_files` to true and explicitly resolves same-name upload races by deleting and re-uploading the competing asset. On Swift, that can leave the zip and checksum sourced from different runs, while parallel Kotlin deploy jobs can both attempt the same immutable Maven version. Derive the locks from one canonical identity such as the run's validated tag ref, enforce that a dispatch input matches that ref, and disable release-asset overwrites as a final defense.

In `docs/RELEASE_WORKFLOWS_CONSOLIDATION_SPEC.md`:
- [BLOCKING] docs/RELEASE_WORKFLOWS_CONSOLIDATION_SPEC.md:257-262: The documented v4.1.0-rc.1 dispatch cannot load these workflows
  The `v4.1.0-rc.1` tag resolves to `08152ea51e`, and that tree contains neither `release-kotlin-sdk.yml` nor `release-swift-sdk.yml`; both files are introduced by this PR. A `workflow_dispatch --ref` run uses the workflow at the selected branch or tag, so the proposed post-merge dispatches at `v4.1.0-rc.1` cannot start. This also makes the advertised emergency path unavailable for every platform tag predating the consolidation. Provide a secure rerun path whose workflow definition exists at the dispatch ref, or explicitly restrict reruns to tags containing these workflows and replace the rc.1 verification plan with an executable test.

In `.claude/skills/release/SKILL.md`:
- [SUGGESTION] .claude/skills/release/SKILL.md:97: Require release-owner confirmation before SDK reruns
  The newly documented commands can attach public release assets and, for Kotlin, automatically publish an irrevocable Maven Central version because the environment has no approval gate. Unlike the initial release command, this rerun guidance does not require human confirmation immediately before execution. Require explicit confirmation from the release owner before dispatching either SDK workflow, while retaining the existing tag-ref commands.
```
</details>

.github/workflows/release-kotlin-sdk.yml
shumkov: Fixed in 51f7f33d8d — the dispatch input now accepts only the plain tag name; any slashed spelling (incl. `refs/tags/...`) is rejected before checkout, so the job-level concurrency groups always key on the one canonical `inputs.tag`. Both attach steps additionally pass `overwrite_files: false`, so a residual same-name upload race fails loudly instead of replacing an asset the guard vouched for.

.github/workflows/release-kotlin-sdk.yml
shumkov: Fixed in 51f7f33d8d — the AAR now attaches together with `dash-sdk-android-<version>.commit.txt` recording the commit it was built from. When both assets already exist, the run may proceed to the Maven deploy only if that recorded commit equals the current run's built commit; a mismatch (force-moved tag + recovery build) or a half-attached pair hard-fails. The GitHub AAR and the Maven Central artifact can no longer come from different commits.

docs/RELEASE_WORKFLOWS_CONSOLIDATION_SPEC.md
shumkov: Addressed in 51f7f33d8d — correct, a dispatch loads the workflow from the selected ref, so the tag-ref re-run path exists only for tags created after this lands. Made honest: the Maven deploy job now skips on non-tag run refs (instead of failing at the environment tag policy), so a dev-branch dispatch — the only option for pre-consolidation tags — builds and attaches assets green, with Maven going through the manual runbook. Spec dry-run plan, PUBLISHING.md and the release skill rewritten accordingly.

.claude/skills/release/SKILL.md
shumkov: Added in 51f7f33d8d — the release skill (and PUBLISHING.md) now require explicit release-owner confirmation immediately before dispatching either SDK re-run, since the Kotlin path publishes irrevocably to Maven Central with no approval gate.

.github/workflows/release-kotlin-sdk.yml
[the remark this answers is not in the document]
shumkov: Fixed in 51f7f33d8d — the `refs/tags/` input form is now rejected outright (plain tag name only), so the concurrency group always keys on the canonical spelling; attach steps also pass `overwrite_files: false`. (Same fix as the sibling thepastaclaw finding.)

.github/workflows/release-swift-sdk.yml
```diff
@@ -0,0 +1,222 @@
...
+    runs-on: macos-15
+    timeout-minutes: 45
+    permissions:
+      contents: write # attach the xcframework to the platform release
+    # Serialize same-tag runs (e.g. an emergency dispatch racing the
+    # release-triggered run) so the asset-exists guard cannot be bypassed by
+    # two concurrent builds. Never cancel a release build in progress.
+    concurrency:
+      group: release-swift-sdk-${{ inputs.tag }}
+      cancel-in-progress: false
```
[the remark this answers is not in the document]
shumkov: Fixed in 51f7f33d8d — same as the Kotlin workflow: the `refs/tags/` input form is rejected outright, so the concurrency group keys on the canonical tag; `overwrite_files: false` added on the attach step.

.claude/skills/release/SKILL.md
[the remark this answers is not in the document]
shumkov: Added in 51f7f33d8d — the skill now requires explicit release-owner confirmation immediately before dispatching either SDK re-run workflow.

.github/workflows/release-swift-sdk.yml
```diff
@@ -0,0 +1,232 @@
...
+      # body/prerelease/generate_release_notes here would overwrite the
+      # platform release's own notes, title or prerelease flag.
+      # overwrite_files: false makes a same-name upload race fail loudly
+      # instead of silently replacing an asset the guard above vouched for.
+      - name: Attach XCFramework to the platform release
+        if: steps.tag-guard.outputs.asset_exists != 'true'
+        uses: softprops/action-gh-release@3bb12739c298aeb8a4eeaf626c5b8d85266b0e65 # v2
+        with:
+          tag_name: ${{ steps.release-ref.outputs.tag_name }}
+          overwrite_files: false
```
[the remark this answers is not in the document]
shumkov: Fixed in 54dfd529a3 — the Swift workflow now mirrors the Kotlin provenance guard: the zip attaches together with `DashSDKFFI-<version>.commit.txt`, and existing assets are reused only when the recorded commit equals the commit the current run built. A force-moved tag now hard-fails the re-run instead of leaving a green run with a different commit's assets; a partial asset set (interrupted attach) also hard-fails.

.github/workflows/release-kotlin-sdk.yml
```diff
@@ -0,0 +1,425 @@
...
+    name: Deploy to Maven Central (version from tag)
+    needs: build-and-release
+    # Runs only when the RUN's ref is a tag: the maven-central environment's
+    # v* tag policy would reject a branch ref anyway, but rejection fails the
+    # job — skipping keeps a branch-ref dispatch (the only way to re-attach
+    # assets for a tag that predates these workflow files, since a dispatch
+    # loads the workflow from the selected ref) green while still never
+    # deploying from it. Maven for such tags goes through the manual runbook
+    # in packages/kotlin-sdk/PUBLISHING.md.
+    if: startsWith(github.ref, 'refs/tags/')
```
thepastaclaw: **🔴 Blocking: Require the deploy ref to match the target release tag**

This condition verifies only that the workflow run uses some tag ref; it does not bind that ref to `inputs.tag`, which controls the checked-out commit and Maven version. A manual dispatch can therefore run the workflow from tag A while passing tag B, causing the protected `maven-central` job to publish B under authorization granted by A's `v*` environment policy. This also bypasses the documented restriction for pre-consolidation tags: an operator can select a newer tag containing the workflow and publish an older input tag through CI instead of using the manual runbook. Maven versions are immutable, so require the run ref to equal the exact target tag.

```suggestion
    if: ${{ github.ref == format('refs/tags/{0}', inputs.tag) }}
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

All four carried-forward findings are fixed at the current head: Kotlin asset provenance is validated, tag aliases are rejected, pre-consolidation reruns have an executable branch-ref path, and release-owner confirmation is required. The current head also resolves CodeRabbit's Swift provenance finding by treating the zip, checksum, and commit record as one validated asset set. However, the latest delta introduces one new blocking authorization gap because Maven deployment accepts any tag run ref instead of requiring the run ref to match the release tag being published.

### Review provenance

🔴 1 blocking

<details>

```

In `.github/workflows/release-kotlin-sdk.yml`:
- [BLOCKING] .github/workflows/release-kotlin-sdk.yml:295: Require the deploy ref to match the target release tag
  This condition verifies only that the workflow run uses some tag ref; it does not bind that ref to `inputs.tag`, which controls the checked-out commit and Maven version. A manual dispatch can therefore run the workflow from tag A while passing tag B, causing the protected `maven-central` job to publish B under authorization granted by A's `v*` environment policy. This also bypasses the documented restriction for pre-consolidation tags: an operator can select a newer tag containing the workflow and publish an older input tag through CI instead of using the manual runbook. Maven versions are immutable, so require the run ref to equal the exact target tag.
```
</details>

.github/workflows/release-kotlin-sdk.yml
shumkov: Fixed in a8867d1fc1 — applied exactly as suggested: `if: ${{ github.ref == format('refs/tags/{0}', inputs.tag) }}`, so the deploy runs only when the run's ref is the target tag itself. A dispatch started at tag A with input B (or at any branch) now skips the deploy instead of publishing under another ref's environment authorization; the pre-consolidation-tag restriction can no longer be bypassed via a newer tag's workflow file.

thepastaclaw (commented):

The only carried-forward blocking finding (prior-1: Maven deploy job must bind to the exact target release tag, not merely any tag ref) is verified fixed at head a8867d1fc17d — release-kotlin-sdk.yml:297 now reads `github.ref == format('refs/tags/{0}', inputs.tag)`. The CodeRabbit Swift provenance finding was likewise already fixed in an earlier commit (54dfd529a3) and remains correct at head. The latest delta (54dfd529a3..a8867d1fc1) introduces no new defects — it is a single-line hardening of the Maven deploy guard. One suggestion-level architecture observation carries over from the FFI review round: the Kotlin AAR attached to the GitHub release and the AAR staged for Maven Central are two independent Gradle builds with no byte-parity check between them, though the native .so boundary is pinned identical via a shared artifact upload.

### Review provenance
- Sonnet reviewers: `claude-sonnet-5` — general (failed), `claude-sonnet-5` — ffi-engineer (failed), `claude-sonnet-5` — general (completed), `claude-sonnet-5` — ffi-engineer (completed)

🟡 1 suggestion(s)

<details>

```

In `.github/workflows/release-kotlin-sdk.yml`:
- [SUGGESTION] .github/workflows/release-kotlin-sdk.yml:188-409: GitHub-attached AAR and Maven Central AAR are independently rebuilt, not reused byte-for-byte
  The `build-and-release` job runs `./gradlew :sdk:assembleRelease` (line 188) to produce the AAR attached to the GitHub release. The `maven-central-deploy` job checks out the same commit, restores the same native `.so` files via `actions/download-artifact` (lines 384-387), but re-invokes Gradle from scratch (`:sdk:publishReleasePublicationToStagingRepository`, lines 407-409) to build the AAR actually staged and pushed to Maven Central. These are two separate AGP `bundleReleaseAar` invocations on two different runners. `verifyStagedAarForRemotePublish` only checks that the staged AAR contains the required JNI `.so` entries — it never compares the staged AAR against the GitHub-attached one. Since the native `.so` boundary is explicitly pinned identical via the shared artifact, the exposure is limited to non-native packaging (classes.jar, resources, R.txt, POM/manifest), so this is unlikely to break the FFI contract in practice, but there is no guarantee today that a consumer verifying the GitHub asset gets byte-parity with the immutable Maven Central artifact.
```
</details>

.github/workflows/release-kotlin-sdk.yml
```diff
@@ -0,0 +1,427 @@
...
+        working-directory: packages/kotlin-sdk
+        env:
+          SDK_VERSION: ${{ needs.build-and-release.outputs.version }}
+          # Same ASCII-armored GPG key/passphrase jreleaser uses below; Gradle
+          # signing reads them as the in-memory signingKey/signingPassword
+          # project properties.
+          ORG_GRADLE_PROJECT_signingKey: ${{ secrets.JRELEASER_GPG_SECRET_KEY }}
+          ORG_GRADLE_PROJECT_signingPassword: ${{ secrets.JRELEASER_GPG_PASSPHRASE }}
+        run: >-
+          ./gradlew :sdk:publishReleasePublicationToStagingRepository
```
thepastaclaw, line 421: **🟡 Suggestion: GitHub-attached AAR and Maven Central AAR are independently rebuilt, not reused byte-for-byte**

The `build-and-release` job runs `./gradlew :sdk:assembleRelease` (line 188) to produce the AAR attached to the GitHub release. The `maven-central-deploy` job checks out the same commit, restores the same native `.so` files via `actions/download-artifact` (lines 384-387), but re-invokes Gradle from scratch (`:sdk:publishReleasePublicationToStagingRepository`, lines 407-409) to build the AAR actually staged and pushed to Maven Central. These are two separate AGP `bundleReleaseAar` invocations on two different runners. `verifyStagedAarForRemotePublish` only checks that the staged AAR contains the required JNI `.so` entries — it never compares the staged AAR against the GitHub-attached one. Since the native `.so` boundary is explicitly pinned identical via the shared artifact, the exposure is limited to non-native packaging (classes.jar, resources, R.txt, POM/manifest), so this is unlikely to break the FFI contract in practice, but there is no guarantee today that a consumer verifying the GitHub asset gets byte-parity with the immutable Maven Central artifact.

<sub>source: ['claude']</sub>

shumkov: Acknowledged — this is a real, deliberate limitation rather than an oversight. The guaranteed contract is: both AARs come from the same SHA-pinned commit with byte-identical native libraries (handed over via the jniLibs artifact and enforced by `verifyJniLibsForRemotePublish`/`verifyStagedAarForRemotePublish`); only non-native packaging could differ between the two AGP invocations. Guaranteeing byte-parity would require rewiring the Gradle `release` publication to consume a prebuilt AAR file instead of the module build — out of scope for this CI consolidation, but a reasonable follow-up if byte-level verification of the GitHub asset against Maven Central becomes a requirement. The `.commit.txt` provenance asset ties both channels to the same source commit, which is the invariant this PR enforces.

.github/workflows/release-kotlin-sdk.yml
```diff
@@ -0,0 +1,427 @@
...
+        if: steps.secrets-gate.outputs.proceed == 'true'
+        env:
+          SDK_VERSION: ${{ needs.build-and-release.outputs.version }}
+        run: |
+          URL="https://repo1.maven.org/maven2/org/dashj/dash-sdk-android/${SDK_VERSION}/dash-sdk-android-${SDK_VERSION}.pom"
+          if curl -sfIo /dev/null "$URL"; then
+            echo "::notice::org.dashj:dash-sdk-android:${SDK_VERSION} is already on Maven Central — deploy skipped (Central versions are immutable)."
+            echo "already_published=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "already_published=false" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
shumkov: Fixed in 42f7956d1f — applied as suggested (kept as a HEAD request): the probe now branches on the HTTP status with bounded timeouts. 200 → skip (already published), 404 → proceed, anything else (curl failure, rate limit, 5xx) → hard-fail as inconclusive instead of blindly attempting an immutable version.

.github/workflows/release-swift-sdk.yml
```diff
@@ -0,0 +1,247 @@
...
+          key: cargo-registry-${{ hashFiles('**/Cargo.lock') }}
+
+      - name: Add iOS Rust targets
+        run: |
+          rustup target add aarch64-apple-ios aarch64-apple-ios-sim
+
+      - name: Install cbindgen (for header generation)
+        run: |
+          brew install cbindgen || true
+
```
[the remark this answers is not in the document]
shumkov: Fixed in 42f7956d1f — the install step now verifies `command -v cbindgen` after `brew install` and fails the job if the binary is missing. (The `|| true` on the install itself is kept only to tolerate brew's benign non-zero exits, e.g. already-installed; the availability check is what gates.)

.github/workflows/release-swift-sdk.yml
```diff
@@ -0,0 +1,251 @@
...
+        run: |
+          rustup target add aarch64-apple-ios aarch64-apple-ios-sim
+
+      - name: Install cbindgen (for header generation)
+        run: |
+          # brew returns non-zero for benign cases (already installed), so the
+          # install itself is tolerated — but a genuinely missing binary must
+          # fail here, not surface later as a broken header generation.
+          brew install cbindgen || true
+          command -v cbindgen >/dev/null || { echo "::error::cbindgen is not available after brew install."; exit 1; }
```
thepastaclaw: **🟡 Suggestion: New cbindgen executable check gates on a binary the build never uses**

The new hard-fail (`command -v cbindgen >/dev/null || { ... exit 1; }`) requires a standalone `cbindgen` CLI on PATH, but `packages/swift-sdk/build_ios.sh` — the script this workflow actually runs — never shells out to `cbindgen`. Header generation happens inside each crate's `build.rs` (e.g. `packages/rs-sdk-ffi/build.rs`, `packages/rs-platform-wallet-ffi/build.rs`) via `cbindgen::Builder`/`cbindgen::Config`, which is a Cargo build-dependency (`cbindgen = "0.27"` in each Cargo.toml) resolved and compiled by Cargo itself, independent of any binary on PATH. As written, a flaky or rate-limited `brew install cbindgen` now aborts an otherwise fully-buildable release for a tool that isn't consumed. Either drop the `command -v` check (and the now-vestigial install step) or confirm there's a legitimate consumer of the CLI elsewhere in the pipeline before keeping the hard gate.

<sub>source: ['codex']</sub>

thepastaclaw (commented):
## Final validation — Codex + Opus

This delta hardens two CI preflight checks: the Maven Central existence probe now fails closed on any non-200/404 HTTP status instead of silently treating inconclusive responses as "not published," and the Swift SDK workflow now verifies cbindgen is actually on PATH after a tolerant `brew install ... || true`. The Maven check is a solid, narrowly-scoped fix with no issues found. The new cbindgen check, however, gates the Swift release on a binary that the actual build (via Cargo `build.rs` using the `cbindgen` library, not the CLI) never invokes, so a transient Homebrew hiccup can now fail an otherwise-viable release. The previously reported AAR byte-parity gap between the GitHub-release and Maven Central Kotlin artifacts is unchanged by this delta and remains a deliberately deferred, non-blocking limitation per the author.

### Review provenance
- Opus reviewers: `claude-sonnet-5` — general (failed), `claude-sonnet-5` — general (completed)

🟡 1 suggestion(s)

<details>

```

In `.github/workflows/release-swift-sdk.yml`:
- [SUGGESTION] .github/workflows/release-swift-sdk.yml:126-132: New cbindgen executable check gates on a binary the build never uses
  The new hard-fail (`command -v cbindgen >/dev/null || { ... exit 1; }`) requires a standalone `cbindgen` CLI on PATH, but `packages/swift-sdk/build_ios.sh` — the script this workflow actually runs — never shells out to `cbindgen`. Header generation happens inside each crate's `build.rs` (e.g. `packages/rs-sdk-ffi/build.rs`, `packages/rs-platform-wallet-ffi/build.rs`) via `cbindgen::Builder`/`cbindgen::Config`, which is a Cargo build-dependency (`cbindgen = "0.27"` in each Cargo.toml) resolved and compiled by Cargo itself, independent of any binary on PATH. As written, a flaky or rate-limited `brew install cbindgen` now aborts an otherwise fully-buildable release for a tool that isn't consumed. Either drop the `command -v` check (and the now-vestigial install step) or confirm there's a legitimate consumer of the CLI elsewhere in the pipeline before keeping the hard gate.
```
</details>

.github/workflows/release-swift-sdk.yml
shumkov: Fixed in 8303ae8fb8 — verified you're right: `build_ios.sh` never shells out to the cbindgen CLI (headers come from each FFI crate's `build.rs` via the `cbindgen` crate as a Cargo build-dependency), so the install step was vestigial weight inherited from the old standalone workflow and the availability gate could only abort a release over an unused tool. Dropped the whole step.

thepastaclaw (approved):
## Final validation — Codex + Opus

At exact head 8303ae8fb86870b81f642e380395a8954b333d0e, the latest delta fixes prior-1 by removing the unused Homebrew cbindgen installation and executable gate; it introduces no new findings. There are no carried-forward live findings: prior-2 remains technically present but was explicitly and intentionally deferred by the author as outside this CI-consolidation PR's required invariant.
Source: reviewers codex/general=gpt-5.6-sol(completed); claude/general=claude-sonnet-5(failed); claude/general=claude-sonnet-5(failed); claude/general=claude-sonnet-5(completed); verifier=codex/final-verifier=gpt-5.6-sol(completed) fallback_for_sonnet_verifier=true; coordinator=openclaw-agent/cliproxy/gpt-5.6-sol/api-gateway/high(orchestration-only).

### Review provenance
- Verifier: `gpt-5.6-sol` — final-verifier (Opus verifier fallback)
- Opus reviewers: `claude-sonnet-5` — general (failed), `claude-sonnet-5` — general (failed), `claude-sonnet-5` — general (completed)

PR #4231, opened 2026-07-25
fix(sdk): restore proved current-epoch fetch with two-step explicit-start query

QuantumExplorer:

`ExtendedEpochInfo::fetch_current` (and therefore the eager `Sdk::refresh_protocol_version` at startup) always failed proof verification on v4.1-dev with:

```
proved descending epoch queries require an explicit start epoch
```

The proof-context binding change made the verifier reject proved descending epoch queries without an explicit start, because resolving "the last epoch" during verification would require trusting unsigned response metadata. `fetch_current` still sent `EpochQuery { start: None, ascending: false }`. Reproduced on iOS SwiftExampleApp against testnet: every app start logged `no more retries left` and the eager protocol-version refresh silently no-oped.

The obvious fix — an explicit descending start at `MAX_EPOCH` — **does not work**: Drive pre-creates `perpetual_storage_epochs` (2000) empty epoch trees above the current one at genesis, and a proved descending query starting inside that window consumes its GroveDB limit on the empty trees, verifiably returning **zero** epochs. Confirmed against live testnet and pinned by a new `rs-drive` test.

## What was done?

- `fetch_current` now issues two proved, guard-compliant queries:
  1. **Probe**: fetch the genesis epoch (`EpochQuery::genesis()`, explicit ascending start); its response metadata carries the current-epoch hint.
  2. **Fetch**: `EpochQuery::newest_at_or_below(hint)` — descending with an explicit start, which for an honest node is exactly the current epoch tree.

  The hint only shapes the second request; proof verification stays a pure function of the request and quorum-signed state. An inflated hint lands in the provably-empty future window and yields `EpochNotFound` rather than a bogus epoch; a deflated hint can present a stale epoch as current — the same exposure as the pre-guard metadata trust, now documented on the impl.
- `rs-sdk-ffi`'s `platform_status` used the same broken shape via `fetch_many`; it now uses `fetch_current`. wasm-sdk is fixed transitively.
- Mock helpers (`expect_epoch_refresh`, `bootstrap_mock_sdk_to_latest`) register both expectations so they keep matching the queries `fetch_current` issues.
- `test_epoch_fetch_current` expects success again; its offline vectors were regenerated from a real testnet response (current epoch 17673, protocol version 12) using the trusted HTTP context provider, so the offline test replays the full proof-verification path (guard → GroveDB verify → quorum signature).
- New regression tests:
  - rs-sdk unit test: both `fetch_current` query shapes pass the `FromProof` guard and fit Drive's `+ EPOCH_KEY_OFFSET` key encoding.
  - drive-proof-verifier: explicit-start descending requests pass the guard (fail later on proof content, not request shape).
  - rs-drive: descending from `MAX_EPOCH` provably returns no epochs (documents why a single-query "current epoch" is not expressible).
- Fixed a pre-existing drive-proof-verifier test compile break: three tests still referenced `StateTransitionProofResult` after the rename to `StateTransitionProofOutcome`, so `cargo test -p drive-proof-verifier` did not compile on v4.1-dev.

## How Has This Been Tested?

- `cargo test -p dash-sdk` (offline mode): full suite green, including `test_epoch_fetch_current` replaying the regenerated testnet vectors and `test_refresh_ratchets_up_via_proven_query` (the exact eager-refresh path that failed on iOS).
- `cargo test -p drive verify_epoch_infos` and full `cargo test -p drive-proof-verifier`: green.
- `cargo check -p rs-sdk-ffi`: green. `cargo fmt --all` applied.
- End-to-end against live testnet: the two-step fetch returned current epoch 17673 with a verified proof.

## Breaking Changes

None. `EpochQuery` gains `genesis()` and `newest_at_or_below()` constructors; existing fields are unchanged. `fetch_current` performs two round trips instead of one failing one.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added verified current-epoch retrieval with metadata, including protection against stale or inflated epoch hints.
  * Added clear errors when a node reports an epoch that conflicts with cryptographically proven data.
  * Added support for explicit ascending and genesis-based epoch queries.

* **Bug Fixes**
  * Improved platform status retrieval and epoch handling when epochs are missing, unstarted, or queried in descending order.
  * Strengthened validation of epoch proofs and authenticated boundaries.


thepastaclaw:
⛔ Blockers found — Sonnet deferred (commit 94c66196)

packages/rs-sdk/src/platform/types/epoch.rs
```diff
@@ -14,44 +18,84 @@ pub type Epoch = ExtendedEpochInfo;
...
-        };
+        let (_, probe_metadata) =
+            Self::fetch_with_metadata(sdk, current_epoch_probe_query(), None).await?;
 
-        let (epoch, metadata) = Self::fetch_with_metadata(sdk, query, None).await?;
+        let hint = epoch_hint_from_metadata(&probe_metadata);
+        let (epoch, metadata) =
+            Self::fetch_with_metadata(sdk, current_epoch_query(hint), None).await?;
 
         Ok((epoch.ok_or(Error::EpochNotFound)?, metadata))
```
thepastaclaw: **🔴 Blocking: Unsigned probe metadata can select a historical epoch as current**

`probe_metadata.epoch` is not authenticated by the Tenderdash signature. `verify_tenderdash_proof` constructs `StateId` from the protocol version, core height, time, app hash, and platform height, but not the metadata epoch. A DAPI node can therefore return a valid genesis proof while changing only `metadata.epoch` to a historical index such as 0. The second request then explicitly asks for epoch 0, and its valid proof establishes only that epoch 0 exists in the authenticated current state—not that it is current. This is proof-valid stale data, not a forged proof. The same stale result can occur honestly if the epoch advances between the probe and the second request. The signed, upward-only protocol-version ratchet is not downgraded by this attack, but SDK, WASM, and FFI callers can receive stale epoch fee, version, time, and height information under a current-epoch API. Keep this lookup fail-closed until the selected index can be derived from authenticated state or a proof can establish the newest initialized epoch.

<sub>source: ['codex']</sub>

packages/rs-sdk/tests/fetch/epoch.rs
```diff
@@ -163,28 +163,26 @@ async fn test_epoch_fetch_future() {
...
         .await
-        .expect_err("proved current-epoch fetch must fail closed");
+        .expect("fetch current epoch");
 
-    assert!(
-        error.to_string().contains("explicit start epoch"),
-        "expected an explicit-start rejection, got: {error}"
-    );
+    // The returned epoch is a real one, not the MAX_EPOCH query bound.
+    assert!(epoch.index() < dpp::block::epoch::MAX_EPOCH);
```
thepastaclaw: **🟡 Suggestion: Current-epoch regression test also accepts every historical epoch**

`epoch.index() < MAX_EPOCH` is true for genesis and essentially every valid historical epoch, so the test remains green if the implementation returns the probe result or follows a deflated hint. The recorded honest vectors report epoch 17673 on both responses. Compare the returned epoch with the final response metadata to pin the intended two-query orchestration without breaking live vector regeneration, and add a controlled deflated-hint case when the trust-boundary issue is fixed.

```suggestion
let (epoch, metadata) = ExtendedEpochInfo::fetch_current_with_metadata(&sdk)
    .await
    .expect("fetch current epoch");

assert_eq!(u32::from(epoch.index()), metadata.epoch);
```

<sub>source: ['codex']</sub>

packages/rs-sdk/src/platform/types/epoch.rs
```diff
@@ -63,11 +107,43 @@ pub struct EpochQuery {
...
+    }
+
+    /// Descending query with an explicit start, returning the newest started epochs at
+    /// or below `start`.
+    ///
+    /// With `limit: Some(1)` and `start` set to the current epoch index, this returns
+    /// the current epoch. Do not pass a `start` far above the current epoch: Drive
+    /// pre-creates empty epoch trees ahead of the current one, and a proved descending
+    /// query starting inside that empty window provably returns no epochs.
+    pub fn newest_at_or_below(start: EpochIndex) -> Self {
```
thepastaclaw: **🟡 Suggestion: The public newest_at_or_below API does not guarantee its advertised search behavior**

The constructor is named and initially documented as returning the newest started epoch at or below `start`, but a future start can return no epochs because pre-created empty epoch trees consume the query limit. This is an explicit descending range query, not a general at-or-below search. Rename it to describe the actual operation, such as `descending_from`, or make it private and expose an API whose preconditions prevent callers from recreating the `MAX_EPOCH` failure.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The explicit-start query shapes and each individual GroveDB/Tenderdash proof are mechanically valid, and the focused SDK and FFI tests pass. However, the two-step composition still relies on an unsigned epoch selector, so it can return a proof-valid historical epoch as current without forging either proof; this fails the PR's proof-verified current-epoch goal and requires changes.

### Review provenance

<details>

```

In `packages/rs-sdk/src/platform/types/epoch.rs`:
- [BLOCKING] packages/rs-sdk/src/platform/types/epoch.rs:49-56: Unsigned probe metadata can select a historical epoch as current
  `probe_metadata.epoch` is not authenticated by the Tenderdash signature. `verify_tenderdash_proof` constructs `StateId` from the protocol version, core height, time, app hash, and platform height, but not the metadata epoch. A DAPI node can therefore return a valid genesis proof while changing only `metadata.epoch` to a historical index such as 0. The second request then explicitly asks for epoch 0, and its valid proof establishes only that epoch 0 exists in the authenticated current state—not that it is current. This is proof-valid stale data, not a forged proof. The same stale result can occur honestly if the epoch advances between the probe and the second request. The signed, upward-only protocol-version ratchet is not downgraded by this attack, but SDK, WASM, and FFI callers can receive stale epoch fee, version, time, and height information under a current-epoch API. Keep this lookup fail-closed until the selected index can be derived from authenticated state or a proof can establish the newest initialized epoch.
- [SUGGESTION] packages/rs-sdk/src/platform/types/epoch.rs:132-139: The public newest_at_or_below API does not guarantee its advertised search behavior
  The constructor is named and initially documented as returning the newest started epoch at or below `start`, but a future start can return no epochs because pre-created empty epoch trees consume the query limit. This is an explicit descending range query, not a general at-or-below search. Rename it to describe the actual operation, such as `descending_from`, or make it private and expose an API whose preconditions prevent callers from recreating the `MAX_EPOCH` failure.

- [SUGGESTION] packages/rs-sdk/src/sdk.rs:369-380: Refresh documentation promises failure semantics the two-step fetch no longer provides
  The documentation still says refresh issues one query and that a failed query leaves the stored protocol version untouched. The genesis probe is now a complete proven fetch, so `verify_response_metadata` can ratchet the SDK version before the second request begins. If that request fails, `refresh_protocol_version` logs that it is keeping the current version even though the probe may already have increased it, and returns the increased value. Retaining a signed, upward-only version learned by the probe may be desirable, but the public documentation, warning text, and tests must describe this partial-success behavior accurately.

In `packages/rs-sdk/tests/fetch/epoch.rs`:
- [SUGGESTION] packages/rs-sdk/tests/fetch/epoch.rs:182-187: Current-epoch regression test also accepts every historical epoch
  `epoch.index() < MAX_EPOCH` is true for genesis and essentially every valid historical epoch, so the test remains green if the implementation returns the probe result or follows a deflated hint. The recorded honest vectors report epoch 17673 on both responses. Compare the returned epoch with the final response metadata to pin the intended two-query orchestration without breaking live vector regeneration, and add a controlled deflated-hint case when the trust-boundary issue is fixed.

In `packages/rs-sdk-ffi/src/system/queries/platform_status.rs`:
- [SUGGESTION] packages/rs-sdk-ffi/src/system/queries/platform_status.rs:109-115: Platform-status test exercises only the error path
  The test discards the entire `DashSDKResult`, so it passes whether the call succeeds or fails. `create_mock_sdk_handle()` registers no epoch expectations, meaning the new `fetch_current` call cannot complete its genesis probe and explicit descending fetch; the currently passing test therefore does not cover the changed success path or JSON conversion. Register both expected epoch requests and assert a null error, `DashSDKResultDataType::String`, and the expected block and core heights in the returned JSON.
```
</details>

packages/rs-sdk/src/platform/types/epoch.rs
QuantumExplorer: Fixed in 8849680 — the hint is no longer trusted as an answer.

The confirming query is now **ascending from the hint with `count: 2`** instead of descending with `count: 1`. Drive keeps epoch `n+1` as a pre-created *empty* tree until that epoch starts, and an empty tree consumes query limit without contributing elements — so a single-epoch result is itself proof, against the same quorum-signed root, that no epoch above the hint has started.

- Hint **above** the tip → lands in the empty window, provably matches nothing → `EpochNotFound`.
- Hint **below** the tip → contradicted by the node's own proof, which carries the newer epoch. The query repeats from that *proven* index, bounded at 2 refinements (enough for the honest epoch-turnover race you noted), then fails closed with a new `StaleNodeError::Epoch`.
- Hint **correct** → one epoch back, and the proof covers the absence above it.

So `metadata.epoch` is now only a request-shaping input, and every accepted outcome is derived from the proof. Round trips are unchanged (2 in the happy path).

Pinned by tests at both layers: `rs-drive` `should_prove_that_no_epoch_started_above_the_current_one` proves the empty-tree limit behaviour at, below and above the current epoch; `rs-sdk` mock tests cover confirmed / stale-by-one / persistently-deflated / inflated hints. Verified live against testnet during vector regeneration (epoch 17723, protocol version 13).

packages/rs-sdk/src/platform/types/epoch.rs
QuantumExplorer: Removed in 8849680. The fix for the blocking finding made the descending query unnecessary — the confirming query is now ascending — so `newest_at_or_below` is gone rather than renamed. What remains is `EpochQuery::ascending_from(start)`, which names exactly the range query it builds, and `genesis()` on top of it.

packages/rs-sdk/tests/fetch/epoch.rs
QuantumExplorer: Done in 8849680, along the lines you suggested:

```rust
let (epoch, metadata) = ExtendedEpochInfo::fetch_current_with_metadata(&sdk).await...;
assert_eq!(u32::from(epoch.index()), metadata.epoch);
```

plus an assertion that `fetch(&sdk, epoch.index() + 1)` returns `None` — the end-to-end form of the claim the confirming query proves. Offline vectors regenerated against testnet (epoch 17723, protocol version 13), so all three recorded requests are real.

The deflated-hint case you asked for is now testable and covered as a mock unit test in `packages/rs-sdk/src/platform/types/epoch.rs` (`should_advance_past_a_hint_the_proof_shows_as_stale` and `should_fail_closed_when_the_hint_stays_below_the_proven_tip`), rather than in the vector-replay test where the hint cannot be varied.

QuantumExplorer:
All five review findings addressed in 8849680.

**🔴 Blocking — unsigned probe metadata could select a historical epoch as current.** Fixed by making the second query prove the selection instead of trusting it. It is now **ascending from the hint with `count: 2`** rather than descending with `count: 1`: Drive keeps epoch `n+1` as a pre-created empty tree until it starts, and an empty tree consumes query limit without contributing elements, so a single-epoch result is proof — against the same quorum-signed root — that nothing above the hint has started. An inflated hint lands in the empty window and matches nothing (`EpochNotFound`); a deflated or raced hint is contradicted by the node's own proof, which carries the newer epoch, so the query repeats from that proven index (bounded at 2 refinements) and otherwise fails closed with a new `StaleNodeError::Epoch`. Still 2 round trips in the happy path.

**🟡 `newest_at_or_below` naming.** Removed — the ascending confirmation made it unnecessary. `EpochQuery::ascending_from(start)` names the query it actually builds.

**🟡 `refresh_protocol_version` docs.** Corrected: they no longer promise one query or an untouched version on failure. Each verified response ratchets on its own, so a partial refresh can raise the stored version, and the returned value reflects that; the warning log now says so too.

**🟡 `test_epoch_fetch_current` accepted every historical epoch.** Now compares the returned epoch with the final response metadata and asserts the epoch above it is absent. Offline vectors regenerated against live testnet (epoch 17723, protocol version 13). The deflated-hint case is covered as a mock unit test, where the hint can actually be varied.

**🟡 platform_status test only exercised the error path.** It now registers both epoch expectations and asserts null error, `String` result type and the expected block/core heights in the JSON; a second test pins the error path explicitly.

New regression tests: `rs-drive` pins the empty-tree limit behaviour at, below and above the current epoch (the invariant the confirmation depends on); `rs-sdk` mock tests cover confirmed / stale-by-one / persistently-deflated / inflated hints.

Verification: `cargo test -p dash-sdk` (179 lib + 131 offline integration), `cargo test -p drive-proof-verifier` (246), `cargo test -p rs-sdk-ffi --lib` (305), `cargo test -p drive --lib verify_epoch_infos` — all green; `cargo clippy` clean on drive / dash-sdk / rs-sdk-ffi; `cargo fmt --all` applied. Confirmed live against testnet during vector regeneration.

DashBot-0001:
**Additional impact data: on the operation path this failure exhausts the whole DAPI address pool.**

The PR description frames the symptom as a startup-time no-op (`refresh_protocol_version` silently doing nothing, `no more retries left` in the log). On testnet against v4.1-dev I hit the same error on the **operation** path, where the consequence is larger: each failed epoch fetch causes the client to **ban the address it was talking to**, and the bans accumulate until nothing is left to talk to.

### Measured (single run, today, testnet)

Eight concurrent `shielded_transfer` calls on one wallet:

| log pattern | count |
|---|---|
| `proved descending epoch queries require an explicit start epoch` | 202 |
| `ban address …` | 47 |
| distinct addresses banned | **28** (`68.67.122.1` … `.29`) |
| `exceeds maximum size of 20480 bytes` | 0 |
| nullifier / "in progress" markers | 0 |

All 8 calls then returned:

```
-32004 "Backend task failed: All Dash network servers are temporarily unreachable. Please wait a minute and retry."
```

Chain state was unchanged (`pool_notes_delta 0`, `pool_credits_delta 0`) — clean fail-closed, no partial state.

### It is not caused by concurrency, only accelerated by it

A separate **serial** run the same hour: transfers 1–6 each succeeded and settled on chain, and transfer 7 failed the same way. So the bans accumulate on ordinary sequential use too; eight parallel calls merely reach the exhaustion point faster (8× the epoch lookups).

It is also **recoverable, not a hard failure**: after waiting ~75 s, a single transfer completed normally. That is consistent with pool exhaustion plus ban backoff rather than connectivity loss.

### The secondary point, which outlives this fix

`keeping the cached fee multiplier` shows there is already a fallback for the *value*. What has no fallback is the **health accounting**: a deterministic decode/verification error is attributed to the peer and costs that peer a ban. Every peer returns it, so every peer gets banned in turn. Once #4231 lands this particular trigger disappears, but any future deterministic request-level error would cascade the same way. It may be worth distinguishing "this response was invalid" from "this node is unhealthy" in the ban logic.

---

*Testnet, self-built DET `v1.0-dev` pinning `feat/platform-wallet-storage-rehydration`. Figures are from executed runs in one session; the concurrent and serial runs are separate. Produced with AI assistance (Claude).*

DashBot-0001:
I traced the ordering finding on `packages/rs-sdk/src/platform/types/epoch.rs` against this branch at `34a4273`. **The premise is right, but the conclusion doesn't hold here** — `ExtendedEpochInfos` arrives in ascending `EpochIndex` order regardless of query direction, so `started.next()` in `resolve_current_epoch` does observe the lowest index. An unconditional sort would be dead code today.

The chain from proof to consumer:

1. `verify_epoch_infos_v0` folds GroveDB's result into `BTreeMap::<Vec<Vec<u8>>, BTreeMap<_, _>>` (`packages/rs-drive/src/verify/system/verify_epoch_infos/v0/mod.rs:90-99`). The path's last component is `(epoch_index + EPOCH_KEY_OFFSET).to_be_bytes()`, so lexicographic byte order equals numeric epoch order — and this fold discards whatever order GroveDB returned.
2. `results.into_iter().filter_map(...)` carries that order into the `Vec<ExtendedEpochInfo>`.
3. `ExtendedEpochInfos as FromProof` collects that `Vec` in order (`packages/rs-drive-proof-verifier/src/proof.rs:1593-1603`); `into_option()` only wraps, and `Sdk::parse_proof_with_metadata_and_proof` → `fetch_many_with_metadata_and_proof` pass it through untouched.

The direct counter-example is already in the drive module: `should_prove_and_verify_epoch_infos_descending` proves *descending* from epoch 1 and still asserts `epoch_infos[0].index == 0`, `epoch_infos[1].index == 1`, under the comment "Results are ordered by epoch index (BTreeMap ordering), regardless of query direction".

**There is a smaller real issue underneath, though.** `resolve_current_epoch`'s correctness now rests on an implementation detail of another crate — that `rs-drive` happens to accumulate into a `BTreeMap` — while the type it consumes advertises no such guarantee (`RetrievedObjects` is an `IndexMap`, and its own doc comment says "You can sort the objects by key if you need a specific order").

If `verify_epoch_infos_v0` were ever changed to preserve the query's direction — a natural thing to want for descending queries — the breakage would be silent and remote from the change. With a hint below the tip, `started.next()` would yield the *higher* epoch, `candidate` would be reassigned to the value it already holds, and the loop would spin `MAX_CURRENT_EPOCH_REFINEMENTS` times and fail closed with `StaleNodeError::Epoch` — nothing in that error points at ordering.

So rather than a defensive sort that hides the coupling, it may be worth stating the invariant where it is depended upon: a doc line on `ExtendedEpochInfos as FromProof` that results are ascending by `EpochIndex`, and a `debug_assert!` on ascending order at the point in `resolve_current_epoch` that relies on it. A future change to the drive-side ordering then breaks loudly instead of turning into an unreachable-node report.

The other finding does look valid: `packages/rs-sdk-ffi/src/system/queries/platform_status.rs:79` and `:90` hard-code `"version":10`, while `LATEST_VERSION` is `PROTOCOL_VERSION_13` — so `dash_sdk_get_platform_status` under-reports the protocol version by at least three.

packages/rs-sdk/src/platform/types/epoch.rs
```diff
@@ -1,57 +1,179 @@
...
+            // a started epoch above it cannot happen in a well-formed state.
+            None => {
+                return Err(Error::InvalidProvedResponse(format!(
+                    "epoch {candidate} is not started but epoch {index} is; \
+                     epochs must be initialized contiguously"
+                )))
+            }
+            // The hint was below the chain tip; retry from the proven epoch.
+            Some((newer_index, _)) => candidate = newer_index,
+        }
```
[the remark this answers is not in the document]
QuantumExplorer: Good catch — confirmed and fixed in 94c6619.

`ExtendedEpochInfos` is `RetrievedObjects<EpochIndex, ExtendedEpochInfo>` = `IndexMap`, so order is whatever the response construction produced. Today `verify_epoch_infos_v0` folds through a `BTreeMap` keyed by the encoded epoch path, so insertion happens to be ascending and the code worked — but nothing guaranteed it, and the failure mode was real: a response listing the newer epoch first would be read as a stale hint pointing at the *older* one, and the refinement loop would spin until it failed closed.

Rather than sorting and re-reading positionally, selection now goes by index through a `BTreeMap`, so only the *set* of epochs in the proof matters:

```rust
match started.keys().next_back().copied() {
    None => return Err(Error::EpochNotFound),          // hint above every started epoch
    Some(newest) if newest == candidate => { /* candidate is current */ }
    Some(newest) if newest > candidate => candidate = newest,   // hint below the tip
    Some(newest) => return Err(Error::InvalidProvedResponse(..)), // below the query start
}
```

Pinned by a new mock test (`should_ignore_the_order_epochs_arrive_in`) that feeds the confirmation response with the newer epoch first; it fails against the previous positional logic.

packages/rs-sdk-ffi/src/system/queries/platform_status.rs
```diff
@@ -64,40 +67,29 @@ fn get_platform_status(sdk_handle: *const SDKHandle) -> Result<String, String> {
...
+                );
+                Ok(json)
+            }
+            Err(dash_sdk::Error::EpochNotFound) => {
+                // If no epochs found, return default values
+                let json = format!(
+                    r#"{{"version":{},"network":"{}","blockHeight":0,"coreHeight":0}}"#,
+                    10, network_str
+                );
+                Ok(json)
```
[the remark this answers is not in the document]
QuantumExplorer: Fixed in 94c6619. The hard-coded `10` predates this PR, but the function is being rewritten here anyway and reporting a wrong protocol version under a status API is worth closing out.

Success path uses `epoch.protocol_version()` — the current epoch's version straight off the verified proof. The `EpochNotFound` fallback has no epoch to read, so it uses `sdk.protocol_version_number()` (the highest version the SDK has verified). `ExtendedEpochInfoV0Getters` was already in scope. The success-path test now asserts the reported version too.

QuantumExplorer (commented):
Sure, why not

PR #4232, opened 2026-07-26
fix(drive-abci): refresh the contract cache when the v13 migration rewrites DPNS

QuantumExplorer:

A consensus divergence at the protocol version 13 activation block. Nodes split into two groups producing different app hashes from the same block, and the split did **not** follow node role — it followed cache history.

The v13 transition rewrites the DPNS contract in state:

```rust
self.drive.apply_contract(&dpns_contract, *block_info, true, None, Some(transaction), platform_version)?;
```

`apply_contract` writes straight to state. It does not go through the drive operation batch, and the batch is the only path that carries the `RemoveDataContractFromCache` finalization task. Nothing evicted the pre-migration DPNS from `drive.cache.data_contracts`.

So a node that had been up long enough to have read DPNS once kept serving its cached copy for the rest of the process lifetime, while a node whose cache was cold — freshly restarted, or the entry evicted under capacity pressure — read the migrated contract.

The two copies differ in a way that reaches state. DPNS has been stored with a **v0** `DataContractConfig` ever since it was registered at genesis under protocol version 1; the v13 contract carries a **v1** config. `DocumentV0::serialize` branches on exactly that:

```rust
if matches!(contract, DataContract::V0(_)) || matches!(contract.config(), DataContractConfig::V0(_)) {
    self.serialize_v0(document_type)   // leading varint 00
} else {
    ... 2 => self.serialize_v2(document_type),   // leading varint 02
}
```

A warm node wrote DPNS documents with a leading `00`; a cold node wrote the identical document with a leading `02`. One byte, a different Merk node hash, a different app hash. Everything else — paths, keys, flags, fees, the preceding app hash — matched.

Note that `perform_events_on_first_block_of_protocol_change_v0` already reloads the *separate* `system_data_contracts` cache. It is the ordinary `data_contracts` cache, the one document state transitions actually read through, that was missed.

## What was done?

Added `Drive::refresh_data_contract_cache_from_state`, and called it from `transition_to_version_13` for DPNS. The document history contract needs no refresh: it first comes into existence in this transition, and the cache holds no negative entries, so no node can hold a stale copy of it.

The helper re-reads the contract from state and re-seeds the cache with it. Two details are deliberate:

- **The read bypasses both caches** rather than going through `get_contract_with_fetch_info`. Read-only DAPI queries fetch with no transaction and `add_to_cache_if_pulled = true`, so a concurrent query thread could otherwise race a pre-migration copy back into the global cache between the eviction and the re-seed.
- **The result goes into the block cache, not the global cache.** The migration's write is still uncommitted at that point. The block cache is the first cache transactional reads consult, it is promoted to global when the block commits (`merge_and_clear_block_cache`), and dropped at the start of the next block if the block never commits (`clear_block_cache`).

Billing is unaffected: `fetch_contract_v0` computes the cached `OperationCost` with grovedb value caching disabled precisely so a contract fetch costs the same every time, and derives `fee` from that cost only when an epoch is supplied — so a cache hit seeded here bills identically to the cold fetch it replaces.

### Delayed-query race (review finding)

Review surfaced a residual race: a read-only query thread reads the pre-migration contract from committed state, is descheduled, and performs its `add_to_cache_if_pulled` insert only **after** the migrated contract is promoted to the global cache at finalize — clobbering it with the stale copy, and re-opening the divergence in a later block. The window is millisecond-scale, but the activation height is predictable and query flooding on public DAPI endpoints could widen it. The same window exists on the ordinary contract-update path, where the batch finalization task evicts during block execution.

Closed with a monotonic version guard in `DataContractCache::insert`: an insert carrying a **lower** `version()` than the cached entry is skipped, atomically per key via moka's compute API. Contract versions increase strictly monotonically — the update transition enforces `new == old + 1`, and the v13 DPNS rewrite goes 1 → 2 — so a lower-version insert is always a delayed writer, never fresh information. Same-version inserts still overwrite (the cache-hit fee-recalculation path relies on that). This closes the race for user contract updates as well, not just the migration.

The fix is scoped to the v13 transition rather than pushed down into `apply_contract` / `insert_contract`, so that replay of the v4 through v12 upgrade blocks stays bit-identical. The helper's doc comment states the rule for future migrations: every contract a migration **rewrites** (i.e. that existed before, and so may already be cached) must be refreshed.

## Consensus impact

**This changes the outcome of the v13 upgrade block relative to rc.1.** It is not a breaking change against any well-defined prior behaviour — the rc.1 behaviour is nondeterministic — but it does mean the fixed binary converges every node on the migrated contract, i.e. on `02` serialization. Replaying the preserved testnet state under rc.1 reproduces the bad app hash, confirming there is no intentional beta.2 → rc.1 change involved.

Mainnet has never activated PV13, so mainnet is unaffected. Testnet needs a coordinated rollback to before the activation height and a re-activation on fixed binaries; clearing caches on the current binary would make things worse, not better, because it flips an individual node onto the other side of the fork.

## How Has This Been Tested?

`test_transition_to_version_13_refreshes_warm_dpns_contract_cache` reproduces the testnet divergence: it puts state where testnet actually was (DPNS as written at genesis under protocol version 1, carrying a v0 config), warms the global cache the way a long-lived node does, asserts the pre-migration contract serializes preorders with the `00` prefix, runs the transition, and then asserts the transactional read returns the migrated contract and serializes the identical document with `02`.

Verified the test fails without the fix and passes with it. The pre-existing `test_transition_to_version_13_inserts_document_history_contract_and_updates_dpns` passes either way — it fetches against a cold cache, which is why the bug got through.

Two unit tests on the helper itself cover replacing a stale cached copy and seeding the block cache rather than the global cache. Three cache-level tests cover the version guard, including an end-to-end reproduction of the delayed-query race: stale contract read → migrated contract seeded into the block cache → promotion → delayed stale insert → the promoted contract survives.

- `cargo test -p drive-abci --lib protocol_upgrade` — 44 passed
- `cargo test -p drive --lib contract::` — 227 passed
- `cargo test -p drive --lib refresh_contract_cache` — 2 passed
- `cargo clippy -p drive -p drive-abci --all-targets` — clean
- `cargo check -p drive --no-default-features --features verify` — clean

## Breaking Changes

None for mainnet: PV13 has not activated there. Testnet requires a coordinated reset — see Consensus impact.

## Checklist:

### DIFF
--- a/packages/rs-drive/src/drive/contract/migration/refresh_contract_cache.rs
+++ b/packages/rs-drive/src/drive/contract/migration/refresh_contract_cache.rs
@@ -5,8 +5,11 @@ use grovedb::TransactionArg;
 
 impl Drive {
     /// Re-reads a data contract from state and re-seeds the in-memory data contract cache
-    /// with it. Call this from a protocol-upgrade migration for **every** contract the
-    /// migration writes.
+    /// with it. Call this from a protocol-upgrade migration for every contract the
+    /// migration **rewrites** — i.e. any contract that existed before the migration and
+    /// so may already sit in a node's cache. A contract the migration introduces for the
+    /// first time needs no refresh: the cache holds no negative entries, so no node can
+    /// have a stale copy of a contract that never existed.
     ///
     /// CONSENSUS-CRITICAL. Contracts written by a state transition go through the drive
     /// operation batch, whose `RemoveDataContractFromCache` finalization task evicts the
--- a/packages/rs-drive/src/drive/contract/migration/refresh_contract_cache.rs
+++ b/packages/rs-drive/src/drive/contract/migration/refresh_contract_cache.rs
@@ -1,0 +1,174 @@
+use crate::drive::Drive;
+use crate::error::Error;
+use dpp::version::PlatformVersion;
+use grovedb::TransactionArg;
+
+impl Drive {
+    /// Re-reads a data contract from state and re-seeds the in-memory data contract cache
+    /// with it. Call this from a protocol-upgrade migration for **every** contract the
+    /// migration writes.
+    ///
+    /// CONSENSUS-CRITICAL. Contracts written by a state transition go through the drive
+    /// operation batch, whose `RemoveDataContractFromCache` finalization task evicts the
+    /// superseded copy from the cache. Migrations write contracts directly
+    /// (`insert_contract` / `apply_contract`) and bypass that machinery entirely, so
+    /// without this call a node that already holds the pre-migration contract in its
+    /// global cache keeps serving that copy for the rest of the process lifetime, while a
+    /// node whose cache is cold (freshly restarted, or the entry was evicted under
+    /// capacity pressure) reads the migrated one. The two nodes then serialize documents
+    /// of that contract against different `DataContract`s — a difference that reaches
+    /// state, because `DocumentV0::serialize` picks its serialization version from the
+    /// contract's config version — and produce different app hashes from the same block.
+    ///
+    /// The refreshed contract is placed in the **block** cache, not the global cache: at
+    /// this point the migration's write is still uncommitted, and the block cache is the
+    /// only cache that transactional reads consult first. It is promoted to the global
+    /// cache once the block commits (`merge_and_clear_block_cache`), and dropped at the
+    /// start of the next block if the block never commits (`clear_block_cache`).
+    ///
+    /// The read deliberately bypasses both caches rather than going through
+    /// `get_contract_with_fetch_info`: a concurrent read-only query thread (which reads
+    /// committed state, with no transaction, and does populate the global cache) could
+    /// otherwise race a pre-migration copy back into the global cache between the
+    /// eviction and the re-seed.
+    ///
+    /// Billing is unaffected. `fetch_contract_v0` computes the cached `OperationCost`
+    /// with grovedb value caching disabled precisely so the cost of a contract fetch is
+    /// deterministic, and derives `fee` from that cost only when an epoch is supplied —
+    /// so a cache hit seeded here bills identically to the cold fetch it replaces.
+    pub fn refresh_data_contract_cache_from_state(
+        &self,
+        contract_id: [u8; 32],
+        transaction: TransactionArg,
+        platform_version: &PlatformVersion,
+    ) -> Result<(), Error> {
+        // Cache-bypassing read: the contract exactly as state now holds it.
+        let maybe_fetch_info = self.fetch_contract_and_add_operations(
+            contract_id,
+            None,
+            transaction,
+            &mut vec![],
+            platform_version,
+        )?;
+
+        // Drop the pre-migration copy from both the block and the global cache.
+        self.cache.data_contracts.remove(contract_id);
+
+        // Re-seed with what state now holds. A migration always writes the contract it
+        // asks us to refresh, so `None` here means the contract genuinely is not in
+        // state; leaving the caches empty is then the correct outcome.
+        if let Some(fetch_info) = maybe_fetch_info {
+            self.cache
+                .data_contracts
+                .insert(fetch_info, transaction.is_some());
+        }
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::drive::contract::DataContractFetchInfo;
+    use crate::util::test_helpers::setup::setup_drive_with_initial_state_structure;
+    use dpp::block::block_info::BlockInfo;
+    use dpp::data_contract::accessors::v0::{DataContractV0Getters, DataContractV0Setters};
+    use dpp::system_data_contracts::{load_system_data_contract, SystemDataContract};
+    use dpp::version::PlatformVersion;
+    use std::sync::Arc;
+
+    /// A contract written directly to state must replace a stale cached copy, so that a
+    /// warm node and a cold node read the same contract afterwards.
+    #[test]
+    fn test_refresh_replaces_stale_cached_contract() {
+        let drive = setup_drive_with_initial_state_structure(None);
+        let platform_version = PlatformVersion::latest();
+        let transaction = drive.grove.start_transaction();
+
+        let dpns = load_system_data_contract(SystemDataContract::DPNS, platform_version)
+            .expect("expected to load DPNS");
+        let contract_id = dpns.id().to_buffer();
+
+        // Warm the global cache with a contract that does NOT match state: a distinct
+        // version stands in for the pre-migration copy a long-lived node would hold.
+        let mut stale = dpns.clone();
+        stale.set_version(u32::MAX);
+        drive.cache.data_contracts.insert(
+            Arc::new(DataContractFetchInfo {
+                contract: stale,
+                storage_flags: None,
+                cost: Default::default(),
+                fee: None,
+            }),
+            false,
+        );
+
+        drive
+            .apply_contract(
+                &dpns,
+                BlockInfo::default(),
+                true,
+                None,
+                Some(&transaction),
+                platform_version,
+            )
+            .expect("expected to apply contract");
+
+        // Without the refresh this still reads the stale copy out of the global cache.
+        drive
+            .refresh_data_contract_cache_from_state(
+                contract_id,
+                Some(&transaction),
+                platform_version,
+            )
+            .expect("expected to refresh cache");
+
+        let refreshed = drive
+            .get_contract_with_fetch_info(contract_id, false, Some(&transaction), platform_version)
+            .expect("expected to fetch contract")
+            .expect("expected the contract to be present");
+
+        assert_eq!(refreshed.contract.version(), dpns.version());
+    }
+
+    /// The refresh must seed the block cache, not the global cache: the migration's write
+    /// is still uncommitted, so a non-transactional reader must not see it yet.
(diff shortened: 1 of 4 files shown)


thepastaclaw:
⛔ Blockers found — Sonnet deferred (commit 89d2602b)

QuantumExplorer (commented):
Reviewed

packages/rs-drive/src/drive/contract/migration/refresh_contract_cache.rs
```diff
@@ -0,0 +1,177 @@
...
+        // Drop the pre-migration copy from both the block and the global cache.
+        self.cache.data_contracts.remove(contract_id);
+
+        // Re-seed with what state now holds. A migration always writes the contract it
+        // asks us to refresh, so `None` here means the contract genuinely is not in
+        // state; leaving the caches empty is then the correct outcome.
+        if let Some(fetch_info) = maybe_fetch_info {
+            self.cache
+                .data_contracts
+                .insert(fetch_info, transaction.is_some());
```
thepastaclaw, line 73: **🔴 Blocking: Delayed DAPI cache fill can overwrite migrated DPNS after promotion**

The remove-and-reseed sequence is not synchronized with cache-populating non-transactional reads. After `remove`, a DAPI document query can miss the global cache, fetch the old DPNS from committed GroveDB through `get_contract_with_fetch_info_and_add_to_operations_v0`, and pause before its unconditional global-cache insertion. The migration meanwhile inserts the new contract into the block cache, and `update_drive_cache` promotes it through `merge_and_clear_block_cache`; when the query resumes, its later global-cache insert replaces the migrated entry with the old contract. The query service's committed-height retry does not repair this because the cache mutation occurs before the height recheck and is not rolled back; a retry can then read the same stale cache entry. At the next block, `clear_block_cache` leaves that global entry intact, so transactional execution can again serialize DPNS documents using the stale `00` format instead of `02`, producing node-dependent app hashes. Publication must use per-key synchronization, generation-aware conditional insertion, or equivalent commit/snapshot ordering that prevents a fetch begun against old committed state from overwriting the migrated entry.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The refresh helper fixes the deterministic warm-cache case by bypassing both caches and seeding transactional reads with the migrated DPNS contract. It remains unsafe against a concurrent non-transactional DAPI fetch: a fetch started against pre-migration committed state can publish its stale result after the migrated block-cache entry is promoted, restoring the cache-history-dependent consensus divergence.

### Review provenance

🔴 1 blocking

<details>

```

In `packages/rs-drive/src/drive/contract/migration/refresh_contract_cache.rs`:
- [BLOCKING] packages/rs-drive/src/drive/contract/migration/refresh_contract_cache.rs:57-66: Delayed DAPI cache fill can overwrite migrated DPNS after promotion
  The remove-and-reseed sequence is not synchronized with cache-populating non-transactional reads. After `remove`, a DAPI document query can miss the global cache, fetch the old DPNS from committed GroveDB through `get_contract_with_fetch_info_and_add_to_operations_v0`, and pause before its unconditional global-cache insertion. The migration meanwhile inserts the new contract into the block cache, and `update_drive_cache` promotes it through `merge_and_clear_block_cache`; when the query resumes, its later global-cache insert replaces the migrated entry with the old contract. The query service's committed-height retry does not repair this because the cache mutation occurs before the height recheck and is not rolled back; a retry can then read the same stale cache entry. At the next block, `clear_block_cache` leaves that global entry intact, so transactional execution can again serialize DPNS documents using the stale `00` format instead of `02`, producing node-dependent app hashes. Publication must use per-key synchronization, generation-aware conditional insertion, or equivalent commit/snapshot ordering that prevents a fetch begun against old committed state from overwriting the migrated entry.
```
</details>

QuantumExplorer (commented):
Reviewed

shumkov:
Notes from working the same bug from the other side (system-contract cache redesign, `fix/system-contracts-cache-redesign`). I've adopted the shape of `refresh_data_contract_cache_from_state` there and credited this PR — cache-bypassing read, evict both caches, seed the block cache is the right design. Three findings, the first two consensus-relevant.

### 1. The block-cache seed is wiped before anything reads it

`refresh_data_contract_cache_from_state` is called from `transition_to_version_13`, i.e. inside `perform_events_on_first_block_of_protocol_change_v0`. In the real block flow that runs **before** the block cache is cleared:

- `execution/engine/run_block_proposal/mod.rs:118` → `perform_events_on_first_block_of_protocol_change(...)`
- `execution/engine/run_block_proposal/mod.rs:139` → `run_block_proposal_v0(...)`
- `execution/engine/run_block_proposal/v0/mod.rs:123` → `clear_drive_block_cache(...)` → `data_contracts.clear_block_cache()`

So the seed is discarded before any state transition reads it. Verified empirically: seed the block cache, call `clear_drive_block_cache`, repeat the transactional read — it returns the pre-migration contract.

The PR still fixes the reported bug, but through the `remove()`, not the seed: with the global entry gone, the transactional read falls through to a real grovedb fetch inside the transaction. Worth knowing, because the rejection-safety property attributed to the seed ("dropped at the start of the next block if the block never commits") isn't what's providing it.

### 2. Eviction alone is defeated by a concurrent query

Once the seed is gone the mechanism is eviction-only, and that has a window:

- a transactional lookup falls back to the global cache on a block-cache miss — `cache/data_contract.rs:47`
- read-only queries read committed state with no transaction and populate the global cache — `query/document_query/v0/mod.rs:173`, `add_to_cache_if_pulled = true`
- the batch document transformer passes `add_to_cache_if_pulled = false` — `batch/transformer/v0/mod.rs:420` — so it never populates the block cache and re-runs that lookup on every read

A DPNS `getDocuments` query landing between the eviction and the transformer's read puts the pre-migration definition straight back, and the activation block serializes against it. Deterministic repro on my branch (seed disabled, everything else identical): the transformer read resolves DPNS version 1 / `keeps_transfer_history = false` instead of version 2 / `true`.

Same divergence as the original bug, now decided by query timing rather than restart history. The PR's test can't see it, since it calls `transition_to_version_13` directly and so never runs the clear or an interleaving query.

**Fix that worked for me:** move `clear_drive_block_cache` from `run_block_proposal_v0` up into `run_block_proposal`, above the protocol-change branch, so the order becomes clear → migrate + seed → execute. The seed then survives and transactional reads hit it before ever reaching the global entry. The only other behavioural change is that a proposal rejected by the `next_block_to` height check also clears the block cache, which is harmless.

### 3. The monotonic version guard is unsound as a general mechanism

`DataContractCache::insert` skipping inserts that carry a lower contract version assumes the version changes whenever the bytes do. True for user contract updates (`validate_update/v0/mod.rs:54` requires `old + 1`) and for the v13 DPNS rewrite (1 → 2), but not in general: `drive/contract/migration/strip_unknown_document_schema_properties.rs` rewrites stored contract schemas **without** incrementing the version, and the guard overwrites on equal versions. So a delayed stale writer at version N can still clobber migrated bytes at version N.

The delayed-writer window it targets is real and I haven't closed it either. It probably needs a generation token bumped on `merge_and_clear_block_cache` — readers capture it before their fetch, inserts rejected if it advanced — rather than the contract version, which isn't a validity token.

---

One thing I'd suggest keeping from this PR regardless of which fix lands: the test asserting the literal `00` vs `02` document prefix is better evidence than the property-level assertions I wrote, because it pins the actual consensus-visible byte.

shumkov:
Closing as superseded — #4222 merged into `v4.1-dev` and carries this fix.

The shape of `refresh_data_contract_cache_from_state` came from this PR and is credited in the merged commit: cache-bypassing read, evict both tiers, seed the block cache, and the billing argument for why a seeded hit costs what a cold fetch costs. What #4222 adds on top:

- **Generalised beyond DPNS** — it refreshes every system contract, iterating `SystemDataContract::ALL`, so a future migration rewriting a different contract cannot silently skip it.
- **The ordering fix that makes the block-cache seed actually work.** As noted [above](https://github.com/dashpay/platform/pull/4232#issuecomment-5083017251), the seed here was wiped before anything read it: `perform_events_on_first_block_of_protocol_change` runs at `run_block_proposal/mod.rs:118`, and `run_block_proposal_v0` cleared the block cache at line 123 immediately afterwards. `clear_drive_block_cache` moved up above the protocol-change branch, so the order is now clear → migrate and seed → execute. Regression-pinned at the production entry point, verified by moving the clear back and watching the tests fail.
- **Gated behind a feature version.** The refresh lives in a new `perform_events_on_first_block_of_protocol_change_v1`, selected by protocol version 13 only (via `DRIVE_ABCI_METHOD_VERSIONS_V9`, which no other protocol version uses), rather than changing `v0` in place.

One thing deliberately **not** carried over: the monotonic contract-version guard on `DataContractCache::insert`. It assumes the contract version changes whenever the bytes do, and `drive/contract/migration/strip_unknown_document_schema_properties.rs` rewrites stored schemas without incrementing it — while the guard overwrites on equal versions. So a delayed writer at version N can still clobber migrated bytes at version N.

The delayed-writer race that guard was aimed at is real and still open. It is being worked separately on `fix/data-contract-cache-delayed-writer`, because the generation-token approach both reviews suggested does not work as written either: `merge_and_clear_block_cache` runs at `finalize_block.rs:48` while the grovedb commit is at line 70, so there is no point where "generation bumped" implies "committed state is current".

PR #4233, opened 2026-07-26
feat(dashmate)!: bump gateway Envoy to 1.39.0

shumkov:

The gateway was pinned to `dashpay/envoy:1.35.11-impr.1`. Envoy v1.39.0 fixes an HTTP/2
PRIORITY/WINDOW_UPDATE flood-protection bypass (protection could be evaded by rapidly opening and
closing streams) that was **not** backported to the 1.35–1.38 patch lines, so the fix is only
reachable by moving to the 1.39 line.

Bumping that pin also had no repeatable pre-check: nothing verified that the config rendered from
`envoy.yaml.dot` still loads on a candidate release, or that the pieces of this gateway that only
exist at runtime — the rate limit service wire protocol and the grpc-web over-limit reply — still
work. This PR adds that gate and then performs the bump with it.

### Validation harness — `packages/dashmate/scripts/validate-envoy-config.js`

- Renders `templates/platform/gateway/envoy.yaml.dot` through dashmate's own renderer for 8 config
  variants covering every branch of the template (TLS provider, rate limiter on/off, metrics/admin,
  stdout/stderr/file × text/JSON access logs, all features at once), then runs Envoy's built-in
  validator — `envoy --mode validate` — on each rendering inside a given image. Hard errors and
  deprecated-field warnings are reported per variant.
- `--smoke` additionally boots redis, the pinned `envoyproxy/ratelimit` and the Envoy image under
  test on a throwaway Docker network and asserts: Envoy reaches `starting workers` with no warnings,
  exactly `requestsPerUnit` requests pass the limiter and the next is rejected, the grpc-web
  over-limit reply is `HTTP 200` + `grpc-status: 8` + `ratelimit-reset`, and the native gRPC
  over-limit reply is `HTTP 200` + `grpc-status: 8`.
- With no image argument the baseline is derived from the pinned image, so the default run always
  targets what dashmate currently ships. Bare tags resolve against `envoyproxy/envoy`; full refs
  (including the custom `dashpay/envoy` image, whose entrypoint is the hot-restart supervisor) work too.
- Procedure, the caveats it does not cover, and the v1.35.11 → v1.39.0 assessment are documented in
  `packages/dashmate/docs/services/gateway-envoy-upgrade.md`, linked from `docs/services/gateway.md`.

### The bump

- `configs/defaults/getBaseConfigFactory.js`: `platform.gateway.docker.image` →
  `dashpay/envoy:1.39.0-impr.1`.
- New config-file migration in `configs/getConfigFileMigrationsFactory.js` keyed `4.1.0-rc.2`.
  Existing configs store resolved values, so the base-config change alone leaves operators on 1.35.11.
  Following the `3.0.2` precedent, it re-pins only images matching `^dashpay/envoy:1\.35\.` and pulls
  the new value from the base config, leaving a deliberately customised image (private fork,
  vendor-patched build, floating tag) untouched.
- Docs: `docs/config/gateway.md`, `docs/update.md`.

### Behaviour changes that come with 1.39.0 (no config change needed)

- **Overload actions keyed on `global_downstream_max_connections` start firing.** v1.37.0 fixed that
  monitor — previously its actions never triggered — so `disable_http_keepalive` at 95% and
  `stop_accepting_connections` at 100% of `platform.gateway.maxConnections` become live. At ~950 of
  1000 connections Envoy will GOAWAY-drain HTTP/2 connections. Confirmed by the new
  `overload.envoy.resource_monitors.global_downstream_max_connections.pressure` gauge, absent on
  v1.35.11 and present on v1.39.0.
- **HTTP/2 flood protection scales with active streams.** Worth watching
  `http.ingress_http.http2.inbound_window_update_frames_flood`; revert guard is
  `envoy.reloadable_features.http2_flood_protection_active_streams`.

Both are documented with their revert guards in the new doc, along with lower-risk notes (upstream
`max_concurrent_streams` default, tls_inspector client TLS version enforcement, transport failure
reasons dropped from response bodies, `HeaderMatcher` multi-value matching, ratelimit `timeout: 0s`
semantics) — none of which require config edits.

## How Has This Been Tested?

- **Harness, v1.39.0**: 8/8 variants validate clean and 6/6 smoke checks pass on
  `envoyproxy/envoy:v1.39.0` and on `dashpay/envoy:1.39.0-impr.1` (both its amd64 and arm64
  manifests), identical to the `v1.35.11` baseline, which was also run through
  `dashpay/envoy:1.35.11-impr.1`. No field in the rendered config is deprecated or removed at 1.39.0.
- **Harness can fail** — verified before trusting it: a bogus field injected into the template makes
  the run report `FAILED` and exit 1 (exit 0 once reverted), and a config using a known-deprecated
  field (`UpstreamTlsContext.enforce_rsa_key_usage`) produces the `Deprecated field:` warning the
  harness greps for.
- **Migration** — no automated test: a unit test pinning it was written and verified red→green
  (✖ `-dashpay/envoy:1.35.11-impr.1 / +dashpay/envoy:1.39.0-impr.1` before the migration, ✔ after),
  then dropped at review request. The remaining `migrate v0.25.0 config file to the latest one` test
  does not exercise the 1.35.x match or the customised-image carve-out for the 3.0.0+ cohort, so that
  behaviour is currently unguarded.
- **Hot-restart supervisor** (outside what the harness covers, since it overrides the entrypoint to
  reach Envoy's flags) checked by hand on `dashpay/envoy:1.39.0-impr.1`: boots through
  `hot-restarter.py` at epoch 0, `SIGHUP` forks a child at epoch 1 and the parent exits cleanly —
  the zero-downtime path dashmate uses after certificate renewal.
- **dashmate unit suite**: 157 passing. ESLint clean on all changed files.
- **e2e: incomplete — no full-suite verdict.** The platform test suite was run against a local 3-node
  network with all three gateways recreated on `dashpay/envoy:1.39.0-impr.1`. Everything not dependent
  on block production passed (`getBlockchainStatus`, `getStatus`, `getEpochsInfo`,
  `subscribeToBlockHeadersWithChainLocks` historical streaming), and Envoy access logs show correct
  unary and streaming proxying, including the 300s `stream_idle_timeout` firing with response flag
  `SI`. Five specs failed on `before all` timeouts, but that was an environment fault of my own making
  (the local miner — a detached `docker exec` loop in the seed's core container, recreated only by
  `dashmate group start` — was killed by a `group restart` attempt, so the chain stopped producing
  blocks and no wallet funding could confirm). The miner was restored and the suite restarted, but the
  environment was torn down before it finished. **A clean full-suite run on 1.39.0 is still
  outstanding and should gate the merge.**

## Breaking Changes

No API or config-schema breaks. Marked `!` because upgrading operators get a new gateway binary whose
load-shedding behaviour differs: the overload actions bound to `global_downstream_max_connections`
begin taking effect, so a node at ~95% of `platform.gateway.maxConnections` starts draining HTTP/2
connections instead of silently ignoring the thresholds. Operators who relied on the previous
(non-functioning) behaviour should review `platform.gateway.maxConnections`.

## Checklist:

### DIFF
--- a/packages/dashmate/configs/defaults/getBaseConfigFactory.js
+++ b/packages/dashmate/configs/defaults/getBaseConfigFactory.js
@@ -179,7 +179,7 @@ export default function getBaseConfigFactory() {
         },
         gateway: {
           docker: {
-            image: 'dashpay/envoy:1.35.11-impr.1',
+            image: 'dashpay/envoy:1.39.0-impr.1',
           },
           maxConnections: 1000,
           maxHeapSizeInBytes: 125000000, // 1 Gb
--- a/packages/dashmate/configs/getConfigFileMigrationsFactory.js
+++ b/packages/dashmate/configs/getConfigFileMigrationsFactory.js
@@ -1561,6 +1561,25 @@ export default function getConfigFileMigrationsFactory(homeDir, defaultConfigs) 
 
         return configFile;
       },
+      '4.1.0-rc.2': (configFile) => {
+        // Move the Platform Gateway onto the Envoy 1.39 line. Only configs still
+        // carrying the previously shipped 1.35.x image are re-pinned; an image
+        // the operator chose themselves (private fork, vendor-patched build,
+        // floating tag) is left alone. Pulled from the base config so it tracks
+        // whatever is pinned there.
+        // Keyed at the next release, not the released 4.1.0-rc.1: the runner
+        // skips fromVersion===toVersion, so a key equal to an operator's current
+        // version never fires.
+        Object.entries(configFile.configs)
+          .forEach(([, options]) => {
+            const docker = options.platform?.gateway?.docker;
+            if (docker && /^dashpay\/envoy:1\.35\./.test(docker.image)) {
+              docker.image = base.get('platform.gateway.docker.image');
+            }
+          });
+
+        return configFile;
+      },
     };
   }
 
--- a/packages/dashmate/docs/config/gateway.md
+++ b/packages/dashmate/docs/config/gateway.md
@@ -6,7 +6,7 @@ The `platform.gateway` section configures the Dash Platform Gateway, which serve
 
 | Option | Description | Default | Example |
 |--------|-------------|---------|---------|
-| `platform.gateway.docker.image` | Docker image for Gateway | `dashpay/envoy:1.35.11-impr.1` | `dashpay/envoy:latest` |
+| `platform.gateway.docker.image` | Docker image for Gateway | `dashpay/envoy:1.39.0-impr.1` | `dashpay/envoy:latest` |
 
 ## Listeners
 
--- a/packages/dashmate/docs/services/gateway-envoy-upgrade.md
+++ b/packages/dashmate/docs/services/gateway-envoy-upgrade.md
@@ -1,0 +1,130 @@
+# Validating an Envoy version bump for the Gateway
+
+The gateway runs Envoy with a config rendered from
+[`templates/platform/gateway/envoy.yaml.dot`](../../templates/platform/gateway/envoy.yaml.dot),
+and the image is pinned in [`configs/defaults/getBaseConfigFactory.js`](../../configs/defaults/getBaseConfigFactory.js)
+(`platform.gateway.docker.image`). `dashpay/envoy:<version>-impr.N` is a stock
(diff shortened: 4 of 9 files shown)


thepastaclaw:
⛔ Blockers found — Sonnet deferred (commit 93636a50)
_Canonical validated blockers: 3_

QuantumExplorer (approved):

PR #4237, opened 2026-07-27
fix(platform-wallet): make shielded re-bind non-destructive so it cannot wipe an in-flight sync pass

QuantumExplorer:

On iOS (rc.2, testnet), a shielded note discovered by a sync pass was not spendable until the app restarted — `IdentityCreateFromShieldedPool` / unshield immediately after a shield failed with `ShieldedNoUnspentNotes` even though the host's persisted note row existed and was unspent. A second defect in the same area: after a from-scratch rescan, the sync watermark never advanced in-session, so every subsequent pass rescanned the entire tree until restart.

Both have one root cause. Hosts re-run `bind_shielded` liberally (the example app fires it twice at launch — a direct `rebindWalletScopedServices()` call plus the wallet-set `onChange` observer — and again on Sync Now and wallet navigation). `PlatformWallet::install_shielded_views` ran unregister → purge → restore on **every** bind. Against an in-flight `sync_notes_across` pass (which holds the store write lock for the whole streamed pass, 45–90 s on testnet), the purge/restore lock acquisitions queue behind the pass and land the moment it finishes — wiping the pass's freshly discovered notes and watermark from the coordinator's in-memory store and restoring a persister snapshot loaded *before* the pass ran. The host's SwiftData rows survive (the pass's changeset persists synchronously first), which is why an app restart always healed it.

Log fingerprint from the failing sessions: Core SPV auto-start (main-actor-queued behind the second bind) begins <0.5 s after pass 1 ends; the next pass reads the pre-pass watermark and re-discovers the same note (`new_notes_total=1` twice for one on-chain note). A related interleaving also produced `WARN Shielded sync changeset dropped: no persister registered` — `register_wallet` inserted `accounts` and `persisters` under two separate lock acquisitions, so a pass starting between them dropped its whole changeset.

## What was done?

- **Re-binds are non-destructive by construction.** `install_shielded_views` no longer calls `unregister_wallet` at all. `register_wallet` now purges only the subwallets the new registration *drops or re-keys* (via the new `ShieldedStore::purge_subwallet`; an account whose FVK changed is treated as dropped — its stored notes belong to the old key) and it replaces, never removes, the persister handle. Retained accounts keep their in-memory notes and watermark across **any** re-bind — identical or account-set-changing — so a registration replacement racing an in-flight sync pass cannot wipe the pass's results, and a pass finishing mid-bind always finds a persister for its changeset.
- **Idempotent re-bind fast path**: when the wallet is already registered with the exact same account → FVK map (new `NetworkShieldedCoordinator::wallet_registration_matches`, compared by the canonical 96-byte FVK encoding) *and* its hydration previously succeeded, the bind skips reloading/re-applying the persister snapshot entirely and never touches the store lock. Side effect: the second launch-time bind no longer blocks the main actor behind the in-flight pass (~50 s launch freeze gone).
- **Hydration is tracked separately from registration** (review finding): the coordinator records per-wallet hydration success (set only after a successful load + restore for the current registration shape; cleared on shape change, unregister, and clear). A matching registration alone no longer short-circuits the restore, so a transient persistence failure on the first bind is retried by the next re-bind instead of leaving notes and the watermark absent until a rescan or restart.
- **`restore_for_wallet` is monotonic**: the watermark only advances (`max`), and a snapshot note whose nullifier the store already tracks is skipped — `save_note` is overwrite-by-nullifier, so a stale `is_spent = false` copy would otherwise re-offer a spent note to selection (a double-spend attempt at broadcast).
- **Registry lifecycle mutations are atomic** (review finding): a coordinator-level lifecycle mutex serializes `register_wallet` / `unregister_wallet` / `clear`'s registry phase, and `register_wallet` inserts the persister before the accounts. Together these make "accounts visible ⇒ persister visible" hold at every interleaving — including concurrent register + unregister — so a pass can no longer silently drop its changeset (the only remaining accounts-without-persister window is a wallet genuinely being removed mid-pass, where dropping its changeset is intended).
- **Removed the stale warning** `next_start_index is 0 after scanning N notes — next sync will rescan from the beginning`: since the streaming refactor that value is diagnostic-only (the watermark is `aligned_start + total_notes_scanned`; the partial buffer chunk is re-covered by chunk-boundary realignment on the next pass), and it fired spuriously on every single-partial-batch cold scan — it pointed the field investigation of this bug at the wrong mechanism.

## How Has This Been Tested?

- Six new regression tests in the coordinator test module: a stale pre-pass snapshot can neither rewind the watermark nor clobber a fresh note; a spent note can never be resurrected by restore (only genuinely new nullifiers are added); `wallet_registration_matches` accepts only the identical wallet + account set + FVK and rejects unknown wallets, changed account sets, and changed keys; a changed-set re-register keeps retained-account state (notes + watermark + persister) while purging only the dropped subwallet; a re-keyed account's old-key notes are purged; and the hydration flag follows the registration lifecycle (preserved on identical re-register, cleared on shape change / unregister).
- `cargo test -p platform-wallet --features shielded --lib`: 629 passed, 0 failed.
- `cargo check -p platform-wallet --all-targets --features shielded` and `cargo check -p platform-wallet-ffi --features shielded` clean; `cargo fmt` applied.
- Root cause was verified against device logs from four app sessions (two failing, two healthy restarts): the stomped watermark values and the SPV-start timing match the mechanism exactly in every session.

### End-to-end verification on the iOS simulator (this build)

Rebuilt the sim xcframework + SwiftExampleApp from this branch (`build_info.txt` in the session log confirms commit `da4e45a3`, clean tree) and re-ran the originally failing shield-then-spend flow on testnet, all in **one app session with no restart**:

| Time (UTC) | Event |
|---|---|
| 07:39:24 | Launch; pass 1 starts with restored watermark 2268 |
| 07:42:51 | Shield 0.05 tDASH (Type 18 asset lock) submitted from the UI |
| 07:43:27 | Shield executed on-chain — 4,787,148,800 credits into the pool |
| 07:45:02 | Sync pass discovers the note (`new_notes_total=1`, watermark → 2274) |
| 07:48:47 | `IdentityCreateFromShieldedPool` (denomination 3B credits): **selection succeeds instantly** — `inputs=1 total_input=4787148800`, i.e. exactly the note discovered 3½ min earlier |
| 07:48:54 | Broadcast succeeded → new identity `HVvuKeae6bpLsBNUZK9gN8rYEK7vWxa4Vr7r4pVWD6f1` |
| 07:49:37 | Next pass discovers the change note → watermark 2276 |

The 3B-credit denomination could only be covered by the fresh note (the only other unspent note was 1.787B), so the spend provably selected the note discovered mid-session — the exact case that returned `ShieldedNoUnspentNotes` before this fix.

Both defects confirmed fixed in the session logs:
- **Watermark monotonic across every pass** (2268 → 2272 → 2274 → 2276). The launch even reproduced the original stomp setup — the second bind racing pass 1 with a stale snapshot (2268) — and the following pass read pass 1's advanced 2272, not the snapshot's value.
- **Fast path observable**: Core SPV auto-start began 0.7 s into pass 1 instead of ~50 s after it (pre-fix the second bind blocked the main actor on the store lock for the whole pass).
- Zero WARN/ERROR lines for the entire session (no dropped-changeset warning, no spurious rescan warning), and SwiftData ground truth is coherent afterwards: the fresh note is marked spent, its change note unspent, and the new identity row is present at slot #7.

## Breaking Changes

None. `wallet_registration_matches` is a new public method; existing APIs and persisted formats are unchanged.

## Checklist:

### DIFF
--- a/packages/rs-platform-wallet/src/wallet/platform_wallet.rs
+++ b/packages/rs-platform-wallet/src/wallet/platform_wallet.rs
@@ -114,6 +114,18 @@ pub struct PlatformWallet {
     /// cloned wallet handles share the one lock.
     #[cfg(feature = "shielded")]
     pub(crate) shield_guard: Arc<tokio::sync::Mutex<()>>,
+    /// Set once this wallet has been removed from the manager, to stop
+    /// a handle that outlives the removal from binding shielded state
+    /// back onto the coordinator. Callers resolve an
+    /// `Arc<PlatformWallet>` and then hold it across a bind that can
+    /// take arbitrarily long (it may resolve a mnemonic through the
+    /// host), so a removal can complete in the middle; without this
+    /// flag the bind would re-register the wallet the removal just
+    /// detached, and the next sync pass would re-fetch and re-persist
+    /// shielded history the host believes it deleted. `Arc` so cloned
+    /// wallet handles observe the one flag.
+    #[cfg(feature = "shielded")]
+    pub(crate) shielded_detached: Arc<std::sync::atomic::AtomicBool>,
 }
 
 impl PlatformWallet {
@@ -468,6 +480,8 @@ impl PlatformWallet {
             shielded_keys: Arc::new(RwLock::new(None)),
             #[cfg(feature = "shielded")]
             shield_guard: Arc::new(tokio::sync::Mutex::new(())),
+            #[cfg(feature = "shielded")]
+            shielded_detached: Arc::new(std::sync::atomic::AtomicBool::new(false)),
         }
     }
 
@@ -533,6 +547,36 @@ impl PlatformWallet {
             account_views.insert(account, ks.viewing_keys());
         }
 
+        // Refuse to overwrite a persisted viewing key with a different
+        // one. The durable notes, activity and sync watermark this
+        // wallet already has are keyed by `(wallet_id, account_index)`
+        // alone — nothing records which key produced them — so a row
+        // written under the old key is indistinguishable from one
+        // written under the new. Upserting the new key would leave the
+        // old key's notes attributed to it (unspendable, yet counted)
+        // and its watermark in force (hiding the new key's own
+        // history). No legitimate flow re-keys an account in place, so
+        // treat it like the malformed-row case below: surface it rather
+        // than silently mixing two keys' state. The recovery is a
+        // shielded Clear, which drops both sides at once.
+        let start = self.persister.load().map_err(|e| {
+            PlatformWalletError::ShieldedBuildError(format!(
+                "persister load failed while binding shielded viewing keys: {e}"
+            ))
+        })?;
+        for (account, views) in &account_views {
+            let id = SubwalletId::new(self.wallet_id, *account);
+            if let Some(persisted) = start.shielded.viewing_keys.get(&id) {
+                if persisted.as_slice() != views.to_fvk_bytes().as_slice() {
+                    return Err(PlatformWalletError::ShieldedKeyDerivation(format!(
+                        "persisted shielded viewing key for account {account} differs from the \
+                         one derived from this seed; clear shielded state before binding a \
+                         re-keyed account"
+                    )));
+                }
+            }
+        }
+
         // Persist the viewing keys while the seed is legitimately present, so
         // every later launch can rebind seedlessly. Do not install the in-memory
         // keys when persistence rejects the write: that would advertise a
@@ -555,7 +599,10 @@ impl PlatformWallet {
                 ))
             })?;
 
-        self.install_shielded_views(account_views, coordinator, None)
+        // Hand the snapshot loaded above to the install step: it is the
+        // same start state the restore needs, minus the viewing-key
+        // rows just written (which the restore doesn't consume).
+        self.install_shielded_views(account_views, coordinator, start)
             .await
     }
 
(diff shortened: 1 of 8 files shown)


thepastaclaw:
🕓 Ready for review — 2 ahead in queue (commit 390f0914)
_Queue position: 3/3_

packages/rs-platform-wallet/src/wallet/platform_wallet.rs
```diff
@@ -638,44 +638,75 @@ impl PlatformWallet {
...
-            Ok(start) => {
-                if let Err(e) = coordinator
-                    .restore_for_wallet(self.wallet_id, &start.shielded)
-                    .await
-                {
+            Ok(start) => match coordinator
+                .restore_for_wallet(self.wallet_id, &start.shielded)
+                .await
+            {
+                Ok(()) => coordinator.mark_hydrated(self.wallet_id, true).await,
```
thepastaclaw: **🔴 Blocking: Serialize the complete shielded bind lifecycle**

The coordinator lifecycle mutex protects only `register_wallet`; the wallet-local key publication, registration comparison, restore, and hydration commit remain separate operations. Starting from hydrated registration A, bind A can write local keys A and observe `identical = true`; bind B can then write local keys B and register B; bind A can subsequently register A. The coordinator ends on A while `self.shielded_keys` remains B, so sync decrypts with A while address, balance, and spend paths use B. The same gap affects removal and clear: the FFI bind function retains a wallet `Arc` after releasing manager lookup protection, so a bind can register a wallet after `remove_wallet` has completed. In addition, `restore_for_wallet` snapshots registered IDs before taking the store lock, allowing unregister or clear to purge state before a delayed restore repopulates it and calls `mark_hydrated(true)`. Serialize the entire per-wallet install transaction and coordinate it with unregister/clear, or use a registration generation that is checked while committing restore and hydration.

<sub>source: ['codex']</sub>

thepastaclaw: **🔴 Blocking: Do not restore old-key state after re-keying a subwallet**

For the changed-FVK path introduced here, `register_wallet` correctly identifies the existing `SubwalletId` as stale and purges its old-key in-memory notes and watermark. The bind tail then loads and restores that subwallet's durable snapshot unconditionally. Durable notes, activity, and sync state are keyed only by `(wallet_id, account_index)`; the viewing key is stored separately, and a seed-backed bind upserts the FVK without deleting the old subwallet rows. `restore_for_wallet` checks only the registered `SubwalletId`, not which FVK produced the snapshot, so it immediately restores old-key notes and the old high watermark under the new registration and marks it hydrated. This can expose notes the new spend key cannot spend and skip historical notes belonging to the new key. Reject an FVK change, or atomically invalidate the durable subwallet snapshot and exclude it from restoration before committing the new registration.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The latest commit fixes both previously reported defects: identical re-binds no longer take a destructive unregister path, and the coordinator lifecycle mutex preserves the accounts/persister registration invariant. Three blocking issues remain: the full bind lifecycle is not serialized, a changed FVK is rehydrated from durable state belonging to the old key, and an in-flight old-key sync can persist its changeset after the re-key.

### Review provenance

🔴 3 blocking

<details>

```

- [BLOCKING] packages/rs-platform-wallet/src/wallet/platform_wallet.rs:637-698: Serialize the complete shielded bind lifecycle
  The coordinator lifecycle mutex protects only `register_wallet`; the wallet-local key publication, registration comparison, restore, and hydration commit remain separate operations. Starting from hydrated registration A, bind A can write local keys A and observe `identical = true`; bind B can then write local keys B and register B; bind A can subsequently register A. The coordinator ends on A while `self.shielded_keys` remains B, so sync decrypts with A while address, balance, and spend paths use B. The same gap affects removal and clear: the FFI bind function retains a wallet `Arc` after releasing manager lookup protection, so a bind can register a wallet after `remove_wallet` has completed. In addition, `restore_for_wallet` snapshots registered IDs before taking the store lock, allowing unregister or clear to purge state before a delayed restore repopulates it and calls `mark_hydrated(true)`. Serialize the entire per-wallet install transaction and coordinate it with unregister/clear, or use a registration generation that is checked while committing restore and hydration.
- [BLOCKING] packages/rs-platform-wallet/src/wallet/platform_wallet.rs:662-698: Do not restore old-key state after re-keying a subwallet
  For the changed-FVK path introduced here, `register_wallet` correctly identifies the existing `SubwalletId` as stale and purges its old-key in-memory notes and watermark. The bind tail then loads and restores that subwallet's durable snapshot unconditionally. Durable notes, activity, and sync state are keyed only by `(wallet_id, account_index)`; the viewing key is stored separately, and a seed-backed bind upserts the FVK without deleting the old subwallet rows. `restore_for_wallet` checks only the registered `SubwalletId`, not which FVK produced the snapshot, so it immediately restores old-key notes and the old high watermark under the new registration and marks it hydrated. This can expose notes the new spend key cannot spend and skip historical notes belonging to the new key. Reject an FVK change, or atomically invalidate the durable subwallet snapshot and exclude it from restoration before committing the new registration.

- [BLOCKING] packages/rs-platform-wallet/src/wallet/shielded/coordinator.rs:967-983: Discard old-key sync changesets after a re-key
  A sync pass snapshots the old account/FVK map at lines 843-846 and builds an old-key changeset while holding the store write lock. A concurrent re-key publishes the new registration, waits for that store lock, and purges the old-key in-memory state at lines 418-429. After releasing the lock, however, the old pass still reaches this fan-out and sends its retained changeset through the current wallet persister. The host persistence callbacks receive only the wallet/account IDs and note or watermark data, with no FVK or registration generation that could reject the stale pass. Old-key notes and a high old-key watermark can therefore be persisted after the account was re-keyed, defeating any durable purge and causing a restart to restore the wrong state. Tag sync snapshots with a registration generation or FVK and discard affected subwallet changesets when the current registration no longer matches, or drain the complete pass through persistence before replacing the key.
```
</details>

packages/rs-platform-wallet/src/wallet/platform_wallet.rs
shumkov: Confirmed and fixed in `7959e154ea`.

The lifecycle mutex now covers the whole install transaction, not one registry write: `NetworkShieldedCoordinator::begin_install` returns a guard that holds it across registration comparison → register → restore → hydration commit, and every single-step public method wraps itself in a one-step transaction so external callers stay serialized. `clear()` takes the mutex before the store reset (lock order is `lifecycle` → `store` everywhere), and `remove_wallet` now marks the handle detached before unregistering, which the install refuses on — so a bind holding a stale `Arc<PlatformWallet>` past the removal can no longer re-register the wallet.

Went with the "serialize the transaction" remedy rather than a registration generation, since the generation is redundant once the transaction is atomic.

Regression test: with the transaction released early (the old shape), a second bind publishes its registration inside the first one's transaction — `a_second_bind_cannot_commit_inside_another_binds_transaction` fails `left: 2, right: 1` before the fix, passes after. `install_transaction_blocks_unregister_until_it_commits` and `bind_after_wallet_removal_is_refused` pin the other two halves.

shumkov: Confirmed and fixed in `7959e154ea`.

Took the "reject an FVK change" branch, since durable rows are keyed `(wallet_id, account_index)` with no key attribution — the schema can't express a coherent partial invalidation, and `ShieldedChangeSet::viewing_keys` already documents that an account's FVK never legitimately changes. `bind_shielded` now compares each derived FVK against the persisted row *before* upserting it, and the install compares against the current registration; both fail closed with an error pointing at Clear as the recovery.

As defense in depth for direct callers, `restore_for_wallet` also skips any subwallet whose snapshot carries a viewing key other than the registered one (snapshots predating the FVK rows restore unchanged).

Regression tests: `restore_skips_a_snapshot_produced_under_a_different_viewing_key` fails on the old restore (old-key notes and the 5000 watermark land) and passes now; `bind_rejects_a_viewing_key_change_for_an_already_persisted_account` fails on the old bind (it succeeded) and passes now.

One behavior note: the seed bind now hard-fails if `persister.load()` fails, since it can't tell a re-key from a first bind without it — the seedless path it sits behind in the FFI already did.

PR #4239, opened 2026-07-27
feat(dashmate)!: record whether a platform image was chosen instead of deriving it

shumkov:

dashmate stores **resolved** config values. Two of the ten images in the base config are not authored values — they are derived from the package version in `getBaseConfigFactory.js`:

```js
const dockerImageVersion = `${semver.major(version)}${prereleaseTag}`;   // "4" stable, "4-rc" on an rc
image: `dashpay/drive:${dockerImageVersion}`,
```

Because that value is copied into every operator's config at install time, it goes stale whenever the version crosses a major or a stable/prerelease boundary — and nothing prompts anyone, since the version bump alone changes what the correct value is. **53 of 73 config migrations exist to re-pin an image.** #4235 was the most recent instance: no migration moved operators onto the `4-rc` line at all.

The deeper problem is that a re-pinning migration must not overwrite an image the operator chose, but both are plain strings in one field, so the only available test is a pattern on the tag. That cannot be made to work — a locally built `dashpay/drive:4-local` is indistinguishable in shape from a published tag, and an earlier guard silently overwrote exactly those. **The information needed — did the operator choose this? — was never recorded.**

## What was done?

Record the intent instead of the value. `null` means *"use the image line this dashmate build ships"*; a string is the operator's own and is never touched.

**Reads are effective by default.** `get()` (exact *and* parent paths), `getOptions()`, the `options` property templates render through, and `toJSON()` all resolve. Stored intent comes back only from `getStored()`/`getStoredOptions()`, restricted to persistence, cloning, equality, base→network inheritance and reset — enforced by an architectural test, so a sixth raw read fails CI.

That ordering is the design. An earlier attempt resolved one hard-coded leaf path, so `config get <path>` answered correctly while `dashmate config --format=json` printed `null`. Now new readers are correct by default and exposing `null` takes deliberate effort.

**Nothing changes for operators:**

```console
$ dashmate config get platform.drive.abci.docker.image
dashpay/drive:4-rc                    # what will actually run

$ dashmate config get --raw platform.drive.abci.docker.image
null                                  # "you haven't pinned this"
```

`dashmate config` marks unpinned values `image: 'dashpay/drive:4-rc' (default)` — note outside the quotes so it can't be copied as part of the value. `--format=json` is deliberately left unannotated because other tools parse it. `dashmate config set <path> null` returns to tracking.

**After this, no release needs an image migration for these two services again.**

### Also fixed, because this depends on them

- **45 unguarded image re-pins.** Migrations across the table overwrote whatever was there, so a config older than the first guard lost a custom image long before any guard could inspect it. All platform re-pins now move a tag only when a release published it. Scoped to `drive.abci`/`dapi.api`/`dapi.rsDapi` deliberately: core, tenderdash and the gateway publish differently-shaped tags (`dashd:23`, `tenderdash:1.6.0`, `envoy:1.39.0-impr.1`) that the published-tag pattern doesn't match, so guarding them with it would silently stop those migrations firing. Each needs its own pattern — separate work.
- **The migration runner discarded its accumulator.** It passed the original object rather than the reduction, working only because every migration happens to mutate in place and return the same object.
- **Migration target.** The runner applied bundled migrations while recording the released version, so a config already stamped at that version short-circuited on `fromVersion === toVersion` and never received them. Target is now `max(released version, newest bundled migration key)`, recorded, and shared with fresh-config creation so a new config isn't immediately re-migrated.
- **Platform-only reset** read effective defaults and persisted them, converting a tracking config into an explicit pin.

## How Has This Been Tested?

168 unit tests passing, lint clean (0 errors).

New coverage, each verified load-bearing by mutation:

| test | fails when |
| --- | --- |
| effective reads on every accessor | any read path stops resolving |
| stored stays `null` through `toObject()` | a resolved value could reach disk |
| writes through a whole-config read throw | the snapshot stops being frozen |
| raw-access allowlist (architectural) | a new file reads stored state — verified by adding one to `config/envs.js` |
| nullable paths == derived-default keys | another image is made nullable with nothing to resolve to |
| stock tag → unset, operator image untouched | the guard widens or narrows |
| operator image survives oldest→newest migration | **without the guard the image is destroyed (`null`)** |

Behaviour verified end-to-end against the real DI container:

| | tracking | pinned |
| --- | --- | --- |
| `config get` / parent read / `--format=json` / compose env | `dashpay/drive:4-rc` | the operator's image |
| persisted to disk | `null` | the operator's image |

**Not tested:** no local-network e2e run. The changed `local-network` action is exercised by this PR's own CI, which is the right place for it.

## Breaking Changes

`platform.drive.abci.docker.image` and `platform.dapi.rsDapi.docker.image` accept `null`, and configs store `null` once migrated. Tooling that reads `config.json` directly must resolve it, or use `dashmate config get`, which does.

This repo's own `local-network` CI action was one such consumer — it matched image strings with `sed` and now sets them through `dashmate config set`.

## Notes for the release

**Installing an RC dashmate moves every configured network, mainnet included, onto the RC line** — and now does so on every command rather than once at a boundary, since the effective value follows the invoking binary. This was already true of a fresh install; document it in the RC notes.

**Downgrade is out of scope.** dashmate has no down migrations, and an older dashmate reading `image: null` will fail schema validation.

**One ambiguity is accepted, once.** An operator who deliberately pinned exactly `dashpay/drive:4-rc` is indistinguishable from an inherited default and will be converted to tracking by the one-time migration. Confining that guess to a single migration is the point of the change.

Spec, including where implementing it disproved the design: `docs/dashmate/DERIVED_IMAGE_DEFAULTS_SPEC.md`.

## Checklist:


thepastaclaw:
⛔ Blockers found — Sonnet deferred (commit b4f07147)

packages/dashmate/src/config/stockImages.js
```diff
@@ -28,3 +28,25 @@ export function stockImagePattern(repository, major) {
...
+ * @return {RegExp}
+ */
+export function stockImagePatternAnyVersion(repository) {
+  const escapedRepository = repository.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+
+  // Published tags only ever took two shapes: the 0.x line used a major.minor
+  // tag (`0.25`), every line since uses the major alone. Anything more
+  // permissive matches an operator's own exact pin such as `dashpay/drive:3.1.5`,
+  // which was never a published default.
+  return new RegExp(`^${escapedRepository}:(0\\.\\d+|\\d+)${stockPrereleaseSuffix}$`);
```
thepastaclaw: **🔴 Blocking: Match the complete historical stock-tag inventory**

Released Dashmate v1.0.0 and v1.0.1 builds derived Drive tags as `major.minor`, including `dashpay/drive:1.0` and `dashpay/drive:1.0-rc`; the derivation changed to major-only in v1.0.2. This matcher rejects those v1.0 tags, so guarded historical re-pins leave them unchanged and the final migration treats them as operator pins instead of converting them to tracking. A direct upgrade can therefore remain on the obsolete Drive 1.0 image. Conversely, the unrestricted integer branch accepts `dashpay/drive:5` or `dashpay/drive:999`, although no such release existed when this migration was introduced, defeating the PR's operator-pin preservation guarantee. Freeze the matcher to the actual historical shapes and add migration cases for `1.0`, `1.0-rc`, and an unpublished integer tag.

```suggestion
  // The 0.x and 1.0 lines used major.minor tags; releases from 1.0.2
  // through major 4 used the major alone. Keep this frozen to the inventory
  // that existed when this migration shipped so operator pins are preserved.
  return new RegExp(`^${escapedRepository}:(0\\.\\d+|1\\.0|[1-4])${stockPrereleaseSuffix}$`);
```

<sub>source: ['codex']</sub>

.github/actions/local-network/action.yaml
```diff
@@ -82,8 +82,12 @@ runs:
...
         docker tag ${{ inputs.image_org }}/dashmate-helper:$SHA_TAG dashpay/dashmate-helper:$VERSION
 
-        # Replace Drive and RS-DAPI images with new org and tag in dashmate config
-        sed -i -E "s/dashpay\/(drive|rs-dapi):[^\"]+/${{ inputs.image_org }}\/\1:${SHA_TAG}/g" ${{ env.HOME }}/.dashmate/config.json
+        # Point Drive and RS-DAPI at the images built for this commit. Set them
+        # through dashmate rather than editing config.json: these options are
+        # stored unset by default, meaning "use the line this dashmate build
+        # ships", so there is no image string in the file for a pattern to match.
+        yarn dashmate config set --config=local platform.drive.abci.docker.image "${{ inputs.image_org }}/drive:${SHA_TAG}"
+        yarn dashmate config set --config=local platform.dapi.rsDapi.docker.image "${{ inputs.image_org }}/rs-dapi:${SHA_TAG}"
```
thepastaclaw: **🔴 Blocking: Pin the images on the local group nodes that CI starts**

`setup local` creates `local_1`, `local_2`, `local_3`, and `local_seed`, assigns them to the `local` group, and leaves the separate `local` template config outside that group. The root `yarn start` command runs `dashmate group start`, so these two commands modify a config that is not started while every running node retains its tracking default and pulls the stock `dashpay` images. The previous `sed` replaced images across all configs, so this change causes local-network CI to stop testing the Drive and rs-dapi images built for the commit. Apply the overrides to every config in the local group.

```suggestion
        while IFS= read -r config_name; do
          yarn dashmate config set --config="$config_name" platform.drive.abci.docker.image "${{ inputs.image_org }}/drive:${SHA_TAG}"
          yarn dashmate config set --config="$config_name" platform.dapi.rsDapi.docker.image "${{ inputs.image_org }}/rs-dapi:${SHA_TAG}"
        done < <(jq -r '.configs | to_entries[] | select(.value.group == "local") | .key' "${HOME}/.dashmate/config.json")
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The stored-intent/effective-value implementation is coherent across persistence, inheritance, reset, CLI/JSON rendering, templates, and setup mutation paths, and the focused derived-default tests pass. Two blocking edge cases remain: the historical matcher misclassifies released and unpublished tags, and the local-network action pins only the template config rather than the group nodes that CI starts.

### Review provenance

🔴 2 blocking

<details>

```

In `packages/dashmate/src/config/stockImages.js`:
- [BLOCKING] packages/dashmate/src/config/stockImages.js:47-51: Match the complete historical stock-tag inventory
  Released Dashmate v1.0.0 and v1.0.1 builds derived Drive tags as `major.minor`, including `dashpay/drive:1.0` and `dashpay/drive:1.0-rc`; the derivation changed to major-only in v1.0.2. This matcher rejects those v1.0 tags, so guarded historical re-pins leave them unchanged and the final migration treats them as operator pins instead of converting them to tracking. A direct upgrade can therefore remain on the obsolete Drive 1.0 image. Conversely, the unrestricted integer branch accepts `dashpay/drive:5` or `dashpay/drive:999`, although no such release existed when this migration was introduced, defeating the PR's operator-pin preservation guarantee. Freeze the matcher to the actual historical shapes and add migration cases for `1.0`, `1.0-rc`, and an unpublished integer tag.

In `.github/actions/local-network/action.yaml`:
- [BLOCKING] .github/actions/local-network/action.yaml:89-90: Pin the images on the local group nodes that CI starts
  `setup local` creates `local_1`, `local_2`, `local_3`, and `local_seed`, assigns them to the `local` group, and leaves the separate `local` template config outside that group. The root `yarn start` command runs `dashmate group start`, so these two commands modify a config that is not started while every running node retains its tracking default and pulls the stock `dashpay` images. The previous `sed` replaced images across all configs, so this change causes local-network CI to stop testing the Drive and rs-dapi images built for the commit. Apply the overrides to every config in the local group.
```
</details>

packages/dashmate/src/config/stockImages.js
shumkov: Confirmed against git history and fixed in cb7671f554.

Verified rather than assumed — `v1.0.0` derived `${semver.major(version)}.${semver.minor(version)}${prereleaseTag}`, and `v1.0.2` changed it to the major alone. So `dashpay/drive:1.0` and `:1.0-rc` were genuinely published, my matcher read them as operator pins, and an operator on one would have been stranded on an obsolete Drive 1.0 image.

Took the suggested form. The bounded major matters for the reason you give — these are historical migrations, so a config carrying a later major's tag never reaches them, and accepting one could only ever match something the operator built themselves:

```js
return new RegExp(`^${escapedRepository}:(0\\.\\d+|1\\.0|[1-4])${stockPrereleaseSuffix}$`);
```

Also renamed it `historicalStockImagePattern` — `AnyVersion` actively misdescribed something now frozen to an inventory.

Test cases added as asked. `should unset a stock tag carried across majors` covers `1.0`, `1.0-rc`, `0.24` and `0.25`; the operator-image table covers `dashpay/drive:5` and `:999`. Verified load-bearing against the previous matcher:

```
stock drive image dashpay/drive:1.0 was not unset for base: expected 'dashpay/drive:1.0' to equal null
```

168 passing, lint clean.

.github/actions/local-network/action.yaml
shumkov: Correct, and this was the serious one — fixed in cb7671f554 with your suggested loop.

I confirmed the mechanism: `scripts/setup_local_network.sh` runs `dashmate setup local --node-count=3`, which creates `local_1..3` and `local_seed` in the `local` group, and the root `start` script is `dashmate group start`. The standalone `local` template is never started, so my two `--config=local` calls pinned a config that does not run, and every started node would have kept its tracking default and pulled the published `dashpay` images.

Worth stating plainly: that failure is **silent**. CI would have gone green while testing the released images instead of the ones built for the commit. I had argued against dispatching an e2e run on the grounds that a mistake here would fail loudly — it would not have, and this is exactly what that run would have caught.

The loop now pins every config in the group, which is what the `sed` it replaces effectively did:

```bash
while IFS= read -r config_name; do
  yarn dashmate config set --config="$config_name" platform.drive.abci.docker.image "…/drive:${SHA_TAG}"
  yarn dashmate config set --config="$config_name" platform.dapi.rsDapi.docker.image "…/rs-dapi:${SHA_TAG}"
done < <(jq -r '.configs | to_entries[] | select(.value.group == "local") | .key' "${HOME}/.dashmate/config.json")
```

`jq` is already used in this step for the package version, and `dashmate config set --config=…` is the form used earlier in this same action, so both are proven here.

Still unexecuted: E2E is gated on `version-changed || schedule || workflow_dispatch`, so this PR does not run it. It will first execute on a nightly or a release PR.

PR #4244, opened 2026-07-28
fix(wallet): reconcile Platform balance after unshield

llbartekll:
## Summary

- preserve the confirmed state transition proof and response height for shielded unshield
- reconcile the expected Platform address through the existing height-aware wallet/provider persistence path
- keep public Rust, C FFI, and Swift APIs unchanged
- treat incomplete or unexpected confirmed proofs as a successful operation with BLAST fallback

## Safety

The proof is accepted for immediate reconciliation only when it contains exactly one expected address with nonce and balance. Any other shape is logged and falls back to sync without retrying the spend or releasing notes.

## Validation

- cargo test -p platform-wallet --features shielded
- cargo clippy -p platform-wallet --features shielded -- -D warnings
- cargo check -p platform-wallet-ffi
- cargo fmt --all -- --check
- packages/swift-sdk/build_ios.sh for iOS device and simulator
- SwiftExampleApp build

### DIFF
--- a/packages/rs-platform-wallet/src/wallet/platform_wallet.rs
+++ b/packages/rs-platform-wallet/src/wallet/platform_wallet.rs
@@ -1163,7 +1163,7 @@ impl PlatformWallet {
             .map_err(|e| {
                 PlatformWalletError::ShieldedBuildError(format!("invalid platform address: {e}"))
             })?;
-        super::shielded::operations::unshield(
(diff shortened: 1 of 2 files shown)


llbartekll:
Closing after verification on clean v4.2-dev: Shielded → Platform balance updates correctly without this client-side reconciliation change, indicating the issue has been resolved on the backend.

PR #4185, opened 2026-07-21
feat(kotlin-sdk): split build/broadcast with reservation release for BIP70-style deferred submission

bfoss765:
Splits transaction build from broadcast for BIP70-style deferred submission: a signed-payment registry with reservation release, a deferred-payment token bounded to the reservation's lifetime, native code 27 for stale reservation tokens, token sweeping only when the final wallet write wins, and routing of deferred builds through the atomic finalize-and-register path — across `rs-platform-wallet`, `platform-wallet-ffi`, `rs-unified-sdk-jni`, and the Kotlin SDK surface.

Re-opens #4090 which was auto-closed when the #3999 base branch was deleted; rebased onto `v4.1-dev`. All seven original commits replayed cleanly — no hunks needed to be dropped as already-absorbed.

Verified: `cargo test -p platform-wallet -p platform-wallet-ffi -p rs-unified-sdk-jni` all pass (504 / 229 / 10); `:sdk:assembleRelease` + sdk unit tests pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added support for deferred signed payments, enabling transactions to be built and signed before later broadcast or release.
  * Added payment details including transaction ID, raw transaction data, fees, and reservation tokens.
  * Added idempotent reservation release for abandoned payments.
  * Added typed reservation, wallet-mismatch, and transaction-rejection errors across supported SDKs.
* **Bug Fixes**
  * Improved cleanup when wallets close while deferred payments remain active.
  * Prevented reservations from being used with the wrong wallet, after expiration, or more than once.
  * Improved handling of removed wallets and definitive broadcast failures.

---

### Why the token registry instead of the V2 handle surface

[...]

PR #4247, opened 2026-07-28
feat(kotlin-sdk): single-account build_signed_payment send API (funding_path)

bfoss765:
## What
Adds the SDK's L1 **send primitive** `build_signed_payment` and plumbs it through FFI → JNI → Kotlin.

- `feat(platform-wallet): union-funding build_signed_payment core primitive` — builds + signs an L1 payment transaction, union-funding across all funding accounts.
- `feat(sdk): plumb build_signed_payment through FFI, JNI, and Kotlin` — exposes it to the Android/Kotlin SDK.

## Why
This is the **send half of the wallet SDK cutover**: the Dash Wallet Android app calls it (`SdkL1SendService` / `CoreSendAllNative`) for post-cutover L1 sends. Once dashj's L1 engine is held after cutover, this primitive is what actually builds/signs/sends transactions — so it is load-bearing for the cutover.

## Notes
- Split out of the QA integration line onto a clean `v4.2-dev` base (2 self-contained commits); mirrors the other v4.1 Kotlin-SDK feature PRs.
- **No rust-dashcore dependency change** required.
- `cargo check -p platform-wallet -p platform-wallet-ffi` is clean.

## Summary by CodeRabbit

* **New Features**
  * Added APIs to build and sign standard L1 payments without broadcasting or recording a debit.
  * Supports multiple recipients, configurable fees, automatic change calculation, and account-specific funding.
  * Added deferred payment workflows for finalization, broadcasting, and reservation release.
  * Account balance data now includes derivation paths when available.

* **Bug Fixes**
  * Excludes watch-only funds and reports available and required amounts when funds are insufficient.
  * Improved wallet lifecycle handling to prevent stale or mismatched payment operations.
  * Added clear errors for invalid, expired, consumed, or rejected payment requests.

### DIFF
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
@@ -306,11 +306,15 @@ class ManagedPlatformWallet internal constructor(
     /**
      * Broadcast the deferred payment behind [token] (from [buildSignedPayment])
      * and return its broadcast txid — the "merchant server acked" arm. Consumes
-     * the token: a second [broadcastSigned] with the same token, or one for a
-     * re-created wallet, throws
-     * [org.dashfoundation.dashsdk.errors.DashSdkError.PlatformWallet.StaleReservationToken]
-     * rather than double-broadcasting. Operates on the token directly (the
-     * inputs are already reserved).
+     * the token. Rather than double-broadcasting, an unusable token throws one
+     * of three sibling errors: already consumed / unknown
+     * ([org.dashfoundation.dashsdk.errors.DashSdkError.PlatformWallet.ReservationTokenConsumed],
+     * e.g. a second [broadcastSigned] with the same token), a different wallet
+     * generation
+     * ([org.dashfoundation.dashsdk.errors.DashSdkError.PlatformWallet.ReservationWalletMismatch],
+     * e.g. a re-created wallet), or aged out
+     * ([org.dashfoundation.dashsdk.errors.DashSdkError.PlatformWallet.StaleReservationToken]).
+     * Operates on the token directly (the inputs are already reserved).
      */
     suspend fun broadcastSigned(token: Long): String = withContext(Dispatchers.IO) {
         mapNativeErrors {
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
@@ -180,6 +180,19 @@ class ManagedPlatformWallet internal constructor(
      * flows that must sign now, POST the raw bytes to a merchant server, and
      * broadcast only on the server's ack.
      *
+     * **Owns the reservation token.** The blocking native registration mints the
+     * token before this object exists, so if the object were then discarded —
+     * the caller drops it, or a coroutine cancellation is observed after
+     * [buildSignedPayment]'s native call returned — the token (and its funding
+     * reservation) would be orphaned until key-wallet's TTL. This type is
+     * therefore [AutoCloseable] with a [NativeCleaner] GC backstop: [close], or
+     * GC if you never call it, releases the token exactly once. Release is
+     * idempotent native-side and tokens are process-unique (never reused), so
+     * releasing a token already consumed by [broadcastSigned] /
+     * [releaseReservation] — or releasing twice — is a harmless no-op. A caller
+     * that broadcasts or releases can still `use`/close this object; a caller
+     * that abandons it is covered by GC.
+     *
      * @property txidHex the transaction id (lowercase hex) the broadcast will
      *   return — computed from the signed bytes Rust-side so it matches exactly.
      * @property rawTxBytes the consensus-serialized signed transaction, to hand
@@ -187,14 +200,29 @@ class ManagedPlatformWallet internal constructor(
      * @property feeDuffs the fee the build charged, in duffs.
      * @property reservationToken the opaque token for [broadcastSigned] /
      *   [releaseReservation]. Valid only for this wallet instance and only until
-     *   consumed by one of those calls.
+     *   consumed by one of those calls (or released by [close] / GC).
      */
     class SignedCoreTransaction internal constructor(
         val txidHex: String,
         val rawTxBytes: ByteArray,
         val feeDuffs: Long,
         val reservationToken: Long,
-    ) {
+    ) : AutoCloseable {
+
+        // GC backstop: releases the token if it was neither broadcast nor
+        // released. The action must not reference this object (it would never
+        // become phantom-reachable), so it captures the token by value.
+        private val cleanable = NativeCleaner.register(this, TokenRelease(reservationToken))
+
+        /**
+         * Release the funding reservation if this payment was neither broadcast
+         * nor released, and drop the token. Idempotent — safe to call after a
+         * [broadcastSigned] / [releaseReservation] (native no-op) and safe to
+         * call twice. The [NativeCleaner] backstop runs the same release on GC
+         * if you never call [close].
+         */
+        override fun close() = cleanable.clean()
+
         override fun equals(other: Any?): Boolean =
             other is SignedCoreTransaction &&
                 txidHex == other.txidHex &&
@@ -213,6 +241,13 @@ class ManagedPlatformWallet internal constructor(
         override fun toString(): String =
             "SignedCoreTransaction(txidHex=$txidHex, feeDuffs=$feeDuffs, " +
                 "reservationToken=$reservationToken, rawTxBytes=${rawTxBytes.size} bytes)"
+
+        /** Releases the reservation token exactly once, on [close] or GC. */
+        private class TokenRelease(private val token: Long) : Runnable {
+            override fun run() {
+                WalletManagerNative.coreWalletReleaseSignedPayment(token)
+            }
+        }
 
         internal companion object {
             /**
@@ -256,6 +291,14 @@ class ManagedPlatformWallet internal constructor(
      * buildSigned split had), so once this returns the reservation holds the
      * inputs and [broadcastSigned] / [releaseReservation] operate on the token
      * later.
+     *
+     * The returned [SignedCoreTransaction] OWNS the token: it is [AutoCloseable]
+     * with a GC/[NativeCleaner] backstop, so a token that is neither broadcast
+     * nor released is never orphaned — even if the caller drops the object or a
+     * cancellation discards it after this call's blocking native registration
+     * already minted the token. The backstop releases the reservation on GC (or
+     * on an explicit [SignedCoreTransaction.close]); consuming the token via
+     * [broadcastSigned] / [releaseReservation] makes that release a native no-op.
      *
      * Process-death note: the reservation is in-memory. An app crash between
      * this call and [broadcastSigned] drops the reservation on restart (the
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
@@ -359,10 +359,40 @@ class ManagedPlatformWallet internal constructor(
      * e.g. a re-created wallet), or aged out
      * ([org.dashfoundation.dashsdk.errors.DashSdkError.PlatformWallet.StaleReservationToken]).
      * Operates on the token directly (the inputs are already reserved).
+     *
+     * Callers holding a [SignedCoreTransaction] should prefer the object
+     * overload: with the bare token, the source object must stay strongly
+     * reachable until this call returns, or its GC backstop can release the
+     * reservation mid-broadcast.
      */
     suspend fun broadcastSigned(token: Long): String = withContext(Dispatchers.IO) {
         mapNativeErrors {
             coreWallet().use { core -> core.broadcastSignedPayment(token) }
+        }
+    }
+
+    /**
+     * Broadcast [payment] and return its txid — the object-owning form of
+     * [broadcastSigned]. Prefer this over passing the bare
+     * [SignedCoreTransaction.reservationToken]: the token's lifetime is coupled
+     * to the object's GC-reachability (the [NativeCleaner] backstop releases the
+     * reservation when the object is collected), so a caller that extracts the
+     * `Long` and drops the object races GC and can find the reservation gone.
+     * This overload keeps the object reachable for the whole native call and
+     * disarms the backstop once the token is consumed.
+     */
+    suspend fun broadcastSigned(payment: SignedCoreTransaction): String {
+        try {
+            val txid = broadcastSigned(payment.reservationToken)
+            // Token consumed: close() disarms the GC backstop (the underlying
+            // native release is an idempotent no-op on a consumed token).
+            payment.close()
+            return txid
+        } finally {
+            // The object must stay reachable across the suspend/native call —
+            // without this, GC could run the backstop mid-broadcast and release
+            // the reservation out from under it.
+            java.lang.ref.Reference.reachabilityFence(payment)
         }
     }
 
@@ -378,6 +408,20 @@ class ManagedPlatformWallet internal constructor(
             mapNativeErrors {
                 WalletManagerNative.coreWalletReleaseSignedPayment(token)
             }
+        }
+    }
+
+    /**
+     * Release [payment]'s funding reservation — the object-owning form of
+     * [releaseReservation]; see [broadcastSigned] for why it is preferred over
+     * the bare-token form.
+     */
+    suspend fun releaseReservation(payment: SignedCoreTransaction) {
+        try {
+            releaseReservation(payment.reservationToken)
+            payment.close()
+        } finally {
+            java.lang.ref.Reference.reachabilityFence(payment)
         }
     }
 
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
@@ -522,6 +522,55 @@ class ManagedPlatformWallet internal constructor(
     }
 
     /**
+     * Release the UTXO reservation a [buildSignedPayment] call took, for a
+     * build this caller has decided **not** to broadcast.
+     *
+     * [buildSignedPayment] deliberately leaves its selected inputs reserved on
+     * success, because the expected next step is a broadcast. If the build is
+     * abandoned instead — the user backed out of the confirmation screen, an
+     * upstream check failed, the send screen is being torn down — this must be
+     * called or those coins stay unselectable.
+     *
+     * **Why it matters (dashpay/platform#4247 review).** Without it the inputs
+     * are stranded until key-wallet's TTL backstop reclaims them 24 blocks
+     * (~1 hour) later. Worse, that backstop does not merely run late but does
+     * not run at all before the first sync completes: the sweep early-returns
+     * while the wallet has no processed height, so a reservation taken at
+     * height 0 is never reclaimed for the life of the process. A single
+     * abandoned build on a freshly restored wallet could otherwise strand the
+     * entire balance. This call consults no height, so it is the one release
+     * path that works pre-sync.
+     *
+     * **Releases only this build's own inputs.** The transaction is the
+     * ownership signal: a reserved outpoint is skipped by every other build's
+     * coin selection, so no concurrent build can hold a reservation on any
+     * input of [txBytes].
+     *
+     * **Idempotent, and safe after a broadcast.** Calling it twice, or on a
+     * transaction that was in fact broadcast, is a silent no-op rather than an
+     * error, and it cannot resurrect a spent coin — coin selection reads the
+     * UTXO set, from which sync removes the spend independently of any
+     * reservation. That makes it safe in an unconditional `finally` without
+     * tracking whether the broadcast succeeded.
+     *
+     * @param txBytes [SignedCorePayment.txBytes] from the build being
+     *   abandoned.
+     * @param fundingPath the **same** [fundingPath] the build was given, so the
+     *   release lands on the account holding the reservation; `null` means the
+     *   unmixed BIP44 account, exactly as it does for the build. A path naming
+     *   a different account is harmless but frees nothing.
+     */
+    suspend fun releasePaymentReservation(
+        txBytes: ByteArray,
+        fundingPath: String? = null,
+    ): Unit = gate.op {
+        require(txBytes.isNotEmpty()) { "txBytes must not be empty" }
+        mapNativeErrors {
+            coreWallet().use { core -> core.releasePaymentReservation(txBytes, fundingPath) }
+        }
+    }
+
+    /**
      * The wallet's Platform-payment addresses that currently hold credits,
      * each as a [FundingInput] whose `credits` is the full cached balance —
      * the funding-input candidates for an identity top-up. Port of the
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
@@ -216,10 +216,9 @@ class ManagedPlatformWallet internal constructor(
 
         internal companion object {
             /**
-             * Decode the big-endian native BLOB the deferred build/register FFI
-             * returns: `u64 token, u64 feeDuffs, u32 txidLen, txid utf8,
-             * u32 txBytesLen, txBytes`. Shared by the atomic
-             * finalize-and-register path and the deprecated register path.
+             * Decode the big-endian native BLOB the atomic
+             * finalize-and-register FFI returns: `u64 token, u64 feeDuffs,
+             * u32 txidLen, txid utf8, u32 txBytesLen, txBytes`.
              */
             internal fun fromRegisterBlob(blob: ByteArray): SignedCoreTransaction {
                 val buffer = java.nio.ByteBuffer.wrap(blob) // big-endian by default
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
@@ -257,14 +257,28 @@ class ManagedPlatformWallet internal constructor(
              */
             internal fun fromRegisterBlob(blob: ByteArray): SignedCoreTransaction {
                 val buffer = java.nio.ByteBuffer.wrap(blob) // big-endian by default
+                // The blob is produced by our own JNI packer, not an attacker, so
+                // a malformed layout is a bug rather than a reachable hostile
+                // input. Validate the two length prefixes anyway: a negative
+                // length (a u32 high bit read as a signed Int) would make
+                // `ByteArray(len)` throw NegativeArraySizeException, and a length
+                // past the buffer would throw BufferUnderflowException — both
+                // opaque. `require` turns either into a precise, greppable message
+                // naming the field and the offending value.
+                fun readLengthPrefixedBytes(field: String): ByteArray {
+                    val len = buffer.int
+                    require(len in 0..buffer.remaining()) {
+                        "register blob: $field length $len out of range " +
+                            "0..${buffer.remaining()} (blob is ${blob.size} bytes)"
+                    }
+                    val bytes = ByteArray(len)
+                    buffer.get(bytes)
+                    return bytes
+                }
                 val token = buffer.long
                 val feeDuffs = buffer.long
-                val txidLen = buffer.int
-                val txidBytes = ByteArray(txidLen)
-                buffer.get(txidBytes)
-                val txBytesLen = buffer.int
-                val rawTxBytes = ByteArray(txBytesLen)
-                buffer.get(rawTxBytes)
+                val txidBytes = readLengthPrefixedBytes("txid")
+                val rawTxBytes = readLengthPrefixedBytes("txBytes")
                 return SignedCoreTransaction(
                     txidHex = String(txidBytes, Charsets.UTF_8),
                     rawTxBytes = rawTxBytes,
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
@@ -294,11 +294,14 @@ class ManagedPlatformWallet internal constructor(
      *
      * The returned [SignedCoreTransaction] OWNS the token: it is [AutoCloseable]
      * with a GC/[NativeCleaner] backstop, so a token that is neither broadcast
-     * nor released is never orphaned — even if the caller drops the object or a
-     * cancellation discards it after this call's blocking native registration
-     * already minted the token. The backstop releases the reservation on GC (or
-     * on an explicit [SignedCoreTransaction.close]); consuming the token via
-     * [broadcastSigned] / [releaseReservation] makes that release a native no-op.
+     * nor released is never orphaned. If a cancellation discards the result
+     * *after* the blocking native registration already minted the token, this
+     * call closes it deterministically on the way out (the gate's
+     * cancellation-cleanup handoff) rather than leaving the reservation to the
+     * GC backstop or the reservation TTL. Otherwise the backstop releases on GC,
+     * or the caller releases via an explicit [SignedCoreTransaction.close];
+     * consuming the token via [broadcastSigned] / [releaseReservation] makes
+     * that release a native no-op.
      *
      * Process-death note: the reservation is in-memory. An app crash between
      * this call and [broadcastSigned] drops the reservation on restart (the
@@ -314,7 +317,17 @@ class ManagedPlatformWallet internal constructor(
         coreSignerHandle: Long,
         accountType: AccountType = AccountType.BIP44,
         accountIndex: Int = 0,
-    ): SignedCoreTransaction = gate.op {
+    ): SignedCoreTransaction = gate.opWithCleanupOnCancellation(
+        // Native finalization mints the token and transfers reservation ownership
+        // to it before the blocking JNI call returns, so the token already exists
+        // by the time `withContext` dispatches back to the caller. That handoff is
+        // a prompt-cancellation point: if the caller was cancelled while JNI ran,
+        // the completed SignedCoreTransaction is discarded before anyone can hold
+        // it, leaving only the GC/NativeCleaner backstop — the reservation would
+        // then sit until an unpredictable GC cycle or the reservation TTL.
+        // Closing the discarded result releases the token deterministically.
+        cleanup = { payment: SignedCoreTransaction -> payment.close() },
+    ) {
         require(accountIndex >= 0) { "accountIndex must be non-negative, got $accountIndex" }
         require(recipients.isNotEmpty()) { "recipients must not be empty" }
         require(recipients.all { it.second > 0 }) {
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
@@ -213,6 +213,32 @@ class ManagedPlatformWallet internal constructor(
         override fun toString(): String =
             "SignedCoreTransaction(txidHex=$txidHex, feeDuffs=$feeDuffs, " +
                 "reservationToken=$reservationToken, rawTxBytes=${rawTxBytes.size} bytes)"
+
+        internal companion object {
+            /**
+             * Decode the big-endian native BLOB the deferred build/register FFI
+             * returns: `u64 token, u64 feeDuffs, u32 txidLen, txid utf8,
+             * u32 txBytesLen, txBytes`. Shared by the atomic
+             * finalize-and-register path and the deprecated register path.
+             */
+            internal fun fromRegisterBlob(blob: ByteArray): SignedCoreTransaction {
+                val buffer = java.nio.ByteBuffer.wrap(blob) // big-endian by default
+                val token = buffer.long
+                val feeDuffs = buffer.long
+                val txidLen = buffer.int
+                val txidBytes = ByteArray(txidLen)
+                buffer.get(txidBytes)
+                val txBytesLen = buffer.int
+                val rawTxBytes = ByteArray(txBytesLen)
+                buffer.get(rawTxBytes)
+                return SignedCoreTransaction(
+                    txidHex = String(txidBytes, Charsets.UTF_8),
+                    rawTxBytes = rawTxBytes,
+                    feeDuffs = feeDuffs,
+                    reservationToken = token,
+                )
+            }
+        }
     }
 
     /**
@@ -256,25 +282,24 @@ class ManagedPlatformWallet internal constructor(
             AccountType.BIP32 -> CoreTransactionBuilder.AccountType.BIP32
         }
         mapNativeErrors {
-            coreWallet().use { core ->
-                val builder = CoreTransactionBuilder(network)
-                // `buildSigned` consumes the builder; `use` still safely
-                // destroys it on the pre-build failure paths.
-                val signedTx = builder.use {
-                    for ((address, amount) in recipients) {
-                        it.addOutput(address, amount)
-                    }
-                    it.setFunding(this@ManagedPlatformWallet, builderAccountType, accountIndex)
-                    it.buildSigned(
-                        this@ManagedPlatformWallet,
-                        builderAccountType,
-                        accountIndex,
-                        coreSignerHandle,
-                    )
+            // One atomic native operation: select + reserve + sign + register.
+            // `finalizeSignedPayment` consumes the builder on every path, so
+            // `use` only needs to destroy it on the pre-finalize failure paths
+            // (adding outputs). Selection and reservation commit as a single unit
+            // under the wallet-manager lock, so a concurrent deferred build — or a
+            // deferred build racing an immediate send — can no longer double-
+            // select the same input, restoring the atomicity the removed Kotlin
+            // per-wallet send mutex used to provide.
+            CoreTransactionBuilder(network).use { builder ->
+                for ((address, amount) in recipients) {
+                    builder.addOutput(address, amount)
                 }
-                // Register the signed tx (holding its reservation) before the
-                // native transaction is freed; `use` frees it afterward.
-                signedTx.use { tx -> core.registerSignedPayment(tx) }
+                builder.finalizeSignedPayment(
+                    this@ManagedPlatformWallet,
+                    builderAccountType,
+                    accountIndex,
+                    coreSignerHandle,
+                )
             }
         }
     }
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
@@ -383,11 +383,9 @@ class ManagedPlatformWallet internal constructor(
     }
 
     /**
-     * Build and sign a Core L1 payment to [recipients], funding it from the
-     * UNION of this wallet's signable funds accounts (BIP44 + BIP32 + CoinJoin
-     * + DashPay receiving; watch-only DashPay external accounts excluded), and
-     * return the signed raw transaction bytes plus the fee and change —
-     * **WITHOUT broadcasting**.
+     * Build and sign a Core L1 payment to [recipients], funding it from a
+     * **single** funds account, and return the signed raw transaction bytes
+     * plus the fee and change — **WITHOUT broadcasting**.
      *
      * This is the transition-era "give me signed bytes" primitive: the Android
      * wallet hands [SignedCorePayment.txBytes] to dashj for commit + broadcast
@@ -395,9 +393,20 @@ class ManagedPlatformWallet internal constructor(
      * confidence listeners), while the SDK owns coin selection and signing. It
      * does not broadcast and does not persist a debit; the selected inputs are
      * only reserved in memory (released when the spend is later observed by sync
-     * or by the reservation-TTL backstop). Coin selection auto-spans every
-     * signable account, so — unlike [sendToAddresses] — the caller does not
-     * pick a funding account.
+     * or by the reservation-TTL backstop).
+     *
+     * **Funding-domain isolation (dashpay/platform#4184).** Coin selection never
+     * spans accounts. [fundingPath] names the one funds account to draw from;
+     * `null` (the default) draws from the unmixed BIP44 account. Passing an
+     * explicit account-level path — e.g. the DIP-9 CoinJoin account path — spends
+     * previously-mixed coins deliberately, and only those. Unioning ordinary,
+     * CoinJoin, and DashPay-receiving coins into one transaction would
+     * irreversibly link those privacy domains on chain, so it is never done
+     * implicitly: if the named account cannot cover the payment this throws
+     * [org.dashfoundation.dashsdk.errors.DashSdkError.PlatformWallet.CoreInsufficientFunds]
+     * rather than reaching into another account — whose `available` figure is
+     * that ONE account's balance, so the actionable response is to pick a
+     * different [fundingPath], not to retry the same one.
      *
      * Runs through the manager's [TeardownGate] like every other native op.
      * A concurrent build cannot select the same UTXO because the underlying
@@ -411,11 +420,14 @@ class ManagedPlatformWallet internal constructor(
      *   (`PlatformWalletManager.mnemonicResolverHandle`); no private key crosses
      *   the boundary.
      * @param feePerKb fee rate in duffs/kB, or 0 for the SDK default.
+     * @param fundingPath optional UTF-8 BIP32 derivation-path string naming the
+     *   single funds account to fund from; `null` = the unmixed BIP44 account.
      */
     suspend fun buildSignedPayment(
         recipients: List<Pair<String, Long>>,
         coreSignerHandle: Long,
         feePerKb: Long = 0,
+        fundingPath: String? = null,
     ): SignedCorePayment = gate.op {
         require(recipients.isNotEmpty()) { "recipients must not be empty" }
         require(recipients.all { it.second > 0 }) { "every recipient amount must be positive" }
@@ -425,7 +437,7 @@ class ManagedPlatformWallet internal constructor(
         mapNativeErrors {
             coreWallet().use { core ->
                 decodeSignedPayment(
-                    core.buildSignedPayment(outputsBlob, feePerKb, coreSignerHandle),
+                    core.buildSignedPayment(outputsBlob, feePerKb, coreSignerHandle, fundingPath),
                 )
             }
         }
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
@@ -361,6 +361,77 @@ class ManagedPlatformWallet internal constructor(
     }
 
     /**
+     * A built-and-signed Core L1 payment that was NOT broadcast — the output of
+     * [buildSignedPayment]. [txBytes] is the consensus-serialized signed
+     * transaction the caller commits/broadcasts itself (dashj during the
+     * dashj→SDK transition; the SDK's own broadcast afterwards). [fee] and
+     * [change] are duffs.
+     */
+    data class SignedCorePayment(
+        val txBytes: ByteArray,
+        val fee: Long,
+        val change: Long,
+    ) {
+        override fun equals(other: Any?): Boolean =
+            other is SignedCorePayment &&
+                txBytes.contentEquals(other.txBytes) &&
+                fee == other.fee &&
+                change == other.change
+
+        override fun hashCode(): Int =
+            (31 * txBytes.contentHashCode() + fee.hashCode()) * 31 + change.hashCode()
+    }
+
+    /**
+     * Build and sign a Core L1 payment to [recipients], funding it from the
+     * UNION of this wallet's signable funds accounts (BIP44 + BIP32 + CoinJoin
+     * + DashPay receiving; watch-only DashPay external accounts excluded), and
+     * return the signed raw transaction bytes plus the fee and change —
+     * **WITHOUT broadcasting**.
+     *
+     * This is the transition-era "give me signed bytes" primitive: the Android
+     * wallet hands [SignedCorePayment.txBytes] to dashj for commit + broadcast
+     * (keeping dashj's `maybeCommitTx` bookkeeping — CrowdNode, memos,
+     * confidence listeners), while the SDK owns coin selection and signing. It
+     * does not broadcast and does not persist a debit; the selected inputs are
+     * only reserved in memory (released when the spend is later observed by sync
+     * or by the reservation-TTL backstop). Coin selection auto-spans every
+     * signable account, so — unlike [sendToAddresses] — the caller does not
+     * pick a funding account.
+     *
+     * Runs through the manager's [TeardownGate] like every other native op.
+     * A concurrent build cannot select the same UTXO because the underlying
+     * `build_signed_payment` holds the wallet-manager write lock across coin
+     * selection and signing (the same native serialization [sendToAddresses]
+     * relies on).
+     *
+     * @param recipients `(address, amountDuffs)` pairs; must be non-empty and
+     *   every amount positive.
+     * @param coreSignerHandle the manager's `MnemonicResolverHandle`
+     *   (`PlatformWalletManager.mnemonicResolverHandle`); no private key crosses
+     *   the boundary.
+     * @param feePerKb fee rate in duffs/kB, or 0 for the SDK default.
+     */
+    suspend fun buildSignedPayment(
+        recipients: List<Pair<String, Long>>,
+        coreSignerHandle: Long,
+        feePerKb: Long = 0,
+    ): SignedCorePayment = gate.op {
+        require(recipients.isNotEmpty()) { "recipients must not be empty" }
+        require(recipients.all { it.second > 0 }) { "every recipient amount must be positive" }
+        require(feePerKb >= 0) { "feePerKb must be non-negative, got $feePerKb" }
+
+        val outputsBlob = encodePaymentOutputs(recipients)
+        mapNativeErrors {
+            coreWallet().use { core ->
+                decodeSignedPayment(
+                    core.buildSignedPayment(outputsBlob, feePerKb, coreSignerHandle),
+                )
+            }
+        }
+    }
+
+    /**
      * Broadcast the deferred payment behind [token] (from [buildSignedPayment])
      * and return its broadcast txid — the "merchant server acked" arm. Consumes
      * the token. Rather than double-broadcasting, an unusable token throws one
@@ -888,6 +959,41 @@ class ManagedPlatformWallet internal constructor(
             dos.writeLong(o.credits)
         }
         return out.toByteArray()
+    }
+
+    /**
+     * Encode [recipients] to the payment-outputs blob
+     * `core_wallet_build_signed_payment` reads: `u32 count` then per row
+     * `u32 addrLen, addr utf8 bytes, u64 amount` (all big-endian, matching
+     * `DataOutputStream`'s wire order and the Rust `from_be_bytes` decoder).
+     */
+    private fun encodePaymentOutputs(recipients: List<Pair<String, Long>>): ByteArray {
+        val out = java.io.ByteArrayOutputStream()
+        val dos = java.io.DataOutputStream(out)
+        dos.writeInt(recipients.size)
+        for ((address, amount) in recipients) {
+            val addrBytes = address.toByteArray(Charsets.UTF_8)
+            dos.writeInt(addrBytes.size)
+            dos.write(addrBytes)
+            dos.writeLong(amount)
+        }
+        return out.toByteArray()
+    }
+
+    /**
+     * Decode the packed [SignedCorePayment] the native build returns:
+     * `u64 fee, u64 change,` then the signed transaction bytes (big-endian).
+     */
+    private fun decodeSignedPayment(packed: ByteArray): SignedCorePayment {
+        require(packed.size >= 16) {
+            "signed-payment result too short (${packed.size} bytes, need >= 16)"
+        }
+        val buffer = java.nio.ByteBuffer.wrap(packed) // big-endian by default
+        val fee = buffer.long
+        val change = buffer.long
+        val txBytes = ByteArray(buffer.remaining())
+        buffer.get(txBytes)
+        return SignedCorePayment(txBytes = txBytes, fee = fee, change = change)
     }
 
     /**
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
@@ -371,15 +371,33 @@ class ManagedPlatformWallet internal constructor(
         val txBytes: ByteArray,
         val fee: Long,
         val change: Long,
+        /**
+         * Opaque handle for the key-wallet reservation token this build stamped
+         * on its inputs (0 = it reserved nothing).
+         *
+         * Pass it to [releasePaymentReservation] together with [txBytes] when
+         * abandoning the build. It is the ownership proof that stops the release
+         * from freeing a *different* build's reservation on the same outpoint —
+         * possible whenever key-wallet's TTL swept this build's reservation and
+         * another build re-took it (dashpay/platform#4247 review). The token
+         * itself never crosses the native boundary: it has no public constructor
+         * precisely so it cannot be forged.
+         *
+         * Not a resource: there is nothing to close, and holding a stale handle
+         * is harmless — it simply stops matching.
+         */
+        val reservationHandle: Long,
     ) {
         override fun equals(other: Any?): Boolean =
             other is SignedCorePayment &&
                 txBytes.contentEquals(other.txBytes) &&
                 fee == other.fee &&
-                change == other.change
+                change == other.change &&
+                reservationHandle == other.reservationHandle
 
         override fun hashCode(): Int =
-            (31 * txBytes.contentHashCode() + fee.hashCode()) * 31 + change.hashCode()
+            ((31 * txBytes.contentHashCode() + fee.hashCode()) * 31 + change.hashCode()) * 31 +
+                reservationHandle.hashCode()
     }
 
     /**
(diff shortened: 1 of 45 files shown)


thepastaclaw:
🔍 Review in progress — actively reviewing now (commit 5752e211)
_ETA: complete ~06:48 UTC (median 23m across 30 recent reviews)_
_Running 4m · Last checked: 2026-08-05 06:30 UTC_

packages/rs-platform-wallet/src/wallet/core/send.rs
```diff
@@ -0,0 +1,619 @@
...
+                })?;
+            let mut builder = TransactionBuilder::new()
+                .set_fee_rate(fee_rate)
+                .set_current_height(height)
+                // See the doc-comment: LargestFirst, not the default
+                // BranchAndBound, to keep CoinJoin's many small denominations
+                // from blowing up the exact-match subset-sum search.
+                .set_selection_strategy(SelectionStrategy::LargestFirst)
+                .set_funding(primary_funds, &primary_account)
+                .add_inputs(extra_inputs);
```
thepastaclaw: **🔴 Blocking: Secondary-account inputs are not reserved in their owning accounts**

`set_funding` filters against and captures only the primary BIP44 account's `ReservationSet`, while `add_inputs` appends UTXOs gathered from every other account without reservation filtering. `TransactionBuilder::assemble_unsigned` consequently records every selected outpoint in the primary set. A later union build re-adds the same secondary-account UTXO because `spendable_utxos` does not consult the primary set, and an account-specific build checks only the secondary account's own set. The manager lock serializes builds while each call is running but does not make the first build's secondary reservations visible after it returns, so sequential calls can return signed transactions that spend the same CoinJoin, BIP32, or DashPay UTXO. Implement wallet-wide or multi-account reservation tracking that filters and reserves every selected outpoint in the ledger consulted by all funding paths.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/core/send.rs
```diff
@@ -0,0 +1,619 @@
...
+    use dashcore::hashes::Hash;
+    use dashcore::{Address as DashAddress, Network, OutPoint, TxOut, Txid};
+    use key_wallet::account::account_type::StandardAccountType;
+    use key_wallet::account::AccountType;
+    use key_wallet::managed_account::ManagedCoreFundsAccount;
+    use key_wallet::Utxo;
+
+    use crate::test_support::{
+        funded_wallet_manager, split_funded_wallet_manager, AlwaysRejectedBroadcaster,
+    };
```
thepastaclaw, line 920: **🔴 Blocking: The new payment tests reference a missing fixture**

`split_funded_wallet_manager` is imported and used by the new union-funding tests, but it is not defined in `crate::test_support` or elsewhere at this head. Running `cargo test -p platform-wallet --lib --no-run` fails with E0432 at this import, followed by cascading type-inference errors, so the crate's unit-test target cannot compile. Add the omitted split-account fixture or rewrite these tests using an existing fixture.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/core_wallet/send.rs
```diff
@@ -0,0 +1,169 @@
...
+        b.copy_from_slice(&buf[*at..end]);
+        *at = end;
+        Ok(u64::from_be_bytes(b))
+    };
+
+    let count = read_u32(blob, &mut cursor)? as usize;
+    let mut outputs = Vec::with_capacity(count);
+    for _ in 0..count {
+        let addr_len = read_u32(blob, &mut cursor)? as usize;
+        let end = cursor + addr_len;
```
thepastaclaw: **🔴 Blocking: Untrusted recipient count can abort the process through an enormous allocation**

The decoder passes the caller-controlled `u32` count directly to `Vec::with_capacity` before checking whether the blob contains that many rows. A four-byte blob declaring `u32::MAX` outputs can therefore request hundreds of gigabytes and enter Rust's process-aborting allocation-failure path instead of returning the documented decode error. On 32-bit targets, capacity or `cursor + addr_len` overflow can also panic inside the nested `extern "C"` function, where the JNI guard cannot safely recover it. Bound the count using the remaining blob length and the standard-transaction output limit, use `try_reserve_exact`, and use checked cursor arithmetic.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/core/send.rs
```diff
@@ -0,0 +1,619 @@
...
+            return Err(PlatformWalletError::TransactionBuild(
+                "at least one output is required".to_string(),
+            ));
+        }
+        if outputs.iter().any(|(_, amount)| *amount == 0) {
+            return Err(PlatformWalletError::TransactionBuild(
+                "every output amount must be greater than zero".to_string(),
+            ));
+        }
+        let outputs_total: u64 = outputs.iter().map(|(_, amount)| *amount).sum();
```
thepastaclaw: **🔴 Blocking: Unchecked output aggregation can build an invalid transaction**

The requested output total is calculated with unchecked `u64` summation, and key-wallet independently performs the same unchecked sum while building the transaction. Kotlin can supply four positive amounts of `1L shl 62`; their mathematical total is `2^64`, which wraps to zero in release builds. Coin selection can then fund only the fee while retaining all four enormous outputs, returning a signed transaction that consensus and relay will reject and reporting meaningless fee/change metadata. Overflow-checking builds panic inside the C FFI path. Use checked aggregation and reject totals above Dash's `MAX_MONEY` before invoking the builder.

```suggestion
let outputs_total = outputs
    .iter()
    .try_fold(0u64, |total, (_, amount)| total.checked_add(*amount))
    .ok_or_else(|| {
        PlatformWalletError::TransactionBuild("output amount total overflow".to_string())
    })?;
if outputs_total > dashcore::blockdata::constants::MAX_MONEY {
    return Err(PlatformWalletError::TransactionBuild(
        "output amount total exceeds MAX_MONEY".to_string(),
    ));
}
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/core/send.rs
```diff
@@ -0,0 +1,619 @@
...
+
+        let mut wm = self.wallet_manager.write().await;
+        let (wallet, info) = wm
+            .get_wallet_and_info_mut(&self.wallet_id)
+            .ok_or_else(|| PlatformWalletError::WalletNotFound(hex::encode(self.wallet_id)))?;
+
+        let height = info.core_wallet.last_processed_height();
+        let fee_rate = FeeRate::new(fee_per_kb.unwrap_or(DEFAULT_FEE_PER_KB));
```
thepastaclaw: **🔴 Blocking: Unbounded fee rates overflow key-wallet fee calculation**

The public Kotlin and FFI APIs accept any non-negative fee rate, including values near `Long.MAX_VALUE`, and pass it directly to `FeeRate::new`. Key-wallet computes fees as `sat_per_kb * size_bytes` with unchecked `u64` multiplication. Large caller-supplied rates therefore panic in overflow-checking Android builds or wrap in release builds, potentially turning an astronomical requested rate into a tiny fee. Validate the rate against a monetary and transaction-size bound or change the fee calculation call chain to return an error from checked arithmetic.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/core/send.rs
```diff
@@ -0,0 +1,619 @@
...
+            ));
+        }
+        if outputs.iter().any(|(_, amount)| *amount == 0) {
+            return Err(PlatformWalletError::TransactionBuild(
+                "every output amount must be greater than zero".to_string(),
+            ));
+        }
+        let outputs_total: u64 = outputs.iter().map(|(_, amount)| *amount).sum();
+
+        let mut wm = self.wallet_manager.write().await;
```
thepastaclaw, line 347: **🟡 Suggestion: External signing holds the wallet-manager write lock**

The write guard acquired here remains live through `build_signed(...).await`. A union-funded payment can select hundreds of inputs, and the production signer resolves and parses the mnemonic, derives the seed and BIP32 key, and invokes the host callback separately for each input. This exclusively blocks wallet sync and every other manager-backed operation for the entire signing phase. The existing `finalize_transaction` path reserves and snapshots the unsigned transaction, UTXOs, and paths under the lock, drops the guard, and signs afterward; this path should follow that model once reservations cover every funding account, releasing all reservations if signing fails.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The new payment API has six blocking defects: secondary-account inputs can be signed twice, malformed recipient metadata can abort the process, unchecked output and fee arithmetic can wrap, insufficient-funds errors lose their Kotlin type, and the added unit tests do not compile. The signing path also holds the wallet-manager write lock throughout repeated host-backed key derivation and signing, unnecessarily blocking wallet sync and access.

### Review provenance

🔴 6 blocking | 🟡 1 suggestion(s)

<details>

```

In `packages/rs-platform-wallet/src/wallet/core/send.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/wallet/core/send.rs:237-238: Secondary-account inputs are not reserved in their owning accounts
  `set_funding` filters against and captures only the primary BIP44 account's `ReservationSet`, while `add_inputs` appends UTXOs gathered from every other account without reservation filtering. `TransactionBuilder::assemble_unsigned` consequently records every selected outpoint in the primary set. A later union build re-adds the same secondary-account UTXO because `spendable_utxos` does not consult the primary set, and an account-specific build checks only the secondary account's own set. The manager lock serializes builds while each call is running but does not make the first build's secondary reservations visible after it returns, so sequential calls can return signed transactions that spend the same CoinJoin, BIP32, or DashPay UTXO. Implement wallet-wide or multi-account reservation tracking that filters and reserves every selected outpoint in the ledger consulted by all funding paths.
- [BLOCKING] packages/rs-platform-wallet/src/wallet/core/send.rs:329-331: The new payment tests reference a missing fixture
  `split_funded_wallet_manager` is imported and used by the new union-funding tests, but it is not defined in `crate::test_support` or elsewhere at this head. Running `cargo test -p platform-wallet --lib --no-run` fails with E0432 at this import, followed by cascading type-inference errors, so the crate's unit-test target cannot compile. Add the omitted split-account fixture or rewrite these tests using an existing fixture.
- [BLOCKING] packages/rs-platform-wallet/src/wallet/core/send.rs:147: Unchecked output aggregation can build an invalid transaction
  The requested output total is calculated with unchecked `u64` summation, and key-wallet independently performs the same unchecked sum while building the transaction. Kotlin can supply four positive amounts of `1L shl 62`; their mathematical total is `2^64`, which wraps to zero in release builds. Coin selection can then fund only the fee while retaining all four enormous outputs, returning a signed transaction that consensus and relay will reject and reporting meaningless fee/change metadata. Overflow-checking builds panic inside the C FFI path. Use checked aggregation and reject totals above Dash's `MAX_MONEY` before invoking the builder.
- [BLOCKING] packages/rs-platform-wallet/src/wallet/core/send.rs:154-155: Unbounded fee rates overflow key-wallet fee calculation
  The public Kotlin and FFI APIs accept any non-negative fee rate, including values near `Long.MAX_VALUE`, and pass it directly to `FeeRate::new`. Key-wallet computes fees as `sat_per_kb * size_bytes` with unchecked `u64` multiplication. Large caller-supplied rates therefore panic in overflow-checking Android builds or wrap in release builds, potentially turning an astronomical requested rate into a tiny fee. Validate the rate against a monetary and transaction-size bound or change the fee calculation call chain to return an error from checked arithmetic.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/core/send.rs:149: External signing holds the wallet-manager write lock
  The write guard acquired here remains live through `build_signed(...).await`. A union-funded payment can select hundreds of inputs, and the production signer resolves and parses the mnemonic, derives the seed and BIP32 key, and invokes the host callback separately for each input. This exclusively blocks wallet sync and every other manager-backed operation for the entire signing phase. The existing `finalize_transaction` path reserves and snapshots the unsigned transaction, UTXOs, and paths under the lock, drops the guard, and signs afterward; this path should follow that model once reservations cover every funding account, releasing all reservations if signing fails.

In `packages/rs-platform-wallet-ffi/src/core_wallet/send.rs`:
- [BLOCKING] packages/rs-platform-wallet-ffi/src/core_wallet/send.rs:62-66: Untrusted recipient count can abort the process through an enormous allocation
  The decoder passes the caller-controlled `u32` count directly to `Vec::with_capacity` before checking whether the blob contains that many rows. A four-byte blob declaring `u32::MAX` outputs can therefore request hundreds of gigabytes and enter Rust's process-aborting allocation-failure path instead of returning the documented decode error. On 32-bit targets, capacity or `cursor + addr_len` overflow can also panic inside the nested `extern "C"` function, where the JNI guard cannot safely recover it. Bound the count using the remaining blob length and the standard-transaction output limit, use `try_reserve_exact`, and use checked cursor arithmetic.

- [BLOCKING] packages/rs-platform-wallet-ffi/src/error.rs:327-339: Payment insufficient-funds errors lose their Kotlin type
  The new `PlatformWalletError::PaymentInsufficientFunds` variant has no conversion arm and falls through to `ErrorUnknown` (99). JNI consequently throws code 1099, which Kotlin maps to `DashSdkError.PlatformWallet.Generic`, instead of the existing `PlatformWallet.CoreInsufficientFunds` type mapped from FFI code 22. This erases the expected shortfall classification from the new API even though the Rust variant preserves `available` and `required` in its message. Map both Core and payment shortfalls to `ErrorCoreInsufficientFunds`, or introduce a dedicated ABI code and Kotlin type.
```
</details>

bfoss765:
This PR reintroduced coin selection across the union of all signable funds accounts, which was blocked on #4184 by @shumkov on 2026-07-21 and replaced by the single-selected-account design (4d3e1322bc, signed off 2026-07-23). The send-raw-tx code predates that re-scope; when it was extracted into this PR on 2026-07-28 it was not diffed against the design decision already settled on the sibling PR, and the extraction was treated as mechanical.

Nothing surfaced it: the code compiled, `cargo check` passed, and the PR carried a test (`payment_funds_from_bip44_and_coinjoin_union`) asserting the union as correct behavior. Both automated reviews then flagged unrelated mechanical defects while treating cross-account funding as the intended premise — one blocker asks for secondary-account inputs to be *reserved* rather than questioning why they exist.

`build_signed_payment` now takes `funding_path: Option<DerivationPath>` and funds from exactly one account, defaulting to unmixed BIP44, matching #4184. The union test is replaced by tests asserting the default never crosses domains and an explicit path selects strictly within one. A new `wallet::funding_privacy` guardrail fails the build if any wallet-wide funds-account iteration is added without an explicit privacy-review marker. The prior reviewer findings are resolved or moot (secondary-account reservation/double-signing are moot with no union; missing fixture added; allocation/overflow/fee-rate bounds added). The write-lock-across-signing item is deliberately deferred to its own PR.

bfoss765:
Added: `accountBalances` rows now expose the account-level `derivationPath` string (Rust row + C-ABI field appended ABI-additively + JNI JSON key). This is the lookup half of the receival-spend fix — the app reads a DashPay receiving-funds account's path here and passes it verbatim as `funding_path` to the single-account send primitive in this PR. Full platform-wallet suite green (504 passed; all 8 send tests + 3 funding-privacy guardrails). The token-minting finalize half (`finalize_signed_payment_from_funding_path`) depends on #4185's signed-payment registry and will land as a separate PR stacked on #4185. Referenced by #4255 (iOS counterpart).

packages/rs-platform-wallet/src/wallet/core/send.rs
```diff
@@ -0,0 +1,921 @@
...
+        if outputs.is_empty() {
+            return Err(PlatformWalletError::TransactionBuild(
+                "at least one output is required".to_string(),
+            ));
+        }
+        if outputs.iter().any(|(_, amount)| *amount == 0) {
+            return Err(PlatformWalletError::TransactionBuild(
+                "every output amount must be greater than zero".to_string(),
+            ));
+        }
```
thepastaclaw, line 256: **🔴 Blocking: Positive dust outputs produce signed transactions that cannot be relayed**

Output validation rejects only zero amounts, so callers can request a one-duff recipient output. `TransactionBuilder::add_output` preserves that amount without applying relay dust policy; for a standard P2PKH destination, `script_pubkey().dust_value()` is 546 duffs. The method can therefore return successfully signed bytes for a transaction that normal Dash nodes reject as nonstandard, despite documenting this as a standard payment primitive intended for later broadcast. Validate every recipient amount against its destination script's dust value before acquiring the wallet lock, reserving inputs, or signing.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/core/send.rs
```diff
@@ -0,0 +1,921 @@
...
+            PlatformWalletError::PaymentInsufficientFunds {
+                available: available_in_account,
+                required: required.max(outputs_total),
+            }
+        }
+        BuilderError::CoinSelection(SelectionError::NoUtxosAvailable) => {
+            PlatformWalletError::PaymentInsufficientFunds {
+                available: available_in_account,
+                required: outputs_total,
+            }
```
thepastaclaw, line 899: **🟡 Suggestion: Insufficient-funds metadata counts reserved inputs as available**

`selectable_value` is calculated from `selected.spendable_utxos(height)`, which applies UTXO spendability rules but does not exclude active reservations. `TransactionBuilder::set_funding`, by contrast, removes reserved outpoints before coin selection. After one build reserves an account's only UTXO, a second build receives `NoUtxosAvailable` but reports that reserved UTXO's full value as `PaymentInsufficientFunds.available`, giving callers an inaccurate actionable balance. Preserve the `available` value carried by the two `InsufficientFunds` variants and report zero for `NoUtxosAvailable` instead of substituting the reservation-blind total.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

Carried-forward prior findings: `prior-fee-overflow`, `prior-dust-output`, and `prior-reserved-balance` are all STILL VALID because the latest commit does not change the affected send path. New findings in the latest delta: none; the derivation-path field consistently uses the same account-path derivation as payment selection and is correctly marshalled and freed across FFI/JNI, while the proposed round-trip test gap does not establish a concrete defect. The two carried-forward transaction-validity blockers require changes.

### Review provenance

<details>

```

In `packages/rs-platform-wallet/src/wallet/core/send.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/wallet/core/send.rs:175-179: Positive dust outputs produce signed transactions that cannot be relayed
  Output validation rejects only zero amounts, so callers can request a one-duff recipient output. `TransactionBuilder::add_output` preserves that amount without applying relay dust policy; for a standard P2PKH destination, `script_pubkey().dust_value()` is 546 duffs. The method can therefore return successfully signed bytes for a transaction that normal Dash nodes reject as nonstandard, despite documenting this as a standard payment primitive intended for later broadcast. Validate every recipient amount against its destination script's dust value before acquiring the wallet lock, reserving inputs, or signing.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/core/send.rs:459-475: Insufficient-funds metadata counts reserved inputs as available
  `selectable_value` is calculated from `selected.spendable_utxos(height)`, which applies UTXO spendability rules but does not exclude active reservations. `TransactionBuilder::set_funding`, by contrast, removes reserved outpoints before coin selection. After one build reserves an account's only UTXO, a second build receives `NoUtxosAvailable` but reports that reserved UTXO's full value as `PaymentInsufficientFunds.available`, giving callers an inaccurate actionable balance. Preserve the `available` value carried by the two `InsufficientFunds` variants and report zero for `NoUtxosAvailable` instead of substituting the reservation-blind total.
```
</details>

packages/rs-platform-wallet/src/wallet/core/send.rs
thepastaclaw: Resolved in this update — _Unchecked output aggregation can build an invalid transaction_ no longer present.

packages/rs-platform-wallet/src/wallet/core/send.rs
thepastaclaw: Resolved in this update — _The new payment tests reference a missing fixture_ no longer present.

packages/rs-platform-wallet/src/wallet/core/send.rs
thepastaclaw: Resolved in this update — _Secondary-account inputs are not reserved in their owning accounts_ no longer present.

packages/rs-platform-wallet-ffi/src/core_wallet/send.rs
thepastaclaw: Resolved in this update — _Untrusted recipient count can abort the process through an enormous allocation_ no longer present.

packages/rs-platform-wallet/src/wallet/core/send.rs
thepastaclaw: Resolved in this update — _External signing holds the wallet-manager write lock_ no longer present.

shumkov:
Reviewed at `fe72da65`. **The re-scope is complete and verified** — no union/multi-account path survives (`send.rs:336-363` iterates only as a lookup, breaking on first match; exactly one `set_funding`), and the `every_union_iteration_is_privacy_reviewed` guardrail would fail on an unmarked reintroduction. Reservation coverage is now structurally complete: `set_funding` clones the shared `Arc<Mutex<..>>` and `assemble_unsigned` reserves every selected outpoint, so **the earlier "secondary-account inputs not reserved" blocker is moot**. Locally: `cargo check` clean, 11/11 new tests pass, clippy clean.

**1. Every `funding_path` failure flattens to `ErrorUnknown(99)`.** `PlatformWalletError::TransactionBuild(..)` has no arm in `rs-platform-wallet-ffi/src/error.rs:271-340`, so "no spendable funds account matches funding derivation path" and "names a watch-only account" both reach Kotlin as `Generic(99, msg)`. Those are precisely the two failure modes the single-account design rests on, and the host can only tell them apart by string-matching. The PR *did* add the arm for `PaymentInsufficientFunds`, so this reads as an oversight rather than a decision. Same applies to the new MAX_MONEY / MAX_FEE_PER_KB rejections and the blob-decode errors.

**2. Six hardening fixes shipped with no tests** — MAX_MONEY output total, MAX_FEE_PER_KB, the `count` bound + `try_reserve_exact` + checked cursor math, UTF-8/network address validation, and `parse_optional_derivation_path`. `core_wallet/send.rs` has no `#[cfg(test)]` module at all. Each of these was a blocking finding; nothing would catch a cleanup that reintroduces `with_capacity`.

**3. No way to release an abandoned build's reservation.** This primitive is build-only, but nothing across FFI/JNI/Kotlin exposes `release_reservation` or `abandon` for its result. If the caller declines to commit, the coins are stranded for `RESERVATION_TTL_BLOCKS = 24` (~1h) — and **indefinitely while `last_processed_height() == 0`**, since `ReservationSet::sweep` early-returns at height 0. The module doc at `send.rs:39-44` names `release_reservation` as the escape hatch, but it doesn't cross the boundary.

**4. `cargo fmt --check` fails on all 5 new files.** Fork PRs never run fmt here, so CI won't tell you.

Minor: the typed shortfall's `available` sums unfiltered `spendable_utxos`, so with a concurrent in-flight build it over-reports and "retry with ≤ available" fails again with the same number; `.unwrap_or(&bip44_acc)` at `send.rs:443` is the same silent-fallback shape that was fixed in #4184 and should be a hard error; the BIP44 change address is derived with `add_to_state = true` before coin selection can fail, burning a pool index; and the fee-direction comment at `send.rs:422-424` has the reasoning backwards (an unpriced input under-reports, not over-).

**Merge order:** this belongs after #4184 (canonical `funding_path` semantics, owns code 29) and #4185 (owns the reservation lifecycle this lacks), rebased on both. At that point `build_signed_payment` should mint a #4185 `ReservationToken` alongside the bytes instead of returning a bare `byte[]` — that closes item 3 and collapses the two "signed payment" concepts into one. Given you've signalled the token-minting half will stack on #4185, holding this until that shape settles may be better than landing an orphan surface.

bfoss765:
Re merge order and the ReservationToken suggestion: agreed on the ordering — #4184 → #4185 → this PR → #4256. On the token shape: rather than reworking this PR's return type, #4256 (stacked on #4185 + this PR) already delivers exactly that — `finalize_signed_payment_from_funding_path` mints a ReservationToken recorded against the resolved funding account, consumed by the existing broadcast/release-by-token flow. Proposal: this PR stays the build-only primitive (used by #4256 internally), and the tokenized API in #4256 is what apps call for receival spends — which is what the Android app now does. The remaining inline items (typed error arm, tests for the hardening fixes, fmt) are being addressed on this branch now. 🤖 Generated with [Claude Code](https://claude.com/claude-code)

packages/rs-platform-wallet/src/wallet/core/send.rs
bfoss765: Fixed in 920b70683c — the finding was valid at `fe72da65`; only zero amounts were rejected.

`build_signed_payment` now checks every recipient against its OWN destination script's `dust_value()` (`wallet/core/send.rs`), not a shared constant — the threshold is script-shaped, which is also why key-wallet's hard-coded `546` change-dust literal was not reusable. Placed after the cheap recipient-count bound (so an absurd list doesn't drive a script serialization per output) but before the wallet lock, before any input is reserved, and before the signer is called.

Tests: `below_dust_outputs_are_rejected` covers 1 duff, `dust - 1`, a below-dust output hidden among valid ones, and that exactly `546` still builds; `a_rejected_dust_request_reserves_nothing` pins that the refusal costs the caller nothing — three refused requests, then an ordinary payment still finds the account's coins selectable.

packages/rs-platform-wallet/src/wallet/core/send.rs
bfoss765: Fixed in 920b70683c — and the finding was sharper than the bound that was already there.

`MAX_FEE_PER_KB` existed at `fe72da65`, but it was derived as `MAX_MONEY / 100`, which is only sound if the transaction stays under the 100 kB standard limit — and the method never enforced that limit. So the product still overflowed at ~878 kB, exactly as the sibling finding on #4256 works out. Two dimensions reach it: an oversized recipient list, and a funding account holding a few thousand small denominations (a realistic CoinJoin account) with no oversized list at all.

Three-part fix in `wallet/core/send.rs`:

- `MAX_STANDARD_TX_SIZE` is now enforced, derived from `dashcore::policy::MAX_STANDARD_TX_WEIGHT / 4` rather than hard-coded. The recipient count is bounded against it with checked arithmetic mirroring key-wallet's own base-size formula (`outputs.len() * 34` is an unchecked `usize` multiply inside key-wallet), requiring room for at least one 148-byte input.
- `MAX_FEE_PER_KB` is re-derived as `u64::MAX / u32::MAX`. That makes `sat_per_kb * size_bytes` representable for ANY size a `u32` can express, so it does not depend on the input count — which is the part the size limit alone cannot cover. ~4.29e9 duffs/kB is ~43 DASH/kB, three orders of magnitude above the 1_000 default, so nothing legitimate is rejected.
- the signed transaction is re-measured against the limit, since the input count is not knowable until selection has run.

Tests: `excessive_fee_rates_are_rejected` (u64::MAX, u64::MAX/2, bound+1, plus a sane rate that still works), `max_fee_rate_cannot_overflow_key_wallets_fee_product` (asserts no overflow at 100 kB, 878_434, and u32::MAX — it would fail if a cleanup restored `MAX_MONEY / 100`), and `oversized_recipient_lists_are_rejected`, which includes the 25,835-recipient case at the maximum accepted rate.

packages/rs-platform-wallet/src/wallet/core/send.rs
bfoss765: Not changed — deliberately, and I'd like to push back on the suggested shape.

The finding is accurate about the mechanism: `selectable_value` comes from `selected.spendable_utxos(height)`, which does not exclude active reservations, while `set_funding` removes reserved outpoints before selection. So with a concurrent in-flight build, `available` can over-report.

But the proposed remedy — report zero for `NoUtxosAvailable` — trades one wrong number for another. Zero tells the user "this account is empty", which is false and is the more misleading of the two: the coins exist and become selectable again the moment the other build broadcasts, is abandoned, or its reservation TTLs out. `available` is rendered as an actionable balance, and "0" invites the caller to conclude the account is unusable.

Reporting the reservation-aware figure would be the genuinely correct fix, and it is cheap — filter the `spendable_utxos` pass by the account's `ReservationSet`. I did not do it here because at this point in the function the reservation ledger has not been consulted yet (`set_funding` is what captures it), so it needs a small restructure rather than a one-liner, and this PR is already carrying two blocking fixes to funds-critical arithmetic. I'd rather not fold a third behavioural change into the same commit.

Happy to do it as a follow-up, or in this PR if you'd prefer it land together — but I'd want it to be the filtered real number, not zero. Flagging explicitly so this isn't read as silently dropped.

packages/rs-platform-wallet/src/wallet/funding_privacy.rs
```diff
@@ -0,0 +1,359 @@
...
+            .map(|i| i.previous_output)
+            .collect();
+        assert!(
+            spent.iter().all(|op| coinjoin_ops.contains(op)),
+            "every input must come from the named CoinJoin account, spent {spent:?}"
+        );
+        assert!(
+            !spent.iter().any(|op| bip44_ops.contains(op)),
+            "an explicitly-named domain must not pull BIP44 inputs, spent {spent:?}"
+        );
```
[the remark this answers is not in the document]
bfoss765: Added in 920b70683c. `all` is vacuously true on an empty set, so the guardrail could pass while proving nothing — the non-empty assertion is in now, matching the sibling `explicit_coinjoin_path_selects_only_coinjoin` in `wallet::core::send`.

packages/rs-platform-wallet-ffi/src/core_wallet_types.rs
```diff
@@ -491,6 +491,17 @@ fn account_index_of(at: &key_wallet::account::AccountType) -> u32 {
...
+/// `derivation_path` is a heap-owned, NUL-terminated UTF-8 C string
+/// carrying the account-level derivation path (e.g. the DIP-15
+/// `m/9'/coinType'/15'/{index}'/0x<user 64hex>/0x<friend 64hex>` for a
+/// DashPay receiving-funds account). It is `null` for account types that
+/// have no derivable account-level path. Freed by
+/// [`platform_wallet_manager_free_account_balances`]. Appended as the
+/// LAST field so the layout change is purely additive — existing readers
+/// of the earlier fields are byte-compatible, but any consumer that
+/// indexes the array by struct stride (iOS/Android bindings) must be
+/// regenerated against this definition before use.
```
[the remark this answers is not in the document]
bfoss765: Not changed here — this is the ABI-ordering question the maintainer is handling across the v4.1 stack, so I don't want to unilaterally version the struct from this PR. Noting for the record that the `derivation_path` field is appended (not inserted), and that the Kotlin/JNI binding is regenerated and shipped in lockstep with the `.so` in this stack, so there is no mixed-version load path in practice. Deferring the decision rather than closing it.

packages/rs-platform-wallet-ffi/src/core_wallet/send.rs
```diff
@@ -0,0 +1,169 @@
...
+            ));
+        }
+        let mut b = [0u8; 8];
+        b.copy_from_slice(&buf[*at..end]);
+        *at = end;
+        Ok(u64::from_be_bytes(b))
+    };
+
+    let count = read_u32(blob, &mut cursor)? as usize;
+    let mut outputs = Vec::with_capacity(count);
```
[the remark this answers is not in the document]
bfoss765: The bound itself landed earlier in `0809290`; what was missing — and what shumkov flagged as review item 2 — is that it shipped with no test. `core_wallet/send.rs` had no `#[cfg(test)]` module at all, so nothing would have caught a cleanup restoring `Vec::with_capacity(count)`.

Covered now in 920b70683c:

- `an_impossible_count_is_rejected_without_allocating` — `u32::MAX`, `u32::MAX / 2`, 1_000_000 and 1 against a header-only blob, asserting the "holds at most 0" bound rather than an allocation attempt;
- `a_count_exceeding_the_rows_present_is_rejected` — a count that merely overstates a non-empty blob;
- `truncation_at_any_boundary_is_a_clean_error` — every truncation offset of a valid blob, pinning the checked cursor arithmetic (the `cursor + addr_len` overflow reachable on 32-bit armeabi-v7a);
- `an_absurd_address_length_is_rejected`, `a_non_utf8_address_is_rejected`, `an_unparseable_address_is_rejected`, `a_wrong_network_address_is_rejected` (which also asserts the same blob decodes fine against its own network, so the rejection is the network check and not a malformed fixture).

bfoss765:
Pushed `920b70683c`. Taking @shumkov's items in order, plus the two PastaClaw blockers.

**1. `TransactionBuild` flattened to `ErrorUnknown(99)` — fixed.** Added `ErrorTransactionBuild` with an arm in the blanket `From` impl (`rs-platform-wallet-ffi/src/error.rs`) and the matching `DashSdkError.PlatformWallet.TransactionBuild` on the Kotlin side. It covers what you listed: unmatched and watch-only `funding_path`, the MAX_MONEY / fee-rate / dust / size rejections, and the blob-decode errors.

Numbering: I took **32**, not 27. On the integration line 27–31 are all claimed (`ErrorStaleReservationToken` 27 and `ErrorReservationTokenConsumed` 28 from #4185, `ErrorAssetLockInsufficientFunds` 29 from #4184, `ErrorReservationWalletMismatch` 30, `ErrorSigningKeyUnavailable` 31), and #4185's branch and #4184's branch currently both claim 29 in isolation. 32 is free on every branch of the stack, so this needs no renumbering whatever order they land in — I've left the 29 contention alone since it's yours to sequence. Tests: `transaction_build_failures_map_to_a_dedicated_code`, `transaction_build_code_is_thirty_two`, `transactionBuildFailuresGetTheirOwnType` (Kotlin).

**2. Six hardening fixes with no tests — fixed.** `core_wallet/send.rs` now has a test module. 15 new tests across the FFI and wallet crates cover the `count` bound + `try_reserve_exact` + checked cursor math (including every truncation offset), UTF-8 and wrong-network address rejection, `parse_optional_derivation_path` (null/empty vs. garbage vs. non-UTF-8 — a malformed path must not silently become "default account"), the MAX_MONEY aggregation, and the fee-rate bound.

**3. No way to release an abandoned build's reservation — not addressed here**, per your merge-order note; that's the token-minting shape that stacks on #4185 (#4256).

**4. `cargo fmt --check` — fixed**, all five files.

**PastaClaw blockers, verified against `fe72da65` first:**

- **Dust (valid, fixed).** Only zero amounts were rejected. Now checked per output against its own script's `dust_value()`, before the lock, before any reservation, before the signer.
- **Fee-rate overflow (partly valid — the bound existed but was unsound, now fixed).** `MAX_FEE_PER_KB` was there, but derived as `MAX_MONEY / 100`, which only holds if the transaction stays under 100 kB — which this method never enforced. The product still overflowed at ~878 kB. Now: `MAX_STANDARD_TX_SIZE` is actually enforced (derived from `dashcore::policy::MAX_STANDARD_TX_WEIGHT / 4`) with checked arithmetic on the output side, the signed transaction is re-measured, and `MAX_FEE_PER_KB` is re-derived as `u64::MAX / u32::MAX` so the product is safe for any `u32`-expressible size regardless of input count. That last part matters independently: a CoinJoin account with a few thousand small denominations reaches ~878 kB with no oversized recipient list at all.

**Also picked up from your "minor" list:** the inverted fee-direction comment, and the `PaymentInsufficientFunds` doc, which still described the pre-#4184 union semantics ("the union of the wallet's signable funds accounts… total selectable value across those accounts") — wrong in a funds-relevant way after the re-scope.

**Left open, deliberately, flagged so they aren't read as dropped:** the reservation-blind `available` (replied on the thread — I'd want the filtered real number rather than the suggested zero, which is the more misleading of the two); `.unwrap_or(&bip44_acc)`; and the change-address pool-index burn. Happy to fold any of them in if you'd rather they land here.

Suites: `cargo test -p platform-wallet --lib` **510 passed / 0 failed** (504 baseline + 6), `-p platform-wallet-ffi --lib` **222 passed / 0 failed** (207 baseline + 15), `cargo check -p platform-wallet-ffi -p rs-unified-sdk-jni` clean, Kotlin `:sdk:testDebugUnitTest` green, `cargo fmt --check` clean.

bfoss765:
@shumkov — following up on [item 3](https://github.com/dashpay/platform/pull/4247#issuecomment-5145025234). I said earlier this would ride on #4185/#4256; on reflection that left a real hazard sitting in a PR that is otherwise ready, so I've closed it here instead. Fixed in `0dcdc743e7`.

**What landed.** A standalone release/abandon across all four layers:

```
CoreWallet::release_payment_reservation(&Transaction, Option<DerivationPath>)
core_wallet_release_payment_reservation            (FFI)
coreWalletReleasePaymentReservation                (JNI)
ManagedPlatformWallet.releasePaymentReservation    (Kotlin)
```

A caller that declines to commit hands back the transaction it was given and the same `funding_path`, and the inputs become selectable again immediately.

**On the height-0 case specifically** — the part of your item that had no workaround at all: the release consults no height and never calls `sweep`, so it works precisely where the TTL backstop cannot. There's a test that pins both halves: at `last_processed_height() == 0` it runs 30 rebuild attempts to show the reservation is genuinely never reclaimed, then releases and rebuilds successfully.

**Scoping.** The transaction is the ownership signal — a reserved outpoint is skipped by every other build's coin selection, so no concurrent build can hold a reservation on any input of the transaction being released. That's the same signal the internal `release_reservation_after_rejected_broadcast` cleanup already uses, so this exposes an existing primitive rather than adding a new notion of ownership. It's idempotent and a silent no-op after a successful broadcast (it can't resurrect a spent coin — selection reads the UTXO set, which sync has already updated), so it's safe in an unconditional `finally` without the caller tracking broadcast success. Tests cover release-then-reselect, release twice, release after a processed broadcast, and a release aimed at the wrong account freeing nothing.

**On collapsing the two "signed payment" concepts** — agreed, and tracked in #4263. Deferring rather than doing it here for two reasons: minting a `ReservationToken` from `build_signed_payment` would make this PR unmergeable until #4185 is approved, and it would reshape the bare-bytes return that the Android wallet already consumes, where dashj deliberately still owns commit/broadcast. That migration is app-side work with its own testing. With the stranding hole closed independently, the unification is an API-shape cleanup that can land cleanly on top of #4185 rather than a gating concern for this PR. The issue lays out the sequencing.

Suites: `cargo test -p platform-wallet --lib` **516 passed / 0 failed** (510 + 6 new), `-p platform-wallet-ffi --lib` **222 passed / 0 failed**, `cargo check -p platform-wallet-ffi -p rs-unified-sdk-jni` clean, `cargo fmt --check` clean, Kotlin `:sdk:compileDebugKotlin` + `:sdk:testDebugUnitTest` green.

bfoss765:
@thepastaclaw the review has been showing "Codex precheck starting" at `0dcdc743` since 2026-08-01 13:50 UTC with no verdict. The last completed verdict is against `fe72da6517`, which was superseded by `920b70683c`. Could you re-run at the current head?

bfoss765:
Pushed a new head **f940296118** — a one-word module-doc tightening in `core_wallet/send.rs` (`cargo check -p platform-wallet-ffi` clean). No code changes vs `0dcdc743e7`; the review gate has shown "Codex precheck starting" at that head since Aug 1 and did not respond to mentions, so this push is to hand it a fresh head.

@coderabbitai full review

bfoss765:
@thepastaclaw please re-review at head `f940296118` — the previous verdict (2 blocking) is pinned to `fe72da6517` (Jul 31), which was superseded by `920b7068` and later heads; shumkov's human review at that same sha judged the re-scope complete. The Aug 3 02:46 comment mistakenly pinged only CodeRabbit, hence this explicit request.

bfoss765:
## Restacked onto #4185

**Stack order (bottom → top):** #4185 → **#4247** → #4256

Previously this PR branched from upstream independently of #4185, while #4256 carried its own divergent copy of this work. That made any pair unmergeable. The three are now a linear stack.

**Review only this PR's own commits** — `3dec774929..HEAD` (the `#4185` commits below it are that PR's, already reviewed there):

| commit | |
|---|---|
| `c75dafb5e7` | union-funding `build_signed_payment` core primitive |
| `4c1b99284e` | plumb through FFI, JNI, Kotlin |
| `bdd3377eda` | re-scope to single-account funding |
| `f1657dcf2e` | account-level `derivationPath` in `accountBalances` |
| `d01c859251` | dust + size bounds, typed build errors |
| `a1ba070328` | fail closed when no account matches the funding path |
| `615f0d14be` | expose `release_payment_reservation` |
| `ebbafa49da` | doc fix |
| `db8f5e7429` | **new** — adopt #4185's `WalletGeneration` in the test fixtures |
| `6e87562af5` | **new** — mirror `ErrorTransactionBuild` (32) into Swift |
| `1606ec0cc0` | **new** — rustfmt |
| merge | current `v4.2-dev` (post-#4268) |

### Blocking findings — both now resolved
- **`3db3f420804b` — "the new payment tests reference a missing fixture"**: fixed *by the restack itself*. `split_funded_wallet_manager` lives in #4185's `test_support.rs`; sitting on top of #4185 makes it resolve. `cargo test -p platform-wallet -p platform-wallet-ffi` now compiles and passes **554 + 248**.
- **`d1898625728f` — "positive dust outputs produce unrelayable transactions"**: every recipient amount is validated against `address.script_pubkey().dust_value()` *before* the wallet lock, reservation or signing, with a regression test.

### Also fixed while restacking
`ErrorTransactionBuild = 32` was added to the Rust enum and the Kotlin decoder by this PR but never to the Swift mirror, so every typed build rejection reached iOS through the `default:` arm as `errorUnknown` (99) — reintroducing on iOS exactly the string-matching this code was split out of `ErrorUnknown` to end. Swift now has the enum case, the `init(ffi:)` mapping and the `PlatformWalletError.transactionBuild` arm.

### Merge with #4268
`ErrorShutdownIncomplete = 27` (#4268) vs `ErrorTransactionBuild = 32` (this PR) — additive, both kept, no renumbering. `mergeable` is back to **MERGEABLE**.

### Not addressed here
The non-blocking suggestions `f33dd4dd99a9` / `1a6cdcdc8b3e` (hold the manager write lock across signing; reservation-blind `available` in the shortfall metadata) are unchanged — both are behavioural reworks of the funding critical section, out of scope for a restack and worth their own commit.

packages/rs-platform-wallet/src/wallet/core/send.rs
bfoss765: Fixed by the restack. This PR now sits on top of #4185 (stack: #4185 → #4247 → #4256), and `split_funded_wallet_manager` is defined in #4185's `test_support.rs`, so the import resolves.

`cargo test -p platform-wallet -p platform-wallet-ffi` compiles and passes **554 + 248** at head `540def16a0`.

packages/rs-platform-wallet/src/wallet/core/send.rs
bfoss765: Fixed. Every recipient amount is now validated against its own destination script's relay dust threshold — `address.script_pubkey().dust_value()` — **before** the wallet lock is taken, before any input is reserved and before signing, so a rejected request reserves nothing:

```rust
let dust = address.script_pubkey().dust_value().to_sat();
if *amount < dust {
    return Err(PlatformWalletError::TransactionBuild(format!(
        "output {amount} duffs to {address} is below the {dust}-duff dust ..."
    )));
}
```

Covered by `below_dust_outputs_are_rejected` and `a_rejected_dust_request_reserves_nothing`. Head `540def16a0`.

packages/rs-platform-wallet/src/wallet/core/send.rs
```diff
@@ -0,0 +1,1756 @@
...
+        // whose recipient list already passed the output-side bound but whose
+        // funding account then contributed enough small inputs to push the
+        // whole transaction over the limit.
+        let signed_size = transaction.size();
+        if signed_size > MAX_STANDARD_TX_SIZE {
+            return Err(PlatformWalletError::TransactionBuild(format!(
+                "the signed transaction is {signed_size} bytes, over the \
+                 {MAX_STANDARD_TX_SIZE}-byte standard transaction limit; it would not relay. \
+                 Send a smaller amount (fewer inputs) or fewer recipients"
+            )));
```
thepastaclaw, line 632: **🔴 Blocking: Post-signing size rejection strands selected inputs**

`build_signed` has already reserved the selected inputs when this size check runs, but this error path returns without releasing them and without returning the transaction bytes the caller would need for `release_payment_reservation`. The branch is reachable when the recipient-side estimate passes but multiple small funding inputs push the signed transaction over the standard limit. Those inputs remain unavailable until the TTL sweep, and can remain unavailable for the process lifetime while the wallet height is zero. Use `build_signed_reserved`, retain its reservation token, and call `release_reservation_if_owner` before returning this error.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/core/send.rs
```diff
@@ -0,0 +1,1756 @@
...
+                .to_account_type()
+                .derivation_path(network)
+                .map_err(|e| {
+                    PlatformWalletError::TransactionBuild(format!(
+                        "failed to derive account-level path for a funds account: {e}"
+                    ))
+                })?;
+            if account_path == funding_path {
+                account.release_reservation(transaction);
+                return Ok(());
```
thepastaclaw: **🔴 Blocking: Abandonment can release another payment's inputs**

`release_reservation` removes entries by outpoint without checking their owner, while the exact key-wallet dependency explicitly restricts that operation to known-spent inputs and provides `release_reservation_if_owner` for abandonment. If build A expires and build B re-reserves the same still-unspent input, a delayed release of A removes B's reservation and permits another conflicting build. The documented unconditional-cleanup behavior is also unsafe after a successful external broadcast but before SDK sync removes the input from its UTXO set: releasing the reservation immediately makes that input selectable again. Preserve the `ReservationToken` returned by `build_signed_reserved` through the Rust, FFI, JNI, and Kotlin APIs, release only if that token still owns the inputs, and do not release merely because sync has not yet observed a successfully broadcast transaction.

<sub>source: ['codex']</sub>

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
```diff
@@ -172,6 +172,403 @@ class ManagedPlatformWallet internal constructor(
...
+        require(recipients.isNotEmpty()) { "recipients must not be empty" }
+        require(recipients.all { it.second > 0 }) { "every recipient amount must be positive" }
+        require(feePerKb >= 0) { "feePerKb must be non-negative, got $feePerKb" }
+
+        val outputsBlob = encodePaymentOutputs(recipients)
+        mapNativeErrors {
+            coreWallet().use { core ->
+                decodeSignedPayment(
+                    core.buildSignedPayment(outputsBlob, feePerKb, coreSignerHandle, fundingPath),
+                )
```
thepastaclaw, line 496: **🔴 Blocking: Coroutine cancellation can orphan a completed payment reservation**

`gate.op` executes the blocking JNI build through `withContext(Dispatchers.IO)`. If the caller is cancelled while native signing is running, the native call can still complete and reserve inputs, after which `withContext` applies prompt cancellation and discards the completed `SignedCorePayment` before the caller receives its transaction bytes. The explicit release API is then unusable, leaving the reservation until its TTL or indefinitely at height zero. The token-owning overload above uses `opWithCleanupOnCancellation` for this exact ownership handoff; this raw-payment API likewise needs an owning native reservation handle whose synchronous cleanup is established before the result can be discarded.

<sub>source: ['codex']</sub>

packages/rs-unified-sdk-jni/src/wallet_manager.rs
```diff
@@ -1006,6 +1006,195 @@ pub extern "system" fn Java_org_dashfoundation_dashsdk_ffi_WalletManagerNative_c
...
+        packed.extend_from_slice(&out_fee.to_be_bytes());
+        packed.extend_from_slice(&out_change.to_be_bytes());
+        packed.extend_from_slice(tx_bytes);
+        unsafe {
+            platform_wallet_ffi::core_wallet_free_payment_bytes(out_tx_bytes, out_tx_len);
+        }
+
+        env.byte_array_from_slice(&packed)
+            .map(|a| a.into_raw())
+            .unwrap_or(ptr::null_mut())
```
thepastaclaw: **🟡 Suggestion: JNI result allocation failure loses the reservation cleanup handle**

The native build has successfully reserved its inputs before this code creates the JVM `byte[]`. If `byte_array_from_slice` fails, the packed transaction is dropped and Kotlin receives no bytes or token with which to release the reservation. This can leave the account unavailable until the TTL, or indefinitely at height zero. Retain the reservation owner token through marshalling and release it if ownership transfer into the JVM fails.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The exact head fixes the prior fee-overflow and dust-output blockers, while the reservation-blind insufficient-funds metadata remains. The new post-signing size rejection, abandonment API, and Kotlin coroutine handoff can strand reservations or release inputs belonging to another payment, so changes are required before this send API is safe to ship.

### Review provenance

<details>

```

In `packages/rs-platform-wallet/src/wallet/core/send.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/wallet/core/send.rs:550-556: Post-signing size rejection strands selected inputs
  `build_signed` has already reserved the selected inputs when this size check runs, but this error path returns without releasing them and without returning the transaction bytes the caller would need for `release_payment_reservation`. The branch is reachable when the recipient-side estimate passes but multiple small funding inputs push the signed transaction over the standard limit. Those inputs remain unavailable until the TTL sweep, and can remain unavailable for the process lifetime while the wallet height is zero. Use `build_signed_reserved`, retain its reservation token, and call `release_reservation_if_owner` before returning this error.
- [BLOCKING] packages/rs-platform-wallet/src/wallet/core/send.rs:657-658: Abandonment can release another payment's inputs
  `release_reservation` removes entries by outpoint without checking their owner, while the exact key-wallet dependency explicitly restricts that operation to known-spent inputs and provides `release_reservation_if_owner` for abandonment. If build A expires and build B re-reserves the same still-unspent input, a delayed release of A removes B's reservation and permits another conflicting build. The documented unconditional-cleanup behavior is also unsafe after a successful external broadcast but before SDK sync removes the input from its UTXO set: releasing the reservation immediately makes that input selectable again. Preserve the `ReservationToken` returned by `build_signed_reserved` through the Rust, FFI, JNI, and Kotlin APIs, release only if that token still owns the inputs, and do not release merely because sync has not yet observed a successfully broadcast transaction.

- [BLOCKING] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt:425-440: Coroutine cancellation can orphan a completed payment reservation
  `gate.op` executes the blocking JNI build through `withContext(Dispatchers.IO)`. If the caller is cancelled while native signing is running, the native call can still complete and reserve inputs, after which `withContext` applies prompt cancellation and discards the completed `SignedCorePayment` before the caller receives its transaction bytes. The explicit release API is then unusable, leaving the reservation until its TTL or indefinitely at height zero. The token-owning overload above uses `opWithCleanupOnCancellation` for this exact ownership handoff; this raw-payment API likewise needs an owning native reservation handle whose synchronous cleanup is established before the result can be discarded.

- [SUGGESTION] packages/rs-unified-sdk-jni/src/wallet_manager.rs:1112-1118: JNI result allocation failure loses the reservation cleanup handle
  The native build has successfully reserved its inputs before this code creates the JVM `byte[]`. If `byte_array_from_slice` fails, the packed transaction is dropped and Kotlin receives no bytes or token with which to release the reservation. This can leave the account unavailable until the TTL, or indefinitely at height zero. Retain the reservation owner token through marshalling and release it if ownership transfer into the JVM fails.
```
</details>

packages/rs-platform-wallet/src/wallet/core/send.rs
thepastaclaw: Resolved in this update — _Unbounded fee rates overflow key-wallet fee calculation_ no longer present.

packages/rs-platform-wallet/src/wallet/core/send.rs
bfoss765: Confirmed and fixed in 3da99396de.

The path is reachable exactly as described: `outputs.len()` passes the output-side pre-check, coin selection then contributes enough small inputs to push the *signed* transaction over `MAX_STANDARD_TX_SIZE`. By then `build_signed` has already reserved every selected input, and the error path returned without releasing and without handing back the transaction the caller would need for `release_payment_reservation` — so the caller had no way to recover them either. And as noted, the TTL backstop is not a fallback here: `ReservationSet::sweep` early-returns at height 0, so on a freshly restored wallet the inputs were stranded for the life of the process. Because the trigger is "many small inputs", the stranded amount is typically the whole funding account.

Changes:

- `build_signed` → `build_signed_reserved`, retaining the `ReservationToken`.
- The size-rejection path releases via `release_reservation_if_owner` before returning the error.
- The release is owner-guarded even though the manager write lock has been held continuously since selection (so neither a sweep nor a re-reservation can interleave). The guard is free and keeps the path correct by construction rather than by an argument about the current locking — which is the reasoning that failed review in the sibling finding on this same method. A failure to resolve the account is logged rather than propagated, so it cannot mask the size error the caller needs to see.

Also folded into the same helper the explicit abandon path uses, so the guard cannot be applied to one release site and forgotten on the other.

New test `an_oversized_signed_transaction_strands_no_reservation`: eight small UTXOs plus the largest recipient count the pre-check still admits, asserting the failure is the post-signing size rejection (not a pre-flight bound), and then that an ordinary payment succeeds and reselects the released inputs. Before the fix that second build failed with `PaymentInsufficientFunds` against a plainly funded wallet.

packages/rs-platform-wallet/src/wallet/core/send.rs
bfoss765: Confirmed and fixed in 3da99396de. Both clauses were real.

**The race.** `release_reservation` removes by outpoint with no owner check, and the doc block on this method argued that was safe because "a reserved outpoint is skipped by every subsequent coin selection — so no concurrent build can hold a reservation on any input of `transaction`". That reasoning ignores the TTL sweep: once key-wallet reclaims the reservation the outpoint *is* selectable again, a second build can legitimately re-reserve it under a new token, and the late release then frees that build's inputs. key-wallet's own docs on `release_reservation_if_owner` describe this hazard and state the platform layer "cannot make this safe on its own", since the sweep happens inside key-wallet invisibly. **That doc block has been corrected, not just the code** — it now says what actually makes the release safe and records that the old argument was wrong.

**Threading the token.** key-wallet's `ReservationToken` is deliberately unforgeable — private counter, no public constructor, no accessor — so it cannot cross the C ABI by value, and serializing it would defeat its purpose. It is threaded as follows, mirroring how `SignedPaymentRegistry` keeps its funding token Rust-side while a `u64` payment handle crosses the boundary:

- Rust: `SignedCorePayment` carries the token; `release_payment_reservation` takes it and uses `release_reservation_if_owner`, plus the `Arc::ptr_eq` generation guard from `release_transaction_reservation` so a same-id wallet recreation cannot have its new reservation freed.
- FFI: new `out_reservation_handle` out-param on the build, new `reservation_handle` param on the release. The handle is allocated from a bounded FIFO table (new `core_wallet/payment_reservation.rs`) that holds the real token. An unknown or aged-out handle is **refused**, never downgraded to the unguarded release. The table is bounded because this primitive does not broadcast, so a build that goes out on the wire never reports back — reaching eviction needs thousands of un-released builds, all of them broadcast ones.
- JNI: the packed result blob is now `u64 fee, u64 change, u64 reservationHandle,` then tx bytes; the release takes the handle. **All three layout docs updated in step** (JNI, `WalletManagerNative`, `decodeSignedPayment`).
- Kotlin: `SignedCorePayment.reservationHandle`, `releasePaymentReservation(txBytes, fundingPath, reservationHandle)`, plus an overload taking the `SignedCorePayment` so bytes and handle cannot be mismatched between two in-flight builds.

**The broadcast clause.** The "## Idempotent, and safe after a broadcast" section advertised this as safe to wire into an unconditional cleanup path — a `finally`, a teardown hook — on the grounds that sync removes a broadcast spend from the UTXO set anyway. **That guidance is removed** at every layer (Rust, FFI, JNI, Kotlin KDoc). The gap is the window *before* sync observes it: this primitive does not broadcast, so between the caller's successful external broadcast and sync processing that spend, the inputs are still in the UTXO set and the reservation is the only thing keeping a second build off them. The contract is now: release a build you decided **not** to broadcast; never one you did. Idempotency is still documented, but scoped to repeated releases of the same abandoned build.

Tests:

- `an_owner_guarded_release_cannot_free_a_re_reserved_input` reproduces the window end to end — reserve, age past `RESERVATION_TTL_BLOCKS`, let a second build re-take the outpoint — and asserts the guarded release leaves it intact. It then releases the same transaction *without* a token as a control, showing the unguarded release does free it, so the guard is demonstrably what closes the window.
- `releasing_a_broadcast_build_before_sync_reopens_its_inputs` pins the second clause: the next build reselects the same inputs and produces a distinct transaction spending the same coins.
- The pre-existing `abandoning_after_broadcast_is_a_no_op` is retained but rescoped to the post-sync case it actually proves, and no longer claims the `finally` guidance.

One fixture note surfaced by the generation guard: `split_funded_wallet_manager` minted a `WalletGeneration` but did not return it, so every caller invented a fresh one. That made the wallets foreign to their own `CoreWallet` and would have made every generation-bound assertion vacuous. The fixture now returns the real handle.

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
bfoss765: Confirmed and fixed in 647dbba660.

The diagnosis holds: the blocking JNI build cannot observe cancellation once native signing has started, and it reserves the funding UTXOs before returning, so the reservation exists by the time `withContext` dispatches back. That handoff is a prompt-cancellation point — a caller cancelled while JNI ran never receives the `SignedCorePayment`, and with it never receives the tx bytes or the reservation handle that are the only way to release. The explicit release API is then unusable and the reservation sits until the TTL, or indefinitely at height 0 where the sweep never runs.

`buildSignedPayment` now uses `gate.opWithCleanupOnCancellation`, the same handoff guard the token-owning overload above it uses, with a cleanup that releases the discarded payment's reservation using the handle threaded in for the sibling finding — so the cleanup is owner-guarded too, not a bare by-outpoint release.

The cleanup is deliberately synchronous and best-effort:

- Blocking rather than suspending, because it runs from `opWithCleanupOnCancellation`'s `finally` where the coroutine is already cancelled and any suspension would be refused. The underlying native call is an in-memory map operation.
- It never throws. `Throwable` is caught (not `Exception`) because the native boundary can surface `UnsatisfiedLinkError` and friends, and logged at warn. Letting anything escape would replace the caller's `CancellationException` with an unrelated error thrown out of a `finally`; the reservation is recoverable by other means, a corrupted cancellation signal is not.

Verified with `./gradlew :sdk:compileDebugKotlin :sdk:compileDebugUnitTestKotlin` (both green).

packages/rs-platform-wallet/src/wallet/core/send.rs
bfoss765: Fixed in `51a31c5406` (`fix(platform-wallet): re-scope build_signed_payment to single-account`).

The union across accounts is gone. `build_signed_payment` now confines selection to a single funding account — the one named by `funding_path`, defaulting to the unmixed BIP44 account — so there is no set of secondary-account UTXOs to reserve in a foreign ledger. The module contract states it directly (`wallet/core/send.rs` header: "Selection is confined to a single funding account … never a union across accounts"), and every selected input is reserved in that one account's `ReservationSet`.

packages/rs-platform-wallet-ffi/src/core_wallet/send.rs
bfoss765: Fixed in `51a31c5406`.

The decoder no longer hands the caller-controlled `u32` to `Vec::with_capacity`. It computes `max_possible` from the remaining blob length (min bytes per entry) and rejects any `count` exceeding it *before* allocating (`core_wallet/send.rs:85`). Two regression tests pin it: `an_impossible_count_is_rejected_without_allocating` and `a_count_exceeding_the_rows_present_is_rejected`, plus a comment guarding against a future reintroduction of `Vec::with_capacity(count)`.

packages/rs-platform-wallet/src/wallet/core/send.rs
bfoss765: Fixed in `51a31c5406`. Output aggregation and the fee/amount sums are now checked (`checked_add`, with the per-amount consensus cap enforced before aggregation), so an overflowing request is rejected with a typed error rather than silently wrapping into an invalid transaction. key-wallet's own summation is reached only after this validation passes.

packages/rs-platform-wallet/src/wallet/core/send.rs
bfoss765: Holding the wallet-manager write guard across the signer await is deliberate, and releasing it there would reintroduce the exact race this PR exists to close.

The guard spans coin selection → reservation → sign as one critical section. `build_signed_reserved` selects inputs, reserves them, and signs under a single hold; the reservation lives in the manager-owned `ReservationSet`. If the guard were dropped for the signer round-trip, another wallet-manager operation could interleave between reserve and sign — the funding/signing selection race the old `setFunding` + `buildSigned` split had, which the atomic finalize was introduced to eliminate (see `ManagedPlatformWallet.kt`, "closing the funding/signing selection race").

Splitting the hold so signing happens lock-free is not a clean change: it means teaching `build_signed_reserved` to reserve under the lock, drop it, sign, then re-acquire to register — reintroducing a coherence window and a substantial refactor of a primitive whose whole contract is atomicity. The sibling deferred-token path (#4256) makes the same choice for the same reason.

The real cost is latency: a keystore/biometric prompt serializes other wallet-manager work for its duration. That is an accepted tradeoff here, not an oversight — a single wallet's operations are already logically serial, and payment finalization is user-initiated and foreground. Leaving as-is by design.

packages/rs-unified-sdk-jni/src/wallet_manager.rs
bfoss765: Fixed in `8c906d85ff` (and the build-export half in `684f06f41b`).

The build-signed-payment JNI export's `byte_array_from_slice` failure path now releases the reservation before returning null: it calls `core_wallet_release_payment_reservation` with the tx bytes (`packed[24..]`), the in-scope `funding_path_ptr`/`funding_path_len`, and `out_reservation_handle`. The finalize export's null-txid arm additionally releases its minted token (`core_wallet_signed_payment_release(token)`) — the second post-commit path the review flagged. Both now match the sibling register-blob Err arm, so a JVM-side allocation failure can no longer strand reserved inputs to the TTL backstop.

packages/rs-platform-wallet-ffi/src/core_wallet/signed_payment.rs
```diff
@@ -0,0 +1,145 @@
...
+    core_handle: Handle,
+    token: u64,
+    out_txid: *mut *mut c_char,
+) -> PlatformWalletFFIResult {
+    check_ptr!(out_txid);
+
+    let core = unwrap_option_or_return!(CORE_WALLET_STORAGE.with_item(core_handle, |w| w.clone()));
+
+    let result =
+        runtime().block_on(SIGNED_PAYMENT_REGISTRY.broadcast(ReservationToken::from(token), &core));
```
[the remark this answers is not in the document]
bfoss765: Fixed in `684f06f41b`. `*out_txid` is now null-initialized at function entry (`signed_payment.rs:87`), before any fallible work, so all five error arms leave a defined pointer instead of whatever the caller allocated.

packages/rs-platform-wallet-ffi/src/core_wallet/transaction_builder.rs
```diff
@@ -134,6 +180,185 @@ pub unsafe extern "C" fn core_wallet_tx_builder_finalize(
...
+) -> PlatformWalletFFIResult {
+    check_ptr!(builder);
+    check_ptr!(core_signer_handle);
+    check_ptr!(out_token);
+    check_ptr!(out_fee);
+    check_ptr!(out_txid);
+    check_ptr!(out_tx);
+    check_ptr!(out_bytes_ptr);
+    check_ptr!(out_bytes_len);
+    *out_token = 0;
```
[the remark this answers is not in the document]
bfoss765: Fixed in `684f06f41b`. All six out-params are now zero-initialized at entry (`transaction_builder.rs:240-249`): `out_token`, `out_fee`, `out_txid` (null), `out_tx` (an `FFICoreTransaction` with null `tx_bytes`), `out_bytes_ptr` (null), and `out_bytes_len`. A caller that ignores the result code after an error return now reads a null transaction rather than a dangling `out_tx.tx_bytes`/`out_bytes_ptr`.

packages/rs-platform-wallet-ffi/src/error.rs
```diff
@@ -172,8 +172,99 @@ pub enum PlatformWalletFFIResultCode {
...
+    //   28  (free — vacated by this PR)
+    //   29  ErrorAssetLockInsufficientFunds dashpay/platform#4184
+    //   30  (free — vacated by this PR)
+    //   31  ErrorSigningKeyUnavailable      dashpay/platform#4183, #4259
+    //   32  ErrorTransactionBuild           dashpay/platform#4247, #4256
+    //   33  ErrorTransactionSigning         dashpay/platform#4256
+    //
+    // This trio previously sat at 26-28, then 27/28/30. It moved to 34-36 after
+    // #4268 merged `ErrorShutdownIncomplete = 27` into the v4.2-dev ABI; the
+    // contiguous block above every current claim ends the renumbering churn.
```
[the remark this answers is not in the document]
bfoss765: Fixed in `684f06f41b`. The two doc comments that still placed the deferred-token trio in the 27–31 window were corrected to defer to the authoritative in-file allocation block (and `ERROR_CODE_REGISTRY.md`, #4261): the trio sits at 34–36, `ErrorTransactionBuild` at 32, `ErrorTransactionSigning` at 33, with 28/30 vacated-but-reserved. The `transaction_build_code_is_thirty_two` test doc was updated to match.

packages/rs-platform-wallet/src/wallet/core/send.rs
```diff
@@ -0,0 +1,1756 @@
...
+        // funding account then contributed enough small inputs to push the
+        // whole transaction over the limit.
+        let signed_size = transaction.size();
+        if signed_size > MAX_STANDARD_TX_SIZE {
+            return Err(PlatformWalletError::TransactionBuild(format!(
+                "the signed transaction is {signed_size} bytes, over the \
+                 {MAX_STANDARD_TX_SIZE}-byte standard transaction limit; it would not relay. \
+                 Send a smaller amount (fewer inputs) or fewer recipients"
+            )));
+        }
```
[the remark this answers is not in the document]
bfoss765: Fixed in `3505571203` (owner-guarded release) — the post-build size rejection no longer strands the reservation. When the signed transaction exceeds `MAX_STANDARD_TX_SIZE`, the path now releases the inputs it reserved before returning the error (`send.rs:586`+), owner-guarded so it frees only its own reservation and cannot touch a concurrent build's. The comment there records why the TTL backstop is not a fallback (`ReservationSet::sweep` early-returns at height 0 pre-first-sync).

packages/rs-platform-wallet/src/wallet/core/send.rs
```diff
@@ -0,0 +1,1756 @@
...
+    /// the one release path that works pre-sync.
+    ///
+    /// ## What is released — only this build's own inputs
+    ///
+    /// The transaction *is* the ownership signal. `build_signed_payment`
+    /// reserves exactly the outpoints it selected, and a reserved outpoint is
+    /// skipped by every subsequent coin selection — so no concurrent build can
+    /// hold a reservation on any input of `transaction`. Releasing precisely
+    /// `transaction`'s inputs therefore releases precisely this build's own
+    /// reservation and can never free a competing in-flight build's coins.
```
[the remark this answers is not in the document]
bfoss765: Addressed. The reservation is now owner-guarded on release, not by-outpoint.

`build_signed_payment` returns the `ReservationToken` (surfaced as the `out_reservation_handle` in the packed result and registered in `PAYMENT_RESERVATIONS`). `core_wallet_release_payment_reservation` resolves that handle back to the token and passes it into `wallet.release_payment_reservation(&transaction, funding_path, reservation_token)`, which releases owner-guarded — freeing only inputs still owned by that token. A reservation that key-wallet's TTL swept and a concurrent build re-took is therefore left untouched, closing the unbounded-abandonment window the finding described. Same owner-guard the `SignedCoreTransaction` handle path uses.

packages/rs-unified-sdk-jni/src/wallet_manager.rs
```diff
@@ -1013,6 +1013,195 @@ pub extern "system" fn Java_org_dashfoundation_dashsdk_ffi_WalletManagerNative_c
...
+    })
```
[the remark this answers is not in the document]
bfoss765: Fixed across both paths.

- Build-signed-payment export (`wallet_manager.rs`, the `byte_array_from_slice` failure arm): `684f06f41b` releases via `core_wallet_release_payment_reservation` with the tx bytes (`packed[24..]` — the packed prefix is `fee|change|handle`, 24 BE bytes), the in-scope `funding_path_ptr`/`funding_path_len`, and `out_reservation_handle`, then returns null.
- Finalize export null-txid arm: `8c906d85ff` releases the minted token via `core_wallet_signed_payment_release(token)` alongside `core_wallet_transaction_free` before throwing.

Both now mirror the sibling register-blob Err arm, so neither post-commit JVM failure leaves inputs reserved to the TTL backstop.

bfoss765:
Moved to #4310 - in-repo branch rebased onto v4.2-dev (post-#4305). Review history preserved here.

PR #4204, opened 2026-07-22
feat(kotlin-sdk): one-time Orchard key shielded-invite API (inviter + claim)

bfoss765:
## What

Adds the one-time Orchard key **shielded-invite** API to the Kotlin SDK. Client-side only — **no L2 protocol / consensus changes** (nothing under `rs-dpp`, `rs-drive`, `dapi`).

- **Inviter side** — `generateOneTimeOrchardKey()` / `orchardAddressFromSpendingKey()` + the `OneTimeOrchardKey` type: generate a one-time Orchard spending key and the raw address the inviter funds a note to.
- **Claim side** — `shieldedIdentityCreateFromOneTimeKey(...)`: a claimer, handed the one-time spending key, spends the funded note to create/top-up a shielded identity.

Backing Rust: `rs-platform-wallet` (`shielded/keys.rs`, `operations.rs`, `sync.rs`, `platform_wallet.rs`), `rs-platform-wallet-ffi` (`shielded_send.rs`), `rs-unified-sdk-jni` (`funding.rs`).

## ⚠️ Stacked on #4183

The claim side consumes `decode_registration_pubkeys_blob` + `IdentityPubkeyCodec`, both introduced by #4183. **This branch is stacked on #4183**, so until #4183 merges the diff below also contains #4183's changes. It will retarget to a clean diff once #4183 lands. **Net-new files to review here:**

- `packages/rs-platform-wallet/src/wallet/shielded/keys.rs`
- `packages/rs-platform-wallet/src/wallet/shielded/operations.rs`
- `packages/rs-platform-wallet/src/wallet/shielded/sync.rs`
- `packages/rs-platform-wallet/src/wallet/shielded/mod.rs`
- `packages/rs-platform-wallet/src/wallet/platform_wallet.rs`
- `packages/rs-platform-wallet-ffi/src/shielded_send.rs`
- `packages/rs-unified-sdk-jni/src/funding.rs`
- `packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/ffi/FundingNative.kt`
- `packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/PlatformWalletManager.kt`
- `packages/rs-platform-wallet/Cargo.toml` (optional `rand` dep for the `shielded` feature)

## Security note — identity key roles

[...]

PR #4251, opened 2026-07-30
feat(kotlin-sdk): tx-label & asset-lock-kind DAO resolver queries

bfoss765:
Adds the two wallet-side transaction-display persistence resolver queries the Android cutover needs:

- `TransactionDao.transactionKindForDisplayTxid` — tx-label (Shielded / Invitation / Unshielded) resolution
- `AssetLockDao.fundingTypeForTxid` — asset-lock funding-type (identity registration / top-up / invitation) resolution

Read-only queries; reuse the existing DashDatabase v7 schema (no migration; this branch does not bump the schema version). Extracted from the QA integration line into its own PR per the feature-branch organization.

Note: the two shielded anchored-note queries the app also uses — `ShieldedDao.getUnspentAnchoredNotesByWallet` and `minUnspentAnchoredBlockHeight` — are **not** in this PR; they ship with the shielded work in #4204 (they are shielded-note queries and belong there). Together, #4204 + this PR provide the four DAO symbols the app requires to compile against the SDK.

## Summary by CodeRabbit

* **New Features**
  * Added transaction lookup by both raw and explorer-formatted transaction IDs.
  * Added asset lock funding type lookup using a transaction ID.
  * Transaction IDs are validated and converted automatically between supported formats.
  * Lookups safely return no result when the ID is invalid or no matching record exists.

### DIFF
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/AssetLockDao.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/AssetLockDao.kt
@@ -77,16 +77,28 @@ interface AssetLockDao {
     /**
      * Transaction-label resolver probe: the `fundingTypeRaw` of the asset
      * lock whose outpoint belongs to [txidHex]. [txidHex] is the explorer
-     * DISPLAY txid hex (64 lowercase chars, wire order reversed) — the
-     * prefix of the `outPointHex` PK (`<txidDisplayHex>:<vout>`), matched
-     * via `LIKE '<txidHex>:%'`. A txid is a pure-hex string (no `%`/`_`),
-     * so the LIKE pattern carries no wildcards of its own. Returns null
-     * when no asset lock funds this tx (e.g. plain send, or an
-     * AssetUnlock/unshield — see [TransactionDao.transactionKindForTxid]).
+     * DISPLAY txid hex (64 chars, wire order reversed; uppercase input is
+     * canonicalized via `lower()`) — the prefix of the `outPointHex` PK
+     * (`<txidDisplayHex>:<vout>`). The query enforces the 64-hex input
+     * contract itself (length + hex-only GLOB, matching the
+     * null-on-malformed behavior of
+     * [TransactionDao.transactionKindForDisplayTxid]) and compares the
+     * exact 65-char `<txid>:` prefix rather than a LIKE pattern, so
+     * `%`/`_` in malformed input can never match arbitrary rows. Returns
+     * null for malformed input or when no asset lock funds this tx (e.g.
+     * plain send, or an AssetUnlock/unshield — see
+     * [TransactionDao.transactionKindForTxid]).
+     *
+     * DIP-0027 permits several asset-lock outputs in one transaction;
+     * those rows are created by the same funding flow and share a
+     * `fundingTypeRaw`, so the unordered `LIMIT 1` pick is value-stable.
      */
     @Query(
         "SELECT fundingTypeRaw FROM asset_locks " +
-            "WHERE outPointHex LIKE :txidHex || ':%' LIMIT 1"
+            "WHERE length(:txidHex) = 64 " +
+            "AND lower(:txidHex) NOT GLOB '*[^0-9a-f]*' " +
+            "AND substr(outPointHex, 1, 65) = lower(:txidHex) || ':' " +
+            "LIMIT 1"
     )
     suspend fun fundingTypeForTxid(txidHex: String): Int?
 
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/AssetLockDao.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/AssetLockDao.kt
@@ -74,6 +74,22 @@ interface AssetLockDao {
     @Query("SELECT * FROM asset_locks WHERE outPointHex = :outPointHex")
     suspend fun getByOutPointHex(outPointHex: String): AssetLockEntity?
 
+    /**
+     * Transaction-label resolver probe: the `fundingTypeRaw` of the asset
+     * lock whose outpoint belongs to [txidHex]. [txidHex] is the explorer
+     * DISPLAY txid hex (64 lowercase chars, wire order reversed) — the
+     * prefix of the `outPointHex` PK (`<txidDisplayHex>:<vout>`), matched
+     * via `LIKE '<txidHex>:%'`. A txid is a pure-hex string (no `%`/`_`),
+     * so the LIKE pattern carries no wildcards of its own. Returns null
+     * when no asset lock funds this tx (e.g. plain send, or an
+     * AssetUnlock/unshield — see [TransactionDao.transactionKindForTxid]).
+     */
+    @Query(
+        "SELECT fundingTypeRaw FROM asset_locks " +
+            "WHERE outPointHex LIKE :txidHex || ':%' LIMIT 1"
+    )
+    suspend fun fundingTypeForTxid(txidHex: String): Int?
+
     @Upsert
     suspend fun upsert(assetLock: AssetLockEntity)
 
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/TransactionDao.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/TransactionDao.kt
@@ -27,6 +27,35 @@ interface TransactionDao {
     @Query("SELECT * FROM transactions WHERE txid = :txid")
     suspend fun getByTxid(txid: ByteArray): TransactionEntity?
 
+    /**
+     * Transaction-label resolver probe for the withdraw/unshield case,
+     * which has NO `asset_locks` row and is identified solely by
+     * `transactionTypeKind == 7` (AssetUnlock).
+     *
+     * [txidWire] is the raw little-endian WIRE txid — the exact
+     * [TransactionEntity.txid] PK form (a BLOB), i.e. the explorer display
+     * hex reversed then hex-decoded. When the caller holds the display hex
+     * (as the resolver does), prefer [transactionKindForDisplayTxid], which
+     * does the reversal. Returns null when no such tx is stored;
+     * `transactionTypeKind == 0xFF` (255) means "not yet populated".
+     */
+    @Query("SELECT transactionTypeKind FROM transactions WHERE txid = :txidWire LIMIT 1")
+    suspend fun transactionKindForTxid(txidWire: ByteArray): Int?
+
+    /**
+     * Convenience over [transactionKindForTxid] keyed by the explorer
+     * DISPLAY txid hex (64 lowercase chars, wire order reversed) the
+     * resolver already holds — the same display form used by
+     * [AssetLockDao.fundingTypeForTxid]. Reverses to wire order before the
+     * BLOB PK match. Returns null for malformed hex (not 64 hex chars) or
+     * when no such tx is stored. Not a Room query — a plain default method
+     * delegating to [transactionKindForTxid].
+     */
+    suspend fun transactionKindForDisplayTxid(txidDisplayHex: String): Int? {
+        val wire = displayHexToWireTxid(txidDisplayHex) ?: return null
+        return transactionKindForTxid(wire)
+    }
+
     /** Timeline join helper — parents of a wallet's TXO set. */
     @Query("SELECT * FROM transactions WHERE txid IN (:txids) ORDER BY firstSeen DESC")
     fun observeByTxids(txids: List<ByteArray>): Flow<List<TransactionEntity>>
@@ -89,4 +118,26 @@ interface TransactionDao {
     /** StorageExplorer row count. */
     @Query("SELECT COUNT(*) FROM transactions")
     fun count(): Flow<Long>
+
+    companion object {
+        /**
+         * Explorer DISPLAY txid hex (wire order reversed, 64 lowercase
+         * chars) → raw little-endian WIRE txid, the [TransactionEntity.txid]
+         * PK form. Mirrors the txid half of [decodeOutPointHex]. Returns
+         * null for any non-64-char or non-hex input.
+         */
+        private fun displayHexToWireTxid(displayHex: String): ByteArray? {
+            if (displayHex.length != 64) return null
+            val display = ByteArray(32)
+            for (i in 0 until 32) {
+                val hi = Character.digit(displayHex[i * 2], 16)
+                val lo = Character.digit(displayHex[i * 2 + 1], 16)
+                if (hi < 0 || lo < 0) return null
+                display[i] = ((hi shl 4) or lo).toByte()
+            }
+            val wire = ByteArray(32)
+            for (i in 0 until 32) wire[i] = display[31 - i]
+            return wire
+        }
+    }
 }
--- a/packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/persistence/TransactionLabelResolverDaoTest.kt
+++ b/packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/persistence/TransactionLabelResolverDaoTest.kt
@@ -1,0 +1,183 @@
+package org.dashfoundation.dashsdk.persistence
+
+import androidx.test.core.app.ApplicationProvider
+import kotlinx.coroutines.test.runTest
+import org.dashfoundation.dashsdk.persistence.entities.AssetLockEntity
+import org.dashfoundation.dashsdk.persistence.entities.TransactionEntity
+import org.junit.After
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNull
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+/**
+ * In-memory Room contract tests for the three transaction-label resolver
+ * queries: [org.dashfoundation.dashsdk.persistence.dao.TransactionDao]'s
+ * `transactionKindForTxid` / `transactionKindForDisplayTxid` and
+ * [org.dashfoundation.dashsdk.persistence.dao.AssetLockDao]'s
+ * `fundingTypeForTxid`.
+ *
+ * The fixture txid is deliberately NON-palindromic (`0x01..0x20`), so the
+ * display↔wire byte reversal in `displayHexToWireTxid` is pinned: deleting
+ * (or double-applying) the reversal loop flips these assertions, which a
+ * symmetric fixture like `"ab" * 32` would not catch.
+ */
+@RunWith(RobolectricTestRunner::class)
+class TransactionLabelResolverDaoTest {
+
+    private lateinit var db: DashDatabase
+
+    private val walletId = ByteArray(32) { 1 }
+
+    /** Raw little-endian WIRE txid — the `transactions.txid` PK form. */
+    private val wireTxid = ByteArray(32) { (it + 1).toByte() }
+
+    /** Explorer DISPLAY hex: wire bytes reversed, lowercase. */
+    private val displayHex = hex(wireTxid.reversedArray())
+
+    /** The unreversed hex of the wire bytes — display form of a DIFFERENT tx. */
+    private val wireHexUnreversed = hex(wireTxid)
+
+    /** `TransactionType` discriminant for AssetUnlock (withdraw/unshield). */
+    private val kindAssetUnlock = 7
+
+    /** `AssetLockFundingType` discriminant for AssetLockAddressTopUp. */
+    private val fundingTypeAddressTopUp = 4
+
+    @Before
+    fun setUp() {
+        db = DashDatabase.createInMemory(ApplicationProvider.getApplicationContext())
+    }
+
+    @After
+    fun tearDown() {
+        db.close()
+    }
+
+    private fun hex(bytes: ByteArray): String =
+        bytes.joinToString("") { "%02x".format(it) }
+
+    private suspend fun insertTransaction() {
+        db.transactionDao().upsert(
+            TransactionEntity(
+                txid = wireTxid,
+                transactionData = ByteArray(4),
+                transactionTypeKind = kindAssetUnlock,
+            ),
+        )
+    }
+
+    private suspend fun insertAssetLock(outPointHex: String, fundingTypeRaw: Int) {
+        db.assetLockDao().upsert(
+            AssetLockEntity(
+                outPointHex = outPointHex,
+                walletId = walletId,
+                transactionBytes = ByteArray(4),
+                fundingTypeRaw = fundingTypeRaw,
+                identityIndexRaw = 0,
+                amountDuffs = 10_000,
+                statusRaw = 1,
+            ),
+        )
+    }
+
+    // ── transactionKindForTxid (wire BLOB PK) ─────────────────────────
+
+    @Test
+    fun transactionKindForTxidMatchesWireBytes() = runTest {
+        insertTransaction()
+        assertEquals(kindAssetUnlock, db.transactionDao().transactionKindForTxid(wireTxid))
+    }
+
+    @Test
+    fun transactionKindForTxidReturnsNullForUnknownTxid() = runTest {
+        insertTransaction()
+        assertNull(db.transactionDao().transactionKindForTxid(ByteArray(32) { 9 }))
+    }
+
+    // ── transactionKindForDisplayTxid (display hex → wire reversal) ───
+
+    @Test
+    fun displayTxidLookupReversesToWireOrder() = runTest {
+        insertTransaction()
+        // Display hex resolves the row — pins the byte reversal on a
+        // non-palindromic txid.
+        assertEquals(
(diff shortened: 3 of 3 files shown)


thepastaclaw:
✅ Final review complete — no blockers (commit 176f8ed3)

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/AssetLockDao.kt
```diff
@@ -74,6 +74,22 @@ interface AssetLockDao {
...
+     * via `LIKE '<txidHex>:%'`. A txid is a pure-hex string (no `%`/`_`),
+     * so the LIKE pattern carries no wildcards of its own. Returns null
+     * when no asset lock funds this tx (e.g. plain send, or an
+     * AssetUnlock/unshield — see [TransactionDao.transactionKindForTxid]).
+     */
+    @Query(
+        "SELECT fundingTypeRaw FROM asset_locks " +
+            "WHERE outPointHex LIKE :txidHex || ':%' LIMIT 1"
+    )
+    suspend fun fundingTypeForTxid(txidHex: String): Int?
```
thepastaclaw, line 103: **🟡 Suggestion: Validate txids before using them as a LIKE pattern**

Unlike `transactionKindForDisplayTxid`, this resolver does not enforce its documented 64-character hexadecimal input contract. SQLite interprets `%` and `_` in the bound value as wildcards, so malformed inputs such as `%` or 64 underscores match unrelated asset-lock rows and `LIMIT 1` returns one row's funding type rather than null. Validate and canonicalize the input in SQL, then compare the exact 64-character outpoint prefix; generated `outPointHex` values are lowercase, while `lower(:txidHex)` preserves support for uppercase canonical txids.

```suggestion
@Query(
    "SELECT fundingTypeRaw FROM asset_locks " +
        "WHERE length(:txidHex) = 64 " +
        "AND lower(:txidHex) NOT GLOB '*[^0-9a-f]*' " +
        "AND substr(outPointHex, 1, 65) = lower(:txidHex) || ':' " +
        "LIMIT 1"
)
suspend fun fundingTypeForTxid(txidHex: String): Int?
```

<sub>source: ['codex']</sub>

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/TransactionDao.kt
```diff
@@ -27,6 +27,35 @@ interface TransactionDao {
...
+     * DISPLAY txid hex (64 lowercase chars, wire order reversed) the
+     * resolver already holds — the same display form used by
+     * [AssetLockDao.fundingTypeForTxid]. Reverses to wire order before the
+     * BLOB PK match. Returns null for malformed hex (not 64 hex chars) or
+     * when no such tx is stored. Not a Room query — a plain default method
+     * delegating to [transactionKindForTxid].
+     */
+    suspend fun transactionKindForDisplayTxid(txidDisplayHex: String): Int? {
+        val wire = displayHexToWireTxid(txidDisplayHex) ?: return null
+        return transactionKindForTxid(wire)
```
thepastaclaw, line 56: **🟡 Suggestion: Add Room tests for the new resolver contracts**

No SDK test invokes `transactionKindForTxid`, `transactionKindForDisplayTxid`, or `fundingTypeForTxid`. Add in-memory Room coverage using a non-palindromic txid to pin the display-to-wire byte reversal, plus missing, malformed, and uppercase input cases. The asset-lock tests should also cover multiple vouts sharing a txid and define the intended behavior if their funding types differ, because the current nullable scalar query uses `LIMIT 1` and otherwise leaves that selection dependent on row scan order.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The txid display-to-wire reversal matches the persisted transaction and outpoint encodings, the raw transaction/funding discriminants are consistent with their Rust definitions, and these read-only queries require no Room migration. However, the exact head omits two DAO methods explicitly promised by the PR and required by the stated Android cutover, so this PR does not yet deliver its advertised API surface.

### Review provenance

<details>

```

In `packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/ShieldedDao.kt`:
- [BLOCKING] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/ShieldedDao.kt:52-56: Add the two promised anchored-note DAO methods
  The PR description explicitly includes `getUnspentAnchoredNotesByWallet` and `minUnspentAnchoredBlockHeight` among the four symbols required for the Android cutover, but this exact head proceeds directly from `observeUnspentNotesByWallet` to `upsertNote` without defining either method. Commit `7bc8a845c67c6010e72b171a1d8f70f498220574` in still-open PR #4204 adds those methods, but it is not an ancestor of this head; the corresponding pre-commit `ShieldedDao` section is identical to this PR's base. Merging PR #4251 alone therefore does not provide the advertised API and leaves the cutover with unresolved references. Include the queries here, or explicitly declare and enforce PR #4204 as a merge prerequisite and update this PR's stated deliverables.

In `packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/AssetLockDao.kt`:
- [SUGGESTION] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/AssetLockDao.kt:87-91: Validate txids before using them as a LIKE pattern
  Unlike `transactionKindForDisplayTxid`, this resolver does not enforce its documented 64-character hexadecimal input contract. SQLite interprets `%` and `_` in the bound value as wildcards, so malformed inputs such as `%` or 64 underscores match unrelated asset-lock rows and `LIMIT 1` returns one row's funding type rather than null. Validate and canonicalize the input in SQL, then compare the exact 64-character outpoint prefix; generated `outPointHex` values are lowercase, while `lower(:txidHex)` preserves support for uppercase canonical txids.

In `packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/TransactionDao.kt`:
- [SUGGESTION] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/TransactionDao.kt:42-56: Add Room tests for the new resolver contracts
  No SDK test invokes `transactionKindForTxid`, `transactionKindForDisplayTxid`, or `fundingTypeForTxid`. Add in-memory Room coverage using a non-palindromic txid to pin the display-to-wire byte reversal, plus missing, malformed, and uppercase input cases. The asset-lock tests should also cover multiple vouts sharing a txid and define the intended behavior if their funding types differ, because the current nullable scalar query uses `LIMIT 1` and otherwise leaves that selection dependent on row scan order.
```
</details>

shumkov:
Confirmed no migration is needed — both columns pre-exist and Room's compile-time validation accepted the queries, so that claim is verified rather than assumed.

**1. `fundingTypeForTxid` doesn't enforce its own documented contract.** `AssetLockDao.kt:87-91` binds `:txidHex` as the **LIKE pattern**, so `fundingTypeForTxid("%")` returns some arbitrary asset lock's funding type and 64 `_` chars matches every row — a silently *wrong* label, not injection. The KDoc's "a txid is pure hex so the pattern carries no wildcards" is a statement about callers, and there are no callers yet to hold to it. The asymmetry is the bug: the sibling `transactionKindForDisplayTxid` (`TransactionDao.kt:54-57`) *does* validate and return null. Cheapest fix is the same 64-hex guard `displayHexToWireTxid` already applies.

**2. Zero tests for all three new methods.** Delete the byte-reversal loop at `TransactionDao.kt:139` and nothing in CI turns red — every withdraw/unshield tx would silently lose its "Asset Unlock" label on device. The harness already exists (`DashDatabase.createInMemory` under Robolectric, as used by `WalletDeletionTest`). A non-palindromic txid fixture is the important case; `"ab".repeat(32)` would pass even with the reversal removed.

**3. The description says DashDatabase v9; this branch is at v7** (`DashDatabase.kt:104`, schemas stop at `7.json`). The "no migration needed" conclusion is still correct, but the version claim will send reviewers looking for a schema that isn't there — and #4240 concurrently adds `8.json`.

Minor: `LIMIT 1` with no `ORDER BY` over a one-to-many (DIP-0027 allows up to 255 credit outputs, so the row returned is scan-order dependent — worth documenting that all vouts share a funding type, or making it deterministic); this is the only query in `AssetLockDao` not scoped by `walletId`; and `TransactionDao.kt:129-141` re-implements the txid half of `decodeOutPointHex`, which is `internal` in the same module and reachable.

Process note: the 🔴 blocker on this PR is filed against `ShieldedDao.kt`, which this PR doesn't touch, and the current PR body already states those queries ship with #4204. Because it was treated as blocking, the second review stage never ran — so this diff has had only partial automated review.

thepastaclaw:
You're right about the preliminary blocker. I revalidated the exact head (`715aad23`) against the full PR description and #4204: the two `ShieldedDao` methods are explicitly out of scope here and already assigned to #4204, so they should not block this PR.

I've queued a same-SHA automated re-review with this thread context so the deferred second stage can run and supersede the incorrect `CHANGES_REQUESTED` review. The remaining two inline items are still valid author-facing suggestions: validate `fundingTypeForTxid` inputs before matching, and add focused Room coverage for the endian conversion/resolver contracts.

GitHub would not allow this account to directly dismiss the existing review, so the fresh review must supersede it.

@coderabbitai review

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/AssetLockDao.kt
thepastaclaw, line 103: **🟡 Suggestion: fundingTypeForTxid doesn't enforce its own hex-only contract before using input as a LIKE pattern**

fundingTypeForTxid binds txidHex directly into `outPointHex LIKE :txidHex || ':%'` without validating it is a 64-char pure-hex string, unlike the sibling transactionKindForDisplayTxid which enforces exactly that via displayHexToWireTxid and returns null on malformed input. SQLite treats `%`/`_` in the bound value as wildcards, so `fundingTypeForTxid("%")` matches an arbitrary asset-lock row and 64 `_` chars matches every row — with `LIMIT 1` this silently returns a wrong funding type instead of null. The in-code comment's claim that "a txid is a pure-hex string ... so the LIKE pattern carries no wildcards of its own" is an assumption about callers, not an invariant this public method enforces. Separately, even once the pattern is exact-matched, `LIMIT 1` with no `ORDER BY` over a genuinely one-to-many relationship (DIP-0027 permits multiple asset-lock outputs sharing a txid) makes the returned row scan-order dependent; if all vouts for a txid always share the same funding type this is harmless, but that invariant should be documented or the query made deterministic.

```suggestion
@Query(
    "SELECT fundingTypeRaw FROM asset_locks " +
        "WHERE length(:txidHex) = 64 " +
        "AND lower(:txidHex) NOT GLOB '*[^0-9a-f]*' " +
        "AND substr(outPointHex, 1, 65) = lower(:txidHex) || ':' " +
        "LIMIT 1"
)
suspend fun fundingTypeForTxid(txidHex: String): Int?
```

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/TransactionDao.kt
thepastaclaw, line 56: **🟡 Suggestion: No test coverage for the three new resolver methods**

Nothing in the repo invokes transactionKindForTxid, transactionKindForDisplayTxid, or fundingTypeForTxid. The hand-written byte-reversal in displayHexToWireTxid (TransactionDao.kt:129-141) has no pinned regression test — deleting the reversal loop entirely would not fail CI, silently mislabeling every withdraw/unshield transaction on-device. The in-memory Room harness already used by WalletDeletionTest is available for this. Add cases with a non-palindromic txid (to catch endian regressions a symmetric fixture like "ab".repeat(32) would miss), uppercase input, malformed/short hex, no-match, and — for fundingTypeForTxid — multiple asset-lock rows sharing a txid prefix to pin down the LIMIT 1 selection behavior.

thepastaclaw (commented):

This PR adds two read-only Room resolver queries (TransactionDao.transactionKindForDisplayTxid/transactionKindForTxid and AssetLockDao.fundingTypeForTxid) reusing existing DashDatabase columns with no migration; the display-to-wire txid reversal is correct and Room's compile-time validation accepted the queries against the actual schema. The prior blocking ShieldedDao finding is confirmed misfiled and is dropped: the PR body explicitly scopes those two anchored-note methods to #4204, and the diff never touches ShieldedDao.kt. Two suggestion-level issues persist on the exact head — fundingTypeForTxid doesn't enforce its own documented hex-only contract before using the input as a LIKE pattern, and none of the three new resolver methods have test coverage — plus a minor doc nitpick that the PR description cites schema v9 while DashDatabase.kt declares version 7. Source: reviewers `gpt-5.6-sol` and `claude-sonnet-5` (general); verifier `claude-sonnet-5`.

### Review provenance
- Sonnet reviewers: `claude-sonnet-5` — general (completed)

<details>

```

In `packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/AssetLockDao.kt`:
- [SUGGESTION] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/AssetLockDao.kt:87-91: fundingTypeForTxid doesn't enforce its own hex-only contract before using input as a LIKE pattern
  fundingTypeForTxid binds txidHex directly into `outPointHex LIKE :txidHex || ':%'` without validating it is a 64-char pure-hex string, unlike the sibling transactionKindForDisplayTxid which enforces exactly that via displayHexToWireTxid and returns null on malformed input. SQLite treats `%`/`_` in the bound value as wildcards, so `fundingTypeForTxid("%")` matches an arbitrary asset-lock row and 64 `_` chars matches every row — with `LIMIT 1` this silently returns a wrong funding type instead of null. The in-code comment's claim that "a txid is a pure-hex string ... so the LIKE pattern carries no wildcards of its own" is an assumption about callers, not an invariant this public method enforces. Separately, even once the pattern is exact-matched, `LIMIT 1` with no `ORDER BY` over a genuinely one-to-many relationship (DIP-0027 permits multiple asset-lock outputs sharing a txid) makes the returned row scan-order dependent; if all vouts for a txid always share the same funding type this is harmless, but that invariant should be documented or the query made deterministic.

In `packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/TransactionDao.kt`:
- [SUGGESTION] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/TransactionDao.kt:42-56: No test coverage for the three new resolver methods
  Nothing in the repo invokes transactionKindForTxid, transactionKindForDisplayTxid, or fundingTypeForTxid. The hand-written byte-reversal in displayHexToWireTxid (TransactionDao.kt:129-141) has no pinned regression test — deleting the reversal loop entirely would not fail CI, silently mislabeling every withdraw/unshield transaction on-device. The in-memory Room harness already used by WalletDeletionTest is available for this. Add cases with a non-palindromic txid (to catch endian regressions a symmetric fixture like "ab".repeat(32) would miss), uppercase input, malformed/short hex, no-match, and — for fundingTypeForTxid — multiple asset-lock rows sharing a txid prefix to pin down the LIMIT 1 selection behavior.
```
</details>

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/TransactionDao.kt
thepastaclaw: Resolved in this update — _Add Room tests for the new resolver contracts_ no longer present.

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/AssetLockDao.kt
thepastaclaw: Resolved in this update — _Validate txids before using them as a LIKE pattern_ no longer present.

bfoss765: Fixed in 176f8ed3eb — adopted the suggested query shape: the SQL now enforces the 64-hex contract itself (`length(:txidHex) = 64` + hex-only GLOB, `lower()`-canonicalized) and compares the exact 65-char `<txid>:` prefix via `substr`, so there is no LIKE pattern left for `%`/`_` to exploit — `fundingTypeForTxid("%")` and 64 underscores now return null instead of an arbitrary row's funding type. The multi-vout `LIMIT 1` point is addressed in the KDoc: DIP-0027 rows for one txid come from the same funding flow and share `fundingTypeRaw`, so the unordered pick is value-stable (and the new test pins that with two vouts sharing a txid).

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/TransactionDao.kt
bfoss765: Added in 176f8ed3eb: `TransactionLabelResolverDaoTest` (in-memory Room via the same Robolectric harness as `WalletDeletionTest`) covers all three resolvers with a NON-palindromic txid (`0x01..0x20`), pinning the display→wire byte reversal — the display hex must resolve while the unreversed wire hex must miss, so deleting the reversal loop in `displayHexToWireTxid` now fails CI. Also covered: uppercase input, malformed/short/empty hex, no-match, wildcard inputs (`%`, 64 underscores) against `fundingTypeForTxid`, and multiple asset-lock vouts sharing a txid to pin the `LIMIT 1` value-stability.

bfoss765:
Following up on the retracted preliminary blocker: confirmed the `ShieldedDao` methods remain out of scope here (they belong to #4204), and the two remaining author-facing suggestions are now addressed in 176f8ed3eb:

1. `AssetLockDao.fundingTypeForTxid` enforces its 64-hex contract in-query (length + hex-only GLOB + exact `substr` prefix match — no LIKE), returning null for wildcard or malformed input.
2. `TransactionLabelResolverDaoTest` adds in-memory Room coverage for all three resolver queries with a non-palindromic txid pinning the display→wire byte reversal, plus uppercase/malformed/wildcard/no-match/multi-vout cases. Verified green via `:sdk:testDebugUnitTest`.

The PR description's stale "DashDatabase v9" claim is also corrected — this branch is on schema v7 with no version bump.

Since the stale CHANGES_REQUESTED review could not be dismissed directly, requesting the queued re-review run against this head and supersede it.

@coderabbitai review

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/TransactionDao.kt
```diff
@@ -89,4 +118,26 @@ interface TransactionDao {
...
+            for (i in 0 until 32) {
+                val hi = Character.digit(displayHex[i * 2], 16)
+                val lo = Character.digit(displayHex[i * 2 + 1], 16)
+                if (hi < 0 || lo < 0) return null
+                display[i] = ((hi shl 4) or lo).toByte()
+            }
+            val wire = ByteArray(32)
+            for (i in 0 until 32) wire[i] = display[31 - i]
+            return wire
+        }
```
thepastaclaw, line 141: **🟡 Suggestion: displayHexToWireTxid accepts non-ASCII Unicode hex digits via Character.digit**

Java's `Character.digit(char, 16)` is documented to accept Unicode decimal digits (any script where `isDigit()` is true and the decimal value fits the radix, e.g. Arabic-Indic `١`) and fullwidth Latin letters (`Ａ`-`Ｆ`, `ａ`-`ｆ`) in addition to ASCII `0-9a-fA-F`. A 64-character string built from these non-ASCII equivalents therefore decodes successfully instead of returning null, contradicting the method's documented contract ("Returns null for malformed hex (not 64 hex chars)") and diverging from `AssetLockDao.fundingTypeForTxid`'s new ASCII-only `NOT GLOB '*[^0-9a-f]*'` guard. Practical exposure is limited today since no caller yet exists in the tree, but this is new SDK-public surface. Note this exact `Character.digit` pattern is intentionally mirrored from the pre-existing `decodeOutPointHex` in `PlatformWalletPersistenceHandler.kt` (which the doc comment explicitly cites), so patching only this copy would leave an inconsistent sibling; the two hex-parsing loops (already flagged separately as duplicated) should be consolidated into one ASCII-only helper shared by both.

```suggestion
private fun displayHexToWireTxid(displayHex: String): ByteArray? {
    if (displayHex.length != 64) return null
    val display = ByteArray(32)
    for (i in 0 until 32) {
        val hi = asciiHexDigit(displayHex[i * 2]) ?: return null
        val lo = asciiHexDigit(displayHex[i * 2 + 1]) ?: return null
        display[i] = ((hi shl 4) or lo).toByte()
    }
    val wire = ByteArray(32)
    for (i in 0 until 32) wire[i] = display[31 - i]
    return wire
}

private fun asciiHexDigit(char: Char): Int? = when (char) {
    in '0'..'9' -> char - '0'
    in 'a'..'f' -> char - 'a' + 10
    in 'A'..'F' -> char - 'A' + 10
    else -> null
}
```

thepastaclaw (commented):

This cumulative review (base ed4116b26c..176f8ed3eb, latest delta 715aad2367..176f8ed3eb) confirms all three prior findings are fixed at exact head 176f8ed3eb2f70b18aa7b270e4fb5e6ab56cdd44: `AssetLockDao.fundingTypeForTxid` now enforces its 64-hex/ASCII-only contract via `length()` + `NOT GLOB '*[^0-9a-f]*'` and an exact `substr(outPointHex,1,65)` comparison instead of a raw `LIKE` pattern (verified directly against the SQL); the new `TransactionLabelResolverDaoTest.kt` exercises all three resolver methods with a genuinely non-palindromic fixture that pins the display↔wire byte reversal and covers wildcard/malformed/uppercase/multi-vout cases; and the PR description now correctly states the queries reuse `DashDatabase v7`, matching `DashDatabase.kt`'s `version = 7`. One new, delta-independent (present since 715aad2367 but newly surfaced by Codex) low-severity finding remains: `TransactionDao.displayHexToWireTxid`'s use of `Character.digit(char, 16)` accepts non-ASCII Unicode digit/letter forms (e.g. fullwidth Latin, Arabic-Indic digits) in addition to ASCII hex, which is inconsistent with the new `AssetLockDao` ASCII-only GLOB guard — this mirrors a pre-existing, intentionally-copied pattern from `decodeOutPointHex` in `PlatformWalletPersistenceHandler.kt`, so a proper fix should consolidate both parsers rather than patch only the new copy. No blocking issues found. Source: reviewers `gpt-5.6-sol` and `claude-sonnet-5` (general); verifier `claude-sonnet-5`.

### Review provenance
- Sonnet reviewers: `claude-sonnet-5` — general (completed)

🟡 1 suggestion(s)

<details>

```

In `packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/TransactionDao.kt`:
- [SUGGESTION] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/TransactionDao.kt:129-141: displayHexToWireTxid accepts non-ASCII Unicode hex digits via Character.digit
  Java's `Character.digit(char, 16)` is documented to accept Unicode decimal digits (any script where `isDigit()` is true and the decimal value fits the radix, e.g. Arabic-Indic `١`) and fullwidth Latin letters (`Ａ`-`Ｆ`, `ａ`-`ｆ`) in addition to ASCII `0-9a-fA-F`. A 64-character string built from these non-ASCII equivalents therefore decodes successfully instead of returning null, contradicting the method's documented contract ("Returns null for malformed hex (not 64 hex chars)") and diverging from `AssetLockDao.fundingTypeForTxid`'s new ASCII-only `NOT GLOB '*[^0-9a-f]*'` guard. Practical exposure is limited today since no caller yet exists in the tree, but this is new SDK-public surface. Note this exact `Character.digit` pattern is intentionally mirrored from the pre-existing `decodeOutPointHex` in `PlatformWalletPersistenceHandler.kt` (which the doc comment explicitly cites), so patching only this copy would leave an inconsistent sibling; the two hex-parsing loops (already flagged separately as duplicated) should be consolidated into one ASCII-only helper shared by both.
```
</details>

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/AssetLockDao.kt
thepastaclaw: Resolved in `176f8ed` — _fundingTypeForTxid doesn't enforce its own hex-only contract before using input as a LIKE pattern_ no longer present.

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/dao/TransactionDao.kt
thepastaclaw: Resolved in this update — _No test coverage for the three new resolver methods_ no longer present.

PR #4258, opened 2026-07-31
feat(sdk): masternodes-by-voting-key lookup (contested-username voting post-cutover)

bfoss765:
## Why

The wallet's contested-username voting feature asks "which masternode(s) can this voting key vote for?" Today that answer comes from dashj's `MasternodeListManager.getMasternodesByVotingKey(votingKeyId)`. Post-cutover the app no longer runs dashj's masternode-list manager, so **that list comes back empty and voting silently stops working**.

Nothing new has to be synced to fix this: the SPV engine already syncs the full SML, and every entry already carries both `key_id_voting` and `pro_reg_tx_hash`. This PR just surfaces the lookup that was already possible over data we already hold — Rust → FFI → JNI → Kotlin.

## What

The chain, one layer per hop:

| Layer | Symbol | File |
|---|---|---|
| SPV runtime | `SpvRuntime::masternodes_by_voting_key_blocking` | `packages/rs-platform-wallet/src/spv/runtime.rs` |
| FFI | `platform_wallet_manager_masternodes_by_voting_key` | `packages/rs-platform-wallet-ffi/src/spv.rs` |
| FFI types | `IdentifierArray::from_hashes` | `packages/rs-platform-wallet-ffi/src/types.rs` |
| JNI | `Java_..._masternodesByVotingKey` + `read_id20` | `packages/rs-unified-sdk-jni/src/wallet_manager.rs` |
| Kotlin | `external fun masternodesByVotingKey` | `packages/kotlin-sdk/.../ffi/WalletManagerNative.kt` |

The runtime accessor reads the current tip via `latest_masternode_list()` and delegates to the new `MasternodeList::masternodes_by_voting_key` helper in rust-dashcore.

### Byte contract

- **In:** the 20-byte voting key id (hash160 of a voting public key).
- **Out:** a flat array of 32-byte proTxHashes in **internal byte order** — the same `pro_reg_tx_hash.as_ref()` convention as the existing `masternode_validity_snapshot_blocking`. Kotlin receives one `ByteArray` and splits it into 32-byte rows.
- **Empty, not an error:** an unsynced or stopped SPV client, an unavailable DML, or a voting key no masternode uses all return an *empty* result rather than throwing. Callers must treat "empty" as "no answer yet", not "no masternodes".
- `IdentifierArray` is reused as the transport for raw hashes (these are proTxHashes, not platform `Identifier`s — hence the separate `from_hashes` constructor, with the same leak/free ownership contract as `new`). Freed via `platform_wallet_identifier_array_free`.
- The FFI publishes the empty sentinel into `*out_array` *before* any fallible work, so an error return can never leave the out-param holding uninitialized stack bytes.

Blocking, like the sibling `*_blocking` accessors: it takes the client + engine `tokio::RwLock`s via `blocking_read` and must run off the async runtime. Lock ordering (clone the engine `Arc` out under the client lock, drop it, then read the engine) matches `connected_peers`.

## ⚠️ rust-dashcore pin swap — please read

The helper this PR calls, `MasternodeList::masternodes_by_voting_key` (with a unit test), **is not yet in dashpay/rust-dashcore**. It currently lives only on the fork.

So the second commit here repoints all 8 workspace rust-dashcore git deps:

- from `dashpay/rust-dashcore` @ `70d4bf8e36057c58e02d56769a6e9760f701dd06` (what `v4.2-dev` pins)
- to `bfoss765/rust-dashcore` @ `4d927c155b6740ba7e1f271144217f7e82b69990`

The fork rev is a **strict fast-forward superset** of the rev `v4.2-dev` already pins — **4 commits ahead, 0 behind, no divergence**. For full transparency, those 4 commits are the voting-key helper, a `cargo fmt`, and two owner-tagged-reservation changes related to platform#4185 — i.e. the pin drags in slightly more than just the helper.

**This pin is temporary and should not merge as-is.** The intended sequence is: land `masternodes_by_voting_key` upstream in dashpay/rust-dashcore, then revert the pin commit and re-pin to an upstream rev. Happy to hold this PR until that lands, or to split the pin out — reviewer's call.

## Verification

`cargo check -p platform-wallet -p platform-wallet-ffi -p rs-unified-sdk-jni` — clean (only a pre-existing workspace `default-features` manifest warning that is also present on `v4.2-dev`).

## App wiring

The consuming side is already written and running on `dashpay/dash-wallet` @ `feat/kotlin-sdk-phase1`; this PR is the SDK half it binds against.

---

## Summary by CodeRabbit

* **New Features**
  * Added support for retrieving masternode transaction hashes associated with a voting key.
  * Exposed this capability through the Kotlin SDK and platform wallet interfaces.
  * Returns matching masternodes in a consistent internal order, or an empty result when none are available.

* **Bug Fixes**
  * Improved handling and cleanup of returned identifier arrays, including empty and multi-item results.
  * Added validation for invalid voting-key inputs and output pointers.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,14 +52,14 @@ members = [
 ]
 
 [workspace.dependencies]
-dashcore = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
-dash-network-seeds = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
-dash-spv = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
-key-wallet = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
-key-wallet-ffi = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
-key-wallet-manager = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
-dash-network = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
-dashcore-rpc = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
+dashcore = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
+dash-network-seeds = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
+dash-spv = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
+key-wallet = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
+key-wallet-ffi = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
+key-wallet-manager = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
+dash-network = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
+dashcore-rpc = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
 
 tokio-metrics = "0.5"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,14 +52,14 @@ members = [
 ]
 
 [workspace.dependencies]
-dashcore = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
-dash-network-seeds = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
-dash-spv = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
-key-wallet = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
-key-wallet-ffi = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
-key-wallet-manager = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
-dash-network = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
-dashcore-rpc = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
+dashcore = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
+dash-network-seeds = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
+dash-spv = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
+key-wallet = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
+key-wallet-ffi = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
+key-wallet-manager = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
+dash-network = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
+dashcore-rpc = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
 
 tokio-metrics = "0.5"
 
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/ffi/WalletManagerNative.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/ffi/WalletManagerNative.kt
@@ -496,6 +496,25 @@ internal object WalletManagerNative {
     external fun spvIsRunning(managerHandle: Long): Boolean
     external fun spvStop(managerHandle: Long)
 
+    /**
+     * The proTxHashes of every masternode in the current-tip deterministic
+     * masternode list whose voting-key hash matches the 20-byte [votingKeyId]
+     * (hash160 of a voting public key), as a flat `byte[]` of concatenated
+     * 32-byte proTxHashes (internal byte order) — the caller splits into
+     * 32-byte rows. Replaces dashj's
+     * `MasternodeListManager.getMasternodesByVotingKey(votingKeyId)` used by
+     * contested-username voting. Returns an EMPTY (non-null) `byte[]` when the
+     * masternode list hasn't synced (SPV client not running / DML unavailable)
+     * or no masternode uses the key; throws only on a structural FFI error.
+     * JNI symbol:
+     * `Java_org_dashfoundation_dashsdk_ffi_WalletManagerNative_masternodesByVotingKey`;
+     * bridges `platform_wallet_manager_masternodes_by_voting_key`.
+     */
+    external fun masternodesByVotingKey(
+        managerHandle: Long,
+        votingKeyId: ByteArray,
+    ): ByteArray
+
     // ── Wallet-memory snapshots (Wave-1B) ─────────────────────────────
 
     /**
--- a/packages/rs-platform-wallet-ffi/src/spv.rs
+++ b/packages/rs-platform-wallet-ffi/src/spv.rs
@@ -3,7 +3,9 @@
 use std::ffi::{CStr, CString};
 use std::os::raw::c_char;
 
+use dashcore::hashes::Hash;
 use dashcore::sml::llmq_type::LlmqDevnetParams;
+use dashcore::PubkeyHash;
 use platform_wallet::spv::{
     ClientConfig, DevnetConfig, ProgressPercentage, SpvPeerNodeType, SyncProgress, SyncState,
 };
@@ -11,7 +13,7 @@ use platform_wallet::spv::{
 use crate::error::*;
 use crate::handle::*;
 use crate::runtime::{block_on_worker, runtime};
-use crate::types::FFINetwork;
+use crate::types::{FFINetwork, IdentifierArray};
 use crate::{check_ptr, unwrap_option_or_return, unwrap_result_or_return};
 
 pub const SPV_SYNC_STATE_WAIT_FOR_EVENTS: u32 = 0;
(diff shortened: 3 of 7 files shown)


thepastaclaw:
✅ Final review complete — no blockers (commit ce8233ed)

packages/rs-platform-wallet-ffi/src/types.rs
```diff
@@ -136,6 +136,23 @@ impl IdentifierArray {
...
+        let count = hashes.len();
+        if count == 0 {
+            return Self::empty();
+        }
+
+        let mut items = hashes;
+        let ptr = items.as_mut_ptr();
+        std::mem::forget(items);
+
+        Self { items: ptr, count }
```
thepastaclaw: **🔴 Blocking: Preserve the exact allocation layout for returned hash arrays**

`from_hashes` leaks a `Vec<[u8; 32]>` but exports only its pointer and length. The paired free function reconstructs it with `Vec::from_raw_parts(items, count, count)`, which is valid only when the original capacity equals `count`. That invariant is false on the production path: collecting one match from the filtered masternode map produces a vector with spare capacity, and the same-layout mapping in the runtime preserves it. JNI unconditionally invokes the free function, so an ordinary one-result lookup can deallocate with a different layout than was allocated, causing undefined behavior. Convert the vector to an exact-length boxed slice before transferring ownership.

```suggestion
let mut items = hashes.into_boxed_slice();
let ptr = items.as_mut_ptr();
std::mem::forget(items);

Self { items: ptr, count }
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/spv.rs
```diff
@@ -152,6 +154,42 @@ pub unsafe extern "C" fn platform_wallet_manager_sync_progress(
...
+pub unsafe extern "C" fn platform_wallet_manager_masternodes_by_voting_key(
+    handle: Handle,
+    voting_key_id: *const u8,
+    out_array: *mut IdentifierArray,
+) -> PlatformWalletFFIResult {
+    check_ptr!(voting_key_id);
+    check_ptr!(out_array);
+    // Publish the empty sentinel before any fallible work so an error return
+    // never leaves the out-param holding uninitialized stack bytes.
+    *out_array = IdentifierArray::empty();
```
thepastaclaw: **🔴 Blocking: Initialize the output before validating the input pointer**

`check_ptr!(voting_key_id)` can return `ErrorNullPointer` before `*out_array` receives the empty sentinel. A valid but initially uninitialized output therefore remains garbage on this error path, despite the function's explicit guarantee that errors leave the output initialized. A C or Swift caller performing unconditional cleanup can pass that garbage to `platform_wallet_identifier_array_free`, which may attempt to free an arbitrary pointer. Validate `out_array`, publish the sentinel, and only then validate the input pointer.

```suggestion
check_ptr!(out_array);
*out_array = IdentifierArray::empty();
check_ptr!(voting_key_id);
```

<sub>source: ['codex']</sub>

Cargo.toml
```diff
@@ -52,14 +52,14 @@ members = [
...
-dashcore-rpc = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
+dashcore = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
+dash-network-seeds = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
+dash-spv = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
+key-wallet = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
+key-wallet-ffi = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
+key-wallet-manager = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
+dash-network = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
+dashcore-rpc = { git = "https://github.com/bfoss765/rust-dashcore", rev = "4d927c155b6740ba7e1f271144217f7e82b69990" }
```
thepastaclaw: **🔴 Blocking: Remove the temporary personal-fork dependency pin before merge**

All eight workspace rust-dashcore dependencies are redirected to `bfoss765/rust-dashcore`. The pinned revision is four commits beyond the previous Dash-owned revision and includes 528 changed lines of unrelated owner-tagged-reservation and asset-lock behavior in addition to the lookup helper. The PR description explicitly says this pin is temporary and must not merge as-is; upstream rust-dashcore PR #916 is still open and currently mixes these changes. Re-pin `Cargo.toml` and `Cargo.lock` to a Dash-owned revision containing the helper, or implement the small voting-key filtering loop locally until a suitable upstream revision exists.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/spv/runtime.rs
```diff
@@ -349,6 +349,53 @@ impl SpvRuntime {
...
+
+        list.masternodes_by_voting_key(voting_key_id)
+            .into_iter()
+            .map(|pro_tx| {
+                let mut out = [0u8; 32];
+                out.copy_from_slice(pro_tx.as_ref());
+                out
+            })
+            .collect()
+    }
```
thepastaclaw, line 391: **🟡 Suggestion: Add regression coverage for the new lookup boundaries**

The fork's helper test covers filtering, multi-match, and no-match behavior, but this PR has no tests for its newly introduced layers. Add coverage for unavailable DML state and preservation of internal proTxHash byte order in the runtime, plus FFI/JNI coverage for the empty sentinel and concatenation of multiple 32-byte rows. These are the contracts consumed by Kotlin and are not exercised by the reported `cargo check`.

thepastaclaw (changes requested):

The Rust-to-Kotlin lookup is structurally coherent, but three blocking issues remain: the returned hash buffer can be freed with the wrong allocation layout, one error path violates the documented initialized-out-parameter contract, and the workspace is pinned to an explicitly temporary personal fork containing unrelated wallet changes. The new runtime and FFI/JNI byte contracts also lack regression coverage.

### Review provenance

<details>

```

In `packages/rs-platform-wallet-ffi/src/types.rs`:
- [BLOCKING] packages/rs-platform-wallet-ffi/src/types.rs:144-154: Preserve the exact allocation layout for returned hash arrays
  `from_hashes` leaks a `Vec<[u8; 32]>` but exports only its pointer and length. The paired free function reconstructs it with `Vec::from_raw_parts(items, count, count)`, which is valid only when the original capacity equals `count`. That invariant is false on the production path: collecting one match from the filtered masternode map produces a vector with spare capacity, and the same-layout mapping in the runtime preserves it. JNI unconditionally invokes the free function, so an ordinary one-result lookup can deallocate with a different layout than was allocated, causing undefined behavior. Convert the vector to an exact-length boxed slice before transferring ownership.

- [BLOCKING] packages/rs-platform-wallet-ffi/src/spv.rs:176-180: Initialize the output before validating the input pointer
  `check_ptr!(voting_key_id)` can return `ErrorNullPointer` before `*out_array` receives the empty sentinel. A valid but initially uninitialized output therefore remains garbage on this error path, despite the function's explicit guarantee that errors leave the output initialized. A C or Swift caller performing unconditional cleanup can pass that garbage to `platform_wallet_identifier_array_free`, which may attempt to free an arbitrary pointer. Validate `out_array`, publish the sentinel, and only then validate the input pointer.

In `Cargo.toml`:
- [BLOCKING] Cargo.toml:55-62: Remove the temporary personal-fork dependency pin before merge
  All eight workspace rust-dashcore dependencies are redirected to `bfoss765/rust-dashcore`. The pinned revision is four commits beyond the previous Dash-owned revision and includes 528 changed lines of unrelated owner-tagged-reservation and asset-lock behavior in addition to the lookup helper. The PR description explicitly says this pin is temporary and must not merge as-is; upstream rust-dashcore PR #916 is still open and currently mixes these changes. Re-pin `Cargo.toml` and `Cargo.lock` to a Dash-owned revision containing the helper, or implement the small voting-key filtering loop locally until a suitable upstream revision exists.

- [SUGGESTION] packages/rs-platform-wallet/src/spv/runtime.rs:370-397: Add regression coverage for the new lookup boundaries
  The fork's helper test covers filtering, multi-match, and no-match behavior, but this PR has no tests for its newly introduced layers. Add coverage for unavailable DML state and preservation of internal proTxHash byte order in the runtime, plus FFI/JNI coverage for the empty sentinel and concatenation of multiple 32-byte rows. These are the contracts consumed by Kotlin and are not exercised by the reported `cargo check`.
```
</details>

packages/rs-platform-wallet-ffi/src/types.rs
bfoss765: Fixed in b1abb827e3.

Confirmed the invariant really is false on the production path rather than only in principle. The `masternodes_by_voting_key` collect goes through a `filter`, so its `size_hint` lower bound is 0 and the vector grows by doubling: a one-match lookup gives **len 1, capacity 4**. I ran it under an allocator that records the `Layout` each pointer was allocated with and compares it at `dealloc` — the old path allocates **128 bytes and frees as 32**. JNI invokes the free unconditionally, so that is reachable from an ordinary single-result lookup.

Took your `into_boxed_slice` suggestion, and applied it to `new` as well: `new` now delegates to `from_hashes`, and the free reclaims a `Box<[[u8; 32]]>` of the same length, so creation and release are the same layout by construction rather than by a capacity coincidence. One ownership contract instead of two.

Tests: a round trip built through the same filtered collect, pinning `capacity > len` as an explicit precondition so the test keeps exercising the mismatch if the producer changes; plus a multi-row byte-for-byte round trip and the empty-sentinel case.

packages/rs-platform-wallet-ffi/src/spv.rs
bfoss765: Fixed in b1abb827e3, exactly as suggested — `check_ptr!(out_array)` and the sentinel now come before every other guard.

Added a regression test that pre-poisons the out-param the way an uninitialized C local looks (non-null pointer, non-zero count) and asserts the `(null, 0)` sentinel survives a null `voting_key_id`, plus one for the null-`out_array` rejection. I verified the test actually catches this by restoring the old guard order first — it fails with `got items=0xdeadbeef, count=9`, i.e. precisely the arbitrary pointer a cleanup-on-error caller would have handed to `platform_wallet_identifier_array_free`.

While in this function: it also had a pre-existing rustfmt violation that was failing CI's format check independently of these fixes, so that's corrected in the same commit.

Cargo.toml
bfoss765: Fixed in ce8233edb7 — the pin is gone entirely.

All eight workspace rust-dashcore dependencies are back on the Dash-owned revision `70d4bf8e36057c58e02d56769a6e9760f701dd06`, which is exactly what `v4.2-dev` pins. `Cargo.toml` and `Cargo.lock` no longer appear in this PR's diff at all, and `grep bfoss765` over the full PR diff returns nothing.

I took the second option you offered rather than waiting for upstream. rust-dashcore#916 carries the helper only as the last of four commits, is `CHANGES_REQUESTED`, conflicts against `dev`, and its own reviewer has asked for it to be split — so waiting on it was open-ended.

**The filter is now local.** A 14-line private free function in `packages/rs-platform-wallet/src/spv/runtime.rs`, over the engine's current-tip `MasternodeList`. I checked both fields it reads (`key_id_voting`, `pro_reg_tx_hash`) are public on `70d4bf8e` before relying on them. It also folds in the `ProTxHash -> [u8; 32]` internal-byte-order conversion the caller was previously doing itself, so `masternodes_by_voting_key_blocking` comes out net shorter than it was with the helper.

**Tested.** Four new unit tests cover multi-match, single-match, no-match and the empty-list case. Against the Dash-owned pin: `cargo test -p platform-wallet --lib` is 497 passed / 0 failed, and `cargo check -p platform-wallet-ffi -p rs-unified-sdk-jni` is clean.

**Tracked, so the duplication doesn't get orphaned:** #4262 records why the local copy exists, names the exact function to delete and the upstream call to restore in its place. That issue number is cited in a doc comment directly above the function, so anyone reading the code finds the plan, and I left a note on rust-dashcore#916 so whoever merges it sees the downstream cleanup it enables.

thepastaclaw (commented):

All three prior blocking findings and the fork-pin issue are confirmed fixed at head ce8233edb7: IdentifierArray now allocates and frees an exact-length boxed slice on both paths (verified in types.rs), the FFI out-parameter sentinel is published before the input-pointer guard (verified in spv.rs), and the temporary personal-fork Cargo pin is fully reverted with a net-zero Cargo.toml/Cargo.lock diff since the PR base, replaced by a locally-implemented, well-documented, and tested voting-key filter. One suggestion-level test-coverage gap remains only partially closed (the blocking accessor's unavailable-client/engine/list branches and the JNI/Kotlin round trip are still untested), and a new suggestion surfaced in this pass: the new native binding has no public Kotlin wrapper on PlatformWalletManager, so SDK consumers outside the kotlin-sdk module cannot actually call it through supported API despite the PR's stated goal of surfacing this lookup to Kotlin.

### Review provenance
- Sonnet reviewers: `claude-sonnet-5` — general (completed), `claude-sonnet-5` — security-auditor (failed), `claude-sonnet-5` — rust-quality (completed), `claude-sonnet-5` — ffi-engineer (completed), `claude-sonnet-5` — security-auditor (completed)

🟡 1 suggestion(s)

<details>

```

In `packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/ffi/WalletManagerNative.kt`:
- [SUGGESTION] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/ffi/WalletManagerNative.kt:496-516: Expose masternodesByVotingKey through the public PlatformWalletManager API
  WalletManagerNative is declared `internal object` (line 23), and masternodesByVotingKey is added here as a raw `external fun` with no corresponding public wrapper in PlatformWalletManager.kt. Every other native call in this class — createWallet, removeWallet, platformAddressSyncStart, identitySyncStart, shieldedSyncStart, trackedIdentityRecoveryAssetLocks, etc. — has a matching public `suspend fun` on PlatformWalletManager that validates input, dispatches via `withContext(Dispatchers.IO)`, and calls through `mapNativeErrors`. This lookup breaks that pattern: it is unreachable from outside the kotlin-sdk module (managerHandle is also private on PlatformWalletManager), and even the raw binding bypasses the IO-dispatch convention even though the underlying Rust accessor documents itself as blocking via two tokio::RwLock::blocking_read calls. Since this PR's stated purpose is surfacing the lookup to Kotlin/dash-wallet consumers, the feature isn't actually usable through supported API without this wrapper.
```
</details>

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/ffi/WalletManagerNative.kt
```diff
@@ -496,6 +496,25 @@ internal object WalletManagerNative {
...
+     * masternode list hasn't synced (SPV client not running / DML unavailable)
+     * or no masternode uses the key; throws only on a structural FFI error.
+     * JNI symbol:
+     * `Java_org_dashfoundation_dashsdk_ffi_WalletManagerNative_masternodesByVotingKey`;
+     * bridges `platform_wallet_manager_masternodes_by_voting_key`.
+     */
+    external fun masternodesByVotingKey(
+        managerHandle: Long,
+        votingKeyId: ByteArray,
+    ): ByteArray
```
thepastaclaw, line 516: **🟡 Suggestion: Expose masternodesByVotingKey through the public PlatformWalletManager API**

WalletManagerNative is declared `internal object` (line 23), and masternodesByVotingKey is added here as a raw `external fun` with no corresponding public wrapper in PlatformWalletManager.kt. Every other native call in this class — createWallet, removeWallet, platformAddressSyncStart, identitySyncStart, shieldedSyncStart, trackedIdentityRecoveryAssetLocks, etc. — has a matching public `suspend fun` on PlatformWalletManager that validates input, dispatches via `withContext(Dispatchers.IO)`, and calls through `mapNativeErrors`. This lookup breaks that pattern: it is unreachable from outside the kotlin-sdk module (managerHandle is also private on PlatformWalletManager), and even the raw binding bypasses the IO-dispatch convention even though the underlying Rust accessor documents itself as blocking via two tokio::RwLock::blocking_read calls. Since this PR's stated purpose is surfacing the lookup to Kotlin/dash-wallet consumers, the feature isn't actually usable through supported API without this wrapper.

```suggestion
suspend fun masternodesByVotingKey(votingKeyId: ByteArray): List<ByteArray> = withContext(Dispatchers.IO) {
    require(votingKeyId.size == 20) { "votingKeyId must be exactly 20 bytes" }
    val flat = mapNativeErrors { WalletManagerNative.masternodesByVotingKey(managerHandle, votingKeyId) }
    check(flat.size % 32 == 0) { "masternodesByVotingKey returned a non-multiple-of-32 buffer" }
    flat.toList().chunked(32).map { it.toByteArray() }
}
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/spv.rs
thepastaclaw: Resolved in `b1abb82` — _Initialize the output before validating the input pointer_ no longer present.

packages/rs-platform-wallet-ffi/src/types.rs
thepastaclaw: Resolved in `b1abb82` — _Preserve the exact allocation layout for returned hash arrays_ no longer present.

Cargo.toml
thepastaclaw: Resolved in `ce8233e` — _Remove the temporary personal-fork dependency pin before merge_ no longer present.

QuantumExplorer (approved):

PR #4259, opened 2026-07-31
feat(platform-wallet): classic Dash message signing (signMessage) over FFI, JNI, Kotlin, and Swift

HashEngineering:
## What this adds

`CoreWallet::sign_message` — classic Dash signed messages ("proof you own this
address"), exposed through the full binding chain. Given a P2PKH address the
wallet holds keys for and an arbitrary UTF-8 message, it returns the 65-byte
recoverable signature, base64-encoded — the same wire format as dashj's
`ECKey.signMessage` and Dash Core's `signmessage` RPC, so existing verifiers
(`verifymessage`, dashj `signedMessageToKey`) accept it as-is.

Primary consumer: the Android/iOS wallets' CrowdNode integrations, where API
withdrawal and email registration are signed-message proofs of address
ownership (currently done in dashj on Android; this makes the kotlin-sdk /
swift-sdk route possible).

## API surface

| Layer | Signature |
|---|---|
| Rust | `CoreWallet::sign_message(&self, address, message, signer) -> Result<String>` |
| C FFI | `core_wallet_sign_message(handle, address_ptr/len, message_ptr/len, core_signer_handle, out_signature)` |
| Kotlin | `ManagedPlatformWallet.signMessage(address: String, message: String, coreSignerHandle: Long): String` (suspend) |
| Swift | `ManagedCoreWallet.signMessage(address: String, message: String) throws -> String` |

(Swift takes no signer parameter deliberately — the swift-sdk convention is a
per-call internal `MnemonicResolver`, as at every other seed-backed call site.)

## Design decisions

- **Recovery id by trial, not recoverable signing.** The digest and
  serialization come from `dashcore::sign_message`; the recovery id is found
  by trying the four candidates against the signer's public key — the same
  approach dashj itself uses (`ECKey.findRecoveryId`). This means every
  `Signer` backend (soft seed, future hardware/HSM) gets signed-message
  support without needing a recoverable-signing method. Cost: at most four
  recovery attempts on one 32-byte digest, off the hot path.
- **Key material never crosses the FFI.** `core_wallet_sign_message` takes the
  caller's existing `MnemonicResolverHandle`, exactly like the send paths.
- **Signable accounts only.** Watch-only DashPay external accounts are
  refused with a typed error; BIP44/BIP32/CoinJoin and DashPay receiving
  accounts sign. The predicate is a local copy of the QA integration branch's
  `funding_privacy::is_signable_funding_account` (identical body), so when
  that module lands the port converges by pure deletion.
- **Typed errors.** Unknown/watch-only address → new FFI code 31
  (`ErrorSigningKeyUnavailable`); malformed address → `ErrorInvalidParameter`;
  both mirrored in Kotlin (`DashSdkError.PlatformWallet.SigningKeyUnavailable`)
  and Swift (`signingKeyUnavailable`).

## Why error code 31, and why the 27–30 gap

Code 31 is what the keystore rework (#4183) assigns to
`ErrorSigningKeyUnavailable` on the QA integration line. This PR carries the
variant at the same number so the enums converge instead of colliding; the
27–30 gap is intentional and reserved. Until #4183's guarded catch-all lands,
`MessageSigningFailed` (signer-internal failure) falls through to
`ErrorUnknown` — noted in a comment at the mapping site.

## One contract fix included

Swift's empty-string marshalling (`Array("".utf8).withUnsafeBufferPointer`)
yields a nil base address, which the FFI's null-check rejected — despite an
empty message being legitimately signable (matches dashj and Core). `read_utf8`
now short-circuits at `len == 0` without dereferencing; `# Safety` docs state
the contract precisely. `address_ptr` remains required non-null.

## Verification

- **Byte-for-byte dashj parity**, two directions:
  - RFC6979 golden: same mnemonic/path/message signed via dashj
    `ECKey.signMessage` and via this Rust produce the identical base64
    (`HzW1qnS7pYhopcbz1ZbiQY+axMMDQ2cMXBjey37IQS15OOluF6l/rfEre7Bl8AzUOwYdANkLYRelpKJmIH4kfZM=`).
  - dashj's own signed-message test vector verifies against this implementation.
- `platform-wallet` lib suite 498/498; `platform-wallet-ffi` 207/207
  (including the new mapping tests); goldens unchanged across the v4.2-dev port.
- Kotlin `:sdk:compileDebugKotlin` clean; `DashSdkErrorTest` passes.
- iOS: `build_ios.sh --target sim` succeeds — cbindgen header carries
  `..._ERROR_SIGNING_KEY_UNAVAILABLE = 31`, `ManagedCoreWallet.swift` compiles,
  and SwiftExampleApp installs and runs on the iOS 26.4 simulator.
- Not covered: no example-app demo screen for signMessage (neither example app
  has one today); end-to-end CrowdNode use happens in the wallet integrations.

## Summary by CodeRabbit

* **New Features**
  * Added classic Dash message signing for wallet-owned addresses in the Kotlin and Swift SDKs.
  * Signatures are returned as Base64-encoded recoverable signatures.
  * Empty messages and embedded content are supported while preserving exact message data.

* **Bug Fixes**
  * Improved validation and error reporting for invalid addresses, malformed messages, unavailable keys, and signing failures.
  * Clarified guidance for watch-only or unowned addresses and unsupported signing requests.

### DIFF
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
@@ -174,6 +174,79 @@ class ManagedPlatformWallet internal constructor(
     }
 
     /**
+     * Sign [message] with the private key behind [address] and return the
+     * signature as base64 — a **classic Dash signed message**, byte-for-byte
+     * compatible with dashj's `ECKey.signMessage` and Dash Core's `signmessage`
+     * RPC, and verifiable by `verifymessage`, `ECKey.verifyMessage`, and
+     * CrowdNode's server-side check.
+     *
+     * **The format.** The signed digest is
+     * `SHA256d(prefix ‖ varint(message.length) ‖ message)`, where the prefix is
+     * the historical `"\x19DarkCoin Signed Message:\n"` — *not* `"Dash"`. Dash
+     * inherited that string from before the rename and every existing verifier
+     * depends on it, so it can never change. The returned signature is the
+     * 65-byte BIP-137-style recoverable form (`header ‖ r ‖ s`, with
+     * `header = 27 + recoveryId + 4`, the `+ 4` marking a compressed public
+     * key), base64-encoded. A verifier recovers the public key from the digest
+     * and compares its hash to [address] — which is why only P2PKH addresses
+     * can sign: no other payload has a defined recovery comparison.
+     *
+     * **The CrowdNode use case.** This is the primitive the Android wallet's
+     * CrowdNode integration needs: it signs short strings — a withdrawal amount,
+     * the account email — which CrowdNode verifies against the address that
+     * funded the account. It is a **proof of address ownership, not a spend**:
+     * no UTXO is selected, reserved, spent, or broadcast, no balance changes, and
+     * nothing is persisted. Calling it repeatedly is free and side-effect-free.
+     *
+     * **Which addresses can sign.** [address] must be a P2PKH address of *this*
+     * wallet, on this wallet's network, already derived into one of its address
+     * pools, and belonging to a **signable funds account** (BIP44 / BIP32 /
+     * CoinJoin / DashPay-*receiving*). A watch-only DashPay **external** account
+     * holds a contact's receiving addresses whose private keys we never had, so
+     * those are refused exactly like any other address the wallet does not own:
+     * [org.dashfoundation.dashsdk.errors.DashSdkError.PlatformWallet.SigningKeyUnavailable].
+     *
+     * An unparseable, wrong-network, or non-P2PKH address is caller input and
+     * surfaces natively as `ErrorInvalidParameter` (2), which has no dedicated
+     * Kotlin arm today and therefore arrives as
+     * [org.dashfoundation.dashsdk.errors.DashSdkError.PlatformWallet.Generic]
+     * with `code == 2`; the message names which of the three it was. Branch on
+     * [org.dashfoundation.dashsdk.errors.DashSdkError.PlatformWallet.SigningKeyUnavailable]
+     * to tell "wallet does not own this address" apart from "this is not a
+     * usable address".
+     *
+     * **Determinism.** Signing is RFC6979 deterministic and low-s normalized, so
+     * the same ([address], [message]) pair on the same seed always returns the
+     * same string. Two calls yielding different signatures means the key or the
+     * message differed.
+     *
+     * Runs through the manager's [TeardownGate] like every other native op.
+     *
+     * @param address the P2PKH address whose key signs; must be one this wallet
+     *   owns and has derived.
+     * @param message the string to sign, **verbatim**. It is length-prefixed
+     *   into the digest, so trailing whitespace and newlines are significant and
+     *   the verifier must receive the identical bytes. An empty string is valid.
+     * @param coreSignerHandle the manager's `MnemonicResolverHandle`
+     *   (`PlatformWalletManager.mnemonicResolverHandle`); no private key crosses
+     *   the boundary.
+     * @return the base64 signature (88 characters for the 65-byte payload).
+     */
+    suspend fun signMessage(
+        address: String,
+        message: String,
+        coreSignerHandle: Long,
+    ): String = gate.op {
+        require(address.isNotEmpty()) { "address must not be empty" }
+
+        mapNativeErrors {
+            coreWallet().use { core ->
+                core.signMessage(address, message, coreSignerHandle)
+            }
+        }
+    }
+
+    /**
      * The wallet's Platform-payment addresses that currently hold credits,
      * each as a [FundingInput] whose `credits` is the full cached balance —
      * the funding-input candidates for an identity top-up. Port of the
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
@@ -181,7 +181,10 @@ class ManagedPlatformWallet internal constructor(
      * CrowdNode's server-side check.
      *
      * **The format.** The signed digest is
-     * `SHA256d(prefix ‖ varint(message.length) ‖ message)`, where the prefix is
+     * `SHA256d(prefix ‖ varint(bytes.size) ‖ bytes)` over
+     * `bytes = message.toByteArray(Charsets.UTF_8)`: the length prefix counts
+     * **UTF-8 bytes**, not `String.length`, which counts UTF-16 code units and
+     * diverges for any non-ASCII text. The prefix is
      * the historical `"\x19DarkCoin Signed Message:\n"` — *not* `"Dash"`. Dash
      * inherited that string from before the rename and every existing verifier
      * depends on it, so it can never change. The returned signature is the
@@ -224,9 +227,13 @@ class ManagedPlatformWallet internal constructor(
      *
      * @param address the P2PKH address whose key signs; must be one this wallet
      *   owns and has derived.
-     * @param message the string to sign, **verbatim**. It is length-prefixed
-     *   into the digest, so trailing whitespace and newlines are significant and
-     *   the verifier must receive the identical bytes. An empty string is valid.
+     * @param message the string to sign, **verbatim** as UTF-8. It is
+     *   length-prefixed into the digest, so trailing whitespace and newlines are
+     *   significant and the verifier must receive the identical bytes. An empty
+     *   string is valid. Must be well-formed text: a string holding an unpaired
+     *   UTF-16 surrogate has no UTF-8 encoding and is rejected with
+     *   [IllegalArgumentException] rather than signed after a silent
+     *   substitution.
      * @param coreSignerHandle the manager's `MnemonicResolverHandle`
      *   (`PlatformWalletManager.mnemonicResolverHandle`); no private key crosses
      *   the boundary.
@@ -238,12 +245,48 @@ class ManagedPlatformWallet internal constructor(
         coreSignerHandle: Long,
     ): String = gate.op {
         require(address.isNotEmpty()) { "address must not be empty" }
+        // The digest commits to the message's UTF-8 bytes, but a Kotlin String
+        // is an unvalidated UTF-16 sequence and may hold an unpaired surrogate,
+        // which has no UTF-8 encoding at all. Every conversion below is LENIENT
+        // and they do not even agree: the JNI bridge's String read substitutes
+        // U+FFFD, while `toByteArray(Charsets.UTF_8)` substitutes '?'. Either
+        // way the wallet would sign bytes the caller never wrote and hand back a
+        // signature that verifies for a different message — silently. Rejected
+        // here, the one layer that still has the exact UTF-16 and can say why.
+        require(message.hasNoUnpairedSurrogate()) {
+            "message must be well-formed text: it contains an unpaired UTF-16 surrogate, " +
+                "which has no UTF-8 encoding and would be silently substituted before signing"
+        }
 
         mapNativeErrors {
             coreWallet().use { core ->
                 core.signMessage(address, message, coreSignerHandle)
             }
         }
+    }
+
+    /**
+     * Whether every UTF-16 surrogate in this string is part of a well-formed
+     * high/low pair — i.e. whether the string has an exact UTF-8 encoding.
+     *
+     * Scanned directly rather than via a strict `CharsetEncoder` to keep the
+     * check allocation-free on the hot path; the two agree on exactly which
+     * strings are encodable.
+     */
+    private fun String.hasNoUnpairedSurrogate(): Boolean {
+        var i = 0
+        while (i < length) {
+            val c = this[i]
+            when {
+                c.isHighSurrogate() -> {
+                    if (i + 1 >= length || !this[i + 1].isLowSurrogate()) return false
+                    i += 2
+                }
+                c.isLowSurrogate() -> return false
+                else -> i++
+            }
+        }
+        return true
     }
 
     /**
--- a/packages/rs-platform-wallet-ffi/src/core_wallet/sign_message.rs
+++ b/packages/rs-platform-wallet-ffi/src/core_wallet/sign_message.rs
@@ -31,7 +31,11 @@ use std::os::raw::c_char;
 /// the same shape for symmetry.
 ///
 /// `on_error` builds the typed error, so each argument reports itself rather
-/// than borrowing the other's phrasing.
+/// than borrowing the other's phrasing: a malformed address yields
+/// `MessageSigningAddressInvalid` and malformed message bytes yield
+/// `MessageSigningMessageInvalid`. Both are caller input and both surface as
+/// [`PlatformWalletFFIResultCode::ErrorInvalidParameter`] (2) — neither is an
+/// internal failure, so neither may reach `ErrorUnknown`.
 ///
 /// `len == 0` yields the empty string WITHOUT dereferencing `ptr`, which may
 /// legitimately be null: Swift's established marshalling
@@ -73,14 +77,20 @@ unsafe fn read_utf8(
 ///   significant and the verifier must receive the identical bytes. An empty
 ///   message is valid and signable, and `message_ptr` MAY be null when
 ///   `message_len` is 0 — the shape host marshalling naturally produces for an
-///   empty string.
+///   empty string. Bytes that are not valid UTF-8 fail with
+///   [`PlatformWalletFFIResultCode::ErrorInvalidParameter`] (2), the same code
+///   a malformed address yields: there is no string to sign, and it is the
+///   caller's argument to fix.
 /// * `core_signer_handle` — the caller's `MnemonicResolverHandle`; ownership is
 ///   retained by the caller (this function does NOT destroy it).
 /// * `out_signature` — receives a heap-allocated C string holding the base64
 ///   signature. Free with [`super::core_wallet_free_address`].
 ///
 /// # Safety
-/// `address_ptr` must be non-null and readable for `address_len` bytes;
+/// Encoding is CHECKED, not assumed: either buffer failing to be valid UTF-8 is
+/// a safe, typed `ErrorInvalidParameter` return rather than undefined behavior.
+/// What the caller must guarantee is memory validity — `address_ptr` must be
+/// non-null and readable for `address_len` bytes;
 /// `message_ptr` must be readable for `message_len` bytes when that length is
 /// non-zero (it may be null when the length is 0). Both must stay valid for the
 /// duration of the call; `out_signature` must point to writable memory for one
@@ -133,9 +143,11 @@ pub unsafe extern "C" fn core_wallet_sign_message(
         }
     }));
     // A non-UTF-8 message is reported against the (now known) address, so the
-    // error names the signing target the caller asked about.
+    // error names the signing target the caller asked about — but as a MESSAGE
+    // error, not an address one: the address parsed fine, and this maps to
+    // ErrorInvalidParameter like any other malformed argument.
     let message = unwrap_result_or_return!(read_utf8(message_ptr, message_len, |e| {
-        PlatformWalletError::MessageSigningFailed {
+        PlatformWalletError::MessageSigningMessageInvalid {
             address: address.clone(),
             reason: format!("message is not valid UTF-8: {e}"),
         }
--- a/packages/rs-platform-wallet-ffi/src/core_wallet/sign_message.rs
+++ b/packages/rs-platform-wallet-ffi/src/core_wallet/sign_message.rs
@@ -15,7 +15,7 @@
 use crate::error::*;
 use crate::handle::{Handle, CORE_WALLET_STORAGE};
 use crate::runtime::runtime;
-use crate::{check_ptr, unwrap_option_or_return, unwrap_result_or_return};
+use crate::{check_ptr, unwrap_result_or_return};
 use platform_wallet::PlatformWalletError;
 use rs_sdk_ffi::{MnemonicResolverCoreSigner, MnemonicResolverHandle};
 use std::ffi::CString;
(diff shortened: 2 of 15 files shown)


thepastaclaw:
✅ Final review complete — no blockers (commit 09a27ac6)

packages/rs-platform-wallet/src/wallet/core/sign_message.rs
```diff
@@ -0,0 +1,431 @@
...
+        };
+
+        let hash = signed_msg_hash(message);
+        let (signature, public_key) = signer
+            .sign_ecdsa(&path, hash.to_byte_array())
+            .await
+            .map_err(|e| PlatformWalletError::MessageSigningFailed {
+                address: target.to_string(),
+                reason: format!("signer rejected the digest at {path}: {e}"),
+            })?;
```
thepastaclaw, line 247: **🔴 Blocking: Honor the signer's digest capability before signing**

`key_wallet::signer::Signer` explicitly assigns capability dispatch to callers and documents `sign_ecdsa` as valid only when the signer advertises `SignerMethod::Digest`. This public generic method calls it unconditionally. The production mnemonic resolver supports digest signing, but a legitimate transaction-only hardware signer may reject the call, panic because the method is unreachable, or blind-sign a host-computed digest despite advertising a policy that forbids that operation. Fail before entering the backend, and add a regression signer that advertises only `Transaction(...)` and panics if `sign_ecdsa` is invoked.

```suggestion
        if !signer.supports(key_wallet::signer::SignerMethod::Digest) {
            return Err(PlatformWalletError::MessageSigningFailed {
                address: target.to_string(),
                reason: "signer does not support digest signing".to_string(),
            });
        }

        let hash = signed_msg_hash(message);
        let (signature, public_key) = signer
            .sign_ecdsa(&path, hash.to_byte_array())
            .await
            .map_err(|e| PlatformWalletError::MessageSigningFailed {
                address: target.to_string(),
                reason: format!("signer rejected the digest at {path}: {e}"),
            })?;
```

<sub>source: ['codex']</sub>

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
```diff
@@ -172,6 +172,79 @@ class ManagedPlatformWallet internal constructor(
...
+    /**
+     * Sign [message] with the private key behind [address] and return the
+     * signature as base64 — a **classic Dash signed message**, byte-for-byte
+     * compatible with dashj's `ECKey.signMessage` and Dash Core's `signmessage`
+     * RPC, and verifiable by `verifymessage`, `ECKey.verifyMessage`, and
+     * CrowdNode's server-side check.
+     *
+     * **The format.** The signed digest is
+     * `SHA256d(prefix ‖ varint(message.length) ‖ message)`, where the prefix is
+     * the historical `"\x19DarkCoin Signed Message:\n"` — *not* `"Dash"`. Dash
```
thepastaclaw, line 188: **🟡 Suggestion: Document the signed message's UTF-8 byte length**

`signed_msg_hash` prefixes `msg.len()`, which is the serialized UTF-8 byte count. Kotlin's `message.length` counts UTF-16 code units, so the documented formula is wrong for non-ASCII messages; for example, `"é"` has length 1 but contributes 2 UTF-8 bytes. State that the varint contains `message.toByteArray(Charsets.UTF_8).size`. The analogous Swift documentation at `packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/CoreWallet/ManagedCoreWallet.swift:145` must use `message.utf8.count` instead of `message.count`, which counts extended grapheme clusters.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/core_wallet/sign_message.rs
```diff
@@ -0,0 +1,143 @@
...
+        let signer = MnemonicResolverCoreSigner::new(
+            signer_addr as *mut MnemonicResolverHandle,
+            wallet_id,
+            network,
+        );
+        runtime().block_on(wallet.sign_message(&address, &message, &signer))
+    });
+
+    let result = unwrap_option_or_return!(option);
+    let signature = unwrap_result_or_return!(result);
```
thepastaclaw: **🟡 Suggestion: Release the global Core-wallet storage lock before signing**

`HandleStorage::with_item` retains the global storage map's read guard throughout the closure. This closure performs `runtime().block_on(...)` and invokes the host-owned mnemonic resolver callback, so a potentially slow signing operation prevents insertion or removal of every Core-wallet handle. A callback that re-enters an operation requiring the write lock can deadlock. `CoreWallet` is an inexpensive Arc-backed clone, and the broadcast entry points already clone it out of storage before blocking; use the same pattern here.

```suggestion
    let wallet = unwrap_option_or_return!(CORE_WALLET_STORAGE.with_item(handle, Clone::clone));

    let address = unwrap_result_or_return!(read_utf8(address_ptr, address_len, |e| {
        PlatformWalletError::MessageSigningAddressInvalid {
            address: "<non-UTF-8>".to_string(),
            reason: format!("address is not valid UTF-8: {e}"),
        }
    }));
    // A non-UTF-8 message is reported against the known address, so the
    // error identifies the signing target requested by the caller.
    let message = unwrap_result_or_return!(read_utf8(message_ptr, message_len, |e| {
        PlatformWalletError::MessageSigningFailed {
            address: address.clone(),
            reason: format!("message is not valid UTF-8: {e}"),
        }
    }));
    let network = wallet.network();
    let wallet_id = wallet.wallet_id();
    // SAFETY: `signer_addr` came from `core_signer_handle`, which the caller
    // keeps alive for this call; the signer is dropped before returning.
    let signer = MnemonicResolverCoreSigner::new(
        signer_addr as *mut MnemonicResolverHandle,
        wallet_id,
        network,
    );
    let signature = unwrap_result_or_return!(
        runtime().block_on(wallet.sign_message(&address, &message, &signer))
    );
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/core_wallet/sign_message.rs
```diff
@@ -0,0 +1,143 @@
...
+    address_len: usize,
+    message_ptr: *const u8,
+    message_len: usize,
+    core_signer_handle: *mut MnemonicResolverHandle,
+    out_signature: *mut *mut c_char,
+) -> PlatformWalletFFIResult {
+    // An address is never legitimately empty, so its pointer must be present.
+    // `message_ptr` is deliberately NOT checked: a null pointer with
+    // `message_len == 0` is the empty message, which is signable (see
+    // [`read_utf8`]). It is only dereferenced when the length is non-zero.
```
thepastaclaw, line 111: **🟡 Suggestion: Cover the null-pointer empty-message ABI contract**

The new FFI contract intentionally accepts `message_ptr == NULL` when `message_len == 0`, which is required by the Swift empty-array marshalling path. No endpoint test exercises this behavior, and the Rust wallet tests only sign `"hello"`. Add a direct `core_wallet_sign_message` regression test that passes a null message pointer with zero length, asserts success, and verifies the returned signature against `signed_msg_hash("")`; this prevents a future unconditional `check_ptr!(message_ptr)` from silently breaking empty messages.

<sub>source: ['codex']</sub>

packages/rs-unified-sdk-jni/src/wallet_manager.rs
```diff
@@ -1013,6 +1013,96 @@ pub extern "system" fn Java_org_dashfoundation_dashsdk_ffi_WalletManagerNative_c
...
+                let _ = env.exception_clear();
+                throw_sdk_exception(env, 1, "message string was invalid");
+                return ptr::null_mut();
+            }
+        };
+
+        // Both cross as UTF-8 bytes + length (no trailing NUL), so an embedded
+        // NUL cannot truncate what actually gets signed.
+        let address_bytes = address.as_bytes();
+        let message_bytes = message.as_bytes();
```
thepastaclaw, line 1068: **🟡 Suggestion: Do not silently alter malformed Java UTF-16 messages**

`env.get_string(...).into()` decodes JNI modified UTF-8 through `jni::JavaStr`. For a Java/Kotlin `String` containing an unpaired UTF-16 surrogate, the JNI crate's decoder fails and falls back to `String::from_utf8_lossy`, silently replacing the input before its bytes are signed. A verifier encoding the same Java string through the platform UTF-8 encoder can therefore hash different replacement bytes, contradicting the API's claim that the message is signed verbatim. Marshal a Kotlin-produced UTF-8 `ByteArray`, or read the UTF-16 code units strictly and reject unpaired surrogates rather than signing transformed text.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The signed-message implementation is cryptographically aligned with Dash Core and dashj, and its address-ownership and recovery-ID checks are sound. One blocking issue remains: the public generic Rust API ignores the signer's advertised capabilities and can invoke blind digest signing on a backend that explicitly disallows it. Non-blocking follow-ups cover binding documentation, FFI lock lifetime and ABI regression coverage, and malformed JVM strings.

### Review provenance

<details>

```

In `packages/rs-platform-wallet/src/wallet/core/sign_message.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/wallet/core/sign_message.rs:189-196: Honor the signer's digest capability before signing
  `key_wallet::signer::Signer` explicitly assigns capability dispatch to callers and documents `sign_ecdsa` as valid only when the signer advertises `SignerMethod::Digest`. This public generic method calls it unconditionally. The production mnemonic resolver supports digest signing, but a legitimate transaction-only hardware signer may reject the call, panic because the method is unreachable, or blind-sign a host-computed digest despite advertising a policy that forbids that operation. Fail before entering the backend, and add a regression signer that advertises only `Transaction(...)` and panics if `sign_ecdsa` is invoked.

- [SUGGESTION] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt:182-184: Document the signed message's UTF-8 byte length
  `signed_msg_hash` prefixes `msg.len()`, which is the serialized UTF-8 byte count. Kotlin's `message.length` counts UTF-16 code units, so the documented formula is wrong for non-ASCII messages; for example, `"é"` has length 1 but contributes 2 UTF-8 bytes. State that the varint contains `message.toByteArray(Charsets.UTF_8).size`. The analogous Swift documentation at `packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/CoreWallet/ManagedCoreWallet.swift:145` must use `message.utf8.count` instead of `message.count`, which counts extended grapheme clusters.

In `packages/rs-platform-wallet-ffi/src/core_wallet/sign_message.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/core_wallet/sign_message.rs:109-138: Release the global Core-wallet storage lock before signing
  `HandleStorage::with_item` retains the global storage map's read guard throughout the closure. This closure performs `runtime().block_on(...)` and invokes the host-owned mnemonic resolver callback, so a potentially slow signing operation prevents insertion or removal of every Core-wallet handle. A callback that re-enters an operation requiring the write lock can deadlock. `CoreWallet` is an inexpensive Arc-backed clone, and the broadcast entry points already clone it out of storage before blocking; use the same pattern here.
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/core_wallet/sign_message.rs:98-101: Cover the null-pointer empty-message ABI contract
  The new FFI contract intentionally accepts `message_ptr == NULL` when `message_len == 0`, which is required by the Swift empty-array marshalling path. No endpoint test exercises this behavior, and the Rust wallet tests only sign `"hello"`. Add a direct `core_wallet_sign_message` regression test that passes a null message pointer with zero length, asserts success, and verifies the returned signature against `signed_msg_hash("")`; this prevents a future unconditional `check_ptr!(message_ptr)` from silently breaking empty messages.

- [SUGGESTION] packages/rs-unified-sdk-jni/src/wallet_manager.rs:1063-1075: Do not silently alter malformed Java UTF-16 messages
  `env.get_string(...).into()` decodes JNI modified UTF-8 through `jni::JavaStr`. For a Java/Kotlin `String` containing an unpaired UTF-16 surrogate, the JNI crate's decoder fails and falls back to `String::from_utf8_lossy`, silently replacing the input before its bytes are signed. A verifier encoding the same Java string through the platform UTF-8 encoder can therefore hash different replacement bytes, contradicting the API's claim that the message is signed verbatim. Marshal a Kotlin-produced UTF-8 `ByteArray`, or read the UTF-16 code units strictly and reject unpaired surrogates rather than signing transformed text.
```
</details>

packages/rs-platform-wallet/src/wallet/core/sign_message.rs
HashEngineering: Addressed in 2965337f95. The method now checks that the signer advertises `SignerMethod::Digest` (via `supports()`) before calling `sign_ecdsa`, mirroring the check key-wallet itself performs in `TransactionSigner::sig_and_pubkey`. A refusal maps to the typed `MessageSigningFailed` with a precise reason — reused rather than a new variant because the production resolver advertises Digest (no host can currently hit this), and a new FFI code would spend one of the 27–30 slots reserved for the keystore rework numbering. The requested regression test is included: `TransactionOnlySigner` advertises only `Transaction(Classical)` and its `sign_ecdsa` panics, so a dropped guard fails loudly; it signs against a wallet-owned address so the refusal cannot be confused with the key-unavailable path.

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
HashEngineering: Addressed in 4270d827c2 (docs only). The Kotlin doc now states the varint prefix is `message.toByteArray(Charsets.UTF_8).size`, and the Swift doc at ManagedCoreWallet.swift uses `message.utf8.count` — each with a note on why the language-native count (UTF-16 code units / grapheme clusters) diverges for non-ASCII input. The Rust module doc was already byte-accurate.

packages/rs-platform-wallet-ffi/src/core_wallet/sign_message.rs
HashEngineering: Addressed in 2965337f95. The entry point now clones the Arc-backed `CoreWallet` out of storage via `with_item(handle, Clone::clone)` and runs `block_on` outside the guard, matching the pattern at broadcast.rs:43 and :82. Agreed the impact was real: the resolver callback can block on a biometric prompt, so the old shape stalled every handle operation for that duration. Note broadcast.rs:169 and the addresses.rs entry points share the old shape — left out of scope for this PR, tracked as a follow-up.

packages/rs-platform-wallet-ffi/src/core_wallet/sign_message.rs
HashEngineering: Addressed in 2965337f95, split across layers rather than as a single FFI test: building the described test inside the FFI crate would need a full `CoreWallet<SpvBroadcaster>` in handle storage plus a live resolver (promoting the test wallet fixture to `test-utils`). Instead: (1) `empty_message_is_signable` in platform-wallet signs `""` end-to-end and verifies it against `signed_msg_hash("")`; (2) `empty_message_is_not_a_null_pointer_error` in the FFI crate pins the null-pointer/zero-length acceptance by discrimination — reintroducing `check_ptr!(message_ptr)` makes it fail (verified by temporarily injecting it); (3) `read_utf8_accepts_a_null_pointer_at_zero_length` pins the primitive. Together these cover both the semantic and ABI halves of the contract.

packages/rs-unified-sdk-jni/src/wallet_manager.rs
HashEngineering: Addressed in 4270d827c2, with one deviation from the suggested options: marshalling a Kotlin-produced UTF-8 `ByteArray` does not fix this — JDK's `String.getBytes(UTF_8)` also substitutes unpaired surrogates (with `?` at 0x3f, verified on JDK 17, vs U+FFFD in the JNI path), so it would relocate and change the corruption rather than remove it. The fix validates surrogate pairing strictly at the Kotlin entry point (allocation-free scan behind `require(...)`, consistent with the existing address check) — the last layer that still holds the exact UTF-16 form. Well-formed strings are unaffected and the JNI signature is unchanged. The lenient decode in the shared `read_cstring_required` affects every JNI string parameter in the crate and is flagged as a crate-wide follow-up rather than changed here.

packages/rs-platform-wallet-ffi/src/core_wallet/sign_message.rs
```diff
@@ -0,0 +1,244 @@
...
+        }
+    }));
+    // A non-UTF-8 message is reported against the (now known) address, so the
+    // error names the signing target the caller asked about.
+    let message = unwrap_result_or_return!(read_utf8(message_ptr, message_len, |e| {
+        PlatformWalletError::MessageSigningFailed {
+            address: address.clone(),
+            reason: format!("message is not valid UTF-8: {e}"),
+        }
+    }));
```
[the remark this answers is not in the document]
HashEngineering: Addressed in 5b77dfd8f1. Agreed the asymmetry was real: malformed message bytes are caller input, same as malformed address bytes. Fixed with a dedicated `MessageSigningMessageInvalid` variant mapping to `ErrorInvalidParameter` rather than reusing `MessageSigningAddressInvalid` — that variant renders "message-signing address {address} is unusable", so reusing it would blame a valid address for a message problem in logs and host-facing text; the crate already separates variants by which argument the caller must fix. No new numeric code (`ErrorInvalidParameter` is already mirrored in Kotlin/Swift, so no mirror churn). The `# Safety`/parameter docs now state the message-encoding failure code, `read_utf8`'s contract notes encoding is checked rather than assumed, and a new mapping test asserts both the code and that the rendered error blames the message.

packages/rs-platform-wallet/src/error.rs
```diff
@@ -108,6 +108,81 @@ pub enum PlatformWalletError {
...
+    /// its `Display`, which the FFI conversion's catch-all promotes to
+    /// `ErrorSigningKeyUnavailable`. The remaining causes are genuine internal
+    /// invariant breaks and should surface as unknown rather than as a
+    /// key-repair prompt.
+    ///
+    /// [`CoreWallet::sign_message`]: crate::wallet::core::CoreWallet::sign_message
+    /// [`Signer`]: key_wallet::signer::Signer
+    /// [`SignerMethod::Digest`]: key_wallet::signer::SignerMethod::Digest
+    #[error("message signing failed for address {address}: {reason}")]
+    MessageSigningFailed { address: String, reason: String },
```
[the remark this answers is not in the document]
HashEngineering: Investigated in depth — the helper (`preserve_signer_key_unavailable_or`, shipped by #4183) exists and is the right tool where it is used, but it cannot apply at this call site, and the finding's premise does not hold: `sign_ecdsa` never produces a `Protocol(Generic(...))` to preserve.

The exact type chain: `CoreWallet::sign_message<S: Signer>` receives `S::Error`, which key-wallet bounds as `Display + Send + Sync + 'static` only — an opaque type with nothing structural to match. The sole production impl (`MnemonicResolverCoreSigner`) uses a typed enum that never stamps the key-unavailable marker; the marker is stamped only in the state-transition completion callback (`SignResult = Result<Vec<u8>, ProtocolError>`), a different signing surface from key-wallet's `sign_ecdsa`. Wiring the helper in produces `error[E0308]: expected enum dash_sdk::Error` — and a `contains(...)` fallback is the substring sniff the #4183 review explicitly rejected (test-pinned).

What f597866a5a does instead: documents the full type chain at both the producer and the FFI mapping, and adds `signer_key_unavailable_is_not_preserved_during_message_signing` — a signer double that emits the marker at position 0 of its own Display, pinning that it survives only mid-string in `reason`. The test is written to fail if an upstream change ever makes code 31 reachable here, with instructions to flip the assertion.

The real fix is upstream, one of: (a) `MnemonicResolverCoreSigner` rendering key-unavailable variants with the marker at position 0, after which a position-0 check here becomes principled; or (b) key-wallet tightening `Signer::Error` to something matchable. Both touch shared surfaces and deserve their own reviewed change.

packages/rs-platform-wallet/src/error.rs
```diff
@@ -108,6 +108,81 @@ pub enum PlatformWalletError {
...
+    /// unreachable with any signer that ships today — the production mnemonic
+    /// resolver advertises `Digest` — so it does not warrant a new FFI code and
+    /// the host mirror-enum churn that follows one.
+    ///
+    /// Deliberately NOT given a dedicated FFI code: a signer that reports its
+    /// typed key-unavailable completion carries the stable machine prefix in
+    /// its `Display`, which the FFI conversion's catch-all promotes to
+    /// `ErrorSigningKeyUnavailable`. The remaining causes are genuine internal
+    /// invariant breaks and should surface as unknown rather than as a
+    /// key-repair prompt.
```
thepastaclaw: **🟡 Suggestion: MessageSigningFailed doc claims a code-31 promotion the FFI layer explicitly does not perform**

This doc comment says a signer's typed key-unavailable completion "carries the stable machine prefix in its Display, which the FFI conversion's catch-all promotes to ErrorSigningKeyUnavailable." That is the opposite of the current, carefully-documented behavior: the FFI conversion at packages/rs-platform-wallet-ffi/src/error.rs:466-528 explicitly states 'MessageSigningFailed is deliberately NOT matched, so it falls to the ErrorUnknown catch-all below' and spends ~40 lines explaining exactly why a signer's key-unavailable failure cannot currently reach code 31 during message signing. The sign_message.rs implementation in this same crate (lines 216-220) says the same thing: 'Every signer failure becomes MessageSigningFailed (→ ErrorUnknown), including a key-unavailable one. That is a known limitation, not an oversight.' The new regression test signer_key_unavailable_is_not_preserved_during_message_signing pins this exact behavior. Since hosts branch on the numeric FFI code to decide between a key-repair UX flow and a generic-error flow, this stale claim on the public Rust type is misleading to anyone reading the type docs without also reading the FFI conversion source.

```suggestion
    /// Deliberately NOT given a dedicated FFI code: `Signer::Error` is generic
    /// and bounded only by `Display`, so it cannot be classified structurally
    /// here. Signer failures — including a signer-reported key-unavailable
    /// completion — remain `MessageSigningFailed` and fall through to
    /// `ErrorUnknown`. Only `MessageSigningKeyUnavailable` (address resolution
    /// failing before a signer is ever invoked) reaches FFI code 31. See the
    /// `MessageSigningFailed` arm's NOTE in `platform-wallet-ffi`'s error
    /// conversion for the full chain and the upstream change that would be
    /// needed to close this gap.
```

<sub>source: ['codex']</sub>

thepastaclaw (commented):

At head f597866a, all five prior findings from the previous review round are confirmed FIXED against the actual source: the digest-capability guard runs before sign_ecdsa, the FFI clones the wallet out of HandleStorage before block_on/the host callback, a direct null-pointer/zero-length regression test exists at the FFI boundary, the Kotlin/Swift docs correctly describe UTF-8 byte length, and Kotlin's public signMessage rejects unpaired UTF-16 surrogates before any native call. All 8 platform-wallet sign_message tests and 2 platform-wallet-ffi sign_message tests pass locally, and clippy shows no new warnings in files touched by this PR. One new latest-delta finding from this round: the MessageSigningFailed doc comment in platform-wallet/src/error.rs still claims the FFI catch-all promotes signer key-unavailable completions to code 31, directly contradicted by the FFI conversion's own explicit comment (which says MessageSigningFailed is deliberately left unmapped and always falls to ErrorUnknown) and by sign_message.rs's own inline note added in this same PR. This is a self-contradictory public doc, not a functional defect, so it does not block the PR.

### Review provenance
- Sonnet reviewers: `claude-sonnet-5` — general (completed), `claude-sonnet-5` — security-auditor (completed), `claude-sonnet-5` — rust-quality (failed), `claude-sonnet-5` — ffi-engineer (completed), `claude-sonnet-5` — rust-quality (completed)

🟡 1 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet/src/error.rs:173-178: MessageSigningFailed doc claims a code-31 promotion the FFI layer explicitly does not perform
  This doc comment says a signer's typed key-unavailable completion "carries the stable machine prefix in its Display, which the FFI conversion's catch-all promotes to ErrorSigningKeyUnavailable." That is the opposite of the current, carefully-documented behavior: the FFI conversion at packages/rs-platform-wallet-ffi/src/error.rs:466-528 explicitly states 'MessageSigningFailed is deliberately NOT matched, so it falls to the ErrorUnknown catch-all below' and spends ~40 lines explaining exactly why a signer's key-unavailable failure cannot currently reach code 31 during message signing. The sign_message.rs implementation in this same crate (lines 216-220) says the same thing: 'Every signer failure becomes MessageSigningFailed (→ ErrorUnknown), including a key-unavailable one. That is a known limitation, not an oversight.' The new regression test signer_key_unavailable_is_not_preserved_during_message_signing pins this exact behavior. Since hosts branch on the numeric FFI code to decide between a key-repair UX flow and a generic-error flow, this stale claim on the public Rust type is misleading to anyone reading the type docs without also reading the FFI conversion source.
```
</details>

packages/rs-platform-wallet/src/wallet/core/sign_message.rs
thepastaclaw: Resolved in this update — _Honor the signer's digest capability before signing_ no longer present.

packages/rs-platform-wallet-ffi/src/core_wallet/sign_message.rs
thepastaclaw: Resolved in this update — _Cover the null-pointer empty-message ABI contract_ no longer present.

packages/rs-unified-sdk-jni/src/wallet_manager.rs
thepastaclaw: Resolved in this update — _Do not silently alter malformed Java UTF-16 messages_ no longer present.

packages/rs-platform-wallet-ffi/src/core_wallet/sign_message.rs
thepastaclaw: Resolved in this update — _Release the global Core-wallet storage lock before signing_ no longer present.

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/ManagedPlatformWallet.kt
thepastaclaw: Resolved in this update — _Document the signed message's UTF-8 byte length_ no longer present.

packages/rs-platform-wallet/src/error.rs
HashEngineering: Addressed in 09a27ac60d — adopted your wording with two additions: intra-doc links to `MessageSigningKeyUnavailable` and `Signer::Error`, and a pointer to the FFI conversion NOTE carrying the full type chain.

thepastaclaw (approved):

Carried-forward prior findings: the five earlier implementation and API findings in the review history remain fixed at exact head 09a27ac60db94baf942d2123817f85c8eb8d2371, and the sole active prior finding about the MessageSigningFailed documentation is now fixed as well. New latest-delta findings: none; the delta changes only that Rust doc comment, which now agrees with the FFI mapping and regression test. `cargo doc -p platform-wallet --no-deps` completed successfully, and none of its pre-existing warnings involve the edited documentation or links.

### Review provenance
- Sonnet reviewers: `claude-sonnet-5` — general (completed), `claude-sonnet-5` — security-auditor (completed), `claude-sonnet-5` — rust-quality (failed), `claude-sonnet-5` — ffi-engineer (completed), `claude-sonnet-5` — rust-quality (failed), `claude-sonnet-5` — rust-quality (completed)

packages/rs-platform-wallet/src/error.rs
thepastaclaw: Resolved in `09a27ac` — _MessageSigningFailed doc claims a code-31 promotion the FFI layer explicitly does not perform_ no longer present.

QuantumExplorer:
superceeded

PR #4265, opened 2026-08-02
fix(drive)!: make shared-prefix aggregate indexes insertable at protocol v14

### LINKED ISSUE #3960: Data contract with shared-prefix count+sum and range-count indexes registers but document inserts fail
A data contract can define two indexes on the same document type that share a prefix but require incompatible Drive tree wrapping:

- `[resourceId]` with `countable: "countable"` and `summable: "rating"`
- `[resourceId, rating]` with `countable: "countable"` and `rangeCountable: true`

The contract is accepted at registration, but inserting any document for that type fails when Drive tries to materialize the index trees.

This appears to be a contract validation gap or an unsupported Drive layout case. The bad state is reachable only after publishing the contract, and index definitions are immutable afterward.

## Example Contract Shape

```json
{
  "review": {
    "type": "object",
    "documentsMutable": true,
    "documentsKeepHistory": true,
    "canBeDeleted": false,
    "properties": {
      "resourceId": {
        "type": "string",
        "minLength": 1,
        "maxLength": 63,
        "position": 0
      },
      "rating": {
        "type": "integer",
        "minimum": 1,
        "maximum": 5,
        "position": 1
      },
      "reviewText": {
        "type": "string",
        "maxLength": 1000,
        "position": 2
      }
    },
    "required": ["$createdAt", "$updatedAt", "resourceId", "rating"],
    "additionalProperties": false,
    "indices": [
      {
        "name": "ownerAndResource",
        "unique": true,
        "properties": [{ "$ownerId": "asc" }, { "resourceId": "asc" }]
      },
      {
        "name": "ownerReviews",
        "properties": [{ "$ownerId": "asc" }, { "$updatedAt": "asc" }]
      },
      {
        "name": "resourceRatingAggregate",
        "properties": [{ "resourceId": "asc" }],
        "countable": "countable",
        "summable": "rating"
      },
      {
        "name": "resourceRatingDistribution",
        "properties": [{ "resourceId": "asc" }, { "rating": "asc" }],
        "countable": "countable",
        "rangeCountable": true
      }
    ]
  }
}
```

## Observed Behavior

1. Contract registration succeeds.
2. Creating a `review` document fails during Drive insertion.

Reported decoded error:

`storage: drive: not supported error: NotCountedOrSummed-wrapping is only supported for the six sum-bearing tree variants — see for_known_path_key_empty_not_summed_tree.`

## Expected Behavior

Either:

1. Contract validation rejects this index combination before publish, or
2. Drive supports this index layout during insert.

The current behavior is the worst case: the unsupported layout is accepted at publish time and fails later on every document write.

## Root Cause Notes

DPP builds shared index levels and attaches index aggregation metadata only where each index terminates. It rejects exact duplicate index paths, but does not appear to validate cross-index aggregation compatibility for shared prefixes.

Relevant code:

- `packages/rs-dpp/src/data_contract/document_type/index_level/mod.rs`
  `IndexLevel::try_from_indices_v0` builds the shared prefix tree and only rejects duplicate terminal levels.
- `packages/rs-dpp/src/data_contract/document_type/index/mod.rs`
  Index validation enforces local invariants like `rangeCountable` requiring `countable`.
- `packages/rs-dpp/src/data_contract/document_type/class_methods/try_from_schema/v2/mod.rs`
  Summable validation checks same summable property, integer type, and required field, but not shared-prefix tree compatibility.

Drive insertion then derives this layout:

- `resourceRatingAggregate` terminates at `resourceId` with count+sum, so the `resourceId/<value>` value tree is a `CountSumTree`.
- `resourceRatingDistribution` continues below the same `resourceId/<value>` path.
- Its `rating` property-name continuation has `rangeCountable: true` and no `rangeSummable`, so it is a `ProvableCountTree`.
- Because the parent is a `CountSumTree`, Drive tries to wrap the child with `NotCountedOrSummed`.
- `NotCountedOrSummed` currently accepts only sum-bearing inner tree variants, so `ProvableCountTree` is rejected.

Relevant Drive code:

- `packages/rs-drive/src/drive/document/insert/add_indices_for_index_level_for_contract_operations/v1/mod.rs`
  Derives `property_name_tree_type` as `ProvableCountTree` for `(range_countable=true, range_summable=false)`, then wraps child continuation trees when the parent value tree aggregates.
- `packages/rs-drive/src/fees/op.rs`
  `wrap_in_non_aggregated_for_parent_tree_type` dispatches count+sum parents to `for_known_path_key_empty_not_counted_or_summed_tree`, which rejects non-sum-bearing inner tree variants with the observed error.

## Test Coverage Gap

There are tests for summable indexes, range-countable/range-summable tree variants, and keep-history + summable behavior, but I could not find coverage for this specific cross-index case:

- one index ending at prefix `[a]` with count+sum
- another index continuing as `[a, b]` with count+rangeCountable only

A regression test should cover either the validation decision or Drive layout support:

- If the intended behavior is rejection: schema/contract creation should reject the index pair.
- If the intended behavior is support: Drive should successfully insert a document and materialize the expected trees.

QuantumExplorer:

A data contract can declare an aggregating (`countable` / `summable`) index that terminates at a property which is also the prefix of a compound index — e.g. a summable `[a]` next to `[a, b]`. Index levels merge into one trie keyed by property name, so the compound continuation tree for `b` hangs *inside* the aggregating per-value trees of `[a]`. The contract registers successfully (contract validation performs no cross-index shape analysis), but for most flag combinations every document insert is then rejected: grovedb's `NotSummed`/`NotCountedOrSummed` wrappers only accept sum-bearing inners, the `NonCounted` helper only accepted count-ish inners, and provable count-bearing value trees reject count-suppressed children entirely.

### Example: count + sum aggregate next to a compound index

A star-rating app wants "how many reviews does this resource have, and what do they sum to" as a one-lookup aggregate, plus the usual compound index for listing reviews of a resource by rating:

```json
{
  "review": {
    "type": "object",
    "properties": {
      "resourceId": { "type": "string", "maxLength": 63, "position": 0 },
      "rating":     { "type": "integer", "minimum": 1, "maximum": 5, "position": 1 }
    },
    "required": ["resourceId", "rating"],
    "additionalProperties": false,
    "indices": [
      {
        "name": "resourceRatingAggregate",
        "properties": [{ "resourceId": "asc" }],
        "countable": "countable",
        "summable": "rating"
      },
      {
        "name": "resourceRatingDistribution",
        "properties": [{ "resourceId": "asc" }, { "rating": "asc" }]
      }
    ]
  }
}
```

Today this contract registers fine, and then **every** `review` insert fails with:

```
drive: not supported error: NotCountedOrSummed-wrapping is only supported for the
six sum-bearing tree variants — see `for_known_path_key_empty_not_summed_tree`
```

because both indexes merge into one trie: the `rating` continuation tree of `resourceRatingDistribution` must live *inside* the per-resource `CountSumTree` that `resourceRatingAggregate` creates, and Drive had no legal way to make it contribute zero to that tree's count. (Three registered testnet contracts are bricked in exactly this state.)

At v14, inserting two reviews for `resource-1` (ratings 5 and 3) produces:

```
review
└─ resourceId                                (property-name tree)
   └─ "resource-1"      CountSumTree(count: 2, sum: 8)   ← the one-lookup aggregate
      ├─ [0]            ref bucket: doc1, doc2           ← the only contributor
      └─ rating         NonCounted(Tree)                 ← continuation, contributes 0
         ├─ "3" ─ [0] ─ doc2 ref
         └─ "5" ─ [0] ─ doc1 ref
```

The `resource-1` element answers count/sum point lookups directly, and `[resourceId, rating]` queries walk the `rating` subtree as usual. Deleting a review subtracts exactly its contribution (delete doc2 → `count: 1, sum: 5`), and deleting the last one cleans the whole branch away.

### Example: provable range aggregates next to a compound index

The `range*` flags buy per-node commitments in the *property-name* tree, so range aggregates verify in O(log n). A sales ledger wants "provable count + total of sales over any day range", plus a compound index for drilling into a day by region:

```json
{
  "sale": {
    "type": "object",
    "properties": {
      "day":    { "type": "integer", "minimum": 0, "position": 0 },
      "region": { "type": "string", "maxLength": 32, "position": 1 },
      "amount": { "type": "integer", "minimum": 0, "maximum": 1000000, "position": 2 }
    },
    "required": ["day", "region", "amount"],
    "additionalProperties": false,
    "indices": [
      {
        "name": "dailySales",
        "properties": [{ "day": "asc" }],
        "countable": "countable",
        "summable": "amount",
        "rangeCountable": true,
        "rangeSummable": true
      },
      {
        "name": "dailySalesByRegion",
        "properties": [{ "day": "asc" }, { "region": "asc" }]
      }
    ]
  }
}
```

Today this registers and then rejects every insert, same as the first example. At v14, after three sales — day 100: (US, 500), (EU, 200); day 101: (EU, 250) — the layout is:

```
sale
└─ day                 ProvableCountProvableSumTree     ← every merk node commits (count, sum):
   │                                                      "count + total for days 100..=200"
   │                                                      is one O(log n) range proof
   ├─ "100"     CountSumTree(count: 2, sum: 700)      ← demoted (would be PCPST without the
   │  ├─ [0]    ref bucket: sale1, sale2                 compound sibling): provable count
   │  └─ region NonCounted(Tree)                         parents refuse zero-contribution
   │     ├─ "EU" ─ [0] ─ sale2 ref                       children, so the per-day trees drop
   │     └─ "US" ─ [0] ─ sale1 ref                       to the element-carried variant
   └─ "101"     CountSumTree(count: 1, sum: 250)
      ├─ [0]    ref bucket: sale3
      └─ region NonCounted(Tree)
         └─ "EU" ─ [0] ─ sale3 ref
```

The provable machinery lives one level up from the demotion and is untouched: each per-day `CountSumTree` element feeds its (count, sum) into the `day` tree's per-node commitments exactly as a provable child would, so `AggregateCountOnRange` / `AggregateSumOnRange` over day ranges keep their O(log n) proofs. What the demotion changes is only the *inside* of each per-day tree — per-value point lookups read (count, sum) off the element either way, and nothing range-aggregates over a value tree's own children. Without the compound sibling there is no demotion: the per-day value trees stay `ProvableCountProvableSumTree`, bit-identical to v13.

The empirically pinned pre-v14 matrix: count-only value trees accept only non-sum continuations, sum-bearing value trees accept only sum-bearing continuations, and everything else errors. (Provable count-bearing value trees accept sum-bearing continuations only through an unenforced path: grovedb's wrapper-vs-provable batch guard fires solely when the parent merk pre-exists, and the walker always creates parent and wrapped child in one batch.)

A full scan of registered contracts via the platform-explorer API found three bricked contracts on testnet (all matching the issue's review schema — registered, but every insert fails), none on mainnet, and zero contracts on either network using range-flagged prefixes with compound siblings.

Fixes #3960

## What was done?

Stacked on #4267, which introduces the protocol version 14 gate. This PR consumes it: `PLATFORM_V14.drive` moves to `DRIVE_VERSION_V9` → `DRIVE_DOCUMENT_METHOD_VERSIONS_V4`; all pre-v14 walker code stays byte-identical.

- **v2 index walkers** (insert/delete × top-level/recursive). Tree types now derive through a single shared helper, `index_level_tree_types.rs`, so the insert walkers and the delete walkers' estimation layer info cannot drift. The helper demotes provable count-bearing value trees (`ProvableCountSumTree` / `ProvableCountProvableSumTree`) to `CountSumTree` when compound continuations exist, since grovedb rejects count-suppressed children under provable count parents by design and a plain continuation has no legal wrapper there at all.
- **Completed zero-contribution wrapper matrix** (`LowLevelDriveOperation::for_known_path_key_empty_tree_contributing_zero_to_parent`): non-sum continuations under `CountSumTree` parents get `Element::NonCounted`; non-sum continuations under sum-only parents are inserted **unwrapped** (a non-sum child contributes 0 to a sum naturally); count-only parents accept any inner via an extended `NonCounted` builder. The v0 dispatcher and helpers are untouched.

Every value tree keeps the invariant that its count/sum equals exactly the `[0]` ref-bucket contribution, never the structural overhead of sibling continuations. Shapes insertable at v13 produce bit-identical operations; the one intentional difference is that provable value trees whose continuations were all sum-bearing get `CountSumTree` value trees for values first seen at v14+ (readers are indifferent — both variants serialize their aggregates into the element and contribute identically to the property-name tree's per-node commitments). No such shape exists on any network today.

**Reviewer note — a deliberate design choice to weigh in on:** instead of demoting, we could keep PCPST value trees and upgrade plain continuations to a sum-bearing inner, but with grovedb v5.0.1 that permanently depends on the unenforced in-batch guard path; making it official would need a grovedb-side policy change (allowing `NonCounted`/`NotCountedOrSummed` under provable count parents, mirroring how `NotSummed` is already allowed under provable sum parents). No reader of per-node commitments *inside* value trees was found — range aggregates walk the property-name tree above, point lookups prove the element — so the demotion loses nothing observable today. Happy to switch if the ranked-aggregates roadmap wants provable value trees preserved.

## How Has This Been Tested?

- `shared_prefix_aggregation_e2e_tests.rs` rewritten (previously `#[ignore]`d): a 32-combination matrix (8 aggregating prefix flag-sets × 4 child flag-sets) proving at v14: contract applies, two documents insert, the value tree carries exactly (count = 2, sum = ratings) with the exact expected element type, the continuation tree lands with the exact expected wrapper, deleting one document subtracts exactly its contribution, and deleting the last document cleans the trees away entirely.
- A `..._frozen_at_v13` suite pins the empirical pre-v14 accept/reject matrix per combination, so the consensus-locked v1 walkers cannot drift.
- Full suites green: rs-drive lib (3251) + integration tests including the deterministic root-hash pins, dpp (3806), drive-abci lib (2629), `cargo check --workspace --all-targets`, all-features clippy, and the `--no-default-features --features verify` build of rs-drive.

## Breaking Changes

Consensus-breaking by design: first consumer of the v14 gate from #4267. At activation, shared-prefix aggregate index shapes become insertable (previously every such insert failed), and provable count-bearing value trees with compound continuations are created as `CountSumTree` for newly seen values. Pre-v14 behavior is bit-identical and pinned by the frozen-at-v13 test.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Improved document index insertion and deletion for shared-prefix aggregate indexes.
  * Added support for continuation trees and accurate aggregate values across nested indexes.
  * Introduced updated platform-version handling for protocol v14.

* **Bug Fixes**
  * Corrected index references, null-field state, and zero-contribution behavior during document removal.
  * Added validation for oversized indexed fields and fee-estimation overflow conditions.

* **Tests**
  * Expanded end-to-end coverage for aggregate indexes, insertions, deletions, compatibility, and continuation behavior.

### DIFF
--- a/packages/rs-drive/src/drive/contract/insert/insert_contract/v0/tests/shared_prefix_aggregation_e2e_tests.rs
+++ b/packages/rs-drive/src/drive/contract/insert/insert_contract/v0/tests/shared_prefix_aggregation_e2e_tests.rs
@@ -16,17 +16,27 @@
 //! when continuations exist (grovedb's stated design rejects
 //! count-suppressed children under provable count parents; pre-v14
 //! those shapes only worked at all through an unenforced in-batch
-//! creation path).
+//! creation path). Key-changing updates materialize index branches
+//! through their own walker, which is bumped to v1 at v14 with the
+//! same demotion + zero-contribution treatment.
 //!
-//! Two suites:
-//! - `..._insert_and_delete_at_latest` proves the entire matrix
-//!   inserts at v14+, that the value trees carry exactly the `[0]`
+//! Suites:
+//! - `..._insert_update_delete_at_v14` proves the entire matrix
+//!   inserts at v14, that the value trees carry exactly the `[0]`
 //!   ref-bucket's (count, sum) — never the structural overhead of
-//!   continuations — through insert AND delete, and that the
-//!   continuation trees land with the exact expected wrapper.
+//!   continuations — through insert, key-changing update (which
+//!   materializes a fresh branch via the update walker), and delete,
+//!   and that the continuation trees land with the exact expected
+//!   wrapper on both the insert- and update-materialized branches.
 //! - `..._frozen_at_v13` pins the pre-v14 behavior (which combos
 //!   insert, which error) so the consensus-locked v1 walkers cannot
 //!   drift.
+//! - `..._estimated_costs_do_not_write_state` exercises the v2
+//!   walkers' stateless-estimation branches (`apply: false`).
+//! - `..._v13_and_v14_layouts_coexist` proves a provable value tree
+//!   created at v13 keeps working at v14 next to newly-demoted
+//!   `CountSumTree` siblings, through inserts and full-cleanup
+//!   deletes.
 
 use crate::drive::Drive;
 use crate::util::object_size_info::DocumentInfo::DocumentRefInfo;
@@ -104,7 +114,7 @@ impl IndexFlags {
         summable: true,
         range_summable: true,
     };
-    const RANGE_COUNT_SUM: Self = Self {
+    const RANGE_COUNT_RANGE_SUM: Self = Self {
         countable: true,
         range_countable: true,
         summable: true,
@@ -178,13 +188,13 @@ fn all_cases() -> Vec<SharedPrefixCase> {
         ("range_sum", IndexFlags::RANGE_SUM),
         ("count_sum_range_count", IndexFlags::COUNT_SUM_RANGE_COUNT),
         ("count_sum_range_sum", IndexFlags::COUNT_SUM_RANGE_SUM),
-        ("range_count_sum", IndexFlags::RANGE_COUNT_SUM),
+        ("range_count_range_sum", IndexFlags::RANGE_COUNT_RANGE_SUM),
     ];
     let children: [(&'static str, IndexFlags); 4] = [
         ("plain", IndexFlags::PLAIN),
         ("range_count", IndexFlags::RANGE_COUNT),
         ("range_sum", IndexFlags::RANGE_SUM),
-        ("range_count_sum", IndexFlags::RANGE_COUNT_SUM),
+        ("range_count_range_sum", IndexFlags::RANGE_COUNT_RANGE_SUM),
     ];
 
     let mut cases = Vec::new();
@@ -331,11 +341,20 @@ fn apply_contract(
         .map_err(|error| format!("{error:?}"))
 }
 
+/// Inserts a review document. `apply: false` runs the stateless
+/// estimation path instead of writing state.
+///
+/// Seeds must be distinct across calls: the `ownerAndResource` index
+/// is unique on `($ownerId, resourceId)` and all documents share a
+/// per-test `resourceId`, so only the seed-derived random owner keeps
+/// the inserts from colliding on that index.
 fn insert_review_document(
     drive: &Drive,
     contract: &DataContract,
     seed: u64,
+    resource: &str,
     rating: u8,
+    apply: bool,
     platform_version: &PlatformVersion,
 ) -> Result<Document, String> {
     let document_type = contract
@@ -345,10 +364,7 @@ fn insert_review_document(
         .random_document(Some(seed), platform_version)
         .expect("random review document");
     let mut properties = BTreeMap::new();
-    properties.insert(
-        "resourceId".to_string(),
-        Value::Text("resource-1".to_string()),
-    );
+    properties.insert("resourceId".to_string(), Value::Text(resource.to_string()));
     properties.insert("rating".to_string(), Value::U8(rating));
     properties.insert(
         "reviewText".to_string(),
@@ -372,7 +388,7 @@ fn insert_review_document(
             },
             false,
             BlockInfo::default(),
-            true,
+            apply,
             None,
             platform_version,
             None,
@@ -381,17 +397,77 @@ fn insert_review_document(
         .map_err(|error| format!("{error:?}"))
 }
 
+/// Moves a stored review document to a new `resourceId` through the
+/// document UPDATE path — the update walker materializes the new
+/// index branch itself, which is exactly the surface the v14
+/// update-walker bump covers.
+fn update_review_document_resource(
+    drive: &Drive,
+    contract: &DataContract,
+    document: &Document,
+    new_resource: &str,
+    platform_version: &PlatformVersion,
+) -> Result<Document, String> {
+    let document_type = contract
+        .document_type_for_name("review")
+        .expect("review document type exists");
+    let mut updated = document.clone();
+    updated.set("resourceId", Value::Text(new_resource.to_string()));
+    updated.set_revision(updated.revision().map(|revision| revision + 1));
+
+    drive
+        .update_document_for_contract(
+            &updated,
+            contract,
+            document_type,
+            None,
+            BlockInfo::default(),
+            true,
+            None,
+            None,
+            platform_version,
+            None,
+        )
+        .map(|_| updated)
+        .map_err(|error| format!("{error:?}"))
+}
+
+fn delete_review_document(
+    drive: &Drive,
+    contract: &DataContract,
+    document: &Document,
+    platform_version: &PlatformVersion,
+) -> Result<(), String> {
+    drive
+        .delete_document_for_contract(
+            document.id(),
+            contract,
+            "review",
+            BlockInfo::default(),
+            true,
+            None,
+            platform_version,
+            None,
+        )
+        .map(|_| ())
+        .map_err(|error| format!("{error:?}"))
+}
+
 /// Reads the raw element at `@/contract/1/review/resourceId` +
-/// `"resource-1"` — the prefix index's value tree — or None if absent.
-fn probe_value_tree(drive: &Drive, contract: &DataContract) -> Option<Element> {
-    probe(drive, contract, &["resourceId"], b"resource-1")
+/// `resource` — the prefix index's value tree — or None if absent.
+fn probe_value_tree(drive: &Drive, contract: &DataContract, resource: &str) -> Option<Element> {
+    probe(drive, contract, &["resourceId"], resource.as_bytes())
 }
 
 /// Reads the raw element at
-/// `@/contract/1/review/resourceId/resource-1` + `"rating"` — the
+/// `@/contract/1/review/resourceId/<resource>` + `"rating"` — the
 /// compound index's continuation property-name tree.
-fn probe_continuation_tree(drive: &Drive, contract: &DataContract) -> Option<Element> {
-    probe(drive, contract, &["resourceId", "resource-1"], b"rating")
+fn probe_continuation_tree(
+    drive: &Drive,
+    contract: &DataContract,
+    resource: &str,
+) -> Option<Element> {
+    probe(drive, contract, &["resourceId", resource], b"rating")
 }
 
 fn probe(drive: &Drive, contract: &DataContract, sub_path: &[&str], key: &[u8]) -> Option<Element> {
@@ -421,51 +497,55 @@ fn probe(drive: &Drive, contract: &DataContract, sub_path: &[&str], key: &[u8]) 
         .expect("probe must succeed")
 }
 
-/// Asserts the value tree carries exactly the `[0]` bucket's
+/// Checks the value tree carries exactly the `[0]` bucket's
 /// (count, sum) on the axes the prefix aggregates — with the
 /// continuation demotion, always the non-provable variant.
-fn assert_value_tree_aggregates(
-    case_name: &str,
+fn check_value_tree_aggregates(
     element: &Element,
     axes: ParentAxes,
     expected_count: u64,
     expected_sum: i64,
-) {
+) -> Result<(), String> {
     match (axes, element) {
         (ParentAxes::Count, Element::CountTree(_, count, _)) => {
-            assert_eq!(
-                *count, expected_count,
-                "{case_name}: value tree count must equal the ref-bucket contribution"
-            );
+            if *count != expected_count {
+                return Err(format!(
+                    "value tree count {count} != expected ref-bucket contribution {expected_count}"
+                ));
+            }
         }
         (ParentAxes::Sum, Element::SumTree(_, sum, _)) => {
-            assert_eq!(
-                *sum, expected_sum,
-                "{case_name}: value tree sum must equal the ref-bucket contribution"
-            );
+            if *sum != expected_sum {
+                return Err(format!(
+                    "value tree sum {sum} != expected ref-bucket contribution {expected_sum}"
+                ));
+            }
         }
         (ParentAxes::CountSum, Element::CountSumTree(_, count, sum, _)) => {
-            assert_eq!(
-                (*count, *sum),
-                (expected_count, expected_sum),
-                "{case_name}: value tree (count, sum) must equal the ref-bucket contribution"
-            );
+            if (*count, *sum) != (expected_count, expected_sum) {
+                return Err(format!(
+                    "value tree (count, sum) ({count}, {sum}) != expected ref-bucket \
+                     contribution ({expected_count}, {expected_sum})"
+                ));
+            }
         }
-        (_, other) => panic!(
-            "{case_name}: unexpected value tree element (demotion or axis mismatch): {other:?}"
-        ),
+        (_, other) => {
+            return Err(format!(
+                "unexpected value tree element (demotion or axis mismatch): {other:?}"
+            ))
+        }
     }
+    Ok(())
 }
 
-/// Asserts the continuation property-name tree landed with the exact
+/// Checks the continuation property-name tree landed with the exact
 /// wrapper the zero-contribution matrix specifies for this
 /// parent-axes / child-tree-type combination.
-fn assert_continuation_shape(
-    case_name: &str,
+fn check_continuation_shape(
     element: &Element,
     axes: ParentAxes,
     child_flags: &IndexFlags,
-) {
+) -> Result<(), String> {
     let expected_inner = child_flags.continuation_tree_type();
     let child_is_sum_bearing = child_flags.is_sum_bearing_continuation();
 
@@ -480,7 +560,7 @@ fn assert_continuation_shape(
         // children are stored unwrapped (they contribute 0 naturally).
         (ParentAxes::Sum, true, Element::NotSummed(inner)) => inner,
         (ParentAxes::Sum, false, plain) => plain,
-        (_, _, other) => panic!("{case_name}: unexpected continuation wrapper: {other:?}"),
+        (_, _, other) => return Err(format!("unexpected continuation wrapper: {other:?}")),
     };
 
     let inner_tree_type = match inner {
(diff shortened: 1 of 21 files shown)


thepastaclaw:
ℹ️ Review skipped (commit 7ca8f9cf)
_Last checked: 2026-08-02 20:30 UTC_

packages/rs-platform-version/src/version/drive_versions/drive_document_method_versions/v4.rs
```diff
@@ -0,0 +1,110 @@
...
+            add_contested_vote_subtree_for_non_identities_operations: 0,
+        },
+        update: DriveDocumentUpdateMethodVersions {
+            add_update_multiple_documents_operations: 0,
+            update_document_for_contract: 0,
+            update_document_for_contract_apply_and_add_to_operations: 0,
+            update_document_for_contract_id: 0,
+            update_document_for_contract_operations: 0,
+            update_document_with_serialization_for_contract: 0,
+            update_serialized_document_for_contract: 0,
```
thepastaclaw, line 98: **🔴 Blocking: Apply the v14 index layout to key-changing document updates**

Protocol v14 activates the corrected v2 insert and delete walkers, but `update_document_for_contract_operations` remains on v0. A key-changing update can independently materialize a previously unseen index branch through that implementation, which still derives the pre-v14 value-tree type and inserts continuation property-name trees with the unwrapped `batch_insert_empty_tree_if_not_exists` helper. For example, with a count-and-sum `[resourceId]` index alongside plain `[resourceId, rating]`, moving one document to a new `resourceId` creates an unwrapped structural `rating` child under the new `CountSumTree`; that child contributes to the count, so the authenticated aggregate can report two entries for one document. Other combinations can leak continuation sums or retain a provable parent that is incompatible with the v14 wrapper rules. Add a v14-dispatched update implementation that shares the continuation-demotion and zero-contribution wrapper logic with the v2 walkers, while preserving v0 for earlier protocols, and cover a shared-prefix key-changing update end to end.

<sub>source: ['codex']</sub>

packages/rs-drive/src/util/grove_operations/batch_insert_empty_tree_if_not_exists/mod.rs
```diff
@@ -111,6 +129,53 @@ impl Drive {
...
+    /// that contribute zero naturally, e.g. a plain continuation under a
+    /// sum-only value tree).
+    ///
+    /// Not dispatched on a drive feature version of its own: the platform
+    /// gate is carried by its only callers, the v14+ v2 index walkers, so
+    /// pre-v14 behavior can't reach this path.
+    #[allow(clippy::too_many_arguments)]
+    pub fn batch_insert_empty_tree_contributing_zero_to_aggregating_parent_if_not_exists<
+        const N: usize,
+    >(
```
thepastaclaw: **🟡 Suggestion: Keep the v14-only insertion helper crate-private**

This helper directly invokes the v0 body with `ContributingZeroToParent` and performs no version dispatch of its own. Its documented safety invariant depends on calls coming exclusively from the protocol-gated v14 v2 walkers, but `pub fn` exposes it to downstream crates that can bypass that gate. The only current caller is inside `rs-drive`, so crate-private visibility enforces the intended ownership boundary without restricting legitimate use.

```suggestion
    pub(crate) fn batch_insert_empty_tree_contributing_zero_to_aggregating_parent_if_not_exists<
        const N: usize,
    >(
```

<sub>source: ['codex']</sub>

packages/rs-drive/src/drive/contract/insert/insert_contract/v0/tests/shared_prefix_aggregation_e2e_tests.rs
```diff
@@ -242,124 +360,257 @@ fn insert_review_document_for_case(case: &SharedPrefixCase) -> SharedPrefixOutco
...
                 document_type,
             },
             false,
             BlockInfo::default(),
             true,
             None,
-            pv,
+            platform_version,
             None,
         )
```
thepastaclaw, line 395: **🟡 Suggestion: Exercise the stateless insert and delete estimation branches**

The matrix helper always inserts with `apply: true`, and both deletion calls also apply their operations. Consequently, the newly changed v2 stateless-estimation branches remain unexercised, including their post-demotion `tree_type`, `EstimatedSumTrees`, and stateless wrapper-operation construction. Add representative `apply: false` cases for demoted and non-demoted layouts, assert that fee estimation succeeds, and verify that each dry run leaves state unchanged.

<sub>source: ['codex']</sub>

packages/rs-drive/src/drive/contract/insert/insert_contract/v0/tests/shared_prefix_aggregation_e2e_tests.rs
```diff
@@ -242,124 +360,257 @@ fn insert_review_document_for_case(case: &SharedPrefixCase) -> SharedPrefixOutco
...
+        let drive = setup_drive_with_initial_state_structure(None);
+        let contract = build_review_contract(case.prefix_flags, case.child_flags)
+            .unwrap_or_else(|error| panic!("{}: contract must build: {error}", case.name));
+        apply_contract(&drive, &contract, platform_version)
+            .unwrap_or_else(|error| panic!("{}: contract must apply: {error}", case.name));
+
+        let first_document = insert_review_document(&drive, &contract, 1, 5, platform_version)
+            .unwrap_or_else(|error| panic!("{}: first insert must succeed: {error}", case.name));
+        let second_document = insert_review_document(&drive, &contract, 2, 3, platform_version)
+            .unwrap_or_else(|error| panic!("{}: second insert must succeed: {error}", case.name));
```
thepastaclaw: **🟡 Suggestion: Pin protocol v14 and cover values created before activation**

The activation matrix uses `PlatformVersion::latest()` rather than protocol 14, so once a later protocol is introduced it will no longer verify the exact v14-to-Drive-v9-to-v2-walker dispatch chain. The v13 suite also uses separate fresh databases and therefore does not test the mixed layout explicitly supported by this change: a provable count-sum value tree created at v13 remains in state while newly seen values use `CountSumTree` at v14. Pin the activation matrix with `PlatformVersion::get(14)` and add a single-database scenario that creates a `ProvableCountSumTree` or `ProvableCountProvableSumTree` value at v13, then inserts into and deletes from that shared prefix at v14.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

Carried-forward prior findings: the key-changing update path still bypasses the v14 shared-prefix layout and remains blocking; the helper-visibility, stateless-estimation coverage, and protocol-upgrade coverage suggestions also remain valid. The prior mixed-version estimation blocker is outdated because pinned GroveDB assigns identical feature and operation costs to all three count-sum tree variants. Genuinely new findings: none.

### Review provenance

<details>

```

In `packages/rs-platform-version/src/version/drive_versions/drive_document_method_versions/v4.rs`:
- [BLOCKING] packages/rs-platform-version/src/version/drive_versions/drive_document_method_versions/v4.rs:86-93: Apply the v14 index layout to key-changing document updates
  Protocol v14 activates the corrected v2 insert and delete walkers, but `update_document_for_contract_operations` remains on v0. A key-changing update can independently materialize a previously unseen index branch through that implementation, which still derives the pre-v14 value-tree type and inserts continuation property-name trees with the unwrapped `batch_insert_empty_tree_if_not_exists` helper. For example, with a count-and-sum `[resourceId]` index alongside plain `[resourceId, rating]`, moving one document to a new `resourceId` creates an unwrapped structural `rating` child under the new `CountSumTree`; that child contributes to the count, so the authenticated aggregate can report two entries for one document. Other combinations can leak continuation sums or retain a provable parent that is incompatible with the v14 wrapper rules. Add a v14-dispatched update implementation that shares the continuation-demotion and zero-contribution wrapper logic with the v2 walkers, while preserving v0 for earlier protocols, and cover a shared-prefix key-changing update end to end.

In `packages/rs-drive/src/util/grove_operations/batch_insert_empty_tree_if_not_exists/mod.rs`:
- [SUGGESTION] packages/rs-drive/src/util/grove_operations/batch_insert_empty_tree_if_not_exists/mod.rs:152-154: Keep the v14-only insertion helper crate-private
  This helper directly invokes the v0 body with `ContributingZeroToParent` and performs no version dispatch of its own. Its documented safety invariant depends on calls coming exclusively from the protocol-gated v14 v2 walkers, but `pub fn` exposes it to downstream crates that can bypass that gate. The only current caller is inside `rs-drive`, so crate-private visibility enforces the intended ownership boundary without restricting legitimate use.

In `packages/rs-drive/src/drive/contract/insert/insert_contract/v0/tests/shared_prefix_aggregation_e2e_tests.rs`:
- [SUGGESTION] packages/rs-drive/src/drive/contract/insert/insert_contract/v0/tests/shared_prefix_aggregation_e2e_tests.rs:334-379: Exercise the stateless insert and delete estimation branches
  The matrix helper always inserts with `apply: true`, and both deletion calls also apply their operations. Consequently, the newly changed v2 stateless-estimation branches remain unexercised, including their post-demotion `tree_type`, `EstimatedSumTrees`, and stateless wrapper-operation construction. Add representative `apply: false` cases for demoted and non-demoted layouts, assert that fee estimation succeeds, and verify that each dry run leaves state unchanged.
- [SUGGESTION] packages/rs-drive/src/drive/contract/insert/insert_contract/v0/tests/shared_prefix_aggregation_e2e_tests.rs:503-521: Pin protocol v14 and cover values created before activation
  The activation matrix uses `PlatformVersion::latest()` rather than protocol 14, so once a later protocol is introduced it will no longer verify the exact v14-to-Drive-v9-to-v2-walker dispatch chain. The v13 suite also uses separate fresh databases and therefore does not test the mixed layout explicitly supported by this change: a provable count-sum value tree created at v13 remains in state while newly seen values use `CountSumTree` at v14. Pin the activation matrix with `PlatformVersion::get(14)` and add a single-database scenario that creates a `ProvableCountSumTree` or `ProvableCountProvableSumTree` value at v13, then inserts into and deletes from that shared prefix at v14.
```
</details>

QuantumExplorer:
Review responses — addressed in 7ca8f9cf65:

**thepastaclaw blocker (key-changing updates bypass the v14 layout): confirmed and fixed.** The internal update walker does materialize index branches itself, with pre-demotion tree types and — worse — *unwrapped* continuation property-name trees. Added `update_document_for_contract_operations` v1 (gated at v14 via `DRIVE_DOCUMENT_METHOD_VERSIONS_V4`) that derives types through the shared continuation-demotion helper and wraps continuations through the zero-contribution batch helper; the estimation branch already delegated to the insert walkers. The v14 e2e matrix now runs insert → key-changing update → delete for all 32 combinations, asserting exact value-tree aggregates and wrapper shapes on the update-materialized branch.

**thepastaclaw suggestions: all applied.** The zero-contribution helper is now `pub(crate)` and dispatches the grove feature version like its siblings; the matrix pins `PlatformVersion::get(14)` instead of `latest()`; new suites cover the stateless-estimation branches (`apply: false` produces a fee and writes nothing) and a single-database scenario where a `ProvableCountSumTree` value tree created at v13 coexists with v14-demoted `CountSumTree` siblings through inserts and full-cleanup deletes.

**CodeRabbit actionable comments:**
- Insert-walker overflow message "on delete" wording: fixed in both v2 walkers.
- `for_known_path_key_empty_non_counted_any_tree` accepting sum-bearing inners: **intentionally kept.** grovedb v5.0.1's `Element::new_non_counted` rejects only wrapper-in-wrapper inners — sum-bearing trees are accepted — and count-only parents have no sum axis to pollute, so `NonCounted(ProvableSumTree)` under a `CountTree` value tree is exactly the shape a `[a]`-rangeCountable + `[a,b]`-rangeSummable contract needs. The e2e matrix exercises those cells end to end, and the new `zero_contribution_dispatcher_full_matrix` unit test pins every parent×inner cell.
- `build_op` closure indentation: this is rustfmt-stable output (`cargo fmt --all` reproduces it), so `--check` passes.

**CodeRabbit nitpicks:** applied the failure-collection pattern in the v14 suite, the distinct-seeds comment, the `RANGE_COUNT_RANGE_SUM` rename, the `changed in vN` marker rescope in `v9.rs`, the derivation/dispatcher invariant unit tests, the dispatcher matrix unit test, and wrapping-mode coverage in the batch helper tests. Skipped two: the shared empty-element-mapping extraction (it would touch the consensus-frozen v0 helpers' text for a pure refactor — the duplication is this codebase's deliberate versioned-module style), and `#[inline(always)]` → `#[inline]` (the v0/v1 siblings carry `#[inline(always)]`; keeping it for consistency).

Suites after the changes: rs-drive lib 3258, dpp 3807, drive-abci 2629 — all green.

QuantumExplorer:
Reviewed

PR #4266, opened 2026-08-02
feat: ranked aggregate indexes with provable top-K queries (protocol v14)

QuantumExplorer:

Platform's 4.0 aggregates can compute counts, sums and averages over document groups, but answering "which groups rank highest?" (e.g. *top 5 restaurants by average grade*) requires scanning every group — `O(n)` work with an `O(n)` proof, so it was never offered.

GroveDB's indexed-tree family (dashpay/grovedb#657, follow-ups in dashpay/grovedb#781) pairs each aggregate-bearing tree with per-axis ordered secondaries, making provable top-K reads `O(log n + k)`. This PR integrates that at the contract level and activates it at protocol v14.

## What was done?

**Contract grammar (rs-dpp, meta-schema v3, `document_type_schema >= 3`)**
- New index keywords `rankedCountable` / `rankedSummable` / `rankedAverageable`, each requiring its `range*` counterpart; parsed into `Index` / `IndexLevelTypeInfo`
- Restricted to single-property, non-unique, non-contested indexes; immutable on contract update (`find_first_ranked_change`); pre-v14 contracts get the same "unexpected property name" rejection as before

**Storage (rs-drive)**
- The index's terminal property-name tree upgrades to `ProvableCountIndexedTree` / `ProvableSumIndexedTree` / `ProvableCountProvableSumIndexedTree(axes)` — a byte-compatible mirror of the tree it replaces, so existing range-aggregate queries keep working on it
- Resolver `drive/document/ranked_index_tree_type.rs` used by contract insert/update and all document index walkers; three new `batch_insert_empty_provable_*_indexed_tree` grove-op helpers; estimation maps indexed trees to their non-indexed mirror weights (estimated ≥ actual verified per variant)

**Queries (rs-drive → rs-drive-abci → wire)**
- `query/drive_document_ranked_query/`: validation (`detect_ranked_mode`, versioned), executors over `indexed_{count,sum,avg}_top_k` and `prove_indexed_axis_top_k`, `MAX_RANKED_LIMIT = 100`; `verify/document_ranked/` wraps `GroveDb::verify_indexed_axis_top_k`
- drive-abci `document_query/v1`: `compute_aggregate_mode_and_check_limit` v1 routes a single `HAVING` ranking clause (`TOP(n)`/`BOTTOM(n)`/`MAX`/`MIN`) to `dispatch_ranked_v1`; protocol v13 keeps feature version 0 and still rejects every `HAVING`, so mixed-version networks agree
- Proto: additive only — `RankedEntry`/`RankedEntries` messages and a `ResultData.ranked` arm; requests reuse the existing `selects`/`group_by`/`having` fields of `GetDocumentsRequestV1`. Avg entries carry the grovedb fixed-point i128 (scale `AVG_FIXED_POINT_SCALE`, currently 10^19) as 16 BE bytes; clients divide

**Clients (rs-drive-proof-verifier, rs-sdk)**
- `DocumentRankedEntries: Fetch` with proof verification bound to the signed app hash via `verify_tenderdash_proof`; `ranking_having` constructor; empty-ranking-with-proof surfaces as an explicit error (grovedb has no absence envelope for indexed-axis proofs)

**Versioning (rs-platform-version)**
- Protocol v14: `CONTRACT_VERSIONS_V6` (meta-schema v3), `DRIVE_VERSION_V9` (`detect_ranked_mode` + grove/verify method slots), `DRIVE_ABCI_QUERY_VERSIONS_V2` (ranked routing gate). Behaviorally identical to v13 until a contract uses the new grammar
- GroveDB pins move `v5.0.1` → develop `4c6720b` (contains #657 + #781; #781 also fixes a subset-verification regression that #657 introduced for pre-existing shielded proofs)

## How Has This Been Tested?

- `rs-dpp` index grammar/immutability/gating tests (28) — parsing, desugaring, per-flag rejections, PV13 vs PV14 acceptance
- `rs-drive` `insert_contract/v0/tests/ranked_index_e2e_tests.rs` — all three variants through real batches: registration shape (axes TLV), ranking across inserts/updates/deletes, `verify_grovedb` integrity after every test; `restaurants` fixture added
- `rs-drive` `drive_document_ranked_query/tests.rs` (28) — validation matrix, all axes/kinds, prove→verify round-trips against the live root hash, tamper rejection, tie ordering, multi-index prover/verifier index agreement
- `rs-drive-abci` `document_query/v1/tests.rs` ranked module (13) — wire-level requests per axis, proof responses, PV13 rejection, empty-ranking-prove clean error, malformed-shape rejections
- `rs-drive-proof-verifier` (12) + `rs-sdk` (7 unit + doctest) — wire round-trip against the abci ground-truth shape, 16-byte avg decode incl. negatives, order preservation
- Full suites: `cargo test -p drive --lib` 3300 · `-p dpp --lib` 3834 · `-p drive-abci --lib query::` 604 · `-p platform-version` 13 · `-p drive-proof-verifier` 257 · `-p dash-sdk --lib/--doc` 181/12 — all 0 failures; `cargo check --workspace --all-targets`, `cargo check -p drive --no-default-features --features verify`, clippy clean on all touched crates, `cargo fmt` clean

Not yet covered (needs a live PV14 network, which doesn't exist): recorded SDK `tests/vectors` fixtures and live e2e; JS/Java/ObjC/Python generated gRPC clients still need regeneration (`yarn workspace @dashevo/dapi-grpc build`, docker).

## Breaking Changes

None for existing networks or clients: the new consensus behavior is entirely gated behind protocol v14 (a v13→v14 upgrade is a no-op until a contract uses the new grammar), the wire change is additive, and existing aggregate queries are unaffected. Merging does make nodes signal v14 as their desired version.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added ranked aggregate queries using `HAVING TOP(n)` and `BOTTOM(n)` for count, sum, and average results.
  * Ranked responses preserve ordering and support proof generation and verification.
  * Added ranked-index configuration for document schemas and compatible response formats.
  * Enabled support through protocol version 14.
* **Bug Fixes**
  * Improved validation for unsupported ranking forms, malformed queries, and incompatible indexes.
  * Fixed aggregate index handling for shared-prefix and ranked layouts.
* **Documentation**
  * Clarified ranking limits, ordering, protocol requirements, and `MIN`/`MAX` limitations.


thepastaclaw:
🕓 Ready for review — 10 ahead in queue (commit 745581d8)
_Queue position: 11/11 · 2 reviews active_
_ETA: start ~22:59 UTC · complete ~23:23 UTC (median 23m across 30 recent reviews; 2 slots)_
_Queued 25m ago · Last checked: 2026-08-04 20:50 UTC_

packages/rs-drive/src/query/drive_document_ranked_query/mode_detection.rs
```diff
@@ -0,0 +1,298 @@
...
+                ))));
+            }
+            if !matches!(clause.operator, HavingOperator::Equal) {
+                return Err(Error::Query(QuerySyntaxError::Unsupported(format!(
+                    "`{:?}` ranking is supported with the `=` having operator only; got \
+                     {:?}",
+                    ranking.kind, clause.operator
+                ))));
+            }
+            1u16
```
thepastaclaw: **🔴 Blocking: MAX and MIN omit groups tied at the extreme**

The wire and Drive contracts define `HAVING COUNT(*) EQ MAX` as selecting every group whose aggregate equals the maximum scalar. This branch instead converts MAX/MIN into a directional top-K read with `k = 1`. The indexed secondary breaks equal aggregates by group key, so only one tied group is returned and the other groups satisfying the predicate are silently omitted. Reject MAX/MIN until a tie-aware primitive is available, or retrieve and prove every entry tied at the extreme.

<sub>source: ['codex']</sub>

packages/dapi-grpc/protos/platform/v0/platform.proto
```diff
@@ -1280,6 +1346,15 @@ message GetDocumentsResponse {
...
         AverageResults averages = 4;
+        // Ranked-aggregate result. Routed when the request carries a
+        // `having` clause whose right operand is a `HavingRanking`
+        // (`TOP(n)` / `BOTTOM(n)` / `MAX` / `MIN`) matching the
+        // single `select`. Answered from the per-axis secondary of
+        // an indexed tree, so the index must declare the matching
+        // `rankedCountable` / `rankedSummable` / `rankedAverageable`
+        // keyword (meta-schema v3, protocol version 14+). Entry
+        // order is the ranking order — see `RankedEntries`.
+        RankedEntries ranked = 5;
```
thepastaclaw: **🔴 Blocking: Regenerate the committed gRPC clients for ranked results**

The protobuf adds `RankedEntry`, `RankedEntries`, and oneof field 5, but the maintained bindings under `packages/dapi-grpc/clients/platform/v0/` contain no ranked symbols. For example, the Node decoder's oneof still lists only `documents|counts|sums|averages` and skips field 5 as unknown, while the Web, Python, and Objective-C APIs expose no ranked accessor. A non-proof ranked response therefore loses its result when decoded by the shipped clients. Run the existing dapi-grpc generation script and commit every maintained binding.

<sub>source: ['codex']</sub>

packages/rs-dpp/schema/meta_schemas/document/v3/document-meta.json
```diff
@@ -0,0 +1,733 @@
...
+          "name"
+        ],
+        "dependentRequired": {
+          "rangeCountable": ["countable"],
+          "rangeSummable": ["summable"],
+          "rangeAverageable": ["averageable"],
+          "rankedCountable": ["rangeCountable"],
+          "rankedSummable": ["rangeSummable"],
+          "rankedAverageable": ["rangeAverageable"]
+        },
```
thepastaclaw, line 490: **🔴 Blocking: Explicit false ranked flags incorrectly require range axes**

JSON Schema's `dependentRequired` is triggered by property presence rather than its boolean value. Consequently, `"rankedCountable": false` still requires `rangeCountable`, and `"rankedAverageable": false` still requires `rangeAverageable`. This contradicts the fields' “When true” semantics and `Index::try_from_value_map`, which parses false as an ordinary opt-out and only enforces the range prerequisite when the resolved ranked flag is true. Replace these dependencies with value-sensitive `if`/`then` conditions, or declare the optional ranked properties as `const: true` if explicit false is intentionally unsupported.

<sub>source: ['codex']</sub>

packages/rs-dpp/schema/meta_schemas/document/v3/document-meta.json
```diff
@@ -0,0 +1,733 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://github.com/dashpay/platform/blob/master/packages/rs-dpp/schema/meta_schemas/document/v1/document-meta.json",
+  "$comment": "EDITABLE UNTIL 4.1 RELEASE — FROZEN AFTER. This v2 document meta-schema activates with protocol v13 (CONTRACT_VERSIONS_V5). It is v1 plus the keepsTransferHistory, keepsPurchaseHistory and keepsPricingHistory document type configuration flags, and admits every v13+ contract written to disk. Once the release carrying protocol v13 ships, mutating it would change historical validation results and break consensus replay. After release, any new top-level property or rule MUST go in a newer meta-schema version (v3+).",
```
thepastaclaw: **💬 Nitpick: Correct the v3 meta-schema header**

The new v3 file still identifies itself as v2, protocol v13, and `CONTRACT_VERSIONS_V5`, and directs future changes to v3. Update the comment to describe the ranked keywords, protocol v14, and `CONTRACT_VERSIONS_V6`, while preserving the existing v1 `$id` compatibility identifier.

```suggestion
  "$comment": "EDITABLE UNTIL RELEASE — FROZEN AFTER. This v3 document meta-schema activates with protocol v14 (CONTRACT_VERSIONS_V6). It is v2 plus the rankedCountable, rankedSummable and rankedAverageable index keywords, and admits every v14+ contract written to disk. Once the release carrying protocol v14 ships, mutating it would change historical validation results and break consensus replay. After release, any new top-level property or rule MUST go in a newer meta-schema version (v4+).",
```

packages/rs-platform-version/src/version/dpp_versions/dpp_contract_versions/v6.rs
```diff
@@ -0,0 +1,80 @@
...
+        index_versions: DocumentTypeIndexVersions {
+            index_levels_from_indices: 0,
+        },
+        class_method_versions: DocumentTypeClassMethodVersions {
+            try_from_schema: 2,
+            create_document_types_from_document_schemas: 1,
+        },
+        structure_version: 0,
+        schema: DocumentTypeSchemaVersions {
+            document_type_schema: 3, // changed: v3 document meta-schema (currently a byte-identical copy of v2; ranked index keywords land on it)
```
thepastaclaw: **💬 Nitpick: Remove the stale byte-identical v3 schema comment**

The inline comment says meta-schema v3 is currently byte-identical to v2, but v3 already contains the three ranked keywords and their prerequisite rules. This contradicts the accurate module-level explanation immediately above it.

```suggestion
            document_type_schema: 3, // changed: v3 document meta-schema with ranked index keywords
```

packages/rs-drive/src/drive/contract/insert/insert_contract/v0/tests/ranked_index_e2e_tests.rs
```diff
@@ -0,0 +1,1061 @@
...
+        })
+        .collect()
+}
+
+/// The fixed-point average grovedb's Avg axis orders by:
+/// `floor(sum * AVG_FIXED_POINT_SCALE / count)` with euclidean (toward
+/// -inf) division. The scale is grovedb's constant — 10^19 as of the
+/// merged PR #657 — and is deliberately not spelled as a literal here.
+fn expected_avg_fixed_point(sum: i64, count: u64) -> i128 {
+    (sum as i128) * AVG_FIXED_POINT_SCALE / (count as i128)
```
thepastaclaw: **💬 Nitpick: Use Euclidean division in the average test helper**

The helper documents floor division toward negative infinity, but Rust's `/` truncates toward zero for negative sums. Existing callers happen to use positive values, but a negative-value test would compute the wrong expected fixed-point average and could hide a regression.

```suggestion
fn expected_avg_fixed_point(sum: i64, count: u64) -> i128 {
    ((sum as i128) * AVG_FIXED_POINT_SCALE).div_euclid(count as i128)
}
```

thepastaclaw (changes requested):

The ranked storage and proof paths are extensively implemented, but three correctness issues block merge: MAX/MIN omit tied extrema, explicit false ranked flags are rejected by the meta-schema, and the committed generated gRPC clients cannot decode ranked responses. The WASM projection also needs two compatibility fixes, while several smaller documentation and test-helper issues remain.

### Review provenance

🔴 3 blocking | 🟡 2 suggestion(s) | 💬 3 nitpick(s)

<details>

```

In `packages/rs-drive/src/query/drive_document_ranked_query/mode_detection.rs`:
- [BLOCKING] packages/rs-drive/src/query/drive_document_ranked_query/mode_detection.rs:187-205: MAX and MIN omit groups tied at the extreme
  The wire and Drive contracts define `HAVING COUNT(*) EQ MAX` as selecting every group whose aggregate equals the maximum scalar. This branch instead converts MAX/MIN into a directional top-K read with `k = 1`. The indexed secondary breaks equal aggregates by group key, so only one tied group is returned and the other groups satisfying the predicate are silently omitted. Reject MAX/MIN until a tie-aware primitive is available, or retrieve and prove every entry tied at the extreme.

In `packages/dapi-grpc/protos/platform/v0/platform.proto`:
- [BLOCKING] packages/dapi-grpc/protos/platform/v0/platform.proto:1349-1357: Regenerate the committed gRPC clients for ranked results
  The protobuf adds `RankedEntry`, `RankedEntries`, and oneof field 5, but the maintained bindings under `packages/dapi-grpc/clients/platform/v0/` contain no ranked symbols. For example, the Node decoder's oneof still lists only `documents|counts|sums|averages` and skips field 5 as unknown, while the Web, Python, and Objective-C APIs expose no ranked accessor. A non-proof ranked response therefore loses its result when decoded by the shipped clients. Run the existing dapi-grpc generation script and commit every maintained binding.

In `packages/rs-dpp/schema/meta_schemas/document/v3/document-meta.json`:
- [BLOCKING] packages/rs-dpp/schema/meta_schemas/document/v3/document-meta.json:486-493: Explicit false ranked flags incorrectly require range axes
  JSON Schema's `dependentRequired` is triggered by property presence rather than its boolean value. Consequently, `"rankedCountable": false` still requires `rangeCountable`, and `"rankedAverageable": false` still requires `rangeAverageable`. This contradicts the fields' “When true” semantics and `Index::try_from_value_map`, which parses false as an ordinary opt-out and only enforces the range prerequisite when the resolved ranked flag is true. Replace these dependencies with value-sensitive `if`/`then` conditions, or declare the optional ranked properties as `const: true` if explicit false is intentionally unsupported.

In `packages/wasm-sdk/src/queries/system.rs`:
- [SUGGESTION] packages/wasm-sdk/src/queries/system.rs:861-875: Expose indexed-tree sums through the WASM PathElement API
  `PathElementWasm::from_element` obtains its JavaScript `sum` property through this function. Both `ProvableSumIndexedTree` and `ProvableCountProvableSumIndexedTree` carry an `i64` sum, but they currently fall through to `None`, so JavaScript sees a recognized indexed element type with an unexpectedly undefined aggregate. Project these sums in the same way as the corresponding non-indexed variants.
- [SUGGESTION] packages/wasm-sdk/src/queries/system.rs:20-68: Add indexed-tree runtime values to the WASM TypeScript union
  The `elementType` getter now emits six values absent from `GroveElementType`: `provableSumIndexedTree`, `provableCountIndexedTree`, `provableCountProvableSumIndexedTree`, and their three `nonCounted` forms. JavaScript receives these strings, but generated TypeScript declarations claim they are impossible, making exhaustive switches and validators unsound. Add all six strings to the custom TypeScript section.
```
</details>

packages/rs-drive/src/query/drive_document_ranked_query/mode_detection.rs
QuantumExplorer: Fixed in 67fbe47: MAX/MIN are now rejected with an error explaining the tie semantics (every group tied at the extreme cannot be proved — the axis secondary breaks ties by group key). Positional TOP(1)/BOTTOM(1) remain and keep k=1 execution coverage on every axis. Detection, execution, abci, and SDK tests retargeted accordingly.

packages/rs-dpp/schema/meta_schemas/document/v3/document-meta.json
QuantumExplorer: Fixed in 67fbe47: the three ranked rows moved out of dependentRequired into value-sensitive if/then conditionals (allOf), so an explicit false no longer demands the range axis — matching the structural parser's opt-out semantics. Both directions are pinned by tests (explicit-false passes full validation; true without the range axis still fails), and the new test was proven load-bearing by temporarily restoring the presence-based row.

packages/rs-dpp/schema/meta_schemas/document/v3/document-meta.json
QuantumExplorer: Fixed in 67fbe47: header rewritten for v3 / protocol v14 / CONTRACT_VERSIONS_V6 + the ranked keywords, with an explanation of why $id keeps the v1 path (enrich_with_base_schema injects that URI as every PV12+ document schema's $schema).

packages/rs-platform-version/src/version/dpp_versions/dpp_contract_versions/v6.rs
QuantumExplorer: Fixed in 67fbe47: the stale byte-identical comment is gone; the accurate module-level doc stands.

packages/rs-drive/src/drive/contract/insert/insert_contract/v0/tests/ranked_index_e2e_tests.rs
QuantumExplorer: Fixed in 67fbe47: the helper now mirrors grovedb's compute_avg_fixed_point operation-for-operation (0/0=0, saturating_mul, div_euclid), and a new signed-delta doctype in the fixture exercises negative averages where truncating division would provably disagree (the test asserts euclidean == truncating - 1 on that group).

packages/rs-drive-proof-verifier/src/lib.rs
```diff
@@ -14,12 +14,27 @@ pub use proof::document_count::{
     verify_distinct_count_proof, verify_point_lookup_count_proof,
     verify_primary_key_count_tree_proof, DocumentCount,
 };
+/// Verified ranked (`HAVING … TOP(n)` / `BOTTOM(n)` / `MAX` / `MIN`)
+/// result types. `DocumentRankedEntries` carries one entry per
+/// returned group **in ranking order**;
+/// [`verify_ranked_top_k_proof`](proof::document_ranked::verify_ranked_top_k_proof)
+/// is the tenderdash-composition wrapper that binds the proof's
+/// reconstructed root hash to the signed app hash.
+pub use proof::document_ranked::{verify_ranked_top_k_proof, DocumentRankedEntries};
```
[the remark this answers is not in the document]
QuantumExplorer: Fixed in ce0cb64: the doc link now uses the public re-export (same-scope item link) instead of routing through the private proof:: path; the same comment's stale MAX/MIN mention was corrected while there.

packages/rs-drive/src/drive/document/update/internal/update_document_for_contract_operations/v0/mod.rs
```diff
@@ -489,14 +492,15 @@ impl Drive {
...
-                        let inserted = self.batch_insert_empty_tree_if_not_exists(
+                        let (property_name_tree_type, ranked_axes) =
+                            property_name_tree_type_for_index_level(current_index_level)?;
+                        let inserted = self.batch_insert_empty_index_tree_if_not_exists(
                             PathKeyInfo::PathKeyRef::<0>((
                                 index_path.clone(),
                                 index_property.name.as_bytes(),
                             )),
                             property_name_tree_type,
+                            &ranked_axes,
```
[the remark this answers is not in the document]
QuantumExplorer: Moot as of merge 0b89421: update_document_for_contract_operations v0 was reverted to its upstream (pre-ranked) shape, and ranked re-materialization now lives in the new v1 update walker (from #4265), which resolves the terminal level through the shared ranked resolver — exactly what this comment asked for, in the implementation protocol v14 actually selects. Covered by ranked_index_ranks_correctly_next_to_a_compound_index_sharing_its_property, which exercises the v1 update walker on a ranked index.

packages/rs-drive/src/query/drive_document_ranked_query/execute_top_k.rs
```diff
@@ -0,0 +1,169 @@
...
+        // Same destructure-don't-unwrap rationale as the no-proof arm.
+        let CostContext { value, cost: _ } = drive.grove.prove_indexed_axis_top_k(
+            path_refs.as_slice(),
+            self.axis.into(),
+            self.k,
+            self.descending,
+            transaction,
+            grove_version,
+        );
+        value.map_err(|e| Error::GroveDB(Box::new(e)))
```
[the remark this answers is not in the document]
QuantumExplorer: Not actionable today: grovedb flattens merk's empty-tree error into Error::CorruptedData(String) at this boundary, so there is no typed variant to match ('when available' does not hold yet). The substring match is centralized in exactly one place (drive-abci's empty_ranking_proof_rejection, with the marker as a named constant) rather than scattered. If grovedb grows a dedicated variant we will switch to it — noted in the code. Keeping the pass-through here preserves generic grovedb failures unchanged.

packages/dapi-grpc/protos/platform/v0/platform.proto
QuantumExplorer: Still open — acknowledged as the remaining blocker. Regeneration needs the docker-based build (packages/dapi-grpc/scripts/build.sh via yarn workspace @dashevo/dapi-grpc build), which is not available in this environment; the regenerated Node/Web/Java/ObjC/Python clients will be committed before merge. The Rust side compiles the proto at build time and already carries the new messages.

QuantumExplorer: Done in 63f91a3: all maintained bindings regenerated with the repo's pinned docker toolchain (yarn workspace @dashevo/dapi-grpc build) — RankedEntry/RankedEntries and the ResultData.ranked oneof arm are now present in the Node (protoc + pbjs), Web (js + d.ts), Python and Objective-C clients, and drive_pbjs.js picks them up through its platform.proto import. The diff is purely additive (the only removals are the four-variant oneof lists becoming five-variant and proto doc comments flowing into generated jsdoc); dapi-grpc JS tests pass (32).

packages/rs-drive-abci/Cargo.toml
```diff
@@ -82,7 +82,7 @@ derive_more = { version = "1.0", features = ["from", "deref", "deref_mut"] }
 async-trait = "0.1.77"
 console-subscriber = { version = "0.4", optional = true }
 bls-signatures = { git = "https://github.com/dashpay/bls-signatures", rev = "0842b17583888e8f46c252a4ee84cdfd58e0546f", optional = true }
-grovedb-commitment-tree = { git = "https://github.com/dashpay/grovedb", tag = "v5.0.1" }
+grovedb-commitment-tree = { git = "https://github.com/dashpay/grovedb", rev = "d473818154dac668676fcb54678287845032d9a7" }
```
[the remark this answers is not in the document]
QuantumExplorer: Confirmed and fixed in 2404e74 — this was a genuine gap. GROVE_V4 gates exactly the indexed-tree batch cleanup behaviors (overwrite inspection + delete-tree actual-type cleanup), and indexed trees only exist from protocol v14, so PV14 selecting GROVE_V3 would have let a batch overwrite of a ranked index orphan its per-axis secondary storage. DRIVE_VERSION_V9 now selects GROVE_V4; the latest-version fee-constant tests moved up by the gate's documented one-read cost while first-version tests keep released values, pinning that the change activates exactly at v14. One residual noted for follow-up: grovedb's estimation does not yet model the inspection read, so dry-run estimates undershoot applied fees by that read on overwrite-capable ops.

packages/rs-drive/src/query/having.rs
```diff
@@ -129,9 +144,11 @@ pub enum HavingRightOperand {
...
     Value(Value),
     /// Cross-group ranking reference. Operator compatibility:
-    /// scalar comparison operators work with `Min` / `Max` /
-    /// `Top(1)` / `Bottom(1)`; `In` works with `Top(N)` /
-    /// `Bottom(N)` (membership in the top-N / bottom-N set).
+    /// scalar comparison operators work with `Top(1)` /
+    /// `Bottom(1)`; `In` works with `Top(N)` / `Bottom(N)`
+    /// (membership in the top-N / bottom-N set). `Min` / `Max` are
+    /// wire-stable but rejected by evaluation whatever the
+    /// operator — see [`HavingRankingKind::Min`].
```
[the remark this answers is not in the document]
QuantumExplorer: Fixed in 2404e74: the doc now states Equal works with Top(1)/Bottom(1) only and In with Top(N)/Bottom(N), no broader scalar-operator claim; the Min/Max rejection note is preserved.

packages/rs-drive/src/drive/identity/update/mod.rs
```diff
@@ -59,7 +59,7 @@ mod tests {
                 storage_fee: 14202000,
                 // 2 extra loaded bytes because the token tree is no longer empty
                 // these 2 loaded bytes cost 20 credits each
-                processing_fee: 1098300,
+                processing_fee: 1102300,
```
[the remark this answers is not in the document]
QuantumExplorer: Fixed (in the next push): the comment now explains both components of the latest-version delta — the 2 loaded token-tree bytes and the grove v4 overwrite-inspection read charged from protocol v14. The increase itself is intended; it is the documented cost of the GROVE_V4 cleanup gate activated at v14 in this PR.

book/src/drive/document-ranked-trees.md
```diff
@@ -0,0 +1,353 @@
...
+
+Both gates cost **one extra stored-element read per overwrite-capable operation** — which is why they could not be applied to released versions, and why v14's fee constants move. The shape of the change is worth internalising because it is easy to misread as a general fee increase:
+
+- **Fresh-key inserts pay nothing extra.** There is no stored element to inspect. In the address-funds-transfer fee regression suite the pins for simple single-input/single-output transfers, multisig spends and the user-fee-increase case are all **unchanged**; only the cases that overwrite existing keys move.
+- **Overwrite-capable operations pay one seek each.** The identity-balance and token-balance update pins move `385 160 → 389 460` and `260 540 → 264 840` processing credits, with storage fees unchanged throughout — the delta is pure read cost.
+- **The boundary itself is pinned.** `run_chain_one_identity_in_solitude` sits exactly **2 000 credits** (one seek) below its new protocol-version-13 sibling, which asserts the pre-grove-v4 figure. The pair brackets the v13 → v14 transition, so a future change that moves either side has to say which one it meant to move.
```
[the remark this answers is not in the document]
QuantumExplorer: Fixed in the next push: the bullets now state the full read composition (seek + 20/byte, so deltas vary with element size), identify the +4,300 examples as a seek plus a 115-byte element, and describe the solitude boundary delta as the 2,000-credit gate cost of that run's single gated op rather than equating the read with a bare seek. Storage-fee claim preserved. Note: if the pending grovedb change making these gates cost-neutral lands, this whole section gets rewritten to describe zero marginal cost.

packages/rs-drive/src/query/drive_document_ranked_query/mode_detection.rs
```diff
@@ -0,0 +1,303 @@
...
+/// SELECT SUM(f)     GROUP BY p HAVING SUM(f)     IN TOP(n) | IN BOTTOM(n) | EQ MAX | EQ MIN
+/// SELECT AVG(f)     GROUP BY p HAVING AVG(f)     IN TOP(n) | IN BOTTOM(n) | EQ MAX | EQ MIN
+/// ```
+///
+/// with no `WHERE`, no `LIMIT` / `OFFSET` / `START AT`, `1 ≤ n ≤`
+/// [`MAX_RANKED_LIMIT`], and the `HAVING` aggregate byte-equal to the
+/// `SELECT` projection. `EQ TOP(1)` / `EQ BOTTOM(1)` are accepted as
+/// synonyms of `EQ MAX` / `EQ MIN` because
+/// [`HavingRightOperand::Ranking`]'s own contract documents them as
+/// equivalent.
```
thepastaclaw: **💬 Nitpick: Remove the remaining MAX/MIN acceptance claim**

The public documentation says the v0 grammar accepts `EQ MAX` and `EQ MIN` and describes `EQ TOP(1)` and `EQ BOTTOM(1)` as synonyms. The implementation at lines 182-203 now correctly rejects MAX/MIN because their value-based semantics require returning every group tied at the extreme, while TOP(1)/BOTTOM(1) are positional and may truncate ties. Update the accepted-shape table and explanation so callers are not told these semantically different forms are equivalent.

```suggestion
/// Accepts exactly:
///
/// ```text
/// SELECT COUNT(*)   GROUP BY p HAVING COUNT(*)   IN TOP(n) | IN BOTTOM(n)
/// SELECT SUM(f)     GROUP BY p HAVING SUM(f)     IN TOP(n) | IN BOTTOM(n)
/// SELECT AVG(f)     GROUP BY p HAVING AVG(f)     IN TOP(n) | IN BOTTOM(n)
/// ```
///
/// with no `WHERE`, no `LIMIT` / `OFFSET` / `START AT`, `1 ≤ n ≤`
/// [`MAX_RANKED_LIMIT`], and the `HAVING` aggregate byte-equal to the
/// `SELECT` projection. `EQ TOP(1)` / `EQ BOTTOM(1)` are also accepted
/// for the positional single best- or worst-ranked group. `MAX` and `MIN`
/// are wire-decodable but rejected because they require returning every
/// group tied at the extreme, which the indexed top-k proof cannot attest.
```

<sub>source: ['codex']</sub>

thepastaclaw (commented):

All eight findings carried forward from SHA d2af6ae81d82d8bf758e0b65bcd021f1a28f7cae are fixed at the current head, including the three former blockers. The latest delta introduces one new documentation nitpick: the public ranked-mode grammar still describes MAX/MIN as accepted and equivalent to TOP(1)/BOTTOM(1), despite the implementation now correctly rejecting them. No functional or security blocker remains.

### Review provenance
- Sonnet reviewers: `claude-sonnet-5` — general (completed), `claude-sonnet-5` — security-auditor (failed), `claude-sonnet-5` — security-auditor (failed), `claude-sonnet-5` — security-auditor (completed)

💬 1 nitpick(s)

packages/rs-drive/src/query/drive_document_ranked_query/mode_detection.rs
QuantumExplorer: Fixed in the next push using your suggested wording: the grammar lines now list EQ TOP(1)/EQ BOTTOM(1), and the prose states MAX/MIN are wire-decodable but rejected because every group tied at the extreme cannot be attested by the indexed top-k proof.

thepastaclaw (approved):

The sole carried-forward finding, "Remove the remaining MAX/MIN acceptance claim," is fixed at the current head: the ranked-mode grammar now documents only TOP/BOTTOM forms as accepted and explicitly distinguishes the rejected, tie-sensitive MAX/MIN semantics. The latest delta is documentation-only, matches the existing rejection logic, and introduces no new in-scope findings.

### Review provenance
- Sonnet reviewers: `claude-sonnet-5` — general (completed), `claude-sonnet-5` — security-auditor (completed)

packages/rs-dpp/src/data_contract/document_type/class_methods/try_from_schema/v3/mod.rs
```diff
@@ -0,0 +1,1692 @@
...
+                                                index_property.name.to_owned(),
+                                                "maxLength".to_string(),
+                                                format!(
+                                                    "should be less or equal {}",
+                                                    MAX_INDEXED_STRING_PROPERTY_LENGTH
+                                                ),
+                                            )
+                                            .into(),
+                                        )))
+                                    }
```
thepastaclaw: **🔴 Blocking: Enforce the shorter ranked-secondary key limits**

This validation still applies the generic index limits of 255 encoded bytes for byte arrays and 63 characters for strings when an index enables a ranked axis. The pinned GroveDB revision constructs each secondary key as `sort_key || item_key` and explicitly limits item keys to 247 bytes for Count/Sum axes and 239 bytes for Avg. `serialize_value_for_key` can therefore produce a schema-valid 255-byte byte-array key or a 63-character string encoded as 240–252 UTF-8 bytes that passes DPP validation but causes the ranked indexed-tree batch to fail with `InvalidInput`. Validate the property's worst-case encoded key length against the strictest configured ranked axis during contract registration so a valid document state transition cannot reach an internal storage failure.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

No prior active findings required carry-forward. The cumulative review found two new blocking correctness gaps in ranked-index validation and null handling, plus one stale protobuf-documentation issue. The latest GroveDB gate and fee-constant delta introduces no separate defect and does not resolve these findings.

### Review provenance

<details>

```

In `packages/rs-dpp/src/data_contract/document_type/class_methods/try_from_schema/v3/mod.rs`:
- [BLOCKING] packages/rs-dpp/src/data_contract/document_type/class_methods/try_from_schema/v3/mod.rs:607-646: Enforce the shorter ranked-secondary key limits
  This validation still applies the generic index limits of 255 encoded bytes for byte arrays and 63 characters for strings when an index enables a ranked axis. The pinned GroveDB revision constructs each secondary key as `sort_key || item_key` and explicitly limits item keys to 247 bytes for Count/Sum axes and 239 bytes for Avg. `serialize_value_for_key` can therefore produce a schema-valid 255-byte byte-array key or a 63-character string encoded as 240–252 UTF-8 bytes that passes DPP validation but causes the ranked indexed-tree batch to fail with `InvalidInput`. Validate the property's worst-case encoded key length against the strictest configured ranked axis during contract registration so a valid document state transition cannot reach an internal storage failure.

In `packages/rs-drive/src/drive/document/insert/add_indices_for_top_index_level_for_contract_operations/v2/mod.rs`:
- [BLOCKING] packages/rs-drive/src/drive/document/insert/add_indices_for_top_index_level_for_contract_operations/v2/mod.rs:149-158: Do not index branches excluded by nullSearchable
  The value tree is created before the recursive terminal handler evaluates `should_insert_with_all_null`. For a missing property on a single-property ranked index, `document_top_field` is the empty key; this call inserts its value tree, then `add_reference_for_index_level_for_contract_operations_v0` returns without adding a reference because `nullSearchable` is false. GroveDB treats the empty value tree as a present indexed-tree child and mirrors its zero aggregate into the ranked secondary, so TOP/BOTTOM queries can return an authenticated empty-key group that the index explicitly excludes. The same phantom group occurs when a ranked `[a]` index shares its prefix with `[a,b]`, because the branch may be needed structurally while its terminal reference remains suppressed. Reject `nullSearchable: false` for ranked indexes or change the layout so branches without an index reference are not mirrored into ranked secondaries.

In `packages/dapi-grpc/protos/platform/v0/platform.proto`:
- [SUGGESTION] packages/dapi-grpc/protos/platform/v0/platform.proto:1065-1071: Update the proto comments for v14 ranked HAVING support
  This field-level comment still says every non-empty HAVING clause is rejected, contradicting the message-level documentation at lines 834–840 and the protocol-v14 implementation that accepts ranked TOP/BOTTOM forms. The rejected-shapes list around line 866 repeats the blanket-rejection claim, while response comments around `RankedEntry`, `RankedEntries`, and `ResultData.ranked` still describe MAX/MIN as returned forms even though mode detection rejects them because tied extremes cannot be proved. Update these comments consistently: protocol versions before v14 reject every HAVING clause, while v14 accepts supported TOP/BOTTOM forms and rejects MAX/MIN.
```
</details>

packages/rs-dpp/src/data_contract/document_type/class_methods/try_from_schema/v3/mod.rs
QuantumExplorer: Fixed in 823a59a: generation-3 parsing now validates the ranked property's worst-case encoded key length against the grovedb ceilings (247 bytes for count/sum, 239 for avg — verified against the pinned rev's MAX_CIDX_ITEM_KEY_LEN / max_item_key_len_for_axis) using the same max_byte_size derivation the generic 255/63 limits come from: strings cap at 61 chars (count/sum) / 59 (avg) via the 4-bytes-per-char worst case, byte arrays at the raw ceilings, avg winning when multiple axes are declared. Boundary tests on both sides of every limit. The same commit also addresses the second blocker (nullSearchable: false now rejected on ranked indexes at parse time — the value-tree-before-reference ordering would mirror an excluded null group into the secondary as a zero-aggregate phantom; with the default true, null docs are a real referenced group and no phantom shape is reachable) and the proto comment sweep including HavingRanking's operator table (clients regenerated, comment-only diff).

thepastaclaw (approved):

All three cumulative prior findings are fixed at the current head: ranked key ceilings are enforced during schema parsing, ranked indexes reject nullSearchable=false, and the protobuf documentation now matches protocol-v14 HAVING behavior. No prior finding remains active, and the latest-delta review found no new actionable defect; the only newly reported suggestion is refuted by the parser invariant and contract-insertion call order.

### Review provenance
- Sonnet reviewers: `claude-sonnet-5` — general (completed), `claude-sonnet-5` — security-auditor (completed), `claude-sonnet-5` — rust-quality (completed)

QuantumExplorer:
Conflicts resolved in 0bbe28e (v4.2-dev's txMetadata revert deleted a file this branch had only mechanically touched — the deletion wins) and CodeRabbit's PV14-pin nitpick is addressed in 94cd599: v3's tests now pin `PlatformVersion::get(14)` via a `pv14()` helper instead of `latest()`, matching the generation-pin pattern used for v0/v1/v2.

QuantumExplorer (commented):
Reviewed

PR #3954, opened 2026-06-22
feat(platform-wallet)!: shared ThreadRegistry for coordinator lifecycle + shutdown UAF/data-loss fixes

Claudius-Maginificent:
## Human-generated TL;DR

dash-evo-tool crashes on shutdown, and mismanages spv sync start/stop, because some start/stop coordination between tokio and OS threads is not working well. 

## Issue being fixed or feature implemented

The four background sync coordinators (platform-address/BLAST, per-identity
token state, DashPay, shielded) each run a `!Send` sync loop on a detached OS
thread. At manager teardown / FFI `destroy` **nothing joined those threads**:
a loop mid-pass could keep calling `persister.store(...)` or fire host
completion callbacks through a context the host frees the instant `destroy`
returns — a use-after-free. Cancellation + join were also racy: a `stop()` →
quick `start()` could leave a newer loop uncancellable, and a `start()` racing
`shutdown()` could leave a live, never-cancelled loop the registry then joins
with a 30s timeout while it keeps firing callbacks.

## What was done?

Merged onto current `v4.2-dev`; the new FFI error code below takes slot `27`
(slots `21`–`26` are already used by the address-nonce / core-funds /
asset-lock / broadcast-rejected codes on the base branch).

**`dash-async`: new shared `ThreadRegistry<K>` lifecycle engine** (`registry.rs`).
Tracks background OS-thread / tokio-task workers, owns their spawn +
cancellation + join, and reports teardown as a `ShutdownReport`. Surface used
by the wallet: `start_thread` (registry spawns, owns cancellation, and joins
— one slot lock covers check → spawn → install, no gap to race), weight-ordered
`shutdown` → `ShutdownReport`, `quiesce`, per-key clearing latch
(`hold_clearing`/`is_clearing`), and the teardown latch `is_closing`. Orphan
reap keeps a still-draining prior generation UAF-safe (parked, never
dropped-and-detached). `WorkerConfig` gained a `stack_size` field (`None` =
platform default) so a caller with a deep-recursion loop body (DashPay's
GroveDB proof descent — see below) can request a larger OS-thread stack
without hand-rolling its own `Builder`. The `AtomicFlagGuard`/`RefcountedFlagGuard`
helper types originally drafted alongside the registry had zero callers
anywhere in this diff or in the wallet, so they were removed rather than kept
speculatively.

**Wallet: all four coordinators spawn through `registry.start_thread`, full stop.**
Each coordinator's `start()` used to spawn its own `std::thread`, install a
[...]

PR #4268, opened 2026-08-02
feat(platform-wallet): registry-owned coordinator lifecycle with Rust-owned FFI callback contexts

QuantumExplorer:
## Human-generated TL;DR

Supersedes #3954 (same underlying work by @lklimek / Claudius, carried
forward here). dash-evo-tool crashes on shutdown and the mobile hosts
could UAF their persistence/event callbacks, because the four background
sync coordinators ran `!Send` loops on detached OS threads that nothing
joined or owned at teardown.

Three distinct problems, fixed by construction rather than by defensive
checking:

1. **Runtime-drop crash (dash-evo-tool).** The coordinator loops run
   `Handle::block_on` on the host's tokio runtime; at exit a detached
   loop mid-pass touches `tokio::time` on a dropped runtime. Nothing
   joined those threads.
2. **Callback use-after-free (Swift/Kotlin hosts).** Callback contexts
   were *borrowed* (`Unmanaged.passUnretained` / caller-owned `GlobalRef`
   boxes): Rust-side `Arc` clones of the persister/event wrappers could
   outlive `destroy` and fire callbacks into memory only the host
   controlled.
3. **Clear/reset races.** `clear_shielded` and the platform-address reset
   mutate state a concurrently-running sync pass also touches (commitment
   tree, watermarks, balances); stop was cancel-only and couldn't prevent
   a pass from re-persisting what was just wiped.

## What was done?

**`dash-async`: `ThreadRegistry<K>`** (from #3954, slimmed). Owns the
coordinator OS-thread lifecycle end to end: the closing/clearing-latch
check, the spawn, and the cancellation-token install happen under one
slot lock (no check-then-spawn gap); restart is generation-guarded with
the prior thread parked and bounded-joined, never detached;
`shutdown()` cancels + joins everything concurrently and returns a
`ShutdownReport` so a wedged thread is *reported* (`Timeout`/`Detached`)
instead of silently detached — the signal a runtime-owning host needs
before dropping its runtime. `WorkerConfig::stack_size` covers DashPay's
deep GroveDB proof descent. Deliberately absent (present in earlier
revisions of this work, deleted here as dead weight): weight-ordered
shutdown tiers, `DrainHook`, `start_task`, `register_thread`,
`merged_with_retry`, `any_alive*` — none had a consumer.

**Wallet: all four coordinators spawn through `registry.start_thread`**,
replacing four hand-rolled copies of the same generation-guard
(`loop_cancel.rs` deleted). Every pass claims an `is_syncing` slot (RAII,
panic-safe) and honors a shared `QuiesceGate`; `quiesce_held()` drains an
in-flight pass **and keeps admission shut until its guard drops**, which
`clear_shielded` / `reset_platform_address_sync_state` hold across their
whole quiesce → wipe section — a drain that times out fails closed with
the typed `ShutdownIncomplete` error instead of wiping under a live pass.
`PlatformWalletManager::shutdown` seals the gates (terminal), stops SPV
(bounded, abort-escalating, still-live run loop re-parked), drains the
DashPay payment-hook tasks (one shared abort deadline), joins the
coordinator threads via the registry, then the event adapter — every
phase bounded, so the FFI `destroy` can never hang.

**FFI: Rust owns the callback contexts.** Both vtables
(`PersistenceCallbacks`, `EventHandlerCallbacks`) gain a trailing
`release_fn`; setting it transfers ownership of `context` to Rust, which
releases it **exactly once, when the manager and its last worker are
provably done calling into the host** (the wrappers are built once per
manager into the `Arc` every worker clones; `Drop` fires the release). A
worker straggling past destroy keeps the host object alive through its
own Arc and frees it on exit. Because of this, `destroy` needs no retry,
no error return, and no host-side leak discipline: it runs one bounded
shutdown, logs a non-clean join, and returns `Success`. Swift hands its
handlers over with `passRetained` + release trampolines (balancing the
retain on the create-failure path); JNI transfers its boxed `GlobalRef`
contexts the same way (jni 0.21's `GlobalRef::drop` self-attaches on a
detached thread). Null `release_fn` keeps the legacy borrowed contract.

This also fixes a latent hazard by construction: a `PlatformWallet`
handle outliving manager destroy previously kept persister clones
pointing at host-freed memory.

## How Has This Been Tested?

- `cargo test -p dash-async` → 28/28 on the slimmed registry
  (generation guard, orphan park/reap, spawn-failure rollback,
  closing/clearing latches, custom stack size).
- `cargo test -p platform-wallet --features shielded --lib` → 650/650,
  incl. new gate tests: `quiesce_held_bars_passes_until_the_guard_drops`,
  `overlapping_quiesce_guards_reopen_only_after_the_last_drop`,
  `quiesce_sealed_never_reopens_admission`,
  `sync_wallet_is_refused_while_admission_is_shut`,
  `reset_platform_address_state_fails_closed_on_a_wedged_pass`,
  `spv_task_surviving_abort_is_returned_for_reparking`.
- `cargo test -p platform-wallet-ffi --lib` → 207/207 (both feature
  configs), incl. `release_fires_once_when_the_last_arc_clone_drops`
  (persister + event variants) and
  `destroy_releases_owned_callback_contexts_exactly_once` (end to end
  through the public FFI, stale handle cannot double-release); the vtable
  layout pin proves `release_fn` is the terminal field.
- `cargo check --all-targets` clean for `rs-unified-sdk-jni`; clippy
  clean on every touched file; `cargo fmt --check` clean;
  `RUSTDOCFLAGS="-D warnings" cargo doc -p dash-async --no-deps` clean.
- Swift compiles in CI against the regenerated cbindgen header (the
  identical Swift-side pattern passed the Swift SDK job on the
  predecessor branch, run 30757672140).

## Breaking Changes

- The four coordinator constructors gain a leading
  `registry: Arc<ThreadRegistry<WalletWorker>>` parameter.
- `PlatformWalletManager::shutdown(&self)` return type: `()` →
  `ShutdownReport<WalletWorker>`.
- `dash_async::WorkerConfig` is now `{ join_budget, stack_size }`.
- `PersistenceCallbacks` / `EventHandlerCallbacks` (`#[repr(C)]`) gain a
  trailing `release_fn` field; in-tree Swift/Kotlin consumers are
  regenerated in lockstep.
- New FFI result code
  `PlatformWalletFFIResultCode::ErrorShutdownIncomplete = 27`, returned
  by the clear/reset/sync-stop drain barriers (NOT by `destroy`, which
  always returns `Success` for a live handle).

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved wallet shutdown reliability with bounded cleanup and safer handling of background tasks.
  * Prevented callback and handler lifetime issues during shutdown, reset, and manager creation failures.
  * Ensured late callbacks are safely rejected and retried during the next launch.
  * Fixed cleanup state after interrupted or failed synchronization operations.

* **New Features**
  * Added a distinct shutdown-incomplete error for Kotlin, Swift, and native integrations.
  * Shutdown now reports which operations did not complete cleanly.


thepastaclaw:
🕓 Ready for review — next in queue (commit 429667e7)
_Queue position: 1/14 · 2 reviews active_
_ETA: start ~19:30 UTC · complete ~19:50 UTC (median 20m across 30 recent reviews; 2 slots)_
_Queued 43m ago · Last checked: 2026-08-02 19:20 UTC_

packages/rs-platform-wallet-ffi/src/manager.rs
```diff
@@ -422,50 +432,44 @@ pub unsafe extern "C" fn platform_wallet_manager_get_wallet(
...
-                    ),
-                );
-            }
         }
+        // Dropping the manager here releases its persister/event-handler
+        // references; the host contexts are released (via `release_fn`)
+        // as soon as the last worker's reference drops — typically right
+        // now, or later if a straggler is still draining.
     }
     PlatformWalletFFIResult::ok()
```
thepastaclaw, line 506: **🔴 Blocking: Non-clean destroy is still unsafe for borrowed callback contexts**

Both callback vtables explicitly preserve the legacy borrowed-context contract when `release_fn` is null, but this path now removes the manager and returns `Success` even when shutdown reports a callback-capable straggler. An `Arc<FFIPersister>` or `Arc<FFIEventHandler>` retains only the Rust wrapper and its raw `context` pointer in borrowed mode; it does not retain the host allocation behind that pointer. A legacy caller can therefore release its context after the successful destroy while a surviving worker later invokes a callback through freed memory. Preserve `ErrorShutdownIncomplete` or another actionable failure whenever either context is borrowed, reject borrowed contexts at creation, or retain the previous retry/fail-closed behavior for borrowed managers.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The PR’s ownership inversion is sound for callback vtables that provide release functions, but two teardown/admission races remain. Borrowed callback contexts can still be used after a successful destroy, and concurrent drain acquisition can briefly reopen sync admission during a clear/reset barrier; both issues require changes before merge.

### Review provenance

🔴 2 blocking

<details>

```

In `packages/rs-platform-wallet-ffi/src/manager.rs`:
- [BLOCKING] packages/rs-platform-wallet-ffi/src/manager.rs:457-474: Non-clean destroy is still unsafe for borrowed callback contexts
  Both callback vtables explicitly preserve the legacy borrowed-context contract when `release_fn` is null, but this path now removes the manager and returns `Success` even when shutdown reports a callback-capable straggler. An `Arc<FFIPersister>` or `Arc<FFIEventHandler>` retains only the Rust wrapper and its raw `context` pointer in borrowed mode; it does not retain the host allocation behind that pointer. A legacy caller can therefore release its context after the successful destroy while a surviving worker later invokes a callback through freed memory. Preserve `ErrorShutdownIncomplete` or another actionable failure whenever either context is borrowed, reject borrowed contexts at creation, or retain the previous retry/fail-closed behavior for borrowed managers.

In `packages/rs-platform-wallet/src/manager/mod.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/manager/mod.rs:259-268: Concurrent drains can reopen admission before the second barrier is established
  A drain closes the atomic gate without recording itself in `GateBookkeeping`, and increments `holds` only after observing `is_syncing == false`. With two concurrent drains, the first can acquire and immediately drop its guard while the second is between its final `is_syncing` load and `gate.hold()`. That drop sets `closed` to false, allowing a direct sync to claim `is_syncing` and pass its gate check; the second drain then installs its hold and returns without rechecking the slot. A clear/reset using that guard can consequently mutate or wipe state while the newly admitted pass is running and later re-persisting stale state. Track active drain attempts as closing reasons, or serialize drain acquisition and recheck `is_syncing` after establishing the exclusive hold.
```
</details>

QuantumExplorer:
Pushed `429667e723`, fixing both blockers — each was a genuine finding.

**Blocker 2 (gate race) — fixed structurally.** `drain_pass` now takes its
gate hold at **entry**, so the gate is closed continuously from the
drain's first instruction to the returned guard's drop; a concurrent
holder's release can no longer open an admission window between another
drain's final `is_syncing` observation and its hold. The timeout path
latches the gate closed (cleared only by a later successful drain)
*before* the drain's own hold is released, and `closed` is now recomputed
from a single bookkeeping state (`holds`/`latched`/`sealed`) under the
lock — there is no ad-hoc writer left to lose an interleaving.
Regression tests:
`concurrent_drain_keeps_gate_closed_across_another_holders_drop` (fails
against the previous close-then-hold-at-the-end gate) and
`timed_out_drain_latches_gate_closed_until_a_successful_drain`.

**Blocker 1 (borrowed contexts) — fixed by rejecting the unsound mode.**
Creation now fails with `ErrorInvalidParameter` when either vtable
carries a non-null `context` without a `release_fn`; `None` remains valid
only alongside a null context (the `configure(modelContainer: nil)`
shape). Rationale: `destroy` returns `Success` without proving every
worker joined, which is only sound when a straggler's `Arc` keeps the
host context alive via ownership — so a borrowed context isn't a legacy
mode to preserve, it's the UAF this PR exists to remove. A caller whose
context needs no cleanup passes a no-op `release_fn`. All in-tree hosts
(Swift, JNI) already pass one. Test:
`create_rejects_context_carrying_vtable_without_release_fn`, covering
both vtables and the null-context shape staying valid.

Verification: `platform-wallet` 652/652 (incl. the new gate tests),
`platform-wallet-ffi` 208/208, clippy clean on touched files,
`cargo fmt --check` clean.

QuantumExplorer (commented):
I guess this is fine.

PR #4269, opened 2026-08-02
fix(drive): derive wasm-drive-verify platform version cap from rs-platform-version

QuantumExplorer:

`packages/wasm-drive-verify` hardcoded `MAX_PLATFORM_VERSION = 9` while the workspace's latest protocol version is 13 (`rs-platform-version` `LATEST_VERSION`). `get_platform_version_with_validation` gates the wasm verify entry points (document and identity proof verification), so any caller passing platform versions 10–13 was rejected with "exceeds maximum supported version". Nothing in the crate is version-specific — 9 was simply the latest version when the file was written, and the constant went stale.

The associated unit tests couldn't catch the drift: besides looping up to the constant itself, they were gated `#[cfg(all(test, target_arch = "wasm32"))]` with plain `#[test]` — a combination neither the native test harness nor the wasm-bindgen test runner ever executes.

## What was done?

- Derive `MIN_PLATFORM_VERSION`/`MAX_PLATFORM_VERSION` from `rs-platform-version`'s `INITIAL_PROTOCOL_VERSION`/`LATEST_VERSION` (via `dpp::version` re-exports), so the cap moves with the workspace and can no longer drift. `PLATFORM_VERSIONS` is contiguous and `PlatformVersion::get` indexes `version - 1`, so the range check and the getter stay in agreement by construction.
- Split the range checks and version resolution into a private `JsValue`-free `validate_platform_version` helper; the public function is now a thin wrapper that maps error details through `format_error`. Error messages are unchanged. This is what makes the error path testable natively, since `JsValue` construction aborts on non-wasm targets.
- Un-gated the tests to plain `#[cfg(test)]` so they run in the native workspace test job (which strips `cdylib` and runs this crate under nextest). The valid-range test now genuinely resolves every version from 1 through `LATEST_VERSION`, so it fails loudly if `LATEST_VERSION` ever advances past the `PLATFORM_VERSIONS` list.

## How Has This Been Tested?

- `cargo test -p wasm-drive-verify --lib` — both new tests execute natively and pass (previously this module contributed zero running tests).
- `cargo check -p wasm-drive-verify --target wasm32-unknown-unknown` — the crate's real target still compiles.
- `cargo fmt --all --check` and `cargo clippy -p wasm-drive-verify` are clean.

## Breaking Changes

None. The public API is unchanged; versions 1–9 validate exactly as before, and 10–13 now validate instead of being rejected.

## Checklist:

### DIFF
--- a/packages/wasm-drive-verify/src/utils/platform_version.rs
+++ b/packages/wasm-drive-verify/src/utils/platform_version.rs
@@ -1,75 +1,59 @@
 //! Platform version validation utilities
 
 use crate::utils::error::{format_error, ErrorCategory};
-use dpp::version::PlatformVersion;
+use dpp::version::{PlatformVersion, INITIAL_PROTOCOL_VERSION, LATEST_VERSION};
 use wasm_bindgen::JsValue;
 
 /// Minimum supported platform version
-pub const MIN_PLATFORM_VERSION: u32 = 1;
+pub const MIN_PLATFORM_VERSION: u32 = INITIAL_PROTOCOL_VERSION;
 
 /// Maximum supported platform version
-/// This should be updated when new versions are released
-pub const MAX_PLATFORM_VERSION: u32 = 9;
+pub const MAX_PLATFORM_VERSION: u32 = LATEST_VERSION;
 
 /// Validate and get a platform version with range checks
 pub fn get_platform_version_with_validation(
     version_number: u32,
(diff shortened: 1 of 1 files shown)


thepastaclaw:
✅ Final review complete — no blockers (commit 9d23da7d)

QuantumExplorer (commented):
Approved

PR #4272, opened 2026-08-03
feat(dpp)!: dashpay payment detection keys and stealth derivation

QuantumExplorer:

Second slice of DIP-33 (dashpay/dips#188), stacked on #4380 (the public tips tier — review that first). This PR adds the notified tier's key material and math: identity payment detection keys and stealth one-time address derivation for the Core and Platform transparent rails.

## What was done?

- **Key purposes** `PAYMENT_SCAN = 7` / `PAYMENT_SPEND = 8` (protocol version 14): ECDSA_SECP256K1 only, exactly 33-byte compressed SEC1 points, non-signing, no contract bounds, not searchable, at most one active key per purpose per identity.
- **Validation**: structure validation v1 (per-transition purpose/type/size checks, gated at PV14 via state-transition method versions v2) + the against-state uniqueness check in the identity-update path, with the same-transition rotation exception.
- **Consensus errors** `InvalidKeyPurposeKeyTypeError` (10535) and `TooManyPublicKeysOfPurposeError` (10536), appended at the BasicError tail with discriminant-pinning tests.
- **Stealth math**: `platform-encryption::stealth` (pure curve ops, KAT vectors matching DIP-33) and the platform-wallet module (DIP-9 feature 33' derivation; payer/receiver/spender roles; rail-domain-separated one-time destinations for Core P2PKH and Platform addresses; typed retryable errors).
- **Binding mirrors**: wasm-dpp/wasm-dpp2 (incl. string aliases and strict numeric parsing), Swift `KeyPurpose` (with unknown-discriminant skip instead of coercion), Kotlin `KeyPurpose`.

## How Has This Been Tested?

- dpp: structure-validation tests (type, size incl. 65-byte uncompressed rejection, per-transition duplicates) and BasicError discriminant pins.
- drive-abci: identity-update state tests — add-only conflict, same-transition rotation, disabled-historical-key replacement, scan/spend independence, PV13 rejection / PV14 acceptance through the full pipeline.
- Workspace clippy (all-features) clean; Kotlin `:sdk:compileDebugKotlin` passes; Swift covered by CI.

## Breaking Changes

Consensus-breaking at protocol version 14 (new key purposes accepted by structure validation). Pre-release; no migration concerns.

## Checklist:


thepastaclaw:
🕓 Review not started yet because this PR is a draft.

- [ ] **Request normal review** — click when the PR is ready for review.
- [ ] **Request priority review** — click to move this review to the front of the queue.

_Commit 79154a2f. Normal review starts when eligible; priority review starts as soon as a slot is available._

QuantumExplorer:
## CI fixes + one design change worth a look

Three CI rounds fixed. The third surfaced something I think is a genuine improvement, so flagging it explicitly rather than burying it in a commit.

### Dropped the Base58 `pattern` from `corePaymentAddress`

The pattern **was** working — I verified documents carrying a non-Base58 value were rejected at consensus with a `JsonSchemaError`. But:

* it could only check the **character class**, not the checksum or network byte, so clients had to fully validate the address regardless — the on-chain check bought a partial guarantee for a field the client must re-validate anyway;
* it made every random-document generator using `FillIfNotRequired` emit schema-invalid dashpay profiles, breaking **16 unrelated document tests** (creation, replacement, deletion) and taxing anyone who generates a profile fixture in future.

The tell is the asymmetry: `platformPaymentAddress` (a 21-byte `byteArray`) broke **nothing**, because any random 21 bytes is schema-valid. Consensus now constrains length only, and full validation is explicitly the client's job. [dashpay/dips#188](https://github.com/dashpay/dips/pull/188) updated to match.

If you'd rather keep stricter on-chain validation, the cleaner version is to store the Core address as 21 bytes (version byte + hash160) exactly like the platform one — symmetric across both rails, smaller, and random-fill-safe. That's a bigger DIP change so I didn't take it unilaterally; happy to switch.

### Rebaselined PV14 expectations

The larger dashpay v2 contract shifts grovedb node layout, so several pinned values moved. **All updates are `latest`-version only — every historical protocol-version pin is untouched**, so chain replay stays bit-for-bit reproducible:

* `check_tx` contract create/update processing fees
* deterministic root hash after contract insertion
* happy-path replace and delete fee baselines
* expected profile document strings (now include the two new optional fields)

### Verification

`drive-abci` 2630 passed / 0 failed · `dpp` 3884 / 0 · `drive` 3261+ / 0 · `platform-wallet` green · workspace `cargo fmt --check` and CI-style `cargo clippy --workspace --all-features -D warnings` both clean.

Earlier rounds: PR title needed a lowercase subject; the wasm `PurposeWasm` mirror enums needed `#[allow(non_camel_case_types)]` (underscored variants trip the lint where single-word all-caps don't); and a stealth test used `pubkey_hash()`, which `key_wallet::Address` doesn't have — that one only broke the test target, which `cargo check`/`clippy` don't build.

packages/rs-dpp/src/errors/consensus/basic/basic_error.rs
```diff
@@ -603,6 +604,12 @@ pub enum BasicError {
     #[error(transparent)]
     InvalidKeyPurposeForContractBoundsError(InvalidKeyPurposeForContractBoundsError),
 
+    #[error(transparent)]
+    InvalidKeyPurposeKeyTypeError(InvalidKeyPurposeKeyTypeError),
+
+    #[error(transparent)]
+    TooManyPublicKeysOfPurposeError(TooManyPublicKeysOfPurposeError),
```
thepastaclaw, line 611: **🔴 Blocking: Append the new BasicError variants to preserve wire discriminants**

BasicError derives bincode Encode/Decode and PlatformSerialize/PlatformDeserialize, so its enum discriminants are positional. Inserting these variants before StateTransitionNotActiveError shifts every existing variant through TokenPricingScheduleEmptyError, causing previously encoded errors to decode as the wrong variant or fail. This directly violates the ordering invariant documented at the enum tail. Move both new variants after TokenPricingScheduleEmptyError without changing the order of any existing variant, and add a compatibility test that pins a pre-existing variant from after the current insertion point.

<sub>source: ['codex']</sub>

packages/rs-dpp/src/state_transition/state_transitions/identity/public_key_in_creation/methods/validate_identity_public_keys_structure/v1/mod.rs
```diff
@@ -0,0 +1,314 @@
...
+            return Ok(SimpleConsensusValidationResult::new_with_error(
+                BasicError::InvalidKeyPurposeKeyTypeError(InvalidKeyPurposeKeyTypeError::new(
+                    invalid_type_key.id(),
+                    invalid_type_key.purpose(),
+                    invalid_type_key.key_type(),
+                    vec![KeyType::ECDSA_SECP256K1],
+                ))
+                .into(),
+            ));
+        }
```
thepastaclaw, line 161: **🔴 Blocking: Require compressed 33-byte encodings for payment keys**

This check constrains payment keys to ECDSA_SECP256K1 but does not constrain their encoding. DPP's ECDSA public-key parser explicitly accepts both 33-byte compressed and 65-byte uncompressed SEC1 keys, and the generic added-key signature verifier passes either representation to PublicKey::from_slice. A signed transition can therefore register a 65-byte PAYMENT_SCAN or PAYMENT_SPEND key even though DIP-33 specifies compressed SEC1 points and this validator's own comment requires the full compressed key. Besides producing noncanonical persisted state, compressed and uncompressed encodings of the same point also evade the in-transition duplicate check, which compares raw bytes. Reject payment-purpose keys unless data().len() is exactly 33 and add a regression test using a valid 65-byte key.

<sub>source: ['codex']</sub>

packages/wasm-dpp2/src/enums/keys/purpose.rs
```diff
@@ -59,6 +62,8 @@ impl TryFrom<&JsValue> for PurposeWasm {
                 "system" => Ok(PurposeWasm::SYSTEM),
                 "voting" => Ok(PurposeWasm::VOTING),
                 "owner" => Ok(PurposeWasm::OWNER),
+                "paymentscan" => Ok(PurposeWasm::PAYMENT_SCAN),
+                "paymentspend" => Ok(PurposeWasm::PAYMENT_SPEND),
```
thepastaclaw, line 66: **🟡 Suggestion: Accept the payment-purpose strings emitted by the WASM getter**

The IdentityPublicKey purpose getter emits PAYMENT_SCAN and PAYMENT_SPEND. The setter lowercases those strings to payment_scan and payment_spend, but this parser accepts only paymentscan and paymentspend. Consequently, assigning key.purpose = key.purpose or copying a key through its exposed properties fails only for the new variants. Accept the underscore forms as aliases so getter values round-trip through constructors and setters.

```suggestion
                "paymentscan" | "payment_scan" => Ok(PurposeWasm::PAYMENT_SCAN),
                "paymentspend" | "payment_spend" => Ok(PurposeWasm::PAYMENT_SPEND),
```

<sub>source: ['codex']</sub>

packages/rs-drive-abci/src/execution/validation/state_transition/common/validate_payment_key_uniqueness/v0/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+                && !public_key_ids_to_disable.contains(&key.id())
+        });
+        if conflicting_active_key_exists {
+            return Ok(SimpleConsensusValidationResult::new_with_error(
+                TooManyPublicKeysOfPurposeError::new(purpose, 1).into(),
+            ));
+        }
+    }
+
+    Ok(SimpleConsensusValidationResult::new())
```
thepastaclaw, line 84: **🟡 Suggestion: Cover payment-key uniqueness against persisted identity state**

This helper implements the consensus-critical against-state half of the one-active-key invariant, including the state-sensitive rotation exception, but rs-drive-abci contains no PAYMENT_SCAN or PAYMENT_SPEND tests. Add tests proving that an add-only update conflicts with an existing active key, disabling that exact key while adding a replacement succeeds, disabled historical keys do not block replacement, and scan/spend purposes remain independent. A pipeline-level test should also pin rejection before protocol version 14 and acceptance at version 14.

<sub>source: ['codex']</sub>

packages/wasm-dpp2/src/enums/keys/purpose.rs
```diff
@@ -75,6 +80,8 @@ impl TryFrom<&JsValue> for PurposeWasm {
                 4 => Ok(PurposeWasm::SYSTEM),
                 5 => Ok(PurposeWasm::VOTING),
                 6 => Ok(PurposeWasm::OWNER),
+                7 => Ok(PurposeWasm::PAYMENT_SCAN),
+                8 => Ok(PurposeWasm::PAYMENT_SPEND),
```
thepastaclaw, line 84: **🟡 Suggestion: Reject fractional numeric payment-purpose values**

PurposeLike's TypeScript union is not a runtime constraint. Casting an arbitrary JavaScript number directly to u8 truncates fractional values, so 7.5 is accepted as PAYMENT_SCAN and 8.5 as PAYMENT_SPEND. Validate that the number is finite, integral, and within the supported 0 through 8 range before casting. Rust's float-to-integer cast saturates rather than wraps, so positive overflow currently becomes 255 and is rejected; the newly introduced defect is specifically the acceptance of fractional values that truncate to 7 or 8.

packages/rs-platform-wallet/src/wallet/identity/crypto/stealth.rs
```diff
@@ -0,0 +1,346 @@
...
+    spend_secret: &SecretKey,
+    ephemeral_public: &PublicKey,
+    rail: PaymentRail,
+    n: u32,
+) -> Result<SecretKey, PlatformWalletError> {
+    let secp = Secp256k1::new();
+    let shared = stealth_shared_point(&secp, scan_secret, ephemeral_public)
+        .map_err(|e| PlatformWalletError::InvalidIdentityData(e.to_string()))?;
+    one_time_secret_key(spend_secret, &shared, ephemeral_public, rail.into(), n)
+        .map_err(|e| PlatformWalletError::InvalidIdentityData(e.to_string()))
```
thepastaclaw, line 229: **🟡 Suggestion: Preserve retryable stealth crypto errors**

The lower crypto layer exposes ZeroStealthTweak as a recoverable condition that requires the payer to choose a fresh ephemeral key, but every stealth operation converts CryptoError into InvalidIdentityData(String). Callers cannot distinguish the retry condition without parsing display text. Add a typed PlatformWalletError source variant for platform_encryption::CryptoError, or a dedicated zero-tweak variant, and propagate it with ?. Errors constructing or deriving the DIP-9 path should likewise use the existing KeyDerivation category rather than InvalidIdentityData.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The DIP-33 implementation is generally well-structured and version-dispatched, but three blocking issues remain: positional BasicError compatibility is broken, noncanonical 65-byte payment keys are accepted, and Swift rewrites the new key-purpose discriminants during persistence. Additional WASM and Kotlin boundary issues, retryable error flattening, and missing state-validation coverage should also be addressed.

### Review provenance

🔴 3 blocking | 🟡 5 suggestion(s)

<details>

```

In `packages/rs-dpp/src/errors/consensus/basic/basic_error.rs`:
- [BLOCKING] packages/rs-dpp/src/errors/consensus/basic/basic_error.rs:607-611: Append the new BasicError variants to preserve wire discriminants
  BasicError derives bincode Encode/Decode and PlatformSerialize/PlatformDeserialize, so its enum discriminants are positional. Inserting these variants before StateTransitionNotActiveError shifts every existing variant through TokenPricingScheduleEmptyError, causing previously encoded errors to decode as the wrong variant or fail. This directly violates the ordering invariant documented at the enum tail. Move both new variants after TokenPricingScheduleEmptyError without changing the order of any existing variant, and add a compatibility test that pins a pre-existing variant from after the current insertion point.

In `packages/rs-dpp/src/state_transition/state_transitions/identity/public_key_in_creation/methods/validate_identity_public_keys_structure/v1/mod.rs`:
- [BLOCKING] packages/rs-dpp/src/state_transition/state_transitions/identity/public_key_in_creation/methods/validate_identity_public_keys_structure/v1/mod.rs:148-161: Require compressed 33-byte encodings for payment keys
  This check constrains payment keys to ECDSA_SECP256K1 but does not constrain their encoding. DPP's ECDSA public-key parser explicitly accepts both 33-byte compressed and 65-byte uncompressed SEC1 keys, and the generic added-key signature verifier passes either representation to PublicKey::from_slice. A signed transition can therefore register a 65-byte PAYMENT_SCAN or PAYMENT_SPEND key even though DIP-33 specifies compressed SEC1 points and this validator's own comment requires the full compressed key. Besides producing noncanonical persisted state, compressed and uncompressed encodings of the same point also evade the in-transition duplicate check, which compares raw bytes. Reject payment-purpose keys unless data().len() is exactly 33 and add a regression test using a valid 65-byte key.

- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift:1798-1804: Preserve payment-key purposes across the Swift FFI boundary
  Rust emits each DPP purpose as its raw u8 value, so the new PAYMENT_SCAN and PAYMENT_SPEND keys reach Swift as 7 and 8. Swift's KeyPurpose enum still ends at 6, and this fallback silently persists either new value as authentication. The cold-start restore path later sends that stored raw value back to Rust, changing the key's meaning, while ManagedIdentity.getPublicKeys() rejects the same values as unknown. Add paymentScan/paymentSpend cases with raw values 7 and 8, update exhaustive switches, and replace this semantic fallback with an unsupported-discriminant error rather than coercing an unknown purpose to authentication.

In `packages/wasm-dpp2/src/enums/keys/purpose.rs`:
- [SUGGESTION] packages/wasm-dpp2/src/enums/keys/purpose.rs:65-66: Accept the payment-purpose strings emitted by the WASM getter
  The IdentityPublicKey purpose getter emits PAYMENT_SCAN and PAYMENT_SPEND. The setter lowercases those strings to payment_scan and payment_spend, but this parser accepts only paymentscan and paymentspend. Consequently, assigning key.purpose = key.purpose or copying a key through its exposed properties fails only for the new variants. Accept the underscore forms as aliases so getter values round-trip through constructors and setters.
- [SUGGESTION] packages/wasm-dpp2/src/enums/keys/purpose.rs:74-84: Reject fractional numeric payment-purpose values
  PurposeLike's TypeScript union is not a runtime constraint. Casting an arbitrary JavaScript number directly to u8 truncates fractional values, so 7.5 is accepted as PAYMENT_SCAN and 8.5 as PAYMENT_SPEND. Validate that the number is finite, integral, and within the supported 0 through 8 range before casting. Rust's float-to-integer cast saturates rather than wraps, so positive overflow currently becomes 255 and is rejected; the newly introduced defect is specifically the acceptance of fractional values that truncate to 7 or 8.

In `packages/rs-drive-abci/src/execution/validation/state_transition/common/validate_payment_key_uniqueness/v0/mod.rs`:
- [SUGGESTION] packages/rs-drive-abci/src/execution/validation/state_transition/common/validate_payment_key_uniqueness/v0/mod.rs:37-84: Cover payment-key uniqueness against persisted identity state
  This helper implements the consensus-critical against-state half of the one-active-key invariant, including the state-sensitive rotation exception, but rs-drive-abci contains no PAYMENT_SCAN or PAYMENT_SPEND tests. Add tests proving that an add-only update conflicts with an existing active key, disabling that exact key while adding a replacement succeeds, disabled historical keys do not block replacement, and scan/spend purposes remain independent. A pipeline-level test should also pin rejection before protocol version 14 and acceptance at version 14.

In `packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/identity/IdentityUpdates.kt`:
- [SUGGESTION] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/identity/IdentityUpdates.kt:27-35: Expose the new DPP key purposes in Kotlin
  The generic Kotlin identity registration and update codecs serialize KeyPurpose.ffiValue, but this mirror stops at OWNER=6. Kotlin callers therefore cannot construct PAYMENT_SCAN=7 or PAYMENT_SPEND=8 rows even though the Rust decoder now supports them. Add both enum cases and update the example application's exhaustive when expressions and purpose-selection logic so Android callers can use the protocol-version-14 feature.

In `packages/rs-platform-wallet/src/wallet/identity/crypto/stealth.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/crypto/stealth.rs:171-229: Preserve retryable stealth crypto errors
  The lower crypto layer exposes ZeroStealthTweak as a recoverable condition that requires the payer to choose a fresh ephemeral key, but every stealth operation converts CryptoError into InvalidIdentityData(String). Callers cannot distinguish the retry condition without parsing display text. Add a typed PlatformWalletError source variant for platform_encryption::CryptoError, or a dedicated zero-tweak variant, and propagate it with ?. Errors constructing or deriving the DIP-9 path should likewise use the existing KeyDerivation category rather than InvalidIdentityData.
```
</details>

QuantumExplorer:
Addressed the review in b2b51fbf77 + 07c5630ce4.

**Blockers**
- **BasicError wire discriminants**: both new variants moved to the enum tail; added tests pinning the encoded discriminants of `StateTransitionNotActiveError` (143) and `TokenPricingScheduleEmptyError` (172) plus the new tail positions (173/174), so any future mid-enum insertion fails CI.
- **65-byte payment keys**: structure validation v1 now requires `data().len() == 33` for payment purposes (rejected as `InvalidIdentityPublicKeyDataError`), with a regression test that re-encodes a valid compressed key uncompressed.
- **Swift purpose coercion**: `KeyPurpose` gains `paymentScan = 7` / `paymentSpend = 8`; the persistence path no longer coerces unknown discriminants to `.authentication` — it skips the row with a warning (purpose, level, and type are all guarded), so a stored raw value can never change meaning on cold-start restore.

**Suggestions**
- wasm-dpp2 parser accepts `payment_scan`/`payment_spend` (the lowercased getter output) and rejects non-finite/fractional/out-of-range numeric purposes instead of truncating.
- Kotlin `KeyPurpose` mirror gains both purposes (+ example-app display arms).
- platform-wallet stealth ops propagate `platform_encryption::CryptoError` through a typed `StealthCrypto` variant (retryable `ZeroStealthTweak` is now matchable); derivation-path errors use `KeyDerivation`.
- drive-abci state tests for the one-active-payment-key invariant: add-only conflict, same-transition rotation, disabled-historical-key replacement, scan/spend independence, and PV13 rejection / PV14 acceptance through the full pipeline.

**Also in this push** (07c5630ce4): `corePaymentAddress` is now the 21-byte storage form (variant byte + HASH160), symmetric with `platformPaymentAddress`, per the open question on this PR — DIP-33 updated to match in dashpay/dips#188. PV14 fee and root-hash baselines regenerated.

Verified locally: dpp 3893 tests, drive-abci full suite, drive, platform-wallet/encryption, dashpay-contract all green; workspace clippy (all-features) clean; Kotlin `:sdk:compileDebugKotlin` passes; Swift changes parse (xcframework build is CI's).

QuantumExplorer:
**PR split**: the public tips tier (dashpay contract v2 + `corePaymentAddress`/`platformPaymentAddress` profile fields, PV14 plumbing, cache guard, and all fee/root-hash rebaselines) has been extracted to #4380, and this PR is now stacked on it (base: `feat/dashpay-tips-jar`). This PR narrows to the notified tier: `PAYMENT_SCAN`/`PAYMENT_SPEND` key purposes with structure + state validation, the stealth derivation modules, and the language-binding mirrors. The merge from the new base also folds in ~9 days of v4.2-dev (the only interaction was the new `drive_abci_validation_versions/v10.rs`, which gained the `validate_payment_key_uniqueness` field). Review order: #4380 first, then this.

packages/rs-dpp/src/state_transition/state_transitions/identity/public_key_in_creation/methods/validate_identity_public_keys_structure/v1/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+                    InvalidIdentityPublicKeyDataError::new(
+                        invalid_size_key.id(),
+                        PublicKeyValidationError::new(
+                            "payment keys must be 33-byte compressed SEC1 points",
+                        ),
+                    ),
+                )
+                .into(),
+            ));
+        }
```
thepastaclaw, line 182: **🔴 Blocking: Reject contract bounds on payment keys during structure validation**

The payment-key checks enforce key type and compressed encoding but do not reject `contract_bounds`. Identity updates eventually call `validate_identity_public_keys_contract_bounds`, but the identity-create, identity-create-from-addresses, and identity-create-from-shielded-pool state paths do not run that validation. A creation transition can therefore pass consensus validation with a `PAYMENT_SCAN` or `PAYMENT_SPEND` key whose bounds are present. During identity insertion, `add_potential_contract_info_for_contract_bounded_key_v0` rejects every purpose other than `ENCRYPTION` or `DECRYPTION` with the internal `IdentityKeyBoundsError("purpose not available for key bounds")`. Reject payment-purpose keys with bounds in this shared structure validator using `InvalidKeyPurposeForContractBoundsError`, and add creation-path regression coverage so user-controlled input produces a deterministic consensus error instead of an execution failure.

<sub>source: ['codex']</sub>

packages/rs-drive-abci/src/execution/validation/state_transition/common/validate_payment_key_uniqueness/v0/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+        identity_id: identity_id.to_buffer(),
+        request_type: KeyRequestType::AllKeys,
+        limit: None,
+        offset: None,
+    };
+    let existing_keys = drive.fetch_identity_keys::<KeyIDIdentityPublicKeyPairBTreeMap>(
+        identity_key_request,
+        transaction,
+        platform_version,
+    )?;
```
thepastaclaw, line 69: **🔴 Blocking: Meter the all-keys lookup used by payment-key uniqueness validation**

Every update adding a payment key performs an `AllKeys` query over the identity's complete key history. `fetch_identity_keys` builds GroveDB cost operations and then discards them, while this helper deliberately ignores `_execution_context`, so the potentially large read and deserialization work is never charged. Repeated payment-key rotations can therefore force validators to rescan a large identity cheaply. Use `fetch_identity_keys_with_costs`, pass the committed epoch, and add the resulting fee to the actual processor-owned context as `ValidationOperation::PrecalculatedOperation`. The identity-update implementation currently creates a separate local context while ignoring the outer context passed by the processor, so that context must also be threaded through; otherwise the operation will still be absent from the resulting `ExecutionEvent`.

<sub>source: ['codex']</sub>

packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/identity_update/state/v0/mod.rs
```diff
@@ -120,6 +121,34 @@ impl IdentityUpdateStateTransitionStateValidationV0 for IdentityUpdateTransition
...
+                self.identity_id(),
+                self.public_keys_to_add(),
+                self.public_key_ids_to_disable(),
+                drive,
+                &mut state_transition_execution_context,
+                tx,
+                platform_version,
+            )?
+            .errors,
+        );
```
thepastaclaw, line 139: **🔴 Blocking: Introduce a new identity-update state-validation version**

This inserts the payment-key uniqueness rule directly into `validate_state_v0`, while every Drive-ABCI validation table still selects identity-update state version 0. Consensus-critical version implementations must remain behaviorally immutable. Keep v0 unchanged, add a `state/v1` implementation containing the new rule and the correctly threaded execution context, extend the dispatcher to versions 0 and 1, and select version 1 only for protocol version 14. Tests should explicitly show that version 13 continues through v0 and version 14 uses v1.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/identity/crypto/stealth.rs
```diff
@@ -0,0 +1,342 @@
...
+
+/// DIP-9 feature index for DashPay payment detection keys (matches DIP-33).
+pub const FEATURE_PURPOSE_PAYMENT_DETECTION: u32 = 33;
+
+/// `key_class'` for the scan key.
+pub const KEY_CLASS_SCAN: u32 = 0;
+/// `key_class'` for the spend key.
+pub const KEY_CLASS_SPEND: u32 = 1;
+/// `key_class'` for the notification out-key.
+pub const KEY_CLASS_NOTIF_OUT: u32 = 2;
```
thepastaclaw, line 56: **🟡 Suggestion: Represent the closed payment key classes with an enum**

DIP-33 defines the scan, spend, and notification-out key classes, but the public derivation functions accept any `u32`. Values outside 0 through 2 still produce valid hardened BIP32 paths, so callers can derive nonstandard keys without receiving an error. Represent the role as a `PaymentKeyClass` enum and convert it to the hardened index internally. This makes invalid roles unrepresentable and gives callers an exhaustive API if another role is introduced later.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/identity/crypto/stealth.rs
```diff
@@ -0,0 +1,342 @@
...
+) -> Result<SecretKey, PlatformWalletError> {
+    let secp = Secp256k1::new();
+    let shared = stealth_shared_point(&secp, scan_secret, ephemeral_public)?;
+    Ok(one_time_secret_key(
+        spend_secret,
+        &shared,
+        ephemeral_public,
+        rail.into(),
+        n,
+    )?)
```
thepastaclaw, line 225: **🟡 Suggestion: Reuse the secp256k1 context during destination scanning**

Each payer, receiver, and spender entry point constructs a fresh `Secp256k1` context. `recognize_one_time_destination` is intended to run repeatedly while scanning candidate notifications and output counters, making repeated context initialization unnecessary work on the hottest path. Accept a caller-owned context, provide a scanner object that owns one, or use the crate's shared context, then pass it through to the lower functions that already accept a borrowed context.

<sub>source: ['codex']</sub>

packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/identity_update/mod.rs
```diff
@@ -1845,4 +1845,388 @@ mod tests {
...
+            )
+            .await;
+            assert!(
+                !matches!(
+                    results.as_slice(),
+                    [StateTransitionExecutionResult::SuccessfulExecution { .. }]
+                ),
+                "payment keys must be rejected before protocol version 14, got: {:?}",
+                results
+            );
```
thepastaclaw, line 2229: **🟡 Suggestion: Pin the activation test to protocol version 14**

The acceptance test uses `PlatformVersion::latest()`, so it will silently move to later protocol versions and stop protecting the version-14 table. Select `PlatformVersion::get(14)` explicitly. The version-13 test should also assert the expected paid `InvalidIdentityPublicKeySecurityLevelError` rather than merely checking that execution was unsuccessful; the current assertion passes for unrelated signature, nonce, serialization, balance, or internal failures and therefore does not prove the activation boundary.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The payment-key model and stealth derivation are generally well structured, but three consensus blockers remain: creation paths accept contract-bounded payment keys that later fail internally, the uniqueness scan discards its query costs, and the new state rule modifies the existing identity-update v0 implementation. The Swift legacy C bridge also cannot decode the new purposes, and several public-API and activation-test improvements remain advisable.
Source: reviewer backends gpt-5.6-sol (general, security-auditor, rust-quality, ffi-engineer); final verifier backend gpt-5.6-sol. openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

### Review provenance

🔴 3 blocking | 🟡 4 suggestion(s)

<details>

```

In `packages/rs-dpp/src/state_transition/state_transitions/identity/public_key_in_creation/methods/validate_identity_public_keys_structure/v1/mod.rs`:
- [BLOCKING] packages/rs-dpp/src/state_transition/state_transitions/identity/public_key_in_creation/methods/validate_identity_public_keys_structure/v1/mod.rs:150-182: Reject contract bounds on payment keys during structure validation
  The payment-key checks enforce key type and compressed encoding but do not reject `contract_bounds`. Identity updates eventually call `validate_identity_public_keys_contract_bounds`, but the identity-create, identity-create-from-addresses, and identity-create-from-shielded-pool state paths do not run that validation. A creation transition can therefore pass consensus validation with a `PAYMENT_SCAN` or `PAYMENT_SPEND` key whose bounds are present. During identity insertion, `add_potential_contract_info_for_contract_bounded_key_v0` rejects every purpose other than `ENCRYPTION` or `DECRYPTION` with the internal `IdentityKeyBoundsError("purpose not available for key bounds")`. Reject payment-purpose keys with bounds in this shared structure validator using `InvalidKeyPurposeForContractBoundsError`, and add creation-path regression coverage so user-controlled input produces a deterministic consensus error instead of an execution failure.

In `packages/rs-drive-abci/src/execution/validation/state_transition/common/validate_payment_key_uniqueness/v0/mod.rs`:
- [BLOCKING] packages/rs-drive-abci/src/execution/validation/state_transition/common/validate_payment_key_uniqueness/v0/mod.rs:37-69: Meter the all-keys lookup used by payment-key uniqueness validation
  Every update adding a payment key performs an `AllKeys` query over the identity's complete key history. `fetch_identity_keys` builds GroveDB cost operations and then discards them, while this helper deliberately ignores `_execution_context`, so the potentially large read and deserialization work is never charged. Repeated payment-key rotations can therefore force validators to rescan a large identity cheaply. Use `fetch_identity_keys_with_costs`, pass the committed epoch, and add the resulting fee to the actual processor-owned context as `ValidationOperation::PrecalculatedOperation`. The identity-update implementation currently creates a separate local context while ignoring the outer context passed by the processor, so that context must also be threaded through; otherwise the operation will still be absent from the resulting `ExecutionEvent`.

In `packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/identity_update/state/v0/mod.rs`:
- [BLOCKING] packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/identity_update/state/v0/mod.rs:124-139: Introduce a new identity-update state-validation version
  This inserts the payment-key uniqueness rule directly into `validate_state_v0`, while every Drive-ABCI validation table still selects identity-update state version 0. Consensus-critical version implementations must remain behaviorally immutable. Keep v0 unchanged, add a `state/v1` implementation containing the new rule and the correctly threaded execution context, extend the dispatcher to versions 0 and 1, and select version 1 only for protocol version 14. Tests should explicitly show that version 13 continues through v0 and version 14 uses v1.

In `packages/rs-sdk-ffi/src/identity/create_from_components.rs`:
- [SUGGESTION] packages/rs-sdk-ffi/src/identity/create_from_components.rs:111-125: Synchronize the legacy Rust C API with the new Swift purpose values
  Swift now exposes `paymentScan = 7` and `paymentSpend = 8`, and `identityToHandle` forwards those raw values through `DashSDKPublicKeyData.purpose`. This decoder only accepts purposes 0 through 6, so converting a `DPPIdentity` containing either new key into a native identity handle fails with `Invalid key purpose`; document and token operations that call `identityToHandle` consequently cannot use such identities. The standalone decoder in `packages/rs-sdk-ffi/src/identity/keys.rs:222-235` is also stale. Use the existing `Purpose::try_from` conversion, or explicitly accept 7 and 8 in both C entry points, so the Swift mirror and Rust boundary share the same discriminants.

In `packages/rs-platform-wallet/src/wallet/identity/crypto/stealth.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/crypto/stealth.rs:51-56: Represent the closed payment key classes with an enum
  DIP-33 defines the scan, spend, and notification-out key classes, but the public derivation functions accept any `u32`. Values outside 0 through 2 still produce valid hardened BIP32 paths, so callers can derive nonstandard keys without receiving an error. Represent the role as a `PaymentKeyClass` enum and convert it to the hardened index internally. This makes invalid roles unrepresentable and gives callers an exhaustive API if another role is introduced later.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/crypto/stealth.rs:157-225: Reuse the secp256k1 context during destination scanning
  Each payer, receiver, and spender entry point constructs a fresh `Secp256k1` context. `recognize_one_time_destination` is intended to run repeatedly while scanning candidate notifications and output counters, making repeated context initialization unnecessary work on the hottest path. Accept a caller-owned context, provide a scanner object that owns one, or use the crate's shared context, then pass it through to the lower functions that already accept a borrowed context.

In `packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/identity_update/mod.rs`:
- [SUGGESTION] packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/identity_update/mod.rs:2010-2229: Pin the activation test to protocol version 14
  The acceptance test uses `PlatformVersion::latest()`, so it will silently move to later protocol versions and stop protecting the version-14 table. Select `PlatformVersion::get(14)` explicitly. The version-13 test should also assert the expected paid `InvalidIdentityPublicKeySecurityLevelError` rather than merely checking that execution was unsuccessful; the current assertion passes for unrelated signature, nonce, serialization, balance, or internal failures and therefore does not prove the activation boundary.
```
</details>

packages/rs-platform-wallet/src/wallet/identity/crypto/stealth.rs
thepastaclaw: Resolved in `b2b51fb` — _Preserve retryable stealth crypto errors_ no longer present.

packages/rs-drive-abci/src/execution/validation/state_transition/common/validate_payment_key_uniqueness/v0/mod.rs
thepastaclaw: Resolved in this update — _Cover payment-key uniqueness against persisted identity state_ no longer present.

packages/wasm-dpp2/src/enums/keys/purpose.rs
thepastaclaw: Resolved in `b2b51fb` — _Accept the payment-purpose strings emitted by the WASM getter_ no longer present.

packages/rs-dpp/src/state_transition/state_transitions/identity/public_key_in_creation/methods/validate_identity_public_keys_structure/v1/mod.rs
thepastaclaw: Resolved in `b2b51fb` — _Require compressed 33-byte encodings for payment keys_ no longer present.

packages/rs-dpp/src/errors/consensus/basic/basic_error.rs
thepastaclaw: Resolved in `b2b51fb` — _Append the new BasicError variants to preserve wire discriminants_ no longer present.

packages/wasm-dpp2/src/enums/keys/purpose.rs
thepastaclaw: Resolved in `b2b51fb` — _Reject fractional numeric payment-purpose values_ no longer present.

packages/rs-platform-encryption/src/stealth.rs
```diff
@@ -0,0 +1,313 @@
...
+
+    /// DIP-33 test vector inputs (test-only scalars).
+    const B_SCAN: &str = "61aafd85dbca17133515038343b05ed2019ca465d1ba93dcbffbf9534d2f436c";
+    const B_SPEND: &str = "8f54f1cfa054739a1d4a87847bdf6a024caf78ff4b2c91aeb75cb843280957ff";
+    const R_EPHEMERAL: &str = "37ca08877b0beea1b1e7655649274689e89944edae537ecad42764e59ac31e3e";
+
+    /// Known-answer test pinning the full DIP-33 vector table: shared point,
+    /// tweaks, one-time public keys, and one-time secret keys for both rails.
+    /// A mismatch here means we broke interop with every other implementation
+    /// of the DIP, not just our own round-trip.
```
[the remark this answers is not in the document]
thepastaclaw: The comment already labels these as DIP-33 test-vector inputs. I also verified that the three input scalars and all expected shared-point, tweak, public-key, and secret-key outputs match the `Test Vectors` section of dashpay/dips#188 byte-for-byte, so the interoperability claim is supported rather than being a locally generated regression lock.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -1795,9 +1795,16 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
...
+                // Never coerce an unknown discriminant: the stored raw value is
+                // sent back to Rust on cold-start restore, so persisting a
+                // substitute silently changes the key's meaning. Skip instead;
+                // the key re-arrives once the enum mirror learns the value.
+                guard let purposeEnum = KeyPurpose(rawValue: entry.purpose),
+                      let levelEnum = SecurityLevel(rawValue: entry.securityLevel),
+                      let keyTypeEnum = KeyType(rawValue: entry.keyType) else {
+                    print("⚠️ persistIdentityKeys: skipped key \(entry.keyId) for identity \(identityHex.prefix(8))… — unsupported purpose/level/type discriminant (\(entry.purpose)/\(entry.securityLevel)/\(entry.keyType))")
+                    continue
+                }
```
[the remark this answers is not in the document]
thepastaclaw: Purpose, security level, and key type are immutable for a given identity-key ID, so existing rows intentionally refresh only mutable fields. The merge base did not support payment purposes, meaning no released/base persistence path could have created the proposed legacy payment rows. In the final diff, new rows validate all three discriminants and preserve 7/8 without coercion.

QuantumExplorer:
**Parked (converted to draft), not abandoned.**

The non-public payment tier has been re-scoped: the public tips tier (#4380, merged) plus shielded-pool payments cover the product need, so the notified/stealth tier this PR implements — `PAYMENT_SCAN`/`PAYMENT_SPEND` identity key purposes, their structure/state validation, and the stealth one-time-address derivation — has no consumer for now, and we don't want to ship dormant consensus surface.

The deciding argument: any payment-time Platform artifact (the payment notification that stealth discovery depends on) is timing-correlatable with the settlement transaction by a global observer, which defeats the unlinkability goal. The clean alternatives are either chain-native scanning (a BIP352-style DAPI tweak index — real new infrastructure) or simply routing private payments through the shielded pool, which already exists and hides the amount as well. We chose the pool.

Everything here remains reviewable and revivable: the pastaclaw round-1 findings are all addressed, the branch stays, and the published detection keys were deliberately made BIP352-shaped so a future scanning-based tier can adopt them unchanged. DIP-33 (dashpay/dips#188) is being redrafted to the reduced scope with this design recorded as considered-and-deferred.

thepastaclaw (changes requested):

The payment-purpose model and stealth derivation are generally coherent, but three consensus blockers remain: creation paths accept contract-bounded payment keys that later fail internally, the uniqueness scan is unmetered, and the new uniqueness behavior was added directly to identity-update state-validation v0. Four API, FFI, performance, and activation-test suggestions also remain valid; Opus review is deferred until the Codex blocker gate clears.

### Review provenance

_6 carried-forward finding(s) already raised on this PR; not re-posting as new inline comments._

<details>

```

In `packages/rs-sdk-ffi/src/identity/create_from_components.rs`:
- [SUGGESTION] packages/rs-sdk-ffi/src/identity/create_from_components.rs:111-125: Synchronize the legacy Rust C API with the new Swift purpose values
  Swift now exposes `paymentScan = 7` and `paymentSpend = 8`, and `identityToHandle` forwards every public key's raw purpose through `DashSDKPublicKeyData`. This decoder still accepts only values 0 through 6, so converting a `DPPIdentity` containing a payment key into a native identity handle fails with `Invalid key purpose`. Document and token operations using this bridge consequently cannot use such identities. The separate decoder in `packages/rs-sdk-ffi/src/identity/keys.rs` is also stale. Use the canonical `Purpose::try_from` conversion, or explicitly accept discriminants 7 and 8 at both C entry points.

In `packages/rs-dpp/src/state_transition/state_transitions/identity/public_key_in_creation/methods/validate_identity_public_keys_structure/v1/mod.rs`:
- [BLOCKING] packages/rs-dpp/src/state_transition/state_transitions/identity/public_key_in_creation/methods/validate_identity_public_keys_structure/v1/mod.rs:150-182: Reject contract bounds on payment keys during structure validation
  (existing thread: https://github.com/dashpay/platform/pull/4272#discussion_r3766545358)
  The payment-purpose checks enforce key type and compressed encoding but do not reject `contract_bounds`, despite the PR defining payment keys as unbounded. Identity updates later call `validate_identity_public_keys_contract_bounds`, but identity-create, identity-create-from-addresses, and identity-create-from-shielded-pool do not perform that validation. Those creation transitions can therefore pass consensus validation with a bounded `PAYMENT_SCAN` or `PAYMENT_SPEND` key and reach Drive insertion, where `add_potential_contract_info_for_contract_bounded_key_v0` rejects every purpose other than encryption or decryption with the internal `IdentityKeyBoundsError("purpose not available for key bounds")`. Reject bounded payment keys here with `InvalidKeyPurposeForContractBoundsError`, and add creation-pipeline coverage so malformed user input produces a deterministic consensus error rather than an internal execution failure.

In `packages/rs-drive-abci/src/execution/validation/state_transition/common/validate_payment_key_uniqueness/v0/mod.rs`:
- [BLOCKING] packages/rs-drive-abci/src/execution/validation/state_transition/common/validate_payment_key_uniqueness/v0/mod.rs:37-69: Meter the all-keys lookup used by payment-key uniqueness validation
  (existing thread: https://github.com/dashpay/platform/pull/4272#discussion_r3766545363)
  Every update adding a payment key performs an unbounded `AllKeys` query over the identity's complete key history. `fetch_identity_keys` executes the GroveDB query and collects its low-level operations, but then discards those operations, while this helper explicitly ignores `_execution_context`. Repeated payment-key rotations can therefore force validators to read and deserialize a growing key history without charging for that work. Use `fetch_identity_keys_with_costs` with the last committed epoch and record its `FeeResult` as `ValidationOperation::PrecalculatedOperation`. The identity-update state implementation must also receive and use the processor-owned execution context instead of its private local context; otherwise the new charge will not be included in the resulting `ExecutionEvent`.

In `packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/identity_update/state/v0/mod.rs`:
- [BLOCKING] packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/identity_update/state/v0/mod.rs:124-139: Introduce a new identity-update state-validation version
  (existing thread: https://github.com/dashpay/platform/pull/4272#discussion_r3766545366)
  The payment-key uniqueness rule was inserted directly into `validate_state_v0`. Both the protocol-13 and protocol-14 Drive-ABCI validation tables still select identity-update state version 0, so the PR changes an existing consensus implementation in place. Restore v0 to its prior behavior, add a `state/v1` implementation containing the uniqueness rule and correctly threaded execution context, extend the dispatcher to support versions 0 and 1, and select version 1 only in the protocol-14 validation table. Add dispatch coverage showing that protocol 13 retains v0 while protocol 14 activates v1.

In `packages/rs-platform-wallet/src/wallet/identity/crypto/stealth.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/crypto/stealth.rs:51-123: Represent the closed payment key classes with an enum
  (existing thread: https://github.com/dashpay/platform/pull/4272#discussion_r3766545370)
  The module defines the three DIP-33 key classes—scan, spend, and notification out—as constants, but both public derivation APIs accept an unrestricted `u32`. Any other hardened value therefore produces a valid-looking path for a nonstandard key class instead of being rejected. Introduce a `PaymentKeyClass` enum and convert its variants to hardened indices internally. That makes invalid roles unrepresentable and gives callers exhaustive handling if another standardized class is added later.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/crypto/stealth.rs:157-225: Reuse the secp256k1 context during destination scanning
  (existing thread: https://github.com/dashpay/platform/pull/4272#discussion_r3766545373)
  Each payer, receiver, and spender entry point constructs and randomizes a fresh `Secp256k1` context. `recognize_one_time_destination` is intended to run repeatedly while scanning candidate notifications and output counters, so this repeatedly allocates and initializes a context on the hottest path. Accept a caller-owned context, provide a scanner that owns one, or use an available shared context, then pass it into the lower functions that already borrow a context.

In `packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/identity_update/mod.rs`:
- [SUGGESTION] packages/rs-drive-abci/src/execution/validation/state_transition/state_transitions/identity_update/mod.rs:2008-2229: Pin the activation test to protocol version 14
  (existing thread: https://github.com/dashpay/platform/pull/4272#discussion_r3766545380)
  The protocol-14 acceptance tests use `PlatformVersion::latest()`, so they will silently move to later protocol versions and stop protecting the version-14 dispatch table. Select `PlatformVersion::get(14)` explicitly. The protocol-13 test also asserts only that execution was not successful; it should match the expected paid `InvalidIdentityPublicKeySecurityLevelError`, because unrelated signature, nonce, serialization, balance, or internal failures currently satisfy the assertion without proving the activation boundary.
```
</details>

packages/rs-platform-encryption/src/stealth.rs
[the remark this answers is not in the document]
thepastaclaw: The current comment already identifies these as DIP-33 test-vector inputs, and direct comparison confirms that all inputs and expected outputs match the `Test Vectors` section of dashpay/dips#188 at commit `18e0756bfeeab16432ffa4f197941f24c6b482f4`. This is not a locally generated regression vector.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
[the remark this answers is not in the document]
thepastaclaw: Purpose, security level, and key type are immutable for a given identity-key ID, so existing rows intentionally refresh only mutable fields. The merge base could not have persisted payment-purpose rows through the old fallback, while new rows now validate and preserve purposes 7 and 8 without coercion.

PR #4060, opened 2026-07-09
fix(kotlin-sdk): android host-app integration fixes (key security policy, unmanaged-identity reads, typed signing error)

bfoss765:
Kotlin-only fixes for the three SDK issues blocking the Dash Android wallet's Platform migration (dashpay/dash-wallet#1507). No native/Rust changes; the prebuilt library is untouched. Defaults are unchanged everywhere.

## 1. Configurable identity-key security policy + queryable pending-key state (fixes #4053)

Identity private keys were always wrapped under the auth-gated Keystore alias (`KeystoreManager.KEYS_ALIAS`, `setUserAuthenticationRequired(true)`, ~30 s post-unlock validity) with no configurability — a host app that gates wallet access behind its own auth model (the Dash wallet gates by PIN app-side) fails every sign outside the window, and a failed derive/store during persistence was silently swallowed.

- New `KeySecurityPolicy` enum: `AUTH_GATED` (default — exact historical behavior, same alias, same auth window) and `DEVICE_BOUND` (a separate non-auth-gated but still hardware-backed / StrongBox-preferring, non-exportable RSA wrapping alias, `org.dashfoundation.wallet.keys.devicebound`). Distinct aliases because Keystore auth parameters are fixed at key generation; the KDoc documents the pick-once-per-install stability requirement.
- `KeystoreManager(keySecurityPolicy)` selects the alias (`keysAlias`); the RSA helpers are parameterized per alias, and the user-auth gate is only applied when generating the `AUTH_GATED` alias.
- `WalletStorage(context, keySecurityPolicy)` convenience constructor; identity-key encrypt/decrypt routes through `keystore.keysAlias`.
- `PlatformWalletPersistenceHandler` no longer skips a failed identity-key derive/store silently: it logs at ERROR (the key is persisted watch-only and cannot sign) and records a queryable `PendingIdentityKey` on a `StateFlow`, cleared when a later persist round stores the key. Exposed as `PlatformWalletManager.pendingIdentityKeys`, repairable via the existing `repairIdentityKey`.

## 2. `Dashpay` reads return null/empty for unmanaged identities (fixes #4051)

[...]

PR #4273, opened 2026-08-03
feat(platform-wallet): secp256k1 primitives, identity-update parsing, and scoped signing keys for DashConnect

romchornyi:

The iOS wallet is porting DashConnect — the `dash-key:` / `dash-st:` passwordless login flow that already ships on Android. Four gaps in `platform-wallet-ffi` / `swift-sdk` block it:

- **No raw secp256k1 primitives.** The key exchange needs an ephemeral keypair and the *unhashed* ECDH X coordinate. The only ECDH in the FFI is DashPay-specific (`dashpay.rs: ecdh_shared_secret`) — it takes a derivation path and returns a finished secret, so neither the arbitrary ephemeral key nor the raw X is reachable.
- **No way to inspect an app-supplied state transition.** A `dash-st:` QR carries a serialized `IdentityUpdateTransition` built by the dApp. Without deserialization the wallet can only broadcast foreign bytes blind; it needs to verify the transition adds exactly the two login keys it derived for its own identity.
- **No signing with keys that are never persisted.** Those login keys are derived on demand and must not touch the keychain, but adding them requires proof of possession at signing time.
- **The FFI was stricter than consensus.** `decode_contract_bounds` rejected `kind == 0` (no bounds) for ENCRYPTION / DECRYPTION keys, while consensus accepts unbounded keys for every purpose (`validate_identity_public_key_contract_bounds/v1`). Real Yappr testnet `dash-st:` transitions carry an unbounded ENCRYPTION key, so the flow failed on a key Platform considers valid.

### Rust FFI — `packages/rs-platform-wallet-ffi`

- **`src/secp256k1_primitives.rs` (new)** — three handle-free functions over the already-linked `dashcore::secp256k1`, exported from `lib.rs`:
  - `platform_wallet_secp256k1_verify_compressed_point`
  - `platform_wallet_secp256k1_compressed_public_key`
  - `platform_wallet_secp256k1_ecdh_shared_x` — raw affine X via `mul_tweak`, deliberately unhashed
- **`src/identity_update.rs`** — parse-only entry points `platform_wallet_parse_identity_update_transition` / `..._free`, returning `ParsedIdentityUpdateFFI` and `ParsedIdentityUpdatePublicKeyFFI`. Deserialization only: no signing, no broadcast, no state change.
- **`src/identity_registration_with_signer.rs`** — `decode_contract_bounds` no longer requires bounds for ENCRYPTION / DECRYPTION. `kind == 0` now decodes to `None` for every purpose; `kind == 1` / `kind == 2` decode and validate exactly as before. Doc comments corrected to describe the consensus rule rather than the old assumption.

### Swift SDK — `packages/swift-sdk`

- **`Sources/SwiftDashSDK/Security/Secp256k1Primitives.swift` (new)** — `isValidCompressedPoint(_:)`, `compressedPublicKey(privateKey:)`, `ecdhSharedX(privateKey:publicKey:)` over the new FFI.
- **`Sources/SwiftDashSDK/PlatformWallet/ManagedPlatformWallet.swift`** — `parseIdentityUpdateTransition(_:)` wrapper mapping the parsed C structs into Swift values.
- **`Sources/SwiftDashSDK/FFI/KeychainSigner.swift`** — `withAdditionalSigningKeys(_:)`: a scoped in-memory registry of `(publicKey, privateKey)` pairs that the sign trampoline consults *before* the platform-address / breadcrumb / persisted-key paths. Entries are popped and their private keys zeroed on scope exit. Mirrors Android's `signingKeys`.

## How Has This Been Tested?

`cargo test -p platform-wallet-ffi` — 26 lib + 6 integration tests pass, and `cargo fmt` is clean. `cargo clippy -p platform-wallet-ffi --all-targets -- -D warnings` reports nothing in the files this PR touches; its findings are all pre-existing on `v4.2-dev` in `src/persistence.rs` and `src/core_wallet_types.rs`, which are untouched here.

- `src/secp256k1_primitives.rs` — 6 tests: known-vector public key and ECDH X, invalid-point rejection, out-of-range scalar, wrong-length inputs.
- `src/identity_update.rs` — 4 tests: parses a tagged transition, parses a tagless one captured from the live Yappr testnet app, rejects malformed and truncated bytes.
- `src/identity_registration_with_signer.rs` — 4 tests covering the bounds change: an unbounded ENCRYPTION key decodes, `kind == 0` decodes to `None` for AUTHENTICATION / ENCRYPTION / DECRYPTION, and `kind == 1` / `kind == 2` still reject null ids and payloads.
- `SwiftTests/SwiftDashSDKTests/KeychainSignerAdditionalSigningKeysTests.swift` (new) — scope push/pop, lookup by public key, and zeroing on exit.

Manual, on **testnet**, from the iOS wallet against the live Yappr instance: scan the `dash-key:` QR → publish `loginKeyResponse`; scan the app's real `dash-st:` QR → the transition is parsed here, validated against locally derived keys, and its two login keys are added to the identity through `updateIdentity` with the scoped signer. Signing out on the Yappr website and signing back in with the `dash-key:` QR alone was also verified end to end.

## Breaking Changes

None. The new FFI entry points are additive. The only behaviour change is `decode_contract_bounds` accepting `kind == 0` for ENCRYPTION / DECRYPTION, which previously returned `ErrorInvalidParameter` for a key consensus accepts — strictly a widening.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added secp256k1 utilities for public-key derivation, point validation, and ECDH shared-secret computation.
  * Added support for parsing identity update transitions in tagged and tagless formats.
  * Added scoped temporary signing keys with automatic secure cleanup.
* **Improvements**
  * Unbounded encryption and decryption keys are now accepted.
  * Added borrowed private-key signing support.
  * Improved validation and error handling for malformed identity updates and cryptographic inputs.
* **Tests**
  * Expanded coverage for parsing, signing-key cleanup, contract bounds, and cryptographic operations.


thepastaclaw:
✅ Final review complete — no blockers (commit 80960f64)

packages/rs-platform-wallet-ffi/src/secp256k1_primitives.rs
```diff
@@ -0,0 +1,269 @@
...
+    let scalar_bytes = Zeroizing::new(secret_key.0.secret_bytes());
+    let scalar = unwrap_result_or_return!(Scalar::from_be_bytes(*scalar_bytes)
+        .map(WipingScalar)
+        .map_err(|error| invalid_parameter(format!("Invalid secp256k1 scalar: {error}"))));
+
+    let shared_point = unwrap_result_or_return!(public_key
+        .mul_tweak(&secp, &scalar.0)
+        .map_err(|_| wallet_operation("ECDH produced the point at infinity")));
+    let uncompressed = shared_point.serialize_uncompressed();
+    std::ptr::copy_nonoverlapping(uncompressed[1..33].as_ptr(), out_shared_x, 32);
```
thepastaclaw: **🔴 Blocking: ECDH uses variable-time public-tweak multiplication with a secret scalar**

`PublicKey::mul_tweak` calls libsecp256k1's `secp256k1_ec_pubkey_tweak_mul`. That path reaches `eckey_pubkey_tweak_mul` and the ordinary WNAF-based `ecmult`, whose branches and table selection depend on the scalar; it is intended for public tweaks. Here the tweak is the private ECDH scalar and the peer controls the input point, so the new DashConnect primitive exposes secret-dependent timing/cache behavior. Use `dashcore::secp256k1::ecdh::shared_secret_point(&public_key, &secret_key.0)`, which invokes `secp256k1_ecdh` and `ecmult_const`; zeroize the returned 64-byte `x || y` buffer after copying its first 32 bytes to `out_shared_x`.

<sub>source: ['codex']</sub>

packages/swift-sdk/Sources/SwiftDashSDK/FFI/KeychainSigner.swift
```diff
@@ -84,6 +84,52 @@ import SwiftData
...
+        }
+
+        func sign(data: Data, network: Network) -> Result<Data, KeychainSigner.Error> {
+            do {
+                return .success(
+                    try RawKeySigner.sign(
+                        data: data,
+                        privateKey: Data(privateKeyBytes),
+                        network: network
+                    )
```
thepastaclaw: **🟡 Suggestion: Scoped private keys leave an unzeroized Data copy after signing**

`Data(privateKeyBytes)` creates a separate immutable buffer containing the scoped private scalar. `RawKeySigner.sign` zeroes only the later mutable `[UInt8]` copy it creates, while `popAdditionalSigningKeys` zeroes only the entry's `privateKeyBytes`; neither can scrub this temporary `Data` before Foundation releases it. This avoidable internal copy can therefore remain in allocator memory after the scope exits, weakening the cleanup guarantee for login keys that are intentionally never persisted. Add a signing overload that borrows the entry's mutable byte buffer directly, or explicitly place every temporary key-bearing buffer in wipe-on-drop storage.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/identity_update.rs
```diff
@@ -102,3 +373,203 @@ pub unsafe extern "C" fn platform_wallet_update_identity_with_signer(
...
+
+        unsafe { platform_wallet_parse_identity_update_transition_free(&mut out) };
+        assert!(out.add_public_keys.is_null());
+        assert!(out.disable_public_key_ids.is_null());
+    }
+
+    #[test]
+    fn parses_yappr_tagless_identity_update_transition() {
+        let tagged = fixture_transition_bytes();
+        let tagless = tagged[1..].to_vec();
```
thepastaclaw, line 561: **🟡 Suggestion: The Yappr framing test does not exercise a captured Yappr payload**

The PR states that this test parses a tagless transition captured from the live Yappr testnet app, but the test serializes the local Rust fixture and removes byte zero. That validates the parser against its own framing assumption and cannot detect an interoperability mismatch with Yappr's actual encoder. Check in the captured tagless payload as fixed bytes and assert the projected identity, key fields, unbounded ENCRYPTION key, contract bounds, and disabled IDs.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The identity-update parser, FFI ownership paths, secp256k1 vectors, and contract-bounds widening are otherwise consistent with the stated DashConnect behavior, but the raw ECDH implementation uses a variable-time multiplication API with a secret scalar and must be replaced before merge. The scoped signer also leaves an avoidable private-key copy unwiped, the claimed captured Yappr fixture is not present, and the Swift contract-bounds documentation contradicts the new consensus-compatible behavior.

Source: reviewers codex/general=gpt-5.6-sol(completed), codex/security-auditor=gpt-5.6-sol(completed), codex/ffi-engineer=gpt-5.6-sol(completed); final verifier codex/verifier=gpt-5.6-sol(completed); sonnet=not run (deferred by blocker gate); coordinator=openclaw-agent/cliproxy/gpt-5.6-sol (orchestration-only).

### Review provenance

<details>

```

In `packages/rs-platform-wallet-ffi/src/secp256k1_primitives.rs`:
- [BLOCKING] packages/rs-platform-wallet-ffi/src/secp256k1_primitives.rs:132-136: ECDH uses variable-time public-tweak multiplication with a secret scalar
  `PublicKey::mul_tweak` calls libsecp256k1's `secp256k1_ec_pubkey_tweak_mul`. That path reaches `eckey_pubkey_tweak_mul` and the ordinary WNAF-based `ecmult`, whose branches and table selection depend on the scalar; it is intended for public tweaks. Here the tweak is the private ECDH scalar and the peer controls the input point, so the new DashConnect primitive exposes secret-dependent timing/cache behavior. Use `dashcore::secp256k1::ecdh::shared_secret_point(&public_key, &secret_key.0)`, which invokes `secp256k1_ecdh` and `ecmult_const`; zeroize the returned 64-byte `x || y` buffer after copying its first 32 bytes to `out_shared_x`.

In `packages/swift-sdk/Sources/SwiftDashSDK/FFI/KeychainSigner.swift`:
- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/FFI/KeychainSigner.swift:103-107: Scoped private keys leave an unzeroized Data copy after signing
  `Data(privateKeyBytes)` creates a separate immutable buffer containing the scoped private scalar. `RawKeySigner.sign` zeroes only the later mutable `[UInt8]` copy it creates, while `popAdditionalSigningKeys` zeroes only the entry's `privateKeyBytes`; neither can scrub this temporary `Data` before Foundation releases it. This avoidable internal copy can therefore remain in allocator memory after the scope exits, weakening the cleanup guarantee for login keys that are intentionally never persisted. Add a signing overload that borrows the entry's mutable byte buffer directly, or explicitly place every temporary key-bearing buffer in wipe-on-drop storage.

In `packages/rs-platform-wallet-ffi/src/identity_update.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/identity_update.rs:471-473: The Yappr framing test does not exercise a captured Yappr payload
  The PR states that this test parses a tagless transition captured from the live Yappr testnet app, but the test serializes the local Rust fixture and removes byte zero. That validates the parser against its own framing assumption and cannot detect an interoperability mismatch with Yappr's actual encoder. Check in the captured tagless payload as fixed bytes and assert the projected identity, key fields, unbounded ENCRYPTION key, contract bounds, and disabled IDs.

In `packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/ManagedPlatformWallet.swift`:
- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/ManagedPlatformWallet.swift:176-181: Swift still documents encryption bounds as mandatory
  This public API still says Encryption and Decryption keys require contract bounds and that `nil` is only valid for other purposes. The Rust decoder was intentionally widened because consensus accepts unbounded keys for every purpose, including the real DashConnect ENCRYPTION key described by this PR. The contradictory documentation can cause SDK consumers to reject or rewrite a valid transition before calling the now-correct FFI.
```
</details>

packages/rs-platform-wallet-ffi/src/secp256k1_primitives.rs
romchornyi: Addressed in 7d998b8d16.

`platform_wallet_secp256k1_ecdh_shared_x` now goes through `dashcore::secp256k1::ecdh::shared_secret_point(&public_key, &secret_key.0)` — i.e. `secp256k1_ecdh` / `ecmult_const` — and holds the returned 64-byte `x || y` in `Zeroizing`, copying only the first 32 bytes into `out_shared_x`.

Agreed on the substance: the tweak here is the wallet's own ephemeral scalar and the point arrives from the peer, so `mul_tweak`'s WNAF `ecmult` gave the party choosing the input a secret-dependent timing signal. `mul_tweak`, the `Scalar` / `WipingScalar` conversion it required, and the now-unreachable point-at-infinity error path are all gone.

`ecdh_shared_x_matches_known_vector` is deliberately unchanged and still passes — that is the equivalence proof for the swap.

packages/swift-sdk/Sources/SwiftDashSDK/FFI/KeychainSigner.swift
romchornyi: Addressed in 7d998b8d16.

`RawKeySigner` gained a `sign(data:privateKeyBuffer:network:)` overload taking an `UnsafeBufferPointer<UInt8>`, and `AdditionalSigningKeyEntry.sign` calls it inside `privateKeyBytes.withUnsafeBufferPointer`, so no `Data` copy of the scalar is created on that path at all.

The `Data`-taking overload is kept as a thin wrapper over the same body, so no existing call site changed. The defensive `keyCopy` + `memset_s` inside stays as well: the FFI signer still never reads directly from caller-owned storage, and that copy is the one that was already being wiped.

packages/rs-platform-wallet-ffi/src/identity_update.rs
romchornyi: Addressed in 7d998b8d16.

`parses_yappr_tagless_identity_update_transition` now decodes the payload captured from the live testnet app (`YAPPR_TAGLESS_IDENTITY_UPDATE_FIXTURE_HEX` — the same bytes the iOS side asserts in `PlatformDashConnectDataSourceTests`) and checks identity id `89fd6ddb…`, no disabled ids, and both keys' id / purpose / security level / key type / data: key 6 AUTHENTICATION · HIGH · ECDSA_HASH160, key 7 ENCRYPTION · MEDIUM · ECDSA_SECP256K1, with `contract_bounds_kind == 0` on both. That unbounded ENCRYPTION key is exactly the regression guard for the `decode_contract_bounds` widening in this PR, and there is a comment saying so. Enum variants are compared via `try_from`, not raw discriminants.

The self-serialized case is kept — it still proves the tagged and prefixed paths agree — but renamed to `parses_tagless_framing_by_prepending_the_variant_tag`, which is what it actually demonstrates. The inaccurate part was the PR description; the test now lives up to it.

romchornyi:
All four findings from the preliminary review are addressed in 7d998b8d16; the three inline threads have per-finding replies. The fourth was reported outside the diff, so it is answered here:

**Swift still documents encryption bounds as mandatory** — fixed. `ManagedPlatformWallet.IdentityPublicKeyInfo.contractBounds` no longer claims bounds are required for Encryption / Decryption; it now states that bounds are optional for every purpose and are only checked by consensus when present. The rest of the file was checked for the same stale claim — the only other mentions are length validations for `ContractBounds` ids, which are unrelated.

### Verification

- `cargo fmt -p platform-wallet-ffi -- --check` — clean
- `cargo test -p platform-wallet-ffi` — 228 lib + 26 + 6 integration, all green, including the unchanged `ecdh_shared_x_matches_known_vector`
- `swift build --build-tests` + `swift test --filter KeychainSigner` — 4/4
- `xcodebuild build -scheme SwiftExampleApp` — succeeds (it compiles these sources without region-based isolation, which SwiftPM alone does not catch)

CI on this commit is green, Swift SDK build and Rust wallet tests included.

packages/swift-sdk/Sources/SwiftDashSDK/Security/Secp256k1Primitives.swift
```diff
@@ -0,0 +1,66 @@
...
+                        UInt(publicBytes.count),
+                        outputBuffer.baseAddress
+                    )
+                }
+            }
+        }
+
+        try result.check()
+        return Data(output)
+    }
```
thepastaclaw, line 68: **🟡 Suggestion: ECDH wrapper leaves an extra unwiped shared-secret copy**

`platform_wallet_secp256k1_ecdh_shared_x` writes the raw ECDH X coordinate into the mutable `[UInt8]` buffer, and `Data(output)` then copies that session secret into a second allocation for the return value. The original array is released without being scrubbed, leaving an avoidable secret-bearing allocation in allocator memory. Allocate the returned `Data` directly and let Rust write into that buffer so only the intentionally returned copy remains.

```suggestion
    public static func ecdhSharedX(privateKey: Data, publicKey: Data) throws -> Data {
        var output = Data(count: 32)

        let result = output.withUnsafeMutableBytes { outputBuffer -> PlatformWalletFFIResult in
            let outputBytes = outputBuffer.bindMemory(to: UInt8.self)
            return privateKey.withUnsafeBytes { privateBuffer -> PlatformWalletFFIResult in
                let privateBytes = privateBuffer.bindMemory(to: UInt8.self)
                return publicKey.withUnsafeBytes { publicBuffer in
                    let publicBytes = publicBuffer.bindMemory(to: UInt8.self)
                    return platform_wallet_secp256k1_ecdh_shared_x(
                        privateBytes.baseAddress,
                        UInt(privateBytes.count),
                        publicBytes.baseAddress,
                        UInt(publicBytes.count),
                        outputBytes.baseAddress
                    )
                }
            }
        }

        try result.check()
        return output
    }
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/identity_update.rs
```diff
@@ -102,3 +373,287 @@ pub unsafe extern "C" fn platform_wallet_update_identity_with_signer(
...
+        // The real DashConnect ENCRYPTION key is intentionally unbounded, so
+        // this guards the decoder widening that now accepts absent bounds.
+        assert_eq!(keys[1].contract_bounds_kind, 0);
+        let key1_data = unsafe { slice::from_raw_parts(keys[1].data_ptr, keys[1].data_len) };
+        assert_eq!(
+            key1_data,
+            hex::decode("035e8cfb0785b54e8902a3dc17bdaad8a5738c6019a18ebc527f79d1c64a27826a")
+                .expect("valid encryption key hex")
+                .as_slice()
+        );
```
thepastaclaw, line 553: **🟡 Suggestion: Captured Yappr fixture does not assert the read-only field**

The captured fixture verifies each key’s ID, purpose, security level, type, data, and bounds, but never checks `read_only`. That field is projected at `project_parsed_identity_update` and then mapped into Swift’s `IdentityPubkey`; hardcoding or dropping it would leave this interoperability test green. The captured keys both decode as `read_only == false`; assert those values, and extend the synthetic tagged-fixture assertions to verify that its second key remains `read_only == true`.

<sub>source: ['codex']</sub>

thepastaclaw (commented):

All four carried-forward prior findings are fixed at the current head, so none remain live. The final combined review confirms one new latest-delta test-coverage finding (the captured Yappr fixture omits `read_only` assertions) and two cumulative full-PR security suggestions concerning avoidable secret-bearing allocations; no blocking issue remains.

Source: reviewers codex/general=gpt-5.6-sol(completed), codex/security-auditor=gpt-5.6-sol(completed), codex/ffi-engineer=gpt-5.6-sol(completed), claude/general=claude-sonnet-5(completed), claude/security-auditor=claude-sonnet-5(completed), claude/ffi-engineer=claude-sonnet-5(completed); final verifier codex/final-verifier=gpt-5.6-sol(completed, fallback_for_sonnet_verifier=true); coordinator=openclaw-agent/cliproxy/gpt-5.6-sol (orchestration-only).

### Review provenance

🟡 3 suggestion(s)

<details>

```

In `packages/swift-sdk/Sources/SwiftDashSDK/Security/Secp256k1Primitives.swift`:
- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/Security/Secp256k1Primitives.swift:44-65: ECDH wrapper leaves an extra unwiped shared-secret copy
  `platform_wallet_secp256k1_ecdh_shared_x` writes the raw ECDH X coordinate into the mutable `[UInt8]` buffer, and `Data(output)` then copies that session secret into a second allocation for the return value. The original array is released without being scrubbed, leaving an avoidable secret-bearing allocation in allocator memory. Allocate the returned `Data` directly and let Rust write into that buffer so only the intentionally returned copy remains.

In `packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/KeyManager.swift`:
- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/KeyManager.swift:81-98: Scoped signing still leaves Rust-side private-key copies unwiped
  The new scoped-key path correctly borrows the registry array and wipes Swift’s `keyCopy`, but `dash_sdk_signer_create_from_private_key` then constructs a Rust `SingleKeySigner`. `SingleKeySigner::new_from_slice` copies the scalar into a plain stack `[u8; 32]` and into a `dashcore::PrivateKey`; destroying the handle only drops the `Arc`, while `SingleKeySigner` has no `Drop` implementation and the copyable `SecretKey` is not automatically erased. Signing also calls `secret_bytes()`, creating another ordinary 32-byte array before `dashcore::sign_hash` parses another non-erasing `SecretKey`. Because these DashConnect login keys are deliberately non-persisted and promised to be cleared at scope exit, use zeroizing construction/signing storage and erase the signer-owned key on drop, or replace the temporary-handle path with a one-shot Rust export whose secret values are guarded end-to-end.

In `packages/rs-platform-wallet-ffi/src/identity_update.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/identity_update.rs:503-549: Captured Yappr fixture does not assert the read-only field
  The captured fixture verifies each key’s ID, purpose, security level, type, data, and bounds, but never checks `read_only`. That field is projected at `project_parsed_identity_update` and then mapped into Swift’s `IdentityPubkey`; hardcoding or dropping it would leave this interoperability test green. The captured keys both decode as `read_only == false`; assert those values, and extend the synthetic tagged-fixture assertions to verify that its second key remains `read_only == true`.
```
</details>

packages/rs-platform-wallet-ffi/src/identity_update.rs
thepastaclaw: Resolved in `7d998b8` — _The Yappr framing test does not exercise a captured Yappr payload_ no longer present.

packages/swift-sdk/Sources/SwiftDashSDK/FFI/KeychainSigner.swift
thepastaclaw: Resolved in `7d998b8` — _Scoped private keys leave an unzeroized Data copy after signing_ no longer present.

packages/rs-platform-wallet-ffi/src/secp256k1_primitives.rs
thepastaclaw: Resolved in `7d998b8` — _ECDH uses variable-time public-tweak multiplication with a secret scalar_ no longer present.

packages/simple-signer/src/single_key_signer.rs
```diff
@@ -34,9 +35,9 @@ impl SingleKeySigner {
         if private_key_data.len() != 32 {
             return Err("Private key must be 32 bytes".to_string());
         }
-        let mut arr = [0u8; 32];
+        let mut arr = Zeroizing::new([0u8; 32]);
         arr.copy_from_slice(private_key_data);
-        let private_key = PrivateKey::from_byte_array(&arr, network)
+        let private_key = PrivateKey::from_byte_array(&*arr, network)
```
thepastaclaw: **🔴 Blocking: Latest zeroization edits fail the required Clippy gate**

With the repository-pinned Rust 1.92 toolchain, `cargo clippy --package simple-signer --all-features --locked -- --no-deps -D warnings` fails on `clippy::explicit_auto_deref`, using the same lint flags as the workspace CI job. This line passes `&*arr`, and the same latest-delta pattern appears at lines 153 and 170 as `&*secret_bytes`; all three expressions were introduced by commit `473b54c205`. Because warnings are denied, the required Clippy gate cannot pass until these use ordinary auto-deref (`&arr` and `&secret_bytes`).

```suggestion
let private_key = PrivateKey::from_byte_array(&arr, network)
```

<sub>source: ['codex']</sub>

packages/simple-signer/src/single_key_signer.rs
```diff
@@ -95,7 +105,8 @@ impl Signer<IdentityPublicKey> for SingleKeySigner {
             KeyType::ECDSA_SECP256K1 | KeyType::ECDSA_HASH160 => {
                 // Do not log private key material. Log data fingerprint only.
                 debug!(data_hex = %hex::encode(data), "SingleKeySigner: signing data");
-                let signature = signer::sign(data, &self.private_key.inner.secret_bytes())?;
+                let secret_bytes = Zeroizing::new(self.private_key.inner.secret_bytes());
+                let signature = signer::sign(data, &secret_bytes[..])?;
```
thepastaclaw, line 118: **🟡 Suggestion: Scoped signing still leaves Rust-side private-key copies unwiped**

The latest delta now zeroizes explicit byte arrays and erases the signer-owned key in `Drop`, but this path still hands the scalar bytes to the pinned `dashcore::signer::sign`. In rust-dashcore revision `70d4bf8e`, `sign_hash` creates `let pk = SecretKey::from_slice(private_key)` and returns without calling `non_secure_erase()`; `SecretKey` is `Copy` and has no erasing `Drop`. The `can_sign_with` branches at lines 151-176 likewise parse local `SecretKey` values that are never erased. Destroying the scoped signer therefore does not explicitly clear all avoidable Rust-side scalar values. Sign with `&self.private_key.inner` through a local helper and derive the public key directly from that stored key, or place every parsed local `SecretKey` behind an erasing guard.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

Carried-forward prior findings: one security suggestion remains because scoped signing still creates unmanaged Rust `SecretKey` temporaries; the other two prior suggestions are fixed. New latest-delta findings: one blocking issue was confirmed because the repository-pinned Rust 1.92 Clippy gate fails on three newly introduced explicit auto-dereferences; no other in-scope issue was found.

Source: reviewers codex/general=gpt-5.6-sol(completed), codex/security-auditor=gpt-5.6-sol(completed), codex/ffi-engineer=gpt-5.6-sol(completed); final verifier codex/verifier=gpt-5.6-sol(completed); sonnet=not run (deferred by blocker gate); coordinator=openclaw-agent/cliproxy/gpt-5.6-sol (orchestration-only).

### Review provenance

<details>

```

In `packages/simple-signer/src/single_key_signer.rs`:
- [BLOCKING] packages/simple-signer/src/single_key_signer.rs:40: Latest zeroization edits fail the required Clippy gate
  With the repository-pinned Rust 1.92 toolchain, `cargo clippy --package simple-signer --all-features --locked -- --no-deps -D warnings` fails on `clippy::explicit_auto_deref`, using the same lint flags as the workspace CI job. This line passes `&*arr`, and the same latest-delta pattern appears at lines 153 and 170 as `&*secret_bytes`; all three expressions were introduced by commit `473b54c205`. Because warnings are denied, the required Clippy gate cannot pass until these use ordinary auto-deref (`&arr` and `&secret_bytes`).
- [SUGGESTION] packages/simple-signer/src/single_key_signer.rs:108-109: Scoped signing still leaves Rust-side private-key copies unwiped
  The latest delta now zeroizes explicit byte arrays and erases the signer-owned key in `Drop`, but this path still hands the scalar bytes to the pinned `dashcore::signer::sign`. In rust-dashcore revision `70d4bf8e`, `sign_hash` creates `let pk = SecretKey::from_slice(private_key)` and returns without calling `non_secure_erase()`; `SecretKey` is `Copy` and has no erasing `Drop`. The `can_sign_with` branches at lines 151-176 likewise parse local `SecretKey` values that are never erased. Destroying the scoped signer therefore does not explicitly clear all avoidable Rust-side scalar values. Sign with `&self.private_key.inner` through a local helper and derive the public key directly from that stored key, or place every parsed local `SecretKey` behind an erasing guard.
```
</details>

packages/swift-sdk/Sources/SwiftDashSDK/Security/Secp256k1Primitives.swift
thepastaclaw: Resolved in `473b54c` — _ECDH wrapper leaves an extra unwiped shared-secret copy_ no longer present.

packages/rs-platform-wallet-ffi/src/identity_update.rs
thepastaclaw: Resolved in `473b54c` — _Captured Yappr fixture does not assert the read-only field_ no longer present.

packages/simple-signer/src/single_key_signer.rs
romchornyi: Addressed in fdbe1b9630.

All three sites are plain auto-deref now — `&arr` at the `PrivateKey::from_byte_array` call, and `&secret_bytes` in both `can_sign_with` branches.

Confirmed against the same invocation the workspace job uses, on the pinned 1.92 toolchain:

```
cargo clippy -p simple-signer --all-features --locked -- --no-deps -D warnings
```

which now finishes clean. `Rust workspace tests / Tests (macOS)` is green on that commit.

Fair catch on process, too: the round-3 spec asked for `cargo fmt` and the test suites but not for Clippy, so `473b54c205` was never run through the gate it had to pass.

packages/simple-signer/src/single_key_signer.rs
romchornyi: Confirmed, and deliberately **not** fixed here — with the doc comment corrected so it no longer implies otherwise (fdbe1b9630).

You are right about the pinned revision: at `70d4bf8e`, `signer::sign` builds its own `SecretKey` from the bytes it is handed and returns without erasing it, and the `can_sign_with` branches parse local `SecretKey` values the same way. `SecretKey` is `Copy` with no erasing `Drop`, so destroying the scoped signer cannot reach those copies. Nothing inside `simple-signer` can close that gap; it needs zeroizing parse/sign storage in `rust-dashcore` itself, followed by a pin bump here.

What changed instead is the honesty of the claim. `Drop` previously read as though the scalar was gone after the handle died. It now states that it clears only the copies this type owns, names the two `dashcore` paths it cannot reach, and carries a `TODO(dashconnect-key-hygiene)` for the upstream change — rather than leaving a comment that promises more than the code delivers.

Happy to open the `rust-dashcore` issue for the upstream half if that is the preferred route; it is out of scope for this PR.

romchornyi:
Both round-3 findings are answered in their threads; fdbe1b9630 also merges `v4.2-dev` and resolves the conflict this PR had picked up.

**Merge conflict — `keychainSignerSignAsyncTrampoline`.** The base moved its identity-key fallback (`lookupIdentityPrivateKey` + `ffiSign`) into the trampoline, while this branch had already moved that same chain inside `signOnDemand`, which returns a non-optional `Result`. Kept this branch's shape and carried the base's behaviour across: the failure arm maps through `keychainSignerCompletionErrorCode`, so hosts still get the typed `signingKeyUnavailable` rather than sniffing the message (#4060 finding 7). Taking "our side" wholesale would have dropped that code.

### Verification on fdbe1b9630

- `cargo clippy -p simple-signer --all-features --locked -- --no-deps -D warnings` — clean
- `cargo fmt --check` — clean
- `cargo test -p platform-wallet-ffi -p simple-signer` — 239 + 26 + 6 + 5, all green
- `swift test --filter KeychainSigner` — 5/5
- `xcodebuild build -scheme SwiftExampleApp` — succeeds

CI: `Rust workspace tests / Tests (macOS)` passes; the Swift job is still running at the time of writing. `codecov/project` is red on coverage delta only and does not gate the merge.

packages/simple-signer/src/single_key_signer.rs
```diff
@@ -137,12 +148,12 @@ impl Signer<IdentityPublicKey> for SingleKeySigner {
...
-                ) {
-                    Ok(sk) => sk,
-                    Err(_) => return false,
-                };
+                let secret_bytes = Zeroizing::new(self.private_key.inner.secret_bytes());
+                let secret_key =
+                    match dashcore::secp256k1::SecretKey::from_byte_array(&*secret_bytes) {
+                        Ok(sk) => sk,
+                        Err(_) => return false,
+                    };
```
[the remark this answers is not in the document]
romchornyi: Addressed in 51994290b3.

Both `can_sign_with` branches now derive the public key directly from the signer-owned `self.private_key.inner`. They no longer call `secret_bytes()` or parse a second, copyable `SecretKey`, so the avoidable local scalar temporaries identified here do not exist and there is no post-comparison cleanup path to miss.

The `Drop` documentation was narrowed accordingly: the signer-owned key is still erased there, while the one remaining limitation is explicitly the separate `SecretKey` constructed inside the pinned `dashcore::signer::sign`, tracked by the existing upstream TODO.

Regression coverage now exercises matching and mismatching keys for both `ECDSA_SECP256K1` and `ECDSA_HASH160`.

Validation on the merged head 154b382180:

- `cargo fmt --all -- --check` — clean
- `cargo test -p simple-signer --all-features --locked` — 6/6
- `cargo clippy -p simple-signer --all-features --locked -- --no-deps -D warnings` — clean for `simple-signer`

packages/rs-platform-wallet-ffi/src/secp256k1_primitives.rs
```diff
@@ -0,0 +1,250 @@
...
+    if private_key_len != 32 {
+        return Err(invalid_parameter(format!(
+            "secp256k1 private keys must be exactly 32 bytes, got {private_key_len}"
+        )));
+    }
+
+    let bytes = unsafe { slice::from_raw_parts(private_key, private_key_len) };
+    SecretKey::from_slice(bytes)
+        .map(WipingSecretKey)
+        .map_err(|error| invalid_parameter(format!("Invalid secp256k1 private key: {error}")))
```
thepastaclaw: **🟡 Suggestion: Private-key parsing leaves an unwiped Rust stack copy**

Both new DashConnect private-key primitives call this parser. Although `WipingSecretKey` erases the returned key, secp256k1 0.30.0 implements `SecretKey::from_slice` by converting the slice into a local `[u8; 32]` and then copying that array into the returned `SecretKey`; the intermediate array is not guarded or erased. Public-key derivation and ECDH therefore create an avoidable Rust-owned scalar copy. Stage the FFI input in `Zeroizing<[u8; 32]>` and call `from_byte_array` directly so the explicit staging buffer and returned key are both scrubbed.

```suggestion
    let mut bytes = Zeroizing::new([0u8; 32]);
    bytes.copy_from_slice(unsafe { slice::from_raw_parts(private_key, private_key_len) });
    SecretKey::from_byte_array(&bytes)
        .map(WipingSecretKey)
        .map_err(|error| invalid_parameter(format!("Invalid secp256k1 private key: {error}")))
```

<sub>source: ['codex']</sub>

thepastaclaw (commented):

Carried-forward prior findings: the required Clippy blocker is fixed, while the scoped signing path still has one in-scope security suggestion because pinned rust-dashcore creates an unmanaged private-scalar copy. New latest-delta findings: none; cumulative review found one additional suggestion in the PR's new secp256k1 parser, where `SecretKey::from_slice` stages the scalar in an unwiped stack array. Focused secp256k1 and simple-signer tests pass, and the required simple-signer Clippy invocation completes successfully.

Source: reviewers codex/general=gpt-5.6-sol(completed), codex/security-auditor=gpt-5.6-sol(completed), codex/ffi-engineer=gpt-5.6-sol(completed), claude/general=claude-sonnet-5(completed), claude/security-auditor=claude-sonnet-5(completed), claude/ffi-engineer=claude-sonnet-5(completed); final verifier codex/final-verifier=gpt-5.6-sol(completed, fallback_for_sonnet_verifier=true); coordinator=openclaw-agent/cliproxy/gpt-5.6-sol (orchestration-only).

### Review provenance

🟡 1 suggestion(s)

<details>

```

In `packages/rs-platform-wallet-ffi/src/secp256k1_primitives.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/secp256k1_primitives.rs:37-40: Private-key parsing leaves an unwiped Rust stack copy
  Both new DashConnect private-key primitives call this parser. Although `WipingSecretKey` erases the returned key, secp256k1 0.30.0 implements `SecretKey::from_slice` by converting the slice into a local `[u8; 32]` and then copying that array into the returned `SecretKey`; the intermediate array is not guarded or erased. Public-key derivation and ECDH therefore create an avoidable Rust-owned scalar copy. Stage the FFI input in `Zeroizing<[u8; 32]>` and call `from_byte_array` directly so the explicit staging buffer and returned key are both scrubbed.
```
</details>

packages/simple-signer/src/single_key_signer.rs
thepastaclaw: Resolved in `fdbe1b9` — _Scoped signing still leaves Rust-side private-key copies unwiped_ no longer present.

packages/simple-signer/src/single_key_signer.rs
thepastaclaw: Resolved in `fdbe1b9` — _Latest zeroization edits fail the required Clippy gate_ no longer present.

packages/simple-signer/src/single_key_signer.rs
```diff
@@ -95,7 +114,8 @@ impl Signer<IdentityPublicKey> for SingleKeySigner {
             KeyType::ECDSA_SECP256K1 | KeyType::ECDSA_HASH160 => {
                 // Do not log private key material. Log data fingerprint only.
                 debug!(data_hex = %hex::encode(data), "SingleKeySigner: signing data");
-                let signature = signer::sign(data, &self.private_key.inner.secret_bytes())?;
+                let secret_bytes = Zeroizing::new(self.private_key.inner.secret_bytes());
+                let signature = signer::sign(data, &secret_bytes[..])?;
                 Ok(signature.to_vec().into())
```
thepastaclaw, line 119: **🟡 Suggestion: Scoped signing still creates an unmanaged SecretKey copy inside pinned dashcore**

`secret_bytes` is correctly wrapped in `Zeroizing`, and the signer's own key is erased in `Drop`, but `signer::sign(data, &secret_bytes[..])` hands the scalar to the pinned `dashcore::signer::sign`, which parses its own local `SecretKey` and returns without erasing it. `SecretKey` is `Copy` with no erasing `Drop`, so this scoped-signer path — used for DashConnect login keys that are intentionally never persisted — can't clear every Rust-side copy of the scalar. This is already acknowledged in the type's own `Drop` doc comment with a `TODO(dashconnect-key-hygiene)` noting it needs an upstream `rust-dashcore` fix, which is a legitimate reason it isn't resolved in this PR; flagging so it stays tracked, not blocking.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/secp256k1_primitives.rs
```diff
@@ -0,0 +1,250 @@
...
+    public_key: *const u8,
+    public_key_len: usize,
+) -> Result<PublicKey, PlatformWalletFFIResult> {
+    if public_key_len != 33 {
+        return Err(invalid_parameter(format!(
+            "Compressed secp256k1 public keys must be exactly 33 bytes, got {public_key_len}"
+        )));
+    }
+
+    let bytes = unsafe { slice::from_raw_parts(public_key, public_key_len) };
```
thepastaclaw, line 59: **💬 Nitpick: Raw-pointer dereferences hide inside otherwise-safe helper functions**

`parse_secret_key` and `parse_compressed_public_key` are ordinary safe Rust functions that dereference raw pointers internally via `slice::from_raw_parts`. Their signatures don't express the pointer-validity/lifetime contract they actually depend on. The current `extern "C"` callers all run `check_ptr!` first, so this isn't exploitable today, but any future safe caller added inside this module could trigger UB by passing an invalid pointer. Moving slice construction into the `unsafe extern "C"` entry points and having these helpers take validated slices/array refs would make the safety contract explicit at the type level.

<sub>source: ['codex']</sub>

thepastaclaw (commented):

This delta (154b3821..2449ed1f) is a merge of ~50 already-merged v4.2-dev commits into the DashConnect feature branch; none of the PR's own files (secp256k1_primitives.rs, identity_update.rs, identity_registration_with_signer.rs, single_key_signer.rs, KeychainSigner.swift, Secp256k1Primitives.swift) changed a single byte in this specific slice — confirmed with byte-identical `git diff`. Evaluating the PR's real cumulative diff against its true merge-base (bce107b61a..2449ed1f), the DashConnect surface is sound: the ECDH primitive correctly uses the constant-time `shared_secret_point` (not `mul_tweak`), the CodeRabbit contract-bounds downgrade bug is fixed, and the Yappr framing/contract-bounds widening behave as documented. Three low-severity, previously-raised suggestions remain genuinely present in the current code: two avoidable unmanaged secret-scalar copies (one in the new secp256k1 FFI parser, one already-acknowledged-in-code-comments limitation of the pinned `dashcore::signer::sign`), and one minor architectural note about raw-pointer dereferences hiding inside otherwise-safe helper functions. No blocking issues found.
Source: reviewers Codex general/security-auditor/rust-quality/ffi-engineer = gpt-5.6-sol, Claude general/security-auditor/rust-quality/ffi-engineer = claude-sonnet-5; final verifier = claude-sonnet-5. Orchestration-only: openclaw-agent/cliproxy/gpt-5.6-sol.

### Review provenance
- Sonnet reviewers: `claude-sonnet-5` — general (completed), `claude-sonnet-5` — security-auditor (completed), `claude-sonnet-5` — rust-quality (completed), `claude-sonnet-5` — ffi-engineer (completed)

<details>

```

In `packages/simple-signer/src/single_key_signer.rs`:
- [SUGGESTION] packages/simple-signer/src/single_key_signer.rs:117-119: Scoped signing still creates an unmanaged SecretKey copy inside pinned dashcore
  `secret_bytes` is correctly wrapped in `Zeroizing`, and the signer's own key is erased in `Drop`, but `signer::sign(data, &secret_bytes[..])` hands the scalar to the pinned `dashcore::signer::sign`, which parses its own local `SecretKey` and returns without erasing it. `SecretKey` is `Copy` with no erasing `Drop`, so this scoped-signer path — used for DashConnect login keys that are intentionally never persisted — can't clear every Rust-side copy of the scalar. This is already acknowledged in the type's own `Drop` doc comment with a `TODO(dashconnect-key-hygiene)` noting it needs an upstream `rust-dashcore` fix, which is a legitimate reason it isn't resolved in this PR; flagging so it stays tracked, not blocking.
```
</details>

packages/rs-platform-wallet-ffi/src/secp256k1_primitives.rs
thepastaclaw: Resolved in this update — _Private-key parsing leaves an unwiped Rust stack copy_ no longer present.

romchornyi: Addressed in 389dc4ce29.

`parse_secret_key` now stages the caller's bytes in `Zeroizing<[u8; 32]>` and calls `SecretKey::from_byte_array(&staged)`, so the unguarded local array that `from_slice` built on its way to the key is gone. Both DashConnect primitives go through this parser, so it covers public-key derivation and ECDH alike.

The length check still runs before the slice is formed, and the error messages are unchanged. `compressed_public_key_matches_known_vector` and `ecdh_shared_x_matches_known_vector` pass untouched — the parsed key is byte-identical, only its staging changed.

packages/rs-platform-wallet-ffi/src/secp256k1_primitives.rs
romchornyi: Agreed on the reasoning, and deliberately left as is for now.

Moving `slice::from_raw_parts` into the `extern "C"` entry points would make the contract explicit, but it also moves the slice construction *ahead* of the length check that currently guards it: today `parse_secret_key` returns `ErrorInvalidParameter` for a wrong `len` before any slice exists, so a bad length can never reach `from_raw_parts`. Reproducing that ordering at three call sites means duplicating the check and its message, and getting it subtly wrong there turns a clean error into UB — a worse trade than the current shape, which as you note is not reachable today because every caller runs `check_ptr!` first.

Worth doing as a focused refactor of this module rather than inside a PR whose subject is the DashConnect primitives; happy to open a follow-up issue for it.

packages/rs-platform-wallet-ffi/src/identity_update.rs
```diff
@@ -102,3 +373,291 @@ pub unsafe extern "C" fn platform_wallet_update_identity_with_signer(
...
+
+        // `ParsedIdentityUpdateFFI` is a raw-pointer C struct with no `Debug`,
+        // so the success case is rejected by hand rather than with `expect_err`.
+        let error = match project_parsed_identity_update(&transition) {
+            Ok(_) => panic!("a document type with an interior NUL must not be narrowed"),
+            Err(error) => error,
+        };
+
+        assert_eq!(
+            error.code,
```
thepastaclaw, line 661: **🟡 Suggestion: Projection cleanup test never owns a previously projected key**

`rejects_contract_bounds_document_type_that_cannot_cross_the_ffi` supplies only the single key whose document-type conversion fails. Because `encode_contract_bounds` runs before that key's data is boxed and pushed, the error path in `project_parsed_identity_update` calls `free_parsed_public_keys(&mut add_public_keys_vec)` with an empty vector. A regression that leaked the data buffer or contract-bounds string of a key successfully projected before a later failure would leave this test green, since the non-empty cleanup branch is never exercised. Put a valid, successfully-boxed key before the invalid key in the fixture so the cleanup path actually frees prior allocations — this matters most under Miri or an allocator/leak-checking configuration, since this code manually reconstructs boxes from raw pointers.

<sub>source: ['codex']</sub>

thepastaclaw (commented):

This delta (2449ed1f..389dc4ce) contains one substantive commit scoped to fixing the previously-flagged unwiped stack copy in parse_secret_key: the FFI-borrowed private-key bytes are now staged in a Zeroizing<[u8;32]> buffer and parsed with SecretKey::from_byte_array instead of from_slice, closing that gap (finding 872eadfb2808 is FIXED, verified directly in secp256k1_primitives.rs). The two remaining carried-forward suggestions/nitpicks (an inherent, already-documented dashcore-pinning limitation in simple-signer's Drop, and a pointer-safety-contract nitpick in the same file) are untouched and still present. One new low-severity test-coverage gap was confirmed in identity_update.rs's contract-bounds-rejection test, which never exercises the non-empty cleanup path. The CodeRabbit contract-bounds-downgrade finding is verified fixed at this head (encode_contract_bounds now returns a Result and propagates the error). No blocking issues found.
Source: Codex general/security-auditor/rust-quality/ffi-engineer (gpt-5.6-sol), Claude general/security-auditor/rust-quality/ffi-engineer (claude-sonnet-5); final verifier: claude-sonnet-5. Orchestration-only: openclaw-agent/cliproxy/gpt-5.6-sol.

### Review provenance
- Sonnet reviewers: `claude-sonnet-5` — general (completed), `claude-sonnet-5` — security-auditor (completed), `claude-sonnet-5` — rust-quality (completed), `claude-sonnet-5` — ffi-engineer (completed)

🟡 1 suggestion(s)

<details>

```

In `packages/rs-platform-wallet-ffi/src/identity_update.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/identity_update.rs:598-639: Projection cleanup test never owns a previously projected key
  `rejects_contract_bounds_document_type_that_cannot_cross_the_ffi` supplies only the single key whose document-type conversion fails. Because `encode_contract_bounds` runs before that key's data is boxed and pushed, the error path in `project_parsed_identity_update` calls `free_parsed_public_keys(&mut add_public_keys_vec)` with an empty vector. A regression that leaked the data buffer or contract-bounds string of a key successfully projected before a later failure would leave this test green, since the non-empty cleanup branch is never exercised. Put a valid, successfully-boxed key before the invalid key in the fixture so the cleanup path actually frees prior allocations — this matters most under Miri or an allocator/leak-checking configuration, since this code manually reconstructs boxes from raw pointers.
```
</details>

romchornyi:
Addressed in 574e194f65.

Good catch — the test proved the error is returned, but not that the cleanup frees anything. `rejects_contract_bounds_document_type_that_cannot_cross_the_ffi` now puts a valid key (id 16, a `SingleContractDocumentType` bound with a real `"profile"` name) ahead of the failing one, so by the time `encode_contract_bounds` rejects the interior-NUL key, the first key's boxed data buffer and `CString` are live and `free_parsed_public_keys` runs over a non-empty vector.

`cargo test -p platform-wallet-ffi` and `cargo clippy --all-targets -- -D warnings` are both clean at that commit.

packages/rs-platform-wallet-ffi/src/identity_update.rs
thepastaclaw: Resolved in `574e194` — _Projection cleanup test never owns a previously projected key_ no longer present.

packages/rs-platform-wallet-ffi/src/identity_update.rs
```diff
@@ -20,6 +28,238 @@ use crate::runtime::block_on_worker;
...
+fn encode_contract_bounds(bounds: Option<&ContractBounds>) -> (u8, [u8; 32], *mut c_char) {
+    match bounds {
+        Some(ContractBounds::SingleContract { id }) => (1u8, id.to_buffer(), ptr::null_mut()),
+        Some(ContractBounds::SingleContractDocumentType {
+            id,
+            document_type_name,
+        }) => match CString::new(document_type_name.as_str()) {
+            Ok(value) => (2u8, id.to_buffer(), value.into_raw()),
+            Err(_) => (1u8, id.to_buffer(), ptr::null_mut()),
+        },
```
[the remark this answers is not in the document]
thepastaclaw: This no longer applies at head 80960f64b9: `encode_contract_bounds` returns a `Result`, and lines 136-145 return `ErrorInvalidParameter` when `CString::new` rejects the document type. The caller propagates that error after freeing previously projected keys, so no contract-bound downgrade remains.

shumkov (approved):

packages/simple-signer/src/single_key_signer.rs
shumkov: Acknowledged residual, unfixable downstream: SecretKey is Copy with no erasing Drop in pinned dashcore. Folding into the existing upstream zeroization work (rust-dashcore#843 — key-wallet Seed has the same defect); the in-code TODO(dashconnect-key-hygiene) references it. Not blocking per the finding itself.

shumkov (approved):
Reviewed with triage + codex-xhigh crypto verification (secp delegation, scoped-key hygiene, parse-only identity-update surface all verified; sole residual is the acknowledged upstream SecretKey erase, folded into rust-dashcore#843).

PR #4277, opened 2026-08-04
feat(platform-wallet): add encrypted txMetadata document support

QuantumExplorer:
## Summary

Adds the wallet-compatible encrypted `txMetadata` create/fetch path directly on the current `v4.2-dev` branch. This is the standalone replacement for the stacked work in #4186 and #4195, rebased after #4183 merged.

The change:

- implements the legacy-compatible `version | IV | AES-256-CBC(payload)` envelope and HD derivation in `platform-wallet`
- supports both resident-key and external-signable wallets through the mnemonic resolver, with key/plaintext zeroization around FFI boundaries
- exposes thin Kotlin/JNI create and decrypt-on-fetch APIs guarded by the manager teardown fence
- validates wire versions and payload size before key derivation, JNI/FFI copies, or network work
- uses a Rust-owned, non-zero 31-bit CSPRNG-generated `encryptionKeyIndex` while retaining the explicit-index ABI for migration and compatibility tests
- fetches and verifies owner-scoped documents with pagination and returns decrypted opaque payloads to the host

## Index redesign

The earlier #4195 design seeded a process-local high-water map from `1 + Platform document count`. That looked authoritative but could not provide an atomic cross-device allocation invariant, added a paginated network scan to the first create, and put mutable orchestration state on the `IdentityWallet` facade.

`encryptionKeyIndex` is a derivation input stored on each document, not a protocol sequence number. Readers derive from each document's stored fields, and every ciphertext carries a fresh IV, so a repeated index is non-lossy. Generating a valid child index at the Rust encryption boundary keeps the host thin without claiming uniqueness the protocol does not enforce.

## Rebase and review

- rebased the complete #4186 + #4195 + redesign range onto `v4.2-dev` at the #4183 merge
- resolved the only conflict by preserving both #4183's structured `SigningKeyUnavailable` mapping and the txMetadata oversized-payload mapping
- range-diff confirmed the pre-existing patches were otherwise unchanged
- removed broad JNI exception logging discovered during review; it affected unrelated operations, duplicated platform-wallet errors, and could persist full error details to Android logcat
- reviewed the final `v4.2-dev` diff with no remaining correctness findings

## Compatibility and impact

- legacy dashj and Rust wire/key derivation vectors are pinned in network-free tests
- explicit caller-supplied indices remain supported
- legacy documents remain readable because decryption uses their stored `keyIndex` and `encryptionKeyIndex`
- auto-index creates do not scan Platform or hold allocator state
- the global FFI wallet-handle store is released before resolver callbacks and network waits

## Validation

- `cargo test -p platform-wallet --lib` — 528 passed
- `cargo test -p platform-wallet-ffi --lib` — 224 passed
- `cargo clippy -p platform-wallet -p platform-wallet-ffi -p rs-unified-sdk-jni --lib -- -D warnings`
- `./gradlew :sdk:testDebugUnitTest --tests org.dashfoundation.dashsdk.documents.DocumentTransactionsVersionValidationTest`
- `./gradlew :sdk:lintDebug`

The live-testnet fetch regression remains `#[ignore]` and is intentionally a manual network gate.


thepastaclaw:
ℹ️ Review skipped (commit bfc9de23)
_Last checked: 2026-08-04 05:30 UTC_

PR #4280, opened 2026-08-04
fix(dpp): derive deterministic names for unnamed document type indexes

QuantumExplorer:

`Index::try_from` assigned a **random** 24-character alphanumeric name to any index declared without one, so two parses of the same contract produced different in-memory index names and a different iteration order of the name-keyed `BTreeMap<String, Index>`. This was found by a differential-equivalence harness where two runs of the same code on the same fixture corpus produced different transcripts (dozens of test fixture contracts, e.g. the family and dashpay corpora, declare unnamed indexes).

This is not consensus- or storage-visible: every document meta-schema version (v0/v1/v2) requires index `name`, block execution (`ValidationMode::Validator`) validates against the meta-schema before index parsing, and stored contracts persist the raw schema `Value` (never the parsed `Index`, which has no bincode derives). The random name was reachable only where validation is skipped — check_tx, client-side/SDK parses, and test fixtures — but there it broke parse reproducibility run-to-run and node-to-node.

## What was done?

- `packages/rs-dpp/src/data_contract/document_type/index/mod.rs`: replaced the random fallback name with a deterministic derivation from the index definition — properties joined as `{property}_{asc|desc}` (e.g. `ownerId_asc_createdAt_desc`), with `"index"` as the fallback for an empty property list. Removed the now-unused `rand::distributions::{Alphanumeric, DistString}` import.
- Documented the collision semantics: two unnamed indexes over identical properties and directions now derive the same name and collapse to one entry in the indices map — the correct outcome for a duplicate index declaration. (Verified no fixture in the repo has such duplicates, nor an explicit name colliding with a derived name.)
- No protocol version gate: the generated name is in-memory only and unreachable in any validated (consensus) path at every protocol version, so no shipped behavior changes.

## How Has This Been Tested?

- Replaced `test_index_try_from_without_name_generates_random` with `test_index_try_from_without_name_derives_deterministic_name` (asserts the derived name and that a second parse yields the same name) and added `test_index_try_from_without_name_multi_property_directions` (multi-property asc/desc derivation).
- `cargo test -p dpp --lib` — 3809 passed.
- `cargo test -p drive` — full suite passed (heavy consumer of the unnamed-index family/dashpay fixtures).
- `cargo clippy -p dpp --all-features --tests` clean (one pre-existing warning in `factory/v0/mod.rs`), `cargo fmt --all` applied.

## Breaking Changes

None.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency when creating data contract indexes by using stable, predictable names.
  * Enhanced support for indexes involving multiple properties and sort directions.
  * Added consistent naming for indexes without specified properties.

### DIFF
--- a/packages/rs-dpp/src/data_contract/document_type/index/mod.rs
+++ b/packages/rs-dpp/src/data_contract/document_type/index/mod.rs
@@ -19,7 +19,6 @@ use crate::data_contract::document_type::ContestedIndexResolution::MasternodeVot
 #[cfg(feature = "validation")]
 use crate::data_contract::errors::DataContractError::RegexError;
 use platform_value::{Value, ValueMap};
-use rand::distributions::{Alphanumeric, DistString};
 use regex::Regex;
 use std::cmp::Ordering;
 use std::sync::OnceLock;
@@ -1223,8 +1222,36 @@ impl Index {
             ));
         }
 
-        // if the index didn't have a name let's make one
-        let name = name.unwrap_or_else(|| Alphanumeric.sample_string(&mut rand::thread_rng(), 24));
+        // If the index didn't have a name, derive one deterministically from
+        // its properties and their directions. Every document meta-schema
+        // (v0/v1/v2) requires `name`, so an unnamed index can only reach this
+        // point when schema validation is skipped (check_tx, legacy fixtures,
+        // client-side parses of contracts that could never register); a random
+        // name here would make two parses of the same contract disagree on the
+        // index name and on the iteration order of the name-keyed indices map.
+        // Properties are joined with `|`, which cannot appear in a validated
+        // property path (path segments match `^[a-zA-Z0-9-_]{1,64}$`, joined
+        // by `.`, plus `$`-prefixed system properties), so two distinct index
+        // declarations can never derive the same name; only true duplicate
+        // declarations collide and collapse to one entry in the name-keyed
+        // map, which is the right outcome for a duplicate index.
+        let name = name.unwrap_or_else(|| {
+            if index_properties.is_empty() {
+                "index".to_string()
+            } else {
+                index_properties
+                    .iter()
+                    .map(|property| {
+                        format!(
+                            "{}_{}",
+                            property.name,
+                            if property.ascending { "asc" } else { "desc" }
+                        )
+                    })
+                    .collect::<Vec<_>>()
+                    .join("|")
+            }
+        });
 
         Ok(Index {
             name,
@@ -1795,7 +1822,7 @@ mod tests {
     }
 
     #[test]
-    fn test_index_try_from_without_name_generates_random() {
+    fn test_index_try_from_without_name_derives_deterministic_name() {
         let index_map: Vec<(Value, Value)> = vec![(
             Value::Text("properties".to_string()),
             Value::Array(vec![Value::Map(vec![(
@@ -1804,8 +1831,70 @@ mod tests {
             )])]),
         )];
         let index = Index::try_from(index_map.as_slice()).unwrap();
-        assert!(!index.name.is_empty());
-        assert_eq!(index.name.len(), 24); // Alphanumeric.sample_string with len 24
+        assert_eq!(index.name, "fieldA_asc");
+
+        // Parsing the same definition again must produce the same name
+        let again = Index::try_from(index_map.as_slice()).unwrap();
+        assert_eq!(again.name, index.name);
+    }
+
+    #[test]
+    fn test_index_try_from_without_name_multi_property_directions() {
+        let index_map: Vec<(Value, Value)> = vec![(
+            Value::Text("properties".to_string()),
+            Value::Array(vec![
+                Value::Map(vec![(
+                    Value::Text("ownerId".to_string()),
+                    Value::Text("asc".to_string()),
+                )]),
+                Value::Map(vec![(
+                    Value::Text("createdAt".to_string()),
+                    Value::Text("desc".to_string()),
+                )]),
+            ]),
+        )];
+        let index = Index::try_from(index_map.as_slice()).unwrap();
+        assert_eq!(index.name, "ownerId_asc|createdAt_desc");
+    }
+
+    #[test]
+    fn test_index_try_from_without_name_empty_properties_falls_back() {
+        let index_map: Vec<(Value, Value)> =
+            vec![(Value::Text("properties".to_string()), Value::Array(vec![]))];
+        let index = Index::try_from(index_map.as_slice()).unwrap();
+        assert_eq!(index.name, "index");
+    }
+
+    #[test]
+    fn test_index_try_from_derived_names_distinct_for_distinct_definitions() {
+        // A single property literally named "a_asc_b" (desc) must not derive
+        // the same name as a compound index over "a" (asc) + "b" (desc):
+        // the `|` joiner cannot appear in a validated property path.
+        let single: Vec<(Value, Value)> = vec![(
+            Value::Text("properties".to_string()),
+            Value::Array(vec![Value::Map(vec![(
+                Value::Text("a_asc_b".to_string()),
+                Value::Text("desc".to_string()),
+            )])]),
+        )];
+        let compound: Vec<(Value, Value)> = vec![(
+            Value::Text("properties".to_string()),
+            Value::Array(vec![
+                Value::Map(vec![(
+                    Value::Text("a".to_string()),
+                    Value::Text("asc".to_string()),
+                )]),
+                Value::Map(vec![(
+                    Value::Text("b".to_string()),
+                    Value::Text("desc".to_string()),
+                )]),
+            ]),
+        )];
+        let single_index = Index::try_from(single.as_slice()).unwrap();
+        let compound_index = Index::try_from(compound.as_slice()).unwrap();
+        assert_eq!(single_index.name, "a_asc_b_desc");
+        assert_eq!(compound_index.name, "a_asc|b_desc");
+        assert_ne!(single_index.name, compound_index.name);
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


thepastaclaw:
🔍 Review in progress — actively reviewing now (commit b22d10a1)

packages/rs-dpp/src/data_contract/document_type/index/mod.rs
```diff
@@ -981,8 +980,33 @@ impl TryFrom<&[(Value, Value)]> for Index {
...
+                        format!(
+                            "{}_{}",
+                            property.name,
+                            if property.ascending { "asc" } else { "desc" }
+                        )
+                    })
+                    .collect::<Vec<_>>()
+                    .join("_")
+            }
+        });
```
thepastaclaw: **🟡 Suggestion: Derived index name is not injective — property-set collisions silently drop indexes from the BTreeMap**

The derivation joins each property's `{name}_{asc|desc}` token with `_`, using the same character as both the intra-token separator and the inter-property joiner. The document meta-schema's `propertyNames` pattern (`^[a-zA-Z0-9-_]{1,64}$`, present in v0/v1/v2 `document-meta.json`) explicitly permits `_` in property names, and that constraint isn't enforced on the unvalidated code path this fallback targets, so the scheme is not collision-free between *different* index definitions.

Concrete collision: a compound index over properties `a` (asc) and `b` (desc) derives `"a_asc_b_desc"`. A single-property index over a property literally named `a_asc_b` with direction `desc` also derives `"a_asc_b_desc"` (verified against the code — a single property just contributes one `{name}_{dir}` token with no extra join needed). These are two semantically distinct index declarations — different property sets, different tree shapes — that end up as the same key.

`parse_indices` (packages/rs-dpp/src/data_contract/document_type/class_methods/try_from_schema/common/mod.rs:885, feeding the `.collect::<BTreeMap<String, Index>>()` at line 728) collects `(index.name.clone(), index)` pairs directly, so a name collision silently overwrites one entry instead of erroring — this contradicts the PR's own commentary that collapsing here 'is the right outcome for a duplicate index declaration,' since the collision case above is not a duplicate declaration at all.

Impact is bounded: this fallback is unreachable for any contract that has passed meta-schema validation (a name is mandatory there), so it can't affect consensus, storage, or proof correctness. It can, however, cause check_tx-time or SDK/client-side parsing of not-yet-validated contract JSON to silently lose one of two distinct index declarations that a client believes are both present.

```suggestion
let name = name.unwrap_or_else(|| {
            if index_properties.is_empty() {
                "index".to_string()
            } else {
                // Length-prefix each property name so the joined string stays
                // injective even when a property name itself contains '_',
                // 'asc', or 'desc' -- this fallback runs exactly where the
                // meta-schema's property-name pattern isn't enforced.
                index_properties
                    .iter()
                    .map(|property| {
                        format!(
                            "{}:{}_{}",
                            property.name.len(),
                            property.name,
                            if property.ascending { "asc" } else { "desc" }
                        )
                    })
                    .collect::<Vec<_>>()
                    .join("_")
            }
        });
```

packages/rs-dpp/src/data_contract/document_type/index/mod.rs
```diff
@@ -1556,8 +1580,30 @@ mod tests {
...
+                    Value::Text("ownerId".to_string()),
+                    Value::Text("asc".to_string()),
+                )]),
+                Value::Map(vec![(
+                    Value::Text("createdAt".to_string()),
+                    Value::Text("desc".to_string()),
+                )]),
+            ]),
+        )];
+        let index = Index::try_from(index_map.as_slice()).unwrap();
```
thepastaclaw, line 1856: **💬 Nitpick: No test covers the empty-properties fallback (`"index"`) or the name-collision edge case**

The two new tests cover the happy-path single- and multi-property derivations and correctly assert re-parse stability, but two edge cases from this same change are untested: (1) the `index_properties.is_empty()` branch that falls back to the literal `"index"` — reachable when `properties` is `[]` or absent, since nothing before this point requires a non-empty property list; and (2) the collision case from the sibling finding, where a single property literally named e.g. `a_asc_b` (desc) derives the same name as two properties `a:asc, b:desc`. Adding regression tests for both would close this coverage gap and make the collapsing behavior an explicit, verifiable contract rather than an implicit side effect of the join scheme.

<sub>source: ['claude']</sub>

thepastaclaw (commented):

This PR correctly replaces a random 24-char index-name fallback with a deterministic derivation, fixing run-to-run/node-to-node reproducibility for unnamed indexes on the check_tx/client/fixture-only parse path (verified: every meta-schema version requires `name`, so the fallback never fires on the validated/consensus path). The one real gap, independently identified and converged upon by all Claude and Codex reviewer lanes, is that the `{prop}_{asc|desc}` join reuses `_` as both the intra-token and inter-property separator, so two structurally different unnamed index declarations (e.g. a property literally named `a_asc_b` with direction `desc` vs. two properties `a:asc, b:desc`) derive the identical name string and silently collapse into one `BTreeMap` entry — verified against the meta-schema's `^[a-zA-Z0-9-_]{1,64}$` property-name pattern (underscores are legal) and the `parse_indices` collect-into-BTreeMap call site. This contradicts the PR's own stated invariant that collapsing only happens for true duplicate declarations, though it remains bounded to the non-consensus-visible fallback path.

Source: reviewers codex/general=gpt-5.6-sol(completed); codex/security-auditor=gpt-5.6-sol(completed); codex/rust-quality=gpt-5.6-sol(completed); claude/general=claude-sonnet-5(completed); claude/security-auditor=claude-sonnet-5(completed); claude/rust-quality=claude-sonnet-5(failed); claude/rust-quality=claude-sonnet-5(completed); verifier=claude/final-verifier=claude-sonnet-5(completed); coordinator=openclaw-agent/cliproxy/gpt-5.6-sol(orchestration-only).

### Review provenance
- Sonnet reviewers: `claude-sonnet-5` — general (completed), `claude-sonnet-5` — security-auditor (completed), `claude-sonnet-5` — rust-quality (failed), `claude-sonnet-5` — rust-quality (completed)

<details>

```

In `packages/rs-dpp/src/data_contract/document_type/index/mod.rs`:
- [SUGGESTION] packages/rs-dpp/src/data_contract/document_type/index/mod.rs:992-1009: Derived index name is not injective — property-set collisions silently drop indexes from the BTreeMap
  The derivation joins each property's `{name}_{asc|desc}` token with `_`, using the same character as both the intra-token separator and the inter-property joiner. The document meta-schema's `propertyNames` pattern (`^[a-zA-Z0-9-_]{1,64}$`, present in v0/v1/v2 `document-meta.json`) explicitly permits `_` in property names, and that constraint isn't enforced on the unvalidated code path this fallback targets, so the scheme is not collision-free between *different* index definitions.

  Concrete collision: a compound index over properties `a` (asc) and `b` (desc) derives `"a_asc_b_desc"`. A single-property index over a property literally named `a_asc_b` with direction `desc` also derives `"a_asc_b_desc"` (verified against the code — a single property just contributes one `{name}_{dir}` token with no extra join needed). These are two semantically distinct index declarations — different property sets, different tree shapes — that end up as the same key.

  `parse_indices` (packages/rs-dpp/src/data_contract/document_type/class_methods/try_from_schema/common/mod.rs:885, feeding the `.collect::<BTreeMap<String, Index>>()` at line 728) collects `(index.name.clone(), index)` pairs directly, so a name collision silently overwrites one entry instead of erroring — this contradicts the PR's own commentary that collapsing here 'is the right outcome for a duplicate index declaration,' since the collision case above is not a duplicate declaration at all.

  Impact is bounded: this fallback is unreachable for any contract that has passed meta-schema validation (a name is mandatory there), so it can't affect consensus, storage, or proof correctness. It can, however, cause check_tx-time or SDK/client-side parsing of not-yet-validated contract JSON to silently lose one of two distinct index declarations that a client believes are both present.
```
</details>

packages/rs-dpp/src/data_contract/document_type/index/mod.rs
QuantumExplorer: Fixed in b22d10a112: the inter-property joiner is now `|`, which cannot appear in a validated property path (segments match `^[a-zA-Z0-9-_]{1,64}$`, joined by `.`, plus $-prefixed system properties), so distinct definitions can no longer derive the same name; added a regression test proving `a_asc_b` (desc) ≠ `a` asc + `b` desc.

packages/rs-dpp/src/data_contract/document_type/index/mod.rs
QuantumExplorer: Added both tests in b22d10a112: `test_index_try_from_without_name_empty_properties_falls_back` (asserts the `"index"` fallback) and `test_index_try_from_derived_names_distinct_for_distinct_definitions` (asserts the collision pair now derives distinct names under the `|` joiner).

packages/rs-dpp/src/data_contract/document_type/index/mod.rs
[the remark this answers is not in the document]
QuantumExplorer: Addressed in b22d10a112: the joiner is now `|` (illegal in validated property paths), making the derivation injective for any registrable contract, and the empty-properties fallback plus the collision edge case are covered by new regression tests.

QuantumExplorer:
Note: the "Rust workspace tests / Tests" failure on b22d10a112 was infrastructure, not a test failure — the job landed on a self-hosted runner without `cargo-nextest` and died with `no such command: nextest` before running anything (the codecov failures were downstream of the missing upload). Commit 3c2bc32950 adds a `cargo install cargo-nextest` guard next to the existing `cargo-llvm-cov` install so the job is runner-independent. Happy to split it into a separate ci PR if preferred.

QuantumExplorer (commented):
Reviewed

PR #4281, opened 2026-08-04
fix(drive-abci)!: cross-check the contested index of a prefunded voting balance at protocol v14

QuantumExplorer:

A document create transition carries the index its prefunded voting balance is for, in `DocumentCreateTransitionV0::prefunded_voting_balance: Option<(String, Credits)>`. That name is what keys the contested vote poll, its stored info, its end-date entry and its prefunded specialized balance — but the contested index tree the contender is actually inserted into always comes from `DocumentType::find_contested_index()`. Nothing tied the two together:

- the action transformer resolved whatever index was named, as long as it existed on the document type (`document_type_indexes.get(index_name)`);
- the advanced structure check validated only the amount and dropped the name (`if let Some((_, paid_amount)) = ...`).

Three consequences, all verified against the pre-fix code by flipping only the validation version and re-running the tests added here:

**1. Reachable chain halt.** On a contract whose non-contested index is a strict *prefix* of the contested one (e.g. `[normalizedParentDomainName]` next to the contested `[normalizedParentDomainName, normalizedLabel]`), the mismatched vote poll's contenders path is a tree the contested insert creates anyway. The transition therefore **executes successfully** and registers the poll one level above the contest it describes. When that poll's end date arrives, the tally query finds no lock/abstain sum trees and `check_for_ended_vote_polls` returns `Err(Drive(CorruptedCodeExecution("expected a locked tally")))`. That is a per-block event handler with a bare `?` (`run_dao_platform_events` → `run_block_proposal`), so it is not contained into a per-state-transition result — it halts every validator at that block. The end-date entry is never cleaned up, so every subsequent block proposal fails identically. Any user can register such a contract.

**2. Contest checks silently skipped.** `current_store_contest_info` is fetched under the vote poll built from the submitter's own index name, so a mismatched name makes that lookup return `None`, and `state_v0`/`state_v1` then skip the `Locked`, join-window and already-a-contestant checks entirely while the contender is still inserted into the real contest.

**3. Unnecessary contests.** In the other direction, a document that resolves to no contested index at all could still supply a prefunded voting balance, storing an ordinary document as a contender that only becomes registered if it wins a masternode vote.

On the live DPNS contract the mismatch currently dies for an unrelated reason: `Index::extract_values` is a flat map lookup, so the `identityId` index's dotted `records.identity` property yields `Value::Null` → an empty path segment → an unpaid `InternalError` from the estimated-cost path cache. That is an accident of DPNS's index shape, not a defense.

## What was done?

`document_create_transition_structure_validation` v1 (new `advanced_structure_v1` module) compares the full expected `ContestedDocumentResourceVotePoll` — derived from the document's own properties via `contested_vote_poll_for_document_properties` — against the one built from the prefunded voting balance, and rejects a prefunded voting balance on a document that resolves to no contested index. Everything else is carried over from v0 unchanged, including the testnet-before-epoch-2080 carve-out.

Two new consensus errors:

| Code | Error |
| --- | --- |
| 40118 | `DocumentContestIndexMismatchError` |
| 40119 | `DocumentContestNotRequiredError` |

The check changes accept/reject behavior, so it is gated by a new `DRIVE_ABCI_VALIDATION_VERSIONS_V10` wired into `v14.rs`. `DRIVE_ABCI_VALIDATION_VERSIONS_V9` is shared with protocol version 13 and stays byte-identical for chain replay; `advanced_structure_v0` is untouched.

## How Has This Been Tested?

Five unit tests on the validation dispatcher (`advanced_structure_v1::tests`): mismatched index name, prefunded balance on a non-contested document, the preserved not-paid-for behavior, the accepted matching case, and a test pinning that protocol version 13 does **not** cross-check — the last one is what documents the gap this PR closes.

Four end-to-end tests through `process_raw_state_transitions`:

- `test_document_creation_on_contested_unique_index_should_fail_if_prefunding_another_index` — DPNS contest funded for `identityId`; asserts the paid consensus error and that neither a contender nor a document was created.
- `test_contest_can_not_be_joined_after_the_join_window_by_prefunding_another_index` — opens a real contest, advances past the one-week join window, and shows a mismatched late join is rejected and the contest keeps exactly its two contenders.
- `contested_index_mismatch_chain_halt::prefunding_a_prefix_index_is_rejected` — the halt scenario above, on a new `dpns-contract-contested-unique-index-and-prefix-index.json` fixture. Run against structure validation v0 this same test shows the transition executing and the resolver returning `Err(... "expected a locked tally")`.
- `contested_index_mismatch_chain_halt::prefunding_the_contested_index_resolves_successfully` — control: the identical contest funded for the contested index executes and its poll resolves cleanly, so the index name is the only variable.

Suites: `cargo test -p drive-abci --lib` 2638 passed / 0 failed, `cargo test -p dpp` 3808 passed / 0 failed, `cargo check --workspace --all-targets` clean, `cargo clippy` clean on the touched crates, and `cargo check -p wasm-dpp --target wasm32-unknown-unknown` clean.

Worth knowing for anyone extending these tests: the default test `PlatformConfig` network is Testnet, where the whole contested structure-validation block is skipped before epoch 2080 — a contested-validation test that forgets `network: Network::Mainnet` passes while proving nothing.

## Breaking Changes

Consensus-breaking, gated at protocol version 14. From v14 on, a create transition whose prefunded voting balance names an index other than the contested index its document resolves to, or that supplies one for a non-contested document, is rejected with a paid consensus error instead of being accepted. Clients that derive the field through `DocumentType::prefunded_voting_balance_for_document` (which returns the contested index) are unaffected. Protocol version 13 and below replay unchanged.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added document-creation validation to ensure prefunded voting balances reference the correct contested index.
  * Added clear errors for mismatched or unnecessary contested-document funding.
  * Enabled the updated validation rules in platform version 14.

* **Bug Fixes**
  * Prevented invalid contested-document configurations that could disrupt vote-poll processing.
  * Added compatibility handling for testnet and earlier validation versions.

* **Tests**
  * Added comprehensive coverage for valid funding, mismatched indexes, missing funding, and contest restrictions.


thepastaclaw:
🕓 Ready for review — next in queue (commit 04bf3f10)
_Queue position: 1/2 · 2 reviews active_
_ETA: start ~06:53 UTC · complete ~07:17 UTC (median 23m across 30 recent reviews; 2 slots)_
_Queued 33m ago · Last checked: 2026-08-05 06:40 UTC_

packages/rs-dpp/src/errors/consensus/state/state_error.rs
```diff
@@ -96,6 +98,12 @@ pub enum StateError {
         DocumentContestDocumentWithSameIdAlreadyPresentError,
     ),
 
+    #[error(transparent)]
+    DocumentContestIndexMismatchError(DocumentContestIndexMismatchError),
+
+    #[error(transparent)]
+    DocumentContestNotRequiredError(DocumentContestNotRequiredError),
```
thepastaclaw: **🔴 Blocking: Append the new StateError variants to preserve wire discriminants**

StateError is encoded by ordinal variant position: its bincode and platform-serialization implementations write the variant index as a u32, and the enum explicitly prohibits reordering without a new serialization version. Inserting these variants here changes DocumentNotFoundError from ordinal 8 to 10 and shifts every subsequent existing variant by two. ErrorWithCode values do not protect this representation. Consensus errors are serialized into ABCI/DAPI responses and decoded by WASM and JavaScript clients, so upgraded and older components can fail decoding or interpret an existing error as a different variant. Move both new variants after InsufficientShieldedFeeError so every pre-existing discriminator remains unchanged; add a frozen-byte compatibility test for an existing variant after the current insertion point.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The protocol-v14 vote-poll cross-check is correctly version-dispatched and has strong regression coverage. One blocking compatibility issue remains: the new StateError variants were inserted into a positionally serialized enum, shifting the wire discriminants of all existing variants after the insertion point. The pre-2080 Testnet carve-out does not expose the fix on a reachable historical configuration because validation v1 exists only at protocol v14, while the carve-out is intentional compatibility behavior covered by an existing test.

Source: reviewers codex/general=gpt-5.6-sol(completed); codex/security-auditor=gpt-5.6-sol(completed); codex/rust-quality=gpt-5.6-sol(completed); codex/ffi-engineer=gpt-5.6-sol(completed); verifier=codex/verifier=gpt-5.6-sol(completed); coordinator=openclaw-agent/cliproxy/gpt-5.6-sol(orchestration-only).

### Review provenance

🔴 1 blocking

<details>

```

In `packages/rs-dpp/src/errors/consensus/state/state_error.rs`:
- [BLOCKING] packages/rs-dpp/src/errors/consensus/state/state_error.rs:101-105: Append the new StateError variants to preserve wire discriminants
  StateError is encoded by ordinal variant position: its bincode and platform-serialization implementations write the variant index as a u32, and the enum explicitly prohibits reordering without a new serialization version. Inserting these variants here changes DocumentNotFoundError from ordinal 8 to 10 and shifts every subsequent existing variant by two. ErrorWithCode values do not protect this representation. Consensus errors are serialized into ABCI/DAPI responses and decoded by WASM and JavaScript clients, so upgraded and older components can fail decoding or interpret an existing error as a different variant. Move both new variants after InsufficientShieldedFeeError so every pre-existing discriminator remains unchanged; add a frozen-byte compatibility test for an existing variant after the current insertion point.
```
</details>

packages/rs-dpp/src/errors/consensus/state/state_error.rs
QuantumExplorer: Fixed in 6058ba5c1a — both variants moved to the end of `StateError` (after `InsufficientShieldedFeeError`), so `DocumentNotFoundError` is back at ordinal 8 and no pre-existing discriminant moves; added `state_error_discriminants_are_frozen` pinning the bincode discriminants of the first variant, `DocumentNotFoundError`, and the last three.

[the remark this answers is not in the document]
QuantumExplorer: Fixed in 6058ba5c1a — both new variants moved to the end of `StateError`, restoring every pre-existing ordinal (`DocumentNotFoundError` back to 8), plus a `state_error_discriminants_are_frozen` test so a future mid-enum insertion fails loudly.

QuantumExplorer (commented):
Reviewed

PR #4282, opened 2026-08-04
fix(dashmate)!: give Debian packages versions apt can order

shumkov:

`apt` cannot order dashmate's Debian packages, and nothing we publish can be verified.

oclif builds the deb version as `<upstream>.<git sha>-1`. That discards the semver prerelease tag and makes the git sha an ordering component — and under dpkg's comparison digits sort below letters, so ordering between builds is effectively random. Verified against the real published 4.1.0 series:

| transition | deb versions | apt verdict |
|---|---|---|
| beta.2 → rc.1 | `4.1.0.08152ea51e-1` vs `4.1.0.ae554fdd83-1` | **refused as a downgrade** |
| rc.1 → rc.2 | `4.1.0.3de436123d-1` vs `4.1.0.08152ea51e-1` | **refused as a downgrade** |
| rc.2 → rc.3 | `4.1.0.61be67f7bf-1` vs `4.1.0.3de436123d-1` | upgrade |
| rc.3 → 4.1.0 | `4.1.0.bfc80249b9-1` vs `4.1.0.61be67f7bf-1` | upgrade |

Two of four real transitions are read as downgrades. Worse, a **same-version security rebuild never ships**: with `4.1.0.bfc80249b9-1` installed, a hotfix built from a later commit compares *lower*, so apt reports the package as already newest and the operator believes they are patched.

Separately, nothing in the release is verifiable — no checksums, no signatures beyond the macOS notarisation, and the apt metadata oclif generates is uploaded unsigned and unserved.

## What was done?

**Monotonic Debian versions.** `4.1.0~rc.3-1` sorts below `4.1.0-1`; rebuilds bump the Debian revision; the git sha moves into the package description where it cannot affect ordering. Filenames deliberately diverge from the control version, because GitHub rewrites `~` and `:` in release asset names — the control field keeps them, the filename drops them, and `dpkg-name` strips epochs from filenames for the same reason.

**A release-blocking ordering gate.** It reads every version from the published package via `dpkg-deb -f` rather than deriving it from a tag, and the packaging job then asserts the built package carries the version that was gated — so the check binds to the bytes that ship.

**Verifiable releases.** Every published asset is hashed into a deterministic `SHA256SUMS`, in a job that checks out nothing and installs nothing, so no dependency lifecycle script can run beside the signing key that job will later hold. Each packaging leg records the hashes it produced and the checksum job refuses any asset it did not build. npm packages publish with provenance where the registry accepts it. Third-party actions are pinned to commit SHAs, Binaryen is checksummed, and the jobs holding credentials name environments so tag and reviewer policies attach to them.

## How Has This Been Tested?

**Version mapping and ordering** — 846 unit tests, plus every transition in the real 4.1.0 series checked against actual `dpkg`: `4.1.0~beta.2-1 → ~rc.1 → ~rc.2 → ~rc.3 → 4.1.0-1 → 4.1.1-1` all sort strictly upward, where two of those four transitions are read as downgrades under the current scheme. The mapping is injection-proof (shell metacharacters, command substitution and embedded newlines are rejected) and `~` ordering agrees with semver precedence across every prerelease form it can produce.

**Repack, index and install — end to end in a Debian container, against the real published `v4.2.0-dev.7` package.** This is the part that had never executed anywhere, since `apt-ftparchive` and `dpkg-deb` do not exist on macOS:

- control `Version:` rewritten `4.2.0.17a2962f91-1` → `4.2.0~dev.7-1`
- filename `dashmate_4.2.0.dev.7-1_amd64.deb` — the tilde dropped, because GitHub rewrites it in release asset names, while the control field keeps it
- git sha moved into the package description, where it cannot affect ordering
- all 27,942 payload entries preserved, with permissions and symlinks unchanged (compared by checksum before and after)
- regenerated index self-consistent: `Filename:` resolves to a real file and `SHA256:` matches the rewritten package
- **apt read that index, offered the package as a candidate, and installed it**, reporting `4.2.0~dev.7-1`; a later version then sorts above it under real `dpkg`

**Baseline selection** can be run against live GitHub data without publishing anything — that was a reason for extracting it from the workflow:

```
packages/dashmate/scripts/check-release-deb-version.sh v4.2.0-dev.8
```

**Not exercised.** The signing environments, Apple notarization, npm provenance acceptance, and the checksum record-and-verify round trip only run on a real release. The version gate fails closed, so the failure mode there is a red release job rather than a bad package. The gate job is also reachable via `workflow_dispatch` once this merges — it has no `needs` and `contents: read`, and the publish jobs skip on dispatch, so it can be exercised in real CI without publishing.

## Breaking Changes

Debian package filenames and versions change shape. A rebuild of an already-published version needs `DASHMATE_DEB_REVISION`; re-releasing a version published under the old scheme needs `DASHMATE_DEB_EPOCH`, because `4.1.0-1` sorts *below* the old `4.1.0.bfc80249b9-1`.

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Debian installation instructions automatically download the latest package for the system architecture.
  - Release packages include validated Debian-compatible versions and recorded checksums.
  - npm packages are published with provenance information.

- **Bug Fixes**
  - Release validation helps prevent Debian packages from receiving versions that cannot upgrade existing installations.

- **Documentation**
  - Updated installation guidance and package repository metadata across the project.
  - Updated package versions for the latest development release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -381,8 +381,9 @@ jobs:
           set -euo pipefail
           compare="${GITHUB_WORKSPACE}/scripts/check_deb_version.sh"
           translate="${GITHUB_WORKSPACE}/scripts/deb_version.js"
-          if [ ! -x "${compare}" ] || [ ! -f "${translate}" ]; then
-            echo "::error::${compare} or ${translate} is missing"
+          select_baseline="${GITHUB_WORKSPACE}/scripts/deb_release_baseline.js"
+          if [ ! -x "${compare}" ] || [ ! -f "${translate}" ] || [ ! -f "${select_baseline}" ]; then
+            echo "::error::${compare}, ${translate} or ${select_baseline} is missing"
             exit 1
           fi
           # The comparison runs on Debian versions, never on the semver tags:
@@ -393,68 +394,21 @@ jobs:
           # independently predicting it from the tag.
           echo "validated_version=${new_version}" >> "${GITHUB_OUTPUT}"
 
-          # Which release line a tag belongs to. Used only to group releases, so
-          # an older-line hotfix is measured against its own predecessor instead
-          # of a higher line it was never meant to supersede. Every version that
-          # actually gets compared still comes from a deb's control field.
-          line_of_tag() {
-            local tag="${1#v}"
-            printf '%s' "${tag%%-*}" | cut -d. -f1,2
-          }
-          current_line="$(line_of_tag "${CURRENT_TAG}")"
+          # Every release is fetched, not just the first page: the repository has
+          # several hundred of them, and a single page silently hides older
+          # lines. Which one becomes the baseline is decided by the script rather
+          # than by the API's default ordering, so the choice of predecessor is
+          # not an undocumented implementation detail.
+          baseline="$(gh api --paginate --slurp "repos/${GITHUB_REPOSITORY}/releases?per_page=100" \
+            | node "${select_baseline}" "${CURRENT_TAG}")"
 
-          # Candidates are non-draft releases that shipped a deb, newest first;
-          # prereleases count, because they go to the same channel and apt
-          # compares against whatever the operator installed last. The current
-          # tag is included: a deb already attached to it is what an earlier run
-          # of this same release shipped, and it is exactly what apt measures a
-          # rebuild against when the revision was not bumped.
-          # Paginated and sorted explicitly: the repository has several hundred
-          # releases, so a single page silently hides older lines, and relying on
-          # the API's default ordering would make the choice of predecessor an
-          # undocumented implementation detail.
-          candidates="${RUNNER_TEMP}/deb-release-candidates"
-          gh api --paginate --slurp "repos/${GITHUB_REPOSITORY}/releases?per_page=100" \
-            | jq -r '
-                add
-                | sort_by(.created_at)
-                | reverse
-                | .[]
-                | select(.draft == false)
-                | { tag: .tag_name, asset: ([.assets[].name | select(endswith("_amd64.deb"))] | first) }
-                | select(.asset != null)
-                | "\(.tag)\t\(.asset)"' > "${candidates}"
-
-          # The newest release on this line, falling back to the newest of any
-          # line when this line has not shipped a deb yet. Deliberately NOT "the
-          # newest release sorting below this one": that would choose the
-          # baseline using the very condition the comparison goes on to assert,
-          # leaving a gate that cannot fail.
-          baseline_tag=""
-          baseline_asset=""
-          newest_tag=""
-          newest_asset=""
-          while IFS=$'\t' read -r tag asset; do
-            if [ -z "${newest_tag}" ]; then
-              newest_tag="${tag}"
-              newest_asset="${asset}"
-            fi
-            if [ "$(line_of_tag "${tag}")" = "${current_line}" ]; then
-              baseline_tag="${tag}"
-              baseline_asset="${asset}"
-              break
-            fi
-          done < "${candidates}"
-
-          if [ -z "${baseline_tag}" ]; then
-            baseline_tag="${newest_tag}"
-            baseline_asset="${newest_asset}"
-          fi
-
-          if [ -z "${baseline_tag}" ]; then
+          if [ -z "${baseline}" ]; then
             echo "::notice::No published deb to compare ${CURRENT_TAG} against"
             exit 0
           fi
+
+          baseline_tag="${baseline%%$'\t'*}"
+          baseline_asset="${baseline#*$'\t'}"
 
           # Read the baseline from the deb's own control field: the only version
           # apt looks at, the only one a server-side rename of the asset cannot
@@ -617,14 +571,25 @@ jobs:
             hash_cmd=(shasum -a 256)
           fi
           cd packages/dashmate/dist
-          find . -type f | sed 's|^\./||' | LC_ALL=C sort > "${RUNNER_TEMP}/built-files"
+          # Hidden files are skipped to match the glob that uploads this
+          # directory, which does not match them either. Recording a file that
+          # never gets uploaded would fail the release for a package that was
+          # never meant to ship.
+          find . -type f -not -path '*/.*' | sed 's|^\./||' | LC_ALL=C sort > "${RUNNER_TEMP}/built-files"
           if [ ! -s "${RUNNER_TEMP}/built-files" ]; then
             echo "::error::No built packages found to record"
             exit 1
           fi
+          # Recorded under the name the file will carry on the release rather
+          # than its path in the build tree: a release has no directories, so
+          # the upload flattens every file to its basename. A record keyed by
+          # path could only ever be matched against a published asset by hash,
+          # which is what lets bytes built for one target be served under
+          # another target's name.
           : > "${RUNNER_TEMP}/built.sha256"
           while IFS= read -r file; do
-            "${hash_cmd[@]}" "${file}" >> "${RUNNER_TEMP}/built.sha256"
+            hash="$("${hash_cmd[@]}" "${file}" | cut -d' ' -f1)"
+            printf '%s  %s\n' "${hash}" "${file##*/}" >> "${RUNNER_TEMP}/built.sha256"
           done < "${RUNNER_TEMP}/built-files"
           cat "${RUNNER_TEMP}/built.sha256"
 
@@ -680,67 +645,93 @@ jobs:
             exit 1
           fi
 
-      - name: Download published release assets
-        env:
-          GH_TOKEN: ${{ github.token }}
-          TAG: ${{ github.event.release.tag_name }}
-        run: |
-          set -euo pipefail
-          mkdir -p assets
-          gh release download "${TAG}" --repo "${GITHUB_REPOSITORY}" --dir assets --clobber
-          # Drop the output of an earlier run of this job so a re-run hashes
-          # only the packages.
-          rm -f assets/SHA256SUMS
-
       - name: Download built package checksums
         uses: actions/download-artifact@v4
         with:
           pattern: dashmate-built-checksums-*
           path: built-checksums
 
+      # The names and hashes the packaging jobs recorded are the whole
+      # expectation this job holds: SHA256SUMS covers the packages this workflow
+      # built, not whatever the release happens to serve. Other jobs attach
+      # assets of their own, and this job has no build record for those - listing
+      # them would attest artifacts it knows nothing about, and would depend on
+      # whether those jobs had finished yet.
+      - name: Collect the packages built in this run
+        run: |
+          set -euo pipefail
+          cat built-checksums/*/built.sha256 | LC_ALL=C sort -u > "${RUNNER_TEMP}/built.sha256"
+          if [ ! -s "${RUNNER_TEMP}/built.sha256" ]; then
+            echo "::error::No built package checksums were recorded"
+            exit 1
+          fi
+          cut -d' ' -f3- "${RUNNER_TEMP}/built.sha256" | LC_ALL=C sort > "${RUNNER_TEMP}/built-names"
+
+          # A release holds one asset per name, so two built files sharing a name
+          # means one of them silently replaced the other on the way up and there
+          # is no longer any way to say which bytes are published under it.
+          duplicates="$(LC_ALL=C uniq -d "${RUNNER_TEMP}/built-names")"
+          if [ -n "${duplicates}" ]; then
+            echo "::error::More than one built file would be published under the same name: $(echo "${duplicates}" | tr '\n' ' ')"
+            exit 1
+          fi
+          cat "${RUNNER_TEMP}/built.sha256"
+
+      # Fetched by name, one asset at a time, so that a package built here but
+      # missing from the release is reported as missing rather than quietly
+      # dropped from the listing.
+      - name: Download the published packages
+        env:
+          GH_TOKEN: ${{ github.token }}
+          TAG: ${{ github.event.release.tag_name }}
+        run: |
+          set -euo pipefail
+          mkdir -p assets
+          missing=0
+          while IFS= read -r name; do
+            gh release download "${TAG}" --repo "${GITHUB_REPOSITORY}" \
+              --pattern "${name}" --dir assets --clobber > /dev/null 2>&1 || true
+            if [ ! -f "assets/${name}" ]; then
+              echo "::error::${name} was built in this run but is not published on ${TAG}"
+              missing=1
+            fi
+          done < "${RUNNER_TEMP}/built-names"
+          if [ "${missing}" -ne 0 ]; then
+            echo "::error::The release is missing packages this run built; refusing to publish checksums"
+            exit 1
+          fi
+
       # A maintainer signs SHA256SUMS out of band and attaches the detached
       # signature as SHA256SUMS.asc; apt publication then refuses to publish any
       # package whose hash is not in the signed file, so a compromised CI run
-      # alone cannot ship a package. The listing is sorted byte-wise so
-      # regenerating it reproduces exactly what was signed.
+      # alone cannot ship a package. The listing is generated in byte-wise name
+      # order so regenerating it reproduces exactly what was signed.
       - name: Generate SHA256SUMS
         run: |
           set -euo pipefail
           cd assets
-          find . -maxdepth 1 -type f -printf '%P\n' | LC_ALL=C sort > "${RUNNER_TEMP}/release-assets"
-          if [ ! -s "${RUNNER_TEMP}/release-assets" ]; then
-            echo "::error::No release assets found to checksum"
-            exit 1
-          fi
-          xargs -a "${RUNNER_TEMP}/release-assets" -d '\n' sha256sum > SHA256SUMS
+          xargs -a "${RUNNER_TEMP}/built-names" -d '\n' sha256sum > SHA256SUMS
           cat SHA256SUMS
 
       # Hashing whatever the release currently serves would attest availability,
       # not origin: anyone able to swap an asset between the packaging job and
-      # this one would simply get the replacement blessed. Every published hash
-      # therefore has to have been produced by a packaging job.
+      # this one would simply get the replacement blessed. The comparison is on
+      # whole (hash, name) pairs and runs in both directions, because each half
+      # alone leaves a way through: matching hashes only would let bytes built
+      # for one target be published under another target's name, and matching
+      # published assets only would let a package be dropped from the release
+      # entirely without anything noticing.
       - name: Verify published assets against the built packages
         run: |
           set -euo pipefail
-          cat built-checksums/*/built.sha256 > "${RUNNER_TEMP}/built-all.sha256"
-          if [ ! -s "${RUNNER_TEMP}/built-all.sha256" ]; then
-            echo "::error::No built package checksums were recorded"
-            exit 1
+          LC_ALL=C sort assets/SHA256SUMS > "${RUNNER_TEMP}/published.sha256"
+          if diff -u "${RUNNER_TEMP}/built.sha256" "${RUNNER_TEMP}/published.sha256" \
+            --label built --label published; then
+            echo "All published packages match the packages built in this run"
+            exit 0
           fi
-          cut -d' ' -f1 "${RUNNER_TEMP}/built-all.sha256" | LC_ALL=C sort -u > "${RUNNER_TEMP}/built-hashes"
-
-          unmatched=0
-          while read -r hash name; do
-            if ! grep -qxF "${hash}" "${RUNNER_TEMP}/built-hashes"; then
-              echo "::error::${name} (${hash}) was not produced by any packaging job"
-              unmatched=1
-            fi
-          done < assets/SHA256SUMS
-          if [ "${unmatched}" -ne 0 ]; then
-            echo "::error::Published assets do not match the built packages; refusing to publish checksums"
-            exit 1
-          fi
-          echo "All published assets match packages built in this run"
+          echo "::error::Published packages do not match the packages built in this run; refusing to publish checksums"
+          exit 1
 
       # Uploaded with the preinstalled CLI rather than a third-party action:
       # code running in this job runs beside the signing material, and an action
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -224,13 +224,8 @@ jobs:
           echo "NPM suffix: ${{ steps.suffix.outputs.result }}"
           echo "NPM release tag: ${{ steps.tag.outputs.result }}"
 
-      # npm rejects a provenance attestation unless the package declares a public
-      # `repository` matching, case-sensitively, the repository the build runs
-      # in, so the flag is applied per workspace rather than to the whole set: a
-      # manifest that does not qualify publishes exactly as it does today, with a
-      # warning naming the file to fix, instead of failing the release.
-      # Attested packages go first: publishing cannot be undone, so the batch
-      # carrying the newer machinery is the one to fail early.
+      # Every publishable workspace declares this repository, which is what npm
+      # requires before it will accept a provenance attestation.
       - name: Publish NPM packages
         env:
           NPM_RELEASE_TAG: ${{ steps.tag.outputs.result }}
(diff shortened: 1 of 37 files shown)


thepastaclaw:
⛔ Blockers found — Phase 2 deferred (commit 810b852b) · triage: critical

.github/workflows/release.yml
```diff
@@ -309,8 +579,151 @@ jobs:
...
+            if ! grep -qxF "${hash}" "${RUNNER_TEMP}/built-hashes"; then
+              echo "::error::${name} (${hash}) was not produced by any packaging job"
+              unmatched=1
+            fi
+          done < assets/SHA256SUMS
+          if [ "${unmatched}" -ne 0 ]; then
+            echo "::error::Published assets do not match the built packages; refusing to publish checksums"
+            exit 1
+          fi
+          echo "All published assets match packages built in this run"
```
thepastaclaw, line 720: **🔴 Blocking: Bind published checksums to filenames and require completeness**

The verification reduces the build records to a set of hashes and only checks that each published hash appears in that set. Removing a built asset therefore still passes, and bytes built for one architecture can be published under another architecture's filename because that hash was produced by some matrix leg. This can produce and later sign an incomplete or mislabeled SHA256SUMS. Normalize the recorded paths to release basenames and compare the complete `(hash, filename)` sets in both directions.

```suggestion
          cat built-checksums/*/built.sha256 > "${RUNNER_TEMP}/built-all.sha256"
          if [ ! -s "${RUNNER_TEMP}/built-all.sha256" ]; then
            echo "::error::No built package checksums were recorded"
            exit 1
          fi

          while read -r hash name; do
            printf '%s  %s\n' "${hash}" "${name##*/}"
          done < "${RUNNER_TEMP}/built-all.sha256" \
            | LC_ALL=C sort > "${RUNNER_TEMP}/built-assets.sha256"
          LC_ALL=C sort assets/SHA256SUMS > "${RUNNER_TEMP}/published-assets.sha256"

          if ! diff -u "${RUNNER_TEMP}/built-assets.sha256" "${RUNNER_TEMP}/published-assets.sha256"; then
            echo "::error::Published assets do not exactly match the packages built in this run"
            exit 1
          fi
          echo "All published assets match packages built in this run"
```

<sub>source: ['codex']</sub>

.github/workflows/release.yml
```diff
@@ -221,13 +333,141 @@ jobs:
...
+          gh api --paginate --slurp "repos/${GITHUB_REPOSITORY}/releases?per_page=100" \
+            | jq -r '
+                add
+                | sort_by(.created_at)
+                | reverse
+                | .[]
+                | select(.draft == false)
+                | { tag: .tag_name, asset: ([.assets[].name | select(endswith("_amd64.deb"))] | first) }
+                | select(.asset != null)
+                | "\(.tag)\t\(.asset)"' > "${candidates}"
```
thepastaclaw, line 403: **🔴 Blocking: Order Debian baselines by publication time**

The candidate list is intended to represent releases in the order operators could install them, but it sorts by `created_at`. A release can be created as a draft and published after another release created later. On a rerun, an older-created current release can then lose to its predecessor even though the current release already has a Debian asset, allowing a same-version rebuild to pass without incrementing the Debian revision. Sort by `published_at` so the gate compares against the most recently exposed package.

```suggestion
                | sort_by(.published_at)
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The workflow improves privilege separation and release verification, but three blocking gaps remain. The checksum gate does not bind hashes to filenames or require all built artifacts to be present, the Debian baseline can be selected using release creation order instead of publication order, and mutable actions inside local composites undermine the PR's action-pinning guarantee.

### Review provenance

🔴 3 blocking

<details>

```

In `.github/workflows/release.yml`:
- [BLOCKING] .github/workflows/release.yml:699-720: Bind published checksums to filenames and require completeness
  The verification reduces the build records to a set of hashes and only checks that each published hash appears in that set. Removing a built asset therefore still passes, and bytes built for one architecture can be published under another architecture's filename because that hash was produced by some matrix leg. This can produce and later sign an incomplete or mislabeled SHA256SUMS. Normalize the recorded paths to release basenames and compare the complete `(hash, filename)` sets in both directions.
- [BLOCKING] .github/workflows/release.yml:394-403: Order Debian baselines by publication time
  The candidate list is intended to represent releases in the order operators could install them, but it sorts by `created_at`. A release can be created as a draft and published after another release created later. On a rerun, an older-created current release can then lose to its predecessor even though the current release already has a Debian asset, allowing a same-version rebuild to pass without incrementing the Debian revision. Sort by `published_at` so the gate compares against the most recently exposed package.
- [BLOCKING] .github/workflows/release.yml:58-72: Pin third-party actions used by local composites
  The PR explicitly claims that third-party actions are pinned, but the local Rust and sccache composites invoked after DockerHub login still resolve `dtolnay/rust-toolchain@master` and `mozilla-actions/sccache-action@v0.0.6`. Those mutable external actions execute in the credential-bearing build job and can modify the workspace before `yarn build`; the resulting artifact is then packaged, hashed, and published by the new release path. Pin the external `uses` references in `.github/actions/rust/action.yaml:42` and `.github/actions/sccache/action.yaml:80` to full commit SHAs.
```
</details>

.github/workflows/release.yml
```diff
@@ -18,33 +18,45 @@ concurrency:
...
-    name: Release NPM packages
+    name: Build NPM packages
     runs-on: ubuntu-24.04
     timeout-minutes: 60
     if: github.event_name != 'workflow_dispatch'
+    permissions:
+      contents: read
     steps:
       - name: Check out repo
         uses: actions/checkout@v4
```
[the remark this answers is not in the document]
thepastaclaw: The three cited checkouts persist only a `contents: read` token in this public repository, no `.git` data is uploaded, and no later step needs authenticated Git. The separate write-scoped packaging checkout predates this PR and is not covered by the comment.

packages/dashmate/docs/installation.md
```diff
@@ -20,22 +20,26 @@ Installing the Linux, MacOS, or Windows packages from the [GitHub releases page]
...
+  | xargs curl -fLO
 ```
 
 Install dashmate using apt:
 
 ```bash
 sudo apt update
-sudo apt install ./dashmate_1.8.0.e4e156c86-1_amd64.deb
+sudo apt install ./dashmate_*.deb
 ```
```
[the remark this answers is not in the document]
thepastaclaw: The wildcard can match stale packages, but these instructions were introduced by parent commit ef65cf03b3049d6518a5a7c4d760c9574b7d9e5c and are unchanged by PR #4282, so the issue is outside this PR's scope.

scripts/pack_dashmate.sh
```diff
@@ -39,6 +39,136 @@ FULL_PATH=$(realpath "$0")
...
+    bzip2 -c Packages > Packages.bz2
+    xz -c Packages > Packages.xz
+    apt-ftparchive -c "$FTPARCHIVE_CONF" release . > Release
+
+    # The signatures oclif made cover the metadata from before the rewrite.
+    if [ -n "$DASHMATE_DEB_KEY" ]
+    then
+      gpg --digest-algo SHA512 --clearsign -u "$DASHMATE_DEB_KEY" -o InRelease Release
+      gpg --digest-algo SHA512 -abs -u "$DASHMATE_DEB_KEY" -o Release.gpg Release
+    fi
```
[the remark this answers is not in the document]
thepastaclaw: This script uses `set -e`, not `set -u`; an unset `DASHMATE_DEB_KEY` expands to an empty string and the condition safely evaluates false. The cited block also predates this PR.

.github/workflows/release.yml
shumkov: Fixed in `aa33949`, and the defect was worse than described: the record step wrote build-tree paths (`deb/dashmate_..._amd64.deb`) while GitHub serves flat basenames, so name matching never ran at all — only hashes were ever compared.

Records are now keyed by release basename and the complete `(hash, name)` sets are compared in both directions, so a built package that was not published and a published package that was not built both fail. Added a duplicate-basename guard, since two build paths sharing a basename would silently clobber on upload.

Verified with an old-vs-new harness: removing a built package and serving amd64 bytes under an arm64 filename both passed the old gate and now fail, naming the offending pair.

One related change worth flagging: the job is now scoped to the packages this workflow builds. Upstream #4228 attaches Kotlin and Swift SDK assets to the same release from jobs with no dependency edge to this one, so the previous "reject any asset without a build record" rule would have failed every release and raced on content.

.github/workflows/release.yml
shumkov: Fixed in `aa33949`. Sorting moved to `published_at`, with drafts filtered before the sort since they have no publication date.

Since this is the same class of defect the gate itself exists to catch, it did not stay a one-word change in untestable YAML: the baseline selection moved into `scripts/deb_release_baseline.js` with unit tests, which also removed ~50 lines of embedded jq from the workflow. Flipping the module back to `created_at` fails the rerun test — it picks `dashmate_4.1.0-1_amd64.deb` instead of `dashmate_4.1.1-1_amd64.deb`.

.github/workflows/release.yml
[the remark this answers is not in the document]
shumkov: Applied in `c14ba26`. `persist-credentials: false` added to all four checkouts in this workflow — none of the jobs uses authenticated git after checking out, and the packaging jobs hold signing material, so leaving the token available to every later step in the job was worth removing.

packages/dashmate/docs/installation.md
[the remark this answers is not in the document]
shumkov: Applied in `030a4f4`. The snippet now captures the selected asset and installs that exact file rather than globbing:

```bash
DASHMATE_DEB="$(curl -fsSL .../releases/latest | grep -o "https://[^\"]*_$(dpkg --print-architecture)\.deb" | head -n 1)"
curl -fLO "$DASHMATE_DEB"
sudo apt install "./$(basename "$DASHMATE_DEB")"
```

Worth fixing rather than a nitpick: this PR changes the package filenames, so an operator upgrading is unusually likely to have the previous release still sitting in the same directory.

scripts/pack_dashmate.sh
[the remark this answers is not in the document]
shumkov: Applied in `030a4f4`. To be accurate about the reasoning: the script uses `set -e` only, not `set -u`, so this is not a live break today. `${DASHMATE_DEB_KEY:-}` costs nothing and stops a later `set -u` from silently failing every unsigned local build, so it went in as cheap insurance rather than as a fix.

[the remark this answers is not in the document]
shumkov: Addressed in `c14ba26` — see the reply on the original thread for this finding.

packages/dashmate/docs/installation.md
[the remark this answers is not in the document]
shumkov: Addressed in `030a4f4` — see the reply on the original thread for this finding.

.github/workflows/release.yml
```diff
@@ -128,38 +223,59 @@ jobs:
...
-      - name: Upload the archive of built files
-        uses: actions/upload-artifact@v4
-        with:
-          name: js-build-${{ github.sha }}
-          path: ${{ steps.diff.outputs.files }}
-          retention-days: 1
-          if-no-files-found: error
-          include-hidden-files: true
-        if: ${{ steps.check-artifact.outputs.exists != 'true' }}
+          set -euo pipefail
```
shumkov: Why so difficult?

.github/workflows/release.yml
```diff
@@ -18,33 +18,48 @@ concurrency:
...
 
+# Debian version knobs, read by packages/dashmate/scripts/deb-version.mjs in both
+# the version gate and the packaging job. They have to stay in lockstep: a value
+# seen by only one of them makes the gate validate a version that is not the one
+# that ships.
+# Bump the revision to rebuild an already published version; set the epoch only
+# to outrank a version whose upstream part carried a git sha.
+env:
+  DASHMATE_DEB_REVISION: "1"
+  DASHMATE_DEB_EPOCH: ""
```
thepastaclaw, line 35: **🔴 Blocking: Expose the Debian epoch and revision knobs to release runs**

The workflow fixes `DASHMATE_DEB_REVISION` to `1` and `DASHMATE_DEB_EPOCH` to empty, and no job overrides them. The last 4.2 package was produced under the old SHA-based scheme as `4.2.0.729847495c-1`; Debian orders `4.2.0~dev.9-1`, `4.2.0~rc.1-1`, and `4.2.0-1` below that version. Consequently, the next release on the 4.2 line will fail the new gate even though the gate's error directs the operator to set `DASHMATE_DEB_EPOCH=1`. A rerun of an already-published release has the same problem with `DASHMATE_DEB_REVISION`: the baseline selector intentionally compares against that release's existing package, but the workflow cannot increment the revision. Read these values from repository variables so release events and reruns can use the documented escape hatches without requiring a preparatory code change.

```suggestion
env:
  DASHMATE_DEB_REVISION: ${{ vars.DASHMATE_DEB_REVISION || '1' }}
  DASHMATE_DEB_EPOCH: ${{ vars.DASHMATE_DEB_EPOCH || '' }}
```

<sub>source: ['codex']</sub>

packages/dashmate/package.json
```diff
@@ -16,7 +16,7 @@
   },
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/dashevo/dashmate.git"
+    "url": "https://github.com/dashpay/platform"
   },
```
thepastaclaw, line 20: **🟡 Suggestion: Identify Dashmate's directory in the monorepo metadata**

This PR changes Dashmate's repository from its retired standalone repository to the platform monorepo, but unlike the other migrated publishable manifests it does not identify the package's subdirectory. The repository URL is sufficient for npm provenance, but omitting `repository.directory` leaves source-linking and package-index tooling pointing only at the monorepo root.

```suggestion
"repository": {
  "type": "git",
  "url": "https://github.com/dashpay/platform",
  "directory": "packages/dashmate"
},
```

<sub>source: ['codex']</sub>

packages/dashmate/scripts/check-release-deb-version.sh
```diff
@@ -0,0 +1,110 @@
...
+baseline_asset="${baseline#*$'\t'}"
+
+# Read the baseline from the deb's own control field: the only version
+# apt looks at, the only one a server-side rename of the asset cannot
+# alter, and what operators actually installed rather than what the tag
+# would produce today.
+mkdir -p baseline-deb
+gh release download "${baseline_tag}" --repo "${GITHUB_REPOSITORY}" \
+  --pattern "${baseline_asset}" --dir baseline-deb --clobber
+baseline_version="$(dpkg-deb -f "baseline-deb/${baseline_asset}" Version)"
```
thepastaclaw, line 98: **🟡 Suggestion: Download the baseline package into temporary storage**

The baseline directory is relative to the caller's current directory and is never removed. The PR explicitly documents invoking this script manually from the repository, so each invocation leaves a large downloaded package and an untracked `baseline-deb` directory in the working tree. Use runner or system temporary storage and remove it when the script exits.

```suggestion
baseline_dir="$(mktemp -d "${RUNNER_TEMP:-${TMPDIR:-/tmp}}/dashmate-baseline.XXXXXX")"
trap 'rm -rf "${baseline_dir}"' EXIT
gh release download "${baseline_tag}" --repo "${GITHUB_REPOSITORY}" \
  --pattern "${baseline_asset}" --dir "${baseline_dir}" --clobber
baseline_version="$(dpkg-deb -f "${baseline_dir}/${baseline_asset}" Version)"
```

<sub>source: ['codex']</sub>

packages/dashmate/scripts/deb-version.mjs
```diff
@@ -0,0 +1,129 @@
...
+  if (!version) {
+    console.error('Usage: deb-version.mjs [--file-name] SEMVER_VERSION\n\n'
+      + '  Prints the Debian package version for a semver version, or the version as it\n'
+      + '  appears in the package file name.\n\n'
+      + '  DASHMATE_DEB_REVISION  Debian revision, default 1. Bump it to rebuild a version\n'
+      + '                         that was already published.\n'
+      + '  DASHMATE_DEB_EPOCH     Debian epoch, unset by default.\n');
+
+    process.exit(1);
+  }
```
thepastaclaw, line 115: **🟡 Suggestion: Reject unexpected Debian version CLI arguments**

The command-line entry point reads only the first one or two arguments and silently ignores everything else. For example, `deb-version.mjs 4.1.0 1 2` succeeds with `4.1.0-1`, while placing `--file-name` after the version silently returns the control version instead of the requested filename version. Since this is also an operator-facing release utility, reject invalid arity instead of producing output for a mistyped command.

```suggestion
const args = process.argv.slice(2);
const forFileName = args[0] === '--file-name';
const version = forFileName ? args[1] : args[0];
const expectedArgumentCount = forFileName ? 2 : 1;

if (!version || args.length !== expectedArgumentCount) {
  console.error('Usage: deb-version.mjs [--file-name] SEMVER_VERSION\n\n'
    + '  Prints the Debian package version for a semver version, or the version as it\n'
    + '  appears in the package file name.\n\n'
    + '  DASHMATE_DEB_REVISION  Debian revision, default 1. Bump it to rebuild a version\n'
    + '                         that was already published.\n'
    + '  DASHMATE_DEB_EPOCH     Debian epoch, unset by default.\n');

  process.exit(1);
}
```

<sub>source: ['codex']</sub>

packages/dashmate/docs/installation.md
```diff
@@ -20,22 +20,26 @@ Installing the Linux, MacOS, or Windows packages from the [GitHub releases page]
...
+Download the newest dashmate installation package for your architecture from the [GitHub releases page](https://github.com/dashpay/platform/releases/latest).
+The file name contains the version, so it changes with every release; this downloads the one matching the architecture you are on:
 
 ```bash
-wget https://github.com/dashpay/platform/releases/download/v1.8.0/dashmate_1.8.0.e4e156c86-1_amd64.deb
+DASHMATE_DEB="$(curl -fsSL https://api.github.com/repos/dashpay/platform/releases/latest \
+  | grep -o "https://[^\"]*_$(dpkg --print-architecture)\.deb" \
+  | head -n 1)"
+curl -fLO "$DASHMATE_DEB"
 ```
```
thepastaclaw, line 31: **🟡 Suggestion: Fail clearly when no Debian release asset is found**

The assignment pipeline reports the status of `head`, so an API failure, rate limit, or missing architecture-specific asset can still leave `DASHMATE_DEB` empty while the command succeeds. The next line then fails with the unrelated message `curl: no URL specified`, and the later install command resolves to `./`. Check the selected URL before downloading so the documented procedure reports the actual problem.

```suggestion
DASHMATE_DEB="$(curl -fsSL https://api.github.com/repos/dashpay/platform/releases/latest \
  | grep -o "https://[^\"]*_$(dpkg --print-architecture)\.deb" \
  | head -n 1)"
if [ -z "${DASHMATE_DEB}" ]; then
  echo "Could not find a dashmate .deb for this architecture; try again later." >&2
  exit 1
fi
curl -fLO "$DASHMATE_DEB"
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):
## Preliminary review — Phase 1 blocker gate

The Debian version translation, package-byte validation, and checksum pipeline are generally coherent and well tested. However, the workflow hard-codes the revision and epoch knobs, so any subsequent 4.2 release cannot outrank the last old-scheme package and rerunning an existing release cannot increment its revision without first changing the workflow. The remaining findings are minor metadata, cleanup, CLI-validation, and installation-documentation issues.

Source: reviewer 1: `glm-5.3-flash` (agent: `phase1-reviewer`, role: `general`); reviewer 2: `glm-5.3-flash` (agent: `phase1-reviewer`, role: `security-auditor`); final verifier: `gpt-5.6-sol` (agent: `sol-verifier`, role: `verifier`)

Validated blockers were found by the Phase-1 review and confirmed by a fresh verifier. Phase 2 is deferred until a fresh same-head revalidation clears the blocker gate.

### Review provenance
- Triage: `critical` by `gpt-6-astra` (effort low) — This broad release-pipeline change affects Debian upgrade ordering, package integrity verification, artifact provenance, and credential isolation, where defects could block security updates or compromise published releases.
- Phase 1 reviewers: `glm-5.3-flash` — general (completed, effort max); agent `phase1-reviewer`, `glm-5.3-flash` — security-auditor (completed, effort max); agent `phase1-reviewer`
- Fresh verifier: `gpt-5.6-sol` — verifier; agent `sol-verifier`
- Phase 2 reviewers: **not run (deferred by blocker gate)**

<details>

```

In `.github/workflows/release.yml`:
- [BLOCKING] .github/workflows/release.yml:33-35: Expose the Debian epoch and revision knobs to release runs
  The workflow fixes `DASHMATE_DEB_REVISION` to `1` and `DASHMATE_DEB_EPOCH` to empty, and no job overrides them. The last 4.2 package was produced under the old SHA-based scheme as `4.2.0.729847495c-1`; Debian orders `4.2.0~dev.9-1`, `4.2.0~rc.1-1`, and `4.2.0-1` below that version. Consequently, the next release on the 4.2 line will fail the new gate even though the gate's error directs the operator to set `DASHMATE_DEB_EPOCH=1`. A rerun of an already-published release has the same problem with `DASHMATE_DEB_REVISION`: the baseline selector intentionally compares against that release's existing package, but the workflow cannot increment the revision. Read these values from repository variables so release events and reruns can use the documented escape hatches without requiring a preparatory code change.

In `packages/dashmate/package.json`:
- [SUGGESTION] packages/dashmate/package.json:17-20: Identify Dashmate's directory in the monorepo metadata
  This PR changes Dashmate's repository from its retired standalone repository to the platform monorepo, but unlike the other migrated publishable manifests it does not identify the package's subdirectory. The repository URL is sufficient for npm provenance, but omitting `repository.directory` leaves source-linking and package-index tooling pointing only at the monorepo root.

In `packages/dashmate/scripts/check-release-deb-version.sh`:
- [SUGGESTION] packages/dashmate/scripts/check-release-deb-version.sh:95-98: Download the baseline package into temporary storage
  The baseline directory is relative to the caller's current directory and is never removed. The PR explicitly documents invoking this script manually from the repository, so each invocation leaves a large downloaded package and an untracked `baseline-deb` directory in the working tree. Use runner or system temporary storage and remove it when the script exits.

In `packages/dashmate/scripts/deb-version.mjs`:
- [SUGGESTION] packages/dashmate/scripts/deb-version.mjs:101-115: Reject unexpected Debian version CLI arguments
  The command-line entry point reads only the first one or two arguments and silently ignores everything else. For example, `deb-version.mjs 4.1.0 1 2` succeeds with `4.1.0-1`, while placing `--file-name` after the version silently returns the control version instead of the requested filename version. Since this is also an operator-facing release utility, reject invalid arity instead of producing output for a mistyped command.

In `packages/dashmate/docs/installation.md`:
- [SUGGESTION] packages/dashmate/docs/installation.md:26-31: Fail clearly when no Debian release asset is found
  The assignment pipeline reports the status of `head`, so an API failure, rate limit, or missing architecture-specific asset can still leave `DASHMATE_DEB` empty while the command succeeds. The next line then fails with the unrelated message `curl: no URL specified`, and the later install command resolves to `./`. Check the selected URL before downloading so the documented procedure reports the actual problem.
```
</details>

<details>
<summary>Out-of-scope follow-up suggestions (3)</summary>

These are valid observations, but they are outside this PR's scope and should be handled in separate issues or author/maintainer-requested PRs rather than blocking this review.

- **Remove the unused DockerHub login from the npm build job** — The npm build job logs into DockerHub before running repository build and dependency code, although the job never builds or pushes a Docker image. This unnecessarily exposes a DockerHub credential to that code, but the login predates this PR and was only SHA-pinned here rather than introduced or expanded.
  - Follow-up: Open a separate security-hardening issue to remove the DockerHub login from `release-npm` or move it into the reusable image-publishing workflows that need it.
- **Rewritten apt Release file preserves oclif's 'Origin: undefined' wart** — Out of scope and low impact: oclif already emitted this tolerated metadata before the PR, and correcting the repository-origin label is unrelated to the version-ordering and release-verification changes.
  - Follow-up: Consider creating a separate issue or author/maintainer-requested PR for this.
- **Installation docs download the deb without offering the new integrity verification** — Out of scope as a separate documentation enhancement: the PR publishes deterministic checksums and verifies assets against build records, while documenting end-user signature verification depends on the future out-of-band signing flow described by the workflow.
  - Follow-up: Consider creating a separate issue or author/maintainer-requested PR for this.

</details>

PR #4283, opened 2026-08-04
fix(dashmate): keep the node up when an image pull fails, and stop reporting success when it did not

shumkov:

Three independent defects found while working on dashmate's update path. Each is reachable today and none depends on the others.

**A failed image pull can leave a masternode down.** `restart` stops every service and only then lets compose pull whatever is missing, so a pull that fails in that window leaves the node stopped. For a masternode that is missed blocks and PoSe-penalty exposure.

**`dashmate update` reports success when it updated nothing.** Docker reports pull failures — rate limiting, a full disk — as in-band error objects on an otherwise successful stream, and the old parser looked only for status lines. Failures surfaced as a coloured word in a table, with the reason printed only under `DEBUG`, and an exit code of 0. Automation could not distinguish "updated everything" from "updated nothing".

**The GitHub release lookup throws on almost any network failure.** It caught node-fetch's error names, but dashmate runs on Node's native fetch, which rejects with `TypeError('fetch failed')` and a `TimeoutError` DOMException. Neither matched, so its `null` branch was unreachable. It also stripped one character from `tag_name` unconditionally, mangling any tag without a `v` prefix.

Plus two smaller ones: the gateway's TLS private key was written world-readable, and `dashmate config default` with no argument threw instead of printing the default, because oclif never applies an args `default: null`.

## What was done?

Required images are confirmed present **before** anything is stopped. Pull-stream parsing moves to docker-modem's `followProgress`, which buffers across chunk boundaries and splits on the separator Docker actually emits — the previous parser split each chunk on CRLF and parsed every fragment, so an ordinary TCP boundary threw inside the stream handler where nothing could catch it. `group restart` gains the same pre-stop guarantee across every node.

The release lookup now returns unknown instead of throwing, enforces its declared response-size cap, and requires a valid semver before returning a version — which keeps package-manager specifiers and terminal control sequences out of a value callers may act on.

The TLS key is written `0600` by both writers, the two historical migrations that copy it no longer carry the old mode forward, permissions are restricted when starting the node so existing installations are actually fixed rather than only new ones, and `doctor` reports a key other users can read.

## How Has This Been Tested?

Failing-test-first throughout; each fix has a test shown red against the unfixed code. The suites were then mutation-tested — deliberately weakened implementations were written to confirm the tests catch them, which is how several of these defects were found in the first place. 257 unit tests pass.

Not executed: no live gateway or masternode was exercised (no Docker daemon available), so container-readability of the tightened key rests on the compose wiring — `ENVOY_UID` is the operator's own uid — rather than a running container.

## Breaking Changes

None.

## Checklist:

### DIFF
--- a/packages/dashmate/src/commands/group/restart.js
+++ b/packages/dashmate/src/commands/group/restart.js
@@ -1,6 +1,7 @@
 import { Listr } from 'listr2';
 import GroupBaseCommand from '../../oclif/command/GroupBaseCommand.js';
 import MuteOneLineError from '../../oclif/errors/MuteOneLineError.js';
+import isServiceBuildRequired from '../../util/isServiceBuildRequired.js';
 
 export default class GroupRestartCommand extends GroupBaseCommand {
   static description = 'Restart group nodes';
@@ -20,6 +21,7 @@ export default class GroupRestartCommand extends GroupBaseCommand {
    * @param {DockerCompose} dockerCompose
    * @param {stopNodeTask} stopNodeTask
    * @param {startGroupNodesTask} startGroupNodesTask
+   * @param {buildServicesTask} buildServicesTask
    * @param {Config[]} configGroup
    * @return {Promise<void>}
    */
@@ -32,15 +34,34 @@ export default class GroupRestartCommand extends GroupBaseCommand {
     dockerCompose,
     stopNodeTask,
     startGroupNodesTask,
+    buildServicesTask,
     configGroup,
   ) {
     const groupName = configGroup[0].get('group');
+
+    // The whole group shares one set of locally built images, so one config
+    // describes the build for all of them
+    const buildConfig = configGroup.find(isServiceBuildRequired);
 
     const tasks = new Listr(
       {
         title: `Restart ${groupName} nodes`,
         task: async () => (
           new Listr([
+            {
+              // An image built from local sources is in no registry, so the
+              // pull below cannot confirm it. Building before the group is
+              // stopped is what keeps a build failure from leaving it down
+              enabled: () => Boolean(buildConfig),
+              title: 'Build services',
+              task: (ctx) => {
+                // The group start builds the same images, and would otherwise
+                // repeat the whole build on every restart
+                ctx.skipBuildServices = true;
+
+                return buildServicesTask(buildConfig);
+              },
+            },
             {
               // Every node's images must be fetched before the first node is
               // stopped, otherwise a failed pull leaves the group stopped
--- a/packages/dashmate/src/commands/group/restart.js
+++ b/packages/dashmate/src/commands/group/restart.js
@@ -42,6 +42,21 @@ export default class GroupRestartCommand extends GroupBaseCommand {
         task: async () => (
           new Listr([
             {
+              // Every node's images must be fetched before the first node is
+              // stopped, otherwise a failed pull leaves the group stopped
+              title: 'Pull missing images',
+              task: () => (
+                new Listr(configGroup.map((config) => ({
+                  task: (ctx, task) => dockerCompose.pullMissingImages(config, {
+                    onProgress: (message) => {
+                      // eslint-disable-next-line no-param-reassign
+                      task.output = message;
+                    },
+                  }),
+                })))
+              ),
+            },
+            {
               title: 'Stop nodes',
               task: () => (
                 // So we stop the miner first, as there's a chance that MNs will get banned
--- a/packages/dashmate/src/commands/update.js
+++ b/packages/dashmate/src/commands/update.js
@@ -118,7 +118,7 @@ export default class UpdateCommand extends ConfigBaseCommand {
 
       if (!result.ok) {
         // Nothing was fetched at all - not a per-image failure, which resolves
-        // as an error row and has always exited 0. Retained so it can be
+        // as an error row and is reported further down. Retained so it can be
         // raised once the certificate has had its say: returning quietly here
         // hands `update && start` a node whose images were never downloaded,
         // with no exit code for the caller to catch.
@@ -137,16 +137,29 @@ export default class UpdateCommand extends ConfigBaseCommand {
       const colors = {
         updated: chalk.yellow,
         'up to date': chalk.green,
+        'built locally': chalk.gray,
         error: chalk.red,
       };
 
       printArrayOfObjects(result.info.map(({
-        name, title, updated, image,
+        name, title, updated, image, error,
       }) => (format === OUTPUT_FORMATS.PLAIN
         ? { Service: title, Image: image, Updated: colors[updated](updated) }
         : {
-          name, title, updated, image,
+          name, title, updated, image, error,
         })), format);
+
+      const failedServices = result.info.filter(({ updated }) => updated === 'error');
+
+      if (failedServices.length > 0) {
+        const reasons = failedServices
+          .map(({ title, image, error }) => `  ${title} (${image}): ${error}`)
+          .join('\n');
+
+        // Reported on stderr so machine-readable output on stdout stays parseable
+        // eslint-disable-next-line no-console
+        console.error(`\nFailed to update ${failedServices.length} of ${result.info.length} images:\n\n${reasons}\n`);
+      }
     };
 
     const tasks = new Listr(
@@ -260,6 +273,9 @@ export default class UpdateCommand extends ConfigBaseCommand {
       throw new MuteOneLineError(unresolved);
     }
 
-    process.exitCode = 0;
+    // An image that failed to download is reported as a row in the table rather
+    // than thrown, so the exit code is the only thing that tells a caller apart
+    // an update that fetched everything from one that fetched some of it.
+    process.exitCode = this.pullResult?.failed > 0 ? 1 : 0;
   }
 }
--- a/packages/dashmate/src/status/providers.js
+++ b/packages/dashmate/src/status/providers.js
@@ -1,33 +1,188 @@
 import https from 'https';
+import semver from 'semver';
 
 const MAX_REQUEST_TIMEOUT = 5000;
 const MAX_RESPONSE_SIZE = 1 * 1024 * 1024; // 1 MB
 
-const request = async (url) => {
+// A remote version string is printed to the operator's terminal, included in JSON
+// output and passed to package managers, so only a strict semver shape is accepted.
+// The anchors and character classes leave no room for control or ANSI escape
+// characters, nor for the specifiers a package manager would treat as a location to
+// install from (git+https://…, file:…, https://….tgz, npm: aliases).
+const VERSION_REGEX = /^\d+\.\d+\.\d+(-[0-9A-Za-z.-]+)?(\+[0-9A-Za-z.-]+)?$/;
+
+const MAX_LOG_LENGTH = 200;
+
+// Characters that let remote text hijack a terminal line or disguise itself in the
+// output: C0 and C1 controls (including the escape that opens an ANSI sequence), zero
+// width characters, line separators, and the bidirectional overrides and isolates that
+// reorder what an operator reads. Matching them is the point, hence the disabled rule.
+// eslint-disable-next-line no-control-regex
+const UNSAFE_LOG_CHARACTERS_REGEX = /[\u0000-\u001f\u007f-\u009f\u200b-\u200f\u2028-\u2029\u202a-\u202e\u2066-\u2069\ufeff]/g;
+
+/**
+ * Make text received from a remote host safe to print
+ *
+ * Error messages quote the payload that failed to parse, so remote bytes reach the
+ * terminal through diagnostics even when the value itself is rejected. Both the
+ * dangerous characters and the length are bounded here.
+ *
+ * @param {*} text
+ * @returns {string}
+ */
+const sanitizeForLog = (text) => (typeof text === 'string'
+  ? text.replace(UNSAFE_LOG_CHARACTERS_REGEX, '').slice(0, MAX_LOG_LENGTH)
+  : '[unprintable]');
+
+const request = async (url, options = {}) => {
   try {
     return await fetch(url, {
+      ...options,
       signal: AbortSignal.timeout(MAX_REQUEST_TIMEOUT),
     });
   } catch (e) {
-    if (e.name === 'FetchError' || e.name === 'AbortError') {
-      if (process.env.DEBUG) {
-        // eslint-disable-next-line no-console
-        console.warn(`Could not fetch: ${e}`);
-      }
-      return null;
+    // Every transport failure (DNS, connection reset, timeout, abort) is reported as
+    // an unknown result. Callers use these providers to enrich output, so an
+    // unreachable remote host must never fail the command that called them.
+    if (process.env.DEBUG) {
+      // eslint-disable-next-line no-console
+      console.warn(`Could not fetch ${url}: ${e.name}: ${sanitizeForLog(e.message)}`);
     }
-    throw e;
+
+    return null;
   }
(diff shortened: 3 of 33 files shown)


thepastaclaw:
⛔ Blockers found — Phase 2 deferred (commit ba2ec49e)
_Last checked: 2026-09-06 19:11 UTC_

packages/dashmate/src/commands/group/restart.js
```diff
@@ -41,6 +41,21 @@ export default class GroupRestartCommand extends GroupBaseCommand {
...
+              title: 'Pull missing images',
+              task: () => (
+                new Listr(configGroup.map((config) => ({
+                  task: (ctx, task) => dockerCompose.pullMissingImages(config, {
+                    onProgress: (message) => {
+                      // eslint-disable-next-line no-param-reassign
+                      task.output = message;
+                    },
+                  }),
+                })))
```
thepastaclaw, line 56: **🔴 Blocking: Build local images before stopping group nodes**

This pre-stop phase only calls `pullMissingImages`, which intentionally excludes services marked `isBuiltLocally`. When local builds are enabled, `startGroupNodesTask` runs `buildServicesTask` as its first task, but that happens only after every node has been stopped. A missing local image or any build failure therefore still leaves the entire group down, despite the commit's stated guarantee that required images are prepared before anything stops. Run the shared local build before the stop phase and arrange for the subsequent group start to skip the duplicate build.

<sub>source: ['codex']</sub>

packages/dashmate/src/listr/tasks/ssl/saveCertificateTask.js
```diff
@@ -31,7 +31,12 @@ export default function saveCertificateTaskFactory(homeDir) {
           fs.writeFileSync(crtFile, ctx.certificateFile, 'utf8');
 
           const keyFile = path.join(certificatesDir, 'private.key');
-          fs.writeFileSync(keyFile, ctx.privateKeyFile, 'utf8');
+
+          // The private key must not be readable by other users on the host.
+          // chmod is required in addition to the write mode because the mode is
+          // applied only when the file is created, and the key is usually rewritten
+          fs.writeFileSync(keyFile, ctx.privateKeyFile, { encoding: 'utf8', mode: 0o600 });
+          fs.chmodSync(keyFile, 0o600);
```
thepastaclaw, line 39: **🔴 Blocking: Replace an exposed private key through a fresh restricted inode**

The `mode` option only affects file creation. If `private.key` already exists with mode 0644, `writeFileSync` writes the replacement secret through that world-readable inode and restricts it only afterward. A local user can also open the old inode before this operation and retain access through that descriptor after `chmodSync`, allowing them to read the newly generated key. Write the key to a temporary file in the same directory created as 0600, then atomically rename it over `private.key`; descriptors for the old inode will then expose only the old key.

<sub>source: ['codex']</sub>

packages/dashmate/src/commands/update.js
```diff
@@ -38,23 +38,38 @@ export default class UpdateCommand extends ConfigBaseCommand {
...
+    const failedServices = updateInfo.filter(({ updated }) => updated === 'error');
+
+    if (failedServices.length > 0) {
+      const reasons = failedServices
+        .map(({ title, image, error }) => `  ${title} (${image}): ${error}`)
+        .join('\n');
+
+      // Report to stderr to keep machine-readable output on stdout intact
+      // eslint-disable-next-line no-console
+      console.error(`\nFailed to update ${failedServices.length} of ${updateInfo.length} images:\n\n${reasons}\n`);
```
thepastaclaw, line 70: **🔴 Blocking: Sanitize registry error text before terminal output**

The `error` field can come directly from a Docker registry pull-stream `errorDetail.message`, but the stderr summary interpolates it unchanged. A registry response can therefore inject newlines, carriage returns, ANSI escapes, bidirectional controls, or an excessively long message into the operator's terminal. This path is unconditional, including when `--format=json` is selected. Strip unsafe control characters and bound the rendered length before writing remote error text to terminal streams, using protections equivalent to the new remote-diagnostic sanitization in `status/providers.js`.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The PR improves pull failure reporting and TLS key permissions, but three in-scope blockers remain. Group restart still performs local builds after stopping every node, replacement of a historically exposed key writes the new secret through the old inode, and registry-controlled error text is printed directly to the terminal.

### Review provenance

🔴 3 blocking

<details>

```

In `packages/dashmate/src/commands/group/restart.js`:
- [BLOCKING] packages/dashmate/src/commands/group/restart.js:47-56: Build local images before stopping group nodes
  This pre-stop phase only calls `pullMissingImages`, which intentionally excludes services marked `isBuiltLocally`. When local builds are enabled, `startGroupNodesTask` runs `buildServicesTask` as its first task, but that happens only after every node has been stopped. A missing local image or any build failure therefore still leaves the entire group down, despite the commit's stated guarantee that required images are prepared before anything stops. Run the shared local build before the stop phase and arrange for the subsequent group start to skip the duplicate build.

In `packages/dashmate/src/listr/tasks/ssl/saveCertificateTask.js`:
- [BLOCKING] packages/dashmate/src/listr/tasks/ssl/saveCertificateTask.js:35-39: Replace an exposed private key through a fresh restricted inode
  The `mode` option only affects file creation. If `private.key` already exists with mode 0644, `writeFileSync` writes the replacement secret through that world-readable inode and restricts it only afterward. A local user can also open the old inode before this operation and retain access through that descriptor after `chmodSync`, allowing them to read the newly generated key. Write the key to a temporary file in the same directory created as 0600, then atomically rename it over `private.key`; descriptors for the old inode will then expose only the old key.

In `packages/dashmate/src/commands/update.js`:
- [BLOCKING] packages/dashmate/src/commands/update.js:64-70: Sanitize registry error text before terminal output
  The `error` field can come directly from a Docker registry pull-stream `errorDetail.message`, but the stderr summary interpolates it unchanged. A registry response can therefore inject newlines, carriage returns, ANSI escapes, bidirectional controls, or an excessively long message into the operator's terminal. This path is unconditional, including when `--format=json` is selected. Strip unsafe control characters and bound the rendered length before writing remote error text to terminal streams, using protections equivalent to the new remote-diagnostic sanitization in `status/providers.js`.
```
</details>

packages/dashmate/src/commands/group/restart.js
shumkov: Fixed in `5b6cf98`. Confirmed: `pullMissingImages` filters `!service.isBuiltLocally`, and `startGroupNodesTaskFactory` ran `buildServicesTask` as its first task — after `Stop nodes`.

The shared local build now runs before the pull in `group/restart.js` and sets `ctx.skipBuildServices`, and `startGroupNodesTaskFactory` gained `enabled: (ctx) => Boolean(platformBuildConfig) && !ctx.skipBuildServices` so the restart does not build twice. That is the same `skipBuildServices` contract `restartNodeTaskFactory` already uses for a single node.

Red→green: with the build moved back after `Stop nodes`, `should build local images before stopping the first node` and `should not stop any node when a local image can not be built` both fail.

packages/dashmate/src/commands/update.js
shumkov: Fixed in `20009a2`, and extended in `68d467e`.

Registry-controlled text now passes through `src/util/sanitizeRemoteText.js` at both intake points — `findPullStreamError` (which also covers `DockerCompose.pullMissingImages`) and `updateNodeFactory` — so the stderr reasons and the `error` field in `--format=json` are both covered.

A follow-up review probe found the first version stripped the sequences that rewrite a terminal but not several that hide text: the Arabic letter mark, zero-width space and joiner, the line and paragraph separators, and the byte order mark all survived into a message an operator reads to decide what to do next. The class now covers those too, pinned by a test per code point.

packages/dashmate/src/listr/tasks/ssl/saveCertificateTask.js
shumkov: Not applied — the blocker is already addressed upstream, and the suggested fix would break certificate renewal.

**Already addressed:** on `v4.2-dev`, `saveCertificateTask.js` chmods an existing key to `0600` *before* the in-place write and restores the chosen mode in a `finally`, so the replacement secret never lands behind a world-readable inode. This branch touches that file 0 lines — the version that motivated the finding was dropped during the rebase in favour of upstream's, which is strictly better (it also preserves an owner's deliberate `0400`, where the dropped version would have forced `0600`).

**Why write-to-temp-then-rename would be a regression:** the gateway bind-mounts `private.key` as an individual file, and a file bind mount follows the **inode**. A rename installs a new inode, so the running gateway keeps reading the old one and goes on serving the expired certificate until it is recreated. Upstream rejected that approach deliberately and has a test named `should install a renewal into the files the gateway already has mounted` asserting the inode is preserved.

**What was added instead** (`1a512d8`, test-only): the existing specs assert only the mode the file is left with, which stays correct whichever order the chmod and write happen in — so the window this finding describes could be silently reopened later without any test noticing. There is now an ordering test that fails (`expected [420] to deeply equal [384]`) if the chmod moves after the write.

packages/dashmate/src/commands/group/restart.js
```diff
@@ -32,15 +34,49 @@ export default class GroupRestartCommand extends GroupBaseCommand {
...
+              enabled: () => Boolean(buildConfig),
+              title: 'Build services',
+              task: (ctx) => {
+                // The group start builds the same images, and would otherwise
+                // repeat the whole build on every restart
+                ctx.skipBuildServices = true;
+
+                return buildServicesTask(buildConfig);
+              },
+            },
```
thepastaclaw, line 64: **🟡 Suggestion: Pre-stop group build trusts a single config to describe every node's local builds**

`configGroup.find(isServiceBuildRequired)` builds from only the first matching node. Group membership is based only on the `group` value, and no invariant requires every member to enable the same locally built services; `generateEnvs` selects the Drive, rs-dapi, and helper build compose files independently for each config. If one member enables a Drive build and another enables only an rs-dapi build, the first config does not build the second member's required image. The pull phase excludes that locally built service, and `ctx.skipBuildServices` suppresses all later per-node builds, so startup fails only after every node has stopped. Build the union of requirements from all group members before setting the shared skip flag.

<sub>source: ['codex']</sub>

packages/dashmate/src/status/providers.js
```diff
@@ -1,33 +1,188 @@
...
+const VERSION_REGEX = /^\d+\.\d+\.\d+(-[0-9A-Za-z.-]+)?(\+[0-9A-Za-z.-]+)?$/;
+
+const MAX_LOG_LENGTH = 200;
+
+// Characters that let remote text hijack a terminal line or disguise itself in the
+// output: C0 and C1 controls (including the escape that opens an ANSI sequence), zero
+// width characters, line separators, and the bidirectional overrides and isolates that
+// reorder what an operator reads. Matching them is the point, hence the disabled rule.
+// eslint-disable-next-line no-control-regex
+const UNSAFE_LOG_CHARACTERS_REGEX = /[\u0000-\u001f\u007f-\u009f\u200b-\u200f\u2028-\u2029\u202a-\u202e\u2066-\u2069\ufeff]/g;
```
thepastaclaw, line 21: **🟡 Suggestion: providers.js sanitizer misses U+061C (Arabic letter mark) that sanitizeRemoteText covers**

The new diagnostic sanitizer removes bidi overrides, isolates, and directional marks, but omits U+061C, an invisible strong right-to-left mark. Remote release-tag text can therefore put U+061C into DEBUG output at line 182 and alter how the diagnostic is displayed in bidi-aware terminals. The sibling `sanitizeRemoteText` utility includes this code point for the same threat model, so the two new sanitizers should provide equivalent coverage.

```suggestion
const UNSAFE_LOG_CHARACTERS_REGEX = /[\u0000-\u001f\u007f-\u009f\u061c\u200b-\u200f\u2028-\u2029\u202a-\u202e\u2066-\u2069\ufeff]/g;
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):
## Preliminary review — GLM Flash blocker gate

The current head fixes the prior group build ordering and registry-error sanitization blockers, but the TLS remediation still writes a replacement key through the previously exposed inode, allowing a descriptor opened before chmod to read the new key. Two additional in-scope suggestions remain: build requirements are derived from only one group member, and the GitHub diagnostic sanitizer omits U+061C. Source: Codex reviewers (exact model identifier unavailable in the supplied evidence), Claude reviewer (exact model identifier unavailable in the supplied evidence), and Claude final verifier (exact model identifier unavailable in the supplied evidence).

Source: reviewer 1: `glm-5.3-flash` (agent: `phase1-reviewer`, role: `general`); reviewer 2: `glm-5.3-flash` (agent: `phase1-reviewer`, role: `security-auditor`); final verifier: `gpt-5.6-sol` (agent: `sol-verifier`, role: `verifier`)

Validated blockers were found by the Phase-1 GLM Flash review and confirmed by a fresh Sol verifier. Phase 2 is deferred until a fresh same-head revalidation clears the blocker gate.

### Review provenance
- Phase 1 reviewers (GLM Flash): `glm-5.3-flash` — general (completed); agent `phase1-reviewer`, `glm-5.3-flash` — security-auditor (completed); agent `phase1-reviewer`
- Fresh verifier (Sol): `gpt-5.6-sol` — verifier; agent `sol-verifier`
- Phase 2 reviewers (Sol): **not run (deferred by blocker gate)**

<details>

```

In `packages/dashmate/src/commands/group/restart.js`:
- [SUGGESTION] packages/dashmate/src/commands/group/restart.js:42-64: Pre-stop group build trusts a single config to describe every node's local builds
  `configGroup.find(isServiceBuildRequired)` builds from only the first matching node. Group membership is based only on the `group` value, and no invariant requires every member to enable the same locally built services; `generateEnvs` selects the Drive, rs-dapi, and helper build compose files independently for each config. If one member enables a Drive build and another enables only an rs-dapi build, the first config does not build the second member's required image. The pull phase excludes that locally built service, and `ctx.skipBuildServices` suppresses all later per-node builds, so startup fails only after every node has stopped. Build the union of requirements from all group members before setting the shared skip flag.

In `packages/dashmate/src/status/providers.js`:
- [SUGGESTION] packages/dashmate/src/status/providers.js:21: providers.js sanitizer misses U+061C (Arabic letter mark) that sanitizeRemoteText covers
  The new diagnostic sanitizer removes bidi overrides, isolates, and directional marks, but omits U+061C, an invisible strong right-to-left mark. Remote release-tag text can therefore put U+061C into DEBUG output at line 182 and alter how the diagnostic is displayed in bidi-aware terminals. The sibling `sanitizeRemoteText` utility includes this code point for the same threat model, so the two new sanitizers should provide equivalent coverage.

In `packages/dashmate/src/listr/tasks/ssl/saveCertificateTask.js`:
- [BLOCKING] packages/dashmate/src/listr/tasks/ssl/saveCertificateTask.js:57-68: Replace an exposed private key through a fresh restricted inode
  (existing thread: https://github.com/dashpay/platform/pull/4283#discussion_r3714726980)
  Moving chmod before write prevents new opens through the old 0644 mode, but it does not revoke descriptors that another local user opened while the key was readable. Because `writeFileSync` truncates and rewrites the same inode, such a descriptor can seek back to offset zero and read the replacement private key after chmod; this behavior was reproduced directly against the current implementation's filesystem sequence. The individual-file gateway bind mount explains why a simple rename would leave the running container on the old inode, but it does not eliminate the exposure. Restoring confidentiality requires installing the replacement through a restricted inode and then remounting or recreating the gateway, or changing the mount arrangement so an atomic replacement is visible inside the container.
```
</details>

PR #4286, opened 2026-08-04
feat(sdk): expose OP_RETURN, output-order and change-to-VIN0 controls

romchornyi:

The Dash iOS wallet is restoring MAYACHAIN swap routes, which requires the DASH
deposit to carry the swap memo in an `OP_RETURN`. `CoreTransactionBuilder` could
not express that, so Maya was disabled during the DashSync unlink.

MAYAChain's UTXO deposit contract ([docs](https://docs.mayaprotocol.com/mayachain-dev-docs/concepts/sending-transactions),
"UTXO Chains") demands a specific shape: `VOUT0` = Asgard vault, `VOUT1` = the
memo as a zero-value `OP_RETURN`, `VOUT2` = change paid back to the **VIN0
address**, and no output reordering. The change rule matters because MAYAChain
identifies the depositor by VIN0 and pays refunds there — routing change to a
fresh HD address fails silently, with only a later refund going astray.

Because `finalize`/`build_signed` fund and sign inside a single FFI call, none of
this can be applied after the fact. It has to be expressible on the builder.

## What was done?

- `packages/rs-platform-wallet-ffi/src/core_wallet/transaction_builder.rs`:
  `core_wallet_tx_builder_add_op_return`, `..._preserve_output_order` and
  `..._change_to_first_input`, following the existing setter style. An over-long
  payload is rejected **before** `take_builder()` runs, so a refused memo cannot
  leave the slot holding a `mem::take` default and silently drop outputs the
  caller already configured.
- `packages/rs-platform-wallet-ffi/src/core_wallet/broadcast.rs`:
  `core_wallet_signed_transaction_v2_bytes` — read a finalized transaction's
  bytes without broadcasting, so the deposit shape can be asserted pre-broadcast.
- `packages/swift-sdk/.../CoreTransactionBuilder.swift`: `addOpReturn(_:)`,
  `preserveOutputOrder()`, `changeToFirstInput()`, and
  `FinalizedCoreTransaction.serializedData()`.
- `.github/workflows/tests-rs-workspace.yml`: fail the workflow if a local
  `[patch."https://github.com/dashpay/rust-dashcore"]` override is left in
  `Cargo.toml` — that override is invisible in review and produces a build that
  only works on one machine.

Depends on dashpay/rust-dashcore#922, which adds the underlying `add_op_return`,
`preserve_output_order` and `change_to_first_input` to `key-wallet`. Until that
merges and the `rev` in `Cargo.toml` is bumped, building this locally needs the
patch override — deliberately **not** committed, which is what the new CI guard
enforces.

## How Has This Been Tested?

Added `packages/swift-sdk/SwiftTests/SwiftDashSDKIntegrationTests/Core/MayaDepositVerificationIntegrationTests.swift`,
which builds short- and long-memo deposits against a local dashmate devnet and
asserts output count and order, the `OP_RETURN` payload, `VOUT2` == VIN0
scriptPubKey, the 80-byte memo ceiling, Maya's dust floor and a ≥ 1 duff/byte
fee — then checks fee parity for ordinary, multi-recipient, selected-input,
drain and asset-lock shapes so the precise output sizing does not move existing
fees.

Also verified: `cargo check -p platform-wallet-ffi`,
`./build_ios.sh --target ios --target sim`, and a green `dashpay` build of the
consuming wallet app.

Known gap: the integration test currently stalls in SPV bootstrap on a
22k-block devnet (compact filters lag past the 180 s wait) and has not yet run
its assertions end to end.

## Breaking Changes

None. All three builder controls are opt-in and default behaviour is unchanged.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added support for including OP_RETURN data in wallet transactions.
  * Added options to preserve transaction output order and route change to the first selected input.
  * Added the ability to retrieve serialized bytes from finalized transactions without consuming them.
* **Bug Fixes**
  * Improved validation and error reporting for invalid transaction data and oversized OP_RETURN payloads.
  * Preserved transaction builder state when rejecting oversized OP_RETURN payloads.


thepastaclaw:
⛔ Blockers found — Opus deferred (commit b3f28014)

packages/swift-sdk/SwiftTests/SwiftDashSDKIntegrationTests/Core/MayaDepositVerificationIntegrationTests.swift
```diff
@@ -0,0 +1,602 @@
...
+        XCTAssertLessThanOrEqual(observation.outputCount, 3, "\(observation.name) output count above Maya maximum")
+        XCTAssertEqual(observation.actualVaultAmount, observation.depositAmount, "\(observation.name) VOUT0 amount mismatch")
+        XCTAssertTrue(observation.outputOneIsOpReturn, "\(observation.name) VOUT1 is not OP_RETURN")
+        XCTAssertEqual(observation.decodedMemo, observation.memo, "\(observation.name) memo payload mismatch")
+        XCTAssertGreaterThanOrEqual(observation.depositAmount, Constants.minimumDepositDuffs, "\(observation.name) deposit fell below Maya dust floor")
+        XCTAssertLessThanOrEqual(observation.memoBytes, Constants.maxMemoBytes, "\(observation.name) memo exceeded 80 bytes")
+        XCTAssertGreaterThanOrEqual(observation.feeDuffs, UInt64(observation.serializedSize), "\(observation.name) fee fell below 1 duff/byte")
+        if observation.outputCount == 3 {
+            XCTAssertTrue(observation.outputTwoMatchesInputZeroScript, "\(observation.name) VOUT2 does not return change to VIN0 scriptPubKey")
+        }
```
thepastaclaw: **🟡 Suggestion: Require VOUT2 for fixtures that necessarily produce change**

Both deposit fixtures leave millions of duffs after the vault payment and fee, so each transaction must contain exactly three outputs: vault, memo, and change. Allowing `outputCount == 2` and conditionally skipping the VIN0-script assertion lets a regression that suppresses change pass, even though change-to-VIN0 is the load-bearing behavior this test claims to verify. Require exactly three outputs and always validate VOUT2 against VIN0.

```suggestion
        XCTAssertEqual(
            observation.outputCount,
            3,
            "\(observation.name) must contain vault, memo, and change outputs"
        )
        XCTAssertEqual(observation.actualVaultAmount, observation.depositAmount, "\(observation.name) VOUT0 amount mismatch")
        XCTAssertTrue(observation.outputOneIsOpReturn, "\(observation.name) VOUT1 is not OP_RETURN")
        XCTAssertEqual(observation.decodedMemo, observation.memo, "\(observation.name) memo payload mismatch")
        XCTAssertGreaterThanOrEqual(observation.depositAmount, Constants.minimumDepositDuffs, "\(observation.name) deposit fell below Maya dust floor")
        XCTAssertLessThanOrEqual(observation.memoBytes, Constants.maxMemoBytes, "\(observation.name) memo exceeded 80 bytes")
        XCTAssertGreaterThanOrEqual(observation.feeDuffs, UInt64(observation.serializedSize), "\(observation.name) fee fell below 1 duff/byte")
        XCTAssertTrue(observation.outputTwoMatchesInputZeroScript, "\(observation.name) VOUT2 does not return change to VIN0 scriptPubKey")
```

<sub>source: ['codex']</sub>

packages/swift-sdk/SwiftTests/SwiftDashSDKIntegrationTests/Core/MayaDepositVerificationIntegrationTests.swift
```diff
@@ -0,0 +1,602 @@
...
+        static let bip44TypeTag: UInt8 = 0
+        static let bip44StandardTag: UInt8 = 0
+        static let bip44AccountIndex: UInt32 = 0
+        static let feeRateSatPerKb: UInt64 = 1_000
+        static let minimumDepositDuffs: UInt64 = 10_000
+        static let maxMemoBytes = 80
+        static let longMemo =
+            "=:ARB.GLD:0x51a1449b3B6D635EddeC781cD47a99221712De97:344233230e4/1/0:_/def:15/0"
+        static let shortMemo =
+            "=:r:thor166n4w5039meulfa3p6ydg60ve6ueac7tlt0jws:669458827/1/0:_/def:15/0"
```
thepastaclaw, line 17: **🟡 Suggestion: Exercise the OP_RETURN 80/81-byte boundary**

The long memo fixture is 79 UTF-8 bytes, and the later `memoBytes <= 80` assertion only verifies the fixture rather than the new API's boundary behavior. No test proves that an 80-byte payload succeeds, an 81-byte payload is rejected, or that a rejection preserves outputs and options already stored in the FFI builder. Add deterministic boundary coverage that rejects 81 bytes and then successfully finalizes the same builder with its earlier configuration intact, which directly tests the pre-`take_builder()` guarantee introduced by this PR.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The FFI allocation and Swift ownership plumbing follow existing repository conventions, but the exact PR head is not buildable because the pinned key-wallet revision lacks every newly referenced builder API and the imported constant has a different name upstream. The Maya integration test also does not require the expected VOUT2 change output and does not exercise the 80/81-byte OP_RETURN boundary or builder-state preservation after rejection.

### Review provenance

<details>

```

In `Cargo.toml`:
- [BLOCKING] Cargo.toml:55-62: Pin a key-wallet revision that provides the new builder API
  The workspace still pins rust-dashcore and key-wallet to `70d4bf8e36057c58e02d56769a6e9760f701dd06`, which does not provide `TransactionBuilder::add_op_return`, `preserve_output_order`, or `change_to_first_input`, nor the imported OP_RETURN limit constant. At this exact head, `cargo check -p platform-wallet-ffi --locked` fails with E0432/E0599, so the native library and the new Swift API cannot be built. Upstream rust-dashcore PR #922 currently provides the methods at `eebacae3d1152609b04e9c9af05acc87ea9b32ad`, but exports the limit as `DEFAULT_MAX_OP_RETURN_BYTES`, not `MAX_STANDARD_OP_RETURN_BYTES`. After the upstream change is merged, pin all rust-dashcore workspace dependencies to a compatible revision, update `Cargo.lock`, and align the import with the exported name.

In `packages/swift-sdk/SwiftTests/SwiftDashSDKIntegrationTests/Core/MayaDepositVerificationIntegrationTests.swift`:
- [SUGGESTION] packages/swift-sdk/SwiftTests/SwiftDashSDKIntegrationTests/Core/MayaDepositVerificationIntegrationTests.swift:181-191: Require VOUT2 for fixtures that necessarily produce change
  Both deposit fixtures leave millions of duffs after the vault payment and fee, so each transaction must contain exactly three outputs: vault, memo, and change. Allowing `outputCount == 2` and conditionally skipping the VIN0-script assertion lets a regression that suppresses change pass, even though change-to-VIN0 is the load-bearing behavior this test claims to verify. Require exactly three outputs and always validate VOUT2 against VIN0.
- [SUGGESTION] packages/swift-sdk/SwiftTests/SwiftDashSDKIntegrationTests/Core/MayaDepositVerificationIntegrationTests.swift:13-17: Exercise the OP_RETURN 80/81-byte boundary
  The long memo fixture is 79 UTF-8 bytes, and the later `memoBytes <= 80` assertion only verifies the fixture rather than the new API's boundary behavior. No test proves that an 80-byte payload succeeds, an 81-byte payload is rejected, or that a rejection preserves outputs and options already stored in the FFI builder. Add deterministic boundary coverage that rejects 81 bytes and then successfully finalizes the same builder with its earlier configuration intact, which directly tests the pre-`take_builder()` guarantee introduced by this PR.
```
</details>

romchornyi:
Pushed `8370b6a2` — all four review comments addressed.

**Require exactly three outputs, always assert VOUT2 → VIN0**

Agreed, and this was the important one. Both fixtures leave millions of duffs after the vault payment and fee, so change is mandatory; `outputCount == 2` was never reachable for them, and allowing it meant a regression that suppressed change would pass the very test that exists to prove change goes to VIN0. Now `XCTAssertEqual(outputCount, 3)` and the VOUT2 assertion is unconditional.

(For contrast, the app-side assertion in dashpay/dashwallet-ios#916 deliberately accepts 2 or 3, because a real deposit *can* drop change below the 546-duff dust threshold. Here the fixture amounts rule that out.)

**Guard counts before indexing**

Done. Shape assertions now run before any subscripting, so a wrong shape produces a readable failure instead of trapping and killing the test process.

**Exercise the 80/81-byte boundary**

Added `testOpReturnCeilingBoundaryAndRejectionPreservesBuilder`. It rejects 81 bytes, accepts exactly 80, and — the part worth having — reuses the *same* builder after the rejection and finalizes it, asserting the vault output and ordering flags configured before the failed call are still there. That directly pins the guarantee this branch introduces: `core_wallet_tx_builder_add_op_return` validates the payload before `take_builder()`, so a refused memo can't leave a `mem::take` default behind and silently discard the caller's configuration.

**Keep the suite from hanging local Swift SDK CI**

Gated behind `MAYA_DEPOSIT_VERIFICATION=1`. You are right that `run_tests.sh` picks this bundle up and that a bootstrap stall would hang rather than fail the job — that stall is exactly what has stopped these tests reaching their assertions so far. Also confirmed no local `stopSpv` is needed, per your note about `IntegrationTestCase.tearDown` / `cleanupSpvCache()`.

---

**On the red CI**

Every failing check has one root cause: `Cargo.toml` still pins rust-dashcore at `70d4bf8e`, which predates the `key-wallet` API this PR wraps.

```
error[E0432]: unresolved import `...transaction_builder::MAX_STANDARD_OP_RETURN_BYTES`
error[E0599]: no method named `add_op_return` found for struct `TransactionBuilder`
```

Same failure in *Rust workspace tests (macOS)* and *Kotlin SDK build*; *CodeRabbit or PastaClaw review* failed only because no review had been submitted at that point and re-runs on its own. It clears when dashpay/rust-dashcore#922 merges and the rev here is bumped.

One thing to flag before that bump, because it is not a one-liner: current `dev` in rust-dashcore is 6 commits ahead of the pinned rev, and two of them are breaking for this repo — #818 removed `AddressInfo.used` / `generated_at` / `used_at`, and #919 changed the `select_coins_with_size` arity. Building `platform-wallet-ffi` against a rebased `key-wallet` today fails in `platform-wallet` itself, on those, entirely independently of anything in this PR:

```
error[E0609]: no field `used` on type `key_wallet::AddressInfo`
error[E0061]: this method takes 6 arguments but 5 arguments were supplied
error: could not compile `platform-wallet` (lib) due to 7 previous errors
```

So the rev bump wants either a companion change adapting `platform-wallet` to those two upstream changes, or a rev that carries #922 without them. Happy to take the adaptation if that is the preferred route.

**Verification status, honestly:** the Swift changes are not compile-verified locally — the SDK cannot be built here right now, since `DashSDKFFI` needs `platform-wallet-ffi`, which is blocked by exactly the mismatch above. They are assertion and gating changes over existing helpers, but I would not call them proven until CI or a local build runs them.

packages/swift-sdk/SwiftTests/SwiftDashSDKIntegrationTests/Core/MayaDepositVerificationIntegrationTests.swift
romchornyi: Agreed, and this was the one worth catching. Fixed in `8370b6a2`.

You are right that the fixtures make change mandatory: `0.5 DASH` funding against a `200_000`-duff deposit, and `0.4 DASH` against `35_000_000`, both leave a change output far above the 546-duff dust threshold. `outputCount == 2` was unreachable for them, so the range check bought nothing and cost the assertion — a regression that dropped change entirely would have satisfied `>= 2 && <= 3` and then skipped the VIN0 check under `if observation.outputCount == 3`, which is precisely the behaviour this test exists to prove.

Now:

```swift
XCTAssertEqual(observation.outputCount, 3, "\(observation.name) must be vault + memo + change")
...
XCTAssertTrue(observation.outputTwoMatchesInputZeroScript, "\(observation.name) VOUT2 does not return change to VIN0 scriptPubKey")
```

with the VOUT2 assertion no longer conditional. I also moved the shape check ahead of the subscripting in `buildDepositObservation`, so a wrong shape fails readably instead of trapping on `decoded.outputs[2]`.

One deliberate asymmetry worth recording, since it looks like the same check in two places: the app-side guard in dashpay/dashwallet-ios#916 (`assertSwapDepositShape`) still accepts **2 or 3** outputs. There the transaction is real, so change genuinely can fall below dust and be dropped by the builder — MAYAChain still identifies the depositor by VIN0 in that case, so a two-output deposit is valid. It is only here, where the fixture amounts rule that out, that three is the correct fixed expectation.

packages/swift-sdk/SwiftTests/SwiftDashSDKIntegrationTests/Core/MayaDepositVerificationIntegrationTests.swift
romchornyi: Good catch on both halves — the fixture proved nothing about the boundary, and nothing at all covered the rejection path. Added `testOpReturnCeilingBoundaryAndRejectionPreservesBuilder` in `8370b6a2`.

It does exactly the two things you asked for:

```swift
let builder = try CoreTransactionBuilder(network: .regtest)
try builder.addOutput(address: vaultAddress, amountDuffs: depositAmount)
try builder.preserveOutputOrder()
try builder.changeToFirstInput()

XCTAssertThrowsError(try builder.addOpReturn(overCeiling))   // 81 bytes

// Same builder instance: if the rejection had consumed it, this would build a
// transaction missing the vault output and the ordering flags.
try builder.addOpReturn(atCeiling)                            // exactly 80 bytes
let tx = try builder.finalizeAtomic(...)
```

then asserts three outputs, that VOUT0 is still the vault payment for the original amount, and that the 80-byte payload round-trips verbatim out of VOUT1.

The reuse-after-rejection part is the one I care about most, because it pins a guarantee that is otherwise invisible: `core_wallet_tx_builder_add_op_return` takes the builder by value on the Rust side, so the naive implementation drops it on error and `take_builder`'s `mem::take` leaves a default in the slot — the vault output and both ordering flags would vanish silently and the next `finalize` would produce a plain send. The FFI validates the payload *before* `take_builder()` specifically to avoid that, and until now nothing exercised it.

Two caveats I would rather state than have you discover: the new test is gated behind `MAYA_DEPOSIT_VERIFICATION=1` along with the rest of the suite (CodeRabbit flagged that `run_tests.sh` runs this bundle and a bootstrap stall would hang the job), and it has **not been run yet** — the SDK does not currently build locally because `platform-wallet-ffi` needs the `key-wallet` API from dashpay/rust-dashcore#922, which is still open. So treat it as written-and-reviewed, not as passing.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/CoreWallet/CoreTransactionBuilder.swift
```diff
@@ -274,6 +308,22 @@ public final class CoreTransactionBuilder {
...
+        try core_wallet_tx_builder_preserve_output_order(handle).check()
+        return self
+    }
+
+    /// Route change to the first selected input address (VIN0) for a MAYACHAIN-style deposit.
+    /// See https://docs.mayaprotocol.com/mayachain-dev-docs/concepts/sending-transactions
+    @discardableResult
+    public func changeToFirstInput() throws -> CoreTransactionBuilder {
+        try core_wallet_tx_builder_change_to_first_input(handle).check()
+        return self
```
thepastaclaw, line 324: **🔴 Blocking: Guard the new Swift setters after Rust consumes the builder**

`finalizeAtomic`, `buildSigned`, and `finalizeSignedPayment` set `consumed` after Rust reclaims the `FFITransactionBuilder` and its inner builder with `Box::from_raw`, but the new `addOpReturn`, `preserveOutputOrder`, and `changeToFirstInput` methods do not check that state. Calling one of these public Swift methods after any consuming finalizer passes the retained non-null but dangling `handle` through the C ABI; `check_ptr!` only rejects null and the FFI then dereferences freed memory. Add the same `consumed` guard used by the finalizers before each new method crosses the FFI boundary.

<sub>source: ['codex']</sub>

packages/swift-sdk/SwiftTests/SwiftDashSDKIntegrationTests/Core/MayaDepositVerificationIntegrationTests.swift
```diff
@@ -0,0 +1,685 @@
...
+    private func skipUnlessEnabled() throws {
+        try XCTSkipUnless(
+            Self.isEnabled,
+            "Set MAYA_DEPOSIT_VERIFICATION=1 to run the Maya deposit verification suite "
+                + "(requires a local dashmate devnet and a completed SPV bootstrap)."
+        )
+    }
+
+    func testPrompt04StaticProofAndLegacyFeeParity() async throws {
+        try skipUnlessEnabled()
```
thepastaclaw, line 71: **🟡 Suggestion: Run the Maya verification coverage in an automated test path**

Both tests in this new suite skip unless `MAYA_DEPOSIT_VERIFICATION=1`, and repository workflows and test scripts never set that variable. Normal Swift and CI runs therefore exercise none of the deposit output ordering, VIN0 change routing, 80/81-byte boundary, rejected-builder preservation, or fee-parity assertions. Add hermetic lower-level FFI coverage for these guarantees, or enable the integration suite in a dedicated reliable job once its SPV bootstrap dependency is stable.

<sub>source: ['codex']</sub>

packages/swift-sdk/SwiftTests/SwiftDashSDKIntegrationTests/Core/MayaDepositVerificationIntegrationTests.swift
```diff
@@ -0,0 +1,685 @@
...
+        // Both fixtures fund far more than the deposit plus fee, so change is always well above
+        // dust and the transaction must be exactly vault + memo + change. Assert the shape here,
+        // before any subscripting: a wrong shape must surface as a readable failure rather than
+        // trapping on an out-of-range index and taking the whole test process down.
+        XCTAssertEqual(
+            decoded.outputs.count, 3,
+            "\(name) must contain vault, memo and change outputs"
+        )
+        XCTAssertFalse(decoded.inputs.isEmpty, "\(name) has no inputs")
+        guard decoded.outputs.count == 3, let firstInput = decoded.inputs.first else {
```
thepastaclaw, line 235: **🟡 Suggestion: Validate the output shape before indexing VOUT1**

`buildDepositObservation` reads `decoded.outputs[1]` before its output-count guard, despite the comment claiming that shape validation precedes subscripting. The boundary test similarly continues to index outputs after `XCTAssertEqual`, which records a failure but does not stop execution. A transaction-shape regression can therefore trap instead of producing a readable test failure. Guard the count before every output subscript, and throw a normal test error rather than `XCTSkip` when an invalid transaction shape is observed.

<sub>source: ['codex']</sub>

packages/swift-sdk/SwiftTests/SwiftDashSDKIntegrationTests/Core/MayaDepositVerificationIntegrationTests.swift
```diff
@@ -0,0 +1,685 @@
...
+    private static let isEnabled =
+        ProcessInfo.processInfo.environment["MAYA_DEPOSIT_VERIFICATION"] == "1"
+
+    private func skipUnlessEnabled() throws {
+        try XCTSkipUnless(
+            Self.isEnabled,
+            "Set MAYA_DEPOSIT_VERIFICATION=1 to run the Maya deposit verification suite "
+                + "(requires a local dashmate devnet and a completed SPV bootstrap)."
+        )
+    }
```
thepastaclaw, line 68: **💬 Nitpick: Use the configured two-space Swift indentation**

The repository root `.editorconfig` applies two-space indentation to Swift files, while the newly added test code uses four spaces per nesting level. Reindent the new test file consistently to match the configured repository style.

thepastaclaw (changes requested):

The exact head does not compile because the pinned rust-dashcore revision lacks every key-wallet builder API added by this PR. The new Swift setters also permit use-after-free after finalization; test execution and failure-reporting gaps should be addressed alongside those blockers.

Source: reviewers codex/general=gpt-5.6-sol(completed); codex/ffi-engineer=gpt-5.6-sol(completed); verifier=codex/verifier=gpt-5.6-sol(completed); coordinator=openclaw-agent/cliproxy/gpt-5.6-sol(orchestration-only).

### Review provenance

🔴 2 blocking | 🟡 2 suggestion(s) | 💬 1 nitpick(s)

<details>

```

In `Cargo.toml`:
- [BLOCKING] Cargo.toml:55-62: Pin a key-wallet revision that provides the new builder API
  The workspace pins rust-dashcore revision `08bf729de819f52973002f754342fedac14a06db`, which does not provide `DEFAULT_MAX_OP_RETURN_BYTES`, `TransactionBuilder::add_op_return`, `preserve_output_order`, or `change_to_first_input`. Running `cargo check -p platform-wallet-ffi` at this exact head fails with an unresolved import and missing-method errors at all four new call sites. Update every rust-dashcore workspace dependency and `Cargo.lock` to a revision containing these APIs before merging.

In `packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/CoreWallet/CoreTransactionBuilder.swift`:
- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/CoreWallet/CoreTransactionBuilder.swift:292-324: Guard the new Swift setters after Rust consumes the builder
  `finalizeAtomic`, `buildSigned`, and `finalizeSignedPayment` set `consumed` after Rust reclaims the `FFITransactionBuilder` and its inner builder with `Box::from_raw`, but the new `addOpReturn`, `preserveOutputOrder`, and `changeToFirstInput` methods do not check that state. Calling one of these public Swift methods after any consuming finalizer passes the retained non-null but dangling `handle` through the C ABI; `check_ptr!` only rejects null and the FFI then dereferences freed memory. Add the same `consumed` guard used by the finalizers before each new method crosses the FFI boundary.

In `packages/swift-sdk/SwiftTests/SwiftDashSDKIntegrationTests/Core/MayaDepositVerificationIntegrationTests.swift`:
- [SUGGESTION] packages/swift-sdk/SwiftTests/SwiftDashSDKIntegrationTests/Core/MayaDepositVerificationIntegrationTests.swift:55-71: Run the Maya verification coverage in an automated test path
  Both tests in this new suite skip unless `MAYA_DEPOSIT_VERIFICATION=1`, and repository workflows and test scripts never set that variable. Normal Swift and CI runs therefore exercise none of the deposit output ordering, VIN0 change routing, 80/81-byte boundary, rejected-builder preservation, or fee-parity assertions. Add hermetic lower-level FFI coverage for these guarantees, or enable the integration suite in a dedicated reliable job once its SPV bootstrap dependency is stable.
- [SUGGESTION] packages/swift-sdk/SwiftTests/SwiftDashSDKIntegrationTests/Core/MayaDepositVerificationIntegrationTests.swift:174-235: Validate the output shape before indexing VOUT1
  `buildDepositObservation` reads `decoded.outputs[1]` before its output-count guard, despite the comment claiming that shape validation precedes subscripting. The boundary test similarly continues to index outputs after `XCTAssertEqual`, which records a failure but does not stop execution. A transaction-shape regression can therefore trap instead of producing a readable test failure. Guard the count before every output subscript, and throw a normal test error rather than `XCTSkip` when an invalid transaction shape is observed.
```
</details>

packages/swift-sdk/SwiftTests/SwiftDashSDKIntegrationTests/Core/MayaDepositVerificationIntegrationTests.swift
thepastaclaw: Resolved in `8370b6a` — _Exercise the OP_RETURN 80/81-byte boundary_ no longer present.

packages/swift-sdk/SwiftTests/SwiftDashSDKIntegrationTests/Core/MayaDepositVerificationIntegrationTests.swift
thepastaclaw: Resolved in `8370b6a` — _Require VOUT2 for fixtures that necessarily produce change_ no longer present.

packages/rs-platform-wallet-ffi/src/core_wallet/broadcast.rs
```diff
@@ -173,6 +173,28 @@ pub unsafe extern "C" fn core_wallet_signed_transaction_v2_fee(
...
+
+    let bytes = unwrap_option_or_return!(CORE_SIGNED_TRANSACTION_V2_STORAGE
+        .with_item(transaction_handle, |tx| dashcore::consensus::serialize(
+            tx.transaction.transaction()
+        )));
+    let len = bytes.len();
+    let boxed = bytes.into_boxed_slice();
+    *out_bytes = Box::into_raw(boxed) as *mut u8;
+    *out_len = len;
+    PlatformWalletFFIResult::ok()
```
thepastaclaw, line 195: **💬 Nitpick: Declare ownership of serialized byte buffers in the C ABI**

`core_wallet_signed_transaction_v2_bytes` returns an allocation created with `Box::into_raw`, but its cbindgen-visible documentation does not state that the caller takes ownership or must release the pointer using `platform_wallet_bytes_free` with the returned length. The current Swift wrapper correctly copies and frees it, but direct C and future binding consumers cannot infer the allocator/deallocator contract and may leak the buffer or incorrectly call the platform allocator's `free()`.

```suggestion
/// Copy the consensus-serialized finalized transaction into a newly allocated buffer.
///
/// On success, `out_bytes` receives a Rust-allocated buffer of `out_len` bytes.
/// The caller takes ownership and must release it exactly once with
/// `platform_wallet_bytes_free(*out_bytes, *out_len)`. Errors leave the outputs
/// initialized to null and zero.
///
/// # Safety
/// `out_bytes` and `out_len` must be valid writable pointers.
#[no_mangle]
pub unsafe extern "C" fn core_wallet_signed_transaction_v2_bytes(
    transaction_handle: Handle,
    out_bytes: *mut *mut u8,
    out_len: *mut usize,
) -> PlatformWalletFFIResult {
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

Carried-forward prior findings: the incompatible key-wallet pin and the consumed-builder use-after-free remain blocking, the output-shape guard remains a valid test suggestion, and automated Maya coverage remains intentionally deferred; the indentation finding is outdated as an actionable review item. The latest delta only renames the OP_RETURN constant and introduces no new defect, but it does not fix the dependency blocker: `cargo check -p platform-wallet-ffi` still fails at the exact head. One additional cumulative PR nitpick was identified: the newly exported byte-buffer function does not document its C ownership contract.

Source: reviewers codex/general=gpt-5.6-sol(completed); codex/ffi-engineer=gpt-5.6-sol(completed); verifier=codex/verifier=gpt-5.6-sol(completed); coordinator=openclaw-agent/cliproxy/gpt-5.6-sol(orchestration-only).

### Review provenance

🔴 2 blocking | 💬 1 nitpick(s)

<details>

```

In `Cargo.toml`:
- [BLOCKING] Cargo.toml:55-62: Pin a key-wallet revision that provides the new builder API
  STILL VALID. The workspace and Cargo.lock remain pinned to rust-dashcore revision `08bf729de819f52973002f754342fedac14a06db`. Direct inspection of that revision confirms it lacks `MAX_STANDARD_OP_RETURN_BYTES` and `TransactionBuilder::{add_op_return,preserve_output_order,change_to_first_input}`. Running `cargo check -p platform-wallet-ffi` at the exact reviewed head fails with E0432 and E0599 for those symbols, so the native library and dependent Swift SDK cannot build. Revision `399066748ef5d64e5132d07232b6a9000270450c` is a descendant of the current pin and contains all four required API elements; update every rust-dashcore workspace dependency consistently and refresh Cargo.lock.
```
</details>

packages/swift-sdk/SwiftTests/SwiftDashSDKIntegrationTests/Core/MayaDepositVerificationIntegrationTests.swift
thepastaclaw: Resolved in this update — _Use the configured two-space Swift indentation_ no longer present.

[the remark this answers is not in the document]
thepastaclaw: The indentation differs from `.editorconfig`, but this is a formatting-only editor-tooling concern and is not actionable under the current correctness-focused review scope.

PR #4287, opened 2026-08-04
ci: schedule Rust workspace tests on macOS or Linux self-hosted runners

QuantumExplorer:

We added a self-hosted Linux runner, but `Rust workspace tests / Tests (macOS)` was pinned to `runs-on: [self-hosted, macOS, ARM64]` and could never schedule onto it. Labels in a `runs-on` array are AND-ed — there is no OR syntax — so no edit to that list alone can express "macOS or Linux".

Current runners on this repo:

| runner | labels |
|---|---|
| mac-runner-1, mac-runner-2 | `self-hosted, macOS, ARM64` |
| ubuntu-runner-1 | `self-hosted, Linux, X64, ubuntu, ms-a2` |

## What was done?

Select on a shared custom label instead:

```yaml
runs-on: [self-hosted, rust-ci]
```

Whichever of the three runners is idle takes the job. Keeping `self-hosted` in the list means a GitHub-hosted Linux VM can never match it, so Linux stays self-hosted-only.

Making the job body portable:

- **Build deps** — the Homebrew step is gated to macOS. Linux gets an apt equivalent covering only what `./.github/actions/rust` doesn't already install (it handles clang, llvm and libsnappy on Linux), plus `gh` for the immutable-structure check and `rustup` for `dtolnay/rust-toolchain`. Each branch is guarded, so it's a no-op once the persistent runner is provisioned by the first run.
- **Disk prune** — the absolute 120 GB / 60 GB-free caps are now clamped to a share of the volume (`TOTAL/3` and `TOTAL/5`). On the large macOS volumes this is exactly today's behavior; it stops a smaller volume from tripping the floor on every run and forcing a cold rebuild each time.
- **GPG reset** — a macOS keyboxd workaround, gated to macOS.

Linux builds rocksdb from source the same way macOS already does, rather than linking the prebuilt static lib the old Ubuntu jobs used. The persistent `target/` makes that a one-time cost and it avoids their `sed` pass rewriting cdylib crate-types in `Cargo.toml`, which static rocksdb forced.

**Removed the three Ubuntu backup jobs** (`test-ubuntu`, `lint-ubuntu`, and the whole of `tests-rs-doctests.yml`) along with their `sleep 15` + poll-the-Mac-job-status dance. `UBUNTU_BACKUP_ENABLED` is `false` on this repo, so they have been dead code; a second self-hosted runner is the redundancy they were standing in for. Doctests already run inside both Rust jobs under `doctests-changed`, so dropping the standalone `rs-doctests` job loses no coverage.

Also added `.github/actionlint.yaml` declaring `rust-ci`, so local `actionlint` runs don't flag it as an unknown label.

## Required before merge

**The `rust-ci` label must be added to all three runners** (Settings → Actions → Runners → *runner* → Labels). Until then the job has nothing to schedule onto and will sit queued.

## How Has This Been Tested?

- `actionlint` (with shellcheck) clean across the three touched workflows; the remaining shellcheck notes are pre-existing and unrelated — the same `SC2086` fires on the untouched Homebrew line in `tests-rs-wallet.yml`.
- YAML parse + job-list check on all three files.
- Swept for stale references to the deleted workflow, the old `Tests (macOS)` job name, and `UBUNTU_BACKUP_ENABLED`; the three comment mentions left behind were updated.
- Branch protection on `v4.2-dev` has no required status checks configured, so the job rename breaks no gate.
- CI on this PR is the real test of the Linux path — it can't be exercised locally.

## Breaking Changes

None.

## Notes for reviewers

- `tests-rs-wallet.yml` stays pinned to the macOS runners. That fast path is short and mac-only by design; floating it too is a reasonable follow-up but I kept this PR to the workspace job. Its header comment is updated to say so.
- `UBUNTU_BACKUP_ENABLED` is now unreferenced and can be deleted from repo variables.
- Fork PRs get no Rust coverage — unchanged. The `if:` guard blocked them from the mac runners before and now guards both platforms; the Ubuntu backup that could theoretically have covered them was disabled.

## Checklist:

- [x] I have checked there is no other PR doing the same thing

## Summary by CodeRabbit

* **Chores**
  * Updated CI configuration to support self-hosted runners labeled `rust-ci`.
  * Consolidated Rust workspace testing onto self-hosted runners with automatic environment setup and disk-capacity checks.
  * Removed the standalone Rust doctest workflow and Ubuntu fallback jobs.
  * Updated wallet workflow documentation to reflect revised macOS runner requirements and testing scope.
  * Improved workflow safeguards for fork-based pull requests and runner environments.


thepastaclaw:
⛔ Blockers found — Sonnet deferred (commit 0444662c)

.github/workflows/tests-rs-workspace.yml
```diff
@@ -53,6 +71,44 @@ jobs:
...
+      # Linux; this covers what the rest of the job needs and what a bare
+      # self-hosted image doesn't ship. Every branch is a no-op once the
+      # persistent runner has been provisioned by the first run.
+      - name: Install build dependencies (Linux)
+        if: runner.os == 'Linux'
+        run: |
+          set -euo pipefail
+
+          MISSING=()
+          for pkg in build-essential cmake libgmp-dev pkg-config jq zip; do
```
thepastaclaw: **🔴 Blocking: Install OpenSSL headers on the Linux runner**

The Linux provisioning list installs `pkg-config` but not the OpenSSL development package. The Linux workspace build activates the default, non-vendored `openssl-sys` feature through `native-tls` and reqwest, including during `cargo clippy --workspace --all-features`. On a fresh Ubuntu runner, `openssl-sys` therefore cannot find the required OpenSSL headers and `openssl.pc` metadata. Add `libssl-dev` to the packages installed during first-run provisioning.

```suggestion
          for pkg in build-essential cmake libgmp-dev libssl-dev pkg-config jq zip; do
```

<sub>source: ['codex']</sub>

.github/workflows/tests.yml
```diff
@@ -156,11 +155,11 @@ jobs:
...
-          # must take the full workspace path (whose Ubuntu backup jobs cover
-          # them when UBUNTU_BACKUP_ENABLED is set, while the wallet
-          # workflow's only job skips fork PRs). Push and schedule runs have
-          # no reliable base SHA — scope computed from the last commit alone
-          # could silently downgrade the nightly / post-merge full runs.
+          # must take the full workspace path (both Rust workflows skip fork
+          # PRs outright, since neither may run untrusted code on a persistent
+          # runner). Push and schedule runs have no reliable base SHA — scope
+          # computed from the last commit alone could silently downgrade the
+          # nightly / post-merge full runs.
```
thepastaclaw, line 162: **💬 Nitpick: Document the trusted fork exception in the scope comment**

The comment says the fast path applies only to same-repository pull requests and that both Rust workflows skip fork PRs, but the condition immediately below explicitly trusts forks owned by `thepastaclaw`. Those pull requests continue through wallet-scope classification and can reach `tests-rs-wallet.yml`. Update the comment to distinguish untrusted forks from this trusted exception.

```suggestion
          # The fast path applies to same-repo pull requests and to the trusted
          # thepastaclaw fork. Untrusted fork PRs are classified as full-scope
          # changes, and both Rust workflows skip them because neither may run
          # untrusted code on a persistent runner. Push and schedule runs have
          # no reliable base SHA — scope computed from the last commit alone
          # could silently downgrade the nightly / post-merge full runs.
```

thepastaclaw (changes requested):

The shared runner label and OS-specific workflow gates correctly allow the workspace job to target either macOS or Linux. However, the Linux bootstrap omits both the OpenSSL development files required by the workspace dependency graph and the cargo-nextest executable required by the test command, so a newly provisioned runner cannot reliably complete the job. The Rust scope comment also inaccurately describes the explicitly trusted fork exception.

Source: reviewers codex/general=gpt-5.6-sol(completed); verifier=codex/verifier=gpt-5.6-sol(completed); coordinator=openclaw-agent/cliproxy/gpt-5.6-sol(orchestration-only).

### Review provenance

🔴 2 blocking | 💬 1 nitpick(s)

<details>

```

In `.github/workflows/tests-rs-workspace.yml`:
- [BLOCKING] .github/workflows/tests-rs-workspace.yml:84: Install OpenSSL headers on the Linux runner
  The Linux provisioning list installs `pkg-config` but not the OpenSSL development package. The Linux workspace build activates the default, non-vendored `openssl-sys` feature through `native-tls` and reqwest, including during `cargo clippy --workspace --all-features`. On a fresh Ubuntu runner, `openssl-sys` therefore cannot find the required OpenSSL headers and `openssl.pc` metadata. Add `libssl-dev` to the packages installed during first-run provisioning.
- [BLOCKING] .github/workflows/tests-rs-workspace.yml:118-119: Provision cargo-nextest before invoking llvm-cov nextest
  The test step runs `cargo llvm-cov nextest`, which delegates execution to the separately installed `cargo-nextest` binary. Installing `cargo-llvm-cov` does not provide that binary. The installer was previously removed from the macOS path only because nextest was already installed on those persistent runners, while the deleted Ubuntu fallback installed it explicitly. The new Linux first-run setup does neither, so a fresh Linux runner will fail when it reaches the non-shielded test step.
```
</details>

QuantumExplorer (commented):
Reviewed

PR #4289, opened 2026-08-05
fix(platform-wallet): batch wallet-event persistence + expose sync_fault (watermark-freeze mitigations)

bfoss765:
## Summary

Makes reviewable the two watermark-freeze commits that already shipped in the downstream QA build (v41int15) but were never opened as a PR. Both harden the `dashpay/platform#4069` durable-watermark path.

### 1. Batch wallet-event persistence (`3e8c27fbd0`)
The wallet-event adapter issued one `persister.store(..)` per `WalletEvent`. On Android that store is a JNI hop into a Room transaction (milliseconds), while projecting an event into a `CoreChangeSet` is microseconds — so the drain rate was pinned at the store rate. The upstream producer publishes fire-and-forget onto a bounded broadcast ring (capacity 1000); a historical SPV catch-up outruns a consumer that slow, the ring overflows, `recv()` returns `Lagged`, and the `#4069` guard freezes `synced_height` for the rest of the process lifetime (a **permanent** latch). In the field this presented as a mainnet sync that climbs, then "rolls back" on every relaunch.

Fix: block for the first event of a batch, then fold everything already buffered behind it via non-waiting `try_recv` (bounded by `ADAPTER_STORE_BATCH_LIMIT`) into one `store()` per wallet. The drain rate is now bounded by the ring rather than the persister.

### 2. Expose `sync_fault_detected` to the host (`e64662205b`)
`PlatformWalletManager::sync_fault_detected()` existed since the `#4069` guard landed but stopped at the Rust boundary. Surface it through the full C-FFI → JNI → Kotlin path so the app can show "verification failed / rescan pending" instead of silently re-scanning from a stale height. Unconditional (outside the `shielded` feature gate), since the fault is a core-persistence signal.

## Note
Batching raises the burst threshold but a single `Lagged` still freezes forever. The **root-cause lossless fix** (dedicated unbounded persistence channel; producer side is dashpay/rust-dashcore#924) is a **follow-up PR stacked on this branch**.

## Test
`cargo test -p rs-platform-wallet` green; built and shipped in v41int15.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added APIs to check whether wallet synchronization has encountered a persistent fault.
  * Applications can now detect when a rescan may be required and surface the appropriate status.

* **Bug Fixes**
  * Improved wallet-event processing by batching updates and reducing redundant persistence operations.
[...]

PR #4290, opened 2026-08-05
fix(platform-wallet): lossless mpsc persistence drain — root-cause fix for the sync-watermark freeze

bfoss765:
## Summary

Root-cause fix for the mainnet sync-watermark freeze (dashpay/platform#4069). **Stacked on #4289** (batching + `sync_fault` exposure) — the first two commits here are #4289; review the top commit. **Requires the producer PR dashpay/rust-dashcore#924** to land.

Batching (#4289) raised the burst threshold but a single `broadcast::Lagged` still froze a wallet's durable sync watermark permanently. The producer (#924) now offers a dedicated, **unbounded** `mpsc` persistence channel alongside its lossy broadcast; this switches the consumer onto it.

## Changes (top commit)
- **core_bridge.rs**: `spawn_/run_wallet_event_adapter` take `mpsc::UnboundedReceiver<WalletEvent>` instead of `broadcast::Receiver`. The batched `try_recv` fold is kept verbatim. The `Lagged` / `missed` / global `fault_all` path is removed — an unbounded channel can never lag. `AdapterFaultState` keeps only the per-wallet store-rejection freeze as a fail-closed backstop.
- **manager/mod.rs**: take the receiver via `take_persistence_receiver()` instead of `subscribe_events()`. The mpsc buffers events emitted before the task's first poll, so there is no subscribe-before-publish race.
- **Diagnostics** via the `log` facade (android_logger forwards `log` to logcat at Info; `tracing` may not): one `log::info!("wallet-event batch: folded=.. wallets=.. synced_height_persisted=.. faulted=.. missed=0")` per drain, and a one-shot `log::error!("SYNC WATERMARK FROZEN …")` if the freeze ever latches — so the next tester logcat is unambiguous.
- **Cargo.toml**: re-pin `key-wallet-manager` (and the sibling rust-dashcore crates, kept consistent to avoid a duplicate-crate type mismatch) to the fork rev carrying #924. *(The shipping v41int16 AAR builds against a rev of #924 rebased onto the integration branch's rust-dashcore base; this branch pins the v4.2-dev-based rev for a minimal, compilable review.)*

## #4069-safety
The channel is lossless and in-order, so every `TransactionDetected` / `BlockProcessed` row event reaches the persister before the `SyncHeightAdvanced` watermark that implies it — the durable watermark can never outrun its rows. The freeze guard stays as a fail-closed backstop but should now never fire.

## Why unbounded, not bounded back-pressure
Several producer emit sites run inside the manager's `RwLock` **write** guard, while this consumer needs a `read()` lock on the same manager to project each event. A bounded `send().await`/`blocking_send` parked under the write guard would deadlock this consumer. Unbounded keeps the producer lock-safe while still lossless. See #924 for the full argument.

## Test
Broadcast-driven adapter tests ported to the mpsc; the `Lagged` test is replaced by `lossless_burst_never_freezes_and_watermark_reaches_tip` (a 3000-event burst — 3× the old ring — advances the watermark to the tip with no freeze). `cargo test -p platform-wallet` (531) and `-p platform-wallet-ffi` (224) green; `cargo fmt --check` clean.

## Summary by CodeRabbit

- **New Features**
  - Added a wallet synchronization health check to detect persistent sync faults.
  - Exposed the sync-fault status through the Kotlin wallet manager API.

- **Improvements**
  - Improved event persistence reliability by preventing event loss during bursts and startup.
  - Added batching for wallet updates and isolated persistence failures to affected wallets.
  - Preserved wallet synchronization progress when unrelated persistence operations fail.

- **Bug Fixes**
  - Prevented synchronization issues caused by dropped or delayed wallet events.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,14 +52,14 @@ members = [
 ]
 
 [workspace.dependencies]
-dashcore = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
-dash-network-seeds = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
-dash-spv = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
-key-wallet = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
-key-wallet-ffi = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
-key-wallet-manager = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
-dash-network = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
-dashcore-rpc = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
+dashcore = { git = "https://github.com/bfoss765/rust-dashcore", rev = "b5dff6de05e4a354680e5b01a54bae9e642a6ad0" }
+dash-network-seeds = { git = "https://github.com/bfoss765/rust-dashcore", rev = "b5dff6de05e4a354680e5b01a54bae9e642a6ad0" }
+dash-spv = { git = "https://github.com/bfoss765/rust-dashcore", rev = "b5dff6de05e4a354680e5b01a54bae9e642a6ad0" }
+key-wallet = { git = "https://github.com/bfoss765/rust-dashcore", rev = "b5dff6de05e4a354680e5b01a54bae9e642a6ad0" }
+key-wallet-ffi = { git = "https://github.com/bfoss765/rust-dashcore", rev = "b5dff6de05e4a354680e5b01a54bae9e642a6ad0" }
+key-wallet-manager = { git = "https://github.com/bfoss765/rust-dashcore", rev = "b5dff6de05e4a354680e5b01a54bae9e642a6ad0" }
+dash-network = { git = "https://github.com/bfoss765/rust-dashcore", rev = "b5dff6de05e4a354680e5b01a54bae9e642a6ad0" }
+dashcore-rpc = { git = "https://github.com/bfoss765/rust-dashcore", rev = "b5dff6de05e4a354680e5b01a54bae9e642a6ad0" }
 
 tokio-metrics = "0.5"
 
--- a/packages/rs-platform-wallet/src/changeset/core_bridge.rs
+++ b/packages/rs-platform-wallet/src/changeset/core_bridge.rs
@@ -38,14 +38,48 @@ use key_wallet::transaction_checking::{DerivedAddressInfo, TransactionContext};
 use key_wallet::Utxo;
 use key_wallet_manager::{WalletEvent, WalletId, WalletManager};
 use tokio::sync::broadcast;
-use tokio::sync::broadcast::error::RecvError;
+use tokio::sync::broadcast::error::{RecvError, TryRecvError};
 use tokio::sync::RwLock;
 use tokio::task::JoinHandle;
 use tokio_util::sync::CancellationToken;
 
 use crate::changeset::changeset::{CoreChangeSet, HighestUsedIndexes, PlatformWalletChangeSet};
+use crate::changeset::merge::Merge;
 use crate::changeset::traits::PlatformWalletPersistence;
 use crate::wallet::platform_wallet::PlatformWalletInfo;
+
+/// Maximum number of `WalletEvent`s folded into a single
+/// `persister.store(..)` round-trip by [`run_wallet_event_adapter`].
+///
+/// # Why batch at all (dashpay/platform#4069 follow-up)
+///
+/// This adapter's per-event cost is overwhelmingly the persister call: on
+/// Android that is a JNI hop into a Room transaction (milliseconds), while
+/// projecting a `WalletEvent` into a [`CoreChangeSet`] is microseconds.
+/// Storing one event per `store()` therefore pinned the drain rate at
+/// roughly the *store* rate — low hundreds per second — which is far below
+/// what a historical SPV catch-up emits. The upstream producer publishes
+/// fire-and-forget onto a bounded ring (`DEFAULT_WALLET_EVENT_CAPACITY`,
+/// 1000), so a consumer that slow overflows it, `recv()` returns
+/// `Lagged`, and the durable-watermark guard freezes the wallet's
+/// `synced_height` for the rest of the process lifetime. In the field that
+/// presented as a sync that climbs toward completion and then "rolls back"
+/// to near the install position on every relaunch: the watermark froze
+/// after the first batch, so every later restart resumed from that same
+/// frozen height.
+///
+/// Folding every event *already buffered* in the ring into one changeset
+/// per wallet collapses a burst of N events into a single store, so the
+/// drain rate is bounded by the ring rather than by the persister. This is
+/// exactly the fold [`Merge`] was specified for — `CoreChangeSet` merging
+/// is commutative and associative, and its doc comment already anticipates
+/// "a flush can fold multiple events together (TransactionDetected +
+/// BlockProcessed for the same wallet over a sync round)".
+///
+/// The cap bounds the worst-case size of a single merged changeset (and
+/// hence one Room transaction), and keeps a saturated producer from
+/// starving the cancellation branch of the select below.
+const ADAPTER_STORE_BATCH_LIMIT: usize = 512;
 
 /// Session fault state for the durable-watermark guard
 /// (dashpay/platform#4069).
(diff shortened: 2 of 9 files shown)


thepastaclaw:
⛔ Blockers found — Opus deferred (commit ce9cc1ac)

Cargo.toml
```diff
@@ -52,14 +52,14 @@ members = [
...
-dashcore-rpc = { git = "https://github.com/dashpay/rust-dashcore", rev = "70d4bf8e36057c58e02d56769a6e9760f701dd06" }
+dashcore = { git = "https://github.com/bfoss765/rust-dashcore", rev = "b5dff6de05e4a354680e5b01a54bae9e642a6ad0" }
+dash-network-seeds = { git = "https://github.com/bfoss765/rust-dashcore", rev = "b5dff6de05e4a354680e5b01a54bae9e642a6ad0" }
+dash-spv = { git = "https://github.com/bfoss765/rust-dashcore", rev = "b5dff6de05e4a354680e5b01a54bae9e642a6ad0" }
+key-wallet = { git = "https://github.com/bfoss765/rust-dashcore", rev = "b5dff6de05e4a354680e5b01a54bae9e642a6ad0" }
+key-wallet-ffi = { git = "https://github.com/bfoss765/rust-dashcore", rev = "b5dff6de05e4a354680e5b01a54bae9e642a6ad0" }
+key-wallet-manager = { git = "https://github.com/bfoss765/rust-dashcore", rev = "b5dff6de05e4a354680e5b01a54bae9e642a6ad0" }
+dash-network = { git = "https://github.com/bfoss765/rust-dashcore", rev = "b5dff6de05e4a354680e5b01a54bae9e642a6ad0" }
+dashcore-rpc = { git = "https://github.com/bfoss765/rust-dashcore", rev = "b5dff6de05e4a354680e5b01a54bae9e642a6ad0" }
```
thepastaclaw, line 62: **🔴 Blocking: Repin the temporary contributor-fork dependencies before merge**

All eight rust-dashcore workspace dependencies now use `bfoss765/rust-dashcore` at `b5dff6de05e4a354680e5b01a54bae9e642a6ad0`. The head commit explicitly describes this as a temporary pin and states that dashpay/rust-dashcore#924 must land; that producer PR is currently open and unmerged, with this revision as its contributor-fork head. Merging this exact head would make clean production builds depend on a contributor-controlled repository rather than the project's governed upstream. After #924 lands, repin these entries to the resulting revision in `https://github.com/dashpay/rust-dashcore` and update `Cargo.lock` in the same commit.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/changeset/core_bridge.rs
```diff
@@ -131,162 +157,205 @@ where
...
+                        hex::encode(wallet_id),
+                        e
+                    );
+                }
+                tracing::error!(
+                    wallet_id = %hex::encode(wallet_id),
+                    error = %e,
+                    "Persister rejected core changeset; freezing this wallet's sync watermark so the next scan re-persists the missing rows (dashpay/platform#4069)"
+                );
+            }
```
thepastaclaw: **🟡 Suggestion: Count the watermark as persisted only after the store succeeds**

`synced_height_persisted` is updated before `persister.store(...)` runs. When the store rejects that changeset, the batch log still reports the rejected height as persisted even though the code faults the wallet precisely because those rows and their watermark were not accepted. This contradicts the diagnostic's stated purpose of showing whether the durable watermark advanced. Capture the proposed height before moving `core`, but update the diagnostic only in the successful store arm.

```suggestion
            let synced_height = core.synced_height;
            let cs = PlatformWalletChangeSet {
                core: Some(core),
                ..PlatformWalletChangeSet::default()
            };
            match persister.store(wallet_id, cs) {
                Ok(()) => {
                    if let Some(h) = synced_height {
                        synced_height_persisted =
                            Some(synced_height_persisted.map_or(h, |cur| cur.max(h)));
                    }
                }
                Err(e) => {
                    // A rejected changeset means these rows are not on disk. Fault
                    // THIS wallet's watermark so it can't outrun them; the next
                    // scan re-emits and the idempotent upserts recover the state.
                    fault.fault_wallet(wallet_id, &sync_fault);
                    faulted_in_batch += 1;
                    // One-shot, unambiguous logcat marker via the `log` facade
                    // (android_logger forwards `log` to logcat; `tracing` may not).
                    if !freeze_logged {
                        freeze_logged = true;
                        log::error!(
                            "SYNC WATERMARK FROZEN: persister rejected a changeset for wallet {} ({}); \
                             its durable sync height is now held so the next scan re-persists the \
                             missing rows (dashpay/platform#4069). syncFaultDetected() is latched.",
                            hex::encode(wallet_id),
                            e
                        );
                    }
                    tracing::error!(
                        wallet_id = %hex::encode(wallet_id),
                        error = %e,
                        "Persister rejected core changeset; freezing this wallet's sync watermark so the next scan re-persists the missing rows (dashpay/platform#4069)"
                    );
                }
            }
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The lossless persistence-channel change and FFI query appear sound, but this exact head is not ready to merge because all rust-dashcore dependencies remain pinned to a contributor fork while the required upstream producer PR is still open. The new batch diagnostic also counts a rejected watermark as persisted, which makes the failure trace internally inconsistent.

### Review provenance

<details>

```

In `Cargo.toml`:
- [BLOCKING] Cargo.toml:55-62: Repin the temporary contributor-fork dependencies before merge
  All eight rust-dashcore workspace dependencies now use `bfoss765/rust-dashcore` at `b5dff6de05e4a354680e5b01a54bae9e642a6ad0`. The head commit explicitly describes this as a temporary pin and states that dashpay/rust-dashcore#924 must land; that producer PR is currently open and unmerged, with this revision as its contributor-fork head. Merging this exact head would make clean production builds depend on a contributor-controlled repository rather than the project's governed upstream. After #924 lands, repin these entries to the resulting revision in `https://github.com/dashpay/rust-dashcore` and update `Cargo.lock` in the same commit.

- [SUGGESTION] packages/rs-platform-wallet/src/changeset/core_bridge.rs:293-323: Count the watermark as persisted only after the store succeeds
  `synced_height_persisted` is updated before `persister.store(...)` runs. When the store rejects that changeset, the batch log still reports the rejected height as persisted even though the code faults the wallet precisely because those rows and their watermark were not accepted. This contradicts the diagnostic's stated purpose of showing whether the durable watermark advanced. Capture the proposed height before moving `core`, but update the diagnostic only in the successful store arm.
```
</details>

thepastaclaw (changes requested):

Both carried-forward prior findings remain valid at head 3c57cb1a706133bd6d672f664117b1e102522a5e: the contributor-fork dependency pin remains blocking, and the batch diagnostic still counts a rejected watermark as persisted. The latest delta only relocates the `log` dependency declaration and introduces no new findings; no current review reply resolves either prior thread.
Source: Codex reviewer backends `gpt-5.6-sol` (general) and `gpt-5.6-sol` (ffi-engineer); Codex verifier backend `gpt-5.6-sol`; the `openclaw-agent` coordinator is orchestration-only.

### Review provenance

🔴 1 blocking

packages/rs-platform-wallet/src/changeset/core_bridge.rs
bfoss765: Fixed in ce9cc1acc03085abad3fd4294917141efeb00791.

You were right, and it mattered more than the "suggestion" label suggests: we are reading these lines off a mainnet tester's logcat right now to decide whether the watermark is advancing, so a drain that logs `synced_height_persisted=Some(h)` while faulting the wallet *because* `h`'s rows were not accepted points the diagnosis at the wrong subsystem.

I took your fix and went one step further, because there was a second way the line under-reported: when the fail-closed guard strips `synced_height` from an already-faulted wallet, the old line just showed `persisted=None`, which is indistinguishable from a drain that carried no watermark at all.

The commit path is now `commit_batch(...) -> BatchDiagnostics`, separating the three fates a height can meet within one drain:

- `synced_height_persisted` — `store()` returned `Ok`. The only field that means the durable watermark advanced.
- `synced_height_frozen` — proposed by the batch, stripped by `freeze_synced_height_if_faulted` before it ever reached the store.
- `synced_height_rejected` — offered to the store, which returned an error.

Each is the monotonic max across the wallets in the drain, so a batch spanning a healthy wallet and a faulted one reports both instead of collapsing into a single number.

The guard itself is untouched — this is a reporting change only. `freeze_synced_height_if_faulted` still runs after the fold, per-wallet fault scoping is unchanged, and the one-shot `SYNC WATERMARK FROZEN` `log::error!` plus the `sync_fault` latch behave exactly as before (both asserted in the new tests).

Two related changes in the same line:

- Dropped the hardcoded `missed=0`. It reported a number the code never measured — the same defect class you flagged. The `wallet-event batch:` prefix testers grep for is unchanged, and nothing in the repo parses this line.
- Extracted `commit_batch` out of the async loop specifically so the diagnostic is unit-testable against a real `store()` rejection, rather than only reachable through the channel plumbing.

7 new tests drive the real `commit_batch` (production commit path, guard included): accepted / rejected / guard-stripped / watermark-only-stripped / mixed batch / monotonic max / exact line format. The regression test is mutation-verified — reintroducing the pre-fix ordering makes `rejected_store_is_not_reported_as_persisted` fail with `left: Some(500), right: None`, exactly the value you predicted.

`cargo test -p platform-wallet` green (538 + 9), rustfmt clean, and no new clippy findings in the touched file.

Cargo.toml
bfoss765: Acknowledged, and the finding is correct — but it resolves upstream rather than by a change on this branch, so the pin is deliberately unchanged.

State verified just now via the API:

- **dashpay/rust-dashcore#924** — the producer that adds the lossless `mpsc` persistence channel this consumer drains — is **OPEN**, base `dev`, head `bfoss765:fix/wallet-event-persistence-mpsc` @ `2af691ba`, `mergeStateStatus: BLOCKED`, `REVIEW_REQUIRED`.
- **dashpay/rust-dashcore#909** is also **OPEN** (`CHANGES_REQUESTED`); its content is carried in the same fork revision.

There is no governed-upstream revision to point at yet: both producer PRs originate from the contributor fork into `dashpay/rust-dashcore:dev`, so every revision containing `take_persistence_receiver()` exists only on the fork today. Repinning now would mean reverting this PR's root-cause fix back onto the lossy broadcast path that caused #4069 in the first place.

So this is a genuine merge blocker and should stay flagged — the resolution is sequencing, not a code change here. Once #909 and #924 land on `dashpay/rust-dashcore:dev`, this branch repins all eight workspace entries to `https://github.com/dashpay/rust-dashcore` at the merged revision, with `Cargo.lock` updated in the same commit. Happy to keep this thread open as that gate.

One note for whoever performs the repin, found while verifying the above: the pinned revision `b5dff6de` has **diverged** from #924's current head (20 ahead / 1 behind). #924 has since gained `f0ab7d89`, which makes the persistence channel opt-in — the send half is `None` until `take_persistence_receiver()` lazily creates it. This consumer already satisfies that contract (the manager takes the receiver exactly once, before it is published to producers), so no adaptation is needed here, but the repin should target the merged `dev` revision rather than a rebase of `b5dff6de`.

thepastaclaw (changes requested):

Carried-forward prior findings: the contributor-fork dependency pin remains a blocking merge gate, while the watermark diagnostic has been fixed. New latest-delta findings: none. Cumulative reconciliation is complete.
Source: reviewer backend `gpt-5.6-sol`; verifier backend `gpt-5.6-sol`; coordinator `cliproxy/gpt-5.6-sol` is orchestration-only and not review evidence.

### Review provenance

🔴 1 blocking

bfoss765:
Moved to #4315 - in-repo branch rebased onto v4.2-dev (post-#4305). The rust-dashcore#924 dependency is repinned to the same commit rebased onto the new key-wallet rev (rebase42/lossless-persistence-channel-on-916 @ d72e71bf87 on the fork); the fork-pin clears when #924 merges. Review history preserved here.

PR #4291, opened 2026-08-05
fix(dpp): make document type index update validation name-order independent

QuantumExplorer:
## Issue being fixed or feature implemented

`IndexLevel::validate_update` compares old/new index trees by `level_identifier` equality, but identifiers are assigned by an incrementing counter that walks the document type's `indices` — a `BTreeMap` keyed by index **name**. As a result, whether a semantically identical index change was rejected with a proper consensus error or with something opaque depended purely on where the changed index's name sorted relative to the document type's other indexes:

- A change whose name sorted **before** an existing index renumbered every level → rejected with a useless `"Invalid path"` message.
- The identical change under a **late-sorting** name passed the tree comparison, then hard-errored in the JSON-schema compatibility check (which has no rule for the `indices` keyword) → surfaced as an `InternalError` (code 1) tx result instead of a clean consensus error.
- Renames of an unchanged index and `unique`/`nullSearchable` flips (which the tree comparison is blind to) hit the same lottery.

Notably, **no index modification was ever acceptable end-to-end under v0** — everything that survived the tree diff was killed by the `indices` schema-compatibility hard error. The name only determined *which kind of failure* you got. Since paid-invalid transitions record their error codes in blocks, the outcome is consensus-visible and the fix must be protocol-gated.

## What was done?

- Added a **v1 generation** of `DocumentTypeRef::validate_update` (protocol v14) that replaces the `IndexLevel` tree comparison with a name-keyed comparison of the parsed index definitions: any added, removed or modified index is rejected with a deterministic `DataContractInvalidIndexDefinitionUpdateError` whose payload names the offending index (`added index 'x'` / `removed index 'x'` / `changed index 'x'`), independent of name sort order.
- The accepted set is **unchanged**: v1 rejects all index modifications, exactly like v0 did end-to-end — only the rejections become deterministic, clean, and correctly labeled. Chain history is unaffected: PV ≤ 13 replays through the byte-for-byte frozen v0 flow, and system-contract index rewrites (e.g. the DPNS v2 update at PV13) bypass `validate_update` entirely.
- Gated via new `DPP_VALIDATION_VERSIONS_V5` (`document_type.validate_update: 1`) wired into v14 only; v13 stays on V4.
- Moved the generation-independent helpers (`validate_config`, byte-array encoding stability, `validate_schema`) verbatim to a shared `common` module, following the `try_from_schema` extraction precedent (#4276). The v0 flow itself is untouched.

[...]

PR #4295, opened 2026-08-05
fix(dpp): stop hard-erroring on index-order-only contract updates

QuantumExplorer:

The JSON-schema compatibility validator has no keyword rule for `indices`, so any contract-update schema diff under `/indices` that survived the dedicated index checks returned `Err(UnsupportedSchemaKeywordError → SchemaCompatibilityValidationError)` instead of a clean consensus validation result. drive-abci surfaces that as `StateTransitionExecutionResult::InternalError` (code 1) for user-triggerable input.

At protocol v14 (after #4291's name-keyed index comparison) the reachable case is reordering the `indices` array without changing the definition set: the name-keyed comparison passes — a reorder is a semantic no-op, since indices are keyed by name — but the JSON diff under `/indices` still hard-errored.

## What was done?

- Added `validate_schema_compatibility` v1 (`packages/rs-dpp/src/data_contract/document_type/schema/validate_schema_compatibility/v1/mod.rs`): a new generation that strips the top-level `indices` key from both schemas before diffing. Index definitions are validated in exactly one place — `validate_update` v1's name-keyed comparison, which rejects any real index change with a clean `DataContractInvalidIndexDefinitionUpdateError` before schema compatibility runs. A *property* named `indices` (under `/properties/indices`) is still validated.
- Wired the new generation into the dispatcher (`validate_schema_compatibility/mod.rs`) and bumped `validate_schema_compatibility: 1` in `CONTRACT_VERSIONS_V6`, which only protocol v14 uses — the shipped v0 generation stays byte-identical and protocol ≤13 behavior is unchanged.
- Documented the gate in the `CONTRACT_VERSIONS_V6` and `PLATFORM_V14` changelog comments.

Stripping was chosen over an "any `/indices` change is incompatible" rule because the only diff that can reach this check at v14 is a reorder-only update, which is harmless and should be accepted rather than cleanly rejected.

## How Has This Been Tested?

- `validate_update/v1/mod.rs`: `should_pass_when_indices_are_reordered_without_changes` — reorder-only contract update through the public `validate_update` dispatcher at the latest protocol version now returns a valid result instead of an `Err`.
- `validate_schema_compatibility/v1/mod.rs`: reorder-only diff is ignored; an incompatible property change alongside an `/indices` diff is still reported; a property literally named `indices` is still validated; replay-safety pin that protocol v13 (v0) still hard-errors on an `/indices` diff with the exact unsupported-keyword message.
- `cargo test -p dpp` (full suite), `cargo test -p platform-version`, `cargo check -p drive-abci`, `cargo clippy -p dpp -p platform-version --all-targets`, `cargo check -p dpp --all-features --tests` and `--no-default-features --tests` — all green.

## Breaking Changes

None. Gated at protocol v14 (unreleased); v0 stays byte-identical and pre-v14 outcomes are pinned by tests. Note this makes a reorder-only update the first index-touching contract update ever accepted end-to-end — previously every such update was rejected one way or another, so no historical accepted-set changes.

## Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
  - Schema compatibility checks now allow document indexes to be reordered without incorrectly rejecting otherwise compatible updates.
  - Genuine schema incompatibilities and index changes continue to be detected and reported.
  - Properties named `indices` within document schemas remain fully validated.
  - Document update validation now confirms reordered unchanged indexes are accepted.

- **Documentation**
  - Updated protocol documentation to reflect the improved schema compatibility behavior and version transition.


thepastaclaw:
🔍 Review in progress — actively reviewing now (commit 3a3f4dcc)
_ETA: complete ~08:57 UTC (median 22m across 30 recent reviews)_
_Running 5m · Last checked: 2026-08-05 08:40 UTC_

packages/rs-platform-version/src/version/v14.rs
```diff
@@ -72,6 +72,12 @@ pub const PROTOCOL_VERSION_14: ProtocolVersion = 14;
 ///   (`rankedCountable` / `rankedSummable` / `rankedAverageable`). v13 keeps
 ///   validating against meta-schema v2, where those keys are rejected as
 ///   unknown properties, so a pre-v14 contract cannot smuggle them in.
+///   It also bumps `validate_schema_compatibility` to 1, which strips the
+///   top-level `indices` key before diffing the old and new document type
+///   schemas: index immutability is enforced by `validate_update` v1's
+///   name-keyed comparison, so a contract update that merely reorders the
+///   `indices` array validates cleanly instead of hitting the
+///   unsupported-keyword hard error (an internal error under v13).
```
thepastaclaw, line 91: **💬 Nitpick: "Only v14 behavior change" framing is now contradicted by the bullet this PR adds**

The topic sentence at line 67-68 says "Until a contract uses the ranked grammar, the only v14 behavior change is the shared-prefix fix; everything else matches v13." The PR then adds lines 75-80 to the very next bullet describing the `validate_schema_compatibility` v1 change (stripping `indices` before diffing), which applies to *any* contract update that reorders `indices` — independent of the ranked grammar or the shared-prefix fix. That's a third, orthogonal behavior change, so the topic sentence is now inaccurate. This file is treated as authoritative behavior-change documentation for consensus versioning in this codebase, so the wording should be updated to acknowledge the schema-compatibility fix as a separate v14 change rather than folding it silently under "shared-prefix fix."

```suggestion
/// Until a contract uses the ranked grammar, the only ranked-adjacent v14
/// behavior changes are the shared-prefix fix and the schema-compatibility
/// index-reorder fix below; everything else matches v13:
```

<sub>source: unknown</sub>

packages/rs-dpp/src/data_contract/document_type/schema/validate_schema_compatibility/v1/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+use crate::ProtocolError;
+use json_schema_compatibility_validator::{
+    validate_schemas_compatibility, CompatibilityRulesCollection, Options,
+    KEYWORD_COMPATIBILITY_RULES,
+};
+use once_cell::sync::Lazy;
+use std::borrow::Cow;
+use std::ops::Deref;
+
+static OPTIONS: Lazy<Options> = Lazy::new(|| {
```
thepastaclaw, line 33: **💬 Nitpick: `OPTIONS` static and result-mapping logic duplicated verbatim from v0**

`v1/mod.rs` re-declares the entire `OPTIONS` `Lazy<Options>` (including the two `.expect()` calls tied to the `required` compatibility rule) and re-implements the identical `validate_schemas_compatibility(...).map(...).map_err(...)` pipeline already present in `v0/mod.rs`, with the only functional delta being the `without_indices` preprocessing step. Some duplication is expected under this codebase's "each generation is a byte-identical snapshot" convention, but `OPTIONS` construction and the result-mapping closure are pure implementation detail, not consensus-pinned wording — factoring them into a shared helper both generations call would remove ~30 duplicated lines and one drift risk (a future fix to the `required`-rule override would otherwise need to be applied by hand in two places).

<sub>source: ['claude']</sub>

packages/rs-dpp/src/data_contract/document_type/schema/validate_schema_compatibility/v1/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+
+        let error = validate_schema_compatibility(&original_schema, &new_schema, platform_version)
+            .expect_err("an indices diff must hard-error under v0");
+
+        assert_matches!(
+            error,
+            ProtocolError::DataContractError(DataContractError::JsonSchema(
+                JsonSchemaError::SchemaCompatibilityValidationError(message)
+            )) if message == "schema keyword 'indices' at path '/indices/0/unique' is not supported"
+        );
```
thepastaclaw, line 250: **💬 Nitpick: Replay-safety pin asserts an exact error string owned by another workspace crate**

`v0_should_error_on_indices_diff` matches on the literal message `"schema keyword 'indices' at path '/indices/0/unique' is not supported"`, which is produced by `rs-json-schema-compatibility-validator`'s `Display` impl for `UnsupportedSchemaKeywordError` (confirmed in `packages/rs-json-schema-compatibility-validator/src/error.rs:55`), not by code in this crate. If that crate's error-formatting wording ever changes for unrelated reasons, this test breaks even though v0's actual behavior (still hard-erroring on an `/indices` diff) hasn't changed. Since the intent is a replay-safety pin on *behavior* rather than exact wording, matching on the `Error::UnsupportedSchemaKeyword(_)` variant (or asserting a substring like `message.contains("indices")`) would be more robust while still catching a real regression.

<sub>source: ['claude']</sub>

packages/rs-dpp/src/data_contract/document_type/schema/validate_schema_compatibility/v1/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+                .collect::<Vec<_>>();
+
+            SimpleValidationResult::new_with_errors(errors)
+        })
+        .map_err(|error| {
+            ProtocolError::DataContractError(DataContractError::JsonSchema(
+                JsonSchemaError::SchemaCompatibilityValidationError(error.to_string()),
+            ))
+        })
+}
```
thepastaclaw, line 94: **🟡 Suggestion: v1's correctness relies on an unenforced cross-version pairing invariant**

`validate_schema_compatibility_v1`'s safety (unconditionally stripping `indices` is fine) depends on `DocumentTypeRef::validate_update` — gated by the *separately versioned* `platform_version.dpp.validation.document_type.validate_update` field — always rejecting any real index change before this check runs, for whichever `validate_update` generation is paired with schema-compat v1 in a given `PlatformVersion`. I confirmed this holds today: v13 pairs `validate_update` v0 with schema-compat v0, and v14 pairs `validate_update` v1 with schema-compat v1 (`DPP_VALIDATION_VERSIONS_V5` + `CONTRACT_VERSIONS_V6`). But nothing in the code ties these two independent version numbers together — a future protocol version that bumps schema-compat without also bumping (or with a looser) `validate_update` would silently let real index changes bypass compatibility checking, since `indices` is unconditionally stripped. This isn't a defect in this PR (the current wiring is correct and tested), but a short code comment on `validate_schema_compatibility_v1` calling out the dependency on `validate_update` v1's guarantee would help future maintainers avoid breaking the pairing.

<sub>source: ['claude']</sub>

thepastaclaw (commented):

This PR adds a well-scoped, protocol-v14-gated `validate_schema_compatibility` v1 that strips the top-level `indices` key before diffing document-type schemas, closing a real gap where an index-order-only contract update hard-errored into an internal error. I verified the version wiring end-to-end (v13 = validate_update v0 + schema-compat v0, v14 = validate_update v1 + schema-compat v1), confirmed the `$defs` call site is unaffected (no top-level `indices` key there) and that a genuine `/properties/indices` property is still validated, and ran the new tests locally — all pass. No blocking issues; remaining findings are minor documentation/duplication/test-robustness nitpicks.

### Review provenance
- Sonnet reviewers: `claude-sonnet-5` — general (failed), `claude-sonnet-5` — security-auditor (failed), `claude-sonnet-5` — rust-quality (completed), `claude-sonnet-5` — ffi-engineer (completed), `claude-sonnet-5` — general (completed), `claude-sonnet-5` — security-auditor (completed)

🟡 1 suggestion(s) | 💬 3 nitpick(s)

<details>

```

In `packages/rs-dpp/src/data_contract/document_type/schema/validate_schema_compatibility/v1/mod.rs`:
- [SUGGESTION] packages/rs-dpp/src/data_contract/document_type/schema/validate_schema_compatibility/v1/mod.rs:1-87: v1's correctness relies on an unenforced cross-version pairing invariant
  `validate_schema_compatibility_v1`'s safety (unconditionally stripping `indices` is fine) depends on `DocumentTypeRef::validate_update` — gated by the *separately versioned* `platform_version.dpp.validation.document_type.validate_update` field — always rejecting any real index change before this check runs, for whichever `validate_update` generation is paired with schema-compat v1 in a given `PlatformVersion`. I confirmed this holds today: v13 pairs `validate_update` v0 with schema-compat v0, and v14 pairs `validate_update` v1 with schema-compat v1 (`DPP_VALIDATION_VERSIONS_V5` + `CONTRACT_VERSIONS_V6`). But nothing in the code ties these two independent version numbers together — a future protocol version that bumps schema-compat without also bumping (or with a looser) `validate_update` would silently let real index changes bypass compatibility checking, since `indices` is unconditionally stripped. This isn't a defect in this PR (the current wiring is correct and tested), but a short code comment on `validate_schema_compatibility_v1` calling out the dependency on `validate_update` v1's guarantee would help future maintainers avoid breaking the pairing.
```
</details>

packages/rs-platform-version/src/version/v14.rs
QuantumExplorer: Fixed in eac2ac02df — the topic sentence now names both pre-ranked v14 behavior changes (shared-prefix fix and the index-reorder schema-compatibility fix).

packages/rs-dpp/src/data_contract/document_type/schema/validate_schema_compatibility/v1/mod.rs
QuantumExplorer: Added in eac2ac02df — a doc comment on validate_schema_compatibility_v1 states the pairing invariant (schema-compat v1 requires validate_update >= 1 in the same version table) and the consequence of bumping one without the other.

packages/rs-dpp/src/data_contract/document_type/schema/validate_schema_compatibility/v1/mod.rs
QuantumExplorer: Not changed: per this repo's versioned-generation convention, shipped vN files stay byte-identical and new generations are clones — extracting OPTIONS/result-mapping into a shared helper would modify the shipped v0 module, which is exactly what the convention avoids.

packages/rs-dpp/src/data_contract/document_type/schema/validate_schema_compatibility/v1/mod.rs
QuantumExplorer: Not changed: the exact-message pin deliberately matches the style of the existing v0 pin from #4291 (validate_update/v0/mod.rs) — if the validator crate's Display wording changes, both pins fail together and get updated together, and the exact message also pins the keyword and path.

QuantumExplorer:
Re @coderabbitai's suggestion to pin the `validate_schema_compatibility` v1 tests to `PlatformVersion::get(14)`: not changed — this repo's convention is that tests for the current (tip) generation use `PlatformVersion::latest()`, and only tests pinning a *frozen older* generation name an explicit protocol version (see the v0 pin `v0_should_error_on_indices_diff`, which uses `PlatformVersion::get(13)`). When a future protocol version introduces a v2 generation, the latest-PV tests move with the tip and the v1 behavior gets its own explicit pin at that point.

PR #4256, opened 2026-07-31
feat(platform-wallet): token-minting finalize from a funding path (spendable DashPay receival accounts)

bfoss765:
## What this does

Bridges two flows that were previously disconnected in `platform-wallet`:

- `build_signed_payment`'s **single-account-by-derivation-path** selection (#4247) — the only build path that can reach a `DashpayReceivingFunds` account; and
- the **reservation/broadcast** flow (#4185) — `token → broadcastSigned` / `releaseReservation`.

New `finalize_signed_payment_from_funding_path` returns a `FinalizedCorePayment` (signed tx, fee, change, the *resolved* `FundingAccountRef`, and the reservation token). The key correctness property: the reservation is recorded against the **resolved** account, never a BIP44 default, so release/broadcast bookkeeping lands in the same ledger the build reserved into. `FundingAccountRef::Path` makes non-standard accounts (DashPay receival) nameable to the release machinery, which previously could only address `StandardAccountType`s.

Surface:

- `wallet/core/transaction.rs` — `FundingAccountRef` enum + `release_reservation_for`
- `wallet/core/send.rs` — `finalize_signed_payment_from_funding_path`, `FinalizedCorePayment`, `abandon_payment`
- `wallet/signed_payment_registry.rs` — `register_funded_by`
- FFI `core_wallet_build_signed_payment_with_token`
- JNI `coreWalletBuildSignedPaymentWithToken`
- Kotlin `ManagedPlatformWallet.buildSignedPaymentWithToken(recipients, coreSignerHandle, feePerKb, fundingPath)`, returning the existing `SignedCoreTransaction` (now carrying `changeDuffs`)

Guardrails preserved and covered by tests: single-account funding only (the funding-privacy guardrails still pass — no cross-account union), change goes to BIP44/0, watch-only accounts are refused, and the fee is taken from the signed transaction rather than re-estimated.

## Stacked on

**Stacked on #4185, and includes #4247's commits.**

The branch is based on #4185's head (`port/v4.1/split-build-broadcast`) with #4247's three commits (`port/v4.1/send-raw-tx`) cherry-picked beneath this change, because it needs both:

- **#4185** supplies the signed-payment registry and the `bfoss765/rust-dashcore` pin at rev `e99959ce` — the `ReservationToken` / `build_unsigned_reserved` API this work is built on.
- **#4247** supplies the `build_signed_payment(funding_path)` primitive itself, which is *not* present on #4185's branch.

[...]

PR #4196, opened 2026-07-21
fix(platform-wallet): age-guard the V2 finalized-transaction handle broadcast (stacked on #4185)

bfoss765:
**Stacked on #4185 (split build/broadcast) — merge after it.** The branch includes #4185's commits until it lands.

Follow-up to #4185 (requested by @shumkov): the V2 finalized-transaction
handle surface (`core_wallet_tx_builder_finalize` →
`broadcast_finalized_transaction`) retained a stale-release hazard — a pinned
V2 handle had no age guard, so a long-held handle could broadcast against
funding inputs that key-wallet's `ReservationSet` TTL sweep may already have
released and re-selected for an unrelated build. This goes live the moment iOS
starts issuing deferred sends.

This mirrors the deferred registry-token age policy on the V2 handle path:

- **Shared bound.** `RESERVATION_MAX_AGE_BLOCKS` (20; key-wallet TTL 24) and
  `reservation_expired()` are hoisted to `wallet::reservations`, so the
  registry and the V2 handle path measure a reservation's age against the same
  number.
- **Guarded op.** `broadcast_finalized_transaction` refuses — *before*
  touching the broadcaster — once `current_height − reservation_height >=` the
  shared bound, using the reservation's own stamp height already carried on
  `SignedCoreTransaction::reservation_height`. The stale reservation is left
  for key-wallet's TTL to reclaim (never released by outpoint). The check runs
  after the existing generation-identity check, matching the registry order.
- **Error code.** A new token-less `PlatformWalletError::StaleReservation`
  reuses the existing FFI `ErrorStaleReservationToken` (26); no new code is allocated
  (27/28 untouched; 29/30 remain reserved for the asset-lock PR #4184's branch)
  (asset-lock PR) touched. Reuse is documented on both sides.
- **Abandon/free stay unguarded** — releasing an aged reservation is always
  safe.

Tests: fresh handle broadcasts; aged handle refuses with `StaleReservation`
yet abandons cleanly and frees its inputs; exact threshold boundary
(BIP44/BIP32); FFI mapping to the shared code.

Stacked on `port/v4.1/split-build-broadcast` (#4185) — review/merge that
first.

PR #4299, opened 2026-08-05
chore(platform-wallet): bump key-wallet pin and migrate to new AddressState/asset-lock API

bfoss765:
## What & why

This is **baseline maintenance**, not a feature. Repointing the workspace's
`rust-dashcore` pins to the canonical `dev` commit that carries
key-wallet #916 (`8f78baa6`, the tip #4185 already points to) unavoidably
also pulls in two older, already-merged `dev` PRs the `v4.2-dev` platform
baseline predates. Their key-wallet API changes leave the baseline calling
the old API, so `v4.2-dev` itself no longer compiles against the new
key-wallet. That inherited break — not anything in the feature PRs — is why
**#4185 / #4247 / #4256 / #4196 are CI-red**. This PR fixes the baseline so
they inherit a green tree once rebased.

Pin change: all 8 `rust-dashcore` crates `70d4bf8e` → `8f78baa6`
(identical to what #4185 pins).

Surfaced upstream changes:
- **#818** — `AddressInfo`'s flat `used` / `generated_at` / `used_at` fields
  became `state: AddressState { Available | Reserved { at } | Used }`.
- **#915** — `build_asset_lock[_with_signer]` gained an explicit
  `AssetLockFundingAccount` funding source and a `drain` flag.
- **#916** — `TransactionBuilder::build_unsigned` was replaced by
  `build_unsigned_reserved` (additionally returns an `Option<ReservationToken>`).

## #818 — AddressState (mechanical, behavior-preserving)

`used == true` → `AddressState::Used`; unused → `AddressState::Available`.
Platform never hands out the `Reserved` state — it is written only by
`AddressPool::next_unused_and_reserve` and
`ManagedCoreFundsAccount::next_receive_address_and_reserve`, and the
workspace has zero callers of either — so `Available`/`Used` is the full
reachable range.

The removed `generated_at` / `used_at` timestamps have **no equivalent** on
the new variants (`Used` carries no timestamp), and dropping them is
lossless: they were never persisted (`CoreAddressEntryFFI` has no field for
either, and `address_info_from_ffi` was their only producer, from the
literals `0` / `Some(0)`), and upstream never had real values either —
every key-wallet `AddressInfo` constructor set `generated_at: 0`
("Should use actual timestamp").

`Reserved { at }` *is* unrepresentable in the persisted address schema and
would reload as `Available`. That is unreachable today per the above, so
rather than change the FFI schema in a baseline PR the emit path now spells
the `Reserved` arm out and warns instead of silently flattening it, the
schema field documents the constraint, and a test pins the behavior so the
first reserving caller lands on an explicit schema decision.

15 sites across 6 files:
- `rs-platform-wallet/src/manager/accessors.rs` — 2 (keys-used count; `is_used` snapshot)
- `rs-platform-wallet/src/changeset/core_bridge.rs` — 1 prod write + 2 test asserts
- `rs-platform-wallet/src/wallet/provider_key_at_index.rs` — 1 construct (`Available`)
- `rs-platform-wallet/src/wallet/asset_lock/build.rs` — 2 test reads
- `rs-platform-wallet/src/changeset/changeset.rs` — 1 test stub
- `rs-platform-wallet-ffi/src/persistence.rs` — 1 prod read + 1 prod round-trip construct + 1 prod pool-restore read + 3 test sites
- `swift-sdk/Sources/SwiftDashSDK/KeyWallet/AddressPool.swift` — 1 read of the
  removed `generated_at` on the regenerated cbindgen header (would have broken
  the Apple framework build; see the review reply on `f49f49e0`)

## #916 — build_unsigned_reserved (behavior-preserving)

`build_unsigned_reserved` has the **same body** as the old `build_unsigned`,
only additionally surfacing the token `assemble_unsigned` already produced
(mirroring how the retained `build_signed` wraps `build_signed_reserved` and
discards the token). The baseline releases abandoned reservations via the
**unchanged** unconditional `release_reservation(&tx)`, so the token is
discarded here to preserve pre-#916 behavior. Owner-guarded release via the
token is the TOCTOU fix threaded by **#4185**, not baseline work.
1 site: `rs-platform-wallet/src/wallet/core/transaction.rs`.

## Assumptions needing maintainer / QE confirmation

- **#915 `drain` flag → defaulted to `false`.** The single call site
  (`rs-platform-wallet/src/wallet/asset_lock/build.rs`) funds from the
  standard BIP44 account (its `account_index` is documented as a BIP44
  account index), so it maps to `AssetLockFundingAccount::Bip44 { account_index }`.
  `drain = false` reproduces the pre-#915 non-drain behavior that the #915
  commit itself documents as the historical default
  ("keeps its behavior (BIP44, non-drain)"). Marked in-code with
  `// TODO(baseline-bump): confirm drain intent with QE`. **Please confirm**
  this platform asset-lock path is intended to remain non-drain BIP44.

## Verification

- `cargo check --workspace --all-targets` — clean (only the pre-existing
  `rs-dpp` `default-features` note).
- `platform-wallet` lib tests: 513 passed / 0 failed.
- `platform-wallet-ffi` lib tests: 221 passed / 0 failed.
- rustfmt clean; scoped clippy exactly as `tests-rs-wallet.yml` runs it
  (`--all-features --locked -- --no-deps -D warnings`) exits 0.
- **Swift:** the whole `SwiftDashSDK` module type-checks clean under
  `-swift-version 6 -warnings-as-errors` against the cbindgen headers
  regenerated from the new pin, using the same umbrella header + modulemap
  `build_ios.sh` injects. The `Swift SDK build` CI job is gated to
  `dashpay/platform` and `thepastaclaw`-owned forks, so it skips here.
- **#4185 on top:** merging this baseline into #4185's head produced a single
  conflict — `transaction.rs`, where #4185's token-threading rewrite
  supersedes the minimal baseline shim (the expected rebase outcome). With
  #4185's `transaction.rs` kept, the merged tree compiles clean, confirming
  #4185 inherits the fix. #4247 / #4256 / #4196 pin the identical `8f78baa6`
  and carry the same unmigrated baseline sites, so they inherit the same fix.

## Scope

Does **not** touch #4185 / #4247 / #4256 / #4196 — they inherit this once it
lands in `v4.2-dev`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved wallet address-state handling across available, used, and reserved addresses.
  * Address usage counts and funded-address tracking now report more accurately.
  * Address pools now restore saved states consistently between sessions.
  * Transaction building now handles temporary address reservations reliably.

* **Documentation**
  * Clarified address usage and reservation behavior, including compatibility limitations for reservation details.
  * Updated address timestamp documentation for compatibility with existing integrations.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1229,7 +1229,7 @@ version = "3.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "faf9468729b8cbcea668e36183cb69d317348c2e08e994829fb56ebfdfbaac34"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -1662,7 +1662,7 @@ dependencies = [
 [[package]]
 name = "dash-network"
 version = "0.45.0"
-source = "git+https://github.com/dashpay/rust-dashcore?rev=70d4bf8e36057c58e02d56769a6e9760f701dd06#70d4bf8e36057c58e02d56769a6e9760f701dd06"
+source = "git+https://github.com/dashpay/rust-dashcore?rev=8f78baa6b7979b9bea56501ad75b5a7b7150a711#8f78baa6b7979b9bea56501ad75b5a7b7150a711"
 dependencies = [
  "bincode",
  "bincode_derive",
@@ -1673,7 +1673,7 @@ dependencies = [
 [[package]]
 name = "dash-network-seeds"
 version = "0.45.0"
-source = "git+https://github.com/dashpay/rust-dashcore?rev=70d4bf8e36057c58e02d56769a6e9760f701dd06#70d4bf8e36057c58e02d56769a6e9760f701dd06"
+source = "git+https://github.com/dashpay/rust-dashcore?rev=8f78baa6b7979b9bea56501ad75b5a7b7150a711#8f78baa6b7979b9bea56501ad75b5a7b7150a711"
 dependencies = [
  "dash-network",
 ]
@@ -1750,7 +1750,7 @@ dependencies = [
 [[package]]
 name = "dash-spv"
 version = "0.45.0"
-source = "git+https://github.com/dashpay/rust-dashcore?rev=70d4bf8e36057c58e02d56769a6e9760f701dd06#70d4bf8e36057c58e02d56769a6e9760f701dd06"
+source = "git+https://github.com/dashpay/rust-dashcore?rev=8f78baa6b7979b9bea56501ad75b5a7b7150a711#8f78baa6b7979b9bea56501ad75b5a7b7150a711"
 dependencies = [
  "async-trait",
  "chrono",
@@ -1779,7 +1779,7 @@ dependencies = [
 [[package]]
 name = "dashcore"
 version = "0.45.0"
-source = "git+https://github.com/dashpay/rust-dashcore?rev=70d4bf8e36057c58e02d56769a6e9760f701dd06#70d4bf8e36057c58e02d56769a6e9760f701dd06"
+source = "git+https://github.com/dashpay/rust-dashcore?rev=8f78baa6b7979b9bea56501ad75b5a7b7150a711#8f78baa6b7979b9bea56501ad75b5a7b7150a711"
 dependencies = [
  "anyhow",
  "base64-compat",
@@ -1805,12 +1805,12 @@ dependencies = [
 [[package]]
 name = "dashcore-private"
 version = "0.45.0"
-source = "git+https://github.com/dashpay/rust-dashcore?rev=70d4bf8e36057c58e02d56769a6e9760f701dd06#70d4bf8e36057c58e02d56769a6e9760f701dd06"
+source = "git+https://github.com/dashpay/rust-dashcore?rev=8f78baa6b7979b9bea56501ad75b5a7b7150a711#8f78baa6b7979b9bea56501ad75b5a7b7150a711"
 
 [[package]]
 name = "dashcore-rpc"
 version = "0.45.0"
-source = "git+https://github.com/dashpay/rust-dashcore?rev=70d4bf8e36057c58e02d56769a6e9760f701dd06#70d4bf8e36057c58e02d56769a6e9760f701dd06"
+source = "git+https://github.com/dashpay/rust-dashcore?rev=8f78baa6b7979b9bea56501ad75b5a7b7150a711#8f78baa6b7979b9bea56501ad75b5a7b7150a711"
 dependencies = [
  "dashcore-rpc-json",
  "hex",
@@ -1823,7 +1823,7 @@ dependencies = [
 [[package]]
 name = "dashcore-rpc-json"
 version = "0.45.0"
-source = "git+https://github.com/dashpay/rust-dashcore?rev=70d4bf8e36057c58e02d56769a6e9760f701dd06#70d4bf8e36057c58e02d56769a6e9760f701dd06"
+source = "git+https://github.com/dashpay/rust-dashcore?rev=8f78baa6b7979b9bea56501ad75b5a7b7150a711#8f78baa6b7979b9bea56501ad75b5a7b7150a711"
 dependencies = [
  "bincode",
  "dashcore",
@@ -1838,7 +1838,7 @@ dependencies = [
 [[package]]
 name = "dashcore_hashes"
 version = "0.45.0"
-source = "git+https://github.com/dashpay/rust-dashcore?rev=70d4bf8e36057c58e02d56769a6e9760f701dd06#70d4bf8e36057c58e02d56769a6e9760f701dd06"
+source = "git+https://github.com/dashpay/rust-dashcore?rev=8f78baa6b7979b9bea56501ad75b5a7b7150a711#8f78baa6b7979b9bea56501ad75b5a7b7150a711"
 dependencies = [
  "bincode",
  "dashcore-private",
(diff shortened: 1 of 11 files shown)


thepastaclaw:
ℹ️ Review skipped (commit f49f49e0)
_Last checked: 2026-08-05 18:30 UTC_

thepastaclaw (changes requested):

The Rust-side baseline migrations preserve the parent commit's behavior, and the owner-token cleanup work is explicitly deferred to the stacked #4185 change. However, the key-wallet-ffi pin removes a field still referenced by the Swift wrapper, so regenerating the Apple framework makes the Swift SDK fail to compile.

### Review provenance

🔴 1 blocking

<details>

```

In `packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/AddressPool.swift`:
- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/AddressPool.swift:117: Update the Swift wrapper for the removed FFI timestamp field
  The newly pinned key-wallet-ffi revision removes `generated_at` and `used_at` from the `#[repr(C)] FFIAddressInfo` structure, while `build_ios.sh` regenerates and includes that crate's header in `DashSDKFFI.xcframework`. This initializer still accesses `ffiInfo.generated_at`, so a fresh framework build leaves the Swift package referencing a nonexistent member and compilation fails. The old address-pool constructors populated this timestamp with the synthetic value zero, so assigning the Unix epoch preserves the previous observable value without breaking the public Swift property.
```
</details>

bfoss765:
Blocking finding fixed in `f49f49e01846d6d2fc700e90187f81a422370da6`.

### The break

`packages/swift-sdk/Sources/SwiftDashSDK/KeyWallet/AddressPool.swift:117` read `ffiInfo.generated_at`. key-wallet #818 removed `generated_at` and `used_at` from `#[repr(C)] FFIAddressInfo` (`key-wallet-ffi/src/address_pool.rs`), and `build_ios.sh` regenerates that crate's cbindgen header into `DashSDKFFI.xcframework`, so a fresh framework build failed with:

```
AddressPool.swift:144:77: error: value of type 'FFIAddressInfo' has no member 'generated_at'
```

This was the only Swift reference to a removed member — `used` is still emitted by the new header (now derived as `info.is_used()`), and the other two upstream deltas in this pin (`error.rs`, `transaction.rs`) do not change the C surface.

### The fix

`generatedAt` is kept as a public property and pinned to the Unix epoch. That is not a substituted value: every `AddressInfo` constructor in key-wallet set `generated_at: 0` (annotated *"Should use actual timestamp"*), so the property has always evaluated to the epoch — the observable value is byte-identical to before the bump. Removing it would be a Swift API break beyond a baseline pin bump; the doc comment states plainly that it is a retained placeholder, not a real generation time.

`used` also gained a doc comment: it now means exactly `state == Used`, so a *reserved* address reports `used == false`. The C struct carries no reservation flag, so reservation state is not observable from Swift at all; exposing it would need a new field in key-wallet-ffi upstream first, which is out of scope here.

### On the persistence round-trip (raised as a concern during triage) — definitively: no loss

`generated_at` / `used_at` were **never serialized**. `CoreAddressEntryFFI` (`core_address_types.rs`) has no field for either; `address_info_from_ffi` was the only thing that produced them, from the literals `0` / `Some(0)`. A save/load cycle already destroyed them before this PR. Dropping them is lossless.

`Reserved { at }` is a real gap in that schema — `is_used` is a single bool, so a reserved address would reload as `Available` and could be handed out twice. It is **unreachable today**: `AddressState::Reserved` is written only by `AddressPool::next_unused_and_reserve` and `ManagedCoreFundsAccount::next_receive_address_and_reserve`, and the workspace has zero callers of either. So this is a boundary, not a live bug — but it is no longer silent:

- the emit path spells the `Reserved` arm out and `tracing::warn!`s instead of folding it into a catch-all,
- `CoreAddressEntryFFI::is_used` documents the constraint (its old doc referenced the now-deleted `AddressInfo.used`),
- new test `address_state_round_trip_keeps_used_and_flattens_reserved` pins the behavior, so the first reserving caller lands on an explicit schema decision.

### Other diff claims, re-verified against both pins

- **#916** — `build_unsigned_reserved`'s body is identical to the old `build_unsigned` plus surfacing the token `assemble_unsigned` already produced. Discarding it preserves pre-#916 behavior.
- **#915** — old `build_asset_lock_with_signer` took `account_index: u32` and resolved it via `wallet.get_bip44_account(account_index)`; the new `AssetLockFundingAccount::Bip44 { account_index }` resolves through the same call. `drain: false` matches what key-wallet-ffi's own migration of this API chose. The BIP44/non-drain mapping is exact — the open item is still just QE confirming that non-drain is *intended* going forward.

### Verification

- `cargo check --workspace --all-targets` — clean (only the pre-existing `rs-dpp` `default-features` note).
- `platform-wallet` 513 passed / 0 failed; `platform-wallet-ffi` **221** passed / 0 failed (was 220; +1 is the new round-trip test).
- `cargo fmt --check` clean; scoped clippy exactly as `tests-rs-wallet.yml` runs it (`--all-features --locked -- --no-deps -D warnings`) exits 0.
- **Swift**: the whole `SwiftDashSDK` module type-checks clean under `-swift-version 6 -warnings-as-errors` against the cbindgen headers regenerated from the new pin, with the same umbrella header + modulemap `build_ios.sh` injects. Negative control: reverting just this one line reproduces the reviewer's exact error against those same headers. Note the `Swift SDK build` CI job cannot cover this — it is gated to `dashpay/platform` and `thepastaclaw`-owned forks, so it skips on this PR.

QuantumExplorer (approved):

QuantumExplorer (changes requested):
Actually no

bfoss765:
Superseded by #4305 (merged as b703f82ab) — same key-wallet bump and AddressState/asset-lock API migration with a minimal lockfile diff. Closing as agreed. One loose end carried here that #4305 dropped: the parity-manifest Swift test `testInvitationVouchersAreExcludedFromResumableList` (see the comment on #4305) — if the Kotlin SDK job on v4.2-dev is red at that step, cherry-picking that test from this branch fixes it.

PR #4301, opened 2026-08-05
feat(platform-wallet): multi-output shielded transfers + output-aware fee predictor (two-note invites)

bfoss765:
## What this closes

A shielded invite is funded as **one** note. When that note is later spent to
claim the invite, Orchard's `BundleType::DEFAULT` pads the single-spend bundle
up to `MIN_ACTIONS = 2`, and the padding action's dummy nullifier is **randomly
generated** (`orchard` `builder.rs:37,76-99`; `note.rs:227-243`;
`nullifier.rs:53-55`).

The new identity's id is `double_sha256(sorted PUBLISHED nullifiers)` —
`identity_id_from_nullifiers`, computed over *every* action's nullifier,
padding included, because consensus re-derives it the same way and dummies are
indistinguishable by design.

So with one real spend the claim's identity id contains fresh randomness. It
cannot be predicted before the build, and — the part that actually hurts — it
cannot be **re-derived on a retry**: a second attempt builds a different dummy
and therefore a different id. Any idempotent claim-recovery step that asks "did
my earlier attempt already create this identity?" has no expected id to compare
against and must fail closed.

With **two or more real spends** no padding action is added, every published
nullifier is the deterministic nullifier of a real note, and the id is a pure
function of the spent note set — predictable and reproducible.

## The fix: fund with two notes, not one

Fund the one-time address with **two notes that each hold less than the
target**, in one atomic transfer — `D` split as `floor(D/2) + ceil(D/2)`.

Note selection is greedy largest-first and breaks as soon as the accumulated
value covers the target (`note_selection.rs:117-135`). Neither half covers `D`
on its own, so **both are structurally forced into the claim bundle**. There is
no heuristic to tune and no way for the selector to pick just one.

This is why it is two sub-target notes rather than "a main note plus a small
anchor": with a main note that already covers the target, the selector would
stop after it and the padding action would come straight back.

**The claim path needs no change.** It already branches on
`selected_notes.len() >= 2` when deciding whether an expected identity id can
be derived. This PR changes the note *layout* so that branch is always taken;
the recovery logic itself is untouched.

## The fee predictor — why it must ship in the same PR

`build_shielded_transfer_transition` sized its fee from
`spends.len().max(2)`, **ignoring the output count**.

An Orchard action is a joined spend/output slot: the on-wire action count is
`max(num_spends, num_outputs)`, padded to `MIN_ACTIONS = 2`. A
`ShieldedTransfer`'s `value_balance` **is** its fee, and consensus pins it to
`compute_minimum_shielded_fee(actions.len())` *exactly* —
`validate_minimum_shielded_fee` rejects under-payment *and* over-payment for
this transition (`amount_is_pure_fee`).

Two-note funding means 2 recipient outputs + change = **3 outputs**. A
spends-only predictor would carve `min_fee(2)` while consensus demanded
`min_fee(3)`, and the transfer would be rejected on chain. So the two-note
layout is simply not constructible until this is fixed — the two changes cannot
be split.

The old form was correct *by accident* for every existing caller, because
`max(n, 1).max(2) == max(n, 2).max(2)` — with at most two outputs the output
side can never set the action count. That is why the bug is latent today rather
than a live failure.

Both builders now size the fee through a shared `shielded_bundle_action_count`,
which delegates to Orchard's own `BundleType::num_actions` rather than
re-deriving the rule, so the predictor cannot drift from the builder that
actually lays out the bundle.

## Deterministic bundle shape

The multi-output builder **always** emits a change output and requires the
spent value to *strictly* exceed `sum(amounts) + fee`.

That removes a genuine circularity: whether a change output exists depends on
the fee, and the fee depends on the output count. Pinning the change output as
unconditional makes the action count — and therefore the fee — a pure function
of the inputs: `max(spends, recipients + 1, 2)`. Note selection reserves
against the same `recipients + 1` floor, so the reserved fee and the carved fee
cannot diverge. A caller spending *exactly* `sum + fee` is rejected with a
clear error rather than silently re-shaped into a differently-priced bundle.

## Cost

Creation side: one extra Orchard action, `min_fee(3) - min_fee(2)`
= 31,425,600 credits = **+0.000314256 DASH** per invite
(0.001628512 → 0.001942768 DASH).

Claim side: **unchanged.** The claim spends two notes instead of one, but
`max(2 spends, 1 change output, 2)` = 2 actions either way — the padding action
it replaces was already being paid for.

## Legacy one-note invites are intentionally unsupported

No transitional or back-compat path is included. One-note invites have never
existed on mainnet, so there is nothing to migrate. `shielded_identity_id_is_reproducible`
states the rule as a single predicate next to the id derivation it guards:
callers that must recognise an identity their earlier attempt created gate on
the note count — no chain lookup, decided before any proving work — and treat a
non-reproducible set as unrecoverable rather than computing an id that will
never match.

## Layers

| Layer | Added |
|---|---|
| `rs-dpp` | `shielded_bundle_action_count`, `ShieldedTransferOutput`, `build_shielded_transfer_transition_multi`, `shielded_identity_id_is_reproducible` |
| `rs-platform-wallet` | `operations::transfer_multi`, `PlatformWallet::shielded_transfer_multi_to` |
| `rs-platform-wallet-ffi` | `platform_wallet_manager_shielded_transfer_multi` |
| `rs-unified-sdk-jni` + `kotlin-sdk` | `shieldedTransferMulti` |

Invite link format, `fundingCredits` and the exit-denomination ladder are
unchanged — this PR changes how a value is laid out across notes, not the
value. The V13 denomination set constrains the *exit* amount, not individual
note values.

## Tests

- **`multi_output_transfer_fee_matches_on_wire_action_count`** — builds a real
  2-spend / 3-output bundle and pins
  `value_balance == fee == min_fee(actions.len()) == min_fee(3)`, explicitly
  asserting it is *not* `min_fee(2)`. Also asserts the two outputs paid to the
  *same* address become two distinct note commitments.
- **`single_output_transfer_fee_matches_on_wire_action_count`** — pins the
  single-output builder against a real bundle so the shared helper cannot
  regress it.
- **`shielded_bundle_action_count_is_max_spends_outputs_padded_to_two`** and
  **`..._matches_a_real_bundle`** — pin the predictor, including the
  output-dominated shapes a spends-only predictor gets wrong.
- **`test_two_sub_denomination_notes_are_both_selected`** /
  **`test_single_full_denomination_note_selects_alone`** — pin the selector
  behaviour the whole design rests on, for both shipped denominations.
- **`test_select_notes_with_fee_reserves_multi_output_action_floor`** — the
  wallet reserves the 3-action fee, not the 2-action floor.
- The existing padding tests now also assert
  `shielded_identity_id_is_reproducible`, tying the predicate to observed
  builder behaviour rather than leaving it a bare constant.

Results: 216 `dpp` shielded tests and 662 `platform-wallet` lib tests pass.
`rustfmt` clean; `clippy --all-targets -D warnings` introduces no new findings
(the pre-existing findings in `recovery.rs`, `withdrawal.rs`, `persistence.rs`
and `core_wallet_types.rs` are identical on the untouched base).

## Follow-ups (deliberately not in this PR)

- **Swift parity** for `platform_wallet_manager_shielded_transfer_multi`. The
  cbindgen header is generated at build time and nothing in the Swift SDK
  references the new symbol, so the Swift build is unaffected.
- **App-side adoption** — switching invite funding to two outputs needs an SDK
  artifact carrying `shieldedTransferMulti`, so it lands with the next AAR.
- **Wiring the guard into the claim scan** — the claim path lives in a separate
  in-flight PR; this PR supplies the predicate it should call.

## Summary by CodeRabbit

- **New Features**
  - Added atomic shielded transfers to multiple recipients in a single transaction.
  - Added shared optional memo support for multi-recipient transfers.
  - Added Kotlin SDK support for initiating multi-recipient shielded transfers.
  - Added support for up to 16 recipients per transfer.

- **Bug Fixes**
  - Improved validation for recipient addresses, amounts, account indexes, and transaction limits.
  - Improved fee calculation, change handling, note selection, and transaction consistency for multi-output transfers.

### DIFF
--- a/packages/rs-dpp/src/shielded/builder/mod.rs
+++ b/packages/rs-dpp/src/shielded/builder/mod.rs
@@ -43,7 +43,10 @@ pub use identity_create_from_shielded_pool::{
 pub use shield_from_asset_lock::build_shield_from_asset_lock_transition;
 #[cfg(feature = "core_key_wallet")]
 pub use shield_from_asset_lock::build_shield_from_asset_lock_transition_with_signer;
-pub use shielded_transfer::build_shielded_transfer_transition;
+pub use shielded_transfer::{
+    build_shielded_transfer_transition, build_shielded_transfer_transition_multi,
+    ShieldedTransferOutput,
+};
 pub use shielded_withdrawal::build_shielded_withdrawal_transition;
 pub use unshield::build_unshield_transition;
 
@@ -101,6 +104,36 @@ impl From<&OrchardAddress> for PaymentAddress {
     fn from(address: &OrchardAddress) -> Self {
         *address.inner()
     }
+}
+
+/// The number of Orchard actions a `BundleType::DEFAULT` bundle built from `num_spends` spends
+/// and `num_outputs` outputs will publish **on the wire**.
+///
+/// Every shielded fee predictor MUST size its fee with this function, because consensus prices
+/// the fee off the on-wire `actions.len()` (see
+/// `StateTransitionShieldedMinimumFeeValidationV0::validate_minimum_shielded_fee`, which reads
+/// `v0.actions.len()`), and an Orchard action is a *joined* spend/output slot: the action count
+/// is `max(num_spends, num_outputs)`, then padded up to Orchard's `MIN_ACTIONS = 2`.
+///
+/// The output side matters. A predictor that looks only at the spend count is correct **only**
+/// while `num_outputs <= 2`, because `max(n, 1).max(2) == max(n, 2).max(2)`. As soon as a
+/// transition publishes three or more outputs (a multi-recipient transfer plus change), a
+/// spends-only predictor under-counts and carves a fee below the one consensus computes — fatal
+/// for `ShieldedTransfer`, whose `value_balance` must equal the minimum fee **exactly**.
+///
+/// This delegates to Orchard's own [`BundleType::num_actions`] rather than re-deriving the rule,
+/// so the predictor cannot drift from the builder that actually lays out the bundle.
+pub fn shielded_bundle_action_count(
+    num_spends: usize,
+    num_outputs: usize,
+) -> Result<usize, ProtocolError> {
+    BundleType::DEFAULT
+        .num_actions(num_spends, num_outputs)
+        .map_err(|e| {
+            ProtocolError::ShieldedBuildError(format!(
+                "invalid Orchard bundle shape ({num_spends} spends, {num_outputs} outputs): {e}"
+            ))
+        })
 }
 
 /// Serializes an authorized Orchard bundle into the raw fields used by
@@ -779,6 +812,59 @@ mod mod_tests {
                 assert_eq!(msg, "closure rejected", "closure error must pass through");
             }
             other => panic!("expected the closure's error to propagate, got {:?}", other),
+        }
+    }
+
+    // ------------------------------------------------------------------
+    // `shielded_bundle_action_count` — the shared fee-sizing predictor.
+    // ------------------------------------------------------------------
+
+    /// The predictor must be `max(num_spends, num_outputs)` padded to Orchard's 2-action
+    /// minimum — for the OUTPUT side as well as the spend side. The `num_outputs >= 3` rows are
+    /// the ones a spends-only predictor gets wrong.
+    #[test]
+    fn shielded_bundle_action_count_is_max_spends_outputs_padded_to_two() {
+        for (spends, outputs, expected) in [
+            (0usize, 1usize, 2usize),
+            (1, 1, 2),
+            (1, 2, 2),
+            (2, 2, 2),
+            // Output-dominated shapes: the spend count no longer determines the fee.
+            (1, 3, 3),
+            (2, 3, 3),
+            (1, 4, 4),
+            (5, 3, 5),
+            (3, 7, 7),
+        ] {
+            let actual = shielded_bundle_action_count(spends, outputs)
+                .expect("DEFAULT bundles accept any spend/output mix");
+            assert_eq!(
+                actual, expected,
+                "action count for {spends} spends / {outputs} outputs"
+            );
+        }
+    }
+
+    /// A real bundle's on-wire `actions.len()` — the number consensus prices the fee off — must
+    /// equal what the predictor said. Exercised through the output-only builder because it is
+    /// the cheapest real bundle to construct at several output counts.
+    #[test]
+    fn shielded_bundle_action_count_matches_a_real_bundle() {
+        let recipient = test_orchard_address();
+        // (dummy_outputs, total outputs = 1 real + dummies)
+        for dummies in [0usize, 1, 4] {
+            let num_outputs = 1 + dummies;
+            let bundle =
+                build_output_only_bundle(&recipient, 10_000, [0u8; 36], None, dummies, &TestProver)
+                    .expect("bundle should build");
+            let predicted =
+                shielded_bundle_action_count(0, num_outputs).expect("valid bundle shape");
+            assert_eq!(
+                bundle.actions().len(),
+                predicted,
+                "predicted action count must match the real bundle's on-wire count for \
+                 {num_outputs} outputs"
+            );
         }
     }
 }
--- a/packages/rs-dpp/src/shielded/builder/mod.rs
+++ b/packages/rs-dpp/src/shielded/builder/mod.rs
@@ -55,6 +55,7 @@ use grovedb_commitment_tree::{
     FullViewingKey, MerklePath, Note, NoteValue, OutgoingViewingKey, PaymentAddress, ProvingKey,
     Scope, SpendAuthorizingKey, SpendingKey,
 };
+use platform_version::version::PlatformVersion;
 use rand::rngs::OsRng;
 use rand::RngCore;
 
@@ -107,7 +108,8 @@ impl From<&OrchardAddress> for PaymentAddress {
 }
 
 /// The number of Orchard actions a `BundleType::DEFAULT` bundle built from `num_spends` spends
-/// and `num_outputs` outputs will publish **on the wire**.
+/// and `num_outputs` outputs will publish **on the wire**, validated against the consensus
+/// action ceiling.
 ///
 /// Every shielded fee predictor MUST size its fee with this function, because consensus prices
 /// the fee off the on-wire `actions.len()` (see
@@ -123,17 +125,41 @@ impl From<&OrchardAddress> for PaymentAddress {
 ///
 /// This delegates to Orchard's own [`BundleType::num_actions`] rather than re-deriving the rule,
 /// so the predictor cannot drift from the builder that actually lays out the bundle.
+///
+/// # The consensus ceiling
+///
+/// Every shielded transition's `validate_structure` rejects a bundle whose `actions.len()`
+/// exceeds `platform_version.system_limits.max_shielded_transition_actions` (via
+/// `validate_actions_count`), but the `try_from_bundle` constructors do NOT run structural
+/// validation — so without this gate an over-sized bundle is built, proved (~30 s of Halo 2),
+/// and only then rejected on chain. Because the action count is `max(spends, outputs)`, bounding
+/// it here bounds BOTH sides: a fragmented wallet spending too many notes and a caller asking
+/// for too many outputs are rejected by the same comparison, before any proving work starts.
 pub fn shielded_bundle_action_count(
     num_spends: usize,
     num_outputs: usize,
+    platform_version: &PlatformVersion,
 ) -> Result<usize, ProtocolError> {
-    BundleType::DEFAULT
+    let num_actions = BundleType::DEFAULT
         .num_actions(num_spends, num_outputs)
         .map_err(|e| {
             ProtocolError::ShieldedBuildError(format!(
                 "invalid Orchard bundle shape ({num_spends} spends, {num_outputs} outputs): {e}"
             ))
-        })
+        })?;
+
+    let max_actions = platform_version
+        .system_limits
+        .max_shielded_transition_actions as usize;
+    if num_actions > max_actions {
+        return Err(ProtocolError::ShieldedBuildError(format!(
+            "a bundle of {num_spends} spends and {num_outputs} outputs publishes {num_actions} \
+             Orchard actions, exceeding the consensus limit of {max_actions} \
+             (max_shielded_transition_actions); consensus would reject the proved transition"
+        )));
+    }
+
+    Ok(num_actions)
 }
 
 /// Serializes an authorized Orchard bundle into the raw fields used by
@@ -824,6 +850,7 @@ mod mod_tests {
     /// the ones a spends-only predictor gets wrong.
     #[test]
     fn shielded_bundle_action_count_is_max_spends_outputs_padded_to_two() {
+        let platform_version = PlatformVersion::latest();
         for (spends, outputs, expected) in [
             (0usize, 1usize, 2usize),
             (1, 1, 2),
@@ -836,7 +863,7 @@ mod mod_tests {
             (5, 3, 5),
             (3, 7, 7),
         ] {
-            let actual = shielded_bundle_action_count(spends, outputs)
+            let actual = shielded_bundle_action_count(spends, outputs, platform_version)
                 .expect("DEFAULT bundles accept any spend/output mix");
             assert_eq!(
                 actual, expected,
@@ -850,6 +877,7 @@ mod mod_tests {
     /// the cheapest real bundle to construct at several output counts.
     #[test]
     fn shielded_bundle_action_count_matches_a_real_bundle() {
+        let platform_version = PlatformVersion::latest();
         let recipient = test_orchard_address();
         // (dummy_outputs, total outputs = 1 real + dummies)
         for dummies in [0usize, 1, 4] {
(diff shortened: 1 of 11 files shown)


thepastaclaw:
⛔ Blockers found — Opus deferred (commit 9f11d66c)

packages/rs-dpp/src/shielded/builder/shielded_transfer.rs
```diff
@@ -134,12 +142,464 @@ pub fn build_shielded_transfer_transition<P: OrchardProver>(
...
+            )
+        })?;
+
+    // A change output is always emitted (see the doc comment), so the output count — and with it
+    // the action count and the fee — is fixed before any value arithmetic.
+    let num_outputs = outputs.len().checked_add(1).ok_or_else(|| {
+        ProtocolError::ShieldedBuildError("output count overflows usize".to_string())
+    })?;
+    let num_actions = shielded_bundle_action_count(spends.len(), num_outputs)?;
+    let fee = compute_minimum_shielded_fee(num_actions, platform_version)?;
```
thepastaclaw: **🔴 Blocking: Reject bundles that exceed the consensus action limit**

`shielded_bundle_action_count` computes the bundle shape but never compares it with `platform_version.system_limits.max_shielded_transition_actions`. `ShieldedTransferTransitionV0::validate_structure` rejects action counts above that limit, while `try_from_bundle` only constructs the transition and performs no structural validation. The new FFI accepts 16 recipients, which become 17 outputs after adding the unconditional change output and therefore at least 17 actions, exceeding the current limit of 16. A fragmented wallet can also exceed the limit through its spend count. Reject this before proof generation so callers cannot spend roughly 30 seconds proving a transition that consensus must reject.

```suggestion
    let num_actions = shielded_bundle_action_count(spends.len(), num_outputs)?;
    let max_actions = usize::from(
        platform_version
            .system_limits
            .max_shielded_transition_actions,
    );
    if num_actions > max_actions {
        return Err(ProtocolError::ShieldedBuildError(format!(
            "shielded transfer requires {num_actions} actions, exceeding the protocol maximum of {max_actions}"
        )));
    }
    let fee = compute_minimum_shielded_fee(num_actions, platform_version)?;
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/shielded/operations.rs
```diff
@@ -989,6 +990,198 @@ pub async fn transfer<S: ShieldedStore, P: OrchardProver>(
...
+    let num_outputs = builder_outputs.len() + 1;
+    let (selected_notes, total_input, exact_fee) = reserve_unspent_notes(
+        sdk,
+        store,
+        id,
+        total_amount,
+        num_outputs,
+        ShieldedFeeKind::Base,
+    )
+    .await?;
```
thepastaclaw, line 1087: **🔴 Blocking: Reserve enough input to guarantee positive change**

`reserve_unspent_notes` accepts a selection where `total_input == total_amount + exact_fee`, but `build_shielded_transfer_transition_multi` unconditionally emits a positive-valued change output and rejects equality. With notes valued `[total_amount + exact_fee, 1]`, largest-first selection reserves only the exact-coverage note and the builder fails even though selecting the remaining credit would satisfy its invariant. The failure path releases the pending reservation, so funds are not stranded, but a valid wallet balance is still rejected. Reserve one additional credit so every successful selection satisfies the builder's strict-change precondition and the existing fee-convergence loop can account for any additional selected action.

```suggestion
    // Reserve against the SAME output count the builder sizes its fee from: every recipient
    // output plus the unconditional change output.
    let num_outputs = builder_outputs.len() + 1;
    let reserve_amount = total_amount.checked_add(1).ok_or_else(|| {
        PlatformWalletError::ShieldedBuildError(
            "multi-output shielded transfer amounts overflow u64".to_string(),
        )
    })?;
    let (selected_notes, total_input, exact_fee) = reserve_unspent_notes(
        sdk,
        store,
        id,
        reserve_amount,
        num_outputs,
        ShieldedFeeKind::Base,
    )
    .await?;
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/shielded_send.rs
```diff
@@ -350,6 +350,153 @@ pub unsafe extern "C" fn platform_wallet_manager_shielded_transfer(
...
+pub unsafe extern "C" fn platform_wallet_manager_shielded_transfer_multi(
+    handle: Handle,
+    wallet_id_bytes: *const u8,
+    mnemonic_resolver_handle: *mut MnemonicResolverHandle,
+    account: u32,
+    recipients_raw_43: *const u8,
+    amounts: *const u64,
+    num_recipients: usize,
+    memo_text: *const c_char,
+) -> PlatformWalletFFIResult {
```
thepastaclaw, line 484: **🟡 Suggestion: Catch panics before crossing the C ABI boundary**

The JNI export wraps this call in `support::guard`, but this function uses `extern "C"`. A panic cannot unwind through that ABI, so Rust aborts at the C function boundary before the caller's `catch_unwind` can translate it into a Java exception. A concrete path exists in `block_on_worker`: if the proving future panics, the task returns `JoinError` and `expect("tokio worker panicked")` panics inside this export. Android profiles deliberately retain unwinding so JNI panics can become Java exceptions, but this intermediate ABI boundary defeats that guarantee. Catch panics inside the C export and convert them to an FFI error, or move the implementation into a normal Rust function that JNI invokes directly under `guard`. Because a panic can occur after note reservation, the error path must conservatively preserve potentially in-flight spend semantics.

<sub>source: ['codex']</sub>

packages/rs-unified-sdk-jni/src/funding.rs
```diff
@@ -845,6 +845,118 @@ pub extern "system" fn Java_org_dashfoundation_dashsdk_ffi_FundingNative_shielde
...
+                1,
+                &format!(
+                    "recipientsRaw43 must be 43 bytes per amount ({} expected), got {}",
+                    amount_len * 43,
+                    recipients.len()
+                ),
+            );
+            return;
+        }
+        let mut amount_buf = vec![0i64; amount_len];
```
thepastaclaw: **🟡 Suggestion: Enforce the recipient bound before allocating JNI buffers**

The C export rejects more than 16 recipients before allocating, but this JNI adapter first copies the complete Java recipient array into a native `Vec`, then allocates `amount_buf`, and later allocates another `amounts_u64` vector. The public Kotlin wrapper likewise constructs flattened recipient and amount arrays without checking the native ceiling. An accidental or hostile oversized call can therefore trigger several caller-sized allocations and allocator OOM before reaching the clean `ErrorInvalidParameter` response. Read `amounts.length` first, reject counts above the supported ceiling, and only then convert either array; mirror the same check in `PlatformWalletManager.shieldedTransferMulti` before constructing the flattened buffers.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The output-aware fee calculation is well structured, but the new multi-output path does not enforce the protocol action ceiling before proving and can reject otherwise fundable transfers because note selection permits exact fee coverage despite the builder requiring positive change. The new JNI path also needs its allocation bound enforced before copying caller-sized arrays, and its call through an `extern "C"` export prevents the outer JNI panic guard from handling worker panics.

### Review provenance

<details>

```

In `packages/rs-dpp/src/shielded/builder/shielded_transfer.rs`:
- [BLOCKING] packages/rs-dpp/src/shielded/builder/shielded_transfer.rs:241-242: Reject bundles that exceed the consensus action limit
  `shielded_bundle_action_count` computes the bundle shape but never compares it with `platform_version.system_limits.max_shielded_transition_actions`. `ShieldedTransferTransitionV0::validate_structure` rejects action counts above that limit, while `try_from_bundle` only constructs the transition and performs no structural validation. The new FFI accepts 16 recipients, which become 17 outputs after adding the unconditional change output and therefore at least 17 actions, exceeding the current limit of 16. A fragmented wallet can also exceed the limit through its spend count. Reject this before proof generation so callers cannot spend roughly 30 seconds proving a transition that consensus must reject.

- [BLOCKING] packages/rs-platform-wallet/src/wallet/shielded/operations.rs:1050-1061: Reserve enough input to guarantee positive change
  `reserve_unspent_notes` accepts a selection where `total_input == total_amount + exact_fee`, but `build_shielded_transfer_transition_multi` unconditionally emits a positive-valued change output and rejects equality. With notes valued `[total_amount + exact_fee, 1]`, largest-first selection reserves only the exact-coverage note and the builder fails even though selecting the remaining credit would satisfy its invariant. The failure path releases the pending reservation, so funds are not stranded, but a valid wallet balance is still rejected. Reserve one additional credit so every successful selection satisfies the builder's strict-change precondition and the existing fee-convergence loop can account for any additional selected action.

- [SUGGESTION] packages/rs-platform-wallet-ffi/src/shielded_send.rs:396-405: Catch panics before crossing the C ABI boundary
  The JNI export wraps this call in `support::guard`, but this function uses `extern "C"`. A panic cannot unwind through that ABI, so Rust aborts at the C function boundary before the caller's `catch_unwind` can translate it into a Java exception. A concrete path exists in `block_on_worker`: if the proving future panics, the task returns `JoinError` and `expect("tokio worker panicked")` panics inside this export. Android profiles deliberately retain unwinding so JNI panics can become Java exceptions, but this intermediate ABI boundary defeats that guarantee. Catch panics inside the C export and convert them to an FFI error, or move the implementation into a normal Rust function that JNI invokes directly under `guard`. Because a panic can occur after note reservation, the error path must conservatively preserve potentially in-flight spend semantics.

In `packages/rs-unified-sdk-jni/src/funding.rs`:
- [SUGGESTION] packages/rs-unified-sdk-jni/src/funding.rs:884-916: Enforce the recipient bound before allocating JNI buffers
  The C export rejects more than 16 recipients before allocating, but this JNI adapter first copies the complete Java recipient array into a native `Vec`, then allocates `amount_buf`, and later allocates another `amounts_u64` vector. The public Kotlin wrapper likewise constructs flattened recipient and amount arrays without checking the native ceiling. An accidental or hostile oversized call can therefore trigger several caller-sized allocations and allocator OOM before reaching the clean `ErrorInvalidParameter` response. Read `amounts.length` first, reject counts above the supported ceiling, and only then convert either array; mirror the same check in `PlatformWalletManager.shieldedTransferMulti` before constructing the flattened buffers.
```
</details>

bfoss765:
Pushed `9f11d66c` addressing all four findings from the last review round:

- **Action limit (blocker):** `shielded_bundle_action_count` now takes `platform_version` and rejects `num_actions > max_shielded_transition_actions` before any proving — one comparison bounds both spend and output sides since the count is `max(spends, outputs)` padded to 2. Called ahead of the `add_spend` loop in both the single and multi builders. Tests: 16-action boundary accepted, 17-action output-side and 17-spend spend-side rejected in milliseconds.
- **Exact-fit selection (blocker):** new `ChangeRequirement` folded into `select_notes_with_fee`'s convergence loop, so the strictly-positive-change requirement is enforced against the *recomputed* fee whenever selection adds a note. The `[amount+fee, 1]` case now selects both notes. Fixed on the selector side deliberately — the builder's strict-change invariant is what keeps its action count a pure function of inputs.
- **Panic across `extern "C"` (suggestion):** the multi-output export is now a `catch_unwind` wrapper; caught panics map to `ErrorShieldedSpendUnconfirmed`, whose documented contract (may have been broadcast, notes stay reserved, do not retry) matches an unknown-state failure exactly.
- **JNI allocation ordering (suggestion):** array lengths are read and validated (count ceiling + amounts/addresses length match) before any `convert_byte_array` — this also closed an unflagged variant on the address-blob path. `MAX_SHIELDED_TRANSFER_RECIPIENTS` is shared with the JNI adapter, and the Kotlin side mirrors the check.

1,110 tests pass across the three touched crates (dpp 213, platform-wallet 665, platform-wallet-ffi 232), fmt/clippy clean.

packages/rs-dpp/src/shielded/builder/mod.rs
```diff
@@ -103,6 +107,61 @@ impl From<&OrchardAddress> for PaymentAddress {
...
+    let max_actions = platform_version
+        .system_limits
+        .max_shielded_transition_actions as usize;
+    if num_actions > max_actions {
+        return Err(ProtocolError::ShieldedBuildError(format!(
+            "a bundle of {num_spends} spends and {num_outputs} outputs publishes {num_actions} \
+             Orchard actions, exceeding the consensus limit of {max_actions} \
+             (max_shielded_transition_actions); consensus would reject the proved transition"
+        )));
+    }
```
thepastaclaw, line 160: **🔴 Blocking: Reject bundles over the effective transition-size limit before proving**

This gate accepts any bundle with at most `max_shielded_transition_actions`, currently 16, but the versioned `max_state_transition_size` of 20,480 bytes is already the tighter acceptance limit. A seven-action `ShieldedTransfer` serializes to 21,594 bytes at this head, so six recipients plus the mandatory change output—or a fragmented wallet selecting seven spends—passes this check and reaches `prove_and_sign_bundle`, only to be rejected later by DAPI's raw-byte prefilter and Drive-ABCI's consensus decoding check. The new Kotlin/FFI API admits up to 16 recipients, making this guaranteed-failure path externally reachable and unnecessarily consuming the expensive Halo 2 proving worker. Add a version-aware pre-proving limit derived from both the transition byte ceiling and the structural action ceiling, cover output- and spend-dominated bundles, and align the binding recipient ceiling with the effective limit; under the current 20 KiB limit, at most five recipients fit alongside mandatory change.

thepastaclaw (changes requested):

Carried-forward prior findings: all four supplied findings are fixed at the current head—the action-limit gate, strict-change note selection, C-ABI panic guard, and pre-allocation JNI recipient bound are present and tested. New latest-delta finding: the action-count gate still admits seven-action transfers that necessarily exceed the separate versioned 20 KiB transition-size limit, allowing expensive proof generation for transitions that DAPI and Drive-ABCI will reject. This remaining in-scope blocker requires changes.

Source: reviewers codex/general=gpt-5.6-sol(completed); codex/security-auditor=gpt-5.6-sol(completed); codex/rust-quality=gpt-5.6-sol(completed); codex/ffi-engineer=gpt-5.6-sol(completed); verifier=codex/verifier=gpt-5.6-sol(completed); coordinator=openclaw-agent/cliproxy/gpt-5.6-sol(orchestration-only).

### Review provenance

🔴 1 blocking

<details>

```

In `packages/rs-dpp/src/shielded/builder/mod.rs`:
- [BLOCKING] packages/rs-dpp/src/shielded/builder/mod.rs:151-160: Reject bundles over the effective transition-size limit before proving
  This gate accepts any bundle with at most `max_shielded_transition_actions`, currently 16, but the versioned `max_state_transition_size` of 20,480 bytes is already the tighter acceptance limit. A seven-action `ShieldedTransfer` serializes to 21,594 bytes at this head, so six recipients plus the mandatory change output—or a fragmented wallet selecting seven spends—passes this check and reaches `prove_and_sign_bundle`, only to be rejected later by DAPI's raw-byte prefilter and Drive-ABCI's consensus decoding check. The new Kotlin/FFI API admits up to 16 recipients, making this guaranteed-failure path externally reachable and unnecessarily consuming the expensive Halo 2 proving worker. Add a version-aware pre-proving limit derived from both the transition byte ceiling and the structural action ceiling, cover output- and spend-dominated bundles, and align the binding recipient ceiling with the effective limit; under the current 20 KiB limit, at most five recipients fit alongside mandatory change.
```
</details>

packages/rs-platform-wallet/src/wallet/shielded/operations.rs
thepastaclaw: Resolved in `9f11d66` — _Reserve enough input to guarantee positive change_ no longer present.

packages/rs-platform-wallet-ffi/src/shielded_send.rs
thepastaclaw: Resolved in `9f11d66` — _Catch panics before crossing the C ABI boundary_ no longer present.

packages/rs-dpp/src/shielded/builder/shielded_transfer.rs
thepastaclaw: Resolved in `9f11d66` — _Reject bundles that exceed the consensus action limit_ no longer present.

packages/rs-unified-sdk-jni/src/funding.rs
thepastaclaw: Resolved in `9f11d66` — _Enforce the recipient bound before allocating JNI buffers_ no longer present.

bfoss765:
Moved to #4312 - in-repo branch rebased onto v4.2-dev (post-#4305), carrying the latest review-round fixes (9f11d66c). Review history preserved here.

PR #4261, opened 2026-08-01
docs(platform-wallet): error-code registry for the FFI result space

bfoss765:
## Problem

`PlatformWalletFFIResultCode` (`packages/rs-platform-wallet-ffi/src/error.rs`) is public ABI — `cbindgen` emits the integers into the generated C header, Swift mirrors them in `PlatformWalletResultCode`, and Kotlin branches on them in `fromPlatformWalletNative`. Four open PRs are currently allocating adjacent integers in that enum with no shared source of truth.

A duplicate discriminant in two branches produces **no textual merge conflict**. The second merge just silently misclassifies errors on every host. That is not hypothetical here: `29` is already claimed by two PRs, and review on #4185 had to issue a renumber directive (`ErrorReservationWalletMismatch` 29 → 30) to resolve it. Three separate PRs now carry hand-written comments trying to reserve ranges on each other's behalf, and one of those reservations is for a variant that no longer exists.

This PR adds a registry so the allocation is decided in one file instead of in four diffs.

## What's here

`packages/rs-platform-wallet-ffi/ERROR_CODE_REGISTRY.md`:

- the merged allocation table (0–26, plus the 98/99 sentinels),
- the proposed allocations with their owning PRs (27–33),
- the rule for claiming a new code — next free integer, recorded here in the same PR that adds it, mirrors updated alongside,
- the ABI-stability rule — never renumber after release; deprecate and allocate a new integer,
- the contested/pending allocations, written up below as well.

Plus a six-line pointer from the crate README's Error Handling section.

**Docs-only. Zero code risk.** No enum, no `From` mapping, no Swift/Kotlin mirror is touched, and nothing is renumbered — that stays the owning PRs' work. This PR only writes down what the branches already say.

## Codes claimed by the in-flight PRs

Reconciling in one place, so this is the table to correct:

| Code | Name | Owning PR |
| ---: | --- | --- |
| 27 | `ErrorStaleReservationToken` | #4185 (also carried by #4256) |
| 28 | `ErrorReservationTokenConsumed` | #4185 (also carried by #4256) |
| 29 | `ErrorReservationWalletMismatch` | #4185 — **collides**, see below |
| 29 | `ErrorAssetLockInsufficientFunds` | #4184 — **collides**, see below |
| 30 | *unallocated* | reserved in sibling comments only, see below |
| 31 | `ErrorSigningKeyUnavailable` | #4183 (also carried by #4204) |
| 32 | `ErrorTransactionBuild` | #4247 (also carried by #4256) |
[...]

PR #4308, opened 2026-08-06
feat(kotlin-sdk): split build/broadcast with reservation release for BIP70-style deferred submission

bfoss765:
Continues #4185 — moved from a fork branch to an in-repo branch (rebased onto v4.2-dev post-#4305) so maintainers can push changes directly, per review request. Full review history on #4185.

---

Splits transaction build from broadcast for BIP70-style deferred submission: a signed-payment registry with reservation release, a deferred-payment token bounded to the reservation's lifetime, native code 27 for stale reservation tokens, token sweeping only when the final wallet write wins, and routing of deferred builds through the atomic finalize-and-register path — across `rs-platform-wallet`, `platform-wallet-ffi`, `rs-unified-sdk-jni`, and the Kotlin SDK surface.

Re-opens #4090 which was auto-closed when the #3999 base branch was deleted; rebased onto `v4.1-dev`. All seven original commits replayed cleanly — no hunks needed to be dropped as already-absorbed.

Verified: `cargo test -p platform-wallet -p platform-wallet-ffi -p rs-unified-sdk-jni` all pass (504 / 229 / 10); `:sdk:assembleRelease` + sdk unit tests pass.

## Summary by CodeRabbit

- **New Features**
  - Added support for building and signing payments before broadcasting.
  - Added APIs to broadcast deferred payments or release their reservations later.
  - Added Kotlin and Swift error handling for stale, consumed, or mismatched reservation tokens.
- **Bug Fixes**
  - Improved wallet lifecycle handling for removed or recreated wallets.
  - Added safer reservation cleanup for rejected, failed, cancelled, and abandoned payments.
  - Prevented stale operations from affecting newer wallet instances.
  - Improved handling of ambiguous broadcast failures and reservation expiration.

---

### Why the token registry instead of the V2 handle surface

The deferred BIP70/BIP270 flow uses the reservation-token registry rather than the V2 finalized-transaction handle for two concrete reasons. First, ownership and cleanup: the token is wrapped in an owning, `AutoCloseable` Kotlin object with a GC/`Cleaner` backstop, so a payment that is signed but then abandoned — the merchant server never acks, the user backs out, or the coroutine is cancelled after the native registration returned — always releases its funding reservation, for free, without the caller having to remember to abandon a handle. Second, the token path carries a lifetime bound the V2 handle does not: it stamps each token with the reservation's own pre-signing height and refuses to act once that reservation could have aged into key-wallet's TTL sweep, so a slow external signer can never let a stale token spend against an outpoint the wallet already swept and re-selected. A pinned V2 `CoreWallet` handle has no such age guard and would keep the old wallet actionable indefinitely. Both paths now share one wallet-generation identity and one teardown policy, so the V2 surface stays correct for the immediate send it was built for while the deferred flow gets the GC-safe, age-bounded ownership it needs. A follow-up adds the age guard to the V2 handle path itself (it becomes live the moment iOS does deferred sends).

All five lifetime findings addressed as merge blockers, one commit each, with regression tests:
1. **Destroy vs teardown**: final-alias `platform_wallet_destroy` releases the generation's reservations against the still-live wallet; actual generation removal drops tokens and V2 handles — token cleanup is now tied to wallet-generation removal.
2. **Height carry**: the pre-signing reservation height travels on `SignedCoreTransaction` and `register` uses it — no post-signing resample; boundary test pins the TTL margin.
3. **One generation identity**: `CoreWallet::is_same_generation` (per-generation identity) is checked by BOTH the V2-handle and registry-token paths, with one teardown policy.
4. **Cancellation-safe ownership**: `SignedCoreTransaction` is an owning `AutoCloseable` with a `NativeCleaner` backstop; round-2 adds object-owning `broadcastSigned`/`releaseReservation` overloads that hold the object reachable across the native call and disarm the backstop on consumption (the bare-token forms remain but document the reachability requirement).
5. **Validate-under-lock**: `broadcast` peeks and consumes atomically under one lock hold (network I/O outside the lock); a wrong-wallet caller leaves the owner's token untouched, pinned by test.

Also per review: the dead `core_wallet_signed_payment_register` four-layer chain is deleted; the single stale-token code is split into typed siblings **27 `ErrorStaleReservationToken` / 28 `ErrorReservationTokenConsumed` / 29 `ErrorReservationWalletMismatch`** (code 26 is not used by this PR — upstream now owns it as `ErrorTransactionBroadcastRejected`; both the Kotlin and Swift enums on this branch map 27/28/29 explicitly); the stale `buildSignedPayment` KDoc is fixed.

> **Error-code allocation note.** These 27 / 28 / 29 allocations are being reconciled repo-wide in the error-code registry PR #4261. That registry records that `ErrorReservationWalletMismatch = 29` on this branch currently collides with `ErrorAssetLockInsufficientFunds = 29` on the asset-lock PR #4184, and that code 30 is free after #4184's re-scope (the variant previously reserved at 30 is not defined anywhere). The resolution of record is that #4184 keeps 29 and this PR moves its mismatch code to 30; that renumber has not yet landed on this head.

Local test evidence (fork PRs skip the Rust CI suite): `platform-wallet --lib` 508 passed, `platform-wallet-ffi --lib` 197 passed, clippy/fmt clean, Kotlin `:sdk:testDebugUnitTest` green.


thepastaclaw:
🔍 Review in progress — actively reviewing now (commit ebaca83a)
_ETA: complete ~12:40 UTC (median 19m across 30 recent reviews)_
_Running 9m · Last checked: 2026-08-06 12:30 UTC_

packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs
```diff
@@ -0,0 +1,1921 @@
...
+        }
+        entry
+            .core
+            .release_transaction_reservation(
+                entry.account_type,
+                entry.account_index,
+                &entry.tx,
+                entry.funding_reservation_token,
+            )
+            .await;
```
thepastaclaw: **🔴 Blocking: Release stale entries with their retained owner token**

Both the stale-broadcast branch and `reconcile_removed_entry` discard an expired payment without releasing its reservation. The exact key-wallet revision used here retains per-reservation ownership and implements `release_reservation_if_owner` atomically: it removes inputs only while they still belong to this payment and becomes a no-op after a TTL sweep or re-reservation transfers ownership. The current tests advance 22 blocks, below key-wallet's 24-block TTL, prove that the original reservation is still held, and require an immediate rebuild to fail. Consequently, the stale-token error instructs callers to rebuild while leaving the required inputs unavailable for several more blocks, and explicit `releaseReservation` reports success without releasing them. Run the owner-guarded reconciliation for stale entries whenever `funding_reservation_token` is present; retain the no-release fallback only for the unowned `None` case.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/manager/wallet_lifecycle.rs
```diff
@@ -640,6 +816,12 @@ impl<P: PlatformWalletPersistence + 'static> PlatformWalletManager<P> {
                 .await;
```
thepastaclaw: **🔴 Blocking: Prevent old-generation teardown from deleting recreated wallet state**

Removing G1 from the inner wallet manager frees `wallet_id`, allowing the deliberately supported concurrent G2 registration to publish a new generation. Teardown then unconditionally calls `coordinator.unregister_wallet(*wallet_id)` and unregisters G1's identity IDs, but both side registries are keyed only by wallet or identity ID rather than `WalletGeneration`. If G2 binds shielded state or registers the same identity-sync row before G1 reaches these awaited removals, G1 deletes G2's newly installed state. The generation check around `self.wallets.remove` protects only the public wallet map. Keep same-ID registration fenced until all ID-keyed teardown is complete, or make these side registries and unregister operations generation-aware; extend the midpoint recreation test to install G2 side-manager state and verify that it survives.

<sub>source: ['codex']</sub>

packages/rs-unified-sdk-jni/src/wallet_manager.rs
```diff
@@ -1225,6 +1225,200 @@ pub extern "system" fn Java_org_dashfoundation_dashsdk_ffi_WalletManagerNative_c
...
+        if take_pwffi_error(env, result) {
+            // The FFI freed the builder on the error path and left the out struct
+            // zeroed (null tx_bytes); dropping `boxed` frees only the box.
+            return ptr::null_mut();
+        }
+        if out_txid.is_null() {
+            unsafe { platform_wallet_ffi::core_wallet_transaction_free(out_tx) };
+            throw_sdk_exception(env, 1, "finalize returned a NULL txid");
+            return ptr::null_mut();
+        }
```
thepastaclaw, line 1527: **🟡 Suggestion: Release the registered token when defensive txid validation fails**

A successful `core_wallet_signed_payment_finalize` has already inserted the payment into the process-global registry and transferred reservation ownership to `token`. If this defensive null-txid branch fires because of an FFI regression, it frees the transaction buffer but returns without releasing that token, orphaning the registry entry and reservation until the TTL backstop. The later byte-array allocation failure path correctly releases the token; this post-registration failure path must do the same.

```suggestion
        if out_txid.is_null() {
            unsafe { platform_wallet_ffi::core_wallet_transaction_free(out_tx) };
            let _ = unsafe {
                platform_wallet_ffi::core_wallet_signed_payment_release(token)
            };
            throw_sdk_exception(env, 1, "finalize returned a NULL txid");
            return ptr::null_mut();
        }
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/core_wallet/signed_payment.rs
```diff
@@ -0,0 +1,145 @@
...
+/// `core_handle` must be a valid core-wallet handle; `out_txid` must be writable.
+#[no_mangle]
+pub unsafe extern "C" fn core_wallet_signed_payment_broadcast(
+    core_handle: Handle,
+    token: u64,
+    out_txid: *mut *mut c_char,
+) -> PlatformWalletFFIResult {
+    check_ptr!(out_txid);
+
+    let core = unwrap_option_or_return!(CORE_WALLET_STORAGE.with_item(core_handle, |w| w.clone()));
```
thepastaclaw, line 87: **🟡 Suggestion: Clear the deferred-broadcast txid output before validation**

This public C ABI validates `out_txid` but does not publish a null sentinel before wallet lookup and token validation. Invalid handles, consumed tokens, mismatches, and stale tokens therefore leave the caller's previous output value untouched. The sibling `core_wallet_broadcast_transaction` explicitly nulls this output before fallible work and documents that operational errors leave it null. Matching that convention prevents direct C or Swift consumers from observing or cleaning up a stale pointer when an error is returned.

```suggestion
    check_ptr!(out_txid);
    *out_txid = std::ptr::null_mut();

    let core = unwrap_option_or_return!(CORE_WALLET_STORAGE.with_item(core_handle, |w| w.clone()));
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/core_wallet/transaction_builder.rs
```diff
@@ -134,6 +180,185 @@ pub unsafe extern "C" fn core_wallet_tx_builder_finalize(
...
+) -> PlatformWalletFFIResult {
+    check_ptr!(builder);
+    check_ptr!(core_signer_handle);
+    check_ptr!(out_token);
+    check_ptr!(out_fee);
+    check_ptr!(out_txid);
+    check_ptr!(out_tx);
+    check_ptr!(out_bytes_ptr);
+    check_ptr!(out_bytes_len);
+    *out_token = 0;
```
thepastaclaw, line 237: **🟡 Suggestion: Initialize every deferred-finalize output before fallible work**

Only `out_token` is initialized before wallet resolution, network validation, signing, and registration. On an error, `out_fee`, `out_txid`, `out_tx`, `out_bytes_ptr`, and `out_bytes_len` retain caller-supplied contents. The in-tree JNI caller independently zeroes its storage, but direct C or Swift consumers must otherwise know this hidden cleanup precondition. Publish deterministic zero/null values for every output immediately after validating the pointers, while preserving the existing success assignments.

```suggestion
    *out_token = 0;
    *out_fee = 0;
    *out_txid = std::ptr::null_mut();
    out_tx.write(FFICoreTransaction {
        tx_bytes: std::ptr::null_mut(),
        tx_len: 0,
        fee: 0,
    });
    *out_bytes_ptr = std::ptr::null();
    *out_bytes_len = 0;
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The deferred-payment ownership model is strong, but two blocking lifecycle defects remain: stale tokens can strand reservations that are still safely owner-releasable, and same-ID recreation can lose the new generation's identity or shielded side-manager state during old-generation teardown. Four non-blocking FFI/JNI/documentation issues should also be corrected for deterministic cleanup and consistent error reporting.
Source: reviewer backend models: gpt-5.6-sol (codex-general), gpt-5.6-sol (codex-rust-quality), gpt-5.6-sol (codex-ffi-engineer); final verifier backend model: gpt-5.6-sol. Orchestration-only, not reviewer evidence: openclaw-agent/cliproxy/gpt-5.6-sol.

### Review provenance

🔴 2 blocking | 🟡 3 suggestion(s) | 💬 1 nitpick(s)

<details>

```

In `packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs:504-554: Release stale entries with their retained owner token
  Both the stale-broadcast branch and `reconcile_removed_entry` discard an expired payment without releasing its reservation. The exact key-wallet revision used here retains per-reservation ownership and implements `release_reservation_if_owner` atomically: it removes inputs only while they still belong to this payment and becomes a no-op after a TTL sweep or re-reservation transfers ownership. The current tests advance 22 blocks, below key-wallet's 24-block TTL, prove that the original reservation is still held, and require an immediate rebuild to fail. Consequently, the stale-token error instructs callers to rebuild while leaving the required inputs unavailable for several more blocks, and explicit `releaseReservation` reports success without releasing them. Run the owner-guarded reconciliation for stale entries whenever `funding_reservation_token` is present; retain the no-release fallback only for the unowned `None` case.

In `packages/rs-platform-wallet/src/manager/wallet_lifecycle.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/manager/wallet_lifecycle.rs:721-816: Prevent old-generation teardown from deleting recreated wallet state
  Removing G1 from the inner wallet manager frees `wallet_id`, allowing the deliberately supported concurrent G2 registration to publish a new generation. Teardown then unconditionally calls `coordinator.unregister_wallet(*wallet_id)` and unregisters G1's identity IDs, but both side registries are keyed only by wallet or identity ID rather than `WalletGeneration`. If G2 binds shielded state or registers the same identity-sync row before G1 reaches these awaited removals, G1 deletes G2's newly installed state. The generation check around `self.wallets.remove` protects only the public wallet map. Keep same-ID registration fenced until all ID-keyed teardown is complete, or make these side registries and unregister operations generation-aware; extend the midpoint recreation test to install G2 side-manager state and verify that it survives.

- [SUGGESTION] packages/rs-unified-sdk-jni/src/wallet_manager.rs:1319-1323: Release the registered token when defensive txid validation fails
  A successful `core_wallet_signed_payment_finalize` has already inserted the payment into the process-global registry and transferred reservation ownership to `token`. If this defensive null-txid branch fires because of an FFI regression, it frees the transaction buffer but returns without releasing that token, orphaning the registry entry and reservation until the TTL backstop. The later byte-array allocation failure path correctly releases the token; this post-registration failure path must do the same.

In `packages/rs-platform-wallet-ffi/src/core_wallet/signed_payment.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/core_wallet/signed_payment.rs:81-83: Clear the deferred-broadcast txid output before validation
  This public C ABI validates `out_txid` but does not publish a null sentinel before wallet lookup and token validation. Invalid handles, consumed tokens, mismatches, and stale tokens therefore leave the caller's previous output value untouched. The sibling `core_wallet_broadcast_transaction` explicitly nulls this output before fallible work and documents that operational errors leave it null. Matching that convention prevents direct C or Swift consumers from observing or cleaning up a stale pointer when an error is returned.

- [SUGGESTION] packages/rs-platform-wallet-ffi/src/core_wallet/transaction_builder.rs:225-233: Initialize every deferred-finalize output before fallible work
  Only `out_token` is initialized before wallet resolution, network validation, signing, and registration. On an error, `out_fee`, `out_txid`, `out_tx`, `out_bytes_ptr`, and `out_bytes_len` retain caller-supplied contents. The in-tree JNI caller independently zeroes its storage, but direct C or Swift consumers must otherwise know this hidden cleanup precondition. Publish deterministic zero/null values for every output immediately after validating the pointers, while preserving the existing success assignments.
```
</details>

packages/rs-unified-sdk-jni/src/wallet_manager.rs
```diff
@@ -1225,6 +1225,200 @@ pub extern "system" fn Java_org_dashfoundation_dashsdk_ffi_WalletManagerNative_c
...
+        let txid = unsafe { CStr::from_ptr(out_txid) }
+            .to_string_lossy()
+            .into_owned();
+        unsafe { platform_wallet_ffi::core_wallet_free_address(out_txid) };
+        env.new_string(txid)
+            .map(|s| s.into_raw())
+            .unwrap_or(ptr::null_mut())
+    })
```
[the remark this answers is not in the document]
thepastaclaw: `JNIEnv::new_string` uses checked `NewStringUTF`; allocation failure leaves a pending JVM exception, normally `OutOfMemoryError`, while returning `Err`. The current null return therefore propagates that exception rather than producing a bare null result. `throw_sdk_exception` also immediately returns when an exception is pending, so the suggested change would not alter the outcome.

packages/rs-platform-wallet/src/wallet/signed_payment_registry.rs
QuantumExplorer: Fixed in 9d4027e688 (and mirrored onto the V2 aged-abandon path on #4309, 656aa70166 + b9200c075f): the stale-broadcast branch and `reconcile_removed_entry` now release owner-guarded whenever `funding_reservation_token` is present — safe at any age since `release_reservation_if_owner` no-ops once ownership transferred — and only a token-less entry keeps the drop-without-release fallback. The stale module doc claiming key-wallet exposes no ownership check is corrected, and the rebuild-fails tests flipped to rebuild-succeeds on both surfaces.

packages/rs-platform-wallet/src/manager/wallet_lifecycle.rs
QuantumExplorer: Fixed in aa5d4e8960: all id-keyed side-registry teardown (shielded-coordinator unregister, identity-sync unregisters) now runs BEFORE the inner-manager removal frees the wallet id — `insert_wallet` is the create path's commit point, so a same-id G2 cannot commit until the id frees, making the deletion window unreachable rather than narrow. Identities added mid-removal are surfaced with a warning instead of unregistered post-removal (which would reopen the window for G2's rows). The midpoint recreation test now threads a shared identity through both generations and asserts G2's re-registered identity-sync row survives.

packages/rs-unified-sdk-jni/src/wallet_manager.rs
QuantumExplorer: Fixed in 9f994eb5cd: the defensive null-txid branch now releases the registered token before throwing, same policy as the byte-array failure path.

packages/rs-platform-wallet-ffi/src/core_wallet/signed_payment.rs
QuantumExplorer: Fixed in 9f994eb5cd: `core_wallet_signed_payment_broadcast` publishes the null `out_txid` sentinel before wallet lookup and token validation.

packages/rs-platform-wallet-ffi/src/core_wallet/transaction_builder.rs
QuantumExplorer: Fixed in 9f994eb5cd: `core_wallet_signed_payment_finalize` initializes every out param (token, fee, txid, tx struct, bytes ptr/len) before any fallible work.

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/errors/DashSdkError.kt
```diff
@@ -345,6 +420,16 @@ sealed class DashSdkError(
...
+            26 -> PlatformWallet.TransactionBroadcastRejected(message, cause) // ErrorTransactionBroadcastRejected
+            // The deferred-token trio sits at the contiguous block 34-36 because
+            // 27-33 are claimed elsewhere: 27 ErrorShutdownIncomplete
+            // (dashpay/platform#4268, merged), 29 ErrorAssetLockInsufficientFunds
+            // (#4184), 31 ErrorSigningKeyUnavailable (#4183/#4259), 32
+            // ErrorTransactionBuild (#4247/#4256), 33 ErrorTransactionSigning
+            // (#4256). See packages/rs-platform-wallet-ffi/ERROR_CODE_REGISTRY.md.
+            34 -> PlatformWallet.StaleReservationToken(message, cause) // ErrorStaleReservationToken
+            35 -> PlatformWallet.ReservationTokenConsumed(message, cause) // ErrorReservationTokenConsumed
+            36 -> PlatformWallet.ReservationWalletMismatch(message, cause) // ErrorReservationWalletMismatch
```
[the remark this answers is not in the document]
QuantumExplorer: Fixed in 9f994eb5cd: the stale 'codes 26-30 reserved' comment now states the actual claim map (26-33 claimed by merged/sibling PRs; the deferred trio at 34-36) and defers to `PlatformWalletFFIResultCode` as the authoritative registry.

QuantumExplorer:
Conflicts and review findings are all addressed on head `9f994eb5cd`:

- **Merged `v4.2-dev`** (`d79c4ab0b2`) — union-resolved the both-sides-adds with #4319's signMessage and adapted its merged-in fixtures to this branch's `PlatformWalletInfo.generation`.
- **Both blockers fixed**: stale deferred entries now reconcile owner-guarded instead of stranding still-held inputs (`9d4027e688`, mirrored on #4309's V2 aged-abandon), and id-keyed side-registry teardown completes before the wallet id frees, with the midpoint recreation test extended to prove a recreated generation's identity-sync row survives (`aa5d4e8960`).
- **All three FFI-hygiene suggestions + the stale code comment** fixed (`9f994eb5cd`).

549 + 238 lib tests green on this branch (and on the stacked #4309 after its base merge), clippy `-D warnings` clean, fmt clean. One note for the description: the error-code section is out of date — the renumber it says is pending landed earlier; the deferred trio sits at 34/35/36 with the 27-33 claim map documented in `PlatformWalletFFIResultCode`.

_🤖 Posted by [Claude Code](https://claude.com/claude-code)_

QuantumExplorer (dismissed):

QuantumExplorer (approved):

PR #4333, opened 2026-08-07
fix(rs-sdk-ffi): build the voter identifier from ProTxHash byte order

QuantumExplorer:
## The bug

Masternode votes could not be cast at all. Platform rejected every one, and after the diagnostics in this PR the reason read:

```
No voting identity exists on Platform for masternode 9HsdR37… with this
voting key (expected voter identity EbKFWo8…)
```

The identity genuinely did not exist — because the FFI asked for the wrong one.

**`ProTxHash` and `Txid` use opposite display conventions**, declared a few lines apart in `hash_types.rs`:

```rust
pub struct Txid(sha256d::Hash);        // no attribute ⇒ backward
#[hash_newtype(forward)]
pub struct ProTxHash(sha256d::Hash);   // ⇒ forward
```

`rpc-json`'s `MasternodeListItem` carries both side by side (`pro_tx_hash: ProTxHash`, `collateral_hash: Txid`), and drive-abci derives the voter identity from `masternode.pro_tx_hash.to_byte_array()` — display order. Callers hand this FFI **wire** order, which is what `reg.txid()` yields and what the iOS wallet stores. The two are exact reverses.

That broke both consumers of the value: `create_voter_identifier`, and the `pro_tx_hash` the `MasternodeVoteTransition` itself carries. Reversed once at entry so they cannot drift apart.

Verified arithmetically against a live device failure: the wallet's key hashes to the node's registered voting address `XgYHAS…`, and `SHA256(reversed protx ‖ hash160)` is the identity Platform holds, where the wire-order hash produced one it does not.

## Also in here

**Diagnosis on the failure path.** Platform reports both "no voter identity" and "rotation disabled key 0" as the same opaque `Public key 0 doesn't exist`. The identity is now fetched *only after a broadcast fails*, to say which — the key itself is built locally, since `create_voter_identity_v0` always assigns id 0 and a rotation creates a *different* identity whose key is also 0.

An earlier revision fetched before every cast. That bought no correctness and cost a round trip per (node, contest) — a bulk vote of 6 nodes across 10 names paid 60 fetches on runs that overwhelmingly succeed. An unrelated failure (network, fees, closed poll) now survives unchanged rather than being recast as a key problem, and a diagnostic fetch that itself fails leaves the original error intact.

**Whitespace in the diagnostics.** Both new messages were multi-line literals without `\` continuations, so source indentation rendered as blank gaps mid-sentence on device.

**A corrected claim.** The original description said the key id was unreliable. It isn't — review was right, and the module doc now states what actually fails.

## Testing

9/9 in this module. `select_voting_key` is covered directly: the matching enabled key is selected; a key at a **non-zero id** behind an `AUTHENTICATION` key at 0 is still found (so selection can't regress to picking by position); a disabled key is rejected with the rotation diagnostic; a wrong address and a matching address under the wrong key type are both rejected. Plus one pinning that the missing-identity message names both identifiers and contains no whitespace runs.

**Not verified:** no vote has been confirmed broadcasting end to end. The byte-order fix is established arithmetically and against a real device failure, but the successful path is untested. The byte orientation itself has no test — the two conventions living in one struct makes that worth adding, and the existing tests all short-circuit before reaching `create_voter_identifier`.

Depends on nothing; #4338 (merged) fixed a separate derivation bug in the same flow.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved contested-resource vote casting by correctly handling transaction hash byte order.
  - Added clearer diagnostics when voter identities or suitable enabled voting keys are missing.
  - Preserved original error details for unrelated broadcast failures.
  - Improved handling of voting-key rotation and mismatches.

- **Documentation**
  - Clarified that transaction hashes must use 32-byte wire order rather than Core display order when casting votes.

### DIFF
--- a/packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
+++ b/packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
@@ -287,9 +287,26 @@ unsafe fn cast_vote_inner(
     }));
 
     // ---- pro_tx_hash ---------------------------------------------------------
+    // Callers pass WIRE order — the orientation `Txid` stores, which is what a
+    // parsed ProRegTx yields (`reg.txid()`) and what the iOS wallet holds.
+    //
+    // Platform identifies masternodes by the OTHER orientation. `ProTxHash` is
+    // declared `#[hash_newtype(forward)]` while `Txid` is not, so
+    // `ProTxHash::to_byte_array()` is display order — the reverse of a `Txid`'s
+    // bytes for the same transaction. `rpc-json`'s `MasternodeListItem` holds
+    // both conventions side by side (`pro_tx_hash: ProTxHash`,
+    // `collateral_hash: Txid`), and drive-abci builds the voter identity from
+    // `masternode.pro_tx_hash.to_byte_array()`.
+    //
+    // Feeding wire order to `create_voter_identifier` therefore asks Platform
+    // for an identity that has never existed, and the vote is rejected as
+    // having no voter identity — with the real one sitting under the reversed
+    // hash. Reverse once, here, and use the corrected value for both the
+    // identifier and the transition so they cannot drift apart.
     let pro_tx_hash_slice = std::slice::from_raw_parts(voter_pro_tx_hash, 32);
     let mut pro_tx_hash_arr = [0u8; 32];
     pro_tx_hash_arr.copy_from_slice(pro_tx_hash_slice);
+    pro_tx_hash_arr.reverse();
     let pro_tx_hash = Identifier::new(pro_tx_hash_arr);
 
     // ---- Derive the masternode voting key + signer --------------------------
--- a/packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
+++ b/packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
@@ -20,12 +20,18 @@
 //! This FFI therefore takes the raw 32-byte **voting private key** plus the
 //! 32-byte `pro_tx_hash`. From the private key it derives:
 //!   * a [`SingleKeySigner`] that signs the transition, and
-//!   * the matching `ECDSA_HASH160` [`IdentityPublicKey`] (the masternode
-//!     voting key) whose `data` is `hash160(pubkey)`.
+//!   * `hash160(pubkey)` — the voting address, which identifies both the voter
+//!     identity (`create_voter_identifier`) and the key on it.
+//!
+//! The voting [`IdentityPublicKey`] is then **fetched from the voter
+//! identity**, not fabricated: its key id is only 0 for an identity Platform
+//! created and never rotated, so assuming 0 fails with an opaque
+//! "Public key 0 doesn't exist" whenever the identity is absent or the key
+//! moved. Looking it up also lets a mismatched key be reported as such.
 //!
 //! The `SingleKeySigner::can_sign_with` check for `ECDSA_HASH160` recomputes
-//! `hash160(pubkey)` from the same private key, so the derived key and signer
-//! always agree.
+//! `hash160(pubkey)` from the same private key, so the fetched key and the
+//! signer agree.
 //!
 //! A regular wallet is **not** a masternode and has no voting key, so a vote
 //! broadcast from such a wallet reaches a deterministic *authorization*
@@ -39,9 +45,11 @@ use crate::types::{FFINetwork, Network, SDKHandle};
 use crate::{DashSDKResult, FFIError};
 use dash_sdk::dpp::dashcore::hashes::{hash160, Hash};
 use dash_sdk::dpp::dashcore::secp256k1::{PublicKey, Secp256k1, SecretKey};
-use dash_sdk::dpp::identity::identity_public_key::v0::IdentityPublicKeyV0;
-use dash_sdk::dpp::identity::{IdentityPublicKey, KeyType, Purpose, SecurityLevel};
-use dash_sdk::dpp::platform_value::{BinaryData, Identifier, Value};
+use dash_sdk::dpp::identifier::MasternodeIdentifiers;
+use dash_sdk::dpp::identity::accessors::IdentityGettersV0;
+use dash_sdk::dpp::identity::identity_public_key::accessors::v0::IdentityPublicKeyGettersV0;
+use dash_sdk::dpp::identity::{Identity, IdentityPublicKey, KeyType, Purpose};
+use dash_sdk::dpp::platform_value::{Identifier, Value};
 use dash_sdk::dpp::voting::vote_choices::resource_vote_choice::ResourceVoteChoice;
 use dash_sdk::dpp::voting::vote_polls::contested_document_resource_vote_poll::ContestedDocumentResourceVotePoll;
 use dash_sdk::dpp::voting::vote_polls::VotePoll;
@@ -49,6 +57,7 @@ use dash_sdk::dpp::voting::votes::resource_vote::v0::ResourceVoteV0;
 use dash_sdk::dpp::voting::votes::resource_vote::ResourceVote;
 use dash_sdk::dpp::voting::votes::Vote;
 use dash_sdk::platform::transition::vote::PutVote;
+use dash_sdk::platform::Fetch;
 use simple_signer::SingleKeySigner;
 use std::ffi::{c_char, CStr};
 use zeroize::Zeroizing;
@@ -285,9 +294,7 @@ unsafe fn cast_vote_inner(
         .map_err(|e| invalid(&format!("Invalid voting private key: {}", e)))?;
 
     // The masternode voting key is the 20-byte hash160 of the (compressed)
-    // public key, packaged as an ECDSA_HASH160 / VOTING / HIGH key with id 0 —
-    // exactly the shape Platform assigns to voter identities
-    // (`get_voter_identity_key_v0`). `MasternodeVoteTransition` calls
+    // public key. `MasternodeVoteTransition` calls
     // `masternode_voting_key.public_key_hash()` to derive the voter
     // identifier, and `SingleKeySigner::can_sign_with` recomputes the same
     // hash160 from the private key, so the two agree by construction.
@@ -297,22 +304,51 @@ unsafe fn cast_vote_inner(
     let public_key = PublicKey::from_secret_key(&secp, &secret_key);
     let voting_address = hash160::Hash::hash(&public_key.serialize()).to_byte_array();
 
-    let masternode_voting_key: IdentityPublicKey = IdentityPublicKeyV0 {
-        id: 0,
-        purpose: Purpose::VOTING,
-        security_level: SecurityLevel::HIGH,
-        key_type: KeyType::ECDSA_HASH160,
-        read_only: true,
-        data: BinaryData::new(voting_address.to_vec()),
-        disabled_at: None,
-        contract_bounds: None,
-    }
-    .into();
-
-    // ---- Broadcast ----------------------------------------------------------
+    // ---- Resolve the real voting key off the voter identity -----------------
     let wrapper = &*(sdk_handle as *const SDKWrapper);
     let sdk = &wrapper.sdk;
 
+    let voter_identifier = Identifier::create_voter_identifier(&pro_tx_hash_arr, &voting_address);
+
+    let masternode_voting_key: IdentityPublicKey = wrapper.runtime.block_on(async {
+        // The key id is NOT reliably 0. Platform assigns 0 when it *creates* a
+        // voter identity, but a voting-key rotation disables the old
+        // identity's keys and creates a fresh identity for the new address
+        // (`update_voter_identity_v0`), and nothing guarantees the caller's key
+        // matches the identity that actually exists. Fabricating id 0 turned
+        // every one of those cases into Platform's opaque
+        // "Public key 0 doesn't exist" at broadcast time.
+        let identity = Identity::fetch(sdk, voter_identifier)
+            .await
+            .map_err(FFIError::from)?
+            .ok_or_else(|| {
+                FFIError::InvalidParameter(format!(
+                    "No voting identity exists on Platform for masternode {} with this voting key                      (expected voter identity {}). Either the voting key does not match the                      masternode's registered voting address, or Platform has not created the                      voter identity yet.",
+                    pro_tx_hash, voter_identifier
+                ))
+            })?;
+
+        // Match on the key's own data, not on position: that is what the
+        // signer can actually sign with.
+        identity
+            .public_keys()
+            .values()
+            .find(|key| {
+                key.purpose() == Purpose::VOTING
+                    && key.key_type() == KeyType::ECDSA_HASH160
+                    && key.data().as_slice() == voting_address
+                    && key.disabled_at().is_none()
+            })
+            .cloned()
+            .ok_or_else(|| {
+                FFIError::InvalidParameter(format!(
+                    "Voter identity {} has no enabled ECDSA_HASH160 voting key matching this                      private key. The masternode's voting key may have been rotated.",
+                    voter_identifier
+                ))
+            })
+    })?;
+
+    // ---- Broadcast ----------------------------------------------------------
     wrapper.runtime.block_on(async {
         vote.put_to_platform_and_wait_for_response(
             pro_tx_hash,
--- a/packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
+++ b/packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
@@ -23,25 +23,32 @@
 //!   * `hash160(pubkey)` — the voting address, which identifies both the voter
 //!     identity (`create_voter_identifier`) and the key on it.
 //!
-//! The voting [`IdentityPublicKey`] is then **fetched from the voter
-//! identity** rather than fabricated. Platform does always assign the voting
-//! key id 0 (`create_voter_identity_v0` passes 0, and a rotation creates a
-//! *different* identity — the identifier includes the voting address — whose
-//! key is likewise 0), so the id itself was never the problem. Fabricating a
-//! key meant never checking the two things that do fail:
+//! The voting [`IdentityPublicKey`] is **built locally, not fetched**.
+//! Platform always assigns a voter identity's voting key id 0:
+//! `create_voter_identity_v0` passes 0, and a rotation creates a *different*
+//! identity — the identifier includes the voting address — whose key is
+//! likewise 0. So the key Platform holds is knowable without a round trip, and
+//! `SingleKeySigner::can_sign_with` for `ECDSA_HASH160` recomputes the same
+//! `hash160(pubkey)` from the private key, so key and signer agree by
+//! construction.
 //!
-//!   * the voter identity may not exist for this `(pro_tx_hash, voting
-//!     address)` pair at all, and
-//!   * after a rotation `update_voter_identity_v0` **disables** the old
-//!     identity's keys, so key 0 can exist but be unusable.
+//! # Diagnosis is deferred to the failure path
 //!
-//! Both surfaced as Platform's opaque "Public key 0 doesn't exist" at
-//! broadcast time. Fetching the identity and matching on the key's own data
-//! (purpose, key type, `data`, not-disabled) reports each case for what it is.
+//! Two things do fail, and Platform reports both as the same opaque
+//! "Public key 0 doesn't exist":
 //!
-//! The `SingleKeySigner::can_sign_with` check for `ECDSA_HASH160` recomputes
-//! `hash160(pubkey)` from the same private key, so the fetched key and the
-//! signer agree.
+//!   * no voter identity exists for this `(pro_tx_hash, voting address)` pair,
+//!     or
+//!   * after a rotation `update_voter_identity_v0` **disabled** the old
+//!     identity's keys, so key 0 exists but is unusable.
+//!
+//! Telling those apart needs the identity, but fetching it before every cast
+//! would spend a Platform round trip per (node, contest) on runs that
+//! overwhelmingly succeed — a bulk vote of 6 nodes across 10 names is 60
+//! fetches. So the fetch happens only after a broadcast has already failed,
+//! where the cost is paid on a path that is already lost. A diagnosis replaces
+//! the opaque error; an unrelated failure (network, fees, a closed poll)
+//! survives unchanged rather than being recast as a key problem.
 //!
 //! A regular wallet is **not** a masternode and has no voting key, so a vote
 //! broadcast from such a wallet reaches a deterministic *authorization*
@@ -58,7 +65,9 @@ use dash_sdk::dpp::dashcore::secp256k1::{PublicKey, Secp256k1, SecretKey};
 use dash_sdk::dpp::identifier::MasternodeIdentifiers;
 use dash_sdk::dpp::identity::accessors::IdentityGettersV0;
 use dash_sdk::dpp::identity::identity_public_key::accessors::v0::IdentityPublicKeyGettersV0;
-use dash_sdk::dpp::identity::{Identity, IdentityPublicKey, KeyType, Purpose};
+use dash_sdk::dpp::identity::identity_public_key::v0::IdentityPublicKeyV0;
+use dash_sdk::dpp::identity::{Identity, IdentityPublicKey, KeyType, Purpose, SecurityLevel};
+use dash_sdk::dpp::platform_value::BinaryData;
 use dash_sdk::dpp::platform_value::{Identifier, Value};
 use dash_sdk::dpp::voting::vote_choices::resource_vote_choice::ResourceVoteChoice;
 use dash_sdk::dpp::voting::vote_polls::contested_document_resource_vote_poll::ContestedDocumentResourceVotePoll;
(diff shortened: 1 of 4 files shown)


thepastaclaw:
🕓 Ready for review — next in queue (commit 54683fa9)
_Queue position: 1/1_
_ETA: start ~10:10 UTC · complete ~10:24 UTC (median 14m across 30 recent reviews; 2 slots)_
_Queued 56s ago · Last checked: 2026-08-08 10:10 UTC_

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
```diff
@@ -297,22 +304,51 @@ unsafe fn cast_vote_inner(
...
+        // (`update_voter_identity_v0`), and nothing guarantees the caller's key
+        // matches the identity that actually exists. Fabricating id 0 turned
+        // every one of those cases into Platform's opaque
+        // "Public key 0 doesn't exist" at broadcast time.
+        let identity = Identity::fetch(sdk, voter_identifier)
+            .await
+            .map_err(FFIError::from)?
+            .ok_or_else(|| {
+                FFIError::InvalidParameter(format!(
+                    "No voting identity exists on Platform for masternode {} with this voting key                      (expected voter identity {}). Either the voting key does not match the                      masternode's registered voting address, or Platform has not created the                      voter identity yet.",
```
thepastaclaw: **🟡 Suggestion: "No voter identity" diagnostic contains literal embedded whitespace runs**

This format! literal was written across multiple source lines without a `\` continuation, so the line-wrap indentation became literal characters inside the string. Verified directly in the file: line 326 contains runs of ~22 consecutive spaces before `(expected`, before `masternode's`, and before `voter identity yet.`. This string is exactly the diagnostic this PR exists to add (per the PR description: "Errors are now diagnostic") and crosses the FFI boundary as the `message` field of `DashSDKError`, so a Swift/iOS caller displaying it to the user sees large mid-sentence gaps instead of a clean message.

```suggestion
                FFIError::InvalidParameter(format!(
                    concat!(
                        "No voting identity exists on Platform for masternode {} with this voting key ",
                        "(expected voter identity {}). Either the voting key does not match the ",
                        "masternode's registered voting address, or Platform has not created the ",
                        "voter identity yet."
                    ),
                    pro_tx_hash, voter_identifier
                ))
```

<sub>source: ['claude', 'codex', 'coderabbit']</sub>

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
```diff
@@ -297,22 +304,51 @@ unsafe fn cast_vote_inner(
...
+            .find(|key| {
+                key.purpose() == Purpose::VOTING
+                    && key.key_type() == KeyType::ECDSA_HASH160
+                    && key.data().as_slice() == voting_address
+                    && key.disabled_at().is_none()
+            })
+            .cloned()
+            .ok_or_else(|| {
+                FFIError::InvalidParameter(format!(
+                    "Voter identity {} has no enabled ECDSA_HASH160 voting key matching this                      private key. The masternode's voting key may have been rotated.",
```
thepastaclaw: **🟡 Suggestion: "Voting key rotated" diagnostic contains the same literal whitespace defect**

Same root cause as the sibling message at line 326: the format! literal contains a run of ~22 literal spaces between "matching this" and "private key." This is the second of the two new diagnostics the PR advertises as its main improvement, and it will render with the same garbled gap to callers and logs.

```suggestion
                FFIError::InvalidParameter(format!(
                    concat!(
                        "Voter identity {} has no enabled ECDSA_HASH160 voting key matching this ",
                        "private key. The masternode's voting key may have been rotated."
                    ),
                    voter_identifier
                ))
```

<sub>source: ['claude', 'codex', 'coderabbit']</sub>

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
```diff
@@ -297,22 +304,51 @@ unsafe fn cast_vote_inner(
...
+                    voter_identifier
+                ))
+            })
+    })?;
```
thepastaclaw: **🟡 Suggestion: New identity-fetch/key-selection logic has no test coverage**

All three existing tests in this file return before reaching this code (null handle, missing contender, invalid vote choice — all short-circuit before line 307). Nothing exercises: a missing voter identity producing the new `InvalidParameter`, an enabled matching key being correctly selected (propagating its real, possibly non-zero-position id into signing), or a disabled/mismatched key being rejected. This is the core of the fix and the PR description itself flags it as "not yet verified against a live vote" — factoring the lookup into a testable helper (or configuring the mock SDK with an identity fixture) would give real regression coverage for the exact bug this PR fixes.

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
```diff
@@ -20,12 +20,18 @@
...
-//!   * the matching `ECDSA_HASH160` [`IdentityPublicKey`] (the masternode
-//!     voting key) whose `data` is `hash160(pubkey)`.
+//!   * `hash160(pubkey)` — the voting address, which identifies both the voter
+//!     identity (`create_voter_identifier`) and the key on it.
+//!
+//! The voting [`IdentityPublicKey`] is then **fetched from the voter
+//! identity**, not fabricated: its key id is only 0 for an identity Platform
+//! created and never rotated, so assuming 0 fails with an opaque
+//! "Public key 0 doesn't exist" whenever the identity is absent or the key
+//! moved. Looking it up also lets a mismatched key be reported as such.
```
thepastaclaw: **💬 Nitpick: Module doc's key-id/rotation explanation doesn't match actual Platform behavior**

The doc claims the key id is "only 0 for an identity Platform created and never rotated" and that assuming 0 fails when "the key moved." Tracing `create_voter_identity_v0` and `update_voter_identity_v0` in rs-drive-abci shows every voter identity's voting key is always created with id 0 — rotation never produces a non-zero id; it either re-enables an existing (also id-0) identity's key or creates a brand-new identity whose key is also id 0. The real risk this PR fixes is that the *identity itself* may not exist for the given (pro_tx_hash, voting_address) pair, or the matching key may be disabled — not that a valid key's id drifts from 0. This is low-stakes since the actual fix logic is correct regardless, but the doc could mislead a future maintainer about what invariant it's compensating for.

<sub>source: ['codex']</sub>

thepastaclaw (commented):

The fix correctly replaces a fabricated ECDSA_HASH160 key with one resolved via a proof-verified fetch of the actual voter identity, matching by purpose/type/data/enabled-state rather than a hardcoded key id 0. Logic, imports, and identifier derivation are correct and this is a clear net improvement over the previous behavior. Two real defects remain: both new diagnostic error strings (the PR's stated deliverable) contain long runs of literal embedded whitespace from unescaped multi-line format! literals, and the core identity-resolution/key-selection logic has no test coverage. A module-doc nitpick about key-id/rotation semantics is also worth a follow-up. No blocking consensus, security, or FFI-boundary issues were found.
Source: Reviewer backends — gpt-5.6-sol (codex: general, ffi-engineer, security-auditor), claude-sonnet-5 (opus-lane: general, ffi-engineer, security-auditor); final verifier backend — claude-sonnet-5 (final-verifier-sonnet-1-951549e47466468fab7200db619eee5b). openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — ffi-engineer (completed), `gpt-5.6-sol` — security-auditor (completed)
- Sonnet reviewers: `claude-sonnet-5` — general (completed), `claude-sonnet-5` — ffi-engineer (completed), `claude-sonnet-5` — security-auditor (completed)

<details>

```

In `packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs`:
- [SUGGESTION] packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs:326: "No voter identity" diagnostic contains literal embedded whitespace runs
  This format! literal was written across multiple source lines without a `\` continuation, so the line-wrap indentation became literal characters inside the string. Verified directly in the file: line 326 contains runs of ~22 consecutive spaces before `(expected`, before `masternode's`, and before `voter identity yet.`. This string is exactly the diagnostic this PR exists to add (per the PR description: "Errors are now diagnostic") and crosses the FFI boundary as the `message` field of `DashSDKError`, so a Swift/iOS caller displaying it to the user sees large mid-sentence gaps instead of a clean message.
- [SUGGESTION] packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs:345: "Voting key rotated" diagnostic contains the same literal whitespace defect
  Same root cause as the sibling message at line 326: the format! literal contains a run of ~22 literal spaces between "matching this" and "private key." This is the second of the two new diagnostics the PR advertises as its main improvement, and it will render with the same garbled gap to callers and logs.
- [SUGGESTION] packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs:313-349: New identity-fetch/key-selection logic has no test coverage
  All three existing tests in this file return before reaching this code (null handle, missing contender, invalid vote choice — all short-circuit before line 307). Nothing exercises: a missing voter identity producing the new `InvalidParameter`, an enabled matching key being correctly selected (propagating its real, possibly non-zero-position id into signing), or a disabled/mismatched key being rejected. This is the core of the fix and the PR description itself flags it as "not yet verified against a live vote" — factoring the lookup into a testable helper (or configuring the mock SDK with an identity fixture) would give real regression coverage for the exact bug this PR fixes.
```
</details>

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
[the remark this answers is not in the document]
QuantumExplorer: Fixed in `04b6edf8e9`. `Identity::fetch` needs a live SDK so the identity *lookup* stays integration-shaped, but you're right that key *selection* is the part deciding whether a vote can be signed — and it's pure. Extracted as `select_voting_key` and covered by five tests: the matching enabled key is selected; a key at a non-zero id is still found (proving selection is by data, not position); a disabled key is rejected with the rotation diagnostic; a different address is rejected; a matching address under the wrong key type is rejected. Plus one pinning `missing_voter_identity` to name both identifiers. 9/9 pass.

QuantumExplorer: Fixed in `04b6edf8e9` — see the reply on the sibling thread. `select_voting_key` is now extracted and has five tests, including one that places the voting key at id 4 behind an AUTHENTICATION key at id 0, which fails if selection ever regresses to picking by position.

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
QuantumExplorer: Fixed in `ca6463ef7b` — `\` continuations on both literals. Verified the rendered output rather than the source: compiled the format! and asserted no double spaces survive, and there's now a test pinning that the message contains no whitespace runs, since it crosses the FFI boundary verbatim.

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
QuantumExplorer: Fixed in `ca6463ef7b` alongside the sibling literal.

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
QuantumExplorer: You're right and I was wrong. Traced it: `create_voter_identity_v0` passes key id 0, and `update_voter_identity_v0` disables the old identity's keys then creates a *new* identity (the identifier includes the voting address) whose key is also 0. So rotation never yields a non-zero id — the id was never the problem.

The fix still holds, for the two things fabricating a key skipped: the identity may not exist for this (pro_tx_hash, voting address) pair, and after a rotation key 0 can exist but be **disabled**. Both surfaced as "Public key 0 doesn't exist". Module doc corrected in `04b6edf8e9` to say that instead.

Confirmed against a real device failure since: the error now reads "No voting identity exists on Platform for masternode … (expected voter identity …)" — the missing-identity branch, exactly as you describe.

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
```diff
@@ -456,4 +521,152 @@ mod tests {
...
+        // Same 20 bytes, but not a key this signer can sign with.
+        let identity = identity_with(vec![key(
+            0,
+            Purpose::VOTING,
+            KeyType::BIP13_SCRIPT_HASH,
+            VOTING_ADDRESS,
+            None,
+        )]);
+        assert!(select_voting_key(&identity, &VOTING_ADDRESS, &voter_id()).is_err());
+    }
```
thepastaclaw, line 866: **🟡 Suggestion: Cover the voting-purpose predicate independently**

`select_voting_key` checks `key.purpose() == Purpose::VOTING`, but no test isolates that condition: the only `Purpose::AUTHENTICATION` fixture (in `selects_by_data_not_by_position`) also uses `OTHER_ADDRESS`, so it's already rejected by the address check. If the purpose check were deleted entirely, every existing test would still pass. Add a fixture with an `AUTHENTICATION` key at the correct `VOTING_ADDRESS` and `ECDSA_HASH160` type to pin that the purpose predicate is actually load-bearing.

```suggestion
    #[test]
    fn rejects_a_matching_address_under_the_wrong_purpose() {
        let identity = identity_with(vec![key(
            0,
            Purpose::AUTHENTICATION,
            KeyType::ECDSA_HASH160,
            VOTING_ADDRESS,
            None,
        )]);
        assert!(select_voting_key(&identity, &VOTING_ADDRESS, &voter_id()).is_err());
    }


```

thepastaclaw (commented):

This delta (52683dc9..04b6edf8) is a follow-up cleanup of the previously-reviewed fix: it corrects the two whitespace-mangled diagnostic strings with `\` line continuations, extracts key selection into a pure `select_voting_key` helper backed by nine unit tests, and fixes the module doc's inaccurate key-id/rotation claim. All four prior findings are verified fixed against the current head. Two minor issues remain out of the prior set: the module doc's 'who can actually cast a vote' section still describes the old platform-side rejection path even though a missing voter identity now fails locally before broadcast, and the `select_voting_key` test suite lacks a case that isolates the `Purpose::VOTING` predicate (removing it would not currently fail any test). A pre-existing, PR-unrelated Swift-side issue (processVoidResult from PR #3883 discards FFI error codes) is recorded as an out-of-scope follow-up.
Source: reviewer backends — gpt-5.6-sol (Codex general, ffi-engineer, security-auditor), claude-sonnet-5 (Sonnet general, ffi-engineer, security-auditor); final verifier backend — claude-sonnet-5. openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — ffi-engineer (completed), `gpt-5.6-sol` — security-auditor (completed)

<details>

```

In `packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs`:
- [SUGGESTION] packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs:649-660: Cover the voting-purpose predicate independently
  `select_voting_key` checks `key.purpose() == Purpose::VOTING`, but no test isolates that condition: the only `Purpose::AUTHENTICATION` fixture (in `selects_by_data_not_by_position`) also uses `OTHER_ADDRESS`, so it's already rejected by the address check. If the purpose check were deleted entirely, every existing test would still pass. Add a fixture with an `AUTHENTICATION` key at the correct `VOTING_ADDRESS` and `ECDSA_HASH160` type to pin that the purpose predicate is actually load-bearing.
```
</details>

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
```diff
@@ -268,9 +287,26 @@ unsafe fn cast_vote_inner(
...
+    // Feeding wire order to `create_voter_identifier` therefore asks Platform
+    // for an identity that has never existed, and the vote is rejected as
+    // having no voter identity — with the real one sitting under the reversed
+    // hash. Reverse once, here, and use the corrected value for both the
+    // identifier and the transition so they cannot drift apart.
     let pro_tx_hash_slice = std::slice::from_raw_parts(voter_pro_tx_hash, 32);
     let mut pro_tx_hash_arr = [0u8; 32];
     pro_tx_hash_arr.copy_from_slice(pro_tx_hash_slice);
+    pro_tx_hash_arr.reverse();
     let pro_tx_hash = Identifier::new(pro_tx_hash_arr);
```
thepastaclaw, line 329: **🔴 Blocking: Unconditional reversal breaks display-order proTxHash callers**

This now treats every FFI input as raw `Txid` wire-order bytes and reverses it, but the public Rust, Swift, Kotlin, and JNI contracts only describe a 32-byte proTxHash. The shipped Swift form asks users to paste `pro_tx_hash (64 hex chars)` and decodes those characters verbatim, while `PersistentMasternode.proTxHashHex` explicitly exposes the conventional block-explorer/display-order value. Pasting that value causes this code to reverse it into wire order even though Platform creates the voter identity from the forward/display-oriented `MasternodeListItem.pro_tx_hash.to_byte_array()`, so the proof-verified identity lookup misses and a valid vote cannot be cast. Define one byte-order contract across the public boundary: either accept display-order bytes here without reversing, or retain wire-order input for wallet-internal callers while converting the display-order Swift/Kotlin form input before the FFI call. Pin the chosen conversion with an asymmetric hash and expected voter-identifier regression test.

thepastaclaw (changes requested):

The proof-verified identity fetch and key matching are sound, but the latest unconditional proTxHash reversal silently assumes wire-order input and breaks the shipped Swift path that accepts conventional display-order hex, so changes are required. The two previously verified documentation and test-coverage findings also remain valid at this head.
Source: Reviewer backends — gpt-5.6-sol (Codex general, security-auditor, rust-quality); final verifier backend — gpt-5.6-sol (Codex verifier). openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

### Review provenance

🔴 1 blocking | 💬 1 nitpick(s)

<details>

```

In `packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs`:
- [BLOCKING] packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs:306-310: Unconditional reversal breaks display-order proTxHash callers
  This now treats every FFI input as raw `Txid` wire-order bytes and reverses it, but the public Rust, Swift, Kotlin, and JNI contracts only describe a 32-byte proTxHash. The shipped Swift form asks users to paste `pro_tx_hash (64 hex chars)` and decodes those characters verbatim, while `PersistentMasternode.proTxHashHex` explicitly exposes the conventional block-explorer/display-order value. Pasting that value causes this code to reverse it into wire order even though Platform creates the voter identity from the forward/display-oriented `MasternodeListItem.pro_tx_hash.to_byte_array()`, so the proof-verified identity lookup misses and a valid vote cannot be cast. Define one byte-order contract across the public boundary: either accept display-order bytes here without reversing, or retain wire-order input for wallet-internal callers while converting the display-order Swift/Kotlin form input before the FFI call. Pin the chosen conversion with an asymmetric hash and expected voter-identifier regression test.
```
</details>

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
```diff
@@ -268,9 +287,26 @@ unsafe fn cast_vote_inner(
...
+    //
+    // Feeding wire order to `create_voter_identifier` therefore asks Platform
+    // for an identity that has never existed, and the vote is rejected as
+    // having no voter identity — with the real one sitting under the reversed
+    // hash. Reverse once, here, and use the corrected value for both the
+    // identifier and the transition so they cannot drift apart.
     let pro_tx_hash_slice = std::slice::from_raw_parts(voter_pro_tx_hash, 32);
     let mut pro_tx_hash_arr = [0u8; 32];
     pro_tx_hash_arr.copy_from_slice(pro_tx_hash_slice);
+    pro_tx_hash_arr.reverse();
```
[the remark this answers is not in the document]
thepastaclaw: The byte-order contract must be explicit, but documentation alone is insufficient: the shipped Swift form accepts a conventional display-order proTxHash hex string and decodes it verbatim, while Rust now reverses every input as though it were stored wire-order bytes. Preserve wire-order semantics only if the Swift/Kotlin display-input paths convert before calling the FFI; otherwise accept display-order bytes at the boundary. Add an asymmetric regression vector for the resulting voter identifier.

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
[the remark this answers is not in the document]
QuantumExplorer: Fixed in `3c37a1b9de`. You're right that nothing isolated it — the only AUTHENTICATION fixture also used `OTHER_ADDRESS`, so the address check rejected it first and deleting `purpose() == VOTING` would have left every test green. Added `rejects_a_matching_address_under_the_wrong_purpose`: an AUTHENTICATION key at the CORRECT `VOTING_ADDRESS` and `ECDSA_HASH160` type, where purpose is the only thing that can reject it. 10/10 in the module.

QuantumExplorer: Fixed in `3c37a1b9de`. You're right that nothing isolated it — the only AUTHENTICATION fixture also used `OTHER_ADDRESS`, so the address check rejected it first and deleting `purpose() == VOTING` would have left every test green. Added `rejects_a_matching_address_under_the_wrong_purpose`: an AUTHENTICATION key at the CORRECT `VOTING_ADDRESS` and `ECDSA_HASH160` type, where purpose is the only thing that can reject it. 10/10 in the module.

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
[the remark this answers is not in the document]
QuantumExplorer: Fixed in `3c37a1b9de` — and this is the right thing to have caught, since an undocumented orientation is exactly what caused the bug this PR fixes.

Documented at all three boundaries a caller can touch: the FFI `voter_pro_tx_hash` parameter, Swift `castContestedResourceVote`, and Kotlin `castVote`. Each now states the bytes must be WIRE order — the orientation `Txid` stores, which is what a parsed ProRegTx yields (`reg.txid()`) and what a wallet holds — and NOT the reverse that Core displays, plus why they are not interchangeable: Platform identifies masternodes by the opposite orientation because `ProTxHash` is declared `#[hash_newtype(forward)]` while `Txid` is not.

I did not rename the parameters. `proTxHash` is what the value is called everywhere in the ecosystem, and a name like `proTxHashWireOrder` would be the only place using that convention while still not saying which order Core prints — the doc carries the fact more reliably than the identifier can.

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
QuantumExplorer: Fixed in `3c37a1b9de` — and this is the right thing to have caught, since an undocumented orientation is exactly what caused the bug this PR fixes.

Documented at all three boundaries a caller can touch: the FFI `voter_pro_tx_hash` parameter, Swift `castContestedResourceVote`, and Kotlin `castVote`. Each now states the bytes must be WIRE order — the orientation `Txid` stores, which is what a parsed ProRegTx yields (`reg.txid()`) and what a wallet holds — and NOT the reverse that Core displays, plus why they are not interchangeable: Platform identifies masternodes by the opposite orientation because `ProTxHash` is declared `#[hash_newtype(forward)]` while `Txid` is not.

I did not rename the parameters. `proTxHash` is what the value is called everywhere in the ecosystem, and a name like `proTxHashWireOrder` would be the only place using that convention while still not saying which order Core prints — the doc carries the fact more reliably than the identifier can.

packages/rs-sdk-ffi/src/contested_resource/transitions/cast_vote.rs
```diff
@@ -322,16 +381,98 @@ unsafe fn cast_vote_inner(
...
+        sdk,
+        &pro_tx_hash,
+        &voter_identifier,
+        &voting_address,
+    ));
+
+    // A diagnosis replaces the opaque error; anything else (network, fees,
+    // an already-closed poll) must survive unchanged rather than be recast as
+    // a key problem.
+    Err(diagnosis.unwrap_or_else(|| FFIError::from(broadcast_error)))
```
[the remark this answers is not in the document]
QuantumExplorer: Fixed in `776dbd2c13`. You're right, and this was worse than a missed edge case: my own commit message claimed unrelated errors survived unchanged, while the code diagnosed every failure and replaced whenever the identity happened to be absent. A vote that arrived after its poll closed, cast by a node with no voter identity, would have been reported as "no voting identity exists" — true cause discarded, replaced with a plausible-looking wrong one.

Now gated on the typed consensus error, not its text: only `SignatureError::{IdentityNotFoundError, MissingPublicKeyError, PublicKeyIsDisabledError}` are diagnosable, since those are exactly what fetching the identity can explain. Everything else keeps its own error, and the existing fallback still applies when the identity and key both check out.

Matched structurally rather than on the message because a string match would silently start diagnosing unrelated failures the first time a message changed — the same fragility that caused the original bug.

Tests pin both directions: the three key failures are diagnosable; a signature failure that is NOT about the key's existence or state (`BasicECDSAError`) is not, nor is a broadcast error with no consensus cause. 12/12 in the module.

PR #948, opened 2023-04-20
feat(dpp): document `$createdAt` and `$updatedAt` validation

jawid-h:
## Issue being fixed or feature implemented

Check that `$updatedAt` property is not equal to `$createdAt` for document replace transition 

## What was done?

- added additional error
- added additional validation

## How Has This Been Tested?

- unit tests

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added or updated relevant unit/integration/functional/e2e tests
- [x] I have made corresponding changes to the documentation

**For repository code-owners and collaborators only**
- [x] I have assigned this pull request to a milestone

PR #4336, opened 2026-08-08
fix(platform-wallet): make asset-lock spends visible to every balance reader

HashEngineering:

An SDK-built asset lock leaves every balance reader believing its funding TXOs are still spendable, long after the lock chain-locked and its top-up was credited. Observed live on an Android testnet wallet: the home header and MAX displayed **0.21 DASH on a 0.1 wallet**, the phantom figure was persisted as the next launch's holding balance, and each relaunch re-inflated the engine's lock-free balance back to the phantom (0.22 after a later deposit).

Root cause, one sentence: an asset-lock tx burns its value into the special-tx **payload** and often has no wallet-owned standard output, so SPV block matching can miss it — the spender's transaction record never leaves mempool context, and every downstream "spent" signal keyed on in-block context never fires.

Three readers were affected:

1. **Persistence mirrors** (Kotlin Room / Swift SwiftData): the TXO's `isSpent` flips only when the *spending tx* reaches in-block context — which never arrives — so rows sit at `isSpent = false` with `spendingTxid` set, forever.
2. **Per-account balance aggregates** (`account_balances_blocking`): served the cached per-account `WalletCoreBalance`, which refreshes only when transaction processing runs `update_balance()` — stale indefinitely, while coin selection (reading the live UTXO set) was always right.
3. **UTXO restore on relaunch**: the restore guard skips a spent-linked TXO only when its spending tx is in-block — same missing signal — so every relaunch handed the consumed outputs back to Rust as spendable and the lock-free wallet balance re-inflated.

## What was done?

The tracked lock's **own status** is a signal that provably arrives (the proof wait drives Built → Broadcast → InstantSendLocked/ChainLocked → Consumed upserts), so the fixes key on it:

- **Kotlin + Swift persistence**: the asset-lock upsert flips `isSpent` on the TXOs already linked to the lock's funding txid once status reaches `InstantSendLocked` (the network has locked the inputs). A terminal `Consumed` upsert heals rows an earlier missed flip left stale.
- **`account_balances_blocking`**: computes each account's balance read-only from `account.utxos` with the exact bucket rules of `ManagedCoreFundsAccount::update_balance`, instead of serving the cache — the snapshot now derives from the same source selection uses and cannot disagree with it.
- **Kotlin UTXO restore**: the stale-flag guard also consults the tracked lock's status via the spending txid's outpoint key (credit outpoints are always vout 0); from `InstantSendLocked` on, the output is skipped **and its `isSpent` healed in place**, so already-poisoned wallets converge on their next launch — terminal `Consumed` rows never re-upsert, so waiting for a status write would never heal them.

Deeper engine follow-up (deliberately out of scope): promote a self-authored transaction's record context **by txid** when its block is processed, so records don't depend on script matching — that would make these reconciles redundant rather than load-bearing.

## How Has This Been Tested?

- Rust: a regression test that dirties the UTXO set while deliberately leaving the cache stale — the accessor must report the live truth (`cargo test -p platform-wallet --lib`: 572 passed).
- Kotlin/Robolectric: `Broadcast` must **not** flip (a pre-broadcast abort can still release the inputs), `InstantSendLocked` flips, and a terminal `Consumed` heals a stale row (`:sdk:testReleaseUnitTest` green).
- The failure itself was reproduced and diagnosed live on an Android testnet wallet (Room mirror inspected directly; the persisted holding figure decoded at exactly the phantom sum).
- ⚠️ The Swift mirror change is compile-unverified here (no iOS toolchain run) — needs CI / an iOS build.

## Breaking Changes

None. All three changes tighten existing readers; no API or schema changes.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Wallet balances now reflect the current UTXO state instead of potentially stale cached values.
  - Asset-lock funding outputs are correctly marked as spent once transactions reach InstantSend finality.
  - Wallet restoration no longer reintroduces outputs already consumed by finalized asset-lock transactions.
  - Improved handling keeps locked, immature, confirmed, and unconfirmed funds accurately classified.
  - Persistence and restoration now fail safely when required spend or finality information cannot be read.


thepastaclaw:
⛔ Blockers found — Opus deferred (commit 4deaf3bc)

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -280,6 +280,33 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
...
+                if entry.statusRaw >= 2,
+                   let displayTxidHex = entry.outPointHex.split(separator: ":").first,
+                   let displayTxid = Data(hexString: String(displayTxidHex)) {
+                    let wireTxid = Data(displayTxid.reversed())
+                    let staleDescriptor = FetchDescriptor<PersistentTxo>(
+                        predicate: #Predicate {
+                            $0.spendingTransaction?.txid == wireTxid && $0.isSpent == false
+                        }
+                    )
+                    if let stale = try? backgroundContext.fetch(staleDescriptor) {
```
thepastaclaw, line 303: **🔴 Blocking: Previously consumed Swift asset locks never trigger this reconciliation**

The new reconciliation only runs when `persistAssetLocks` receives another upsert. A wallet upgraded with an existing lock at `InstantSendLocked`, `ChainLocked`, or especially terminal `Consumed` status can already have linked TXOs persisted with `isSpent == false`. The Swift load path at lines 4482–4484 still fetches every such row solely by `isSpent == false` and marshals it back to Rust without consulting the persisted asset-lock status. Consumed rows are intentionally retained for history and never advance again, so no future callback is guaranteed to repair them; the phantom UTXO can therefore return on every launch. Add a load-time finalized-lock exclusion and healing step equivalent to Kotlin's `buildUtxoRestoreData` guard.

<sub>source: ['codex']</sub>

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandler.kt
```diff
@@ -2267,6 +2284,23 @@ class PlatformWalletPersistenceHandler(
...
+                // finality signal that provably arrives; from
+                // InstantSendLocked on this output is gone. Skip it, and
+                // heal the flag so isSpent-based readers stop counting it.
+                val lockKey = encodeOutPointHex(spendingTxid + ByteArray(4))
+                val lock = database.assetLockDao().getByOutPointHex(lockKey)
+                if (lock != null && lock.statusRaw >= ASSET_LOCK_STATUS_INSTANT_SEND_LOCKED) {
+                    if (!txo.isSpent) {
+                        database.txoDao().upsert(txo.copy(isSpent = true, lastUpdated = now()))
+                    }
+                    continue
```
thepastaclaw, line 2302: **🟡 Suggestion: Guard the healing write against exceptions inside the load path**

This healing upsert runs inside `onLoadWalletList`, whose entire body is wrapped by `guardedLoad(emptyArray())`. If the single-row write throws, the exception escapes the wallet loop and `guardedLoad` returns an empty array for every wallet, even though excluding this finalized TXO from the current restore does not depend on the repair being durable. Treat the write as opportunistic: log its failure and continue skipping the consumed row so one failed repair cannot discard the complete restore result.

```suggestion
if (lock != null && lock.statusRaw >= ASSET_LOCK_STATUS_INSTANT_SEND_LOCKED) {
    if (!txo.isSpent) {
        try {
            database.txoDao().upsert(txo.copy(isSpent = true, lastUpdated = now()))
        } catch (t: Throwable) {
            Log.w(TAG, "load: failed to heal stale asset-lock-consumed TXO", t)
        }
    }
    continue
}
```

packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandlerTest.kt
```diff
@@ -2762,4 +2764,79 @@ class PlatformWalletPersistenceHandlerTest {
...
+                spendingInputIndex = 0,
+                isSpent = false,
+            ),
+        )
+
+        handler.onPersistAssetLockUpsert(
+            walletId, lockTxid + ByteArray(4), ByteArray(4), 0, 1, 0, 9_999_545, 4, null,
+        )
+        assertTrue(db.txoDao().getByOutpoint(outpoint)!!.isSpent)
+    }
```
thepastaclaw, line 2841: **🟡 Suggestion: Add coverage for the finalized asset-lock restore guard**

The added tests exercise only callback-time flips for statuses 2 and 4. They never invoke `onLoadWalletList()` with the legacy state this new restore branch is meant to repair: an unspent TXO linked to a mempool-context spending transaction and an already-finalized asset-lock row. Add a restore test that seeds this state, verifies the TXO is absent from the returned `utxos`, and verifies its Room row is healed to `isSpent = true`; this also pins the synthetic vout-0 key and txid byte orientation.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/manager/accessors.rs
```diff
@@ -1237,3 +1249,77 @@ mod spv_rescan_tests {
...
+        } else if !utxo.is_mature(last_processed_height) {
+            immature += value;
+        } else if utxo.is_confirmed || utxo.is_instantlocked || utxo.is_trusted {
+            confirmed += value;
+        } else {
+            unconfirmed += value;
+        }
+    }
+    key_wallet::wallet::balance::WalletCoreBalance::new(confirmed, unconfirmed, immature, locked)
+}
```
thepastaclaw, line 1279: **🟡 Suggestion: Balance classification now has two independent implementations**

`computed_core_balance` duplicates the pinned key-wallet `ManagedCoreFundsAccount::update_balance` bucket rules line for line. A future key-wallet change to maturity, locking, trust, or confirmed/unconfirmed classification can compile cleanly while this accessor retains the old behavior, recreating disagreement between balance readers. Add an immutable balance-calculation method in key-wallet and have both its mutating cache update and this accessor call that shared implementation.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/manager/accessors.rs
```diff
@@ -1237,3 +1249,77 @@ mod spv_rescan_tests {
...
+        assert_eq!(
+            account.balance.total(),
+            cached_before.total(),
+            "precondition: the cache must still hold the stale figure"
+        );
+        assert_eq!(
+            computed_core_balance(account, height).total(),
+            0,
+            "the accessor's fold must see the live (empty) UTXO set"
+        );
```
thepastaclaw, line 1323: **🟡 Suggestion: Regression test passes if the calculator always returns zero**

The test checks `computed_core_balance` only after clearing every UTXO and expecting zero. An implementation that always returned `WalletCoreBalance::default()` would therefore pass. Before clearing the account, compare the computed balance with the freshly updated, non-empty cache so the test establishes that the live fold classifies funded UTXOs as well as observing their removal.

```suggestion
account.update_balance(height);
let cached_before = account.balance;
assert!(cached_before.total() > 0, "fixture must be funded");
assert_eq!(
    computed_core_balance(account, height),
    cached_before,
    "the live fold must reproduce a freshly updated non-empty balance"
);
account.utxos.clear();

assert_eq!(
    account.balance.total(),
    cached_before.total(),
    "precondition: the cache must still hold the stale figure"
);
assert_eq!(
    computed_core_balance(account, height).total(),
    0,
    "the accessor's fold must see the live (empty) UTXO set"
);
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The Rust live-balance calculation and callback-time asset-lock reconciliation are sound, but the Swift restore path still rehydrates stale funding TXOs for locks finalized before this fix, leaving upgraded iOS wallets with persistent phantom balances. The remaining suggestions cover failure isolation and regression coverage for the Kotlin restore repair, shared ownership of the Rust balance rules, and a stronger non-empty-balance assertion.
Source: reviewers gpt-5.6-sol (general) and gpt-5.6-sol (rust-quality); final verifier gpt-5.6-sol. Orchestration-only: openclaw-agent/cliproxy/gpt-5.6-sol (not reviewer evidence).

### Review provenance

<details>

```

- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift:294-303: Previously consumed Swift asset locks never trigger this reconciliation
  The new reconciliation only runs when `persistAssetLocks` receives another upsert. A wallet upgraded with an existing lock at `InstantSendLocked`, `ChainLocked`, or especially terminal `Consumed` status can already have linked TXOs persisted with `isSpent == false`. The Swift load path at lines 4482–4484 still fetches every such row solely by `isSpent == false` and marshals it back to Rust without consulting the persisted asset-lock status. Consumed rows are intentionally retained for history and never advance again, so no future callback is guaranteed to repair them; the phantom UTXO can therefore return on every launch. Add a load-time finalized-lock exclusion and healing step equivalent to Kotlin's `buildUtxoRestoreData` guard.

- [SUGGESTION] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandler.kt:2298-2302: Guard the healing write against exceptions inside the load path
  This healing upsert runs inside `onLoadWalletList`, whose entire body is wrapped by `guardedLoad(emptyArray())`. If the single-row write throws, the exception escapes the wallet loop and `guardedLoad` returns an empty array for every wallet, even though excluding this finalized TXO from the current restore does not depend on the repair being durable. Treat the write as opportunistic: log its failure and continue skipping the consumed row so one failed repair cannot discard the complete restore result.

In `packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandlerTest.kt`:
- [SUGGESTION] packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandlerTest.kt:2768-2841: Add coverage for the finalized asset-lock restore guard
  The added tests exercise only callback-time flips for statuses 2 and 4. They never invoke `onLoadWalletList()` with the legacy state this new restore branch is meant to repair: an unspent TXO linked to a mempool-context spending transaction and an already-finalized asset-lock row. Add a restore test that seeds this state, verifies the TXO is absent from the returned `utxos`, and verifies its Room row is healed to `isSpent = true`; this also pins the synthetic vout-0 key and txid byte orientation.

- [SUGGESTION] packages/rs-platform-wallet/src/manager/accessors.rs:1258-1279: Balance classification now has two independent implementations
  `computed_core_balance` duplicates the pinned key-wallet `ManagedCoreFundsAccount::update_balance` bucket rules line for line. A future key-wallet change to maturity, locking, trust, or confirmed/unconfirmed classification can compile cleanly while this accessor retains the old behavior, recreating disagreement between balance readers. Add an immutable balance-calculation method in key-wallet and have both its mutating cache update and this accessor call that shared implementation.
- [SUGGESTION] packages/rs-platform-wallet/src/manager/accessors.rs:1309-1323: Regression test passes if the calculator always returns zero
  The test checks `computed_core_balance` only after clearing every UTXO and expecting zero. An implementation that always returned `WalletCoreBalance::default()` would therefore pass. Before clearing the account, compare the computed balance with the freshly updated, non-empty cache so the test establishes that the live fold classifies funded UTXOs as well as observing their removal.
```
</details>

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandler.kt
```diff
@@ -1523,6 +1523,23 @@ class PlatformWalletPersistenceHandler(
...
+            // does arrive (the proof wait drives it): once it reaches
+            // InstantSendLocked (2) the network has locked the inputs, so
+            // flip the linked TXOs here. Monotonic, and keyed strictly to
+            // TXOs already linked to THIS lock's funding txid.
+            if ((status.toInt() and 0xFF) >= ASSET_LOCK_STATUS_INSTANT_SEND_LOCKED) {
+                val fundingTxid = outPoint.copyOfRange(0, 32)
+                for (txo in db.txoDao().getUnspentBySpendingTxid(fundingTxid)) {
+                    db.txoDao().upsert(txo.copy(isSpent = true, lastUpdated = now()))
+                }
+            }
```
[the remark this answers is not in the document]
thepastaclaw: InstantSend finality is a property of the spending transaction, not the wallet that delivered the callback. Every matched TXO is linked through `spendingTxid` to this exact finalized funding transaction, so it is genuinely consumed even when owned by another wallet; wallet-scoping the update could leave another input of the same transaction stale.

QuantumExplorer (changes requested):
Something seems weird here.

QuantumExplorer:
I'll work on this later

HashEngineering:
> Something seems weird here.

@QuantumExplorer 
You are right.  The asset locks generated were draining the wallet using other code that I didn't submit a PR for.

Within the mobile apps, many functions have a MAX button to send the entire wallet balance and I was going that here.

The same issue was found when doing a drain on a Maya Sell Swap (outputs are OP_RETURN and an address outside the wallet only).

In both cases, there was no address in the outputs of the transaction that are owned by the wallet, and for some reason such transactions are not accounted for when determining the balance or marking inputs as spent.

Two PR's were submitted to do the drain on Maya swaps (OP_RETURN)
1. https://github.com/dashpay/rust-dashcore/pull/928
2. https://github.com/dashpay/platform/pull/4324

When testing these two, which allow for draining the wallet balance with a TX with an OP_RETURN output, I found the same problem that this PR was attempting to fix -- that drain TX was not counted in the balance or it wasn't counting the inputs as spent.

In theory, a regular send that drains the entire balance would result in the same problem.  The balance would remain the same as it was before the TX was sent.

HashEngineering:
## The rust-dashcore fix behind this, and whether this PR is still needed after it

### What actually needs to change in rust-dashcore

The engine follow-up this PR deferred turns out to be simpler — and stronger — than "promote a self-authored tx's record context by txid when its block is processed." Context promotion only helps if the block gets downloaded at all; a spender in an otherwise-irrelevant block never triggers block processing to promote from. The real gap is one step earlier, in compact-filter **matching**:

- Dash Core's BIP158 basic filter carries each input's **prevout scriptPubKey** from undo data (`BasicFilterElements` in `blockfilter.cpp`), alongside the block's output scripts. So a block that spends our UTXOs *is* discoverable — if the query includes the spent outputs' scripts.
- `dash-spv`'s block/filter matching (`sync/filters/manager.rs`, `scan_batch`) queries only `monitored_script_pubkeys_for(wallet_id)` — the monitored *address* set. A transaction with no wallet-owned standard output (an asset lock burning into its payload, or a MAX Maya drain paying destination + OP_RETURN with no change) contributes nothing to that query from its outputs, and if the spent UTXO's script isn't in the monitored set, the block never matches. The mempool path doesn't have this hole — its bloom filter already watches `wallet.watched_outpoints()` — which is exactly why these rows end up with `spendingTxid` set (mempool saw it) but `isSpent = 0` forever (in-block context never arrives).

The fix: have filter matching watch the wallet's **UTXO scripts** alongside the monitored scripts, mirroring the bloom path's outpoint watching:

1. `WalletInterface::watched_utxo_script_pubkeys_for(wallet_id)` — the scriptPubKeys of the wallet's UTXOs, sourced from the UTXO records themselves (not the address pools, so it holds even when an address has fallen out of the monitored set).
2. `WalletManager` implements it from `info.utxos()` → `utxo.txout.script_pubkey`.
3. `scan_batch` merges those scripts (deduped) into each behind-wallet's filter query, feeding both the union match and per-wallet attribution.

With that, any block spending a known UTXO matches regardless of the transaction's outputs — asset locks and Maya drains alike get their in-block context, and the ordinary `isSpent` flip runs. A useful side effect: **rescans become effective for already-poisoned wallets**, because the stuck UTXOs are still tracked, so their scripts are watched and the old block matches on re-scan.

This is implemented with regression coverage (a drain-shaped block matched via UTXO scripts and missed without them; a confirmed drain settling the balance to zero), pending a rust-dashcore PR.

### Is this PR still needed after the engine fix? Yes.

- **`account_balances_blocking` computing live from `account.utxos`**: needed regardless — the stale-cache accessor bug is independent of block matching.
- **The status-keyed heals and the restore-time heal**: no longer *load-bearing* for transactions processed after both fixes ship — the engine path delivers the in-block flip on its own — but they stay necessary for:
  - **already-poisoned rows**: the engine fix doesn't rewrite history, and users can't be counted on to rescan; the `Consumed`/restore heals converge affected wallets on their next launch;
  - **the deployment window**: the engine fix rides a rust-dashcore release → platform dependency bump → app release, while this PR heals data written by every engine already in the field;
  - **earlier finality**: the InstantSendLocked flip fires seconds after broadcast, versus the engine path's flip at confirmation — and it covers any other future path that leaves a spender stuck pre-in-block.

In short: the engine fix makes these reconciles what this PR's description hoped they'd become — redundant rather than load-bearing — for *new* transactions. They remain the convergence path for existing data. Merge as-is; engine fix follows as the root-cause companion.

HashEngineering:
CoinJoin mixing fee transactions that only have an OP_RETURN output may also not be counted in the balance calculation.

shumkov:
@QuantumExplorer — fair reaction. Here is the part that is established, and the part
that isn't, kept separate.

**What the client is actually seeing.** An asset-lock transaction is a Core special tx
(type 8) whose value is burned into the `AssetLockPayload` credit outputs, not into a
wallet-owned standard output; on a MAX/drain funding there is no change output either,
so the tx has zero outputs paying a script this wallet owns. In the field the mirrored
row ends up in a very specific state: `spendingTxid` **set** — the mempool sighting
landed — and `isSpent = 0` **permanently**, i.e. the in-block sighting never did.

Every downstream "spent" signal is keyed on in-block context, so all of them stay
silent:
- Kotlin `PlatformWalletPersistenceHandler.kt:861` — `isSpent = txo.isSpent || context >= CONTEXT_IN_BLOCK`
- Swift `PlatformWalletPersistenceHandler.swift:1487` → `spendIsInBlock` (`:1201`)
- the UTXO-restore guard, Kotlin `…kt:2476` — `spending.context >= CONTEXT_IN_BLOCK`

The consumed output therefore stays "unspent" in the mirror and the restore path hands
it back to Rust as spendable on **every** relaunch — that is the 0.21-on-a-0.1-wallet
figure, re-inflating at each launch. That chain is verifiable in this repo and is what
the persistence half of this PR addresses.

**Why the reconciles key on the lock's own status.** The one signal that provably
arrives is the tracked lock's status: the proof wait drives Built → Broadcast →
InstantSendLocked → ChainLocked → Consumed through `advance_asset_lock_status`
(`rs-platform-wallet/src/wallet/asset_lock/sync/tracking.rs:274`), persisting at each
step. From `InstantSendLocked` (2) on the network has locked the inputs, so the linked
TXOs are gone — hence the flip at `…kt:1683` / `…swift:339` and the load-time
exclusion + heal at `…kt:2497` / `…swift:4904`. `Consumed` is terminal and never
re-upserts, which is exactly why the load-time pass has to exist: it is the only thing
that converges rows an older build already poisoned.

**The Rust change is a separate, ordinary bug — no SPV involved.**
`account_balances_blocking` served the cached `ManagedCoreFundsAccount::balance` field,
which only refreshes when transaction processing calls `update_balance()`. Coin
selection reads `account.utxos` directly and was always right, so the reader and the
selector could disagree indefinitely with or without any of the above. It now folds
`account.utxos` with identical bucket rules (`accessors.rs:413`, `:432`, `:1297`).

**Where I'd push back on the stated root cause.** The comment above says block matching
misses these because the filter query watches only *address* scripts. I checked that
against the pinned engine (`4db5c367`) and it is only half right:

- The scan query really is address-derived — `scan_batch`
  (`dash-spv/src/sync/filters/manager.rs:857`) feeds
  `scan_script_pubkeys_for(wallet_id)` plus `monitored_filter_elements_for(wallet_id)`,
  the latter being just provider owner/voting `hash160`s
  (`key-wallet-manager/src/wallet_interface.rs:102,114`). No UTXO scripts, no outpoints.
  The outpoint watching is bloom-only — `watched_outpoints()` is documented as "used for
  bloom filter construction to detect spends of our UTXOs" (`…:119-120`). That asymmetry is
  real, and it does explain mempool-yes / block-no.
- **But** for a Standard/BIP44 account the query is `all_script_pubkeys()` over every
  address in every pool, used ones included
  (`key-wallet/src/wallet/managed_wallet_info/wallet_info_interface.rs:395`). The
  address that funded the lock is in that set. Since a BIP158 basic filter also carries
  each input's prevout `scriptPubKey`, the block *should* match on the spent output's
  own script — so "no wallet-owned output" alone should not be enough to miss it.

So one of these has to be true, and I could not settle it from the platform side:
(i) Dash Core's basic filter does not actually include prevout scripts, or (ii) the
block does match and is processed, but the self-authored transaction's *record* is
never promoted out of mempool context — which was the original hypothesis in this PR's
description. You'd know which faster than I would.

One case I *can* confirm from the pinned code: for a **CoinJoin** account the scan
query is pruned to `unspent_or_unused_script_pubkeys()`
(`key-wallet/src/managed_account/managed_core_funds_account.rs:177-187`,
`!info.is_used() || funded.contains(script)`, added by dashpay/rust-dashcore#948). The
moment the lock spends the last UTXO on a used CoinJoin address, that address is
`is_used()` and no longer funded — so it drops out of the forward-scan query and the
block carrying its spender genuinely stops matching. That path fits the reported
symptom exactly.

**On scope.** The Rust accessor fix stands on its own either way. The status-keyed
heals stop being load-bearing for *new* transactions once whatever the engine gap turns
out to be is closed, but they stay necessary for (i) rows already written by every
engine in the field — an engine fix does not rewrite history and users can't be relied
on to rescan, (ii) the deployment window (rust-dashcore release → dep bump → app
release), and (iii) earlier finality, since the InstantSendLocked flip lands seconds
after broadcast rather than at confirmation. And the shape is not asset-lock-specific:
the same "no wallet-owned output" transaction appears in a Maya sell-swap drain
(destination + `OP_RETURN`), in a CoinJoin mixing-fee tx with only an `OP_RETURN`, and
in principle in any ordinary send that drains the whole balance.

Happy to split the client heals into their own PR if you'd rather nail the engine
question first — but as written they are convergence logic for already-written data,
not a substitute for the engine fix.

shumkov:
Follow-up to the analysis above — we ran both experiments and the question is settled.

**(i) is false: the filters are fine.** Dash Core's BASIC filter inserts the scriptPubKey of every spent coin (`src/blockfilter.cpp`, `BasicFilterElements`, verified at master `728f5055`), and a live testnet BASIC filter matched a real type-8 AssetLock input's prevout script that appears in none of the block's outputs. Blocks containing these spends ARE fetched.

**(ii) is true, with a subtler mechanism than either of us guessed: the transaction is starved of relevance by its own mempool sighting.** Reproduced against the exact platform-pinned engine (`4db5c367`) with a real DIP-2 type-8 `AssetLockPayload` driven through the production path:
1. The mempool delivery creates the record **and removes the live UTXO it spends**.
2. When the block arrives (filter matched — the funding script stays in the scan query), `WalletTransactionChecker::check_transaction` recomputes fresh relevance from **output scripts + current live UTXOs** (`wallet_checker.rs:54-61` → `account_checker.rs:645-672`). Outputs pay nobody we own; the UTXO is already gone — deleted by step 1.
3. Fresh relevance = false → early return **before any lookup of the existing txid record**. The record stays `Mempool` forever; no `BlockProcessed.updated` is emitted. Hence exactly mempool-yes / block-no, every time: relevance is true precisely once, and the first sighting consumes it.

**Minimal engine fix** (scoped in the diagnostic report): when fresh relevance is false and the context is `InBlock`/`InChainLockedBlock`, fall back to a txid lookup of already-stored records across routed accounts and promote only their confirmation context (existing monotonic rules), marking the wallet affected and emitting the updated record. Narrower and safer than trusting filter attribution — an existing record is direct wallet attribution; a filter match is not.

We have a ready-made failing promotion test (asserts the expected-success promotion and fails on the current engine) that becomes the red regression for that fix — happy to hand it over or open the rust-dashcore PR, whichever @HashEngineering prefers for the promised companion. The CoinJoin scan-query pruning (#948) remains a real second path for CoinJoin-funded locks and is unaffected by this fix.

This PR's heals remain necessary under this outcome for the reasons above: already-poisoned rows, the deployment window, and earlier finality.

shumkov:
Engine companion is up: dashpay/rust-dashcore#998 — implements the confirmed-context fallback for the root cause settled above (the relevance gate dropping a wallet's own spend-only tx after the mempool sighting consumed its UTXO), with the diagnostic red→green regression. It reaches platform at the next rust-dashcore pin bump; this PR's heals remain necessary independently (already-poisoned rows, deployment window, earlier finality).

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandler.kt
```diff
@@ -2267,6 +2284,23 @@ class PlatformWalletPersistenceHandler(
...
+                // InstantSendLocked on this output is gone. Skip it, and
+                // heal the flag so isSpent-based readers stop counting it.
+                val lockKey = encodeOutPointHex(spendingTxid + ByteArray(4))
+                val lock = database.assetLockDao().getByOutPointHex(lockKey)
+                if (lock != null && lock.statusRaw >= ASSET_LOCK_STATUS_INSTANT_SEND_LOCKED) {
+                    if (!txo.isSpent) {
+                        database.txoDao().upsert(txo.copy(isSpent = true, lastUpdated = now()))
+                    }
+                    continue
+                }
```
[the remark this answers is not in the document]
shumkov: Addressed in 255d231174/e959e949eb: both Kotlin heals are now column-scoped conditional UPDATEs inside try/catch — a failed heal can no longer clobber healthier rows nor abort the load; a throwing per-candidate lookup drops only its own candidate (regression with unrelated-wallet sentinel).

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
shumkov: Addressed in 255d231174: the Swift load-path heal the thread asked for — previously consumed locks are reconciled at load; Swift compiled against the xcframework, and thrown-fetch regressions (34aa2a0b7c) pin the fail-closed paths.

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandler.kt
shumkov: Addressed in 255d231174/e959e949eb: both Kotlin heals are now column-scoped conditional UPDATEs inside try/catch — a failed heal can no longer clobber healthier rows nor abort the load; a throwing per-candidate lookup drops only its own candidate (regression with unrelated-wallet sentinel).

packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandlerTest.kt
shumkov: Addressed in 255d231174/e959e949eb: the finalized-lock restore guard is vout-independent (MAX(statusRaw) by txid prefix, DIP-0027 multi-output rows match) and covered by a Kotlin restore test proven red plus a throwing-lookup regression.

packages/rs-platform-wallet/src/manager/accessors.rs
shumkov: Addressed in e959e949eb: account_balances_blocking now folds account.utxos with the identical bucket rules coin selection uses — one classification semantics; the regression drives the production accessor and fails against the cached read.

packages/rs-platform-wallet/src/manager/accessors.rs
shumkov: Addressed in 255d231174/e959e949eb: the regression is now proven red against BOTH the pre-fix cached read and a WalletCoreBalance::default() zero stub, and drives account_balances_blocking through a real manager — an always-zero calculator fails it.

PR #4338, opened 2026-08-08
feat(platform-wallet): derive owner/voting provider keys Rust-side

QuantumExplorer:
## Why

The secp256k1 provider families (masternode **owner** and **voting** keys) were the only key material the iOS wallet derived in Swift, and the only ones sourced from a throwaway key-wallet rebuilt from the mnemonic. Everything else already derives Rust-side from the running wallet:

| what | route |
|---|---|
| payments (BIP44/BIP32/CoinJoin) | `CoreTransactionBuilder` |
| identity / DPNS | platform-wallet identity signer |
| operator (BLS), platform node (Ed25519) | `derive_provider_key_at_index` |
| **owner, voting (secp256k1)** | **Swift `Account.derivePrivateKeyWIF`** ← the outlier |

Owner and voting were on the odd path only because `ProviderKeyKind` never had variants for them. That outlier turned out to be broken: the Swift wrapper pre-derived the account root and passed it as the "master", applying the DIP-3 path twice, so every owner/voting key came from `m/9'/5'/3'/1'/9'/5'/3'/1'/index`. Nothing failed locally — the keys were well-formed — and it surfaced only as Platform rejecting masternode votes as having no voter identity.

## What

- `ProviderKeyKind::{Owner, Voting}` (account tags 9 / 8) with a secp256k1 branch: public key and P2PKH address off the account xpub (non-hardened, no seed — as with BLS operator), private key derived from the raw seed.
- The branch runs **the same seed-vs-xpub cross-check the operator family does**. That guard is the point: the two sides derive independently, and a mismatched pair has no local symptom — it signs for an address nobody expects. On the voting family that shows up only as Platform rejecting the vote.
- `ProviderDerivedKey` / the FFI struct gain `address` and `private_key_wif`, non-null only for the ECDSA families. WIF is encoded Rust-side so the network byte and compression flag have one home; it is zeroized on free like the hex scalar.

### The watch-only gate

The first cut used `Account::derive_from_seed_private_key_at`, which gates on `is_watch_only` and so refused exactly the external-signable wallets the seed argument exists to serve — on device: *"Watch-only wallet: private keys not available"*. This module's own docs already warn about that asymmetry, and are why the BLS/Ed25519 families use key-wallet's gate-free entry points. secp256k1 has none, so the derivation is inline in the same shape as `BLSAccount::operator_private_key_at`: raw seed → master xpriv → the account type's DIP-3 path → non-hardened child.

## Testing

`platform-wallet` provider-key tests: **9/9**.

- owner and voting pinned against explicit `m/9'/{5'|1'}/3'/{2'|1'}/index` at indexes 0, 1, 19, on mainnet and testnet
- a guard that the doubled path is *not* what we derive
- **a watch-only test**, which is the one that mattered: the seed-bearing tests passed the entire time the app was failing, because they never reach the gate. It pins both halves — the gated wrapper is expected to refuse a watch-only account, and the gate-free derivation must still match the DIP-3 path.

## Notes

- `#[repr(C)]` change — consumers need a rebuilt xcframework.
- Supersedes #4334 for our callers (it removes the app's only use of `derivePrivateKeyWIF`), but #4334 is still worth landing: that API stays public and wrong-by-default for the next caller.

## Summary by CodeRabbit

* **New Features**
  * Added support for deriving voting and owner ECDSA provider keys.
  * Exposed optional legacy addresses and private keys in WIF format for applicable derived keys.
  * Added corresponding provider key types to the Swift SDK.

* **Bug Fixes**
  * Improved key derivation validation and handling for watch-only wallets.
  * Prevented incorrect account-path duplication during derivation.

### DIFF
--- a/packages/rs-platform-wallet-ffi/src/provider_key_at_index.rs
+++ b/packages/rs-platform-wallet-ffi/src/provider_key_at_index.rs
@@ -163,8 +163,10 @@ pub unsafe extern "C" fn platform_wallet_provider_key_at_index(
             return PlatformWalletFFIResult::err(
                 PlatformWalletFFIResultCode::ErrorInvalidParameter,
                 format!(
-                    "unknown provider key kind {other} (expected {PROVIDER_KEY_KIND_OPERATOR} \
-                     operator or {PROVIDER_KEY_KIND_PLATFORM_NODE} platform node)"
+                    "unknown provider key kind {other} (expected \
+                     {PROVIDER_KEY_KIND_VOTING} voting, {PROVIDER_KEY_KIND_OWNER} owner, \
+                     {PROVIDER_KEY_KIND_OPERATOR} operator or \
+                     {PROVIDER_KEY_KIND_PLATFORM_NODE} platform node)"
                 ),
             );
         }
--- a/packages/rs-platform-wallet-ffi/src/provider_key_at_index.rs
+++ b/packages/rs-platform-wallet-ffi/src/provider_key_at_index.rs
@@ -50,6 +50,10 @@ use rs_sdk_ffi::MnemonicResolverHandle;
 pub const PROVIDER_KEY_KIND_OPERATOR: u8 = 10;
 /// See [`PROVIDER_KEY_KIND_OPERATOR`].
 pub const PROVIDER_KEY_KIND_PLATFORM_NODE: u8 = 11;
+/// secp256k1 masternode voting keys (`ProviderVotingKeys`, tag 8).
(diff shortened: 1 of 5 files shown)


thepastaclaw:
🕓 Ready for review — 1 ahead in queue (commit db24fe30)
_Queue position: 2/3_
_ETA: start ~08:40 UTC · complete ~08:54 UTC (median 14m across 30 recent reviews; 2 slots)_
_Queued 24m ago · Last checked: 2026-08-08 08:40 UTC_

PR #4340, opened 2026-08-08
feat(rs-sdk): own the masternode voting-key facts

QuantumExplorer:
Follow-up to #4333, which fixed a masternode voting bug — and in doing so put a pile of Platform domain knowledge in the FFI shim.

## The problem

Casting a contested-resource vote needs three things that are properties of **Platform**, not of any binding:

1. which key a voter identity holds,
2. how to recognise that Platform rejected that key, and
3. what the rejection actually meant.

All three live in `rs-sdk-ffi`. So Swift and Kotlin callers get them — they route through the FFI — while a Rust caller of `PutVote` does not, and has to rediscover them.

That is not hypothetical. `get_voting_identity_id` derives the voter identity from `voter_pro_tx_hash.as_bytes()` verbatim, and until #4333 nothing said which orientation that had to be. A caller holding wire/`Txid` bytes addresses an identity that has never existed and gets "no voter identity" back. That cost real debugging time, and the FFI's `.reverse()` only protects callers who go through the FFI.

## What this adds

`platform::transition::masternode_vote_keys`:

| | |
|---|---|
| `voter_identity_id` | takes a typed **`ProTxHash`**, not bytes |
| `voter_identity_voting_key` | builds the key Platform holds, so no fetch is needed to broadcast |
| `select_voting_key` | picks the usable key off a fetched identity **by its own data** |
| `is_voting_key_failure` | gates diagnosis on the typed consensus error |
| `diagnose_voting_key_failure` | returns a typed `VotingKeyProblem` |

Two choices worth calling out:

**The type is the guard, not the doc.** `voter_identity_id` takes `ProTxHash` because the derivation is orientation-sensitive and `Txid` bytes for the same transaction are its exact reverse. A doc comment already proved insufficient — this bug happened once with the contract undocumented, and documenting it (#4333) helps only readers. The type helps everyone.

**`diagnose_voting_key_failure` returns a fact, not a sentence.** `VotingKeyProblem` is an enum; bindings phrase it for their users. Formatting messages in the SDK would make every caller inherit English strings they cannot localise.

**Signing stays with the caller.** Producing a `Signer` from a raw key is a binding concern, and pulling `simple-signer` into the public SDK to do it here would be the wrong trade — so this covers the *key facts*, not a full `cast_vote`.

## Testing

10 tests, all in-process:

- the fabricated key matches what `create_voter_identity_v0` assigns (id 0, VOTING, ECDSA_HASH160, enabled)
- **the identity id is orientation-sensitive** — reversing the `ProTxHash` changes it, which is the property that made votes fail
- selection is by data not position (a voting key at id 4 behind another key at id 0 is still found), and rejects a disabled key, a different address, a wrong purpose, and a wrong key type — each isolated so no predicate is dead weight
- the three key failures are diagnosable; a signature failure that is *not* about the key's existence or state, and a broadcast error with no consensus cause, are not

## Scope

No behaviour change on its own — this is the API. Moving `rs-sdk-ffi` onto it is a separate PR so this can be reviewed as the surface it is, and so #4333 (already several rounds in) is not reopened.

## Summary by CodeRabbit

* **New Features**
  * Added public utilities for resolving masternode voting identities and keys.
  * Added support for selecting enabled, matching voting keys.
  * Added diagnostics for missing identities, unavailable keys, and relevant voting failures.
  * Added comprehensive validation for voting-key construction, selection, and failure detection.

### DIFF
--- a/packages/rs-sdk/src/platform/transition.rs
+++ b/packages/rs-sdk/src/platform/transition.rs
@@ -10,6 +10,7 @@ pub(crate) mod broadcast_identity;
 pub mod broadcast_request;
 #[cfg(feature = "shielded")]
 pub mod identity_create_from_shielded_pool;
+pub mod masternode_vote_keys;
 pub mod purchase_document;
 pub mod put_contract;
 pub mod put_document;
--- a/packages/rs-sdk/src/platform/transition/masternode_vote_keys.rs
+++ b/packages/rs-sdk/src/platform/transition/masternode_vote_keys.rs
@@ -1,0 +1,397 @@
+//! Voting-key facts for masternode contested-resource votes.
+//!
+//! Casting a vote needs three things that are properties of Platform, not of
+//! any particular binding: which key a voter identity holds, how to recognise
+//! that Platform rejected that key, and what the rejection actually meant.
+//! They lived in `rs-sdk-ffi`, which meant Swift and Kotlin callers got them
+//! (they route through the FFI) while Rust callers of [`PutVote`] did not, and
+//! had to rediscover them — including the byte-order rule that made a vote
+//! address an identity that never existed.
+//!
+//! Signing stays with the caller: producing a [`Signer`] from a raw key is a
+//! binding concern, and pulling `simple-signer` into the public SDK to do it
+//! here would be the wrong trade.
+//!
+//! [`PutVote`]: super::vote::PutVote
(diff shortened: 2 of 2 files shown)


thepastaclaw:
🕓 Ready for review — 1 ahead in queue (commit e1465edd)
_Queue position: 2/2_

PR #4343, opened 2026-08-08
chore: update rust-dashcore to b056d07c

QuantumExplorer:

The `rust-dashcore` pin was two commits behind `dev`, missing two fixes — one of
which the platform tree already works around and refers to by issue number.

`rs-platform-wallet` carries a defensive duplicate-prevout check that cites
`rust-dashcore#931` as the upstream fix
([transaction.rs](https://github.com/dashpay/platform/blob/v4.2-dev/packages/rs-platform-wallet/src/wallet/core/transaction.rs#L445-L450)),
but the pinned revision predates that fix, so the comment described behavior the
build did not actually have. This bump makes it true.

## What was done?

Moved all eight `rust-dashcore` git dependencies from `944e53a5` to `b056d07c`
(current `dev` tip) and regenerated the lockfile.

The two commits picked up, neither of which changes a public signature:

- **`fix(key-wallet): never offer an outpoint the builder already holds`**
  ([#931](https://github.com/dashpay/rust-dashcore/pull/931)) — `add_funding`
  appends every unreserved UTXO of a funding account to the candidate set. If
  the builder was already seeded with one of them (via `add_inputs`, or an
  earlier `add_funding` of an overlapping account), that outpoint became two
  candidates. Coin selection does not deduplicate by outpoint, so
  `SelectionStrategy::All` was guaranteed to put both copies in the
  transaction, and the ordinary strategies could pick both and double-count
  them toward the target. Either way the transaction spends one prevout twice
  and Core rejects it.

- **`fix(dash-spv): stop the filter rescan loop when the chain anchors above the wallet`**
  ([#933](https://github.com/dashpay/rust-dashcore/pull/933)) — a wallet whose
  `synced_height` sits below the chain anchor could never be certified and the
  rescan never terminated. On mainnet `hd_wallet_sync_floor` pushes the anchor
  to the checkpoint at 200000 while the CLI creates wallets at birth height 0,
  so `synced_height` stayed at 0 for the whole run — thousands of restarts and
  batch scans producing zero sync-height advances at ~350% CPU, indefinitely.

No platform source changes were required.

## How Has This Been Tested?

- `cargo update` moved only the eight `rust-dashcore` crates and their internal
  members — no incidental dependency churn in the lockfile.
- `cargo check --workspace --all-targets` — passes (exit 0).
- `cargo clippy --workspace --all-targets` — passes (exit 0).

The upstream fixes carry their own regression tests in `rust-dashcore` (#933
adds four cases that fail before the fix); this PR is the pin bump only.

## Breaking Changes

None. Both commits are behavior fixes internal to `dash-spv` and `key-wallet`;
no public signature changed, and nothing in the platform tree needed adjusting.

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1229,7 +1229,7 @@ version = "3.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "faf9468729b8cbcea668e36183cb69d317348c2e08e994829fb56ebfdfbaac34"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -1662,7 +1662,7 @@ dependencies = [
 [[package]]
 name = "dash-network"
 version = "0.45.0"
-source = "git+https://github.com/dashpay/rust-dashcore?rev=944e53a503a88daa0d7706d7d7c5f5350213d295#944e53a503a88daa0d7706d7d7c5f5350213d295"
+source = "git+https://github.com/dashpay/rust-dashcore?rev=b056d07c61f8618f05082552bbb88072290d57c1#b056d07c61f8618f05082552bbb88072290d57c1"
 dependencies = [
  "bincode",
  "bincode_derive",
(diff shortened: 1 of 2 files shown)


thepastaclaw:
🔍 Review in progress — actively reviewing now (commit 4f59528b)
_Stage: Opus review + final verification_
_ETA: complete ~17:45 UTC (median 14m across 30 recent reviews)_
_Running 19m · Last checked: 2026-08-08 17:40 UTC_

PR #4344, opened 2026-08-08
feat(sdk): add transport feature to dapi-grpc for types-only consumers

PastaPastaPasta:
## Issue being fixed or feature implemented

`dapi-grpc` unconditionally built tonic with its native channel and TLS transport stack, so consumers that only need generated message types or proof verification also pulled hyper, rustls, ring, and tower into their dependency trees. `drive-proof-verifier` is a concrete synchronous consumer that does not open network connections.

This is the transport-feature prerequisite extracted from #4335. It lets embedders such as Dash Core use Platform proof verification without inheriting a networking implementation they do not use.

## What was done?

- Added an opt-in `transport` feature to `dapi-grpc` for tonic channel, transport, and TLS support.
- Kept `dapi-grpc` defaults transport-free because Cargo features are additive and cannot be target-scoped.
- Made `build.rs` emit tonic transport helpers only when `transport` is enabled and never for `wasm32`.
- Kept native networking available through `rs-dapi-client`'s non-wasm target dependency.
- Made the `server` feature imply `transport`, preserving the DAPI server build.
- Added CI guards for:
  - hyper, rustls, and tower in `drive-proof-verifier`'s native dependency tree;
  - hyper, rustls, tower, and mio in the default `dash-sdk` and `wasm-sdk` wasm32 dependency trees.
- Added `dapi-grpc` to the nightly per-feature package matrix.

Types-only consumers can use:

```toml
dapi-grpc = { default-features = false, features = ["platform", "client"] }
```

## How Has This Been Tested?

Validated after rebasing onto current `v4.2-dev`:

- `cargo fmt --check --all`
- native, wasm32, and explicit types-only `dapi-grpc` checks
- `drive-proof-verifier`, `rs-dapi-client`, and default `dash-sdk` checks
- native and wasm32 dependency leak guards
- feature-tree assertions proving native `dash-sdk` includes transport while wasm32 `dash-sdk` does not

The corrected full #4335 series is rebased and validated separately on top of this prerequisite.

## Breaking Changes

SDK users and in-repository consumers retain native transport through the appropriate native client or server dependency.

External native consumers that depend directly on `dapi-grpc` and call generated `connect()` methods must explicitly enable `features = ["transport"]`. wasm32 consumers can now use the default feature set without pulling an incompatible native transport graph.

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added or updated relevant unit/integration/functional/e2e tests
[...]

PR #4335, opened 2026-08-07
feat(sdk): transport-free verification and query core for embedders

PastaPastaPasta:
## Issue being fixed or feature implemented

Embedders that bring their own transport and trust context cannot currently consume Platform's query and proof-verification layer without depending on the networked SDK or reconstructing Drive query shapes themselves.

Dash Core is the concrete consumer: it already owns endpoint selection, transport, locally synchronized LLMQ keys, and wallet key custody. It needs Platform's canonical request construction, response decoding, and proof verification without adopting `rs-dapi-client` or tonic's native channel/TLS stack.

## What was done?

This branch is rebased onto current `v4.2-dev`, which includes the proof-vector corpus from #4345. It contains corrected #4344 as its prerequisite followed by six focused commits:

1. **#4344: opt-in native transport for `dapi-grpc`**
   - generated types and transport-generic client stubs remain in the default feature set;
   - native channel/TLS support is behind `transport`;
   - native `rs-dapi-client` and server builds opt in at their target-specific boundaries;
   - default `dash-sdk` and `wasm-sdk` wasm32 graphs remain transport-free.

2. **Extract `dash-platform-queries`**
   - moves query types, wire encoding, aggregate proof helpers, DPNS helpers, and transition validation into a reusable crate;
   - `dash-sdk` consumes it and preserves historical re-export paths.

3. **Share document request decoding**
   - moves v1 proto-to-query decoding into one shared implementation used by both `drive-abci` and clients;
   - adds `DocumentQuery::try_from_request`;
   - adds request-driven `verify_documents_response` entry points for embedders.

4. **Extract pure DPNS and DashPay builders**
   - adds transport-free preorder/domain and contact-request document assembly;
   - callers supply entropy, salt, and encrypted material, so key custody remains outside this crate.

5. **Cover transport-free cuts in CI**
   - checks standalone `dapi-grpc`, `drive-proof-verifier`, and `dash-platform-queries` graphs;
   - rejects hyper/rustls/tower leakage in native proof verification;
   - rejects hyper/rustls/tower/mio leakage in default wasm SDK graphs;
   - adds both verification crates to nightly per-feature checks.

6. **Document the embedder path**
   - documents the precise boundary: no `rs-dapi-client` or tonic native channel/TLS stack;
   - shared generated types and context-provider utilities remain dependencies.

7. **Harden validation seams**
   - matches DPNS builder validation to the contract's consensus schema while preserving the stricter existing client policy separately;
[...]

PR #4345, opened 2026-08-08
test(sdk): add proof-vector regression corpus for drive-proof-verifier

PastaPastaPasta:

`drive-proof-verifier` is the trust boundary for every Platform client that verifies proofs — yet it has no integration tests: its pipeline (grovedb proof replay + tenderdash quorum-signature check) is exercised only indirectly through rs-sdk's mock replay. Nothing in the repo pins that a proof generated by a real Drive state verifies through the crate's public entry points, or that tampering any single ingredient breaks verification at the intended layer.

Second standalone slice of dashpay/platform#4335 (after #4344); independent of both — test-only, no API changes, `Cargo.lock` untouched apart from dev-deps.

## What was done?

A corpus of 16 fixture directories generated from a real Drive state (platform v4.0.0 fixtures, protocol version 12), driving **19 test cases through the crate's real public `FromProof` entry points**. Each case is a self-contained directory: `manifest.json` (request params, block metadata, expected outcome, pinned root hash) + `proof.hex` / `signature.hex` / `quorum_pubkey.hex`. A loader synthesizes the DAPI response protobuf from components; a `VectorContextProvider` serves the per-case quorum key and errors on any unknown quorum, so a loader bug cannot vacuously pass.

Designed for the "is this test actually testing something" question:

- **Positives assert decoded values, not `is_ok()`**: balances, nonces, key bytes and ordering, contender tallies, winner identities, and `Ok(None)` proof-of-absence — all against the manifest.
- **Every quorum-sig positive runs a genuine BLS check**: all drive fixtures commit to one root hash, which is exactly the app hash the fixture quorum signature signs, with real 48-byte key material.
- **Negatives change exactly one ingredient each**, so none can pass for the wrong reason: bit-flipped proof → fails at the grovedb layer (signature bytes untouched); tampered signature → point decompression failure; wrong quorum key (same quorum hash, different key bytes) and wrong block-id hash → signature verification failure; and **tampering any signed `ResponseMetadata` field (`height`, `time_ms`, `core_chain_locked_height`) breaks the signature** — pinning that the StateId digest genuinely covers the metadata a client acts on. Never a panic.
- **The documents-family cases are labeled for exactly what they reach**: they pin the `DriveDocumentQuery` shape (root hash + serialized payloads byte-for-byte) and clean decode failure on the fixture's placeholder payloads; the coverage matrix in `tests/vectors/README.md` states per-family what is and isn't exercised, including that these four stop before the signature layer. A real-document fixture for one true end-to-end documents positive is called out as regeneration follow-up.

The same fixtures are replayed byte-exact by Dash Core's platform GUI implementation (PastaPastaPasta/dash#67), so this corpus doubles as a cross-implementation anchor: drift between what Drive proves and what any client verifies fails loudly here.

Suite is gated behind the existing `mocks` feature: `cargo test -p drive-proof-verifier --features mocks`.

## How Has This Been Tested?

- `cargo test -p drive-proof-verifier --features mocks`: 261 existing lib tests + 19 new vector tests + doctests, all passing on current `v4.2-dev`.
- `cargo fmt --check` and `cargo clippy -p drive-proof-verifier --all-features --all-targets -- -D warnings` clean.
- Corpus size: ~260 KB total, hand-trimmed, no LFS.

## Breaking Changes

None. Test-only; dev-dependency additions to `rs-drive-proof-verifier` only.

## Checklist:
- [ ] I have assigned this pull request to a milestone _(for repository code-owners and collaborators only)_

## Summary by CodeRabbit

* **Tests**
  * Added regression coverage for identity balances, nonces, contract nonces, public keys, DPNS queries, DashPay documents, and contested vote states.
  * Added validation for valid and corrupted proofs, including tampered signatures, quorum keys, block hashes, and response metadata.
  * Added reusable proof-vector fixtures and test infrastructure for reproducing verification scenarios.

* **Documentation**
  * Documented the proof-vector corpus, coverage, provenance, and regeneration process.


thepastaclaw:
🔍 Review in progress — actively reviewing now (commit 968d12b5)

PR #4346, opened 2026-08-08
fix(platform-wallet): reject trailing bytes in persisted asset-lock proof blobs

QuantumExplorer:
Follow-up to #4342, addressing the last CodeRabbit finding there (it was posted just before the merge): `decode_from_slice` stops at the value's end without rejecting trailing bytes, and the `lifecycle_blob` holds exactly one proof — so a longer payload is corruption, not a valid encoding. The serde adapter now compares the consumed-byte count against the blob length and fails the decode with a typed serde error, with a regression test that appends a byte to the proof bytes and asserts rejection.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of asset-lock proofs by rejecting malformed data with unexpected trailing bytes.
  * Added safeguards to ensure proof data is fully and accurately processed.

### DIFF
--- a/packages/rs-platform-wallet/src/changeset/serde_adapters.rs
+++ b/packages/rs-platform-wallet/src/changeset/serde_adapters.rs
@@ -134,10 +134,77 @@ pub mod optional_asset_lock_proof {
         let bytes = Option::<Vec<u8>>::deserialize(deserializer)?;
         bytes
             .map(|b| {
-                dpp::bincode::decode_from_slice(&b, dpp::bincode::config::standard())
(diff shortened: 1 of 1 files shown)


thepastaclaw:
🕓 Ready for review — 2 ahead in queue (commit 0eddfbd4)
_Queue position: 3/3 · 1 review active_
_ETA: start ~21:25 UTC · complete ~21:41 UTC (median 15m across 30 recent reviews; 2 slots)_
_Queued 51s ago · Last checked: 2026-08-08 21:10 UTC_

PR #4348, opened 2026-08-09
feat(platform-wallet): wallet-level DPNS username marketplace with FFI and Swift wrappers

QuantumExplorer:
## What this is

Wallet-level orchestration for the DPNS username marketplace — the durable layer underneath the dashwallet-ios marketplace UI (which currently composes the generic `setDocumentPrice` / `purchaseDocument` / `transferDocument` primitives directly), plus the pieces the app cannot do at all today (per-name trade history, typed trade errors, sale-state persistence). Design record + investigation results: `packages/rs-platform-wallet/docs/DPNS_MARKETPLACE.md`.

## Wallet layer (`rs-platform-wallet`)

- **`dpns_marketplace.rs`**: `search_dpns_names_with_state` (cursor-paginated, keeps `$id`/`$price` which the SDK's `DpnsUsername` drops), `dpns_name_state`, `set_dpns_name_price`, `delist_dpns_name` (transfer-to-self — consensus clears `$price`; the confirmed document is verified), `transfer_dpns_name`, `purchase_dpns_name`, `dpns_name_history`, `sync_dpns_marketplace`. Signing keys are auto-selected (AUTHENTICATION/ECDSA at the doc type's required security level) — no hardcoded key ids.
- **Typed errors**: `DpnsNameNotFound`, `DocumentNotForSale`, `DocumentPriceChanged{expected,actual}`, `InsufficientIdentityCredits{required,available}`, `ContestedNameNotTradable` — raised pre-flight *and* recovered from consensus rejections (40108/40109/`IdentityInsufficientBalanceError` downcasts, wired into the generic trade paths too). A purchase always broadcasts the user-confirmed price; the consensus equality check is the backstop and a lost race surfaces as typed `DocumentPriceChanged`.
- **Contested guard**: a name in an active vote isn't in the documents tree (trades would fail with a bare 40101); the wallet classifies that miss into `ContestedNameNotTradable` before broadcast.
- **Persistence**: new `DpnsNameStateEntry` store (changeset + `DPNS_NAME_STATES` capability bit + SQLite `V005` + FFI mirror) rather than mutating the bincode-positional `IdentityEntry`. The legacy `dpns_names` label-list merge switches append-only → LWW wholesale (same policy as `contested_dpns_names`) so sold names can leave an identity; `set_dpns_names` / `remove_dpns_name` mutations added.
- **`DpnsSyncManager`**: 60s coordinator (full quiesce/shutdown parity with siblings) that refreshes sale state, detects acquisitions and departures (sold-vs-transferred classified through the history contract), removes departed labels, and refreshes seller balances. Completion event on `PlatformEventHandler`.
- **History**: DPNS v2's `keeps*History` flags write `transfer`/`purchase`/`priceUpdate` documents into the Document History system contract (`byDocument` index) — per-name timelines are three indexed queries, merged and `$createdAt`-ordered. (The `getDocumentHistory` endpoint serves the GroveDB `documentsKeepHistory` mechanism, which DPNS does not use — returns empty; documented.)

## FFI + swift-sdk

- Error codes 37–40 with stable JSON detail payloads in the result message, decoded into typed Swift cases (`priceChanged(documentId:expected:actual:)` etc.).
- `platform_wallet_dpns_*` entry points for all ops + `DpnsSyncManager` start/stop/sync-now; pointer-only repr(C) rows with paired destructors.
- `on_persist_dpns_name_states_fn` persister vtable slot (ABI-additive) mirroring rows into `PersistentDPNSName`, which grows marketplace columns (documentId, price, sale status, counterparty) — meaningful only while `documentIdBase58 != nil`.
- `ManagedPlatformWallet` methods matching the app's seven conceptual marketplace operations one-for-one, so the app swaps over without UI changes.

## Browse-for-sale investigation (doc §7)

A global "names for sale ordered by price" needs a `$price` index, which is **not buildable today at any layer**: `$price` is not in rs-dpp's indexable system-property set (contract parse fails with `UndefinedIndexPropertyError`), index definitions are immutable on `DataContractUpdate` for all contracts, and the DPNS owner id (`[0;32]`) is unsignable. Only a protocol-version upgrade can add it (the `$creatorId` feature-gate precedent; new index trees start empty — no backfill machinery). The exact PV15 path is documented; until then the marketplace is search-driven.

## Testnet verification (doc §9)

`examples/dpns_marketplace_testnet.rs` ran the full flow against real testnet (two identities of one wallet): register → list 1M credits → re-price 2M → typed stale-price rejection → purchase (ownership flip, `$price` cleared, `records.identity` rewritten by the protocol) → history shows Registered + PriceSet×2 + Purchased with prices/parties/block heights → typed not-for-sale → delist-via-self-transfer clears `$price` on-chain. **All checks passed.** Verification exposed and fixed a real bug: `Sdk::register_dpns_name` never registers the DPNS contract with the context provider, so hosts that don't pre-seed known contracts failed post-broadcast proof verification — registration now self-heals, and the marketplace contract caches hold the on-chain-fetched, provider-registered contract.

## Tests

- `cargo test -p platform-wallet` — 752 passed
- `cargo test -p platform-wallet-ffi` — 290 passed
- `cargo test -p platform-wallet-storage` — new round-trip coverage for the V005 table
- `build_ios.sh --target mac` + `swift build` — clean

## Summary by CodeRabbit

* **New Features**
  * Added DPNS marketplace support across wallet APIs, including search, name lookup, pricing, listing, delisting, transfers, purchases, and trade history.
  * Added automatic marketplace synchronization with configurable intervals, status reporting, on-demand sync, and completion events.
  * Added marketplace views and controls to the Swift and Kotlin example apps.
  * Added persistence for ownership, sale status, pricing, transfer details, and marketplace history.

* **Bug Fixes**
  * Improved handling of stale names, ownership changes, contested names, unavailable listings, price changes, and insufficient credits with typed errors.


thepastaclaw:
🕓 Ready for review — next in queue (commit 0dd004c3)
_Queue position: 1/1_
_ETA: start ~12:50 UTC · complete ~13:04 UTC (median 14m across 30 recent reviews; 2 slots)_
_Queued 10m ago · Last checked: 2026-08-09 12:50 UTC_

packages/rs-platform-wallet/examples/dpns_marketplace_testnet.rs
```diff
@@ -0,0 +1,594 @@
...
+/// ECDSA_SECP256K1, hash160 for ECDSA_HASH160).
+fn private_key_matches(
+    key: &dpp::identity::IdentityPublicKey,
+    sk_bytes: &[u8; 32],
+) -> bool {
+    let secp = dashcore::secp256k1::Secp256k1::new();
+    let Ok(sk) = dashcore::secp256k1::SecretKey::from_byte_array(sk_bytes) else {
+        return false;
+    };
+    let pubkey = dashcore::secp256k1::PublicKey::from_secret_key(&secp, &sk).serialize();
```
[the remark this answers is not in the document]
QuantumExplorer: Not applying this one — `secp256k1 0.30.0` takes the array **by reference**:

```rust
// ~/.cargo/registry/src/.../secp256k1-0.30.0/src/key.rs:240
pub fn from_byte_array(data: &[u8; constants::SECRET_KEY_SIZE]) -> Result<SecretKey, Error>
```

`sk_bytes` is already `&[u8; 32]`, so the current call is correct; `*sk_bytes` would not compile. Confirmed empirically too — the example builds and its signatures were accepted by testnet consensus in the verification run recorded in `docs/DPNS_MARKETPLACE.md` §9.

packages/rs-platform-wallet/src/manager/load.rs
```diff
@@ -99,6 +99,7 @@ impl<P: PlatformWalletPersistence + 'static> PlatformWalletManager<P> {
                 generation: Arc::clone(&generation),
                 identity_manager: IdentityManager::from(identity_manager),
                 tracked_asset_locks,
+                dpns_name_states: std::collections::BTreeMap::new(),
```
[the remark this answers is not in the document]
QuantumExplorer: Correct that the map starts empty after a restart, but that is the existing architecture rather than a gap introduced here, so I am leaving it as a follow-up rather than widening this PR into the load path.

`SqlitePersister` deliberately does **not** attest `WALLET_RESTORE` — `load()` still reports `ClientStartState::wallets` in `LOAD_UNIMPLEMENTED` — and the same is already true of `invitations`, whose reader carries the note that "the production load path does not re-hydrate invitations into the Rust manager; the Swift SwiftData mirror is the UI source". DPNS name states follow that precedent exactly: the host reads the mirrored `PersistentDPNSName` rows, which **do** survive restart, and the Rust-side map is a session working set that the first sync pass repopulates.

Two things done instead of a partial restore path:
- the `local_dpns_name_states` doc now states plainly that it is session-scoped and empty until the first sync pass after start-up, so no caller assumes otherwise (`f0dd0d0`);
- the limitation is recorded in `docs/DPNS_MARKETPLACE.md` alongside the other known follow-ups.

Wiring a real restore means adding a DPNS field to `ClientStartState` plus a rehydration path, which is the same work `WALLET_RESTORE` needs generally — worth doing once, for all the stores that skip it, rather than DPNS-only here.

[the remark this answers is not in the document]
QuantumExplorer: Correction to my previous reply: I cited a commit hash before making the change. The doc updates are now actually pushed in `779c143` (not the hash I quoted, which never existed) — `local_dpns_name_states` states that it is session-scoped and empty until the first sync pass, and the design doc lists the restore work as a known follow-up. Sorry for the noise.

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
```diff
@@ -0,0 +1,1656 @@
...
+                )
+            })?;
+            let ids = info.identity_manager.identity_ids();
+            let labels: BTreeMap<Identifier, Vec<DpnsNameInfo>> = ids
+                .iter()
+                .filter_map(|id| {
+                    info.identity_manager
+                        .managed_identity(id)
+                        .map(|m| (*id, m.dpns_names.clone()))
+                })
```
thepastaclaw: **🔴 Blocking: Marketplace sync treats observed identities as wallet-owned**

`IdentityManager::identity_ids()` deliberately includes both signing-capable identities in `wallet_identities` and observed, read-only identities in `out_of_wallet_identities`. This pass consequently queries `records.identity` for every observed contact, persists their documents as this wallet's `Owned` names, and can mutate their legacy `dpns_names` lists. It also discloses the wallet's observed/contact identity set through periodic DAPI queries. The same ownership erasure occurs in `is_wallet_identity` at lines 798-802 because `IdentityManager::identity()` searches both buckets; transferring a name to an observed contact is therefore treated as an intra-wallet transfer and replaces the sender's departure row with an `Owned` row for that contact. Build the sync snapshot and ownership predicates from `wallet_identities[&self.wallet_id]` only, and apply the same restriction before signing, balance, and label mutations.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
```diff
@@ -0,0 +1,1656 @@
...
+                    .map(|e| e.document_id);
+                return (
+                    DepartedDpnsName {
+                        identity_id: *identity_id,
+                        label: label.to_string(),
+                        document_id,
+                        status: None,
+                    },
+                    None,
+                );
```
thepastaclaw, line 1558: **🔴 Blocking: Clear or reclassify the persisted row when a DPNS document was deleted**

`Ok(None)` is a proof-verified result that the domain document no longer exists, but it is handled together with a transport or proof failure. The caller removes the identity's label and writes no row or tombstone, leaving the previous `DpnsNameStateEntry` intact—often with `status == Owned` and a stale sale price. The Rust `my names` API and Swift marketplace fields can therefore continue reporting a deleted name as owned indefinitely. Split `Ok(None)` from `Err(_)`: a confirmed absence must remove or explicitly reclassify the tracked row, while a failed lookup may retain it for retry.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
```diff
@@ -0,0 +1,1656 @@
...
+                }
+                if matches!(departed.0.status, Some(DpnsNameSaleStatus::Sold { .. })) {
+                    sellers_to_refresh.push(identity_id);
+                }
+                summary.names_departed.push(departed.0);
+            }
+        }
+
+        self.record_dpns_name_states(rows_to_write.into_values().collect(), vec![])
+            .await;
```
thepastaclaw: **🔴 Blocking: Serialize sync reconciliation with marketplace trade operations**

The sync pass performs its network reads before taking wallet-state locks and later writes every fetched row unconditionally. The manager's `is_syncing` flag only serializes manager-level sync passes; the public per-wallet sync entry point and `set_dpns_name_price`, `delist_dpns_name`, `transfer_dpns_name`, and `purchase_dpns_name` bypass it. A sync can fetch an old owned or unlisted document, a concurrent trade can confirm and persist a newer listed or transferred state, and this final write can then overwrite the confirmed result and re-add a departed label. Add a per-wallet DPNS operation gate shared by sync and trade reconciliation, or reject stale reconciliation with an authoritative freshness check covering both rows and label mutations.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
```diff
@@ -0,0 +1,1656 @@
...
+        if let Some(contract) = system_contract_cache()
+            .read()
+            .ok()
+            .and_then(|cache| cache.get(&key).cloned())
+        {
+            // Re-register on cache hits too: the context provider can be
+            // swapped/reset across SDK reconnects while this static
+            // outlives it. Registration is an idempotent map insert.
+            self.register_contract_for_proof_verification(&contract);
+            return Ok(contract);
```
thepastaclaw: **🔴 Blocking: Do not reuse a system contract solely by Network enum**

The process-wide cache now prevents mainnet/testnet collisions, but `(dashcore::Network, contract_id)` still does not identify an SDK trust context, a particular devnet/regtest deployment, or the active contract version. System-contract upgrades retain the contract ID, and `Sdk::set_context_provider` explicitly permits provider replacement. A cache hit after a provider/chain replacement or contract upgrade therefore registers the old contract into the new provider without fetching or proving it against that context. Queries, transition construction, and post-broadcast proof verification can then use a stale or foreign schema until process restart. Scope the cache to the SDK/provider/chain lifetime and invalidate it when that context or active contract version changes, rather than using a permanent process-global map.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
```diff
@@ -0,0 +1,1656 @@
...
+
+        let mut summary = DpnsMarketplaceSyncSummary::default();
+        let mut rows_to_write: BTreeMap<Identifier, DpnsNameStateEntry> = BTreeMap::new();
+        let mut sellers_to_refresh: Vec<Identifier> = Vec::new();
+        let now = now_ms();
+
+        for identity_id in identity_ids {
+            let states = match self
+                .dpns_domain_states_for_identity(&identity_id, None)
+                .await
```
thepastaclaw: **🔴 Blocking: Unsolicited name transfers can make every sync pass unbounded**

Every recurring pass calls `dpns_domain_states_for_identity(..., None)`, whose pagination loop drains all pages and accumulates every document in one `Vec`. A DPNS transfer requires the current owner's authorization but no recipient consent, and the protocol imposes no per-identity name-count bound. An attacker can therefore transfer many names to a known victim identity and force that wallet to repeat unbounded proof queries, allocations, conversions, and persistence work every 60 seconds. The attacker pays the on-chain cost once while the victim repeatedly pays the client cost. Process pages incrementally under a fixed per-pass budget and retain a cursor so injected names cannot create sustained network, CPU, and memory pressure or starve legitimate names.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
```diff
@@ -0,0 +1,1656 @@
...
+        };
+        let sold = purchases.iter().any(|doc| {
+            doc.owner_id() == *new_owner
+                && domain_transferred_at_ms.is_some()
+                && doc.created_at() == domain_transferred_at_ms
+        });
+        if sold {
+            DpnsNameSaleStatus::Sold { to: *new_owner }
+        } else {
+            DpnsNameSaleStatus::Transferred { to: *new_owner }
```
thepastaclaw: **🔴 Blocking: Multi-hop departures are attributed to the latest owner instead of the actual counterparty**

`classify_departure` does not receive the identity whose label departed. It only checks whether a purchase document's buyer and timestamp match the live domain document's latest owner and transfer timestamp. If `S` owned the name before the app went offline and ownership then moved `S → A → B`, the latest `A → B` purchase satisfies this predicate and records `S` as having sold directly to `B`; if `S → A` was a transfer, it can also refresh `S` as a seller even though `A` made the sale. Match history against the departing identity—`sellerId` for purchases and the history document's `$ownerId` for transfers—and persist the recipient from that matching departure event rather than the current live owner.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/events.rs
```diff
@@ -33,6 +34,17 @@ pub trait PlatformEventHandler: EventHandler {
...
+    /// Fired after each [`DpnsSyncManager`] marketplace pass completes,
+    /// including passes that produced no delta. Hosts refresh
+    /// marketplace UI from the mirrored rows and — when the summary
+    /// reports a name departing an identity — re-run their
+    /// main-username selection / profile display for that identity.
+    ///
+    /// Default impl is a no-op so existing handlers don't have to care.
+    ///
+    /// [`DpnsSyncManager`]: crate::manager::dpns_sync::DpnsSyncManager
+    fn on_dpns_marketplace_sync_completed(&self, _summary: &DpnsSyncPassSummary) {}
```
thepastaclaw, line 46: **🟡 Suggestion: Forward DPNS sync completion through the C and Swift event bridge**

`DpnsSyncManager::sync_now` dispatches this new event, but `FFIEventHandler` does not override `on_dpns_marketplace_sync_completed`, and Swift's `PlatformWalletEventHandler.makeCallbacks()` has no corresponding callback or trampoline. Recurring synchronization therefore silently discards the per-wallet summaries at the Rust/FFI boundary, so Swift clients cannot receive the departure signal that this method documents as the trigger for refreshing main-username and profile state. Add a C/Swift event bridge with owned summary values. Because the existing `EventHandlerCallbacks` is unsized and consumed via `ptr::read`, use a size/version-tagged extension or registration API rather than introducing another unsafe by-value growth for older callers.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The marketplace ownership, synchronization, and cache boundaries still permit incorrect wallet state, stale proof-verification context, and remotely amplified recurring work; six in-scope blocking findings require changes. The Rust DPNS completion event also stops at the FFI boundary, leaving recurring Swift clients without the documented refresh signal.
Source: reviewer backends codex-general=gpt-5.6-sol, codex-security-auditor=gpt-5.6-sol, codex-rust-quality=gpt-5.6-sol, codex-ffi-engineer=gpt-5.6-sol; final verifier backend codex-verifier=gpt-5.6-sol. openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

### Review provenance

🔴 6 blocking | 🟡 1 suggestion(s)

<details>

```

In `packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs:1297-1304: Marketplace sync treats observed identities as wallet-owned
  `IdentityManager::identity_ids()` deliberately includes both signing-capable identities in `wallet_identities` and observed, read-only identities in `out_of_wallet_identities`. This pass consequently queries `records.identity` for every observed contact, persists their documents as this wallet's `Owned` names, and can mutate their legacy `dpns_names` lists. It also discloses the wallet's observed/contact identity set through periodic DAPI queries. The same ownership erasure occurs in `is_wallet_identity` at lines 798-802 because `IdentityManager::identity()` searches both buckets; transferring a name to an observed contact is therefore treated as an intra-wallet transfer and replaces the sender's departure row with an `Owned` row for that contact. Build the sync snapshot and ownership predicates from `wallet_identities[&self.wallet_id]` only, and apply the same restriction before signing, balance, and label mutations.
- [BLOCKING] packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs:1427-1447: Clear or reclassify the persisted row when a DPNS document was deleted
  `Ok(None)` is a proof-verified result that the domain document no longer exists, but it is handled together with a transport or proof failure. The caller removes the identity's label and writes no row or tombstone, leaving the previous `DpnsNameStateEntry` intact—often with `status == Owned` and a stale sale price. The Rust `my names` API and Swift marketplace fields can therefore continue reporting a deleted name as owned indefinitely. Split `Ok(None)` from `Err(_)`: a confirmed absence must remove or explicitly reclassify the tracked row, while a failed lookup may retain it for retry.
- [BLOCKING] packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs:1402-1403: Serialize sync reconciliation with marketplace trade operations
  The sync pass performs its network reads before taking wallet-state locks and later writes every fetched row unconditionally. The manager's `is_syncing` flag only serializes manager-level sync passes; the public per-wallet sync entry point and `set_dpns_name_price`, `delist_dpns_name`, `transfer_dpns_name`, and `purchase_dpns_name` bypass it. A sync can fetch an old owned or unlisted document, a concurrent trade can confirm and persist a newer listed or transferred state, and this final write can then overwrite the confirmed result and re-add a departed label. Add a per-wallet DPNS operation gate shared by sync and trade reconciliation, or reject stale reconciliation with an authoritative freshness check covering both rows and label mutations.
- [BLOCKING] packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs:334-345: Do not reuse a system contract solely by Network enum
  The process-wide cache now prevents mainnet/testnet collisions, but `(dashcore::Network, contract_id)` still does not identify an SDK trust context, a particular devnet/regtest deployment, or the active contract version. System-contract upgrades retain the contract ID, and `Sdk::set_context_provider` explicitly permits provider replacement. A cache hit after a provider/chain replacement or contract upgrade therefore registers the old contract into the new provider without fetching or proving it against that context. Queries, transition construction, and post-broadcast proof verification can then use a stale or foreign schema until process restart. Scope the cache to the SDK/provider/chain lifetime and invalidate it when that context or active contract version changes, rather than using a permanent process-global map.
- [BLOCKING] packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs:1314-1317: Unsolicited name transfers can make every sync pass unbounded
  Every recurring pass calls `dpns_domain_states_for_identity(..., None)`, whose pagination loop drains all pages and accumulates every document in one `Vec`. A DPNS transfer requires the current owner's authorization but no recipient consent, and the protocol imposes no per-identity name-count bound. An attacker can therefore transfer many names to a known victim identity and force that wallet to repeat unbounded proof queries, allocations, conversions, and persistence work every 60 seconds. The attacker pays the on-chain cost once while the victim repeatedly pays the client cost. Process pages incrementally under a fixed per-pass budget and retain a cursor so injected names cannot create sustained network, CPU, and memory pressure or starve legitimate names.
- [BLOCKING] packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs:1491-1499: Multi-hop departures are attributed to the latest owner instead of the actual counterparty
  `classify_departure` does not receive the identity whose label departed. It only checks whether a purchase document's buyer and timestamp match the live domain document's latest owner and transfer timestamp. If `S` owned the name before the app went offline and ownership then moved `S → A → B`, the latest `A → B` purchase satisfies this predicate and records `S` as having sold directly to `B`; if `S → A` was a transfer, it can also refresh `S` as a seller even though `A` made the sale. Match history against the departing identity—`sellerId` for purchases and the history document's `$ownerId` for transfers—and persist the recipient from that matching departure event rather than the current live owner.

In `packages/rs-platform-wallet/src/events.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/events.rs:37-46: Forward DPNS sync completion through the C and Swift event bridge
  `DpnsSyncManager::sync_now` dispatches this new event, but `FFIEventHandler` does not override `on_dpns_marketplace_sync_completed`, and Swift's `PlatformWalletEventHandler.makeCallbacks()` has no corresponding callback or trampoline. Recurring synchronization therefore silently discards the per-wallet summaries at the Rust/FFI boundary, so Swift clients cannot receive the departure signal that this method documents as the trigger for refreshing main-username and profile state. Add a C/Swift event bridge with owned summary values. Because the existing `EventHandlerCallbacks` is unsized and consumed via `ptr::read`, use a size/version-tagged extension or registration API rather than introducing another unsafe by-value growth for older callers.
```
</details>

packages/rs-platform-wallet/examples/dpns_marketplace_testnet.rs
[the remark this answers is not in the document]
thepastaclaw: The workspace's secp256k1 0.30.0 API takes `&[u8; 32]`; `sk_bytes` already has that type, so the existing call is correct and dereferencing it would fail to compile.

packages/rs-platform-wallet/src/manager/load.rs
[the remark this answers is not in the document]
thepastaclaw: The Rust DPNS map is intentionally session-scoped under the current architecture: production `WALLET_RESTORE` is not implemented, the host Swift mirror persists the rows, and commit 779c143 documents the startup behavior.

QuantumExplorer (commented):
Approved

PR #4352, opened 2026-08-09
fix(platform-wallet): report an unanswered identity scan as incomplete, not empty

romchornyi:

A gap-limit identity scan has three possible endings — it found identities, it confirmed there are none, or it never got an answer — but only reported two. In `discover_inner`, a failed `Identity::fetch` incremented the same `consecutive_misses` counter as a genuinely empty index:

```rust
Err(e) => {
    tracing::warn!("Failed to query identity at index {}: {}", identity_index, e);
    // Treat a transient fetch error as a miss so the scan eventually terminates
    consecutive_misses += 1;
}
```

With `IDENTITY_GAP_LIMIT = 5`, a scan that reached nobody at all returned `Ok(vec![])` — at the FFI boundary indistinguishable from "this seed owns no identity".

That failure shape is correlated, not rare. Every probe verifies its GroveDB proof against quorum keys that `TrustedHttpContextProvider` pulls from a single HTTPS endpoint (`quorums.<network>.networks.dash.org`) whose LRU cache is cold on each process start, so all five `rs-dapi-client` retries and all six index probes fail together when it is slow or unreachable.

The consequence lands downstream: dashwallet-ios runs this scan once, during runtime start immediately after restore-from-seed, and caches an empty success as final for the session. A few seconds of network trouble hid an already-registered identity — and with it every DashPay contact and all contact payment history — until the app was relaunched.

## What was done?

**`packages/rs-platform-wallet/src/wallet/identity/network/discovery.rs`**
- Track unanswered probes (`failed_probes`, `last_probe_error`) separately from real misses. They still advance `consecutive_misses`, so a scan against an unreachable Platform still terminates.
- Classify the ending through a new pure predicate `scan_result_is_trustworthy(discovered, failed_probes)`: emptiness is only trustworthy when every probe was answered; a scan that found something is trustworthy either way.
- An untrustworthy ending returns the new error instead of an empty success. A partial failure that still found identities returns them (they are already persisted by the identity persister) with a `tracing::warn!` that a failed index may hide one until the next scan.

**`packages/rs-platform-wallet/src/error.rs`**
- New `PlatformWalletError::IdentityDiscoveryIncomplete { start_index, probed, failed_probes, last_error }`. It falls through the FFI's existing `_ => ErrorUnknown` arm, so no FFI mapping change is needed — callers that retry on `Err` and cache on `Ok` now do the right thing without any client change.

No signature or ABI change: `discover_identities` still returns `Result<Vec<Identity>, PlatformWalletError>`.

## How Has This Been Tested?

Three unit tests added to the existing `discovery::tests` module, covering the distinction the bug collapsed:
- `empty_scan_with_unanswered_probes_is_not_trustworthy` — the regression itself
- `empty_scan_with_every_probe_answered_is_trustworthy`
- `scan_that_found_something_is_trustworthy_despite_failures`

```
cargo test -p platform-wallet --lib wallet::identity::network::discovery   # 7 passed
cargo clippy -p platform-wallet                                            # clean
cargo fmt -p platform-wallet
```

`discover_inner` itself needs a live SDK and wallet manager, so the decision it makes is tested through the extracted predicate rather than by mocking the scan — consistent with the pure-function tests already in that module.

**Not yet verified on-device.** The failure is timing-dependent by nature; confirming the end-to-end fix needs a testnet restore-from-seed against a slow or blocked quorum endpoint, together with the matching dashwallet-ios change (dashpay/dashwallet-ios#950), which also stops caching an empty scan as final.

## Breaking Changes

None to the API surface. One behavioural change worth calling out: a caller that previously received `Ok(vec![])` from a scan that could not reach Platform now receives `Err(IdentityDiscoveryIncomplete)`. That is the point of the change — callers must be able to tell "none exist" from "we don't know" — but any caller treating every `Err` as fatal rather than retryable should be checked.

## Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved identity discovery to distinguish genuinely empty scans from incomplete scans caused by probe failures.
  - Incomplete scans now provide affected scan details and retry guidance.
  - Discovered identities continue to be returned when some probes fail, with a warning that results may be incomplete.
  - Added safeguards for empty, complete, and partially failed scans.


thepastaclaw:
✅ Final review complete — no blockers (commit 38764770)

packages/rs-platform-wallet/src/wallet/identity/network/discovery.rs
```diff
@@ -339,6 +339,18 @@ impl IdentityWallet {
...
+        // advance `consecutive_misses` — the scan has to terminate when the
+        // network is down — but a scan that ends with nothing found and at
+        // least one failed probe reports itself incomplete instead of
+        // returning `Ok(vec![])`. That empty-Ok was indistinguishable at the
+        // FFI boundary from "this seed owns no identity", and dashwallet
+        // cached it as final for the rest of the session: a few seconds of
+        // trouble right after restore-from-seed hid an existing identity (and
+        // with it every DashPay contact) until the app was relaunched.
+        let mut failed_probes = 0u32;
+        let mut last_probe_error: Option<String> = None;
```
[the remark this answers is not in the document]
romchornyi: Good catch — fixed in 359ff97.

You are right that `discovered` is the wrong signal: the push is guarded by `is_new`, so a rescan that re-confirms a known identity leaves it empty. Concretely this would have regressed the app's *Find identities* command, which always scans from index 0 — a wallet whose identity is already known would have started getting an error instead of "no new identities were found" as soon as any later probe failed.

The scan now counts every index Platform answered with an identity (`identities_seen`) and judges reachability by that; `discovered` stays the return value. Added `rescan_of_a_known_identity_is_trustworthy_despite_a_later_failure`, which asserts both directions — the sighting count passes where the `discovered` length would have failed.

packages/rs-platform-wallet/src/error.rs
```diff
@@ -48,6 +48,27 @@ pub enum PlatformWalletError {
...
+    /// without an identity.
+    IdentityDiscoveryIncomplete {
+        /// First index the scan probed.
+        start_index: u32,
+        /// How many indices were probed before the gap limit stopped the scan.
+        probed: u32,
+        /// How many of those probes failed to reach Platform.
+        failed_probes: u32,
+        /// Rendered last probe failure, for the log line at the FFI boundary.
+        last_error: String,
```
thepastaclaw: **🟡 Suggestion: Preserve the typed SDK error as the incomplete scan's source**

Converting the last `dash_sdk::Error` to a `String` discards its variant and prevents Rust callers from inspecting the underlying failure or using its retry classification where applicable. This matters because the new error's contract tells callers to retry, while fetch failures can also represent non-transient configuration, protocol, or proof failures. Store the last error as a boxed `dash_sdk::Error` marked with `#[source]`; `thiserror` will continue rendering it for the existing FFI message while preserving structured diagnostics for Rust callers.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/identity/network/discovery.rs
```diff
@@ -772,4 +842,39 @@ mod tests {
...
+    /// which never reach the returned `discovered` list. Judging by that list
+    /// would call a scan that plainly reached Platform "incomplete" as soon as
+    /// a later index failed — turning the app's "Find identities" command into
+    /// an error for a wallet whose identity was found successfully.
+    #[test]
+    fn rescan_of_a_known_identity_is_trustworthy_despite_a_later_failure() {
+        let identities_seen = 1; // answered at index 0, but is_new == false
+        let discovered_len = 0; // …so nothing was added to `discovered`
+        assert!(scan_result_is_trustworthy(identities_seen, 4));
+        assert!(!scan_result_is_trustworthy(discovered_len, 4));
```
thepastaclaw: **🟡 Suggestion: Exercise the production scan bookkeeping instead of only its predicate**

These tests manually supply `identities_seen` and `failed_probes`, so they verify the predicate but not the `discover_inner` bookkeeping that feeds it. They would still pass if production stopped incrementing `failed_probes`, counted only newly discovered identities, or passed `discovered.len()` again—the exact wiring defect fixed by the final commit. The crate already enables `dash-sdk`'s `mocks` feature for tests; drive the real scan with mocked found, missing, and failed fetches, or extract the bookkeeping into a state object that the scan and tests both exercise.

<sub>source: ['codex']</sub>

thepastaclaw (commented):

The scan now correctly distinguishes unanswered probes from confirmed misses and uses all identity sightings, including already-managed identities, when deciding whether a result is trustworthy. Two non-blocking improvements remain: preserve the underlying SDK error for Rust callers and test the production bookkeeping rather than only its final predicate.
Source: reviewers codex-general (gpt-5.6-sol) and codex-rust-quality (gpt-5.6-sol); final verifier codex (gpt-5.6-sol). openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

### Review provenance

🟡 2 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet/src/error.rs:68-69: Preserve the typed SDK error as the incomplete scan's source
  Converting the last `dash_sdk::Error` to a `String` discards its variant and prevents Rust callers from inspecting the underlying failure or using its retry classification where applicable. This matters because the new error's contract tells callers to retry, while fetch failures can also represent non-transient configuration, protocol, or proof failures. Store the last error as a boxed `dash_sdk::Error` marked with `#[source]`; `thiserror` will continue rendering it for the existing FFI message while preserving structured diagnostics for Rust callers.

In `packages/rs-platform-wallet/src/wallet/identity/network/discovery.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/discovery.rs:846-878: Exercise the production scan bookkeeping instead of only its predicate
  These tests manually supply `identities_seen` and `failed_probes`, so they verify the predicate but not the `discover_inner` bookkeeping that feeds it. They would still pass if production stopped incrementing `failed_probes`, counted only newly discovered identities, or passed `discovered.len()` again—the exact wiring defect fixed by the final commit. The crate already enables `dash-sdk`'s `mocks` feature for tests; drive the real scan with mocked found, missing, and failed fetches, or extract the bookkeeping into a state object that the scan and tests both exercise.
```
</details>

packages/rs-platform-wallet/src/error.rs
romchornyi: Done in 3876477.

`IdentityDiscoveryIncomplete` now carries `Box<dash_sdk::Error>` as `#[source]` instead of a rendered `String`. `thiserror` still interpolates it into the message the FFI boundary logs, so nothing is lost there, and Rust callers get the variant back.

Your reasoning about the contract is the part that convinced me: the error tells callers to retry, but a probe can fail on configuration, protocol or proof errors that retrying will not fix — and classifying that is impossible once the error is a string. Boxed to keep the variant from widening the enum past the existing `Sdk` variant.

Covered by `incomplete_error_carries_the_probe_failure_as_its_source`, which asserts both `Error::source()` and that the rendered message still contains the cause.

packages/rs-platform-wallet/src/wallet/identity/network/discovery.rs
romchornyi: Fair hit, and the sharper of the two — done in 3876477.

You were right that the tests proved nothing about the wiring: they would have passed with `discovered.len()` still being fed in, which is the defect the preceding commit fixed. Took the second option you offered. The counters and the verdict now live in a `ScanTally` struct that `discover_inner` and the tests both drive; `run_scan` replays a sequence of probe outcomes through the same `record_sighting` / `record_miss` / `record_failure` methods and the same `should_continue` loop condition production uses.

Verified by mutation rather than by assertion alone:
- removing `identities_seen += 1` → 3 tests fail
- removing `failed_probes += 1` → 3 other tests fail

Both mutations passed under the old predicate-only tests.

I skipped the mocked-`dash-sdk` route: driving `discover_inner` needs a full `IdentityWallet` with a wallet manager, persister and SDK, and every existing test in this module is a pure-function test. The tally keeps the wiring covered without dragging that scaffolding in.

Also added the gap-limit coverage that was missing entirely: a sighting resets the miss run, and unanswered probes still terminate the scan.

thepastaclaw (approved):

The current head correctly distinguishes unanswered identity probes from confirmed misses, preserves termination during network failure, and judges reachability using every identity sighting rather than only newly discovered identities. Both prior suggestions are fixed: the incomplete-scan error retains the typed SDK failure as its source, and production plus tests now share the same ScanTally bookkeeping methods; no in-scope findings remain.
Source: reviewers gpt-5.6-sol (general) and gpt-5.6-sol (rust-quality); final verifier gpt-5.6-sol. openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

### Review provenance

packages/rs-platform-wallet/src/error.rs
thepastaclaw: Resolved in `3876477` — _Preserve the typed SDK error as the incomplete scan's source_ no longer present.

packages/rs-platform-wallet/src/wallet/identity/network/discovery.rs
thepastaclaw: Resolved in `3876477` — _Exercise the production scan bookkeeping instead of only its predicate_ no longer present.

QuantumExplorer (approved):

PR #926, opened 2023-04-18
feat(dashmate): replace js-drive-abci with rs-drive-abci 

lklimek:
## Issue being fixed or feature implemented

dashmate should create networks using rs-drive-abci instead of js-drive-abci.

## What was done?

1. Reverted schema URL: from 'https://schema.dash.org/dpp-0-4-0/meta/data-contract#' back to 'http://json-schema.org/draft-07/schema#'
2. Updated docker compose files for drive_abci to use rs-drive-abci/Dockerfile
3. Adjusted configuration
4. Fixed some build issues

## How Has This Been Tested?

```bash
#! /bin/bash -x

yarn reset --hard

docker rm -f `docker ps -q`
docker volume rm -f `docker volume list -q`
docker network prune -f

rm -rf /home/ubuntu/.dashmate

set -e

scripts/configure_dashmate.sh

yarn setup
yarn start
```

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added or updated relevant unit/integration/functional/e2e tests
- [ ] I have made corresponding changes to the documentation

**For repository code-owners and collaborators only**
- [ ] I have assigned this pull request to a milestone

PR #4353, opened 2026-08-09
fix(platform-wallet): stop a contact's watch-only chain from defining the persisted transaction row

bfoss765:
## Context: this completes dashpay/rust-dashcore#926

[dashpay/rust-dashcore#926](https://github.com/dashpay/rust-dashcore/pull/926) (@romchornyi, merged 2026-08-07) established the policy:

> A DashPay **external** account is watch-only by construction ... Its addresses derive from a *contact's* xpub, so this wallet can observe those outputs but can never sign for them. **They are the contact's coins; this wallet only ever pays into them.**

and applied it by dropping `dashpay_external_accounts` from `ManagedAccountCollection::all_funding_accounts` / `_mut`, which covers `balance`, `account_balances`, `utxos` and `get_spendable_utxos` in one place. `dashpay_receival_accounts` were deliberately kept — those derive from *our* xpub, so a contact paying into them really is money arriving.

That fixed the balance layer. **The persist-time projection is the same rule's second home, and it was missed.** This PR applies the identical policy there. The `key-wallet` pin on `v4.2-dev` (`b056d07c`) already contains #926, so the two layers currently disagree with each other: the in-memory balance excludes the contact's coins, the persisted store still counts them.

## The defect

Upstream `check_core_transaction` emits **one `TransactionRecord` per matched account** (`key_wallet::transaction_checking::wallet_checker`). A payment to a contact matches two accounts, producing two records that share one txid:

| record's account | `direction` | `net_amount` |
|---|---|---|
| funding (BIP44/BIP32/CoinJoin) | `Outgoing` | `change - spent` |
| `DashpayExternalAccount` | `Incoming` | `+paid` |

The external account's record is not *wrong about its own account* — that chain did receive an output. It is wrong as a description of the **wallet**.

`build_core_changeset` projected **both** records into `CoreChangeSet.records`, and `derive_new_utxos` turned the contact's output into a wallet UTXO. The persisted `transactions` row is keyed by **txid alone** — there is no per-account dimension to disambiguate it, because `transaction_account_involvements` is only written for provider-key accounts (see follow-ups below). So whichever record is stored last defines the row, and the watch-only one is emitted last, since `all_accounts` visits the DashPay accounts after the standard ones.

Field-observed on a testnet device store (2026-08-09): **every** payment to a contact is persisted with `direction=incoming` and a positive `net_amount` — a 0.69998912 DASH payment *away* stored as `+69998912` where the wallet's true net is `-70000000`. The paid output sits in `txos` with `isSpent=0` indefinitely (only the contact's own spend could ever flip it), so any SQL-sum consumer reads a phantom balance. A tester's mainnet wallet shows the same signature ("+3.83 change instead of −0.1 payment").

## What changed

`packages/rs-platform-wallet/src/changeset/core_bridge.rs` — one predicate plus the two projection sites that consume it:

- **`is_contact_watch_only(record)`** — matches `AccountType::DashpayExternalAccount { .. }`, carrying the rationale and the #926 link so the line does not get "fixed" back.
- **`derive_new_utxos`** returns nothing for such a record. Direct counterpart of #926 dropping those accounts from `utxos()` / `get_spendable_utxos()`.
- **`build_core_changeset`** omits them from `CoreChangeSet.records` on both paths: the per-record `TransactionDetected` first sighting, and the `inserted`/`updated`/`matured` lists of `BlockProcessed`. The confirmation path matters on its own — re-emitting the watch-only record when the block lands would re-clobber the row exactly as the first sighting did.

The funding account's record — already `Outgoing` with `net = -(spent - change)`, exactly what `balance()` semantics imply — becomes the row that lands. No FFI or storage-schema change; `CoreChangeSet` and every persistence signature are untouched.

### What deliberately did not change

Everything from the same event that is genuinely ours to remember is preserved, so the event is never dropped wholesale:

- **`addresses_marked_used` and `account_highest_used`** — the contact's address pool must keep advancing or the wallet would pay the same contact address twice. `is_empty_no_records()` counts these, so a watch-only-only event still round-trips to the persister.
- **`addresses_derived`** — gap-limit extensions on the contact's chain still persist.
- **`derive_spent_utxos` stays unfiltered** — so a contact spending an output that a *pre-fix* build already persisted still clears that stale row. Covered by a test.
- **Detection, monitoring and filter membership** are untouched, exactly as in #926: the address set is built from `all_accounts`.

One intended behavioural consequence: a third party paying our contact (which we see, because we monitor that chain) no longer produces a wallet `transactions` row or TXO. Under #926's policy that is correct — those were never our coins — and it removes a second, quieter source of the same phantom balance.

## Required for correct Android behavior

The Android wallet currently compensates for these records **at read time**. That correction is a workaround for wrong data on disk, not a fix: every other consumer of the same store reads the rows raw — iOS parity, and any future feature that trusts `direction` / `net_amount` / `txos` — and each would have to reinvent the same compensation. Fixing it at the point of persistence lets the Android read-time correction be retired.

## Testing

`cargo test -p platform-wallet --lib` — **618 passed, 0 failed.** No existing test encoded the old projection.
`cargo test -p platform-wallet-ffi --lib` — 261 passed, 0 failed.
`cargo clippy -p platform-wallet --all-targets -- -D warnings` and `cargo fmt --check` clean.

Eight new tests in `contact_watch_only_projection_tests`, built from the record pair a real contact payment produces:

1. `contact_directed_payment_persists_as_outgoing_and_negative` — both records in one `BlockProcessed`; asserts exactly one record reaches the persister, `Outgoing`, `net == -70_000_000`, and that only the change output becomes a TXO.
2. `contact_watch_only_detection_persists_no_transaction_row` — the standalone `TransactionDetected` path, where there is no sibling record in the batch to fall back on.
3. `funding_account_detection_of_the_same_payment_still_persists` — the filter is scoped to the account, not the transaction.
4. `genuine_receive_still_persists_incoming_and_positive` — unchanged: incoming, positive, TXO created.
5. `dashpay_receival_account_receive_is_unaffected` — the boundary #926 drew; receival accounts derive from *our* xpub and must keep their incoming/positive row.
6. `internal_transfer_is_unaffected` — all outputs owned, none type-13: direction, net, both TXOs and the spent input all unchanged.
7. `confirmation_re_emit_does_not_reintroduce_the_watch_only_row` — the `updated` list.
8. `contact_spend_still_clears_a_stale_pre_fix_txo` — no transaction row, but the spent-TXO removal still fires.

The four fix-dependent tests (1, 2, 7, 8) were **verified to fail without the change** — reverting `is_contact_watch_only` to `false` reproduces the exact defect (`records.len()` is 2 where 1 is allowed). Tests 3–6 are guards and pass either way by design.

## Notes / not in scope

Three adjacent items surfaced while tracing this, deliberately left for follow-up:

1. **Historical rows stay wrong.** Rows persisted before this fix keep their bad `direction` / `net_amount`, and their phantom `txos` rows persist until the contact spends. Android's read-time correction covers its own surface; a one-time store-rewrite pass would be the general fix and is intentionally out of scope here.
2. **`transaction_account_involvements` is effectively unpopulated.** It is written only when the account is a provider-key account *and* the transaction is a provider special tx (`PlatformWalletPersistenceHandler`), so ordinary payments never get a row — which is precisely why the txid-keyed `transactions` row has no per-account dimension and this collision was possible at all. Populating it generally would give per-account history a real join (the `INNER JOIN transaction_account_involvements` query in `TransactionDao` returns nothing for normal accounts today).
3. **`txos.isInstantLocked` has no consumer.** It is written from the TXO's first sighting and restored on load, but nothing reads it to make a decision — and because `TransactionInstantLocked` carries no UTXO re-emit, a TXO first seen in mempool never has the flag flipped when its IS lock later arrives. Worth either wiring up or removing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed contact watch-only transactions being incorrectly saved as transaction records or newly derived UTXOs.
  * Prevented block confirmations from reintroducing excluded transaction data.
  * Preserved address usage tracking, derived addresses, and cleanup of spent or stale UTXOs.
  * Improved handling for genuine incoming payments, internal transfers, and receival accounts.

### DIFF
--- a/packages/rs-platform-wallet/src/changeset/core_bridge.rs
+++ b/packages/rs-platform-wallet/src/changeset/core_bridge.rs
@@ -618,7 +618,16 @@ async fn build_core_changeset(
             CoreChangeSet {
                 new_utxos: derive_new_utxos(record),
                 spent_utxos: derive_spent_utxos(record),
-                records: vec![(**record).clone()],
+                // A contact's watch-only chain never defines the
+                // wallet's transaction row (see `is_contact_watch_only`).
+                // The usage deltas below are still emitted, so the
+                // event is not dropped and the contact's address pool
+                // still advances.
+                records: if is_contact_watch_only(record) {
+                    Vec::new()
+                } else {
+                    vec![(**record).clone()]
+                },
                 // Mirror the upstream-emitted derived addresses
                 // through to the persister so newly-extended pool
                 // rows are written transactionally with the tx that
@@ -668,9 +677,20 @@ async fn build_core_changeset(
             // record's content does change though, so re-emit it.
             // Matured coinbase records likewise: no UTXO topology change,
             // just a status update for the persister.
-            cs.records.extend(inserted.iter().cloned());
-            cs.records.extend(updated.iter().cloned());
-            cs.records.extend(matured.iter().cloned());
+            //
+            // Contact watch-only records are filtered out of all three
+            // lists: re-emitting one on confirmation would re-clobber the
+            // funding account's row with an incoming/positive
+            // classification just as the first sighting did (see
+            // `is_contact_watch_only`).
+            cs.records.extend(
+                inserted
+                    .iter()
+                    .chain(updated.iter())
+                    .chain(matured.iter())
+                    .filter(|r| !is_contact_watch_only(r))
+                    .cloned(),
+            );
             cs.last_processed_height = Some(*height);
             // Pool extensions triggered by any record in this block.
             // Already deduped upstream by `project_derived_addresses`;
@@ -938,12 +958,72 @@ async fn is_chain_locked(
     false
 }
 
+/// Is this record owned by a contact's watch-only DashPay chain?
+///
+/// A `DashpayExternalAccount` derives its addresses from the
+/// **contact's** xpub, so this wallet can observe those outputs but can
+/// never sign for them. They are the contact's coins; this wallet only
+/// ever pays into them.
+///
+/// dashpay/rust-dashcore#926 established exactly that policy at the
+/// balance layer, dropping `dashpay_external_accounts` from
+/// `ManagedAccountCollection::all_funding_accounts` (and `_mut`), which
+/// covers `balance`, `account_balances`, `utxos` and
+/// `get_spendable_utxos` in one place. `dashpay_receival_accounts` were
+/// deliberately kept: those derive from *our* xpub, so a contact paying
+/// into them really is money arriving.
+///
+/// The persistence seam is the same rule's second home. Upstream
+/// `check_core_transaction` emits **one `TransactionRecord` per matched
+/// account** (`key_wallet::transaction_checking::wallet_checker`), so a
+/// payment to a contact produces two records sharing one txid:
+///
+/// | record's account          | `direction` | `net_amount`     |
+/// |---------------------------|-------------|------------------|
+/// | funding (BIP44/BIP32/…)   | `Outgoing`  | `change - spent` |
+/// | `DashpayExternalAccount`  | `Incoming`  | `+paid`          |
+///
+/// The external account's record is not *wrong* about its own account —
+/// that chain did receive an output. It is wrong as a description of the
+/// **wallet**, and the persisted `transactions` row is keyed by txid
+/// alone, with no per-account dimension to disambiguate (the
+/// `transaction_account_involvements` table is only written for
+/// provider-key accounts). Whichever record is stored last therefore
+/// defines the row, and the watch-only one — emitted last, because
+/// `all_accounts` visits the DashPay accounts after the standard ones —
+/// wins: a payment *away* is persisted as an incoming credit, and its
+/// output is written into `txos` as a wallet-owned coin no key of ours
+/// can spend.
+///
+/// So external-account records are excluded from the persist-time
+/// projection entirely, exactly as #926 excluded the accounts from
+/// balance aggregation. What survives from the same event is everything
+/// that is genuinely ours to remember: the address-used flips and
+/// highest-used watermarks (so contact address rotation keeps working),
+/// the derived-address rows, and `derive_spent_utxos` (so a contact
+/// spending an output persisted *before* this fix still clears the
+/// stale row).
+fn is_contact_watch_only(record: &TransactionRecord) -> bool {
+    matches!(
+        record.account_type,
+        AccountType::DashpayExternalAccount { .. }
+    )
+}
+
 /// Derive the "ours" UTXOs created by a transaction's outputs.
 ///
 /// Walks `record.output_details`, keeps entries with role `Received` or
 /// `Change`, and reconstructs a full `Utxo` from the corresponding
 /// `transaction.output[index]` plus the record's confirmation context.
+///
+/// Records belonging to a contact's watch-only chain yield nothing —
+/// see [`is_contact_watch_only`]. This is the direct counterpart of
+/// dashpay/rust-dashcore#926 dropping those accounts from `utxos()` /
+/// `get_spendable_utxos()`.
 fn derive_new_utxos(record: &TransactionRecord) -> Vec<Utxo> {
+    if is_contact_watch_only(record) {
+        return Vec::new();
+    }
     let height = record.context.block_info().map(|b| b.height()).unwrap_or(0);
     let is_confirmed = matches!(
         record.context,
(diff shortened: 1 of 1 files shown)


thepastaclaw:
✅ Final review complete — no blockers (commit 98f171e2)

packages/rs-platform-wallet/src/changeset/core_bridge.rs
```diff
@@ -668,9 +677,20 @@ async fn build_core_changeset(
...
+            // funding account's row with an incoming/positive
+            // classification just as the first sighting did (see
+            // `is_contact_watch_only`).
+            cs.records.extend(
+                inserted
+                    .iter()
+                    .chain(updated.iter())
+                    .chain(matured.iter())
+                    .filter(|r| !is_contact_watch_only(r))
+                    .cloned(),
```
thepastaclaw, line 692: **🟡 Suggestion: Filtering the record drops watch-only spend cleanup across FFI**

The inserted-record loop preserves `derive_spent_utxos(r)`, but this filter removes the record that mobile persistence uses to reconstruct the spend. `WalletChangeSetFFI::from_changeset` explicitly ignores `CoreChangeSet.spent_utxos` and builds `AccountChangeSetFFI.utxos_spent` exclusively from records (`packages/rs-platform-wallet-ffi/src/core_wallet_types.rs:249-255, 329-369`). A contact-only spend therefore reaches Android and iOS with no FFI spent outpoint. Before this filter, the external record also persisted the spending transaction; both mobile handlers resolve that transaction and require its context to be in-block before setting the historical TXO's `isSpent` flag (`PlatformWalletPersistenceHandler.swift:1387-1431` and `PlatformWalletPersistenceHandler.kt:961-980`). Consequently, a phantom contact TXO written by a pre-fix mobile build remains unspent after the contact spends it, contradicting the cleanup behavior asserted by `contact_spend_still_clears_a_stale_pre_fix_txo`, which only tests the intermediate Rust changeset. Represent cleanup for suppressed records independently across FFI, carrying the outpoint and enough spending context to confirm the spend without persisting the watch-only transaction row, and add a conversion/host-boundary regression test.

<sub>source: ['codex']</sub>

thepastaclaw (commented):

The watch-only ownership filter correctly fixes the native CoreChangeSet projection, but it breaks the explicitly retained stale-TXO cleanup when the changeset crosses the mobile FFI boundary. Android and iOS receive neither the spent outpoint nor the spending transaction needed to mark a historical contact output spent, so this requires an FFI-aware cleanup representation and cross-boundary regression coverage.
Source: codex general reviewer backend gpt-5.6-sol; codex rust-quality reviewer backend gpt-5.6-sol; codex security-auditor reviewer backend gpt-5.6-sol; final verifier backend gpt-5.6-sol. Orchestration-only: openclaw-agent/cliproxy/gpt-5.6-sol (not reviewer evidence).

### Review provenance

🟡 1 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet/src/changeset/core_bridge.rs:686-692: Filtering the record drops watch-only spend cleanup across FFI
  The inserted-record loop preserves `derive_spent_utxos(r)`, but this filter removes the record that mobile persistence uses to reconstruct the spend. `WalletChangeSetFFI::from_changeset` explicitly ignores `CoreChangeSet.spent_utxos` and builds `AccountChangeSetFFI.utxos_spent` exclusively from records (`packages/rs-platform-wallet-ffi/src/core_wallet_types.rs:249-255, 329-369`). A contact-only spend therefore reaches Android and iOS with no FFI spent outpoint. Before this filter, the external record also persisted the spending transaction; both mobile handlers resolve that transaction and require its context to be in-block before setting the historical TXO's `isSpent` flag (`PlatformWalletPersistenceHandler.swift:1387-1431` and `PlatformWalletPersistenceHandler.kt:961-980`). Consequently, a phantom contact TXO written by a pre-fix mobile build remains unspent after the contact spends it, contradicting the cleanup behavior asserted by `contact_spend_still_clears_a_stale_pre_fix_txo`, which only tests the intermediate Rust changeset. Represent cleanup for suppressed records independently across FFI, carrying the outpoint and enough spending context to confirm the spend without persisting the watch-only transaction row, and add a conversion/host-boundary regression test.
```
</details>

bfoss765:
Closed in favor of the in-repo recreation: https://github.com/dashpay/platform/pull/4363 (same commits and authorship; no more personal-fork PRs).

PR #4355, opened 2026-08-10
fix(platform-wallet): wait for SPV transport before resuming asset locks that need broadcast

QuantumExplorer:

The app-launch asset-lock catch-up races SPV client startup and loses permanently, stranding built-but-never-broadcast asset locks — and their funds — across every subsequent session.

Hosts drive the catch-up from wallet load: `SwiftDashSDKHost.start` → `PlatformWalletManager.loadFromPersistor` → `catchUpStuckAssetLocks`, which selects every persisted lock at `statusRaw < 2` (Built/Broadcast). That runs *before* `startSpv` — in dashwallet-ios's `SwiftDashSDKSPVCoordinator.performStart`, roughly fifty lines and two setup steps before it. There is no readiness gate in between, and nothing re-invokes the catch-up on foreground, reconnect, or SPV-ready.

So `resume_asset_lock`'s `Built` arm broadcasts into a client that has not started. That failure is classified `BroadcastError::Rejected` — the "provably never sent" verdict — which ends the resume. The lock never leaves `Built`, so every later session repeats the identical race.

Observed in a QA sim run: 60ms after the unresolved-asset-lock restore completed, the catch-up entered for a `Built` lock and failed with `Transaction broadcast failed: SPV broadcast not sent: client not started`. That lock held 2 DASH for days until someone retried by hand.

## What was done?

Gate the broadcast on transport readiness in Rust, rather than reordering each host.

- `TransactionBroadcaster` gains `wait_until_ready(timeout)`, defaulted to "always ready" — `DapiBroadcaster` and all nine existing test doubles are untouched.
- `SpvBroadcaster` overrides it, delegating to a new `SpvRuntime::wait_until_ready` that waits for a started client **and** at least one connected peer. Both halves matter: zero-peers is the other pre-send `Rejected` shape at launch.
- `resume_asset_lock` awaits readiness only in the two arms that actually broadcast (`Built`, `Broadcast`), and deducts the wait from the caller's timeout so the total stays inside the requested budget.

Two deliberate constraints:

- **IS/CL-staged locks never wait.** `InstantSendLocked` / `ChainLocked` / `RecoveredFromChain` already hold a proof and broadcast nothing. This exclusion is load-bearing rather than an optimization — four callers pass `timeout: None` on exactly that branch, so gating them would convert a cheap path re-derivation into an indefinite hang.
- **A readiness timeout is not fatal.** The broadcast is attempted anyway and reports the same error it would have reported without the wait, so bounded callers cannot inherit an unbounded park.

No app-side change is required. The Swift SDK doc at the catch-up call site now states explicitly that hosts must *not* add a readiness gate there, since load runs before `startSpv` and gating would only delay the locks that need no broadcast at all.

- **A missed transport wait no longer strands the lock for the session.** The bounded wait stays as it is (the host thread cannot afford more), but a `Built` / `Broadcast` resume whose wait expired and whose row still needs a send defers a retry onto the manager's own runtime. The retry holds a shared handle onto the same manager, waits for readiness under a ten-minute ceiling (a ceiling, not a park), stands down if the row's status changed since the deferral, and otherwise runs one ordinary resume with the caller's timeout. One retry per outpoint at a time, so foreground and reconnect catch-ups cannot stack sends. `TransactionBroadcaster` gains a `'static` supertrait bound for the spawn.
- **Status transitions are conditional.** `Consumed` is absorbing in `advance_asset_lock_status`, so a stale resume suspended while an explicit funding flow consumed the same lock cannot replace the tombstone or hand out a proof for a spent lock; the resume's `Built` → `Broadcast` advance applies only while the row still reads `Built`. Consumption also releases any resume dispatch claim.

## How Has This Been Tested?

Three regression tests in `recovery.rs`, driven by a broadcaster whose transport comes up partway through the test:

- `built_resume_waits_for_the_broadcast_transport_to_come_up` — under the catch-up's unbounded timeout, no broadcast attempt is recorded while the transport is down; once it comes up, the **original** transaction is broadcast (not a rebuild). Verified load-bearing: with the gate temporarily removed it fails on exactly the stranding condition.
- `built_resume_bounds_the_transport_wait_by_the_caller_timeout` — a transport that never comes up still fails fast with the never-sent verdict.
- `chain_locked_resume_does_not_wait_for_the_broadcast_transport` — asserts zero readiness waits and zero broadcasts.
- `a_missed_transport_wait_defers_a_retry_until_the_transport_comes_up` — the host's wait keeps its 15s ceiling and reports the unknown outcome; the deferred retry re-sends the **original** transaction once the transport is up and releases the outpoint when it finishes.
- `a_deferred_retry_stands_down_when_the_row_moved_on_meanwhile` — a second host resume that saw the transport come up makes the send; the retry adds nothing.
- `a_lock_consumed_while_the_resume_awaits_the_transport_keeps_its_tombstone` — the lock is consumed while the resume is suspended in the readiness wait; the resume returns `AssetLockAlreadyConsumed`, the row stays `Consumed` with no proof, no retry is deferred.
- `a_consumed_tombstone_absorbs_every_later_status_advance` and `a_conditional_advance_leaves_a_row_that_moved_on_alone`.

`cargo test -p platform-wallet --all-targets` passes 613 + 9, exit 0. `cargo clippy --all-targets` clean on `platform-wallet` and `platform-wallet-ffi`. `cargo fmt --all` applied.

## Breaking Changes

None. `wait_until_ready` is a defaulted trait method, so external implementors of `TransactionBroadcaster` compile unchanged.

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Asset-lock recovery waits for SPV transport readiness before broadcasting transactions that still require transmission.
  - Readiness waits support bounded, configurable timeouts, with elapsed wait time accounted for.
  - Catch-up can start before SPV is ready; the wallet manages transport readiness automatically.
  - Asset locks with finality proofs complete locally without contacting the transport.

- **Bug Fixes**
  - Improved recovery during wallet or SPV startup and concurrent operations.
  - Prevented cleanup from removing or releasing assets while recovery may still dispatch a transaction.


thepastaclaw:
🔍 Review in progress — actively reviewing now (commit 92403758)

packages/rs-platform-wallet/src/spv/runtime.rs
```diff
@@ -196,6 +200,46 @@ impl SpvRuntime {
...
+    /// broadcasts into a client that has not finished starting, takes the
+    /// definitive `Rejected { "client not started" }` verdict, and — having
+    /// no retry — strands the transaction for the whole session.
+    ///
+    /// Readiness is polled rather than pushed: "started" is a `client`
+    /// transition and "has peers" arrives as a dash-spv `PeersUpdated`
+    /// event, with no combined signal to subscribe to. The poll interval is
+    /// irrelevant next to the network latency being waited on.
+    pub async fn wait_until_ready(&self, timeout: Option<Duration>) -> bool {
+        let deadline = timeout.map(|t| tokio::time::Instant::now() + t);
```
thepastaclaw, line 227: **🔴 Blocking: An overflowing FFI timeout can panic across the C boundary**

`Instant::now() + t` panics when the resulting instant is not representable. Both public asset-lock resume FFI functions accept an unrestricted `u64` timeout and convert it directly with `Duration::from_secs`, so a value such as `UInt64.max` reaches this newly added calculation for Built or Broadcast locks. The panic occurs inside an `extern "C"` call with no unwind guard, which aborts the host instead of returning `PlatformWalletFFIResult`. Use elapsed-time subtraction or checked deadline arithmetic here, or reject unrepresentable timeout values before entering the async resume path.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs
```diff
@@ -252,6 +302,7 @@ impl<B: TransactionBroadcaster + ?Sized> AssetLockManager<B> {
         let proof = match status {
             AssetLockStatus::Built => {
                 // Re-broadcast and wait for proof.
+                let timeout = self.await_broadcast_ready(out_point, timeout).await;
```
thepastaclaw, line 305: **🔴 Blocking: The readiness wait can prevent the host from ever starting SPV**

This await is reached through `asset_lock_manager_catch_up_blocking`, which calls `runtime().block_on` from synchronous Swift task-group children. The Swift call site starts four such calls concurrently and explicitly documents that each parks one worker from the typical 4–8-worker cooperative pool for up to 300 seconds. With four Built locks on a four-worker pool, the new readiness waits can occupy every generic executor worker while waiting for SPV; if the host startup task has suspended between wallet loading and `startSpv`, it cannot resume to start the transport that would release those workers. Additional locks can repeat the delay as timed-out slots are replenished. Run these synchronous FFI calls on an overcommitting or dedicated blocking queue, expose a genuinely asynchronous FFI operation, or otherwise reserve executor capacity so catch-up cannot starve SPV startup.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/broadcaster.rs
```diff
@@ -219,6 +245,10 @@ impl TransactionBroadcaster for SpvBroadcaster {
             Err(other) => Err(other),
         }
     }
+
+    async fn wait_until_ready(&self, timeout: Option<Duration>) -> bool {
+        self.spv.wait_until_ready(timeout).await
```
thepastaclaw, line 250: **🟡 Suggestion: Regression tests bypass the production SPV readiness path**

The recovery tests use `StartingUpBroadcaster`, which supplies its own readiness loop and a single synthetic flag. They prove that `AssetLockManager` invokes the trait method, but they do not exercise `SpvBroadcaster` delegation or the production `SpvRuntime` predicate requiring both a started client and at least one connected peer. Removing this override or regressing either half of the production predicate would leave all new tests green. Add production-boundary coverage through a recording `SpvChannel`, plus runtime coverage for the no-client and no-peer states.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs
```diff
@@ -282,6 +333,7 @@ impl<B: TransactionBroadcaster + ?Sized> AssetLockManager<B> {
                 // rather than failing the resume on a tx that is actually
                 // fine. If the tx really was mined, `wait_for_proof`
                 // resolves immediately from the SPV/persisted record.
+                let timeout = self.await_broadcast_ready(out_point, timeout).await;
                 if let Err(e) = self.broadcaster.broadcast(&tx).await {
```
thepastaclaw, line 337: **🟡 Suggestion: Broadcast-status readiness behavior is not covered**

The PR adds readiness waiting to both broadcasting arms, but the regression suite only constructs a Built lock. The Broadcast arm has materially different behavior because it swallows defensive re-broadcast errors and continues into proof waiting. Add a Broadcast-status test that verifies no re-broadcast occurs while transport is unavailable, that broadcasting begins after readiness, and that proof waiting still proceeds when the defensive re-broadcast returns an error.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The readiness gate addresses the reported SPV startup race, but verification found three blocking regressions: an overflowing FFI timeout can abort the host, the synchronous waits can starve the executor needed to start SPV, and a delayed resume can overwrite newer asset-lock state. Two additional test gaps leave the production readiness delegation and the changed Broadcast-status path unprotected.
Source: reviewers codex general/security-auditor/rust-quality/ffi-engineer (backend model gpt-5.6-sol); final verifier codex (backend model gpt-5.6-sol). Orchestration-only: openclaw-agent/cliproxy/gpt-5.6-sol (not reviewer evidence).

### Review provenance

🔴 3 blocking | 🟡 2 suggestion(s)

<details>

```

- [BLOCKING] packages/rs-platform-wallet/src/spv/runtime.rs:227: An overflowing FFI timeout can panic across the C boundary
  `Instant::now() + t` panics when the resulting instant is not representable. Both public asset-lock resume FFI functions accept an unrestricted `u64` timeout and convert it directly with `Duration::from_secs`, so a value such as `UInt64.max` reaches this newly added calculation for Built or Broadcast locks. The panic occurs inside an `extern "C"` call with no unwind guard, which aborts the host instead of returning `PlatformWalletFFIResult`. Use elapsed-time subtraction or checked deadline arithmetic here, or reject unrepresentable timeout values before entering the async resume path.

- [BLOCKING] packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs:305: The readiness wait can prevent the host from ever starting SPV
  This await is reached through `asset_lock_manager_catch_up_blocking`, which calls `runtime().block_on` from synchronous Swift task-group children. The Swift call site starts four such calls concurrently and explicitly documents that each parks one worker from the typical 4–8-worker cooperative pool for up to 300 seconds. With four Built locks on a four-worker pool, the new readiness waits can occupy every generic executor worker while waiting for SPV; if the host startup task has suspended between wallet loading and `startSpv`, it cannot resume to start the transport that would release those workers. Additional locks can repeat the delay as timed-out slots are replenished. Run these synchronous FFI calls on an overcommitting or dedicated blocking queue, expose a genuinely asynchronous FFI operation, or otherwise reserve executor capacity so catch-up cannot starve SPV startup.
- [BLOCKING] packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs:305-309: Revalidate the asset-lock status after waiting for readiness
  `status` and `tx` are snapshotted before the readiness wait, which may now last minutes. Concurrent catch-up and explicit funding flows are supported, so another flow can advance, consume, or untrack the same lock while this task is parked. The stale Built arm then broadcasts the old transaction and `advance_asset_lock_status` unconditionally overwrites the current entry; it can downgrade a terminal Consumed tombstone to Broadcast or broadcast a transaction after a rejected build removed the row and released its input reservation. Re-read the tracked entry after readiness and make Built-to-Broadcast and final proof updates conditional on the expected current status, so terminal, removed, or newer transitions cannot be overwritten in a subsequent race window.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs:336-337: Broadcast-status readiness behavior is not covered
  The PR adds readiness waiting to both broadcasting arms, but the regression suite only constructs a Built lock. The Broadcast arm has materially different behavior because it swallows defensive re-broadcast errors and continues into proof waiting. Add a Broadcast-status test that verifies no re-broadcast occurs while transport is unavailable, that broadcasting begins after readiness, and that proof waiting still proceeds when the defensive re-broadcast returns an error.

- [SUGGESTION] packages/rs-platform-wallet/src/broadcaster.rs:249-250: Regression tests bypass the production SPV readiness path
  The recovery tests use `StartingUpBroadcaster`, which supplies its own readiness loop and a single synthetic flag. They prove that `AssetLockManager` invokes the trait method, but they do not exercise `SpvBroadcaster` delegation or the production `SpvRuntime` predicate requiring both a started client and at least one connected peer. Removing this override or regressing either half of the production predicate would leave all new tests green. Add production-boundary coverage through a recording `SpvChannel`, plus runtime coverage for the no-client and no-peer states.
```
</details>

packages/rs-platform-wallet/src/spv/runtime.rs
shumkov: Fixed in the reimplementation (72129fe19b): the branch was rebuilt on current v4.2-dev via a supersede-merge + one commit. Timeout arithmetic now uses checked/saturating operations under a single unified timeout contract across the FFI boundary — no overflowing Duration remains.

packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs
shumkov: Fixed in 72129fe19b: the readiness wait is fully async (no blocking waits on the executor), local finality is probed before transport readiness on both broadcast-capable arms (already-final rows complete offline, elapsed == 0 under paused time), and the wait respects the unified timeout contract.

packages/rs-platform-wallet/src/broadcaster.rs
shumkov: Addressed in 72129fe19b: the reimplementation carries deterministic regressions for readiness waits (paused-time, both arms), rejection-cleanup vs parked-resume interleaving, cancellation after an observable send and after observed local proof, and the pre-dispatch release branch — each proven red on the unfixed code.

packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs
shumkov: Addressed in 72129fe19b: Built->Broadcast status transitions are now guarded by a per-outpoint dispatch-claim lifecycle — cleanup is excluded from the moment a broadcast may have side effects until the status transition is durably committed, so no ambiguous-status window remains.

shumkov:
@coderabbitai full review

packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs
```diff
@@ -446,6 +461,52 @@ pub(crate) fn seed_observed_input_conflicts(info: &PlatformWalletInfo) {
...
+        });
+        let started = tokio::time::Instant::now();
+        if !self.broadcaster.wait_until_ready(budget).await {
+            tracing::warn!(
+                outpoint = %out_point,
+                ?budget,
+                "resume_asset_lock: broadcast transport still not ready; \
+                 attempting the broadcast anyway"
+            );
+        }
```
thepastaclaw, line 539: **🔴 Blocking: Retry recovery when readiness misses the 15-second ceiling**

`await_broadcast_ready` caps a 300-second catch-up request at 15 seconds and then deliberately broadcasts even when readiness was not reached. If SPV startup or first-peer connection takes longer than that, the production broadcaster returns its pre-dispatch `Rejected` result and both Built and Broadcast recovery paths terminate without arranging another attempt. The in-tree integration automatically starts catch-up only during wallet loading and discards this ordinary failure; although the public method can be invoked manually, no readiness or reconnect event invokes it. A cold or disrupted startup that exceeds 15 seconds therefore recreates the PR's original failure: SPV can become usable immediately afterward while the transaction remains unsent until another explicit catch-up or restart. Keep the bounded wait needed by the synchronous FFI integration, but requeue unresolved locks when SPV reports readiness or otherwise retain the recovery attempt without occupying Swift's cooperative executor.

packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs
```diff
@@ -1263,6 +1457,7 @@ impl<B: TransactionBroadcaster + ?Sized> AssetLockManager<B> {
             .advance_asset_lock_status(out_point, new_status, Some(proof.clone()))
             .await?;
```
thepastaclaw, line 1680: **🟡 Suggestion: Revalidate the asset-lock status after waiting for readiness**

`resume_asset_lock` snapshots the transaction and status before its readiness, broadcast, and proof awaits, but the final call to `advance_asset_lock_status` replaces the live status without an expected-state check. The new dispatch claim prevents rejected-build cleanup from deleting the row, but it does not coordinate with `consume_asset_lock`. If another explicit funding flow consumes the same lock while this resume is suspended, the stale resume can replace the in-memory terminal `Consumed` tombstone with `Broadcast`, `InstantSendLocked`, or `ChainLocked` and return a proof and derivation path as though the lock remained usable. `AssetLockChangeSet::merge`, the Swift persister, and Platform's one-shot validation protect durable state and prevent a second credit, so this is not a direct fund-loss path; however, the live wallet loses its terminal classification and can attempt to reuse the spent lock until restart or reconciliation. Make the Built-to-Broadcast and final proof transitions conditional on the current status, with `Consumed` remaining absorbing, and add an interleaving test that consumes the lock while readiness or proof waiting is suspended.

thepastaclaw (changes requested):
## Final validation — GLM Flash + Sol

The reimplementation fixes the timeout-overflow, production-readiness coverage, Broadcast-arm coverage, and rejected-build cleanup issues. One blocking startup limitation remains: when SPV takes longer than the hard-coded 15-second readiness window, the only automatic recovery attempt still fails before dispatch and is not retried when SPV subsequently becomes ready. The delayed recovery path also still permits a stale resume to overwrite the in-memory Consumed state, although durable-state guards and Platform's one-shot enforcement contain that issue.

Source: reviewer 1: `glm-5.3-flash` (agent: `phase1-reviewer`, role: `general`); reviewer 2: `glm-5.3-flash` (agent: `phase1-reviewer`, role: `security-auditor`); reviewer 3: `glm-5.3-flash` (agent: `phase1-reviewer`, role: `rust-quality`); reviewer 4: `glm-5.3-flash` (agent: `phase1-reviewer`, role: `ffi-engineer`); reviewer 5: `gpt-5.6-sol` (agent: `phase2-reviewer`, role: `general`); reviewer 6: `gpt-5.6-sol` (agent: `phase2-reviewer`, role: `security-auditor`); reviewer 7: `gpt-5.6-sol` (agent: `phase2-reviewer`, role: `rust-quality`); reviewer 8: `gpt-5.6-sol` (agent: `phase2-reviewer`, role: `ffi-engineer`); final verifier: `gpt-5.6-sol` (agent: `sol-verifier`, role: `final-verifier`)

### Review provenance
- Phase 1 reviewers (GLM Flash): `glm-5.3-flash` — general (completed); agent `phase1-reviewer`, `glm-5.3-flash` — security-auditor (completed); agent `phase1-reviewer`, `glm-5.3-flash` — rust-quality (completed); agent `phase1-reviewer`, `glm-5.3-flash` — ffi-engineer (completed); agent `phase1-reviewer`
- Fresh verifier (Sol): `gpt-5.6-sol` — final-verifier; agent `sol-verifier`
- Phase 2 reviewers (Sol): `gpt-5.6-sol` — general (completed); agent `phase2-reviewer`, `gpt-5.6-sol` — security-auditor (completed); agent `phase2-reviewer`, `gpt-5.6-sol` — rust-quality (completed); agent `phase2-reviewer`, `gpt-5.6-sol` — ffi-engineer (completed); agent `phase2-reviewer`

<details>

```

- [BLOCKING] packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs:489-500: Retry recovery when readiness misses the 15-second ceiling
  `await_broadcast_ready` caps a 300-second catch-up request at 15 seconds and then deliberately broadcasts even when readiness was not reached. If SPV startup or first-peer connection takes longer than that, the production broadcaster returns its pre-dispatch `Rejected` result and both Built and Broadcast recovery paths terminate without arranging another attempt. The in-tree integration automatically starts catch-up only during wallet loading and discards this ordinary failure; although the public method can be invoked manually, no readiness or reconnect event invokes it. A cold or disrupted startup that exceeds 15 seconds therefore recreates the PR's original failure: SPV can become usable immediately afterward while the transaction remains unsent until another explicit catch-up or restart. Keep the bounded wait needed by the synchronous FFI integration, but requeue unresolved locks when SPV reports readiness or otherwise retain the recovery attempt without occupying Swift's cooperative executor.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs:1456-1458: Revalidate the asset-lock status after waiting for readiness
  `resume_asset_lock` snapshots the transaction and status before its readiness, broadcast, and proof awaits, but the final call to `advance_asset_lock_status` replaces the live status without an expected-state check. The new dispatch claim prevents rejected-build cleanup from deleting the row, but it does not coordinate with `consume_asset_lock`. If another explicit funding flow consumes the same lock while this resume is suspended, the stale resume can replace the in-memory terminal `Consumed` tombstone with `Broadcast`, `InstantSendLocked`, or `ChainLocked` and return a proof and derivation path as though the lock remained usable. `AssetLockChangeSet::merge`, the Swift persister, and Platform's one-shot validation protect durable state and prevent a second credit, so this is not a direct fund-loss path; however, the live wallet loses its terminal classification and can attempt to reuse the spent lock until restart or reconciliation. Make the Built-to-Broadcast and final proof transitions conditional on the current status, with `Consumed` remaining absorbing, and add an interleaving test that consumes the lock while readiness or proof waiting is suspended.
```
</details>

packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs
QuantumExplorer: Fixed in 92403758d9. The bounded wait stays as it is (the host thread cannot afford more), but a resume whose wait expired and whose row still needs a send now defers a retry onto the manager's own runtime via `resume_when_transport_ready`: a shared handle onto the same manager waits for readiness under `DEFERRED_RESUME_TRANSPORT_WAIT` (10 min, a ceiling rather than a park), stands down if the row's status changed since the deferral, and otherwise runs one ordinary resume with the caller's timeout. Membership in `deferred_resumes` makes a second deferral while one is waiting a no-op, so foreground/reconnect catch-ups cannot stack sends; a purely blocking caller with no runtime is left exactly as before. Covered by `a_missed_transport_wait_defers_a_retry_until_the_transport_comes_up` (original transaction re-sent once the transport is up, host wait keeps its 15s ceiling) and `a_deferred_retry_stands_down_when_the_row_moved_on_meanwhile`.

packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs
QuantumExplorer: Fixed in 92403758d9. `advance_asset_lock_status` now treats `Consumed` as absorbing (returns `AssetLockAlreadyConsumed`, tombstone untouched), and the resume's Built → Broadcast advance goes through the new `advance_asset_lock_status_if` gated on the row still reading `Built`, so it neither regresses a row a concurrent resume carried to a proof-bearing status nor overwrites a tombstone. `consume_asset_lock` also releases any resume dispatch claim. Interleaving test: `a_lock_consumed_while_the_resume_awaits_the_transport_keeps_its_tombstone` consumes the lock while the resume is suspended in the readiness wait and asserts the resume returns `AssetLockAlreadyConsumed`, the row stays `Consumed` with no proof, and no retry is deferred; plus `a_consumed_tombstone_absorbs_every_later_status_advance` and `a_conditional_advance_leaves_a_row_that_moved_on_alone`.

shumkov (approved):

PR #4359, opened 2026-08-10
feat(platform-wallet): own the DashPay startup ordering instead of each client

romchornyi:

A DashPay contact's DIP-15 payment addresses are derived from its contact account, and an address the wallet is not watching when the compact-filter scan passes its funding height produces **no transaction at all**. So the order in which a wallet brings identity, contacts and contact accounts up decides whether a restored wallet has contact payment history, and a client has to be able to hold Core SPV back until those addresses exist.

That ordering is a policy decision, and it had no home in the SDK. iOS implemented it in Swift (dashpay/dashwallet-ios#961) — a hand-rolled step sequencer with its own wall-clock budget and two poll loops. Review rejected it on the grounds that `packages/swift-sdk/CLAUDE.md` states outright:

> No iteration / gap-limit walks / **policy loops in Swift**. […] If it's deciding anything — how many, which index, which path, which key, **which order** — move the decision to Rust. **If you find a decision that Rust doesn't currently let you ask for by a single call, add the helper in the Rust library first.**

This is that helper. Android would otherwise have had to reimplement the same sequence, and the identity-retry half of it had already regressed once while it lived client-side.

**The step that is easy to miss.** After a DashPay sync pass the contact accounts still do not exist: the recurring sweep runs unattended and holds no signer, so it cannot derive the receiving xpub or run the ECDH for a contact's external account — it only *enqueues* the work (`enqueue_deferred_contact_crypto`). The accounts come into being when a signer-present drain runs. A sequence that stopped after the sync pass would be correctly ordered and still start SPV with nothing extra to watch. Field evidence from the iOS build before this: 350 `Deferred DashPay account build: enqueued` lines, 0 drains, and `payments projection — 0 row(s)` on a wallet with 7 established contacts.

## What was done?

**`manager/startup.rs`** — `PlatformWalletManager::start_wallet_subsystems`:

1. **Identity** — skipped entirely when one is already known locally, so a warm launch costs no network round trip. Otherwise a scan, retried **only** on `IdentityDiscoveryIncomplete`. A scan that returned has an answer from Platform, and an empty answer is a proof of absence that rescanning cannot overturn — the distinction #4352 made expressible, and this is its first consumer.
2. **One contact-request pass**, so the deferred builds exist to drain. Log-and-continue: a prior session may have queued work this call can still finish.
3. **The drain** — `drain_pending_contact_crypto`, plus `drain_auto_accepts` when an identity signer was supplied.

Returns `Err` only for `WalletNotFound`. An unreachable Platform, a failed sync pass and an unfinished drain are all reported in `WalletStartupOutcome`, because failing loudly would trade a data gap for a wallet with no balance. Budget is a parameter, defaulting to 20s.

**`StartupTally`** holds the counters and the classification, mirroring `ScanTally` in `discovery.rs`: the tests drive the same methods production does, so a miswiring fails a test instead of shipping.

**Key material stays per-call.** The master xpriv and contact-crypto provider are borrowed for this call only, matching what `discover_from_master` and the drain already require.

**`platform-wallet-ffi`** — `platform_wallet_manager_start_wallet_subsystems` + `WalletStartupOutcomeFFI`. Uses `run_on_big_stack_thread` rather than `block_on_worker`: the manager is only reachable as a `&PlatformWalletManager` borrowed from the handle store, so the future cannot satisfy the `'static` bound. The scoped thread also supplies the 8 MB stack GroveDB proof verification needs. The master xpriv is erased with `non_secure_erase` before every return path.

**`swift-sdk`** — `PlatformWalletManager.startWalletSubsystems(wallet:budget:gapLimit:)`, plus `WalletStartupStatus.identityIsSettled` so a caller never re-derives which outcomes are worth repeating.

### Considered and rejected

Making the unattended sweep resolve its own signer, which would have removed the need for any client sequencing at all. `MnemonicResolver` is a Swift-owned vtable pinned to a single synchronous FFI call, and `drain_auto_accepts` needs a full `Signer<IdentityPublicKey>`, not just an ECDH/xpub derive. Turning either into a standing capability driven by a 15s background loop is a security-posture change, not a refactor.

## How Has This Been Tested?

```
cargo test -p platform-wallet --lib            # 625 passed
cargo test -p platform-wallet-ffi --lib        # 261 passed
cargo clippy -p platform-wallet -p platform-wallet-ffi   # clean
cargo fmt --check                              # clean
```

8 unit tests in `manager::startup::tests`, no I/O. The first pins the rule that already regressed once client-side: **a scan that came back definitively empty settles as `NoIdentity` and is never retried**, while one that never reached Platform reports `PartialNoIdentity` and is. Others cover the warm-launch path costing zero scans, queued builds surfacing as `PartialAccountsPending`, and an unreachable Platform outranking a finished drain.

**End-to-end on a testnet restore** (iOS 26.5 simulator, wallet with 7 established contacts), driven through the Swift wrapper:

```
👥 DP-READY :: ready for SPV in 8.7s scans=1 drained=14
🛰️ SPVCOORD :: started on 1
```

- 14 contact-account builds drained, against the 350-enqueue / 0-drain baseline above.
- 700 DIP-15 contact addresses registered; `reconcile_sent_payments_from_tx_history` then matched 18 txids and the payment projection went from 0 to 21 rows.
- On a freshly created wallet the sequence correctly no-ops (`no identity for this seed`) and SPV starts immediately — no added launch latency where there is nothing to prepare.

**Not covered:** the async body itself has no integration test. Driving it needs a mock SDK plus a registered wallet, and injecting a failing `Identity::fetch` to exercise the `IdentityDiscoveryIncomplete` branch — I did not confirm `SdkBuilder::new_mock()` supports that. `SeedCryptoProvider` (`contact_requests.rs`) is a ready-made fixture for the provider half if someone picks this up.

## Breaking Changes

None — purely additive. No existing caller changes; the separate `startPlatformAddressSync` / `bindShielded` / `startDashPaySync` / DPNS entry points are untouched. Those were deliberately **not** subsumed: none has a "must exist before the first filter set" ordering requirement, and folding cheap idempotent flag-flips into a budget-bounded call would widen its blast radius for no benefit.

Two notes for reviewers:

- `ensureConfigured`, `modelContainer` and `signerNetwork` on the Swift `PlatformWalletManager` go from `private` to `internal`, so the new extension reuses them rather than duplicating the gate and the signer construction.
- A host that adopts this call accepts up to `budget` of added latency before Core SPV starts. It self-skips when there is nothing to prepare, and on expiry falls through to exactly the previous behaviour.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added wallet startup support across platform wallet APIs.
  * Startup discovers identities, synchronizes contact requests, processes deferred contact-account activity, and reports pending work.
  * Added configurable startup time budgets and contact-account gap limits.
  * Added clear outcomes for ready, missing identity, partial progress, discovery failures, and pending accounts.
  * Added Swift SDK support for asynchronous startup and structured results.
  * Added wallet network and mnemonic availability status accessors.
* **Improvements**
  * Contact-account processing now supports deadlines and incremental progress.

### DIFF
--- a/packages/rs-platform-wallet-ffi/src/wallet_startup.rs
+++ b/packages/rs-platform-wallet-ffi/src/wallet_startup.rs
@@ -89,10 +89,12 @@ impl From<WalletStartupOutcome> for WalletStartupOutcomeFFI {
 /// outcome to `out_outcome`. Intended to be called once per wallet load,
 /// immediately before starting Core SPV.
 ///
-/// Only an invalid manager handle or an unknown `wallet_id` produce an error;
-/// an unreachable Platform, a failed sync pass and an unfinished drain are all
-/// reported through `out_outcome.status`, because a host must be able to start
-/// Core SPV regardless. See [`WalletStartupStatusFFI`].
+/// Errors are limited to the request itself: an invalid manager handle, an
+/// unknown `wallet_id`, or a `budget_secs` so large its deadline is not
+/// representable. Everything about the run — an unreachable Platform, a failed
+/// sync pass, an unfinished drain — is reported through `out_outcome.status`,
+/// because a host must be able to start Core SPV regardless. See
+/// [`WalletStartupStatusFFI`].
 ///
 /// # Arguments
 ///
@@ -143,17 +145,22 @@ pub unsafe extern "C" fn platform_wallet_manager_start_wallet_subsystems(
         );
     };
 
+    // Key material is only needed for a scan, and a wallet whose identity is
+    // already on file will not run one — so a warm launch must not pay for a
+    // Keychain round trip, nor be failed by one that goes wrong. Ask first.
+    let needs_scan_key = !PLATFORM_WALLET_MANAGER_STORAGE
+        .with_item(manager_handle, |m| m.has_local_identity_blocking(&wid))
+        .unwrap_or(false);
+
     // Resolve the master xpriv once, up front. The helper holds the mnemonic
     // and seed in `Zeroizing` buffers and scrubs them before returning; the
     // master itself has no `Drop`, so it is erased explicitly below.
     //
-    // A failure here is NOT returned as an FFI error. This call documents that
-    // only handle and wallet-id problems throw, and the resolver is needed only
-    // when a scan actually runs — a warm launch whose identity is already known
-    // must not be failed by a Keychain hiccup it never needed. If a scan does
-    // turn out to be required, it fails without key material and surfaces as
-    // `DiscoveryFailed`, which is the structured outcome for exactly this.
-    let mut master = if mnemonic_resolver_handle.is_null() {
+    // A failure here is NOT returned as an FFI error: this call documents that
+    // only handle and wallet-id problems throw. Without key material the scan
+    // fails and surfaces as `DiscoveryFailed`, which is the structured outcome
+    // for exactly this.
+    let mut master = if mnemonic_resolver_handle.is_null() || !needs_scan_key {
         None
     } else {
         match resolve_master_from_resolver(mnemonic_resolver_handle, &wid, network) {
--- a/packages/rs-platform-wallet-ffi/src/wallet_startup.rs
+++ b/packages/rs-platform-wallet-ffi/src/wallet_startup.rs
@@ -146,12 +146,26 @@ pub unsafe extern "C" fn platform_wallet_manager_start_wallet_subsystems(
     // Resolve the master xpriv once, up front. The helper holds the mnemonic
     // and seed in `Zeroizing` buffers and scrubs them before returning; the
     // master itself has no `Drop`, so it is erased explicitly below.
+    //
+    // A failure here is NOT returned as an FFI error. This call documents that
+    // only handle and wallet-id problems throw, and the resolver is needed only
+    // when a scan actually runs — a warm launch whose identity is already known
+    // must not be failed by a Keychain hiccup it never needed. If a scan does
+    // turn out to be required, it fails without key material and surfaces as
+    // `DiscoveryFailed`, which is the structured outcome for exactly this.
     let mut master = if mnemonic_resolver_handle.is_null() {
         None
     } else {
         match resolve_master_from_resolver(mnemonic_resolver_handle, &wid, network) {
             Ok(master) => Some(master),
-            Err(e) => return e,
+            Err(_) => {
+                tracing::warn!(
+                    wallet_id = %hex::encode(wid),
+                    "startup: could not resolve the wallet mnemonic; continuing without \
+                     scan key material"
+                );
+                None
+            }
         }
     };
 
--- a/packages/rs-platform-wallet-ffi/src/wallet_startup.rs
+++ b/packages/rs-platform-wallet-ffi/src/wallet_startup.rs
@@ -28,6 +28,7 @@ pub enum WalletStartupStatusFFI {
     NoIdentity = 1,
     PartialNoIdentity = 2,
     PartialAccountsPending = 3,
+    DiscoveryFailed = 4,
 }
 
 impl From<WalletStartupStatus> for WalletStartupStatusFFI {
@@ -37,6 +38,7 @@ impl From<WalletStartupStatus> for WalletStartupStatusFFI {
             WalletStartupStatus::NoIdentity => Self::NoIdentity,
             WalletStartupStatus::PartialNoIdentity => Self::PartialNoIdentity,
             WalletStartupStatus::PartialAccountsPending => Self::PartialAccountsPending,
+            WalletStartupStatus::DiscoveryFailed => Self::DiscoveryFailed,
         }
     }
 }
--- a/packages/rs-platform-wallet-ffi/src/wallet_startup.rs
+++ b/packages/rs-platform-wallet-ffi/src/wallet_startup.rs
@@ -7,7 +7,7 @@
 //! [`platform_wallet::manager::startup`]; this is the marshalling shell.
 
 use platform_wallet::manager::startup::{
-    WalletStartupOptions, WalletStartupOutcome, WalletStartupStatus,
+    ScanKeyError, WalletStartupOptions, WalletStartupOutcome, WalletStartupStatus,
 };
 use rs_sdk_ffi::{MnemonicResolverHandle, SignerHandle, VTableSigner};
 use std::time::Duration;
@@ -16,7 +16,9 @@ use crate::check_ptr;
 use crate::dashpay::resolver_contact_crypto_provider;
 use crate::error::{PlatformWalletFFIResult, PlatformWalletFFIResultCode};
 use crate::handle::{Handle, PLATFORM_WALLET_MANAGER_STORAGE};
-use crate::identity_keys_from_mnemonic::resolve_master_from_resolver;
+use crate::identity_keys_from_mnemonic::{
+    resolve_master_from_resolver_classified, ResolveFailureKind,
+};
 use crate::runtime::run_on_big_stack_thread;
 
 /// Discriminants for [`WalletStartupStatus`] across the boundary.
@@ -153,25 +155,33 @@ pub unsafe extern "C" fn platform_wallet_manager_start_wallet_subsystems(
     // it is valid for the duration of this call, which is the only time the
     // closure can run. A null resolver means the host is telling us the wallet
     // holds resident keys (watch-only / in-process derive), so no resolver is
-    // supplied at all — as opposed to one that fails, which the library treats
-    // as a retryable "come back later".
+    // supplied at all — distinct from one that is supplied and fails, which
+    // the library classifies as retryable or terminal per `ScanKeyError`.
     let resolver_addr = mnemonic_resolver_handle as usize;
     let resolve_scan_key = move || {
-        resolve_master_from_resolver(resolver_addr as *mut MnemonicResolverHandle, &wid, network)
-            .map_err(|e| {
-                tracing::warn!(
-                    wallet_id = %hex::encode(wid),
-                    code = ?e.code,
-                    "startup: could not resolve the wallet mnemonic for the identity scan"
-                );
-                // Carried as an error the library can classify, not as an FFI
-                // throw: the documented contract is that only handle and
-                // wallet-id problems throw, and a locked device is neither.
-                platform_wallet::error::PlatformWalletError::KeyDerivation(format!(
-                    "mnemonic resolver failed ({:?})",
-                    e.code
-                ))
-            })
+        resolve_master_from_resolver_classified(
+            resolver_addr as *mut MnemonicResolverHandle,
+            &wid,
+            network,
+        )
+        .map_err(|failure| {
+            tracing::warn!(
+                wallet_id = %hex::encode(wid),
+                code = ?failure.result.code,
+                kind = ?failure.kind,
+                "startup: could not resolve the wallet mnemonic for the identity scan"
+            );
+            // Carried as an outcome the library classifies, not as an FFI
+            // throw: the documented contract is that only handle and wallet-id
+            // problems throw, and a locked device is neither. The permanence
+            // travels with it — a phrase that is not BIP-39 must not be
+            // reported as something the next launch might fix.
+            let detail = format!("mnemonic resolver failed ({:?})", failure.result.code);
+            match failure.kind {
+                ResolveFailureKind::Unavailable => ScanKeyError::Unavailable(detail),
+                ResolveFailureKind::Permanent => ScanKeyError::Invalid(detail),
+            }
+        })
     };
     let scan_key = (!mnemonic_resolver_handle.is_null())
         .then_some(&resolve_scan_key as &(dyn Fn() -> Result<_, _> + Send + Sync));
--- a/packages/rs-platform-wallet-ffi/src/wallet_startup.rs
+++ b/packages/rs-platform-wallet-ffi/src/wallet_startup.rs
@@ -1,0 +1,213 @@
+//! FFI for the ordered wallet bring-up.
+//!
+//! One call that runs identity discovery → contact-request sync → the
+//! signer-present contact-crypto drain, so the host can start Core SPV knowing
+//! the DIP-15 contact addresses exist and will be in the first filter set. The
+//! ordering policy itself lives in
+//! [`platform_wallet::manager::startup`]; this is the marshalling shell.
+
+use platform_wallet::manager::startup::{
+    WalletStartupOptions, WalletStartupOutcome, WalletStartupStatus,
+};
+use rs_sdk_ffi::{MnemonicResolverHandle, SignerHandle, VTableSigner};
+use std::time::Duration;
+
+use crate::check_ptr;
+use crate::dashpay::resolver_contact_crypto_provider;
+use crate::error::{PlatformWalletFFIResult, PlatformWalletFFIResultCode};
+use crate::handle::{Handle, PLATFORM_WALLET_MANAGER_STORAGE};
+use crate::identity_keys_from_mnemonic::resolve_master_from_resolver;
+use crate::runtime::run_on_big_stack_thread;
+
+/// Discriminants for [`WalletStartupStatus`] across the boundary.
+///
+/// Values are part of the ABI — append, never renumber.
+#[repr(u8)]
+pub enum WalletStartupStatusFFI {
+    Ready = 0,
+    NoIdentity = 1,
+    PartialNoIdentity = 2,
+    PartialAccountsPending = 3,
+}
+
+impl From<WalletStartupStatus> for WalletStartupStatusFFI {
+    fn from(status: WalletStartupStatus) -> Self {
+        match status {
+            WalletStartupStatus::Ready => Self::Ready,
+            WalletStartupStatus::NoIdentity => Self::NoIdentity,
+            WalletStartupStatus::PartialNoIdentity => Self::PartialNoIdentity,
+            WalletStartupStatus::PartialAccountsPending => Self::PartialAccountsPending,
+        }
+    }
+}
+
+/// Flat result of [`platform_wallet_manager_start_wallet_subsystems`].
+#[repr(C)]
+pub struct WalletStartupOutcomeFFI {
+    /// A [`WalletStartupStatusFFI`] discriminant.
+    pub status: u8,
+    /// Whether `identity_id` carries a value.
+    pub has_identity_id: bool,
+    /// The wallet's identity, valid only when `has_identity_id`.
+    pub identity_id: [u8; 32],
+    /// Discovery scans performed; `0` when a local identity was already known.
+    pub discovery_attempts: u32,
+    /// Whether the inline contact-request pass ran.
+    pub dashpay_sync_ran: bool,
+    /// Contact-crypto entries the drain completed.
+    pub contact_accounts_drained: u32,
+    /// Contact-account builds still queued on return.
+    pub contact_accounts_pending: u32,
+    /// Wall-clock duration of the whole sequence.
+    pub elapsed_ms: u64,
+}
+
+impl From<WalletStartupOutcome> for WalletStartupOutcomeFFI {
+    fn from(outcome: WalletStartupOutcome) -> Self {
+        let (has_identity_id, identity_id) = match outcome.identity_id {
+            Some(id) => (true, id.to_buffer()),
+            None => (false, [0u8; 32]),
+        };
+        Self {
+            status: WalletStartupStatusFFI::from(outcome.status) as u8,
+            has_identity_id,
+            identity_id,
+            discovery_attempts: outcome.discovery_attempts,
+            dashpay_sync_ran: outcome.dashpay_sync_ran,
+            contact_accounts_drained: outcome.contact_accounts_drained as u32,
+            contact_accounts_pending: outcome.contact_accounts_pending as u32,
+            elapsed_ms: outcome.elapsed.as_millis() as u64,
+        }
+    }
+}
+
+/// Bring one wallet's DashPay state up in dependency order.
+///
+/// Blocks until the sequence finishes or its budget expires, then writes the
+/// outcome to `out_outcome`. Intended to be called once per wallet load,
+/// immediately before starting Core SPV.
+///
+/// Only an invalid manager handle or an unknown `wallet_id` produce an error;
+/// an unreachable Platform, a failed sync pass and an unfinished drain are all
+/// reported through `out_outcome.status`, because a host must be able to start
+/// Core SPV regardless. See [`WalletStartupStatusFFI`].
+///
+/// # Arguments
+///
+/// * `wallet_id` — 32 bytes.
+/// * `mnemonic_resolver_handle` — nullable. Required for a Keychain-backed
+///   external-signable wallet, whose seed does not live in the wallet manager;
+///   null means the wallet holds resident keys and derives in-process.
+/// * `identity_signer_handle` — nullable; null skips the DIP-15 auto-accept
+///   pass, matching `platform_wallet_drain_pending_contact_crypto`.
+/// * `budget_secs` — `0` means the crate default (20s). Deliberately not
+///   "unbounded": this call gates Core SPV, so it must always terminate.
+/// * `gap_limit` — `0` means the crate default.
+///
+/// # Safety
+///
+/// `wallet_id` must point to 32 readable bytes and `out_outcome` to a writable
+/// [`WalletStartupOutcomeFFI`]. Both handles, when non-null, must be valid for
+/// the duration of this call — the callee borrows and never retains them.
+#[no_mangle]
+pub unsafe extern "C" fn platform_wallet_manager_start_wallet_subsystems(
+    manager_handle: Handle,
+    wallet_id: *const u8,
+    mnemonic_resolver_handle: *mut MnemonicResolverHandle,
+    identity_signer_handle: *mut SignerHandle,
+    budget_secs: u64,
+    gap_limit: u32,
+    out_outcome: *mut WalletStartupOutcomeFFI,
+) -> PlatformWalletFFIResult {
+    check_ptr!(wallet_id);
+    check_ptr!(out_outcome);
+
+    let wid: [u8; 32] = std::ptr::read(wallet_id as *const [u8; 32]);
+
+    // The network is needed before any key material can be built, and the
+    // caller gave us a manager handle rather than a wallet handle.
+    let Some(network_opt) = PLATFORM_WALLET_MANAGER_STORAGE
+        .with_item(manager_handle, |m| m.wallet_network_blocking(&wid))
+    else {
+        return PlatformWalletFFIResult::err(
+            PlatformWalletFFIResultCode::ErrorInvalidHandle,
+            "Manager handle invalid".to_string(),
+        );
+    };
+    let Some(network) = network_opt else {
+        return PlatformWalletFFIResult::err(
+            PlatformWalletFFIResultCode::NotFound,
+            "Wallet not found".to_string(),
+        );
+    };
+
+    // Resolve the master xpriv once, up front. The helper holds the mnemonic
+    // and seed in `Zeroizing` buffers and scrubs them before returning; the
+    // master itself has no `Drop`, so it is erased explicitly below.
+    let mut master = if mnemonic_resolver_handle.is_null() {
+        None
+    } else {
+        match resolve_master_from_resolver(mnemonic_resolver_handle, &wid, network) {
+            Ok(master) => Some(master),
+            Err(e) => return e,
+        }
+    };
+
+    let signer_addr = if identity_signer_handle.is_null() {
+        0usize
+    } else {
+        identity_signer_handle as usize
+    };
+    let provider = resolver_contact_crypto_provider(mnemonic_resolver_handle, wid, network);
+
+    let opts = WalletStartupOptions {
+        budget: if budget_secs == 0 {
+            WalletStartupOptions::default().budget
+        } else {
+            Duration::from_secs(budget_secs)
+        },
+        gap_limit: (gap_limit > 0).then_some(gap_limit),
+    };
+
+    // `run_on_big_stack_thread`, not `block_on_worker`: the manager is only
+    // reachable as a `&PlatformWalletManager` borrowed from the handle store,
+    // so the future cannot satisfy `block_on_worker`'s `'static` bound. The
+    // scoped thread also supplies the 8 MB stack the GroveDB proof
+    // verification inside discovery needs.
+    let result = run_on_big_stack_thread(|| {
+        PLATFORM_WALLET_MANAGER_STORAGE.with_item(manager_handle, |manager| {
+            crate::runtime::runtime().block_on(async {
+                let signer =
+                    (signer_addr != 0).then(|| unsafe { &*(signer_addr as *const VTableSigner) });
+                manager
+                    .start_wallet_subsystems(&wid, master.as_ref(), &provider, signer, opts)
+                    .await
+            })
+        })
+    });
+
+    // Erase the master before any early return below.
+    if let Some(master) = master.as_mut() {
+        master.private_key.non_secure_erase();
+    }
+
+    let outcome = match result {
+        Ok(Some(Ok(outcome))) => outcome,
+        Ok(Some(Err(e))) => return PlatformWalletFFIResult::from(e),
+        Ok(None) => {
+            return PlatformWalletFFIResult::err(
+                PlatformWalletFFIResultCode::ErrorInvalidHandle,
+                "Manager handle invalid".to_string(),
+            )
+        }
+        Err(e) => {
+            return PlatformWalletFFIResult::err(
+                PlatformWalletFFIResultCode::ErrorWalletOperation,
+                format!("failed to spawn the startup thread: {e}"),
+            )
+        }
+    };
+
+    *out_outcome = WalletStartupOutcomeFFI::from(outcome);
+    PlatformWalletFFIResult::ok()
+}
--- a/packages/rs-platform-wallet-ffi/src/wallet_startup.rs
+++ b/packages/rs-platform-wallet-ffi/src/wallet_startup.rs
@@ -7,7 +7,7 @@
 //! [`platform_wallet::manager::startup`]; this is the marshalling shell.
 
 use platform_wallet::manager::startup::{
-    WalletStartupOptions, WalletStartupOutcome, WalletStartupStatus,
+    ScanKey, WalletStartupOptions, WalletStartupOutcome, WalletStartupStatus,
 };
 use rs_sdk_ffi::{MnemonicResolverHandle, SignerHandle, VTableSigner};
 use std::time::Duration;
@@ -156,24 +156,33 @@ pub unsafe extern "C" fn platform_wallet_manager_start_wallet_subsystems(
     // and seed in `Zeroizing` buffers and scrubs them before returning; the
     // master itself has no `Drop`, so it is erased explicitly below.
     //
-    // A failure here is NOT returned as an FFI error: this call documents that
-    // only handle and wallet-id problems throw. Without key material the scan
-    // fails and surfaces as `DiscoveryFailed`, which is the structured outcome
-    // for exactly this.
-    let mut master = if mnemonic_resolver_handle.is_null() || !needs_scan_key {
+    // A failure here is NOT returned as an FFI error — this call documents that
+    // only handle and wallet-id problems throw. It becomes `ScanKey::Unavailable`
+    // instead, so a locked device defers the scan rather than failing it
+    // terminally.
+    //
+    // `None` = no key needed (warm launch) or none exists (resident-key
+    // wallet); `Some(Err(()))` = a scan needs one and we could not get it.
+    // Both no-key cases collapse to `None`: nothing to scan for, or a host that
+    // deliberately supplied no resolver (watch-only / resident-key wallet, where
+    // discovery derives in-process). Neither is a failure to obtain a key.
+    let mut master: Option<Result<_, ()>> = if !needs_scan_key || mnemonic_resolver_handle.is_null()
+    {
         None
     } else {
-        match resolve_master_from_resolver(mnemonic_resolver_handle, &wid, network) {
-            Ok(master) => Some(master),
-            Err(_) => {
-                tracing::warn!(
-                    wallet_id = %hex::encode(wid),
-                    "startup: could not resolve the wallet mnemonic; continuing without \
-                     scan key material"
-                );
-                None
-            }
-        }
+        Some(
+            match resolve_master_from_resolver(mnemonic_resolver_handle, &wid, network) {
+                Ok(master) => Ok(master),
+                Err(_) => {
+                    tracing::warn!(
+                        wallet_id = %hex::encode(wid),
+                        "startup: could not resolve the wallet mnemonic; deferring the \
+                         identity scan to a later start"
+                    );
+                    Err(())
+                }
+            },
+        )
     };
 
     let signer_addr = if identity_signer_handle.is_null() {
@@ -202,20 +211,26 @@ pub unsafe extern "C" fn platform_wallet_manager_start_wallet_subsystems(
     // so the future cannot satisfy `block_on_worker`'s `'static` bound. The
     // scoped thread also supplies the 8 MB stack the GroveDB proof
     // verification inside discovery needs.
+    let scan_key = match master.as_ref() {
+        Some(Ok(master)) => ScanKey::Master(master),
+        Some(Err(())) => ScanKey::Unavailable,
+        None => ScanKey::Resident,
+    };
+
     let result = run_on_big_stack_thread(|| {
         PLATFORM_WALLET_MANAGER_STORAGE.with_item(manager_handle, |manager| {
             crate::runtime::runtime().block_on(async {
                 let signer =
                     (signer_addr != 0).then(|| unsafe { &*(signer_addr as *const VTableSigner) });
                 manager
-                    .start_wallet_subsystems(&wid, master.as_ref(), provider.as_ref(), signer, opts)
+                    .start_wallet_subsystems(&wid, scan_key, provider.as_ref(), signer, opts)
                     .await
             })
         })
     });
 
     // Erase the master before any early return below.
-    if let Some(master) = master.as_mut() {
+    if let Some(Ok(master)) = master.as_mut() {
         master.private_key.non_secure_erase();
     }
 
--- a/packages/rs-platform-wallet-ffi/src/wallet_startup.rs
+++ b/packages/rs-platform-wallet-ffi/src/wallet_startup.rs
@@ -7,7 +7,7 @@
 //! [`platform_wallet::manager::startup`]; this is the marshalling shell.
 
 use platform_wallet::manager::startup::{
-    ScanKey, WalletStartupOptions, WalletStartupOutcome, WalletStartupStatus,
+    WalletStartupOptions, WalletStartupOutcome, WalletStartupStatus,
 };
 use rs_sdk_ffi::{MnemonicResolverHandle, SignerHandle, VTableSigner};
 use std::time::Duration;
@@ -145,45 +145,36 @@ pub unsafe extern "C" fn platform_wallet_manager_start_wallet_subsystems(
         );
     };
 
-    // Key material is only needed for a scan, and a wallet whose identity is
-    // already on file will not run one — so a warm launch must not pay for a
-    // Keychain round trip, nor be failed by one that goes wrong. Ask first.
-    let needs_scan_key = !PLATFORM_WALLET_MANAGER_STORAGE
-        .with_item(manager_handle, |m| m.has_local_identity_blocking(&wid))
-        .unwrap_or(false);
-
-    // Resolve the master xpriv once, up front. The helper holds the mnemonic
-    // and seed in `Zeroizing` buffers and scrubs them before returning; the
-    // master itself has no `Drop`, so it is erased explicitly below.
+    // The master xpriv is resolved lazily, by the library, on the branch that
+    // actually scans — this crate no longer predicts when that is. Erasure is
+    // the library's too: it resolved the key, so it clears it.
     //
-    // A failure here is NOT returned as an FFI error — this call documents that
-    // only handle and wallet-id problems throw. It becomes `ScanKey::Unavailable`
-    // instead, so a locked device defers the scan rather than failing it
-    // terminally.
-    //
-    // `None` = no key needed (warm launch) or none exists (resident-key
-    // wallet); `Some(Err(()))` = a scan needs one and we could not get it.
-    // Both no-key cases collapse to `None`: nothing to scan for, or a host that
-    // deliberately supplied no resolver (watch-only / resident-key wallet, where
-    // discovery derives in-process). Neither is a failure to obtain a key.
-    let mut master: Option<Result<_, ()>> = if !needs_scan_key || mnemonic_resolver_handle.is_null()
-    {
-        None
-    } else {
-        Some(
-            match resolve_master_from_resolver(mnemonic_resolver_handle, &wid, network) {
-                Ok(master) => Ok(master),
-                Err(_) => {
-                    tracing::warn!(
-                        wallet_id = %hex::encode(wid),
-                        "startup: could not resolve the wallet mnemonic; deferring the \
-                         identity scan to a later start"
-                    );
-                    Err(())
-                }
-            },
-        )
+    // The handle is carried as a `usize` so the closure stays `Send + Sync`;
+    // it is valid for the duration of this call, which is the only time the
+    // closure can run. A null resolver means the host is telling us the wallet
+    // holds resident keys (watch-only / in-process derive), so no resolver is
+    // supplied at all — as opposed to one that fails, which the library treats
+    // as a retryable "come back later".
+    let resolver_addr = mnemonic_resolver_handle as usize;
+    let resolve_scan_key = move || {
+        resolve_master_from_resolver(resolver_addr as *mut MnemonicResolverHandle, &wid, network)
+            .map_err(|e| {
+                tracing::warn!(
+                    wallet_id = %hex::encode(wid),
+                    code = ?e.code,
+                    "startup: could not resolve the wallet mnemonic for the identity scan"
+                );
+                // Carried as an error the library can classify, not as an FFI
+                // throw: the documented contract is that only handle and
+                // wallet-id problems throw, and a locked device is neither.
+                platform_wallet::error::PlatformWalletError::KeyDerivation(format!(
+                    "mnemonic resolver failed ({:?})",
+                    e.code
+                ))
+            })
     };
+    let scan_key = (!mnemonic_resolver_handle.is_null())
+        .then_some(&resolve_scan_key as &(dyn Fn() -> Result<_, _> + Send + Sync));
 
     let signer_addr = if identity_signer_handle.is_null() {
         0usize
@@ -211,12 +202,6 @@ pub unsafe extern "C" fn platform_wallet_manager_start_wallet_subsystems(
     // so the future cannot satisfy `block_on_worker`'s `'static` bound. The
     // scoped thread also supplies the 8 MB stack the GroveDB proof
     // verification inside discovery needs.
-    let scan_key = match master.as_ref() {
-        Some(Ok(master)) => ScanKey::Master(master),
-        Some(Err(())) => ScanKey::Unavailable,
-        None => ScanKey::Resident,
-    };
-
     let result = run_on_big_stack_thread(|| {
         PLATFORM_WALLET_MANAGER_STORAGE.with_item(manager_handle, |manager| {
             crate::runtime::runtime().block_on(async {
@@ -228,11 +213,6 @@ pub unsafe extern "C" fn platform_wallet_manager_start_wallet_subsystems(
             })
         })
     });
-
-    // Erase the master before any early return below.
-    if let Some(Ok(master)) = master.as_mut() {
-        master.private_key.non_secure_erase();
-    }
 
     let outcome = match result {
         Ok(Some(Ok(outcome))) => outcome,
--- a/packages/rs-platform-wallet-ffi/src/wallet_startup.rs
+++ b/packages/rs-platform-wallet-ffi/src/wallet_startup.rs
@@ -158,7 +158,12 @@ pub unsafe extern "C" fn platform_wallet_manager_start_wallet_subsystems(
     } else {
         identity_signer_handle as usize
     };
-    let provider = resolver_contact_crypto_provider(mnemonic_resolver_handle, wid, network);
+    // The provider is resolver-backed, so without a resolver every crypto
+    // operation would fail with `NullHandle` and the drain would report zero
+    // while leaving the queue untouched. Pass `None` instead: the sequence then
+    // skips the drain and reports the real pending count.
+    let provider = (!mnemonic_resolver_handle.is_null())
+        .then(|| resolver_contact_crypto_provider(mnemonic_resolver_handle, wid, network));
 
     let opts = WalletStartupOptions {
         budget: if budget_secs == 0 {
@@ -180,7 +185,7 @@ pub unsafe extern "C" fn platform_wallet_manager_start_wallet_subsystems(
                 let signer =
                     (signer_addr != 0).then(|| unsafe { &*(signer_addr as *const VTableSigner) });
                 manager
-                    .start_wallet_subsystems(&wid, master.as_ref(), &provider, signer, opts)
+                    .start_wallet_subsystems(&wid, master.as_ref(), provider.as_ref(), signer, opts)
                     .await
             })
         })
--- a/packages/rs-platform-wallet/src/manager/startup.rs
+++ b/packages/rs-platform-wallet/src/manager/startup.rs
@@ -274,6 +274,13 @@ impl StartupTally {
         if self.contact_accounts_pending > 0 {
             return WalletStartupStatus::PartialAccountsPending;
         }
+        // An empty queue only means "nothing left to build" if a contact pass
+        // actually completed. Without one there may be undiscovered contact
+        // requests whose account builds were never enqueued, and calling that
+        // `Ready` would promise addresses this call never prepared.
+        if !self.dashpay_sync_ran {
+            return WalletStartupStatus::PartialAccountsPending;
+        }
         WalletStartupStatus::Ready
     }
 
@@ -327,7 +334,16 @@ impl<P: PlatformWalletPersistence + Send + Sync + 'static> PlatformWalletManager
         S: Signer<IdentityPublicKey> + Send + Sync,
     {
         let started = Instant::now();
-        let deadline = started + opts.budget;
+        // `Instant + Duration` panics when the sum is not representable, and
+        // this runs under an `extern "C"` caller whose thread wrapper re-raises
+        // a panic as an abort — so an absurd budget would take the host process
+        // down rather than return an error.
+        let deadline = started.checked_add(opts.budget).ok_or_else(|| {
+            PlatformWalletError::InvalidParameter(format!(
+                "startup budget {:?} exceeds the supported duration range",
+                opts.budget
+            ))
+        })?;
         let mut tally = StartupTally::default();
 
         let wallet = self
@@ -342,6 +358,7 @@ impl<P: PlatformWalletPersistence + Send + Sync + 'static> PlatformWalletManager
             tally.record_local_identity(known);
         } else {
             self.discover_identity_with_backoff(
+                wallet_id,
                 identity_wallet,
                 master,
                 opts.gap_limit,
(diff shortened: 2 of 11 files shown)


thepastaclaw:
✅ Final review complete — no blockers (commit 58a43aaf)

packages/rs-platform-wallet/src/manager/startup.rs
```diff
@@ -0,0 +1,550 @@
...
+            .await;
+        tally.record_drain(drained + accepted, pending);
+
+        if pending > 0 {
+            tracing::warn!(
+                wallet_id = %hex::encode(wallet_id),
+                pending,
+                "startup: contact-account builds still queued; the DIP-15 rescan will backfill"
+            );
+        }
```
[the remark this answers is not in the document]
romchornyi: Fixed in d402a18 — this was the most serious of the three, and it is a gap I had flagged in the Swift version and then reproduced here.

Every network step now runs through a `within_budget` helper: the contact-request pass and both drains. Abandoning one is safe by construction — entries a drain did not complete stay queued and the next signer-present action retries them — and the caller needs Core SPV to start far more than it needs any single step to finish.

`pending_contact_crypto_count` is deliberately left unbounded: it is a local queue-length read with no I/O, and its answer matters most precisely when the steps above ran out of time.

Also corrected the `PartialAccountsPending` doc as you noted — a non-zero pending count no longer implies budget expiry, since the drain may have failed on some entries or been skipped for want of a provider.

Three tests on a paused clock (`start_paused = true`) cover the abandon, the fits-in-budget and the already-expired cases.

packages/rs-platform-wallet-ffi/src/wallet_startup.rs
```diff
@@ -0,0 +1,213 @@
...
+            Err(e) => return e,
+        }
+    };
+
+    let signer_addr = if identity_signer_handle.is_null() {
+        0usize
+    } else {
+        identity_signer_handle as usize
+    };
+    let provider = resolver_contact_crypto_provider(mnemonic_resolver_handle, wid, network);
```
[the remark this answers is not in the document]
romchornyi: Fixed in d402a18. You are right: the provider is resolver-backed, so with a null resolver every operation returns `NullHandle` — the drain reported zero while the queue was untouched, which reads as "nothing to do" rather than "could not try".

`contact_crypto` is now `Option<&C>`. The FFI passes `None` when `mnemonic_resolver_handle` is null, and the sequence skips the drain, logs why, and reports the real `contact_accounts_pending` — so a resident-key wallet gets an honest `PartialAccountsPending` instead of a misleading `Ready`.

Worth noting for anyone reading later: a resident-key wallet reaching this path at all is unusual, since dashwallet's wallets are Keychain-backed external-signable. But the FFI documents the null resolver as meaning "resident keys", so the combination is reachable and should not silently do nothing.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerStartup.swift
```diff
@@ -0,0 +1,158 @@
...
+        // derives in-process and the resolver is never consulted.
+        let coreSigner: MnemonicResolver? =
+            storage.hasMnemonic(for: walletId) ? MnemonicResolver(storage: storage) : nil
+        // Identity signer for the DIP-15 auto-accept pass. Nil without a
+        // SwiftData container → the drain runs provider-only, matching
+        // `unlockWalletFromKeychain`.
+        let identitySigner: KeychainSigner? = self.modelContainer.map {
+            KeychainSigner(modelContainer: $0, network: self.signerNetwork ?? .testnet)
+        }
+        let budgetSecs = UInt64((budget ?? 0).rounded())
```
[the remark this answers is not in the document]
romchornyi: Fixed in d402a18 — both halves.

Non-finite and negative budgets are now rejected with `PlatformWalletError.invalidParameter` rather than trapping in the `UInt64` initializer. And you are right about the rounding: `0` means "SDK default" across the boundary, so a positive budget under half a second silently became the longest possible budget instead of the shortest. Clamped to one second.

thepastaclaw:
One CodeRabbit review-body finding still appears valid on `d402a18` (it was not one of the three inline threads): the non-network discovery `Err(e)` arm still calls `record_discovery_gave_up()`, which sets `discovery_unreachable`. That maps to `PartialNoIdentity`, and `identity_is_settled()` returns `false`, so clients are told to retry even though the branch explicitly says the local wallet/persistence failure will not fix itself on another attempt.

Please distinguish this terminal local-discovery failure from an unreachable Platform (or explicitly change/document the status contract) so clients do not schedule a futile rescan.

romchornyi:
Good catch — fixed in 235620b, and you are right that it contradicted its own comment.

The non-network `Err` arm called `record_discovery_gave_up()`, which maps to `PartialNoIdentity` and therefore told clients to scan again, while the comment right above it said a wallet/persistence fault would not fix itself. New terminal `DiscoveryFailed` status for that case; it outranks unreachability in the classification, since both leave the identity question open but only this one settles whether asking again is worth anything.

That case did not fit the single predicate, which is what let the bug through, so there are now two — deliberately **not** inverses:

- `discovery_worth_retrying()` — true only for `PartialNoIdentity`. Answers *should I scan again?*
- `identity_is_settled()` — false for `PartialNoIdentity` **and** `DiscoveryFailed`. Answers *do we know?*

`DiscoveryFailed` needs both readings at once: unanswered, yet futile to retry. Two tests pin the asymmetry, including one that walks every terminal status and asserts none of them asks for a rescan.

FFI discriminant appended as 4, existing values untouched. The iOS caller handles it at error level (dashpay/dashwallet-ios#961).

packages/rs-platform-wallet/src/manager/startup.rs
```diff
@@ -0,0 +1,710 @@
...
+        if !tally.has_identity() {
+            return Ok(tally.into_outcome(started.elapsed()));
+        }
+
+        // 2. One contact-request pass, so the deferred builds exist to drain.
+        //    Log-and-continue: a prior session may already have queued work
+        //    that this call can still complete.
+        match within_budget(deadline, identity_wallet.dashpay().sync_contact_requests()).await {
+            Some(Ok(requests)) => {
+                tally.record_sync_ran();
```
thepastaclaw, line 469: **🟡 Suggestion: A failed contact sync can still be classified as Ready**

The tally records a successful sync whenever `sync_contact_requests` returns `Ok`, but that method catches received and sent fetch failures per identity and ultimately returns `Ok(all_requests)` even if Platform was unreachable for every identity. Explicit errors and timeouts leave `dashpay_sync_ran` false, but `StartupTally::status()` does not inspect that field either. If the existing deferred queue is empty, all three cases become `Ready`, even though no complete contact-document pass occurred and undiscovered requests may require account builds. This contradicts the `Ready` contract and can start SPV without the contact addresses this API exists to prepare. Return a completion summary from the sync operation and add a partial classification whenever any identity's contact fetch did not complete.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/manager/startup.rs
```diff
@@ -0,0 +1,710 @@
...
+        let opts = IdentityDiscoveryOptions {
+            start_index: Some(0),
+            gap_limit: gap_limit.unwrap_or(IdentityDiscoveryOptions::default().gap_limit),
+        };
+
+        for (attempt, backoff) in DISCOVERY_BACKOFF.iter().map(Some).chain([None]).enumerate() {
+            let result = match master {
+                Some(master) => identity_wallet.discover_from_master(opts, master).await,
+                None => identity_wallet.discover(opts).await,
+            };
```
thepastaclaw, line 635: **🟡 Suggestion: Identity discovery is not bounded by the startup deadline**

Each discovery attempt is awaited directly instead of through `within_budget`. One attempt can perform up to `gap_limit` sequential Platform fetches and DPNS lookups, so it can continue well beyond `opts.budget`; after a backoff consumes all remaining time, the loop can also start another unbounded attempt with an already-expired deadline. This breaks the documented whole-sequence ceiling and can keep the synchronous FFI caller from starting Core SPV during a Platform outage. Bound the entire discovery future by the remaining deadline and stop immediately on timeout. Because discovery persists sightings incrementally, re-check wallet-local identities before reporting `PartialNoIdentity` after cancellation.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/manager/startup.rs
```diff
@@ -0,0 +1,710 @@
...
+        contact_crypto: Option<&C>,
+        identity_signer: Option<&S>,
+        opts: WalletStartupOptions,
+    ) -> Result<WalletStartupOutcome, PlatformWalletError>
+    where
+        C: ContactCryptoProvider + Sync,
+        S: Signer<IdentityPublicKey> + Send + Sync,
+    {
+        let started = Instant::now();
+        let deadline = started + opts.budget;
```
thepastaclaw: **🟡 Suggestion: An oversized budget can abort the FFI host**

`WalletStartupOptions::budget` is unrestricted, and the new C entry point converts every nonzero `u64` directly to `Duration`. `Instant + Duration` panics when the resulting instant is not representable, including sufficiently large valid `budget_secs` values. `run_on_big_stack_thread` then re-panics through `join().expect(...)`; because the caller is an `extern "C"` function, this can abort the host process. Construct the deadline with `checked_add` and return `InvalidParameter` for an unsupported duration, then update the method's error documentation accordingly.

```suggestion
let deadline = started.checked_add(opts.budget).ok_or_else(|| {
    PlatformWalletError::InvalidParameter(
        "startup budget exceeds the supported duration range".to_string(),
    )
})?;
```

<sub>source: ['codex']</sub>

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerStartup.swift
```diff
@@ -0,0 +1,190 @@
...
+        let budgetSecs: UInt64
+        switch budget {
+        case .none:
+            budgetSecs = 0
+        case .some(let requested) where !requested.isFinite || requested < 0:
+            throw PlatformWalletError.invalidParameter(
+                "budget must be a finite, non-negative number of seconds, got \(requested)"
+            )
+        case .some(let requested):
+            budgetSecs = max(1, UInt64(requested.rounded()))
```
thepastaclaw: **🟡 Suggestion: A large finite Swift budget traps during UInt64 conversion**

The validation rejects negative and non-finite values, but a finite `TimeInterval` larger than the `UInt64` range still reaches the trapping `UInt64` initializer. For example, `Double.greatestFiniteMagnitude` crashes the caller instead of producing `PlatformWalletError.invalidParameter`. Use the failable exact conversion after rounding. Rust still needs its own independent deadline validation because many representable `UInt64` values cannot be added to an `Instant`.

```suggestion
case .some(let requested):
    let rounded = requested.rounded()
    guard let converted = UInt64(exactly: rounded) else {
        throw PlatformWalletError.invalidParameter(
            "budget is outside the supported UInt64 seconds range: \(requested)"
        )
    }
    budgetSecs = max(1, converted)
```

<sub>source: ['codex']</sub>

thepastaclaw (commented):

The ordered startup flow is directionally sound, and the previously reported sync/drain timeout issue is fixed at the exact head. Six in-scope correctness issues remain: readiness can be overstated, discovery can exceed the deadline, extreme budgets can crash at both language boundaries, and two startup paths can misclassify or prematurely fail.
Source: reviewer backend model gpt-5.6-sol (general, rust-quality, and ffi-engineer lanes); final verifier backend model gpt-5.6-sol. openclaw-agent/cliproxy/gpt-5.6-sol was orchestration-only and is not reviewer evidence.

### Review provenance

🟡 6 suggestion(s)

<details>

```

In `packages/rs-platform-wallet/src/manager/startup.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/manager/startup.rs:363-365: A failed contact sync can still be classified as Ready
  The tally records a successful sync whenever `sync_contact_requests` returns `Ok`, but that method catches received and sent fetch failures per identity and ultimately returns `Ok(all_requests)` even if Platform was unreachable for every identity. Explicit errors and timeouts leave `dashpay_sync_ran` false, but `StartupTally::status()` does not inspect that field either. If the existing deferred queue is empty, all three cases become `Ready`, even though no complete contact-document pass occurred and undiscovered requests may require account builds. This contradicts the `Ready` contract and can start SPV without the contact addresses this API exists to prepare. Return a completion summary from the sync operation and add a partial classification whenever any identity's contact fetch did not complete.
- [SUGGESTION] packages/rs-platform-wallet/src/manager/startup.rs:478-482: Identity discovery is not bounded by the startup deadline
  Each discovery attempt is awaited directly instead of through `within_budget`. One attempt can perform up to `gap_limit` sequential Platform fetches and DPNS lookups, so it can continue well beyond `opts.budget`; after a backoff consumes all remaining time, the loop can also start another unbounded attempt with an already-expired deadline. This breaks the documented whole-sequence ceiling and can keep the synchronous FFI caller from starting Core SPV during a Platform outage. Bound the entire discovery future by the remaining deadline and stop immediately on timeout. Because discovery persists sightings incrementally, re-check wallet-local identities before reporting `PartialNoIdentity` after cancellation.
- [SUGGESTION] packages/rs-platform-wallet/src/manager/startup.rs:329-330: An oversized budget can abort the FFI host
  `WalletStartupOptions::budget` is unrestricted, and the new C entry point converts every nonzero `u64` directly to `Duration`. `Instant + Duration` panics when the resulting instant is not representable, including sufficiently large valid `budget_secs` values. `run_on_big_stack_thread` then re-panics through `join().expect(...)`; because the caller is an `extern "C"` function, this can abort the host process. Construct the deadline with `checked_add` and return `InvalidParameter` for an unsupported duration, then update the method's error documentation accordingly.
- [SUGGESTION] packages/rs-platform-wallet/src/manager/startup.rs:484-490: An empty discovery return can misclassify a concurrently inserted identity
  `IdentityWallet::discover` returns only identities newly inserted by that invocation, while its internal scan separately treats already-managed identity sightings as trustworthy. Two concurrent startup calls can both pass the initial local check; after one inserts the identity, the other sees it on Platform but finds it already managed, returns an empty vector, and records proven absence here. That caller then skips contact sync and reports `NoIdentity` even though the wallet now owns the identity. Re-check local state whenever discovery returns an empty vector before classifying the result as a proof of absence.

In `packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerStartup.swift`:
- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerStartup.swift:145-150: A large finite Swift budget traps during UInt64 conversion
  The validation rejects negative and non-finite values, but a finite `TimeInterval` larger than the `UInt64` range still reaches the trapping `UInt64` initializer. For example, `Double.greatestFiniteMagnitude` crashes the caller instead of producing `PlatformWalletError.invalidParameter`. Use the failable exact conversion after rounding. Rust still needs its own independent deadline validation because many representable `UInt64` values cannot be added to an `Instant`.

In `packages/rs-platform-wallet-ffi/src/wallet_startup.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/wallet_startup.rs:146-155: Eager mnemonic resolution bypasses the startup outcome contract
  The FFI resolves the master xpriv before Rust checks whether the wallet already has an identity locally. A Keychain race, access denial, missing item, or malformed mnemonic therefore returns an immediate FFI error even on a warm launch that requires no discovery and may have no deferred crypto work. This contradicts both the FFI and Swift documentation that only configuration, malformed-ID, invalid-handle, and unknown-wallet failures throw; sync and drain failures are supposed to become structured startup outcomes so Core SPV can proceed. Defer master resolution until discovery is actually required, while retaining the resolver-backed provider for drain operations that genuinely need it, or translate resolver failure into a structured partial outcome.
```
</details>

packages/rs-platform-wallet/src/manager/startup.rs
```diff
@@ -0,0 +1,710 @@
...
+                None => identity_wallet.discover(opts).await,
+            };
+
+            match result {
+                Ok(found) => {
+                    match found.first() {
+                        Some(identity) => tally.record_discovered(identity.id()),
+                        None => tally.record_proven_absent(),
+                    }
+                    return;
```
thepastaclaw, line 651: **🟡 Suggestion: An empty discovery return can misclassify a concurrently inserted identity**

`IdentityWallet::discover` returns only identities newly inserted by that invocation, while its internal scan separately treats already-managed identity sightings as trustworthy. Two concurrent startup calls can both pass the initial local check; after one inserts the identity, the other sees it on Platform but finds it already managed, returns an empty vector, and records proven absence here. That caller then skips contact sync and reports `NoIdentity` even though the wallet now owns the identity. Re-check local state whenever discovery returns an empty vector before classifying the result as a proof of absence.

```suggestion
Ok(found) => {
    let identity_id = match found.first() {
        Some(identity) => Some(identity.id()),
        None => self.local_identity_id(identity_wallet.wallet_id()).await,
    };
    match identity_id {
        Some(identity_id) => tally.record_discovered(identity_id),
        None => tally.record_proven_absent(),
    }
    return;
}
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/wallet_startup.rs
```diff
@@ -0,0 +1,220 @@
...
+    // Resolve the master xpriv once, up front. The helper holds the mnemonic
+    // and seed in `Zeroizing` buffers and scrubs them before returning; the
+    // master itself has no `Drop`, so it is erased explicitly below.
+    let mut master = if mnemonic_resolver_handle.is_null() {
+        None
+    } else {
+        match resolve_master_from_resolver(mnemonic_resolver_handle, &wid, network) {
+            Ok(master) => Some(master),
+            Err(e) => return e,
+        }
```
thepastaclaw: **🟡 Suggestion: Eager mnemonic resolution bypasses the startup outcome contract**

The FFI resolves the master xpriv before Rust checks whether the wallet already has an identity locally. A Keychain race, access denial, missing item, or malformed mnemonic therefore returns an immediate FFI error even on a warm launch that requires no discovery and may have no deferred crypto work. This contradicts both the FFI and Swift documentation that only configuration, malformed-ID, invalid-handle, and unknown-wallet failures throw; sync and drain failures are supposed to become structured startup outcomes so Core SPV can proceed. Defer master resolution until discovery is actually required, while retaining the resolver-backed provider for drain operations that genuinely need it, or translate resolver failure into a structured partial outcome.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/manager/startup.rs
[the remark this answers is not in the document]
thepastaclaw: This no longer applies at the exact head. `sync_contact_requests` is wrapped by `within_budget` at lines 363-385, and both drain operations are wrapped at lines 399-415. `pending_contact_crypto_count` is intentionally outside the timeout because it is a local lock-protected queue-length read with no network or persistence I/O. Identity discovery remains unbounded, but that is a separate call-site issue.

packages/rs-platform-wallet/src/manager/startup.rs
romchornyi: Fixed in 4f04fd3, partially — and I want to be precise about which half.

`Ready` now requires `dashpay_sync_ran`, so an explicit error or a timeout on the contact pass no longer produces it: those cases fall to `PartialAccountsPending`.

What is **not** fixed is the case you name in the middle: `sync_contact_requests` catches per-identity fetch failures internally and returns `Ok(all_requests)` even when Platform was unreachable for every identity, so that still counts as a completed pass. Making it honest means returning a completion summary and changing its signature, which touches the recurring sweep and every other caller — wider than this PR should carry. Worth a follow-up; happy to open one if you agree that is the right split.

packages/rs-platform-wallet/src/manager/startup.rs
romchornyi: Fixed in 4f04fd3. You are right, and this is the same gap I closed for the sync pass and both drains in the previous round while leaving the scans themselves out — one scan walks up to `gap_limit` indices, each a Platform fetch plus a DPNS lookup, so it can outlast the whole budget on its own.

Each attempt now runs through `within_budget`. And per your last sentence: on timeout the local identity set is re-checked before concluding anything, because sightings persist incrementally and an abandoned scan may still have folded one in.

packages/rs-platform-wallet/src/manager/startup.rs
romchornyi: Fixed in 4f04fd3, taking your suggestion. Built with `checked_add` and returning `InvalidParameter`.

The abort path is the part worth spelling out: `run_on_big_stack_thread` re-raises through `join().expect(...)`, and the caller is `extern "C"`, so this was a host-process abort reachable from a plain argument value.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerStartup.swift
romchornyi: Fixed in 4f04fd3 with `UInt64(exactly:)` as suggested.

Noted your point that Rust still needs its own validation — done in the same commit via `checked_add`, since a representable `UInt64` of seconds can still be unaddable to an `Instant`. The two checks are deliberately independent rather than one trusting the other.

packages/rs-platform-wallet/src/manager/startup.rs
romchornyi: Fixed in 4f04fd3. This is the subtlest of the six: `discover` reports only identities that call inserted, so a concurrent bring-up that inserts first leaves the second seeing an already-managed identity and returning empty — which my code read as proof of absence for a wallet that demonstrably owns one.

Local state is now consulted before classifying an empty return, and the wallet id is threaded into the helper for it. Same check after an abandoned scan, for the reason you gave on the neighbouring thread.

packages/rs-platform-wallet-ffi/src/wallet_startup.rs
romchornyi: Fixed in 4f04fd3. The contradiction is real: the FFI and Swift docs both promise that only handle and wallet-id problems throw, while a Keychain hiccup could fail a warm launch that needed no scan at all.

Took the second option you offered — resolver failure is now a warning rather than an early return. If a scan does turn out to be required it runs without key material, fails, and surfaces as `DiscoveryFailed`, which is exactly the structured outcome that status was added for. I preferred this to deferring resolution because the alternative means either duplicating the local-identity check FFI-side or plumbing a lazy resolver through the crate boundary, and neither buys anything the warning path does not.

packages/rs-platform-wallet/src/manager/startup.rs
romchornyi: Fixed in 1af08c9 — and you are right that my previous fix only covered half of it.

Making a resolver failure non-fatal fixed the error contract but left the resolution happening unconditionally, so a warm launch still paid for a Keychain round trip whose result it would never use, and spent that time outside the budget it was about to be measured against.

Key material is only ever needed for a scan, and a wallet with an identity on file does not run one. New `has_local_identity_blocking` answers that from a single lock read, and the resolve is skipped when the answer is yes.

packages/rs-platform-wallet/src/manager/startup.rs
romchornyi: Fixed in 1af08c9. `checked_add` made `InvalidParameter` reachable and I did not update the docs that promised otherwise — all three (Rust, FFI, Swift) now say the same thing: errors are about the request, outcomes are about the run.

packages/rs-platform-wallet/src/manager/startup.rs
thepastaclaw: Thanks for separating the two cases. `4f04fd3` fixes the explicit-error/timeout half, but I don't think the swallowed per-identity failure can be deferred: `Ready` is this API's promise that the contact-document pass completed before SPV starts, and the remaining path can still make that promise without discovering the requests whose accounts/addresses need to be built.

Please carry completion information through `sync_contact_requests` (and update its callers) or otherwise ensure an incomplete per-identity fetch cannot produce `Ready` in this PR. A follow-up would leave the central startup-ordering contract knowingly unsound in the meantime; weakening/documenting `Ready` would also push the policy ambiguity back onto clients, which this PR is intended to remove.

thepastaclaw (commented):

The latest commits correctly bound the Rust network operations, prevent deadline overflow, handle concurrent discovery insertion, and skip unnecessary warm-launch key resolution. Four in-scope correctness issues remain: readiness still accepts internally incomplete contact synchronization, partial identity scans can suppress all later discovery, transient key-material failures become terminal outcomes, and cold key resolution remains outside the documented whole-sequence budget. Source: reviewer backend model gpt-5.6-sol (general, rust-quality, and ffi-engineer lanes); final verifier backend model gpt-5.6-sol. openclaw-agent/cliproxy/gpt-5.6-sol was orchestration-only and is not reviewer evidence.

### Review provenance

🟡 3 suggestion(s)

<details>

```

In `packages/rs-platform-wallet/src/manager/startup.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/manager/startup.rs:360-364: The warm-launch shortcut can strand identities hidden by a partial scan
  A locally persisted identity does not prove that the preceding gap-limit scan completed. Discovery deliberately returns success after any sighting even when later probes failed (`ScanTally::is_trustworthy`), and its warning states that an identity at a failed index may remain missing until the next scan. This shortcut prevents that next startup scan: after the first identity is persisted, every later launch skips discovery and only synchronizes identities already known locally. The timeout path has the same effect when a sighting was persisted before cancellation. Persist scan-completion state, or otherwise distinguish a complete gap-limit scan from a partial scan, so restored multi-identity wallets remain discoverable on later launches.
- [SUGGESTION] packages/rs-platform-wallet/src/manager/startup.rs:385-387: A failed contact sync can still be classified as Ready
  (existing thread: https://github.com/dashpay/platform/pull/4359#discussion_r3752448291)
  `Some(Ok(requests))` does not prove that the contact pass completed. `sync_contact_requests` catches received-fetch failures at `contact_requests.rs:1114-1127` and sent-fetch failures at `1135-1149`, continues processing, and ultimately returns `Ok(all_requests)` at line 1417. If Platform is unreachable for every identity, this arm still records `dashpay_sync_ran`; an empty deferred queue then produces `Ready` even though no contact requests were discovered or enqueued. Because `Ready` is the new API's promise that contact-derived addresses are prepared before SPV starts, propagate per-identity and per-direction completion information from `sync_contact_requests` and record a completed startup pass only when every required fetch and ingest succeeded.

In `packages/rs-platform-wallet-ffi/src/wallet_startup.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/wallet_startup.rs:163-175: Transient key-material lookup failures become terminal discovery failures
  Every resolver error is discarded and converted to `master = None`. For an external-signable wallet requiring a cold scan, that selects resident-key discovery, which fails because the in-process wallet intentionally has no private key; startup then classifies the error as `DiscoveryFailed`, whose `discovery_worth_retrying()` result is false. The Swift preflight has the same loss of information: `WalletStorage.hasMnemonic` returns false for every non-success Security status, and line 130 consequently passes a null resolver for both a genuinely absent mnemonic and transient protected-data or Keychain failures. Preserve an unavailable/error state across both checks and expose it as retryable, rather than routing recoverable Keychain failures through the terminal local-discovery status.
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/wallet_startup.rs:163-177: Cold-start key resolution is outside the startup budget
  `resolve_master_from_resolver` synchronously invokes the Swift-owned Keychain resolver and performs mnemonic parsing, seed derivation, and master-key construction before `start_wallet_subsystems` creates its timer and deadline at `startup.rs:341-346`. Cold-start resolver latency is therefore neither deducted from `opts.budget` nor included in the exported `elapsed_ms`; a blocked callback can also hold the synchronous FFI call beyond the API's stated whole-sequence ceiling. Start deadline accounting before key resolution and include the callback in the bounded orchestration, passing only the remaining duration to subsequent steps. If key resolution is intentionally outside the budget, narrow the Rust, C, and Swift contracts instead of describing the budget and elapsed value as covering the whole sequence.
```
</details>

packages/rs-platform-wallet/src/manager/startup.rs
```diff
@@ -0,0 +1,770 @@
...
+            .get_wallet(wallet_id)
+            .await
+            .ok_or_else(|| PlatformWalletError::WalletNotFound(hex::encode(wallet_id)))?;
+        let identity_wallet = wallet.identity();
+
+        // 1. Local identities first. A warm launch must not pay for a network
+        //    scan it does not need.
+        if let Some(known) = self.local_identity_id(wallet_id).await {
+            tally.record_local_identity(known);
+        } else {
```
thepastaclaw, line 446: **🟡 Suggestion: The warm-launch shortcut can strand identities hidden by a partial scan**

A locally persisted identity does not prove that the preceding gap-limit scan completed. Discovery deliberately returns success after any sighting even when later probes failed (`ScanTally::is_trustworthy`), and its warning states that an identity at a failed index may remain missing until the next scan. This shortcut prevents that next startup scan: after the first identity is persisted, every later launch skips discovery and only synchronizes identities already known locally. The timeout path has the same effect when a sighting was persisted before cancellation. Persist scan-completion state, or otherwise distinguish a complete gap-limit scan from a partial scan, so restored multi-identity wallets remain discoverable on later launches.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/wallet_startup.rs
```diff
@@ -0,0 +1,241 @@
...
+        match resolve_master_from_resolver(mnemonic_resolver_handle, &wid, network) {
+            Ok(master) => Some(master),
+            Err(_) => {
+                tracing::warn!(
+                    wallet_id = %hex::encode(wid),
+                    "startup: could not resolve the wallet mnemonic; continuing without \
+                     scan key material"
+                );
+                None
+            }
```
thepastaclaw: **🟡 Suggestion: Transient key-material lookup failures become terminal discovery failures**

Every resolver error is discarded and converted to `master = None`. For an external-signable wallet requiring a cold scan, that selects resident-key discovery, which fails because the in-process wallet intentionally has no private key; startup then classifies the error as `DiscoveryFailed`, whose `discovery_worth_retrying()` result is false. The Swift preflight has the same loss of information: `WalletStorage.hasMnemonic` returns false for every non-success Security status, and line 130 consequently passes a null resolver for both a genuinely absent mnemonic and transient protected-data or Keychain failures. Preserve an unavailable/error state across both checks and expose it as retryable, rather than routing recoverable Keychain failures through the terminal local-discovery status.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/wallet_startup.rs
```diff
@@ -0,0 +1,241 @@
...
+        }
+    };
```
thepastaclaw: **🟡 Suggestion: Cold-start key resolution is outside the startup budget**

`resolve_master_from_resolver` synchronously invokes the Swift-owned Keychain resolver and performs mnemonic parsing, seed derivation, and master-key construction before `start_wallet_subsystems` creates its timer and deadline at `startup.rs:341-346`. Cold-start resolver latency is therefore neither deducted from `opts.budget` nor included in the exported `elapsed_ms`; a blocked callback can also hold the synchronous FFI call beyond the API's stated whole-sequence ceiling. Start deadline accounting before key resolution and include the callback in the bounded orchestration, passing only the remaining duration to subsequent steps. If key resolution is intentionally outside the budget, narrow the Rust, C, and Swift contracts instead of describing the budget and elapsed value as covering the whole sequence.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/manager/startup.rs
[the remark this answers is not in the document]
thepastaclaw: This no longer applies at the current head. `within_budget` wraps the contact sync and both network drains; `pending_contact_crypto_count` intentionally remains outside the timeout because it is a local queue-length read with no network or persistence I/O.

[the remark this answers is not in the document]
thepastaclaw: This is a mute/meta comment and contains no code-review finding to validate.

packages/rs-platform-wallet-ffi/src/wallet_startup.rs
```diff
@@ -0,0 +1,234 @@
...
+            Ok(master) => Some(master),
+            Err(_) => {
+                tracing::warn!(
+                    wallet_id = %hex::encode(wid),
+                    "startup: could not resolve the wallet mnemonic; continuing without \
+                     scan key material"
+                );
+                None
+            }
+        }
```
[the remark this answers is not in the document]
thepastaclaw: The reported warm-launch behavior is fixed at the current head: `has_local_identity_blocking` runs before mnemonic resolution, and a locally known identity causes the resolver to be skipped. Cold-start resolver work remains outside budget accounting, which is tracked separately.

packages/rs-platform-wallet/src/manager/startup.rs
```diff
@@ -0,0 +1,765 @@
...
+        // `Instant + Duration` panics when the sum is not representable, and
+        // this runs under an `extern "C"` caller whose thread wrapper re-raises
+        // a panic as an abort — so an absurd budget would take the host process
+        // down rather than return an error.
+        let deadline = started.checked_add(opts.budget).ok_or_else(|| {
+            PlatformWalletError::InvalidParameter(format!(
+                "startup budget {:?} exceeds the supported duration range",
+                opts.budget
+            ))
+        })?;
```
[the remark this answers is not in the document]
thepastaclaw: The documentation is updated at the current head. Rust documents `InvalidParameter` for an unrepresentable deadline, and the FFI and Swift contracts both list unsupported budgets among request errors.

QuantumExplorer (commented):
Reviewed the full diff, plus the drain/discovery paths it calls into. The direction is right — this ordering is exactly the kind of policy that belongs in the Rust library, and the `StartupTally` split with the empty-vs-unreachable pin is well done (`manager::startup` tests: 14 passed locally). One must-fix, one structural should-fix, and a few notes.

## Must fix: the drains are not cancel-safe, and this PR is their first caller that cancels them

`within_budget` is `tokio::time::timeout`, i.e. drop-the-future on expiry. Every pre-existing caller of the two drains awaits them to completion (`platform_wallet_drain_pending_contact_crypto`, the `payments.rs` call sites) — the wrap at [startup.rs#L419-L440](https://github.com/dashpay/platform/blob/1af08c9286/packages/rs-platform-wallet/src/manager/startup.rs#L419-L440) introduces cancellation as new surface.

Both drains accumulate their dequeue list in a local `cleared` vec and apply it **once, after the loop** ([contact_requests.rs#L2179](https://github.com/dashpay/platform/blob/1af08c9286/packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs#L2179) for the provider drain, [#L2396](https://github.com/dashpay/platform/blob/1af08c9286/packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs#L2396) for auto-accepts), while per-entry side effects (`register_contact_account`, `mark_contact_channel_broken`, the reciprocal send) commit as they go. So a budget expiry at entry 14 of 14:

- 13 accounts really exist, but all 14 entries stay queued;
- `drained` reports `0` (the `.unwrap_or(0)`), and the unbounded `pending_contact_crypto_count()` returns the full pre-drain count — so the outcome says `drained=0, pending=14` when the truth is 13 built, 1 remaining.

I did verify the damage stops there: `accept_contact_request_with_external_signer` checks `already_established` / `already_reciprocated` before broadcasting ([contact_requests.rs#L2752-L2764](https://github.com/dashpay/platform/blob/1af08c9286/packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs#L2752-L2764)), so the next launch's re-drain resolves those entries locally — no duplicate state transition, no unique-index rejection, no credit spend. And the provider-only drain skips `AutoAccept` entries, so `drained + accepted` doesn't double-count. The queue is at-least-once by design and survives this.

What doesn't survive is the outcome struct — the thing this PR exists to return — which is wrong exactly and only in the case the budget fires. The module doc's "entries it did not complete stay queued" is true but incomplete: entries it *did* complete also stay queued. Fix belongs in `contact_requests.rs`: clear each entry as it lands (or flush in small batches), rather than one apply at the end. That makes the drains honestly abandonable and the counters truthful.

## Should fix: `needs_scan_key` re-implements the library's scan gate in the FFI, and mislabels the failure it can cause

[wallet_startup.rs#L148-L177](https://github.com/dashpay/platform/blob/1af08c9286/packages/rs-platform-wallet-ffi/src/wallet_startup.rs#L148-L177) predicts "key material is only needed if the library will scan, and it won't scan when a local identity exists." That same rule lives in the library at [startup.rs#L360-L374](https://github.com/dashpay/platform/blob/1af08c9286/packages/rs-platform-wallet/src/manager/startup.rs#L360-L374) — two copies in two crates with nothing tying them together, and `has_local_identity_blocking` was added solely so the FFI can make the prediction. The JNI bridge will need a third copy. If the library ever changes when it scans (rescan-after-incomplete, a `force` option, deeper identity indices), the FFI silently supplies `master = None` and the external-signable path breaks.

The sharper consequence is the swallow at L168: a resolver failure becomes `master = None` + a warn, the library then takes the resident-wallet derive, which for an external-signable wallet errors ("External signable wallet has no private key") — not `IdentityDiscoveryIncomplete` — so the tally records a **local, terminal** failure and the client gets `DiscoveryFailed`. But the realistic causes of a resolver failure at launch are a locked device, a cancelled biometric prompt, or the Keychain not being available yet — all transient. `DiscoveryFailed` is documented, in both Rust and Swift, as "another scan will not answer it," and the Swift caller sees `discoveryWorthRetrying == false` and `identityIsSettled == false`: no correct action remains.

Suggestion: make the key source lazy — `start_wallet_subsystems` takes a closure/provider returning `Result<ExtendedPrivKey, _>`, invoked only on the branch that actually scans. Then the FFI drops to pure marshalling, both new accessors can go away, Android inherits the behaviour, and the library sees the resolver error directly and can classify it honestly (a retryable key-material-unavailable outcome rather than terminal `DiscoveryFailed`).

Related, worth a doc line either way: the resolver currently runs before `Instant::now()`, so the Keychain round trip sits outside the budget the FFI doc promises — and note that even moved inside, `tokio::time::timeout` cannot interrupt a synchronous `SecItemCopyMatching`, so a biometric prompt blocks past the deadline regardless.

## Notes (non-blocking)

- **The only production `ContactCryptoProvider` lives in the FFI crate.** `platform-wallet` ships the trait with test-only impls, so a non-FFI consumer can never drain, and this call can't self-supply a provider even for a resident-key wallet (it reports `PartialAccountsPending` forever). Pre-existing, not introduced here — and `SeedCryptoProvider` is already the exact shape a resident-wallet provider would take, if someone wants the follow-up.
- **Swift invents a budget floor.** Because `0` is the "use default" sentinel, Swift needs `max(1, converted)` so a 0.4s request doesn't silently become 20s — a policy with no Rust-side home that the JNI bridge will reinvent differently. Milliseconds across the boundary, or an explicit use-default flag, removes the rule.
- **`PartialAccountsPending` collapses three causes the library can tell apart** (no provider / budget expiry / per-entry failures) — its own doc admits it. A reason field would keep the FFI's `provider = None` decision visible downstream.
- **The manager handle store's read guard is now held for up to `budget`.** That `parking_lot` lock is write-preferring: a teardown arriving mid-startup queues every other manager FFI call behind it for the remainder. Same pattern as `platform_address_sync`, but this is the first hold with a user-configurable duration.
- Warm launch still performs one attribute-only `SecItemCopyMatching` in `hasMnemonic` — cheap and promptless, so the skip achieves what matters; just noting the last commit message slightly overstates it.

Verified by reading source rather than assumed: the discovery error-classification chain, both drains' clear-at-end structure and their full caller set, the accept-path idempotency guards, the `VTableSigner` cast (matches the four existing call sites), and that headers are cbindgen-generated. Ran `cargo test -p platform-wallet --lib manager::startup` (14/14). Did not build the Swift side.

packages/rs-platform-wallet/src/manager/startup.rs
romchornyi: Confirmed, and it is a regression from this PR rather than a pre-existing property — before the warm-launch shortcut this call always scanned. Filed as #4365 rather than fixed here; reasoning below.

The chain is exactly as you describe, and the budget-expiry path shares it: `within_budget` returns `None`, local state is consulted, and a sighting persisted before cancellation is recorded as a warm launch. Worth adding that on iOS the only automatic fallback has the identical gate — `DWCurrentUserIdentityInfo.recoverIfNeeded` scans only when there are no local identities at all — so the sole remedy today is the manual **Settings → Identities → Discover** button.

I looked for a fix that does not need new persisted state, and there isn't one:

- **Inferring it from the local index set** catches the wrong case. A hole below the maximum known index (identities at 0 and 2, nothing at 1) is visible through `IdentityEntry::identity_index`. The case that matters is not: found at 0, failed at 1, identity really at 1 leaves `{0}` — identical to what a clean scan of a single-identity wallet leaves.
- **Reusing existing storage** doesn't work either. `sqlite/schema/blob.rs` is a codec, not a KV table, and `WalletMetadataEntry` is emitted only at registration and requires `network` + `wallet_group_id`, neither of which a scan has to hand.

So it needs a column plus a `refinery` V002 — only V001 exists today — which is a schema-version bump, and a database migrated forward is refused by older app binaries. That is a release-coordination decision, not something to fold into a startup-ordering PR.

The other way to close it is to drop the shortcut and scan unconditionally. That costs ~5 probes (`IDENTITY_GAP_LIMIT`), each a Platform fetch plus a DPNS lookup, plus a Keychain round trip, before every Core SPV start — which is the cost the previous review round asked to remove.

Leaving the shortcut in with the gap tracked in #4365. `ScanTally` already knows `failed_probes` at the point it makes the verdict, so the fix is small once there is somewhere to record it.

packages/rs-platform-wallet/src/manager/startup.rs
```diff
@@ -0,0 +1,837 @@
...
+
+            tally.record_discovery_gave_up();
+        }
+        .await;
+
+        // The key lives no longer than the scan that needed it. `ExtendedPrivKey`
+        // has no `Drop`, so this is the only thing that clears it.
+        if let Some(master) = master.as_mut() {
+            master.private_key.non_secure_erase();
+        }
```
thepastaclaw: **🟡 Suggestion: Resolved scan key is not erased when the startup future is cancelled**

The resolver transfers an owned `ExtendedPrivKey` into this async function, but erasure occurs only in explicit code after the discovery block finishes. If a Rust caller drops or cancels `start_wallet_subsystems` while discovery is awaiting Platform, the future is dropped and lines 629-633 never run. `ExtendedPrivKey` has no erasing `Drop`, as documented by `resolve_master_from_resolver`, so its private scalar remains in freed future memory despite this API's per-call key-material contract. Store the resolved key in an RAII guard whose `Drop` calls `non_secure_erase`, covering completion, cancellation, and unwinding.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/manager/startup.rs
```diff
@@ -0,0 +1,837 @@
...
+        tally.record_drain(1, 0);
+
+        let outcome = tally.into_outcome(Duration::from_secs(3));
+        assert_eq!(outcome.status, WalletStartupStatus::Ready);
+        assert_eq!(outcome.identity_id, Some(identity()));
+        assert!(outcome.dashpay_sync_ran);
+        assert_eq!(outcome.contact_accounts_drained, 1);
+        assert_eq!(outcome.elapsed, Duration::from_secs(3));
+    }
+}
```
thepastaclaw, line 885: **🟡 Suggestion: The ordered orchestration is not exercised by its tests**

Every test in this module exercises `StartupTally`, outcome conversion, or the standalone `within_budget` helper; none constructs a manager or invokes `start_wallet_subsystems`. The PR's central behavior—lazy resolver invocation, identity/contact/drain ordering, retry wiring, deadline propagation, and translation of real operation results into the tally—can therefore regress while these tests remain green. Add a manager-level test with a mock SDK and recording `ScanKeyResolver` and `ContactCryptoProvider` implementations that invokes the public method and asserts the observed order plus representative timeout and failure paths.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/wallet_startup.rs
```diff
@@ -0,0 +1,236 @@
...
+                // Carried as an error the library can classify, not as an FFI
+                // throw: the documented contract is that only handle and
+                // wallet-id problems throw, and a locked device is neither.
+                platform_wallet::error::PlatformWalletError::KeyDerivation(format!(
+                    "mnemonic resolver failed ({:?})",
+                    e.code
+                ))
+            })
+    };
+    let scan_key = (!mnemonic_resolver_handle.is_null())
```
thepastaclaw, line 186: **🟡 Suggestion: Resolver error flattening makes permanent key faults retryable**

The callback adapter converts every failure from `resolve_master_from_resolver` into the same `PlatformWalletError::KeyDerivation`, and startup maps every resolver error to retryable `PartialNoIdentity`. That is appropriate for a locked device or temporary Keychain outage, but the lower layer also reports invalid UTF-8, invalid BIP-39 phrases, invalid callback lengths, an oversized buffer, and master-key construction failures. Those persistent local or callback-contract faults cannot be repaired by another discovery attempt, yet Swift receives `discoveryWorthRetrying == true` and may repeatedly schedule one. Preserve a typed temporary-unavailable versus malformed/permanent distinction across the resolver adapter and classify only the former as retryable.

<sub>source: ['codex']</sub>

thepastaclaw (commented):

The centralized startup ordering is directionally sound, and the latest commits improve lazy key resolution, deadline-aware draining, and cancellation-safe queue bookkeeping. Seven in-scope suggestions remain: incomplete contact or identity discovery can still be presented as settled, synchronous callbacks can overrun the advertised budget, resolved key material is not erased on cancellation, resolver faults lose their permanence classification, and the real orchestration wiring remains untested.
Source: reviewer backends gpt-5.6-sol (general), gpt-5.6-sol (rust-quality), and gpt-5.6-sol (ffi-engineer); final verifier backend gpt-5.6-sol. openclaw-agent/cliproxy/gpt-5.6-sol was orchestration-only and is not reviewer evidence.

### Review provenance

🟡 4 suggestion(s)

<details>

```

In `packages/rs-platform-wallet/src/manager/startup.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/manager/startup.rs:534-633: Resolved scan key is not erased when the startup future is cancelled
  The resolver transfers an owned `ExtendedPrivKey` into this async function, but erasure occurs only in explicit code after the discovery block finishes. If a Rust caller drops or cancels `start_wallet_subsystems` while discovery is awaiting Platform, the future is dropped and lines 629-633 never run. `ExtendedPrivKey` has no erasing `Drop`, as documented by `resolve_master_from_resolver`, so its private scalar remains in freed future memory despite this API's per-call key-material contract. Store the resolved key in an RAII guard whose `Drop` calls `non_secure_erase`, covering completion, cancellation, and unwinding.
- [SUGGESTION] packages/rs-platform-wallet/src/manager/startup.rs:637-837: The ordered orchestration is not exercised by its tests
  Every test in this module exercises `StartupTally`, outcome conversion, or the standalone `within_budget` helper; none constructs a manager or invokes `start_wallet_subsystems`. The PR's central behavior—lazy resolver invocation, identity/contact/drain ordering, retry wiring, deadline propagation, and translation of real operation results into the tally—can therefore regress while these tests remain green. Add a manager-level test with a mock SDK and recording `ScanKeyResolver` and `ContactCryptoProvider` implementations that invokes the public method and asserts the observed order plus representative timeout and failure paths.
- [SUGGESTION] packages/rs-platform-wallet/src/manager/startup.rs:417-425: A failed contact sync can still be classified as Ready
  (existing thread: https://github.com/dashpay/platform/pull/4359#discussion_r3752448291)
  `Some(Ok(requests))` does not prove that the contact pass completed. `sync_contact_requests` catches received-fetch failures at `contact_requests.rs:1150-1163`, sent-fetch failures at `1171-1185`, and per-direction persistence failures around `1299-1353`, then returns `Ok(all_requests)` at line 1453. This arm consequently records `dashpay_sync_ran` even when required directions were skipped. If the deferred queue is empty, the outcome becomes `Ready` despite never discovering or ingesting requests whose account builds should precede SPV startup. Return a typed completion summary from `sync_contact_requests` and record a completed startup pass only when every required fetch and ingest succeeded.
- [SUGGESTION] packages/rs-platform-wallet/src/manager/startup.rs:392-405: The warm-launch shortcut can strand identities hidden by a partial scan
  (existing thread: https://github.com/dashpay/platform/pull/4359#discussion_r3752862202)
  A locally persisted identity does not prove that the preceding gap-limit scan completed. `ScanTally::is_trustworthy()` returns success after any identity sighting even if later probes failed, and discovery explicitly warns that an identity at a failed index may remain missing until a later scan. The timeout path also accepts an identity persisted before cancellation. Once either path leaves one local identity, this shortcut skips every subsequent startup scan, so another seed identity at a failed or unvisited index can remain undiscovered indefinitely. Persist scan-completion state separately from discovered identities, or retain a rescan requirement whenever the preceding scan was partial.

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs:2571-2575: Auto-accept can run past the whole startup budget
  The deadline bounds proof-key derivation and stops the drain between entries, but once an entry reaches this call, `accept_contact_request_with_external_signer` is awaited without a deadline. That operation can fetch or derive account data, invoke signer/provider callbacks, broadcast a reciprocal state transition, and register accounts. The FFI signer alone permits an asynchronous completion to take up to 300 seconds, far beyond the default 20-second startup budget. A valid in-flight accept can therefore hold the synchronous startup call long after its documented ceiling. Thread the deadline through cancel-safe pre-commit stages and define safe post-broadcast completion behavior, or document that a started auto-accept entry is exempt from the ceiling and that the budget is only a stop-between-entries target.

In `packages/rs-platform-wallet-ffi/src/wallet_startup.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/wallet_startup.rs:158-176: Resolver error flattening makes permanent key faults retryable
  The callback adapter converts every failure from `resolve_master_from_resolver` into the same `PlatformWalletError::KeyDerivation`, and startup maps every resolver error to retryable `PartialNoIdentity`. That is appropriate for a locked device or temporary Keychain outage, but the lower layer also reports invalid UTF-8, invalid BIP-39 phrases, invalid callback lengths, an oversized buffer, and master-key construction failures. Those persistent local or callback-contract faults cannot be repaired by another discovery attempt, yet Swift receives `discoveryWorthRetrying == true` and may repeatedly schedule one. Preserve a typed temporary-unavailable versus malformed/permanent distinction across the resolver adapter and classify only the former as retryable.
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/wallet_startup.rs:158-177: Cold-start key resolution is outside the startup budget
  (existing thread: https://github.com/dashpay/platform/pull/4359#discussion_r3752862221)
  The Rust timer now starts before the lazy resolver runs, so resolver latency is included in `elapsed_ms` and deducted from later steps. However, `resolve()` remains a synchronous callback with no interruptible deadline boundary: a delayed Keychain query, mnemonic derivation, or custom resolver holds the synchronous C call until it returns. Swift's `mnemonicAvailability` Keychain lookup also runs before entering the timed Rust call. This still contradicts the Rust and Swift contracts describing `budget` as a ceiling for the whole sequence and the FFI contract saying the call must always terminate. Make key acquisition genuinely asynchronous/cancellable with callback ownership that can outlive an abandoned startup, or narrow all three public contracts to state that synchronous host key access is measured but not bounded.
```
</details>

packages/rs-platform-wallet-ffi/src/wallet_startup.rs
thepastaclaw: Resolved in `482997e` — _Transient key-material lookup failures become terminal discovery failures_ no longer present.

packages/rs-platform-wallet/src/manager/startup.rs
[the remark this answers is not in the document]
thepastaclaw: This is a mute/meta comment and contains no code-review finding to validate.

thepastaclaw (commented):

The centralized startup orchestration is directionally sound, and the latest commits correctly add lazy scan-key resolution, lifetime-based key erasure, typed resolver-failure classification, and incremental deadline-aware drains. Four in-scope suggestions remain: incomplete contact synchronization can still produce Ready, synchronous key resolution and a started auto-accept can overrun the advertised whole-sequence budget, and the production orchestration is not directly tested.
Source: Reviewer backend model gpt-5.6-sol (general, rust-quality, and ffi-engineer lanes); final verifier backend model gpt-5.6-sol. openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

### Review provenance

🟡 1 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs:2571-2574: Auto-accept can run past the whole startup budget
  The drain checks the deadline between entries and bounds proof-key derivation, but after a proof verifies it awaits `accept_contact_request_with_external_signer` without a deadline. That operation can derive account data, invoke provider and signer callbacks, broadcast the reciprocal state transition, fetch the contact identity, and register both accounts. The FFI signer's asynchronous completion timeout alone is 300 seconds, far beyond the default 20-second startup budget, so one started auto-accept can keep the startup call blocked long after its documented ceiling. Thread the deadline through cancellation-safe pre-commit work and define the required post-broadcast completion behavior, or document consistently that started auto-accept entries are exempt and the budget is only checked between entries.

In `packages/rs-platform-wallet/src/manager/startup.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/manager/startup.rs:467-469: A failed contact sync can still be classified as Ready
  (existing thread: https://github.com/dashpay/platform/pull/4359#discussion_r3752448291)
  This records a completed contact pass for every `Ok` from `sync_contact_requests`, but that method deliberately converts several incomplete outcomes into `Ok(all_requests)`. It skips an identity after a received-fetch failure at `contact_requests.rs:1150-1163`, ingests only the received side after a sent-fetch failure at 1171-1185, and catches per-direction persistence failures around 1293-1353. The SDK pagination helper also returns `Ok` after its 5,000-document per-sweep cap even when more documents remain. If no deferred entries are currently queued, `StartupTally::status` therefore returns `Ready` even though requests whose accounts must precede SPV may not have been fetched or persisted. Return a typed synchronization summary carrying per-identity, per-direction, persistence, and pagination-exhaustion state, and call `record_sync_ran` only when every required direction completed.
- [SUGGESTION] packages/rs-platform-wallet/src/manager/startup.rs:685-885: The ordered orchestration is not exercised by its tests
  (existing thread: https://github.com/dashpay/platform/pull/4359#discussion_r3753883576)
  The tests exercise `StartupTally`, outcome conversion, and the standalone `within_budget` helper, but none constructs a `PlatformWalletManager` or invokes `start_wallet_subsystems`. Consequently, the PR's central behavior—lazy resolver invocation, identity/contact/drain ordering, retry wiring, deadline propagation, cancellation cleanup, and translation of real operation results into the tally—can be miswired while every test in this module remains green. Add a manager-level test using a mock SDK and recording `ScanKeyResolver` and `ContactCryptoProvider` implementations, asserting the observed order and representative timeout, resolver-failure, and warm-launch paths.

In `packages/rs-platform-wallet-ffi/src/wallet_startup.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/wallet_startup.rs:161-167: Cold-start key resolution is outside the startup budget
  (existing thread: https://github.com/dashpay/platform/pull/4359#discussion_r3752862221)
  The manager timer now starts before this lazy resolver is invoked, so successful resolver latency is included in `elapsed_ms` and deducted from later steps. However, `resolve()` remains a synchronous host callback and cannot be interrupted at the deadline: a delayed Keychain query, mnemonic derivation, biometric flow, or custom resolver can hold the synchronous C entry point indefinitely. Swift also performs `mnemonicAvailability(for:)` synchronously before entering the timed Rust call. This conflicts with the Rust and Swift contracts describing `budget` as a ceiling for the whole sequence and the FFI documentation stating that the call must always terminate. Make host key acquisition asynchronous and cancellable with callback ownership that can survive abandoning startup, or narrow all public contracts to state that synchronous host key access is measured where applicable but not bounded.
```
</details>

packages/rs-platform-wallet-ffi/src/wallet_startup.rs
thepastaclaw: Resolved in `58a43aa` — _Resolver error flattening makes permanent key faults retryable_ no longer present.

packages/rs-platform-wallet/src/manager/startup.rs
thepastaclaw: Resolved in `58a43aa` — _Resolved scan key is not erased when the startup future is cancelled_ no longer present.

packages/rs-platform-wallet/src/manager/startup.rs
thepastaclaw: Resolved in `482997e` — _The warm-launch shortcut can strand identities hidden by a partial scan_ no longer present.

packages/rs-platform-wallet-ffi/src/wallet_startup.rs
thepastaclaw: Resolved in `482997e` — _Cold-start key resolution is outside the startup budget_ no longer present.

packages/rs-platform-wallet/src/manager/startup.rs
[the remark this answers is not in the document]
thepastaclaw: This finding is no longer present at the exact head. `sync_contact_requests` is wrapped by `within_budget`, both drains receive the shared deadline through their `_until` variants, and `pending_contact_crypto_count` is intentionally unbounded because it is a local queue-length read with no network or persistence I/O.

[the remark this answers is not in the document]
thepastaclaw: This is a mute/meta comment and contains no code-review finding.

packages/rs-platform-wallet/src/manager/startup.rs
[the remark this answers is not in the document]
thepastaclaw: The documentation issue is fixed at the exact head. Rust documents `InvalidParameter` for an unrepresentable deadline, and the FFI and Swift contracts both list unsupported budgets among request errors.

packages/rs-platform-wallet/src/manager/startup.rs
romchornyi: Fixed in 58a43aa — and you are right that moving the erase into the library is what created the window. In the FFI it sat after a `run_on_big_stack_thread` join, so it ran on every path out of that scope; as an explicit call after the discovery loop it runs on every path the loop takes, and on none at all if the whole future is dropped mid-await.

The key is now owned by a `ScanKeyGuard` that erases in `Drop`, so erasure follows the value's lifetime rather than a code path. That also let the scan loop go back to its plain shape — the `async` block around it existed only to funnel the loop's early returns onto the explicit call.

packages/rs-platform-wallet-ffi/src/wallet_startup.rs
romchornyi: Fixed in 58a43aa, taking the classification you asked for.

The blocker was that the FFI result codes cannot express it: `NOT_FOUND`, the Keychain-error bucket, an impossible length, a phrase in no wordlist and a failed master-key build all land on `ErrorWalletOperation`, so nothing downstream can tell them apart without reading the message. The distinction now travels with the failure — `ResolveFailureKind::{Unavailable, Permanent}` on a `ResolveFailure`, produced at each failure site by `_classified` variants of the two resolver helpers. The existing helpers are thin wrappers that drop the kind, so the other ~10 call sites are untouched.

On the library side that becomes `ScanKeyError::{Unavailable, Invalid}`: unavailable keeps the retryable path, invalid records a local discovery failure and settles as terminal `DiscoveryFailed`.

One deliberate departure from your list: `NOT_FOUND` is classified retryable, not permanent. The host filters watch-only wallets before calling (`mnemonicAvailability == .absent` → no resolver at all), so reaching `NOT_FOUND` means the item was expected and was not there — a wipe/restore race — rather than proof this wallet has no seed.

4 tests drive each cause through a fake resolver vtable. The mapping is one that reads fine while behaving wrong in opposite directions, so it is worth pinning.

packages/rs-platform-wallet/src/manager/startup.rs
romchornyi: Accurate, and it is called out as a known gap in the PR description rather than an oversight — every test here drives `StartupTally`, the outcome conversion, or `within_budget`, and none constructs a manager.

What it would take, having looked: a mock SDK via `SdkBuilder::new_mock()`, a registered wallet on a `SqlitePersister` (the `__test-helpers` feature has `wallet_meta::ensure_exists` for exactly this), and a fake `ContactCryptoProvider` — `SeedCryptoProvider` in `contact_requests.rs` is already the right shape for that half. The piece I could not confirm is the one that matters most for this module: whether the mock SDK can be made to fail an `Identity::fetch` in a way that surfaces as `IdentityDiscoveryIncomplete`, which is what the retry wiring and the empty-vs-unreachable distinction turn on. Without that, a manager-level test would cover the ordering but not the branch that has already regressed once.

I have taken the parts that are unit-testable without it: the deadline helper the drains use, and now the resolver's retry classification (4 tests, fake vtable). The orchestration itself I would rather see covered by a real integration test than by a mock deep enough to be its own fiction — filing that separately rather than half-doing it here.

QuantumExplorer (approved):

PR #4360, opened 2026-08-10
feat: expose Platform-to-Shielded capacity preflight

llbartekll:
## Summary

- add a cached, account-aware Platform Payment shield-capacity planner shared by preflight and execution
- expose the preflight result through Rust, C FFI, and Swift SDK APIs
- enforce the versioned minimum-input rules and map live insufficient-capacity races to a typed error

## Why

The wallet previously treated the displayed Platform balance as fully shieldable. A Platform Payment account may need to retain protocol reserve and may contain inputs that cannot participate in the transition, so sending the displayed maximum could fail with `Insufficient shielded balance`.

Companion iOS PR: https://github.com/dashpay/dashwallet-ios/pull/971

## Testing

- `cargo test -p platform-wallet --features shielded shield_input_selection_tests` (11 passed)
- `cargo test -p platform-wallet --features shielded shield_input_fetch_error_tests` (1 passed)
- `cargo test -p platform-wallet-ffi --features shielded shield` (15 passed)
- SwiftDashSDK `ErrorHandlingTests` (45 passed)
- `cargo fmt --all -- --check`
- rebuilt the simulator XCFramework and built the Swift SDK example

## Summary by CodeRabbit

* **New Features**
  * Added shielded transaction preflight checks showing account balance, reserved fees, maximum shieldable amount, availability, and advisory reasons.
  * Preflight checks work without signing, broadcasting, or coordinator access.
  * Added typed handling for insufficient shielded capacity errors across SDKs.

* **Bug Fixes**
  * Improved shielded input selection, fee reservation, minimum-balance handling, and arithmetic-overflow reporting.
  * Clarified address ordering and capacity calculations.

* **Tests**
  * Added coverage for preflight validation, capacity reporting, error mapping, overflow, dust handling, and input selection.

PR #4366, opened 2026-08-11
fix(platform-wallet): derive the shield fee reserve from the versioned fee

QuantumExplorer:

Targets #4360's branch. The preflight/planner introduced there computes `can_shield`, `max_shieldable_credits`, and the input-0 viability threshold from the fixed `SHIELDED_SHIELD_FEE_RESERVE_CREDITS = 1_000_000_000` reserve. That value is ~6× the fee the network actually charges a Type 15 shield (metered note/nullifier storage/processing + `compute_shielded_verification_fee(2)` ≈ 1.65e8 credits at current constants), which turns internal slop into wrong host-visible capacity:

- Any account whose funded addresses each hold ≤ 1e9 credits (0.01 DASH) reports `can_shield == false` with zero capacity, regardless of aggregate balance. Platform payment accounts accumulate many small per-payment addresses, so this shape is the norm, and the loss is unbounded.
- Max understates real capacity by ~8.3e8 credits, and after a Max shield input 0 retains `reserve − actual fee` ≈ 8.3e8 credits — below the 1e9 viability threshold — so every account that shields Max converges to a permanently unshieldable transparent residue.

The reserve's only real job is covering the execution-side fee that drive deducts from input 0's post-reallocation residue (`DeductFromInput(0)`). The structure gate (`Σ claims ≥ amount + fee`) is satisfied by the claimed fee `reserve_shield_fee_on_input_0` loads, which is already versioned-exact — an oversized unclaimed reserve contributes nothing there.

## What was done?

- Replaced the constant with `shield_fee_reserve_credits(platform_version)` = `2 × compute_minimum_shielded_fee(SHIELD_NUM_ACTIONS)` in `operations.rs`, next to the action-count constant it depends on. The 2× keeps one full fee of headroom for the GroveDB metering the client cannot reproduce, and the value now tracks fee-constant bumps instead of freezing (at current constants: 325,702,400 credits, ~0.0033 DASH).
- Threaded the reserve through `plan_shield_inputs` as a parameter, like the versioned `min_input_amount` / `max_address_inputs` the planner already takes from `sdk.version()`. `ShieldedShieldPreflight.fee_reserve_credits` and the FFI struct field now carry the versioned value; no FFI/Swift/Kotlin shape changes.
- Updated the planner tests to derive the reserve from `LATEST_PLATFORM_VERSION`, keeping every scenario parametric; the real-account regression test now guards its own shape (leading address must stay below the reserve).
- Extended the headroom test to bound the reserve from above (`≤ 4× fee`) so an oversized reserve can't silently reintroduce the capacity loss, and refreshed the stale `~1.2e8` fee comment (pre-bump value; it's ~1.63e8 at current constants).

## How Has This Been Tested?

- `cargo test -p platform-wallet --features shielded --lib` — 775 passed, including the 12 planner tests, 4 reserve tests, and the fetch-error mapping test.
- `cargo clippy -p platform-wallet -p platform-wallet-ffi --all-targets --all-features` and no-feature `cargo check` for both crates — clean.
- `cargo fmt --all`.

## Breaking Changes

None on-wire. The `SHIELDED_SHIELD_FEE_RESERVE_CREDITS` export (added on this branch, unreleased) is replaced by the `shield_fee_reserve_credits` function; hosts should keep reading `fee_reserve_credits` from the preflight, which is unchanged in shape.

## Checklist:

### DIFF
--- a/packages/rs-platform-wallet/src/lib.rs
+++ b/packages/rs-platform-wallet/src/lib.rs
@@ -80,8 +80,10 @@ pub use wallet::identity::{
 };
 pub use wallet::platform_wallet::PlatformWalletInfo;
 #[cfg(feature = "shielded")]
-pub use wallet::platform_wallet::{ShieldedShieldPreflight, SHIELDED_SHIELD_FEE_RESERVE_CREDITS};
+pub use wallet::platform_wallet::ShieldedShieldPreflight;
 pub use wallet::provider_key_at_index::{ProviderDerivedKey, ProviderKeyKind};
+#[cfg(feature = "shielded")]
+pub use wallet::shielded::operations::shield_fee_reserve_credits;
 pub use wallet::PlatformAddressTag;
 pub use wallet::PlatformWallet;
 
--- a/packages/rs-platform-wallet/src/wallet/mod.rs
+++ b/packages/rs-platform-wallet/src/wallet/mod.rs
@@ -24,12 +24,14 @@ pub use platform_addresses::{
     PerAccountPlatformAddressState, PerWalletPlatformAddressState, PlatformAddressTag,
     PlatformAddressWallet,
 };
+#[cfg(feature = "shielded")]
+pub use platform_wallet::ShieldedShieldPreflight;
 pub use platform_wallet::{
     PlatformWallet, PlatformWalletInfo, WalletId, WalletStateReadGuard, WalletStateWriteGuard,
 };
-#[cfg(feature = "shielded")]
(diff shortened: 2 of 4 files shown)


thepastaclaw:
🕓 Ready for review — next in queue (commit 0ba8231a)
_Queue position: 1/1_
_ETA: start ~08:10 UTC · complete ~08:21 UTC (median 11m across 30 recent reviews; 2 slots)_
_Queued 9m ago · Last checked: 2026-08-11 08:10 UTC_

llbartekll (approved):

PR #4357, opened 2026-08-10
fix(platform-wallet): preserve reported-consumed asset-lock recovery

llbartekll:
## Summary

- reconcile structured exact-outpoint already-consumed reports in the shared asset-lock orchestration layer
- apply the same conservative handling to identity registration, identity top-up, platform-address funding, and shielded funding
- reuse or obtain a ChainLock proof and durably persist `RecoveredFromChain` while returning stable code 24
- never infer that the requested Platform operation succeeded; unrelated and mismatched errors remain unchanged
- document code 24 as consumption unknown and make Swift/Kotlin invitation examples retain their status and in-flight marker
- rename the composite persistence capability to `ASSET_LOCK_RECONCILIATION` without changing capability bits or FFI/API types

## Tests

- `cargo test -p platform-wallet --features shielded --lib wallet::shielded::fund_from_asset_lock::tests`
- `cargo test -p platform-wallet --lib wallet::asset_lock::orchestration::tests`
- `cargo test -p platform-wallet-ffi --features shielded --lib persistence::tests`
- `cargo test -p platform-wallet-ffi --features shielded --lib error::tests`
- `cargo check -p platform-wallet --lib`
- `cargo check -p platform-wallet --features shielded --lib`
- `cargo check -p platform-wallet-storage`
- `python3 scripts/check_sdk_parity_manifest.py`
- `python3 -m unittest discover -s scripts/tests -p 'test_*.py'`
- `cargo fmt --all -- --check`
- `cargo clippy -p platform-wallet -p platform-wallet-ffi --features shielded --lib --tests -- -D warnings`

Focused Swift source parsing passed. The local Swift unit target could not import the generated `DashSDKFFI` module, and the local Kotlin Gradle test could not configure without an Android SDK; CI covers the repository integration environment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of already-consumed asset locks during funding, retries, and shielded sends.
  * Preserved matching asset-lock errors while surfacing unrelated failures normally.
  * Recovered matching locks with ChainLock proofs when Platform-side consumption remains unknown.
  * Improved persistence, rollback, and consistency across immediate and deferred storage modes.
  * Invitation reclaim attempts now remain unresolved rather than incorrectly indicating success or a third-party claim.

* **Documentation**
  * Clarified recovery states, proof retention, and Platform finality limitations.
  * Updated error descriptions for unconfirmed one-shot funding completion.
  * Added tracked asset-lock persistence capability reporting.

PR #4337, opened 2026-08-08
fix(platform-wallet): settle already-consumed asset locks and survive an ambiguous resume broadcast

HashEngineering:
## Issue being fixed or feature implemented

Two ways a completed or in-flight asset-lock top-up could never finish, both observed live on an Android testnet wallet (MO-998 / dashpay/dash-wallet#1520):

**1. Platform's "already completely used" verdict was dropped.** When a transition is rejected with `IdentityAssetLockTransactionOutPointAlreadyConsumedError` ("Asset lock transaction {txid} output {n} already completely used"), the credits it would have bought already landed — an earlier attempt succeeded and the client never learned. Nothing recorded that locally: `consume_asset_lock` was only ever called on the success path, so the lock stayed in the resumable set and every recovery pass re-submitted it for the same deterministic rejection, forever. A client that blocks new funding while a lock is unresolved could not buy credits at all until it special-cased the error *string* (which the Android wallet currently does — this PR is the typed root fix that lets that workaround be deleted).

**2. A `Built`-status resume aborted on an ambiguous re-broadcast.** The `Built` arm of `resume_asset_lock` propagated every broadcast error, including `MaybeSent`. For a lock stuck at `Built` whose transaction WAS broadcast (the app died between the send and the status advance), `MaybeSent` is the *expected* answer on every retry — DAPI classifies all failures that way — so the resume failed, the lock stayed `Built`, and the next pass repeated it. The top-up never completed.

## What was done?

- Both funded flows (`register_identity_with_funding`, `top_up_identity_with_funding`) classify the already-consumed rejection **from the typed consensus error** (`asset_lock_already_consumed_out_point`, reading the outpoint Platform itself names — never a message match), mark the lock `Consumed` via the existing `consume_asset_lock`, and return the same typed `AssetLockAlreadyConsumed` a resume of a consumed lock already raises — callers need one terminal case.
- The `Built` resume arm now distinguishes verdicts: only a definite `Rejected` stops the resume (nothing is on the network; the lock stays at `Built` for a later re-send). `MaybeSent` advances to `Broadcast` and proceeds to the proof wait — matching what the `Broadcast` arm already does with the identical signal.

## How Has This Been Tested?

- Two new regression tests assert the **status transition**, not just the error: an ambiguous re-broadcast must advance `Built → Broadcast` and reach the proof wait; a definite rejection must keep failing with the lock still resumable at `Built`.
[...]

PR #4367, opened 2026-08-11
fix(platform-wallet): survive an ambiguous re-broadcast when resuming a Built asset lock

QuantumExplorer:
Extracts the still-needed half of #4337 onto current `v4.2-dev` (authorship preserved — the commit is @HashEngineering's, adapted from #4337's head).

## Why #4337 was split

#4337 fixed two independent ways an asset-lock top-up could never finish. Its first half — classifying Platform's "already completely used" rejection and settling the tracked lock — is superseded by #4357 (merged 2026-08-11), whose shared `reconcile_asset_lock_submit_result` covers the same rejection across all four funded flows with outpoint binding and a safer policy: the unauthenticated verdict never tombstones the lock; it is preserved as `RecoveredFromChain` / consumption-unknown with a ChainLock proof, surfacing the same typed code 24.

This PR carries the second half, which #4357 does not touch.

## The remaining bug

The `Built` arm of `resume_asset_lock` propagated every broadcast error, including `MaybeSent`. For a lock stuck at `Built` whose transaction **was** broadcast — the app died between the send and the status advance — `MaybeSent` is the expected answer on every retry: the network already holds the tx, so each re-broadcast draws the same ambiguity. The resume failed, the lock stayed `Built`, every recovery pass repeated the same abort, and the top-up never completed. Observed live on an Android testnet wallet (the same incident that produced #4337).

## The fix

Only a definite `Rejected` (provably never entered the send pipeline) stops the resume. `MaybeSent` logs, advances the lock to `Broadcast`, and proceeds to the proof wait — exactly what the `Broadcast` arm already does with the identical signal. A genuinely un-broadcast tx still fails definitively and stays resumable at `Built`.

Two regression tests cover the ambiguous and definite branches, asserting the resulting lock status as well as the error (`built_resume_survives_an_ambiguous_rebroadcast_and_advances`, `built_resume_still_fails_on_a_definite_rejection`).

Note for reviewers: #4355 (open) touches the same `Built` arm to wait for SPV transport before this broadcast; the changes compose but whichever lands second needs a trivial rebase.

## Verification

- `cargo test -p platform-wallet --lib` — 654 passed, 0 failed (both new tests included)
- `cargo clippy -p platform-wallet --all-targets` — no new warnings (1 pre-existing on the unmodified base)
- `cargo fmt --all` — clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved asset lock recovery when a broadcast may have been sent but its outcome is uncertain.
  * Ambiguous broadcasts now continue processing while awaiting confirmation instead of stopping prematurely.
  * Definite broadcast rejections continue to fail safely without advancing the asset lock.

### DIFF
--- a/packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs
+++ b/packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs
@@ -4,7 +4,7 @@
 //! resolving status from wallet info, resuming interrupted locks,
 //! and re-deriving private keys.
 
-use crate::broadcaster::TransactionBroadcaster;
+use crate::broadcaster::{BroadcastError, TransactionBroadcaster};
 use std::time::Duration;
 
 use dashcore::Address as DashAddress;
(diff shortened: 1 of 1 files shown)


thepastaclaw:
🕓 Ready for review — 1 ahead in queue (commit 6d4bebd4)
_Queue position: 2/2 · 1 review active_
_ETA: start ~10:45 UTC · complete ~10:56 UTC (median 11m across 30 recent reviews; 2 slots)_
_Queued 56s ago · Last checked: 2026-08-11 10:40 UTC_

PR #4370, opened 2026-08-11
fix(platform-wallet): commit wallet events off the async runtime

romchornyi:

`run_wallet_event_adapter` called `commit_batch` — and through it `persister.store()` — **inline on the tokio worker driving it**. `store()` is synchronous and, for the SQLite backend, commits a real transaction per call. Its own trait docs (`traits.rs:200-206, 263-267`) already warn that a slow write blocks every other wallet accessor for its duration; what they do not say, because until now it was not true, is that it also blocks the runtime those accessors run on.

Found while investigating a user report that a restored wallet's transaction history appears only in large, minutes-apart jumps long after Core sync reports 100%.

**Evidence from a testnet restore of a 6663-transaction wallet** (52k-line session log):

- Drains coalesce into ever larger, ever rarer batches — `folded` 1 → 47 → 164 → **512** (the `ADAPTER_STORE_BATCH_LIMIT` ceiling), with gaps of **42s, 144s and finally 1109s** between them.
- The metrics tick covering the 512-event drain:
  ```
  12:30:39.889  wallet-event batch: folded=512 … synced_height_persisted=Some(2179999)
  12:30:40.490  workers=14 busy_ratio=1106.67 mean_poll_us=1397886
  12:30:41.491  workers=14 busy_ratio=0.0089   mean_poll_us=24
  ```
  A 1.4-second **mean** poll on a runtime that reads 24µs one second later.
- `Blocks: … last_activity: 549s` at the same moment — the SPV managers sharing that runtime were starved, not idle.
- The durable watermark topped out at height **2179999** against a chain tip of **2520064** and never caught up within the session.
- The home timeline only advances when a batch lands, so it showed roughly a third of the history ten minutes after core sync had finished.

The escalating `folded` counts are the symptom, not the cause: events pile up in the channel *because* the previous drain's synchronous `store()` is still holding a worker.

## What was done?

`commit_batch` now runs on `tokio::task::spawn_blocking`.

**The handle is awaited rather than raced against `cancel`.** A store that has started must be allowed to finish, and dropping a `spawn_blocking` handle does not stop the thread in any case. Shutdown is observed at the next `recv`, which is where the loop already handles it.

**`AdapterFaultState` and the freeze latch move behind `Arc<Mutex<..>>` / `Arc<AtomicBool>`** instead of being moved into the closure by value. This is the part worth reviewing: moving them would mean a panicking commit thread loses a wallet's frozen watermark — un-freezing a wallet whose verification had failed, which is the single outcome the fail-closed guard exists to prevent. The lock is uncontended by construction (this task is the only writer, and one drain commits at a time), so it carries state rather than arbitrating access.

**A panicking commit thread is now reported and its drain skipped**, rather than taking the adapter down with it.

### Not done here

Nothing about batch sizing, the channel, or `ADAPTER_STORE_BATCH_LIMIT`. With the blocking call off the runtime the coalescing behaves as designed; tuning it before re-measuring would be guessing.

## How Has This Been Tested?

```
cargo test -p platform-wallet --lib          # 662 passed
cargo clippy -p platform-wallet --all-targets  # clean
cargo fmt --check                            # clean
```

**Not covered:** no test reproduces the stall. Doing so needs a persister whose `store()` blocks for a controllable duration plus assertions on runtime poll latency — worth adding, but it would not have caught this class of bug by construction, only this instance of it. The change is behaviour-preserving for the commit itself: the same `commit_batch`, the same inputs, the same diagnostics line.

A before/after restore on device is the measurement that matters, and I have the "before" trace above to compare against.

## Breaking Changes

None. No public API changes; `PlatformWalletPersistence` implementors are unaffected (the trait already requires `Send + Sync`).

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved wallet transaction reliability when persistence operations fail or unexpectedly stop.
  * Affected wallets are marked as faulted while their progress remains safely frozen.
  * Processing continues for unaffected wallets and later record-bearing changes.
  * Improved recovery keeps wallet status accurate after interrupted operations and prevents one wallet’s failure from disrupting others.
  * Improved runtime responsiveness by preventing persistence work from blocking normal processing.
  * Added clearer outcomes for interrupted or unsuccessful wallet operations.

### DIFF
--- a/packages/rs-platform-wallet/src/changeset/core_bridge.rs
+++ b/packages/rs-platform-wallet/src/changeset/core_bridge.rs
@@ -384,7 +384,32 @@ async fn run_wallet_event_adapter<P>(
         // Captured before the batch moves into the closure: if the commit
         // thread panics, these are the wallets whose rows have an unknown fate
         // and whose watermark must therefore be frozen.
-        let batch_wallet_ids: Vec<WalletId> = batch.keys().copied().collect();
+        //
+        // Only wallets `commit_batch` can actually call `store()` for. A wallet
+        // contributes to the batch whenever it produced an event, including
+        // events that project to nothing — a `TransactionInstantLocked` that is
+        // ignored because the transaction is already chain-locked, a
+        // `SyncHeightAdvanced` for an unknown wallet. Those hit the
+        // `is_empty_no_records()` skip and never reach a store, so a sibling's
+        // panic says nothing about them; freezing them would strip a healthy
+        // wallet's watermark for the rest of the session over someone else's
+        // bad batch.
+        //
+        // Deliberately conservative in one direction: `commit_batch` re-tests
+        // emptiness AFTER `freeze_synced_height_if_faulted` has stripped a
+        // faulted wallet's watermark, so a batch that looks non-empty here can
+        // still be skipped there. That wallet is already faulted, so freezing
+        // it again costs nothing — whereas the reverse error, omitting a wallet
+        // whose store did run, would leave a watermark free to advance past
+        // rows nobody can account for.
+        let batch_wallet_ids: Vec<WalletId> = batch
+            .iter()
+            .filter(|(_, wallet_batch)| {
+                !wallet_batch.core.is_empty_no_records()
+                    || !Merge::is_empty(&wallet_batch.asset_locks)
+            })
+            .map(|(wallet_id, _)| *wallet_id)
+            .collect();
         // Filled by `commit_batch` as each wallet's `store()` returns. Lives
         // out here so a panicking commit thread cannot take it down with it:
         // what it holds is the difference between "this wallet's rows are
@@ -466,6 +491,29 @@ async fn run_wallet_event_adapter<P>(
                     for wallet_id in &unsettled {
                         fault.fault_wallet(*wallet_id, &sync_fault);
                     }
+                }
+                // Same one-shot marker the rejection arm emits, and via the same
+                // `log` facade, because this freeze is indistinguishable from
+                // that one as far as the host is concerned: the latch is up and
+                // a rescan is pending. Leaving it to `tracing` alone would hide
+                // a panic-induced freeze from logcat entirely, and — worse —
+                // leave `freeze_logged` clear, so a later rejected store would
+                // emit the supposedly one-shot line as though it were the first
+                // fault of the session.
+                //
+                // `swap` rather than load-then-store: a store panicking inside
+                // the same blocking task can race the flag the task wrote back
+                // on its way out, and emitting this line twice is a worse
+                // outcome than the branch reading its own write.
+                if !unsettled.is_empty() && !freeze_logged.swap(true, Ordering::Relaxed) {
+                    log::error!(
+                        "SYNC WATERMARK FROZEN: the wallet-event commit thread panicked ({}); \
+                         {} wallet(s) whose rows have an unknown outcome are now held so the \
+                         next scan re-persists them (dashpay/platform#4370). \
+                         syncFaultDetected() is latched.",
+                        join_error,
+                        unsettled.len()
+                    );
                 }
                 tracing::error!(
                     error = %join_error,
@@ -2418,6 +2466,97 @@ mod tests {
     /// let the NEXT batch persist a higher `synced_height` for a wallet whose
     /// earlier rows may never have landed, which is exactly the hole
     /// dashpay/platform#4069 closed.
+    /// The commit must not run on a Tokio worker.
+    ///
+    /// This is the behaviour the `spawn_blocking` move exists for, and no
+    /// other test covers it: the outcome-only assertions elsewhere all still
+    /// pass with `commit_batch` called inline, because a blocking store on a
+    /// multi-worker runtime merely steals one worker and the rest of the test
+    /// proceeds. Pinning the runtime to a single worker makes the difference
+    /// observable — inline, that one worker sits inside `store()` and nothing
+    /// else on the runtime can advance.
+    ///
+    /// The release is driven from a plain `std::thread`, and the verdict is
+    /// read there too, because a regression parks the only worker: the test
+    /// body cannot run, and neither can a `tokio::time::timeout` — the timer
+    /// needs the same worker to fire. An in-runtime deadline would therefore
+    /// hang CI instead of failing it, which is exactly what the bounded waits
+    /// elsewhere in this file exist to avoid.
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn a_blocking_store_does_not_park_the_runtime() {
+        let wallet_id = [0xB1u8; 32];
+        let (tx, rx) = unbounded_channel::<WalletEvent>();
+
+        let (obs_tx, mut obs_rx) = unbounded_channel();
+        let persister = Arc::new(ProbePersister::new(obs_tx));
+        let (unblock, blocked) = persister.block_next();
+        let sync_fault = Arc::new(AtomicBool::new(false));
+        let cancel = CancellationToken::new();
+        let handle = tokio::spawn(run_wallet_event_adapter(
+            test_manager(),
+            Arc::clone(&persister),
+            rx,
+            Arc::clone(&sync_fault),
+            cancel.clone(),
+        ));
+
+        // Set before the store blocks, so it is waiting on the runtime rather
+        // than racing to be scheduled: whether it ran is then purely a question
+        // of the worker being free.
+        let progressed = Arc::new(AtomicBool::new(false));
+        let flag = Arc::clone(&progressed);
+        tokio::spawn(async move {
+            flag.store(true, Ordering::Relaxed);
+        });
+
+        tx.send(block_processed_event(wallet_id, 10)).unwrap();
+
+        // Outside the runtime entirely: waits for the store to park, records
+        // whether the unrelated task got to run while it was parked, then
+        // releases it so the test always terminates either way.
+        let observer_progressed = Arc::clone(&progressed);
+        let observer_blocked = Arc::clone(&blocked);
+        let watcher = std::thread::spawn(move || {
+            let deadline = std::time::Instant::now() + std::time::Duration::from_secs(10);
+            while !observer_blocked.load(Ordering::Relaxed) {
+                if std::time::Instant::now() > deadline {
+                    return Err("store never entered its block");
+                }
+                std::thread::sleep(std::time::Duration::from_millis(10));
+            }
+            // Give the runtime a moment it does not need when it is healthy;
+            // parked, no amount of waiting would help.
+            std::thread::sleep(std::time::Duration::from_millis(200));
+            let ran_while_blocked = observer_progressed.load(Ordering::Relaxed);
+            drop(unblock);
+            Ok(ran_while_blocked)
+        });
+
+        let observed = obs_rx
+            .recv()
+            .await
+            .expect("the released store should report");
+        assert_eq!(observed.wallet_id, wallet_id);
+
+        let ran_while_blocked = watcher
+            .join()
+            .expect("watcher thread should not panic")
+            .expect("store never entered its block");
+        assert!(
+            ran_while_blocked,
+            "an unrelated task must make progress while the commit blocks — \
+             the commit is running on a runtime worker instead of a blocking thread"
+        );
+        assert!(
+            !sync_fault.load(Ordering::Relaxed),
+            "a slow store is not a fault"
+        );
+
+        cancel.cancel();
+        drop(tx);
+        let _ = handle.await;
+    }
+
     #[tokio::test]
     async fn a_panicking_commit_freezes_the_batch_wallets() {
         let wallet_id = [0xEEu8; 32];
--- a/packages/rs-platform-wallet/src/changeset/core_bridge.rs
+++ b/packages/rs-platform-wallet/src/changeset/core_bridge.rs
@@ -34,7 +34,7 @@
 
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 
 use dashcore::blockdata::transaction::{txout::TxOut, OutPoint};
 use dashcore::ScriptBuf;
@@ -294,10 +294,16 @@ async fn run_wallet_event_adapter<P>(
     P: PlatformWalletPersistence + 'static,
 {
     tracing::debug!("wallet-event adapter task started");
-    let mut fault = AdapterFaultState::default();
+    // Both live behind handles rather than as locals because the commit runs
+    // on a blocking thread (see the `spawn_blocking` below) and has to be able
+    // to carry its state across drains. Moving them into the closure by value
+    // would lose a wallet's frozen watermark if that thread ever panicked —
+    // un-freezing a wallet that failed verification is the one outcome the
+    // fail-closed guard exists to prevent.
+    let fault = Arc::new(Mutex::new(AdapterFaultState::default()));
     // One-shot latch so the hard "watermark frozen" line hits logcat exactly
     // once per session rather than once per faulted batch.
-    let mut freeze_logged = false;
+    let freeze_logged = Arc::new(AtomicBool::new(false));
 
     loop {
         // Block for the first event of a batch. Everything already sitting in
@@ -360,14 +366,60 @@ async fn run_wallet_event_adapter<P>(
         // Commit the folded batch. The channel is lossless, so the only way a
         // watermark is held back is a rejected `store()` (the fail-closed
         // backstop inside `commit_batch`).
-        let diag = commit_batch(
-            &*persister,
-            batch,
-            folded,
-            &mut fault,
-            &sync_fault,
-            &mut freeze_logged,
-        );
+        // Commit on a blocking thread, never on the async worker.
+        //
+        // `store()` is synchronous and, for the SQLite backend, commits a real
+        // transaction per call — its own docs warn that a slow write blocks
+        // every other wallet accessor for its duration. Called inline here it
+        // blocked a tokio worker instead: a field restore showed one drain of
+        // 512 folded events park the runtime long enough for the metrics tick
+        // covering it to report a 1.4s mean poll, with the whole sync stalled
+        // for minutes at a time and the durable watermark left hundreds of
+        // thousands of blocks behind the chain tip.
+        //
+        // The handle is awaited rather than raced against `cancel`: a store
+        // that has started must be allowed to finish, and dropping the handle
+        // would not stop the thread anyway. Shutdown is observed at the next
+        // `recv` instead.
+        let persister_for_commit = Arc::clone(&persister);
+        let sync_fault_for_commit = Arc::clone(&sync_fault);
+        let fault_for_commit = Arc::clone(&fault);
+        let freeze_for_commit = Arc::clone(&freeze_logged);
+        let committed = tokio::task::spawn_blocking(move || {
+            // The lock is uncontended by construction — this task is the only
+            // writer, and one drain commits at a time — so it never blocks;
+            // it exists to carry the state, not to arbitrate.
+            let mut fault = fault_for_commit
+                .lock()
+                .unwrap_or_else(|poisoned| poisoned.into_inner());
+            let mut freeze_logged = freeze_for_commit.load(Ordering::Relaxed);
+            let diag = commit_batch(
+                &*persister_for_commit,
+                batch,
+                folded,
+                &mut fault,
+                &sync_fault_for_commit,
+                &mut freeze_logged,
+            );
+            freeze_for_commit.store(freeze_logged, Ordering::Relaxed);
+            diag
+        })
+        .await;
+
+        let diag = match committed {
+            Ok(diag) => diag,
+            // The commit thread panicked. The fault state survives (it lives
+            // behind the handle above), but this batch's outcome is unknown,
+            // so it is reported rather than silently folded into the next one.
+            Err(join_error) => {
+                tracing::error!(
+                    error = %join_error,
+                    folded,
+                    "wallet-event commit thread failed; batch outcome unknown"
+                );
+                continue;
+            }
+        };
 
         // One structured line per drain via the `log` facade so a tester
         // logcat is unambiguous about whether the watermark is advancing.
--- a/packages/rs-platform-wallet/src/changeset/core_bridge.rs
+++ b/packages/rs-platform-wallet/src/changeset/core_bridge.rs
@@ -91,11 +91,17 @@ const ADAPTER_STORE_BATCH_LIMIT: usize = 512;
 /// Session fault state for the durable-watermark guard
 /// (dashpay/platform#4069).
 ///
-/// Now that the persistence channel is a lossless unbounded `mpsc`, the only
-/// remaining fault trigger is a **`store()` rejection**, which carries a
-/// `wallet_id`, so only THAT wallet's watermark freezes. A sibling wallet
-/// whose rows are still landing atomically keeps advancing — freezing it too
-/// would force a redundant rescan of a wallet that never lost a row.
+/// Now that the persistence channel is a lossless unbounded `mpsc`, two things
+/// fault a wallet, and both name the wallets they hit, so a sibling whose rows
+/// are still landing atomically keeps advancing — freezing it too would force a
+/// redundant rescan of a wallet that never lost a row.
+///
+/// - A **`store()` rejection**, which carries a `wallet_id`: that wallet's rows
+///   are known not to be on disk.
+/// - A **panic in the blocking commit thread**, which faults every wallet with
+///   something to persist that is absent from `settled` — the one that panicked
+///   plus every wallet the loop never reached. Their outcome is unknown rather
+///   than known-bad, and unknown must fail closed the same way.
 ///
 /// The old global (`broadcast::Lagged`) latch is gone: the unbounded channel
 /// can never `Lagged`, so there is no more "dropped events of unknown wallet"
@@ -103,7 +109,8 @@ const ADAPTER_STORE_BATCH_LIMIT: usize = 512;
 /// fail-closed backstop — in a healthy run it never fires.
 #[derive(Default)]
 struct AdapterFaultState {
-    /// Set by a `store()` rejection: freezes only the named wallet.
+    /// Set by a `store()` rejection, or by the panic-recovery branch for a
+    /// wallet whose commit outcome is unknown: freezes only the named wallets.
     per_wallet: HashMap<WalletId, bool>,
 }
 
@@ -113,8 +120,9 @@ impl AdapterFaultState {
         self.per_wallet.get(wallet_id).copied().unwrap_or(false)
     }
 
-    /// Fault a single wallet after its `store()` was rejected, and raise
-    /// the host-visible hard-fault signal.
+    /// Fault a single wallet — after its `store()` was rejected, or after a
+    /// commit panic left its outcome unknown — and raise the host-visible
+    /// hard-fault signal.
     fn fault_wallet(&mut self, wallet_id: WalletId, hard_signal: &AtomicBool) {
         self.per_wallet.insert(wallet_id, true);
         hard_signal.store(true, Ordering::Relaxed);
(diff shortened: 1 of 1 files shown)


thepastaclaw:
✅ Final review complete — no blockers (commit 75ac76c7)

packages/rs-platform-wallet/src/changeset/core_bridge.rs
```diff
@@ -360,14 +366,60 @@ async fn run_wallet_event_adapter<P>(
...
+            // The commit thread panicked. The fault state survives (it lives
+            // behind the handle above), but this batch's outcome is unknown,
+            // so it is reported rather than silently folded into the next one.
+            Err(join_error) => {
+                tracing::error!(
+                    error = %join_error,
+                    folded,
+                    "wallet-event commit thread failed; batch outcome unknown"
+                );
+                continue;
```
thepastaclaw, line 539: **🔴 Blocking: Continuing after a commit panic can advance the watermark past lost rows**

`commit_batch` consumes the folded batch and calls `store()` for each wallet. If `store()` panics before its outcome is known, unwinding bypasses the `Err` arm that calls `fault_wallet`, drops the remainder of the consumed batch, and returns a `JoinError`. This branch then continues with an unaffected fault state, so a later event for the same wallet can successfully persist a higher `synced_height` even though rows from the panicked batch may be absent. That violates the adapter's fail-closed invariant; before this PR, the panic terminated the adapter and prevented later watermark advancement. Stop the adapter and latch `sync_fault`, or capture every batch wallet ID before moving the batch and fault all of them before continuing.

```suggestion
Err(join_error) => {
    tracing::error!(
        error = %join_error,
        folded,
        "wallet-event commit thread failed; stopping adapter because batch outcome is unknown"
    );
    sync_fault.store(true, Ordering::Relaxed);
    break;
}
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/changeset/core_bridge.rs
```diff
@@ -360,14 +366,60 @@ async fn run_wallet_event_adapter<P>(
...
+                batch,
+                folded,
+                &mut fault,
+                &sync_fault_for_commit,
+                &mut freeze_logged,
+            );
+            freeze_for_commit.store(freeze_logged, Ordering::Relaxed);
+            diag
+        })
+        .await;
```
thepastaclaw, line 459: **🟡 Suggestion: The off-runtime persistence boundary has no regression test**

The existing `ProbePersister` returns immediately and checks only persistence outcomes. Those tests still pass if `commit_batch` is moved back inline onto the Tokio worker, so they do not protect the primary behavior introduced by this PR. Add a controllably blocking persister and run the adapter on a current-thread or single-worker runtime, then verify that another future makes progress while `store()` remains blocked. The fixture should also cover a panicking store and assert that no later store for the affected wallet carries `synced_height`.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

Moving synchronous persistence into `spawn_blocking` correctly prevents wallet commits from parking Tokio workers, but the new panic branch continues after consuming a batch whose persistence outcome is unknown, allowing a later watermark to advance past missing rows. The adapter must fail closed after a commit panic, and the off-runtime boundary should have deterministic regression coverage.
Source: reviewer backend gpt-5.6-sol (Codex general and Rust-quality lanes); final verifier backend gpt-5.6-sol (Codex). openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

### Review provenance

<details>

```

- [BLOCKING] packages/rs-platform-wallet/src/changeset/core_bridge.rs:409-420: Continuing after a commit panic can advance the watermark past lost rows
  `commit_batch` consumes the folded batch and calls `store()` for each wallet. If `store()` panics before its outcome is known, unwinding bypasses the `Err` arm that calls `fault_wallet`, drops the remainder of the consumed batch, and returns a `JoinError`. This branch then continues with an unaffected fault state, so a later event for the same wallet can successfully persist a higher `synced_height` even though rows from the panicked batch may be absent. That violates the adapter's fail-closed invariant; before this PR, the panic terminated the adapter and prevented later watermark advancement. Stop the adapter and latch `sync_fault`, or capture every batch wallet ID before moving the batch and fault all of them before continuing.
- [SUGGESTION] packages/rs-platform-wallet/src/changeset/core_bridge.rs:388-407: The off-runtime persistence boundary has no regression test
  The existing `ProbePersister` returns immediately and checks only persistence outcomes. Those tests still pass if `commit_batch` is moved back inline onto the Tokio worker, so they do not protect the primary behavior introduced by this PR. Add a controllably blocking persister and run the adapter on a current-thread or single-worker runtime, then verify that another future makes progress while `store()` remains blocked. The fixture should also cover a panicking store and assert that no later store for the affected wallet carries `synced_height`.
```
</details>

packages/rs-platform-wallet/src/changeset/core_bridge.rs
romchornyi: Fixed in `0499b9c9` — and you are right that this PR introduced the hole rather than merely failing to close it.

Before the move, a panic inside `store()` unwound the adapter task itself. Violent, but safe in one specific way: the writer was gone, so nothing could persist a higher `synced_height` afterwards. `spawn_blocking` converts the same panic into a recoverable `JoinError`, and my branch logged it and continued with the fault state untouched — so the next batch could advance the watermark past rows whose fate is unknown. Exactly what #4069 closed.

The batch's wallet ids are now captured before the batch moves into the closure, and a `JoinError` faults every one of them.

**On stopping the adapter versus faulting per wallet** — I took the per-wallet route rather than the `break` you suggested. Reasoning, and I am happy to be overruled: a rejected `store()` already freezes only the wallet it affected and lets its siblings keep syncing, and stopping the adapter would freeze every wallet on the manager over one wallet's bad batch. It would also undo the reason the commit moved off the runtime — one bad batch would again take down the writer for the whole session, just via a different mechanism. `sync_fault` is latched either way, so the host still sees the hard fault. If you would rather have the harder guarantee, say so and I will switch it to `break`.

The test covers all three halves of the contract: the hard-fault signal is raised, the adapter survives, and no later store for that wallet carries a `synced_height`.

On the second point — the off-runtime boundary having no regression coverage — this is a partial answer, not a complete one. `ProbePersister` gained a `panic_next` mode, which covers the panicking-store half you asked for. What it does not yet cover is the blocking half: a controllably-blocking persister on a single-worker runtime, asserting another future makes progress while `store()` is held. That is the test that would actually fail if someone moved `commit_batch` back inline, and it is worth having. I would rather add it as a follow-up than bolt a timing-sensitive fixture on at the end of this PR — but if you want it here before merge, I will write it.

One note on the test's shape, since it looks over-engineered otherwise: the wait for `sync_fault` is bounded by an explicit timeout. My first version spun unbounded, which meant a regression *hung* CI instead of failing it. With the fault path removed the test now fails in 5s with "a panicked commit must raise the hard-fault signal" — verified by actually removing it.

```
cargo test -p platform-wallet --lib   # 663 passed
cargo clippy --all-targets + cargo fmt --check   # clean
```

packages/rs-platform-wallet/src/changeset/core_bridge.rs
```diff
@@ -360,14 +366,84 @@ async fn run_wallet_event_adapter<P>(
...
+            // adapter keeps the existing design: a wallet whose commit is in
+            // doubt freezes, its siblings keep syncing.
+            Err(join_error) => {
+                {
+                    let mut fault = fault
+                        .lock()
+                        .unwrap_or_else(|poisoned| poisoned.into_inner());
+                    for wallet_id in &batch_wallet_ids {
+                        fault.fault_wallet(*wallet_id, &sync_fault);
+                    }
```
thepastaclaw: **🟡 Suggestion: Do not freeze wallets whose stores completed before the panic**

`commit_batch` processes the `BTreeMap` serially, but the `JoinError` branch faults every wallet in the drain. If wallet A's `store()` returns `Ok` and wallet B's later store panics, A's outcome is already known and its rows were accepted, yet A is marked faulted and all of its later `synced_height` updates are stripped for the rest of the manager session. Track completed wallet IDs across the blocking boundary, or isolate panic handling per wallet, so recovery faults only the panicking wallet and wallets that were not attempted after it. Add a multi-wallet test with a successful wallet ordered before the panicking wallet.

thepastaclaw (commented):

The current head correctly restores the fail-closed watermark invariant after a commit-thread panic, so the prior blocking issue is fixed. Two non-blocking gaps remain: the primary off-runtime behavior lacks a direct progress regression test, and panic recovery unnecessarily freezes wallets whose stores already completed successfully.
Source: reviewer backend model gpt-5.6-sol; final verifier backend model gpt-5.6-sol. openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

### Review provenance

🟡 1 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet/src/changeset/core_bridge.rs:433-435: Do not freeze wallets whose stores completed before the panic
  `commit_batch` processes the `BTreeMap` serially, but the `JoinError` branch faults every wallet in the drain. If wallet A's `store()` returns `Ok` and wallet B's later store panics, A's outcome is already known and its rows were accepted, yet A is marked faulted and all of its later `synced_height` updates are stripped for the rest of the manager session. Track completed wallet IDs across the blocking boundary, or isolate panic handling per wallet, so recovery faults only the panicking wallet and wallets that were not attempted after it. Add a multi-wallet test with a successful wallet ordered before the panicking wallet.
- [SUGGESTION] packages/rs-platform-wallet/src/changeset/core_bridge.rs:392-411: The off-runtime persistence boundary has no regression test
  (existing thread: https://github.com/dashpay/platform/pull/4370#discussion_r3758900980)
  The new panic test verifies panic isolation and fail-closed recovery, but every non-panicking `ProbePersister::store()` still returns immediately. It therefore does not directly enforce the PR's primary guarantee that a slow synchronous store cannot park the Tokio worker; an inline implementation with equivalent panic isolation would still pass. Add a controllably blocking persister and use a current-thread or single-worker runtime with an external watchdog/release mechanism, then assert that an unrelated future makes progress before the store is released.
```
</details>

packages/rs-platform-wallet/src/changeset/core_bridge.rs
thepastaclaw: Resolved in `0499b9c` — _Continuing after a commit panic can advance the watermark past lost rows_ no longer present.

packages/rs-platform-wallet/src/changeset/core_bridge.rs
```diff
@@ -360,14 +366,118 @@ async fn run_wallet_event_adapter<P>(
...
+        // `recv` instead.
+        // Captured before the batch moves into the closure: if the commit
+        // thread panics, these are the wallets whose rows have an unknown fate
+        // and whose watermark must therefore be frozen.
+        let batch_wallet_ids: Vec<WalletId> = batch.keys().copied().collect();
+        // Filled by `commit_batch` as each wallet's `store()` returns. Lives
+        // out here so a panicking commit thread cannot take it down with it:
+        // what it holds is the difference between "this wallet's rows are
+        // accounted for" and "nobody knows".
+        let settled: Arc<Mutex<Vec<WalletId>>> = Arc::new(Mutex::new(Vec::new()));
```
thepastaclaw, line 429: **🟡 Suggestion: Exclude no-op wallets from panic faulting**

`batch_wallet_ids` includes every wallet that contributed an event, even when its projected `WalletBatch` is empty and `commit_batch` consequently skips `store()` at lines 555-559. This occurs, for example, when `TransactionInstantLocked` is intentionally ignored because the transaction is already chain-locked. If another wallet's store panics, an empty wallet does not appear in `settled`, regardless of whether it was skipped before the panic or ordered after it, so the panic handler freezes that healthy wallet's watermark for the remainder of the session even though it had no rows to persist. Build the panic-recovery candidate list from batches that contain persistable changes.

```suggestion
let batch_wallet_ids: Vec<WalletId> = batch
    .iter()
    .filter_map(|(wallet_id, wallet_batch)| {
        (!wallet_batch.core.is_empty_no_records()
            || !Merge::is_empty(&wallet_batch.asset_locks))
        .then_some(*wallet_id)
    })
    .collect();
```

<sub>source: ['codex']</sub>

thepastaclaw (commented):

The main off-runtime persistence change is sound, and the current head fixes both previously verified gaps with direct regression coverage and per-wallet completion tracking. Two non-blocking panic-recovery issues remain: projected no-op wallets can be frozen despite having nothing to persist, and panic-induced freezes bypass the documented one-shot logcat fault marker.

### Review provenance

🟡 2 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet/src/changeset/core_bridge.rs:384-392: Exclude no-op wallets from panic faulting
  `batch_wallet_ids` includes every wallet that contributed an event, even when its projected `WalletBatch` is empty and `commit_batch` consequently skips `store()` at lines 555-559. This occurs, for example, when `TransactionInstantLocked` is intentionally ignored because the transaction is already chain-locked. If another wallet's store panics, an empty wallet does not appear in `settled`, regardless of whether it was skipped before the panic or ordered after it, so the panic handler freezes that healthy wallet's watermark for the remainder of the session even though it had no rows to persist. Build the panic-recovery candidate list from batches that contain persistable changes.
- [SUGGESTION] packages/rs-platform-wallet/src/changeset/core_bridge.rs:462-478: Emit the one-shot fault marker for panic-induced freezes
  The panic branch calls `fault_wallet`, which raises the host-visible hard-fault latch, but reports the freeze only through `tracing::error!` and never updates `freeze_logged`. This contradicts the adapter's documented observability contract at lines 282-286: `SYNC WATERMARK FROZEN` deliberately uses the `log` facade because `tracing` may not reach Android logcat. A commit panic can therefore freeze watermarks without the operational marker, and a later rejected store can emit that supposedly one-shot marker as though it were the first fault. Route rejection and panic recovery through a shared atomic one-shot logging helper so the latch is updated immediately, including when a later store in the same blocking task panics.
```
</details>

packages/rs-platform-wallet/src/changeset/core_bridge.rs
```diff
@@ -360,14 +366,118 @@ async fn run_wallet_event_adapter<P>(
...
+                }
+                tracing::error!(
+                    error = %join_error,
+                    folded,
+                    settled = settled_ids.len(),
+                    frozen = unsettled.len(),
+                    "wallet-event commit thread failed; freezing the wallets whose \
+                     rows have an unknown outcome"
+                );
+                continue;
```
thepastaclaw, line 539: **🟡 Suggestion: Emit the one-shot fault marker for panic-induced freezes**

The panic branch calls `fault_wallet`, which raises the host-visible hard-fault latch, but reports the freeze only through `tracing::error!` and never updates `freeze_logged`. This contradicts the adapter's documented observability contract at lines 282-286: `SYNC WATERMARK FROZEN` deliberately uses the `log` facade because `tracing` may not reach Android logcat. A commit panic can therefore freeze watermarks without the operational marker, and a later rejected store can emit that supposedly one-shot marker as though it were the first fault. Route rejection and panic recovery through a shared atomic one-shot logging helper so the latch is updated immediately, including when a later store in the same blocking task panics.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/changeset/core_bridge.rs
thepastaclaw: Resolved in `0499b9c` — _The off-runtime persistence boundary has no regression test_ no longer present.

packages/rs-platform-wallet/src/changeset/core_bridge.rs
thepastaclaw: Resolved in `4e5a193` — _Do not freeze wallets whose stores completed before the panic_ no longer present.

packages/rs-platform-wallet/src/changeset/core_bridge.rs
```diff
@@ -2253,6 +2455,334 @@ mod tests {
...
+                }
+                std::thread::sleep(std::time::Duration::from_millis(10));
+            }
+            // Give the runtime a moment it does not need when it is healthy;
+            // parked, no amount of waiting would help.
+            std::thread::sleep(std::time::Duration::from_millis(200));
+            let ran_while_blocked = observer_progressed.load(Ordering::Relaxed);
+            drop(unblock);
+            Ok(ran_while_blocked)
+        });
```
thepastaclaw: **🟡 Suggestion: Remove the race-prone duplicate blocking test**

This test schedules the `progressed` task before the adapter has entered the blocked store. When the test future first yields, Tokio can poll that task before the adapter, setting the flag before any blocking occurs; the test can therefore pass even if `commit_batch` is moved back inline. Its fixed 200 ms observation window can also fail on a loaded host. The stronger `a_blocked_store_does_not_park_the_runtime` test at lines 2638-2701 first observes the blocked store from outside the runtime and only then schedules its sentinel with a five-second deadline. Remove this duplicate and retain the later deterministic test.

<sub>source: ['codex']</sub>

thepastaclaw (commented):

The off-runtime commit boundary and fail-closed panic recovery are fundamentally sound, including the filtering of no-op and already-settled wallets. Three non-blocking issues remain: a rejection followed by a panic can emit the documented one-shot freeze marker twice, several fault-state comments omit panic-induced faults, and one blocking regression test is race-prone and duplicates a stronger test.
Source: reviewers gpt-5.6-sol (general and rust-quality); final verifier gpt-5.6-sol. openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

### Review provenance

🟡 2 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet/src/changeset/core_bridge.rs:94-117: Document commit panics as a fault trigger
  These invariant comments say a rejected `store()` is the only remaining fault trigger and that `fault_wallet` is called only after rejection, but the PR now deliberately faults persistable wallets whose outcomes are unknown after a commit panic. The same stale assertion appears in the adapter documentation around line 269, the commit-loop comment around line 366, and the watermark-guard comment around line 674. Update the documentation to distinguish rejected stores, which fault one known wallet, from commit panics, which fault each persistable wallet whose store outcome was not settled.
- [SUGGESTION] packages/rs-platform-wallet/src/changeset/core_bridge.rs:2503-2533: Remove the race-prone duplicate blocking test
  This test schedules the `progressed` task before the adapter has entered the blocked store. When the test future first yields, Tokio can poll that task before the adapter, setting the flag before any blocking occurs; the test can therefore pass even if `commit_batch` is moved back inline. Its fixed 200 ms observation window can also fail on a loaded host. The stronger `a_blocked_store_does_not_park_the_runtime` test at lines 2638-2701 first observes the blocked store from outside the runtime and only then schedules its sentinel with a five-second deadline. Remove this duplicate and retain the later deterministic test.
- [SUGGESTION] packages/rs-platform-wallet/src/changeset/core_bridge.rs:430-443: Emit the one-shot fault marker for panic-induced freezes
  (existing thread: https://github.com/dashpay/platform/pull/4370#discussion_r3779128332)
  The panic branch now emits the log-facade marker, but the rejection and panic paths do not spend the same atomic latch immediately. `commit_batch` copies `freeze_logged` into a local `bool`; if wallet A's `store()` returns `Err`, lines 651-659 emit `SYNC WATERMARK FROZEN` and update only that local value. If a later wallet's store panics in the same drain, unwinding skips the atomic write-back at line 443, so the `swap` at line 508 observes `false` and emits the supposedly one-shot marker again. Pass the shared atomic latch into `commit_batch` and use one atomic compare-and-set or `swap` helper from both the rejection and panic paths so the first emission is recorded even if a later store unwinds the closure.
```
</details>

packages/rs-platform-wallet/src/changeset/core_bridge.rs
thepastaclaw: Resolved in `4354b16` — _Exclude no-op wallets from panic faulting_ no longer present.

packages/rs-platform-wallet/src/changeset/core_bridge.rs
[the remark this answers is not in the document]
thepastaclaw: This is no longer applicable at the current head. Panic recovery now captures persistable batch wallet IDs, excludes stores whose outcomes settled, and faults every remaining wallet at lines 405-493 before continuing.

thepastaclaw (commented):

The off-runtime commit boundary and fail-closed panic recovery are sound: panic handling now excludes no-op and settled wallets, faults only persistable wallets with unknown outcomes, and shares the one-shot freeze latch with store-rejection handling. One in-scope documentation suggestion remains because several invariant comments still describe store rejection as the only fault trigger even though commit panics now fault unsettled wallets.
Source: reviewers gpt-5.6-sol (general, rust-quality, security-auditor); final verifier gpt-5.6-sol. openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

### Review provenance

🟡 1 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet/src/changeset/core_bridge.rs:94-117: Document commit panics as a fault trigger
  These invariant comments still state that a rejected `store()` is the only remaining fault trigger and that `fault_wallet` is called only after rejection. The adapter now also faults every persistable wallet whose outcome remains unknown after the blocking commit thread panics. The same stale invariant appears in the adapter documentation at lines 269-280, the commit-loop comment at lines 366-368, the watermark-guard documentation at lines 672-681, and the test comment at lines 2806-2807. Update these comments to distinguish a returned store error, which faults the wallet with a known rejection, from a commit panic, which faults each persistable wallet absent from `settled` because its outcome is unknown.
```
</details>

packages/rs-platform-wallet/src/changeset/core_bridge.rs
thepastaclaw: Resolved in `4354b16` — _Emit the one-shot fault marker for panic-induced freezes_ no longer present.

packages/rs-platform-wallet/src/changeset/core_bridge.rs
thepastaclaw: Resolved in `b8cb023` — _Remove the race-prone duplicate blocking test_ no longer present.

packages/rs-platform-wallet/src/changeset/core_bridge.rs
```diff
@@ -2253,6 +2468,339 @@ mod tests {
...
+    /// `commit_batch` walks the batch serially (a `BTreeMap`, so in wallet-id
+    /// order). If an earlier wallet's `store()` returned and a later one
+    /// panics, the earlier wallet's rows are on disk and its watermark is
+    /// safe — freezing it would strip its `synced_height` for the rest of the
+    /// session over a sibling's bad batch.
+    ///
+    /// Guards the fix for the first version of the panic handler, which
+    /// faulted every wallet in the drain.
+    #[tokio::test]
+    async fn a_panicking_commit_spares_the_wallets_it_already_stored() {
```
thepastaclaw: **🟡 Suggestion: Cover wallets the commit never reaches after a panic**

The panic tests cover the wallet whose `store()` panics and a wallet successfully settled before that panic, but none includes a persistable wallet ordered after the panicking wallet. That case exercises the other load-bearing half of `batch_wallet_ids - settled`: unwinding drops its consumed changes before `store()` is attempted, so panic recovery must fault it before a later watermark can advance. A regression that faults only the directly panicking wallet would pass the current tests. Add a third wallet whose ID sorts after `doomed`, include it in the same drain, and verify that a later record-bearing store for it has `synced_height == None`. Also correct the opening comment: wallets the commit never reached must be faulted; only wallets whose stores already returned are spared.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/changeset/core_bridge.rs
```diff
@@ -2253,6 +2468,339 @@ mod tests {
...
+        );
+        assert!(
+            !sync_fault.load(Ordering::Relaxed),
+            "a slow store is not a fault"
+        );
+
+        cancel.cancel();
+        drop(tx);
+        let _ = handle.await;
+    }
```
thepastaclaw: **🟡 Suggestion: Remove the duplicate off-runtime progress test**

`a_blocking_store_does_not_park_the_runtime` and the later `a_blocked_store_does_not_park_the_runtime` enforce the same guarantee with a single-worker Tokio runtime, a controllably blocked synchronous `store()`, an externally scheduled sentinel, and an out-of-runtime deadline and release mechanism. Both are now deterministic, but retaining both duplicates a heavyweight thread/runtime fixture and leaves two tests to maintain as the adapter evolves. Their comments also each claim that no other test covers the behavior. Keep the later plain `#[test]` version, which explicitly owns the runtime and expresses the discriminator more directly, and remove this duplicate.

<sub>source: ['codex']</sub>

thepastaclaw (commented):

The production adapter correctly moves synchronous persistence off Tokio workers and preserves the fail-closed watermark invariant across commit panics. Two non-blocking test-quality issues remain: the panic suite does not exercise a persistable wallet ordered after the panicking wallet, and the off-runtime progress guarantee is covered by two nearly identical heavyweight fixtures.
Source: reviewers gpt-5.6-sol (general, rust-quality); final verifier gpt-5.6-sol. openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

### Review provenance

🟡 2 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet/src/changeset/core_bridge.rs:2721-2732: Cover wallets the commit never reaches after a panic
  The panic tests cover the wallet whose `store()` panics and a wallet successfully settled before that panic, but none includes a persistable wallet ordered after the panicking wallet. That case exercises the other load-bearing half of `batch_wallet_ids - settled`: unwinding drops its consumed changes before `store()` is attempted, so panic recovery must fault it before a later watermark can advance. A regression that faults only the directly panicking wallet would pass the current tests. Add a third wallet whose ID sorts after `doomed`, include it in the same drain, and verify that a later record-bearing store for it has `synced_height == None`. Also correct the opening comment: wallets the commit never reached must be faulted; only wallets whose stores already returned are spared.
- [SUGGESTION] packages/rs-platform-wallet/src/changeset/core_bridge.rs:2471-2576: Remove the duplicate off-runtime progress test
  `a_blocking_store_does_not_park_the_runtime` and the later `a_blocked_store_does_not_park_the_runtime` enforce the same guarantee with a single-worker Tokio runtime, a controllably blocked synchronous `store()`, an externally scheduled sentinel, and an out-of-runtime deadline and release mechanism. Both are now deterministic, but retaining both duplicates a heavyweight thread/runtime fixture and leaves two tests to maintain as the adapter evolves. Their comments also each claim that no other test covers the behavior. Keep the later plain `#[test]` version, which explicitly owns the runtime and expresses the discriminator more directly, and remove this duplicate.
```
</details>

packages/rs-platform-wallet/src/changeset/core_bridge.rs
thepastaclaw: Resolved in `75ac76c` — _Remove the duplicate off-runtime progress test_ no longer present.

packages/rs-platform-wallet/src/changeset/core_bridge.rs
thepastaclaw: Resolved in `75ac76c` — _Cover wallets the commit never reaches after a panic_ no longer present.

packages/rs-platform-wallet/src/changeset/core_bridge.rs
shumkov: Addressed in 4e5a1939: JoinError fault scoping is per-wallet (core_bridge.rs:417-424, 489-506, 634-641) — cross-model verified (codex gpt-5.6) at head 75ac76c7.

packages/rs-platform-wallet/src/changeset/core_bridge.rs
shumkov: Addressed in 75ac76c7: the race-prone duplicate test is removed and replaced (core_bridge.rs:2560-2608) — cross-model verified at head.

packages/rs-platform-wallet/src/changeset/core_bridge.rs
shumkov: Addressed in 75ac76c7: wallets the commit never reached are covered (core_bridge.rs:2645-2652, 2667-2672, 2716-2732) — cross-model verified at head.

shumkov (approved):
Cross-model verified (codex gpt-5.6 via Orca, card on file): spawn_blocking handle awaited, per-wallet panic fault scoping sound, no interaction with the new rust-dashcore pin; all three suggestion threads confirmed addressed at head 75ac76c7.

PR #4372, opened 2026-08-11
fix(platform-wallet): accept legacy dashj key purposes on inbound contact requests

romchornyi:

A mainnet user could not pay **any** DashPay contact established before the iOS client existed — contacts brought over by importing an Android/dashj seed. Every attempt failed with:

```
Invalid identity data: No DashpayExternalAccount found for contact <id>
  — call register_external_contact_account first
```

Contacts created fresh on iOS worked normally, which is what made this look like a restore bug rather than a validation one.

Device logs (mainnet, one 29-contact wallet) show the split exactly:

| outcome | contacts |
|---|---|
| `drain: contact request key-purpose mismatch` | **27** |
| `Registered DashpayExternalAccount` | 2 (both established on iOS) |

The rejected purposes are all on **our own** key, as referenced by the inbound legacy document:

| our key id | purpose | warns |
|---|---|---|
| 3 | `TRANSFER` | 131 |
| 2 | `AUTHENTICATION` | 117 |
| 1 | `AUTHENTICATION` | 90 |
| 0 | `AUTHENTICATION` | 43 |
| 1 (+ sender key 1 also `AUTHENTICATION`) | `AUTHENTICATION` | 15 |

**Root cause.** `validate_contact_request` required `ENCRYPTION`/`DECRYPTION` for `recipientKeyIndex`, but legacy dashj `contactRequest` documents point at the recipient's `AUTHENTICATION` (key ids 0-2) or `TRANSFER` (key id 3) key. The drain classified this as a purpose-only mismatch and — correctly — refused to mark the channel broken, so it retried forever and never succeeded.

Two things kept it invisible:

- Contacts restored from seed are built **only** by the deferred path (the signerless sweep enqueues `RegisterExternal`; the signer-backed drain completes it). Contacts established live build inline and surface real errors. That is the entire difference between the two cohorts.
- `send_payment` calls `drain_pending_contact_crypto` at `payments.rs:1099` and discards its result, so every drain failure reaches the user as the same generic lookup error.

The previous policy was calibrated on a 368-document **testnet** census that contains no dashj-era cohort.

## What was done?

**Split the purpose policy in two** rather than widening the existing predicate — the existing one is also called from `create_contact_request` (`rs-sdk/src/platform/dashpay/contact_request.rs:243,265`), so widening it would have silently relaxed the key we pick for our own outgoing requests too.

- `recipient_key_purpose_is_valid` — unchanged, governs what we **mint**. `select_recipient_key_index` still prefers `DECRYPTION` then `ENCRYPTION`, so key separation on new documents is preserved.
- `recipient_key_purpose_is_acceptable_on_receive` / `sender_key_purpose_is_acceptable_on_receive` (new, `rs-sdk/src/platform/dashpay/contact_request.rs`) — govern what we **accept** from immutable history.

Accepted on receive:

- recipient: `DECRYPTION`, `ENCRYPTION`, `AUTHENTICATION`, `TRANSFER`
- sender: `ENCRYPTION`, `AUTHENTICATION`

`validate_contact_request` (`rs-platform-wallet/src/wallet/identity/crypto/validation.rs`) now uses the receive-side predicates. The `ECDSA_SECP256K1` key-type gate and the disabled-key check are untouched — those are the checks that actually protect the ECDH. Purpose is not a security boundary here: ECDH is defined over the secp256k1 keypair, and DIP-9 indexes the identity-key tree by key *type* and id, never by purpose, so the same derivation reaches all of them.

`SYSTEM`/`VOTING`/`OWNER` stay rejected, and stay a **non-permanent** purpose mismatch, so a later evidence-driven widening can still recover those contacts instead of finding their channels broken.

Doc comments on `validate_contact_request` and `select_recipient_key_index` were updated to record the legacy cohort and why the mint-side rule deliberately stays stricter than the receive-side one.

## How Has This Been Tested?

`cargo test -p platform-wallet -p dash-sdk` — 664 + 203 tests pass. `cargo clippy --all-targets` clean, `cargo fmt` applied.

New tests in `validation.rs`:

- `legacy_dashj_recipient_key_purposes_are_accepted` — `AUTHENTICATION` and `TRANSFER` recipient keys validate. This is the regression guard for the 27 unpayable contacts.
- `legacy_dashj_authentication_pair_is_accepted` — the full `AUTHENTICATION`/`AUTHENTICATION` pair, the exact shape 15 of the logged failures took.
- `recipient_node_operational_key_is_rejected_as_purpose_mismatch` — `SYSTEM`/`VOTING`/`OWNER` still rejected, and still non-permanent (`purpose_mismatch && !hard_error`).

New test in `contact_request.rs`:

- `receive_side_accepts_the_legacy_dashj_cohort` / `receive_side_still_refuses_node_operational_purposes` — pin the two new predicates.

Rewritten because they asserted the behaviour being reversed: `mint_side_still_refuses_authentication_and_transfer` (was `recipient_key_purpose_rejects_authentication`, now explicitly scoped to the mint side), `unaccepted_sender_purpose_is_a_purpose_mismatch` and `test_sender_wrong_purpose` (both now use a purpose that is still rejected).

**Not yet verified, and worth gating the merge on:** our ECDH convention (`rs-platform-encryption/src/ecdh.rs`) has only a self-generated known-answer test — nothing pins it against real dashj output. If it diverges, this change does **not** risk funds (`parse_compact_xpub` requires exactly 69 bytes, so a silently-accepted bogus xpub is ~2⁻⁵⁵), but a clean `Permanent` decrypt error would flip those 27 contacts from "queued" to `payment_channel_broken`, which only a superseding contact request can heal. Decrypting 2-3 of the real logged documents against the user's seed would settle it.

## Breaking Changes

None. The change only widens what is accepted from inbound documents; the documents we create are unaffected.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved receive-side contact validation for legacy requests using authentication and transfer key purposes.
  * Continued enforcing key type, enabled-key, and sender-purpose requirements.
  * Correctly rejects unsupported node-operational key purposes as non-permanent mismatches.
  * Prevented compatible legacy registration failures from incorrectly breaking contact channels or losing pending operations.
* **Compatibility**
  * Maintained strict requirements for outgoing requests while improving compatibility with legacy incoming requests.
* **Tests**
  * Added coverage for legacy compatibility, validation failures, and successful legacy account registration.

### DIFF
--- a/packages/rs-platform-wallet/src/wallet/identity/crypto/validation.rs
+++ b/packages/rs-platform-wallet/src/wallet/identity/crypto/validation.rs
@@ -3,10 +3,12 @@
 //! Validates that the sender and recipient identities have the correct key
 //! types and purposes before a contact request is submitted to the platform.
 
-use dash_sdk::platform::dashpay::recipient_key_purpose_is_valid;
+use dash_sdk::platform::dashpay::{
+    recipient_key_purpose_is_acceptable_on_receive, sender_key_purpose_is_acceptable_on_receive,
+};
 use dpp::identity::accessors::IdentityGettersV0;
 use dpp::identity::identity_public_key::accessors::v0::IdentityPublicKeyGettersV0;
-use dpp::identity::{Identity, KeyType, Purpose};
+use dpp::identity::{Identity, KeyType};
 
 /// Result of validating a contact request before it is sent.
 #[derive(Debug, Clone)]
@@ -110,31 +112,50 @@ impl ContactRequestValidation {
 
 /// Validate a contact request against the verified on-chain envelope.
 ///
-/// The empirical testnet census (368 docs) shows two live
-/// honest cohorts: the dominant mobile population references an **unbound
-/// ENCRYPTION key for BOTH indices** (mobile identities carry no DECRYPTION
-/// key), and the newest cohort uses bound **ENCRYPTION(sender) /
-/// DECRYPTION(recipient)** — our original convention. Consensus enforces
-/// neither purpose nor boundedness on these integer fields. This validator is
-/// therefore *liberal on receive*: it accepts the purposes mobile actually
-/// uses while keeping the ECDSA key-*type* gate (every observed key is
-/// ECDSA_SECP256K1) and the disabled-key check.
+/// Consensus enforces neither purpose nor boundedness on `senderKeyIndex` /
+/// `recipientKeyIndex`, and a `contactRequest` document is immutable — so this
+/// validator is *liberal on receive* by necessity. Rejecting a document is not
+/// a retry, it is a permanent sentence on a contact relationship the user
+/// cannot renegotiate. What it keeps strict is what actually carries weight:
+/// the ECDSA key-*type* gate (ECDH needs the full secp256k1 key) and the
+/// disabled-key check.
+///
+/// Three live cohorts are known:
+/// - **Newest** — bound `ENCRYPTION(sender)` / `DECRYPTION(recipient)`, our
+///   original convention.
+/// - **Mobile** — an unbound `ENCRYPTION` key for BOTH indices; these
+///   identities carry no DECRYPTION key at all. (Both of the above come from
+///   a 368-document *testnet* census.)
+/// - **Legacy Android/dashj** — references the recipient's `AUTHENTICATION`
+///   (key ids 0-2) or `TRANSFER` (key id 3) key, sometimes pairing it with an
+///   `AUTHENTICATION` sender key. Absent from the testnet census and
+///   discovered only from **mainnet** device logs (2026-08): 27 of one
+///   wallet's 29 contacts, every one of them established before the iOS
+///   client existed. Under the previous, testnet-calibrated policy all 27
+///   were permanently unpayable.
 ///
 /// # Checks performed
 ///
 /// **Sender key:**
 /// - Key at `sender_key_index` exists on the sender identity.
 /// - Key type is `ECDSA_SECP256K1` (required for ECDH).
-/// - Key purpose is `ENCRYPTION` (bound or unbound) — a non-ENCRYPTION
-///   purpose is flagged as a `purpose_mismatch` (non-permanent).
+/// - Key purpose is `ENCRYPTION` or `AUTHENTICATION` (bound or unbound) —
+///   anything else is flagged as a `purpose_mismatch` (non-permanent).
 /// - Key is not disabled.
 ///
 /// **Recipient key (our key):**
 /// - Key at `recipient_key_index` exists on the recipient identity.
 /// - Key type is compatible (`ECDSA_SECP256K1` or `ECDSA_HASH160`).
-/// - Key purpose is `ENCRYPTION` **or** `DECRYPTION` — anything else
-///   (AUTHENTICATION/MASTER/TRANSFER) is flagged as a `purpose_mismatch`.
+/// - Key purpose is `DECRYPTION`, `ENCRYPTION`, `AUTHENTICATION` or
+///   `TRANSFER` — the node-operational purposes (`SYSTEM`, `VOTING`,
+///   `OWNER`) are flagged as a `purpose_mismatch`.
 /// - Key is not disabled.
+///
+/// The accepted sets live in `dash_sdk::platform::dashpay` as
+/// `*_key_purpose_is_acceptable_on_receive`, deliberately separate from the
+/// stricter `recipient_key_purpose_is_valid` that governs the requests we
+/// *mint*: accepting history is not the same decision as choosing a key for a
+/// new document, and only the latter can still practice key separation.
 ///
 /// A failure whose *only* cause is a purpose mismatch sets
 /// [`ContactRequestValidation::purpose_mismatch`], signalling callers to skip
@@ -161,12 +182,15 @@ pub fn validate_contact_request(
                 ));
             }
 
-            // Must have ENCRYPTION purpose (bound or unbound — both live
-            // cohorts use ENCRYPTION for the sender). A non-ENCRYPTION
-            // purpose is a non-permanent purpose mismatch.
-            if key.purpose() != Purpose::ENCRYPTION {
+            // ENCRYPTION is the modern convention; legacy dashj documents
+            // reference an AUTHENTICATION key. Both are accepted on receive —
+            // the document is immutable, so rejecting it is a permanent
+            // sentence on a contact the user cannot appeal. Anything else is
+            // still a non-permanent purpose mismatch (skip and retry).
+            if !sender_key_purpose_is_acceptable_on_receive(key.purpose()) {
                 validation.add_purpose_error(format!(
-                    "Sender key {} has purpose {:?}, but ENCRYPTION is required for contact requests",
+                    "Sender key {} has purpose {:?}, but ENCRYPTION or AUTHENTICATION is \
+                     required for contact requests",
                     sender_key_index,
                     key.purpose(),
                 ));
@@ -214,18 +238,22 @@ pub fn validate_contact_request(
                 }
             }
 
-            // Purpose must be ENCRYPTION or DECRYPTION: the mobile
-            // cohort's recipientKeyIndex points at an ENCRYPTION key, the
-            // newest cohort's at a DECRYPTION key — both honest. Anything
-            // else (AUTHENTICATION/MASTER/TRANSFER) is a non-permanent purpose
-            // mismatch: legacy 2024 docs reference AUTHENTICATION keys, so we
-            // skip-and-retry rather than permanently break the channel. The
-            // accepted cohort is owned by the shared SDK predicate so this
-            // validator and the recipient-key selector cannot disagree.
-            if !recipient_key_purpose_is_valid(key.purpose()) {
+            // Four honest cohorts reach this point: the newest references our
+            // DECRYPTION key, the mobile population our ENCRYPTION key, and
+            // the legacy Android/dashj population our AUTHENTICATION (key ids
+            // 0-2) or TRANSFER (key id 3) key. Purpose is not a security
+            // boundary here — ECDH is defined over the secp256k1 keypair and
+            // DIP-9 indexes the identity-key tree by type and id, never by
+            // purpose — so the type and disabled-key gates around this block
+            // are what actually protect the derivation. The node-operational
+            // purposes (SYSTEM/VOTING/OWNER) stay out — nothing on chain
+            // references them for DashPay — and remain a non-permanent purpose
+            // mismatch: skip and retry, never break the channel, so a later
+            // evidence-driven widening can still pick those contacts up.
+            if !recipient_key_purpose_is_acceptable_on_receive(key.purpose()) {
                 validation.add_purpose_error(format!(
-                    "Recipient key {} has purpose {:?}, but ENCRYPTION or DECRYPTION is \
-                     required for contact requests",
+                    "Recipient key {} has purpose {:?}, which is not accepted for contact \
+                     requests",
                     recipient_key_index,
                     key.purpose(),
                 ));
@@ -373,11 +401,11 @@ mod tests {
 
     #[test]
     fn test_sender_wrong_purpose() {
-        let sender = make_identity(vec![make_key(
-            0,
-            KeyType::ECDSA_SECP256K1,
-            Purpose::AUTHENTICATION,
-        )]);
+        // VOTING, not AUTHENTICATION: the legacy dashj cohort pairs an
+        // AUTHENTICATION sender key with an AUTHENTICATION recipient key and
+        // is now accepted on receive, so AUTHENTICATION no longer exercises
+        // the sender-side rejection this test is about.
+        let sender = make_identity(vec![make_key(0, KeyType::ECDSA_SECP256K1, Purpose::VOTING)]);
         let recipient = make_identity(vec![make_key(
             0,
             KeyType::ECDSA_SECP256K1,
@@ -460,9 +488,13 @@ mod tests {
     // references an UNBOUND ENCRYPTION key for BOTH senderKeyIndex and
     // recipientKeyIndex (mobile identities carry no DECRYPTION key); the
     // newest cohort uses bound ENCRYPTION(sender)/DECRYPTION(recipient).
-    // Consensus enforces neither purpose nor boundedness. So the validator
-    // must accept ENCRYPTION for the sender and ENCRYPTION-or-DECRYPTION for
-    // the recipient, keep the ECDSA type gate, and reject AUTHENTICATION.
+    // Consensus enforces neither purpose nor boundedness, and mainnet adds a
+    // third, legacy Android/dashj cohort referencing AUTHENTICATION/TRANSFER
+    // keys. So the validator accepts ENCRYPTION-or-AUTHENTICATION for the
+    // sender and everything but the node-operational purposes for the
+    // recipient, while keeping the
+    // ECDSA type gate and the disabled-key check — those are the checks that
+    // actually protect the ECDH.
     // -----------------------------------------------------------------------
 
     /// Mobile-cohort shape: sender references an ENCRYPTION key, recipient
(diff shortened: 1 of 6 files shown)


thepastaclaw:
🔍 Review in progress — actively reviewing now (commit ee984ad4)
_ETA: complete ~19:25 UTC (median 9m across 30 recent reviews)_
_Running 4m · Last checked: 2026-08-11 19:20 UTC_

packages/rs-platform-wallet/src/wallet/identity/crypto/validation.rs
```diff
@@ -493,35 +525,81 @@ mod tests {
...
-        );
-        assert!(
-            result.purpose_mismatch,
-            "an AUTHENTICATION recipient is a PURPOSE mismatch (non-permanent skip), not a hard/permanent failure"
-        );
-        assert!(result.errors.iter().any(|e| e.contains("ENCRYPTION")
-            || e.contains("DECRYPTION")
-            || e.contains("purpose")));
+        let result = validate_contact_request(&sender, 1, &recipient, 1);
+        assert!(result.is_valid, "errors: {:?}", result.errors);
```
thepastaclaw, line 575: **🟡 Suggestion: Add a DashJ interoperability fixture for the newly accepted path**

The new tests use synthetic identities and stop after `validate_contact_request`, so they only prove that the widened purpose predicates accept AUTHENTICATION and TRANSFER. This PR now allows those requests to continue through the deferred `RegisterExternal` path, where the wallet derives the recipient key at the legacy key ID, performs ECDH, decrypts `encryptedPublicKey`, and parses the compact xpub. There is no DashJ-produced known-answer fixture covering those steps, so these tests would remain green if the actual legacy payload still failed because of a derivation-path or ECDH/AES convention mismatch. Such a failure is classified as permanent by `register_external_contact_account`; the drain then marks `payment_channel_broken` and removes the queued operation. Add a sanitized fixed seed and DashJ-generated payload—covering at least an AUTHENTICATION recipient and preferably TRANSFER key ID 3—and run it through the production derivation/decryption path, asserting that the expected compact xpub or external account is produced without marking the channel broken.

<sub>source: ['codex']</sub>

thepastaclaw (commented):

The receive-side policy is correctly separated from the stricter mint-side policy, and the existing key-type and disabled-key checks remain intact. One in-scope test gap remains: the regression tests stop at purpose validation and do not prove that a DashJ-produced legacy payload succeeds through key derivation, ECDH, decryption, and compact-xpub parsing before the permanent-failure path can mark the channel broken. Source: reviewer backends — general: gpt-5.6-sol; security-auditor: gpt-5.6-sol; rust-quality: gpt-5.6-sol; final verifier backend — gpt-5.6-sol. Orchestration only (not reviewer evidence): openclaw-agent/cliproxy/gpt-5.6-sol.

### Review provenance

🟡 1 suggestion(s)

<details>

```

In `packages/rs-platform-wallet/src/wallet/identity/crypto/validation.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/crypto/validation.rs:539-575: Add a DashJ interoperability fixture for the newly accepted path
  The new tests use synthetic identities and stop after `validate_contact_request`, so they only prove that the widened purpose predicates accept AUTHENTICATION and TRANSFER. This PR now allows those requests to continue through the deferred `RegisterExternal` path, where the wallet derives the recipient key at the legacy key ID, performs ECDH, decrypts `encryptedPublicKey`, and parses the compact xpub. There is no DashJ-produced known-answer fixture covering those steps, so these tests would remain green if the actual legacy payload still failed because of a derivation-path or ECDH/AES convention mismatch. Such a failure is classified as permanent by `register_external_contact_account`; the drain then marks `payment_channel_broken` and removes the queued operation. Add a sanitized fixed seed and DashJ-generated payload—covering at least an AUTHENTICATION recipient and preferably TRANSFER key ID 3—and run it through the production derivation/decryption path, asserting that the expected compact xpub or external account is produced without marking the channel broken.
```
</details>

romchornyi:
Addressed in afec36d01c — thanks, the harm you named is the right one to design around.

**The test gap is real and I cannot fully close it.** A dashj known-answer fixture needs a dashj-generated payload plus the matching seed, and neither exists in this repo. Rather than write a test that looks like it covers that and does not, I split the response:

**1. What I could pin, I pinned.** `legacy_key_id_and_purpose_survive_the_whole_external_build` runs key id 3 — the TRANSFER slot the legacy cohort references — through the production path end to end: `identity_auth_derivation_path` at the real DIP-9 path, `ContactCryptoProvider::ecdh_shared_secret`, AES decrypt, compact-xpub parse, `register_external_contact_account`. The two ECDH sides are derived independently (ours through the provider, the sender's by hand from our public key at the same path) and asserted equal, so the symmetry is real rather than one value handed to both halves. That establishes nothing downstream of the predicate is purpose- or id-sensitive. Its doc comment states plainly what it does not prove.

**2. What I could not pin, I made non-destructive.** Your finding's sharpest point is the consequence, not the gap: decrypt and compact-xpub parse are the only gates on the plaintext, so a convention mismatch is indistinguishable from a corrupt document, and the permanent classification then breaks every legacy channel at once — healable only by the contact sending a fresh request, which the user cannot make them do.

So a permanent register fault on a request accepted **only** by the widened policy now leaves the entry queued instead of marking the channel broken. Requests that would pass the mint-side policy keep the original permanent classification. Cost is a retry; the alternative costs the user a relationship they cannot repair.

Worth noting the failure mode is skewed the safe way regardless: `parse_compact_xpub` requires exactly 69 bytes, so a wrong shared secret producing a plausible-looking xpub is on the order of 2^-55. A convention mismatch shows up as a clean failure, not a silently bogus account.

**Still open, and I think it should gate the merge:** decrypting 2–3 of the real logged mainnet documents against the affected user's seed. That is the only thing that actually settles dashj interop, and it needs material only the user has.

packages/rs-platform-wallet/src/wallet/identity/network/payments.rs
```diff
@@ -5026,6 +5026,121 @@ mod tests {
...
+    /// production `ContactCryptoProvider::ecdh_shared_secret` at the real
+    /// DIP-9 auth path, the sender's side by hand from our public key at that
+    /// same path — so the asserted symmetry is real and not one value handed
+    /// to both halves.
+    ///
+    /// What this does NOT prove: that a payload produced by **dashj** decrypts
+    /// under our ECDH/AES conventions. That needs a dashj-generated known
+    /// answer, which no fixture in this repo has. It is why the drain treats a
+    /// permanent register fault on a legacy-cohort request as "leave queued"
+    /// rather than "break the channel".
```
thepastaclaw, line 5045: **🟡 Suggestion: Add a DashJ interoperability fixture for the newly accepted path**

The new end-to-end test verifies key ID 3 derivation, ECDH, AES decryption, compact-xpub parsing, and registration, but both sides still use this repository's Rust implementations. As the test comment acknowledges, two internally consistent Rust paths cannot detect disagreement with the historical DashJ implementation over the derivation path, ECDH shared-secret representation, AES convention, or compact payload. The new queue-preserving branch prevents a recipient-purpose legacy request from being marked broken on such a failure, but the external account would still never be registered and payment attempts would continue failing indefinitely. Add a known-answer payload produced by the relevant DashJ version from synthetic fixed key material and pass it through the production derivation, decryption, parsing, and registration path, covering at least an AUTHENTICATION recipient and preferably TRANSFER key ID 3.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs
```diff
@@ -2095,6 +2105,21 @@ impl<B: TransactionBroadcaster + ?Sized> DashPayView<'_, B> {
...
+                    // dashj-produced payload. Used below to keep a decrypt
+                    // failure from being treated as the document's fault.
+                    let accepted_by_legacy_widening = our_identity
+                        .get_public_key_by_id(*our_decryption_key_index)
+                        .map(|k| {
+                            !dash_sdk::platform::dashpay::recipient_key_purpose_is_valid(
+                                k.purpose(),
+                            )
+                        })
+                        .unwrap_or(false);
```
thepastaclaw: **🟡 Suggestion: Include sender-only widening in the legacy failure classification**

`accepted_by_legacy_widening` is described as identifying requests admitted only by the wider receive policy, but it checks only the recipient key. This PR also widens the sender policy from ENCRYPTION to AUTHENTICATION, so an AUTHENTICATION sender paired with a mint-valid ENCRYPTION or DECRYPTION recipient passes validation solely because of the receive-side widening while this flag remains false. If decryption or compact-xpub parsing then returns a permanent error, the ordinary arm marks the channel broken and clears the queued operation, preserving the destructive behavior that the new classification is intended to avoid for unverified legacy payloads. Derive the flag from both referenced keys and add a drain-level test for the sender-only widened shape.

```suggestion
let recipient_required_legacy_widening = our_identity
    .get_public_key_by_id(*our_decryption_key_index)
    .map(|key| {
        !dash_sdk::platform::dashpay::recipient_key_purpose_is_valid(key.purpose())
    })
    .unwrap_or(false);
let sender_required_legacy_widening = contact_identity
    .get_public_key_by_id(*contact_encryption_key_index)
    .map(|key| key.purpose() != Purpose::ENCRYPTION)
    .unwrap_or(false);
let accepted_by_legacy_widening =
    recipient_required_legacy_widening || sender_required_legacy_widening;
```

<sub>source: ['codex']</sub>

thepastaclaw (commented):

The receive-side purpose widening remains correctly separated from the stricter mint-side policy, and the current head adds useful Rust end-to-end coverage plus non-destructive handling for recipient-purpose legacy failures. Actual DashJ interoperability is still unverified, and the new failure classification does not protect requests admitted solely by the widened sender-purpose policy.
Source: reviewers gpt-5.6-sol (general), gpt-5.6-sol (security-auditor), and gpt-5.6-sol (rust-quality); final verifier gpt-5.6-sol. Orchestration-only, not reviewer evidence: openclaw-agent/cliproxy/gpt-5.6-sol.

### Review provenance

🟡 2 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/payments.rs:5041-5045: Add a DashJ interoperability fixture for the newly accepted path
  The new end-to-end test verifies key ID 3 derivation, ECDH, AES decryption, compact-xpub parsing, and registration, but both sides still use this repository's Rust implementations. As the test comment acknowledges, two internally consistent Rust paths cannot detect disagreement with the historical DashJ implementation over the derivation path, ECDH shared-secret representation, AES convention, or compact payload. The new queue-preserving branch prevents a recipient-purpose legacy request from being marked broken on such a failure, but the external account would still never be registered and payment attempts would continue failing indefinitely. Add a known-answer payload produced by the relevant DashJ version from synthetic fixed key material and pass it through the production derivation, decryption, parsing, and registration path, covering at least an AUTHENTICATION recipient and preferably TRANSFER key ID 3.

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs:2115-2122: Include sender-only widening in the legacy failure classification
  `accepted_by_legacy_widening` is described as identifying requests admitted only by the wider receive policy, but it checks only the recipient key. This PR also widens the sender policy from ENCRYPTION to AUTHENTICATION, so an AUTHENTICATION sender paired with a mint-valid ENCRYPTION or DECRYPTION recipient passes validation solely because of the receive-side widening while this flag remains false. If decryption or compact-xpub parsing then returns a permanent error, the ordinary arm marks the channel broken and clears the queued operation, preserving the destructive behavior that the new classification is intended to avoid for unverified legacy payloads. Derive the flag from both referenced keys and add a drain-level test for the sender-only widened shape.
```
</details>

packages/rs-platform-wallet/src/wallet/identity/crypto/validation.rs
thepastaclaw: Resolved in this update — _Add a DashJ interoperability fixture for the newly accepted path_ no longer present.

romchornyi:
Both addressed in 38e6c918a2.

**Sender-only widening — you found a real hole in the classification, thank you.** The flag inspected only the recipient key while the PR widens the sender rule too, so an AUTHENTICATION sender with a mint-valid recipient reached the decrypt purely on the receive-side policy and would still have had a convention gap charged to it. It is now the OR of both referenced keys against their respective mint-side rules.

`sender_only_legacy_shape_is_not_charged_for_a_decrypt_failure` pins exactly the shape you named. Verified it catches the defect rather than merely passing — with the sender term removed it fails `drained 1 vs 0` with the channel marked broken.

**DashJ fixture — I cannot close this one, and I would rather say so than simulate it.** Producing a known answer needs a payload generated by the historical dashj build plus its key material; neither exists in this repo, and I have no way to generate one here. Writing a fixture from our own implementation would be worse than no fixture, because it would look like interop coverage while proving only self-consistency — which is precisely the limitation the test's own doc comment states.

Your point that the queue-preserving branch does not *fix* a convention gap is correct, and I want to be plain about it: it bounds the damage to "still unpayable, recoverable later" instead of "unpayable forever, unrecoverable without the contact's action". Given the alternative is shipping nothing to 27 contacts that are already 100% broken today, I think bounded-and-recoverable is the right posture — but it is not a fix for that scenario and should not be read as one.

What actually settles it is empirical, and the material exists: decrypting 2-3 of the real logged mainnet documents against the affected user's seed. I have flagged that in the description as a merge gate. If you would prefer this PR block until that runs, I have no objection — it needs the user's seed, so it is a coordination question, not an engineering one.

packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs
thepastaclaw: Resolved in `38e6c91` — _Include sender-only widening in the legacy failure classification_ no longer present.

packages/rs-platform-wallet/src/wallet/identity/network/payments.rs
```diff
@@ -5026,6 +5026,268 @@ mod tests {
...
+        assert!(
+            !managed
+                .dashpay()
+                .established_contacts()
+                .get(&contact)
+                .expect("contact resident")
+                .payment_channel_broken,
+            "a sender-only legacy shape must not have a possible convention gap charged \
+             to it — the channel stays recoverable"
+        );
```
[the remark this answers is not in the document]
thepastaclaw: `drained` is not merely a success-report counter: the method documents it as the number of entries completed and removed from the queue, and `flush_drained_contact_crypto` computes it from the entries actually removed. With no concurrent mutation in this test, clearing this `RegisterExternal` entry through the drain would make the result 1, so the existing `drained == 0` assertion already pins queue retention; the channel assertion separately verifies that the contact was not marked broken.

QuantumExplorer (approved):

bfoss765:
Confirming this from the Android side: we hit the same failure class during kotlin-sdk cutover field testing on mainnet, from the same legacy-dashj cohort. A long-lived wallet (~150 contacts, dashj-era history) had 3 established contacts whose inbound requests the SDK rejected with the verbatim

```
Sender key 1 has purpose AUTHENTICATION, but ENCRYPTION is required
```

— contacts that work normally in the shipped dashj-based app, where both sides see identical transaction history. Our own interop analysis of v11.9-era Android and shipped iOS independently reached the same conclusion this PR implements: reaching pre-SDK mobile contacts requires key-purpose tolerance (the AUTHENTICATION fallback) on the receive path.

The shape here looks right to us — liberal on receive, while the stricter `recipient_key_purpose_is_valid` still governs requests we create. We support the change and will pick it up on the Android integration line once merged.

romchornyi:
## The dashj interoperability gap is now closed — by real mainnet data, not a fixture

The reporting user ran a build carrying this branch against their live mainnet wallet. Result:

```
Registered DashpayExternalAccount  ×29 distinct contacts
DashPay payment broadcast          ×8
drain failures                      0
channels marked broken              0
```

Before the fix that same wallet built **2** external accounts out of 29 and logged 396 `key-purpose mismatch` rejections in one session.

This answers the finding I could not close with a test. Those 27 payloads are genuine dashj-produced `encryptedPublicKey` blobs, minted years ago by the client we could not reproduce — so their decrypting under our ECDH shared-secret convention, AES-256-CBC/PKCS7 layout and 69-byte compact parse is exactly the known-answer evidence a synthetic fixture could not provide. Eight of those contacts then took a real payment end to end.

I still think a committed fixture would be worth having for regression safety, but it is no longer the difference between merging and not: the interop claim now rests on 27 independent live samples rather than on reading the code.

The queue-preserving branch for legacy-cohort permanent faults never fired, which is the outcome it was designed for — it stays as insurance for a cohort we have not seen yet.

**Note on scope:** the same run surfaced a *separate*, pre-existing bug — after 8 contact payments the wallet reports `Insufficient funds: available 41505, required 100000` while the home balance still shows the pre-payment amount. That is the known DIP-15 balance-accounting issue (contact-payment addresses are watched by our own wallet, so outbound value is never debited from the displayed balance), not a regression from this PR. Tracking it separately.

PR #4373, opened 2026-08-11
perf(platform-wallet): keep the contact fetch off the drain's repeating-rejection path

romchornyi:
> Stacked on #4372 — review that one first. This PR's diff against it is two files.

A `RegisterExternal` entry that the key-purpose policy turns away **stays queued on purpose**: the `contactRequest` document is immutable, so what might change is our acceptance policy, and marking the channel broken instead would need a superseding request from the contact to heal — an appeal the user cannot file. The consequence is that the sweep re-offers the same entry every pass, for the life of the wallet.

The drain fetched the contact identity from Platform **before** it could validate, so every one of those passes spent a network round trip to reach a verdict it already had locally. The dominant rejection is on `recipientKeyIndex` — **our own** key, always resident, no fetch required.

Mainnet logs from a single wallet, one session:

- 27 contacts blocked, **396** fetch-then-reject cycles
- 396 WARN lines for what is really 4 distinct reasons
- and a second session in the same export showing 486 more

#4372 makes those 27 contacts succeed, but the mechanism stays: any future cohort our policy does not yet accept lands in the same forever-loop.

## What was done?

**Ordering — get the network off the rejection path**

- Split `validate_contact_request` into `validate_sender_key` (needs the counterparty identity) and `validate_recipient_key` (needs only our own). The composed `validate_contact_request` is unchanged and still calls both via the existing `ContactRequestValidation::merge`, so every other caller keeps the identical contract.
- `drain_pending_contact_crypto_until` now runs the recipient half first and only performs `Identity::fetch` when it passes.

**One failure policy, applied by both halves**

- New `apply_drain_validation_failure` helper: purpose-only → count it and leave queued; anything else (missing key, wrong type, disabled) → mark the channel broken and clear the entry. Both halves route through it, so they cannot drift.
- `validate_sender_key` / `validate_recipient_key` are `pub(crate)` — drain-internal decomposition, and both the `crypto` and `validation` modules are public, so `pub` would have committed them as external API.

**Logging — keep the signal, drop the flood**

- Per-entry purpose rejections drop to `debug!`.
- One end-of-drain `warn!` carries the entry count plus the distinct reasons and how many entries each blocked.

This deliberately does **not** add a persisted "policy-blocked" marker on the contact. That would need a SQLite migration plus the FFI persistence bridge, and — more importantly — a frozen contact needs a correct un-freeze trigger; getting that wrong reproduces the bug #4372 just fixed. Removing the cost of the retry is the smaller and safer lever.

## How Has This Been Tested?

`cargo test -p platform-wallet` — 667 pass. `cargo clippy --all-targets` clean, `cargo fmt` applied.

New test `drain_decides_our_own_key_fault_without_fetching_the_contact` (`payments.rs`): a wallet-owned but keyless owner means `recipientKeyIndex` 0 cannot resolve — a hard fault that must break the channel. The mock SDK has **no** contact-identity fetch configured, so the verdict is only reachable if the recipient half ran before the fetch.

Verified it is a real guard, not a passenger: checked out the parent branch's `contact_requests.rs` and re-ran it against the old ordering —

```
assertion `left == right` failed: a hard validation fault must clear the entry
rather than retry it forever
  left: 0
 right: 1
```

Under the old ordering the fetch failed first and the verdict was never reached.

`unaccepted_recipient_purpose_never_fetches_and_stays_recoverable` covers the case that actually repeats — an unaccepted *purpose*, not a hard fault. The contact identity **is** configured on the mock, with a hard sender-side fault, and the drain runs twice. Under the old ordering the fetch succeeds, the merge escalates to broken, and the entry is cleared; verified against the parent branch's file —

```
assertion `left == right` failed: pass 1: a purpose-rejected entry must stay
queued, not be cleared
  left: 1
 right: 0
```

So passing proves no fetch was spent: an implementation that short-circuited only hard recipient faults, and still fetched for purpose mismatches, would fail it.

## Breaking Changes

No API breakage. But there **is** a deliberate behaviour change, and an earlier revision of this description wrongly claimed there was none — review caught it:

**Mixed failures now resolve in favour of leaving the entry queued.** When our key is purpose-rejected *and* the contact's key carries a hard fault (missing / disabled / wrong type), the composed validator merged both halves, saw `hard_error`, and marked the channel permanently broken. Deciding on the recipient half alone classifies it purpose-only and leaves it queued.

Kept, as a stated decision: the `hard_error` precedence exists to stop a permanent fault becoming a retry-forever loop, and that loop was expensive *because each retry fetched*. With the fetch gone it costs a map lookup per sweep, while marking the channel broken is unappealable by the user — only a fresh request from the contact clears it. A sender-side hard fault still breaks the channel as soon as our own key stops being the blocker. Documented on `validate_recipient_key` and at the call site.

Also worth a reviewer's eye: a hard *recipient*-side fault now marks the channel broken without fetching the contact. That is correct — a missing, wrong-type or disabled key on our own identity is permanent independent of the contact — but the broken mark can now be set where a Platform round trip previously had to succeed first.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved contact validation for legacy key purposes and unsupported operational keys.
  * Missing contact keys are now treated as retryable, preventing channels from becoming permanently broken.
  * Improved deferred contact processing, recipient and sender validation, and retry handling.
  * Improved payment funding across accounts and cleaned up reservations after rejected transactions.
  * Restored payment addresses after failed or rejected sends.
  * Improved handling of disabled and non-ECDSA keys while preserving preferred key selection.
* **Diagnostics**
  * Added clearer encryption and decryption error details, including likely key mismatches, corrupted data, and data lengths.

### DIFF
--- a/packages/rs-platform-wallet/src/wallet/identity/crypto/validation.rs
+++ b/packages/rs-platform-wallet/src/wallet/identity/crypto/validation.rs
@@ -71,6 +71,14 @@ impl ContactRequestValidation {
         self.hard_error = true;
     }
 
+    /// Add an ABSENT-KEY error: the referenced key id does not exist on the
+    /// identity *today*. Sets `is_valid = false` but NOT `hard_error`, because
+    /// identities gain keys — see [`is_permanent`](Self::is_permanent).
+    pub fn add_absent_key_error(&mut self, error: String) {
+        self.errors.push(error);
+        self.is_valid = false;
+    }
+
     /// Add a key-PURPOSE error: sets `is_valid = false` AND flags
     /// `purpose_mismatch` so callers can downgrade a *purpose-only* failure
     /// to a non-permanent skip rather than a permanent broken-channel mark.
@@ -89,9 +97,29 @@ impl ContactRequestValidation {
     /// Whether the *sole* cause of invalidity is a key-purpose mismatch —
     /// the only case that may be downgraded to a non-permanent skip.
     /// A purpose mismatch that co-occurs with a hard error (disabled /
-    /// missing / wrong-type key) is NOT purpose-only and must stay permanent.
+    /// wrong-type key) is NOT purpose-only and must stay permanent.
     pub fn is_purpose_only(&self) -> bool {
         self.purpose_mismatch && !self.hard_error
+    }
+
+    /// Whether this failure can never resolve on its own — the only kind that
+    /// may permanently break a contact's payment channel.
+    ///
+    /// The distinction is not "did validation fail" but "can the world change
+    /// such that it stops failing". A `contactRequest` clears consensus without
+    /// consensus checking anything about the keys it names, so a document can
+    /// reference a key id our identity does not have *yet*: identities gain
+    /// keys (that is what the DashPay enablement flow does, and what
+    /// dashwallet-ios#981 exists to notice when it happened on another device).
+    /// Recording that as permanent turns a temporary gap into a relationship
+    /// the user cannot repair — only a fresh request from the CONTACT clears
+    /// the flag.
+    ///
+    /// So an absent key is retryable, alongside a purpose mismatch. What stays
+    /// permanent is what immutable facts make impossible: a key whose *type*
+    /// cannot do ECDH, and a key we have deliberately disabled.
+    pub fn is_permanent(&self) -> bool {
+        self.hard_error
     }
 
     /// Merge another validation result into this one.
@@ -221,7 +249,7 @@ pub(crate) fn validate_sender_key(
             }
         }
         None => {
-            validation.add_error(format!(
+            validation.add_absent_key_error(format!(
                 "Sender key index {} not found on identity {}",
                 sender_key_index,
                 sender_identity.id(),
@@ -319,7 +347,7 @@ pub(crate) fn validate_recipient_key(
             }
         }
         None => {
-            validation.add_error(format!(
+            validation.add_absent_key_error(format!(
                 "Recipient key index {} not found on identity {}",
                 recipient_key_index,
                 recipient_identity.id(),
@@ -625,6 +653,51 @@ mod tests {
         let result = validate_contact_request(&sender, 1, &recipient, 1);
         assert!(result.is_valid, "errors: {:?}", result.errors);
         assert!(!result.purpose_mismatch);
+    }
+
+    /// A key id the identity does not have **yet** must not be permanent.
+    ///
+    /// Identities gain keys — that is what the DashPay enablement flow does,
+    /// and dashwallet-ios#981 exists to notice it happening on another device.
+    /// A `contactRequest` clears consensus without consensus checking anything
+    /// about the keys it names, and it can never be re-minted, so recording
+    /// "we have no key 5 today" as a permanent verdict ends a relationship over
+    /// a gap that may close on its own — and only the CONTACT can clear the
+    /// flag, so the user cannot appeal it.
+    #[test]
+    fn an_absent_key_is_not_a_permanent_fault() {
+        let sender = make_identity(vec![make_key(
+            0,
+            KeyType::ECDSA_SECP256K1,
+            Purpose::ENCRYPTION,
+        )]);
+        let recipient = make_identity(vec![]);
+
+        let result = validate_contact_request(&sender, 0, &recipient, 0);
+        assert!(!result.is_valid, "an absent key still fails validation");
+        assert!(
+            !result.is_permanent(),
+            "but it must be retryable: the identity can gain the key later"
+        );
+    }
+
+    /// A key type that cannot do ECDH is permanent — a key's type is fixed for
+    /// its lifetime, so no future state makes this request usable.
+    #[test]
+    fn a_non_ecdh_key_type_is_a_permanent_fault() {
+        let sender = make_identity(vec![make_key(
+            0,
+            KeyType::ECDSA_SECP256K1,
+            Purpose::ENCRYPTION,
+        )]);
+        let recipient = make_identity(vec![make_key(0, KeyType::BLS12_381, Purpose::ENCRYPTION)]);
+
+        let result = validate_contact_request(&sender, 0, &recipient, 0);
+        assert!(!result.is_valid);
+        assert!(
+            result.is_permanent(),
+            "a BLS key can never do secp256k1 ECDH, so this one may break the channel"
+        );
     }
 
     /// The node-operational purposes are the ones still refused for a
--- a/packages/rs-platform-wallet/src/wallet/identity/crypto/validation.rs
+++ b/packages/rs-platform-wallet/src/wallet/identity/crypto/validation.rs
@@ -3,10 +3,12 @@
 //! Validates that the sender and recipient identities have the correct key
 //! types and purposes before a contact request is submitted to the platform.
 
-use dash_sdk::platform::dashpay::recipient_key_purpose_is_valid;
+use dash_sdk::platform::dashpay::{
+    recipient_key_purpose_is_acceptable_on_receive, sender_key_purpose_is_acceptable_on_receive,
+};
 use dpp::identity::accessors::IdentityGettersV0;
 use dpp::identity::identity_public_key::accessors::v0::IdentityPublicKeyGettersV0;
-use dpp::identity::{Identity, KeyType, Purpose};
+use dpp::identity::{Identity, KeyType};
 
 /// Result of validating a contact request before it is sent.
 #[derive(Debug, Clone)]
@@ -110,31 +112,50 @@ impl ContactRequestValidation {
 
 /// Validate a contact request against the verified on-chain envelope.
 ///
-/// The empirical testnet census (368 docs) shows two live
-/// honest cohorts: the dominant mobile population references an **unbound
-/// ENCRYPTION key for BOTH indices** (mobile identities carry no DECRYPTION
-/// key), and the newest cohort uses bound **ENCRYPTION(sender) /
-/// DECRYPTION(recipient)** — our original convention. Consensus enforces
-/// neither purpose nor boundedness on these integer fields. This validator is
-/// therefore *liberal on receive*: it accepts the purposes mobile actually
-/// uses while keeping the ECDSA key-*type* gate (every observed key is
-/// ECDSA_SECP256K1) and the disabled-key check.
+/// Consensus enforces neither purpose nor boundedness on `senderKeyIndex` /
+/// `recipientKeyIndex`, and a `contactRequest` document is immutable — so this
+/// validator is *liberal on receive* by necessity. Rejecting a document is not
+/// a retry, it is a permanent sentence on a contact relationship the user
+/// cannot renegotiate. What it keeps strict is what actually carries weight:
+/// the ECDSA key-*type* gate (ECDH needs the full secp256k1 key) and the
+/// disabled-key check.
+///
+/// Three live cohorts are known:
+/// - **Newest** — bound `ENCRYPTION(sender)` / `DECRYPTION(recipient)`, our
+///   original convention.
+/// - **Mobile** — an unbound `ENCRYPTION` key for BOTH indices; these
+///   identities carry no DECRYPTION key at all. (Both of the above come from
+///   a 368-document *testnet* census.)
+/// - **Legacy Android/dashj** — references the recipient's `AUTHENTICATION`
+///   (key ids 0-2) or `TRANSFER` (key id 3) key, sometimes pairing it with an
+///   `AUTHENTICATION` sender key. Absent from the testnet census and
+///   discovered only from **mainnet** device logs (2026-08): 27 of one
+///   wallet's 29 contacts, every one of them established before the iOS
+///   client existed. Under the previous, testnet-calibrated policy all 27
+///   were permanently unpayable.
 ///
 /// # Checks performed
 ///
 /// **Sender key:**
 /// - Key at `sender_key_index` exists on the sender identity.
 /// - Key type is `ECDSA_SECP256K1` (required for ECDH).
-/// - Key purpose is `ENCRYPTION` (bound or unbound) — a non-ENCRYPTION
-///   purpose is flagged as a `purpose_mismatch` (non-permanent).
+/// - Key purpose is `ENCRYPTION` or `AUTHENTICATION` (bound or unbound) —
+///   anything else is flagged as a `purpose_mismatch` (non-permanent).
 /// - Key is not disabled.
 ///
 /// **Recipient key (our key):**
 /// - Key at `recipient_key_index` exists on the recipient identity.
 /// - Key type is compatible (`ECDSA_SECP256K1` or `ECDSA_HASH160`).
-/// - Key purpose is `ENCRYPTION` **or** `DECRYPTION` — anything else
-///   (AUTHENTICATION/MASTER/TRANSFER) is flagged as a `purpose_mismatch`.
+/// - Key purpose is `DECRYPTION`, `ENCRYPTION`, `AUTHENTICATION` or
+///   `TRANSFER` — the node-operational purposes (`SYSTEM`, `VOTING`,
+///   `OWNER`) are flagged as a `purpose_mismatch`.
 /// - Key is not disabled.
+///
+/// The accepted sets live in `dash_sdk::platform::dashpay` as
+/// `*_key_purpose_is_acceptable_on_receive`, deliberately separate from the
+/// stricter `recipient_key_purpose_is_valid` that governs the requests we
+/// *mint*: accepting history is not the same decision as choosing a key for a
+/// new document, and only the latter can still practice key separation.
 ///
 /// A failure whose *only* cause is a purpose mismatch sets
 /// [`ContactRequestValidation::purpose_mismatch`], signalling callers to skip
@@ -161,12 +182,15 @@ pub fn validate_contact_request(
                 ));
             }
 
-            // Must have ENCRYPTION purpose (bound or unbound — both live
-            // cohorts use ENCRYPTION for the sender). A non-ENCRYPTION
-            // purpose is a non-permanent purpose mismatch.
-            if key.purpose() != Purpose::ENCRYPTION {
+            // ENCRYPTION is the modern convention; legacy dashj documents
+            // reference an AUTHENTICATION key. Both are accepted on receive —
+            // the document is immutable, so rejecting it is a permanent
+            // sentence on a contact the user cannot appeal. Anything else is
+            // still a non-permanent purpose mismatch (skip and retry).
+            if !sender_key_purpose_is_acceptable_on_receive(key.purpose()) {
                 validation.add_purpose_error(format!(
-                    "Sender key {} has purpose {:?}, but ENCRYPTION is required for contact requests",
+                    "Sender key {} has purpose {:?}, but ENCRYPTION or AUTHENTICATION is \
+                     required for contact requests",
                     sender_key_index,
                     key.purpose(),
                 ));
@@ -214,18 +238,22 @@ pub fn validate_contact_request(
                 }
             }
 
-            // Purpose must be ENCRYPTION or DECRYPTION: the mobile
-            // cohort's recipientKeyIndex points at an ENCRYPTION key, the
-            // newest cohort's at a DECRYPTION key — both honest. Anything
-            // else (AUTHENTICATION/MASTER/TRANSFER) is a non-permanent purpose
-            // mismatch: legacy 2024 docs reference AUTHENTICATION keys, so we
-            // skip-and-retry rather than permanently break the channel. The
-            // accepted cohort is owned by the shared SDK predicate so this
-            // validator and the recipient-key selector cannot disagree.
-            if !recipient_key_purpose_is_valid(key.purpose()) {
+            // Four honest cohorts reach this point: the newest references our
+            // DECRYPTION key, the mobile population our ENCRYPTION key, and
+            // the legacy Android/dashj population our AUTHENTICATION (key ids
+            // 0-2) or TRANSFER (key id 3) key. Purpose is not a security
+            // boundary here — ECDH is defined over the secp256k1 keypair and
+            // DIP-9 indexes the identity-key tree by type and id, never by
+            // purpose — so the type and disabled-key gates around this block
+            // are what actually protect the derivation. The node-operational
+            // purposes (SYSTEM/VOTING/OWNER) stay out — nothing on chain
+            // references them for DashPay — and remain a non-permanent purpose
+            // mismatch: skip and retry, never break the channel, so a later
+            // evidence-driven widening can still pick those contacts up.
+            if !recipient_key_purpose_is_acceptable_on_receive(key.purpose()) {
                 validation.add_purpose_error(format!(
-                    "Recipient key {} has purpose {:?}, but ENCRYPTION or DECRYPTION is \
-                     required for contact requests",
+                    "Recipient key {} has purpose {:?}, which is not accepted for contact \
+                     requests",
                     recipient_key_index,
                     key.purpose(),
                 ));
@@ -373,11 +401,11 @@ mod tests {
 
     #[test]
     fn test_sender_wrong_purpose() {
-        let sender = make_identity(vec![make_key(
-            0,
-            KeyType::ECDSA_SECP256K1,
-            Purpose::AUTHENTICATION,
-        )]);
+        // VOTING, not AUTHENTICATION: the legacy dashj cohort pairs an
+        // AUTHENTICATION sender key with an AUTHENTICATION recipient key and
+        // is now accepted on receive, so AUTHENTICATION no longer exercises
+        // the sender-side rejection this test is about.
+        let sender = make_identity(vec![make_key(0, KeyType::ECDSA_SECP256K1, Purpose::VOTING)]);
         let recipient = make_identity(vec![make_key(
             0,
             KeyType::ECDSA_SECP256K1,
@@ -460,9 +488,13 @@ mod tests {
     // references an UNBOUND ENCRYPTION key for BOTH senderKeyIndex and
     // recipientKeyIndex (mobile identities carry no DECRYPTION key); the
     // newest cohort uses bound ENCRYPTION(sender)/DECRYPTION(recipient).
-    // Consensus enforces neither purpose nor boundedness. So the validator
-    // must accept ENCRYPTION for the sender and ENCRYPTION-or-DECRYPTION for
-    // the recipient, keep the ECDSA type gate, and reject AUTHENTICATION.
+    // Consensus enforces neither purpose nor boundedness, and mainnet adds a
+    // third, legacy Android/dashj cohort referencing AUTHENTICATION/TRANSFER
+    // keys. So the validator accepts ENCRYPTION-or-AUTHENTICATION for the
+    // sender and everything but the node-operational purposes for the
+    // recipient, while keeping the
+    // ECDSA type gate and the disabled-key check — those are the checks that
+    // actually protect the ECDH.
     // -----------------------------------------------------------------------
 
     /// Mobile-cohort shape: sender references an ENCRYPTION key, recipient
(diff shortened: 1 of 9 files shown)


thepastaclaw:
✅ Final review complete — no blockers (commit 52fe0d96)

packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs
```diff
@@ -2044,6 +2053,48 @@ impl<B: TransactionBroadcaster + ?Sized> DashPayView<'_, B> {
...
+                            .apply_drain_validation_failure(
+                                entry,
+                                &recipient_validation,
+                                &mut policy_blocked,
+                            )
+                            .await
+                        {
+                            cleared.push(entry.key());
+                        }
+                        continue;
```
thepastaclaw, line 2181: **🟡 Suggestion: Preserve hard-error precedence across the split validation**

Applying the recipient result before fetching and validating the sender changes the failure policy for mixed faults. If the recipient key has only an unaccepted purpose but the sender key is missing, disabled, or the wrong type, this branch classifies the request as purpose-only and leaves it queued without examining the sender. The previous composed `validate_contact_request` merged both halves, propagated `hard_error`, and made `is_purpose_only()` false, so the channel was marked broken and the queue entry was cleared. This contradicts the documented mixed-failure invariant in `ContactRequestValidation`, the merge regression tests, and the PR's statement that only validation order changes. Preserve aggregate classification for mixed failures, or explicitly redefine the policy and update its documentation and regression coverage.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/identity/network/payments.rs
```diff
@@ -5026,6 +5026,92 @@ mod tests {
...
+    /// nothing — a hard, permanent fault that must break the channel. The mock
+    /// SDK has NO contact-identity fetch configured, so this can only pass if
+    /// the recipient half of the validation ran *before* the fetch: the old
+    /// ordering fetched first, failed transiently, and left the channel intact.
+    ///
+    /// That ordering is what keeps a purpose-rejected entry — which stays
+    /// queued by design, and so is retried on every sweep forever — from
+    /// spending a network round trip each time.
+    #[tokio::test]
+    async fn drain_decides_our_own_key_fault_without_fetching_the_contact() {
```
thepastaclaw, line 5296: **🟡 Suggestion: Exercise the purpose-only no-fetch path directly**

This test uses a missing recipient key, which is a hard failure. It proves that at least one local fault is checked before `Identity::fetch`, but it does not pin the PR's principal performance behavior: an unaccepted recipient purpose must remain queued while causing zero contact fetches across repeated drains. An implementation that short-circuits hard recipient faults but still fetches for purpose mismatches would pass this test and reintroduce the reported repeating network cost. Add a test with a valid ECDSA recipient key using an unaccepted purpose, drain more than once, and assert that no identity fetch occurs, the entry remains queued, and the channel remains unbroken.

<sub>source: ['codex']</sub>

thepastaclaw (commented):

The recipient-first split removes the contact fetch from local rejection paths, but it changes the established classification of requests that combine a recipient-purpose mismatch with a hard sender-key fault. The main purpose-only performance path also lacks a direct regression test, and the internal validation halves unnecessarily expand the public API. Source: reviewer backends: codex general — gpt-5.6-sol; codex rust-quality — gpt-5.6-sol; final verifier backend: codex — gpt-5.6-sol; orchestration-only, not reviewer evidence: openclaw-agent/cliproxy/gpt-5.6-sol.

### Review provenance

🟡 3 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs:2084-2095: Preserve hard-error precedence across the split validation
  Applying the recipient result before fetching and validating the sender changes the failure policy for mixed faults. If the recipient key has only an unaccepted purpose but the sender key is missing, disabled, or the wrong type, this branch classifies the request as purpose-only and leaves it queued without examining the sender. The previous composed `validate_contact_request` merged both halves, propagated `hard_error`, and made `is_purpose_only()` false, so the channel was marked broken and the queue entry was cleared. This contradicts the documented mixed-failure invariant in `ContactRequestValidation`, the merge regression tests, and the PR's statement that only validation order changes. Preserve aggregate classification for mixed failures, or explicitly redefine the policy and update its documentation and regression coverage.

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/payments.rs:5029-5043: Exercise the purpose-only no-fetch path directly
  This test uses a missing recipient key, which is a hard failure. It proves that at least one local fault is checked before `Identity::fetch`, but it does not pin the PR's principal performance behavior: an unaccepted recipient purpose must remain queued while causing zero contact fetches across repeated drains. An implementation that short-circuits hard recipient faults but still fetches for purpose mismatches would pass this test and reintroduce the reported repeating network cost. Add a test with a valid ECDSA recipient key using an unaccepted purpose, drain more than once, and assert that no identity fetch occurs, the entry remains queued, and the channel remains unbroken.

In `packages/rs-platform-wallet/src/wallet/identity/crypto/validation.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/crypto/validation.rs:182-244: Keep the drain-specific validation halves crate-private
  Both `wallet::identity::crypto` and its `validation` module are public, so declaring `validate_sender_key` and `validate_recipient_key` as `pub` exposes the new decomposition helpers to downstream crates. Their documentation identifies the internal deferred-crypto drain as the reason for the split, and the only callers are inside this crate. Use `pub(crate)` for these helpers so external callers continue through the complete `validate_contact_request` contract and the crate does not acquire an unnecessary public API commitment.
```
</details>

packages/rs-platform-wallet/src/wallet/identity/crypto/validation.rs
```diff
@@ -213,9 +227,26 @@ pub fn validate_contact_request(
...
+///
+/// Split out because the deferred-crypto drain would otherwise pay a Platform
+/// round trip (`Identity::fetch` of the contact) before it could discover that
+/// the request is unusable for a reason it could have known locally. A
+/// purpose-rejected entry stays queued by design — the policy, not the
+/// immutable document, is what might change — so that fetch was repeating on
+/// every sweep, forever. Mainnet logs from one wallet show 27 contacts and 396
+/// such fetch-then-reject cycles in a single session. Running this half first
+/// costs nothing and removes the network entirely from that loop.
+pub fn validate_recipient_key(
```
thepastaclaw: **🟡 Suggestion: Keep the drain-specific validation halves crate-private**

Both `wallet::identity::crypto` and its `validation` module are public, so declaring `validate_sender_key` and `validate_recipient_key` as `pub` exposes the new decomposition helpers to downstream crates. Their documentation identifies the internal deferred-crypto drain as the reason for the split, and the only callers are inside this crate. Use `pub(crate)` for these helpers so external callers continue through the complete `validate_contact_request` contract and the crate does not acquire an unnecessary public API commitment.

<sub>source: ['codex']</sub>

romchornyi:
All three addressed in 608dac581e (plus d822704a63 merging the parent branch forward).

**1. Hard-error precedence — you were right, and this was a real defect.** I changed the mixed-failure policy and then wrote "only validation order changes" in the description. It was not just ordering: our key purpose-rejected + the contact's key hard-faulted went from *broken* to *queued*.

I kept the new behaviour, but as a stated decision rather than an accident. The `hard_error` precedence exists to stop a permanent fault becoming a retry-forever loop — and that loop was expensive **because each retry fetched**. With the fetch gone it costs a map lookup per sweep, while marking the channel broken is unappealable by the user: only a fresh request from the contact clears it. So the invariant's own rationale is what this PR removes. A sender-side hard fault still breaks the channel the moment our own key stops being the blocker. Documented on `validate_recipient_key` and at the call site, and the Breaking Changes section now says so instead of denying it.

If you would rather have the original semantics preserved exactly, the only way is to fetch even when the recipient half rejects — which gives back the entire cost this PR removes. I think that trade is wrong, but it is your call to overrule.

**2. Purpose-only no-fetch path — good catch, my test did not discriminate.** `unaccepted_recipient_purpose_never_fetches_and_stays_recoverable`: our key is valid ECDSA with an unaccepted purpose, the contact identity **is** configured on the mock with a hard sender-side fault, and the drain runs twice. Under the old ordering the fetch succeeds, the merge escalates to broken, and the entry is cleared — verified against the parent branch's file:

```
assertion `left == right` failed: pass 1: a purpose-rejected entry must stay queued, not be cleared
  left: 1
 right: 0
```

Passing therefore proves no round trip was spent, and the "short-circuits hard faults but still fetches for purpose mismatches" implementation you described fails it. It covers finding 1 at the same time.

**3. `pub(crate)`.** Done — both helpers. They are drain-internal decomposition and both enclosing modules are public, so `pub` was an unintended API commitment.

667 tests pass, clippy clean.

thepastaclaw (approved):

The current head explicitly documents and tests the intentional recipient-first mixed-failure policy, directly covers repeated purpose-only drains without fetching the sender identity, and keeps the split validation helpers crate-private. The remaining clone suggestion identifies a possible micro-optimization, but the same full-identity clone already occurred on every retry before this PR and is not required to achieve the PR's stated goal of eliminating the repeated Platform fetch, so it is outside this review's scope.
Source: Codex general, rust-quality, and security-auditor reviewers — gpt-5.6-sol; final verifier — gpt-5.6-sol. Orchestration-only, not reviewer evidence: openclaw-agent/cliproxy/gpt-5.6-sol.

### Review provenance

packages/rs-platform-wallet/src/wallet/identity/crypto/validation.rs
thepastaclaw: Resolved in `608dac5` — _Keep the drain-specific validation halves crate-private_ no longer present.

packages/rs-platform-wallet/src/wallet/identity/network/payments.rs
thepastaclaw: Resolved in `d822704` — _Exercise the purpose-only no-fetch path directly_ no longer present.

packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs
thepastaclaw: Resolved in `d822704` — _Preserve hard-error precedence across the split validation_ no longer present.

packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs
```diff
@@ -2300,9 +2463,74 @@ impl<B: TransactionBroadcaster + ?Sized> DashPayView<'_, B> {
...
+    async fn apply_drain_validation_failure(
+        &self,
+        entry: &crate::changeset::PendingContactCrypto,
+        validation: &crate::wallet::identity::crypto::validation::ContactRequestValidation,
+        policy_blocked: &mut std::collections::BTreeMap<String, usize>,
+    ) -> bool {
+        if validation.is_purpose_only() {
+            for reason in &validation.errors {
+                *policy_blocked.entry(reason.clone()).or_default() += 1;
+            }
```
thepastaclaw: **🟡 Suggestion: Move validation errors into the summary instead of cloning them**

`apply_drain_validation_failure` borrows the validation result, so every purpose-only reason is cloned before the map lookup. These entries intentionally remain queued and revisit this path on every sweep; consequently, an equivalent reason already present in `policy_blocked` still causes another temporary heap allocation. Neither caller uses its validation result afterward, so the helper can take `ContactRequestValidation` by value and move its strings after logging. Alternatively, retain the borrowed API but use `get_mut` first, cloning only when a distinct reason is inserted for the first time.

```suggestion
if validation.is_purpose_only() {
    for reason in &validation.errors {
        if let Some(count) = policy_blocked.get_mut(reason) {
            *count += 1;
        } else {
            policy_blocked.insert(reason.clone(), 1);
        }
    }
```

<sub>source: ['codex']</sub>

thepastaclaw (commented):

The recipient-first validation split correctly avoids fetching the counterparty for locally decidable failures, preserves the composed validator for existing callers, and directly tests the repeated purpose-only path. One in-scope performance suggestion remains: the new summary path clones every rejection reason on every sweep, including reasons already represented in the map.
Source: Codex general reviewer — gpt-5.6-sol; Codex rust-quality reviewer — gpt-5.6-sol; final verifier — gpt-5.6-sol.
Orchestration-only, not reviewer evidence: openclaw-agent/cliproxy/gpt-5.6-sol.

### Review provenance

🟡 1 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs:2513-2516: Move validation errors into the summary instead of cloning them
  `apply_drain_validation_failure` borrows the validation result, so every purpose-only reason is cloned before the map lookup. These entries intentionally remain queued and revisit this path on every sweep; consequently, an equivalent reason already present in `policy_blocked` still causes another temporary heap allocation. Neither caller uses its validation result afterward, so the helper can take `ContactRequestValidation` by value and move its strings after logging. Alternatively, retain the borrowed API but use `get_mut` first, cloning only when a distinct reason is inserted for the first time.
```
</details>

packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs
```diff
@@ -1923,7 +1933,66 @@ impl<B: TransactionBroadcaster + ?Sized> DashPayView<'_, B> {
...
         }
 
+        // Our identity's key inventory, once per drain that has external
+        // builds queued. The whole legacy-cohort bug is a statement about this
+        // layout — an identity minted before DashPay encryption keys existed
+        // carries only AUTHENTICATION/TRANSFER slots, so inbound requests
+        // reference those ids and nothing downstream makes sense without
+        // knowing that. Reading it back off an exported log beats asking the
+        // user to query Platform. On-chain public metadata only; no key data.
+        {
```
thepastaclaw: **🟡 Suggestion: Skip key-inventory construction when INFO logging is disabled**

The key-inventory block runs before `tracing::info!` can apply its level filter. Every drain containing a `RegisterExternal` entry therefore builds a `BTreeSet`, acquires the wallet-manager read lock, allocates one formatted string per identity key, and joins those strings even when INFO events are disabled. Purpose-rejected entries intentionally remain queued and revisit this path on every sweep, so this adds recurring preparation cost to the path this PR is optimizing. Guard the entire block with `tracing::enabled!(tracing::Level::INFO)`.

```suggestion
if tracing::enabled!(tracing::Level::INFO) {
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/identity/network/contacts.rs
```diff
@@ -479,9 +491,18 @@ impl<B: TransactionBroadcaster + ?Sized> DashPayView<'_, B> {
...
+                    // length is the discriminator, so it leads the message. The
+                    // bytes themselves are never logged: they are the contact's
+                    // payment xpub, and this text reaches an exported log.
                     Permanent(PlatformWalletError::InvalidIdentityData(format!(
-                        "Decrypted contact xpub is neither a 69-byte DIP-15 compact form \
-                         nor a 78/107-byte BIP32/DIP-14 serialization: {e}"
+                        "Decrypted contact xpub is {} bytes — neither a 69-byte DIP-15 compact \
+                         form nor a 78/107-byte BIP32/DIP-14 serialization ({e}). The decrypt \
+                         itself SUCCEEDED, so the shared secret matched and it is the plaintext \
+                         layout that differs",
```
thepastaclaw: **🟡 Suggestion: Do not infer shared-secret correctness from PKCS7 padding**

`decrypt_extended_public_key` performs unauthenticated AES-CBC decryption and validates only PKCS7 padding. A padding failure can result from malformed or corrupted ciphertext as well as a mismatched shared secret, while an incorrect key has roughly a 1/256 chance of producing acceptable padding and reaching the parsing fallback. The exported errors nevertheless state conclusively that a failure means the shared secret did not match and that successful unpadding means it matched and only the plaintext layout differs. This can misdirect investigation of the legacy interoperability issue. Describe these outcomes as likely or possible causes, unless the ciphertext format is authenticated before making a definitive attribution.

<sub>source: ['codex']</sub>

thepastaclaw (commented):

The recipient-first validation split removes the repeated contact fetch while preserving the intended queue policy, and the prior reason-cloning finding is fixed at the current head. Two in-scope suggestions remain: avoid eagerly constructing INFO-only key inventories, and qualify the unauthenticated AES-CBC diagnostics instead of presenting PKCS7 results as conclusive evidence about the shared secret.
Source: Codex general reviewer — gpt-5.6-sol; Codex rust-quality reviewer — gpt-5.6-sol; final verifier — gpt-5.6-sol. Orchestration-only (not reviewer evidence): openclaw-agent/cliproxy/gpt-5.6-sol.

### Review provenance

🟡 2 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs:1943: Skip key-inventory construction when INFO logging is disabled
  The key-inventory block runs before `tracing::info!` can apply its level filter. Every drain containing a `RegisterExternal` entry therefore builds a `BTreeSet`, acquires the wallet-manager read lock, allocates one formatted string per identity key, and joins those strings even when INFO events are disabled. Purpose-rejected entries intentionally remain queued and revisit this path on every sweep, so this adds recurring preparation cost to the path this PR is optimizing. Guard the entire block with `tracing::enabled!(tracing::Level::INFO)`.

In `packages/rs-platform-wallet/src/wallet/identity/network/contacts.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/contacts.rs:453-504: Do not infer shared-secret correctness from PKCS7 padding
  `decrypt_extended_public_key` performs unauthenticated AES-CBC decryption and validates only PKCS7 padding. A padding failure can result from malformed or corrupted ciphertext as well as a mismatched shared secret, while an incorrect key has roughly a 1/256 chance of producing acceptable padding and reaching the parsing fallback. The exported errors nevertheless state conclusively that a failure means the shared secret did not match and that successful unpadding means it matched and only the plaintext layout differs. This can misdirect investigation of the legacy interoperability issue. Describe these outcomes as likely or possible causes, unless the ciphertext format is authenticated before making a definitive attribution.
```
</details>

packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs
thepastaclaw: Resolved in `abe80c6` — _Move validation errors into the summary instead of cloning them_ no longer present.

romchornyi:
Both addressed in d9afc996c0, plus the merge of `v4.2-dev` now that #4372 has landed (8415a507ba).

**Padding as evidence — you are right, and this was the more important of the two.** The messages asserted their conclusions outright ("the ECDH shared secret did not match", "the decrypt itself SUCCEEDED, so the shared secret matched"). Neither follows from unauthenticated CBC with padding as the only check: a rejection is equally consistent with a corrupted ciphertext, and a wrong key clears PKCS7 roughly 1 in 256 times and lands in the length branch instead.

That overclaim is worse than a normal wording slip, because these messages exist precisely to steer the legacy-interop investigation — a confident wrong pointer costs more than a vague right one. Both now name the likely cause **and** the alternative that produces the same symptom.

**Key inventory now gated on `tracing::enabled!(INFO)`.** Correct: it allocates a set, a string per key and a join, and takes the wallet-manager read lock, and purpose-rejected entries revisit that path every sweep — recurring cost on the path this PR exists to make cheap.

668 tests, clippy clean.

romchornyi:
Addressed in 0eac01e41c — you are right about the part that matters, and it exposed a hole I had walked past.

**Where the principle is exactly right.** `payment_channel_broken` is permanent, and it heals *only* when the contact sends a fresh request. The user cannot appeal it. So recording a local inability as a verdict about a consensus-valid document is the wrong default, and this stack is itself the proof: the key-purpose policy was wrong for two years, and the only reason 27 mainnet contacts stayed recoverable is that someone had classified that branch as a soft skip. Had it been "hard", they would have been permanently unpayable with no path back.

**The concrete hole.** An **absent key id** was classified permanent. But consensus checks nothing about the keys a `contactRequest` names, and identities *gain* keys — that is what the DashPay enablement flow does, and what dashwallet-ios#981 exists to notice when it happened on another device. A document referencing a key we do not have *yet* was ending the relationship over a gap that may close on its own. Same on the contact's side, so the drain's "contact encryption key missing" arm now leaves the entry queued too.

**Where I would not go all the way.** "Never reject locally" cannot be absolute: ECDH over a BLS or hash-only key is impossible whatever consensus says. So the predicate is no longer "how badly did validation fail" but "can this ever resolve":

| fault | before | now |
|---|---|---|
| key id absent (either side) | permanent | **retryable** — identities gain keys |
| key type cannot do ECDH | permanent | permanent — a key's type is fixed for its lifetime |
| key disabled | permanent | permanent — we revoked it deliberately |
| purpose mismatch | retryable | retryable |

`is_permanent()` replaces `is_purpose_only()` as the drain's predicate so the question asked is the question being answered. The co-occurrence guard is unchanged: a purpose mismatch alongside a genuinely permanent fault stays permanent.

**On scope** — this changes policy rather than ordering, so it arguably wanted its own PR, but it is a direct correction to the failure policy this PR already redefines, and splitting it would have left the wrong behaviour on the branch under review.

Two ordering tests were pinned on an absent key being permanent and stopped discriminating; both re-armed on a BLS key at the referenced index. `an_absent_key_is_not_a_permanent_fault` pins the new boundary — verified it fails when the absent key is put back on the permanent classification.

672 tests, clippy clean. #4378 is merged into this branch, so it now carries the pooled-funding fix as well.

packages/rs-platform-wallet/src/wallet/identity/network/payments.rs
```diff
@@ -1291,16 +1335,28 @@ impl<B: TransactionBroadcaster + ?Sized> DashPayView<'_, B> {
...
+            Err(e) if matches!(e, crate::broadcaster::BroadcastError::Rejected { .. }) => {
+                crate::wallet::reservations::release_reservation_after_rejected_broadcast(
+                    &self.wallet_manager,
+                    &self.wallet_id,
+                    &funding_accounts,
+                    &tx,
+                    // This path does not thread the build's reservation token
+                    // either; keep the historical unconditional release.
+                    None,
+                )
```
thepastaclaw, line 1353: **🟡 Suggestion: Retain the pooled reservation token through pre-broadcast exits**

`build_signed` discards the `ReservationToken` stamped across the contributing accounts. If `persister.store` fails at line 1330, the function returns without releasing the signed transaction's reserved inputs. If broadcast is rejected, passing `None` performs an unconditional release by outpoint. A TTL sweep can release the original reservation and allow another build to reserve the same inputs while broadcast is awaiting completion; the old build's unconditional cleanup can then clear the newer reservation. Use `build_signed_reserved`, carry its token alongside `tx` and `funding_accounts`, release with that token before returning a persistence error, and pass it to `release_reservation_after_rejected_broadcast` after a definitive rejection.

packages/rs-platform-wallet/src/wallet/identity/crypto/validation.rs
```diff
@@ -89,11 +97,31 @@ impl ContactRequestValidation {
     /// Whether the *sole* cause of invalidity is a key-purpose mismatch —
     /// the only case that may be downgraded to a non-permanent skip.
     /// A purpose mismatch that co-occurs with a hard error (disabled /
-    /// missing / wrong-type key) is NOT purpose-only and must stay permanent.
+    /// wrong-type key) is NOT purpose-only and must stay permanent.
     pub fn is_purpose_only(&self) -> bool {
         self.purpose_mismatch && !self.hard_error
     }
```
thepastaclaw, line 103: **🟡 Suggestion: Align the public retryability gate with absent-key classification**

`add_absent_key_error` intentionally leaves `hard_error` false because an absent key is retryable, but the public documentation still says missing keys are permanent and directs callers to `is_purpose_only()` for skip-versus-break decisions. That predicate returns false for a lone absent-key failure, while returning true for a purpose mismatch combined with an absent key even though purpose is not then the sole cause. The in-crate drain correctly uses `is_permanent()`, but downstream callers following the advertised public contract can retain the obsolete permanent treatment. Make `is_permanent()` the documented canonical disposition predicate and deprecate or explicitly restrict `is_purpose_only()` to diagnostics, or track absent-key failures separately so its stated sole-cause semantics remain exact. Add coverage for lone-absent and purpose-plus-absent results.

<sub>source: ['codex']</sub>

thepastaclaw (commented):

The recipient-first drain behavior and both previously reported diagnostics/performance issues are correct at the current head. Two in-scope suggestions remain: retain the pooled funding reservation token through all pre-broadcast exits, and align the public validation API with the new retryable absent-key policy.
Source: Codex general, rust-quality, and security-auditor reviewers — gpt-5.6-sol; final verifier — gpt-5.6-sol. Orchestration-only, not reviewer evidence: openclaw-agent/cliproxy/gpt-5.6-sol.

### Review provenance

🟡 2 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/payments.rs:1277-1353: Retain the pooled reservation token through pre-broadcast exits
  `build_signed` discards the `ReservationToken` stamped across the contributing accounts. If `persister.store` fails at line 1330, the function returns without releasing the signed transaction's reserved inputs. If broadcast is rejected, passing `None` performs an unconditional release by outpoint. A TTL sweep can release the original reservation and allow another build to reserve the same inputs while broadcast is awaiting completion; the old build's unconditional cleanup can then clear the newer reservation. Use `build_signed_reserved`, carry its token alongside `tx` and `funding_accounts`, release with that token before returning a persistence error, and pass it to `release_reservation_after_rejected_broadcast` after a definitive rejection.

In `packages/rs-platform-wallet/src/wallet/identity/crypto/validation.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/crypto/validation.rs:29-103: Align the public retryability gate with absent-key classification
  `add_absent_key_error` intentionally leaves `hard_error` false because an absent key is retryable, but the public documentation still says missing keys are permanent and directs callers to `is_purpose_only()` for skip-versus-break decisions. That predicate returns false for a lone absent-key failure, while returning true for a purpose mismatch combined with an absent key even though purpose is not then the sole cause. The in-crate drain correctly uses `is_permanent()`, but downstream callers following the advertised public contract can retain the obsolete permanent treatment. Make `is_permanent()` the documented canonical disposition predicate and deprecate or explicitly restrict `is_purpose_only()` to diagnostics, or track absent-key failures separately so its stated sole-cause semantics remain exact. Add coverage for lone-absent and purpose-plus-absent results.
```
</details>

packages/rs-platform-wallet/src/wallet/identity/network/contact_requests.rs
thepastaclaw: Resolved in `d9afc99` — _Skip key-inventory construction when INFO logging is disabled_ no longer present.

packages/rs-platform-wallet/src/wallet/identity/network/contacts.rs
thepastaclaw: Resolved in `d9afc99` — _Do not infer shared-secret correctness from PKCS7 padding_ no longer present.

packages/rs-platform-wallet/src/wallet/identity/network/payments.rs
```diff
@@ -5717,6 +6005,110 @@ mod tests {
...
+        // funds while a funded receiving account sits right there.
+        if let Err(e) = &result {
+            let msg = e.to_string();
+            assert!(
+                !msg.contains("Insufficient funds") && !msg.contains("No UTXOs available"),
+                "the contact-receiving account's 1_000_000 duffs must be offered to \
+                 selection — BIP44-only funding is the bug this pins, got: {msg}"
+            );
+        }
+    }
```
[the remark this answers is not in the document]
thepastaclaw: The separate `rejected_broadcast_releases_every_pooled_funding_account` test already proves this behavior more strongly: BIP44 and the DashPay receiving account each hold only 60,000 duffs for a 100,000-duff payment, and the test asserts that execution reaches a rejecting broadcaster. Omitting the receiving account, failing derivation, or failing signing would prevent the expected `TransactionBroadcast` result.

QuantumExplorer (approved):

PR #4329, opened 2026-08-07
feat(platform-wallet): pool BIP44 + BIP32 + DashPay receiving funds on the send path

QuantumExplorer:
## Issue being fixed or feature implemented

When sending funds, the wallet previously spent from exactly one account family the caller named — so coins sitting in DashPay contact-receiving accounts (or the "other" standard family) were unspendable without account gymnastics. rust-dashcore#925 made the transaction builder multi-account (additive `add_funding`, one process-wide reservation token spanning every contributing account) and rust-dashcore#929 added the DashPay receiving-side sources. This PR adopts both: **a plain send pools BIP44 + BIP32 + every DashPay contact-receiving account.**

## What was done?

- **Pin bump `dca5b05b` → `944e53a5`** — exactly three upstream commits: #926 (watch-only coins excluded from balances), #925 (multi-account funding), #929 (DashPay sources). Lock regenerated minimally.
- **`SEND_FUNDING_SOURCES = [BIP44, BIP32, AllDashpayReceivingFunds]`** — funding order is meaningful: the first source supplies the change address, so change from a pooled send always returns to the transparent primary. CoinJoin is deliberately excluded (spending mixed coins alongside transparent ones undoes the mixing — same reasoning as upstream's `DEFAULT`), and a contact's watch-only `DashpayExternalAccount` coins are excluded upstream by the receiving-side selector.
- **`finalize_transaction` takes a source list.** A single-element list keeps the old strict one-account contract (asking for a missing account errors); a pooled list skips sources that resolve to nothing (a wallet with no BIP32 account or no contacts still sends) and errors only if nothing funds. Source resolution mirrors key-wallet's own `fund()` semantics, including the overlap dedup that prevents double-offering one account's UTXOs.
- **Multi-account reservation lifecycle.** `SignedCoreTransaction` and the deferred-payment registry now record every contributing `AccountType`; abandon / rejected-broadcast / stale-token reconciliation run the one build token against each contributing account's `ReservationSet` (upstream reserves per account, all stamped with the same token — the #925 design).
- **FFI/host surface**: `CoreAccountTypeFFI::AllSpendable` (single-account APIs like gap limits reject it with a typed parameter error), JNI value 3, Kotlin `ALL_SPENDABLE` + Swift `.allSpendable` on the builder enums, and the send entry points (`sendToAddresses`, `buildSignedPayment`, `finalizeAtomic`) **default to pooled**. Explicit single-family selection still works everywhere.

## How Has This Been Tested?

[...]

PR #4378, opened 2026-08-11
fix(platform-wallet): pool the same funding sources for a contact payment as for a plain send

romchornyi:
> Stacked on #4373 (which is stacked on #4372). Its own diff is one file plus a test.

#4329 made sends draw from BIP44 + BIP32 + every DashPay receiving account. Its own words:

> A user holding funds on BIP32 or a DashPay contact-receiving account saw them in the total and got "insufficient funds" trying to spend them.

The DashPay **pay-to-contact** path never picked that up. Its entire share of #4329 was a mechanical rename:

```diff
-                .set_funding(managed_account, account)
+                .add_funding(managed_account, account)
```

The funding set stayed pinned to BIP44 account 0. So the bug #4329 fixed still reproduces — on the one path where contact-received funds are most likely to be sitting.

Reported from mainnet, straight after the legacy-contact fix in #4372 made 29 contacts payable: eight contact payments drained BIP44, and the ninth failed with

```
Transaction building failed: Coin selection error:
Insufficient funds: available 41505, required 100000
```

on a home screen showing plenty. The money was in the contact-receiving accounts — spendable, signable, and simply not offered to selection.

## What was done?

`send_payment` now walks `SEND_FUNDING_SOURCES` through the same `resolve_source_accounts` the core path uses, rather than reimplementing the resolution (it is `pub(crate)` now, and re-exported from `wallet::core`).

- **BIP44 is still offered first**, so the builder keeps taking the change address from it — change returns to BIP44 exactly as before.
- **CoinJoin stays excluded** by construction: spending mixed outputs beside transparent ones links them and undoes the mixing.
- **Contact *external* accounts stay excluded** too — they hold the counterparty's xpub and no key this wallet can sign with.
- A source the wallet does not have contributes nothing rather than failing the send: a wallet with no BIP32 account, or no contacts, still pays from BIP44.

Two consequences of pooling, both handled:

1. The signer closure resolved derivation paths from the one funding account (`managed_account.address_derivation_path`). It now consults a map built across every offered account.
2. A rejected broadcast released reservations on BIP44 only. Inputs drawn from a BIP32 or receiving account would have stayed reserved until the TTL backstop, and an immediate retry would fail with a spurious insufficient-funds. Release now covers every account that offered inputs — a superset of the contributors, which is the safe direction (releasing on an account that supplied nothing is a no-op).

## How Has This Been Tested?

`cargo test -p platform-wallet` — 670 pass. `cargo clippy --all-targets` clean, `cargo fmt` applied.

New test `contact_payment_funds_from_a_dashpay_receiving_account`: leaves BIP44 empty, registers a contact-receiving account and gives it the wallet's only 1,000,000 duffs, then pays the contact 100,000. Reaching the signer at all proves the receiving account was offered to selection.

Verified it catches the defect rather than merely passing — pinned back to BIP44-only funding it fails on the insufficient-funds assertion:

```
the contact-receiving account's 1_000_000 duffs must be offered to selection —
BIP44-only funding is the bug this pins, got: ... Insufficient funds ...
```

`rejected_broadcast_releases_every_pooled_funding_account` covers the release path: BIP44 and a contact-receiving account hold 60,000 duffs each and the payment is 100,000, so neither covers it alone — selection must take from both, and the retry after the rejection can only succeed if both reservations were released. The retry is the assertion.

Also verified by falsification — pinned back to the BIP44-only release it fails exactly as a user would experience it:

```
an immediate retry must reselect every pooled input ...:
TransactionBuild("Coin selection error: Insufficient funds: available 60000, required 100000")
```

The contact-receiving account's 60,000 stayed reserved until the TTL backstop.

## Breaking Changes

None. A contact payment that succeeded before still succeeds and still takes its change to BIP44; the change only widens what the selector may draw from.

## Checklist:

### DIFF
--- a/packages/rs-platform-wallet/src/wallet/core/mod.rs
+++ b/packages/rs-platform-wallet/src/wallet/core/mod.rs
@@ -10,5 +10,6 @@ pub mod wallet;
 pub use balance::WalletBalance;
 pub use balance_handler::BalanceUpdateHandler;
 pub use generation::WalletGeneration;
+pub(crate) use transaction::resolve_source_accounts;
 pub use transaction::{SignedCoreTransaction, ASSET_LOCK_FUNDING_SOURCES, SEND_FUNDING_SOURCES};
 pub use wallet::CoreWallet;
--- a/packages/rs-platform-wallet/src/wallet/core/transaction.rs
+++ b/packages/rs-platform-wallet/src/wallet/core/transaction.rs
@@ -266,7 +266,7 @@ pub const ASSET_LOCK_FUNDING_SOURCES: [AccountTypePreference; 3] = SEND_FUNDING_
 /// DashPay source. A set selector matching nothing resolves to an empty list,
 /// not an error — a wallet with no contacts still sends from its standard
 /// accounts.
-fn resolve_source_accounts(
+pub(crate) fn resolve_source_accounts(
     accounts: &key_wallet::account::ManagedAccountCollection,
     preference: AccountTypePreference,
     source_index: u32,
--- a/packages/rs-platform-wallet/src/wallet/identity/network/payments.rs
+++ b/packages/rs-platform-wallet/src/wallet/identity/network/payments.rs
@@ -6219,6 +6219,72 @@ mod tests {
         );
     }
 
+    /// A rejected broadcast releases the UTXO reservation on EVERY account
+    /// that funded the payment, not just BIP44.
+    ///
+    /// Pooling made this reachable: before it, one account funded the send and
+    /// releasing that one was complete. Now inputs can come from a BIP32 or
+    /// contact-receiving account too, and a release that still named only
+    /// BIP44 would leave those reserved until the TTL backstop — so the
+    /// immediate retry a user makes after "payment rejected" would fail with a
+    /// spurious insufficient-funds on money that is demonstrably theirs.
+    ///
+    /// Neither account can cover the payment alone here, so a successful retry
+    /// is only possible if BOTH were released.
+    #[tokio::test]
+    async fn rejected_broadcast_releases_every_pooled_funding_account() {
+        use crate::wallet::identity::network::contact_requests::SeedCryptoProvider;
+
+        let (manager, _persister, wallet_id, owner_id, contact_id) =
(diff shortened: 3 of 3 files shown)


thepastaclaw:
🕓 Ready for review — 3 ahead in queue (commit ebe696f2)
_Queue position: 4/5 · 1 review active_
_ETA: start ~22:15 UTC · complete ~22:25 UTC (median 10m across 30 recent reviews; 2 slots)_
_Queued 5m ago · Last checked: 2026-08-11 22:00 UTC_

romchornyi:
## Confirmed on testnet

Two runs, each decisive in one direction. They are **different wallets** — the failure was captured on one before this PR existed, the success on another after — so this is not a same-wallet A/B; each half stands on its own arithmetic.

**Without this PR** — wallet holding 90,998,870 duffs in BIP44 and 17,498,551 elsewhere (contact-receiving + BIP32); the pay sheet offered `Available: 1.08497421 DASH`, i.e. the pooled total:

```
Transaction building failed: Coin selection error:
Insufficient funds: available 90998870, required 105000000
```

`available` is BIP44 to the duff. The UI promised the pooled balance and the selector was given one account.

**With this PR** — wallet holding 6,899,172 duffs in BIP44 plus a 500,000 duff contact payment received minutes earlier:

```
TXSEND :: pay-to-contact starting — 7000000 duffs
DashPay payment broadcast amount_duffs=7000000
  txid=4a8bb77e8402f2615184ede7dd4dec8d25d73773955a61c3e028369e56dfd0bc
TXSEND :: pay-to-contact broadcast, fee 374 duffs
```

7,000,000 exceeds that wallet's BIP44 by 100,828 duffs, so it is only satisfiable by reaching the contact-receiving account — which is exactly what `contact_payment_funds_from_a_dashpay_receiving_account` asserts in a unit test, here reproduced end to end against live testnet.

An earlier run on the same stack also confirmed the parent PRs on **mainnet**: 29/29 legacy contacts built their external accounts and 8 contact payments went through — which is what exhausted BIP44 and surfaced this bug in the first place.

QuantumExplorer (approved):

PR #4383, opened 2026-08-12
test(drive): pin the batch-transition cap and land the phantom-group evidence

shumkov:

`max_transitions_in_documents_batch` has been `1` at every protocol version since the first mainnet release, and nothing in the codebase said why that matters. It reads as a throughput knob. It is not — it is the only thing keeping a consensus-visible index defect off the network, and raising it (an attractive, plausible future change) would arm that defect silently.

This PR does not fix the defect. It documents the cap, pins it with tests, and lands the evidence so the next person to touch it cannot miss what it is holding up.

## What was done?

**The mechanism.** `BatchTransitionAction::into_high_level_drive_operations` flattens every transition of a batch into one `Vec<DriveOperation>`, and `apply_drive_operations` turns that into a single GroveDB batch. Within that batch the ordinary document Add/Update/Delete conversions are blind to each other: `batch_delete_up_tree_while_empty` decides whether an index group has become empty from committed state plus only its own operations. Two operations that jointly empty a group each observe the other's document still committed, neither removes the group tree, and on a **ranked** index the leftover tree is mirrored into the aggregate secondary — so the group keeps ranking at zero with nothing behind it, sorting ahead of every group with a positive aggregate. Primary and secondary agree the empty group exists, so `verify_grovedb` is clean and the proof reconstructs the live root hash. **A wrong ranking that verifies.**

**Why it is not fixed here.** Fixing it at the correct site (the flatten loop in `apply_drive_operations_v0`, not the five `&mut None` call sites — those are a no-op) changes the number of emitted GroveDB operations, hence fees, hence identity balances, hence the app hash. It would also break `Drive::update_contract_keywords_operations`, which concatenates N deletes then M adds over a shared `byContractId` group and is correct *because* its deletes are blind — sibling awareness would make the second delete conclude the group is empty and delete it, then the adds would insert into a tree the same batch deleted. That needs a protocol-version gate and its own PR.

Landed instead:

- **The rationale**, as a doc comment on `SystemLimits::max_transitions_in_documents_batch`, with pointers from all three `SYSTEM_LIMITS_V*` constants and from the one hand-written `SystemLimits` literal in the mocks. It also names the two *other* guards holding the same line, because "the cap is the only thing" is not true in general:
  - the keyword path's caller guard (below);
  - `MultipleDocumentOperationsForSameContractDocumentType`, which *does* thread accumulated operations and is why the withdrawal paths batch many documents safely — and which is **not** a drop-in for batch transitions, because it carries no delete variant. That is the detail that makes the obvious fix non-obvious.
- **Cap tripwires**: a test asserting the value is 1 across `PLATFORM_VERSIONS`, and (under `mock-versions`) across the mock registry, since one mock hand-writes its `SystemLimits` and the main loop cannot see it.
- **`rs-drive-abci` reachability coverage** (`ranked_group_drain.rs`): a real signed two-transition `BatchTransition` is refused by the cap before any drive operation is built, and the reachable near-miss — two transitions draining the same group in one block — produces the correct index, because `execute_event` calls `apply_drive_operations` once per state transition.
- **`rs-drive` mechanism coverage** (`batched_group_drain.rs`): five cases that expose the defect land `#[ignore]`d rather than deleted or weakened, so raising the cap has something concrete to un-ignore. They fail today, for the right reason. Everything else in the file runs: a characterization test asserting the phantom exactly as it is on all three axes (including that it verifies and proves), the fail-loud inverse shapes, and sequential controls for both the delete and move paths.
- **The keyword path**, previously argued safe only by code reading, is now executed. Replacing a whole keyword set keeps the `byContractId` group holding exactly the new members. Clearing it entirely strands an empty group tree — and the only thing preventing that from the network is `!contract.keywords().is_empty()` in `update_contract_v1`. That guard is a shield, not a fix: it also skips the deletes, so a contract that clears its keywords keeps being returned by keyword search under the old ones. **Both halves are commented at the guard and pinned by characterization tests**, because removing the guard to fix the stale index would silently arm the stranded group tree.

## How Has This Been Tested?

Rebased onto `v4.2-dev` @ `f05bf82dc9` and re-run there. No conflicts (this branch touches no `Cargo.lock`).

| Gate | Result |
|---|---|
| `cargo clippy --workspace --all-features --all-targets` | exit 0, no findings |
| `cargo fmt --check --all` | exit 0 |
| `cargo test -p drive --lib` | 3333 passed, 0 failed, 5 ignored |
| `cargo test -p drive --lib batched_group_drain -- --ignored` | **5 failed, as designed** — the defect is real |
| `cargo test -p drive-abci --lib batch::tests` | 320 passed, 0 failed |
| `cargo test -p platform-version --lib` | 17 passed; 18 with `--features mock-versions` |

Not overstating this: **the clippy gate is green only after retries.** Two third-party build scripts download GitHub artifacts (`tenderdash-proto` fetching the Tenderdash source zip, `grovedb`'s `grovedbg` feature fetching a release zip) and both failed repeatedly from this machine with `Peer disconnected` / HTTP2 `REFUSED_STREAM`. I retried until the downloads succeeded, verified the grovedbg artifact against the SHA-256 the build script pins, seeded both build scripts' own documented caches, and then got a clean full-workspace run. Nothing was skipped or narrowed to work around it — the green run above is the real full-feature, all-targets workspace lint. Flagging it because it is the exact command a reviewer will re-run, and it may flake for them too.

Three independent review passes were run over the diff (correctness, adversarial test-quality, consensus/version-safety) and their must-fixes are folded in. Two are worth naming because the first version was genuinely weaker than it looked:

- The two "never lands silently wrong" tests took the rejection branch in all 24 combinations, so they asserted nothing beyond "returned `Err`". They now assert the refusal *kind* per batching configuration — Drive's pre-flight check when `batching_consistency_verification` is on, grovedb's batch applier on the shipped default, which is the only guard a real node has — and that a refused batch leaves state unchanged.
- The documentation's strongest claim (that the phantom passes integrity verification and proves against the live root hash) was prose only: every ignored test dies at its first assertion on the Count axis and never reaches the proof checks. That claim is now executed by a green test on all three axes.

One verification limit, stated rather than papered over: `clearing_a_contracts_keywords_leaves_the_old_ones_indexed` drives Drive-layer `update_contract`, not a signed `DataContractUpdate` end to end. Reachability of the stale-keyword behaviour from the network is strongly indicated by the call path but is not executed here.

## Breaking Changes

None. No wire format, serialization, fee, or state change. Every non-test hunk in this PR is a comment; no value, version mapping, or dispatch entry is touched, and the `1` literals themselves are unchanged.

## Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved ranked-index consistency when multiple document operations empty the same group.
  - Prevented stale or phantom entries after document deletions, updates, or keyword removal.
  - Preserved accurate ranked queries and proofs across count, sum, and average indexes.

- **Documentation**
  - Clarified the single-document batch limit and its role in maintaining state consistency.

- **Tests**
  - Added coverage for batched and sequential operations, index draining and refilling, keyword updates, and state integrity.

### DIFF
--- a/packages/rs-platform-version/src/version/system_limits/mod.rs
+++ b/packages/rs-platform-version/src/version/system_limits/mod.rs
@@ -39,9 +39,12 @@ pub struct SystemLimits {
     /// reader raising it needs to know what the other two are:
     ///
     /// * `Drive::update_contract_keywords_operations` puts N blind document deletes and M adds
-    ///   in one batch over a single shared index group. It stays correct only because its
-    ///   caller skips it entirely when the new keyword set is empty, so any batch that empties
-    ///   the group also refills it. That guard is load-bearing in the same way this cap is.
+    ///   in one batch over a single shared index group. Every batch it actually emits refills
+    ///   the group it empties, but only because its caller skips it outright when the new
+    ///   keyword set is empty — and that skip is a shield, not a fix. Called directly with an
+    ///   empty set it does strand the group, and the skip leaves the old keyword documents in
+    ///   place, so a contract that clears its keywords keeps being found under them. Both
+    ///   halves are pinned by tests; see the call site in `update_contract_v1`.
     /// * `DocumentOperationType::MultipleDocumentOperationsForSameContractDocumentType` threads
     ///   the accumulated operations through, so document operations in *that* variant do see
     ///   their siblings — which is why the withdrawal paths batch many documents safely. It is
--- a/packages/rs-platform-version/src/version/system_limits/mod.rs
+++ b/packages/rs-platform-version/src/version/system_limits/mod.rs
@@ -14,6 +14,42 @@ pub struct SystemLimits {
     ///
     /// NOTE: This must be equal to the `max-tx-bytes` in the Tenderdash config
     pub max_state_transition_size: u64,
+    /// Maximum number of batched transitions (document and token transitions counted together)
+    /// one batch state transition may carry.
+    ///
+    /// This cap is load-bearing for state correctness, not merely a size or throughput limit.
+    /// `BatchTransitionAction::into_high_level_drive_operations` flattens every transition of a
+    /// batch into one `Vec<DriveOperation>`, and `apply_drive_operations` turns that vector into
+    /// a single GroveDB batch. Within one such batch the ordinary document Add/Update/Delete
+    /// conversions are blind to each other: the check that decides whether an index group tree
+    /// has become empty and should be removed sees committed state plus only the operations of
+    /// its own conversion.
+    ///
+    /// While the cap is 1, no two document operations can share a GroveDB batch *by way of a
+    /// batch state transition*, so on that route the blindness has nothing to act on. Raise it
+    /// and two operations that jointly empty a group each observe the other's document still
+    /// committed, each conclude the group is not yet empty, and the group tree survives with no
+    /// documents behind it. On a ranked index that leftover tree is mirrored into the aggregate
+    /// secondary, so the group keeps ranking with a zero aggregate — sorting ahead of every
+    /// group with a positive one — and, because primary and secondary agree that the empty
+    /// group exists, the state is internally consistent: integrity verification passes and
+    /// proofs attest the wrong ranking against the live root hash.
+    ///
+    /// The cap is not the only thing standing between that machinery and a live path, and a
+    /// reader raising it needs to know what the other two are:
+    ///
+    /// * `Drive::update_contract_keywords_operations` puts N blind document deletes and M adds
+    ///   in one batch over a single shared index group. It stays correct only because its
+    ///   caller skips it entirely when the new keyword set is empty, so any batch that empties
+    ///   the group also refills it. That guard is load-bearing in the same way this cap is.
+    /// * `DocumentOperationType::MultipleDocumentOperationsForSameContractDocumentType` threads
+    ///   the accumulated operations through, so document operations in *that* variant do see
+    ///   their siblings — which is why the withdrawal paths batch many documents safely. It is
+    ///   not a drop-in for batch transitions: it carries no delete variant.
+    ///
+    /// Five cases in `rs-drive`'s `batched_group_drain` suite are `#[ignore]`d for exactly this
+    /// reason; the rest of that suite runs. Anyone raising this cap should un-ignore those five
+    /// first and make them pass.
     pub max_transitions_in_documents_batch: u16,
     pub withdrawal_transactions_per_block_limit: u16,
     pub retry_signing_expired_withdrawal_documents_per_block_limit: u16,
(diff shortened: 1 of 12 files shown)


thepastaclaw:
✅ Final review complete — no blockers (commit e2f44808)

shumkov:
Thanks for the review. Responses to the three nitpicks and the flagged finding, in order.

---

**`system_limits/mod.rs:84` — derive the registry-size floor instead of hardcoding `14`**

Not applying this one — I think it inverts the guard. `PlatformVersion::latest()` is `PLATFORM_VERSIONS.last()` (`protocol_version.rs:167`), not the `LATEST_PLATFORM_VERSION` constant, so the proposed `assert_eq!(PLATFORM_VERSIONS.len(), PlatformVersion::latest().protocol_version)` compares the array against its own last element. Truncate the registry to `[V1, V2, V3]` and it becomes `assert_eq!(3, 3)` and passes — which is precisely the case the assertion exists to catch.

The premise is also slightly off: a `>= 14` floor doesn't need bumping to stay *correct* when a protocol version is added, only to stay tight, and the loop below already covers every entry the registry does hold.

---

**`batched_group_drain.rs:1072-1235` — extract the shared driver for the two "never lands silently wrong" cases**

Leaving these as two explicit tests. They differ in the departing operation, the order labels, and the expected `departure_stored`, so a single driver ends up taking a boolean that selects the success expectation — which is the thing that makes a failure hard to read at 2am. The drift risk you name is real (both sets of assertions were strengthened together during review), but I'd rather pay it than obscure which case failed.

---

**`ranked_group_drain.rs:544-586` — extract the repeated platform and fixture setup**

Leaving as is. The substantive helpers are already extracted — `register_restaurants`, `create_visit`, `process_block`, `signed_delete_batch`. What's left is platform construction plus a seeding loop, and a combined fixture would have to hand back the `platform_state` guard so each caller can re-derive `platform_version` from it. That's lifetime plumbing in exchange for a few lines of setup.

---

**`system_limits/mod.rs:41-44` — verify the keyword-guard claim against the keyword-clearing behaviour** *(your flagged "unverified finding")*

Good catch, and correct — fixed in c05b9b3155. You inferred it from the two documents disagreeing and you were right: the guard-site comment in `update_contract_v1` already stated both halves, but this bullet still read as though the guard made the keyword path safe.

It now says that every batch the keyword update actually emits refills the group it empties, but that the skip is a shield rather than a fix — called directly with an empty set it does strand the group, and the skip leaves the previous keyword documents in place, so a contract that clears its keywords keeps being found under them.

packages/rs-platform-version/src/version/system_limits/mod.rs
```diff
@@ -34,8 +73,71 @@ pub struct SystemLimits {
...
+    /// been 1 since the first mainnet release; a version that relaxes it must be a deliberate,
+    /// reviewed decision rather than a copy-paste into a new `SYSTEM_LIMITS_V*`.
+    #[test]
+    fn documents_batch_is_capped_at_one_transition_at_every_protocol_version() {
+        // Without this the loop below asserts nothing if the registry is ever
+        // emptied or truncated.
+        assert!(
+            PLATFORM_VERSIONS.len() >= 14,
+            "the protocol version registry lost entries; this test only covers what it holds"
+        );
```
thepastaclaw: **🟡 Suggestion: Derive the registry completeness check from LATEST_VERSION**

The fixed `>= 14` floor only detects truncation below the current registry size. If protocol version 15 is declared by updating `LATEST_VERSION` but `PLATFORM_V15` is accidentally omitted from `PLATFORM_VERSIONS`, the registry will still contain 14 entries, this assertion will pass, and the loop will never inspect version 15. Compare the registry length with the independent, authoritative `LATEST_VERSION` constant so the test continues to verify every declared protocol version. Unlike `PlatformVersion::latest()`, that constant is not derived from the registry being checked.

```suggestion
assert_eq!(
    PLATFORM_VERSIONS.len(),
    crate::version::LATEST_VERSION as usize,
    "the protocol version registry does not contain every declared protocol version"
);
```

<sub>source: ['codex']</sub>

thepastaclaw (commented):

The PR documents and characterizes the existing batch-index defect without changing consensus behavior, but the new registry-wide cap test does not remain complete when another protocol version is declared. Its independent completeness assertion should use the authoritative latest protocol version rather than a fixed floor.
Source: reviewer backend gpt-5.6-sol; final verifier backend gpt-5.6-sol. Orchestration-only: openclaw-agent/cliproxy/gpt-5.6-sol (not reviewer evidence).

### Review provenance

🟡 1 suggestion(s)

<details>

```

In `packages/rs-platform-version/src/version/system_limits/mod.rs`:
- [SUGGESTION] packages/rs-platform-version/src/version/system_limits/mod.rs:87-90: Derive the registry completeness check from LATEST_VERSION
  The fixed `>= 14` floor only detects truncation below the current registry size. If protocol version 15 is declared by updating `LATEST_VERSION` but `PLATFORM_V15` is accidentally omitted from `PLATFORM_VERSIONS`, the registry will still contain 14 entries, this assertion will pass, and the loop will never inspect version 15. Compare the registry length with the independent, authoritative `LATEST_VERSION` constant so the test continues to verify every declared protocol version. Unlike `PlatformVersion::latest()`, that constant is not derived from the registry being checked.
```
</details>

thepastaclaw (approved):

The sole prior suggestion is fixed at the exact head: the cap test now compares `PLATFORM_VERSIONS.len()` against the independently declared `LATEST_VERSION`, so a declared but unregistered protocol version fails the completeness guard. No additional in-scope findings were reported or identified, and the PR remains documentation-and-test-only with respect to runtime behavior.
Source: Reviewer backend gpt-5.6-sol; final verifier backend gpt-5.6-sol. Orchestration-only: openclaw-agent/cliproxy/gpt-5.6-sol (not reviewer evidence).

### Review provenance

packages/rs-platform-version/src/version/system_limits/mod.rs
thepastaclaw: Resolved in `e2f4480` — _Derive the registry completeness check from LATEST_VERSION_ no longer present.

PR #4385, opened 2026-08-12
perf(swift-sdk): linear wallet-changeset rounds via per-round bulk-prefetch cache

PastaPastaPasta:

Restoring a wallet with a large transaction history made the app pin a CPU core for hours and grow memory without bound until the OS killed it (observed: **59 GB footprint** on a mainnet wallet whose SPV scan matches ~8,000 transactions, with only 3,884 of them ever reaching disk).

The root cause is how a persistence round applies its rows. Each Rust `store()` round maps to one `beginChangeset` → per-kind callbacks → `endChangeset` bracket, with a single `save()` at the end. During SPV catch-up one round can carry thousands of transaction records, and the apply helpers (`upsertTransaction`, `resolveInputOutpoint`, `upsertUtxo`, `markUtxoSpent`, …) issued an individual `ModelContext.fetch` for every row, every input, and every UTXO. SwiftData evaluates each of those fetches against **all objects staged so far in the unsaved round**, so the more rows a round had already staged, the more expensive every following fetch became:

- fetch # 1 scans ~0 staged objects, fetch # 100,000 scans ~100,000 → total cost grows with the **square** of the round size;
- measured: ~2.3 µs × (staged objects) per fetch — the first 1,000 upserts took 1.3 s, the eighth 1,000 took 20.9 s;
- an 8k-record round with per-input work extrapolates to **hours of pinned CPU**, which is why the persistence drain stalled and the app died before finishing.

## What was done?

One idea, applied consistently: **fetch once per round, not once per row.**

- `PlatformWalletPersistenceHandler.persistWalletChangeset` now builds a `WalletChangesetRoundCache` before applying anything: it walks the changeset once, collects every txid / outpoint / address the round could touch, and bulk-fetches the matching `PersistentTransaction` / `PersistentTxo` / `PersistentPendingInput` / `PersistentCoreAddress` rows with chunked `IN` predicates (≤900 keys per chunk, under SQLite's bind-variable limit).
- All apply helpers (`upsertTransaction`, `resolveInputOutpoint`, `removePendingInputs`, `upsertUtxo`, `markUtxoSpent`, `markUtxoInstantLocked`) look rows up in the cache dictionaries instead of fetching. Inserts and deletes update the cache in place, so later rows in the same batch observe them exactly as they previously observed staged objects through per-row fetches.
- A key the prefetch covered but found no row for is an authoritative miss; the rare key discovered mid-round (e.g. a stale pending row's `spendingTxid` from a prior session) falls back to a single-row fetch.
- `persistAccountAddresses` gets the same treatment — its per-address row fetch and per-address TXO-backfill fetch (a second hot loop in the same rounds during restore) are now two chunked bulk fetches.

Result: a 4,000-record round drops from minutes to **under a second**, and the end-to-end restore that previously died at 59 GB completes a full mainnet genesis→tip sync in ~16 minutes with a ~1.2 GB peak (header download, not persistence; ~430 MB settled).

## How Has This Been Tested?

New unit tests (`swift test`, 354 passing):
- `BulkFetchPredicateTests` — pins the two SwiftData behaviors the cache depends on: `[Data].contains($0.column)` translating to SQL `IN` with >900 keys chunked, and staged (unsaved) rows staying visible to bulk fetches.
- `WalletChangesetRoundTests` — drives real `WalletChangeSetFFI` structs through a full begin→persist→end round: a same-round chain of spends resolves every TXO↔spender linkage and drains all pending-input rows; an input with unknown funding still writes its pending-input row (the out-of-order spend-repair mechanism); and a scaling regression test asserts a 4× larger round costs near-linearly more (fails on any quadratic regression).
- `FFIFixtures` — shared test helpers (deduplicates `tuple32` copies that existed in `DashPayPersistenceTests`).

Manual end-to-end: restored a mainnet wallet reproducing the incident workload (~8k matched transactions) in SwiftExampleApp on the iOS simulator. Full chain scan completed in ~16 minutes; all matched transactions and TXOs durably persisted; sync watermark reached the chain tip; memory sampled every 30 s never exceeded ~1.25 GB; app restart came back clean with the watermark intact.

## Breaking Changes

None. No public API or schema changes; the persistence semantics (round atomicity, pending-input repair, spend gating) are unchanged — only the lookup strategy inside a round.

## Checklist:

## Summary by CodeRabbit

* **Performance**
  * Improved wallet synchronization and transaction processing efficiency, especially for larger wallets.
  * Reduced repeated data lookups during wallet updates and address reconciliation.

* **Reliability**
  * Improved handling of pending inputs, transaction links, unspent outputs, and wallet changeset updates.
  * Ensured unknown funding inputs remain available for later reconciliation.

* **Testing**
  * Added coverage for bulk data retrieval, wallet changeset processing, pending-input cleanup, and performance scaling.

### DIFF
--- a/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
@@ -827,13 +827,20 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
                     let tx = txsPtr[t]
                     let txid = hashData(tx.txid)
                     cache.prefetchedTxids.insert(txid)
+                    // Only the input OUTPOINTS are collected here — the
+                    // apply helpers look inputs up as TXOs / pending
+                    // rows, never as transactions, so pulling every
+                    // prevout's parent tx row would only inflate the
+                    // `IN` fetch (hundreds of foreign parents per
+                    // CoinJoin record).
                     if let inPtr = tx.input_outpoints, tx.input_outpoints_count > 0 {
                         for j in 0..<Int(tx.input_outpoints_count) {
                             let entry = inPtr[j]
-                            let prevTxid = hashData(entry.txid)
-                            cache.prefetchedTxids.insert(prevTxid)
                             cache.prefetchedOutpoints.insert(
-                                PersistentTxo.makeOutpoint(txid: prevTxid, vout: entry.vout)
+                                PersistentTxo.makeOutpoint(
+                                    txid: hashData(entry.txid),
+                                    vout: entry.vout
+                                )
                             )
                         }
                     }
@@ -872,34 +879,48 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
             }
         }
 
+        // A failed chunk fetch must NOT leave its keys in the
+        // `prefetched*` sets: membership there declares dictionary
+        // absence authoritative, so an error would silently read as
+        // "these ~900 rows don't exist" and the upsert paths would
+        // insert duplicates over `.unique` columns. Dropping the
+        // chunk's keys instead routes every lookup through the
+        // single-row fallback fetch — the pre-cache behavior.
         for chunk in Self.chunked(Array(cache.prefetchedTxids)) {
             let descriptor = FetchDescriptor<PersistentTransaction>(
                 predicate: #Predicate { chunk.contains($0.txid) }
             )
-            for row in (try? backgroundContext.fetch(descriptor)) ?? [] {
-                cache.transactions[row.txid] = row
+            if let rows = try? backgroundContext.fetch(descriptor) {
+                for row in rows { cache.transactions[row.txid] = row }
+            } else {
+                cache.prefetchedTxids.subtract(chunk)
             }
         }
         for chunk in Self.chunked(Array(cache.prefetchedOutpoints)) {
             let txoDescriptor = FetchDescriptor<PersistentTxo>(
                 predicate: #Predicate { chunk.contains($0.outpoint) }
             )
-            for row in (try? backgroundContext.fetch(txoDescriptor)) ?? [] {
-                cache.txos[row.outpoint] = row
-            }
             let pendingDescriptor = FetchDescriptor<PersistentPendingInput>(
                 predicate: #Predicate { chunk.contains($0.outpoint) }
             )
-            for row in (try? backgroundContext.fetch(pendingDescriptor)) ?? [] {
-                cache.pendingInputs[row.outpoint, default: []].append(row)
+            if let txoRows = try? backgroundContext.fetch(txoDescriptor),
+               let pendingRows = try? backgroundContext.fetch(pendingDescriptor) {
+                for row in txoRows { cache.txos[row.outpoint] = row }
+                for row in pendingRows {
+                    cache.pendingInputs[row.outpoint, default: []].append(row)
+                }
+            } else {
+                cache.prefetchedOutpoints.subtract(chunk)
             }
         }
         for chunk in Self.chunked(Array(cache.prefetchedAddresses)) {
             let descriptor = FetchDescriptor<PersistentCoreAddress>(
                 predicate: #Predicate { chunk.contains($0.address) }
             )
-            for row in (try? backgroundContext.fetch(descriptor)) ?? [] {
-                cache.coreAddresses[row.address] = row
+            if let rows = try? backgroundContext.fetch(descriptor) {
+                for row in rows { cache.coreAddresses[row.address] = row }
+            } else {
+                cache.prefetchedAddresses.subtract(chunk)
             }
         }
 
@@ -3394,12 +3415,20 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
         let allAddresses = entries.map(\.address)
         var existingRows: [String: PersistentCoreAddress] = [:]
         var txosByAddress: [String: [PersistentTxo]] = [:]
+        // Addresses whose bulk row fetch FAILED (threw) — a miss for
+        // these is not authoritative, so the upsert loop falls back to
+        // a single-row fetch instead of inserting over the `.unique`
+        // address column. A failed TXO-backfill fetch just skips the
+        // backfill for the chunk, matching the old per-row `try?`.
+        var unresolvedAddresses: Set<String> = []
         for chunk in Self.chunked(allAddresses) {
             let rowDescriptor = FetchDescriptor<PersistentCoreAddress>(
                 predicate: #Predicate { chunk.contains($0.address) }
             )
-            for row in (try? backgroundContext.fetch(rowDescriptor)) ?? [] {
-                existingRows[row.address] = row
+            if let rows = try? backgroundContext.fetch(rowDescriptor) {
+                for row in rows { existingRows[row.address] = row }
+            } else {
+                unresolvedAddresses.formUnion(chunk)
             }
             let txoDescriptor = FetchDescriptor<PersistentTxo>(
                 predicate: #Predicate { chunk.contains($0.address) }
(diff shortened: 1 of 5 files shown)


thepastaclaw:
🕓 Ready for review — 2 ahead in queue (commit 25dfd8c7)
_Queue position: 3/8 · 2 reviews active_
_ETA: start ~00:52 UTC · complete ~01:08 UTC (median 16m across 30 recent reviews; 2 slots)_
_Queued 26m ago · Last checked: 2026-08-13 00:30 UTC_

thepastaclaw (commented):

The cache preserves the staged-row behavior on normal reads, but two in-scope suggestions remain: bulk-fetch failures must not become authoritative misses, and input funding transaction IDs should not widen transaction prefetches when only their outpoints are consumed. No blocking issue was verified.
Source: reviewer backend model gpt-5.6-sol; final verifier backend model gpt-5.6-sol; CodeRabbit inline findings independently verified. openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

### Review provenance

🟡 2 suggestion(s)

<details>

```

- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift:875-904: Do not turn failed bulk fetches into authoritative cache misses
  Each bulk fetch uses `try?` and substitutes an empty array, while the chunk's keys remain in the corresponding `prefetched*` set. A fetch failure is therefore indistinguishable from a successful query that found no rows: `cachedTransaction`, `cachedTxo`, `cachedPendingInputs`, and `cachedCoreAddress` suppress their fallback lookups and can insert duplicates, leave existing pending inputs unresolved, or omit address relationships. The address-persistence prefetch at lines 3397–3410 has the same amplification: one failed chunk is treated as hundreds of missing addresses or TXOs. Handle each error explicitly by either failing the persistence round or removing the affected keys from authoritative coverage so row-wise lookups can retry; for outpoints, retain coverage only when both the TXO and pending-input queries succeeded.
- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift:833-834: Avoid prefetching transaction rows for every input parent
  An input's previous txid is only used to construct the outpoint passed to `cachedTxo`; none of the input-resolution helpers looks up the funding `PersistentTransaction`. Transaction rows that are actually consumed are already covered by transaction records, added UTXOs, and `spending_txid` values. Adding every input parent therefore widens the chunked transaction `IN` queries and can materialize unrelated historical transaction rows on the input-heavy restore path this PR is optimizing. Keep the outpoint prefetch but omit the transaction-prefetch insertion.
```
</details>

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -766,6 +766,222 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
...
+            }
+        }
+        for chunk in Self.chunked(Array(cache.prefetchedAddresses)) {
+            let descriptor = FetchDescriptor<PersistentCoreAddress>(
+                predicate: #Predicate { chunk.contains($0.address) }
+            )
+            for row in (try? backgroundContext.fetch(descriptor)) ?? [] {
+                cache.coreAddresses[row.address] = row
+            }
+        }
```
thepastaclaw, line 925: **🟡 Suggestion: Do not turn failed bulk fetches into authoritative cache misses**

Each bulk fetch uses `try?` and substitutes an empty array, while the chunk's keys remain in the corresponding `prefetched*` set. A fetch failure is therefore indistinguishable from a successful query that found no rows: `cachedTransaction`, `cachedTxo`, `cachedPendingInputs`, and `cachedCoreAddress` suppress their fallback lookups and can insert duplicates, leave existing pending inputs unresolved, or omit address relationships. The address-persistence prefetch at lines 3397–3410 has the same amplification: one failed chunk is treated as hundreds of missing addresses or TXOs. Handle each error explicitly by either failing the persistence round or removing the affected keys from authoritative coverage so row-wise lookups can retry; for outpoints, retain coverage only when both the TXO and pending-input queries succeeded.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -766,6 +766,222 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
...
+            if acc.transactions_count > 0, let txsPtr = acc.transactions {
+                for t in 0..<Int(acc.transactions_count) {
+                    let tx = txsPtr[t]
+                    let txid = hashData(tx.txid)
+                    cache.prefetchedTxids.insert(txid)
+                    if let inPtr = tx.input_outpoints, tx.input_outpoints_count > 0 {
+                        for j in 0..<Int(tx.input_outpoints_count) {
+                            let entry = inPtr[j]
+                            let prevTxid = hashData(entry.txid)
+                            cache.prefetchedTxids.insert(prevTxid)
```
thepastaclaw: **🟡 Suggestion: Avoid prefetching transaction rows for every input parent**

An input's previous txid is only used to construct the outpoint passed to `cachedTxo`; none of the input-resolution helpers looks up the funding `PersistentTransaction`. Transaction rows that are actually consumed are already covered by transaction records, added UTXOs, and `spending_txid` values. Adding every input parent therefore widens the chunked transaction `IN` queries and can materialize unrelated historical transaction rows on the input-heavy restore path this PR is optimizing. Keep the outpoint prefetch but omit the transaction-prefetch insertion.

```suggestion
let prevTxid = hashData(entry.txid)
```

<sub>source: ['codex']</sub>

packages/swift-sdk/SwiftTests/SwiftDashSDKTests/WalletChangesetRoundTests.swift
```diff
@@ -0,0 +1,237 @@
...
+        _ = try measureRound(count: 50)
+
+        let small = try measureRound(count: 1_000)
+        let large = try measureRound(count: 4_000)
+        XCTAssertLessThan(
+            large,
+            max(small, 0.05) * 10,
+            "4× records cost \(large)s vs \(small)s — superlinear scaling regression"
+        )
+    }
```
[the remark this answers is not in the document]
thepastaclaw: The 50 ms floor provides extra absolute headroom: when 1,000 records finish below 50 ms, a linear 4,000-record round should be below roughly 200 ms but receives a 500 ms budget. Production timing from another device does not establish that this runner's threshold is too tight, while skipping the assertion on fast machines would disable the regression guard.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
PastaPastaPasta: Fixed in 25dfd8c7b4. A failed chunk fetch now removes its keys from the prefetched set (round cache) — or records the addresses for a single-row fallback fetch in persistAccountAddresses — so a thrown bulk fetch degrades to the pre-cache per-row behavior instead of reading as an authoritative miss.

---
🤖 Posted autonomously by Claude on behalf of pasta.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
PastaPastaPasta: Done in 25dfd8c7b4 — input prevout txids are no longer collected into the transaction prefetch. They were only needed by the CoinJoin funding-known gate, which was split out of this PR into a follow-up; that PR will reintroduce the collection alongside its consumer.

---
🤖 Posted autonomously by Claude on behalf of pasta.

PastaPastaPasta:
Superseded by #4392 — same branch pushed in-repo so the full CI matrix (including the fork-gated Swift SDK job) runs. All review feedback from this PR is already incorporated there (head 25dfd8c7b4).

---
🤖 Posted autonomously by Claude on behalf of pasta.

PR #4391, opened 2026-08-13
feat(drive)!: multiple IN clauses on consecutive index properties in document queries

QuantumExplorer:

Drive's document-query grammar allows at most one `IN` clause per query — `WhereClause::group_clauses` rejects a second one with `MultipleInClauses`, treating `IN` as a range-class operator. But an `IN` over several *consecutive* properties of a compound index is not a range at all: it is a bounded cross-product of point lookups — `|list1| × |list2|` subtrees of the index — a shape grovedb path queries natively express (a key set at one path level, per-key subqueries carrying another key set at the next level), proof generation included. The single-`IN` cap was a drive grammar restriction, not a storage limitation.

This PR relaxes the grammar so queries like

```sql
WHERE identityId IN [...] AND class IN [math, physics]
ORDER BY identityId ASC, class ASC
```

work on a compound index over those properties. Plain document queries only (`SELECT` documents) — the grouped-aggregate surfaces (count/sum/average/ranked) keep rejecting multiple `IN`s.

## What was done?

**Grammar (structural, unversioned).** `InternalClauses.in_clause: Option<WhereClause>` became `in_clauses: Vec<WhereClause>`, and `group_clauses` now groups any number of `IN` clauses (still rejecting duplicate fields and equality/`IN` overlap). The field rename forced a compile-time audit of every consumer: filters, uniqueness validation, withdrawal queries, data triggers, wasm-drive-verify (whose JS `in_clause` key stays accepted for back-compat, with a new `in_clauses` array form), and the CBOR/gRPC round-trip `From` impl, which now emits all `IN` clauses.

**Consensus gate (protocol version 14).** Which query shapes are accepted is part of the consensus query contract, so acceptance is decided at path-query lowering — the single choke point shared by execution, proof generation, and client proof verification (`construct_path_query*` → `get_non_primary_key_path_query`). Following the repo's versioned-module convention, a new `DriveDocumentQueryMethodVersions.non_primary_key_path_query` feature version dispatches the lowering:

- **v0** (all tables through protocol version 13): rejects more than one `IN` with the historical `MultipleInClauses` error.
- **v1** (`DRIVE_DOCUMENT_METHOD_VERSIONS_V4`, protocol version 14, unreleased): lowers multiple `IN`s to a multi-level key-set path query. Single-`IN` queries route through the v0 body under both versions, byte-identically.

**v1 semantics (deliberately conservative):**

- The `IN` clauses must sit on **consecutive index properties** immediately after the equality prefix; an optional single range clause may follow the last `IN`. Index selection only considers conforming indexes (the existing `Index::matches` tail and order-by continuity rules still apply, with the deepest `IN` field playing the in-field role).
- Every `IN`'d property and the trailing range property require an `orderBy` entry; results return in index traversal order with per-level direction (grovedb supports mixed asc/desc per level).
- **Cross-product cap:** `Π |list_i| ≤ 100` (`defaults::MAX_IN_CROSS_PRODUCT_SIZE`) — the same worst-case branch enumeration as one maximal single `IN`; each list keeps its 100-value cap.
- **`startAt`/`startAfter` with more than one `IN` is rejected** (`Unsupported`) rather than shipped broken: the existing cross-branch cursor machinery bakes the cursor's per-level start keys into the *default* subquery applied to every sibling branch, which is only correct under a single-branch (equality) ancestry.
- Fees need no new accounting: processing fees derive from the operations of the actual grovedb traversal, so cost scales with the enumerated branches.

The count/sum dispatcher's shared validator gained an explicit multi-`IN` guard so the aggregate surfaces keep their existing contract.

## How Has This Been Tested?

- **Grammar units** (`conditions.rs`): multiple `IN`s on distinct fields group structurally; same-field `IN`s and equality overlap still reject.
- **Execution + proof round-trips** (`query_tests.rs`, family compound indexes): 2-`IN`, 3-`IN`, equality prefix + 2-`IN` (on the 4-property index), 2-`IN` + trailing range, and a descending first level — each cross-checked against a brute-force filter over all stored documents and round-tripped through `execute_with_proof_only_get_elements`, asserting the verified root hash equals the live grovedb root hash and proof results equal no-proof results.
- **Rejections:** protocol version 13 rejects the same query with `MultipleInClauses` on both the no-proof and prove paths (and v14 accepts it); 120-branch cross product; non-consecutive `IN` properties; cursor pagination; missing `orderBy` on an `IN` field.
- **Wire level** (`drive-abci` `getDocuments` v1 handler): multi-`IN` documents select returns the expected documents and a proof at protocol version 14, and surfaces `MultipleInClauses` as a query error at protocol version 13.
- **Version tables:** a freeze test pins `non_primary_key_path_query` to 0 at v13 and 1 at v14.
- Suites run with real exit codes: `cargo check --workspace --all-targets`, full `drive` tests (`server,verify,cbor_query`), `drive-abci` document-query tests, `drive-proof-verifier`, `dash-sdk` lib tests, `platform-version` tests, and clippy on `drive` — all clean.

## Breaking Changes

Consensus query contract: protocol version 14 nodes accept a query shape (multiple `IN` clauses) that v13 nodes reject, gated behind the new versioned lowering so mixed-version networks agree until the upgrade activates. Rust API: `InternalClauses.in_clause` is now `in_clauses: Vec<WhereClause>` (wasm-drive-verify keeps accepting the JS `in_clause` key).

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added support for multiple `IN` clauses in compound document queries starting with protocol version 14.
  * Added support for equality prefixes, trailing ranges, ordering, proof verification, and cross-product limits.
  * Added protocol-version-aware withdrawal lookups by transaction indices.
  * Preserved compatibility with legacy single-`IN` query inputs.

* **Bug Fixes**
  * Improved validation for unsupported, malformed, duplicate, or incorrectly ordered query clauses.
  * Count queries and cursor-based requests now correctly reject unsupported multiple-`IN` combinations.

* **Documentation**
  * Updated query restrictions and protocol-version guidance for multiple `IN` clauses.


thepastaclaw:
🕓 Ready for review — next in queue (commit c43cb109)
_Queue position: 1/1_

packages/rs-drive/src/query/mod.rs
```diff
@@ -2076,54 +2212,284 @@ impl<'a> DriveDocumentQuery<'a> {
...
+                }
+            };
+        Self::recursive_insert_on_query(
+            &mut child_query,
+            left_over_index_properties.as_slice(),
+            index.unique,
+            None,
+            deepest_left_to_right,
+            Some(&self.order_by),
+            platform_version,
```
thepastaclaw: **🔴 Blocking: Descending order on the first leftover index property is ignored**

The v1 multi-IN path delegates the index tail to `recursive_insert_on_query`. That helper computes the requested direction for the first leftover property at lines 1984-1991, but its no-cursor branch constructs that level with `Query::new_with_direction(first.ascending)` at line 2086 instead. An accepted query over `[a, b, c]`, such as `a IN (...) AND b IN (...) ORDER BY a ASC, b ASC, c DESC`, therefore traverses `c` in the index's ascending direction. `Index::matches` permits this shape because the deepest IN field is penultimate and the three order fields are continuous. Implement the requested tail direction in the v1 lowering without editing historical v0 behavior in place, and add execution/proof coverage for descending order on the leftover property.

<sub>source: ['codex']</sub>

packages/rs-drive/src/query/conditions.rs
```diff
@@ -601,25 +606,21 @@ impl<'a> WhereClause {
...
-                    Ok(Some(clause.clone()))
+                    Ok(clause)
                 }
-            }
-            _ => Err(Error::Query(QuerySyntaxError::MultipleInClauses(
-                "There should only be one in clause",
-            ))),
-        }?;
+            })
+            .collect::<Result<Vec<WhereClause>, Error>>()?;
```
thepastaclaw: **🔴 Blocking: The unversioned grammar changes v13 multi-IN errors**

Before this PR, `group_clauses` checked the number of non-primary-key IN clauses before checking field overlap, so every query with more than one such clause returned `MultipleInClauses`. The new unversioned loop instead returns `DuplicateNonGroupableClauseSameField` when two IN clauses share a field or when one overlaps an equality clause. Parsing completes before `get_non_primary_key_path_query_v0` dispatches, so the v0 length guard cannot preserve the historical protocol-v13 result. Preserve duplicate and overlap information structurally until version-aware validation: v0 must reject any multi-IN shape with `MultipleInClauses`, while v1 can apply the new duplicate/overlap checks.

<sub>source: ['codex']</sub>

packages/rs-drive/src/query/mod.rs
```diff
@@ -2076,54 +2212,284 @@ impl<'a> DriveDocumentQuery<'a> {
...
+        starts_at_document: Option<(Document, bool)>,
+        platform_version: &PlatformVersion,
+    ) -> Result<PathQuery, Error> {
+        // Conservative v1: the cross-branch cursor machinery is not wired
+        // for key-set branching at more than one level, so reject cursors
+        // instead of shipping silently wrong pagination.
+        if starts_at_document.is_some() || self.start_at.is_some() {
+            return Err(Error::Query(QuerySyntaxError::Unsupported(
+                "startAt/startAfter is not supported with multiple in clauses".to_string(),
+            )));
```
thepastaclaw: **🔴 Blocking: Cursor rejection occurs after cursor storage or proof processing**

The multi-IN cursor check runs only inside path-query lowering. `construct_path_query_operations` first reads and deserializes `self.start_at` from GroveDB at lines 1212-1268, while proof verification first calls `verify_start_at_document_in_proof` before constructing the query. Consequently, a v14 request for an unsupported multi-IN cursor shape can return `StartDocumentNotFound` or a proof error instead of `Unsupported`, and an existing cursor performs unnecessary state/proof work. Add a shared version-aware shape preflight before cursor lookup or proof extraction. It must retain the historical v0 precedence by returning `MultipleInClauses` before processing a cursor.

<sub>source: ['codex']</sub>

packages/rs-drive/src/query/mod.rs
```diff
@@ -2076,54 +2212,284 @@ impl<'a> DriveDocumentQuery<'a> {
...
+
+    #[cfg(any(feature = "server", feature = "verify"))]
+    /// v1 of the non-primary-key path query lowering (protocol version 14):
+    /// accepts multiple `In` clauses. Single-`In` shapes lower exactly as v0.
+    fn get_non_primary_key_path_query_v1(
+        &self,
+        document_type_path: Vec<Vec<u8>>,
+        starts_at_document: Option<(Document, bool)>,
+        platform_version: &PlatformVersion,
+    ) -> Result<PathQuery, Error> {
```
thepastaclaw: **🔴 Blocking: Consensus-critical versions are implemented inline instead of versioned modules**

The dispatcher selects feature versions correctly, but `get_non_primary_key_path_query_v0`, `get_non_primary_key_path_query_v1`, and the v1-specific lowering remain in the monolithic `query/mod.rs`. Consensus-versioned Drive methods must keep dispatch in `mod.rs` and implementations in separate `v0/mod.rs` and `v1/mod.rs` modules so already-live behavior is isolated from later edits. Move both implementations into version directories and leave only dispatch at the parent boundary.

<sub>source: ['codex']</sub>

packages/wasm-drive-verify/src/document/verify_proof.rs
```diff
@@ -156,10 +156,24 @@ fn parse_internal_clauses(where_clauses: &JsValue) -> Result<InternalClauses, Js
...
+
+    // Parse in_clauses (array form; protocol version 14+ accepts several)
+    if let Ok(clauses) = Reflect::get(&obj, &JsValue::from_str("in_clauses")) {
+        if !clauses.is_null() && !clauses.is_undefined() {
+            let clauses_array = Array::from(&clauses);
+            for i in 0..clauses_array.length() {
+                internal_clauses
+                    .in_clauses
+                    .push(parse_where_clause(&clauses_array.get(i))?);
+            }
```
thepastaclaw, line 182: **🟡 Suggestion: Reject non-array in_clauses values in WASM proof parsers**

`Array::from` accepts array-like values rather than requiring a JavaScript array; for example, `{}` becomes an empty array. A caller that supplies malformed `in_clauses` can therefore have its constraints silently discarded, causing the verifier to reconstruct and verify a broader query than requested. Require `Array::is_array(&clauses)` before conversion in this parser and in `verify_proof_keep_serialized.rs` and `verify_start_at_document_in_proof.rs`, returning an invalid-input error for non-array values.

packages/rs-drive/SECONDARY_INDEX_QUERIES.md
```diff
@@ -32,7 +32,7 @@ The `InternalClauses` struct organizes WHERE clauses into categories:
 pub struct InternalClauses {
     pub primary_key_in_clause: Option<WhereClause>,      // IN clause on $id
     pub primary_key_equal_clause: Option<WhereClause>,   // == clause on $id
-    pub in_clause: Option<WhereClause>,                  // Single IN clause on indexed field
+    pub in_clauses: Vec<WhereClause>,                    // IN clauses on indexed fields (several allowed from protocol version 14)
```
thepastaclaw: **🟡 Suggestion: Scope the documented v14 multi-IN allowance to plain document queries**

The documentation currently says several indexed-field IN clauses are allowed from protocol version 14 without limiting that statement to plain document queries. The PR intentionally keeps count, sum, average, and ranked/grouped aggregate surfaces on the single-IN contract. Identify these as non-primary-key IN clauses for plain document queries here and in the restrictions section at lines 241-246, and explicitly state that grouped aggregate queries continue to reject multiple IN clauses.

thepastaclaw (changes requested):

The protocol-v14 multi-IN lowering has four blocking issues: it ignores descending order on the first leftover index property, changes historical v13 error precedence before versioned dispatch, rejects cursor combinations only after storage/proof work, and keeps consensus-versioned implementations inline rather than in immutable version modules. Two additional suggestions address misleading scope documentation and permissive WASM parsing that can silently broaden a proof query.
Source: reviewer backend model gpt-5.6-sol; final verifier backend model gpt-5.6-sol; openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

### Review provenance

<details>

```

In `packages/rs-drive/src/query/mod.rs`:
- [BLOCKING] packages/rs-drive/src/query/mod.rs:2371-2378: Descending order on the first leftover index property is ignored
  The v1 multi-IN path delegates the index tail to `recursive_insert_on_query`. That helper computes the requested direction for the first leftover property at lines 1984-1991, but its no-cursor branch constructs that level with `Query::new_with_direction(first.ascending)` at line 2086 instead. An accepted query over `[a, b, c]`, such as `a IN (...) AND b IN (...) ORDER BY a ASC, b ASC, c DESC`, therefore traverses `c` in the index's ascending direction. `Index::matches` permits this shape because the deepest IN field is penultimate and the three order fields are continuous. Implement the requested tail direction in the v1 lowering without editing historical v0 behavior in place, and add execution/proof coverage for descending order on the leftover property.
- [BLOCKING] packages/rs-drive/src/query/mod.rs:2286-2292: Cursor rejection occurs after cursor storage or proof processing
  The multi-IN cursor check runs only inside path-query lowering. `construct_path_query_operations` first reads and deserializes `self.start_at` from GroveDB at lines 1212-1268, while proof verification first calls `verify_start_at_document_in_proof` before constructing the query. Consequently, a v14 request for an unsupported multi-IN cursor shape can return `StartDocumentNotFound` or a proof error instead of `Unsupported`, and an existing cursor performs unnecessary state/proof work. Add a shared version-aware shape preflight before cursor lookup or proof extraction. It must retain the historical v0 precedence by returning `MultipleInClauses` before processing a cursor.
- [BLOCKING] packages/rs-drive/src/query/mod.rs:2220-2259: Consensus-critical versions are implemented inline instead of versioned modules
  The dispatcher selects feature versions correctly, but `get_non_primary_key_path_query_v0`, `get_non_primary_key_path_query_v1`, and the v1-specific lowering remain in the monolithic `query/mod.rs`. Consensus-versioned Drive methods must keep dispatch in `mod.rs` and implementations in separate `v0/mod.rs` and `v1/mod.rs` modules so already-live behavior is isolated from later edits. Move both implementations into version directories and leave only dispatch at the parent boundary.

In `packages/rs-drive/src/query/conditions.rs`:
- [BLOCKING] packages/rs-drive/src/query/conditions.rs:609-623: The unversioned grammar changes v13 multi-IN errors
  Before this PR, `group_clauses` checked the number of non-primary-key IN clauses before checking field overlap, so every query with more than one such clause returned `MultipleInClauses`. The new unversioned loop instead returns `DuplicateNonGroupableClauseSameField` when two IN clauses share a field or when one overlaps an equality clause. Parsing completes before `get_non_primary_key_path_query_v0` dispatches, so the v0 length guard cannot preserve the historical protocol-v13 result. Preserve duplicate and overlap information structurally until version-aware validation: v0 must reject any multi-IN shape with `MultipleInClauses`, while v1 can apply the new duplicate/overlap checks.

In `packages/wasm-drive-verify/src/document/verify_proof.rs`:
- [SUGGESTION] packages/wasm-drive-verify/src/document/verify_proof.rs:168-176: Reject non-array in_clauses values in WASM proof parsers
  `Array::from` accepts array-like values rather than requiring a JavaScript array; for example, `{}` becomes an empty array. A caller that supplies malformed `in_clauses` can therefore have its constraints silently discarded, causing the verifier to reconstruct and verify a broader query than requested. Require `Array::is_array(&clauses)` before conversion in this parser and in `verify_proof_keep_serialized.rs` and `verify_start_at_document_in_proof.rs`, returning an invalid-input error for non-array values.

In `packages/rs-drive/SECONDARY_INDEX_QUERIES.md`:
- [SUGGESTION] packages/rs-drive/SECONDARY_INDEX_QUERIES.md:35: Scope the documented v14 multi-IN allowance to plain document queries
  The documentation currently says several indexed-field IN clauses are allowed from protocol version 14 without limiting that statement to plain document queries. The PR intentionally keeps count, sum, average, and ranked/grouped aggregate surfaces on the single-IN contract. Identify these as non-primary-key IN clauses for plain document queries here and in the restrictions section at lines 241-246, and explicitly state that grouped aggregate queries continue to reject multiple IN clauses.
```
</details>

packages/rs-drive/src/drive/identity/withdrawals/document/find_withdrawal_documents_by_status_and_transaction_indices/v0/mod.rs
```diff
@@ -74,7 +74,7 @@ impl Drive {
             internal_clauses: InternalClauses {
                 primary_key_in_clause: None,
                 primary_key_equal_clause: None,
-                in_clause: None,
+                in_clauses: Vec::new(),
                 range_clause: None,
                 equal_clauses: where_clauses,
             },
```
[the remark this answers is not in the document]
thepastaclaw: The clause does not lose its IN behavior in this direct construction. The v0 lowering selects the final clause from `equal_clauses` and passes it to `WhereClause::to_path_query`, which dispatches on the retained `WhereOperator::In` and inserts all supplied transaction-index keys. This PR only replaced the pre-existing `in_clause: None` initializer with an empty vector.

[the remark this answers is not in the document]
QuantumExplorer: Moved the transaction-index IN clause into `in_clauses` in 6030d33246 (it previously lowered correctly only because `to_path_query` dispatches on the operator; the lowered path query is unchanged).

packages/wasm-drive-verify/src/document/verify_proof.rs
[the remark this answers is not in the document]
QuantumExplorer: Added an `Array::is_array` check before `Array::from` in all three parsers in 6030d33246, so a non-array `in_clauses` value errors instead of silently dropping the IN constraints.

packages/rs-drive/SECONDARY_INDEX_QUERIES.md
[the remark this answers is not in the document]
QuantumExplorer: Reworded both sites in 6030d33246: scoped to non-primary-key IN clauses in plain document queries, with grouped aggregates keeping the multiple-IN rejection.

packages/rs-drive/src/query/conditions.rs
QuantumExplorer: Fixed in 3c5172f560: the grouping is now versioned (`where_clause_grouping` feature version) — constructors take the platform version and dispatch to a v0 grouping restored verbatim (MultipleInClauses precedence ahead of duplicate/overlap/range checks) or the v1 structural grouping; a grammar test pins the v13 precedence across the affected shapes.

packages/rs-drive/src/query/mod.rs
QuantumExplorer: Fixed in 3c5172f560: a versioned shape preflight now runs before the cursor fetch on both construct paths and in verify_start_at_document_in_proof — v0 returns MultipleInClauses and v1 returns Unsupported before any state or proof work; a test pins that a nonexistent startAfter no longer surfaces as StartDocumentNotFound.

packages/rs-drive/src/query/mod.rs
QuantumExplorer: Fixed in 3c5172f560: the v1 multi-IN lowering uses its own left-over recursion that takes each level's direction from orderBy (falling back to the index property), leaving the v0 helper untouched; added a unit test on the lowered query direction and an execution + proof round-trip with a descending left-over level.

packages/rs-drive/src/query/mod.rs
QuantumExplorer: Done in 3c5172f560: the v0/v1 lowerings and their helpers moved to query/non_primary_key_path_query/{v0,v1}/mod.rs and the grouping implementations to query/where_clause_grouping/{v0,v1}/mod.rs, with only dispatch left at the parent boundary.

packages/wasm-drive-verify/src/document/verify_proof.rs
QuantumExplorer: Fixed in 6030d33246: all three parsers require Array::is_array before Array::from and return an invalid-input error for non-array values.

packages/rs-drive/SECONDARY_INDEX_QUERIES.md
QuantumExplorer: Fixed in 6030d33246: both sites now scope the allowance to non-primary-key IN clauses in plain document queries and state that grouped aggregates keep rejecting multiples.

QuantumExplorer:
Re CodeRabbit's outside-diff note (routing the sum and average prove paths through `validate_and_canonicalize_where_clauses`): leaving that out of this PR. Those paths never ran the shared validator, before or after this change — the sum executors independently reject multiple IN clauses (`in_clauses.len() != 1` guards), so this PR does not open the aggregate surfaces to multi-IN — and unifying their validation would change the sum endpoint's accepted shapes and error surface (e.g. `>`/`<` pair canonicalization), which is its own consensus-facing change and deserves its own PR.

packages/rs-drive/src/drive/identity/withdrawals/document/find_withdrawal_documents_by_status_and_transaction_indices/v0/mod.rs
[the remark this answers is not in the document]
QuantumExplorer: Follow-up in 8cd86f360f: rather than editing the live v0 body, v0 is restored to its historical construction and the in_clauses form now lives in a v1 gated at protocol version 14 via DRIVE_IDENTITY_METHOD_VERSIONS_V2; a test pins that both builders lower to the identical path query at v13 and v14.

QuantumExplorer:
Reviewed

Issue #100, opened 2021-02-09
Fedora support


strophy:
mn-bootstrap needs to be tested on non-Ubuntu platforms. Let's test CentOS/Fedora/Redhat first.

shumkov:
Let's do it per user request.

PR #4356, opened 2026-08-10
fix(platform-wallet): fail a double-spending asset lock with a typed terminal error

QuantumExplorer:
## Issue being fixed or feature implemented

A tracked asset lock whose funding input was already spent by a different confirmed transaction can never confirm. Peers reject it as a double spend at the mempool boundary and relay nothing back, and Core has not sent BIP61 `reject` messages by default since 0.17, so the drop is completely silent.

`resume_asset_lock` had no way to see this. It would re-broadcast into the void and then sit in `wait_for_proof` — unbounded for the user-facing funding flows — leaving the condition indistinguishable from a slow network. The app had no basis on which to offer discarding the lock, so the funds it was meant to move stayed stranded with no error surfaced anywhere.

Seen on testnet: a restored wallet built an identity top-up asset lock spending an outpoint that one of its own earlier asset locks had already consumed at height 1510203.

## What was done?

`resume_asset_lock` now screens its `Built` and `Broadcast` arms for a **confirmed** transaction in the wallet's own history that spends one of the lock's inputs, and returns a new terminal `PlatformWalletError::AssetLockInputConflict { out_point, input, spent_by, height }` naming the conflicting input and the transaction that actually spent it.

- The check runs inside the existing read-lock snapshot, so it costs no extra lock acquisition.
- The status match is exhaustive, so settled states (`InstantSendLocked` / `ChainLocked` / `RecoveredFromChain` / `Consumed`) are explicitly excluded and a future status variant forces a decision here.
- Confirmation is required rather than mere presence: an unconfirmed sibling spending the same outpoint is a competing candidate, not a verdict, and is often the transaction the user actually wants to push through.
- FFI result code **41** (`ErrorAssetLockInputConflict`, next free above the highest in-tree claim of 40; the nominally-free 28/30 are left vacated per the ledger convention in that file), with the ledger comment extended and a dedicated arm added to the `From<PlatformWalletError>` mapping so it no longer falls through to `ErrorUnknown`. Mirrored through `PlatformWalletResult.swift` to a typed Swift case so a host can key a discard affordance off the case rather than off message text.

[...]

PR #4392, opened 2026-08-13
perf(swift-sdk): linear wallet-changeset rounds via per-round bulk-prefetch cache

PastaPastaPasta:

Restoring a wallet with a large transaction history made the app pin a CPU core for hours and grow memory without bound until the OS killed it (observed: **59 GB footprint** on a mainnet wallet whose SPV scan matches ~8,000 transactions, with only 3,884 of them ever reaching disk).

The root cause is how a persistence round applies its rows. Each Rust `store()` round maps to one `beginChangeset` → per-kind callbacks → `endChangeset` bracket, with a single `save()` at the end. During SPV catch-up one round can carry thousands of transaction records, and the apply helpers (`upsertTransaction`, `resolveInputOutpoint`, `upsertUtxo`, `markUtxoSpent`, …) issued an individual `ModelContext.fetch` for every row, every input, and every UTXO. SwiftData evaluates each of those fetches against **all objects staged so far in the unsaved round**, so the more rows a round had already staged, the more expensive every following fetch became:

- fetch #1 scans ~0 staged objects, fetch #100,000 scans ~100,000 → total cost grows with the **square** of the round size;
- measured: ~2.3 µs × (staged objects) per fetch — the first 1,000 upserts took 1.3 s, the eighth 1,000 took 20.9 s;
- an 8k-record round with per-input work extrapolates to **hours of pinned CPU**, which is why the persistence drain stalled and the app died before finishing.

## What was done?

One idea, applied consistently: **fetch once per round, not once per row.**

- `PlatformWalletPersistenceHandler.persistWalletChangeset` now builds a `WalletChangesetRoundCache` before applying anything: it walks the changeset once, collects every txid / outpoint / address the round could touch, and bulk-fetches the matching `PersistentTransaction` / `PersistentTxo` / `PersistentPendingInput` / `PersistentCoreAddress` rows with chunked `IN` predicates (≤900 keys per chunk, under SQLite's bind-variable limit).
- All apply helpers (`upsertTransaction`, `resolveInputOutpoint`, `removePendingInputs`, `upsertUtxo`, `markUtxoSpent`, `markUtxoInstantLocked`) look rows up in the cache dictionaries instead of fetching. Inserts and deletes update the cache in place, so later rows in the same batch observe them exactly as they previously observed staged objects through per-row fetches.
- A key the prefetch covered but found no row for is an authoritative miss; the rare key discovered mid-round (e.g. a stale pending row's `spendingTxid` from a prior session) falls back to a single-row fetch.
- `persistAccountAddresses` gets the same treatment — its per-address row fetch and per-address TXO-backfill fetch (a second hot loop in the same rounds during restore) are now two chunked bulk fetches.

Result: a 4,000-record round drops from minutes to **under a second**, and the end-to-end restore that previously died at 59 GB completes a full mainnet genesis→tip sync in ~16 minutes with a ~1.2 GB peak (header download, not persistence; ~430 MB settled).

## How Has This Been Tested?

New unit tests (`swift test`, 354 passing):
- `BulkFetchPredicateTests` — pins the two SwiftData behaviors the cache depends on: `[Data].contains($0.column)` translating to SQL `IN` with >900 keys chunked, and staged (unsaved) rows staying visible to bulk fetches.
- `WalletChangesetRoundTests` — drives real `WalletChangeSetFFI` structs through a full begin→persist→end round: a same-round chain of spends resolves every TXO↔spender linkage and drains all pending-input rows; an input with unknown funding still writes its pending-input row (the out-of-order spend-repair mechanism); and a scaling regression test asserts a 4× larger round costs near-linearly more (fails on any quadratic regression).
- `FFIFixtures` — shared test helpers (deduplicates `tuple32` copies that existed in `DashPayPersistenceTests`).

Manual end-to-end: restored a mainnet wallet reproducing the incident workload (~8k matched transactions) in SwiftExampleApp on the iOS simulator. Full chain scan completed in ~16 minutes; all matched transactions and TXOs durably persisted; sync watermark reached the chain tip; memory sampled every 30 s never exceeded ~1.25 GB; app restart came back clean with the watermark intact.

## Breaking Changes

None. No public API or schema changes; the persistence semantics (round atomicity, pending-input repair, spend gating) are unchanged — only the lookup strategy inside a round.

## Checklist:

## Summary by CodeRabbit

* **Performance**
  * Improved wallet synchronization efficiency by reducing repeated data lookups during changeset processing.
  * Added bulk handling for transactions, outputs, pending inputs, and addresses.

* **Reliability**
  * Improved reconciliation of transaction relationships, pending inputs, spends, instant locks, and wallet addresses.
  * Preserved staged wallet data during inserts and deletions.

* **Tests**
  * Added coverage for bulk lookups, wallet changeset processing, spend linkage, pending inputs, and performance scaling.

### DIFF
--- a/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
@@ -827,13 +827,20 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
                     let tx = txsPtr[t]
                     let txid = hashData(tx.txid)
                     cache.prefetchedTxids.insert(txid)
+                    // Only the input OUTPOINTS are collected here — the
+                    // apply helpers look inputs up as TXOs / pending
+                    // rows, never as transactions, so pulling every
+                    // prevout's parent tx row would only inflate the
+                    // `IN` fetch (hundreds of foreign parents per
+                    // CoinJoin record).
                     if let inPtr = tx.input_outpoints, tx.input_outpoints_count > 0 {
                         for j in 0..<Int(tx.input_outpoints_count) {
                             let entry = inPtr[j]
-                            let prevTxid = hashData(entry.txid)
-                            cache.prefetchedTxids.insert(prevTxid)
                             cache.prefetchedOutpoints.insert(
-                                PersistentTxo.makeOutpoint(txid: prevTxid, vout: entry.vout)
+                                PersistentTxo.makeOutpoint(
+                                    txid: hashData(entry.txid),
+                                    vout: entry.vout
+                                )
                             )
                         }
                     }
@@ -872,34 +879,48 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
             }
         }
 
+        // A failed chunk fetch must NOT leave its keys in the
+        // `prefetched*` sets: membership there declares dictionary
+        // absence authoritative, so an error would silently read as
+        // "these ~900 rows don't exist" and the upsert paths would
+        // insert duplicates over `.unique` columns. Dropping the
+        // chunk's keys instead routes every lookup through the
+        // single-row fallback fetch — the pre-cache behavior.
         for chunk in Self.chunked(Array(cache.prefetchedTxids)) {
             let descriptor = FetchDescriptor<PersistentTransaction>(
                 predicate: #Predicate { chunk.contains($0.txid) }
             )
-            for row in (try? backgroundContext.fetch(descriptor)) ?? [] {
-                cache.transactions[row.txid] = row
+            if let rows = try? backgroundContext.fetch(descriptor) {
+                for row in rows { cache.transactions[row.txid] = row }
+            } else {
+                cache.prefetchedTxids.subtract(chunk)
             }
         }
         for chunk in Self.chunked(Array(cache.prefetchedOutpoints)) {
             let txoDescriptor = FetchDescriptor<PersistentTxo>(
                 predicate: #Predicate { chunk.contains($0.outpoint) }
             )
-            for row in (try? backgroundContext.fetch(txoDescriptor)) ?? [] {
-                cache.txos[row.outpoint] = row
-            }
             let pendingDescriptor = FetchDescriptor<PersistentPendingInput>(
                 predicate: #Predicate { chunk.contains($0.outpoint) }
             )
-            for row in (try? backgroundContext.fetch(pendingDescriptor)) ?? [] {
-                cache.pendingInputs[row.outpoint, default: []].append(row)
+            if let txoRows = try? backgroundContext.fetch(txoDescriptor),
+               let pendingRows = try? backgroundContext.fetch(pendingDescriptor) {
+                for row in txoRows { cache.txos[row.outpoint] = row }
+                for row in pendingRows {
+                    cache.pendingInputs[row.outpoint, default: []].append(row)
+                }
+            } else {
+                cache.prefetchedOutpoints.subtract(chunk)
             }
         }
         for chunk in Self.chunked(Array(cache.prefetchedAddresses)) {
             let descriptor = FetchDescriptor<PersistentCoreAddress>(
                 predicate: #Predicate { chunk.contains($0.address) }
             )
-            for row in (try? backgroundContext.fetch(descriptor)) ?? [] {
-                cache.coreAddresses[row.address] = row
+            if let rows = try? backgroundContext.fetch(descriptor) {
+                for row in rows { cache.coreAddresses[row.address] = row }
+            } else {
+                cache.prefetchedAddresses.subtract(chunk)
             }
         }
 
@@ -3394,12 +3415,20 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
         let allAddresses = entries.map(\.address)
         var existingRows: [String: PersistentCoreAddress] = [:]
         var txosByAddress: [String: [PersistentTxo]] = [:]
+        // Addresses whose bulk row fetch FAILED (threw) — a miss for
+        // these is not authoritative, so the upsert loop falls back to
+        // a single-row fetch instead of inserting over the `.unique`
+        // address column. A failed TXO-backfill fetch just skips the
+        // backfill for the chunk, matching the old per-row `try?`.
+        var unresolvedAddresses: Set<String> = []
         for chunk in Self.chunked(allAddresses) {
             let rowDescriptor = FetchDescriptor<PersistentCoreAddress>(
                 predicate: #Predicate { chunk.contains($0.address) }
             )
-            for row in (try? backgroundContext.fetch(rowDescriptor)) ?? [] {
-                existingRows[row.address] = row
+            if let rows = try? backgroundContext.fetch(rowDescriptor) {
+                for row in rows { existingRows[row.address] = row }
+            } else {
+                unresolvedAddresses.formUnion(chunk)
             }
             let txoDescriptor = FetchDescriptor<PersistentTxo>(
                 predicate: #Predicate { chunk.contains($0.address) }
@@ -3411,6 +3440,14 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
 
         for entry in entries {
             let address = entry.address
+            if existingRows[address] == nil, unresolvedAddresses.contains(address) {
+                let fallbackDescriptor = FetchDescriptor<PersistentCoreAddress>(
+                    predicate: #Predicate { $0.address == address }
+                )
+                if let row = try? backgroundContext.fetch(fallbackDescriptor).first {
+                    existingRows[address] = row
+                }
+            }
             let row: PersistentCoreAddress
             if let existing = existingRows[address] {
                 row = existing
--- a/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
+++ b/packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
@@ -766,6 +766,222 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
 
     // MARK: - Wallet Changeset (transactions, utxos, accounts, balance, chain)
 
+    /// Per-round lookup cache for the wallet-changeset apply path.
+    ///
+    /// A single changeset can carry thousands of transaction records
+    /// (an SPV catch-up folds many blocks into one `store()` round),
+    /// and the apply helpers used to issue an individual
+    /// `ModelContext.fetch` per row, per input, and per UTXO. Each of
+    /// those fetches re-evaluates its predicate against every object
+    /// staged (unsaved) in the open begin/end changeset bracket, so
+    /// the round's cost grew quadratically with its size — hours of
+    /// CPU for an 8k-record round on a large wallet.
+    ///
+    /// Instead, `buildWalletChangesetRoundCache` walks the changeset
+    /// once, bulk-fetches every row the round could touch with
+    /// chunked `IN` predicates, and the helpers hit these
+    /// dictionaries. Inserts and deletes performed during the round
+    /// update the cache in place so later rows observe them, exactly
+    /// as they observed staged objects through per-row fetches.
+    ///
+    /// A key found in a dictionary is a hit. A key absent from the
+    /// dictionary but present in the corresponding `prefetched*` set
+    /// is an authoritative miss (the bulk fetch covered it). A key in
+    /// neither (rare: values discovered mid-round, e.g. a pending
+    /// row's `spendingTxid` loaded from the store) falls back to a
+    /// single-row fetch.
+    private final class WalletChangesetRoundCache {
+        /// txid → transaction row (records, stubs, spending txs).
+        var transactions: [Data: PersistentTransaction] = [:]
+        /// 36-byte outpoint → TXO row.
+        var txos: [Data: PersistentTxo] = [:]
+        /// 36-byte outpoint → unresolved pending-input rows. A key
+        /// present with an empty array is authoritative: the rows
+        /// were deleted this round (or a fallback fetch found none).
+        var pendingInputs: [Data: [PersistentPendingInput]] = [:]
+        /// Base58Check address → core-address row.
+        var coreAddresses: [String: PersistentCoreAddress] = [:]
+
+        /// Keys covered by the bulk prefetch — absence from the
+        /// dictionaries above is authoritative for these.
+        var prefetchedTxids: Set<Data> = []
+        var prefetchedOutpoints: Set<Data> = []
+        var prefetchedAddresses: Set<String> = []
+    }
+
+    /// Walk the changeset's account buckets, collect every txid /
+    /// outpoint / address the apply helpers could look up, and
+    /// bulk-fetch the matching rows in chunks (staying under SQLite's
+    /// bind-variable limit). One fetch per entity per ~900 keys
+    /// replaces one fetch per row.
+    private func buildWalletChangesetRoundCache(
+        accountsPtr: UnsafePointer<AccountChangeSetFFI>,
+        count: Int
+    ) -> WalletChangesetRoundCache {
+        let cache = WalletChangesetRoundCache()
+
+        for i in 0..<count {
+            let acc = accountsPtr[i]
+            if acc.transactions_count > 0, let txsPtr = acc.transactions {
+                for t in 0..<Int(acc.transactions_count) {
+                    let tx = txsPtr[t]
+                    let txid = hashData(tx.txid)
+                    cache.prefetchedTxids.insert(txid)
+                    if let inPtr = tx.input_outpoints, tx.input_outpoints_count > 0 {
+                        for j in 0..<Int(tx.input_outpoints_count) {
+                            let entry = inPtr[j]
+                            let prevTxid = hashData(entry.txid)
+                            cache.prefetchedTxids.insert(prevTxid)
+                            cache.prefetchedOutpoints.insert(
+                                PersistentTxo.makeOutpoint(txid: prevTxid, vout: entry.vout)
+                            )
+                        }
+                    }
+                }
+            }
+            if acc.utxos_added_count > 0, let utxosPtr = acc.utxos_added {
+                for u in 0..<Int(acc.utxos_added_count) {
+                    let utxo = utxosPtr[u]
+                    let txid = hashData(utxo.outpoint.txid)
+                    cache.prefetchedTxids.insert(txid)
+                    cache.prefetchedOutpoints.insert(
+                        PersistentTxo.makeOutpoint(txid: txid, vout: utxo.outpoint.vout)
+                    )
+                    if let addrPtr = utxo.address {
+                        cache.prefetchedAddresses.insert(String(cString: addrPtr))
+                    }
+                }
+            }
+            if acc.utxos_spent_count > 0, let spentPtr = acc.utxos_spent {
+                for s in 0..<Int(acc.utxos_spent_count) {
+                    let entry = spentPtr[s]
+                    let txid = hashData(entry.outpoint.txid)
+                    cache.prefetchedOutpoints.insert(
+                        PersistentTxo.makeOutpoint(txid: txid, vout: entry.outpoint.vout)
+                    )
+                    cache.prefetchedTxids.insert(hashData(entry.spending_txid))
+                }
+            }
+            if acc.utxos_instant_locked_count > 0, let ilPtr = acc.utxos_instant_locked {
+                for l in 0..<Int(acc.utxos_instant_locked_count) {
+                    let op = ilPtr[l]
+                    cache.prefetchedOutpoints.insert(
+                        PersistentTxo.makeOutpoint(txid: hashData(op.txid), vout: op.vout)
+                    )
+                }
+            }
+        }
+
+        for chunk in Self.chunked(Array(cache.prefetchedTxids)) {
+            let descriptor = FetchDescriptor<PersistentTransaction>(
+                predicate: #Predicate { chunk.contains($0.txid) }
+            )
+            for row in (try? backgroundContext.fetch(descriptor)) ?? [] {
+                cache.transactions[row.txid] = row
+            }
+        }
+        for chunk in Self.chunked(Array(cache.prefetchedOutpoints)) {
+            let txoDescriptor = FetchDescriptor<PersistentTxo>(
+                predicate: #Predicate { chunk.contains($0.outpoint) }
+            )
+            for row in (try? backgroundContext.fetch(txoDescriptor)) ?? [] {
+                cache.txos[row.outpoint] = row
+            }
+            let pendingDescriptor = FetchDescriptor<PersistentPendingInput>(
+                predicate: #Predicate { chunk.contains($0.outpoint) }
+            )
+            for row in (try? backgroundContext.fetch(pendingDescriptor)) ?? [] {
+                cache.pendingInputs[row.outpoint, default: []].append(row)
+            }
+        }
+        for chunk in Self.chunked(Array(cache.prefetchedAddresses)) {
+            let descriptor = FetchDescriptor<PersistentCoreAddress>(
+                predicate: #Predicate { chunk.contains($0.address) }
+            )
+            for row in (try? backgroundContext.fetch(descriptor)) ?? [] {
+                cache.coreAddresses[row.address] = row
+            }
+        }
+
+        return cache
+    }
+
+    /// Split `keys` into slices below SQLite's historical 999
+    /// bind-variable limit so each `IN` predicate stays translatable.
+    private static func chunked<T>(_ keys: [T], size: Int = 900) -> [[T]] {
+        stride(from: 0, to: keys.count, by: size).map {
+            Array(keys[$0..<min($0 + size, keys.count)])
+        }
+    }
+
+    /// Cache-first transaction lookup. Falls back to a single-row
+    /// fetch only for keys outside the prefetched set.
+    private func cachedTransaction(
+        txid: Data,
+        cache: WalletChangesetRoundCache
+    ) -> PersistentTransaction? {
+        if let hit = cache.transactions[txid] { return hit }
+        if cache.prefetchedTxids.contains(txid) { return nil }
+        let descriptor = FetchDescriptor<PersistentTransaction>(
+            predicate: #Predicate { $0.txid == txid }
+        )
+        guard let row = try? backgroundContext.fetch(descriptor).first else { return nil }
+        cache.transactions[txid] = row
+        return row
+    }
+
+    /// Cache-first TXO lookup, same fallback contract as
+    /// `cachedTransaction`.
+    private func cachedTxo(
+        outpoint: Data,
+        cache: WalletChangesetRoundCache
+    ) -> PersistentTxo? {
+        if let hit = cache.txos[outpoint] { return hit }
+        if cache.prefetchedOutpoints.contains(outpoint) { return nil }
+        let descriptor = FetchDescriptor<PersistentTxo>(
+            predicate: #Predicate { $0.outpoint == outpoint }
+        )
+        guard let row = try? backgroundContext.fetch(descriptor).first else { return nil }
+        cache.txos[outpoint] = row
+        return row
+    }
+
+    /// Cache-first core-address lookup, same fallback contract as
+    /// `cachedTransaction`.
+    private func cachedCoreAddress(
+        address: String,
+        cache: WalletChangesetRoundCache
+    ) -> PersistentCoreAddress? {
+        if let hit = cache.coreAddresses[address] { return hit }
+        if cache.prefetchedAddresses.contains(address) { return nil }
+        let descriptor = FetchDescriptor<PersistentCoreAddress>(
+            predicate: #Predicate { $0.address == address }
+        )
+        guard let row = try? backgroundContext.fetch(descriptor).first else { return nil }
+        cache.coreAddresses[address] = row
+        return row
+    }
+
+    /// Cache-first pending-input lookup. Always leaves an entry for
+    /// `outpoint` in the dictionary afterwards, so the result is
+    /// authoritative on subsequent hits (including "no rows").
+    private func cachedPendingInputs(
+        outpoint: Data,
+        cache: WalletChangesetRoundCache
+    ) -> [PersistentPendingInput] {
+        if let rows = cache.pendingInputs[outpoint] { return rows }
+        if cache.prefetchedOutpoints.contains(outpoint) {
+            cache.pendingInputs[outpoint] = []
+            return []
+        }
+        let descriptor = FetchDescriptor<PersistentPendingInput>(
+            predicate: #Predicate { $0.outpoint == outpoint }
+        )
+        let rows = (try? backgroundContext.fetch(descriptor)) ?? []
+        cache.pendingInputs[outpoint] = rows
+        return rows
+    }
+
     /// Apply a full `WalletChangeSetFFI` to SwiftData.
     ///
     /// Called from the Rust persister when an SPV round produces core-
@@ -813,11 +1029,17 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
                 wallet.lastUpdated = Date()
             }
 
-            // Per-account: transactions, UTXOs, pool state.
+            // Per-account: transactions, UTXOs, pool state. All row
+            // lookups go through a per-round bulk-prefetched cache —
+            // see `WalletChangesetRoundCache`.
             if cs.accounts_count > 0, let accountsPtr = cs.accounts {
+                let cache = buildWalletChangesetRoundCache(
+                    accountsPtr: accountsPtr,
+                    count: Int(cs.accounts_count)
+                )
                 for i in 0..<Int(cs.accounts_count) {
                     let acc = accountsPtr[i]
-                    applyAccountChangeset(walletRecord: wallet, acc: acc)
+                    applyAccountChangeset(walletRecord: wallet, acc: acc, cache: cache)
                 }
             }
 
@@ -883,7 +1105,8 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
     /// Apply a single account changeset to SwiftData.
     private func applyAccountChangeset(
         walletRecord: PersistentWallet,
-        acc: AccountChangeSetFFI
+        acc: AccountChangeSetFFI,
+        cache: WalletChangesetRoundCache
     ) {
         let accountIndex = acc.account_index
         // Stable account-type discriminants from the FFI. Used as the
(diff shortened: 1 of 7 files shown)


thepastaclaw:
✅ Final review complete — no blockers (commit 86c9033d) · triage: critical · Phase 2 only (queue backlog)

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -766,6 +766,243 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
...
+        if let rows = cache.pendingInputs[outpoint] { return rows }
+        if cache.prefetchedOutpoints.contains(outpoint) {
+            cache.pendingInputs[outpoint] = []
+            return []
+        }
+        let descriptor = FetchDescriptor<PersistentPendingInput>(
+            predicate: #Predicate { $0.outpoint == outpoint }
+        )
+        let rows = (try? backgroundContext.fetch(descriptor)) ?? []
+        cache.pendingInputs[outpoint] = rows
```
thepastaclaw: **🟡 Suggestion: Do not make a failed pending-input fetch authoritative**

The bulk-prefetch path removes a failed chunk from `prefetchedOutpoints` so later accesses can fall back to individual fetches, but this fallback converts its own fetch failure into `[]` and caches that value. Subsequent operations therefore treat the outpoint as having no pending rows: `upsertUtxo` can skip deferred-spend reconciliation, while `removePendingInputs` can leave persisted rows behind. Preserve a distinct fetch-failure state and only cache successfully fetched rows. The callers also need to avoid replacing that failure state with an authoritative empty or partial entry: `resolveInputOutpoint` may insert a new staged row without claiming it is the complete set, and `removePendingInputs` should only cache `[]` after a successful lookup and deletion.

thepastaclaw (commented):

The per-round cache preserves the intended reconciliation behavior and removes the main quadratic lookup path, but two minor issues remain: failed pending-input fallback fetches are cached as authoritative misses, and a new test performs an alignment-dependent typed load from Data. Neither issue is blocking, but both should be corrected before relying on the fallback and regression coverage.
Source: reviewer backend model gpt-5.6-sol; final verifier backend model gpt-5.6-sol. openclaw-agent/cliproxy/gpt-5.6-sol was orchestration-only and is not reviewer evidence.

### Review provenance

🟡 2 suggestion(s)

<details>

```

- [SUGGESTION] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift:1001-1002: Do not make a failed pending-input fetch authoritative
  The bulk-prefetch path removes a failed chunk from `prefetchedOutpoints` so later accesses can fall back to individual fetches, but this fallback converts its own fetch failure into `[]` and caches that value. Subsequent operations therefore treat the outpoint as having no pending rows: `upsertUtxo` can skip deferred-spend reconciliation, while `removePendingInputs` can leave persisted rows behind. Preserve a distinct fetch-failure state and only cache successfully fetched rows. The callers also need to avoid replacing that failure state with an authoritative empty or partial entry: `resolveInputOutpoint` may insert a new staged row without claiming it is the complete set, and `removePendingInputs` should only cache `[]` after a successful lookup and deletion.

In `packages/swift-sdk/SwiftTests/SwiftDashSDKTests/WalletChangesetRoundTests.swift`:
- [SUGGESTION] packages/swift-sdk/SwiftTests/SwiftDashSDKTests/WalletChangesetRoundTests.swift:166: Use an unaligned load when decoding the Data-backed txid
  `UnsafeRawBufferPointer.load(as:)` requires the buffer address to satisfy `UInt64` alignment, which `Data.withUnsafeBytes` does not guarantee. The test can therefore trap before checking the spend linkage. The package uses Swift tools 6.0, where `loadUnaligned(as:)` is available, so decode these bytes without imposing an alignment precondition.
```
</details>

packages/swift-sdk/SwiftTests/SwiftDashSDKTests/WalletChangesetRoundTests.swift
```diff
@@ -0,0 +1,237 @@
...
+        }
+        runRound(handler: handler, txs: txs)
+
+        let transactions = try fetchAll(PersistentTransaction.self, in: container)
+        XCTAssertEqual(transactions.count, count)
+
+        let txos = try fetchAll(PersistentTxo.self, in: container)
+        XCTAssertEqual(txos.count, count)
+        for txo in txos {
+            let fundingIndex = txo.outpoint.withUnsafeBytes { $0.load(as: UInt64.self) }
```
thepastaclaw: **🟡 Suggestion: Use an unaligned load when decoding the Data-backed txid**

`UnsafeRawBufferPointer.load(as:)` requires the buffer address to satisfy `UInt64` alignment, which `Data.withUnsafeBytes` does not guarantee. The test can therefore trap before checking the spend linkage. The package uses Swift tools 6.0, where `loadUnaligned(as:)` is available, so decode these bytes without imposing an alignment precondition.

```suggestion
            let fundingIndex = txo.outpoint.withUnsafeBytes {
                $0.loadUnaligned(as: UInt64.self)
            }
```

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -766,6 +766,243 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
...
+        return rows
+    }
```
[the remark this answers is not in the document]
thepastaclaw: The failure is valid, but changing only `cachedPendingInputs` is incomplete. `resolveInputOutpoint` can immediately cache only its newly inserted row, and `removePendingInputs` unconditionally replaces the failed lookup with `[]`; both callers must preserve the non-authoritative failure state so a later lookup can retry and discover persisted rows.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
PastaPastaPasta: Fixed in 9219275f64: a thrown fallback fetch now leaves the cache unpopulated (reads retry), the insert path skips seeding an entry for a failed-fetch outpoint so a staged row never reads as the complete set, and removePendingInputs only writes the authoritative empty after a successful lookup.

---
🤖 Posted autonomously by Claude on behalf of pasta.

packages/swift-sdk/SwiftTests/SwiftDashSDKTests/WalletChangesetRoundTests.swift
PastaPastaPasta: Fixed in 9219275f64 — switched to loadUnaligned(as:).

---
🤖 Posted autonomously by Claude on behalf of pasta.

PastaPastaPasta (commented):
Automated deep review of this PR (8 independent finder angles, each candidate adversarially re-verified against the PR head before posting). 9 findings survived verification and are posted inline below; the remaining candidates were dropped as duplicates of existing review threads or as unconfirmed on inspection (notably a claimed prefetch-coverage gap on `upsertUtxo`'s pending-resolve path — the `chosen.spendingTransaction` relationship preference makes that fallback fetch rare in practice).

---
🤖 Posted autonomously by Claude on behalf of pasta.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -766,6 +766,243 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
...
+    private func cachedTransaction(
+        txid: Data,
+        cache: WalletChangesetRoundCache
+    ) -> PersistentTransaction? {
+        if let hit = cache.transactions[txid] { return hit }
+        if cache.prefetchedTxids.contains(txid) { return nil }
+        let descriptor = FetchDescriptor<PersistentTransaction>(
+            predicate: #Predicate { $0.txid == txid }
+        )
+        guard let row = try? backgroundContext.fetch(descriptor).first else { return nil }
```
PastaPastaPasta: **🟡 Thrown fallback fetches in `cachedTransaction`/`cachedTxo`/`cachedCoreAddress` read as "row absent" and license inserts over `.unique` columns**

All three single-row fallbacks use `guard let row = try? backgroundContext.fetch(descriptor).first else { return nil }`, which collapses a thrown fetch into the same `nil` as a genuinely missing row. Callers take `nil` as license to insert (`upsertTransaction`'s new-record branch, `upsertUtxo`'s TXO and stub-tx inserts), and `PersistentTransaction.txid`, `PersistentTxo.outpoint`, and `PersistentCoreAddress.address` are all `@Attribute(.unique)` — so a transient fetch error becomes a duplicate-key insert whose failure only surfaces at `endChangeset`'s `save()`, rolling back the entire round and reporting a persistence failure to Rust; if the underlying error recurs, the replayed round wedges the same way. The bulk-path comment above (lines 882-888) names exactly this hazard and routes failures to this fallback, but the fallback has the same collapse, and the resolution being applied to `cachedPendingInputs` in the existing thread doesn't extend here. Consider distinguishing a thrown fetch from an empty result in these three as well (fail the round early, or track the key as unresolved rather than absent).

---
🤖 Posted autonomously by Claude on behalf of pasta.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -766,6 +766,243 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
...
+            }
+        }
+        for chunk in Self.chunked(Array(cache.prefetchedOutpoints)) {
+            let txoDescriptor = FetchDescriptor<PersistentTxo>(
+                predicate: #Predicate { chunk.contains($0.outpoint) }
+            )
+            let pendingDescriptor = FetchDescriptor<PersistentPendingInput>(
+                predicate: #Predicate { chunk.contains($0.outpoint) }
+            )
+            if let txoRows = try? backgroundContext.fetch(txoDescriptor),
```
PastaPastaPasta: **💬 A thrown pending-input fetch discards the successful TXO chunk results**

The joint `if let txoRows = try? ..., let pendingRows = try? ...` couples the two fetches: when the pending fetch throws, the already-fetched `txoRows` are thrown away and the whole chunk is subtracted from `prefetchedOutpoints`, demoting ~900 keys of *both* entities to per-row fallback fetches for the rest of the round. Since one `prefetchedOutpoints` set serves both caches, confining the damage means splitting coverage per entity — e.g. separate prefetched-outpoint sets for TXOs and pending inputs, so a transient error in one fetch doesn't reinstate per-row fetching for the other.

---
🤖 Posted autonomously by Claude on behalf of pasta.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -1314,14 +1557,9 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
...
-        if record.coreAddress == nil, !record.address.isEmpty {
-            let addressLookup = record.address
-            let coreAddressDescriptor = FetchDescriptor<PersistentCoreAddress>(
-                predicate: #Predicate { $0.address == addressLookup }
-            )
-            if let coreAddr = try? backgroundContext.fetch(coreAddressDescriptor).first {
-                record.coreAddress = coreAddr
-            }
+        if record.coreAddress == nil, !record.address.isEmpty,
+           let coreAddr = cachedCoreAddress(address: record.address, cache: cache) {
```
PastaPastaPasta, line 1925: **💬 Existing-TXO core-address lookups can miss the prefetch and fall back per row**

`prefetchedAddresses` is collected from `utxo.address` (FFI value, only when non-nil), but this lookup keys on `record.address` — the stored row value. When a known outpoint is re-emitted without an FFI address (or with a different one) and `record.coreAddress` is still nil, the stored address misses the prefetched set and takes a single-row fallback fetch; a negative result is never memoized, so every such TXO in the round re-fetches. Since the TXO bulk fetch runs before the core-address chunk loop in `buildWalletChangesetRoundCache`, unioning the fetched TXO rows' `address` values into `prefetchedAddresses` there (or memoizing negative fallback results) would close the gap.

---
🤖 Posted autonomously by Claude on behalf of pasta.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -766,6 +766,243 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
...
+                }
+            }
+            if acc.utxos_spent_count > 0, let spentPtr = acc.utxos_spent {
+                for s in 0..<Int(acc.utxos_spent_count) {
+                    let entry = spentPtr[s]
+                    let txid = hashData(entry.outpoint.txid)
+                    cache.prefetchedOutpoints.insert(
+                        PersistentTxo.makeOutpoint(txid: txid, vout: entry.outpoint.vout)
+                    )
+                    cache.prefetchedTxids.insert(hashData(entry.spending_txid))
```
PastaPastaPasta, line 1118: **💬 All-zero `spending_txid` sentinel is seeded into `prefetchedTxids`**

The consumer (`markUtxoSpent`) guards the lookup with `!spendingTxid.allSatisfy { $0 == 0 }`, but this insert is unconditional, so the "no spending tx" sentinel lands in the bulk `IN` fetch and the authoritative-miss set. Harmless today because the only guarded consumer never looks it up, but mirroring the zero-check here keeps the key set meaningful and the fetch lists minimal.

---
🤖 Posted autonomously by Claude on behalf of pasta.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -3187,14 +3407,49 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
             return true
         }
 
+        // Bulk-prefetch the address rows and the TXO-backfill rows in
+        // chunked `IN` fetches instead of two per-entry fetches — a
+        // restore emits thousands of entries per round, and each
+        // per-row fetch would re-scan the round's staged objects
+        // (same quadratic the wallet-changeset round cache removes).
+        let allAddresses = entries.map(\.address)
```
PastaPastaPasta, line 3933: **🟡 `persistPlatformPaymentAddresses` still does one fetch per entry — the same n² this block removes**

This bulk prefetch fixes the base58 branch, but the platform branch this function delegates to a few lines up (`persistPlatformPaymentAddresses`, ~line 3527) still runs a `FetchDescriptor<PersistentPlatformAddress>` per entry against a context full of staged rows. The justification in this comment applies verbatim: a DIP-17 platform-account restore emits thousands of entries per round, and each per-entry fetch re-scans the round's staged objects — keeping exactly the quadratic this PR removes for core addresses. Applying the same chunked-`IN` prefetch keyed on `PersistentPlatformAddress.address` there would finish the job.

---
🤖 Posted autonomously by Claude on behalf of pasta.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -3187,14 +3407,49 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
...
+            )
+            if let rows = try? backgroundContext.fetch(rowDescriptor) {
+                for row in rows { existingRows[row.address] = row }
+            } else {
+                unresolvedAddresses.formUnion(chunk)
+            }
+            let txoDescriptor = FetchDescriptor<PersistentTxo>(
+                predicate: #Predicate { chunk.contains($0.address) }
+            )
+            for txo in (try? backgroundContext.fetch(txoDescriptor)) ?? [] {
```
PastaPastaPasta, line 3954: **💬 Failed TXO-backfill chunk now silently skips ~900 addresses; `allAddresses` keeps duplicates**

The comment says a failed TXO-backfill fetch "match[es] the old per-row `try?`", but the granularity changed: the old code lost one address's backfill per thrown fetch, while a thrown chunk here drops it for up to 900 addresses, with no `unresolvedAddresses`-style fallback like the address-row fetch immediately above. Separately, `entries.map(\.address)` keeps duplicate addresses in the chunk `IN` lists (a `Set` would be minimal by construction). Both are minor since the backfill is a display-relationship sweep, but a symmetric per-address fallback set would restore parity with the row fetch beside it.

---
🤖 Posted autonomously by Claude on behalf of pasta.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -766,6 +766,243 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
...
+        }
+
+        // A failed chunk fetch must NOT leave its keys in the
+        // `prefetched*` sets: membership there declares dictionary
+        // absence authoritative, so an error would silently read as
+        // "these ~900 rows don't exist" and the upsert paths would
+        // insert duplicates over `.unique` columns. Dropping the
+        // chunk's keys instead routes every lookup through the
+        // single-row fallback fetch — the pre-cache behavior.
+        for chunk in Self.chunked(Array(cache.prefetchedTxids)) {
```
PastaPastaPasta, line 1139: **🟡 Round cache pins every key and fetched row for the whole round — memory peak lands on exactly the rounds this PR targets**

Each prefetched outpoint is a 36-byte `Data`, and the builder collects every input outpoint (CoinJoin records carry hundreds of foreign parents each), so a large round can hold 10^5–10^6 keys across the `prefetched*` sets plus a reference to every fetched/inserted row in the dictionaries until the round ends; `Self.chunked(Array(...))` then materializes the full key set again as arrays. Staged inserts are pinned by the `ModelContext` regardless, so the delta is the key sets, the fetched-row maps, and the array copies — concentrated on the huge rounds the PR optimizes, which is where iOS memory pressure (jetsam) bites. Draining the cache per account/sub-batch and chunking over `ArraySlice` instead of copied arrays would flatten the peak without giving up the linear fetch count.

---
🤖 Posted autonomously by Claude on behalf of pasta.

packages/swift-sdk/SwiftTests/SwiftDashSDKTests/WalletChangesetRoundTests.swift
```diff
@@ -0,0 +1,237 @@
...
+            return Date().timeIntervalSince(start)
+        }
+
+        // Warm-up so one-time SwiftData/SQLite setup cost doesn't
+        // pollute the small-round baseline.
+        _ = try measureRound(count: 50)
+
+        let small = try measureRound(count: 1_000)
+        let large = try measureRound(count: 4_000)
+        XCTAssertLessThan(
```
PastaPastaPasta: **🟡 Wall-clock ratio assertion is simultaneously loose and flaky — consider counting fetches instead**

`large < max(small, 0.05) * 10` passes a 2–3× superlinear regression outright, while on a loaded CI host `large` can spike past 10× of a fast `small` run and flake — and the natural response to flakes is loosening the threshold, eroding the only regression guard this PR adds. A deterministic proxy would serve better: a fetch-call counter on the handler (increment around each `backgroundContext.fetch`) asserting the round executes O(chunks + fallback misses) fetches rather than O(rows). That fails on any reintroduced per-row fetch, runs in milliseconds, and is immune to host load.

---
🤖 Posted autonomously by Claude on behalf of pasta.

packages/swift-sdk/SwiftTests/SwiftDashSDKTests/BulkFetchPredicateTests.swift
```diff
@@ -0,0 +1,84 @@
...
+            let tx = PersistentTransaction(txid: makeTxid(i), transactionData: Data())
+            context.insert(tx)
+            let txo = PersistentTxo(transaction: tx, vout: 0, amount: UInt64(i), address: "addr\(i)")
+            context.insert(txo)
+            outpoints.append(txo.outpoint)
+        }
+        try context.save()
+
+        var fetched: [Data: PersistentTxo] = [:]
+        for chunk in stride(from: 0, to: outpoints.count, by: 900).map({
```
PastaPastaPasta, line 43: **💬 Test re-implements the chunking instead of exercising `chunked(_:size:)`**

Because `PlatformWalletPersistenceHandler.chunked` is `private`, this test hand-rolls the same stride/slice logic inline — so it validates a copy of the algorithm, not the shipped helper, and the two can drift (say, a future chunk-size or slicing change) without this pin noticing. Widening `chunked` to `internal` (the suite already imports `@testable`) and calling it here would make the contract test bind to the real code; this PR's new FFIFixtures.swift shows the pattern of promoting shared test plumbing when a second user appears.

---
🤖 Posted autonomously by Claude on behalf of pasta.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
PastaPastaPasta: Fixed in f5f4a4b75c. The round's reads now go through the existing `ModelFetching` seam: a thrown row lookup in `cachedTransaction` / `cachedTxo` / `cachedCoreAddress` records `cache.fetchFailure`, the apply loops stop, `persistWalletChangeset` returns `false`, and the C shim returns a non-zero code so Rust closes the round as failed and `endChangeset` rolls the staged writes back. Review of that fix found two worse instances of the same collapse in the same call chain, and they are covered too: the `PersistentAccount` lookup in `applyAccountChangeset` (no unique column, so the duplicate would have committed) and the `PersistentWallet` lookup in `persistWalletChangeset` (a thrown read reported the round as a success while dropping it). Pending inputs keep the retry treatment from 5d0a0f1a14, since a duplicate pending row resolves to the same TXO. Regression test: `testThrownFallbackFetchRejectsTheRound`.

---
🤖 Posted autonomously by Claude on behalf of pasta.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
PastaPastaPasta: Fixed in f5f4a4b75c. TXO and pending-input prefetch coverage are now tracked in separate sets (`prefetchedTxoOutpoints` / `prefetchedPendingOutpoints`) and the two chunk fetches run independently, so a thrown pending-input chunk fetch keeps the already-fetched TXO rows and only demotes pending-input lookups for that chunk to the per-row fallback (and vice versa).

---
🤖 Posted autonomously by Claude on behalf of pasta.

packages/swift-sdk/SwiftTests/SwiftDashSDKTests/WalletChangesetRoundTests.swift
PastaPastaPasta: Fixed in 86c9033d6b. The wall-clock ratio is replaced by `testRoundFetchCountIsIndependentOfRecordCount`, which counts reads through the `ModelFetching` seam (f5f4a4b75c routed the round's reads through it) and asserts the exact count `2 + 4 * ceil(records / 900)` for 100 and 2,000 records: the wallet and account lookups plus one bulk fetch per entity per chunk. Any reintroduced per-row fetch scales that with the record count.

---
🤖 Posted autonomously by Claude on behalf of pasta.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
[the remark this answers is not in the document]
PastaPastaPasta: Fixed in 5d0a0f1a14 (9219275f64 before the rebase), covering all three paths: a thrown fallback fetch leaves `cache.pendingInputs[outpoint]` absent and records the outpoint in `pendingFetchFailed`; `resolveInputOutpoint` only appends its newly staged row when the lookup was authoritative; `removePendingInputs` only writes the authoritative `[]` after a successful lookup. f5f4a4b75c keeps this retry treatment for pending inputs (no unique column) while the unique-keyed lookups now reject the round instead.

---
🤖 Posted autonomously by Claude on behalf of pasta.

packages/swift-sdk/SwiftTests/SwiftDashSDKTests/WalletChangesetRoundTests.swift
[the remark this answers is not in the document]
PastaPastaPasta: Fixed in 5d0a0f1a14 (9219275f64 before the rebase): the test decodes the `Data`-backed outpoint with `loadUnaligned(as:)`.

---
🤖 Posted autonomously by Claude on behalf of pasta.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
PastaPastaPasta: After the rebase that fix is 5d0a0f1a14; f5f4a4b75c extends the same non-authoritative treatment to the rest of the round (unique-keyed lookups reject the round rather than retry).

---
🤖 Posted autonomously by Claude on behalf of pasta.

packages/swift-sdk/SwiftTests/SwiftDashSDKTests/WalletChangesetRoundTests.swift
PastaPastaPasta: After the rebase that fix is 5d0a0f1a14.

---
🤖 Posted autonomously by Claude on behalf of pasta.

PastaPastaPasta:
Rebased onto `v4.2-dev` (`49931091af`) and pushed `86c9033d6b`. Commits:

- `593f48cabb` / `cd00e679f9` / `5d0a0f1a14` — the three existing commits, replayed. One conflicting file (`PlatformWalletPersistenceHandler.swift`, first commit only); range-diff shows `!` for that commit (resolved hunks only) and `=` for the other two. Resolutions all keep the base's intent and thread the `cache:` argument through: the per-row `autoreleasepool` from #4414 wraps the cached upserts; `resolveInputOutpoint`, the `upsertUtxo` pending-input pass and `markUtxoSpent` keep the `reconcileSpendObservation` finality rule from #4336 / #4356 (the PR's older "newest wins" / `spendIsInBlock` block and the unconditional `spendingTransaction` write are superseded by it) with only the fetches swapped for cache lookups.
- `f5f4a4b75c` **fix: reject the changeset round when a round read throws** — addresses the line-949 and line-906 threads. Round reads go through the `ModelFetching` seam; a thrown row lookup (`cachedTransaction` / `cachedTxo` / `cachedCoreAddress`, plus the account lookup in `applyAccountChangeset` and the wallet lookup in `persistWalletChangeset`, which had the same collapse with no unique backstop) sets `cache.fetchFailure`, the apply loops stop, and the changeset callback returns non-zero so Rust closes the round as failed and `endChangeset` rolls back. Pending inputs keep the retry semantics from `5d0a0f1a14`. TXO and pending-input prefetch coverage are tracked separately so a failed chunk of one no longer demotes the other. The four apply loops fold into `applyEntries` (one home for the per-row pool and the rejection guard). `FetchFaultInjector` is shared between suites; new test `testThrownFallbackFetchRejectsTheRound`.
- `86c9033d6b` **test: pin the changeset round's fetch count instead of its wall-clock ratio** — addresses the line-231 thread. Asserts the exact seam read count `2 + 4 * ceil(records / 900)` for 100 and 2,000 records.

Tests: `swift build --build-tests` clean; full `SwiftDashSDKTests` suite 429 tests, 14 skipped, 0 failures (macOS, prebuilt FFI from the same base; the simulator leg of `run_tests.sh` was not run locally).

Resolved: line 949, line 906, line 231, and the pending-input / `loadUnaligned` threads already fixed by `5d0a0f1a14`.

Deferred to follow-ups, left open on purpose (self-review findings outside the scope of this round):
- line 1561 — existing-TXO core-address lookups can miss the prefetch
- line 869 — all-zero `spending_txid` sentinel seeded into `prefetchedTxids`
- line 3415 — `persistPlatformPaymentAddresses` still fetches per entry
- line 3436 — failed TXO-backfill chunk granularity / duplicate addresses in `IN` lists
- line 889 — round-cache memory peak on large rounds
- `BulkFetchPredicateTests` line 43 — test re-implements `chunked(_:size:)`

---
🤖 Posted autonomously by Claude on behalf of pasta.

packages/swift-sdk/SwiftTests/SwiftDashSDKTests/WalletChangesetRoundTests.swift
```diff
@@ -0,0 +1,280 @@
...
+    /// surface as a failed `save()` at `endChangeset`. Faulting every
+    /// `PersistentTransaction` read fails the bulk prefetch (which
+    /// demotes the chunk to per-row fetches) and then the first
+    /// fallback, so this pins both halves of the contract.
+    func testThrownFallbackFetchRejectsTheRound() throws {
+        let injector = FetchFaultInjector(faulting: PersistentTransaction.self)
+        let (handler, container) = try makeHandler(modelFetcher: injector)
+
+        XCTAssertFalse(
+            runRound(handler: handler, txs: spendChain(count: 3), expectPersisted: false),
```
thepastaclaw, line 239: **🟡 Suggestion: Exercise transient pending-input fetch failures followed by recovery**

This test faults every PersistentTransaction read, so the round aborts before exercising the distinct pending-input retry contract. The new tests do not cover pendingFetchFailed or the guards in resolveInputOutpoint and removePendingInputs that prevent an unsuccessful lookup from becoming an authoritative partial or empty cache entry. Add a scripted fetcher that fails the pending-input bulk fetch and first fallback, then succeeds on another lookup for the same outpoint within the round. Seed a persisted confirmed pending spender and stage a competing mempool observation before the funding TXO arrives; assert that the retry recovers both persisted and staged rows, preserves the confirmed spend linkage, and drains the pending rows. Also exercise a failed lookup from removePendingInputs followed by another access, so removing its empty-cache guard would fail a test. The existing happy-path and transaction-rejection tests would still pass if these authority guards regressed.

<sub>source: ['claude']</sub>

thepastaclaw (commented):
## Final validation — Phase 2 only (queue backlog)

Verified the supplied Phase-2 finding against head 86c9033d6ba7d0887665a79b8403cdacbc01cb50. The pending-input retry behavior and cache-authority guards are implemented, but the new tests do not exercise transient pending-input failures followed by recovery. Retained one non-blocking test-coverage suggestion; tests were not rerun during this verification.

Source: reviewer 1: `gpt-6-astra` (agent: `phase2-reviewer`, role: `general`); reviewer 2: `gpt-6-astra` (agent: `phase2-reviewer`, role: `ffi-engineer`); final verifier: `gpt-6-astra` (agent: `astra-verifier`, role: `final-verifier`)

### Review provenance
- Triage: `critical` by `gpt-6-astra` (effort low) — This substantial wallet-persistence refactor changes transaction, UTXO, and pending-input storage semantics, including cache coherence and fetch-failure handling, where bugs could corrupt persisted wallet state or misrepresent spendable funds.
- Phase 1 reviewers: **not run (skipped for throughput: 35 PRs queued, above the 10 limit)**
- Fresh verifier: `gpt-6-astra` — final-verifier; agent `astra-verifier`
- Phase 2 reviewers: `gpt-6-astra` — general (completed, effort xhigh); agent `phase2-reviewer`, `gpt-6-astra` — ffi-engineer (completed, effort xhigh); agent `phase2-reviewer`

🟡 1 suggestion(s)

<details>

```

In `packages/swift-sdk/SwiftTests/SwiftDashSDKTests/WalletChangesetRoundTests.swift`:
- [SUGGESTION] packages/swift-sdk/SwiftTests/SwiftDashSDKTests/WalletChangesetRoundTests.swift:234-239: Exercise transient pending-input fetch failures followed by recovery
  This test faults every PersistentTransaction read, so the round aborts before exercising the distinct pending-input retry contract. The new tests do not cover pendingFetchFailed or the guards in resolveInputOutpoint and removePendingInputs that prevent an unsuccessful lookup from becoming an authoritative partial or empty cache entry. Add a scripted fetcher that fails the pending-input bulk fetch and first fallback, then succeeds on another lookup for the same outpoint within the round. Seed a persisted confirmed pending spender and stage a competing mempool observation before the funding TXO arrives; assert that the retry recovers both persisted and staged rows, preserves the confirmed spend linkage, and drains the pending rows. Also exercise a failed lookup from removePendingInputs followed by another access, so removing its empty-cache guard would fail a test. The existing happy-path and transaction-rejection tests would still pass if these authority guards regressed.
```
</details>

llbartekll (changes requested):
Reviewed the diff at 86c9033 against the current `v4.2-dev`. The code on its own merge-base looks sound and the earlier threads (non-authoritative pending fetch, `loadUnaligned`, split TXO/pending prefetch coverage, rejecting the round on a thrown unique-key lookup, fetch-count regression test) are addressed. Requesting changes because the branch no longer fits the base it would land on.

**Blocking**

1. **Conflicts with `v4.2-dev`.** A local `merge-tree` shows 19 conflict regions in `PlatformWalletPersistenceHandler.swift` plus one in `AssetLockSpendVisibilityTests.swift`. Five commits touched the handler since this branch's merge-base: #4589, #4586, #4638, #4587, #4659. The green CI run is from Sep 8 against a base that predates all of them.

2. **#4589 already introduced a per-round index for the same hot path.** `ChangesetRoundIndex` (8bd3e53032) is a read-through index: the first keyed lookup in a round fetches store-only (`includePendingChanges = false`, so it never scans the round's staged objects), registers the hit, and every later lookup of that key answers from memory. That removes the same quadratic this PR targets, by a different mechanism. More importantly, the sweep phase from #4589/#4638 depends on the invariant documented above `fetchTransactionRow`: no keyed store-only lookup may ever refresh an object carrying staged state, and every first touch of a key goes through a registering helper. `WalletChangesetRoundCache` bulk-fetches with the default `includePendingChanges = true` and has its own "authoritative miss" semantics, so rebasing means reconciling two caches, not resolving hunks. Two options I see:
   - fold the bulk `IN` prefetch into `ChangesetRoundIndex` as a priming step at `beginChangeset`/start of `persistWalletChangeset` (populate `transactionsByTxid` / `txosByOutpoint` / `pendingInputsByOutpoint` + `pendingInputsFetched` / `coreAddressesByAddress` once per round, keep the existing helpers as the only read path), or
   - narrow this PR to what #4589 does not cover: the chunked prefetch in `persistAccountAddresses`, plus the shared test fixtures (`FFIFixtures`, `FetchFaultInjector`, `BulkFetchPredicateTests`) and the fetch-count regression test, re-targeted at the base helpers.

   Either way it would help to re-measure the incident workload on current `v4.2-dev` first, to see how much of the 59 GB / hours-of-CPU problem #4589 already removes.

**Non-blocking, carried over from the earlier automated threads (still open)**

- `upsertUtxo`: the core-address lookup keys on `record.address`, but `prefetchedAddresses` is built from the FFI `utxo.address`, and a negative fallback is not memoized, so an existing TXO whose stored address differs re-fetches on every round.
- `persistAccountAddresses`: a thrown TXO-backfill chunk now silently skips up to 900 addresses, where the old per-row `try?` lost one; a per-address fallback like the one for the row fetch beside it would restore parity. `allAddresses` also keeps duplicates in the `IN` lists.
- `persistPlatformPaymentAddresses` still fetches per entry.
- No test exercises the `pendingFetchFailed` retry path (thepastaclaw's Sep 9 suggestion).
- The all-zero `spending_txid` sentinel is inserted into `prefetchedTxids` unconditionally.

PR #4393, opened 2026-08-13
feat(drive): per-prefix ranked aggregates on compound indexes

QuantumExplorer:

Ranked aggregate index flags (`rankedCountable` / `rankedSummable` / `rankedAverageable`) were restricted to single-property indexes. This lifts the restriction: a compound index like `[identityId, class]` can now declare ranked flags with **per-prefix semantics** — the ranked flags land on the index's terminal property-name level, so each prefix value (each `identityId`) gets its own ordered secondary ranking only that prefix's `class` groups. There is deliberately no global cross-prefix ordering.

Both query surfaces gain **equality-prefix routing** to serve it:

- having-range: `WHERE identityId = X GROUP BY class HAVING AVG(grade) > 80 LIMIT n`
- ranked top-k: `WHERE identityId = X GROUP BY class ORDER BY AVG(grade) DESC LIMIT 5`

Everything is gated at protocol version 14 (unreleased): the ranked contract grammar is meta-schema v3 / parser generation 3, and both query surfaces route only through `DRIVE_ABCI_QUERY_VERSIONS_V3` — all PV14-only and unreleased, so the relaxation is edited in place in those modules. PV13 keeps rejecting every new shape (pinned by tests). No version slots were added or renumbered, so the v14 gate tests are unchanged.

**Stacked on #4384** (the having-range surface) — that PR's three commits are included here; merge it first (or merge this one into it).

### rs-dpp

- Removed the single-property rejection in `Index::try_from` (generation-3 grammar, PV14-only). Flag-dependency rules (`ranked*` requires the matching `range*`), the unique-index rejection and the `nullSearchable: false` rejection apply to compound indexes unchanged.
- Added the one genuine structural rejection as a cross-index check where the document type's full index set is visible (`validate_no_ranked_prefix_overlap` in `try_from_schema::common`, gated on the generation's `admit_ranked`): a compound ranked index whose **full leading prefix** also terminates a countable/summable index is refused — its aggregating value trees would demand the `NonCounted`/`NotSummed` shell grovedb structurally rejects around indexed trees (the `INDEXED_INNER_UNWRAPPABLE` fail-closed guard in drive remains as the backstop). Only the exact `n-1` prefix conflicts; shorter-prefix aggregating indexes and extensions past the ranked terminal are fine and covered by tests. The check runs on validating and non-validating parses alike, so check_tx/cache-warm paths cannot smuggle the shape past it.

### Write path (rs-drive)

No walker changes were needed: the v2 document index walkers are arity-generic and already emit the indexed tree type at a compound index's terminal level, and the pinned grovedb rev explicitly supports creating an indexed primary and populating it in the same batch (only overwrite-with-descendants is rejected) — so the per-prefix terminal trees, created lazily by the first document insert under each prefix, maintain their secondaries through the ordinary write path. This is verified end to end by the new integration tests (documents inserted through the real write path, per-prefix reads and proofs against the live root hash). Stale comments claiming grovedb rejects same-batch create+populate were corrected.

### Query surfaces (rs-drive)

- **Grammar** (`detect_ranked_mode_v0` / `detect_having_mode_v0`, both PV14-only-reachable): `where` clauses are now accepted as **equality pins** — each clause must be `==` on a distinct property (shared `equality_pins_from_where_clauses`). `IN` on a prefix is rejected loudly with its own not-yet message (one walk per element layers on future multi-`IN` branching); range operators and duplicate pins are rejected; multi-field `group_by` stays rejected, with the message steering to the pinned form.
- **Resolution** is now a single shared path per surface (`resolve_ranked_query_for_mode` / `resolve_having_query_for_mode`), used by the server executors AND the SDK proof helpers: the covering-index picker matches `pins + trailing group_by` against the index's properties (exact cover — partial pins match nothing), and the pins are encoded into prefix path segments with `DocumentType::serialize_value_for_key` — the same encoding the write path used to key the prefix value trees.
- **Path builder** (`indexed_property_name_tree_path_for_index`) extended with the prefix-value segments; prover and verifier both go through it (and through the shared resolver), so they cannot drift on which subtree a proof is about. Arity mismatches fail closed with a typed error.
- A pin on a prefix value that never saw a document addresses a nonexistent subtree and errors (read and prove alike) rather than fabricating an empty page — same contract as the existing empty-secondary limitation, pinned by a test.

### abci

No routing code changes: the v2 `compute_aggregate_mode_and_check_limit` already routes any grouped single-clause having / aggregate-ordered shape regardless of `where`, and both dispatchers forward `where_clauses` to drive untouched — drive owns the grammar. PV13 (query table v1 / helper v0) never routes either surface, so it rejects all new shapes before any contract fetch.

### Clients (rs-sdk / rs-drive-proof-verifier)

The SDK's ranked and having proof helpers now call drive's shared resolvers instead of hand-building the query structs, so pinned requests verify against the same prefixed path the prover used. Docs updated (request-shape sections, book chapter `document-ranked-trees.md`).

## How Has This Been Tested?

- **dpp**: compound+ranked accepted (validating + structural parse paths); prefix-overlap still rejected on both paths with a message naming both indexes and the conflict; only the exact prefix conflicts (aggregating index elsewhere, and plain prefix index, both accepted); `ranked*`-requires-`range*` enforced on compound indexes.
- **drive, new `grades-compound-ranked-contract.json` fixture** (`[identityId asc, class asc]`, averageable/rangeAverageable/rankedAverageable on `grade`), all through the real write path with proof round-trips against the live grovedb root hash:
  - having pinned prefix: exact-threshold exclusion (avg exactly 80 stays out under `>`), fractional average just above (80, 81 → 80.5 included), byte-exact string class keys, ascending and descending walks, proof round-trips for each — and isolation: identity Y's qualifying classes never leak into X's result, including a class *name* collision (`math` fails X's bound at avg 80 but passes Y's at 92.5).
  - ranked pinned prefix: per-prefix top-k with paginated proof round-trip, plus the same isolation.
  - rejections: unpinned prefix (no covering index, message names the needed shape), `IN` prefix (not-yet message), wrong-property pin, duplicate pins, range-operator pins, unknown prefix value (error, not empty page).
- **abci wire level**: pinned-prefix having request end to end (`ResultData.ranked`, `skipped` unset, isolation asserted) + proved variant; unpinned two-field `group_by` still rejected (flipped from the old single-property pin, now asserting the steering message); the pinned shape still rejected wholesale at protocol version 13.
- Full suites, real exit codes: `cargo check --workspace --all-targets` ✓, `drive` lib (3361 tests) ✓, `drive-abci` lib ✓, `dpp` (3873) ✓, `drive-proof-verifier` ✓, `dash-sdk` lib (208) ✓, `platform-version` (16) ✓, clippy clean on drive/dpp, `cargo fmt --all`.

## Breaking Changes

None for deployed networks — everything activates at protocol version 14, which is unreleased. Contracts that were invalid at PV14-in-development (compound ranked) become valid; the newly-rejected prefix-overlap shape was never registrable (it was covered by the broader single-property rejection).

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added ranked queries over compound indexes, including equality-pinned prefix filtering.
  * Added HAVING range queries for COUNT, SUM, and AVG, with optional ordering, limits, and proof generation.
  * Added support for compound-index HAVING queries with equality-pinned prefixes.
  * Added verified HAVING proof support in the SDK.

* **Bug Fixes**
  * Improved index validation and conflict detection for ranked compound indexes.

* **Documentation**
  * Updated query and protocol documentation with supported syntax, limitations, pagination rules, and protocol version requirements.


thepastaclaw:
🕓 Ready for review — 1 ahead in queue (commit 650416f5)
_Queue position: 2/2_
_ETA: start ~14:40 UTC · complete ~14:54 UTC (median 14m across 30 recent reviews; 2 slots)_
_Queued 9m ago · Last checked: 2026-08-13 14:40 UTC_

packages/rs-drive/src/query/drive_document_ranked_query/index_picker.rs
```diff
@@ -73,10 +103,165 @@ pub fn find_ranked_index_for_mode<'b>(
...
+                })?;
+            document_type
+                .serialize_value_for_key(&property.name, value, platform_version)
+                .map_err(|e| {
+                    Error::Query(QuerySyntaxError::InvalidParameter(format!(
+                        "the equality pin on `{}` does not encode as that property's \
+                         index key: {e}",
+                        property.name
+                    )))
+                })
```
thepastaclaw, line 275: **🔴 Blocking: Encode null prefix pins as empty path segments**

Optional system properties such as `$updatedAt`, `$transferredAt`, and `$creatorId` can legally be leading properties of a compound ranked index. When one is absent, the write walker turns `get_raw_for_document_type(...).unwrap_or_default()` into an empty path segment. This query path instead sends `Value::Null` through `serialize_value_for_key`; the system-property branches attempt identifier or integer conversion and fail before reaching the user-property null encoding. Consequently, a valid query such as `WHERE $updatedAt = null GROUP BY ...` cannot address the empty-key prefix subtree populated by the write path. Handle null before system-property serialization so server execution and SDK proof verification reconstruct the stored path exactly, and cover this case in ranked and HAVING proof tests.

```suggestion
            if value.is_null() {
                Ok(Vec::new())
            } else {
                document_type
                    .serialize_value_for_key(&property.name, value, platform_version)
                    .map_err(|e| {
                        Error::Query(QuerySyntaxError::InvalidParameter(format!(
                            "the equality pin on `{}` does not encode as that property's \
                             index key: {e}",
                            property.name
                        )))
                    })
            }
```

<sub>source: ['codex']</sub>

packages/rs-platform-version/src/version/drive_abci_versions/drive_abci_query_versions/v3.rs
```diff
@@ -0,0 +1,31 @@
...
+/// helper rejects every non-empty `having` ("HAVING clause is not yet
+/// implemented"); the v2 helper routes a grouped aggregate carrying
+/// exactly one `having` clause (`GROUP BY p HAVING <agg> <op> <value>
+/// LIMIT n`) to the having-range executor, which serves it as a
+/// value-bounded range read of the covering ranked index's axis
+/// secondary. Everything else — including multi-clause `having` and
+/// `having` on a select with no ranked axis — keeps the v1 behavior.
+///
+/// Same mixed-network rationale as the v1 → v2 flip: earlier protocol
+/// versions keep the v2 table and keep rejecting the shape, so nodes
```
thepastaclaw: **💬 Nitpick: Name the query tables actually selected before PV14**

The mixed-network rationale says earlier protocol versions keep the V2 query table, but no shipped protocol version selects it: protocol versions 1–11 select `DRIVE_ABCI_QUERY_VERSIONS_V0`, and versions 12–13 select `DRIVE_ABCI_QUERY_VERSIONS_V1`. Both use aggregate helper version 0 and reject ranked and HAVING requests, so the rationale is sound but the table attribution is inaccurate.

```suggestion
/// Mixed-network safety comes from the shipped tables: protocol versions
/// 1–11 select `DRIVE_ABCI_QUERY_VERSIONS_V0`, and versions 12–13 select
/// `DRIVE_ABCI_QUERY_VERSIONS_V1`. Both use helper version 0 and reject
/// ranked and `HAVING` shapes, so nodes agree until the PV14 upgrade carries.
/// The wire surface is unchanged —
```

packages/rs-sdk/src/platform/documents/document_having_entries.rs
```diff
@@ -0,0 +1,326 @@
...
+            let mut clause = count_over_100();
+            clause.operator = operator;
+            let query = base.clone().with_having(vec![clause]);
+            assert!(assert_having_shape(&query, platform_version()).is_err());
+        }
+    }
+
+    /// The generic FromProof guard in drive-proof-verifier must not be
+    /// reachable from the SDK path: this impl (on `DocumentQuery`) is
+    /// the one `fetch` resolves, and it runs the real verification.
```
thepastaclaw: **💬 Nitpick: Describe the limit contract tested by the HAVING test**

The comment discusses `FromProof` implementation resolution, but the test directly calls `assert_having_shape` and only verifies rejection of limits 0 and 101. Describe the inclusive client-side limit range and rejection behavior that the test actually exercises.

```suggestion
    /// HAVING limits must be within the inclusive range `1..=100`;
    /// values outside that range are rejected client side rather than
    /// clamped.
```

thepastaclaw (changes requested):

The compound-ranked implementation has two correctness blockers: contract validation applies the ranked item-key limit to leading prefix properties, and null equality pins for optional system-property prefixes do not reproduce the write path's empty key encoding. The remaining findings correct inaccurate public SDK and maintenance documentation.
Source: reviewer backend model gpt-5.6-sol; final verifier backend model gpt-5.6-sol; orchestration-only openclaw-agent/cliproxy/gpt-5.6-sol (not reviewer evidence).

### Review provenance

🔴 2 blocking | 🟡 1 suggestion(s) | 💬 2 nitpick(s)

<details>

```

In `packages/rs-dpp/src/data_contract/document_type/class_methods/try_from_schema/v3/mod.rs`:
- [BLOCKING] packages/rs-dpp/src/data_contract/document_type/class_methods/try_from_schema/v3/mod.rs:113-115: Apply the ranked key-length ceiling only to the terminal property
  `validate_index_properties` calls this check for every property in the index, but only the terminal property's encoded value becomes the indexed primary's item key and is concatenated with the ranked secondary's 8- or 16-byte sort key. Leading property values remain ordinary GroveDB path keys and should retain the generic 63-character or 255-byte limit. As written, a valid average-ranked index such as `[region, class]`, with `region.maxLength = 60` and a short terminal `class`, is rejected because the leading prefix exceeds the 59-character ranked limit even though it is never part of the average secondary key. Restrict this check to the terminal property and add a boundary test with a long leading prefix.

In `packages/rs-drive/src/query/drive_document_ranked_query/index_picker.rs`:
- [BLOCKING] packages/rs-drive/src/query/drive_document_ranked_query/index_picker.rs:256-264: Encode null prefix pins as empty path segments
  Optional system properties such as `$updatedAt`, `$transferredAt`, and `$creatorId` can legally be leading properties of a compound ranked index. When one is absent, the write walker turns `get_raw_for_document_type(...).unwrap_or_default()` into an empty path segment. This query path instead sends `Value::Null` through `serialize_value_for_key`; the system-property branches attempt identifier or integer conversion and fail before reaching the user-property null encoding. Consequently, a valid query such as `WHERE $updatedAt = null GROUP BY ...` cannot address the empty-key prefix subtree populated by the write path. Handle null before system-property serialization so server execution and SDK proof verification reconstruct the stored path exactly, and cover this case in ranked and HAVING proof tests.

In `packages/rs-sdk/src/platform/documents/document_query.rs`:
- [SUGGESTION] packages/rs-sdk/src/platform/documents/document_query.rs:293-299: Document the PV14 HAVING support on the public builder
  The public `DocumentQuery::with_having` documentation still states that every non-empty value is rejected and that the builder only exists ahead of server support. This contradicts both the updated `having` field documentation and the PV14 having-range API introduced by this PR, so SDK users are told that the new supported request is unusable.

In `packages/rs-platform-version/src/version/drive_abci_versions/drive_abci_query_versions/v3.rs`:
- [NITPICK] packages/rs-platform-version/src/version/drive_abci_versions/drive_abci_query_versions/v3.rs:18-20: Name the query tables actually selected before PV14
  The mixed-network rationale says earlier protocol versions keep the V2 query table, but no shipped protocol version selects it: protocol versions 1–11 select `DRIVE_ABCI_QUERY_VERSIONS_V0`, and versions 12–13 select `DRIVE_ABCI_QUERY_VERSIONS_V1`. Both use aggregate helper version 0 and reject ranked and HAVING requests, so the rationale is sound but the table attribution is inaccurate.

In `packages/rs-sdk/src/platform/documents/document_having_entries.rs`:
- [NITPICK] packages/rs-sdk/src/platform/documents/document_having_entries.rs:313-315: Describe the limit contract tested by the HAVING test
  The comment discusses `FromProof` implementation resolution, but the test directly calls `assert_having_shape` and only verifies rejection of limits 0 and 101. Describe the inclusive client-side limit range and rejection behavior that the test actually exercises.
```
</details>

packages/rs-platform-version/src/version/drive_versions/drive_document_method_versions/v2.rs
```diff
@@ -20,6 +20,7 @@ pub const DRIVE_DOCUMENT_METHOD_VERSIONS_V2: DriveDocumentMethodVersions =
             detect_count_mode: 0,
             detect_sum_mode: 0,
             detect_ranked_mode: 0,
+            detect_having_mode: 0,
```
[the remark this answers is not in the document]
thepastaclaw: `DriveDocumentQueryMethodVersions` is not serialized, and adding a field requires every historical constant to initialize it. The value is dormant for v10/v11 because those versions never route HAVING requests. This also follows the existing precedent from commit 60dbfa1dc1e, which added `detect_ranked_mode: 0` to the same V2 table. The freeze comment protects `primary_key_tree_type` and its historical storage behavior, not the source-level shape of the query-version struct.

thepastaclaw (changes requested):

Two correctness blockers remain: compound ranked-index validation rejects valid long leading prefixes, and null equality pins cannot reproduce the empty path segments written for absent optional system properties. Three documentation inaccuracies also remain.
Source: reviewer backend model gpt-5.6-sol; final verifier backend model gpt-5.6-sol; openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and is not reviewer evidence. Opus review is deferred by the preliminary blocker gate.

### Review provenance

<details>

```

In `packages/rs-dpp/src/data_contract/document_type/class_methods/try_from_schema/common/mod.rs`:
- [BLOCKING] packages/rs-dpp/src/data_contract/document_type/class_methods/try_from_schema/common/mod.rs:1052-1105: Apply the ranked key-length ceiling only to the terminal property
  `validate_index_properties` invokes the generation-3 ranked key-length callback for every property in the index. Only the terminal property's encoded value becomes an item key in the indexed primary and is concatenated with the ranked secondary's 8- or 16-byte sort key; leading values are ordinary GroveDB path segments and should retain the generic 63-character or 255-byte limit. This did not affect the previous single-property-only grammar, but this PR exposes it by admitting compound ranked indexes: an average-ranked `[region, class]` index with `region.maxLength = 60` and a short terminal `class` is rejected under the 59-character ranked ceiling even though `region` never becomes a ranked secondary item key. Restrict the callback to the terminal property and add a boundary test with a long leading prefix.

In `packages/rs-sdk/src/platform/documents/document_query.rs`:
- [SUGGESTION] packages/rs-sdk/src/platform/documents/document_query.rs:293-302: Document the PV14 HAVING support on the public builder
  The public `DocumentQuery::with_having` documentation still states that every non-empty value is rejected and that the builder only exists ahead of server support. That contradicts the updated `having` field documentation and the PV14 single-clause HAVING-range API introduced by this PR, so SDK users are told not to use the newly supported API.

In `packages/rs-drive/src/query/drive_document_ranked_query/index_picker.rs`:
- [BLOCKING] packages/rs-drive/src/query/drive_document_ranked_query/index_picker.rs:256-264: Encode null prefix pins as empty path segments
  (existing thread: https://github.com/dashpay/platform/pull/4393#discussion_r3772731001)
  Optional system properties such as `$updatedAt`, `$transferredAt`, and `$creatorId` can legally be leading properties of a compound ranked index. The write walker stores an absent value under an empty path segment because `get_raw_for_document_type(...).unwrap_or_default()` converts `None` to `Vec::new()`. This resolver instead sends `Value::Null` through `serialize_value_for_key`; the system-property branches immediately attempt identifier or integer conversion and reject null. Consequently, a valid pinned request such as `WHERE $updatedAt = null GROUP BY ...` cannot address the subtree populated by the write path. Handle null before system-property serialization and cover the case in ranked and HAVING proof round trips so server execution and SDK verification reconstruct the same path.

In `packages/rs-platform-version/src/version/drive_abci_versions/drive_abci_query_versions/v3.rs`:
- [NITPICK] packages/rs-platform-version/src/version/drive_abci_versions/drive_abci_query_versions/v3.rs:19-21: Name the query tables actually selected before PV14
  (existing thread: https://github.com/dashpay/platform/pull/4393#discussion_r3772731016)
  The mixed-network rationale says earlier protocol versions retain the V2 query table, but no shipped protocol version selects it. Protocol versions 1–11 select `DRIVE_ABCI_QUERY_VERSIONS_V0`, while versions 12–13 select `DRIVE_ABCI_QUERY_VERSIONS_V1`; both configure aggregate helper version 0 and reject ranked and HAVING requests. The safety conclusion is correct, but the table attribution is inaccurate.

In `packages/rs-sdk/src/platform/documents/document_having_entries.rs`:
- [NITPICK] packages/rs-sdk/src/platform/documents/document_having_entries.rs:313-315: Describe the limit contract tested by the HAVING test
  (existing thread: https://github.com/dashpay/platform/pull/4393#discussion_r3772731021)
  The comment discusses `FromProof` implementation resolution, but the test directly invokes `assert_having_shape` and verifies only that limits 0 and 101 are rejected. Describe the inclusive `1..=100` client-side limit contract and the fact that out-of-range values are rejected rather than clamped.
```
</details>

packages/rs-drive/src/query/drive_document_ranked_query/index_picker.rs
thepastaclaw: Resolved in this update — _Encode null prefix pins as empty path segments_ no longer present.

thepastaclaw (changes requested):

The compound-ranked query path is directionally sound, but two correctness blockers remain: validation applies the ranked item-key ceiling to leading path properties, and null pins cannot reproduce empty path segments written for absent optional system properties. Two documentation comments also remain inaccurate.
Source: reviewer backend model gpt-5.6-sol; final verifier backend model gpt-5.6-sol; openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

### Review provenance

🔴 2 blocking

<details>

```

In `packages/rs-dpp/src/data_contract/document_type/class_methods/try_from_schema/common/mod.rs`:
- [BLOCKING] packages/rs-dpp/src/data_contract/document_type/class_methods/try_from_schema/common/mod.rs:1099-1105: Apply the ranked key-length ceiling only to the terminal property
  `validate_index_properties` still invokes the generation-3 ranked key-length callback for every user property in the index. The storage walker creates the indexed primary only at the terminal property-name level, so only the terminal property's encoded value is concatenated with the ranked secondary's 8- or 16-byte sort key. Leading values are ordinary GroveDB path segments and retain the generic 63-character or 255-byte limit. This PR exposes the mismatch by admitting compound ranked indexes: an average-ranked `[region, class]` index with `region.maxLength = 60` and a short terminal `class` is rejected under the 59-character ranked ceiling even though `region` is not an indexed-primary item key. Restrict the ranked callback to the terminal property and add a boundary test for a long leading prefix.

In `packages/rs-drive/src/query/drive_document_ranked_query/index_picker.rs`:
- [BLOCKING] packages/rs-drive/src/query/drive_document_ranked_query/index_picker.rs:256-264: Encode null prefix pins as empty path segments
  (existing thread: https://github.com/dashpay/platform/pull/4393#discussion_r3772731001)
  Optional system properties such as `$updatedAt`, `$transferredAt`, and `$creatorId` can legally lead a compound ranked index. When such a property is absent, the write walker calls `get_raw_for_document_type(...).unwrap_or_default()` and stores the next level under an empty path segment. This resolver instead passes `Value::Null` to `serialize_value_for_key`; the system-property branches immediately attempt identifier or integer conversion and reject null. A valid pinned ranked or HAVING request such as `WHERE $updatedAt = null GROUP BY ...` therefore cannot address the subtree populated by the write path. Handle null before system-property serialization and cover ranked and HAVING proof round trips so reads, proofs, and SDK verification all reconstruct the stored path.

In `packages/rs-platform-version/src/version/drive_abci_versions/drive_abci_query_versions/v3.rs`:
- [NITPICK] packages/rs-platform-version/src/version/drive_abci_versions/drive_abci_query_versions/v3.rs:18-20: Name the query tables actually selected before PV14
  (existing thread: https://github.com/dashpay/platform/pull/4393#discussion_r3772731016)
  The mixed-network rationale says earlier protocol versions retain the V2 query table, but no earlier protocol version selects that table. Protocol versions 1–11 select `DRIVE_ABCI_QUERY_VERSIONS_V0`, while versions 12–13 select `DRIVE_ABCI_QUERY_VERSIONS_V1`; both use aggregate helper version 0 and reject ranked and HAVING requests. The safety conclusion is correct, but the table attribution is inaccurate.

In `packages/rs-sdk/src/platform/documents/document_having_entries.rs`:
- [NITPICK] packages/rs-sdk/src/platform/documents/document_having_entries.rs:313-315: Describe the limit contract tested by the HAVING test
  (existing thread: https://github.com/dashpay/platform/pull/4393#discussion_r3772731021)
  The comment discusses `FromProof` implementation resolution, but this test directly invokes `assert_having_shape` and verifies that limits 0 and 101 are rejected. Describe the inclusive `1..=100` client-side limit contract and that out-of-range values are rejected rather than clamped.
```
</details>

packages/rs-drive/src/query/drive_document_ranked_query/index_picker.rs
QuantumExplorer: Fixed in b17dfcf620: a null pin short-circuits to an empty encoded segment before the system-property encoders run, matching the write walkers' `unwrap_or_default()`. Round-tripped (read + prove + client-side verify, with `equality_prefix_values == [vec![]]` asserted) on both surfaces via a new optional-leading-`tag` doctype in the compound fixture: `a_null_pin_addresses_the_absent_value_prefix` (ranked) and `a_null_pin_bounds_the_absent_value_prefix` (having).

packages/rs-platform-version/src/version/drive_abci_versions/drive_abci_query_versions/v3.rs
QuantumExplorer: Fixed in b17dfcf620 with the suggested attribution: the mixed-network sentence now names `DRIVE_ABCI_QUERY_VERSIONS_V0` (protocol 1–11) and `V1` (12–13), both on helper 0 rejecting ranked and HAVING shapes.

packages/rs-sdk/src/platform/documents/document_having_entries.rs
QuantumExplorer: Fixed in b17dfcf620: the doc comment now describes the inclusive `1..=100` client-side contract the test pins — 0 (the unset sentinel) and 101 rejected rather than clamped, because the limit is echoed in the proof envelope.

packages/rs-sdk/src/platform/documents/document_having_entries.rs
```diff
@@ -0,0 +1,326 @@
...
+    #[test]
+    fn limit_is_required_and_capped_client_side() {
+        for limit in [0u32, 101] {
+            let query = hashtags_over_100().with_limit(limit);
+            assert!(
+                assert_having_shape(&query, platform_version()).is_err(),
+                "LIMIT {limit} is outside 1..=100 and must be rejected, not clamped"
+            );
+        }
+    }
```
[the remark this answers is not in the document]
QuantumExplorer: Fixed in b17dfcf620 (same change as the duplicate thepastaclaw nitpick): the comment now states the 1..=100 reject-not-clamp contract.

packages/rs-platform-version/src/version/drive_abci_versions/drive_abci_query_versions/v3.rs
```diff
@@ -0,0 +1,31 @@
...
+/// agree until the upgrade carries. The wire surface is unchanged —
```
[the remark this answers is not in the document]
QuantumExplorer: Fixed in b17dfcf620: the v3 doc now names the shipped tables (V0 for protocol 1–11, V1 for 12–13) instead of implying a shipped version selects V2.

QuantumExplorer:
Reviewed

PR #4394, opened 2026-08-13
chore: bump rust-dashcore to the latest dev revision

QuantumExplorer:

Moves the workspace's rust-dashcore pins up to the current head of `dev`, picking up the key rotation ChainLock signature fix ([rust-dashcore#934](https://github.com/dashpay/rust-dashcore/pull/934)).

That fix matters for us: `feed_qr_info` was collapsing each diff's `quorumsCLSigs` down to a single signature taken from the first rotating quorum it saw, then stamping the same 4-signature tuple onto every entry of the batch. The active rotation set can span several cycles at once, because a failed DKG leaves the previous cycle's quorum in place at that index, so every straggler got quarter modifiers built from another cycle's signatures. The reconstructed member sets then disagreed with the real signers and the whole QRInfo hard-failed with `AllCommitmentAggregatedSignatureNotValid` on the first fresh-sync response — permanently wedging masternode sync, ChainLock validation and InstantSend verification.

## What was done?

Bumped all eight rust-dashcore workspace pins from `36b49cb7` to `173ffac0` and regenerated the affected `Cargo.lock` entries. `Cargo.toml` and `Cargo.lock` are the only files touched; both pin sites are workspace-level (`packages/rs-platform-wallet` only inherits via `workspace = true`), and no stale revisions remain anywhere in the tree.

The range is exactly one upstream commit. Alongside the signature keying it hardens several peer-controlled inputs in the same paths — an out-of-range `quorum_index` that sign-extended and underflowed the cycle base, a truncated `lastCommitmentPerIndex` that passed the trust gate, and whole-map cycle replacement that let one genuine commitment drop the other indices of a stored cycle. It also degrades individual quorums to `Skipped` rather than poisoning an entire feed when context is missing.

**No platform source changes were needed.** The reshaped surface — `QRInfoFeedResult`'s new fields, `find_rotated_masternodes_for_quorums` returning a per-quorum `Result`, the new `QuorumValidationError` variants and the new `qrinfo-capture` feature — has zero call sites in this repo; every consumer lives inside rust-dashcore and dash-spv.

## How Has This Been Tested?

- `cargo check --workspace --all-targets` — clean
- `cargo check --workspace --all-features --all-targets` — clean
- `cargo test -p dpp --all-features` — 3955 passed, 0 failed
- `cargo test -p drive-abci --lib` — 2670 passed, 1 failed

The single `drive-abci` failure is pre-existing and unrelated to this bump. `test_data_contract_creation_fails_with_more_than_fifty_keywords` fails with `config version 0 is not supported, minimum version is 1`; it passes in isolation, and the identical suite run against the base pin `36b49cb7` fails on the same test with the same counts (2670 passed / 1 failed). It is the latent `PlatformVersion::set_current` global test race — a concurrent test resets the global version mid-run and the contract serializes at config v0. Worth fixing, but separately from this bump.

## Breaking Changes

None. No consensus-relevant platform behavior changes; the upstream fix corrects quorum reconstruction that previously failed outright.

## Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas (n/a — no source changes)
- [ ] I have added or updated relevant unit/integration/functional/e2e tests (n/a — pin bump; existing suites run above)
- [ ] I have made corresponding changes to the documentation if needed (n/a)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1229,7 +1229,7 @@ version = "3.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "faf9468729b8cbcea668e36183cb69d317348c2e08e994829fb56ebfdfbaac34"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -1662,7 +1662,7 @@ dependencies = [
 [[package]]
 name = "dash-network"
 version = "0.45.0"
-source = "git+https://github.com/dashpay/rust-dashcore?rev=36b49cb7f9c07ce8f2ed63f965489c50968d19f7#36b49cb7f9c07ce8f2ed63f965489c50968d19f7"
+source = "git+https://github.com/dashpay/rust-dashcore?rev=173ffac0fdc0c73dda0626cf385bbcfcf2437aeb#173ffac0fdc0c73dda0626cf385bbcfcf2437aeb"
 dependencies = [
  "bincode",
  "bincode_derive",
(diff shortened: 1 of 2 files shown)


thepastaclaw:
ℹ️ Review superseded (commit fa6aaece)
_Reason: PR merged before the review completed._
_Last checked: 2026-08-13 09:19 UTC_

PR #4396, opened 2026-08-13
fix(drive)!: cursor pagination over multi-branch query levels dropped sibling-branch documents

QuantumExplorer:

Cursor pagination (`startAt`/`startAfter`) over a document query whose best index level is multi-branch — a single `In` clause (or a range clause) with leftover index properties — returned wrong result sets:

- branches ordered **after** the cursor's branch silently **dropped** every document whose leftover-property value sorted below the cursor's key;
- branches ordered **before** the cursor's branch wrongly **included** documents above the cursor's key.

Example (family contract, index `[firstName, lastName]`): `firstName IN [Adam, Ben, Cara] ORDER BY firstName, lastName` with `startAfter` a document `(Adam, Moore)` dropped `(Ben, Barber)` and `(Cara, Abbott)`.

The cause: the v0 lowering builds the outer multi-branch level with no cursor awareness and then `recursive_insert_on_query` bakes the cursor's per-level start keys into the **default** subquery applied to *every* sibling branch. The block comment in that function ("the start at document is used only on the conditional subquery and not on the main query") already described the correct construction; the code didn't implement it at the multi-branch top level.

A second latent defect surfaced while testing the reversed direction: for **cursorless** queries, the leftover-level recursion built each level with the index property's own direction, ignoring an opposite-direction `orderBy` — `orderBy [lastName, desc]` on a leftover property returned within-branch results ascending. All existing desc-on-leftover tests used cursors, which route through a different (direction-correct) arm, so this was never caught.

This is the machinery that led #4391 to reject `startAt`/`startAfter` combined with multiple `In` clauses.

## What was done?

Rebased over #4391 (merged), which introduced the shared `DriveDocumentQueryMethodVersions.non_primary_key_path_query` slot and the `query/non_primary_key_path_query/{v0,v1}` module split. Both PRs' v14 semantics now live under that one slot, as planned when the slot name was deliberately shared.

The v1 lowering (protocol v14, unreleased) now only routes by shape: both shape lowerings are versioned methods with their own `DriveDocumentQueryMethodVersions` slots (`non_primary_key_single_in_path_query`, `non_primary_key_multiple_in_path_query`, 0 in every table, exact-match dispatch erroring on unknown values) and frozen `v0` modules. The new at-most-one-`In` lowering, `single_in_path_query/v0`, replaces v1's previous fallthrough to the pre-v14 lowering for those shapes. When the last clause is range-typed with leftover index properties and a cursor is present:

- the outer level is built with the cursor document marked included, so the existing branch trimming removes branches ordered before the cursor's branch;
- the default subquery is unfiltered (`insert_all`, direction derived from `orderBy`), so branches after the cursor return everything;
- the cursor's own branch is refined by a conditional subquery at its branch key.

The cursor-recursion helpers live beside their owner in `single_in_path_query/v0` as `recursive_insert_on_query_ordered_with_cursor` / `recursive_conditional_insert_on_query_ordered` / `recursive_create_query_ordered`, identical to their pre-v14 counterparts except that every cursorless level derives its direction from `orderBy` (falling back to the index property's own) — the same rule #4391's `recursive_insert_on_query_ordered` applies on the multi-`In` path, so v1 direction semantics are uniform. The v0 module is untouched, per the frozen-version-module convention.

Not covered (still v0-shaped inside v1, same bug family, can be follow-ups): the `In` + range combination (`subquery_clause` path) with a cursor, and the no-where-clause path (`orderBy`-only with a cursor and leftover properties).

## How Has This Been Tested?

New integration test `test_family_single_in_clause_with_cursor_keeps_sibling_branches_intact` (rs-drive `tests/query_tests.rs`) with a fixed 7-person dataset (new `setup_family_tests_with_people` helper) and brute-force expected orderings. Five scenarios — page-one asc and desc, cursor in the first branch, cursor in a middle branch, and a descending cursor — each asserted on both the no-proof execution path and the proof path (proof results must byte-match, root hash checked). The test fails on the v0 lowering exactly as described and passes with the fix.

A second test, `test_family_single_in_clause_with_cursor_v0_lowering_frozen_at_protocol_v13`, runs the same scenarios at `PlatformVersion::get(13)` and pins the old defective outputs on both execution and proof paths, plus asserts protocol v13 selects method version 0 — proving the gate empirically in both directions (v14 fixed, ≤v13 byte-for-byte frozen).

Regression: full `cargo test -p drive` (3336 lib tests plus all integration suites, including #4391's new multi-`In` tests) green; rs-drive compiles with `--no-default-features --features verify`; clippy clean on both touched crates (`--all-targets` and `--all-features --tests`); drive-abci `document_query` tests green.

## Breaking Changes

Result sets and ordering of accepted-and-answered document queries change for the shapes above. The change only activates at protocol v14 via `non_primary_key_path_query: 1` (shared with #4391's multi-`In` support); all released protocol versions retain the v0 lowering.

## Checklist:


thepastaclaw:
🕓 Ready for review — 4 ahead in queue (commit 5882f9f6)
_Queue position: 5/7 · 2 reviews active_
_ETA: start ~12:01 UTC · complete ~12:19 UTC (median 17m across 30 recent reviews; 2 slots)_
_Queued 26m ago · Last checked: 2026-08-13 11:20 UTC_

packages/rs-drive/src/query/mod.rs
```diff
@@ -1947,7 +1947,23 @@ impl<'a> DriveDocumentQuery<'a> {
...
+                    // direction, ignoring an opposite-direction order_by on a
+                    // leftover property; from v1 of non_primary_key_path_query
+                    // the order_by direction wins.
+                    let direction = if platform_version
+                        .drive
+                        .methods
+                        .document
+                        .query
+                        .non_primary_key_path_query
+                        >= 1
```
thepastaclaw: **🔴 Blocking: Dispatch the lowering through exact method versions**

`non_primary_key_path_query` is a new consensus-relevant method-version slot, but this check and the second `>= 1` check at lines 2230–2236 embed v0/v1 selection inside the existing recursive implementation. An unsupported future value such as `2` therefore executes v1 instead of returning an unknown-version error, which can hide an incomplete or inconsistent platform-version table. It also leaves the released v0 behavior interwoven with code that future changes can modify accidentally. Route the entry point through an exact match (`0 => v0`, `1 => v1`, unknown => `DriveError::UnknownVersionMismatch`) and keep the version-specific lowering implementations separately frozen.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The pagination fix is protocol-gated and has strong regression coverage for both the corrected v14 behavior and frozen v13 outputs. However, the new consensus-relevant method slot is consumed through inline `>= 1` checks rather than exact version dispatch, so unsupported versions silently inherit v1 behavior and the implementations are not independently frozen.
Source: reviewer backend model gpt-5.6-sol; final verifier backend model gpt-5.6-sol. Orchestration-only model openclaw-agent/cliproxy/gpt-5.6-sol is explicitly not reviewer evidence.

### Review provenance

🔴 1 blocking

<details>

```

In `packages/rs-drive/src/query/mod.rs`:
- [BLOCKING] packages/rs-drive/src/query/mod.rs:1954-1960: Dispatch the lowering through exact method versions
  `non_primary_key_path_query` is a new consensus-relevant method-version slot, but this check and the second `>= 1` check at lines 2230–2236 embed v0/v1 selection inside the existing recursive implementation. An unsupported future value such as `2` therefore executes v1 instead of returning an unknown-version error, which can hide an incomplete or inconsistent platform-version table. It also leaves the released v0 behavior interwoven with code that future changes can modify accidentally. Route the entry point through an exact match (`0 => v0`, `1 => v1`, unknown => `DriveError::UnknownVersionMismatch`) and keep the version-specific lowering implementations separately frozen.
```
</details>

packages/rs-drive/src/query/mod.rs
QuantumExplorer: Resolved by the rebase onto #4391 plus a follow-up commit. The inline `>= 1` checks this comment pointed at no longer exist:

- `get_non_primary_key_path_query` now dispatches through an exact match on the slot (`0 => v0`, `1 => v1`, unknown => `DriveError::UnknownVersionMismatch` with `known_versions: [0, 1]`), as does `validate_in_clause_shape`, the only other read of the slot.
- The released v0 lowering lives untouched in `query/non_primary_key_path_query/v0/`; the v14 changes are isolated in `v1/` and its shape submodules, so future edits can't accidentally modify v0. The `test_family_single_in_clause_with_cursor_v0_lowering_frozen_at_protocol_v13` freeze test additionally pins v0's outputs byte-for-byte on both the execution and proof paths.
- As of 5882f9f69e the two shape lowerings are themselves versioned methods behind their own slots (`non_primary_key_single_in_path_query`, `non_primary_key_multiple_in_path_query`), each with the same exact-match-or-error dispatch and a frozen `v0` module owning its recursion helpers.

QuantumExplorer:
Reviewed

PR #4399, opened 2026-08-13
feat: register funded identities via a ChainLock asset lock

shumkov:

Bootstrapping a funded Platform identity from a Core faucet had no working tool on this branch, and the earlier InstantSend-based approach it supersedes does not merely fail — **it burns the funds on every run.**

That approach proves the asset lock by subscribing to DAPI's transaction stream filtered on the **one-time asset-lock address**. That address exists only inside the asset-lock **special-transaction payload** (`credit_outputs`), and rs-dapi's stream matcher (`matches_transaction`) never inspects the special-tx payload — it looks only at the transaction's regular inputs, output scripts, and txid. So the InstantSend lock is never forwarded to the subscriber, and the client times out **~120 s after broadcast, with the money already committed on-chain.** The DASH is stranded on a one-time key that is then discarded: a silent, unrecoverable loss.

## What was done?

Adds `register_identity`, a bin that funds and registers a Platform identity from a Dash Core faucet wallet using a **ChainLock** asset-lock proof instead. It builds the asset-lock funding transaction from the faucet's UTXOs, has Core sign and broadcast it, obtains the proof, and registers the identity — printing only the new identity id.

- **ChainLock, not InstantSend.** After broadcast it polls Core (`getrawtransaction`) until the funding tx is chain-locked, waits for Platform's `coreChainLockedHeight` to reach that height, then builds a `ChainAssetLockProof` directly. No stream, so the proof cannot be silently dropped. On any network that chain-locks promptly (devnets, testnet) this costs a couple of minutes.
- **Recoverable on failure.** The one-time asset-lock WIF (and the identity's private keys) are written to the credentials file **before** the funding transaction is broadcast. An asset lock is irreversible once on-chain, so if any later step fails the locked funds are still recoverable via that one-time key (retry identity-create or top-up with the same outpoint). The prior bin captured nothing pre-broadcast — a failure there was an outright loss.
- **No secrets on stdout.** Only the identity id is printed; key material goes solely to the mode-600 credentials file. The Core RPC password is read from `CORE_RPC_PASSWORD` so it never appears in process arguments. Core does the signing (`signrawtransactionwithwallet`, with a hard full-sign check before broadcast).
- The module docs state the InstantSend failure mode plainly and say **do not reintroduce it** — the comment is the stop sign for the next person.

Note: `register_identity` was **not present on v4.2-dev** (the branch had no funding bin at all), so this adds the working tool rather than replacing anything on-branch. The broken InstantSend version lived only in local load-tool worktrees that were never merged upstream, so there is nothing to delete here.

## How Has This Been Tested?

Exercised end-to-end against **devnet-moutai** (protocol 14, rs-dapi 4.2.0-dev.1):

- A **1 DASH** de-risk identity — ChainLock proof obtained in ~107 s, resolved on-chain (proved `getIdentity`).
- The **~2.5T-credit campaign identity** (25 DASH → 2,499,898,360,300 credits) — ChainLock proof in ~266 s, balance confirmed on-chain (proved + unproved `getIdentityBalance`).

Both obtained a ChainLock asset-lock proof (never InstantSend) and registered cleanly. `cargo fmt` and `cargo clippy -p rs-scripts --bin register_identity` are clean; the bin compiles green from the `v4.2-dev` base.

## Breaking Changes

None. Net-new bin plus a one-line `rs-scripts` Cargo.toml change enabling `simple-signer`'s `state-transitions` feature (needed for the multi-key `SimpleSigner`).

## Checklist:

- [ ] I have added "!" to the title and described breaking changes in the corresponding section if my code contains any (no breaking changes)

## Summary by CodeRabbit

* **New Features**
  * Added a command-line workflow for creating and registering funded Platform identities.
  * Supports configurable networks, funding wallets, key counts, credential storage, and Core/DAPI settings.
  * Can resume registration using a previously broadcast funding transaction.
  * Waits for required ChainLocks before completing registration.
  * Saves recovery credentials and identity metadata.
  * Adds funding, validation, timeout handling, and clear identity output for streamlined registration.

---

### Update: resume mode + non-stranding proof timeout (2nd commit)

Field-hardening after a slow platform ChainLock ingestion exceeded the original 300s proof wait and left 25 DASH locked in a broadcast, chainlocked asset lock while the tool exited with an error — the same **stranded-funds** failure class as the InstantSend bug this bin replaces. Fixed in the follow-up commit:

- **`PROOF_TIMEOUT` raised to 600s** so the wait outlasts a normal core-ChainLock ingestion lag.
- **Non-stranding timeout:** on timeout the tool no longer exits with a bare error. It states that the asset lock is broadcast and on-chain (funds pending, not lost), points at the saved one-time WIF in `--out-env`, and prints the exact `--resume-txid <txid>` command to finish registration against the existing lock — spending nothing new.
- **`--resume-txid`:** reuse an already-broadcast asset lock (one-time WIF read from `--out-env`) and register against it. The identity id derives from the asset-lock outpoint, so the identity keys are generated fresh on resume.

Verified on devnet-moutai: recovered a real stranded 25-DASH lock into a ~2.5T-credit identity via `--resume-txid`, no new funds spent.

### DIFF
--- a/packages/rs-scripts/src/bin/register_identity.rs
+++ b/packages/rs-scripts/src/bin/register_identity.rs
@@ -53,9 +53,18 @@
 //! the one-time asset-lock WIF and the identity's private keys — is
 //! written to the `--out-env` file **before** the transaction is
 //! broadcast. If any later step fails, the locked funds are still
-//! recoverable via the one-time key (retry the identity-create or
-//! top-up with the same outpoint). Only the final identity id is ever
+//! recoverable via the one-time key. Only the final identity id is ever
 //! printed to stdout; key material never touches stdout or the logs.
+//!
+//! The ChainLock wait can legitimately take minutes (platform ingests a
+//! fresh core ChainLock with some lag), so the proof timeout is generous.
+//! If it is nonetheless exceeded, the tool does NOT strand the funds: it
+//! reports that the asset lock is broadcast and on-chain, points at the
+//! saved one-time WIF, and prints the exact command to finish the
+//! registration against the existing lock — re-run with `--resume-txid
+//! <txid>`, which reuses the lock and spends nothing new. Re-running
+//! without `--resume-txid` would broadcast a second asset lock and strand
+//! the first, so the resume path exists precisely to avoid that.
 //!
 //! # Endpoints
 //!
@@ -69,6 +78,7 @@
 use std::io::Write;
 use std::os::unix::fs::OpenOptionsExt;
 use std::process::ExitCode;
+use std::str::FromStr;
 use std::time::{Duration, Instant};
 
 use clap::Parser;
@@ -80,7 +90,9 @@ use dpp::dashcore::consensus::encode::{deserialize, serialize};
 use dpp::dashcore::secp256k1::Secp256k1;
 use dpp::dashcore::transaction::special_transaction::asset_lock::AssetLockPayload;
 use dpp::dashcore::transaction::special_transaction::TransactionPayload;
-use dpp::dashcore::{Address, Network, OutPoint, PrivateKey, ScriptBuf, Transaction, TxIn, TxOut};
+use dpp::dashcore::{
+    Address, Network, OutPoint, PrivateKey, ScriptBuf, Transaction, TxIn, TxOut, Txid,
+};
 use dpp::dashcore_rpc::{Auth, Client, RpcApi};
 use dpp::identity::accessors::IdentityGettersV0;
 use dpp::identity::identity_public_key::accessors::v0::IdentityPublicKeyGettersV0;
@@ -102,8 +114,12 @@ const ASSET_LOCK_FEE_DUFFS: u64 = 100_000;
 /// Minimum change worth returning; below this, fold into the fee.
 const DUST_DUFFS: u64 = 10_000;
 /// How long to wait for the asset-lock tx to be ChainLocked and for
-/// platform to reach that core height before giving up.
-const PROOF_TIMEOUT: Duration = Duration::from_secs(300);
+/// platform to reach that core height before giving up. Generous, because
+/// platform's ingestion of a fresh core ChainLock can lag by minutes; a
+/// short budget here risks the tool giving up on a lock that is about to
+/// become usable. On timeout the tool still prints a resume recipe rather
+/// than stranding the funds (see the timeout handling in `run`).
+const PROOF_TIMEOUT: Duration = Duration::from_secs(600);
 
 #[derive(Parser, Debug)]
 #[command(
@@ -153,6 +169,14 @@ struct Args {
     /// here (never printed to stdout).
     #[arg(long = "out-env")]
     out_env: String,
+
+    /// Resume a previously-broadcast asset lock instead of creating a new
+    /// one. Pass the asset-lock txid; the one-time WIF is read from the
+    /// `--out-env` file (`ASSET_LOCK_ONE_TIME_WIF`). Registers the identity
+    /// against the existing chainlocked asset lock — NO new funds are spent.
+    /// Use this to recover after a proof-wait timeout stranded a lock.
+    #[arg(long = "resume-txid")]
+    resume_txid: Option<String>,
 }
 
 fn parse_network(s: &str) -> Result<Network, String> {
@@ -180,6 +204,34 @@ fn append_creds(path: &str, lines: &[(String, String)]) -> Result<(), String> {
         writeln!(f, "{k}={v}").map_err(|e| format!("failed to write creds file: {e}"))?;
     }
     Ok(())
+}
+
+/// Read the last value for `key` from a `KEY=value` credentials file.
+/// Last-wins, matching how the file is sourced by a shell.
+fn read_creds_value(path: &str, key: &str) -> Result<Option<String>, String> {
+    let contents = std::fs::read_to_string(path)
+        .map_err(|e| format!("failed to read creds file {path}: {e}"))?;
+    let prefix = format!("{key}=");
+    Ok(contents
+        .lines()
+        .filter_map(|l| l.strip_prefix(&prefix))
+        .next_back()
+        .map(str::to_string))
+}
+
+/// Message for a proof-wait timeout that makes the failure NON-stranding:
+/// the asset lock is already broadcast and on-chain, so the funds are not
+/// lost — they are pending. It names the recoverable one-time key and the
+/// exact command to resume without spending new funds.
+fn strand_safe_timeout(txid: &Txid, out_env: &str, waiting_for: &str) -> String {
+    format!(
+        "timed out waiting for {waiting_for}.\n\
+         The asset lock IS broadcast and on-chain — the funds are NOT lost, only pending.\n\
+         The one-time key that owns it is saved in {out_env} (ASSET_LOCK_ONE_TIME_WIF).\n\
+         Resume WITHOUT spending new funds by re-running the same command plus:\n\
+         \x20   --resume-txid {txid}\n\
+         Do NOT re-run without --resume-txid — that broadcasts a second asset lock and strands this one."
+    )
 }
 
 #[tokio::main(flavor = "multi_thread")]
@@ -252,8 +304,10 @@ async fn run() -> Result<(), String> {
     )
     .map_err(|e| format!("failed to connect to Core RPC: {e}"))?;
 
-    // --- Generate the identity (with private keys) and a one-time
-    //     asset-lock key ---
+    // --- Generate the identity keys ---
+    // The identity id derives from the asset-lock outpoint, not from these
+    // keys, so the key set is generated the same way in both the fresh and
+    // resume paths.
     let mut rng = StdRng::from_entropy();
     let (identity, key_material): (Identity, Vec<(IdentityPublicKey, [u8; 32])>) =
         Identity::random_identity_with_main_keys_with_private_key(
@@ -263,101 +317,128 @@ async fn run() -> Result<(), String> {
         )
         .map_err(|e| format!("failed to generate identity: {e}"))?;
 
-    let secp = Secp256k1::new();
-    let mut secp_rng = dpp::dashcore::secp256k1::rand::thread_rng();
-    let one_time_secret = dpp::dashcore::secp256k1::SecretKey::new(&mut secp_rng);
-    let one_time_private_key = PrivateKey::new(one_time_secret, network);
-    let one_time_public_key = one_time_private_key.public_key(&secp);
-    let one_time_key_hash = one_time_public_key.pubkey_hash();
-    let one_time_address = Address::p2pkh(&one_time_public_key, network);
-
-    // --- Build the asset-lock funding transaction ---
-    let tx = build_asset_lock_transaction(&core, amount_duffs, &one_time_key_hash)?;
-    let unsigned_hex = hex::encode(serialize(&tx));
-
-    eprintln!(
-        "Signing asset-lock tx ({} inputs, locking {} DASH) via Core wallet...",
-        tx.input.len(),
-        args.fund_dash
-    );
-    let signed = core
-        .sign_raw_transaction_with_wallet(unsigned_hex.as_str(), None, None)
-        .map_err(|e| format!("signrawtransactionwithwallet failed: {e}"))?;
-    if !signed.complete {
-        return Err(
-            "Core could not fully sign the asset-lock tx (signrawtransactionwithwallet returned \
-             incomplete — is the faucet wallet loaded and are the inputs spendable?)"
-                .to_string(),
-        );
-    }
-    let signed_hex = hex::encode(&signed.hex);
-    let signed_tx: Transaction =
-        deserialize(&signed.hex).map_err(|e| format!("failed to parse signed tx: {e}"))?;
-    let txid = signed_tx.txid();
-
-    // --- Capture recovery credentials BEFORE broadcast ---
-    // Once the asset lock is on-chain, the locked funds are recoverable
-    // only via the one-time key + tx outpoint (retry identity-create or
-    // top-up). Persist everything needed for that before we spend.
-    let mut pre_creds: Vec<(String, String)> = vec![
-        ("ASSET_LOCK_TXID".to_string(), txid.to_string()),
-        (
-            "ASSET_LOCK_ONE_TIME_WIF".to_string(),
-            one_time_private_key.to_wif(),
-        ),
-        (
-            "ASSET_LOCK_ONE_TIME_ADDRESS".to_string(),
-            one_time_address.to_string(),
-        ),
-        (
-            "ASSET_LOCK_FUND_DASH".to_string(),
-            args.fund_dash.to_string(),
-        ),
-    ];
+    // Encode the identity key material for the credentials file.
+    let mut identity_creds: Vec<(String, String)> = Vec::new();
     for (pk, secret) in &key_material {
         let wif = PrivateKey::from_byte_array(secret, network)
             .map(|k| k.to_wif())
             .map_err(|e| format!("failed to encode identity key {} as WIF: {e}", pk.id()))?;
-        pre_creds.push((format!("IDENTITY_KEY_{}_WIF", pk.id()), wif));
+        identity_creds.push((format!("IDENTITY_KEY_{}_WIF", pk.id()), wif));
     }
     if let Some((critical_key, critical_secret)) = find_critical_auth_key(&identity, &key_material)
     {
         let wif = PrivateKey::from_byte_array(&critical_secret, network)
             .map(|k| k.to_wif())
             .map_err(|e| format!("failed to encode critical key as WIF: {e}"))?;
-        pre_creds.push((
+        identity_creds.push((
             "IDENTITY_CRITICAL_AUTH_KEY_ID".to_string(),
             critical_key.id().to_string(),
         ));
-        pre_creds.push(("IDENTITY_CRITICAL_AUTH_KEY_WIF".to_string(), wif));
+        identity_creds.push(("IDENTITY_CRITICAL_AUTH_KEY_WIF".to_string(), wif));
     }
-    append_creds(&args.out_env, &pre_creds)?;
-    eprintln!(
-        "Recovery credentials written to {} before broadcast.",
-        args.out_env
-    );
 
-    // --- Broadcast ---
-    let broadcast_txid = core
-        .send_raw_transaction(signed_hex.as_str())
-        .map_err(|e| format!("sendrawtransaction failed: {e}"))?;
-    if broadcast_txid != txid {
-        return Err(format!(
-            "broadcast txid {broadcast_txid} does not match signed txid {txid}"
-        ));
-    }
-    eprintln!("Broadcast asset-lock tx {txid}; waiting for ChainLock proof...");
+    // --- Obtain the asset lock: resume an existing one, or create+broadcast ---
+    let (txid, one_time_private_key): (Txid, PrivateKey) = if let Some(resume) = &args.resume_txid {
+        // RESUME: reuse an already-broadcast asset lock — NO new funds spent.
+        // The one-time key that owns the lock is read from the credentials
+        // file written before the original broadcast; the freshly generated
+        // identity keys are recorded now so the resumed identity is
+        // recoverable too.
+        let txid =
+            Txid::from_str(resume.trim()).map_err(|e| format!("invalid --resume-txid: {e}"))?;
+        let wif = read_creds_value(&args.out_env, "ASSET_LOCK_ONE_TIME_WIF")?.ok_or_else(|| {
+            format!(
+                "resume needs ASSET_LOCK_ONE_TIME_WIF in {} (the one-time key that owns the asset lock)",
+                args.out_env
+            )
+        })?;
+        let one_time_private_key = PrivateKey::from_wif(wif.trim())
+            .map_err(|e| format!("failed to parse ASSET_LOCK_ONE_TIME_WIF: {e}"))?;
+        append_creds(&args.out_env, &identity_creds)?;
+        eprintln!(
+            "Resuming existing asset-lock tx {txid} (no new broadcast; no new funds spent)..."
+        );
+        (txid, one_time_private_key)
+    } else {
+        // FRESH: generate a one-time key, build + Core-sign + broadcast the
+        // asset lock, capturing all recovery credentials BEFORE broadcast.
+        let secp = Secp256k1::new();
+        let mut secp_rng = dpp::dashcore::secp256k1::rand::thread_rng();
+        let one_time_secret = dpp::dashcore::secp256k1::SecretKey::new(&mut secp_rng);
+        let one_time_private_key = PrivateKey::new(one_time_secret, network);
+        let one_time_public_key = one_time_private_key.public_key(&secp);
+        let one_time_key_hash = one_time_public_key.pubkey_hash();
+        let one_time_address = Address::p2pkh(&one_time_public_key, network);
+
+        let tx = build_asset_lock_transaction(&core, amount_duffs, &one_time_key_hash)?;
+        let unsigned_hex = hex::encode(serialize(&tx));
+
+        eprintln!(
+            "Signing asset-lock tx ({} inputs, locking {} DASH) via Core wallet...",
+            tx.input.len(),
+            args.fund_dash
+        );
+        let signed = core
+            .sign_raw_transaction_with_wallet(unsigned_hex.as_str(), None, None)
+            .map_err(|e| format!("signrawtransactionwithwallet failed: {e}"))?;
+        if !signed.complete {
+            return Err(
+                "Core could not fully sign the asset-lock tx (signrawtransactionwithwallet \
+                 returned incomplete — is the faucet wallet loaded and are the inputs spendable?)"
+                    .to_string(),
+            );
+        }
+        let signed_hex = hex::encode(&signed.hex);
+        let signed_tx: Transaction =
+            deserialize(&signed.hex).map_err(|e| format!("failed to parse signed tx: {e}"))?;
+        let txid = signed_tx.txid();
+
+        // Capture recovery credentials BEFORE broadcast: once on-chain, the
+        // locked funds are recoverable only via the one-time key + outpoint.
+        let mut pre_creds: Vec<(String, String)> = vec![
+            ("ASSET_LOCK_TXID".to_string(), txid.to_string()),
+            (
+                "ASSET_LOCK_ONE_TIME_WIF".to_string(),
+                one_time_private_key.to_wif(),
+            ),
+            (
+                "ASSET_LOCK_ONE_TIME_ADDRESS".to_string(),
+                one_time_address.to_string(),
+            ),
+            (
+                "ASSET_LOCK_FUND_DASH".to_string(),
+                args.fund_dash.to_string(),
+            ),
+        ];
+        pre_creds.extend(identity_creds.iter().cloned());
+        append_creds(&args.out_env, &pre_creds)?;
+        eprintln!(
+            "Recovery credentials written to {} before broadcast.",
+            args.out_env
+        );
+
+        let broadcast_txid = core
+            .send_raw_transaction(signed_hex.as_str())
+            .map_err(|e| format!("sendrawtransaction failed: {e}"))?;
+        if broadcast_txid != txid {
+            return Err(format!(
+                "broadcast txid {broadcast_txid} does not match signed txid {txid}"
+            ));
+        }
+        eprintln!("Broadcast asset-lock tx {txid}; waiting for ChainLock proof...");
+        (txid, one_time_private_key)
+    };
 
     // --- ChainLock asset-lock proof (no DAPI stream) ---
     let started = Instant::now();
     // 1. Wait for Core to report the tx as ChainLocked and give its height.
     let core_chain_locked_height: u32 = loop {
         if started.elapsed() > PROOF_TIMEOUT {
-            return Err(
-                "timed out waiting for the asset-lock tx to be ChainLocked (the ChainLock proof \
-                 path did not complete — do not retry on this path)"
-                    .to_string(),
-            );
+            return Err(strand_safe_timeout(
+                &txid,
+                &args.out_env,
+                "the asset-lock tx to be ChainLocked",
+            ));
         }
         let info = core
             .get_raw_transaction_info(&txid, None)
@@ -379,11 +460,11 @@ async fn run() -> Result<(), String> {
     //    drive will accept the proof.
     loop {
         if started.elapsed() > PROOF_TIMEOUT {
-            return Err(
-                "timed out waiting for platform to reach the ChainLocked core height (do not retry \
-                 on this path)"
-                    .to_string(),
-            );
+            return Err(strand_safe_timeout(
+                &txid,
+                &args.out_env,
+                "platform to reach the ChainLocked core height",
+            ));
         }
         let (_epoch, metadata) = Epoch::fetch_current_with_metadata(&sdk)
             .await
--- a/packages/rs-scripts/src/bin/register_identity.rs
+++ b/packages/rs-scripts/src/bin/register_identity.rs
@@ -633,3 +633,93 @@ fn find_critical_auth_key(
         .find(|(pk, _)| pk.id() == public_key.id())
         .map(|(pk, secret)| (pk.clone(), *secret))
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn parse_network_accepts_known_and_rejects_unknown() {
+        assert!(matches!(parse_network("devnet"), Ok(Network::Devnet)));
+        assert!(matches!(parse_network("DevNet"), Ok(Network::Devnet)));
+        assert!(matches!(parse_network("testnet"), Ok(Network::Testnet)));
+        assert!(matches!(parse_network("mainnet"), Ok(Network::Mainnet)));
+        assert!(matches!(parse_network("regtest"), Ok(Network::Regtest)));
+        assert!(parse_network("bogus").is_err());
+    }
+
+    #[test]
+    fn creds_roundtrip_returns_last_value_and_file_is_mode_600() {
+        // A resume run appends fresh identity keys after an earlier attempt's,
+        // so read_creds_value MUST return the LAST value — otherwise the caller
+        // would sign with a stale key that does not match the live identity.
+        use std::os::unix::fs::PermissionsExt;
+        let path =
+            std::env::temp_dir().join(format!("register_identity_test_{}.env", std::process::id()));
+        let p = path.to_str().unwrap();
+        let _ = std::fs::remove_file(p);
+
+        append_creds(p, &[("K".to_string(), "first".to_string())]).unwrap();
+        append_creds(
+            p,
+            &[
+                ("OTHER".to_string(), "x".to_string()),
+                ("K".to_string(), "second".to_string()),
+            ],
+        )
+        .unwrap();
+
+        assert_eq!(read_creds_value(p, "K").unwrap().as_deref(), Some("second"));
+        assert_eq!(read_creds_value(p, "OTHER").unwrap().as_deref(), Some("x"));
+        assert_eq!(read_creds_value(p, "MISSING").unwrap(), None);
+
+        let mode = std::fs::metadata(p).unwrap().permissions().mode() & 0o777;
+        assert_eq!(mode, 0o600, "credentials file must be created mode 600");
+        let _ = std::fs::remove_file(p);
+    }
+
+    #[test]
+    fn strand_message_is_non_stranding_and_actionable() {
+        let txid =
+            Txid::from_str("c913da3655688c10c79e0d8b8e059c94625b939cfa99e848f0f24dc48ec4f685")
+                .unwrap();
+        let msg = strand_safe_timeout(&txid, "/tmp/creds.env", "the tx to be ChainLocked");
+        // Must name the txid, the creds path, and the exact resume flag, and must
+        // reassure the funds are not lost — that is the whole point of the change.
+        assert!(msg.contains("c913da3655688c10c79e0d8b8e059c94625b939cfa99e848f0f24dc48ec4f685"));
+        assert!(msg.contains("/tmp/creds.env"));
+        assert!(msg.contains("--resume-txid"));
+        assert!(msg.to_lowercase().contains("not lost"));
+    }
+
+    #[test]
+    fn finds_the_critical_authentication_signing_key_and_its_secret() {
+        // The whole tool is useless if it cannot hand back a CRITICAL-level
+        // authentication key (the one that can sign both contracts and
+        // documents) together with the private key that matches it.
+        let platform_version = PlatformVersion::latest();
+        let mut rng = StdRng::seed_from_u64(1);
+        let (identity, key_material): (Identity, Vec<(IdentityPublicKey, [u8; 32])>) =
+            Identity::random_identity_with_main_keys_with_private_key(
+                3,
+                &mut rng,
+                platform_version,
+            )
+            .expect("generate identity");
+
+        let (pk, secret) = find_critical_auth_key(&identity, &key_material)
+            .expect("a 3-key identity must expose a CRITICAL AUTHENTICATION ECDSA key");
+
+        assert_eq!(pk.purpose(), Purpose::AUTHENTICATION);
+        assert_eq!(pk.security_level(), SecurityLevel::CRITICAL);
+        assert_eq!(pk.key_type(), KeyType::ECDSA_SECP256K1);
+
+        // The returned secret is really THIS key's material, not some other key's.
+        let expected = key_material
+            .iter()
+            .find(|(k, _)| k.id() == pk.id())
+            .map(|(_, s)| *s)
+            .unwrap();
+        assert_eq!(secret, expected);
+    }
+}
--- a/packages/rs-scripts/src/bin/register_identity.rs
+++ b/packages/rs-scripts/src/bin/register_identity.rs
@@ -86,6 +86,7 @@ use dash_sdk::platform::fetch_current_no_parameters::FetchCurrent;
 use dash_sdk::platform::transition::put_identity::PutIdentity;
 use dash_sdk::platform::types::epoch::Epoch;
 use dash_sdk::{Sdk, SdkBuilder};
+use dpp::balances::credits::CREDITS_PER_DUFF;
 use dpp::dashcore::consensus::encode::{deserialize, serialize};
 use dpp::dashcore::secp256k1::Secp256k1;
 use dpp::dashcore::transaction::special_transaction::asset_lock::AssetLockPayload;
@@ -194,12 +195,19 @@ fn parse_network(s: &str) -> Result<Network, String> {
 /// Append `KEY=value` lines to the mode-600 credentials file. Creates
 /// the file if absent; never truncates.
 fn append_creds(path: &str, lines: &[(String, String)]) -> Result<(), String> {
+    use std::os::unix::fs::PermissionsExt;
     let mut f = std::fs::OpenOptions::new()
         .create(true)
         .append(true)
         .mode(0o600)
         .open(path)
         .map_err(|e| format!("failed to open creds file {path}: {e}"))?;
+    // `.mode(0o600)` only applies to a freshly created inode. If the file
+    // already existed with looser permissions, tighten it to 0600 BEFORE
+    // writing any secret, so a pre-existing world-readable file cannot
+    // silently keep exposing the keys we are about to append.
+    f.set_permissions(std::fs::Permissions::from_mode(0o600))
+        .map_err(|e| format!("failed to enforce mode 600 on {path}: {e}"))?;
     for (k, v) in lines {
         writeln!(f, "{k}={v}").map_err(|e| format!("failed to write creds file: {e}"))?;
     }
@@ -219,15 +227,60 @@ fn read_creds_value(path: &str, key: &str) -> Result<Option<String>, String> {
         .map(str::to_string))
 }
 
-/// Message for a proof-wait timeout that makes the failure NON-stranding:
-/// the asset lock is already broadcast and on-chain, so the funds are not
-/// lost — they are pending. It names the recoverable one-time key and the
-/// exact command to resume without spending new funds.
-fn strand_safe_timeout(txid: &Txid, out_env: &str, waiting_for: &str) -> String {
+/// Minimum asset-lock funding, in duffs, that an identity-create with
+/// `key_count` keys needs to clear its required-balance validation. Mirrors
+/// `IdentityCreateTransition::calculate_min_required_fee` (versioned): the
+/// asset-lock processing-start floor plus, on fee-calc v1+, the base
+/// identity-create cost and the per-key creation cost. Locking below this
+/// broadcasts and mines a transaction whose identity-create then fails —
+/// stranding the funds, the exact failure this tool exists to avoid.
+fn min_asset_lock_duffs(key_count: u32, pv: &PlatformVersion) -> u64 {
+    let identities = &pv.dpp.state_transitions.identities;
+    let floor_credits = identities
+        .asset_locks
+        .required_asset_lock_duff_balance_for_processing_start_for_identity_create
+        .saturating_mul(CREDITS_PER_DUFF);
+    let required_credits = match identities.calculate_min_required_fee_on_identity_create_transition
+    {
+        0 => floor_credits,
+        _ => {
+            let min_fees = &pv.fee_version.state_transition_min_fees;
+            min_fees.identity_create_base_cost.saturating_add(
+                floor_credits.saturating_add(
+                    min_fees
+                        .identity_key_in_creation_cost
+                        .saturating_mul(key_count as u64),
+                ),
+            )
+        }
+    };
+    // Convert the required credits back to the minimum asset-lock duffs
+    // (ceil, so rounding never lands us a duff short).
+    required_credits.div_ceil(CREDITS_PER_DUFF)
+}
+
+/// Message for a proof-wait timeout that makes the failure NON-stranding.
+/// The funds are recoverable via the saved one-time key either way, but the
+/// finality claim must match what Core actually reported: `chainlocked=false`
+/// means the tx may still be unconfirmed/in the mempool (verify in Core
+/// first), while `chainlocked=true` means only Platform's catch-up timed out.
+fn strand_safe_timeout(txid: &Txid, out_env: &str, chainlocked: bool) -> String {
+    let state = if chainlocked {
+        "The asset lock IS broadcast and ChainLocked on L1 — only Platform's catch-up to that \
+         core height timed out. The funds are NOT lost; re-running will finish once Platform \
+         has caught up."
+            .to_string()
+    } else {
+        format!(
+            "The asset lock was broadcast but Core has NOT yet reported it as ChainLocked — it may \
+             still be unconfirmed or in the mempool. The funds are NOT lost. First verify in Core \
+             (`getrawtransaction {txid} 1` → chainlock:true); resume only once it is chainlocked."
+        )
+    };
     format!(
-        "timed out waiting for {waiting_for}.\n\
-         The asset lock IS broadcast and on-chain — the funds are NOT lost, only pending.\n\
-         The one-time key that owns it is saved in {out_env} (ASSET_LOCK_ONE_TIME_WIF).\n\
+        "timed out waiting for the ChainLock proof.\n\
+         {state}\n\
+         The one-time key that owns the lock is saved in {out_env} (ASSET_LOCK_ONE_TIME_WIF__{txid}).\n\
          Resume WITHOUT spending new funds by re-running the same command plus:\n\
          \x20   --resume-txid {txid}\n\
          Do NOT re-run without --resume-txid — that broadcasts a second asset lock and strands this one."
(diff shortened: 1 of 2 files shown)


thepastaclaw:
⛔ Blockers found — Opus deferred (commit a03f1e29)

packages/rs-scripts/src/bin/register_identity.rs
```diff
@@ -0,0 +1,554 @@
...
+
+/// Append `KEY=value` lines to the mode-600 credentials file. Creates
+/// the file if absent; never truncates.
+fn append_creds(path: &str, lines: &[(String, String)]) -> Result<(), String> {
+    let mut f = std::fs::OpenOptions::new()
+        .create(true)
+        .append(true)
+        .mode(0o600)
+        .open(path)
+        .map_err(|e| format!("failed to open creds file {path}: {e}"))?;
```
thepastaclaw, line 248: **🔴 Blocking: Enforce mode 600 when the credentials file already exists**

`OpenOptionsExt::mode(0o600)` only affects a newly created inode. If `--out-env` already exists with mode 0644, this function appends the one-time asset-lock WIF and every identity private key without removing group or world access, violating the command's stated secret-storage boundary. Set the opened file's permissions before writing any new secrets.

```suggestion
fn append_creds(path: &str, lines: &[(String, String)]) -> Result<(), String> {
    use std::os::unix::fs::PermissionsExt;

    let mut f = std::fs::OpenOptions::new()
        .create(true)
        .append(true)
        .mode(0o600)
        .open(path)
        .map_err(|e| format!("failed to open creds file {path}: {e}"))?;
    f.set_permissions(std::fs::Permissions::from_mode(0o600))
        .map_err(|e| format!("failed to set mode 0600 on creds file {path}: {e}"))?;
    for (k, v) in lines {
        writeln!(f, "{k}={v}").map_err(|e| format!("failed to write creds file: {e}"))?;
    }
    Ok(())
}
```

packages/rs-scripts/src/bin/register_identity.rs
```diff
@@ -0,0 +1,554 @@
...
+    let mut secp_rng = dpp::dashcore::secp256k1::rand::thread_rng();
+    let one_time_secret = dpp::dashcore::secp256k1::SecretKey::new(&mut secp_rng);
+    let one_time_private_key = PrivateKey::new(one_time_secret, network);
+    let one_time_public_key = one_time_private_key.public_key(&secp);
+    let one_time_key_hash = one_time_public_key.pubkey_hash();
+    let one_time_address = Address::p2pkh(&one_time_public_key, network);
+
+    // --- Build the asset-lock funding transaction ---
+    let tx = build_asset_lock_transaction(&core, amount_duffs, &one_time_key_hash)?;
+    let unsigned_hex = hex::encode(serialize(&tx));
```
thepastaclaw: **🔴 Blocking: Reject funding transactions above Platform's asset-lock input cap**

Coin selection has no input limit. On the latest platform version used by this command, `max_asset_lock_transaction_inputs` is 100, and `validate_asset_lock_transaction_structure_v0` rejects a transaction with 101 or more inputs. Dash Core can sign, broadcast, and mine such a transaction, but every Platform use of its asset-lock proof remains invalid, making the locked output unusable. Validate the versioned cap before signing or broadcasting.

```suggestion
    // --- Build the asset-lock funding transaction ---
    let tx = build_asset_lock_transaction(&core, amount_duffs, &one_time_key_hash)?;
    let max_asset_lock_inputs = platform_version
        .dpp
        .state_transitions
        .identities
        .asset_locks
        .max_asset_lock_transaction_inputs as usize;
    if tx.input.len() > max_asset_lock_inputs {
        return Err(format!(
            "asset-lock transaction requires {} inputs, but Platform accepts at most {}; consolidate the faucet wallet's UTXOs first",
            tx.input.len(), max_asset_lock_inputs
        ));
    }
    let unsigned_hex = hex::encode(serialize(&tx));
```

<sub>source: ['codex']</sub>

packages/rs-scripts/src/bin/register_identity.rs
```diff
@@ -0,0 +1,554 @@
...
+        .or(args.core_rpc_password.clone())
+        .ok_or_else(|| {
+            "Core RPC password required: set CORE_RPC_PASSWORD or pass --core-rpc-password"
+                .to_string()
+        })?;
+
+    let amount_duffs = (args.fund_dash * DUFFS_PER_DASH as f64) as u64;
+    if amount_duffs == 0 {
+        return Err("--fund-dash must be > 0".to_string());
+    }
```
thepastaclaw, line 384: **🔴 Blocking: Validate the identity-create minimum before locking funds**

The CLI accepts any amount that truncates to at least one duff, but the latest identity-create fee calculation requires the 200,000-duff asset-lock floor plus 2,000 duffs of base cost and 6,500 duffs per key. The default three-key identity therefore requires at least 221,500 duffs. A smaller asset lock is broadcast and mined before identity registration fails its required-balance validation; very small locks also fall below the alternative top-up and address-funding floors. Compute the versioned minimum for the requested key count before constructing the transaction. Parse `--fund-dash` with Dash Core's exact decimal amount parser as well, because multiplying an `f64` and casting to `u64` silently truncates and can misrepresent decimal input.

<sub>source: ['codex']</sub>

packages/rs-scripts/src/bin/register_identity.rs
```diff
@@ -0,0 +1,554 @@
...
+}
+
+async fn run() -> Result<(), String> {
+    let args = Args::parse();
+    let network = parse_network(&args.network)?;
+    let platform_version = PlatformVersion::latest();
+
+    if args.key_count < 3 {
+        return Err("--key-count must be >= 3 (need MASTER + CRITICAL + HIGH keys)".to_string());
+    }
```
thepastaclaw, line 360: **🔴 Blocking: Bound key count by Platform's creation limit**

Only the lower bound is checked, while the latest platform version permits at most six public keys in an identity-create transition. The generator can produce more than six keys, so the command signs and broadcasts the irreversible funding transaction before the SDK's structural validation rejects the identity transition. Validate the versioned upper bound before generating keys or performing any funding work.

```suggestion
    let platform_version = PlatformVersion::latest();
    let max_key_count = platform_version
        .dpp
        .state_transitions
        .identities
        .max_public_keys_in_creation as u32;

    if !(3..=max_key_count).contains(&args.key_count) {
        return Err(format!(
            "--key-count must be between 3 and {max_key_count}"
        ));
    }
```

<sub>source: ['codex']</sub>

packages/rs-scripts/src/bin/register_identity.rs
```diff
@@ -0,0 +1,554 @@
...
+
+    let core_password = std::env::var("CORE_RPC_PASSWORD")
+        .ok()
+        .or(args.core_rpc_password.clone())
+        .ok_or_else(|| {
+            "Core RPC password required: set CORE_RPC_PASSWORD or pass --core-rpc-password"
+                .to_string()
+        })?;
```
thepastaclaw, line 379: **🟡 Suggestion: Treat an empty CORE_RPC_PASSWORD as unset**

`std::env::var("CORE_RPC_PASSWORD").ok()` produces `Some("")` for an exported but empty variable. That value takes precedence over a valid `--core-rpc-password`, causing an authentication failure instead of using the documented fallback. Discard empty environment values before resolving the CLI option.

```suggestion
    let core_password = std::env::var("CORE_RPC_PASSWORD")
        .ok()
        .filter(|password| !password.is_empty())
        .or(args.core_rpc_password.clone())
        .ok_or_else(|| {
            "Core RPC password required: set CORE_RPC_PASSWORD or pass --core-rpc-password"
                .to_string()
        })?;
```

packages/rs-scripts/src/bin/register_identity.rs
```diff
@@ -0,0 +1,554 @@
...
+        if started.elapsed() > PROOF_TIMEOUT {
+            return Err(
+                "timed out waiting for platform to reach the ChainLocked core height (do not retry \
+                 on this path)"
+                    .to_string(),
+            );
+        }
+        let (_epoch, metadata) = Epoch::fetch_current_with_metadata(&sdk)
+            .await
+            .map_err(|e| format!("failed to fetch platform metadata: {e}"))?;
```
thepastaclaw: **🟡 Suggestion: Retry transient polling errors after the asset lock is broadcast**

Both polling loops immediately return on a Core RPC or DAPI metadata-fetch error, even though the asset lock has already been broadcast and the loops otherwise retry until `PROOF_TIMEOUT`. A brief transport interruption therefore forces the operator into manual recovery using the saved credentials. Treat polling request failures as retryable, log them without secrets, sleep, and let the timeout determine when the command gives up. Recovery data makes this failure non-destructive, so this is a robustness suggestion rather than a blocker.

thepastaclaw (changes requested):

The ChainLock workflow correctly saves recovery material before broadcast, but four accepted states can expose private keys or leave the command unable to consume an irreversible L1 asset lock: an existing permissive credentials file, more than 100 transaction inputs, insufficient funding, or more than six identity keys. Two smaller configuration and polling issues also cause avoidable command failures.
Source: Reviewer backend model: gpt-5.6-sol. Final verifier backend model: gpt-5.6-sol. Orchestration-only: openclaw-agent/cliproxy/gpt-5.6-sol (not reviewer evidence).

### Review provenance

<details>

```

In `packages/rs-scripts/src/bin/register_identity.rs`:
- [BLOCKING] packages/rs-scripts/src/bin/register_identity.rs:172-178: Enforce mode 600 when the credentials file already exists
  `OpenOptionsExt::mode(0o600)` only affects a newly created inode. If `--out-env` already exists with mode 0644, this function appends the one-time asset-lock WIF and every identity private key without removing group or world access, violating the command's stated secret-storage boundary. Set the opened file's permissions before writing any new secrets.
- [BLOCKING] packages/rs-scripts/src/bin/register_identity.rs:274-276: Reject funding transactions above Platform's asset-lock input cap
  Coin selection has no input limit. On the latest platform version used by this command, `max_asset_lock_transaction_inputs` is 100, and `validate_asset_lock_transaction_structure_v0` rejects a transaction with 101 or more inputs. Dash Core can sign, broadcast, and mine such a transaction, but every Platform use of its asset-lock proof remains invalid, making the locked output unusable. Validate the versioned cap before signing or broadcasting.
- [BLOCKING] packages/rs-scripts/src/bin/register_identity.rs:213-216: Validate the identity-create minimum before locking funds
  The CLI accepts any amount that truncates to at least one duff, but the latest identity-create fee calculation requires the 200,000-duff asset-lock floor plus 2,000 duffs of base cost and 6,500 duffs per key. The default three-key identity therefore requires at least 221,500 duffs. A smaller asset lock is broadcast and mined before identity registration fails its required-balance validation; very small locks also fall below the alternative top-up and address-funding floors. Compute the versioned minimum for the requested key count before constructing the transaction. Parse `--fund-dash` with Dash Core's exact decimal amount parser as well, because multiplying an `f64` and casting to `u64` silently truncates and can misrepresent decimal input.
- [BLOCKING] packages/rs-scripts/src/bin/register_identity.rs:199-203: Bound key count by Platform's creation limit
  Only the lower bound is checked, while the latest platform version permits at most six public keys in an identity-create transition. The generator can produce more than six keys, so the command signs and broadcasts the irreversible funding transaction before the SDK's structural validation rejects the identity transition. Validate the versioned upper bound before generating keys or performing any funding work.
- [SUGGESTION] packages/rs-scripts/src/bin/register_identity.rs:205-211: Treat an empty CORE_RPC_PASSWORD as unset
  `std::env::var("CORE_RPC_PASSWORD").ok()` produces `Some("")` for an exported but empty variable. That value takes precedence over a valid `--core-rpc-password`, causing an authentication failure instead of using the documented fallback. Discard empty environment values before resolving the CLI option.
- [SUGGESTION] packages/rs-scripts/src/bin/register_identity.rs:354-390: Retry transient polling errors after the asset lock is broadcast
  Both polling loops immediately return on a Core RPC or DAPI metadata-fetch error, even though the asset lock has already been broadcast and the loops otherwise retry until `PROOF_TIMEOUT`. A brief transport interruption therefore forces the operator into manual recovery using the saved credentials. Treat polling request failures as retryable, log them without secrets, sleep, and let the timeout determine when the command gives up. Recovery data makes this failure non-destructive, so this is a robustness suggestion rather than a blocker.
```
</details>

packages/rs-scripts/src/bin/register_identity.rs
```diff
@@ -0,0 +1,635 @@
...
+            Txid::from_str(resume.trim()).map_err(|e| format!("invalid --resume-txid: {e}"))?;
+        let wif = read_creds_value(&args.out_env, "ASSET_LOCK_ONE_TIME_WIF")?.ok_or_else(|| {
+            format!(
+                "resume needs ASSET_LOCK_ONE_TIME_WIF in {} (the one-time key that owns the asset lock)",
+                args.out_env
+            )
+        })?;
+        let one_time_private_key = PrivateKey::from_wif(wif.trim())
+            .map_err(|e| format!("failed to parse ASSET_LOCK_ONE_TIME_WIF: {e}"))?;
+        append_creds(&args.out_env, &identity_creds)?;
```
thepastaclaw: **🔴 Blocking: Scope recovery credentials to one asset-lock transaction**

The append-only credentials file can contain multiple asset-lock records, but resume parses the requested txid and independently selects the last global `ASSET_LOCK_ONE_TIME_WIF`. Resuming an older lock can therefore sign with an unrelated one-time key. Resume also generates and appends new `IDENTITY_KEY_<id>_WIF` values before determining whether the proof-derived identity already exists. Because the SDK resolves an `AlreadyExists` response by fetching that identity, the command can report an existing identity after the last-wins environment values have been replaced with keys that do not control it; other consumed-lock errors leave the credentials overwritten as well. Store complete recovery records under their asset-lock txid, select the matching one-time WIF on resume, and reload the original identity key metadata rather than generating and appending a new key set.

packages/rs-scripts/src/bin/register_identity.rs
```diff
@@ -0,0 +1,635 @@
...
+/// lost — they are pending. It names the recoverable one-time key and the
+/// exact command to resume without spending new funds.
+fn strand_safe_timeout(txid: &Txid, out_env: &str, waiting_for: &str) -> String {
+    format!(
+        "timed out waiting for {waiting_for}.\n\
+         The asset lock IS broadcast and on-chain — the funds are NOT lost, only pending.\n\
+         The one-time key that owns it is saved in {out_env} (ASSET_LOCK_ONE_TIME_WIF).\n\
+         Resume WITHOUT spending new funds by re-running the same command plus:\n\
+         \x20   --resume-txid {txid}\n\
+         Do NOT re-run without --resume-txid — that broadcasts a second asset lock and strands this one."
```
thepastaclaw: **🟡 Suggestion: Make the timeout message reflect ChainLock state**

`strand_safe_timeout` unconditionally states that the asset lock is on-chain, but it is also called when Core has never reported the transaction as ChainLocked. At that point the transaction may still be in the mempool, so the message claims a finality state the command has not established. Use separate guidance for the pre-ChainLock timeout and the post-ChainLock Platform synchronization timeout.

thepastaclaw (changes requested):

The ChainLock workflow is a useful replacement for the broken InstantSend path, but the current implementation can still expose credentials, irreversibly fund asset locks that Platform cannot consume, and associate resume attempts with the wrong recovery keys. The new resume flow must also preserve the original identity credentials, while the timeout and polling behavior need smaller operational fixes.
Source: Reviewer backend model: gpt-5.6-sol. Final verifier backend model: gpt-5.6-sol. Orchestration-only: openclaw-agent/cliproxy/gpt-5.6-sol (not reviewer evidence).

### Review provenance

🔴 5 blocking | 🟡 1 suggestion(s)

_6 carried-forward finding(s) already raised on this PR; not re-posting as new inline comments._

<details>

```

In `packages/rs-scripts/src/bin/register_identity.rs`:
- [BLOCKING] packages/rs-scripts/src/bin/register_identity.rs:341-357: Scope recovery credentials to one asset-lock transaction
  The append-only credentials file can contain multiple asset-lock records, but resume parses the requested txid and independently selects the last global `ASSET_LOCK_ONE_TIME_WIF`. Resuming an older lock can therefore sign with an unrelated one-time key. Resume also generates and appends new `IDENTITY_KEY_<id>_WIF` values before determining whether the proof-derived identity already exists. Because the SDK resolves an `AlreadyExists` response by fetching that identity, the command can report an existing identity after the last-wins environment values have been replaced with keys that do not control it; other consumed-lock errors leave the credentials overwritten as well. Store complete recovery records under their asset-lock txid, select the matching one-time WIF on resume, and reload the original identity key metadata rather than generating and appending a new key set.
- [SUGGESTION] packages/rs-scripts/src/bin/register_identity.rs:226-233: Make the timeout message reflect ChainLock state
  `strand_safe_timeout` unconditionally states that the asset lock is on-chain, but it is also called when Core has never reported the transaction as ChainLocked. At that point the transaction may still be in the mempool, so the message claims a finality state the command has not established. Use separate guidance for the pre-ChainLock timeout and the post-ChainLock Platform synchronization timeout.
- [BLOCKING] packages/rs-scripts/src/bin/register_identity.rs:197-202: Enforce mode 600 when the credentials file already exists
  (existing thread: https://github.com/dashpay/platform/pull/4399#discussion_r3777361948)
  `OpenOptionsExt::mode(0o600)` only affects a newly created inode. If `--out-env` already exists with mode 0644, this function appends the one-time asset-lock WIF and every identity private key without removing group or world access, despite the command promising a mode-600 credentials file. Explicitly set the opened file's permissions to 0600 before writing any secrets and fail without writing if that operation fails.
- [BLOCKING] packages/rs-scripts/src/bin/register_identity.rs:373-374: Reject funding transactions above Platform's asset-lock input cap
  (existing thread: https://github.com/dashpay/platform/pull/4399#discussion_r3777361965)
  `build_asset_lock_transaction` can select an arbitrary number of UTXOs, and the result is sent for signing without checking its input count. The latest platform version used here sets `max_asset_lock_transaction_inputs` to 100, and `validate_asset_lock_transaction_structure_v0` rejects larger transactions. Core can therefore sign, broadcast, and mine an asset lock with 101 or more inputs that Platform will never accept. Compare `tx.input.len()` with the versioned cap before signing, saving credentials, or broadcasting.
- [BLOCKING] packages/rs-scripts/src/bin/register_identity.rs:265-268: Validate the identity-create minimum before locking funds
  (existing thread: https://github.com/dashpay/platform/pull/4399#discussion_r3777361973)
  The fresh-funding path accepts any amount that truncates to at least one duff. In the latest version, identity creation requires the 200,000-duff asset-lock floor plus 2,000 duffs of base cost and 6,500 duffs per key, so the default three-key identity requires at least 221,500 duffs. A smaller lock can be broadcast and mined before registration fails its required-balance validation. Compute the versioned minimum for the selected key count before constructing a fresh transaction, and parse `--fund-dash` using Dash Core's exact decimal amount parser instead of multiplying and truncating an `f64`.
- [BLOCKING] packages/rs-scripts/src/bin/register_identity.rs:251-255: Bound key count by Platform's creation limit
  (existing thread: https://github.com/dashpay/platform/pull/4399#discussion_r3777361979)
  Only the lower bound is enforced, while the latest platform version permits at most six public keys in an identity-create transition. A larger value can generate a transition that Platform rejects only after the irreversible asset-lock transaction has been broadcast. Read `max_public_keys_in_creation` from `platform_version` and reject larger values before generating keys or doing any funding work.
- [SUGGESTION] packages/rs-scripts/src/bin/register_identity.rs:257-263: Treat an empty CORE_RPC_PASSWORD as unset
  (existing thread: https://github.com/dashpay/platform/pull/4399#discussion_r3777361985)
  `std::env::var("CORE_RPC_PASSWORD").ok()` returns `Some("")` when the variable is exported but empty. That empty value takes precedence over a valid `--core-rpc-password`, causing an avoidable authentication failure. Filter empty environment values before applying the CLI fallback.
- [SUGGESTION] packages/rs-scripts/src/bin/register_identity.rs:443-471: Retry transient polling errors after the asset lock is broadcast
  (existing thread: https://github.com/dashpay/platform/pull/4399#discussion_r3777361990)
  Both polling loops still return immediately on a single Core RPC or DAPI metadata-fetch error, even though they otherwise retry until `PROOF_TIMEOUT` after the asset lock has been broadcast. A temporary transport failure therefore aborts the automatic workflow and forces an unnecessary resume attempt. Treat request failures as retryable within the timeout, log a non-secret diagnostic, sleep, and continue polling.
```
</details>

packages/rs-scripts/src/bin/register_identity.rs
thepastaclaw: Resolved in `e8d2b14` — _Reject funding transactions above Platform's asset-lock input cap_ no longer present.

packages/rs-scripts/src/bin/register_identity.rs
thepastaclaw: Resolved in `e8d2b14` — _Retry transient polling errors after the asset lock is broadcast_ no longer present.

packages/rs-scripts/src/bin/register_identity.rs
thepastaclaw: Resolved in `e8d2b14` — _Validate the identity-create minimum before locking funds_ no longer present.

packages/rs-scripts/src/bin/register_identity.rs
```diff
@@ -0,0 +1,725 @@
...
+        pre_creds.extend(identity_creds.iter().cloned());
+        append_creds(&args.out_env, &pre_creds)?;
+        eprintln!(
+            "Recovery credentials written to {} before broadcast.",
+            args.out_env
+        );
+
+        let broadcast_txid = core
+            .send_raw_transaction(signed_hex.as_str())
+            .map_err(|e| format!("sendrawtransaction failed: {e}"))?;
```
thepastaclaw: **🟡 Suggestion: Provide recovery guidance for an ambiguous broadcast failure**

`sendrawtransaction` can return a transport error after Core has already accepted the transaction. This path exits with only `sendrawtransaction failed`, even though the expected txid and recovery credentials were saved and rerunning normally could create a second asset lock. On an RPC error, explain that acceptance is ambiguous and direct the operator to check the saved txid in Core and use `--resume-txid` if it exists; alternatively query the expected txid before deciding which error to return.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

This preliminary Codex-only revalidation confirms that all eight previously verified issues remain at the exact head; the new tests exercise only successful helper paths and do not fix the underlying behavior. Five blockers can still expose private keys, create asset locks that Platform cannot consume, or associate resumed registrations with the wrong identity credentials; four additional operational issues can cause avoidable failures or misleading recovery guidance. Opus review is deferred because the blocker gate remains open.
Source: Reviewer backend model: gpt-5.6-sol. Final verifier backend model: gpt-5.6-sol. Orchestration-only: openclaw-agent/cliproxy/gpt-5.6-sol (not reviewer evidence).

### Review provenance

🔴 5 blocking | 🟡 1 suggestion(s)

_8 carried-forward finding(s) already raised on this PR; not re-posting as new inline comments._

<details>

```

In `packages/rs-scripts/src/bin/register_identity.rs`:
- [SUGGESTION] packages/rs-scripts/src/bin/register_identity.rs:420-422: Provide recovery guidance for an ambiguous broadcast failure
  `sendrawtransaction` can return a transport error after Core has already accepted the transaction. This path exits with only `sendrawtransaction failed`, even though the expected txid and recovery credentials were saved and rerunning normally could create a second asset lock. On an RPC error, explain that acceptance is ambiguous and direct the operator to check the saved txid in Core and use `--resume-txid` if it exists; alternatively query the expected txid before deciding which error to return.
- [BLOCKING] packages/rs-scripts/src/bin/register_identity.rs:197-204: Enforce mode 600 when the credentials file already exists
  (existing thread: https://github.com/dashpay/platform/pull/4399#discussion_r3777361948)
  `OpenOptionsExt::mode(0o600)` only controls the mode used when creating a new inode. If `--out-env` already exists with mode 0644, `append_creds` preserves that mode while appending the asset-lock WIF and all identity private keys, contrary to the command's documented security boundary. Explicitly set the opened file's permissions to 0600 before writing, and return without writing if that operation fails. The new test only verifies a file created by `append_creds`, so it does not cover an existing permissive file.
- [BLOCKING] packages/rs-scripts/src/bin/register_identity.rs:373-374: Reject funding transactions above Platform's asset-lock input cap
  (existing thread: https://github.com/dashpay/platform/pull/4399#discussion_r3777361965)
  `build_asset_lock_transaction` can select an arbitrary number of UTXOs, and the resulting transaction proceeds directly to Core signing without an input-count check. The latest platform version sets `max_asset_lock_transaction_inputs` to 100, and `validate_asset_lock_transaction_structure_v0` explicitly rejects transactions with more inputs. Core can therefore sign, broadcast, and mine a transaction with 101 or more inputs whose asset-lock proof Platform will never accept. Compare `tx.input.len()` with the versioned limit immediately after construction, before signing, persisting credentials, or broadcasting.
- [BLOCKING] packages/rs-scripts/src/bin/register_identity.rs:265-268: Validate the identity-create minimum before locking funds
  (existing thread: https://github.com/dashpay/platform/pull/4399#discussion_r3777361973)
  The fresh-funding path accepts any floating-point amount that truncates to at least one duff. In the latest version, identity creation requires the 200,000-duff asset-lock floor plus 2,000 duffs of base cost and 6,500 duffs per key, so the default three-key identity needs at least 221,500 duffs. A smaller transaction can be broadcast and mined before registration fails minimum-fee validation. Compute the versioned minimum for the selected key count before constructing a fresh transaction, and parse `--fund-dash` with Dash Core's exact decimal amount representation instead of multiplying and truncating an `f64`.
- [BLOCKING] packages/rs-scripts/src/bin/register_identity.rs:251-255: Bound key count by Platform's creation limit
  (existing thread: https://github.com/dashpay/platform/pull/4399#discussion_r3777361979)
  Only the lower bound is checked, while `PlatformVersion::latest()` permits at most six public keys in an identity-create transition. The random identity helper can generate more than six keys, so values such as seven reach the irreversible funding and broadcast path before Platform's basic-structure validation rejects the transition. Read `max_public_keys_in_creation` from `platform_version` and reject larger values before key generation or funding work.
- [BLOCKING] packages/rs-scripts/src/bin/register_identity.rs:341-357: Scope recovery credentials to one asset-lock transaction
  (existing thread: https://github.com/dashpay/platform/pull/4399#discussion_r3778738671)
  Resume parses the requested txid but independently reads the last global `ASSET_LOCK_ONE_TIME_WIF`; it never checks the saved `ASSET_LOCK_TXID`. Reusing an append-only credentials file can therefore pair an older transaction with a newer lock's private key. Resume also generates and appends a new identity key set before determining whether an earlier registration already succeeded. The SDK handles `AlreadyExists` by fetching and returning the existing proof-derived identity, so the command can report success while the last-wins `IDENTITY_KEY_*_WIF` entries no longer control that identity. Persist complete records scoped by asset-lock txid, select the matching one-time WIF, and reload the original identity public/private key set on resume.
- [SUGGESTION] packages/rs-scripts/src/bin/register_identity.rs:257-263: Treat an empty CORE_RPC_PASSWORD as unset
  (existing thread: https://github.com/dashpay/platform/pull/4399#discussion_r3777361985)
  `std::env::var("CORE_RPC_PASSWORD").ok()` returns `Some("")` when the variable exists but is empty. That empty value takes precedence over a valid `--core-rpc-password`, causing Core authentication to fail instead of using the documented fallback. Filter empty environment values before applying the CLI option.
- [SUGGESTION] packages/rs-scripts/src/bin/register_identity.rs:433-471: Retry transient polling errors after the asset lock is broadcast
  (existing thread: https://github.com/dashpay/platform/pull/4399#discussion_r3777361990)
  Both polling loops still propagate a single Core RPC or DAPI metadata-fetch failure immediately, even though the asset lock has already been broadcast and the loops otherwise retry until `PROOF_TIMEOUT`. A temporary transport interruption therefore aborts the automatic workflow and forces an unnecessary resume attempt. Treat polling request failures as retryable within the timeout, emit a non-secret diagnostic, sleep, and continue; retain immediate failure for deterministic malformed-success responses such as an invalid reported height.
- [SUGGESTION] packages/rs-scripts/src/bin/register_identity.rs:226-233: Make the timeout message reflect ChainLock state
  (existing thread: https://github.com/dashpay/platform/pull/4399#discussion_r3778738685)
  `strand_safe_timeout` unconditionally says the asset lock is on-chain, but the first call occurs specifically because Core has not reported the transaction as ChainLocked. The transaction may still be in the mempool, so the message asserts a finality state the command has not established. Use separate pre-ChainLock guidance that tells the operator to check Core before resuming, and retain the on-chain wording only after Core reported a ChainLock while Platform synchronization was pending.
```
</details>

packages/rs-scripts/src/bin/register_identity.rs
thepastaclaw: Resolved in `e8d2b14` — _Treat an empty CORE_RPC_PASSWORD as unset_ no longer present.

packages/rs-scripts/src/bin/register_identity.rs
thepastaclaw: Resolved in `14c9eda` — _Scope recovery credentials to one asset-lock transaction_ no longer present.

packages/rs-scripts/src/bin/register_identity.rs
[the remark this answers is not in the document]
shumkov: Fixed in 93ee7c6e23: `CORE_RPC_PASSWORD` is now `.filter(|p| !p.is_empty())`, so an exported-but-empty value no longer shadows `--core-rpc-password`.

packages/rs-scripts/src/bin/register_identity.rs
shumkov: Fixed (blocking) in 93ee7c6e23: `build_asset_lock_transaction` now rejects, before signing/broadcast, when the selected inputs exceed the versioned `max_asset_lock_transaction_inputs` — so an over-input, Platform-unusable lock is never created.

packages/rs-scripts/src/bin/register_identity.rs
shumkov: Fixed (blocking) in 93ee7c6e23: added `min_asset_lock_duffs()` mirroring `IdentityCreateTransition::calculate_min_required_fee` (versioned: asset-lock processing-start floor + identity_create_base_cost + identity_key_in_creation_cost*keys). The fresh path rejects a below-minimum `--fund-dash` before locking, so a too-small lock can't be broadcast and then stranded.

packages/rs-scripts/src/bin/register_identity.rs
shumkov: Fixed (blocking) in 93ee7c6e23: `--key-count` is now bounded above by the versioned `max_public_keys_in_creation` before generating keys or doing any funding work.

packages/rs-scripts/src/bin/register_identity.rs
shumkov: Fixed in 93ee7c6e23 (same change as the empty-`CORE_RPC_PASSWORD` thread): `.filter(|p| !p.is_empty())` before the `--core-rpc-password` fallback.

packages/rs-scripts/src/bin/register_identity.rs
shumkov: Fixed in 93ee7c6e23: both post-broadcast polling loops (Core `getrawtransaction` and DAPI metadata) now log a transient error and retry until `PROOF_TIMEOUT` instead of returning immediately, so a brief transport blip no longer forces manual recovery.

packages/rs-scripts/src/bin/register_identity.rs
shumkov: Fixed (blocking) in 93ee7c6e23: `append_creds` now calls `set_permissions(0o600)` right after opening, tightening a pre-existing looser file before any secret is written (the `.mode(0o600)` only applied to a freshly created inode).

packages/rs-scripts/src/bin/register_identity.rs
```diff
@@ -0,0 +1,635 @@
...
+//!
+//! The ChainLock wait can legitimately take minutes (platform ingests a
+//! fresh core ChainLock with some lag), so the proof timeout is generous.
+//! If it is nonetheless exceeded, the tool does NOT strand the funds: it
+//! reports that the asset lock is broadcast and on-chain, points at the
+//! saved one-time WIF, and prints the exact command to finish the
+//! registration against the existing lock — re-run with `--resume-txid
+//! <txid>`, which reuses the lock and spends nothing new. Re-running
+//! without `--resume-txid` would broadcast a second asset lock and strand
+//! the first, so the resume path exists precisely to avoid that.
```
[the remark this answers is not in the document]
shumkov: Fixed in 93ee7c6e23: `strand_safe_timeout` is now ChainLock-state-aware. The pre-ChainLock timeout says the tx may still be unconfirmed/in the mempool and to verify in Core first; the post-ChainLock timeout says only Platform catch-up timed out. No path claims a finality state it hasn't established.

packages/rs-scripts/src/bin/register_identity.rs
```diff
@@ -0,0 +1,635 @@
...
+/// Last-wins, matching how the file is sourced by a shell.
+fn read_creds_value(path: &str, key: &str) -> Result<Option<String>, String> {
+    let contents = std::fs::read_to_string(path)
+        .map_err(|e| format!("failed to read creds file {path}: {e}"))?;
+    let prefix = format!("{key}=");
+    Ok(contents
+        .lines()
+        .filter_map(|l| l.strip_prefix(&prefix))
+        .next_back()
+        .map(str::to_string))
```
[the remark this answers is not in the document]
shumkov: Fixed in 93ee7c6e23 (also the blocking thepastaclaw thread on this): recovery creds are now txid-scoped (`ASSET_LOCK_ONE_TIME_WIF__<txid>`), resume reads the scoped key, and identity keys are written only on a key-matching success — so a reused file can't pair an old txid with the latest WIF, and never advertises keys that don't control the registered identity.

packages/rs-scripts/src/bin/register_identity.rs
shumkov: Fixed (blocking) in 93ee7c6e23: resume reads `ASSET_LOCK_ONE_TIME_WIF__<txid>` (scoped to the requested txid), so it cannot sign with an unrelated lock's one-time key. Identity keys are written post-success only, and if the asset lock had already produced an identity (put_to_platform resolves AlreadyExists by fetching it) whose keys differ from ours, the tool warns and does NOT write our non-controlling keys.

packages/rs-scripts/src/bin/register_identity.rs
shumkov: Fixed in 93ee7c6e23: separate pre- and post-ChainLock timeout messages; the pre-ChainLock one no longer claims on-chain/ChainLock finality and tells the operator to verify in Core before resuming.

packages/rs-scripts/src/bin/register_identity.rs
shumkov: Fixed in 93ee7c6e23: on a `sendrawtransaction` transport error the tool now queries Core for the expected txid — if present it proceeds (Core accepted it), otherwise it reports the acceptance ambiguity and directs the operator to check Core and use `--resume-txid` rather than blindly re-running (which could create a second lock).

packages/rs-scripts/src/bin/register_identity.rs
thepastaclaw: This blocker is not fully resolved at `93ee7c6`. The one-time WIF is now txid-scoped, but lines 394-405 still generate a fresh identity key set on every run, including `--resume-txid`, and lines 648-664 persist those keys only after the registration call returns success. If Platform accepts the identity-create transition but the response is lost, the original controlling keys were never saved; a resume generates different keys, fetches the already-existing identity, and can only warn that the unavailable original keys are required. Please persist the complete identity key set in the txid-scoped recovery record before broadcast and reload that set on resume, as the original finding requested.

packages/rs-scripts/src/bin/register_identity.rs
```diff
@@ -0,0 +1,926 @@
...
+        .address
+        .split(',')
+        .map(str::trim)
+        .filter(|s| !s.is_empty())
+        .map(|s| {
+            s.parse::<DapiAddress>()
+                .map_err(|e| format!("failed to parse address '{s}': {e}"))
+        })
+        .collect::<Result<Vec<DapiAddress>, String>>()?;
+    let address_list = AddressList::from_iter(addresses);
```
thepastaclaw: **🟡 Suggestion: Reject an empty DAPI endpoint list before funding**

Empty components are filtered, but the resulting vector is never checked. `AddressList::from_iter` permits an empty list and `SdkBuilder::build` accepts it, so `--address ''` or `--address ',,'` reaches the fresh funding path. The asset-lock transaction can then be signed and broadcast before the first Platform request fails with no available DAPI addresses. Reject the empty list before constructing the SDK or performing any funding work.

```suggestion
let addresses = args
    .address
    .split(',')
    .map(str::trim)
    .filter(|s| !s.is_empty())
    .map(|s| {
        s.parse::<DapiAddress>()
            .map_err(|e| format!("failed to parse address '{s}': {e}"))
    })
    .collect::<Result<Vec<DapiAddress>, String>>()?;
if addresses.is_empty() {
    return Err("--address must contain at least one DAPI endpoint".to_string());
}
let address_list = AddressList::from_iter(addresses);
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

Seven of the nine prior findings are fixed at the exact head, including permission enforcement, versioned funding limits, polling resilience, timeout guidance, and ambiguous-broadcast recovery. One blocking recovery gap remains because identity keys are not persisted before registration, and two non-blocking input-validation issues remain for exact DASH parsing and empty DAPI endpoint lists.
Source: Reviewer backend model: gpt-5.6-sol. Final verifier backend model: gpt-5.6-sol. Orchestration-only: openclaw-agent/cliproxy/gpt-5.6-sol (not reviewer evidence).

### Review provenance

<details>

```

In `packages/rs-scripts/src/bin/register_identity.rs`:
- [SUGGESTION] packages/rs-scripts/src/bin/register_identity.rs:358-368: Reject an empty DAPI endpoint list before funding
  Empty components are filtered, but the resulting vector is never checked. `AddressList::from_iter` permits an empty list and `SdkBuilder::build` accepts it, so `--address ''` or `--address ',,'` reaches the fresh funding path. The asset-lock transaction can then be signed and broadcast before the first Platform request fails with no available DAPI addresses. Reject the empty list before constructing the SDK or performing any funding work.
- [BLOCKING] packages/rs-scripts/src/bin/register_identity.rs:394-405: Scope recovery credentials to one asset-lock transaction
  (existing thread: https://github.com/dashpay/platform/pull/4399#discussion_r3778738671)
  The one-time asset-lock WIF is now scoped to the txid, but the identity key set is still generated afresh on every invocation and is not persisted until lines 647-664, after registration returns successfully. If Platform accepts the identity-create transition but its successful response is lost, this call returns an error without saving the controlling private keys. A resume then generates different keys; the SDK's `AlreadyExists` path fetches the existing proof-derived identity, the key comparison fails, and the command cannot recover the unavailable original keys. Persist the complete identity key set in the txid-scoped recovery record before the fresh broadcast and reload that exact set on resume.
- [SUGGESTION] packages/rs-scripts/src/bin/register_identity.rs:337: Validate the identity-create minimum before locking funds
  (existing thread: https://github.com/dashpay/platform/pull/4399#discussion_r3777361973)
  The new versioned minimum-funding guard prevents an underfunded identity-create lock, but `--fund-dash` is still parsed as `f64` and converted by multiplication followed by a truncating cast. For example, `0.16483894 * 100_000_000` evaluates to `16483893.999999998` and locks 16,483,893 duffs instead of the requested 16,483,894; values with more than eight decimal places are silently truncated rather than rejected. Parse the original CLI string with `dpp::dashcore::Amount::from_str_in(..., Denomination::Dash)` and use `to_sat()` so the funded amount exactly matches the user's decimal input.
```
</details>

packages/rs-scripts/src/bin/register_identity.rs
```diff
@@ -0,0 +1,554 @@
...
+        .create(true)
+        .append(true)
+        .mode(0o600)
+        .open(path)
+        .map_err(|e| format!("failed to open creds file {path}: {e}"))?;
+    for (k, v) in lines {
+        writeln!(f, "{k}={v}").map_err(|e| format!("failed to write creds file: {e}"))?;
+    }
+    Ok(())
+}
```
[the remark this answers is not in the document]
thepastaclaw: No longer applicable at the exact head: `append_creds` explicitly enforces mode 0600 before writing any secret, including for an existing file.

packages/rs-scripts/src/bin/register_identity.rs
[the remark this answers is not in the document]
thepastaclaw: No longer applicable at the exact head: the environment password is filtered for empty values before falling back to `--core-rpc-password`.

packages/rs-scripts/src/bin/register_identity.rs
```diff
@@ -0,0 +1,554 @@
...
+            .map_err(|e| format!("getrawtransaction failed: {e}"))?;
+        if info.chainlock {
+            let h = info
+                .height
+                .ok_or("ChainLocked tx has no height in getrawtransaction result")?;
+            break u32::try_from(h).map_err(|_| format!("invalid tx height {h}"))?;
+        }
+        eprintln!("asset-lock tx not yet ChainLocked; retrying in 2s...");
+        tokio::time::sleep(Duration::from_secs(2)).await;
+    };
```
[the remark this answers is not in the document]
thepastaclaw: No longer applicable at the exact head: both post-broadcast Core and Platform polling loops now log transient request failures and retry until `PROOF_TIMEOUT`.

packages/rs-scripts/src/bin/register_identity.rs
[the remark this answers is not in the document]
thepastaclaw: No longer applicable at the exact head: timeout guidance now distinguishes a transaction not yet reported ChainLocked by Core from a ChainLocked transaction awaiting Platform catch-up.

packages/rs-scripts/src/bin/register_identity.rs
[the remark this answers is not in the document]
thepastaclaw: The txid-scoped one-time WIF is fixed, but the complete recovery issue is not. Identity keys are still generated afresh on every resume and saved only after registration returns success, so a lost success response can leave an existing identity whose controlling private keys were never persisted.

packages/rs-scripts/src/bin/register_identity.rs
shumkov: Fixed in 8b77f03b2d: extracted `parse_dapi_addresses()` which rejects an empty endpoint list (from `--address ''` or `,,`) with a hard error, called BEFORE the SDK is constructed and before any funding work — so the asset lock is never signed/broadcast against an empty address list. Same spend-before-validate ordering as the identity-create-minimum and input-cap guards. Unit test covers empty/comma-only/whitespace inputs and a valid multi-address list.

packages/rs-scripts/src/bin/register_identity.rs
```diff
@@ -0,0 +1,956 @@
...
+    // keys, so the key set is generated the same way in both the fresh and
+    // resume paths.
+    let mut rng = StdRng::from_entropy();
+    let (identity, key_material): (Identity, Vec<(IdentityPublicKey, [u8; 32])>) =
+        Identity::random_identity_with_main_keys_with_private_key(
+            args.key_count,
+            &mut rng,
+            platform_version,
+        )
+        .map_err(|e| format!("failed to generate identity: {e}"))?;
```
thepastaclaw: **🟡 Suggestion: Do not randomly generate unsupported authentication keys**

For `--key-count` above three, this fixture-oriented helper randomly selects additional authentication key types and security levels. Its V0 generator can label an extra key as `BIP13_SCRIPT_HASH`—for a fourth key, two of the nine selectable combinations have that type. Identity creation can still succeed because BIP13 is not a unique key type and therefore receives no creation witness, but `SimpleSigner::sign` explicitly rejects BIP13 and the generated key data has no redeem script. The resulting identity and credentials therefore contain an authentication key that cannot authorize later identity transitions. Generate production keys from explicitly supported signing types such as `ECDSA_SECP256K1`, `BLS12_381`, or `ECDSA_HASH160`, and test every accepted `--key-count` value.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The txid-scoped one-time WIF and empty-endpoint guard are correct, but identity private keys are still not persisted until after Platform registration returns, so a lost success response can permanently leave an identity without its controlling keys. Exact DASH parsing and randomly generated BIP13 authentication keys also remain non-blocking correctness issues.
Source: Reviewer backend model: gpt-5.6-sol; final verifier backend model: gpt-5.6-sol. Orchestration-only: openclaw-agent/cliproxy/gpt-5.6-sol (not reviewer evidence).

### Review provenance

<details>

```

In `packages/rs-scripts/src/bin/register_identity.rs`:
- [SUGGESTION] packages/rs-scripts/src/bin/register_identity.rs:418-424: Do not randomly generate unsupported authentication keys
  For `--key-count` above three, this fixture-oriented helper randomly selects additional authentication key types and security levels. Its V0 generator can label an extra key as `BIP13_SCRIPT_HASH`—for a fourth key, two of the nine selectable combinations have that type. Identity creation can still succeed because BIP13 is not a unique key type and therefore receives no creation witness, but `SimpleSigner::sign` explicitly rejects BIP13 and the generated key data has no redeem script. The resulting identity and credentials therefore contain an authentication key that cannot authorize later identity transitions. Generate production keys from explicitly supported signing types such as `ECDSA_SECP256K1`, `BLS12_381`, or `ECDSA_HASH160`, and test every accepted `--key-count` value.
- [BLOCKING] packages/rs-scripts/src/bin/register_identity.rs:513-534: Scope recovery credentials to one asset-lock transaction
  (existing thread: https://github.com/dashpay/platform/pull/4399#discussion_r3778738671)
  The fresh-path recovery record still saves only asset-lock data before broadcast. The generated identity keys are appended at lines 667-683 only after `put_to_platform_and_wait_for_response_with_private_key` returns successfully. If Platform accepts the transition but its response is lost, the process exits without saving the keys controlling the newly created identity. A resume generates a different key set at lines 418-424; the SDK's `AlreadyExists` path fetches the existing proof-derived identity, and the command can only warn that the unavailable original keys are required. Persist the complete identity definition and private material in this txid-scoped record before broadcasting, then reload and reconstruct that exact identity and signer on resume.
- [SUGGESTION] packages/rs-scripts/src/bin/register_identity.rs:363-365: Validate the identity-create minimum before locking funds
  (existing thread: https://github.com/dashpay/platform/pull/4399#discussion_r3777361973)
  The versioned minimum-funding guard is present, but `--fund-dash` is still parsed as `f64` and converted by multiplication followed by a truncating cast. For example, `0.16483894 * 100_000_000` evaluates to `16483893.999999998`, so the transaction locks 16,483,893 duffs instead of the requested 16,483,894. Inputs with more than eight decimal places are also silently truncated. Keep the original CLI value as a string, parse it with `dpp::dashcore::Amount::from_str_in(..., Denomination::Dash)`, and use `to_sat()` so excess precision is rejected and the locked amount exactly matches the user's input.
```
</details>

packages/rs-scripts/src/bin/register_identity.rs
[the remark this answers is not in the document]
thepastaclaw: This no longer applies at the current head. `append_creds` explicitly sets mode 0600 before writing any credential, including when the file already exists.

packages/rs-scripts/src/bin/register_identity.rs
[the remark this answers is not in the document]
thepastaclaw: This no longer applies at the current head. Empty `CORE_RPC_PASSWORD` values are filtered before falling back to `--core-rpc-password`.

packages/rs-scripts/src/bin/register_identity.rs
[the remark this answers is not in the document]
thepastaclaw: This no longer applies at the current head. Both the Core and Platform polling loops log transient request failures and retry until `PROOF_TIMEOUT`.

packages/rs-scripts/src/bin/register_identity.rs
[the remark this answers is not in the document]
thepastaclaw: This no longer applies at the current head. Timeout guidance now distinguishes a transaction not yet reported ChainLocked by Core from a ChainLocked transaction awaiting Platform catch-up.

packages/rs-scripts/src/bin/register_identity.rs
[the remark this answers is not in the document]
thepastaclaw: The one-time WIF is now txid-scoped, but the complete recovery problem remains. Identity keys are generated anew on every resume and are not persisted until registration returns successfully, so a lost success response can leave the existing identity's controlling keys unavailable.

packages/rs-scripts/src/bin/register_identity.rs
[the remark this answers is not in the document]
thepastaclaw: This is a mute/meta comment and contains no code-review finding to validate.

packages/rs-scripts/src/bin/register_identity.rs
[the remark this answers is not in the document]
thepastaclaw: This is a mute/meta comment and contains no code-review finding to validate.

packages/rs-scripts/src/bin/register_identity.rs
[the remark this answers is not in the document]
thepastaclaw: This is a mute/meta comment and contains no code-review finding to validate.

packages/rs-scripts/src/bin/register_identity.rs
shumkov: Fixed in a03f1e2930: identity keys are now constrained at the call site (the generator is a shared fixture, so its behaviour is left unchanged for other callers) to key types SimpleSigner can actually sign — ECDSA_SECP256K1, BLS12_381, ECDSA_HASH160 — regenerating any draw that includes another type (bounded loop). I checked simple-signer as you suggested: `sign` rejects only BIP13_SCRIPT_HASH (it handles ECDSA, BLS and EDDSA), so this allowlist is a strict safe subset — no BIP13, and no EDDSA (which we simply don't generate). The three main keys, including the CRITICAL signing key the tool hands off, are always ECDSA_SECP256K1. Unit tests cover the predicate and that a --key-count 6 identity ends up all-signable.

thepastaclaw (changes requested):

The ChainLock workflow now validates endpoint, key-count, funding-floor, and asset-lock input preconditions before spending, and the signer-compatible key-generation fix resolves the latest prior finding. Two issues remain: identity private keys are still not persisted until after registration succeeds, defeating recovery after a lost success response, and DASH amounts still pass through a lossy floating-point conversion.
Source: Reviewer backend model: gpt-5.6-sol; final verifier backend model: gpt-5.6-sol. Orchestration-only: openclaw-agent/cliproxy/gpt-5.6-sol (not reviewer evidence).

### Review provenance

🔴 1 blocking

<details>

```

In `packages/rs-scripts/src/bin/register_identity.rs`:
- [BLOCKING] packages/rs-scripts/src/bin/register_identity.rs:559-580: Scope recovery credentials to one asset-lock transaction
  (existing thread: https://github.com/dashpay/platform/pull/4399#discussion_r3778738671)
  The fresh-path recovery record remains incomplete. Although `identity_creds` is prepared at lines 472-490, this pre-broadcast write stores only the asset-lock txid, one-time WIF, address, and amount. The identity keys are not written until lines 713-729, after `put_to_platform_and_wait_for_response_with_private_key` returns successfully. If Platform accepts the identity-create transition but its response is lost, the process exits without persisting the private keys controlling the new identity. A resume generates a new identity and signer at lines 431-470; the SDK's `AlreadyExists` branch then fetches the proof-derived identity, but the key comparison can only warn that the unavailable original keys differ. Persist the complete identity definition and private material in the txid-scoped record before broadcasting, and reconstruct that exact identity and signer from the record on resume.
- [SUGGESTION] packages/rs-scripts/src/bin/register_identity.rs:381-384: Validate the identity-create minimum before locking funds
  (existing thread: https://github.com/dashpay/platform/pull/4399#discussion_r3777361973)
  The versioned minimum-funding guard is now present, but `--fund-dash` is still parsed as `f64` and converted by multiplication followed by a truncating cast. For example, `0.16483894 * 100_000_000` evaluates to `16483893.999999998`, so this code locks 16,483,893 duffs rather than the requested 16,483,894. Values with more than eight decimal places are also silently truncated. Preserve the CLI value as a string, parse it with `dpp::dashcore::Amount::from_str_in(..., Denomination::Dash)`, and use `to_sat()` so excess precision is rejected and the locked amount exactly matches the decimal input.
```
</details>

packages/rs-scripts/src/bin/register_identity.rs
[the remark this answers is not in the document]
thepastaclaw: This no longer applies at the current head: `append_creds` explicitly enforces mode 0600 before writing, including when the file already exists.

packages/rs-scripts/src/bin/register_identity.rs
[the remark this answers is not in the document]
thepastaclaw: This no longer applies at the current head: empty `CORE_RPC_PASSWORD` values are filtered before falling back to `--core-rpc-password`.

packages/rs-scripts/src/bin/register_identity.rs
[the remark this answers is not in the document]
thepastaclaw: This no longer applies at the current head: both post-broadcast polling loops retry transient request failures until `PROOF_TIMEOUT`.

packages/rs-scripts/src/bin/register_identity.rs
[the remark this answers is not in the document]
thepastaclaw: This no longer applies at the current head: timeout handling now distinguishes a transaction not yet reported ChainLocked by Core from a ChainLocked transaction awaiting Platform catch-up.

packages/rs-scripts/src/bin/register_identity.rs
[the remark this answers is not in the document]
thepastaclaw: This is a mute/meta comment and contains no code-review finding to validate.

packages/rs-scripts/src/bin/register_identity.rs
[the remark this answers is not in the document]
thepastaclaw: This is a mute/meta comment and contains no code-review finding to validate.

packages/rs-scripts/src/bin/register_identity.rs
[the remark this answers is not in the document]
thepastaclaw: This is a mute/meta comment and contains no code-review finding to validate.

PR #4404, opened 2026-08-13
fix(platform-wallet): give the conflict screen a source that survives the load

romchornyi:

Follow-up to #4356, targeting its branch — see https://github.com/dashpay/platform/pull/4356#issuecomment-5283461299 for the original report.

The conflict screen #4356 adds cannot fire on the iOS load path, because it reads state that path does not have. `first_confirmed_input_conflict` scans `info.core_wallet.transaction_history()`, and the FFI load path deliberately leaves `transactions()` empty apart from the unresolved locks' own funding records (`restore_unresolved_asset_lock_tx_records`). `catchUpStuckAssetLocks` runs at app launch, before block sync repopulates anything — so at the one moment the screen runs, it has nothing to scan.

Measured on a testnet device with a temporary diagnostic at the call site:

```
resume_asset_lock: conflict-screen inputs
  outpoint=7ef7773e…:0  status=Broadcast
  history_len=1
  inputs=["8789cd69…:1"]
```

`history_len=1`, and that single record is the lock's own funding transaction, which `record.txid != lock_txid` filters out. Zero candidates.

This was a textbook case for the screen: `8789cd69…:1` had already been taken by `8c19e1c2…`, confirmed and chain-locked at height 1532949 (verified against a testnet node). The screen returned `None`, the resume re-broadcast into the void, and `wait_for_proof` ran the full 300s — the exact behaviour #4356 sets out to eliminate.

The PR's own tests pass because they populate the transaction history first, so the gap is specific to the load path rather than the logic.

## What was done?

The host mirror already knows the answer — the persisted row for a spent outpoint records which transaction took it, with height and context — it just never crosses the FFI.

- New `AssetLockInputSpendFFI` row on `WalletRestoreEntryFFI`: outpoint, spender txid, spender height, and the spender's `TransactionContext` discriminant passed through **verbatim**. The host marshals; deciding which contexts count as final is Rust's call, per `packages/swift-sdk/CLAUDE.md`.
- Decoded into `PlatformWalletInfo::restored_asset_lock_input_spends` via `ClientWalletStartState`. Only a spender that reached a block settles an outpoint — a mempool or InstantSend-only sighting can still be replaced, so it is no basis for calling another transaction dead.
- `first_confirmed_input_conflict` consults that map first, then falls back to the existing history scan.
- Swift side: one additional fetch over `PersistentTxo` rows that carry a `spendingTransaction`, capped at 4096 rows, freed with the rest of the load allocation.

In-session behaviour is unchanged — the fallback scan still runs and still wins whenever the history is populated. A malformed row is skipped rather than failing the load: this is evidence for a screen that degrades to its previous behaviour without it, so a bad row must not cost the user their wallet.

## How Has This Been Tested?

- `cargo test -p platform-wallet --lib asset_lock::` — 51 passed, including #4356's own conflict suite unchanged.
- `cargo check -p platform-wallet --tests`, `cargo check -p platform-wallet-ffi` clean; `./build_ios.sh --target sim` succeeds (Swift compiles against the regenerated header).
- On the testnet device that produced the report, with this applied, the screen fires correctly at catch-up:

```
resume_asset_lock: asset lock double-spends an outpoint already consumed by a
  confirmed transaction; it can never confirm
  input=8789cd69…:1  spent_by=8c19e1c2…  height=Some(1532949)
  spender_chain_locked=true
```

That lock was the root of a three-transaction chain holding 1.57 DASH of phantom balance on that wallet, so the screen firing is what lets the whole chain be discarded — it earns its keep well beyond the error message.

## Breaking Changes

None. Additive fields on an FFI struct that already grows this way; hosts that do not populate them get the previous behaviour exactly.

## Checklist:

Two things I want to flag rather than quietly leave:

**No new test.** The gap is that a real load produces an empty history, and the existing suite constructs its wallets in memory with the history already populated — so a unit test asserting "the map is consulted" would not reproduce the condition that made the screen blind. Reproducing it properly needs a restore-from-persistence harness. I verified on device instead, and would rather say so than add a test that passes for the wrong reason. Happy to add one if you can point me at the right harness.

**I only checked the iOS path.** If the Kotlin load path repopulates `transactions()` on startup, this is iOS-specific and the framing above is broader than it should be.

### DIFF
--- a/packages/rs-platform-wallet-ffi/src/persistence.rs
+++ b/packages/rs-platform-wallet-ffi/src/persistence.rs
@@ -6,6 +6,7 @@
 //! SwiftData on iOS).
 
 use bincode::config;
+use dashcore::prelude::CoreBlockHeight;
 use key_wallet::account::account_collection::AccountCollection;
 use key_wallet::account::{Account, AccountType, BLSAccount, EdDSAAccount, StandardAccountType};
 use key_wallet::bip32::DerivationPath;
@@ -4796,12 +4797,14 @@ fn build_wallet_start_state(
     // was interrupted by an app kill can resume from the latest
     // status without rebroadcasting.
     let unused_asset_locks = build_unused_asset_locks(entry)?;
+    let asset_lock_input_spends = build_asset_lock_input_spends(entry);
 
     let wallet_state = ClientWalletStartState {
         wallet,
         wallet_info,
         identity_manager,
         unused_asset_locks,
+        asset_lock_input_spends,
     };
 
     let platform_address_state = if per_account.is_empty()
@@ -4843,6 +4846,66 @@ fn build_wallet_start_state(
 /// registration whose key-persist round hasn't completed) loads with
 /// an empty map and gets refreshed on the next sync round —
 /// degraded-but-usable for that narrow case.
+/// Decode the host mirror's report of which transaction took each outpoint
+/// an unresolved asset lock spends.
+///
+/// A malformed row is skipped rather than failing the load: the map is
+/// evidence for a screen that degrades to its old behaviour without it, so a
+/// bad row must not cost the user their wallet.
+fn build_asset_lock_input_spends(
+    entry: &WalletRestoreEntryFFI,
+) -> BTreeMap<dashcore::OutPoint, platform_wallet::wallet::platform_wallet::RestoredSpend> {
+    use dashcore::hashes::Hash;
+
+    let mut spends = BTreeMap::new();
+    if entry.asset_lock_input_spends.is_null() || entry.asset_lock_input_spends_count == 0 {
+        return spends;
+    }
+    let rows = unsafe {
+        slice::from_raw_parts(
+            entry.asset_lock_input_spends,
+            entry.asset_lock_input_spends_count,
+        )
+    };
+    for row in rows {
+        let (Ok(prev_txid), Ok(spender_txid)) = (
+            dashcore::Txid::from_slice(&row.prev_txid),
+            dashcore::Txid::from_slice(&row.spender_txid),
+        ) else {
+            tracing::warn!(
+                wallet_id = %hex::encode(entry.wallet_id),
+                "load: skipping asset-lock input-spend row with malformed txid bytes"
+            );
+            continue;
+        };
+        // Both facts are kept, and the readers apply their own bar: the
+        // double-spend screen needs a spender that reached a block, the
+        // abandon cascade needs any spender at all.
+        const CONTEXT_IN_BLOCK: u32 = 2;
+        const CONTEXT_IN_CHAIN_LOCKED_BLOCK: u32 = 3;
+        spends.insert(
+            dashcore::OutPoint {
+                txid: prev_txid,
+                vout: row.vout,
+            },
+            platform_wallet::wallet::platform_wallet::RestoredSpend {
+                spender: spender_txid,
+                height: (row.spender_height != 0).then_some(row.spender_height),
+                in_block: row.spender_context >= CONTEXT_IN_BLOCK,
+                chain_locked: row.spender_context >= CONTEXT_IN_CHAIN_LOCKED_BLOCK,
+            },
+        );
+    }
+    if !spends.is_empty() {
+        tracing::info!(
+            wallet_id = %hex::encode(entry.wallet_id),
+            count = spends.len(),
+            "load: restored asset-lock input-spend conflicts"
+        );
+    }
+    spends
+}
+
 /// Rebuild the `unused_asset_locks` map carried on
 /// [`ClientWalletStartState`] from the `tracked_asset_locks` slice the
 /// Swift load callback hands back. Mirrors the encoding used by
--- a/packages/rs-platform-wallet-ffi/src/persistence.rs
+++ b/packages/rs-platform-wallet-ffi/src/persistence.rs
@@ -58,8 +58,8 @@ use crate::platform_address_types::AddressBalanceEntryFFI;
 use crate::token_persistence::{TokenBalanceRemovalFFI, TokenBalanceUpsertFFI};
 use crate::wallet_registration_persistence::AccountAddressPoolFFI;
 use crate::wallet_restore_types::{
-    AccountSpecFFI, AccountTypeTagFFI, ContactProfileRestoreEntryFFI, IdentityKeyRestoreFFI,
-    IdentityRestoreEntryFFI, LoadWalletListFreeFn, PaymentRestoreEntryFFI,
+    AccountSpecFFI, AccountTypeTagFFI, AssetLockInputSpendFFI, ContactProfileRestoreEntryFFI,
+    IdentityKeyRestoreFFI, IdentityRestoreEntryFFI, LoadWalletListFreeFn, PaymentRestoreEntryFFI,
     ProviderSpecialTxRestoreEntryFFI, StandardAccountTypeTagFFI, UnresolvedAssetLockTxRecordFFI,
     UtxoRestoreEntryFFI, WalletRestoreEntryFFI,
 };
@@ -4878,9 +4878,12 @@ fn build_asset_lock_input_spends(
             );
             continue;
         };
-        // Both facts are kept, and the readers apply their own bar: the
-        // double-spend screen needs a spender that reached a block, the
-        // abandon cascade needs any spender at all.
+        // Match the known discriminants exactly rather than comparing by
+        // order: the contract defines 0..=3, and an unknown value must
+        // degrade to "no evidence" rather than being read as finality. The
+        // screen treats `in_block` as conclusive and returns a terminal code
+        // the host may act on by discarding the lock, so a malformed or
+        // forward-versioned byte manufacturing that verdict would be unsafe.
         const CONTEXT_IN_BLOCK: u32 = 2;
         const CONTEXT_IN_CHAIN_LOCKED_BLOCK: u32 = 3;
         spends.insert(
@@ -4891,8 +4894,11 @@ fn build_asset_lock_input_spends(
             platform_wallet::wallet::platform_wallet::RestoredSpend {
                 spender: spender_txid,
                 height: (row.spender_height != 0).then_some(row.spender_height),
-                in_block: row.spender_context >= CONTEXT_IN_BLOCK,
-                chain_locked: row.spender_context >= CONTEXT_IN_CHAIN_LOCKED_BLOCK,
+                in_block: matches!(
+                    row.spender_context,
+                    CONTEXT_IN_BLOCK | CONTEXT_IN_CHAIN_LOCKED_BLOCK
+                ),
+                chain_locked: row.spender_context == CONTEXT_IN_CHAIN_LOCKED_BLOCK,
             },
         );
     }
@@ -6021,6 +6027,43 @@ mod tests {
     //! exercising the in-memory mutation against synthetic input.
 
     use super::*;
+
+    // --- asset-lock input-spend linkage decode ---
+
+    /// The context byte decides whether persisted evidence may condemn a
+    /// tracked lock, so only the two known block discriminants may read as
+    /// final. An unknown value — corrupt row, forward-versioned host — must
+    /// degrade to "no evidence" rather than manufacture finality.
+    #[test]
+    fn asset_lock_input_spend_context_decodes_only_known_block_discriminants() {
+        for (context, expect_in_block, expect_chain_locked) in [
+            (0u32, false, false), // mempool
+            (1, false, false),    // InstantSend, replaceable
+            (2, true, false),     // in a block
+            (3, true, true),      // chain-locked block
+            (4, false, false),    //unknown / forward-versioned
+            (u32::MAX, false, false),
+        ] {
+            let row = AssetLockInputSpendFFI {
+                prev_txid: [7u8; 32],
+                vout: 1,
+                spender_txid: [9u8; 32],
+                spender_height: 1_532_949,
+                spender_context: context,
+            };
+            // The decoder reads only `wallet_id` (for the log line) and the
+            // spend slice, so a zeroed entry is a sound stand-in for the
+            // ~40 pointer fields it never touches.
+            let mut entry: WalletRestoreEntryFFI = unsafe { std::mem::zeroed() };
+            entry.asset_lock_input_spends = &row;
+            entry.asset_lock_input_spends_count = 1;
+
+            let spends = build_asset_lock_input_spends(&entry);
+            let spend = spends.values().next().expect("row decodes");
+            assert_eq!(spend.in_block, expect_in_block, "context={context}");
+            assert_eq!(spend.chain_locked, expect_chain_locked, "context={context}");
+        }
+    }
 
     // --- persists_durably: the fail-closed durability attestation ---
 
--- a/packages/rs-platform-wallet-ffi/src/persistence.rs
+++ b/packages/rs-platform-wallet-ffi/src/persistence.rs
@@ -6,7 +6,6 @@
 //! SwiftData on iOS).
 
 use bincode::config;
-use dashcore::prelude::CoreBlockHeight;
 use key_wallet::account::account_collection::AccountCollection;
 use key_wallet::account::{Account, AccountType, BLSAccount, EdDSAAccount, StandardAccountType};
 use key_wallet::bip32::DerivationPath;
@@ -58,8 +57,8 @@ use crate::platform_address_types::AddressBalanceEntryFFI;
 use crate::token_persistence::{TokenBalanceRemovalFFI, TokenBalanceUpsertFFI};
 use crate::wallet_registration_persistence::AccountAddressPoolFFI;
 use crate::wallet_restore_types::{
-    AccountSpecFFI, AccountTypeTagFFI, AssetLockInputSpendFFI, ContactProfileRestoreEntryFFI,
-    IdentityKeyRestoreFFI, IdentityRestoreEntryFFI, LoadWalletListFreeFn, PaymentRestoreEntryFFI,
+    AccountSpecFFI, AccountTypeTagFFI, ContactProfileRestoreEntryFFI, IdentityKeyRestoreFFI,
+    IdentityRestoreEntryFFI, LoadWalletListFreeFn, PaymentRestoreEntryFFI,
     ProviderSpecialTxRestoreEntryFFI, StandardAccountTypeTagFFI, UnresolvedAssetLockTxRecordFFI,
     UtxoRestoreEntryFFI, WalletRestoreEntryFFI,
 };
@@ -71,6 +70,17 @@ use dpp::platform_value::BinaryData;
 use dpp::prelude::Identifier;
 use platform_wallet::{DpnsNameInfo, IdentityManagerStartState, IdentityStatus, ManagedIdentity};
 use std::ffi::CStr;
+
+/// The persisted `TransactionContext` discriminant values shared with the
+/// host mirrors (`PersistentTransaction.context` on Swift): `0` mempool,
+/// `1` InstantSend, `2` in a block, `3` in a chain-locked block. Every u32
+/// `context_raw` decoder in this crate matches the confirmed contexts
+/// against these constants — a new context value must be added here first,
+/// so a grep for the constant names finds every decoder that has to learn
+/// it. The sites deliberately differ in their defensive defaults (miss vs
+/// `Mempool` vs no-evidence); see each match's comment.
+pub(crate) const TX_CONTEXT_RAW_IN_BLOCK: u32 = 2;
+pub(crate) const TX_CONTEXT_RAW_IN_CHAIN_LOCKED_BLOCK: u32 = 3;
 
 /// Versioned C projection of [`PersistenceCapabilities`].
 ///
@@ -2948,6 +2958,9 @@ impl PlatformWalletPersistence for FFIPersister {
             return Ok(None);
         }
 
+        // `context_kind` is the u8 out-param twin of the u32
+        // `TX_CONTEXT_RAW_*` discriminants at the top of this file — the
+        // values must stay in lockstep with those constants.
         let context = match context_kind {
             0 => TransactionContext::Mempool,
             1 => {
@@ -4825,33 +4838,16 @@ fn build_wallet_start_state(
     Ok((wallet_state, platform_address_state))
 }
 
-/// Translate the `IdentityRestoreEntryFFI` slice carried on a wallet
-/// entry into the wallet-bucket portion of an
-/// [`IdentityManagerStartState`].
-///
-/// Every entry on a `WalletRestoreEntryFFI` is wallet-owned by
-/// definition, so the returned map is shaped for direct insertion
-/// into `wallet_identities[entry.wallet_id]`. Out-of-wallet identities
-/// (no associated wallet) come from a separate path that today simply
-/// doesn't exist in SwiftData — see the report observation.
-///
-/// The DPP `Identity` is reconstructed from the persisted scalars via
-/// the `IdentityV0` shape — same approach
-/// [`apply_identity_entry`](platform_wallet::IdentityManager::apply_identity_entry)
-/// uses on the changeset replay path. Public keys are now pulled in
-/// from the `keys` array on each `IdentityRestoreEntryFFI` (assembled
-/// from the per-identity `PersistentPublicKey` rows on the Swift
-/// side), so the restored `Identity.public_keys` map is populated at
-/// load time. An identity with no persisted keys (e.g. an in-flight
-/// registration whose key-persist round hasn't completed) loads with
-/// an empty map and gets refreshed on the next sync round —
-/// degraded-but-usable for that narrow case.
 /// Decode the host mirror's report of which transaction took each outpoint
 /// an unresolved asset lock spends.
 ///
 /// A malformed row is skipped rather than failing the load: the map is
 /// evidence for a screen that degrades to its old behaviour without it, so a
-/// bad row must not cost the user their wallet.
+/// bad row must not cost the user their wallet. "Malformed" here means an
+/// all-zero txid on either side of the row — the shape a zero-initialised
+/// struct from a host that never filled the row in would take. (The 32-byte
+/// arrays themselves always parse, so this check is the row validation, not
+/// the `Txid` constructor.)
 fn build_asset_lock_input_spends(
     entry: &WalletRestoreEntryFFI,
 ) -> BTreeMap<dashcore::OutPoint, platform_wallet::wallet::platform_wallet::RestoredSpend> {
@@ -4868,24 +4864,27 @@ fn build_asset_lock_input_spends(
         )
     };
     for row in rows {
-        let (Ok(prev_txid), Ok(spender_txid)) = (
-            dashcore::Txid::from_slice(&row.prev_txid),
-            dashcore::Txid::from_slice(&row.spender_txid),
-        ) else {
+        // A fixed 32-byte array always parses as a `Txid`, so the real
+        // malformed-row check is content: an all-zero txid on either side is
+        // the shape of a row a host zero-initialised and never filled in,
+        // and no genuine transaction hashes to zero.
+        if row.prev_txid == [0u8; 32] || row.spender_txid == [0u8; 32] {
             tracing::warn!(
                 wallet_id = %hex::encode(entry.wallet_id),
-                "load: skipping asset-lock input-spend row with malformed txid bytes"
+                "load: skipping asset-lock input-spend row with zeroed txid bytes"
             );
             continue;
-        };
+        }
+        let prev_txid = dashcore::Txid::from_slice(&row.prev_txid)
+            .expect("32-byte array always parses as Txid");
+        let spender_txid = dashcore::Txid::from_slice(&row.spender_txid)
+            .expect("32-byte array always parses as Txid");
         // Match the known discriminants exactly rather than comparing by
         // order: the contract defines 0..=3, and an unknown value must
         // degrade to "no evidence" rather than being read as finality. The
         // screen treats `in_block` as conclusive and returns a terminal code
         // the host may act on by discarding the lock, so a malformed or
         // forward-versioned byte manufacturing that verdict would be unsafe.
-        const CONTEXT_IN_BLOCK: u32 = 2;
-        const CONTEXT_IN_CHAIN_LOCKED_BLOCK: u32 = 3;
         spends.insert(
             dashcore::OutPoint {
                 txid: prev_txid,
@@ -4896,17 +4895,21 @@ fn build_asset_lock_input_spends(
                 height: (row.spender_height != 0).then_some(row.spender_height),
                 in_block: matches!(
                     row.spender_context,
-                    CONTEXT_IN_BLOCK | CONTEXT_IN_CHAIN_LOCKED_BLOCK
+                    TX_CONTEXT_RAW_IN_BLOCK | TX_CONTEXT_RAW_IN_CHAIN_LOCKED_BLOCK
                 ),
-                chain_locked: row.spender_context == CONTEXT_IN_CHAIN_LOCKED_BLOCK,
+                chain_locked: row.spender_context == TX_CONTEXT_RAW_IN_CHAIN_LOCKED_BLOCK,
             },
         );
     }
     if !spends.is_empty() {
+        // "rows", not "conflicts": the host emits whatever spender the
+        // mirror linked, which for a healthy broadcast lock is the lock's
+        // own transaction — whether a row is a conflict is decided
+        // per-lock by the screen, not here.
         tracing::info!(
             wallet_id = %hex::encode(entry.wallet_id),
             count = spends.len(),
-            "load: restored asset-lock input-spend conflicts"
+            "load: restored asset-lock input-spend rows"
         );
     }
     spends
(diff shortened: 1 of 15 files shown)


thepastaclaw:
🕓 Ready for review — 3 ahead in queue (commit c82369f1)
_Queue position: 4/4_
_ETA: start ~08:23 UTC · complete ~08:37 UTC (median 13m across 30 recent reviews; 2 slots)_
_Queued 54s ago · Last checked: 2026-08-19 08:10 UTC_

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -5417,6 +5430,67 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
...
+    ) -> (UnsafeMutablePointer<AssetLockInputSpendFFI>?, Int) {
+        // Deliberately not filtered on `isSpent`. A row keeps that flag
+        // false until its spender reaches a block, so the outputs of a
+        // transaction that never got there — exactly the chain the abandon
+        // cascade has to follow — sit in the unspent set with their linkage
+        // recorded and nothing else pointing at them.
+        var descriptor = FetchDescriptor<PersistentTxo>(
+            predicate: #Predicate { txo in
+                txo.walletId == walletId && txo.spendingTransaction != nil
+            }
```
thepastaclaw: **🔴 Blocking: Avoid the crash-prone relationship predicate during wallet load**

`PersistentTxo` explicitly documents that predicates traversing its optional `spendingTransaction` relationship enter a SwiftData nested-optional path that crashes, which is why the scalar `isSpent` column exists. This query runs during wallet loading, and `try?` cannot recover from a process-level SwiftData crash. The conflict reader only accepts restored rows whose spender reached a block, while the persistence handler sets `isSpent` under that same condition, so filtering on the scalar retains every row the current Rust conflict screen can use without touching the unsafe predicate path.

```suggestion
predicate: #Predicate { txo in
                txo.walletId == walletId && txo.isSpent == true
            }
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/persistence.rs
```diff
@@ -4843,6 +4846,66 @@ fn build_wallet_start_state(
...
+        spends.insert(
+            dashcore::OutPoint {
+                txid: prev_txid,
+                vout: row.vout,
+            },
+            platform_wallet::wallet::platform_wallet::RestoredSpend {
+                spender: spender_txid,
+                height: (row.spender_height != 0).then_some(row.spender_height),
+                in_block: row.spender_context >= CONTEXT_IN_BLOCK,
+                chain_locked: row.spender_context >= CONTEXT_IN_CHAIN_LOCKED_BLOCK,
```
thepastaclaw: **🔴 Blocking: Do not classify unknown context values as chain-locked**

The FFI contract defines only context discriminants 0 through 3, but these ordered comparisons classify every malformed or future value, including 4 and `u32::MAX`, as both confirmed and chain-locked. That is unsafe because `first_confirmed_input_conflict` treats `in_block` as conclusive evidence and returns the terminal conflict code that allows the host to discard the tracked lock. Match only the known block-context discriminants so invalid persisted data degrades to the previous no-evidence behavior instead of manufacturing finality.

```suggestion
height: (row.spender_height != 0).then_some(row.spender_height),
                in_block: matches!(row.spender_context, 2 | 3),
                chain_locked: row.spender_context == 3,
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs
```diff
@@ -260,6 +260,19 @@ fn first_confirmed_input_conflict(
...
+    // below has nothing to find however dead the lock is. The mirror knows
+    // because it recorded which transaction took the outpoint.
+    if let Some((input, spend)) = lock_inputs.iter().find_map(|input| {
+        info.restored_asset_lock_input_spends
+            .get_key_value(input)
+            .filter(|(_, spend)| spend.spender != lock_txid && spend.in_block)
+    }) {
+        return Some((*input, spend.spender, spend.height, spend.chain_locked));
+    }
+
```
thepastaclaw: **🟡 Suggestion: Cover the newly added empty-history conflict path**

The existing conflict tests insert spender records into `transaction_history()`, while every test fixture initializes `restored_asset_lock_input_spends` as empty. Add focused tests with empty history and a restored spend for the funded input: a confirmed different spender must return `AssetLockInputConflict` without broadcasting, while an unconfirmed spender and the lock's own txid must not trigger the terminal error. The FFI decoder should likewise cover unknown context values, because that boundary determines whether persisted evidence can condemn a lock.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The restore plumbing addresses the startup visibility gap, but the new load query uses a SwiftData relationship predicate that the model explicitly identifies as crash-prone, and the FFI decoder can turn unknown context values into false terminal conflicts. The newly introduced restored-map path and its safety gates also need focused regression coverage.
Source: reviewer backend gpt-5.6-sol; final verifier backend gpt-5.6-sol; openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

### Review provenance

<details>

```

- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift:5457-5459: Avoid the crash-prone relationship predicate during wallet load
  `PersistentTxo` explicitly documents that predicates traversing its optional `spendingTransaction` relationship enter a SwiftData nested-optional path that crashes, which is why the scalar `isSpent` column exists. This query runs during wallet loading, and `try?` cannot recover from a process-level SwiftData crash. The conflict reader only accepts restored rows whose spender reached a block, while the persistence handler sets `isSpent` under that same condition, so filtering on the scalar retains every row the current Rust conflict screen can use without touching the unsafe predicate path.

- [BLOCKING] packages/rs-platform-wallet-ffi/src/persistence.rs:4893-4895: Do not classify unknown context values as chain-locked
  The FFI contract defines only context discriminants 0 through 3, but these ordered comparisons classify every malformed or future value, including 4 and `u32::MAX`, as both confirmed and chain-locked. That is unsafe because `first_confirmed_input_conflict` treats `in_block` as conclusive evidence and returns the terminal conflict code that allows the host to discard the tracked lock. Match only the known block-context discriminants so invalid persisted data degrades to the previous no-evidence behavior instead of manufacturing finality.

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs:268-275: Cover the newly added empty-history conflict path
  The existing conflict tests insert spender records into `transaction_history()`, while every test fixture initializes `restored_asset_lock_input_spends` as empty. Add focused tests with empty history and a restored spend for the funded input: a confirmed different spender must return `AssetLockInputConflict` without broadcasting, while an unconfirmed spender and the lock's own txid must not trigger the terminal error. The FFI decoder should likewise cover unknown context values, because that boundary determines whether persisted evidence can condemn a lock.
```
</details>

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -5417,6 +5430,69 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
...
+        // exists for exactly this reason. No rows are lost here: `isSpent`
+        // flips under the same in-block condition the conflict screen
+        // requires of a spender, so every row it can act on is included.
+        var descriptor = FetchDescriptor<PersistentTxo>(
+            predicate: #Predicate { txo in
+                txo.walletId == walletId && txo.isSpent == true
+            }
+        )
+        descriptor.relationshipKeyPathsForPrefetching = [\.spendingTransaction]
+        descriptor.fetchLimit = Self.assetLockInputSpendCap
```
thepastaclaw: **🔴 Blocking: Do not cap an unfiltered set before selecting the lock inputs**

This query selects an unspecified first 4096 confirmed spent TXOs for the wallet, while only Rust later intersects those rows with the unresolved asset locks' input outpoints. `PersistentTxo` retains spent outputs for wallet history, so a wallet can legitimately exceed this limit, and neither a predicate nor a sort gives relevant lock inputs priority. If the required outpoint falls outside the returned subset, startup again has no evidence of the confirmed conflict and `resume_asset_lock` enters the full proof wait—the load-path failure this PR is intended to fix. Remove the cap or constrain the Swift query to the unresolved locks' input outpoints before applying a bound.

<sub>source: ['codex']</sub>

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -5417,6 +5430,69 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
...
+                Swift.withUnsafeMutableBytes(of: &row.spender_txid) { dst in
+                    dst.copyMemory(from: src)
+                }
+            }
+            row.spender_height = spender.blockHeight
+            row.spender_context = spender.context
+            buf[written] = row
+            written += 1
+        }
+        return written == 0 ? (nil, 0) : (buf, written)
```
thepastaclaw: **🔴 Blocking: Track only initialized spend-linkage buffer elements**

The allocation is registered with `spent.count` before rows with a missing `spendingTransaction` or malformed txid are skipped. `LoadAllocation.release()` consequently calls `deinitialize(count:)` for the full fetched count even though only `written` elements were initialized. When every row is skipped, the function returns `(nil, 0)` but retains a wholly uninitialized buffer for later deinitialization. This violates `UnsafeMutablePointer`'s initialized-memory contract; the allocation tracker elsewhere in this file explicitly documents that deinitializing uninitialized slots is undefined behavior. Deallocate immediately when no rows were written and otherwise register exactly the initialized count.

```suggestion
let buf = UnsafeMutablePointer<AssetLockInputSpendFFI>.allocate(capacity: spent.count)
        var written = 0
        for txo in spent {
            guard let spender = txo.spendingTransaction else { continue }

            guard txo.txid.count == 32, spender.txid.count == 32 else { continue }
            var row = AssetLockInputSpendFFI()
            txo.txid.withUnsafeBytes { src in
                Swift.withUnsafeMutableBytes(of: &row.prev_txid) { dst in
                    dst.copyMemory(from: src)
                }
            }
            row.vout = txo.vout
            spender.txid.withUnsafeBytes { src in
                Swift.withUnsafeMutableBytes(of: &row.spender_txid) { dst in
                    dst.copyMemory(from: src)
                }
            }
            row.spender_height = spender.blockHeight
            row.spender_context = spender.context
            buf[written] = row
            written += 1
        }
        if written == 0 {
            buf.deallocate()
            return (nil, 0)
        }
        allocation.assetLockInputSpendBuffers.append((buf, written))
        return (buf, written)
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The current head fixes all three prior findings: the SwiftData query now uses the safe scalar predicate, the Rust decoder accepts only known context discriminants, and focused regression tests cover the empty-history restore path and decoder boundary. Two new Swift load-path blockers remain: the arbitrary 4096-row fetch limit can omit the conflict this PR needs to restore, and cleanup deinitializes more spend-linkage elements than were initialized.

### Review provenance

🔴 2 blocking

<details>

```

- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift:5463-5464: Do not cap an unfiltered set before selecting the lock inputs
  This query selects an unspecified first 4096 confirmed spent TXOs for the wallet, while only Rust later intersects those rows with the unresolved asset locks' input outpoints. `PersistentTxo` retains spent outputs for wallet history, so a wallet can legitimately exceed this limit, and neither a predicate nor a sort gives relevant lock inputs priority. If the required outpoint falls outside the returned subset, startup again has no evidence of the confirmed conflict and `resume_asset_lock` enters the full proof wait—the load-path failure this PR is intended to fix. Remove the cap or constrain the Swift query to the unresolved locks' input outpoints before applying a bound.
- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift:5469-5493: Track only initialized spend-linkage buffer elements
  The allocation is registered with `spent.count` before rows with a missing `spendingTransaction` or malformed txid are skipped. `LoadAllocation.release()` consequently calls `deinitialize(count:)` for the full fetched count even though only `written` elements were initialized. When every row is skipped, the function returns `(nil, 0)` but retains a wholly uninitialized buffer for later deinitialization. This violates `UnsafeMutablePointer`'s initialized-memory contract; the allocation tracker elsewhere in this file explicitly documents that deinitializing uninitialized slots is undefined behavior. Deallocate immediately when no rows were written and otherwise register exactly the initialized count.
```
</details>

packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs
thepastaclaw: Resolved in `48afef9` — _Cover the newly added empty-history conflict path_ no longer present.

packages/rs-platform-wallet-ffi/src/persistence.rs
thepastaclaw: Resolved in `48afef9` — _Do not classify unknown context values as chain-locked_ no longer present.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
thepastaclaw: Resolved in `48afef9` — _Avoid the crash-prone relationship predicate during wallet load_ no longer present.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -5399,6 +5412,133 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
...
+            guard let outpoint = decodeOutPointHex(lock.outPointHex) else { continue }
+            let txidData = Data(outpoint.prefix(32))
+            var txDescriptor = FetchDescriptor<PersistentTransaction>(
+                predicate: #Predicate { $0.txid == txidData }
+            )
+            txDescriptor.fetchLimit = 1
+            guard let txRow = try? backgroundContext.fetch(txDescriptor).first,
+                  !txRow.transactionData.isEmpty,
+                  let decoded = try? TransactionDecoder.decode(txRow.transactionData, network: network)
+            else { continue }
```
thepastaclaw: **🔴 Blocking: Decode inputs from the asset-lock row instead of requiring a transaction row**

Every `PersistentAssetLock` stores the consensus-encoded funding transaction in `transactionBytes`, and `buildAssetLockRestoreBuffer` uses those bytes to restore the tracked lock. This helper instead requires a separate `PersistentTransaction` row for the lock txid and skips the lock when that row is missing or empty. The same file explicitly recognizes that a Built/Broadcast asset-lock row can exist when its own transaction never reached the transaction table. In that state, the input TXO can still be linked to a different confirmed spender, but this early skip prevents the outpoint lookup and leaves the restored conflict map blind, reproducing the startup proof-wait failure this PR is intended to fix. Decode the authoritative bytes already stored on the asset-lock row.

```suggestion
guard !lock.transactionBytes.isEmpty,
      let decoded = try? TransactionDecoder.decode(
          lock.transactionBytes,
          network: network
      )
else { continue }
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs
```diff
@@ -260,6 +260,19 @@ fn first_confirmed_input_conflict(
...
+    // source that works at app-launch catch-up: `transaction_history()` is
+    // empty then except for the unresolved locks themselves, so the scan
+    // below has nothing to find however dead the lock is. The mirror knows
+    // because it recorded which transaction took the outpoint.
+    if let Some((input, spend)) = lock_inputs.iter().find_map(|input| {
+        info.restored_asset_lock_input_spends
+            .get_key_value(input)
+            .filter(|(_, spend)| spend.spender != lock_txid && spend.in_block)
+    }) {
+        return Some((*input, spend.spender, spend.height, spend.chain_locked));
```
thepastaclaw: **🟡 Suggestion: Apply the restored chain-lock boundary to restored spends**

The restored path reports only `spend.chain_locked`, which reflects the persisted transaction context. A transaction row can remain `InBlock` after `last_applied_chain_lock` advances beyond its block height; that is why the history fallback immediately below combines the record context with `chain_locked_height`. The same restored metadata is available here, so an InBlock restored spend at or below that boundary is final even when its row was never promoted to context 3. Without the same fallback, the terminal error incorrectly exposes `spender_chain_locked: false` to the host.

```suggestion
let spender_chain_locked = spend.chain_locked
    || chain_locked_height
        .zip(spend.height)
        .is_some_and(|(boundary, spender_height)| spender_height <= boundary);
return Some((*input, spend.spender, spend.height, spender_chain_locked));
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The latest commit fixes both prior blockers by selecting only unresolved-lock inputs and initializing exactly the tracked buffer length. The restored conflict path still skips valid locks when their standalone transaction row is absent, and its finality report fails to apply the restored chain-lock boundary.

### Review provenance

<details>

```

- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift:5520-5529: Decode inputs from the asset-lock row instead of requiring a transaction row
  Every `PersistentAssetLock` stores the consensus-encoded funding transaction in `transactionBytes`, and `buildAssetLockRestoreBuffer` uses those bytes to restore the tracked lock. This helper instead requires a separate `PersistentTransaction` row for the lock txid and skips the lock when that row is missing or empty. The same file explicitly recognizes that a Built/Broadcast asset-lock row can exist when its own transaction never reached the transaction table. In that state, the input TXO can still be linked to a different confirmed spender, but this early skip prevents the outpoint lookup and leaves the restored conflict map blind, reproducing the startup proof-wait failure this PR is intended to fix. Decode the authoritative bytes already stored on the asset-lock row.

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs:268-273: Apply the restored chain-lock boundary to restored spends
  The restored path reports only `spend.chain_locked`, which reflects the persisted transaction context. A transaction row can remain `InBlock` after `last_applied_chain_lock` advances beyond its block height; that is why the history fallback immediately below combines the record context with `chain_locked_height`. The same restored metadata is available here, so an InBlock restored spend at or below that boundary is final even when its row was never promoted to context 3. Without the same fallback, the terminal error incorrectly exposes `spender_chain_locked: false` to the host.
```
</details>

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
thepastaclaw: Resolved in `84eddfc` — _Track only initialized spend-linkage buffer elements_ no longer present.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
thepastaclaw: Resolved in `84eddfc` — _Do not cap an unfiltered set before selecting the lock inputs_ no longer present.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
```diff
@@ -5399,6 +5412,140 @@ public final class PlatformWalletPersistenceHandler: @unchecked Sendable {
...
+                predicate: #Predicate { $0.outpoint == key }
+            )
+            descriptor.fetchLimit = 1
+            descriptor.relationshipKeyPathsForPrefetching = [\.spendingTransaction]
+            guard let txo = try? backgroundContext.fetch(descriptor).first,
+                  txo.walletId == walletId,
+                  txo.isSpent,
+                  let spender = txo.spendingTransaction,
+                  txo.txid.count == 32,
+                  spender.txid.count == 32
```
thepastaclaw: **🔴 Blocking: Accept migrated TXOs whose wallet ID has not been backfilled**

`PersistentTxo.walletId` explicitly defaults to empty for rows migrated from the older schema, and the established restore path already resolves those legacy rows through their account relationship. The spend-reconciliation path can set `isSpent` and `spendingTransaction` without backfilling that field, so a confirmed conflicting spender can be present while this strict equality guard still discards the row. Because the point lookup uses the globally unique outpoint decoded from this wallet's unresolved asset-lock transaction, accepting the documented empty migration value cannot introduce an unrelated row; rejecting it leaves the restored conflict map empty and sends startup recovery back into the full proof wait this PR is intended to prevent.

```suggestion
guard let txo = try? backgroundContext.fetch(descriptor).first,
      (txo.walletId == walletId || txo.walletId.isEmpty),
      txo.isSpent,
      let spender = txo.spendingTransaction,
      txo.txid.count == 32,
      spender.txid.count == 32
else { continue }
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The current head resolves both carried findings by decoding unresolved inputs from each asset-lock row's authoritative transaction bytes and applying the persisted chain-lock boundary to restored spends. One blocking load-path issue remains: migrated spent TXOs with an empty denormalized wallet ID are discarded before their valid conflict evidence reaches Rust, reproducing the startup proof wait for existing wallets.

### Review provenance

🔴 1 blocking

<details>

```

- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift:5465-5470: Accept migrated TXOs whose wallet ID has not been backfilled
  `PersistentTxo.walletId` explicitly defaults to empty for rows migrated from the older schema, and the established restore path already resolves those legacy rows through their account relationship. The spend-reconciliation path can set `isSpent` and `spendingTransaction` without backfilling that field, so a confirmed conflicting spender can be present while this strict equality guard still discards the row. Because the point lookup uses the globally unique outpoint decoded from this wallet's unresolved asset-lock transaction, accepting the documented empty migration value cannot introduce an unrelated row; rejecting it leaves the restored conflict map empty and sends startup recovery back into the full proof wait this PR is intended to prevent.
```
</details>

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
thepastaclaw: Resolved in `1fcf166` — _Decode inputs from the asset-lock row instead of requiring a transaction row_ no longer present.

packages/rs-platform-wallet/src/wallet/asset_lock/sync/recovery.rs
thepastaclaw: Resolved in `1fcf166` — _Apply the restored chain-lock boundary to restored spends_ no longer present.

thepastaclaw (approved):

The exact head fixes the prior migrated-TXO ownership blocker by resolving ownership through the account relationship when the denormalized wallet ID is empty, with a focused Swift test exercising the real load callback. The restored conflict path also conservatively handles context discriminants, applies the chain-lock boundary, and has focused Rust coverage; no in-scope defects remain.

### Review provenance

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletPersistenceHandler.swift
thepastaclaw: Resolved in `f962c06` — _Accept migrated TXOs whose wallet ID has not been backfilled_ no longer present.

QuantumExplorer:
## Review — head `f962c06008`

Reviewed with the same multi-angle process as [#4406](https://github.com/dashpay/platform/pull/4406#issuecomment-5335126711). The diagnosis is right and well-evidenced (the on-device trace in the description is exactly the kind of proof these PRs should carry), the Rust screen's own guards are careful (verbatim discriminant decode, self-spend filter, conservative unknown-context handling), and the FFI plumbing checked out on decode/count/lifetime/memory grounds. But the review surfaced a false-condemn family in the evidence lifecycle that I'd resolve before this merges, plus one ABI-policy issue and a CI-breaker.

### The core problem: the evidence is a fossil, and it condemns 🔴

**1. The restored map is never invalidated and outranks live history for the whole session** (`recovery.rs:268`). It's written once at load, consulted *before* the history scan, and returns terminally on a hit. The PR body says "the fallback scan still runs and still wins whenever the history is populated" — it doesn't: the map wins unconditionally. Once catch-up sync has proven a recorded spender absent, every `resume_asset_lock` still condemns from the fossil. Four independent review angles converged on this.

**2. Nothing can ever demote the condemning verdict, and the boundary fallback upgrades it** (`recovery.rs:278`). The verdict is a persist-time context byte; the pinned upstream has no block-disconnect handling, so a spender recorded in-block during a chainlock stall and then reorged out stays `context=2` in the mirror forever — and the restored-path chainlock-boundary fallback promotes the stale height to `spender_chain_locked=true`. Worst case: lock LB's sibling LA lands in a non-chainlocked block, the app is killed, the reorg drops LA — every future launch condemns LB at the API's highest confidence tier, a host auto-discards it, and LB's already-broadcast transaction later confirms into an unclaimed asset lock. Restored evidence at `context=2` should arguably be retryable rather than terminal, and the chain-locked promotion should never run on a snapshot height.

**3. The evidence can also be silently blanked** (`PlatformWalletPersistenceHandler.swift:1173`, missed-conflict direction). `resolveInputOutpoint` on this branch is last-writer-wins: a stale re-relay of the dead lock at mempool context overwrites the confirmed spender's linkage and flips `isSpent` false, so the row fails the builder's gate and the screen is blind again next launch. Note #4406 makes exactly this assignment monotonic in the same function on its branch — this is a merge-ordering hazard: whichever lands second needs both the guard and this feature, ideally with a test pinning the combination.

These three share a root: the evidence bypassed `transactions()` and its promote/demote lifecycle. Which leads to —

**4. The altitude question.** The load path already solves this class of problem: `restore_unresolved_asset_lock_tx_records` projects persisted rows into `transactions()` so the existing machinery just works. Restoring the *spender* records the same way — one query wider — would feed the existing history scan with zero new FFI surface, and promotion/demotion would keep the evidence honest, dissolving findings 1–2 structurally. Neither the code nor the PR body mentions this alternative or why it was rejected. If there's a real blocker (spender records perturbing payment/balance reconstruction?), it should be stated at the struct or the precedence branch; if there isn't, I'd seriously consider the rewrite — it would also shrink this PR substantially.

### Also before merge

- **ABI: the new fields are inserted mid-struct in `WalletRestoreEntryFFI`** (`wallet_restore_types.rs:665`), which crosses as a bare pointer with no size/version tag — the exact shape the #4406 review moved to a size-negotiated extension, and mid-insertion also shifts three existing fields. In-tree builds are lockstep so this doesn't bite today, but nothing detects skew, and a stale-header pairing fabricates condemning evidence (fixed-size txids always parse; a garbage context byte of 2–3 is the only gate). Append at the tail with a size-negotiated read, or version the struct.
- **Four unused imports fail the wallet clippy lane** (`client_wallet_start_state.rs:13/:15`, `ffi persistence.rs:9/:61`) — verified with `cargo check`; the lane simply didn't run on this feature-branch base, so it detonates when this reaches `v4.2-dev`.

### Contract and coverage

- **The FFI evidence contract has drifted in both directions** (`wallet_restore_types.rs:522`): docs promise "different transaction only" but Swift emits self-spends (Rust re-filters; the load-path "restored conflicts" log overcounts on every healthy confirmed lock); docs promise rows "whether or not that spender ever reached a block" but the `txo.isSpent` gate structurally withholds non-in-block spenders — which also means the host is making the finality call the PR body says is Rust's, and the documented future "abandon cascade" reader can never be served by this channel on iOS.
- **The malformed-row skip is dead code** (`ffi persistence.rs:4871`): `Txid::from_slice` on a `[u8;32]` can't fail, so the branch, its log, and the "a bad row must not cost the user their wallet" contract are unreachable — and no real validation exists. Validate something real (all-zero txid?) or drop the claim.
- **Zero content coverage on the one cross-language contract this PR creates** (`AssetLockInputSpendRestoreTests.swift`): the Swift suite asserts only the row *count*; nothing pins which bytes land in `prev_txid` vs `spender_txid` — a swapped copy passes both suites and terminally condemns healthy locks at runtime. One read-back assertion closes it.
- The **Android gap is untracked** — third silently-divergent field in the same JNI struct literal, no TODO/issue; the Kotlin twin never gets scheduled from a struct-literal comment.

### Smaller items (in the findings report)

Nil-network bail silently dropping all evidence for legacy wallets; the ownership check dropping twin-wallet evidence with no log; two doc-comment splices (`PlatformWalletInfo` — the crate's central type — is now undocumented, its doc absorbed by `RestoredSpend`; same pattern in the FFI crate); `prev_txid` copied from the row's computed property instead of the outpoint key it was fetched by; stale 3-tuple field docs and a PR body that describes the superseded 4096-cap revision (the "no new test" caveat is also half-stale now — the Swift harness exists). In passing: `unresolvedAssetLockInputs` is a third fetch of rows the load path already holds, the context-discriminant decode is now inlined at three sites in the FFI crate, and the test's `outPointHex` helper reimplements `PersistentAssetLock.encodeOutPoint`.

### Bottom line

The gap is real, the on-device diagnosis is exemplary, and the screen-side guards are careful — but the evidence source chosen freezes at load, can't be demoted, gets promoted to chain-locked confidence, and can be silently blanked by a stale re-relay. I'd either move the fix down a level (restore spender records through the existing mechanism and delete the side-channel) or, if the side-channel stays, make restored evidence non-terminal unless chain-locked-fresh, fix the precedence so live history wins, and add the payload read-back test. The clippy imports and the mid-struct ABI insertion need fixing regardless.

QuantumExplorer:
## Fixes pushed — `f962c06008..c82369f146`

Pushed four commits addressing the [review](https://github.com/dashpay/platform/pull/4404#issuecomment-5338843189). All verified locally: `platform-wallet` 672/672, `platform-wallet-ffi` 275/275, Swift suite 344/344 (after a fresh mac-slice build so the regenerated header is what's tested), clippy `-D warnings` clean on both wallet crates, `cargo fmt` clean.

**`1809c60f5e` — live history outranks the restored snapshot.** The scan order is reversed (history first), a snapshot row whose spender live history has re-observed *unconfirmed* is treated as stale and skipped, and the chainlock-boundary promotion no longer applies to restored rows — a persisted height can't prove its block survived to be buried, so only the mirror's own observed chainlock context (`context=3`) claims that tier. The conflict still fires for restored in-block evidence, so the measured startup case keeps working; only the confidence flag is withheld. Three new/inverted tests pin this: `live_history_outranks_the_restored_snapshot`, `a_live_unconfirmed_sighting_retracts_the_restored_verdict`, and `restored_spend_below_the_chainlock_boundary_stays_unpromoted` (deliberate inversion of the old promotion test — the old assertion was the false-condemn vector). One residual is documented rather than fixed: a spender that was reorged out and never re-observed is indistinguishable from the load blind spot; that window closes only when the mirror learns to demote spend links.

**`bc327334c7` — FFI tail-append + real validation.** The two spend-row fields moved to the tail of `WalletRestoreEntryFFI` with the append-only discipline documented on the struct; the malformed-row branch now rejects zeroed txids (the shape a broken host actually produces) instead of guarding an infallible parse; the load log says "rows" not "conflicts"; the struct docs describe what the host actually sends (self-spends included, in-block only today); the context discriminants are shared named constants; and the identity-restore doc block is back on `build_wallet_identity_bucket`.

**`9e88c63e5b` — Swift evidence hardening.** `resolveInputOutpoint`'s `isSpent` is now monotonic — this is byte-for-byte the same guard the sweep-persistence branch (#4406) adds to the same line, so whichever merges second gets a clean no-op there. The emitted row's outpoint fields now come from the 36-byte key the fetch matched on (not the row's computed `txid`), ownership skips log instead of vanishing, and a legacy nil-network wallet keeps its evidence.

**`c82369f146` — payload read-back test** plus the fixture's outpoint encoding routed through `PersistentAssetLock.encodeOutPoint`.

**Deliberately not done, your call:**
- The altitude question (restoring spender *records* through `restore_unresolved_asset_lock_tx_records` and deleting the side-channel) — that's a design decision for you; the pushed fixes harden the side-channel enough that it's now a simplification question rather than a correctness one.
- The Kotlin twin — still untracked; I'd file an issue so the Android gap doesn't live only in the JNI struct-literal comment.
- The third `PersistentAssetLock` fetch on the load path — left as-is; it's bounded by lock count and your point-lookup comments show the shape is deliberate.
- The PR body still describes the superseded 4096-cap revision and the pre-harness "no new test" caveat — worth a refresh when you touch it next.

PR #4409, opened 2026-08-17
fix(drive-abci): roll back dropped state transitions on the proposing path

QuantumExplorer:

Hotfix (targeting a 4.1.1 release) for the mainnet evo1 stalls of 2026-08-14/15, after heights 415652 and 415661 (~2h each).

- Execution can write into the shared block transaction before failing: the address-input fee flow is apply-then-check, and the admission estimate can undershoot the actual metered fee for a `Shield` (the keyless commitment-tree append is skipped in estimation — dashpay/grovedb#812).
- A `Shield` funded between the estimated and actual fee passes validation, fails the coverage guard after its writes landed, and is dropped as `InternalError` → `TxAction::Removed` — with nothing rolling the writes back.
- The proposer gossips a block **without** the transition while advertising an app hash computed **with** its writes. No validator can reproduce that hash; every proposer whose mempool carries the transition burns its round. With a 100-validator quorum that is a full ~2h rotation, and the trigger is remotely repeatable by anyone at the cost of one Unshield.

The full fix (roll back on **all** nodes, closing the malicious-proposer variant too) changes what state a received block evaluates to, so it is protocol-v14-gated and rides #4408 / v4.2. This hotfix is the consensus-invisible half that can ship immediately.

## What was done?

- **Proposer-side rollback** in `process_raw_state_transitions_v0`: when `proposing_state_transitions == true`, wrap each executed transition in a GroveDB savepoint (`transaction.set_savepoint()`) and roll back if the result strips it from the block (`InternalError` / `UnpaidConsensusError`). The proposal then omits the transition *and* its app hash omits its writes — exactly what any v4.1.0 validator computes from that block, so mixed networks cannot diverge and no protocol-version gate is needed.
- **Validation path untouched** (`proposing_state_transitions == false`): no savepoint, no rollback. Rolling back there would be a consensus change; it activates at protocol v14 via #4408 instead.
- **Genesis height excluded** (`rollback_dropped_transitions = proposing && height != genesis_height`): the genesis re-proposal path relies on a single-savepoint discipline (init_chain sets one savepoint; each genesis round rewinds to it with one rollback), and savepoints of kept transitions stay on the stack because RocksDB exposes no pop-without-rollback. Excluding genesis keeps that path completely untouched — no drain loops, no compensating logic. At every other height each proposal round runs in a freshly started transaction that is committed (leftover savepoints are inert markers) or dropped at round end, so the residue is provably inert. The gap this leaves — no halt protection for a crafted genesis-block shield — is irrelevant to any running network.
- **Test-only fault hook** (`test_fault_injection::FAIL_NEXT_SUCCESSFUL_EXECUTION`): forces a successful transition to report `InternalError` after its writes applied — modeling the failure without depending on the fee-estimation constants.

Deployment properties: protection is incremental with rollout — each upgraded masternode immediately stops poisoning its own proposals, and a stall triggered mid-rollout ends at the first upgraded proposer's slot (expected ~1/f rounds at adoption fraction f) instead of a full rotation.

## How Has This Been Tested?

New `mod proposer_rollback_hotfix` in `shield/tests.rs` (`cargo test -p drive-abci --lib proposer_rollback_hotfix`); the first two are driven by the fault hook on a fully-funded shield so they are independent of fee constants, the third reproduces the real mainnet failure:

- `proposing_must_not_leave_state_of_dropped_transition` — the fix: a transition dropped while proposing leaves shielded pool, note count, and root hash untouched.
- `validating_must_behave_exactly_as_v4_1_0` — the consensus-invisibility guarantee: the validating path still behaves bit-for-bit like v4.1.0 (leak preserved: pool +5000, notes +2, hash changed). If this test ever fails because the deltas became zero, the hotfix has silently become a fork.
- `proposing_real_underfunded_shield_leaves_no_trace` — the real mainnet scenario, no fault hook: a shield funded one credit below the actual metered fee passes estimated-fee validation, fails the coverage guard at execution ("address-input fee not fully covered"), is dropped — and the root hash is unchanged. The match on the outcome keeps the test valid if fee constants shift.

Full `cargo test -p drive-abci --lib`: 2632 passed / 0 failed. `cargo clippy -p drive-abci --all-targets` clean.

## Breaking Changes

None — deliberately. The change affects only which blocks an upgraded node builds; every block, whoever built it, still evaluates identically on every validator. Interoperates with v4.1.0 nodes in both directions during rolling upgrade.

## Checklist:

### DIFF
--- a/packages/rs-drive-abci/src/execution/platform_events/state_transition_processing/process_raw_state_transitions/v0/mod.rs
+++ b/packages/rs-drive-abci/src/execution/platform_events/state_transition_processing/process_raw_state_transitions/v0/mod.rs
@@ -84,6 +84,33 @@ where
         let state_transition_container =
             self.decode_raw_state_transitions(raw_state_transitions, platform_version)?;
 
+        // PROPOSER-SIDE ONLY (consensus-invisible, hence no protocol-version gate): while
+        // building a proposal, wrap each executed state transition in a savepoint and roll
+        // back if its result strips it from the block (`TxAction::Removed`). Execution can
+        // write into the shared block transaction before failing (the address-input fee flow
+        // is apply-then-check), and without the rollback the gossiped block omits the
+        // transition while the advertised app hash includes its writes — no validator can
+        // reproduce the hash, and every proposer carrying the transition burns its round
+        // (mainnet evo1 stalls of 2026-08-14/15, after heights 415652 and 415661).
+        //
+        // The validation path (`proposing_state_transitions == false`) is deliberately
+        // untouched: rolling back there would change what state a received block evaluates
+        // to, which is a consensus change that must ride a protocol-version gate (it does,
+        // from v14). This proposer-side rollback only changes which blocks this node BUILDS —
+        // the published block and app hash are exactly what any un-upgraded validator
+        // computes from that block, so mixed networks cannot diverge.
+        //
+        // The genesis height is excluded because its re-proposal path relies on a
+        // single-savepoint discipline: init_chain sets one savepoint, and each genesis round
+        // rewinds to it with one `rollback_to_savepoint()` (see prepare_proposal /
+        // process_proposal). Savepoints of KEPT transitions stay on the stack — RocksDB
+        // exposes no pop-without-rollback — and extra savepoints on the genesis transaction
+        // would redirect that rewind. At every other height each proposal round runs in a
+        // freshly started transaction that is either committed (leftover savepoints are inert
+        // markers) or dropped when the round ends, so the residue can affect nothing.
+        let rollback_dropped_transitions =
+            proposing_state_transitions && block_info.height != self.config.abci.genesis_height;
+
         let mut processing_result = StateTransitionsProcessingResult::default();
 
         for decoded_state_transition in state_transition_container.into_iter() {
@@ -123,26 +150,9 @@ where
                             );
                         }
 
-                        // PROPOSER-SIDE ONLY (consensus-invisible, hence no protocol-version
-                        // gate): when building a proposal, mark the state before this
-                        // transition. Execution can write into the shared block transaction
-                        // before failing (the address-input fee flow is apply-then-check), and
-                        // a transition whose result strips it from the block
-                        // (`TxAction::Removed`) must leave no trace in the state the proposal's
-                        // app hash is computed over — otherwise the gossiped block omits the
-                        // transition while the advertised app hash includes its writes, no
-                        // validator can reproduce the hash, and every proposer carrying the
-                        // transition burns its round (mainnet evo1 stalls of 2026-08-14/15,
-                        // after heights 415652 and 415661).
-                        //
-                        // The validation path (`proposing_state_transitions == false`) is
-                        // deliberately untouched: rolling back there would change what state a
-                        // received block evaluates to, which is a consensus change that must
-                        // ride a protocol-version gate (it does, from v14). This proposer-side
-                        // rollback only changes which blocks this node BUILDS — the published
-                        // block and app hash are exactly what any un-upgraded validator
-                        // computes from that block, so mixed networks cannot diverge.
-                        if proposing_state_transitions {
+                        // Mark the state we can return to if this transition's result strips
+                        // it from the block (see `rollback_dropped_transitions` above).
+                        if rollback_dropped_transitions {
                             transaction.set_savepoint();
                         }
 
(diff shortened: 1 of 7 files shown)


thepastaclaw:
🔍 Review in progress — actively reviewing now (commit 3978bdf9)

QuantumExplorer:
Reviewed.

thepastaclaw (commented):

The proposer-only, non-genesis rollback correctly removes state writes for transitions omitted from a proposal while preserving the existing validator and genesis behavior. One maintainability suggestion remains: make the rollback classification exhaustive so future execution-result variants cannot be classified elsewhere without an explicit savepoint decision here.
Source: reviewers codex/general, codex/security-auditor, and codex/rust-quality used backend model gpt-5.6-sol; final verifier used backend model gpt-5.6-sol. openclaw-agent/cliproxy/gpt-5.6-sol was orchestration-only and is not reviewer evidence.

### Review provenance

🟡 1 suggestion(s)

<details>

```

In `packages/rs-drive-abci/src/execution/platform_events/state_transition_processing/process_raw_state_transitions/v0/mod.rs`:
- [SUGGESTION] packages/rs-drive-abci/src/execution/platform_events/state_transition_processing/process_raw_state_transitions/v0/mod.rs:203-224: Keep proposal-removal classification exhaustive
  This match defines the critical correspondence between execution outcomes removed from a proposal and outcomes whose transaction writes must be rolled back, but the wildcard arm does not make Rust enforce that correspondence. Other exhaustive matches, including the metrics match below and the `TxAction` match in `prepare_proposal`, would catch a new enum variant, but updating those matches would not require the maintainer to classify its savepoint behavior here. Enumerate every retained or non-writing outcome so any future `StateTransitionExecutionResult` addition also requires an explicit rollback decision at this point.
```
</details>

packages/rs-drive-abci/src/execution/platform_events/state_transition_processing/process_raw_state_transitions/v0/mod.rs
```diff
@@ -137,6 +185,44 @@ where
...
+                                }
+                                _ => {
+                                    // The transition stays in the block, so its writes stay.
+                                    // Its savepoint is intentionally left on the stack (see
+                                    // `rollback_dropped_transitions` above: no
+                                    // pop-without-rollback exists, and at non-genesis heights
+                                    // the residue is inert).
+                                }
+                            }
+                        }
```
thepastaclaw, line 237: **🟡 Suggestion: Keep proposal-removal classification exhaustive**

This match defines the critical correspondence between execution outcomes removed from a proposal and outcomes whose transaction writes must be rolled back, but the wildcard arm does not make Rust enforce that correspondence. Other exhaustive matches, including the metrics match below and the `TxAction` match in `prepare_proposal`, would catch a new enum variant, but updating those matches would not require the maintainer to classify its savepoint behavior here. Enumerate every retained or non-writing outcome so any future `StateTransitionExecutionResult` addition also requires an explicit rollback decision at this point.

```suggestion
if rollback_dropped_transitions {
    match &execution_result {
        StateTransitionExecutionResult::InternalError(_)
        | StateTransitionExecutionResult::UnpaidConsensusError(_) => {
            // This transition will be stripped from the proposal
            // (`TxAction::Removed`), so none of its writes may remain
            // in the state the app hash is computed over. A rollback
            // failure means the proposal can no longer match the
            // block — fail it rather than continue on leaked state.
            transaction.rollback_to_savepoint().map_err(|e| {
                drive::grovedb::error::Error::StorageError(RocksDBError(e))
            })?;
        }
        StateTransitionExecutionResult::PaidConsensusError { .. }
        | StateTransitionExecutionResult::SuccessfulExecution { .. }
        | StateTransitionExecutionResult::NotExecuted(_) => {
            // No rollback is required for the currently retained or
            // non-writing outcomes. The savepoint is intentionally left
            // on the stack because RocksDB has no pop-without-rollback.
        }
    }
}
```

<sub>source: ['codex']</sub>

QuantumExplorer: Done in 3978bdf955 — wildcard replaced with explicit arms for SuccessfulExecution/PaidConsensusError (kept, savepoint left) and NotExecuted (delayed, nothing written), with a comment stating the match must mirror prepare_proposal's TxAction classification.

shumkov (approved):
We need to refactor this later to get rid of #cfg(test) inside the function, but it's fine for a hot fix

Issue #4354, opened 2026-08-10
Mainnet shielded sync fails repeatedly: bootstrap list dials known-dead evonodes, and one exhausted chunk voids the whole sync pass


bfoss765:
## Summary

On mainnet, shielded wallet sync dies repeatedly with TLS transport errors ("transport error" from tonic) against evonodes with expired certificates. The TLS layer itself is behaving correctly — `rs-dapi-client` classifies these failures as retryable, bans the node for 60s, and fails over across up to 4 distinct nodes per request. The real problems are two things upstream of TLS:

## Problem 1: the bootstrap address list ships nodes it knows are dead

`default_address_list_for_network` loads **all** mainnet evo seeds from the static seed file, but the seed file itself carries a per-node TLS-health column that the SDK throws away. At the time we measured:

- 340 mainnet evo seeds total
- 277 valid
- 58 flagged `plat_ssl=expired`
- 5 no-handshake

So ~18.5% of the addresses the client dials are known-bad **according to data the SDK itself ships**. Testnet is 30/30 valid, which is why this never surfaces in testnet QA and only bites real mainnet users.

**Ask:** filter the bootstrap list to valid-and-reachable nodes using the health column already present in the seed file (or at minimum de-prioritize flagged nodes).

## Problem 2: shielded sync is fail-fast where everything else is one-shot

A one-shot request (e.g. an identity fetch) that exhausts its retry budget fails alone. The shielded chunk stream is different: it runs ~16 concurrent chunk fetches, and the **first chunk** that exhausts its 4-node budget terminates the stream, which propagates up to `fail_all_wallets` — marking every registered wallet failed and discarding the entire pass. Because the sync cooldown is only stamped on **success**, the failed pass re-runs immediately, producing repeated failures in a single session.

With 18.5% known-bad nodes in the pool, the probability that at least one of 16 chunks draws 4 bad/unlucky nodes in a row is high enough that mainnet shielded sync fails outright while individual one-shot calls appear healthy.

**Ask:** make a single failed chunk retry at the stream level (re-draw nodes, resume the stream) instead of voiding the whole pass — and consider stamping a shorter cooldown on failure so a failing pass doesn't hot-loop.

## What we are NOT asking for

We are not proposing any change to certificate validation. Failover on cert errors already works; the fixes above are list hygiene and stream-level retry.

## Environment / evidence

[...]

PR #4420, opened 2026-08-19
fix(dashmate): detect and report gateway certificate problems

shumkov:

A live scan of all 353 registered mainnet evonodes found **88 (25%) serving an expired TLS
certificate**, unreachable by any standards-compliant client. Some had been dark for over a
year. PoSe does not probe port 443, so those nodes keep their score at 0 and are paid on the
same cadence as healthy ones — the tooling was the only place this could surface, and it was
blind.

Two independent reasons it stayed invisible:

1. **`dashmate doctor` could not report a certificate problem for any provider.** Five defects
   stacked on top of each other, so no path reached the operator.
2. **Nothing ever checked what the gateway actually serves.** Every check was file-based or
   provider-API-based, so a certificate that renewed on disk but never reached Envoy — a missed
   reload, an un-copied bundle — read as perfectly healthy everywhere.

Related: #4354 (mainnet shielded wallet sync fails because the SDK bootstrap list dials dead
evonodes).

### Doctor could not report certificate problems at all (`analyseConfigFactory.js`, `collectSamplesTaskFactory.js`)

- `validateZeroSslCertificate` was not awaited, so `error`/`data` destructured off a Promise and
  were both `undefined` — ZeroSSL nodes never produced a problem, whatever their state.
- `{block.cyanBright}` is not a chalk style. The message table is one eagerly-built object
  literal, so this threw `Unknown Chalk style: block` for **every** certificate error of **every**
  provider, including Let's Encrypt renewal failures.
- Two ZeroSSL messages dereferenced `ssl?.data?.certificate.expires` / `.common_name` with the
  chaining stopping at `data`, throwing for any error raised before the certificate is fetched.
- `ZEROSSL_ERRORS` and `LETSENCRYPT_ERRORS` share `CERTIFICATE_EXPIRES_SOON`, so in one literal
  the later Let's Encrypt entry silently overwrote the ZeroSSL one. Messages are now grouped per
  provider and selected by the configured provider.
- Three metrics samples did not await `fetchTextOrError`, storing unresolved Promises that
  serialised into diagnostic archives as `{}`.

### Report what the gateway actually serves (new)

- `src/ssl/probeServedCertificate.js` — opens a TLS connection to the gateway's own listener and
  reports the certificate it presents. Identity is judged against `externalIp` rather than the
  address dialled, and separately from the chain verdict, because a connection surfaces only its
  first verification failure. Has an absolute deadline (the socket `timeout` option is an
  inactivity timer that does not close the connection) and flattens the peer certificate to plain
  values (a verified chain is circular and breaks JSON serialisation and the obfuscation pass).
- `src/ssl/readCertificateBundle.js` — reads the gateway bundle's server certificate via
  `crypto.X509Certificate`, skipping CA blocks so a reversed operator-supplied bundle is not
  compared against an intermediate.
- `src/doctor/analyse/analyseGatewayCertificateFactory.js` — reports: expired with renewal not
  reaching the gateway (restart) vs. renewal itself failing (helper logs); renewed but not picked
  up while still valid; untrusted chain; wrong address; gateway not answering TLS. Validity is
  judged against `samples.date`, not analysis time, so an archive opened days later does not
  report every 6-day certificate as dead.
- Inbound port 80 is collected and reported **only alongside a certificate problem**. The port is
  bound for the seconds a validation takes, so an external check finds it closed on healthy nodes
  too — in the scan, 52 valid, actively-renewing nodes looked identical to 4 expired ones.

### Renewal that never reaches the gateway

- `dashmate ssl obtain` wrote the certificate files and exited; only the helper's scheduled
  renewal signalled the gateway. It now reloads the gateway, skipping with a message when the
  gateway is not running.
- `validateLetsEncryptCertificateFactory.js` already computed `isCertificatePairInstalled` and
  never read it. Now returned as `CERTIFICATE_NOT_INSTALLED`.

### ZeroSSL remediation

Every message told operators to run `dashmate ssl obtain`, which renews with the provider already
configured. Whether that works depends on the operator's ZeroSSL plan, which dashmate cannot see,
so both renewing and switching to Let's Encrypt are offered, with the cost stated (IP certificates
are 6-day and auto-renewing). Where the ZeroSSL API explained the failure — exhausted limit,
unpaid invoice, rejected key — that explanation is now the description; an API failure carrying no
message was previously dropped entirely. Also corrects a suggestion to run
`dashmate ssl zerossl obtain`, which is not a command that exists.

## How Has This Been Tested?

`yarn workspace dashmate test:unit` — **331 passing**, lint clean.

New/updated specs:

- `test/unit/ssl/probeServedCertificate.spec.js` — probe against real `tls.createServer`
  instances: served/expired certificates, identity judged against `externalIp` rather than the
  probed address, identity reported separately from the chain when both fail, and the degradation
  cases (nothing listening; a peer that accepts TCP and never handshakes; a peer that trickles
  bytes to keep an inactivity timer alive).
- `test/unit/doctor/analyse/analyseGatewayCertificateFactory.spec.js` — one case per severity row
  including the negatives, plus judging expiry against sample-collection time and port 80 being
  silent on a healthy node.
- `test/unit/doctor/analyse/analyseConfigFactory.spec.js` — pins that every provider's certificate
  errors produce a problem rather than throwing, plus the ZeroSSL remediation content.
- `test/unit/listr/tasks/doctor/collectSamplesTaskFactory.spec.js` — end-to-end sample collection,
  including a real TLS server so the probe wiring is exercised.
- `test/unit/ssl/letsencrypt/validateLetsEncryptCertificateFactory.spec.js` — a renewed
  certificate that was never installed for the gateway.
- `test/unit/commands/ssl/obtain.spec.js` — gateway reload after obtaining, and skipped when the
  gateway is not running.

Every fix was written test-first and observed failing against the unfixed code before the fix
landed. Certificate fixtures are generated at test time rather than committed, so no test rots on
a date nobody chose.

## Breaking Changes

None. `dashmate doctor` reports problems it previously could not; `dashmate ssl obtain` gains a
reload step. No configuration, schema, or wire format changes.

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Added comprehensive gateway certificate diagnostics, including expiry, identity, trust-chain, reachability, and stale-certificate checks.
  - Added detection for certificates issued but not installed in the gateway.
  - Added validation comparing gateway-served certificates with certificates stored on disk.
  - Added clearer SSL troubleshooting guidance for Let’s Encrypt, ZeroSSL, and file-based certificates, including ACME port checks.

- **Improvements**
  - Automatically reloads a running gateway after successfully obtaining an SSL certificate.
  - SSL and metrics checks now complete more reliably during diagnostics.

### DIFF
--- a/packages/dashmate/src/listr/tasks/doctor/collectSamplesTaskFactory.js
+++ b/packages/dashmate/src/listr/tasks/doctor/collectSamplesTaskFactory.js
@@ -105,7 +105,10 @@ export default function collectSamplesTaskFactory(
                       const {
                         error,
                         data,
-                      } = validateZeroSslCertificate(config, Certificate.EXPIRATION_LIMIT_DAYS);
+                      } = await validateZeroSslCertificate(
+                        config,
+                        Certificate.EXPIRATION_LIMIT_DAYS,
+                      );
 
                       obfuscateObjectRecursive(data, (_field, value) => (typeof value === 'string' ? value.replaceAll(
                         process.env.USER,
@@ -311,7 +314,7 @@ export default function collectSamplesTaskFactory(
 
               const url = `http://${config.get('platform.drive.tenderdash.rpc.host')}:${config.get('platform.drive.tenderdash.rpc.port')}/metrics`;
 
-              const result = fetchTextOrError(url);
+              const result = await fetchTextOrError(url);
 
               ctx.samples.setServiceInfo('drive_tenderdash', 'metrics', result);
             }
@@ -322,7 +325,7 @@ export default function collectSamplesTaskFactory(
 
               const url = `http://${config.get('platform.drive.abci.metrics.host')}:${config.get('platform.drive.abci.metrics.port')}/metrics`;
 
-              const result = fetchTextOrError(url);
+              const result = await fetchTextOrError(url);
 
               ctx.samples.setServiceInfo('drive_abci', 'metrics', result);
             }
@@ -333,7 +336,7 @@ export default function collectSamplesTaskFactory(
 
               const url = `http://${config.get('platform.gateway.metrics.host')}:${config.get('platform.gateway.metrics.port')}/metrics`;
 
-              const result = fetchTextOrError(url);
+              const result = await fetchTextOrError(url);
 
               ctx.samples.setServiceInfo('gateway', 'metrics', result);
             }
--- a/packages/dashmate/src/listr/tasks/doctor/collectSamplesTaskFactory.js
+++ b/packages/dashmate/src/listr/tasks/doctor/collectSamplesTaskFactory.js
@@ -233,23 +233,22 @@ export default function collectSamplesTaskFactory(
                       : null;
                   }
 
-                  obfuscateObjectRecursive(result, (_field, value) => (typeof value === 'string' ? value.replaceAll(
-                    process.env.USER,
-                    hideString(process.env.USER),
-                  ) : value));
-
                   ctx.samples.setServiceInfo('gateway', 'servedCertificate', result);
                 },
               },
               {
-                // Both certificate providers validate this node over inbound port 80.
-                enabled: () => config.get('platform.enable'),
-                title: 'ACME HTTP validation port',
+                // Both obtainable providers reach this node on port 80 to prove it controls
+                // its address before issuing: Let's Encrypt over ACME, ZeroSSL over its own
+                // verification server. A self-signed or operator-supplied certificate is
+                // never validated, so the port means nothing for those.
+                enabled: () => config.get('platform.enable')
+                  && ['zerossl', 'letsencrypt'].includes(config.get('platform.gateway.ssl.provider')),
+                title: 'Certificate validation port',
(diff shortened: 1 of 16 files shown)


thepastaclaw:
🔍 Review in progress — actively reviewing now (commit 49cfba62)

packages/dashmate/src/listr/tasks/doctor/collectSamplesTaskFactory.js
```diff
@@ -187,6 +192,66 @@ export default function collectSamplesTaskFactory(
                   }
                 },
               },
+              {
+                // Every other certificate check reads a file or the provider's API, so a
+                // certificate that was renewed on disk but never reached the gateway looks
+                // healthy to all of them. This connects to the gateway and records what it
+                // actually serves. Doctor is run by an operator on the node, so the gateway's
+                // listener is reached at the address it is published on.
+                enabled: () => config.get('platform.enable')
```
shumkov: we need to test it's actually works on Linux?

PR #4248, opened 2026-07-28
fix(dashmate): stop concurrent commands reverting each other's config changes

shumkov:
## Issue being fixed or feature implemented

Closes #4242.

`Config`'s constructor delegates to `setOptions()`, which marks the config changed — so **every config was dirty the moment it was read off disk**. `BaseCommand.finally()` persists the whole config file whenever anything is dirty, so every successful command, *including pure readers* like `dashmate config get` and `dashmate core cli`, rewrote `config.json` from its own load-time snapshot.

That turns any overlapping pair of commands into a lost update:

1. A reader loads the config at T0.
2. `dashmate config set …docker.image X` loads, writes, prints success and exits 0 at T1.
3. The reader finishes and writes its **T0** snapshot at T2.

The setting silently reverts even though `config set` reported success. Reported from a real rolling masternode image-pin update, where some nodes kept the old image digest.

Two things made it hard to spot:

- **In single-process use the rewrite is byte-identical** (verified against a real 58 KB, 4-config `config.json`), so it never shows up outside real concurrency.
- The worst window is not a CLI command at all — `scripts/helper.js` reads config **once at startup** and writes it back on certificate renewal months later. Its guard is even commented *"Persist config if it was migrated"*, and this bug made it fire on **every** startup.

## What was done?

**Hydration is no longer a mutation.** `Config`'s constructor resets the flag after the initial `setOptions()`. `ConfigFile` already did exactly this; `Config` was just missing it. This alone means read-only commands write nothing — they cannot clobber anyone.

**Changing configuration reads and saves as one locked step.** Reading when a command starts and saving when it exits means the state written may already be out of date. `ConfigFileJsonRepository.update(mutator)` takes the lock, reads the current state, applies the change and saves — the state handed to the mutator is read *after* the lock is held, so there is nothing stale to write:

```js
configFileRepository.update((configFile) => {
  configFile.getConfig(name).set(path, value);
});
```

Used by `config set`, `create`, `remove`, `default` and `group default`. Two commands changing different options now **both succeed**.

**Commands that reconfigure a node hold the lock for their run.** `setup`, `reset`, `group reset` and `ssl obtain` change configuration repeatedly while doing long, partly irreversible work, so a single locked step does not fit. They declare `static mutatesConfig = true`; `BaseCommand.init()` takes the lock *before* reading and releases it in `finally()`. Reading inside the lock is what makes their state current.

Read-only commands never take the lock, so a node can still be inspected while it is being set up.

**Supporting details that matter:**

[...]

PR #4421, opened 2026-08-19
fix(dashmate): deliver renewed certificates to the running gateway

shumkov:

Renewed gateway certificates could stop reaching a running Envoy.

`bundle.crt` and `private.key` are bind-mounted into the gateway container as
individual files:

```yaml
- ${DASHMATE_HOME_DIR}/${CONFIG_NAME}/platform/gateway/ssl/bundle.crt:/etc/ssl/bundle.crt:ro
- ${DASHMATE_HOME_DIR}/${CONFIG_NAME}/platform/gateway/ssl/private.key:/etc/ssl/private.key:ro
```

A file bind mount follows the inode, not the path. Since #4248 the certificate
was installed by writing a replacement file and renaming it over the old one,
which by definition creates a new inode and moves the name onto it. The running
container stays pinned to the inode it mounted at startup — and the mount itself
keeps that inode alive — so the renewal never becomes visible inside the
container. Envoy uses static `tls_certificates` filenames with no SDS, the only
refresh in the renewal path is `kill -SIGHUP 1`, and nothing recreates the
container. A `SIGHUP` therefore re-reads the same stale file.

The failure is silent: both files on disk are current, so `isCertificatePairInstalled`
(which compares two on-disk files) sees nothing wrong. The node would serve the
previous certificate until it expired.

Separately, `dashmate ssl obtain` never reloaded the gateway at all — only the
scheduled renewal signalled Envoy. An operator could run the command dashmate
points them to, see it report success, and find nothing changed on the wire.
That one is independent of #4248 and is present today on every version.

## What was done?

**Install the certificate pair in place again**, so the inode the container holds
is the one that receives the renewal. This removes the temp-file apparatus added
in #4248 from `saveCertificateTask`: temp-file sweeping, the `graceful.on('exit')`
cleanup hook, the rollback path, and the certificate mode juggling.

What that apparatus protected against was a torn write during renewal. Concurrent
writers — the scheduled renewal and a manual `ssl obtain` — are already serialized
by the configuration lock introduced in the same PR, which spans the whole obtain
on both paths (`renewCertificate.js` acquires it; `ObtainCommand` sets
`mutatesConfig = true`). What remained was an unclean shutdown or `ENOSPC` landing
inside a sub-millisecond write that happens once every few days. Weighed against a
delivery failure that is silent, affects every node, and needs a container
recreate to clear, that trade did not hold up.

Kept from #4248:

- the configuration lock, untouched
- the private key permission repair (`& 0o700`), which fixes keys left group- and
  world-readable by older installs

Writing in place cannot open a key an owner hardened to `0400` — rename sidestepped
that by never touching the existing file — so the owner write bit is restored around
the write and the chosen mode put back straight after. Without this, renewal would
have begun failing with `EACCES` on hardened nodes.

**Reload the gateway after `dashmate ssl obtain`.** Skipped when the certificate was
already current, and when the gateway is not running, so first-time setup before
`dashmate start` is unaffected.

## How Has This Been Tested?

Two regression tests, red before the change and green after.

`saveCertificateTask.spec.js` — pins the property the bind mount depends on:

```
should install a renewal into the files the gateway already has mounted
  AssertionError: expected 422747267 to equal 422747265
```

`obtain.spec.js` — pins the missing reload:

```
should reload the gateway after obtaining a certificate
  AssertionError: expected stub to have been called exactly once with
  arguments { … }, 'gateway', 'kill -SIGHUP 1'
```

The inode assertion is the durable one: a future atomic-install change cannot
reintroduce this without the suite going red and explaining why.

Full dashmate unit suite: **291 passing**. Lint clean (0 errors).

**Scope of what was verified.** The test demonstrates that the previous install
created a new inode. That a container's file bind mount then goes stale was
reasoned from Linux mount semantics (moby/moby#6011) rather than measured against
a running gateway container. A containerised check comparing
`stat -c %i` on the host against `stat -c %i` inside the gateway would close that
gap and is worth doing, but the fix does not depend on it: writing in place is
the behaviour every currently healthy node on 4.1.x is already running.

## Breaking Changes

None. No configuration, template or compose changes, so no config migration and
no different behaviour for an operator mid-upgrade.

Note that `4ada750769` is in no released tag — it exists only on `v4.2-dev` and
has never run in production. Restoring the in-place write returns this path to
the code currently running across the fleet.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * SSL certificate renewals now automatically reload the active gateway so updated certificates take effect immediately.
  * Gateway reloads are skipped when the gateway is stopped, with a status message shown.

* **Bug Fixes**
  * Certificate and key files are updated in place during renewal while preserving existing file permissions, improving compatibility with mounted certificate files.
  * Private key permissions are restored securely if an update fails.

### DIFF
--- a/packages/dashmate/src/commands/ssl/obtain.js
+++ b/packages/dashmate/src/commands/ssl/obtain.js
@@ -56,6 +56,7 @@ Certificate will be renewed if it is about to expire (see 'expiration-days' flag
     obtainLetsEncryptCertificateTask,
     configFileRepository,
     configFile,
+    dockerCompose,
   ) {
     const provider = providerFlag || config.get('platform.gateway.ssl.provider');
 
@@ -87,6 +88,25 @@ Certificate will be renewed if it is about to expire (see 'expiration-days' flag
         {
           title: taskTitle,
           task: () => task(config, taskOptions),
+        },
+        {
+          // Envoy reads the certificate files once at startup, so a gateway that
+          // is already up keeps serving the previous certificate until it is
+          // told to reload. Without this the command reports success while
+          // nothing changes on the wire.
+          title: 'Reload gateway',
+          skip: async (ctx) => {
+            if (!ctx.certificateSaved) {
+              return 'Certificate is already up to date';
+            }
+
+            if (!await dockerCompose.isServiceRunning(config, 'gateway')) {
+              return 'Gateway is not running';
+            }
+
+            return false;
+          },
+          task: () => dockerCompose.execCommand(config, 'gateway', 'kill -SIGHUP 1'),
         },
       ],
       {
--- a/packages/dashmate/src/commands/ssl/obtain.js
+++ b/packages/dashmate/src/commands/ssl/obtain.js
@@ -94,12 +94,14 @@ Certificate will be renewed if it is about to expire (see 'expiration-days' flag
           // is already up keeps serving the previous certificate until it is
           // told to reload. Without this the command reports success while
           // nothing changes on the wire.
+          //
+          // This runs whenever the gateway is up, including when the obtain
+          // wrote no new files: providers install the pair by different routes,
+          // and nothing on disk reveals which certificate Envoy currently
+          // holds, so an obtain that skipped the write is also how an operator
+          // retries a reload that failed earlier.
           title: 'Reload gateway',
-          skip: async (ctx) => {
-            if (!ctx.certificateSaved) {
-              return 'Certificate is already up to date';
-            }
-
+          skip: async () => {
             if (!await dockerCompose.isServiceRunning(config, 'gateway')) {
               return 'Gateway is not running';
             }
--- a/packages/dashmate/src/listr/tasks/ssl/saveCertificateTask.js
+++ b/packages/dashmate/src/listr/tasks/ssl/saveCertificateTask.js
@@ -1,7 +1,6 @@
 import { Listr } from 'listr2';
 import path from 'path';
 import fs from 'fs';
-import graceful from 'node-graceful';
 
 /**
  * @param {HomeDir} homeDir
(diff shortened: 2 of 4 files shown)


thepastaclaw:
ℹ️ Review skipped (commit 63c47f86)
_Last checked: 2026-08-19 18:31 UTC_

packages/dashmate/src/commands/ssl/obtain.js
```diff
@@ -88,6 +89,25 @@ Certificate will be renewed if it is about to expire (see 'expiration-days' flag
...
+          skip: async (ctx) => {
+            if (!ctx.certificateSaved) {
+              return 'Certificate is already up to date';
+            }
+
+            if (!await dockerCompose.isServiceRunning(config, 'gateway')) {
+              return 'Gateway is not running';
+            }
+
+            return false;
```
thepastaclaw: **🔴 Blocking: Do not gate the gateway reload on the transient save marker**

The reload only runs when `ctx.certificateSaved` was set by `saveCertificateTask`, but the ZeroSSL pipeline writes `private.key` and `bundle.crt` directly in `obtainZeroSSLCertificateTaskFactory.js` and never sets this property. A successful `dashmate ssl obtain --provider=zerossl` therefore skips the reload and leaves a running Envoy serving its previously loaded certificate. The transient marker also prevents recovery after a Let's Encrypt reload failure: once the pair has been written, a normal retry sees it as installed, does not set the marker, and reports success without retrying SIGHUP. The on-disk comparison cannot determine which certificate Envoy currently has loaded, so reload every running gateway after a successful obtain, or persist reload-pending state durably.

```suggestion
          skip: async () => {
            if (!await dockerCompose.isServiceRunning(config, 'gateway')) {
              return 'Gateway is not running';
            }

            return false;
          },
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The in-place certificate writes preserve the bind-mounted inodes, but the new reload step does not work for the supported ZeroSSL flow and cannot recover from a previously failed reload. Because those paths can still report success while the running gateway serves the stale certificate, changes are required.
Source: reviewers codex general (gpt-5.6-sol) and codex security-auditor (gpt-5.6-sol); final verifier codex (gpt-5.6-sol). openclaw-agent/cliproxy/gpt-5.6-sol was orchestration-only and is not reviewer evidence.

### Review provenance

🔴 1 blocking

<details>

```

In `packages/dashmate/src/commands/ssl/obtain.js`:
- [BLOCKING] packages/dashmate/src/commands/ssl/obtain.js:98-107: Do not gate the gateway reload on the transient save marker
  The reload only runs when `ctx.certificateSaved` was set by `saveCertificateTask`, but the ZeroSSL pipeline writes `private.key` and `bundle.crt` directly in `obtainZeroSSLCertificateTaskFactory.js` and never sets this property. A successful `dashmate ssl obtain --provider=zerossl` therefore skips the reload and leaves a running Envoy serving its previously loaded certificate. The transient marker also prevents recovery after a Let's Encrypt reload failure: once the pair has been written, a normal retry sees it as installed, does not set the marker, and reports success without retrying SIGHUP. The on-disk comparison cannot determine which certificate Envoy currently has loaded, so reload every running gateway after a successful obtain, or persist reload-pending state durably.
```
</details>

packages/dashmate/src/commands/ssl/obtain.js
shumkov: Confirmed both halves and fixed in 3705cbafca.

**ZeroSSL:** verified — `saveCertificateTask` is injected into `obtainZeroSSLCertificateTaskFactory` but never invoked; the flow writes `bundleFilePath`/`privateKeyFilePath` itself, so the marker was never set and the reload always skipped. (Those writes are already in place, so the inode half of this PR was never a ZeroSSL problem — only the reload was.)

**Recovery:** also verified — LE skips the save on `certificateValid && isCertificatePairInstalled`, ZeroSSL on `isBundleFilePresent`, so a re-run after a failed reload set no marker and sent no signal while reporting success.

Took the suggested `skip`, and additionally removed `ctx.certificateSaved` from `saveCertificateTask` since it no longer has a reader.

Three tests, red before and green after:

- `should reload the gateway after obtaining a certificate`
- `should reload the gateway after obtaining a ZeroSSL certificate`
- `should reload the gateway when the certificate was already installed`

The ZeroSSL one pins the reported defect; the third pins recovery and replaces an earlier test that asserted the opposite.

Accepted trade: an obtain that changes nothing now costs a gateway hot restart. That seems right for a manual command, and it makes re-running it the way to retry a failed reload.

packages/dashmate/src/listr/tasks/ssl/saveCertificateTask.js
```diff
@@ -30,69 +29,34 @@ export default function saveCertificateTaskFactory(homeDir) {
...
-            unsubscribe();
+          // An owner who hardened the key to 0400 has removed the write bit that
+          // writing in place needs, so restore it for the write and put the
+          // chosen mode back straight after.
+          if (keyExists) {
+            fs.chmodSync(keyFile, 0o600);
           }
 
+          fs.writeFileSync(keyFile, ctx.privateKeyFile, { encoding: 'utf8', mode: keyMode });
+          fs.chmodSync(keyFile, keyMode);
```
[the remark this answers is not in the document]
shumkov: Taking half of this and declining half, in 63c47f8635.

**Taken — the mode restore.** This was a real defect, and mine: writing in place needs the owner write bit, so a key hardened to `0400` is loosened to `0600` *before* the write, and a write that threw skipped the restore and left it that way. Now restored in a `finally` so it happens on both paths. Test, red before and green after:

```
should keep a hardened private key mode when the write fails
  AssertionError: expected 384 to equal 256
```

**Declined — reinstating the read-and-restore-both-files rollback.** That apparatus was removed deliberately in this PR, not overlooked.

- The window is one sub-millisecond in-place write of ~4 KB, occurring once per renewal. Reaching a mismatched pair needs `ENOSPC` or a kill inside it.
- The resulting failure is loud and recoverable: the gateway refuses to start, the operator sees it immediately, and `ssl obtain --force` repairs it. That is the opposite profile from the bug this PR fixes, which was silent, hit every node, and needed a container recreate.
- The previous code did not actually provide the property either. It performed two separate renames, so a reader landing between them saw the new certificate with the old key exactly the same way. The rollback only covered the second rename *throwing*, not the interleaving.
- Concurrent Dashmate writers are already serialised by the configuration lock, which spans the whole obtain on both the scheduled and manual paths. Envoy is not a lock participant, but the write is followed immediately by the reload, so it reads after the write completes.

Happy to revisit if someone has seen a torn pair in the field — it is a complexity-versus-small-risk call rather than a technical constraint, and an in-place restore would preserve the bind-mounted inode fine.

The `ast-grep` path-traversal warning attached to this is a false positive: the path is built from the Dashmate home directory and the config name, not from request-derived input.

PR #4423, opened 2026-08-19
fix: restore four Swift-parity guards in the Kotlin SDK, plus two DPNS marketplace defects

bfoss765:
Six audit-verified fixes in one batch.

**Items 1-4 are Swift-parity restorations**: each landed in the Swift SDK through review and was never ported to Kotlin, so the Android persister has been running without a guard its own reference implementation has. Every one cites the Swift code it mirrors.

**Items 5-6** are two adjacent defects in the marketplace lane (Rust) surfaced by the same audit. Item 6 is fully fixed; item 5 lands the Rust plumbing and the SQLite backend but does **not** yet reach the mobile hosts — see its section.

---

## 1. Consumed asset locks are terminal again

**Swift reference:** `PlatformWalletPersistenceHandler.swift:270` (upsert), `:310` (removal)

`onPersistAssetLockUpsert` did a plain Room `@Upsert` and `onPersistAssetLockRemoval` an unconditional delete — both last-write-wins. The writers race: the wallet-event adapter's batched drain can deliver a stale reconstruction/enrichment snapshot **after** the live flow's synchronous consumption write, silently regressing a consumed lock. Swift guards both sides; Kotlin did not.

A stored `Consumed` (4) row is now never overwritten by a non-consumed write and never deleted by a non-consumed removal. The guard is deliberately narrow — non-terminal statuses legitimately move both ways and stay last-write-wins.

## 2. `isOwned` / marketplace-column authority split

**Swift reference:** `upsertDPNSNames`, `PlatformWalletPersistenceHandler.swift:1912-1941`

Two halves, both restored:

- **The sweep** deleted departed DPNS rows outright, with no marketplace-history guard. Swift sets `isOwned = false` and deletes **only** when `documentId == null` (a pure label-cache row). This is the audit's Android-only **permanent-destruction** case: a name that departed during a round the marketplace pass could not classify lost the only record of where it went. Swift never did this.
- **The canonical branch** wrote `saleStatusRaw = 0` and `counterpartyIdentityId = null` over live marketplace state on every identity flush. Swift refreshes only `acquiredAt`/`label` (plus `isOwned = true`).

The identity snapshot's authority stops at `isOwned`; the marketplace columns belong to the reconciliation lane.

> Note this pairs directly with item 5 — the unclassifiable departure round that destroyed the row here is the same round item 5 addresses on the Rust side. One consequence of restoring Swift's keep-as-history behavior: on hosts where item 5's fallback is not yet wired (both mobile hosts — see item 5's scope), the restart-orphaned row is no longer destructively erased by this sweep, so it surfaces as a permanent **"Owned · not listed"** card in the Android marketplace list. That is the same stale card the Swift host shows in the same scenario — parity, not a new defect class — but it is newly visible on Android, and it is the visible face of the still-open mobile orphan.

## 3. `isLocal` promotion + startup heal

**Swift reference:** `row.isLocal = true` on wallet linkage (`swift:1827`); `healIdentityIsLocalFlags()` (`swift:4688`, called from `loadWalletList` `:4719`)

The persister hardcoded `isLocal = false` on every row it created, so a wallet's own identities — which are always local — were mis-marked, with no promotion on wallet linkage and no heal for rows already written.

- Promotion is one-way: set as soon as the row carries a wallet link; nothing writes `false` over a `true`.
- A promote-only, idempotent heal (`UPDATE identities SET isLocal = 1 WHERE walletId IS NOT NULL AND isLocal = 0`) runs from the load path — the one guaranteed per-launch pass over the store — and is skipped while a changeset round is open.

**The heal is safe on Android precisely because the Kotlin persister never mislinked `walletId`.** It only ever writes the link the FFI entry declared, so "has a wallet link" is exactly "is wallet-owned" — the condition cannot over-promote.

Also corrects the example app's `LoadIdentityScreen` manual-add to `isLocal = true`: a manual add is an identity the owner deliberately tracks, which is what the flag means. It carries no `walletId`, so neither the promotion nor the heal can reach it — that write is the only thing that can set it.

## 4. Watch-only clears a stale seed-mismatch flag

**Swift reference:** `PlatformWalletManager.swift:764-770`

`unlockWalletFromKeystore` returned on the `hasMnemonic` check **before** any status update. A wallet whose seed failed to bind and whose Keystore entry was then removed kept publishing the unlock banner forever for a seed that is no longer there — nothing downstream of the early return could clear it. No mnemonic is not a mismatch.

Ordering is the whole contract, so the ENTIRE guard sequence — the `hasMnemonic` storage read, the hex status-key derivation, the `seedMismatch = false` transform, and the early-return verdict — lives in an `internal` seam (`isGenuineWatchOnly`; the file's established pattern for native-free unit tests — cf. `initializePlatformWalletNativeManager`, `decodeShieldedCreatePayload`), and the call site in `unlockWalletFromKeystore` is pure delegation. The tests drive the seam through a fake storage probe and a manager-shaped status map and assert the clear provably runs before the watch-only return. An earlier cut of this fix extracted only a Boolean-taking helper, which left the real call-site sequence untested — see the corrected non-vacuity note below.

---

## 5. Orphaned marketplace row after a restart (audit F1) — Rust-side plumbing + SQLite backend; mobile-host wiring pending the persistence-vtable batch — the mobile orphan is NOT yet fixed

`resolve_departed_name` derived a departed name's `previous_document_id` solely from the in-memory `dpns_name_states` map. That map is **session-scoped**: the load path builds it EMPTY on every process start and nothing rehydrates it. So a departure observed in the first sync pass after a restart resolved no document id, emitted **no removal delta** — and still deleted the label. Since the label is what triggers departure detection, no later pass ever revisited it, and the host mirror kept a stale owned/listed row for a name the wallet no longer holds, permanently.

The lookup now falls back to the durable mirror through a new defaulted `PlatformWalletPersistence::get_dpns_name_state`, following the same `Ok(None)`-default shape as the existing `get_core_tx_record` / `list_wallet_core_txids`, so no existing backend breaks. `SqlitePersister` overrides it with a real reader scoped on **all three** of wallet, identity and normalized label — `Sold`/`Transferred` rows are retained in that table, so dropping the identity predicate could remove another identity's document. A persistence read failure degrades to today's behaviour rather than aborting the departure.

**Scope — what this does and does not fix.** The fallback reaches only backends that implement the read, and today that is `SqlitePersister` alone. `FFIPersister` keeps the `Ok(None)` default, and the persistence vtable has **no read slot** for this lookup — there is no callback a host could set. So the Android Room and iOS SwiftData mirrors — the two hosts this section's orphan narrative describes — still resolve nothing after a restart, and **their orphan is still live**. On Android, combined with item 2's keep-as-history sweep, it now renders as a permanent "Owned · not listed" card in the marketplace list (see item 2's note). Closing it requires a `get_dpns_name_state` read callback in the persistence vtable, which is deliberately deferred to the batched vtable/ABI additions rather than shipped piecemeal here; the in-code docs on `previous_document_id_for` and the trait method state this scope plainly.

Both defective arms are fixed at once on implementing backends: the `Ok(None)` arm and the summary in the `Err` arm shared the same resolution.

No migration for a dedicated index — SQLite serves this from the `(wallet_id, document_id)` primary key, scanning one wallet's rows, at most once per departed name per pass. Adding one would bump `max_supported_version` and trip the forward-version gate.

## 6. Zero-price listing guard

`set_dpns_name_price` never validated `price != 0`. Consensus would accept the listing and anyone could then take the name for free. It is now rejected with a typed `InvalidParameter` as the first statement — ahead of the operation gate and any network round-trip. `purchase_dpns_name` likewise refuses a `Some(0)` listing, checked **before** the `expected_price` comparison so the caller is told the listing is invalid rather than that the price moved. Every `> 0` path is unchanged.

There is no legitimate zero-price caller: a deliberate free handover is `transfer_dpns_name`, which names the recipient.

---

## Test evidence

| Suite | Result |
|---|---|
| `:sdk:testDebugUnitTest` (Kotlin) | **325 passed, 0 failed** (16 new; the 3 watch-only tests since rewritten against the call-site seam, count unchanged) |
| `cargo test -p platform-wallet --features shielded` | **855 passed, 0 failed**, 3 ignored (9 new) |
| `cargo test -p platform-wallet-storage` | **336 passed, 0 failed** (3 new, doctests included) |
| `cargo fmt --check` (both crates) | clean |
| `cargo clippy --all-targets -- -D warnings` (both crates) | clean |

**Non-vacuity was verified rather than assumed** — a green suite proves nothing if the tests don't gate the fix:

- **Kotlin, items 1-3:** reverting them fails the 8 tests asserting their changed behavior; the 5 that pin *unchanged* behavior (guard narrowness in both directions, label-cache deletion, observed identities staying non-local) keep passing. No pre-existing test broke under the revert, so none of this rests on relaxed expectations.
- **Kotlin, item 4 — corrected:** the evidence originally claimed here was **circular**. The watch-only test drove the extracted helper directly, so reverting the production call site to the pre-fix `if (!walletStorage.hasMnemonic(walletId)) return false` — the actual regression — kept the whole suite green (verified, 325/0). That gap is closed: the guard sequence now lives whole in the `isGenuineWatchOnly` seam, the call site is pure delegation, and the mutation was re-run against the seam — transplanting the pre-fix call-site shape into it fails 2 of the 3 watch-only tests, while the stored-mnemonic test (pinning unchanged behavior) keeps passing. The one line JVM tests still cannot see is the delegation call itself, which now carries no logic.
- **Rust:** short-circuiting the persister branch fails exactly the **4** fallback-dependent tests — including `resolve_departed_name_recovers_the_document_id_from_the_persister`, which runs the real async fn on a real `IdentityWallet` and so proves the fallback is wired into production rather than merely reachable as a helper. The steady-state test (`..._without_reading_the_persister`) and the unwired-backend test correctly keep passing. These exercise the SQLite-backed path; there is no FFI-host equivalent to test because no FFI read slot exists yet (see item 5's scope).

## Also in scope, from reviewing the above

- `read_all` moved its SQL into a `format!` as part of sharing the row projection with the new reader, which silently escaped the **TC-P1-003 `prepare_cached` lint** — its allow-list probe only inspects the three lines starting at the `.prepare(` call. Switched to `prepare_cached` and **dropped the now-dead allow-list exemption** rather than widening it; a stale exemption would later mask a genuinely uncached writer.
- Added the missing coverage for the SQLite reader itself. The marketplace tests exercise it through a test double, which proves the wiring but not the SQL — and a three-column filter is exactly where a wrong column or param order hides.

## Reviewer notes

- **Item 2 has a small behavior change beyond the stated bug:** the sweep now also deletes non-owned rows with no `documentId`, a shape the old Kotlin condition skipped entirely, leaving them to leak after `onRemoveDpnsNameState` nulls the marketplace columns. This is Swift's exact behavior, so it is parity, but it is worth a look.
- **Item 2 × item 5 interaction:** restoring Swift's keep-as-history sweep means the restart-orphaned row (item 5, still unfixed on mobile hosts) is no longer destructively erased on Android — it persists as a "Owned · not listed" card. Parity with what Swift shows today, but newly visible on Android; it goes away when the persistence-vtable batch adds the `get_dpns_name_state` read callback.
- No Room schema change, so no migration: the heal is a `@Query` `UPDATE`, and `IdentityEntity` is untouched.

## Summary by CodeRabbit

* **New Features**
  * Improved identity persistence and wallet restoration with more reliable local identity handling.
  * Added recovery for DPNS names after app restarts and improved departed-name handling.
  * Added clearer reporting when DPNS departures cannot be completed.

* **Bug Fixes**
  * Fixed stale watch-only and seed-mismatch status during wallet unlocking.
  * Protected consumed asset locks from stale updates or removal.
  * Rejected zero-priced marketplace listings and purchases before processing.
  * Preserved marketplace history during identity synchronization.
  * Improved DPNS state accuracy and validation when recovering persisted names.

### DIFF
--- a/packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandlerTest.kt
+++ b/packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandlerTest.kt
@@ -2912,7 +2912,7 @@ class PlatformWalletPersistenceHandlerTest {
     }
 
     @Test
-    fun consumedAssetLockIsNotRegressedByAStaleNonConsumedUpsert() = runTest {
+    fun shouldNotRegressAConsumedAssetLockOnAStaleNonConsumedUpsert() = runTest {
         val outpoint = makeOutpoint(ByteArray(32) { 60 }, 0)
         persistAssetLock(outpoint, status = 4, proofBytes = ByteArray(8) { 1 })
 
@@ -2927,7 +2927,7 @@ class PlatformWalletPersistenceHandlerTest {
     }
 
     @Test
-    fun consumedAssetLockStillAcceptsAnotherConsumedWrite() = runTest {
+    fun shouldStillAcceptAnotherConsumedWriteOnAConsumedAssetLock() = runTest {
         val outpoint = makeOutpoint(ByteArray(32) { 61 }, 0)
         persistAssetLock(outpoint, status = 4)
         persistAssetLock(outpoint, status = 4, amountDuffs = 250_000, proofBytes = ByteArray(4) { 7 })
@@ -2939,7 +2939,7 @@ class PlatformWalletPersistenceHandlerTest {
     }
 
     @Test
-    fun nonConsumedAssetLockStatusesStayLastWriteWinsInBothDirections() = runTest {
+    fun shouldKeepNonConsumedAssetLockStatusesLastWriteWinsInBothDirections() = runTest {
         val outpoint = makeOutpoint(ByteArray(32) { 62 }, 0)
         val hex = encodeOutPointHex(outpoint)
 
@@ -2954,7 +2954,7 @@ class PlatformWalletPersistenceHandlerTest {
     }
 
     @Test
-    fun consumedAssetLockSurvivesAStaleRemoval() = runTest {
+    fun shouldRetainAConsumedAssetLockThroughAStaleRemoval() = runTest {
         val outpoint = makeOutpoint(ByteArray(32) { 63 }, 0)
         persistAssetLock(outpoint, status = 4)
 
@@ -2971,7 +2971,7 @@ class PlatformWalletPersistenceHandlerTest {
     }
 
     @Test
-    fun nonConsumedAssetLockIsStillRemoved() = runTest {
+    fun shouldStillRemoveANonConsumedAssetLock() = runTest {
         val outpoint = makeOutpoint(ByteArray(32) { 64 }, 0)
         persistAssetLock(outpoint, status = 0) // Built
 
@@ -3033,7 +3033,7 @@ class PlatformWalletPersistenceHandlerTest {
     }
 
     @Test
-    fun identitySweepKeepsAListedNameAsDepartedHistoryInsteadOfDestroyingIt() = runTest {
+    fun shouldKeepAListedNameAsDepartedHistoryThroughTheIdentitySweepInsteadOfDestroyingIt() = runTest {
         handler.onPersistWalletMetadata(walletId, testnet, groupId, 0)
         val identityId = ByteArray(32) { 20 }
         val documentId = ByteArray(32) { 21 }
@@ -3062,7 +3062,7 @@ class PlatformWalletPersistenceHandlerTest {
     }
 
     @Test
-    fun identitySweepKeepsASoldNameWithItsCounterparty() = runTest {
+    fun shouldKeepASoldNameWithItsCounterpartyThroughTheIdentitySweep() = runTest {
         handler.onPersistWalletMetadata(walletId, testnet, groupId, 0)
         val identityId = ByteArray(32) { 22 }
         val documentId = ByteArray(32) { 23 }
@@ -3084,7 +3084,7 @@ class PlatformWalletPersistenceHandlerTest {
     }
 
     @Test
-    fun identitySweepStillDeletesLabelCacheRowsWithNoMarketplaceHistory() = runTest {
+    fun shouldStillDeleteLabelCacheRowsWithNoMarketplaceHistoryOnTheIdentitySweep() = runTest {
         handler.onPersistWalletMetadata(walletId, testnet, groupId, 0)
         val identityId = ByteArray(32) { 25 }
 
@@ -3102,7 +3102,7 @@ class PlatformWalletPersistenceHandlerTest {
     }
 
     @Test
-    fun identitySnapshotDoesNotClobberMarketplaceColumnsOnAStillCarriedLabel() = runTest {
+    fun shouldNotClobberMarketplaceColumnsWhenAnIdentitySnapshotStillCarriesTheLabel() = runTest {
         handler.onPersistWalletMetadata(walletId, testnet, groupId, 0)
         val identityId = ByteArray(32) { 26 }
         val documentId = ByteArray(32) { 27 }
@@ -3136,7 +3136,7 @@ class PlatformWalletPersistenceHandlerTest {
     // `healIdentityIsLocalFlags` (:4688, called from loadWalletList :4719).
 
     @Test
-    fun persisterCreatedWalletLinkedIdentityIsMarkedLocal() = runTest {
+    fun shouldMarkAPersisterCreatedWalletLinkedIdentityLocal() = runTest {
         handler.onPersistWalletMetadata(walletId, testnet, groupId, 0)
         val identityId = ByteArray(32) { 30 }
 
@@ -3148,7 +3148,7 @@ class PlatformWalletPersistenceHandlerTest {
     }
 
     @Test
-    fun observedOutOfWalletIdentityStaysNonLocal() = runTest {
+    fun shouldKeepAnObservedOutOfWalletIdentityNonLocal() = runTest {
         handler.onPersistWalletMetadata(walletId, testnet, groupId, 0)
         val identityId = ByteArray(32) { 31 }
 
@@ -3166,7 +3166,7 @@ class PlatformWalletPersistenceHandlerTest {
     }
 
     @Test
-    fun laterWalletLinkagePromotesAnAlreadyPersistedIdentityToLocal() = runTest {
+    fun shouldPromoteAnAlreadyPersistedIdentityToLocalOnLaterWalletLinkage() = runTest {
         handler.onPersistWalletMetadata(walletId, testnet, groupId, 0)
         val identityId = ByteArray(32) { 32 }
 
@@ -3186,7 +3186,7 @@ class PlatformWalletPersistenceHandlerTest {
     }
 
     @Test
-    fun loadPathHealsLegacyIsLocalFalseOnWalletLinkedRowsOnly() = runTest {
+    fun shouldHealLegacyIsLocalFalseOnWalletLinkedRowsOnlyDuringLoad() = runTest {
         handler.onPersistWalletMetadata(walletId, testnet, groupId, 0)
 
         val legacyOwned = ByteArray(32) { 33 }
--- a/packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandlerTest.kt
+++ b/packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandlerTest.kt
@@ -2880,4 +2880,362 @@ class PlatformWalletPersistenceHandlerTest {
         // Unchanged pre-invitation behavior: no account row conjured.
         assertTrue(db.accountDao().observeByWallet(walletId).first().isEmpty())
     }
+
+    // ── Asset locks: Consumed is terminal ─────────────────────────────
+    //
+    // Swift parity with `persistAssetLocks`
+    // (PlatformWalletPersistenceHandler.swift:270 upsert, :310 removal).
+    // Consumed (4) is the terminal lifecycle state and the writers race:
+    // the wallet-event adapter's batched drain can deliver a stale
+    // reconstruction snapshot AFTER the live flow's consumption write.
+
+    /** Upsert [outpoint] at [status] in its own committed round. */
+    private fun persistAssetLock(
+        outpoint: ByteArray,
+        status: Byte,
+        amountDuffs: Long = 100_000,
+        proofBytes: ByteArray? = null,
+    ) {
+        handler.onChangesetBegin(walletId)
+        handler.onPersistAssetLockUpsert(
+            walletId = walletId,
+            outPoint = outpoint,
+            transactionBytes = ByteArray(20) { 41 },
+            accountIndex = 0,
+            fundingType = 0,
+            identityIndex = 0,
+            amountDuffs = amountDuffs,
+            status = status,
+            proofBytes = proofBytes,
+        )
+        handler.onChangesetEnd(walletId, success = true)
+    }
+
+    @Test
+    fun consumedAssetLockIsNotRegressedByAStaleNonConsumedUpsert() = runTest {
+        val outpoint = makeOutpoint(ByteArray(32) { 60 }, 0)
+        persistAssetLock(outpoint, status = 4, proofBytes = ByteArray(8) { 1 })
+
+        // Stale replay of an older Broadcast snapshot.
+        persistAssetLock(outpoint, status = 1, amountDuffs = 999, proofBytes = null)
+
+        val row = db.assetLockDao().getByOutPointHex(encodeOutPointHex(outpoint))!!
+        assertEquals(4, row.statusRaw)
+        // The whole row is skipped, not merely the status column.
+        assertEquals(100_000L, row.amountDuffs)
+        assertNotNull(row.proofBytes)
+    }
+
+    @Test
+    fun consumedAssetLockStillAcceptsAnotherConsumedWrite() = runTest {
+        val outpoint = makeOutpoint(ByteArray(32) { 61 }, 0)
+        persistAssetLock(outpoint, status = 4)
+        persistAssetLock(outpoint, status = 4, amountDuffs = 250_000, proofBytes = ByteArray(4) { 7 })
+
+        val row = db.assetLockDao().getByOutPointHex(encodeOutPointHex(outpoint))!!
+        assertEquals(4, row.statusRaw)
+        assertEquals(250_000L, row.amountDuffs)
+        assertNotNull(row.proofBytes)
+    }
+
+    @Test
+    fun nonConsumedAssetLockStatusesStayLastWriteWinsInBothDirections() = runTest {
+        val outpoint = makeOutpoint(ByteArray(32) { 62 }, 0)
+        val hex = encodeOutPointHex(outpoint)
+
+        persistAssetLock(outpoint, status = 1) // Broadcast
+        persistAssetLock(outpoint, status = 3) // ChainLocked
+        assertEquals(3, db.assetLockDao().getByOutPointHex(hex)!!.statusRaw)
+
+        // The guard is narrow: non-terminal statuses legitimately move both
+        // ways, so they stay last-write-wins.
+        persistAssetLock(outpoint, status = 1)
+        assertEquals(1, db.assetLockDao().getByOutPointHex(hex)!!.statusRaw)
+    }
+
+    @Test
+    fun consumedAssetLockSurvivesAStaleRemoval() = runTest {
+        val outpoint = makeOutpoint(ByteArray(32) { 63 }, 0)
+        persistAssetLock(outpoint, status = 4)
+
+        handler.onChangesetBegin(walletId)
+        assertEquals(0, handler.onPersistAssetLockRemoval(walletId, outpoint))
+        assertEquals(0, handler.onChangesetEnd(walletId, success = true))
+
+        // Retained for historical lookup: the only removal emitter
+        // (`untrack_asset_lock`) targets rejected Built rows, so a removal
+        // reaching a consumed row is by construction a stale write.
+        val row = db.assetLockDao().getByOutPointHex(encodeOutPointHex(outpoint))
+        assertNotNull(row)
+        assertEquals(4, row!!.statusRaw)
+    }
+
+    @Test
+    fun nonConsumedAssetLockIsStillRemoved() = runTest {
+        val outpoint = makeOutpoint(ByteArray(32) { 64 }, 0)
+        persistAssetLock(outpoint, status = 0) // Built
+
+        handler.onChangesetBegin(walletId)
+        handler.onPersistAssetLockRemoval(walletId, outpoint)
+        handler.onChangesetEnd(walletId, success = true)
+
+        assertNull(db.assetLockDao().getByOutPointHex(encodeOutPointHex(outpoint)))
+    }
+
+    // ── Identity sweep vs marketplace column authority ────────────────
+    //
+    // Swift parity with `upsertDPNSNames`
+    // (PlatformWalletPersistenceHandler.swift:1912-1941). The identity
+    // snapshot owns `isOwned` / `acquiredAt` / `label`; the marketplace
+    // reconciliation lane owns `documentId` / price / sale status /
+    // counterparty. Neither may overwrite the other's columns.
+
+    /** Commit one canonical identity snapshot carrying exactly [names]. */
+    private fun persistIdentitySnapshot(identityId: ByteArray, vararg names: String) {
+        handler.onChangesetBegin(walletId)
+        handler.onPersistIdentityUpsert(
+            walletId, identityId, 1, 0, false, 0, 0, true, walletId,
+            names.toList().toTypedArray(), LongArray(names.size), false, null, null, null,
+            ByteArray(32), false, ByteArray(8), false, null,
+        )
+        handler.onChangesetEnd(walletId, success = true)
+    }
+
+    /** Commit one marketplace row for [documentId]. */
+    private fun persistMarketplaceRow(
+        identityId: ByteArray,
+        documentId: ByteArray,
+        label: String,
+        normalizedLabel: String,
+        status: Byte,
+        counterpartyId: ByteArray? = null,
+        priceCredits: Long? = null,
+    ) {
+        handler.onChangesetBegin(walletId)
+        handler.onPersistDpnsNameState(
+            walletId = walletId,
+            documentId = documentId,
+            walletIdentityId = identityId,
+            hasCounterparty = counterpartyId != null,
+            counterpartyId = counterpartyId ?: ByteArray(32),
+            label = label,
+            normalizedLabel = normalizedLabel,
+            normalizedParentDomainName = "dash",
+            hasPrice = priceCredits != null,
+            priceCredits = priceCredits ?: 0,
+            status = status,
+            createdAtMs = 100,
+            updatedAtMs = 200,
+            transferredAtMs = 300,
+            lastSyncedAtMs = 400,
+        )
+        handler.onChangesetEnd(walletId, success = true)
+    }
+
+    @Test
+    fun identitySweepKeepsAListedNameAsDepartedHistoryInsteadOfDestroyingIt() = runTest {
+        handler.onPersistWalletMetadata(walletId, testnet, groupId, 0)
+        val identityId = ByteArray(32) { 20 }
+        val documentId = ByteArray(32) { 21 }
+
+        persistIdentitySnapshot(identityId, "Alice")
+        // Listed for sale: still owned, and marketplace-tracked.
+        persistMarketplaceRow(
+            identityId, documentId, "Alice", "a11ce",
+            status = 0, priceCredits = 5_000,
+        )
+        assertTrue(db.dpnsNameDao().getByDocumentId(documentId)!!.isOwned)
+
+        // The label leaves the canonical set before the marketplace pass can
+        // classify the departure — the unclassifiable-departure round. The
+        // old sweep deleted the row here, permanently destroying the only
+        // record of where the name went (Android-only; Swift never did).
+        persistIdentitySnapshot(identityId)
+
+        val retained = db.dpnsNameDao().getByDocumentId(documentId)
+        assertNotNull("marketplace history must survive the identity sweep", retained)
+        assertFalse("a departed name must not read as owned", retained!!.isOwned)
+        assertEquals(5_000L, retained.priceCredits)
+        assertEquals(400L, retained.marketplaceUpdatedAt)
+        // ...but owned-name queries and UI selection must not surface it.
+        assertTrue(db.dpnsNameDao().observeByIdentity(identityId).first().isEmpty())
+    }
+
+    @Test
+    fun identitySweepKeepsASoldNameWithItsCounterparty() = runTest {
+        handler.onPersistWalletMetadata(walletId, testnet, groupId, 0)
+        val identityId = ByteArray(32) { 22 }
+        val documentId = ByteArray(32) { 23 }
+        val buyerId = ByteArray(32) { 24 }
+
+        persistIdentitySnapshot(identityId, "Bob")
+        persistMarketplaceRow(
+            identityId, documentId, "Bob", "b0b",
+            status = 1, counterpartyId = buyerId,
+        )
+
+        persistIdentitySnapshot(identityId)
+
+        val retained = db.dpnsNameDao().getByDocumentId(documentId)
+        assertNotNull(retained)
+        assertFalse(retained!!.isOwned)
+        assertEquals(1, retained.saleStatusRaw)
+        assertTrue(buyerId.contentEquals(retained.counterpartyIdentityId!!))
+    }
+
+    @Test
+    fun identitySweepStillDeletesLabelCacheRowsWithNoMarketplaceHistory() = runTest {
+        handler.onPersistWalletMetadata(walletId, testnet, groupId, 0)
+        val identityId = ByteArray(32) { 25 }
+
+        persistIdentitySnapshot(identityId, "Alice", "Bob")
+        assertEquals(2, db.dpnsNameDao().getAllByIdentity(identityId).size)
+
+        persistIdentitySnapshot(identityId, "Alice")
+
+        // No documentId attached → a pure stale label-cache row, removed
+        // entirely (unchanged behavior).
+        assertEquals(
+            listOf("Alice"),
+            db.dpnsNameDao().getAllByIdentity(identityId).map { it.label },
+        )
+    }
+
+    @Test
+    fun identitySnapshotDoesNotClobberMarketplaceColumnsOnAStillCarriedLabel() = runTest {
+        handler.onPersistWalletMetadata(walletId, testnet, groupId, 0)
+        val identityId = ByteArray(32) { 26 }
+        val documentId = ByteArray(32) { 27 }
+        val recipientId = ByteArray(32) { 28 }
+
+        persistIdentitySnapshot(identityId, "Carol")
+        persistMarketplaceRow(
+            identityId, documentId, "Carol", "car01",
+            status = 2, counterpartyId = recipientId, priceCredits = 9_000,
+        )
+
+        // A later identity flush still carries the label — the two lanes
+        // disagree for a round. The canonical branch refreshes only
+        // acquiredAt/label (+ isOwned); it used to blank saleStatusRaw and
+        // counterpartyIdentityId on every such flush.
+        persistIdentitySnapshot(identityId, "Carol")
+
+        val row = db.dpnsNameDao().getByDocumentId(documentId)
+        assertNotNull(row)
+        assertEquals(2, row!!.saleStatusRaw)
+        assertTrue(recipientId.contentEquals(row.counterpartyIdentityId!!))
+        assertEquals(9_000L, row.priceCredits)
+        assertEquals(300L, row.documentTransferredAtMs)
+        assertTrue("the identity lane still asserts ownership", row.isOwned)
+    }
+
+    // ── isLocal: wallet-link promotion + load-path heal ───────────────
+    //
+    // Swift parity with `persistIdentities`
+    // (PlatformWalletPersistenceHandler.swift:1827-1829) and
+    // `healIdentityIsLocalFlags` (:4688, called from loadWalletList :4719).
+
+    @Test
+    fun persisterCreatedWalletLinkedIdentityIsMarkedLocal() = runTest {
+        handler.onPersistWalletMetadata(walletId, testnet, groupId, 0)
+        val identityId = ByteArray(32) { 30 }
+
+        persistIdentitySnapshot(identityId)
+
+        val row = db.identityDao().getByIdentityId(identityId)!!
+        assertTrue(walletId.contentEquals(row.walletId!!))
+        assertTrue("a wallet's own identity is always local", row.isLocal)
+    }
+
+    @Test
+    fun observedOutOfWalletIdentityStaysNonLocal() = runTest {
+        handler.onPersistWalletMetadata(walletId, testnet, groupId, 0)
+        val identityId = ByteArray(32) { 31 }
+
+        handler.onChangesetBegin(walletId)
+        handler.onPersistIdentityUpsert(
+            walletId, identityId, 1, 0, false, 0, 0, false, ByteArray(32),
+            emptyArray(), longArrayOf(), false, null, null, null,
+            ByteArray(32), false, ByteArray(8), false, null,
+        )
+        handler.onChangesetEnd(walletId, success = true)
+
+        val row = db.identityDao().getByIdentityId(identityId)!!
+        assertNull(row.walletId)
+        assertFalse("an observed identity is not local", row.isLocal)
+    }
+
+    @Test
+    fun laterWalletLinkagePromotesAnAlreadyPersistedIdentityToLocal() = runTest {
+        handler.onPersistWalletMetadata(walletId, testnet, groupId, 0)
+        val identityId = ByteArray(32) { 32 }
+
+        handler.onChangesetBegin(walletId)
+        handler.onPersistIdentityUpsert(
+            walletId, identityId, 1, 0, false, 0, 0, false, ByteArray(32),
+            emptyArray(), longArrayOf(), false, null, null, null,
+            ByteArray(32), false, ByteArray(8), false, null,
+        )
+        handler.onChangesetEnd(walletId, success = true)
+        assertFalse(db.identityDao().getByIdentityId(identityId)!!.isLocal)
+
+        // The wallet relationship attaches on a later flush — promote.
+        persistIdentitySnapshot(identityId)
+
+        assertTrue(db.identityDao().getByIdentityId(identityId)!!.isLocal)
+    }
+
+    @Test
+    fun loadPathHealsLegacyIsLocalFalseOnWalletLinkedRowsOnly() = runTest {
+        handler.onPersistWalletMetadata(walletId, testnet, groupId, 0)
+
+        val legacyOwned = ByteArray(32) { 33 }
+        val manualAdd = ByteArray(32) { 34 }
+        val observed = ByteArray(32) { 35 }
+
+        // Rows exactly as the pre-fix persister wrote them: a constant
+        // `false` even on the wallet's own identities.
+        db.identityDao().upsert(
+            IdentityEntity(
+                identityId = legacyOwned,
+                networkRaw = testnet,
+                walletId = walletId,
+                isLocal = false,
+            ),
+        )
+        db.identityDao().upsert(
+            IdentityEntity(
+                identityId = manualAdd,
+                networkRaw = testnet,
+                walletId = null,
+                isLocal = true,
+            ),
+        )
+        db.identityDao().upsert(
+            IdentityEntity(
+                identityId = observed,
+                networkRaw = testnet,
+                walletId = null,
+                isLocal = false,
+            ),
+        )
+
+        handler.onLoadWalletList()
+
+        assertTrue(
+            "a wallet-linked legacy row is promoted",
+            db.identityDao().getByIdentityId(legacyOwned)!!.isLocal,
+        )
+        assertTrue(
+            "a manual add (no wallet link) keeps its flag",
+            db.identityDao().getByIdentityId(manualAdd)!!.isLocal,
+        )
+        assertFalse(
+            "an observed row is never promoted",
+            db.identityDao().getByIdentityId(observed)!!.isLocal,
+        )
+
+        // Promote-only and idempotent: a second pass matches nothing.
+        assertEquals(0, db.identityDao().healIsLocalFlags())
+    }
 }
--- a/packages/rs-platform-wallet-storage/src/sqlite/schema/dpns_name_states.rs
+++ b/packages/rs-platform-wallet-storage/src/sqlite/schema/dpns_name_states.rs
@@ -6,19 +6,17 @@
 //! `counterparty_id` column (NULL for `owned`), with the pairing enforced by
 //! a table CHECK.
 
-use rusqlite::{params, Transaction};
+use dpp::prelude::Identifier;
+use rusqlite::{params, Connection, OptionalExtension, Transaction};
 
-use platform_wallet::changeset::{DpnsNameSaleStatus, DpnsNameStateChangeSet};
+use platform_wallet::changeset::{DpnsNameSaleStatus, DpnsNameStateChangeSet, DpnsNameStateEntry};
 use platform_wallet::wallet::platform_wallet::WalletId;
 
 use crate::sqlite::error::WalletStorageError;
 
-// Imports used only by the test-gated reader below.
+// Import used only by the test-gated whole-table reader below.
 #[cfg(any(test, feature = "__test-helpers"))]
-use {
-    dpp::prelude::Identifier, platform_wallet::changeset::DpnsNameStateEntry, rusqlite::Connection,
-    std::collections::BTreeMap,
-};
+use std::collections::BTreeMap;
 
 pub fn apply(
     tx: &Transaction<'_>,
@@ -111,7 +109,9 @@ pub(crate) fn status_columns(s: &DpnsNameSaleStatus) -> (&'static str, Option<[u
     }
 }
 
-#[cfg(any(test, feature = "__test-helpers"))]
+/// Inverse of [`status_columns`]. Ungated: the production
+/// [`get_by_identity_and_label`] reader decodes rows too, not just the
+/// test-only whole-table reader.
 fn status_from_columns(
     status: &str,
     counterparty: Option<Vec<u8>>,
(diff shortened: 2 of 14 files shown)


thepastaclaw:
✅ Final review complete — no blockers (commit c7ce550b)

bfoss765:
Two corrections on my own claims in this PR, both pushed (0311298b).

**Item 4's test evidence was circular.** The watch-only test exercised the extracted helper against itself: reverting the production call site to the pre-fix `if (!walletStorage.hasMnemonic(walletId)) return false` — the actual regression — kept the whole suite green (verified, 325/0). I've made the call-site shape itself the unit under test: `isGenuineWatchOnly` now performs the entire guard sequence (storage read, status-key derivation, the seedMismatch clear, the verdict) and the call site is pure delegation. Mutation-verified this time in the direction that matters: transplanting the pre-fix shape into the seam fails 2 of the 3 watch-only tests, and the stored-mnemonic test correctly keeps passing. Suite is 325/0 again.

**Item 5's description overstated its reach.** The `get_dpns_name_state` override exists only in `SqlitePersister`. `FFIPersister` keeps the `Ok(None)` default, and the persistence vtable has no read slot for this lookup — there is no callback a host could set — so the Android Room / iOS SwiftData mirrors the section narrates are NOT fixed: their restart orphan is still live until a read callback lands with the batched vtable/ABI additions (deliberately not smuggled into this PR). One knock-on worth flagging for item 2's review: with the Swift-parity sweep no longer destructively deleting departed rows, the unresolved orphan on Android now shows as a permanent "Owned · not listed" card in the marketplace list — same stale card the Swift host shows, but newly visible on Android. I've corrected the in-code docs and the PR description to state the real scope.

packages/rs-platform-wallet-storage/src/sqlite/schema/dpns_name_states.rs
```diff
@@ -133,6 +133,137 @@ fn status_from_columns(
...
+pub fn get_by_identity_and_label(
+    conn: &Connection,
+    wallet_id: &WalletId,
+    wallet_identity_id: &Identifier,
+    normalized_label: &str,
+) -> Result<Option<DpnsNameStateEntry>, WalletStorageError> {
+    let sql = format!(
+        "SELECT {ROW_PROJECTION} FROM dpns_name_states \
+         WHERE wallet_id = ?1 AND identity_id = ?2 AND normalized_label = ?3 LIMIT 1"
+    );
```
thepastaclaw, line 285: **🔴 Blocking: Select the current retained row deterministically**

The schema does not enforce uniqueness for `(wallet_id, identity_id, normalized_label)`; its primary key is `(wallet_id, document_id)`. DPNS domain documents can be deleted and re-registered with a new document ID, while this table deliberately retains earlier `Sold` and `Transferred` rows. A wallet identity can therefore have an old historical row and a newer owned row for the same normalized label. This unordered `LIMIT 1` may select the old row according to primary-key scan order. If the current document is then deleted and the fallback runs after restart, the sync removes the historical row and the identity label while leaving the current persisted row permanently orphaned. Prefer an `Owned` row, then the most recently synchronized row, and update the trait documentation that currently says any matching row is acceptable.

```suggestion
let sql = format!(
    "SELECT {ROW_PROJECTION} FROM dpns_name_states \
     WHERE wallet_id = ?1 AND identity_id = ?2 AND normalized_label = ?3 \
     ORDER BY CASE status WHEN 'owned' THEN 0 ELSE 1 END, \
              last_synced_at_ms DESC, document_id DESC \
     LIMIT 1"
);
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
```diff
@@ -384,6 +384,70 @@ fn insert_sync_row(rows: &mut BTreeMap<Identifier, DpnsNameStateEntry>, entry: D
...
+    match persister.get_dpns_name_state(identity_id, &normalized_label) {
+        Ok(row) => row.map(|entry| entry.document_id),
+        Err(error) => {
+            tracing::warn!(
+                identity = %identity_id,
+                name = label,
+                "persisted DPNS row lookup failed for a departed name; the host mirror may \
+                 keep a stale row for it until the name is re-acquired: {error}"
+            );
+            None
```
thepastaclaw: **🔴 Blocking: Retry departed-name recovery after persistence read failures**

`Ok(None)` means that a backend does not support this lookup or has no row, while `Err` means an attempted persistence read failed; this branch collapses those distinct outcomes. If SQLite returns a transient read error and the subsequent Platform lookup confirms that the domain document is absent, `resolve_departed_name` receives no previous document ID and sets `retry` to false. The caller then removes the label without emitting a row-removal delta. Because that label is the trigger for future departure detection, the durable row is permanently orphaned. `PersistenceError` already preserves transient classification, and this flow already retains pending departures for retryable network errors. Propagate the persistence lookup result into `resolve_departed_name` and retain the pending departure on a transient error instead of treating it as an unsupported lookup.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-storage/src/sqlite/schema/dpns_name_states.rs
```diff
@@ -133,6 +133,137 @@ fn status_from_columns(
...
+        wallet_identity_id,
+        label,
+        normalized_label,
+        normalized_parent_domain_name: normalized_parent,
+        price: price.map(|p| p as u64),
+        status: status_from_columns(&status, counterparty)?,
+        created_at_ms: created_at.map(|v| v as u64),
+        updated_at_ms: updated_at.map(|v| v as u64),
+        transferred_at_ms: transferred_at.map(|v| v as u64),
+        last_synced_at_ms: last_synced as u64,
```
thepastaclaw: **🟡 Suggestion: Use checked conversions for persisted timestamps**

Only `price` has a non-negative schema constraint; the four timestamp columns permit negative SQLite integers. These `as u64` casts silently turn a malformed or externally modified `-1` into `u64::MAX`, despite the storage crate's explicit rule that durable-boundary casts use `safe_cast`. This decoder is now used by a production persistence lookup rather than only the test-gated whole-table helper. Decode every signed database value through `i64_to_u64` so malformed rows return the existing typed `IntegerOverflow` error.

```suggestion
price: price
    .map(|value| {
        crate::sqlite::util::safe_cast::i64_to_u64("dpns_name_states.price", value)
    })
    .transpose()?,
status: status_from_columns(&status, counterparty)?,
created_at_ms: created_at
    .map(|value| {
        crate::sqlite::util::safe_cast::i64_to_u64(
            "dpns_name_states.created_at_ms",
            value,
        )
    })
    .transpose()?,
updated_at_ms: updated_at
    .map(|value| {
        crate::sqlite::util::safe_cast::i64_to_u64(
            "dpns_name_states.updated_at_ms",
            value,
        )
    })
    .transpose()?,
transferred_at_ms: transferred_at
    .map(|value| {
        crate::sqlite::util::safe_cast::i64_to_u64(
            "dpns_name_states.transferred_at_ms",
            value,
        )
    })
    .transpose()?,
last_synced_at_ms: crate::sqlite::util::safe_cast::i64_to_u64(
    "dpns_name_states.last_synced_at_ms",
    last_synced,
)?,
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
```diff
@@ -1076,6 +1165,18 @@ impl IdentityWallet {
...
+        // ahead of the `expected_price` comparison so the caller is told
+        // the listing itself is not purchasable rather than that the
+        // price moved. Only `== 0` is affected; every `> 0` listing
+        // follows the unchanged price-match path below.
+        if listed_price == 0 {
+            return Err(PlatformWalletError::InvalidParameter(format!(
+                "DPNS name {name:?} carries a listed price of 0 credits, which is not a \
+                 valid sale listing and will not be purchased by this wallet"
+            )));
+        }
```
thepastaclaw: **🟡 Suggestion: Add direct coverage for purchase-side zero-price rejection**

The added tests exercise `set_dpns_name_price(0)` and its non-zero path, but no test reaches this separate guard in `purchase_dpns_name`. Add a regression case whose fetched domain state has `price = Some(0)` and whose `expected_price` is non-zero. It should assert `InvalidParameter`, rather than `DocumentPriceChanged`, and verify that signing and broadcast are not reached. This pins the ordering that the new code and API documentation explicitly promise.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The Kotlin parity changes and zero-price guards are generally well scoped, but the new SQLite departed-name recovery can still orphan the current row when multiple historical documents match, and it permanently loses recovery after a transient persistence-read failure. The production SQLite decoder also needs checked timestamp conversions, and the purchase-side zero-price ordering lacks direct regression coverage.
Source: reviewers codex general/security-auditor/rust-quality (backend model gpt-5.6-sol); final verifier codex (backend model gpt-5.6-sol). openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

### Review provenance

<details>

```

In `packages/rs-platform-wallet-storage/src/sqlite/schema/dpns_name_states.rs`:
- [BLOCKING] packages/rs-platform-wallet-storage/src/sqlite/schema/dpns_name_states.rs:249-252: Select the current retained row deterministically
  The schema does not enforce uniqueness for `(wallet_id, identity_id, normalized_label)`; its primary key is `(wallet_id, document_id)`. DPNS domain documents can be deleted and re-registered with a new document ID, while this table deliberately retains earlier `Sold` and `Transferred` rows. A wallet identity can therefore have an old historical row and a newer owned row for the same normalized label. This unordered `LIMIT 1` may select the old row according to primary-key scan order. If the current document is then deleted and the fallback runs after restart, the sync removes the historical row and the identity label while leaving the current persisted row permanently orphaned. Prefer an `Owned` row, then the most recently synchronized row, and update the trait documentation that currently says any matching row is acceptable.
- [SUGGESTION] packages/rs-platform-wallet-storage/src/sqlite/schema/dpns_name_states.rs:211-216: Use checked conversions for persisted timestamps
  Only `price` has a non-negative schema constraint; the four timestamp columns permit negative SQLite integers. These `as u64` casts silently turn a malformed or externally modified `-1` into `u64::MAX`, despite the storage crate's explicit rule that durable-boundary casts use `safe_cast`. This decoder is now used by a production persistence lookup rather than only the test-gated whole-table helper. Decode every signed database value through `i64_to_u64` so malformed rows return the existing typed `IntegerOverflow` error.

In `packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs:437-446: Retry departed-name recovery after persistence read failures
  `Ok(None)` means that a backend does not support this lookup or has no row, while `Err` means an attempted persistence read failed; this branch collapses those distinct outcomes. If SQLite returns a transient read error and the subsequent Platform lookup confirms that the domain document is absent, `resolve_departed_name` receives no previous document ID and sets `retry` to false. The caller then removes the label without emitting a row-removal delta. Because that label is the trigger for future departure detection, the durable row is permanently orphaned. `PersistenceError` already preserves transient classification, and this flow already retains pending departures for retryable network errors. Propagate the persistence lookup result into `resolve_departed_name` and retain the pending departure on a transient error instead of treating it as an unsupported lookup.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs:1168-1179: Add direct coverage for purchase-side zero-price rejection
  The added tests exercise `set_dpns_name_price(0)` and its non-zero path, but no test reaches this separate guard in `purchase_dpns_name`. Add a regression case whose fetched domain state has `price = Some(0)` and whose `expected_price` is non-zero. It should assert `InvalidParameter`, rather than `DocumentPriceChanged`, and verify that signing and broadcast are not reached. This pins the ordering that the new code and API documentation explicitly promise.
```
</details>

packages/rs-platform-wallet-storage/src/sqlite/schema/dpns_name_states.rs
bfoss765: Fixed in 2f646dabba. I applied your ordering as supplied — the column names lined up with V005 exactly, so it went in verbatim.

You're right that the triple isn't unique and that scan order has no relation to recency. I added a test that proves the failure rather than just the fix: it seeds a `Sold` row with a *lower* document id than the current `Owned` row, then asserts the raw unordered `LIMIT 1` really does return the historical row before asserting `get_by_identity_and_label` returns the current one. It does, so this wasn't theoretical. A second test pins the tie-break order when no `owned` row is left, with the fresher row given the lower document id so a reader that sorted on document id alone would fail it.

The trait doc no longer says any matching row is acceptable. `get_dpns_name_state` now mandates the same preference for every backend and spells out why an arbitrary match is worse than it looks: it aims the orphaning at the wrong row instead of preventing it.

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
bfoss765: Fixed in 50002d6e3d. `previous_document_id_for` now returns `Result` and reports the lookup without deciding policy; `resolve_departed_name` tells the two outcomes apart. A transient error returns `retry` before the network call, so the sync loop pushes the departure back and breaks ahead of `remove_dpns_label` — label, queue entry and durable row all survive. A non-transient error still degrades to `None` with a warning: it can't be retried into success, and since `retry` breaks the loop, retaining it would wedge that identity's whole departure queue for the life of the process. Tell me if you'd rather retain on every error class.

Getting the test to mean anything took some doing, and it's worth flagging. Your scenario needs Platform to return *confirmed-absent*, not *fetch failed* — with a bare mock SDK every fetch errors, the retry arm is reached regardless, and the obvious test passes against the unfixed code. So the mock is now primed to answer the DPNS contract fetch and the exact-match domain query. That initially matched nothing, because an unpinned mock seeds at the network-minimum protocol version and ratchets up on the first response, silently re-encoding the later document query from V0 to V1 after the contract fetch; pinning the version fixed it, and there's a comment on the helper so the next person doesn't lose the same hour.

The test asserts the confirmed-absent branch is reached with a healthy mirror (`retry == false`, removal delta emitted), then that the same Platform answer with a transiently failing mirror retains instead with no delta and no claimed document id, then heals the mirror and asserts the retained departure resolves normally. The domain query is now built by a shared named function so the expectation can't drift from the production query.

packages/rs-platform-wallet-storage/src/sqlite/schema/dpns_name_states.rs
bfoss765: Done in 2f646dabba, all five signed columns including `price` as in your diff. Your point about this decoder now backing a production read rather than only the test-gated helper is what makes it worth more than tidiness — a corrupted `-1` decoded to `u64::MAX`, which is a year-584-million timestamp that reads back as perfectly valid marketplace state rather than as corruption.

Test corrupts each of the four timestamp columns in turn (insert valid, then `UPDATE` — the only route to `last_synced_at_ms`, which is `NOT NULL`) and asserts the typed `IntegerOverflow` names the offending column.

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
bfoss765: Added in 50002d6e3d. The price pre-flight is extracted as a pure `preflight_purchase_price` so the rejection order can be pinned directly — not-for-sale, then zero, then price-changed — without a live Platform to serve the domain fetch that precedes it.

On "verify that signing and broadcast are not reached": two end-to-end tests drive the real `purchase_dpns_name` against a mock serving a document listed at `$price = 0`, with a purchaser that is deliberately *not* one of the wallet's identities. Every step after the pre-flight then fails with a different, identifiable error, so a typed `InvalidParameter` coming back is itself the proof that nothing downstream ran. The `expected_price = 0` case is the one I'd point at: the prices match, so without the guard the pre-flight passes and the call runs on into signing and broadcast. A third test is the positive control — a matching non-zero price passes the guard and fails at the credit check with `IdentityNotFound`, pinning that the guard rejects only zero and that the pre-flight sits ahead of the credit and signing stages rather than replacing them.

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
```diff
@@ -384,6 +386,104 @@ fn insert_sync_row(rows: &mut BTreeMap<Identifier, DpnsNameStateEntry>, entry: D
...
+    label: &str,
+    previous_rows: &BTreeMap<Identifier, DpnsNameStateEntry>,
+) -> Result<Option<Identifier>, PersistenceError> {
+    let normalized_label = convert_to_homograph_safe_chars(label);
+    let in_memory = previous_rows
+        .values()
+        .find(|entry| {
+            entry.wallet_identity_id == *identity_id && entry.normalized_label == normalized_label
+        })
+        .map(|entry| entry.document_id);
```
thepastaclaw, line 487: **🔴 Blocking: Select the current row from the in-memory snapshot too**

`previous_rows` is keyed by document ID and retains `Sold` and `Transferred` history, so one identity can have both a historical document and a newer `Owned` document for the same normalized label. This `.find()` follows `BTreeMap` document-ID order rather than the deterministic current-row preference now required of persistence backends, and any match prevents the corrected SQLite fallback from running. If the newer document later disappears, recovery can remove the historical row and drop the identity label while leaving the actual current row permanently orphaned. Apply the same `Owned`, `last_synced_at_ms`, and document-ID ordering to the in-memory snapshot.

```suggestion
let in_memory = previous_rows
    .values()
    .filter(|entry| {
        entry.wallet_identity_id == *identity_id
            && entry.normalized_label == normalized_label
    })
    .max_by_key(|entry| {
        (
            matches!(entry.status, DpnsNameSaleStatus::Owned),
            entry.last_synced_at_ms,
            entry.document_id,
        )
    })
    .map(|entry| entry.document_id);
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
```diff
@@ -1521,20 +1618,71 @@ impl IdentityWallet {
...
+                Err(error) => {
+                    tracing::warn!(
+                        identity = %identity_id,
+                        name = label,
+                        "persisted DPNS row lookup failed unrecoverably for a departed name; \
+                         the host mirror may keep a stale row for it until the name is \
+                         re-acquired: {error}"
+                    );
+                    None
+                }
```
thepastaclaw: **🔴 Blocking: Preserve the departure after non-transient persistence read failures**

This arm converts a failed persistence read into `previous_document_id = None`. If Platform then confirms the domain document is absent, the caller removes the identity label without emitting a row-removal delta and reports a successful sync. That label is the only trigger for future departure detection, so the persisted marketplace row remains permanently orphaned even if the underlying problem is later repaired. This is reachable through the newly checked decoder, for example when a negative timestamp produces a fatal `IntegerOverflow`. A non-transient classification means the operation should not be retried indefinitely without surfacing the failure; it does not establish that no row exists. Preserve the label and propagate the persistence failure, or retain a terminal per-item failure without taking the successful-departure path.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

All four previously verified findings are fixed at the current head: the SQLite reader selects deterministically, persisted integers use checked conversions, transient read failures retain departures, and purchase-side zero-price behavior has direct coverage. Two new in-scope blockers remain in departed-name recovery: the in-memory source can select a historical row instead of the current row, and non-transient persistence failures still discard the only future reconciliation trigger.
Source: Codex reviewers (general, security-auditor, rust-quality): gpt-5.6-sol; final verifier: gpt-5.6-sol. openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

### Review provenance

🔴 2 blocking

<details>

```

In `packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs:437-442: Select the current row from the in-memory snapshot too
  `previous_rows` is keyed by document ID and retains `Sold` and `Transferred` history, so one identity can have both a historical document and a newer `Owned` document for the same normalized label. This `.find()` follows `BTreeMap` document-ID order rather than the deterministic current-row preference now required of persistence backends, and any match prevents the corrected SQLite fallback from running. If the newer document later disappears, recovery can remove the historical row and drop the identity label while leaving the actual current row permanently orphaned. Apply the same `Owned`, `last_synced_at_ms`, and document-ID ordering to the in-memory snapshot.
- [BLOCKING] packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs:1675-1684: Preserve the departure after non-transient persistence read failures
  This arm converts a failed persistence read into `previous_document_id = None`. If Platform then confirms the domain document is absent, the caller removes the identity label without emitting a row-removal delta and reports a successful sync. That label is the only trigger for future departure detection, so the persisted marketplace row remains permanently orphaned even if the underlying problem is later repaired. This is reachable through the newly checked decoder, for example when a negative timestamp produces a fatal `IntegerOverflow`. A non-transient classification means the operation should not be retried indefinitely without surfacing the failure; it does not establish that no row exists. Preserve the label and propagate the persistence failure, or retain a terminal per-item failure without taking the successful-departure path.
```
</details>

packages/rs-platform-wallet-storage/src/sqlite/schema/dpns_name_states.rs
thepastaclaw: Resolved in `2f646da` — _Select the current retained row deterministically_ no longer present.

packages/rs-platform-wallet-storage/src/sqlite/schema/dpns_name_states.rs
thepastaclaw: Resolved in `2f646da` — _Use checked conversions for persisted timestamps_ no longer present.

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
thepastaclaw: Resolved in `50002d6` — _Retry departed-name recovery after persistence read failures_ no longer present.

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
thepastaclaw: Resolved in `50002d6` — _Add direct coverage for purchase-side zero-price rejection_ no longer present.

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
bfoss765: Fixed in 386fe6768f. The in-memory snapshot scan now applies the suggested selection — `filter` on identity + normalized label, then `max_by_key((is Owned, last_synced_at_ms, document_id))` — matching the deterministic current-row preference the persistence contract requires of `get_dpns_name_state` backends, so a retained `Sold`/`Transferred` row can no longer shadow the current `Owned` document in `BTreeMap` id order. Added `departed_document_id_prefers_the_current_owned_row_in_the_snapshot` (both document-id orders, historical row deliberately fresher, cross-identity decoy, failing persister proving the in-memory hit still short-circuits) and `departed_document_id_breaks_snapshot_ties_like_the_persistence_contract` (freshness then greatest-id tie-breaks); the first fails against the previous `.find()` code.

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
bfoss765: Fixed in 386fe6768f. The non-transient arm no longer converts the failed read into `previous_document_id = None`: the error is held, and if Platform then confirms the domain document absent, resolution returns a new terminal outcome (`DepartureResolution::Failed`) instead of the successful-departure path — the label and durable row are preserved (so a later pass re-detects the departure and completes the removal once the backend is repaired), no deltas are applied, and the skip is surfaced on the sync summary as a `FailedDpnsDeparture` rather than reported as a clean sync. Sold/Transferred departures, which never read the persisted id, still resolve normally, and the queue is not parked (the failed item is not requeued within the pass; the retained label guarantees re-detection). Tests: `resolve_departed_name_fails_terminally_when_the_persistence_error_is_not_retryable` (fails against the old degrade-to-`None` arm; includes heal-and-resolve) and `resolve_departed_name_defers_a_fatal_persistence_error_until_the_id_is_needed`.

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
```diff
@@ -1459,9 +1625,29 @@ impl IdentityWallet {
...
+                    // failure within this process cannot succeed, and the
+                    // retained label already guarantees re-detection.
+                    DepartureResolution::Failed(error) => {
+                        departures_processed += 1;
+                        summary.departures_failed.push(FailedDpnsDeparture {
+                            identity_id,
+                            label: previous_name.label,
+                            error: error.to_string(),
+                        });
+                        continue;
```
thepastaclaw, line 1676: **🟡 Suggestion: Exercise terminal departure failure through the sync-loop caller**

The new tests call `resolve_departed_name` directly and prove that a fatal mirror read produces `DepartureResolution::Failed`, but they do not exercise this load-bearing caller branch. A regression here could still remove the label, apply a row delta, omit `departures_failed`, or prevent a later pass from rediscovering the departure while every resolver test remains green. Add a `sync_dpns_marketplace` regression with a managed identity carrying the departed label, a confirmed-absent Platform response, and a fatal mirror read. Assert that the label and persisted row remain, no successful departure or row delta is emitted, and `departures_failed` contains the failure; then heal the mirror and verify a later pass completes the removal.

<sub>source: ['codex']</sub>

packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandlerTest.kt
```diff
@@ -2880,4 +2880,362 @@ class PlatformWalletPersistenceHandlerTest {
...
+        assertFalse(db.identityDao().getByIdentityId(identityId)!!.isLocal)
+
+        // The wallet relationship attaches on a later flush — promote.
+        persistIdentitySnapshot(identityId)
+
+        assertTrue(db.identityDao().getByIdentityId(identityId)!!.isLocal)
+    }
+
+    @Test
+    fun loadPathHealsLegacyIsLocalFalseOnWalletLinkedRowsOnly() = runTest {
```
thepastaclaw: **💬 Nitpick: Name the new Kotlin tests with the required `should` prefix**

All 13 tests added in this section use camel-case names such as `consumedAssetLockIsNotRegressedByAStaleNonConsumedUpsert`. The repository testing guideline in `AGENTS.md` requires descriptive test names starting with `should …`. Rename these newly added methods to Kotlin backtick identifiers beginning with `should` while leaving their behavior unchanged.

thepastaclaw (commented):

The two prior blocking findings are fixed at the exact head: in-memory recovery now uses the deterministic current-row preference, and fatal persistence failures preserve the departure instead of orphaning its durable row. Two non-blocking review items remain: the terminal outcome is not tested through the sync-loop caller, and the newly added Kotlin tests do not follow the repository's required `should …` naming convention.
Source: Codex reviewers (general, rust-quality, security-auditor): `gpt-5.6-sol`; final verifier: `gpt-5.6-sol`. `openclaw-agent/cliproxy/gpt-5.6-sol` is orchestration-only and not reviewer evidence.

### Review provenance

<details>

```

In `packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs:1641-1648: Exercise terminal departure failure through the sync-loop caller
  The new tests call `resolve_departed_name` directly and prove that a fatal mirror read produces `DepartureResolution::Failed`, but they do not exercise this load-bearing caller branch. A regression here could still remove the label, apply a row delta, omit `departures_failed`, or prevent a later pass from rediscovering the departure while every resolver test remains green. Add a `sync_dpns_marketplace` regression with a managed identity carrying the departed label, a confirmed-absent Platform response, and a fatal mirror read. Assert that the label and persisted row remain, no successful departure or row delta is emitted, and `departures_failed` contains the failure; then heal the mirror and verify a later pass completes the removal.

In `packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandlerTest.kt`:
- [NITPICK] packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandlerTest.kt:2915-3189: Name the new Kotlin tests with the required `should` prefix
  All 13 tests added in this section use camel-case names such as `consumedAssetLockIsNotRegressedByAStaleNonConsumedUpsert`. The repository testing guideline in `AGENTS.md` requires descriptive test names starting with `should …`. Rename these newly added methods to Kotlin backtick identifiers beginning with `should` while leaving their behavior unchanged.
```
</details>

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
thepastaclaw: Resolved in `386fe67` — _Select the current row from the in-memory snapshot too_ no longer present.

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
thepastaclaw: Resolved in `386fe67` — _Preserve the departure after non-transient persistence read failures_ no longer present.

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
```diff
@@ -262,6 +283,10 @@ pub struct DpnsMarketplaceSyncSummary {
     pub names_added: Vec<(Identifier, String)>,
     /// Names that left a wallet identity since the local snapshot.
     pub names_departed: Vec<DepartedDpnsName>,
+    /// Departures whose resolution failed terminally this pass; their
+    /// labels and durable rows are untouched, so a later pass retries.
+    /// See [`FailedDpnsDeparture`].
+    pub departures_failed: Vec<FailedDpnsDeparture>,
```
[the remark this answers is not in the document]
thepastaclaw: `departures_failed` represents a pass where no marketplace delta was applied, so including it in `is_empty_delta`/`has_delta` would make those state-change predicates report a change that did not occur. The FFI paths also cannot currently produce this outcome: their manager uses `FFIPersister`, which has no DPNS lookup callback and returns `Ok(None)`. The existing `repr(C)` summaries cannot safely gain fields in place; propagation should be added through a versioned ABI when the explicitly deferred persistence read callback lands.

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
bfoss765: Added in de81acbbc765ec3b759e0d2a92c9a6486b4b2cd9: `sync_pass_surfaces_a_terminal_departure_failure_and_completes_it_once_healed` drives a full `sync_dpns_marketplace` pass with a managed identity still carrying the departed label, a mock SDK answering the identity domain page empty and the exact-match domain query confirmed-absent, and a fatally failing mirror read.

It asserts everything on the caller branch: the failure lands in `departures_failed` (with the underlying error), `names_departed` is empty and `is_empty_delta()` holds, the label survives on the managed identity, no DPNS removal delta reaches the persister (`MirrorPersister` now records the name-state changesets handed to `store()`, so the assertion is on what crossed the persistence boundary, not what the summary claims), and `dpns_sync_progress` carries no parked entry. It then heals the mirror and verifies the next pass completes the removal: `names_departed` carries the document id recovered from the mirror, the label drops, and the removal delta is stored.

Verified the test catches the regression it guards: mutating the confirmed-absent arm back to the old degrade-to-`None`/`Resolved` behavior fails it (`departures_failed` empty, label dropped). `cargo test -p platform-wallet --lib dpns_marketplace`: 23 passed.

packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandlerTest.kt
bfoss765: Renamed in 1d83e8fe17b3226ee7a77874699de1ef17dfc75d: all 13 tests now begin with `should`. Used `shouldCamelCase` rather than backtick identifiers to match this file (its existing tests are all camelCase, no backtick names) and the module's other `should`-prefixed tests (`PlatformWalletManagerInitializationTest`). Behavior unchanged; verified with `:sdk:testDebugUnitTest --tests "...PlatformWalletPersistenceHandlerTest"` — 92 tests, 0 failures, all 13 renamed tests executed.

packages/kotlin-sdk/sdk/src/test/kotlin/org/dashfoundation/dashsdk/persistence/PlatformWalletPersistenceHandlerTest.kt
```diff
@@ -2880,4 +2880,362 @@ class PlatformWalletPersistenceHandlerTest {
...
+            identityIndex = 0,
+            amountDuffs = amountDuffs,
+            status = status,
+            proofBytes = proofBytes,
+        )
+        handler.onChangesetEnd(walletId, success = true)
+    }
+
+    @Test
+    fun consumedAssetLockIsNotRegressedByAStaleNonConsumedUpsert() = runTest {
```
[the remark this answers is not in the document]
bfoss765: Renamed in 1d83e8fe17b3226ee7a77874699de1ef17dfc75d: all 13 tests at lines 2915-3189 now begin with `should`. Went with `shouldCamelCase` instead of backtick identifiers to stay consistent with this file's existing camelCase test names and the module's other `should`-prefixed tests (`PlatformWalletManagerInitializationTest`). Verified with `:sdk:testDebugUnitTest` scoped to the class — 92 tests, 0 failures.

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
```diff
@@ -1603,7 +1883,7 @@ impl IdentityWallet {
             },
             entry: status.map(|sale_status| state.to_entry(*identity_id, sale_status, now)),
             remove_document_id: status.is_none().then_some(state.document_id),
-            retry: false,
+            resolution: DepartureResolution::Resolved,
```
thepastaclaw, line 2046: **🔴 Blocking: Remove the recovered document when the label was re-registered**

`previous_document_id` identifies the current row selected from this identity's in-memory snapshot or durable mirror, but this branch discards it whenever an exact-label document currently exists. DPNS domain documents are deletable, and the new persistence contract explicitly handles a label being re-registered under a fresh document ID. If persisted document A was deleted and unrelated document B was registered under the same normalized label, `dpns_name_state` returns B; `classify_departure(B, old_identity)` returns `None`, so this code reports and removes B. The sync caller then drops the old identity's label, while row A remains with no trigger for a future reconciliation pass. When the recovered ID differs from the live state's ID, resolve the prior incarnation by removing A rather than treating B as the departed document. If the prior-ID lookup failed and the replacement makes that ID necessary, preserve the label through `DepartureResolution::Failed`. Add a regression with persisted A and a live, history-unrelated B that asserts A is the removal ID and the replacement is untouched.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The two prior findings are fixed: terminal persistence failures are now tested through the full marketplace sync loop, and all 13 new Kotlin tests use the required `should` prefix. One in-scope blocker remains: when a deleted label has been re-registered under a new document ID, departure resolution deletes the replacement ID and permanently orphans the recovered historical row. Source: Codex reviewers `gpt-5.6-sol` (general, FFI engineer, Rust quality, and security auditor); final verifier `gpt-5.6-sol`. `openclaw-agent/cliproxy/gpt-5.6-sol` is orchestration-only and not reviewer evidence.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — ffi-engineer (completed), `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — rust-quality (completed), `gpt-5.6-sol` — security-auditor (completed)

🔴 1 blocking

<details>

```

In `packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs:1877-1886: Remove the recovered document when the label was re-registered
  `previous_document_id` identifies the current row selected from this identity's in-memory snapshot or durable mirror, but this branch discards it whenever an exact-label document currently exists. DPNS domain documents are deletable, and the new persistence contract explicitly handles a label being re-registered under a fresh document ID. If persisted document A was deleted and unrelated document B was registered under the same normalized label, `dpns_name_state` returns B; `classify_departure(B, old_identity)` returns `None`, so this code reports and removes B. The sync caller then drops the old identity's label, while row A remains with no trigger for a future reconciliation pass. When the recovered ID differs from the live state's ID, resolve the prior incarnation by removing A rather than treating B as the departed document. If the prior-ID lookup failed and the replacement makes that ID necessary, preserve the label through `DepartureResolution::Failed`. Add a regression with persisted A and a live, history-unrelated B that asserts A is the removal ID and the replacement is untouched.
```
</details>

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
bfoss765: Fixed in 5e2c01655932532f13e0bda0c02b9a49bda0da50.

The confirmed-live branch no longer assumes the live document is the departed row. When `classify_departure` yields no sale status:

- **Recovered id differs from the live document's** — the departure now resolves against the recovered prior incarnation: the removal delta and the `DepartedDpnsName` summary carry the prior document id (A), and the re-registered replacement (B) is left untouched. Previously the code reported and removed B while row A was orphaned with no label left to trigger a future pass.
- **Prior-id lookup failed non-retryably** — the id is load-bearing here exactly as under a confirmed-absent document (it decides which row the removal targets), so the departure now takes `DepartureResolution::Failed`: the label and durable row are preserved, the failure is surfaced on the summary, and a later pass completes the removal once the backend is repaired. Sold/Transferred departures still resolve without the id.
- **Ids match, or no prior row is known** — unchanged: the live document's id is removed.

Two regressions cover the seam, both red on the previous behavior: `resolve_departed_name_removes_the_prior_incarnation_when_the_label_was_re_registered` (persisted A, live history-unrelated B; asserts A is the removal id, B untouched, no entry written) and `resolve_departed_name_fails_terminally_when_a_re_registered_label_needs_the_failed_lookup` (fatal mirror read under a live replacement; asserts the terminal failure, then completion after the backend heals). The history `byDocument` query moved into a named builder (`history_by_source_document_query`) so the tests prime the mock SDK with the exact query the production path issues.

bfoss765:
@HashEngineering requesting your review on this one — it's part of the Android-migration estate and is bot-clean/ready for human review. (GitHub won't accept a formal review request yet: your collaborator access on dashpay/platform hasn't been provisioned — flagged to be fixed alongside the #4449 team setup.)

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
```diff
@@ -1590,20 +1907,83 @@ impl IdentityWallet {
...
                 document_id: Some(state.document_id),
-                status,
+                status: Some(sale_status),
             },
-            entry: status.map(|sale_status| state.to_entry(*identity_id, sale_status, now)),
-            remove_document_id: status.is_none().then_some(state.document_id),
-            retry: false,
+            entry: Some(state.to_entry(*identity_id, sale_status, now)),
+            remove_document_id: None,
+            resolution: DepartureResolution::Resolved,
```
thepastaclaw, line 2046: **🔴 Blocking: Retire the recovered incarnation when a replacement has departure history**

This classified-departure branch ignores `previous_document_id`. If persisted owned document A was deleted, the label was re-registered as B, and B subsequently passed through this wallet identity and departed before the next sync, `classify_departure(B, identity_id)` correctly returns `Sold` or `Transferred`. The branch writes B's historical row with `remove_document_id: None`, after which the caller removes the identity label; recovered row A therefore remains permanently persisted as Owned with no label left to trigger reconciliation. When the recovered ID differs from `state.document_id`, write B's classified history entry and remove A in the same changeset. If the persistence lookup failed non-transiently, return `DepartureResolution::Failed` because an unknown prior incarnation may still need removal. Add a regression for persisted owned A plus live B whose history records a departure from this identity.

```suggestion
let previous_document_id = match previous_document_id {
    Ok(document_id) => document_id,
    Err(error) => {
        tracing::warn!(
            identity = %identity_id,
            name = label,
            document = %state.document_id,
            "persisted DPNS row lookup failed unrecoverably for a classified \
             departure; preserving the label and any prior durable row: {error}"
        );
        return ResolvedDepartedName {
            summary: DepartedDpnsName {
                identity_id: *identity_id,
                label: label.to_string(),
                document_id: None,
                status: None,
            },
            entry: None,
            remove_document_id: None,
            resolution: DepartureResolution::Failed(error),
        };
    }
};
let remove_document_id = match previous_document_id {
    Some(document_id) if document_id != state.document_id => Some(document_id),
    _ => None,
};
ResolvedDepartedName {
    summary: DepartedDpnsName {
        identity_id: *identity_id,
        label: label.to_string(),
        document_id: Some(state.document_id),
        status: Some(sale_status),
    },
    entry: Some(state.to_entry(*identity_id, sale_status, now)),
    remove_document_id,
    resolution: DepartureResolution::Resolved,
}
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The prior history-unrelated re-registration defect is fixed at the exact head, but the classified Sold/Transferred branch still leaves a recovered prior incarnation permanently persisted when the replacement document also departed the wallet identity. All 25 focused marketplace tests pass, but none covers this classified re-registration sequence.
Source: Codex reviewers (general, FFI engineer, Rust quality, security auditor): `gpt-5.6-sol`; final verifier: `gpt-5.6-sol`. `openclaw-agent/cliproxy/gpt-5.6-sol` is orchestration-only and not reviewer evidence.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — rust-quality (completed), `gpt-5.6-sol` — ffi-engineer (completed), `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — security-auditor (completed)

🔴 1 blocking

<details>

```

In `packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs:1977-1986: Retire the recovered incarnation when a replacement has departure history
  This classified-departure branch ignores `previous_document_id`. If persisted owned document A was deleted, the label was re-registered as B, and B subsequently passed through this wallet identity and departed before the next sync, `classify_departure(B, identity_id)` correctly returns `Sold` or `Transferred`. The branch writes B's historical row with `remove_document_id: None`, after which the caller removes the identity label; recovered row A therefore remains permanently persisted as Owned with no label left to trigger reconciliation. When the recovered ID differs from `state.document_id`, write B's classified history entry and remove A in the same changeset. If the persistence lookup failed non-transiently, return `DepartureResolution::Failed` because an unknown prior incarnation may still need removal. Add a regression for persisted owned A plus live B whose history records a departure from this identity.
```
</details>

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
thepastaclaw: Resolved in `5e2c016` — _Remove the recovered document when the label was re-registered_ no longer present.

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
bfoss765: Fixed in c7ce550b51fc7bcce90a4b0201ccad0fedae01d9.

The classified Sold/Transferred branch now consumes the recovered `previous_document_id`:

- When the recovered prior incarnation's id differs from the live document's, the resolver emits `remove_document_id: Some(prior)` alongside the classified historical entry, so B's departure row and A's retirement land in the same changeset instead of the label drop orphaning A as `Owned` forever.
- When the persistence lookup failed non-retryably, the branch returns `DepartureResolution::Failed` like the other id-consuming branches — an unknown prior incarnation may still need removal — preserving the label so a later pass completes the retirement once the backend is repaired.

Regressions added, both proven failing against the previous head: `resolve_departed_name_retires_the_prior_incarnation_when_the_replacement_also_departed` (persisted owned A + live B whose transfer history departs this identity → entry keyed under B, removal targets A; previously `remove_document_id` was `None`) and `resolve_departed_name_fails_terminally_when_a_classified_departure_needs_the_failed_lookup` (same shape over a fatally failing mirror → terminal failure, then heals into entry + retirement; previously it resolved with no removal). Full `dpns_marketplace` suite passes (27 tests).

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
```diff
@@ -253,6 +255,25 @@ pub struct DpnsPriceChange {
...
+///
+/// Emitted when the persistence lookup for the departed name's
+/// `document_id` fails with a NON-retryable error while Platform
+/// confirms the domain document is absent: the removal delta cannot be
+/// built and retrying cannot make it buildable, so the departure is NOT
+/// resolved — the identity keeps its label (which is what lets a later
+/// pass re-detect the departure once the backend is repaired) and
+/// nothing is written or removed. Reported on the summary so a pass
+/// that had to skip a departure is distinguishable from a pass that
+/// resolved everything.
```
thepastaclaw, line 268: **🟡 Suggestion: Update the public failure contract to cover every emission path**

`FailedDpnsDeparture` is publicly re-exported, but its documentation says it is emitted only when Platform confirms that the domain document is absent. The resolver now also emits it when a live document is an unrelated re-registration and when a classified Sold/Transferred departure may coexist with an unknown prior incarnation. Consumers relying on this exported contract could therefore misinterpret the failure as proof that the live document was absent. Document the shared invariant instead: a non-retryable persistence lookup failed at a point where the recovered document ID was required to construct safe deltas.

```suggestion
/// A queued departure whose resolution failed terminally this pass.
///
/// Emitted when a non-retryable persistence lookup fails at a point
/// where the recovered `document_id` is required to construct safe
/// deltas: the live document is absent, is an unrelated re-registration,
/// or has classified departure history while a prior incarnation may
/// still exist. The affected row set is therefore unknown, so the
/// identity keeps its label and nothing is written or removed. Reported
/// on the summary so a pass that skipped a departure is distinguishable
/// from one that resolved everything.
```

<sub>source: ['codex']</sub>

thepastaclaw (commented):

The prior blocking classified-departure defect is fixed: the recovered prior incarnation is now retired alongside the replacement's historical entry, and fatal lookup failures preserve the retry trigger; all 27 focused marketplace tests pass. One non-blocking documentation issue remains, while the non-atomic persistence ordering is a concrete pre-existing follow-up and does not affect this review action. Source: Codex general, security-auditor, and rust-quality reviewers: `gpt-5.6-sol`; final verifier: `gpt-5.6-sol`. `openclaw-agent/cliproxy/gpt-5.6-sol` is orchestration-only and not reviewer evidence.

### Review provenance

🟡 1 suggestion(s)

<details>

```

In `packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs:258-268: Update the public failure contract to cover every emission path
  `FailedDpnsDeparture` is publicly re-exported, but its documentation says it is emitted only when Platform confirms that the domain document is absent. The resolver now also emits it when a live document is an unrelated re-registration and when a classified Sold/Transferred departure may coexist with an unknown prior incarnation. Consumers relying on this exported contract could therefore misinterpret the failure as proof that the live document was absent. Document the shared invariant instead: a non-retryable persistence lookup failed at a point where the recovered document ID was required to construct safe deltas.
```
</details>

packages/rs-platform-wallet/src/wallet/identity/network/dpns_marketplace.rs
thepastaclaw: Resolved in `c7ce550` — _Retire the recovered incarnation when a replacement has departure history_ no longer present.

shumkov (approved):
Triage-verified: all four Swift-parity guards and both DPNS defects confirmed real against v4.2-dev; blocking thread addressed in c7ce550b51 with regression coverage.

PR #4424, opened 2026-08-19
fix(platform-wallet): contain panics before the extern "C" abort shim

bfoss765:

`rs-platform-wallet-ffi` has **zero `catch_unwind`**, and every entry point in it is a `pub unsafe extern "C" fn`. `extern "C"` is a **non-unwind ABI**: rustc plants the abort shim in the **callee**, so a panic that reaches the boundary calls `abort()` from inside this library. It never becomes an unwind that anyone can observe.

That is why the Android shim cannot save us. `rs-unified-sdk-jni/src/support.rs:110-125` wraps every JNI export in `catch_unwind`:

```rust
pub fn guard<T>(env: &mut JNIEnv, default: T, f: impl FnOnce(&mut JNIEnv) -> T) -> T {
    match catch_unwind(AssertUnwindSafe(|| f(env))) { ... }
}
```

but the abort fires one frame **below** it, in the `platform-wallet-ffi` callee. The process is already gone before control would return to the guard. So the workspace policy at **`Cargo.toml:84-87`** (`[profile.release-android]`) was not actually enforced anywhere:

> Unlike iOS, panic stays `unwind`: **a JNI library must never abort the app process** — panics are caught at the shim boundary and rethrown as Java exceptions.

The most direct producer was `runtime.rs`:

```rust
rt.block_on(async move { rt.spawn(future).await.expect("tokio worker panicked") })
```

tokio polls the spawned task inside its **own** `catch_unwind`, so the task's panic never unwinds through our frames — it arrives as a `JoinError` **value**. The `.expect` then turned that value back into a *live panic on the calling thread*, which unwound into the entry point's abort shim and SIGABRTed. It also mislabelled a **cancelled** worker as a panic and killed the process over work that merely did not finish.

The exposure is crate-wide (any `unwrap`/index/slice/overflow-check on network-supplied data under any of ~373 exported entry points), and grew with #4348's 22 new marketplace entry points.

### iOS is deliberately untouched

`Cargo.toml:69-71` and `:79-81` set `panic = "abort"` for `[profile.release-ios]` / `[profile.dev-ios]` on purpose. Under those profiles `catch_unwind` still **compiles** — it simply never observes an `Err`, because the panic aborts before unwinding starts. Nothing here is `cfg`-gated on the panic strategy: the same source builds under both, and the iOS carve-out keeps its documented behavior. `cargo check` was run with and without `--features shielded`.

## What was done?

Guarded the crate's **shared execution choke points** rather than mechanically wrapping every entry point. The ~120 entry points that run async work are covered without being edited.

**1. `runtime.rs`**
- `block_on_worker` handles both `JoinError` shapes gracefully (panicked → error value carrying the payload; cancelled → error value saying the work did not complete) and wraps the whole drive in `catch_unwind`. The mapping lives in `from_join_error` so both shapes are directly testable.
- `runtime()` now returns an `FfiRuntime` newtype whose **inherent** `block_on` shadows `tokio::runtime::Runtime::block_on` with a guarded version; every other `Runtime` method resolves through `Deref`. This is why ~46 existing `runtime().block_on(...)` sites are guarded **without being touched**, and why a future entry point gets containment by default. `.raw()` is the explicit opt-out (used by tests, where a panic must still fail the test, and by code already inside `run_on_big_stack_thread`).
- `run_on_big_stack_thread` reports a panicked join through the **same** `io::Result` channel its call sites already map to `ErrorWalletOperation` — so no call site changed.

**2. `panic_guard.rs` (new)** — payload rendering, the `guard_ffi` helper, and the conversion traits. A caught panic becomes the crate's generic **`ErrorWalletOperation` (code 6)** with a message that starts with a machine-readable marker, carries the panic payload, and names the guarded call site (`#[track_caller]`), plus an `ERROR`-level log on `platform_wallet_ffi::panic`.

A panic deliberately **never** borrows a code that carries retry/outcome semantics (`ErrorShieldedBroadcastUnconfirmed`, `ErrorTransactionBroadcastUnconfirmed`, …). It proves nothing about whether an operation reached the network, so it must read to hosts as "unknown outcome".

**3. `FromCaughtPanic` is implemented only where a failure value is honest.** `PlatformWalletFFIResult`, `Result<T, E>` for the crate's error types, and `()` (documented as a last resort). It is **not** implemented for bare value outputs — a balance `u64`, a peer `Vec`, a sync summary — because fabricating one would turn a crash into silent, plausible-looking wrong data, which is worse than the abort this PR removes. The compiler therefore flagged all 24 such sites, and they moved to the explicit `try_block_on` / `try_block_on_worker` helpers, which surface the panic as `Err`. Six `()`-output entry points whose success the host acts on (`identity_sync` register/unregister/update/sync_now, `platform_address_wallet_restore_sync_state`, `core_wallet_signed_payment_release`) were converted too, so they cannot report a false success.

**4. `PlatformWalletError::InternalPanic(String)`** — a typed carrier, rendered as the bare pre-composed message so the marker stays at position 0, and mapped explicitly to code 6 at the boundary. Nothing in `rs-platform-wallet` constructs it.

The **C ABI and the generated cbindgen header are unchanged**: no new error codes, no new exported symbols, no signature changes. The test-only entry point is `#[cfg(test)]` and was verified absent from the generated header.

### Choke-point coverage / residual (honest survey)

Covered:
- 91 `block_on_worker` + 9 `try_block_on_worker` sites
- ~46 `runtime().block_on(...)` sites (via the newtype) + 25 `try_block_on`
- 5 `run_on_big_stack_thread` sites, including the `runtime().block_on` nested inside them

Residual, deliberately **not** closed here:
1. **Purely synchronous entry-point bodies** — a scan of the crate's `extern "C" fn` definitions puts 120 on a guarded choke point and 294 with no async work at all (handle getters, `*_free`, pointer/string conversions, `publish_array`). A panic in one of those still aborts. Closing them needs a `guard_ffi(|| ...)` at each entry point — ~300 mechanical wraps plus a full-body reindent, which is a separate, reviewable-on-its-own change. `guard_ffi` is in place for it.
2. **Code outside the choke point inside an otherwise-guarded entry point** — the `CStr::from_ptr(...)`/`read_identifier(...)` reads before the `block_on` and the FFI-struct conversions after it are not covered by the same guard.
3. **The `()` fallback** (~19 remaining sites, all best-effort `abandon_transaction`/`release`/stop cleanup on paths that already return an error for their own reasons). The panic is logged at ERROR with payload and call site, but the entry point's own result is unaffected. Documented on the impl, with the rule that anything whose success the host acts on must use `try_block_on`.

## How Has This Been Tested?

`cargo test -p platform-wallet-ffi --features shielded` — **328 passing, 0 failing** (296 lib + 26 + 6).

New tests in `runtime.rs`:

- **`panicking_entry_point_returns_an_error_result_instead_of_aborting`** — injects an out-of-bounds read on "network-supplied" data through a `#[cfg(test)] #[no_mangle] unsafe extern "C" fn` shaped exactly like a real entry point. It has to be a genuine `extern "C"` fn for the test to mean anything, because the abort shim is planted in the callee. **The test process surviving is half the assertion**: before this change the panic re-raised on the calling thread, unwound into that shim, and the test binary would die with SIGABRT before any assertion ran. It then asserts code 6, the marker at position 0, and the payload in the message.
- **`join_error_shapes_both_become_error_values`** — both `JoinError` shapes, built from real tokio errors (a panicking `tokio::spawn`, and an aborted `pending()` task). Asserts the cancelled one is *not* reported as a panic — the case the old `.expect` conflated.
- `panicking_worker_maps_result_outputs_to_internal_panic`, `try_block_on_worker_surfaces_a_panic_as_err`, `runtime_block_on_is_guarded_and_passes_values_through`, `run_on_big_stack_thread_reports_a_panic_as_an_io_error`, plus `panic_guard.rs` unit tests for payload rendering and the guard.

A first run of the entry-point test **caught a real defect** in this PR — `InternalPanic`'s `thiserror` prefix double-stamped the message and pushed the marker off position 0 — which is now fixed.

Also verified: `cargo check -p platform-wallet-ffi` (no features) clean; `cargo clippy -p platform-wallet-ffi --features shielded --all-targets` clean; `cargo fmt` applied.

## Breaking Changes

None. No ABI change, no header change, no error-code change. One additive variant on `PlatformWalletError` (a non-`#[non_exhaustive]` enum already matched with catch-alls in-workspace).

Behavior change worth naming: paths that previously **aborted the host process** now return `ErrorWalletOperation`. On the `()`-output cleanup sites in residual item 3 a panic is now logged rather than fatal, which is a deliberate trade — see the doc comment on that impl.

## Checklist:

- [x] I have added tests that prove my fix is effective
- [x] Any dependent changes have been merged and published in downstream modules

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling across wallet, synchronization, transaction, broadcast, and address operations.
  - Runtime failures, worker errors, cancellations, and internal panics are now reported instead of being treated as successful operations.
  - Wallet shutdown continues safely even when cleanup encounters a panic.
- **Reliability**
  - Added safeguards to prevent panics at the FFI boundary and provide clearer wallet-operation errors.
  - Preserved existing behavior for successful operations and missing-resource handling.

### DIFF
--- a/packages/rs-platform-wallet-ffi/src/core_wallet/broadcast.rs
+++ b/packages/rs-platform-wallet-ffi/src/core_wallet/broadcast.rs
@@ -138,11 +138,17 @@ pub unsafe extern "C" fn core_wallet_abandon_signed_transaction(
             "transaction was finalized by a different wallet generation".to_string(),
         );
     }
-    runtime().block_on(
+    // `try_block_on`: releasing the reservation IS this entry point's job, and
+    // the handle was consumed on entry so there is no retry — a panic must not
+    // fall through to a success the host records as "reservation released"
+    // (same contract as `core_wallet_signed_payment_release`). The two abandon
+    // calls in the error arms above stay on the swallowing `block_on`: they are
+    // best-effort cleanup on paths that already report an error of their own.
+    unwrap_result_or_return!(runtime().try_block_on(
         transaction
             .wallet
             .abandon_transaction(&transaction.transaction),
-    );
+    ));
     PlatformWalletFFIResult::ok()
 }
 
@@ -351,6 +357,54 @@ mod tests {
         assert_eq!(rebroadcast.code, PlatformWalletFFIResultCode::NotFound);
         assert!(txid.is_null());
         assert_released(&core, &signer, 47);
+        CORE_WALLET_STORAGE.remove(core_handle);
+    }
+
+    /// The main-path abandon is this entry point's whole job, and the handle
+    /// is consumed on entry — so a panic during the release must surface as
+    /// the typed panic error, never fall through to `ok()` for the host to
+    /// record as "reservation released".
+    ///
+    /// Driving the call from inside a runtime context makes the guarded
+    /// `block_on` panic in the guarded region of the genuine entry point
+    /// ("Cannot start a runtime from within a runtime") before the abandon
+    /// future is ever polled — a real panic, on the real path. Before the
+    /// fix, `block_on`'s `()` recovery swallowed exactly this panic and the
+    /// function reported `Success`.
+    #[test]
+    fn abandon_reports_a_panic_as_an_error_not_success() {
+        let (core, signer) = runtime()
+            .raw()
+            .block_on(funded_spv_core_wallet(StandardAccountType::BIP44Account));
+        let core_handle = CORE_WALLET_STORAGE.insert(core.clone());
+        let transaction_handle = insert(&core, finalize(&core, &signer, 48));
+
+        let result = tokio::runtime::Builder::new_current_thread()
+            .build()
+            .expect("build helper runtime")
+            .block_on(async {
+                unsafe { core_wallet_abandon_signed_transaction(core_handle, transaction_handle) }
+            });
+
+        assert_eq!(
+            result.code,
+            PlatformWalletFFIResultCode::ErrorWalletOperation,
+            "a panicked abandon must not report success — the host would \
+             record the reservation as released"
+        );
+        let message = unsafe { std::ffi::CStr::from_ptr(result.message) }
+            .to_str()
+            .expect("message is UTF-8");
+        assert!(
+            message.starts_with(crate::panic_guard::FFI_PANIC_PREFIX),
+            "message must carry the panic marker: {message}"
+        );
+
+        // The handle was consumed on entry regardless of the panic — a retry
+        // is a clean not-found error, not a second release attempt.
+        let retry =
+            unsafe { core_wallet_abandon_signed_transaction(core_handle, transaction_handle) };
+        assert_eq!(retry.code, PlatformWalletFFIResultCode::NotFound);
         CORE_WALLET_STORAGE.remove(core_handle);
     }
 }
--- a/packages/rs-platform-wallet-ffi/src/core_wallet/broadcast.rs
+++ b/packages/rs-platform-wallet-ffi/src/core_wallet/broadcast.rs
@@ -76,11 +76,11 @@ pub unsafe extern "C" fn core_wallet_broadcast_signed_transaction(
     // exclusive side, so it cannot interleave between the check and the send.
     // Scoped per generation, so this send — up to the broadcaster's timeout —
     // blocks only THIS wallet's teardown, never an unrelated wallet's.
-    let (_lifecycle, wallet_is_live) = runtime().block_on(async {
+    let (_lifecycle, wallet_is_live) = unwrap_result_or_return!(runtime().try_block_on(async {
         let gate = wallet.generation_payment_guard().await;
         let live = wallet.is_current_generation().await;
         (gate, live)
-    });
+    }));
     if !wallet_is_live {
         runtime().block_on(finalized.wallet.abandon_transaction(&finalized.transaction));
         return PlatformWalletFFIResult::err(
@@ -264,6 +264,7 @@ mod tests {
 
     fn finalize(core: &TestCore, signer: &WalletSigner, tag: u8) -> SignedCoreTransaction {
         runtime()
+            .raw()
             .block_on(core.finalize_transaction(
                 TransactionBuilder::new().add_output(
                     &Address::dummy(Network::Testnet, usize::from(tag)),
@@ -285,13 +286,14 @@ mod tests {
 
     fn assert_released(core: &TestCore, signer: &WalletSigner, tag: u8) {
         let retry = finalize(core, signer, tag);
-        runtime().block_on(core.abandon_transaction(&retry));
+        runtime().raw().block_on(core.abandon_transaction(&retry));
     }
 
     #[test]
     fn double_free_is_safe_and_releases_reservation() {
-        let (core, signer) =
-            runtime().block_on(funded_spv_core_wallet(StandardAccountType::BIP44Account));
+        let (core, signer) = runtime()
+            .raw()
+            .block_on(funded_spv_core_wallet(StandardAccountType::BIP44Account));
         let transaction_handle = insert(&core, finalize(&core, &signer, 40));
 
         core_wallet_signed_transaction_free(transaction_handle);
@@ -302,8 +304,9 @@ mod tests {
 
     #[test]
     fn invalid_or_wrong_wallet_consumes_and_releases() {
-        let (origin, origin_signer) =
-            runtime().block_on(funded_spv_core_wallet(StandardAccountType::BIP44Account));
+        let (origin, origin_signer) = runtime()
+            .raw()
+            .block_on(funded_spv_core_wallet(StandardAccountType::BIP44Account));
         let invalid_transaction = insert(&origin, finalize(&origin, &origin_signer, 42));
         let invalid =
             unsafe { core_wallet_abandon_signed_transaction(u64::MAX, invalid_transaction) };
@@ -313,8 +316,9 @@ mod tests {
         );
         assert_released(&origin, &origin_signer, 43);
 
-        let (other, _) =
-            runtime().block_on(funded_spv_core_wallet(StandardAccountType::BIP44Account));
+        let (other, _) = runtime()
+            .raw()
+            .block_on(funded_spv_core_wallet(StandardAccountType::BIP44Account));
         let other_handle = CORE_WALLET_STORAGE.insert(other);
         let wrong_transaction = insert(&origin, finalize(&origin, &origin_signer, 44));
         let wrong =
@@ -329,8 +333,9 @@ mod tests {
 
     #[test]
     fn abandon_then_free_or_broadcast_cannot_reconsume_handle() {
-        let (core, signer) =
-            runtime().block_on(funded_spv_core_wallet(StandardAccountType::BIP44Account));
+        let (core, signer) = runtime()
+            .raw()
+            .block_on(funded_spv_core_wallet(StandardAccountType::BIP44Account));
         let core_handle = CORE_WALLET_STORAGE.insert(core.clone());
         let transaction_handle = insert(&core, finalize(&core, &signer, 46));
 
--- a/packages/rs-platform-wallet-ffi/src/core_wallet/broadcast.rs
+++ b/packages/rs-platform-wallet-ffi/src/core_wallet/broadcast.rs
@@ -1,23 +1,31 @@
 //! FFI bindings for CoreWallet transaction broadcasting.
 
+use crate::core_wallet::lifecycle::lifecycle_gate_or_release;
 use crate::error::*;
 use crate::handle::*;
+use crate::panic_guard::GuardedError;
 use crate::runtime::runtime;
 use crate::{check_ptr, unwrap_option_or_return, unwrap_result_or_return};
 use platform_wallet::PlatformWalletError;
 use std::os::raw::c_char;
 
 fn classify_broadcast_result(
-    result: Result<dashcore::Txid, PlatformWalletError>,
+    result: Result<dashcore::Txid, GuardedError<PlatformWalletError>>,
     local_txid: dashcore::Txid,
 ) -> (Option<dashcore::Txid>, PlatformWalletFFIResult) {
     match result {
         Ok(_) => (Some(local_txid), PlatformWalletFFIResult::ok()),
-        Err(error @ PlatformWalletError::TransactionBroadcast(_))
-        | Err(error @ PlatformWalletError::TransactionBroadcastUnconfirmed(_)) => {
-            (Some(local_txid), error.into())
-        }
-        Err(error) => (None, error.into()),
+        // A boundary failure (caught panic / no runtime) carries NO outcome
+        // guarantee, so it gets the same shape as any other unclassified
+        // failure: the generic code, its own message verbatim, and no txid —
+        // it must not borrow the `Some(local_txid)` treatment that says
+        // "this reached the network, reconcile by this id".
+        Err(GuardedError::Boundary(error)) => (None, error.into()),
+        Err(GuardedError::Domain(error @ PlatformWalletError::TransactionBroadcast(_)))
+        | Err(GuardedError::Domain(
+            error @ PlatformWalletError::TransactionBroadcastUnconfirmed(_),
+        )) => (Some(local_txid), error.into()),
+        Err(GuardedError::Domain(error)) => (None, error.into()),
     }
 }
 
@@ -76,11 +84,11 @@ pub unsafe extern "C" fn core_wallet_broadcast_signed_transaction(
     // exclusive side, so it cannot interleave between the check and the send.
     // Scoped per generation, so this send — up to the broadcaster's timeout —
     // blocks only THIS wallet's teardown, never an unrelated wallet's.
-    let (_lifecycle, wallet_is_live) = unwrap_result_or_return!(runtime().try_block_on(async {
-        let gate = wallet.generation_payment_guard().await;
-        let live = wallet.is_current_generation().await;
-        (gate, live)
-    }));
+    let (_lifecycle, wallet_is_live) = unwrap_result_or_return!(lifecycle_gate_or_release(
+        &wallet,
+        &finalized.wallet,
+        &finalized.transaction
+    ));
     if !wallet_is_live {
         runtime().block_on(finalized.wallet.abandon_transaction(&finalized.transaction));
         return PlatformWalletFFIResult::err(
@@ -210,6 +218,13 @@ mod outcome_tests {
         dashcore::Txid::from_byte_array([byte; 32])
     }
 
+    /// A domain error in the guarded shape the classifier now takes.
+    fn domain(
+        error: PlatformWalletError,
+    ) -> Result<dashcore::Txid, GuardedError<PlatformWalletError>> {
+        Err(GuardedError::Domain(error))
+    }
+
     #[test]
     fn network_outcomes_all_carry_a_txid() {
         let accepted = classify_broadcast_result(Ok(txid(1)), txid(9));
@@ -217,7 +232,7 @@ mod outcome_tests {
         assert_eq!(accepted.1.code, PlatformWalletFFIResultCode::Success);
 
         let rejected = classify_broadcast_result(
-            Err(PlatformWalletError::TransactionBroadcast(
+            domain(PlatformWalletError::TransactionBroadcast(
                 "rejected".to_string(),
             )),
             txid(2),
@@ -229,7 +244,7 @@ mod outcome_tests {
         );
 
         let unknown = classify_broadcast_result(
-            Err(PlatformWalletError::TransactionBroadcastUnconfirmed(
+            domain(PlatformWalletError::TransactionBroadcastUnconfirmed(
                 "timeout".to_string(),
             )),
             txid(3),
@@ -244,7 +259,7 @@ mod outcome_tests {
     #[test]
     fn operational_error_does_not_carry_a_txid() {
         let outcome = classify_broadcast_result(
-            Err(PlatformWalletError::TransactionBuild("invalid".to_string())),
+            domain(PlatformWalletError::TransactionBuild("invalid".to_string())),
             txid(4),
         );
         assert_eq!(outcome.0, None);
@@ -264,6 +279,7 @@ mod tests {
     use platform_wallet::{CoreWallet, SignedCoreTransaction};
 
     use super::*;
+    use crate::core_wallet::lifecycle::arm_lifecycle_gate_panic;
     use crate::core_wallet::FFICoreSignedTransaction;
 
     type TestCore = CoreWallet<platform_wallet::broadcaster::SpvBroadcaster>;
@@ -406,5 +422,97 @@ mod tests {
             unsafe { core_wallet_abandon_signed_transaction(core_handle, transaction_handle) };
         assert_eq!(retry.code, PlatformWalletFFIResultCode::NotFound);
         CORE_WALLET_STORAGE.remove(core_handle);
+    }
+
+    /// A lifecycle-gate failure must not strand the build's UTXO reservation.
+    ///
+    /// The gate is taken *after* `finalize_transaction` has reserved the
+    /// inputs, and on this path the transaction handle has already been
+    /// consumed on entry — so if the guarded acquisition fails and the entry
+    /// point just returns, the host is left holding neither handle nor token,
+    /// and nothing in the process can ever release those inputs again. The
+    /// wallet would silently lose that much spendable balance until restart
+    /// (`dashpay/platform#4424` review).
+    ///
+    /// `assert_released` is the proof: it finalizes a *second* transaction
+    /// against the same account, which can only fund if the first build's
+    /// inputs came back.
+    #[test]
+    fn a_lifecycle_gate_panic_releases_the_reservation_and_reports_the_panic() {
+        let (core, signer) = runtime()
+            .raw()
+            .block_on(funded_spv_core_wallet(StandardAccountType::BIP44Account));
+        let core_handle = CORE_WALLET_STORAGE.insert(core.clone());
+        let transaction_handle = insert(&core, finalize(&core, &signer, 50));
+        let mut out_txid: *mut c_char = std::ptr::null_mut();
+
+        arm_lifecycle_gate_panic();
+        let result = unsafe {
+            core_wallet_broadcast_signed_transaction(core_handle, transaction_handle, &mut out_txid)
+        };
+
+        assert_eq!(
+            result.code,
+            PlatformWalletFFIResultCode::ErrorWalletOperation,
+            "a caught panic must arrive as the generic code"
+        );
+        let message = unsafe { std::ffi::CStr::from_ptr(result.message) }
+            .to_str()
+            .expect("message is UTF-8");
+        assert_eq!(
+            message.find(crate::panic_guard::FFI_PANIC_PREFIX),
+            Some(0),
+            "the marker must be at position 0: {message}"
+        );
+        assert!(
+            message.contains("injected lifecycle-gate panic"),
+            "the payload must survive: {message}"
+        );
+        assert!(
+            out_txid.is_null(),
+            "nothing was broadcast, so no txid may be published"
+        );
+
+        // The reservation came back: a fresh build on the same account funds.
+        assert_released(&core, &signer, 51);
+
+        // The handle was consumed on entry, panic or not.
+        let retry = unsafe {
+            core_wallet_broadcast_signed_transaction(core_handle, transaction_handle, &mut out_txid)
+        };
+        assert_eq!(retry.code, PlatformWalletFFIResultCode::NotFound);
+        CORE_WALLET_STORAGE.remove(core_handle);
+    }
+
+    /// The same guarantee stated at the seam the other two entry points share
+    /// (`core_wallet_tx_builder_finalize` and
+    /// `core_wallet_signed_payment_finalize` reach it with an unpublished
+    /// handle / unregistered token respectively). Those two need a live
+    /// `MnemonicResolverHandle` vtable to drive end-to-end, so the shared
+    /// helper is pinned directly instead.
+    #[test]
+    fn lifecycle_gate_helper_releases_before_returning_the_failure() {
+        let (core, signer) = runtime()
+            .raw()
+            .block_on(funded_spv_core_wallet(StandardAccountType::BIP44Account));
+        let finalized = finalize(&core, &signer, 52);
+
+        arm_lifecycle_gate_panic();
+        let error = lifecycle_gate_or_release(&core, &core, &finalized)
+            .expect_err("the armed panic must surface as a boundary failure");
+        assert_eq!(
+            error.to_string().find(crate::panic_guard::FFI_PANIC_PREFIX),
+            Some(0)
+        );
+
+        assert_released(&core, &signer, 53);
+
+        // The happy path still hands back a held gate and a live generation.
+        let (_gate, live) = lifecycle_gate_or_release(&core, &core, &finalized)
+            .expect("acquisition must succeed when nothing is armed");
+        assert!(live, "a freshly built wallet is its own current generation");
+        runtime()
+            .raw()
+            .block_on(core.abandon_transaction(&finalized));
     }
 }
--- a/packages/rs-platform-wallet-ffi/src/core_wallet/transaction_builder.rs
+++ b/packages/rs-platform-wallet-ffi/src/core_wallet/transaction_builder.rs
@@ -1,3 +1,4 @@
+use crate::core_wallet::lifecycle::lifecycle_gate_or_release;
 use crate::core_wallet_types::OutPointFFI;
 use crate::error::*;
 use crate::handle::{Handle, CORE_SIGNED_TRANSACTION_STORAGE, PLATFORM_WALLET_STORAGE};
@@ -165,11 +166,11 @@ pub unsafe extern "C" fn core_wallet_tx_builder_finalize(
     // the signer await, never around it: holding it across an open signing prompt
     // would stall this wallet's teardown for as long as the user takes, and the
     // check makes that unnecessary.
-    let (_lifecycle, wallet_is_live) = unwrap_result_or_return!(runtime().try_block_on(async {
-        let gate = wallet.core().generation_payment_guard().await;
-        let live = wallet.core().is_current_generation().await;
-        (gate, live)
-    }));
+    let (_lifecycle, wallet_is_live) = unwrap_result_or_return!(lifecycle_gate_or_release(
+        wallet.core(),
+        wallet.core(),
+        &finalized
+    ));
     if !wallet_is_live {
         // No handle was published, so nothing would ever release this build's
         // reservation. Reconcile it here: the release is generation-bound, so on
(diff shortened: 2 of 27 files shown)


thepastaclaw:
✅ Final review complete — no blockers (commit 84f469e0)

bfoss765:
Follow-up pushed in 8cadc68: I missed a sibling of the `core_wallet_signed_payment_release` conversion. `core_wallet_abandon_signed_transaction`'s main path was still on the swallowing `block_on` — its `()` recovery eats a caught panic and the function falls through to `ok()`, so the host (JNI throws on non-Success; Swift `.check()`s) would affirmatively record "reservation released" for a release that panicked. Since the handle is consumed on entry, there's no retry either. Releasing is this entry point's whole job, so it gets the identical treatment: `try_block_on` + `unwrap_result_or_return!`, surfacing the panic as code 6 with the `FFI_PANIC_PREFIX` marker. The abandons in the invalid-handle / generation-mismatch arms deliberately stay on the swallowing form — they're best-effort cleanup on paths that already return their own error, which is exactly the documented `()` carve-out.

The new regression test drives the real entry point from inside a runtime context so the guarded `block_on` itself panics in the guarded region; before the fix it observes `Success`, after it the typed panic error, and the consumed-handle contract (retry → `NotFound`) holds across the panic. I also left `platform_wallet_manager_spv_stop` as-is per the review discussion (pre-existing lossy contract) but added a comment stating the lossiness is deliberate so the next audit doesn't re-flag it.

packages/rs-platform-wallet-ffi/src/panic_guard.rs
```diff
@@ -0,0 +1,284 @@
...
+impl FromCaughtPanicError for String {
+    fn from_caught_panic_error(message: String) -> Self {
+        message
+    }
+}
+
+impl FromCaughtPanicError for Box<dyn std::error::Error + Send + Sync> {
+    fn from_caught_panic_error(message: String) -> Self {
+        message.into()
+    }
```
thepastaclaw: **🟡 Suggestion: Caught panics lose the documented code-and-prefix contract**

Converting a panic into the future's ordinary error type only preserves the promised `ErrorWalletOperation` code and position-zero marker when every later caller forwards that error unchanged. Existing callers do not. `platform_wallet_load_and_apply_persisted` receives the boxed string produced here, then `From<Box<dyn Error>>` maps it to `ErrorUnknown` (99) and prefixes `unclassified error: `. The `String` path in `core_wallet/transaction_builder.rs` prefixes `add_inputs_from_outpoints failed:`, while catch-all `PlatformWalletError` handlers such as `shielded_send.rs:609-612` prepend operation context. In each case the process survives, but hosts cannot classify the panic using the documented code-6/position-zero contract. Carry caught panics in an outer FFI-local result, such as `try_block_on`/`try_block_on_worker`, and intercept that outer `PlatformWalletError::InternalPanic` before legacy domain-error mapping. Add exported-path tests for both the code and exact prefix position.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/core_wallet/transaction_builder.rs
```diff
@@ -165,11 +165,11 @@ pub unsafe extern "C" fn core_wallet_tx_builder_finalize(
     // the signer await, never around it: holding it across an open signing prompt
     // would stall this wallet's teardown for as long as the user takes, and the
     // check makes that unnecessary.
-    let (_lifecycle, wallet_is_live) = runtime().block_on(async {
+    let (_lifecycle, wallet_is_live) = unwrap_result_or_return!(runtime().try_block_on(async {
         let gate = wallet.core().generation_payment_guard().await;
         let live = wallet.core().is_current_generation().await;
         (gate, live)
-    });
+    }));
```
thepastaclaw: **🟡 Suggestion: Release the reservation before returning a lifecycle-check failure**

These newly fallible lifecycle checks run after finalization has acquired a transaction reservation. If `try_block_on` catches a panic, `unwrap_result_or_return!` consumes the error path without publishing a transaction handle and without calling `abandon_transaction`, leaving the host no token or handle with which to release the reservation. Handle the outer `Err` explicitly and make a best-effort abandonment before returning the original panic error. The same ownership gap exists at `transaction_builder.rs:302-306` and `core_wallet/broadcast.rs:79-83`, using `finalized` and `finalized.transaction` respectively.

packages/rs-platform-wallet/src/error.rs
```diff
@@ -579,6 +579,30 @@ pub enum PlatformWalletError {
...
+    /// (`panic_guard::FromCaughtPanicError`); nothing in this crate returns
+    /// it, and nothing should match on it to make a retry decision.
+    ///
+    /// Rendered as the bare payload: the guard composes the whole message,
+    /// starting with its machine-readable marker
+    /// (`panic_guard::FFI_PANIC_PREFIX`). Adding a `thiserror` prefix here
+    /// would push that marker off position 0 and duplicate the text on every
+    /// path that renders through `Display`.
+    #[error("{0}")]
+    InternalPanic(String),
```
thepastaclaw: **🟡 Suggestion: FFI panic handling introduces a source-breaking lower-layer error variant**

`PlatformWalletError` is a public enum without `#[non_exhaustive]`, so adding `InternalPanic` breaks exhaustive matches in downstream Rust consumers. It also exposes a boundary-specific failure in the lower-level `platform-wallet` domain API even though the documentation says only `platform-wallet-ffi` constructs it. This conflicts with the PR's statement that it has no breaking changes. Keep the panic carrier in an FFI-local outer result instead of extending the wallet domain error enum, or explicitly account for the public API break.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/panic_guard.rs
```diff
@@ -0,0 +1,284 @@
...
+}
+
+/// Compose the FFI-visible message for a caught panic: the marker prefix, the
+/// guarded call site, and the panic payload.
+pub(crate) fn ffi_panic_message(location: &Location<'_>, detail: &str) -> String {
+    format!(
+        "{FFI_PANIC_PREFIX}{detail} (guarded call site: {}:{})",
+        location.file(),
+        location.line()
+    )
```
thepastaclaw, line 129: **🟡 Suggestion: NUL-containing panic payloads discard the machine-readable marker**

Rust panic strings can contain interior NUL bytes. `ffi_panic_message` preserves such bytes, but `PlatformWalletFFIResult::err` cannot construct a `CString` from the result and replaces the entire message with `<invalid UTF-8>`. The panic is contained, but the marker, payload, and call site are all discarded, violating this module's stated contract for synthesized panic messages. Escape NUL bytes before composing the C-facing message and add a regression test using a NUL-containing panic payload.

```suggestion
pub(crate) fn ffi_panic_message(location: &Location<'_>, detail: &str) -> String {
    let detail = detail.replace('\0', "\\0");
    format!(
        "{FFI_PANIC_PREFIX}{detail} (guarded call site: {}:{})",
        location.file(),
        location.line()
    )
}
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The new runtime guards contain the principal Tokio worker and big-stack-thread panic paths, but initial runtime construction still happens before any guard and can therefore abort the host on the first async FFI call. Several result carriers also lose the documented code-6/position-zero panic marker, and newly recoverable lifecycle failures can strand finalized transaction reservations.
Source: reviewer backends gpt-5.6-sol (general, ffi-engineer, rust-quality, security-auditor); final verifier backend gpt-5.6-sol. openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — ffi-engineer (completed), `gpt-5.6-sol` — rust-quality (completed), `gpt-5.6-sol` — security-auditor (completed)

<details>

```

In `packages/rs-platform-wallet-ffi/src/runtime.rs`:
- [BLOCKING] packages/rs-platform-wallet-ffi/src/runtime.rs:115-121: Runtime initialization can still panic before the guard
  The first call to `runtime()` evaluates this `Lazy` before `FfiRuntime::block_on`, `try_block_on`, or `block_on_worker` can enter `guard_ffi`. Runtime construction remains fallible: Tokio's driver initialization can return an `io::Error`, which this `expect` converts to a panic, and Tokio's multi-thread worker launch can itself panic when the OS cannot create a worker thread. Either panic therefore reaches the calling `extern "C"` function's non-unwind abort shim and terminates the host. This leaves every async entry point vulnerable on its first runtime use, directly contradicting the containment goal. Make runtime acquisition fallible and perform or force initialization inside a guarded operation so initialization failures reach the FFI error channel.

In `packages/rs-platform-wallet-ffi/src/panic_guard.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/panic_guard.rs:130-169: Caught panics lose the documented code-and-prefix contract
  Converting a panic into the future's ordinary error type only preserves the promised `ErrorWalletOperation` code and position-zero marker when every later caller forwards that error unchanged. Existing callers do not. `platform_wallet_load_and_apply_persisted` receives the boxed string produced here, then `From<Box<dyn Error>>` maps it to `ErrorUnknown` (99) and prefixes `unclassified error: `. The `String` path in `core_wallet/transaction_builder.rs` prefixes `add_inputs_from_outpoints failed:`, while catch-all `PlatformWalletError` handlers such as `shielded_send.rs:609-612` prepend operation context. In each case the process survives, but hosts cannot classify the panic using the documented code-6/position-zero contract. Carry caught panics in an outer FFI-local result, such as `try_block_on`/`try_block_on_worker`, and intercept that outer `PlatformWalletError::InternalPanic` before legacy domain-error mapping. Add exported-path tests for both the code and exact prefix position.
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/panic_guard.rs:75-80: NUL-containing panic payloads discard the machine-readable marker
  Rust panic strings can contain interior NUL bytes. `ffi_panic_message` preserves such bytes, but `PlatformWalletFFIResult::err` cannot construct a `CString` from the result and replaces the entire message with `<invalid UTF-8>`. The panic is contained, but the marker, payload, and call site are all discarded, violating this module's stated contract for synthesized panic messages. Escape NUL bytes before composing the C-facing message and add a regression test using a NUL-containing panic payload.

- [SUGGESTION] packages/rs-platform-wallet-ffi/src/core_wallet/transaction_builder.rs:168-172: Release the reservation before returning a lifecycle-check failure
  These newly fallible lifecycle checks run after finalization has acquired a transaction reservation. If `try_block_on` catches a panic, `unwrap_result_or_return!` consumes the error path without publishing a transaction handle and without calling `abandon_transaction`, leaving the host no token or handle with which to release the reservation. Handle the outer `Err` explicitly and make a best-effort abandonment before returning the original panic error. The same ownership gap exists at `transaction_builder.rs:302-306` and `core_wallet/broadcast.rs:79-83`, using `finalized` and `finalized.transaction` respectively.

- [SUGGESTION] packages/rs-platform-wallet/src/error.rs:583-605: FFI panic handling introduces a source-breaking lower-layer error variant
  `PlatformWalletError` is a public enum without `#[non_exhaustive]`, so adding `InternalPanic` breaks exhaustive matches in downstream Rust consumers. It also exposes a boundary-specific failure in the lower-level `platform-wallet` domain API even though the documentation says only `platform-wallet-ffi` constructs it. This conflicts with the PR's statement that it has no breaking changes. Keep the panic carrier in an FFI-local outer result instead of extending the wallet domain error enum, or explicitly account for the public API break.
```
</details>

packages/rs-platform-wallet-ffi/src/panic_guard.rs
```diff
@@ -0,0 +1,284 @@
...
+    /// `ERROR` log — payload plus guarded call site — is the whole record,
+    /// and the entry point goes on to report whatever it was going to report.
+    ///
+    /// That makes a swallowed panic look like success to the host, so it is
+    /// **only** acceptable where the call is already best-effort cleanup on a
+    /// path that returns an error for its own reasons. Anything whose success
+    /// the host acts on must use [`crate::runtime::FfiRuntime::try_block_on`]
+    /// (or [`crate::runtime::try_block_on_worker`]) and report the `Err`.
+    fn from_caught_panic(_message: String) -> Self {}
+}
```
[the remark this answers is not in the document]
thepastaclaw: The `SpvRuntime::stop()` result was already discarded at the merge base, before this panic-containment change. This entry point intentionally remains best-effort teardown, and the head now documents that hosts do not act on its outcome. Propagating ordinary stop errors would change that pre-existing API contract rather than fix an issue introduced by this PR.

bfoss765:
On the runtime-initialization blocker (runtime.rs:115-121): fixed in 84f469e. You were right that this was reachable on the first async call: `runtime()` handed out a `Lazy` that `expect`ed the build, and that `Lazy` was evaluated at the top of the entry point, outside every guard.

Acquisition is fallible now — `runtime_checked() -> Result<&'static FfiRuntime, FfiBoundaryError>` — and the one-time construction happens inside `guard_ffi`, which covers both shapes you named: `Builder::build` reports driver init as an `io::Error` (no `expect` left to convert it into a panic), and a panic out of tokio's worker launch is caught rather than propagated.

To make "all entry paths route through the fallible acquisition" structural rather than a review invariant, `runtime()` now returns a zero-sized `RuntimeHandle` that holds no runtime at all, so merely calling it cannot build or panic; every method on it acquires through `runtime_checked` and folds a failure into that entry point's own error channel. The 75 `block_on`/`try_block_on` call sites were untouched; the 19 `enter`/`spawn`/`spawn_blocking`/`raw` sites were each converted to handle the failure. As a side effect the `Lazy` closure is now infallible in the panic sense, so it can no longer poison and re-panic on a later access either. Pinned by `runtime_acquisition_is_fallible_and_succeeds_here` and `an_unavailable_runtime_is_a_generic_error_result_not_an_abort`.

packages/rs-platform-wallet-ffi/src/panic_guard.rs
bfoss765: Fixed in 84f469e, and your diagnosis found more than the three sites you listed.

Caught panics now ride an FFI-local outer result. `guard_ffi` returns `FfiOutcome<T>`; the guarded helpers flatten it into each call site's own channel as `GuardedError::{Boundary, Domain}`, and `From<GuardedError<E>> for PlatformWalletFFIResult` answers the `Boundary` arm itself — code 6, message verbatim — before delegating to the legacy `From<E>` mapping. So `From<Box<dyn Error>>` never sees a panic, and `platform_wallet_load_and_apply_persisted` no longer answers `ErrorUnknown` (99) behind `unclassified error: `.

The hand-written context sites needed more than that, because a panic reaching `format!("{operation} failed: {e}")` would still be re-prefixed. Rather than fix them by inspection, I denied `GuardedError` a `Display` impl: you cannot interpolate a guarded error into a context string at all, you have to say which failure you mean first (via `peel_boundary`, which returns the domain result and hands the boundary failure straight back as a finished FFI result). That turned the whole class into a compile error — and surfaced seven sites beyond the three you named: `asset_lock/sync.rs` catch-up, `shielded_send.rs` seed-pool-notes, four in `shielded_sync.rs`, and `wallet_startup.rs`, which was labelling a caught panic `"failed to spawn the startup thread: "`.

Exported-path tests as requested: `exported_domain_result_entry_point_pins_code_six_and_prefix_at_position_zero` drives a genuine `extern "C"` entry point whose async body returns a domain `Result` and asserts `message.find(FFI_PANIC_PREFIX) == Some(0)` — exact position, not `contains` — alongside code 6. `mappers_pass_a_caught_panic_through_at_code_six_with_the_marker_first` pins the same through both `shielded_send` mappers, including that the operation context is *not* prepended.

packages/rs-platform-wallet/src/error.rs
bfoss765: Agreed, and fixed in 84f469e: `InternalPanic` is removed from `PlatformWalletError` entirely. `packages/rs-platform-wallet/src/error.rs` is now byte-identical to the merge base, so the PR's no-breaking-changes claim holds as written.

The carrier is FFI-local, as you suggested: `FfiOutcome`/`FfiBoundaryError` in `platform-wallet-ffi`, flattened at call sites as `GuardedError<E>`. `FfiBoundaryError` has exactly one conversion — to `PlatformWalletFFIResult`, verbatim, at code 6 — and deliberately no conversion into any domain error type, which is what structurally prevents this from being reintroduced. The mapping table in the FFI `error.rs` keeps a comment recording why the arm is absent.

packages/rs-platform-wallet-ffi/src/core_wallet/transaction_builder.rs
bfoss765: Confirmed and fixed in 84f469e at all three sites you identified (`transaction_builder.rs:168-172` and `:302-306` with `finalized`, `broadcast.rs:79-83` with `finalized.transaction`).

Rather than repeat the compensation three times, the release now lives next to the acquisition that can fail: `core_wallet/lifecycle.rs::lifecycle_gate_or_release` takes the gate, reads liveness, and on a guarded failure abandons the finalized transaction before returning the original error. It takes `gate_on` and `release_on` separately because the broadcast path gates on the caller's handle but must act through the transaction's own originating wallet. The abandon stays on the swallowing `block_on` — this path is already returning an error of its own and a second failure must not mask the first — and the release is generation-bound, so it remains a logged no-op on a genuine removal and correctly declines to touch a re-created generation's inputs.

Two tests. `a_lifecycle_gate_panic_releases_the_reservation_and_reports_the_panic` drives the real exported `core_wallet_broadcast_signed_transaction`: it asserts code 6 with the marker at position 0, that no txid is published, that the handle stays consumed — and that the reservation came back, by finalizing a *second* transaction on the same account, which can only fund if the first build's inputs were released. `lifecycle_gate_helper_releases_before_returning_the_failure` pins the same guarantee at the shared seam for the two finalize entry points.

One thing worth calling out: the panic is injected through a `#[cfg(test)]` thread-local hook. The obvious alternative — driving the entry point from inside a runtime context, as the existing abandon test does — makes the compensating `abandon_transaction` panic as well, so it could never show the reservation returning. The hook makes only the gate acquisition fail, on the real code path.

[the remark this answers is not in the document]
bfoss765: Confirmed and fixed in 84f469e at all three sites you identified (`transaction_builder.rs:168-172` and `:302-306` with `finalized`, `broadcast.rs:79-83` with `finalized.transaction`).

Rather than repeat the compensation three times, the release now lives next to the acquisition that can fail: `core_wallet/lifecycle.rs::lifecycle_gate_or_release` takes the gate, reads liveness, and on a guarded failure abandons the finalized transaction before returning the original error. It takes `gate_on` and `release_on` separately because the broadcast path gates on the caller's handle but must act through the transaction's own originating wallet. The abandon stays on the swallowing `block_on` — this path is already returning an error of its own and a second failure must not mask the first — and the release is generation-bound, so it remains a logged no-op on a genuine removal and correctly declines to touch a re-created generation's inputs.

Two tests. `a_lifecycle_gate_panic_releases_the_reservation_and_reports_the_panic` drives the real exported `core_wallet_broadcast_signed_transaction`: it asserts code 6 with the marker at position 0, that no txid is published, that the handle stays consumed — and that the reservation came back, by finalizing a *second* transaction on the same account, which can only fund if the first build's inputs were released. `lifecycle_gate_helper_releases_before_returning_the_failure` pins the same guarantee at the shared seam for the two finalize entry points.

One thing worth calling out: the panic is injected through a `#[cfg(test)]` thread-local hook. The obvious alternative — driving the entry point from inside a runtime context, as the existing abandon test does — makes the compensating `abandon_transaction` panic as well, so it could never show the reservation returning. The hook makes only the gate acquisition fail, on the real code path.

packages/rs-platform-wallet-ffi/src/panic_guard.rs
bfoss765: Applied your suggestion in 84f469e, plus the same escaping on `FfiBoundaryError::runtime_unavailable` so the other marker cannot be lost the same way. `nul_bytes_in_a_payload_keep_the_marker_and_the_payload` panics with `"boom\0with\0nuls"` and asserts the marker is still at position 0, the payload survives escaped, the call site survives, and the message did not collapse to the `<invalid UTF-8>` fallback; `runtime_unavailable_escapes_nuls_and_keeps_its_marker` covers the second constructor.

packages/rs-platform-wallet-ffi/src/core_wallet/broadcast.rs
```diff
@@ -1,23 +1,31 @@
...
-        | Err(error @ PlatformWalletError::TransactionBroadcastUnconfirmed(_)) => {
-            (Some(local_txid), error.into())
-        }
-        Err(error) => (None, error.into()),
+        // A boundary failure (caught panic / no runtime) carries NO outcome
+        // guarantee, so it gets the same shape as any other unclassified
+        // failure: the generic code, its own message verbatim, and no txid —
+        // it must not borrow the `Some(local_txid)` treatment that says
+        // "this reached the network, reconcile by this id".
+        Err(GuardedError::Boundary(error)) => (None, error.into()),
```
thepastaclaw, line 23: **🟡 Suggestion: Return the local txid for an unknown panic outcome**

A caught panic can occur after `broadcast_finalized_transaction` has submitted the transaction but before it returns. This arm nevertheless discards the already-computed deterministic `local_txid`, after the finalized handle has been consumed, even though the panic contract tells the host to treat the result as an unknown outcome and reconcile against chain state. Supplying the txid does not claim network acceptance—the definitive rejection path already supplies it despite proving that the send failed. Preserve the txid for panic-marked boundary failures and update the Swift and JNI wrappers to expose a code-6, panic-marked result as an unknown outcome carrying that txid; JNI currently throws before reading `out_txid`, while Swift only constructs an outcome for codes 0, 20, and 26.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/core_wallet/broadcast.rs
```diff
@@ -138,11 +146,17 @@ pub unsafe extern "C" fn core_wallet_abandon_signed_transaction(
...
+    // (same contract as `core_wallet_signed_payment_release`). The two abandon
+    // calls in the error arms above stay on the swallowing `block_on`: they are
+    // best-effort cleanup on paths that already report an error of their own.
+    unwrap_result_or_return!(runtime().try_block_on(
         transaction
             .wallet
             .abandon_transaction(&transaction.transaction),
-    );
+    ));
     PlatformWalletFFIResult::ok()
```
thepastaclaw, line 160: **🟡 Suggestion: Keep the reservation recoverable when abandon panics**

The transaction is removed from `CORE_SIGNED_TRANSACTION_STORAGE` before this guarded release runs. If `try_block_on` returns a boundary failure—including the nested-runtime failure exercised by the test—the only `SignedCoreTransaction` ownership object is dropped, even though the release may never have been polled. `SignedCoreTransaction` has no `Drop` cleanup, so the inputs remain reserved until sync or the reservation TTL, and retrying with the original handle returns `NotFound`. Retain or restore a retryable ownership token on this error path, or complete compensation through the originating wallet before permanently consuming the handle. The regression test should verify that the reservation remains recoverable rather than pinning `NotFound` after a release that did not run.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/runtime.rs
```diff
@@ -17,32 +17,216 @@
...
     });
-    &RT
+
+/// Fallible acquisition of the shared runtime — the only way to reach it.
+///
+/// Cheap after the first call (a `Lazy` deref and a clone of nothing on the
+/// success path). The `Err` is cloned rather than borrowed so callers can put
+/// it in their own `Result` without borrowing from the static.
+pub(crate) fn runtime_checked() -> Result<&'static FfiRuntime, FfiBoundaryError> {
+    RT.as_ref().map_err(Clone::clone)
```
thepastaclaw, line 132: **🟡 Suggestion: Transient runtime construction failures are cached for the process lifetime**

`RT` stores `Result<FfiRuntime, FfiBoundaryError>` in a `Lazy`, so the first driver-initialization error or caught worker-launch panic permanently initializes the singleton to `Err`. Resource failures such as file-descriptor or thread exhaustion can be transient, but every later wallet operation only clones the original error and cannot recover after the host releases resources. Store only successful initialization, for example with `OnceCell<FfiRuntime>::get_or_try_init`, so an error or caught construction panic leaves the cell uninitialized and allows a later operation to retry.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/runtime.rs
```diff
@@ -17,32 +17,216 @@
...
+static RT: once_cell::sync::Lazy<Result<FfiRuntime, FfiBoundaryError>> =
+    once_cell::sync::Lazy::new(|| match guard_ffi(build_runtime) {
+        FfiOutcome::Ok(Ok(runtime)) => Ok(runtime),
+        FfiOutcome::Ok(Err(error)) => Err(FfiBoundaryError::runtime_unavailable(&format!(
+            "failed to create the tokio runtime for platform-wallet-ffi: {error}"
+        ))),
+        // Already carries the panic marker at position 0 — a panic is a panic
+        // wherever it happened, and re-labelling it would cost the host the
+        // one thing the marker is for.
+        FfiOutcome::Panicked(message) => Err(FfiBoundaryError::caught_panic(message)),
```
thepastaclaw, line 123: **🟡 Suggestion: Runtime-construction failures are mislabeled as operation panics**

A panic while Tokio constructs its worker pool occurs before any wallet future is submitted, but this arm wraps it with `caught_panic`, placing `FFI_PANIC_PREFIX` at position zero. That marker is documented as an internal wallet-operation panic with an unknown outcome, while `FFI_RUNTIME_UNAVAILABLE_PREFIX` is specifically documented for failure to obtain an execution context, including refusal of an 8 MB worker thread, where the operation provably never started. Report this branch through `runtime_unavailable` while retaining the construction-panic detail so hosts receive the correct machine-readable outcome classification.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/runtime.rs
```diff
@@ -51,13 +235,97 @@ pub(crate) fn runtime() -> &'static tokio::runtime::Runtime {
...
+            panic_payload_message(join_error.into_panic().as_ref())
+        )
+    } else {
+        // Cancellation (runtime shutdown, an explicit `abort()`). The work
+        // definitively did not finish — but that is an error to report, not a
+        // reason to take the host process down with it.
+        format!("tokio worker task did not complete: {join_error}")
+    };
+    FfiBoundaryError::caught_panic(report_panic(location, &detail))
 }
```
thepastaclaw, line 329: **🟡 Suggestion: Cancelled workers are still classified and logged as internal panics**

The code distinguishes a panicked `JoinError` from a cancelled one when composing `detail`, but both branches then pass through `report_panic` and `FfiBoundaryError::caught_panic`. A cancelled task therefore receives `FFI_PANIC_PREFIX` and emits the log message `caught panic`, contradicting the function documentation and the test's claim that cancellation is not reported as a panic. Give incomplete or cancelled work its own boundary constructor and marker, log it as incomplete work, and assert that the cancelled result does not start with `FFI_PANIC_PREFIX`.

<sub>source: ['codex']</sub>

thepastaclaw (commented):

All five prior findings are fixed at the exact head: runtime acquisition is guarded, boundary failures remain outside domain errors, lifecycle failures compensate reservations, the lower-layer error enum is unchanged, and NUL-bearing messages retain their markers. Five in-scope suggestions remain around recoverable runtime initialization, boundary classification, and preserving the information or ownership needed after caught panics; no blocking defect was verified.
Source: Codex general, security-auditor, rust-quality, and ffi-engineer reviewers used gpt-5.6-sol; final verifier backend used gpt-5.6-sol. openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

### Review provenance

🟡 5 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet-ffi/src/core_wallet/broadcast.rs:18-23: Return the local txid for an unknown panic outcome
  A caught panic can occur after `broadcast_finalized_transaction` has submitted the transaction but before it returns. This arm nevertheless discards the already-computed deterministic `local_txid`, after the finalized handle has been consumed, even though the panic contract tells the host to treat the result as an unknown outcome and reconcile against chain state. Supplying the txid does not claim network acceptance—the definitive rejection path already supplies it despite proving that the send failed. Preserve the txid for panic-marked boundary failures and update the Swift and JNI wrappers to expose a code-6, panic-marked result as an unknown outcome carrying that txid; JNI currently throws before reading `out_txid`, while Swift only constructs an outcome for codes 0, 20, and 26.
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/core_wallet/broadcast.rs:149-160: Keep the reservation recoverable when abandon panics
  The transaction is removed from `CORE_SIGNED_TRANSACTION_STORAGE` before this guarded release runs. If `try_block_on` returns a boundary failure—including the nested-runtime failure exercised by the test—the only `SignedCoreTransaction` ownership object is dropped, even though the release may never have been polled. `SignedCoreTransaction` has no `Drop` cleanup, so the inputs remain reserved until sync or the reservation TTL, and retrying with the original handle returns `NotFound`. Retain or restore a retryable ownership token on this error path, or complete compensation through the originating wallet before permanently consuming the handle. The regression test should verify that the reservation remains recoverable rather than pinning `NotFound` after a release that did not run.

In `packages/rs-platform-wallet-ffi/src/runtime.rs`:
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/runtime.rs:114-132: Transient runtime construction failures are cached for the process lifetime
  `RT` stores `Result<FfiRuntime, FfiBoundaryError>` in a `Lazy`, so the first driver-initialization error or caught worker-launch panic permanently initializes the singleton to `Err`. Resource failures such as file-descriptor or thread exhaustion can be transient, but every later wallet operation only clones the original error and cannot recover after the host releases resources. Store only successful initialization, for example with `OnceCell<FfiRuntime>::get_or_try_init`, so an error or caught construction panic leaves the cell uninitialized and allows a later operation to retry.
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/runtime.rs:120-123: Runtime-construction failures are mislabeled as operation panics
  A panic while Tokio constructs its worker pool occurs before any wallet future is submitted, but this arm wraps it with `caught_panic`, placing `FFI_PANIC_PREFIX` at position zero. That marker is documented as an internal wallet-operation panic with an unknown outcome, while `FFI_RUNTIME_UNAVAILABLE_PREFIX` is specifically documented for failure to obtain an execution context, including refusal of an 8 MB worker thread, where the operation provably never started. Report this branch through `runtime_unavailable` while retaining the construction-panic detail so hosts receive the correct machine-readable outcome classification.
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/runtime.rs:317-329: Cancelled workers are still classified and logged as internal panics
  The code distinguishes a panicked `JoinError` from a cancelled one when composing `detail`, but both branches then pass through `report_panic` and `FfiBoundaryError::caught_panic`. A cancelled task therefore receives `FFI_PANIC_PREFIX` and emits the log message `caught panic`, contradicting the function documentation and the test's claim that cancellation is not reported as a panic. Give incomplete or cancelled work its own boundary constructor and marker, log it as incomplete work, and assert that the cancelled result does not start with `FFI_PANIC_PREFIX`.
```
</details>

packages/rs-platform-wallet-ffi/src/panic_guard.rs
thepastaclaw: Resolved in `84f469e` — _Caught panics lose the documented code-and-prefix contract_ no longer present.

packages/rs-platform-wallet-ffi/src/panic_guard.rs
thepastaclaw: Resolved in `84f469e` — _NUL-containing panic payloads discard the machine-readable marker_ no longer present.

packages/rs-platform-wallet-ffi/src/core_wallet/transaction_builder.rs
thepastaclaw: Resolved in `84f469e` — _Release the reservation before returning a lifecycle-check failure_ no longer present.

packages/rs-platform-wallet/src/error.rs
thepastaclaw: Resolved in `84f469e` — _FFI panic handling introduces a source-breaking lower-layer error variant_ no longer present.

bfoss765:
Converting to issue #4437 to keep the open-PR queue focused on migration-critical work — this is preventive hardening with no observed field occurrences. The complete fix remains on `fix/ffi-panic-unwind-guard`; the five open round-4 suggestions are listed on the issue for the reopen.

PR #4425, opened 2026-08-19
fix(platform-wallet): reconcile send_payment's reservation, and carry suppressed spends across the FFI

bfoss765:
Two audit-verified wallet-core defects. Each closes an **unresolved review thread on an already-merged PR**.

---

## 1. `send_payment` discarded its reservation token (MED-HIGH)

`packages/rs-platform-wallet/src/wallet/identity/network/payments.rs`

`send_payment` built with `build_signed`, which upstream implements as a thin wrapper over `build_signed_reserved` that **discards** the `ReservationToken`. The inputs are reserved either way — the send was simply throwing away the handle it needed to reconcile them. Two failure paths were wrong as a result:

**(a) The `persister.store` failure returned `?` with no release at all.** The used-flip store must be durable *before* broadcast, so this abort happens holding a **fully signed** transaction whose inputs are all reserved. They stayed reserved until the TTL backstop, and the user's retry failed with a spurious insufficient-funds.

**(b) The rejected-broadcast release passed `reservation_token = None`**, taking `release_reservation_after_rejected_broadcast`'s unconditional by-outpoint branch — the branch its own doc warns about. That release runs after two `.await`s (build, then broadcast), so a TTL sweep can reclaim the reservation and a **newer** build re-reserve the same outpoints in between. The unconditional release then clobbers that newer owner and re-exposes the inputs of a transaction that may since have been sent (dashpay/platform#4185).

Since **#4373** pooled the funding across BIP44 + BIP32 + every DashPay receiving account, the blast radius of both is the wallet's whole spendable set rather than a single account.

**Fix:** switch to `build_signed_reserved`, thread the token out of the build block, and use it on both failure paths. `release_reservation_after_rejected_broadcast` now delegates to a general `release_build_reservation`, so the persistence abort reuses the identical per-account owner-guarded reconciliation without being misnamed.

The build-failure arm needs no release, and now says so: selection failing never reserved, and a signer failing is already released owner-guarded inside `build_signed_reserved`.

### Closes two unresolved threads on #4373 (merged)

- **CodeRabbit — 🟠 Major**, `payments.rs:1101`: "Keep reservation cleanup owner-guarded on every pre-broadcast failure. … Return the build reservation token with `tx`. Use owner-guarded cleanup for both the persistence-error and rejected-broadcast paths."
- **thepastaclaw**, `payments.rs:1353`: "Retain the pooled reservation token through pre-broadcast exits … Use `build_signed_reserved`."

---

## 2. Suppressed contact spends never crossed the FFI (MODERATE)

`packages/rs-platform-wallet/src/changeset/`, `packages/rs-platform-wallet-ffi/src/core_wallet_types.rs`

**Read the scope note below before the fix description — an earlier revision of this PR overstated what this achieves, and I've corrected it.**

**#4363** stopped a contact's watch-only record from defining the persisted transaction row, and deliberately kept `derive_spent_utxos` unfiltered so a contact spending an output a *pre-fix* build had wrongly persisted still clears the stale row.

`WalletChangeSetFFI::from_changeset` derives every account's `utxos_spent` from `cs.records` alone and states outright that it does not read `cs.spent_utxos` — a documented redundancy that **stopped holding the moment #4363 began suppressing records**. With no record left to re-derive from, a contact-only spend produced *zero* accounts and *zero* spend entries crossing the FFI. The SQLite backend, which reads `cs.spent_utxos` directly, was unaffected and heals correctly.

Reconstructing the spend downstream from `spent_utxos` is not possible: a `Utxo` carries no spending txid, and both host handlers key the `isSpent` flip on resolving the spending transaction.

**Fix:** `CoreChangeSet::unrecorded_spends` carries exactly the residue that record-suppression destroys — outpoint, spending txid, and owning account — populated only for suppressed records, and folded into the per-account `utxos_spent` arrays in the FFI builder. `spent_utxos` and its existing SQLite consumer are untouched. No FFI struct, trampoline, or Kotlin/Swift handler change: entries reuse the existing `SpentOutPointFFI` shape.

### 2b. It was derived from the wrong phases

`unrecorded_spends` was first derived from `BlockProcessed.inserted` only, reasoning that it tracked `spent_utxos`. It does not — it is the stand-in for the emit a *record* would have produced, and records re-emit from all three record-carrying phases (`inserted`, `updated` and `matured` all feed `cs.records`).

So a suppressed spend first sighted in the **mempool** got exactly one emit, at a context that can never satisfy the hosts' `context >= IN_BLOCK` gate on the `isSpent` flip, and then nothing at all when the block confirming it arrived. Now derived from all three phases. `spent_utxos` / `new_utxos` stay `inserted`-only — a confirmation re-emit creates and destroys no outpoint.

### Scope: this is Rust-side plumbing. It is INERT on Android and iOS today.

The entries now cross the FFI carrying everything a host would need. Neither host acts on them: both resolve `spending_txid` to a persisted transaction row before touching `isSpent` (`PlatformWalletPersistenceHandler.kt:973`, `PlatformWalletPersistenceHandler.swift:1413`), and for a suppressed record that row is precisely what never gets written. So for a **pure** contact → third-party spend the entry lands, the lookup misses, and the handler leaves the row alone.

The **mixed** case — the contact spends the stale coin in a transaction that also carries a surviving record of ours — does **not** depend on this field and never did. That record's FFI emit carries `input_outpoints` for *every* input of the transaction, built from `tx.input.iter()` rather than the classified spend slice (`core_wallet_types.rs:1662`), and both hosts reconcile the spend from there (`PlatformWalletPersistenceHandler.kt:852`, `PlatformWalletPersistenceHandler.swift:1120`). It is already healed on both hosts.

**The plumbing stays** because it is the half that cannot be reconstructed downstream once the record is gone, and it becomes load-bearing the moment either downstream piece lands:

1. a host-visible "this spend is final, no transaction row is coming" contract — a new field on `SpentOutPointFFI` plus the JNI signature and both handlers. That is a cross-language ABI change against pinned AAR/framework consumers, so it is not bundled into a PR scoped to these two fixes; or
2. the planned store-reconciliation pass that heals stale rows out of band.

Until one of those lands, the mobile stale-TXO heal does not run. The code says so at every site that touches the field.

### Closes the unresolved thread on #4363 (merged)

**thepastaclaw**, `core_bridge.rs:692`: "Preserve filtered contact spends through the FFI projection … Add an account-routed spent delta that the FFI conversion consumes independently of persisted transaction records, and cover the complete FFI conversion path with the stale-TXO regression fixture."

That is what this implements, fixture included — through the FFI conversion, which is the boundary the thread named. The host-side consumption is tracked separately, above.

---

## How Has This Been Tested?

Every new test was verified to **fail against the pre-fix code** and pass after, so none is vacuous.

| Test | Pre-fix failure |
|---|---|
| `send_payment_store_failure_releases_the_build_reservation` | retry aborts with `TransactionBuild("Coin selection error: No UTXOs available for selection")` — the leak, demonstrated |
| `send_payment_rejected_broadcast_release_is_owner_guarded` | the retry **succeeds**, spending an outpoint a newer build owned — the double-spend window, demonstrated |
| `contact_spend_of_a_stale_txo_crosses_the_ffi` | `accounts_count` `left: 0, right: 1` — nothing crossed the FFI |
| `unrecorded_spends_do_not_duplicate_record_derived_spends` | `left: 0, right: 1` |
| `a_suppressed_spend_re_emits_when_its_block_confirmation_arrives` | `left: 0, right: 1` — the confirming `updated` phase carried nothing |
| `phase_chaining_keeps_suppressed_and_surviving_spends_disjoint` | `left: 0, right: 1` |
| `contact_spend_still_clears_a_stale_pre_fix_txo` (extended) | new assertions on `unrecorded_spends` |
| `a_surviving_record_emits_no_unrecorded_spend` | pins the no-double-count invariant |

`send_payment_rejected_broadcast_release_is_owner_guarded` reproduces the race *inside* the broadcast await: the broadcaster releases the in-flight reservation unconditionally (what the TTL sweep does), lets a newer build re-reserve the same outpoint under a fresh token, then rejects. The assertion is that the newer reservation **survives**.

```
cargo test -p platform-wallet --features shielded   842 passed; 0 failed
cargo test -p platform-wallet-ffi                   275 passed; 0 failed
cargo test -p platform-wallet-storage               133 passed; 0 failed   (SQLite path unaffected)
```

`cargo fmt` and `cargo clippy` clean.

---

## Also out of scope

Back-filling the born-wrong `netAmount` rows written before #4363. There is no backfill pass; a row is last-writer-wins on txid and is corrected only if the funding account's record re-emits.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved payment handling when transaction building, saving, or broadcasting fails.
  - Reservations are now reliably released after failures, while protecting newer reservations from accidental cleanup.
  - Improved tracking of suppressed transaction records so spent outputs are still cleaned up correctly.
  - Prevented duplicate spend entries during transaction updates and confirmations.
- **Reliability**
  - Enhanced handling of stale contact and watch-only transaction data across wallet processing.

### DIFF
--- a/packages/rs-platform-wallet-ffi/src/core_wallet_types.rs
+++ b/packages/rs-platform-wallet-ffi/src/core_wallet_types.rs
@@ -258,11 +258,18 @@ impl WalletChangeSetFFI {
     /// dashpay/platform#4363 began *suppressing* a contact's watch-only
     /// record from `records` while still emitting its spend — the
     /// stale-TXO heal. With nothing left in `records` to re-derive from,
-    /// those spends reached no mobile host at all, so the heal was dead on
-    /// Android and iOS; only the SQLite backend, which reads `spent_utxos`
-    /// directly, ever ran it. `CoreChangeSet::unrecorded_spends` carries
-    /// exactly that residue — outpoint, spending txid and owning account —
-    /// and is folded into the per-account `utxos_spent` arrays below.
+    /// those spends reached no mobile host at all; only the SQLite backend,
+    /// which reads `spent_utxos` directly, ever ran the heal.
+    /// `CoreChangeSet::unrecorded_spends` carries exactly that residue —
+    /// outpoint, spending txid and owning account — and is folded into the
+    /// per-account `utxos_spent` arrays below.
+    ///
+    /// That closes the gap *at this boundary only*. The entries now cross
+    /// the FFI, but neither host acts on them yet: both gate the `isSpent`
+    /// flip on resolving `spending_txid` to a persisted transaction row, and
+    /// for a suppressed record no such row is ever written. Do not read the
+    /// fold below as "the heal runs on mobile" — see
+    /// `CoreChangeSet::unrecorded_spends` for what still has to land.
     ///
     /// `chain` carries `synced_height` from the changeset's
     /// `synced_height` field; `block_hash` is omitted because
@@ -920,6 +927,13 @@ fn record_spent_outpoints_ffi(
 /// The spending txid is carried through rather than zeroed: both host
 /// handlers resolve it to decide whether the spend is final, and a zero txid
 /// makes the emit a no-op on each of them.
+///
+/// A non-zero txid is not sufficient either, and today it is not: the hosts
+/// look the txid up in their own transaction table, and a suppressed record
+/// never puts it there, so these entries currently land and do nothing. The
+/// txid is carried because it is the one fact that cannot be recovered
+/// downstream once the record is gone — see
+/// `CoreChangeSet::unrecorded_spends`.
 ///
 /// [`UnrecordedSpend`]: platform_wallet::changeset::UnrecordedSpend
 fn unrecorded_spend_ffi(spend: &platform_wallet::changeset::UnrecordedSpend) -> SpentOutPointFFI {
@@ -1897,6 +1911,12 @@ mod tests {
     /// (which reads `cs.spent_utxos` directly) and no mobile host.
     ///
     /// Pre-fix this emitted zero accounts and zero spends.
+    ///
+    /// Scope: this pins the *FFI boundary*, not the heal. Both host handlers
+    /// still require a persisted row for `spending_txid` before they touch
+    /// `isSpent`, and a suppressed record produces none — so what this test
+    /// asserts is that the entry arrives with everything a host would need,
+    /// not that any host acts on it today.
     #[test]
     fn contact_spend_of_a_stale_txo_crosses_the_ffi() {
         use dashcore::hashes::Hash;
--- a/packages/rs-platform-wallet-ffi/src/core_wallet_types.rs
+++ b/packages/rs-platform-wallet-ffi/src/core_wallet_types.rs
@@ -254,6 +254,16 @@ impl WalletChangeSetFFI {
     /// (the records are authoritative), and re-deriving keeps the
     /// per-account routing self-contained.
     ///
+    /// That redundancy stopped holding for spends once
+    /// dashpay/platform#4363 began *suppressing* a contact's watch-only
+    /// record from `records` while still emitting its spend — the
+    /// stale-TXO heal. With nothing left in `records` to re-derive from,
+    /// those spends reached no mobile host at all, so the heal was dead on
+    /// Android and iOS; only the SQLite backend, which reads `spent_utxos`
+    /// directly, ever ran it. `CoreChangeSet::unrecorded_spends` carries
+    /// exactly that residue — outpoint, spending txid and owning account —
+    /// and is folded into the per-account `utxos_spent` arrays below.
+    ///
     /// `chain` carries `synced_height` from the changeset's
     /// `synced_height` field; `block_hash` is omitted because
     /// `WalletEvent::SyncHeightAdvanced` doesn't carry it (the upstream
@@ -351,6 +361,17 @@ impl WalletChangeSetFFI {
             }
         }
 
+        // Same for an account carrying only suppressed spend-clears. This is
+        // the common shape of the #4363 heal: a contact spends a stale row
+        // and their watch-only record — the batch's only record for that
+        // account — is dropped, so the bucket would not exist at all and the
+        // spend would have nowhere to be emitted.
+        for spend in &cs.unrecorded_spends {
+            if !by_account.iter().any(|(at, _)| at == &spend.account_type) {
+                by_account.push((spend.account_type, Vec::new()));
+            }
+        }
+
         let mut ffi_accounts = Vec::with_capacity(by_account.len());
         for (account_type, recs) in by_account {
             let type_name = CString::new(format!("{:?}", account_type))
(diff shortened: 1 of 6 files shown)


thepastaclaw:
🔍 Review in progress — actively reviewing now (commit 4087fe17)
_ETA: complete ~00:08 UTC (median 13m across 30 recent reviews)_
_Running 4m · Last checked: 2026-08-20 00:00 UTC_

bfoss765:
Correcting something I got wrong in this PR, and pushing the fix for it (4087fe17).

I described fix 2 as reviving the stale-TXO heal on mobile, and said it "completes the heal when the spending transaction also carries a surviving record of ours." Both halves of that are wrong, and I've now checked the host code rather than reasoning about it.

**The mixed case never needed this field.** When a surviving record of ours exists for the spending transaction, its FFI emit already carries `input_outpoints` for *every* input of that transaction — built from `tx.input.iter()`, not from the classified spend slice — and both hosts walk that list and flip `isSpent` by outpoint (`PlatformWalletPersistenceHandler.kt:852`, `PlatformWalletPersistenceHandler.swift:1120`). That case was healed before this PR. I claimed credit for it.

**The pure case is gated out on both hosts, not one.** Each resolves `spending_txid` to a persisted transaction row before touching `isSpent` (`getByTxid(spendingTxid)` on Android, the `PersistentTransaction` fetch on iOS), and a suppressed record is exactly the row that never gets written. So the entry arrives and the handler leaves the row alone.

Net: `unrecorded_spends` is correct Rust-side and crosses the FFI with everything a host would need, but it is **inert on Android and iOS today**. It is not a partial heal — it is a prerequisite for one. The mobile stale-TXO heal still does not run.

I'm keeping the plumbing, because it is the half that genuinely cannot be reconstructed downstream once the record is suppressed, and it goes live the moment either the host-side "final spend, no transaction row is coming" contract or the planned store-reconciliation pass lands. What changes is every claim about it — the field doc, the FFI builder doc, the derivation doc, the regression test's doc, and the PR description all now say plainly what it does and does not do.

While re-reading it I also found a real defect in the same code. `unrecorded_spends` was derived from `BlockProcessed.inserted` only, on the reasoning that it tracked `spent_utxos`. It doesn't — it stands in for the emit a *record* would have produced, and records re-emit from `inserted`, `updated` and `matured` alike. A suppressed spend first sighted in the mempool therefore got exactly one emit, at a context that can never satisfy the hosts' `context >= IN_BLOCK` gate, and nothing when the block confirming it arrived. Now derived from all three phases, with two tests that fail `left: 0, right: 1` against the old derivation. `spent_utxos` / `new_utxos` stay `inserted`-only, which was always right for them.

Also restored two doc blocks the first commit accidentally re-attached: `CoreChangeSet`'s had been captured by the new `UnrecordedSpend` struct, and `derive_spent_utxos`' by `derive_unrecorded_spends`.

Fix 1 (the reservation token) is unaffected by any of this.

bfoss765:
Converting to issue #4435 to keep the open-PR queue focused on migration-critical work. The complete fix remains on `fix/payment-reservation-and-txo-heal` and this PR can be reopened as-is when there is review bandwidth — ideally alongside the ABI batch that activates the suppressed-spend carry on mobile.

Issue #4365, opened 2026-08-10
fix(platform-wallet): a partial identity scan is never retried once any identity is on file


romchornyi:
## Expected Behavior

A wallet that owns more than one identity should stay fully discoverable across launches. If a gap-limit scan could not probe every index, a later launch should scan again until one scan has completed cleanly.

## Current Behavior

An identity hidden by a failed probe can stay hidden permanently, because nothing re-scans once any identity is on file.

The chain, all in `packages/rs-platform-wallet`:

1. `ScanTally::is_trustworthy` (`wallet/identity/network/discovery.rs:595`) is `identities_seen > 0 || failed_probes == 0`. A scan that saw an identity at index 0 and got no answer at index 1 returns `Ok`, and logs its own caveat: *"an identity at a failed index may be missing until the next scan"* (`discovery.rs:468`). That is correct as far as it goes — the identities it did find are persisted, and discarding them would be worse.
2. `PlatformWalletManager::start_wallet_subsystems` (`manager/startup.rs:394`) skips discovery entirely whenever any identity is already on file. Introduced in #4359 so a warm launch costs no Keychain round trip and no Platform probes.
3. So there is no next scan. The identity at the failed index stays missing for the life of the installation.

The budget-expiry path has the same shape: `within_budget` returns `None`, the code consults local state, finds the sighting that was persisted before cancellation, and records it as a warm launch.

On iOS the only automatic fallback has the identical gate — `DWCurrentUserIdentityInfo.recoverIfNeeded` scans only when there are no local identities at all ("no local identities; scanning seed from index 0"). The one path that recovers from this is the manual **Settings → Identities → Discover** button.

## Possible Solution

Persist whether the last gap-limit scan completed cleanly, and let the warm-launch shortcut apply only when it did. `ScanTally` already knows `failed_probes` at the point the verdict is made; it just has nowhere to record it.

Cheaper alternatives were checked and do not work:

- **Infer it from the local index set.** A hole below the maximum known index (identities at 0 and 2, nothing at 1) is visible via `IdentityEntry::identity_index`. The case that matters is not: found at 0, failed at 1, identity really at 1 leaves the local set `{0}`, which is byte-for-byte what a clean scan of a single-identity wallet leaves.
- **Reuse existing storage.** `sqlite/schema/blob.rs` is a codec, not a KV table. `WalletMetadataEntry` is emitted only at registration and requires `network` + `wallet_group_id`, neither of which a scan has to hand.

[...]

PR #4426, opened 2026-08-19
fix(platform-wallet): stop the startup sequence reporting integrity it did not establish

bfoss765:
Three defects on the wallet bring-up path, all of which end the same way: `start_wallet_subsystems` returns a status that promises a contact's DIP-15 addresses exist before Core SPV starts, when they do not. An address the wallet is not watching when the compact-filter scan passes its funding height produces no transaction at all — so each of these is a silent data gap, not a cosmetic mislabel.

Two come from the automated review of #4359 (findings F1 and F2); the third is the follow-through on a deferral #4368 took deliberately.

## 1. A contact pass that reached nobody was recorded as a completed sync (F1)

`sync_contact_requests` is log-and-continue per identity — right for the recurring sweep, and it collapsed two opposite endings into one return value. "Platform answered, and there is nothing new" and "Platform answered nobody" both arrived as `Ok(vec![])`. With DAPI unreachable every identity's fetch hit the `continue`, the sweep returned an empty success, `startup.rs` called `record_sync_ran`, and `status()` reported `Ready`.

`Ready` is precisely the claim that a contact pass completed, so this fed the persistence-corruption class the audit was tracking: SPV starts against an address set that is silently short.

**Fix.** The pass now reports what it reached rather than only what it found:

- `sync_contact_requests_reporting` returns a `ContactSyncReport` carrying `identities_attempted`, the identities nothing was ingested for, and the identities whose sent side alone failed.
- `sync_contact_requests` keeps its shape and raises `ContactSyncUnreachable` when there were identities to read and not one was read.
- Startup records the sync **only** on a complete pass. A degraded one leaves `dashpay_sync_ran = false`, so `status()` stays `PartialAccountsPending`.

Partial passes keep sensible semantics: what was fetched is real and stays persisted, and the failures retry themselves — a failed fetch leaves that direction's high-water cursor unadvanced, so the next sweep re-requests exactly the range it missed. A partial pass is still not *complete*, because the identities it missed have contact requests nobody looked at and account builds nobody enqueued.

The recurring sweep at `dashpay_sync.rs` already logs-and-continues on an `Err` from this call, so the new error is a strict improvement there too: a total outage used to be recorded as a successful sweep.

## 2. A partial identity scan was never retried once any identity was on file (F2 = #4365)

[...]

PR #4427, opened 2026-08-19
fix(platform-wallet): harden contact-sync startup gating and payment reservation release

bfoss765:
## Summary

Fixes three audit-verified defects on the **merged** contact-request / startup / payment surface of `rs-platform-wallet`. Each was verified at `v4.2-dev` tip `c99872b08b`.

| # | Sev | Origin | One-line |
|---|-----|--------|----------|
| F1 | HIGH | merged #4359 (reviewer **refused deferral** of exactly this) | `sync_contact_requests` reports an unreachable-Platform empty as a settled empty → SPV starts with contact addresses underived |
| F2 | HIGH | merged #4359, filed as open **#4365** (author-confirmed regression) | warm-launch shortcut skips discovery forever after an *incomplete* initial scan → strands a later identity + its contacts |
| F3 | MEDIUM-HIGH | **#4373-widened** | `send_payment` drops the `ReservationToken` → store-failure strands inputs; rejected broadcast can clobber a newer reservation |

### F1 — startup gates `Ready` on an unreachable-Platform empty (#4359)
`sync_contact_requests` caught each identity's received/sent fetch failure with `continue` and returned `Ok(vec![])` **even when Platform was unreachable for every identity**. `manager/startup.rs` then recorded `record_sync_ran()` on any `Some(Ok(_))` and `status()` returned `Ready` — the API's promise that contact addresses exist before SPV starts — so SPV scanned past contacts' funding heights with DIP-15 addresses underived (the persistence-corruption class). The reviewer refused deferral of exactly this.

**Fix:** `sync_contact_requests` now returns `ContactRequestSyncOutcome { requests, fetch_complete }`. `fetch_complete` is cleared by any per-identity fetch failure — the hard errors **and** the swallowed per-identity `continue`s. Startup records the pass only when `fetch_complete`; otherwise the wallet stays out of `Ready` (via the existing `!dashpay_sync_ran` guard) and the pass re-runs. The genuine-empty fast path is preserved.

### F2 — warm-launch shortcut strands identities after an incomplete scan (#4365)
`start_wallet_subsystems` skipped discovery entirely whenever any identity was on file, and `discover_inner` treated a scan as authoritative when `is_trustworthy()` (`identities_seen > 0 || failed_probes == 0`) despite `failed_probes > 0`, with **no scan-completion state persisted**. A warm launch that saw identity 0 but got no answer for identity 1 stranded identity 1 + its contacts permanently (only the manual Discover button recovered).

**Fix:** track a per-wallet `fully_discovered_wallets` set on `IdentityManager`, set only when a scan answers **every** probe (`failed_probes == 0`), and round-trip it through `IdentityManagerStartState`. The warm-launch shortcut is now gated on it (`may_take_warm_shortcut`). An incomplete scan **or** the budget-expiry abandonment path leaves the flag unset, so the next launch re-runs discovery (resuming past the known identities) instead of shortcutting past the unprobed index forever.

### F3 — payment reservation not released / clobbered (#4373-widened)
`send_payment` used `build_signed` (which reserves the selected UTXOs but **drops** the `ReservationToken`). Consequently: (a) a `persister.store(used_flip_changeset)` failure returned early via `?` **without releasing** — signed inputs stranded until TTL; and (b) a rejected broadcast passed `reservation_token = None` to the unconditional-release branch, which can clobber a **newer** reservation a TTL sweep + re-build created on the same outpoints, re-exposing inputs of a possibly-sent tx. #4373 widened the funding set to the whole spendable set, so both now span everything.

**Fix:** thread the token via `build_signed_reserved`, release owner-guarded on the store-failure abort, and pass `Some(token)` to the rejected-broadcast release so it uses `release_reservation_if_owner` (the pattern already at `signed_payment_registry.rs`). Extracted `release_funding_reservations` as the shared, neutrally-named release primitive; `release_reservation_after_rejected_broadcast` is now a thin alias (broadcast-side callers unchanged).

## Test evidence

`cargo test -p platform-wallet --features shielded` → **844 lib + 9 integration tests pass, 0 failures.**

New tests:
- **F1** — `contact_sync_that_could_not_reach_platform_is_not_ready` (was the bug), `contact_sync_genuine_empty_is_ready`.
- **F2** — `completion_requires_every_probe_answered_not_just_trustworthiness`, `warm_shortcut_requires_a_complete_prior_scan`, `discovery_complete_flag_round_trips_through_start_state`, `set_wallet_discovery_complete_toggles_and_persists`.
- **F3** — `send_payment_store_failure_releases_the_reservation` (retry reselects the freed input); the existing `rejected_broadcast_releases_every_pooled_funding_account` / `send_payment_rejected_broadcast_returns_the_address_to_the_pool` still pass with the token threaded. The owner-guard **non-clobber** property is pinned by key-wallet's `release_if_owner_does_not_free_a_reservation_taken_over_after_a_sweep`; this change threads `Some(token)` to invoke that primitive.

## Residual limitations (stated plainly)

- **F2 durable persistence.** The completion flag round-trips through `IdentityManagerStartState` and is emitted on the changeset, but **no current backend restores it across process restart**: the FFI persister vtable has no slot for it (fixed-ABI — needs a slot + native Swift/Kotlin handlers), and the SQLite backend does not restore wallet start-state at all yet (does not attest `WALLET_RESTORE`). So on those hosts the flag is process-lifetime only and every fresh launch defaults it to *not complete* → a warm launch **conservatively re-runs discovery** (bounded by the startup budget, resuming past known identities). This is the safe direction — it never strands — but it does not yet preserve the warm-launch scan-skip optimization across restarts on those hosts. Adding the FFI slot + SQLite restore is the follow-up; the shape mirrors the existing `pending_contact_crypto_added` durability caveat.
- **F3 owner-guard.** The non-clobber property is tested at the key-wallet primitive layer (the correct home, per #4185), not duplicated end-to-end through `send_payment`'s concurrency, which would require simulating a TTL sweep + re-build between build and broadcast.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved identity discovery tracking so incomplete scans are retried instead of incorrectly treated as finished.
  * Wallet startup now waits for complete contact-request synchronization before reporting readiness.
  * Preserved discovered identities when scans encounter temporary probe failures.
  * Improved payment failure cleanup to safely release funding reservations without affecting newer transactions.
* **Reliability**
  * Persisted discovery completion status across wallet restarts.
  * Continued processing available contact requests when individual identity fetches fail.

### DIFF
--- a/packages/rs-platform-wallet-ffi/src/dashpay.rs
+++ b/packages/rs-platform-wallet-ffi/src/dashpay.rs
@@ -254,8 +254,11 @@ pub unsafe extern "C" fn platform_wallet_sync_contact_requests(
         block_on_worker(async move { identity.dashpay().sync_contact_requests().await })
     });
     let result = unwrap_option_or_return!(option);
-    let list = unwrap_result_or_return!(result);
-    unsafe { *out_array = ContactRequestHandleArray::from_requests(list) };
+    let outcome = unwrap_result_or_return!(result);
+    // This on-demand FFI fetch surfaces only the ingested requests; the
+    // `fetch_complete` flag is consumed by the in-Rust ordered-startup gate
+    // (`manager::startup`), not by this C entry point.
+    unsafe { *out_array = ContactRequestHandleArray::from_requests(outcome.requests) };
     PlatformWalletFFIResult::ok()
 }
 
--- a/packages/rs-platform-wallet-ffi/src/persistence.rs
+++ b/packages/rs-platform-wallet-ffi/src/persistence.rs
@@ -4786,6 +4786,13 @@ fn build_wallet_start_state(
     let identity_manager = IdentityManagerStartState {
         out_of_wallet_identities: BTreeMap::new(),
         wallet_identities,
+        // The FFI persister vtable has no slot for the identity-discovery
+        // completion flag (#4365) yet, so it is not restored here — it defaults
+        // to "not fully discovered", the safe direction: a warm launch re-runs
+        // discovery rather than shortcutting past an unprobed index. See the
+        // durability caveat on
+        // `PlatformWalletChangeSet::identity_discovery_complete`.
+        fully_discovered_wallets: std::collections::BTreeSet::new(),
     };
 
     // Rehydrate tracked asset-locks (built / broadcast / IS-locked
--- a/packages/rs-platform-wallet/src/changeset/changeset.rs
+++ b/packages/rs-platform-wallet/src/changeset/changeset.rs
@@ -1636,6 +1636,25 @@ pub struct PlatformWalletChangeSet {
     /// failed). Append-only delta; apply removes matching `(owner, contact,
     /// kind)` from the persisted queue.
     pub pending_contact_crypto_cleared: Vec<PendingContactCryptoKey>,
+    /// Per-wallet identity-discovery completion flag (issue #4365), keyed by
+    /// the changeset's wallet id (the `store(wallet_id, changeset)` argument).
+    /// `Some(true)` when the wallet's most recent gap-limit identity scan
+    /// answered every probe (zero failed probes); `Some(false)` when a scan was
+    /// incomplete (a failed probe left an index unprobed). `None` means no
+    /// change in this delta. Merge policy: last-write-wins (a later `Some`
+    /// overrides an earlier one).
+    ///
+    /// It gates the ordered-startup warm-launch shortcut: only a wallet whose
+    /// last scan was complete may skip the network scan when a local identity
+    /// is already on file, so an incomplete initial scan re-runs discovery next
+    /// launch instead of shortcutting past a failed-probe index forever.
+    ///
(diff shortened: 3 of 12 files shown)


thepastaclaw:
🕓 Ready for review — 2 ahead in queue (commit 0d76ff76)
_Queue position: 3/4 · 2 reviews active_
_ETA: start ~21:34 UTC · complete ~21:49 UTC (median 14m across 30 recent reviews; 2 slots)_
_Queued 1h 15m ago · Last checked: 2026-08-19 21:10 UTC_

bfoss765:
Closing as redundant: this branch duplicated work that landed split across #4425 (send_payment reservation token) and #4426 (contact-sync startup gating + discovery-completion verdict, which additionally carries the seed-binding gate). Those two are the canonical PRs; nothing here is unique. Apologies for the reviewer noise — parallel-agent orchestration error on our side.

PR #4318, opened 2026-08-06
docs(platform-wallet): error-code registry for the FFI result space

bfoss765:
Continues #4261 — moved from a fork branch to an in-repo branch (rebased onto v4.2-dev post-#4305) so maintainers can push changes directly, per review request. Full review history on #4261.

---

## Problem

`PlatformWalletFFIResultCode` (`packages/rs-platform-wallet-ffi/src/error.rs`) is public ABI — `cbindgen` emits the integers into the generated C header, Swift mirrors them in `PlatformWalletResultCode`, and Kotlin branches on them in `fromPlatformWalletNative`. Four open PRs are currently allocating adjacent integers in that enum with no shared source of truth.

A duplicate discriminant in two branches produces **no textual merge conflict**. The second merge just silently misclassifies errors on every host. That is not hypothetical here: `29` is already claimed by two PRs, and review on #4185 had to issue a renumber directive (`ErrorReservationWalletMismatch` 29 → 30) to resolve it. Three separate PRs now carry hand-written comments trying to reserve ranges on each other's behalf, and one of those reservations is for a variant that no longer exists.

This PR adds a registry so the allocation is decided in one file instead of in four diffs.

## What's here

`packages/rs-platform-wallet-ffi/ERROR_CODE_REGISTRY.md`:

- the merged allocation table (0–26, plus the 98/99 sentinels),
- the proposed allocations with their owning PRs (27–33),
- the rule for claiming a new code — next free integer, recorded here in the same PR that adds it, mirrors updated alongside,
- the ABI-stability rule — never renumber after release; deprecate and allocate a new integer,
- the contested/pending allocations, written up below as well.

Plus a six-line pointer from the crate README's Error Handling section.

**Docs-only. Zero code risk.** No enum, no `From` mapping, no Swift/Kotlin mirror is touched, and nothing is renumbered — that stays the owning PRs' work. This PR only writes down what the branches already say.

## Codes claimed by the in-flight PRs

Reconciling in one place, so this is the table to correct:

| Code | Name | Owning PR |
| ---: | --- | --- |
| 27 | `ErrorStaleReservationToken` | #4185 (also carried by #4256) |
| 28 | `ErrorReservationTokenConsumed` | #4185 (also carried by #4256) |
| 29 | `ErrorReservationWalletMismatch` | #4185 — **collides**, see below |
| 29 | `ErrorAssetLockInsufficientFunds` | #4184 — **collides**, see below |
| 30 | *unallocated* | reserved in sibling comments only, see below |
[...]

PR #4428, opened 2026-08-19
fix(platform-wallet): harden FFI runtime vs panic-abort + #4360 preflight findings

bfoss765:
## Summary

Hardens the shielded/platform-wallet FFI surface against a crate-wide panic-abort hazard and fixes the actionable preflight findings from #4360. Four independent fixes, one PR.

Verified with `cargo test -p platform-wallet --features shielded` (**839 passed, 0 failed**) and `cargo test -p platform-wallet-ffi --features shielded` (**all passed**); `cargo clippy` on both crates is clean.

---

### 1. HIGH — FFI panic-abort (`rs-platform-wallet-ffi/src/runtime.rs`)

`block_on_worker` did `rt.spawn(future).await.expect("tokio worker panicked")`. A panic in a spawned future becomes a `JoinError`; `.expect(...)` re-raised it, and that re-panic unwound across the `extern "C"` boundary → **SIGABRT on the unwind (Android/host) build**. The JNI shim's own `catch_unwind` (`rs-unified-sdk-jni`) sits *above* this callee and cannot intercept it, violating the workspace policy at `Cargo.toml` ("a JNI library must never abort the app process"). The crate had zero `catch_unwind`. #4348's 22 new marketplace entry points surfaced it, but it affected the whole async FFI surface.

**Fix.** Both shared runtime helpers now catch the panic and return a typed result instead of re-raising:
- `block_on_worker` wraps the `block_on` body in `catch_unwind` and converts a panicked/cancelled task into the future's own output type via a new fail-closed `RecoverWorkerPanic` trait bound — `Result<_, PlatformWalletError>` → `Err(PlatformWalletError::InternalPanic(msg))`, best-effort sync outputs (`()`, counters, sync summaries) → a logged empty value. `InternalPanic` **reuses** the generic `ErrorUnknown` FFI code (per "add or reuse an internal-panic code"); the panic text rides the message.
- `run_on_big_stack_thread` maps a caught thread panic to an `io::Error` (already its return type — no bound needed); its lone caller already threads that into a `PlatformWalletFFIResult`.

On the iOS `panic = "abort"` profiles (`dev-ios`/`release-ios`) `catch_unwind` is **inert by design** — the process aborts at the panic site before any `JoinError`/`catch_unwind` is observed. This matches the in-tree note that `catch_unwind` cannot protect an abort build; the fix hardens the unwind builds **without pretending to protect iOS**, and says so in a comment. The success path is unchanged (a normally-completing future returns its value directly). The `RecoverWorkerPanic` bound is fail-closed: a future new call site with an unrecoverable output type will not compile until it opts into a recovery.

Test: a panicking future returns `Err(InternalPanic)` (gated to `cfg(panic = "unwind")`); the success path and the big-stack `io::Error` mapping are covered too.

### 2. MEDIUM-HIGH (#4360) — fee-reserve scaling (`rs-platform-wallet/src/wallet/shielded/operations.rs`)

`shield_fee_reserve_credits` returned a flat `2 × compute_minimum_shielded_fee(2)` (~325.7M credits) regardless of input count, while the executor deducts the actual metered fee from input 0 (`DeductFromInput(0)`) and meters one `SetBalanceToAddress` per input, admitting up to 16. Headroom therefore thinned as inputs grew (to ~1.84x at 16, blind to count) — the estimate-vs-actual band that risks the `InternalError`/`TxAction::Removed` app-hash-divergence class (family of the mainnet shield-halt).

**Fix.** The reserve now scales with the admitted input count:
`reserve(n) = F + n × per_input_write + F_headroom`, where `F = compute_minimum_shielded_fee(2)` and `per_input_write = SHIELDED_UNSHIELD_ADDRESS_STORAGE_BYTES × (disk+processing) credits/byte` — priced off the **same versioned constants the executor reads**, never hardcoded. The planner passes `min(funded_candidates, max_address_inputs)`, an upper bound on the inputs any shield from the account uses; preflight and execution stay consistent because both re-plan through the one `shielded_shield_plan_for_account`. At `n = 0` it equals the old flat `2×F` (backward-compatible base).

**Safety statement (per request).** This is consensus-adjacent, so it is deliberately conservative and derived only from version constants. `SHIELDED_UNSHIELD_ADDRESS_STORAGE_BYTES` (222) models a *new-address* `AddBalanceToAddress` write; a shield's `SetBalanceToAddress` reduces an *existing* balance (a value replacement), which meters **no more** than a fresh subtree write — so pricing every input at that figure is a provable upper bound. Numerically (latest constants): `F = 162,851,200`, `per_input = 6,082,800`, so `reserve(16) = 423,027,200 = 2.6×F`, which **stays within the old 4×F guard** (kept, not loosened) while restoring ~2.4x headroom against the modeled worst case at 16 inputs. **Residual flagged for reviewer attention:** the per-input term is an upper bound from an existing calibrated constant, not a re-derivation of drive's exact `SetBalanceToAddress` metering. If drive can ever charge *more* than a new-address write per input, raise the per-input byte model or `SHIELD_FEE_RESERVE_HEADROOM_FEES`. New tests cover the 16-input boundary and strict monotonicity in input count.

### 3. MEDIUM (#4360) — per-input shortfall mis-mapped to account capacity (`operations.rs::map_shield_input_fetch_error`)

A strictly per-address `AddressNotEnoughFundsError` was remapped to `PlatformShieldCapacityExceeded { available: short.balance(), required: short.required_balance() }` — account-capacity semantics — and the offending address was dropped. A host reading `available` as the account max understates capacity by up to the input-count cap (~16x).

**Fix.** Mapped instead to a **distinct** `PlatformShieldInputShortfall { address, available, required }` variant whose `Display` restores the bech32m address and frames the figures as one input's live values. It rides the **same** FFI code (`ErrorShieldedInsufficientBalance` = 41) — the host's corrective action (refresh preflight, retry) is identical, and a distinct FFI *code* was deliberately avoided so as not to collide with the in-flight code-space frontier (46 per #4318); the disambiguation lives in the variant + message. `map_spend_result` routes both variants to code 41 (an or-pattern) so the new variant does not regress to the generic `ErrorWalletOperation`.

### 4. MEDIUM (#4360) — self-contradictory `available` on a fragmented account (`platform_wallet.rs::select_inputs`)

When `usable_candidates` was empty, `available` fell back to `account_balance_credits` while `required` stayed `amount + fee_reserve`, so a *failed* shield could report `available > required` (fragmented account with a large total but no usable input).

**Fix.** `available` is now the usable-for-shield amount (`usable_balance_credits`, i.e. 0 when no candidate can retain the reserve), coherent with `required`. The full account balance remains on the preflight snapshot / `reason` for display. Test added; the one existing test that encoded the old behavior was updated to the coherent value.

---

## Deferred (noted, not in this PR)

- **Max-selection largest-first (#4360 item 4)** — a larger selection-policy design change.
- **Android preflight JNI binding** — larger surface, separate change.
- The ~56 direct `runtime().block_on(...)` FFI call sites (not routed through `block_on_worker`) remain a related, unhardened panic surface. This PR closes the two shared runtime helpers (`block_on_worker` + `run_on_big_stack_thread`), which is the vector #4348 surfaced; converting the direct-`block_on` sites is a follow-up.

### DIFF
--- a/packages/rs-platform-wallet-ffi/src/error.rs
+++ b/packages/rs-platform-wallet-ffi/src/error.rs
@@ -586,6 +586,18 @@ impl From<PlatformWalletError> for PlatformWalletFFIResult {
             PlatformWalletError::PlatformShieldCapacityExceeded { .. } => {
                 PlatformWalletFFIResultCode::ErrorShieldedInsufficientBalance
             }
+            // The per-input sibling of the account-capacity variant above: a
+            // live pre-broadcast per-input shortfall (a stale-snapshot race).
+            // It rides the SAME capacity code — the host's corrective action is
+            // identical (refresh preflight, retry) — but as its OWN wallet
+            // variant so the message names the offending address and the typed
+            // `available`/`required` are understood as that single input's live
+            // figures, not an account maximum. Minting a distinct FFI code was
+            // deliberately avoided to not collide with the in-flight code-space
+            // frontier; the disambiguation lives in the message.
+            PlatformWalletError::PlatformShieldInputShortfall { .. } => {
+                PlatformWalletFFIResultCode::ErrorShieldedInsufficientBalance
+            }
             // The core-transaction sibling of the shielded pair above: the
             // do-not-retry signal must survive the boundary as a typed code
             // so hosts can distinguish it from a definitive rejection.
@@ -708,6 +720,14 @@ impl From<PlatformWalletError> for PlatformWalletFFIResult {
             // `MessageSigningKeyUnavailable` mapped above. By the time a
             // `MessageSigningFailed` reason exists, any marker in it sits
             // mid-string and is deliberately not matched.
+            // A panic recovered at the runtime-helper boundary
+            // (`runtime::block_on_worker` / `run_on_big_stack_thread`) instead
+            // of aborting the host. It is by definition an unexpected internal
+            // failure, so it reuses the generic `ErrorUnknown` code (per the
+            // "add or reuse an internal-panic code" contract); the panic text
+            // rides the message. An explicit arm — rather than the catch-all
+            // below — so the deliberate reuse is greppable.
+            PlatformWalletError::InternalPanic(..) => PlatformWalletFFIResultCode::ErrorUnknown,
             _ => PlatformWalletFFIResultCode::ErrorUnknown,
         };
         PlatformWalletFFIResult::err(code, error.to_string())
--- a/packages/rs-platform-wallet-ffi/src/runtime.rs
+++ b/packages/rs-platform-wallet-ffi/src/runtime.rs
@@ -45,19 +45,144 @@ pub(crate) fn runtime() -> &'static tokio::runtime::Runtime {
     &RT
 }
 
+/// Convert a caught worker/thread panic into a value of the driven future's
+/// own output type, so a panic surfaces as a typed error at the FFI boundary
+/// instead of unwinding across `extern "C"` and aborting the host (workspace
+/// policy — `Cargo.toml`: "a JNI library must never abort the app process").
+///
+/// Implemented for every output type actually driven through
+/// [`block_on_worker`]: any `Result<T, E>` whose error type recovers (the
+/// overwhelmingly common shape — the panic becomes a typed `Err`), plus the
+/// handful of non-`Result`, best-effort outputs whose panic degrades to a
+/// logged, empty value. The [`block_on_worker`] bound is fail-closed: a new
+/// call site whose output does not implement this will not compile until it
+/// opts into an explicit recovery here.
+pub(crate) trait RecoverWorkerPanic {
+    fn recover_from_worker_panic(reason: String) -> Self;
+}
+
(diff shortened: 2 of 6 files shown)


thepastaclaw:
🕓 Ready for review — 3 ahead in queue (commit 7e96d12f)
_Queue position: 4/4 · 2 reviews active_
_ETA: start ~21:34 UTC · complete ~21:49 UTC (median 14m across 30 recent reviews; 2 slots)_
_Queued 1h 12m ago · Last checked: 2026-08-19 21:10 UTC_

bfoss765:
Closing to keep review surfaces clean: the panic-containment half of this PR is superseded by #4424, which covers the same runtime.rs seam plus the ~46 direct runtime().block_on sites this PR's own body notes as unhardened. The unique half — the #4360 preflight fixes (input-count-scaled fee reserve, typed per-input shortfall with the address restored, coherent available/required on fragmented accounts) — is being re-cut as a focused standalone PR so it can be reviewed without the overlapping panic diff. Apologies for the noise.

PR #4429, opened 2026-08-19
fix(platform-wallet): harden shield preflight per #4360 findings

bfoss765:
## Summary

Three `#4360` shield-preflight findings, extracted clean.

**This supersedes the preflight half of closed PR #4428.** #4428 bundled these three fixes with an FFI panic-abort fix; that panic work is carried independently by #4424, so #4428 was closed and only the preflight material is re-landed here. **This branch does not touch `runtime.rs` and contains no panic/unwind changes** — it is orthogonal to #4424 and the two can land in either order.

### 1. MEDIUM-HIGH — the input-0 fee reserve now scales with the admitted input count

The whole transition fee is deducted from input 0's post-reallocation residue (`DeductFromInput(0)`), so input 0 must retain at least the actual metered fee. That fee is the flat Orchard-bundle fee `F = compute_minimum_shielded_fee(2)` **plus one `SetBalanceToAddress` address-balance write per input**, which drive meters as storage. `compute_minimum_shielded_fee` prices none of the per-input writes, so the old flat `2 × F` reserve was blind to input count and **thinned as inputs grew** — a 16-input Max shield fell to ~1.84× the modeled fee, landing in the estimate-vs-actual band that risks the `InternalError` / `TxAction::Removed` app-hash-divergence class (the family of the mainnet shield-halt).

`shield_fee_reserve_credits(version, input_count)` now computes:

```
reserve = F + input_count × per_input_write + 1 × F
```

with `per_input_write` priced off the **same versioned per-byte storage rate the executor reads** (`storage_disk_usage_credit_per_byte + storage_processing_credit_per_byte`) — never a hardcoded credit figure — using the reviewed `SHIELDED_UNSHIELD_ADDRESS_STORAGE_BYTES` address-write model. All arithmetic is checked.

At `input_count == 0` this is exactly the pre-change flat `2 × F`, so the change is a superset of the old behavior at the base. At the protocol-max 16 inputs the reserve stays ≈2.6×F — the old `≤ 4×F` ceiling is retained as a test assertion, so an oversized reserve cannot silently understate preflight capacity and strand funds below the input-0 viability threshold after a Max shield.

The planner passes `min(funded_candidates, max_address_inputs)` — an upper bound on the inputs any shield from that account can use, so the reserve is sufficient for every requested amount while staying tight (no always-16 over-reservation for accounts with few funded addresses). Preflight and execution re-plan through the same helper, so they stay consistent.

> **Residual, flagged for review:** the per-input term is an **upper bound derived from an existing calibrated constant, not a re-derivation of drive's exact `SetBalanceToAddress` metering.** `SHIELDED_UNSHIELD_ADDRESS_STORAGE_BYTES` sizes a *new-address* `AddBalanceToAddress`; a shield's `SetBalanceToAddress` reduces an *existing* address balance (a value replacement), which should meter no more than a fresh subtree write — so pricing every input at the new-address figure is deliberately conservative. A reviewer should confirm drive never charges *more* than a new-address write per input; if it can, raise the per-input byte model or `SHIELD_FEE_RESERVE_HEADROOM_FEES`. This is called out in the doc comment on the function, not just here.

### 2. MEDIUM — a strictly per-input shortfall no longer masquerades as account capacity

`AddressNotEnoughFundsError` is strictly per-address: `balance()` / `required_balance()` describe the **one short input**, not the account. Mapping it onto the account-wide `PlatformShieldCapacityExceeded` let a host misread that single address's `available` as the **account maximum** (understating capacity by up to the versioned input-count cap), and dropped the offending address entirely.

`map_shield_input_fetch_error` now maps to a distinct `PlatformShieldInputShortfall { address, available, required }` that restores the bech32m address in the message and keeps the per-input figures typed as per-input.

**No new FFI code number.** It rides the existing `ErrorShieldedInsufficientBalance` (41) alongside the account-capacity variant — the host's corrective action is identical (refresh preflight, retry) — so it does not collide with the in-flight code-space frontier. The disambiguation lives in the message and the wallet-side variant.

### 3. MEDIUM — a fragmented account reports a coherent `available`

`select_inputs` previously fell back to `account_balance_credits` when `usable_candidates` was empty. On a fragmented account (several addresses each holding exactly the reserve, none strictly above it) that reported the **full account balance** as `available` on a *failed* shield — which can exceed `required`, i.e. the self-contradictory `available > required`. `available` is now `usable_balance_credits` (0 when no candidate can retain the reserve), which is coherent with a nonzero `required = amount + fee_reserve`. The total account balance remains on the preflight snapshot (`account_balance_credits`) and its `reason` for display.

## Tests

Five tests, all new or rewritten:

| Test | Covers |
| --- | --- |
| `reserve_scales_with_admitted_input_count` | strict monotonicity in input count 0..16; exact equality to `modeled_fee(16) + F` from versioned constants |
| `versioned_fee_keeps_input_zero_valid_and_reserve_tracks_the_fee` | `input_count == 0` reproduces the old flat `2 × F`; the `≤ 4×F` ceiling still holds at max inputs |
| `live_per_input_shortfall_maps_to_typed_input_shortfall_with_address` | per-input error maps to the distinct variant, with the bech32m address restored in the message |
| `fragmented_account_reports_coherent_available_not_exceeding_required` | `available == 0 < required` on a fragmented account holding `3 × reserve` |
| `map_spend_result_maps_per_input_shortfall_to_same_code_with_address` | FFI boundary keeps code 41 (no regression to generic `ErrorWalletOperation`) and keeps the address in the message |

## Issue

Addresses the three preflight findings from #4360.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shielded transaction balance checks by accounting for fees, input limits, and available shielding capacity.
  * Added clearer errors when a specific address lacks sufficient balance, including relevant balance details.
  * Improved error reporting for shield input retrieval failures by including network information.
  * Added safeguards for fragmented accounts and fee-reserve calculation edge cases.

### DIFF
--- a/packages/rs-platform-wallet/src/wallet/platform_wallet.rs
+++ b/packages/rs-platform-wallet/src/wallet/platform_wallet.rs
@@ -60,7 +60,11 @@ pub struct ShieldedShieldPreflight {
     /// versioned maximum address-input count.
     pub usable_balance_credits: Credits,
     /// Balance retained on input 0 for the transition fee — the versioned
-    /// [`shield_fee_reserve_credits`] value the plan was computed with.
+    /// [`shield_fee_reserve_credits`] value the plan was computed with, scaled
+    /// to the input count the plan actually ADMITS rather than to the raw
+    /// funded-address count (see `resolve_shield_fee_reserve`). Addresses the
+    /// planner excludes — the leading ones that cannot be input 0, and the
+    /// later ones below the versioned minimum input amount — never inflate it.
     pub fee_reserve_credits: Credits,
     /// Maximum claim accepted by the wallet's deterministic selector:
     /// `usable_balance_credits - fee_reserve_credits`, floored at zero.
@@ -163,6 +167,125 @@ fn checked_credit_sum<'a>(
     })
 }
 
+/// The candidate set [`plan_shield_inputs`] admits at one fee reserve.
+///
+/// This is the SINGLE definition of the admission rule. The fee-reserve
+/// resolver below counts inputs through this same function, so the reserve can
+/// never be priced off a candidate count the planner would not actually admit
+/// — the defect this helper was extracted to close.
+///
+/// `candidates` must already be sorted by Platform address: that order is what
+/// the BTreeMap and the network use to identify input 0.
+#[cfg(feature = "shielded")]
+fn admit_shield_inputs(
+    candidates: &[(PlatformAddress, Credits)],
+    fee_reserve: Credits,
+    min_input_amount: Credits,
+    max_address_inputs: usize,
+) -> Vec<(PlatformAddress, Credits)> {
+    let Some(index) = candidates
+        .iter()
+        .position(|(_, balance)| *balance > fee_reserve)
+    else {
+        return Vec::new();
+    };
+
+    // Keep the fee-bearing input 0 regardless of its post-reserve base
+    // capacity: the shield fee is added to its requested claim before
+    // structure validation. Later addresses get no fee addition, so a
+    // full balance below `min_input_amount` can never form a valid
+    // input and must not inflate preflight capacity. Finally, truncate
+    // the deterministic sequence before deriving capacity so Max can
+    // always be represented by a protocol-valid input count.
+    std::iter::once(candidates[index])
+        .chain(
+            candidates[index + 1..]
+                .iter()
+                .copied()
+                .filter(|(_, balance)| *balance >= min_input_amount),
+        )
+        .take(max_address_inputs)
+        .collect()
+}
+
+/// Resolve the mutually dependent (fee reserve, admitted input set) pair.
+///
+/// Write `A(R)` for the cardinality of [`admit_shield_inputs`] at reserve `R`,
+/// and `R(n)` for `fee_reserve_for_inputs(n)`. The two are circular: the
+/// reserve decides which address can be input 0, and the admitted input count
+/// decides how many per-input `SetBalanceToAddress` writes the reserve must
+/// price. Call a count SOUND when `A(R(n)) <= n` — the reserve prices at least
+/// as many writes as the plan built with it can contain inputs. That is the
+/// property the reserve needs for EVERY requested amount, because
+/// [`ShieldedShieldInputPlan::select_inputs`] never uses more than the admitted
+/// candidates, whatever the amount. This returns the SMALLEST sound count, i.e.
+/// the least post-fixed point of `R |-> R(A(R))`.
+///
+/// Why it is well defined, minimal, and terminating:
+///
+/// * `A` is non-increasing in `R`. Raising `R` can only move the `balance > R`
+///   search for input 0 later (weakly) in the address-sorted list, and a later
+///   input 0 leaves a shorter suffix to draw the `balance >= min_input_amount`
+///   tail from. Neither the tail filter nor the `max_address_inputs` cap
+///   depends on `R`.
+/// * `R` is non-decreasing in `n`: each priced input adds one write.
+/// * So soundness is UPWARD CLOSED in `n` — if `A(R(n)) <= n` and `n' >= n`
+///   then `A(R(n')) <= A(R(n)) <= n <= n'`. The sound counts form an up-set, so
+///   the first hit of an upward scan is the minimum.
+/// * `min(candidates.len(), max_address_inputs)` is always sound, because
+///   `admit_shield_inputs` returns a subsequence of `candidates` truncated to
+///   `max_address_inputs`. That bounds the scan at `max_address_inputs + 1`
+///   evaluations (17 at the current protocol maximum) and makes the
+///   fallthrough below total — there is no unreachable branch and no panic.
+///
+/// A single descent from that upper bound (take `n1 = A(R(n_max))`, then use
+/// `R(n1)`) would be UNSOUND, which is why this scans upward instead. Because
+/// `A` is antitone, iterating it can settle into a two-cycle rather than a
+/// fixed point: an address holding exactly `R(n_max)` is not viable at
+/// `R(n_max)` (the viability test is a strict `>`) but becomes viable at the
+/// lower `R(n1)`, pulling itself and its whole suffix back in, so
+/// `A(R(n1)) > n1` and `R(n1)` under-prices its own plan.
+/// `reserve_fixed_point_does_not_descend_into_an_unsound_cycle` pins that
+/// shape. For the same reason an EXACT fixed point need not exist; the least
+/// sound count is the correct target and over-prices by at most `n - A(R(n))`
+/// writes, which is the conservative direction.
+#[cfg(feature = "shielded")]
+fn resolve_shield_fee_reserve<F>(
+    candidates: &[(PlatformAddress, Credits)],
+    fee_reserve_for_inputs: F,
+    min_input_amount: Credits,
+    max_address_inputs: usize,
+) -> Result<(Credits, Vec<(PlatformAddress, Credits)>), PlatformWalletError>
+where
+    F: Fn(usize) -> Result<Credits, PlatformWalletError>,
+{
+    let upper_bound = candidates.len().min(max_address_inputs);
+
+    for priced_inputs in 0..upper_bound {
+        let fee_reserve = fee_reserve_for_inputs(priced_inputs)?;
+        let admitted = admit_shield_inputs(
+            candidates,
+            fee_reserve,
+            min_input_amount,
+            max_address_inputs,
+        );
+        if admitted.len() <= priced_inputs {
+            return Ok((fee_reserve, admitted));
+        }
+    }
+
+    // `upper_bound` is sound by construction (see the doc comment), so this is
+    // the scan's guaranteed terminal case rather than an error path.
+    let fee_reserve = fee_reserve_for_inputs(upper_bound)?;
+    let admitted = admit_shield_inputs(
+        candidates,
+        fee_reserve,
+        min_input_amount,
+        max_address_inputs,
+    );
+    Ok((fee_reserve, admitted))
+}
+
 /// Analyze funded Platform addresses once for both preflight and execution.
 ///
 /// The representable set is the lexicographically earliest usable prefix,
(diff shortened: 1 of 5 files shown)


thepastaclaw:
⛔ Blockers found — Opus deferred (commit b1b4ace8)

bfoss765:
Addressing the reserve-vs-planner mismatch finding: good catch, and the failure mode is real — fixed in b1b4ace801.

You're right that `reserve_input_count` counted every positive-balance address while the planner then drops the leading addresses that can't be input 0 and the later ones below `min_input_amount`. At current constants each phantom input costs 6,082,800 credits, so the 1-viable + 15-dust shape moved the reserve from 331,785,200 to 423,027,200 — 91,242,000 credits charged to a transition that can only ever hold one input. Since input-0 viability is a strict `>`, that's enough to zero out capacity and reject a shield the one-input transition could fund.

I've made the reserve and the admitted set resolve against each other inside the planner rather than at the call site. `plan_shield_inputs` now takes the versioned reserve *function*, and the admission rule is extracted into `admit_shield_inputs` — the single definition both the planner and the resolver count through, so the count can't drift from the rule that produces it.

`resolve_shield_fee_reserve` returns the least *sound* input count, where sound means `A(R(n)) ≤ n` (`A` = admitted count, `R` = reserve). That's the property the reserve needs across all amounts, because `select_inputs` never uses more than the admitted candidates. `A` is non-increasing in `R` — a larger reserve only pushes the input-0 search later, leaving a shorter suffix for the `>= min_input_amount` tail, and neither the tail filter nor the input cap depends on `R` — and `R` is non-decreasing in `n`, so soundness is upward closed and the first hit of an upward scan is the minimum. `min(candidates.len(), max_address_inputs)` is always sound because `admit_shield_inputs` returns a truncated subsequence of `candidates`, which bounds the scan at 17 evaluations and makes the terminal case total — no unreachable branch.

One amendment to the suggestion: the ≤2-pass descent from the all-candidates upper bound isn't sound, so I scanned instead. `A` is antitone, so iterating settles into a two-cycle rather than a fixed point. Counterexample, now pinned as `reserve_fixed_point_does_not_descend_into_an_unsound_cycle`: four candidates where the head holds exactly `R(4)` — not viable at `R(4)` (strict `>`) so it admits 3, sound; but at the descended `R(3)` the head becomes viable and pulls its whole suffix back in, admitting 4 on a 3-input reserve. For the same reason an exact fixed point isn't guaranteed to exist, which is why least-sound is the target; it over-prices by at most `n - A(R(n))`, the conservative direction.

The reserve can only shrink relative to the old pricing, never grow, so nothing that used to succeed gets tighter.

Tests: the prescribed regression (`one_viable_address_behind_dust_reserves_for_one_input_not_the_input_cap`) asserts the shield succeeds on a one-input reserve with a single-input transition, and runs the same candidates through the pre-fix cap-priced reserve to assert that's what was rejecting it. Plus the two-cycle test above and a soundness sweep over every reserve rung crossed with admitted/dropped tail shapes. `cargo test -p platform-wallet --features shielded` 842 passed, `cargo test -p platform-wallet-ffi --features shielded` green, fmt and clippy `-D warnings` clean.

packages/rs-platform-wallet/src/wallet/platform_wallet.rs
```diff
@@ -159,6 +167,125 @@ fn checked_credit_sum<'a>(
...
+    for priced_inputs in 0..upper_bound {
+        let fee_reserve = fee_reserve_for_inputs(priced_inputs)?;
+        let admitted = admit_shield_inputs(
+            candidates,
+            fee_reserve,
+            min_input_amount,
+            max_address_inputs,
+        );
+        if admitted.len() <= priced_inputs {
+            return Ok((fee_reserve, admitted));
```
thepastaclaw, line 273: **🔴 Blocking: Resolve the reserve against inputs actually needed by a viable plan**

The soundness check prices the entire set returned by `admit_shield_inputs`, while `select_inputs` stops as soon as the requested amount is covered. Consequently, candidates that pass the minimum-input filter but are irrelevant to a small shield can eliminate otherwise valid capacity. For example, if the first address holds `R(1) + 1` credits and the next 15 addresses each hold exactly `min_input_amount`, `R(1)` admits all 16 candidates and is rejected as unsound; at `R(2)` no address can retain the reserve, so the planner reports zero capacity. A one-credit shield using only the first address is nevertheless valid and leaves the required one-input reserve. Evaluate candidate plans capped at each priced input count, retain the viable plan with the greatest executable capacity for preflight, and select an amount-compatible plan whose actual input count does not exceed the count priced by its reserve.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/shielded/operations.rs
```diff
@@ -76,39 +77,90 @@ use tracing::{debug, info, trace, warn};
...
+/// SAFETY / RESIDUAL FLAGGED FOR REVIEW: the per-input term is an upper bound
+/// derived from an existing calibrated constant, NOT a re-derivation of drive's
+/// exact `SetBalanceToAddress` metering. It stays conservative for the
+/// protocol-max input count (16), but a reviewer should confirm drive never
+/// charges MORE than a new-address write per input; if it can, raise the
+/// per-input byte model or `SHIELD_FEE_RESERVE_HEADROOM_FEES`.
 pub fn shield_fee_reserve_credits(
     platform_version: &PlatformVersion,
+    input_count: usize,
 ) -> Result<Credits, PlatformWalletError> {
```
thepastaclaw, line 136: **🟡 Suggestion: Avoid breaking the exported reserve helper's Rust signature**

`shield_fee_reserve_credits` remains publicly re-exported from `wallet::mod` and the crate root, but this PR changes it from one argument to two. The one-argument API is present in the `v4.2.0-dev.1` tag, so downstream Rust callers can already depend on it and will stop compiling. Add a distinctly named input-aware helper and retain the existing one-argument function as a deprecated compatibility wrapper. Because the old function cannot know the eventual input count, the wrapper should conservatively use the versioned maximum input count rather than preserve the unsafe flat reserve.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/shielded/operations.rs
```diff
@@ -2997,15 +3078,83 @@ mod reserve_shield_fee_tests {
...
+            .expect("reserve at 16 inputs");
+        assert_eq!(
+            reserve_16,
+            modeled_fee_16 + shield_fee,
+            "reserve(16) must equal modeled_fee(16) + one F of headroom"
+        );
+        assert!(
+            reserve_16 > modeled_fee_16,
+            "the reserve must exceed the modeled worst-case 16-input fee"
         );
```
thepastaclaw, line 3158: **🟡 Suggestion: Validate the reserve against Drive metering rather than duplicating its formula**

This test reconstructs the wallet formula from the same constant and version fields used by the implementation, so it remains green if Drive's real shield metering changes independently. The central safety property is that the reserve covers the applied shield fee for each supported input count. `SHIELDED_UNSHIELD_ADDRESS_STORAGE_BYTES` was calibrated for a new-address `AddBalanceToAddress`, whereas Shield emits `SetBalanceToAddress` operations and subsequently applies the fee-payer balance adjustment. Add a cross-crate regression in an appropriate integration harness that executes or meters representative Shield actions through Drive for input counts from one through the versioned maximum and asserts the wallet reserve covers the resulting fee; do not introduce Drive's `server` feature into the client wallet crate.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The input-aware fee model and per-input error mapping are directionally sound, but the reserve resolver still rejects valid low-input shields when additional admitted candidates are unnecessary for the requested amount. The public helper signature should remain source-compatible, and the new reserve formula needs coverage against Drive's actual metering path.
Source: Codex reviewer backend (exact model ID was not supplied in the evidence; lanes: general, security-auditor, rust-quality, and ffi-engineer); Claude final verifier backend (exact model ID was not supplied by the runtime context). openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and is not reviewer evidence.

### Review provenance

<details>

```

- [BLOCKING] packages/rs-platform-wallet/src/wallet/platform_wallet.rs:264-273: Resolve the reserve against inputs actually needed by a viable plan
  The soundness check prices the entire set returned by `admit_shield_inputs`, while `select_inputs` stops as soon as the requested amount is covered. Consequently, candidates that pass the minimum-input filter but are irrelevant to a small shield can eliminate otherwise valid capacity. For example, if the first address holds `R(1) + 1` credits and the next 15 addresses each hold exactly `min_input_amount`, `R(1)` admits all 16 candidates and is rejected as unsound; at `R(2)` no address can retain the reserve, so the planner reports zero capacity. A one-credit shield using only the first address is nevertheless valid and leaves the required one-input reserve. Evaluate candidate plans capped at each priced input count, retain the viable plan with the greatest executable capacity for preflight, and select an amount-compatible plan whose actual input count does not exceed the count priced by its reserve.

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/operations.rs:133-136: Avoid breaking the exported reserve helper's Rust signature
  `shield_fee_reserve_credits` remains publicly re-exported from `wallet::mod` and the crate root, but this PR changes it from one argument to two. The one-argument API is present in the `v4.2.0-dev.1` tag, so downstream Rust callers can already depend on it and will stop compiling. Add a distinctly named input-aware helper and retain the existing one-argument function as a deprecated compatibility wrapper. Because the old function cannot know the eventual input count, the wrapper should conservatively use the versioned maximum input count rather than preserve the unsafe flat reserve.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/operations.rs:3138-3158: Validate the reserve against Drive metering rather than duplicating its formula
  This test reconstructs the wallet formula from the same constant and version fields used by the implementation, so it remains green if Drive's real shield metering changes independently. The central safety property is that the reserve covers the applied shield fee for each supported input count. `SHIELDED_UNSHIELD_ADDRESS_STORAGE_BYTES` was calibrated for a new-address `AddBalanceToAddress`, whereas Shield emits `SetBalanceToAddress` operations and subsequently applies the fee-payer balance adjustment. Add a cross-crate regression in an appropriate integration harness that executes or meters representative Shield actions through Drive for input counts from one through the versioned maximum and asserts the wallet reserve covers the resulting fee; do not introduce Drive's `server` feature into the client wallet crate.
```
</details>

bfoss765:
Converting to issue #4436 to keep the open-PR queue focused on migration-critical work. The fix remains on `fix/shield-preflight-hardening`; the unresolved reserve-resolver adjudication is captured on the issue. Reopen when there is bandwidth.

PR #4432, opened 2026-08-20
fix(sdk): mirror the COUNT dispatcher's outer-walk limit when verifying range-outer carrier proofs

PastaPastaPasta:

A grouped COUNT over two range clauses on distinct fields cannot be verified by the SDK once the outer range holds more than ten distinct keys. The server answers with a valid proof; the client rejects it.

The shape is any `rangeCountable` index of at least two properties queried as

```
SELECT COUNT(*) WHERE outer > X AND inner > Y GROUP BY outer   (prove = true)
```

Drive serves this with the range-outer carrier-aggregate proof and caps the outer walk at the compile-time `MAX_CARRIER_AGGREGATE_OUTER_RANGE_LIMIT` (10). When the request leaves the limit unset the dispatcher lowers it to the cap, the prover walks ten outer keys and abridges the rest of the tree, and that `Some(10)` becomes part of the path query the proof commits to.

The SDK verifier rebuilt the path query with `limit: None` for every carrier shape. With eleven or more outer keys the merk walker expected an unbounded walk, reached the abridged region, and failed with:

```
carrier aggregate-count multi-key proof failed to verify: Proof is missing data for query range
```

Ten keys or fewer verified by accident because the prover never truncated, which is why this was not caught: the cap only matters once real data exceeds it, and no prove-then-verify test for this shape existed. The only caller of the carrier count verifier is the SDK, and the SDK's tests for this shape run through a mock transport that never executes it.

### Why it matters

- **Every grouped range-over-range COUNT stops working at scale.** A trending or leaderboard query of this shape is fine in a demo with a handful of groups and breaks in production as soon as an eleventh group appears. The failure surfaces as a proof error, so it looks like a compromised node rather than a client bug.
- **The limit is proof-sensitive.** `SizedQuery::limit` is committed to by the proof. A client that reconstructs it differently from the server cannot verify honest answers, and a client that accepts limits the server refuses can be handed a proof for a narrower query than it asked for. Both sides must derive the value from the same rule.

## What was done?

The SDK COUNT verifier now derives the outer-walk limit exactly as the dispatcher does, keyed on the request's where-clause shape:

- **Range-outer (two range clauses):** unset lowers to `MAX_CARRIER_AGGREGATE_OUTER_RANGE_LIMIT`; explicit limits pass through up to the cap; anything above the cap is refused before a proof is inspected, because the server refuses those requests and no proof can belong to them.
- **In-outer (one In, one range):** unset stays `None` because the In array already bounds the walk; every explicit limit is refused, as on the server.

One new function in `count_proof_helpers.rs`, about thirty lines, replacing the unconditional `None`. SUM and AVG are untouched: their dispatchers do not apply the range-outer lowering, so their verifiers already agree with the server.

### Tests

- **rs-drive, `test_range_outer_carrier_proof_verifies_with_capped_limit_not_none`:** the first prove-then-verify round trip for this proof shape. Twelve outer keys, unset limit. Pins that the dispatcher lowers to the cap, that verifying with the cap returns exactly ten entries, and that the same bytes fail under `limit: None`.
- **dash-platform-queries:** unit tests that the translation matches the dispatcher on both carrier shapes, including the refusals, asserting on the typed `RequestError`.

### Verified on testnet

Reproduced against testnet (Drive 4.1.0, protocol 13) with contract `6TjweyPbxLNB4BhyVBcAzTxo54jDn33s5Y5j78WKM43Z` (`bucketed` document type, index `(bucket, score)` marked `rangeCountable`) holding twelve documents with distinct `bucket` values, query `COUNT(*) WHERE bucket > 0 AND score > 0 GROUP BY bucket`:

| SDK | limit | result |
|---|---|---|
| before (v4.2-dev) | unset | **fails**: "Proof is missing data for query range" |
| after | unset | verified, 10 entries |
| before / after | 5 | verified, 5 entries |
| before / after | 10 | verified, 10 entries |
| either | 11 | rejected by the server with `InvalidLimit` |

## Follow-ups deliberately not in this PR

An earlier revision of this PR also added a client-side gate at the compile-time default of `max_query_limit` for all three aggregate verifiers. Review pointed out that `max_query_limit` is operator-configurable and the server compares against the runtime value, so a fixed client-side cap would reject authentic proofs from nodes configured above the default and still admit limits nodes configured below it refuse. That gate is dropped; the existing `u16` narrowing plus server-side rejection is the honest amount of validation the client can do without an authenticated view of server configuration.

Still open, unchanged from before:

- Wire-version gating for the aggregate verify entry points (COUNT/SUM/AVG and the split variants), which take an already-decoded `DocumentQuery` with no wire-request wrapper.
- Mode-aware limit-absence enforcement for the plain `Aggregate` mode and for SUM/AVG, where a `limit: 50` request still verifies against a proof produced for `limit: None`.

## Breaking Changes

None. Requests an honest server serves verify exactly as before, plus the range-outer shape now verifies past ten keys. The newly rejected requests are ones the server already refuses with `InvalidLimit`.

## Checklist:

- [x] I have assigned this pull request to a milestone _(for repository code-owners and collaborators only)_

## Summary by CodeRabbit

- **Bug Fixes**
  - Carrier aggregate proof verification now applies the correct limit behavior based on query shape.
  - Range-outer queries use the capped default limit and reject values above the cap.
  - In-outer queries remain unlimited by default and reject explicitly provided limits.

- **Tests**
  - Added coverage confirming capped range-outer proofs verify successfully and return the expected entries.
  - Added validation that unbounded verification fails for capped proofs.

### DIFF
--- a/packages/dash-platform-queries/src/documents/aggregate_limit.rs
+++ b/packages/dash-platform-queries/src/documents/aggregate_limit.rs
@@ -23,9 +23,13 @@
 //!   [`DEFAULT_QUERY_LIMIT`], because the server applies
 //!   `limit.unwrap_or(DEFAULT_QUERY_LIMIT)` before building the
 //!   path query;
-//! - carrier walk (`RangeAggregateCarrierProof`): `0` → `None`,
-//!   because the server keeps an unset limit as an unbounded outer
-//!   walk.
+//! - carrier walk (`RangeAggregateCarrierProof`), SUM / AVG: `0` →
+//!   `None`, because those servers keep an unset limit as an
+//!   unbounded outer walk;
+//! - carrier walk, COUNT: shape-dependent — see
+//!   [`ServerCappedLimit::count_carrier_walk_limit`], whose
+//!   range-outer arm enforces the stricter compile-time cap the
+//!   COUNT dispatcher applies.
 
 use drive::config::{DEFAULT_MAX_QUERY_LIMIT, DEFAULT_QUERY_LIMIT};
 
@@ -37,15 +41,24 @@ use drive::config::{DEFAULT_MAX_QUERY_LIMIT, DEFAULT_QUERY_LIMIT};
 // a silent parity break.
 const _: () = assert!(DEFAULT_QUERY_LIMIT <= DEFAULT_MAX_QUERY_LIMIT);
 
+/// A request limit that has passed the server's aggregate
+/// prove-path cap. Constructing one via [`check_within_server_cap`]
+/// is the only way to reach the walk-shape converters, so the
+/// "cap check runs first" ordering — what makes their narrowing
+/// casts exact — is enforced by the type system rather than a
+/// `debug_assert!` that vanishes from release builds.
+#[derive(Debug)]
+pub(crate) struct ServerCappedLimit(u32);
+
 /// Reject a limit the server's aggregate prove paths would refuse
 /// with `InvalidLimit`. Run this before any proof or
-/// context-provider machinery; the walk-shape converters below
-/// assume it has already passed. `0` (unset sentinel) always
+/// context-provider machinery; the returned [`ServerCappedLimit`]
+/// carries the proof that it passed. `0` (unset sentinel) always
 /// passes — its meaning is resolved per walk shape.
 pub(crate) fn check_within_server_cap(
     limit: u32,
     surface: &str,
-) -> Result<(), drive_proof_verifier::Error> {
+) -> Result<ServerCappedLimit, drive_proof_verifier::Error> {
     if limit > u32::from(DEFAULT_MAX_QUERY_LIMIT) {
         return Err(drive_proof_verifier::Error::RequestError {
             error: format!(
@@ -56,34 +69,91 @@ pub(crate) fn check_within_server_cap(
             ),
         });
     }
-    Ok(())
+    Ok(ServerCappedLimit(limit))
 }
 
-/// Distinct-walk (`RangeDistinctProof`) limit: `0` falls back to
-/// [`DEFAULT_QUERY_LIMIT`], mirroring the server's
-/// `limit.unwrap_or(DEFAULT_QUERY_LIMIT)`. Callers must have run
-/// [`check_within_server_cap`] first, which is what makes the
-/// narrowing cast exact.
-pub(crate) fn distinct_walk_limit(limit: u32) -> u16 {
-    debug_assert!(limit <= u32::from(DEFAULT_MAX_QUERY_LIMIT));
-    if limit == 0 {
-        DEFAULT_QUERY_LIMIT
-    } else {
-        limit as u16
+impl ServerCappedLimit {
+    /// Distinct-walk (`RangeDistinctProof`) limit: `0` falls back
+    /// to [`DEFAULT_QUERY_LIMIT`], mirroring the server's
+    /// `limit.unwrap_or(DEFAULT_QUERY_LIMIT)`.
+    pub(crate) fn distinct_walk_limit(&self) -> u16 {
+        if self.0 == 0 {
+            DEFAULT_QUERY_LIMIT
+        } else {
+            self.0 as u16
+        }
     }
-}
 
-/// Carrier-walk (`RangeAggregateCarrierProof`) limit: `0` stays
-/// `None` (unbounded outer walk), mirroring the server keeping an
-/// unset request limit as `None`. Callers must have run
-/// [`check_within_server_cap`] first, which is what makes the
-/// narrowing cast exact.
-pub(crate) fn carrier_walk_limit(limit: u32) -> Option<u16> {
-    debug_assert!(limit <= u32::from(DEFAULT_MAX_QUERY_LIMIT));
-    if limit == 0 {
-        None
-    } else {
-        Some(limit as u16)
+    /// Carrier-walk (`RangeAggregateCarrierProof`) limit for SUM /
+    /// AVG: `0` stays `None` (unbounded outer walk), mirroring
+    /// those servers keeping an unset request limit as `None`.
+    /// COUNT must use [`Self::count_carrier_walk_limit`] instead —
+    /// its dispatcher applies shape-dependent rules this converter
+    /// does not know about.
+    pub(crate) fn carrier_walk_limit(&self) -> Option<u16> {
+        if self.0 == 0 {
+            None
+        } else {
+            Some(self.0 as u16)
+        }
+    }
+
+    /// COUNT carrier-walk (`RangeAggregateCarrierProof`) limit,
+    /// mirroring the two shape-dependent rules of the COUNT
+    /// dispatcher's carrier arm (`drive_document_count_query/
+    /// drive_dispatcher.rs`) exactly — both change the
+    /// proof-sensitive `SizedQuery::limit` bytes, so a mismatch
+    /// either rejects honest proofs or verifies request/proof
+    /// pairings no server produced:
+    ///
+    /// - **Range-outer carrier (G8)** — `GROUP BY` one range field
+    ///   with two range clauses on distinct fields
+    ///   (`has_outer_range`): the server lowers an unset limit to
+    ///   the compile-time cap
+    ///   [`MAX_CARRIER_AGGREGATE_OUTER_RANGE_LIMIT`] and refuses
+    ///   explicit limits above it with `InvalidLimit` before
+    ///   producing proof bytes.
+    /// - **In-outer carrier (G7)**: the `In` array already bounds
+    ///   the walk. An unset limit stays `None`; the server refuses
+    ///   every explicit limit here, so the verifier must too.
+    ///
+    /// [`MAX_CARRIER_AGGREGATE_OUTER_RANGE_LIMIT`]:
+    /// drive::query::drive_document_count_query::MAX_CARRIER_AGGREGATE_OUTER_RANGE_LIMIT
+    pub(crate) fn count_carrier_walk_limit(
+        &self,
+        has_outer_range: bool,
+    ) -> Result<Option<u16>, drive_proof_verifier::Error> {
+        use drive::query::drive_document_count_query::MAX_CARRIER_AGGREGATE_OUTER_RANGE_LIMIT;
+
+        if has_outer_range {
+            if self.0 == 0 {
+                return Ok(Some(MAX_CARRIER_AGGREGATE_OUTER_RANGE_LIMIT));
+            }
+            if self.0 > u32::from(MAX_CARRIER_AGGREGATE_OUTER_RANGE_LIMIT) {
+                return Err(drive_proof_verifier::Error::RequestError {
+                    error: format!(
+                        "limit {} exceeds the carrier-aggregate range-outer cap \
+                         {MAX_CARRIER_AGGREGATE_OUTER_RANGE_LIMIT} (COUNT); the server \
+                         refuses such requests with InvalidLimit before producing proof \
+                         bytes, so no proved response can belong to this request",
+                        self.0
+                    ),
+                });
+            }
+            Ok(Some(self.0 as u16))
+        } else {
+            if self.0 != 0 {
+                return Err(drive_proof_verifier::Error::RequestError {
+                    error: format!(
+                        "limit {} on a carrier-aggregate In-outer COUNT; the server refuses \
+                         every explicit limit here (the In array bounds the walk), so no \
+                         proved response can belong to this request",
+                        self.0
+                    ),
+                });
+            }
+            Ok(None)
+        }
     }
 }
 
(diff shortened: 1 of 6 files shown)


thepastaclaw:
✅ Final review complete — no blockers (commit 7a9991d7) · triage: normal · Phase 2 only (queue backlog)

packages/dash-platform-queries/src/documents/aggregate_limit.rs
```diff
@@ -0,0 +1,267 @@
...
+    limit: u32,
+    surface: &str,
+) -> Result<ServerCappedLimit, drive_proof_verifier::Error> {
+    if limit > u32::from(DEFAULT_MAX_QUERY_LIMIT) {
+        return Err(drive_proof_verifier::Error::RequestError {
+            error: format!(
+                "limit {limit} exceeds the server's max_query_limit {DEFAULT_MAX_QUERY_LIMIT} \
+                 on the prove path ({surface}); the server refuses such requests with \
+                 InvalidLimit before producing proof bytes, so no proved response can \
+                 belong to this request"
```
thepastaclaw, line 68: **🔴 Blocking: The default limit is not the server's actual cap**

`DEFAULT_MAX_QUERY_LIMIT` is only the default for the environment-configurable `DriveConfig::max_query_limit`; `PlatformConfig` flattens that configuration and the COUNT, SUM, and AVG proof dispatchers compare requests against `request.drive_config.max_query_limit`. For example, a server configured with `max_query_limit = 200` can legitimately produce a RangeDistinctProof for an explicit limit of 101, but this verifier rejects that authentic response before verification. Conversely, a server configured with a cap of 50 refuses a limit of 75 while this gate accepts it, so the gate does not provide the claimed server-policy binding. Make the proof-serving policy a shared compile-time ceiling on the server as well, or provide the effective policy to the verifier through authenticated input; using the configuration default as though it were every server's runtime value is neither parity-preserving nor backward-compatible.

<sub>source: ['codex']</sub>

packages/dash-platform-queries/src/documents/count_proof_helpers.rs
```diff
@@ -290,17 +271,21 @@ pub(super) fn verify_count_query(
...
+            // refuses explicit limits above it, while the In-outer
+            // (G7) shape keeps `None` and refuses every explicit
+            // limit. Mirror the server's own shape test.
+            let has_outer_range = request
+                .where_clauses
+                .iter()
+                .filter(|wc| DriveDocumentCountQuery::is_range_operator(wc.operator))
+                .count()
+                == 2;
+            let limit_u16 = capped_limit.count_carrier_walk_limit(has_outer_range)?;
```
thepastaclaw, line 288: **🟡 Suggestion: Exercise COUNT carrier shape translation through the real verifier path**

The proof-sensitive behavior depends on both deriving `has_outer_range` from the request clauses and passing the converted limit into `verify_carrier_aggregate_count_proof`. The new unit tests bypass that composition by supplying `true` or `false` directly, while the integration test only exercises the shared over-100 gate before mode detection. A transposed predicate, incorrect mode route, or wrong argument at the proof call would therefore leave all new tests passing while valid G7/G8 proofs fail. Add entry-point coverage for G8 with unset, 10, and 11 limits and G7 with unset and explicit limits; server-produced or checked-in proof fixtures should also verify that the reconstructed `SizedQuery::limit` matches end to end.

<sub>source: ['codex']</sub>

packages/dash-platform-queries/tests/aggregate_limit_parity.rs
```diff
@@ -0,0 +1,105 @@
...
+                PlatformVersion::latest(),
+                &NeverCalledProvider,
+            )
+            .expect_err("an over-cap limit on an aggregate verify path must be rejected");
+            assert!(
+                error
+                    .to_string()
+                    .contains("exceeds the server's max_query_limit 100"),
+                "unexpected error for {surface} limit {limit}: {error}"
+            );
```
thepastaclaw, line 98: **🟡 Suggestion: Assert the typed RequestError instead of only its Display text**

The regression test checks only the formatted error text, so another error variant containing the same substring would satisfy it. Destructure `drive_proof_verifier::Error::RequestError` first and then inspect its payload; this pins both the typed API result and the intended limit-gate message.

```suggestion
            let message = match error {
                drive_proof_verifier::Error::RequestError { error } => error,
                other => panic!(
                    "expected RequestError for {surface} limit {limit}, got {other:?}"
                ),
            };
            assert!(
                message.contains("exceeds the server's max_query_limit 100"),
                "unexpected RequestError for {surface} limit {limit}: {message}"
            );
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The COUNT carrier conversion now matches the server's G7/G8 path-query shapes, but the new shared gate treats the configurable default of 100 as a universal server ceiling. This rejects valid proofs from servers configured above 100 and cannot establish the claimed request/server parity for servers configured below 100, so the limit policy must be made fixed on the server or conveyed to the verifier through authenticated input. Source: Codex reviewer lanes codex-general, codex-security-auditor, and codex-rust-quality (exact backend model identifiers were not present in the supplied evidence); final verifier: Anthropic Claude Agent SDK (exact model identifier not exposed); openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and not reviewer evidence.

### Review provenance

<details>

```

In `packages/dash-platform-queries/src/documents/aggregate_limit.rs`:
- [BLOCKING] packages/dash-platform-queries/src/documents/aggregate_limit.rs:62-68: The default limit is not the server's actual cap
  `DEFAULT_MAX_QUERY_LIMIT` is only the default for the environment-configurable `DriveConfig::max_query_limit`; `PlatformConfig` flattens that configuration and the COUNT, SUM, and AVG proof dispatchers compare requests against `request.drive_config.max_query_limit`. For example, a server configured with `max_query_limit = 200` can legitimately produce a RangeDistinctProof for an explicit limit of 101, but this verifier rejects that authentic response before verification. Conversely, a server configured with a cap of 50 refuses a limit of 75 while this gate accepts it, so the gate does not provide the claimed server-policy binding. Make the proof-serving policy a shared compile-time ceiling on the server as well, or provide the effective policy to the verifier through authenticated input; using the configuration default as though it were every server's runtime value is neither parity-preserving nor backward-compatible.

In `packages/dash-platform-queries/src/documents/count_proof_helpers.rs`:
- [SUGGESTION] packages/dash-platform-queries/src/documents/count_proof_helpers.rs:282-288: Exercise COUNT carrier shape translation through the real verifier path
  The proof-sensitive behavior depends on both deriving `has_outer_range` from the request clauses and passing the converted limit into `verify_carrier_aggregate_count_proof`. The new unit tests bypass that composition by supplying `true` or `false` directly, while the integration test only exercises the shared over-100 gate before mode detection. A transposed predicate, incorrect mode route, or wrong argument at the proof call would therefore leave all new tests passing while valid G7/G8 proofs fail. Add entry-point coverage for G8 with unset, 10, and 11 limits and G7 with unset and explicit limits; server-produced or checked-in proof fixtures should also verify that the reconstructed `SizedQuery::limit` matches end to end.

In `packages/dash-platform-queries/tests/aggregate_limit_parity.rs`:
- [SUGGESTION] packages/dash-platform-queries/tests/aggregate_limit_parity.rs:93-98: Assert the typed RequestError instead of only its Display text
  The regression test checks only the formatted error text, so another error variant containing the same substring would satisfy it. Destructure `drive_proof_verifier::Error::RequestError` first and then inspect its payload; this pins both the typed API result and the intended limit-gate message.
```
</details>

packages/dash-platform-queries/src/documents/aggregate_limit.rs
PastaPastaPasta: Agreed, and this gate is gone. `max_query_limit` is operator-configurable and the server compares against the runtime value, so a fixed client-side ceiling at the compile-time default can neither reject what a lower-configured node rejects nor accept what a higher-configured node accepts. The PR now carries only the range-outer carrier COUNT fix, where the cap being mirrored (`MAX_CARRIER_AGGREGATE_OUTER_RANGE_LIMIT`) is a compile-time constant on the server side too, so parity there is exact. Rewrote the description to say what changed and why.

---
🤖 Posted autonomously by Codex on behalf of pasta.

packages/dash-platform-queries/src/documents/count_proof_helpers.rs
PastaPastaPasta: Done. The translation is now a single `carrier_walk_limit(where_clauses, limit)` that derives `has_outer_range` from the (already canonicalized) clause list itself rather than taking a boolean, and the unit tests drive it with real `WhereClause` lists for both shapes. On top of that, `rs-drive` now has `test_range_outer_carrier_proof_verifies_with_capped_limit_not_none`: a prove-then-verify round trip over twelve outer keys with the limit unset, asserting the dispatcher lowered to the cap, that verification with the cap returns exactly the first ten buckets in walk order, and that the same proof bytes fail under `limit: None`. That is the first round trip for this proof shape in the tree. The end-to-end wiring through `verify_count_query` was additionally exercised against testnet (table in the description); a fixture-backed test through the SDK entry point would need a populated Drive, which `dash-platform-queries` builds without, so that remains a follow-up.

---
🤖 Posted autonomously by Codex on behalf of pasta.

packages/dash-platform-queries/tests/aggregate_limit_parity.rs
PastaPastaPasta: That test file is gone with the cap gate. The replacement unit tests in `count_proof_helpers.rs` match on `drive_proof_verifier::Error::RequestError { error }` first and only then inspect the message.

---
🤖 Posted autonomously by Codex on behalf of pasta.

packages/dash-platform-queries/src/documents/count_proof_helpers.rs
```diff
@@ -93,6 +95,70 @@ fn limit_to_u16_or_default(limit: u32) -> Result<u16, drive_proof_verifier::Erro
...
+    let has_outer_range = where_clauses
+        .iter()
+        .filter(|wc| DriveDocumentCountQuery::is_range_operator(wc.operator))
+        .count()
+        == 2;
+    if has_outer_range {
+        if limit == 0 {
+            return Ok(Some(MAX_CARRIER_AGGREGATE_OUTER_RANGE_LIMIT));
+        }
+        if limit > u32::from(MAX_CARRIER_AGGREGATE_OUTER_RANGE_LIMIT) {
```
thepastaclaw, line 138: **🟡 Suggestion: Share carrier-limit lowering with the Drive dispatcher**

This helper duplicates the shape predicate and limit defaulting/rejection rules in `rs-drive/src/query/drive_document_count_query/drive_dispatcher.rs:510–552`. The implementations agree today, but sharing only the cap constant leaves two independently maintained versions of the proof-sensitive policy whose disagreement caused this bug. The helper tests and Drive round-trip test cover each implementation separately rather than enforcing their equivalence. Extract this calculation into a pure helper under `rs-drive::query::drive_document_count_query`, available with both `server` and `verify`, and call it from both locations. Accept `Option<u32>` and return a Drive query error, leaving the SDK's zero-to-None conversion and error mapping here. Keep the low-level proof verifier's explicit effective-limit contract unchanged.

<sub>source: `gpt-6-astra` (phase2-reviewer: architecture-layering)</sub>

thepastaclaw (commented):
## Re-review — Final validation — Phase 2 only (queue backlog)

At head 7a9991d762eb6ebd7e39b817446534c49b6df73d, the SDK's carrier-limit translation matches the server's range-outer and In-outer policies; no blocking correctness or security issues were identified. Local validation passed all 59 dash-platform-queries library tests and the new Drive capped-proof regression test. One non-blocking maintainability suggestion remains: share the proof-sensitive limit policy between the dispatcher and SDK adapter.

🟡 1 suggestion(s)

<details>
<summary>Review provenance</summary>

Source: reviewer 1: `gpt-6-astra` (agent: `phase2-reviewer`, role: `general`); reviewer 2: `gpt-6-astra` (agent: `phase2-reviewer`, role: `architecture-layering`); reviewer 3: `gpt-6-astra` (agent: `phase2-reviewer`, role: `rust-quality`); reviewer 4: `gpt-6-astra` (agent: `phase2-reviewer`, role: `security-auditor`); reviewer 5: `gpt-6-astra` (agent: `phase2-reviewer`, role: `general`); reviewer 6: `gpt-6-astra` (agent: `phase2-reviewer`, role: `architecture-layering`); reviewer 7: `gpt-6-astra` (agent: `phase2-reviewer`, role: `rust-quality`); reviewer 8: `gpt-6-astra` (agent: `phase2-reviewer`, role: `security-auditor`); final verifier: `gpt-6-astra` (agent: `astra-verifier`, role: `final-verifier`)

- Triage: `normal` by `gpt-6-astra` (effort low) — This is a contained SDK proof-verification logic fix with substantial targeted tests, but it does not modify consensus rules, cryptography, signatures, funds, peer-facing deserialization, or storage migrations.
- Phase 1 reviewers: **not run (skipped for throughput: 20 PRs queued, above the 10 limit)**
- Fresh final gate: an independent Phase-2 review ran after iterative findings were reconciled
- Fresh verifier: `gpt-6-astra` — final-verifier; agent `astra-verifier`
- Phase 2 reviewers: `gpt-6-astra` — general (completed, effort high); agent `phase2-reviewer`, `gpt-6-astra` — architecture-layering (completed, effort high); agent `phase2-reviewer`, `gpt-6-astra` — rust-quality (completed, effort high); agent `phase2-reviewer`, `gpt-6-astra` — security-auditor (completed, effort high); agent `phase2-reviewer`, `gpt-6-astra` — general (completed, effort high); agent `phase2-reviewer`, `gpt-6-astra` — architecture-layering (completed, effort high); agent `phase2-reviewer`, `gpt-6-astra` — rust-quality (completed, effort high); agent `phase2-reviewer`, `gpt-6-astra` — security-auditor (completed, effort high); agent `phase2-reviewer`

</details>

<details>

```

In `packages/dash-platform-queries/src/documents/count_proof_helpers.rs`:
- [SUGGESTION] packages/dash-platform-queries/src/documents/count_proof_helpers.rs:125-138: Share carrier-limit lowering with the Drive dispatcher
  This helper duplicates the shape predicate and limit defaulting/rejection rules in `rs-drive/src/query/drive_document_count_query/drive_dispatcher.rs:510–552`. The implementations agree today, but sharing only the cap constant leaves two independently maintained versions of the proof-sensitive policy whose disagreement caused this bug. The helper tests and Drive round-trip test cover each implementation separately rather than enforcing their equivalence. Extract this calculation into a pure helper under `rs-drive::query::drive_document_count_query`, available with both `server` and `verify`, and call it from both locations. Accept `Option<u32>` and return a Drive query error, leaving the SDK's zero-to-None conversion and error mapping here. Keep the low-level proof verifier's explicit effective-limit contract unchanged.
```
</details>

PR #4438, opened 2026-08-20
fix(platform-wallet): fold per-account records into one wallet-level row, owned roles winning collisions

### LINKED ISSUE #4387: platform-wallet: multi-account send persists per-account net_amount (+ null fee, stale pending_inputs, block-gated spend reconciliation)
# platform-wallet: multi-account send persists per-account `net_amount` (+ null `fee`, stale `pending_inputs`, INCOMING-until-confirm) — store-consistency, currently masked on Android

## Summary

When a Core (L1) transaction spends inputs owned by **more than one account** (e.g. a plain BIP44 receive + a DashPay-friendship receive, or pooled BIP44/BIP32 funding), the persisted `transactions` row does not describe the **wallet**, only one contributing account. Concretely, on a send:

- `net_amount` is stored as **one account's** `change - spent`, not the wallet total (understated by the other accounts' contribution).
- `fee` is `NULL`.
- The just-broadcast send is first persisted as **`direction = INCOMING, net_amount = +change`** (the change output read as a receive), and only corrected to Outgoing when the tx **confirms** and `txos.isSpent`/`spendingTxid` are written.
- `pending_inputs` reservation rows are **not cleaned up** after confirmation.

None of this is user-visible on Android today — it is fully masked app-side (see below) — but the stored records are wrong until confirmation, and iOS/tooling reading these fields directly will surface the errors.

## Root cause (code)

`packages/rs-platform-wallet/src/changeset/core_bridge.rs:981-996` documents the mechanism precisely: upstream `check_core_transaction` emits **one `TransactionRecord` per matched account**, so a payment touching a contact produces two records sharing one txid —

| record's account         | direction  | net_amount       |
|--------------------------|------------|------------------|
| funding (BIP44/BIP32/…)  | Outgoing   | `change - spent` |
| DashpayExternalAccount   | Incoming   | `+paid`          |

The persisted `transactions` row is **keyed by txid alone, with no per-account dimension** (`transaction_account_involvements` is only written for provider-key accounts), so last-write-wins. The watch-only exclusion (`is_contact_watch_only`, dashpay/rust-dashcore#926/#952) correctly drops the *contact* record, but the surviving *funding* record still reflects only that account's spent/change — not a wallet-total aggregation across every owned account that contributed inputs. `fee` is not computed in this projection. `pending_inputs` (Room table; SDK-side reservation set) is written at broadcast but has no confirmation-time cleanup.

## Reproduction (Dash testnet, mainnet-equivalent code path)

Wallet holds two owned UTXOs in **different accounts**: 0.2 DASH in BIP44 (plain receive) + 0.01 DASH in a DashPay-friendship receiving account (received from a contact). Send-max to an address:

- On-chain: correct — one tx spends both, delivers 0.209993, fee ~700 duffs. Recipient credited correctly.
- `transactions` row: `direction = Outgoing`, `net_amount = -20000000` (**-0.2**, missing the 0.01 input), `fee = NULL`.
- `txos`: both inputs `isSpent = 0` until the tx confirmed (~minutes), then `isSpent = 1` + `spendingTxid` linked.
- `pending_inputs`: 2 reservation rows remained **after** confirmation (isSpent already 1).

(Single-account sends do not exhibit the `net_amount` error — the funding record then *is* the wallet total.)

## User-visibility & how Android masks it (for the iOS team / to size the SDK fix)

**Zero user-facing impact on Android** — two app-side layers absorb it:

1. **`net_amount` / `fee` correction** — `SdkTxStoreWalker` recomputes the true wallet net (and fee) from the transaction's own inputs/outputs and overrides the stored value for display, persisting the correction once confirmed spent-evidence exists (`t.spendingTxid = tx.txid`). So the UI showed the correct **-0.21**, not the stored -0.2.

2. **INCOMING-until-confirm ("lingering receive") correction** — the app treats a received output as spent-for-display iff `txos.isSpent = 0 AND txos.spendingTxid IS NULL AND its outpoint ∈ pending_inputs` (excluding foreign/watch-only rows). This is **display-only** (never persisted from a reservation; persisted corrections still require confirmed spent-evidence), and it **yields automatically** to any SDK authority — the moment the SDK writes `isSpent`/`spendingTxid` (at broadcast, confirmation, or via a future API), the app path contributes 0. Stale post-confirm `pending_inputs` rows are therefore harmless (isSpent=1 wins).

iOS has neither layer today, so it will show the wrong `net_amount`/`fee` and a phantom INCOMING row for a just-broadcast multi-account send until confirmation. iOS can either mirror the two Android workarounds or wait on the SDK fix below.

## Additional finding: spend reconciliation is block-gated, not islock-gated

On Dash, InstantSend lock (~2s) is finality. The **receive** side honors this (`txos.isConfirmed = 1` on islock). The **spend** side does not: after an InstantSend-locked send, the input `txos.isSpent` stayed `0` and the durable `net_amount`/fee write-back did not fire for 30+ minutes — both wait for **block confirmation**. Evidence: an islocked send (`InstantLocked` event logged) whose 5 input UTXOs remained `isSpent = 0, isInstantLocked=1`-then-cleared, `isConfirmed = 1` well past any block interval. This is non-user-facing on Android today (the display uses the islock-independent `pending_inputs` reservation, so balance/amount are correct immediately), but the durable store state — and therefore any consumer reading it, plus the `net_amount`/fee correction — lags for no reason on a chain where the spend is already final. Spend reconciliation should trigger on InstantSend lock, consistent with the receive side.

## Proposed SDK direction (additive, preserves current invariants)

Priority ordered; each is independently useful:

1. **Aggregate at persist time**: compute `net_amount` and `fee` for the `transactions` row across **all owned accounts** that contributed inputs/outputs (wallet-total), rather than storing one account's view. This removes the need for the app-side recompute entirely.
2. **Clean up `pending_inputs`** when the spending tx confirms (or is dropped/re-orged).
3. **Optionally**, expose an additive **`pendingSpend` read** (reserved outpoints for in-flight local spends) so clients can render the correct pre-confirm spent state **without changing `txos.isSpent`'s "spent-and-confirmed" invariant** and without reading the raw reservation table. This is the clean contract that lets both Android and iOS drop their workarounds.

Note: we deliberately did **not** propose flipping `txos.isSpent` at broadcast — that would break its on-chain-truth invariant and is unsafe in the drop/re-org case. (3) is the safe equivalent.

## Not blocking

Android ships the masks above; this is store-hardening, not a user-facing defect on Android. Filed for correctness + to unblock iOS parity.

HashEngineering:
## Issue being fixed
Closes #4387.

Two stacked defects in how one transaction's per-account record slices become the single persisted `transactions` row:

1. **No fold at all (#4387):** upstream emits one record PER MATCHED ACCOUNT, each carrying only its account's net slice. The persisted row is keyed by txid alone, so whichever slice drained last defined the row — a multi-account spend persisted one account's slice as the whole wallet's net (S22 field case: −0.005 stored for a −2.61920199 spend).
2. **The fold's output union let the wrong role win:** a cross-account spend's change output appears in the funding account's slice as `Sent` (its account-local view cannot attribute the sibling account's address) and in the owning account's slice as `Change`. Seeding the union from the funding record kept `Sent` on index collision — and every UTXO projection over the folded record (`record_new_utxos_ffi`, `derive_new_utxos` filter on `Received|Change`) then silently dropped the wallet's own change while the folded net stayed correct. Observed on-device 2026-08-19: corrected record rows landed, their TXO rows never arrived, the store-side reconcile tripwire healed 4 missing TXOs at sync.

## What was done
- First commit (authored by @bfoss765, cherry-picked from the keystore integration line): `fold_same_txid_records` — per txid group, net is the sum of slices, input/output details are unioned, fee from the funding slice, direction recomputed, identity fields from the funding record.
- Second commit: on output-index collision the OWNED role wins unconditionally — ownership is account-scoped knowledge, so exactly one slice can carry `Received|Change` for a given index.

## How this was tested
platform-wallet suite: 675 passed, 0 failed — including the fold's own tests and a new regression test (`fold_prefers_owned_output_role_on_index_collision`) with the exact device shape: funding slice says Sent, owning slice says Change, folded record must say Change. Device-validated as part of the reconcile series: after this fix a CoinJoin-heavy testnet wallet's corrective records deliver their TXOs and the store-side reconcile reports zero heals across restore, rescan, and relaunch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Consolidated multiple records from the same transaction into a single wallet record.
  * Preserved transaction details while correctly combining amounts, inputs, outputs, fees, and transaction direction.
  * Ensured owned change outputs take precedence when output indexes conflict.
  * Kept records from separate transactions distinct and maintained their original ordering.


thepastaclaw:
🕓 Ready for review — 2 ahead in queue (commit 2fbc7b68)
_Queue position: 3/4_
_ETA: start ~23:07 UTC · complete ~23:25 UTC (median 17m across 30 recent reviews; 2 slots)_
_Queued 1m ago · Last checked: 2026-08-28 22:50 UTC_

packages/rs-platform-wallet/src/changeset/changeset.rs
```diff
@@ -230,14 +230,150 @@ impl HighestUsedIndexes {
...
+                drop_idx.insert(i);
+            }
+        }
+        merged.net_amount = net;
+        merged.direction = match net.cmp(&0) {
+            std::cmp::Ordering::Less => TransactionDirection::Outgoing,
+            std::cmp::Ordering::Greater => TransactionDirection::Incoming,
+            std::cmp::Ordering::Equal => records[base_pos].direction,
+        };
+        folded.insert(base_pos, merged);
```
thepastaclaw: **🔴 Blocking: Repeated lifecycle snapshots are summed as if they were account slices**

`CoreChangeSet::merge` combines independent wallet events, including a `TransactionDetected` event and a later `BlockProcessed.updated` snapshot for the same transaction. These records repeat the same account contribution rather than representing disjoint account slices, but the fold groups only by txid and sums both amounts. A -100 mempool record followed by its -100 confirmed snapshot therefore becomes -200. Since `base_pos` selects the first record with inputs, it also retains the earlier `Mempool` context instead of the newer `InBlock` context. Resolve successive observations with latest-snapshot semantics before aggregating distinct account slices, and add a detection-then-confirmation regression test.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/changeset/core_bridge.rs
```diff
@@ -691,6 +691,11 @@ async fn build_core_changeset(
                     .filter(|r| !is_contact_watch_only(r))
                     .cloned(),
             );
+            // One block can insert SEVERAL per-account records for one
+            // transaction (a multi-account spend); fold them into the one
+            // wallet-level record the txid-keyed row needs
+            // (dashpay/platform#4387 — see fold_same_txid_records).
+            crate::changeset::changeset::fold_same_txid_records(&mut cs.records);
```
thepastaclaw: **🔴 Blocking: Mempool account slices are folded only when scheduling puts them in one adapter batch**

`BlockProcessed` carries all account records together and is folded directly, but live mempool matching emits one `TransactionDetected` event per account. Those slices meet only if the adapter's opportunistic `try_recv` drain happens to place them in the same persistence batch. The adapter can store the first event before the producer sends the next, causing each fold to see a singleton and the later txid upsert to replace the earlier slice. That nondeterministically reproduces the incorrect wallet net this PR is intended to fix. Aggregate at a boundary that guarantees all records for one transaction are complete rather than using the persistence drain boundary.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/changeset/changeset.rs
```diff
@@ -230,14 +230,150 @@ impl HighestUsedIndexes {
...
+                }
+                drop_idx.insert(i);
+            }
+        }
+        merged.net_amount = net;
+        merged.direction = match net.cmp(&0) {
+            std::cmp::Ordering::Less => TransactionDirection::Outgoing,
+            std::cmp::Ordering::Greater => TransactionDirection::Incoming,
+            std::cmp::Ordering::Equal => records[base_pos].direction,
+        };
```
thepastaclaw, line 426: **🔴 Blocking: Net-sign recomputation erases Internal and CoinJoin directions**

`TransactionDirection` is not determined solely by `net_amount`. Upstream assigns `CoinJoin` from `transaction_type` and assigns `Internal` when wallet inputs produce only wallet-owned outputs. A cross-account internal transfer normally has a negative wallet net equal to its fee, so the fold relabels it `Outgoing`; even a zero-net transfer retains the funding slice's account-local direction rather than deriving wallet-level `Internal`. A multi-account CoinJoin with a nonzero net is likewise rewritten as `Incoming` or `Outgoing`. Recompute direction from the merged transaction type and input/output roles using the same semantics as upstream.

```suggestion
merged.net_amount = net;
merged.direction = if merged.transaction_type
    == key_wallet::transaction_checking::transaction_router::TransactionType::CoinJoin
{
    TransactionDirection::CoinJoin
} else {
    let has_inputs = !merged.input_details.is_empty();
    let has_sent = merged
        .output_details
        .iter()
        .any(|detail| detail.role == OutputRole::Sent);
    let has_our_outputs = merged.output_details.iter().any(|detail| {
        matches!(detail.role, OutputRole::Received | OutputRole::Change)
    });

    if !has_sent && has_inputs && has_our_outputs {
        TransactionDirection::Internal
    } else if has_inputs {
        TransactionDirection::Outgoing
    } else {
        TransactionDirection::Incoming
    }
};
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/changeset/changeset.rs
```diff
@@ -230,14 +230,150 @@ impl HighestUsedIndexes {
...
+            }
+            if i != base_pos {
+                for d in &r.input_details {
+                    if seen_inputs.insert(d.index) {
+                        merged.input_details.push(d.clone());
+                    }
+                }
+                for d in &r.output_details {
+                    if seen_outputs.insert(d.index) {
+                        merged.output_details.push(d.clone());
```
thepastaclaw, line 356: **🔴 Blocking: The fold erases output ownership required by the C/Swift persistence boundary**

The merged record keeps the funding record's `account_type` while moving sibling-account output details into it, and `OutputDetail` has no owning-account field. `WalletChangeSetFFI::from_changeset` then buckets records solely by `rec.account_type` and derives every added UTXO inside that bucket. Swift stores the enclosing account on `PersistentTxo`, and the restart path emits that account's tags before Rust inserts the UTXO into the corresponding account map. In the regression test's CoinJoin-funded/BIP44-change shape, the owned output is now retained but persisted and restored as a CoinJoin UTXO rather than a BIP44 UTXO, corrupting per-account balances and fund-selection state. Preserve each owned output's original account association through a separate per-account persistence projection while folding only the wallet-level transaction row.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/changeset/changeset.rs
thepastaclaw: **🟡 Suggestion: Keep the folded record at the first group position**

The function documents that a fold keeps the group's first position, but when the first slice has no inputs and a later slice is selected as `base_pos`, line 339 drops the first slice and line 348 inserts the result at the later funding position. Any unrelated records between those slices consequently move ahead of the folded transaction. Use `group[0]` as the output position and retain `base_pos` only as the source of funding metadata.

packages/rs-platform-wallet/src/changeset/changeset.rs
```diff
@@ -230,14 +230,150 @@ impl HighestUsedIndexes {
...
+        // each carrying only its account's `net_amount` slice, and the
+        // txid-keyed persisted row was whichever record landed last
+        // (field case: a 15-input full-balance sweep stored as −0.005
+        // instead of −2.61920199 — the S22 reconciliation). Folding at the
+        // batch seam makes the persisted row describe the WALLET whenever
+        // the per-account events drain together, which is how detection
+        // emits them. Cross-batch stragglers remain the persister's
+        // txid-uniqueness concern, unchanged.
         self.records.extend(other.records);
+        fold_same_txid_records(&mut self.records);
```
thepastaclaw: **🟡 Suggestion: Each buffered event rebuilds the complete txid index**

The adapter calls `CoreChangeSet::merge` once per buffered event, up to `ADAPTER_STORE_BATCH_LIMIT`, and each call now rebuilds a `BTreeMap` over all records accumulated so far. For N distinct record events, a single drain performs O(N² log N) comparisons and repeatedly allocates tree nodes, on the historical catch-up path whose batching exists to drain events at projection speed. Append records while constructing the batch and perform the event-aware fold once immediately before committing each wallet's completed batch.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The owned-output collision rule works for the tested block shape, but the fold still has four blocking data-integrity defects: it conflates lifecycle snapshots with account slices, depends on opportunistic batching for mempool slices, loses richer direction semantics, and erases the owning account needed for UTXO persistence. Two additional issues affect ordering and catch-up performance.
Source: reviewer backend model gpt-5.6-sol (Codex general, rust-quality, and FFI lanes); final verifier backend model gpt-5.6-sol; CodeRabbit inline evidence independently checked; openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only, not reviewer evidence.

### Review provenance

<details>

```

In `packages/rs-platform-wallet/src/changeset/changeset.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/changeset/changeset.rs:289-348: Repeated lifecycle snapshots are summed as if they were account slices
  `CoreChangeSet::merge` combines independent wallet events, including a `TransactionDetected` event and a later `BlockProcessed.updated` snapshot for the same transaction. These records repeat the same account contribution rather than representing disjoint account slices, but the fold groups only by txid and sums both amounts. A -100 mempool record followed by its -100 confirmed snapshot therefore becomes -200. Since `base_pos` selects the first record with inputs, it also retains the earlier `Mempool` context instead of the newer `InBlock` context. Resolve successive observations with latest-snapshot semantics before aggregating distinct account slices, and add a detection-then-confirmation regression test.
- [BLOCKING] packages/rs-platform-wallet/src/changeset/changeset.rs:342-347: Net-sign recomputation erases Internal and CoinJoin directions
  `TransactionDirection` is not determined solely by `net_amount`. Upstream assigns `CoinJoin` from `transaction_type` and assigns `Internal` when wallet inputs produce only wallet-owned outputs. A cross-account internal transfer normally has a negative wallet net equal to its fee, so the fold relabels it `Outgoing`; even a zero-net transfer retains the funding slice's account-local direction rather than deriving wallet-level `Internal`. A multi-account CoinJoin with a nonzero net is likewise rewritten as `Incoming` or `Outgoing`. Recompute direction from the merged transaction type and input/output roles using the same semantics as upstream.
- [BLOCKING] packages/rs-platform-wallet/src/changeset/changeset.rs:289-308: The fold erases output ownership required by the C/Swift persistence boundary
  The merged record keeps the funding record's `account_type` while moving sibling-account output details into it, and `OutputDetail` has no owning-account field. `WalletChangeSetFFI::from_changeset` then buckets records solely by `rec.account_type` and derives every added UTXO inside that bucket. Swift stores the enclosing account on `PersistentTxo`, and the restart path emits that account's tags before Rust inserts the UTXO into the corresponding account map. In the regression test's CoinJoin-funded/BIP44-change shape, the owned output is now retained but persisted and restored as a CoinJoin UTXO rather than a BIP44 UTXO, corrupting per-account balances and fund-selection state. Preserve each owned output's original account association through a separate per-account persistence projection while folding only the wallet-level transaction row.
- [SUGGESTION] packages/rs-platform-wallet/src/changeset/changeset.rs:339-348: Keep the folded record at the first group position
  The function documents that a fold keeps the group's first position, but when the first slice has no inputs and a later slice is selected as `base_pos`, line 339 drops the first slice and line 348 inserts the result at the later funding position. Any unrelated records between those slices consequently move ahead of the folded transaction. Use `group[0]` as the output position and retain `base_pos` only as the source of funding metadata.
- [SUGGESTION] packages/rs-platform-wallet/src/changeset/changeset.rs:375-376: Each buffered event rebuilds the complete txid index
  The adapter calls `CoreChangeSet::merge` once per buffered event, up to `ADAPTER_STORE_BATCH_LIMIT`, and each call now rebuilds a `BTreeMap` over all records accumulated so far. For N distinct record events, a single drain performs O(N² log N) comparisons and repeatedly allocates tree nodes, on the historical catch-up path whose batching exists to drain events at projection speed. Append records while constructing the batch and perform the event-aware fold once immediately before committing each wallet's completed batch.

- [BLOCKING] packages/rs-platform-wallet/src/changeset/core_bridge.rs:694-698: Mempool account slices are folded only when scheduling puts them in one adapter batch
  `BlockProcessed` carries all account records together and is folded directly, but live mempool matching emits one `TransactionDetected` event per account. Those slices meet only if the adapter's opportunistic `try_recv` drain happens to place them in the same persistence batch. The adapter can store the first event before the producer sends the next, causing each fold to see a singleton and the later txid upsert to replace the earlier slice. That nondeterministically reproduces the incorrect wallet net this PR is intended to fix. Aggregate at a boundary that guarantees all records for one transaction are complete rather than using the persistence drain boundary.
```
</details>

packages/rs-platform-wallet/src/changeset/changeset.rs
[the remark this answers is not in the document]
thepastaclaw: The doubling is confirmed. In addition, because both snapshots can contain inputs, `base_pos` selects the earlier mempool record and retains its stale context instead of the confirmed context. Snapshot resolution must happen before account slices are folded, or the representation must preserve event/observation provenance.

HashEngineering:
@bfoss765 — heads-up: the first commit here is your fold fix from the keystore integration branch (`f887cfa1ce`), cherry-picked with your authorship intact so #4387 gets a standalone landing path on v4.2-dev instead of waiting for the full integration merge. The second commit is a follow-up we hit on-device: on an output-index collision between account slices, the funding account's `Sent` view was winning over the owning account's `Received`/`Change`, so every UTXO projection over the folded record silently dropped the wallet's own change (records landed corrected, TXO rows never arrived — the reconcile tripwire in #4439 caught it healing 4). If you'd rather land the fold through your own PR, happy to retarget the role fix onto that instead — this shape just seemed the fastest path to closing #4387. Review welcome on both counts.

HashEngineering:
Composition data point from device testing dashpay/platform#4439's branch (which does NOT include this fold): coin balances and the TXO store converge without this PR, but multi-account transaction-history rows persist a single account slice — the same wallet's history net-sum converges only on bases that include this fold. So #4439 alone fixes funds; this PR is what makes the displayed history sum truthful. Relevant when weighing the review blockers here: the fold's absence is a user-visible history defect, not just an internal nicety.

QuantumExplorer (dismissed):

packages/rs-platform-wallet-ffi/src/core_wallet_types.rs
```diff
@@ -317,24 +317,55 @@ impl WalletChangeSetFFI {
...
+            } else {
+                &cs.account_records
+            };
+        #[allow(clippy::type_complexity)]
         let mut by_account: Vec<(
             AccountType,
             Vec<&key_wallet::managed_account::transaction_record::TransactionRecord>,
+            Vec<&key_wallet::managed_account::transaction_record::TransactionRecord>,
         )> = Vec::new();
         for rec in &cs.records {
```
QuantumExplorer, line 365: [P1] Preserve provider-account transaction involvement

Only `cs.records` populates transaction rows, so after folding, the transaction callback is emitted solely for the funding account; sibling `account_records` buckets contain UTXOs but no transaction. Both Swift and Kotlin create provider-special account involvement from the enclosing transaction callback, and restart restoration selects provider transactions through that involvement. A ProReg/ProUp transaction funded by a Standard account but also matching Provider Owner/Voting accounts is therefore recorded only as Standard and disappears from provider restoration/masternode aggregation after restart until a rescan. Preserve transaction involvement for every owned account slice—either emit the folded row into each involved account bucket or add an explicit involvement projection—while retaining the single wallet-level transaction values.

QuantumExplorer: Fixed in 2fbc7b6884. Verified the finding first: Swift's `upsertTransaction` appends the enclosing bucket's account to `involvedAccounts` and Kotlin writes `TransactionAccountInvolvementEntity(txid, account.id)` — both only from the per-bucket transaction callback, so funding-bucket-only emission did drop payload-only provider involvement.

`from_changeset` now emits each folded row into the bucket of every account that owns a slice of its txid (funder included, deduped). The row's values are identical in every bucket — the persisted row is txid-keyed and account-agnostic, so duplicate upserts converge — and only the enclosing bucket differs, which is exactly what the involvement join records. TXO deltas stay slice-bucketed.

Pinned by two tests: `txos_route_to_their_owning_accounts_bucket` now asserts both buckets carry the row with identical wallet-level values, and the new `payload_only_provider_account_still_receives_the_transaction_row` models the reported shape — a `ProviderOwnerKeys` slice with no input/output details still receives the row. Side benefit: a bucket carrying a TXO now always carries its parent transaction row.

packages/rs-platform-wallet/src/changeset/changeset.rs
shumkov: Addressed by 8b074a47: the fold now recomputes direction from the merged transaction type and detail roles (changeset.rs:400-419), matching upstream key-wallet rule (managed_core_funds_account.rs:521-533). Red-verified: reverting just that block to net-sign fails cross_account_transfer_folds_to_internal_direction and coinjoin_fold_keeps_coinjoin_direction.

packages/rs-platform-wallet/src/changeset/changeset.rs
shumkov: Addressed by 8b074a47 (+ QE 2fbc7b68 on top): per-account ownership rides the new CoreChangeSet.account_records slice vector (changeset.rs:105-123, coalesced on txid+account_type), which WalletChangeSetFFI::from_changeset uses as the UTXO source while transaction rows come from the folded records. Red-verified: pinning the FFI UTXO source back to cs.records fails txos_route_to_their_owning_accounts_bucket.

PR #4397, opened 2026-08-13
feat(platform)!: identity public key references (refersTo identityPublicKey)

QuantumExplorer:
## Issue being fixed or feature implemented

Follows [#2993](https://github.com/dashpay/platform/pull/2993) (refersTo for identities, contracts, tokens) and [#4390](https://github.com/dashpay/platform/pull/4390) (permanent documents). This adds the next reference target that structurally cannot dangle: a specific public key of an identity — identity keys can be disabled but never removed.

## What was done?

Extends `refersTo` with an **`identityPublicKey`** target. The reference property's value holds the identity id; the declaration names a sibling property of the same document type whose value carries the key id:

```json
"toUserId": {
  "type": "array",
  "byteArray": true,
  "minItems": 32,
  "maxItems": 32,
  "contentMediaType": "application/x.dash.dpp.identifier",
  "position": 0,
  "refersTo": {
    "type": "identityPublicKey",
    "keyIdProperty": "toKeyIndex"
  }
},
"toKeyIndex": {
  "type": "integer",
  "position": 1
}
```

This two-property shape matches how DashPay's `contactRequest` already pairs `toUserId` with `recipientKeyIndex`, so per-document key selection works naturally.

### Semantics

- **Write time** (document create, and replace on changed fields via the existing machinery): the key id is read from the declared sibling property, and the identity's key is fetched with a specific-key request (billed as one key-by-id lookup, the same way signature validation fetches keys). The key must exist and **must not be disabled** — writing a fresh reference to a dead key is almost certainly a mistake. A key disabled *later* does not invalidate previously written documents; since keys are never removed, an accepted reference can never dangle.
- **Contract registration** (the create/update declaration validation introduced in #4390): the declared `keyIdProperty` must exist in the same document type and be an integer. The u32 key-id range is enforced at write time by the integer read.

### Changes

- `DocumentPropertyReferenceTarget` gains `IdentityPublicKey { key_id_property }` (appended — the enum is consensus-serialized inside errors and `@append_only`).
- **Meta-schema v3** (still editable until the PV14 release ships): `refersTo.type` admits `identityPublicKey`; `keyIdProperty` (dotted property-path shape, ≤256 chars) is required for it and rejected for every other target. The per-target conditionals are now an `allOf` of `if/then/else` blocks.
- **Parse** (`apply_property_reference_v0`): new arm; missing/oversized `keyIdProperty` is a hard parse error.
[...]

PR #4390, opened 2026-08-13
feat(platform)!: permanent document references (refersTo permanentDocument)

QuantumExplorer:
## Issue being fixed or feature implemented

[PR #2993](https://github.com/dashpay/platform/pull/2993) introduced `refersTo` reference validation for identifier properties, limited to targets that structurally cannot disappear: identities, contracts and tokens. Documents could not be referenced, because an ordinary document can be deleted, which would leave validated references dangling.

## What was done?

Extends `refersTo` with a **`permanentDocument`** target: a reference to a document of a document type whose documents can never be deleted (`canBeDeleted: false`). The name is deliberate — such documents are *permanent* (they can still be replaced if their type is mutable; only deletion is barred, so "immutable" would be wrong, and every stored document is "persistent" in the storage sense).

The declaration names the contract and document type statically in the schema; the property value carries the referenced document's id:

```json
"parentNoteId": {
  "type": "array",
  "byteArray": true,
  "minItems": 32,
  "maxItems": 32,
  "contentMediaType": "application/x.dash.dpp.identifier",
  "position": 0,
  "refersTo": {
    "type": "permanentDocument",
    "contractId": "4Bqs6itzfoDXzmgQibYZQABbqYsXmawVf7SKe3mKDQVd",
    "documentType": "note"
  }
}
```

`contractId` is **optional**: when absent the reference targets a document type of the declaring contract itself. This is also what makes self references usable on contract creation, where the final contract id is derived from the owner and nonce and cannot be named in the schema. `documentType` is always required.

### Soundness

A reference validated at write time can never dangle, because all three legs are frozen:
- the referenced **document** cannot be deleted (`canBeDeleted: false` on its type, enforced at write time);
- the **document type** cannot be removed by a contract update (`"document type can't be removed"` in `DataContract::validate_update`), and `canBeDeleted` itself is immutable in both directions (`DocumentTypeRef::validate_config`);
- the **contract** cannot be deleted.

### Changes

- `DocumentPropertyReferenceTarget` gains a `PermanentDocument { contract_id, document_type_name }` variant, **appended** — the enum is consensus-serialized inside `ReferencedEntityNotFoundError` — and the enum is now marked `@append_only` for the CI structure gate.
[...]

PR #4388, opened 2026-08-13
refactor(sdk)!: extract transport-free query core into dash-platform-queries

PastaPastaPasta:
## Issue being fixed or feature implemented

Dash Core's Platform GUI integration (PastaPastaPasta/dash#67, tracked in dashpay/dash#7512) needs the SDK's query-building, wire-encoding, and proof-related core without the networking stack. Per maintainer guidance to refactor rather than duplicate ("make an SDK with a feature set so small it matches what you want / split up rs-sdk"), this extracts that core instead of letting embedders reimplement it.

Second slice of the `feat/transport-free-embedder-core` series (#4335), following #4344 and #4345.

## What was done?

- New `packages/dash-platform-queries` crate: `DocumentQuery` building and wire encoding for both request versions, aggregate proof helpers (count/sum/average/ranked), DPNS username helpers, and structural state-transition validation — with **no networking transport** in its dependency graph (see the dependency-graph note below for what that does and does not mean).
- `dash-sdk` depends on the new crate and re-exports every moved item at its old path. `wasm-sdk`, `rs-sdk-ffi`, and `platform-wallet` compile unchanged; three rs-sdk surfaces are source-incompatible for *downstream* crates, enumerated under Breaking Changes.
- CI: transport-free feature-cut checks now inspect `dash-platform-queries` alongside `drive-proof-verifier`; `check-features` knows about it.
- Docs: crate README + rs-sdk README pointer describing the embedder consumption path.

Wire-request decoding, request-driven proof verification, and pure DPNS/DashPay document builders follow in the next slice (#4389).

## Dependency graph — what "transport-free" means here

An earlier revision of this description claimed "no tokio, no tonic channel/TLS anywhere in its dependency graph". The tokio half of that was wrong and is corrected here.

What is **absent** from `dash-platform-queries`' normal native graph, and asserted in CI:

- `rs-dapi-client`, `hyper`, `rustls`, `tower` — i.e. no networking transport, and no tonic *transport* feature (its absence is exactly what the missing `hyper`/`tower` prove).

What is **present**, transitively:

- `tokio` (+ `tokio-util`, `mio`) via `dash-context-provider` → `dash-async`.
- `tonic` via `dapi-grpc`, for the generated message/client *types* only.

[...]

PR #4450, opened 2026-08-21
feat: expose protocol v14 ranked queries and document references to JavaScript

### LINKED ISSUE #4402: Protocol v14 features (ranked aggregate indexes, document `refersTo`) have no JS client surface in wasm-sdk / evo-sdk
## Expected Behavior

The protocol v14 features on `v4.2-dev` should be usable from the recommended JavaScript SDK:

- **Ranked aggregate indexes / provable top-K queries** (#4266) — queryable from evo-sdk.
- **Document references, `refersTo`** (#2993, #4390, #4397) — `permanentDocument` and `identityPublicKey` targets supported when building and validating documents.

## Current Behavior

Neither has a JS surface. Both are implemented in Drive/DPP and reachable over gRPC, but the client path stops before JavaScript.

`packages/js-evo-sdk` has exactly one change between `v4.1.0` and `v4.2-dev`: the `package.json` version bump from the release commit (#4379). No source file changed.

Since evo-sdk is a thin wrapper over `@dashevo/wasm-sdk`, the gap starts a layer down: the only ranked-index-related code in wasm-sdk is a `GroveElementType` declaration plus `sum` projection in `src/queries/system.rs` (from #4266). That is low-level GroveDB element typing, not a top-K query API. `refersTo` has no wasm-sdk plumbing at all.

Client-side v14 work landed in `packages/rs-sdk`, so these features are currently Rust-only for clients.

## Possible Solution

1. wasm-sdk: expose ranked/top-K document queries over the existing `RankedEntry` / `RankedEntries` messages in `platform.proto`.
2. wasm-sdk: surface `refersTo` reference metadata and its consensus errors.
3. evo-sdk: wrap both in the `documents` and `contracts` facades.

## Alternatives Considered

N/A

## Additional Context

Filed to track the gap, not to report a defect: `v4.2-dev` is active and these commits landed recently, so the binding work may just be lagging. Close if already planned.

Compared `v4.1.0` .. `v4.2-dev` at `eaf3a4890b`; `v4.2-dev` is a moving target.

QuantumExplorer:

Closes #4402.

Protocol v14 shipped two client-visible features that stopped at the Rust SDK:

- **Ranked aggregate indexes / provable top-K queries** (#4266, #4384, #4393)
- **Document references, `refersTo`** (#2993, #4390, #4397)

Neither had a JavaScript surface. `js-evo-sdk` had exactly one change between `v4.1.0` and `v4.2-dev` — a version bump — because the gap starts a layer down, in `wasm-sdk` / `wasm-dpp2`.

### Ranked and having-range queries — `wasm-sdk`, `js-evo-sdk`

`getDocumentsRanked` / `getDocumentsHaving` plus their `WithProofInfo` twins, wrapped as `documents.ranked()` / `documents.having()` in evo-sdk. Both modes are bound together because they ride the same wire path and share a result type.

**The grammar is not reimplemented.** `detect_ranked_mode` / `detect_having_mode` are `pub` under rs-drive's `verify` feature — which wasm-sdk already enables — and are pure and contract-free. The binding calls them directly, so a malformed query fails locally with rs-drive's own message and cannot drift from what the server's query table and the proof verifier enforce. (rs-sdk's `assert_ranked_shape` was deliberately *not* widened: it phrases errors in terms of Rust builder methods, which is the wrong vocabulary for a JS caller.)

Design decisions worth a reviewer's attention:

- **Dedicated `DocumentsRankedQuery` / `DocumentsHavingQuery`** rather than adding `offset` to `DocumentsQuery`, which feeds four entry points that reject an offset. The dedicated interfaces also make `limit` and `groupBy` non-optional, which ranked requires.
- **`direction: 'asc' | 'desc'` replaces `orderBy`.** This keeps the `$count` sentinel out of the public surface entirely, and structurally prevents the ordering trap documented at `document_query.rs:330` — the parser owns the `with_select`-before-`order_by_selected_aggregate` sequence, so a caller cannot invert it.
- **Results are objects, not `Map`s.** `startingRank` has nowhere to live in a `Map`, and without it `{ limit: 1, offset: 4 }` — "the 5th best" — has no meaning. Entry order is the answer, which a `Map` only conveys implicitly.
- **Entries carry both `groupKeyHex` and a decoded `groupValue`.** Hex correlates with the count/sum/average maps for the same grouping; the decode (via `DocumentType::deserialize_value_for_key`) is what makes a ranking actually name its groups, since index-key decoding is type-directed and not reimplementable in JS. Decoding is best effort — an undecodable key yields `undefined` and never fails the query; an empty key is `null`, the write path's marker for an absent optional value.
- **Averages return exact fixed point plus the scale**, rather than pre-divided. Pre-dividing destroys the integer the proof commits to, and the scale is a build-time constant that has already moved by four orders of magnitude, so it is returned rather than documented.
- **`deny_unknown_fields` on the query inputs**, a deliberate departure from `DocumentsQueryInput`. The expected mistake is pasting a `DocumentsQuery` into a ranked call and dragging `orderBy` along; permissive serde would drop it and still run the query under the default direction, answering a different question silently.

Contract-level checks (does the index declare `rankedCountable` / `rankedSummable` / `rankedAverageable`, do the pins cover a compound index's leading properties) are left to the network on purpose: wasm-sdk serves contracts from a cache, so a stale entry would reject queries the network would happily serve. Shape validation reads no contract and has no such hazard.

### Prefix `IN` pins — follow-up to #4401

#4401 landed after this branch was written and widened the grammar the binding delegates to: a compound index's leading property may now be pinned with a bounded `IN` rather than `==`, entries merged from several branches carry the branch they came from, and a non-zero `OFFSET` is rejected alongside a branching `IN`.

The Rust side needed nothing. Because the binding hands `where` clauses to `detect_ranked_mode` instead of re-implementing the operator grammar, an `IN` pin already flowed through end to end the moment #4401 merged — which is the payoff the delegation was chosen for. What was stale was the JS-facing layer wrapped around it:

- **`DocumentsIndexPin` structurally forbade `in`**, so a shipped capability was unreachable from TypeScript without a cast, and its doc comment explained `in` as something that *would* need one secondary walk per element — now a description of what the server does rather than why it refuses.
- **`RankedEntry.in_key` was dropped on the floor.** A merged page can carry one group key twice, once per pinned prefix, so without the discriminator those two rows are indistinguishable. It surfaces as `branchKeyHex`, set only on a merged page — *absent*, not `undefined`-valued, on a single-prefix page, so `'branchKeyHex' in entry` reads as "this page is a merge".
- **Nothing named `MAX_PREFIX_IN_BRANCHES`**, so a caller discovered the fan-out ceiling by being rejected. It joins `maxRankedLimit` as a static on `WasmSdk` and `EvoSDK`.
- The `offset` docs still promised the skip had no ceiling and no caveats.

`deny_unknown_fields` keeps doing its job here: `timeRange` (#3740) is *not* added to the ranked input, because the ranked dispatcher rejects a time-range selection outright — a document belongs to every bucket containing its timestamp, so it would rank into several groups at once. The omission is the correct surface, and a pasted `timeRange` now fails locally rather than at the node.

### Document references — `wasm-dpp2`

`DataContract.documentTypeReferences(name)` and `.documentReferences` report what a contract's `refersTo` declarations point at, with a `DocumentPropertyReference` TypeScript union mirroring `DocumentPropertyReferenceTarget`.

This lives in wasm-dpp2 rather than wasm-sdk because it is parsed-contract metadata — no network, no `Sdk` handle, no async — and `pub use wasm_dpp2::*` carries it to wasm-sdk and evo-sdk with no changes in either. No `DocumentType` wrapper class was added: `DocumentTypeRef<'_>` cannot cross wasm-bindgen, so one would mean cloning the index map and schema on every access, and it is a ~20-accessor API commitment that deserves its own design pass.

Two details that mirror consensus exactly:

- It walks `flattened_properties()` (dotted paths), which is what both the registration-time and write-time validators walk, and how they build their error `path`. The nested map would produce paths no consensus error matches and would miss nested declarations.
- An omitted `contractId` resolves to the declaring contract's own id, because consensus computes `contract_id.unwrap_or(contract.id())` and treats an explicit self-id identically. `ref.contractId.equals(contract.id)` is the self-reference test, with no null branch.

`refersTo` is only parsed from protocol version 14 onward, so a contract deserialized against an earlier version reports none. That is faithful to what consensus enforced at that version, and it is documented as a trap, since `toJSON()` still shows the raw keyword.

**On the consensus errors: the codes already reach JS.** I traced the full path and confirmed `wasm-sdk/src/error.rs` passes `Some(err.code as i32)` through on the broadcast path, so `e.code === 40123` works today. This PR only names them — a `DocumentReferenceErrorCode` enum (40120-40125) and a `ConsensusError.code` getter — so callers can branch without a message regex.

An exhaustive `match` guard test in rs-dpp makes adding a sixth reference target a compile error in the crate that owns the enum, where whoever adds it will see that the JS mirror needs updating.

## How Has This Been Tested?

Offline only — no functional tests against a running node, since ranked/having need a PV14 network.

- **32 Rust unit tests** in `document_ranked.rs` covering the builder shape (the `with_select` ordering regression, the `$count` sentinel, direction defaults, offset, limit ceilings asserted against `MAX_RANKED_LIMIT` / `MAX_HAVING_LIMIT` rather than literals), the having grammar (derived aggregate, two-operand `between`, rejected non-contiguous operators, optional ordering), the index pins (equality, `null`, rejected range and repeated pins), the serde surface (`orderBy` / cursors / `offset` rejected rather than dropped), and result shaping (hex convention, string decode, absent vs. undecodable keys, the avg scale).
- **13 wasm-dpp2 specs**, including a pre-v14 gate regression asserting the accessor reports nothing while the raw schema still carries the keyword.
- **6 stubbed evo-sdk facade specs** and 2 wasm-sdk specs pinning the `pub use wasm_dpp2::*` fan-out.

Full suites and lints, all exit 0:

- `cargo fmt --check --all`; `cargo clippy -p dpp -p wasm-dpp2 -p wasm-sdk --all-targets --all-features -D warnings`; `cargo test` for all three; `cargo check -p wasm-sdk --target wasm32-unknown-unknown`
- `wasm-dpp2` 1155 tests, `wasm-sdk` 400, `evo-sdk` 219 (mocha + karma); all three package lints clean

I also verified in the generated `.d.ts` that every new type emits properly rather than falling back to `any`.

Rebased onto current `v4.2-dev` and re-verified end to end — first after #4388 restructured `rs-sdk/src/platform/documents/`, and again after #4401, #3740 and #4486 landed. The second rebase is the one that mattered: #4401 added a third field to `RankedEntry`, which broke four struct literals in the host tests (`E0063`), and a fifth in the `wasm32`-gated module that no host-target build compiles. This branch's own green CI run predates all of it.

New coverage for the `IN` surface: 8 host tests (the pin is carried intact, the branch ceiling asserted against `MAX_PREFIX_IN_BRANCHES` rather than a literal, two branching `IN`s rejected, `OFFSET` × branching `IN` rejected, and — so that exclusion cannot pass while being over-broad — a singleton `IN` plus `OFFSET` accepted), 3 host tests on the branch-key decode, 2 `wasm32` tests pinning absent-vs-empty-string at the real JS boundary, and 1 evo-sdk spec on the new static.

Each new assertion was mutation-checked rather than assumed: forcing `branch_key_hex: None` fails exactly the tests that assert it, and emitting the field as `undefined` instead of omitting it fails the boundary test alone — which is the distinction the host target cannot see.

## Breaking Changes

None. Everything here is additive: new query entry points, new accessors on `DataContract`, and a new error-code enum. No existing signature or wire shape changes.

## Checklist:

---

### Notes for reviewers

Two adjacent gaps I found while tracing the error path and deliberately left out of scope:

1. `WasmSdkError` drops `StateTransitionBroadcastError.cause: Option<ConsensusError>`, so the structured consensus fields (`path`, `entityId`, `keyId`) are message text only. Carrying the serialized bytes as `consensusErrorBytes` would let JS use the existing `ConsensusError.deserialize` to read them. That is structured detail rather than branchability, which is what this PR set out to deliver.
2. The check_tx rejection path collapses consensus errors to `Error::Protocol` with `code: -1`, erasing the code. Document transitions cannot take that path (full state validation is gated behind `validates_full_state_on_check_tx()`, false for everything but masternode vote), but other transitions can.

One convention note: `document_type_reference.rs` uses `js_sys::Reflect::set`. `CONVENTIONS.md` forbids that for *conversion shapes* (`toObject` / `toJSON`, where the rs-dpp serde derive is the source of truth); this is a getter assembling a collection, matching `DataContract::tokens` and `::groups`.

## Summary by CodeRabbit

- **New Features**
  - Added ranked top-K and bounded aggregate document queries, including proof-enabled variants.
  - Added helpers for ranked-query limits, branch limits, and average scaling.
  - Added accessors for `refersTo` document-reference metadata.
  - Added typed document-reference error codes and enhanced error details.
- **Documentation**
  - Expanded SDK guidance for ranked queries, aggregate filtering, proofs, document references, protocol version 14, and related errors.
- **Tests**
  - Added coverage for query validation, result handling, proofs, document references, and error-code mappings.


thepastaclaw (commented):

The ranked-query bindings correctly reuse Drive's versioned grammar and proof-verifying SDK fetch paths, but two JavaScript-facing issues remain: valid large integer group keys can reject an entire result, and the evo-sdk README names a helper that the EvoSDK class does not expose. Source: codex-general, codex-ffi-engineer, and codex-rust-quality reviewers; Claude final verifier.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — ffi-engineer (completed), `gpt-5.6-sol` — rust-quality (completed)

🟡 2 suggestion(s)

<details>

```

In `packages/wasm-sdk/src/queries/document_ranked.rs`:
- [SUGGESTION] packages/wasm-sdk/src/queries/document_ranked.rs:886-892: Large integer group keys reject valid results at the WASM boundary
  `deserialize_value_for_key` can legitimately return `Value::I64`, `Value::U64`, `Value::I128`, or `Value::U128` for indexed group-by properties. This conversion first passes those values through `serde_json`; the subsequent JSON-compatible serde-wasm-bindgen serializer rejects 64-bit values outside JavaScript's safe-integer range, while values outside serde_json's 64-bit number domain fail even earlier. A valid verified result can therefore reject the entire Promise instead of returning the documented lossless `groupKeyHex` fallback. Fall back to the non-human-readable object conversion, which emits exact JavaScript `BigInt` values, and add a WASM-runtime regression test with a group key above `Number.MAX_SAFE_INTEGER` so the actual exported result path is exercised.

In `packages/js-evo-sdk/README.md`:
- [SUGGESTION] packages/js-evo-sdk/README.md:120: README points to a nonexistent EvoSDK method
  The README directs callers to `EvoSDK.maxRankedLimit()`, but `maxRankedLimit()` is generated only as a static method on the exported `WasmSdk` class. The `EvoSDK` facade forwards `setLogLevel` and `getLatestVersionNumber`, but does not define this helper, so the documented invocation fails type checking and produces a runtime TypeError. Document `WasmSdk.maxRankedLimit()` instead, or add an initialization-aware forwarding method to `EvoSDK` if that is the intended public API.
```
</details>

packages/wasm-sdk/src/queries/document_ranked.rs
```diff
@@ -0,0 +1,1763 @@
...
+
+    set_field(&entry, "groupKeyHex", &JsValue::from_str(&parts.key_hex));
+
+    let group_value = match (&parts.decoded, parts.key_absent) {
+        (Some(value), _) => platform_value_to_json(value).map_err(WasmSdkError::from)?,
+        // Empty key: the group-by value was absent, which is a real
+        // group with a known meaning, not a decode failure.
+        (None, true) => JsValue::NULL,
+        (None, false) => JsValue::UNDEFINED,
+    };
```
thepastaclaw, line 1030: **🟡 Suggestion: Large integer group keys reject valid results at the WASM boundary**

`deserialize_value_for_key` can legitimately return `Value::I64`, `Value::U64`, `Value::I128`, or `Value::U128` for indexed group-by properties. This conversion first passes those values through `serde_json`; the subsequent JSON-compatible serde-wasm-bindgen serializer rejects 64-bit values outside JavaScript's safe-integer range, while values outside serde_json's 64-bit number domain fail even earlier. A valid verified result can therefore reject the entire Promise instead of returning the documented lossless `groupKeyHex` fallback. Fall back to the non-human-readable object conversion, which emits exact JavaScript `BigInt` values, and add a WASM-runtime regression test with a group key above `Number.MAX_SAFE_INTEGER` so the actual exported result path is exercised.

```suggestion
let group_value = match (&parts.decoded, parts.key_absent) {
    (Some(value), _) => platform_value_to_json(value)
        .or_else(|_| {
            wasm_dpp2::serialization::conversions::platform_value_to_object(value)
        })
        .map_err(WasmSdkError::from)?,
    // Empty key: the group-by value was absent, which is a real
    // group with a known meaning, not a decode failure.
    (None, true) => JsValue::NULL,
    (None, false) => JsValue::UNDEFINED,
};
```

<sub>source: ['codex']</sub>

packages/js-evo-sdk/README.md
```diff
@@ -96,6 +96,63 @@ The SDK organises its API into domain-specific facades, each accessible as a pro
...
+});
+
+for (const entry of page.entries) {
+  // `value` is exact fixed point for the avg axis — divide by `page.valueScale`,
+  // never by a hardcoded constant. `valueAsNumber` is a lossy display helper.
+  console.log(entry.rank, entry.groupValue, Number(entry.value) / Number(page.valueScale));
+}
+```
+
+`limit` is required and capped at `EvoSDK`'s `maxRankedLimit()` (a hard reject, not a clamp). `offset` skips ranks — `{ limit: 1, offset: 4 }` is "the 5th best" — and has no ceiling, because the skipped region is attested rather than walked.
```
thepastaclaw: **🟡 Suggestion: README points to a nonexistent EvoSDK method**

The README directs callers to `EvoSDK.maxRankedLimit()`, but `maxRankedLimit()` is generated only as a static method on the exported `WasmSdk` class. The `EvoSDK` facade forwards `setLogLevel` and `getLatestVersionNumber`, but does not define this helper, so the documented invocation fails type checking and produces a runtime TypeError. Document `WasmSdk.maxRankedLimit()` instead, or add an initialization-aware forwarding method to `EvoSDK` if that is the intended public API.

<sub>source: ['codex']</sub>

packages/wasm-sdk/src/queries/document_ranked.rs
shumkov: @QuantumExplorer sounds reasonable. Except that all looks good!

shumkov (changes requested):

packages/wasm-sdk/src/queries/document_ranked.rs
QuantumExplorer: Confirmed and fixed in 077f2f5 — `json_compatible()` sets `serialize_large_number_types_as_bigints: false`, so `serialize_u64`/`serialize_i64` error past `Number.MAX_SAFE_INTEGER` (they do not round), and `u128`/`i128` fail earlier in `serde_json`. `u64`/`i64`/`u128`/`i128` now cross as exact `BigInt`s via a `group_value_repr` classifier split out so host tests can assert it; narrower integer types keep `number`, so the JS type follows the property's declared type rather than a value's magnitude. Added `tests/unit/ranked-group-value.spec.ts` exercising the JS boundary through a test-only export (a key of `2^53+1`, `u64::MAX`, and a large negative `i64`), plus three host tests covering the variants that spec cannot express.

packages/js-evo-sdk/README.md
QuantumExplorer: Correct — `maxRankedLimit` was generated only as a static on `WasmSdk`, so the documented `EvoSDK` call was a TypeError. Fixed in 077f2f5 by adding the forwarding statics rather than retreating to `WasmSdk` in the docs, since `EvoSDK` was the intended surface: `EvoSDK.maxRankedLimit()` and `EvoSDK.rankedAverageScale()` now use the same initialization-aware pattern as `getLatestVersionNumber()`. README updated to `await EvoSDK.maxRankedLimit()` and both added to the static-helpers list, with specs in `tests/unit/sdk.spec.ts`.

packages/wasm-sdk/src/queries/document_ranked.rs
```diff
@@ -0,0 +1,1908 @@
...
+/// host-target test cannot reach them — `js_sys` panics off-wasm, so only a
+/// WASM-runtime spec can prove that an integer past
+/// `Number.MAX_SAFE_INTEGER` comes back exact instead of throwing. Mirrors
+/// `testJsValueToJson` in wasm-dpp2, which exists for the same reason.
+///
+/// Not part of the supported API surface.
+#[wasm_bindgen(js_name = "testRankedGroupValue")]
+pub fn test_ranked_group_value(value: JsValue) -> Result<JsValue, WasmSdkError> {
+    let decoded = js_value_to_platform_value(&value).map_err(WasmSdkError::from)?;
+    group_value_to_js(&decoded)
```
thepastaclaw: **🟡 Suggestion: Test helper is exported in the production WASM API**

`test_ranked_group_value` is compiled unconditionally with `#[wasm_bindgen]`, so wasm-bindgen includes `testRankedGroupValue` in the production JavaScript bundle and generated TypeScript declarations. The wasm-sdk bundle re-exports every generated symbol, and js-evo-sdk subsequently re-exports that WASM surface, so the comment saying this is unsupported does not keep consumers from discovering and depending on it. Exercise `group_value_to_js` with a `wasm-bindgen-test`, or gate this export behind a non-default testing feature that the test build explicitly enables.

<sub>source: ['codex']</sub>

thepastaclaw (commented):

The current head fixes both prior JavaScript-facing findings: wide integer group keys now cross as exact BigInts, and the documented EvoSDK ranked helpers now exist. One in-scope suggestion remains because the regression helper is exported through the production WASM and TypeScript API despite being documented as test-only.
Source: reviewer backend `gpt-5.6-sol` (general, Rust-quality, and FFI lanes); final verifier backend `claude-opus-4-6`; orchestration-only, not reviewer evidence: `openclaw-agent/cliproxy/gpt-5.6-sol`.

### Review provenance

🟡 1 suggestion(s)

<details>

```

In `packages/wasm-sdk/src/queries/document_ranked.rs`:
- [SUGGESTION] packages/wasm-sdk/src/queries/document_ranked.rs:947-960: Test helper is exported in the production WASM API
  `test_ranked_group_value` is compiled unconditionally with `#[wasm_bindgen]`, so wasm-bindgen includes `testRankedGroupValue` in the production JavaScript bundle and generated TypeScript declarations. The wasm-sdk bundle re-exports every generated symbol, and js-evo-sdk subsequently re-exports that WASM surface, so the comment saying this is unsupported does not keep consumers from discovering and depending on it. Exercise `group_value_to_js` with a `wasm-bindgen-test`, or gate this export behind a non-default testing feature that the test build explicitly enables.
```
</details>

packages/wasm-sdk/src/queries/document_ranked.rs
QuantumExplorer: Agreed — fixed in 2347ba3 by removing the export entirely rather than feature-gating it. A feature gate would have left the spec unrunnable against a normally-built `dist/`, which is what CI tests. Instead the property it checked is now enforced by a signature: `exact_integer_to_js` returns `JsValue` rather than `Result`, so "a wide integer group key can never reject its page" holds by construction — making any of those arms fallible would not compile, and the only fallible arm left is the JSON conversion that wide integers no longer reach. That is stronger than the spec was. Classification, the half that can actually regress, stays covered by three host tests over all twelve `Value` variants (including the narrow ints and `u128`/`i128` the spec could not express, since its input conversion normalized every JS number to `i64`); the residual it did cover is wasm-bindgen's own `JsValue::from` for primitive integers, already exercised via `entry.value` in the evo-sdk specs. `dist/sdk.d.ts` and `dist/sdk.js` now contain zero references to it.

thepastaclaw (commented):

The implementation is sound and the prior production WASM test-helper export is fixed, with no blocking issues found. Two in-scope test-coverage suggestions remain: wide-integer exactness is no longer exercised at the JavaScript boundary, and the WASM error-code test does not compare its mappings with DPP's canonical codes. Targeted host tests passed, and `cargo check -p wasm-sdk --target wasm32-unknown-unknown` passed using the installed LLVM clang.
Source: reviewer backend `gpt-5.6-sol`; final verifier backend `claude-opus-4-6`. Orchestration-only: `openclaw-agent/cliproxy/gpt-5.6-sol` (not reviewer evidence).

### Review provenance

🟡 2 suggestion(s)

<details>

```

In `packages/wasm-sdk/src/queries/document_ranked.rs`:
- [SUGGESTION] packages/wasm-sdk/src/queries/document_ranked.rs:940-946: Wide-integer regression no longer exercises the WASM boundary
  The host tests verify only that wide `Value` variants select `ExactBigInt`; they cannot execute `exact_integer_to_js` because `js_sys` panics off-wasm. Returning `JsValue` makes this path infallible, but it does not guarantee exactness or the required JavaScript `bigint` representation—a future `JsValue::from_f64(inner as f64)` implementation would still satisfy the signature and pass every current test while corrupting group keys. Add a private `wasm-bindgen-test`, or equivalent production-path WASM test, that renders signed and unsigned values above `Number.MAX_SAFE_INTEGER` and asserts both their JavaScript type and exact value without exporting a test-only production API.

In `packages/wasm-dpp2/src/consensus_error.rs`:
- [SUGGESTION] packages/wasm-dpp2/src/consensus_error.rs:109-143: Error-code test validates duplicated literals rather than DPP codes
  This test compares the WASM enum and `from_code` with the same numeric literals duplicated in their implementations. It therefore still passes if a JavaScript name is consistently assigned the wrong protocol code, such as two names being swapped. Construct each corresponding DPP reference-error variant, obtain its canonical code through `ErrorWithCode`, and assert that it maps to the intended `DocumentReferenceErrorCodeWasm` variant; this directly verifies the advertised correspondence with DPP's source of truth.
```
</details>

packages/wasm-sdk/src/queries/document_ranked.rs
```diff
@@ -0,0 +1,1911 @@
...
+/// this whole classification exists to provide, so it is enforced by the
+/// signature rather than left to a test to notice. Making any of these arms
+/// fallible would not compile without changing this return type.
+fn exact_integer_to_js(key: ExactIntegerKey) -> JsValue {
+    match key {
+        ExactIntegerKey::U64(inner) => JsValue::from(inner),
+        ExactIntegerKey::I64(inner) => JsValue::from(inner),
+        ExactIntegerKey::U128(inner) => JsValue::from(inner),
+        ExactIntegerKey::I128(inner) => JsValue::from(inner),
+    }
```
thepastaclaw, line 1004: **🟡 Suggestion: Wide-integer regression no longer exercises the WASM boundary**

The host tests verify only that wide `Value` variants select `ExactBigInt`; they cannot execute `exact_integer_to_js` because `js_sys` panics off-wasm. Returning `JsValue` makes this path infallible, but it does not guarantee exactness or the required JavaScript `bigint` representation—a future `JsValue::from_f64(inner as f64)` implementation would still satisfy the signature and pass every current test while corrupting group keys. Add a private `wasm-bindgen-test`, or equivalent production-path WASM test, that renders signed and unsigned values above `Number.MAX_SAFE_INTEGER` and asserts both their JavaScript type and exact value without exporting a test-only production API.

<sub>source: ['codex']</sub>

packages/wasm-dpp2/src/consensus_error.rs
```diff
@@ -20,6 +78,75 @@ impl ConsensusErrorWasm {
...
+            ),
+        ];
+
+        for (code, expected) in cases {
+            assert_eq!(
+                DocumentReferenceErrorCodeWasm::from_code(code),
+                Some(expected)
+            );
+            assert_eq!(expected as u32, code);
+        }
```
thepastaclaw: **🟡 Suggestion: Error-code test validates duplicated literals rather than DPP codes**

This test compares the WASM enum and `from_code` with the same numeric literals duplicated in their implementations. It therefore still passes if a JavaScript name is consistently assigned the wrong protocol code, such as two names being swapped. Construct each corresponding DPP reference-error variant, obtain its canonical code through `ErrorWithCode`, and assert that it maps to the intended `DocumentReferenceErrorCodeWasm` variant; this directly verifies the advertised correspondence with DPP's source of truth.

<sub>source: ['codex']</sub>

packages/wasm-sdk/src/queries/document_ranked.rs
QuantumExplorer: You are right, and my "stronger than the spec" claim last round was only true on the no-throw axis — a signature cannot pin representation. Added `#[wasm_bindgen_test]` cases in c3e25f3 behind `cfg(all(test, target_arch = "wasm32"))`, with the dependency scoped to that target, so nothing is exported: `dist/sdk.d.ts` and `dist/sdk.js` are unchanged. They call the same `exact_integer_to_js` the production result path calls and assert both `typeof === "bigint"` and exact equality for `2^53+1`, `u64::MAX`, `i64::MIN`/`MAX`, `u128::MAX` and `i128::MIN`, plus one case over the whole entry conversion, and an `assert_ne!` against the f64-rounded value. Verified against your exact hypothetical: with a `JsValue::from_f64(inner as f64)` body the host suite still reports 35 passed while the wasm suite fails 2. Run with `CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_RUNNER=wasm-bindgen-test-runner cargo test -p wasm-sdk --target wasm32-unknown-unknown` (documented in Cargo.toml); as with `packages/wasm-drive-verify`, no CI lane runs wasm tests today.

packages/wasm-dpp2/src/consensus_error.rs
QuantumExplorer: Agreed — the test was self-referential. Rewritten in c3e25f3 to construct all six real DPP errors (`ReferencedEntityNotFoundError`, `…DocumentTypeNotFound…`, `…DocumentTypeDeletable…`, `…IdentityKeyNotFound…`, `…IdentityKeyDisabled…`, `…KeyIdPropertyInvalid…`), read each canonical code through `ErrorWithCode`, and assert it maps to the intended `DocumentReferenceErrorCodeWasm` variant; a second test does the same through the `ConsensusError` wasm getters, and a third asserts the six codes are pairwise distinct. Verified non-vacuous by consistently swapping 40123 and 40124 in both the enum and `from_code`: the old test passed, the new one fails with "ReferencedIdentityKeyNotFound is advertised for an error DPP codes as 40123".

shumkov (dismissed):

packages/wasm-sdk/src/queries/document_ranked.rs
```diff
@@ -0,0 +1,1999 @@
...
+    document_type_name: String,
+    group_by: String,
+    aggregate: AggregateSelectInput,
+    limit: u32,
+    #[serde(default)]
+    direction: Option<String>,
+    #[serde(default)]
+    offset: Option<u32>,
+    #[serde(rename = "where", default)]
+    where_clauses: Option<Vec<JsonValue>>,
```
thepastaclaw, line 501: **🟡 Suggestion: Wide 128-bit equality pins cannot cross the query boundary**

The ranked and having inputs store each `where` clause as `serde_json::Value`, and `deserialize_required_query` routes the JavaScript object through `from_object` and `platform_value_from_object`. The configured serde-wasm-bindgen deserializer handles `deserialize_any` BigInts only within `i64::MIN..=u64::MAX`, while `serde_json::Value` cannot retain integers outside its 64-bit number domain. A JavaScript BigInt representing a valid `u128` value above `u64::MAX` or an `i128` value below `i64::MIN` is therefore rejected before `parse_where_clause` runs. DPP supports `U128` and `I128` indexed properties, and Drive's `encode_equality_prefix_values` serializes their equality pins, so valid compound ranked and having indexes are unreachable through these bindings. Preserve pin operands as `platform_value::Value` through deserialization using a conversion that supports 128-bit BigInts, and add wasm32 boundary tests for both sides of the 64-bit range.

<sub>source: ['codex']</sub>

thepastaclaw (commented):

Both prior test-coverage findings are fixed at the exact head, and the targeted consensus-error and wasm32 boundary tests pass. One in-scope suggestion remains: the new ranked and having query inputs cannot represent valid 128-bit equality pins because those operands are routed through `serde_json::Value` and a BigInt deserialization path limited to 64-bit values.
Source: reviewers `gpt-5.6-sol`; verifier `claude-opus-4-6`. Orchestration-only, not reviewer evidence: `openclaw-agent/cliproxy/gpt-5.6-sol`.

### Review provenance

🟡 1 suggestion(s)

<details>

```

In `packages/wasm-sdk/src/queries/document_ranked.rs`:
- [SUGGESTION] packages/wasm-sdk/src/queries/document_ranked.rs:448-449: Wide 128-bit equality pins cannot cross the query boundary
  The ranked and having inputs store each `where` clause as `serde_json::Value`, and `deserialize_required_query` routes the JavaScript object through `from_object` and `platform_value_from_object`. The configured serde-wasm-bindgen deserializer handles `deserialize_any` BigInts only within `i64::MIN..=u64::MAX`, while `serde_json::Value` cannot retain integers outside its 64-bit number domain. A JavaScript BigInt representing a valid `u128` value above `u64::MAX` or an `i128` value below `i64::MIN` is therefore rejected before `parse_where_clause` runs. DPP supports `U128` and `I128` indexed properties, and Drive's `encode_equality_prefix_values` serializes their equality pins, so valid compound ranked and having indexes are unreachable through these bindings. Preserve pin operands as `platform_value::Value` through deserialization using a conversion that supports 128-bit BigInts, and add wasm32 boundary tests for both sides of the 64-bit range.
```
</details>

packages/wasm-sdk/src/queries/document_ranked.rs
QuantumExplorer: Not changing the behaviour here — the mechanism you describe is exactly right, but I could not find a way to reach it, so I recorded the analysis in 068a645 instead. Pushing back rather than resolving, in case you see a route I missed.

The premise that breaks is "DPP supports U128 and I128 indexed properties": true of the Rust enum, not of the contract schema. The live schema path `DocumentPropertyType::try_from_value_map` sends `"integer"` to `find_integer_type_for_subschema_value` (property/mod.rs:2537), which reads `minimum`/`maximum` as `i64` and whose every branch tops out at `U64`/`I64` — `find_unsigned_integer_type_for_max_value` (:2586) and `find_integer_type_for_min_and_max_values` (:2598) both take `i64` and return at most `U64`/`I64`; `"number"` gives `F64`. The only constructors of `DocumentPropertyType::{U128, I128}` anywhere outside tests are the `#[deprecated]` `try_from_name` (:145) and the random-document-type generator. So no contract can declare a 128-bit property, hence no index property to pin, hence no valid compound ranked or having index is unreachable.

Two things I did confirm: your chain is correct wherever a 128-bit value could occur (serde-wasm-bindgen `deserialize_any` refuses BigInts outside `i64::MIN..=u64::MAX`, de.rs:313-318), and the bound is not specific to this PR — the eight pre-existing entry points fed by `DocumentsQueryInput` carry the identical `Vec<serde_json::Value>` where-clause shape. So if DPP ever gains a schema route to those widths this becomes real for all ten at once, and the fix has to bypass `deserialize_any` for the operand — reading `where` off the raw object with the field declared `IgnoredAny`, which serde-wasm-bindgen skips without inspecting (de.rs:612). That is written down at the input struct so the next person does not have to re-derive it.

Happy to implement it here if you would rather have it now, but it would be a surface-wide change and it would make `documents.ranked()` diverge from `documents.query()`.

thepastaclaw (approved):

The reviewed JavaScript surfaces preserve the SDK proof-verification path and reuse Drive’s versioned ranked/having query grammar; no in-scope correctness, security, or architectural defects remain. The prior 128-bit equality-pin finding is outdated because network contracts serialize raw schemas and reparse integer properties through an i64-bounded classifier that can produce at most U64 or I64, so no valid current index can require a wider pin.
Source: reviewers `gpt-5.6-sol`; final verifier `claude-opus-4-6`; orchestration-only and not reviewer evidence: `openclaw-agent/cliproxy/gpt-5.6-sol`.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — ffi-engineer (completed), `gpt-5.6-sol` — security-auditor (completed)

packages/wasm-sdk/src/queries/document_ranked.rs
thepastaclaw: Resolved in `068a645` — _Wide 128-bit equality pins cannot cross the query boundary_ no longer present.

QuantumExplorer: Correcting the auto-reconcile note above: nothing was fixed in `068a645`. That commit added 20 lines and 0 of them are code — `git show 068a645 -- packages/wasm-sdk/src/queries/document_ranked.rs | grep "^+" | grep -v "^+++" | grep -vE "^\+\s*(///|//)|^\+\s*$"` returns empty, and `where_clauses: Option<Vec<JsonValue>>` is still there on both inputs. The 64-bit operand bound remains, deliberately; the commit only records why it is not reachable (no schema route to a 128-bit property, so no 128-bit index property to pin — detail in my reply above).

Flagging because "no longer present" would otherwise read as the bound having been lifted, which it has not. The thread is still open on my side, so no action needed if you agree the case is unreachable; reopen the argument if you see a route I missed.

shumkov (dismissed):

thepastaclaw (commented):

The implementation preserves the SDK proof-verification path and delegates ranked and having grammar validation to Drive, with no blocking correctness or architectural issues found. Three JavaScript-facing documentation and diagnostic suggestions remain: two descriptions do not fully reflect the newly exposed prefix `in` grammar, and one strict-TypeScript README example accesses an unknown catch value without narrowing it.
Source: reviewers `gpt-5.6-sol`; final verifier `claude-opus-4-6`. Orchestration-only and not reviewer evidence: `openclaw-agent/cliproxy/gpt-5.6-sol`.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — ffi-engineer (completed), `gpt-5.6-sol` — rust-quality (completed)

🟡 3 suggestion(s)

<details>

```

In `packages/wasm-sdk/src/queries/document_ranked.rs`:
- [SUGGESTION] packages/wasm-sdk/src/queries/document_ranked.rs:92-96: Prefix-pin documentation omits the null-plus-IN restriction
  The generated TypeScript documentation enumerates the restrictions on a branching `in` and then describes `null` as generally legal. Drive's shared `encode_prefix_branches` rejects a branching `in` when another prefix property has a singleton `null` pin because the empty segment representing that absent value cannot occur in the branched proof's shared prefix or suffix. This affects both ranked and having queries. Document this combination explicitly while preserving the valid case where `null` is an element of the branching `in` itself.
- [SUGGESTION] packages/wasm-sdk/src/queries/document_ranked.rs:707-712: Ranked-query error still claims every prefix pin must use equality
  `detect_ranked_mode` accepts bounded `in` prefix pins, but the wrapper appended to every classifier error still instructs callers that all `where` entries must use `==`. Consequently, malformed `in` requests—such as two branching pins, an oversized branch list, or a branching pin combined with a non-zero offset—receive Drive's accurate error followed by contradictory guidance that says the supported operator itself is invalid. Update the suffix to describe the current grammar without obscuring Drive's specific diagnostic.

In `packages/js-evo-sdk/README.md`:
- [SUGGESTION] packages/js-evo-sdk/README.md:178-181: Narrow the caught error before reading its code
  The package uses strict TypeScript, under which a catch variable has type `unknown`. The new README example reads `e.code` directly, so callers pasting the documented code into a strict project receive a type error. Add an object-and-property guard before comparing the consensus code.
```
</details>

packages/wasm-sdk/src/queries/document_ranked.rs
```diff
@@ -0,0 +1,2297 @@
...
+ * multiply into a cartesian product of prefix walks inside one proof. A
+ * single-element `in` is normalized to `==` and never spends that
+ * budget. Entries merged from an `in` carry `branchKeyHex` to say which
+ * branch they came from.
+ *
+ * A branching `in` cannot combine with a non-zero `offset` — see
+ * `DocumentsRankedQuery.offset`.
+ *
+ * A `null` value is legal and addresses the subtree the write path
+ * creates for an *absent* optional value.
```
thepastaclaw, line 96: **🟡 Suggestion: Prefix-pin documentation omits the null-plus-IN restriction**

The generated TypeScript documentation enumerates the restrictions on a branching `in` and then describes `null` as generally legal. Drive's shared `encode_prefix_branches` rejects a branching `in` when another prefix property has a singleton `null` pin because the empty segment representing that absent value cannot occur in the branched proof's shared prefix or suffix. This affects both ranked and having queries. Document this combination explicitly while preserving the valid case where `null` is an element of the branching `in` itself.

```suggestion
 * A branching `in` cannot combine with a non-zero `offset` — see
 * `DocumentsRankedQuery.offset`.
 *
 * A branching `in` also cannot combine with a `null` pin on another
 * property, because the branched proof cannot express that empty shared
 * path segment. `null` as an element of the branching `in` itself remains
 * legal.
 *
 * A `null` value is otherwise legal and addresses the subtree the write
 * path creates for an *absent* optional value.
```

<sub>source: ['codex']</sub>

packages/wasm-sdk/src/queries/document_ranked.rs
```diff
@@ -0,0 +1,2297 @@
...
+        platform_version,
+    )
+    .map(|_| ())
+    .map_err(|e| {
+        WasmSdkError::invalid_argument(format!(
+            "not a well-formed ranked query: {e}. A ranked query is \
+             {{ groupBy, aggregate, limit }} plus optional {{ direction, offset, where }}; \
+             `where` entries must be `==` pins on the covering compound index's leading \
+             properties."
+        ))
```
thepastaclaw, line 712: **🟡 Suggestion: Ranked-query error still claims every prefix pin must use equality**

`detect_ranked_mode` accepts bounded `in` prefix pins, but the wrapper appended to every classifier error still instructs callers that all `where` entries must use `==`. Consequently, malformed `in` requests—such as two branching pins, an oversized branch list, or a branching pin combined with a non-zero offset—receive Drive's accurate error followed by contradictory guidance that says the supported operator itself is invalid. Update the suffix to describe the current grammar without obscuring Drive's specific diagnostic.

```suggestion
WasmSdkError::invalid_argument(format!(
    "not a well-formed ranked query: {e}. A ranked query is \
     {{ groupBy, aggregate, limit }} plus optional {{ direction, offset, where }}; \
     `where` entries must pin the covering compound index's leading properties with \
     `==` or bounded `in`; at most one `in` may contain multiple elements."
))
```

<sub>source: ['codex']</sub>

packages/js-evo-sdk/README.md
```diff
@@ -96,6 +101,89 @@ The SDK organises its API into domain-specific facades, each accessible as a pro
...
+
+When a write is rejected because a reference does not resolve, the consensus code reaches JS as `error.code`:
+
+```ts
+import { DocumentReferenceErrorCode } from '@dashevo/evo-sdk';
+
+try {
+  await sdk.documents.create({ document, identityKey, signer });
+} catch (e) {
+  if (e.code === DocumentReferenceErrorCode.ReferencedIdentityKeyDisabled) {
```
thepastaclaw, line 181: **🟡 Suggestion: Narrow the caught error before reading its code**

The package uses strict TypeScript, under which a catch variable has type `unknown`. The new README example reads `e.code` directly, so callers pasting the documented code into a strict project receive a type error. Add an object-and-property guard before comparing the consensus code.

```suggestion
try {
  await sdk.documents.create({ document, identityKey, signer });
} catch (e) {
  if (
    typeof e === 'object'
    && e !== null
    && 'code' in e
    && e.code === DocumentReferenceErrorCode.ReferencedIdentityKeyDisabled
  ) {
```

PR #4451, opened 2026-08-22
feat(platform-wallet): claim masternode credits with the owner or payout key

QuantumExplorer:
## Summary

Implements the masternode (evonode) identity credit withdrawal that `platform_wallet_manager_masternode_withdraw` had stubbed out ("pending verified signer"), with **both** keys Platform accepts for it, plus the preflight a UI needs to offer it correctly:

- **Owner key** (`ProviderOwnerKeys`, identity purpose `OWNER`): may withdraw, but Platform pays the **registered payout address** — an output script signed with it is rejected (`signature_purpose_matches_requirements`). The claim therefore carries no output script.
- **Transfer key** (the pubkey behind the registered P2PKH payout script, identity purpose `TRANSFER`): may withdraw to **any** destination.

### platform-wallet
- `PlatformWallet::masternode_withdrawal_keys` (sync, seedless, no network): which of the two keys this wallet holds — owner via account-xpub derive-and-compare over the `ProviderOwnerKeys` pool depth (same approach as the operator ownership check), transfer via the funding accounts' address-pool lookup (same lookup as `sign_message`) — plus the payout address.
- `PlatformWallet::masternode_withdraw` (async): fetch the owner identity (id = display-order proTxHash), select the matching `OWNER` / `TRANSFER` identity key (**refuses to broadcast** if the identity doesn't carry it), sign through a key-wallet `Signer` at the resolved derivation path via a `DerivedKeyIdentitySigner` adapter (binds the derived pubkey to the identity key hash before emitting the 65-byte recoverable signature, `dashcore::signer::sign` format), then `withdraw_credits_with_signer`.
- `select_transfer_withdrawal_key` next to the existing `select_owner_withdrawal_key`; both are now used in production (dead-code allow removed).

### platform-wallet-ffi
- New `platform_wallet_manager_masternode_withdrawal_keys` preflight.
- `platform_wallet_manager_masternode_withdraw` now takes a `MnemonicResolverHandle` + `use_owner_key` + optional `dest_address` and runs the claim through `MnemonicResolverCoreSigner` — the same resolver-backed signing as `core_wallet_sign_message` / `core_wallet_tx_builder_finalize` (seed never resident; handle-storage guard not held across the network/resolver round-trip). The masternode is resolved from the same aggregation `platform_wallet_manager_list_masternodes` renders.

### Outcome typing / owner-index hint (review follow-ups)
- Broadcast and result wait are split: a definitive rejection stays an ordinary retryable error; an ambiguous outcome is the new `PlatformWalletError::MasternodeWithdrawalUnconfirmed` → FFI code 42 `ErrorMasternodeWithdrawalUnconfirmed` → Swift `.masternodeWithdrawalUnconfirmed` / Kotlin `MasternodeWithdrawalUnconfirmed` (do NOT retry until the claimable balance is re-read — the nonce was consumed). The broadcast-outcome classifiers move out of the `shielded`-gated module into `broadcast_outcome`.
- `masternode_withdrawal_keys` and both FFIs take an optional owner-key index hint (persisted `ownerKeyIndex` / host address join), verified by derivation before the pool-depth scan.

### swift-sdk
- `PlatformWalletManager.masternodeWithdrawalKeys(walletId:proTxHash:)` → `MasternodeWithdrawalKeys` (`ownerKeyIndex`, `transferKeyInWallet`, `payoutAddress`, `canChooseDestination`, `preferredSigningKey`).
- `masternodeWithdraw(walletId:proTxHash:amountCredits:signingKey:destinationAddress:)` is now `async` (detached — the FFI blocks on the Platform round-trip) and returns the remaining balance.
- SwiftExampleApp: the Claim button is no longer gated off and calls the new API.

Consumer: dashpay/dashwallet-ios#1035 (evonode "Withdraw" screen).

## Test plan
- [x] `cargo test -p platform-wallet --lib -- masternode_withdrawal withdrawal::masternode_withdrawal_tests` (8 tests: key selectors incl. decoys, signer adapter signature verifies via `verify_hash_signature`, refuses foreign keys / mismatched derived key, DIP-3 owner path, flags)
- [x] `cargo test -p platform-wallet-ffi --lib -- masternode_withdrawal` (owner-key + destination rejected before touching handles; invalid handle paths)
- [x] `cargo clippy -p platform-wallet -p platform-wallet-ffi --all-targets -- -D warnings`, `cargo fmt`
- [x] `./build_ios.sh --target sim` (xcframework + SwiftExampleApp, warnings-as-errors)
- [ ] Testnet claim from the iOS wallet (see app PR)

## Summary by CodeRabbit

* **New Features**
  * Added masternode reward withdrawals supporting owner and transfer keys.
  * Added preflight checks showing available signing keys, payout addresses, and destination options.
  * Added support for custom withdrawal destinations when using a transfer key.
  * Updated the example app with an enabled claim flow and clearer signing-key information.

* **Bug Fixes**
  * Added explicit handling for withdrawals whose final outcome cannot be confirmed. These withdrawals are not safe to retry.
  * Improved broadcast outcome classification for failed and ambiguous transactions.


packages/rs-platform-wallet/src/wallet/masternode_withdrawal.rs
```diff
@@ -0,0 +1,693 @@
...
+            ))
+        })?;
+
+        let identity_signer = DerivedKeyIdentitySigner {
+            signer,
+            path,
+            expected_key_hash160: expected_hash160,
+        };
+
+        self.identity()
```
QuantumExplorer: [P1] Preserve ambiguous post-broadcast outcomes

`Identity::withdraw` broadcasts first and then waits for an affected-state proof. If broadcasting succeeds but confirmation or proof retrieval fails, this call converts the ambiguous result into an ordinary `InvalidIdentityData` error. The FFI therefore reports a normal failure and the UI permits retry; because the SDK nonce cache was already bumped, the retry can use the next valid nonce and both withdrawals can execute. Please split broadcast from confirmation and preserve a typed ‘may have executed—do not retry’ result until balance/nonce reconciliation completes.

QuantumExplorer (commented):
Requesting changes for one post-broadcast safety issue and two functional regressions. The focused platform-wallet and platform-wallet-ffi tests pass locally; the macOS wallet CI job was canceled by its 30-minute timeout rather than a test assertion.

packages/rs-platform-wallet/src/wallet/masternode_withdrawal.rs
```diff
@@ -0,0 +1,693 @@
...
+        // Under the wallet-manager read lock: the owner pool's depth and the
+        // payout address's derivation path. Dropped before the derive loop
+        // below (which takes the wallet's own state lock).
+        let (owner_scan_max, transfer_key) = {
+            let wm = self.wallet_manager().blocking_read();
+            let info = wm.get_wallet_info(&self.wallet_id()).ok_or_else(|| {
+                PlatformWalletError::WalletNotFound(hex::encode(self.wallet_id()))
+            })?;
+
+            let owner_scan_max = info
```
QuantumExplorer, line 192: [P2] Restored owner keys above index 19 disappear

When an imported/restored provider-owner account has no `highest_generated` watermark, this defaults the scan to 20 and `find_owner_key` checks only indices 0–19. The removed FFI deliberately accepted the persisted owner-key index because restored in-memory pools can be empty. A known owner at index 20 or higher now fails preflight and cannot withdraw. Please accept and verify a durable index hint, or obtain the scan bound from persisted ownership data.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Views/MasternodeDetailView.swift
```diff
@@ -259,13 +246,6 @@ struct MasternodeDetailView: View {
...
     }
 
-    /// Owner-key masternode withdrawal is not yet enabled: the Rust FFI
-    /// `platform_wallet_masternode_withdraw` is a deliberate stub returning
-    /// `ErrorWalletOperation` pending the verified-signer implementation, so
-    /// every confirmed claim would fail. Gate the Claim action off until
-    /// that lands; flip this to re-enable the existing confirmation flow.
-    private static let masternodeWithdrawEnabled = false
-
     /// Claim is enabled only with a positive balance and this wallet's
```
QuantumExplorer: [P2] Use the new key preflight in the example app

The app still gates on persisted `ownerInWallet` and later hard-codes `.owner`, without calling `masternodeWithdrawalKeys`. A transfer-only wallet therefore never receives a Claim action despite that being a newly supported withdrawal path, while stale persisted owner state can expose a claim that runtime resolution rejects. Please load the preflight result, gate on `canWithdraw`, and use its selected key and destination capability.

packages/rs-platform-wallet/src/wallet/masternode_withdrawal.rs
QuantumExplorer: Done in fbdcffc121: the claim now builds/signs locally, then `broadcast()` and `wait_for_affected_state()` are split. A definitive rejection (consensus verdict / transport verdict, via the shared `broadcast_outcome` classifiers moved out of the shielded-gated module) stays an ordinary retryable error; anything ambiguous after the transition left the wallet is a new typed `PlatformWalletError::MasternodeWithdrawalUnconfirmed` → FFI code 42 `ErrorMasternodeWithdrawalUnconfirmed` → Swift `.masternodeWithdrawalUnconfirmed` / Kotlin `MasternodeWithdrawalUnconfirmed`. The example app (and the iOS wallet PR) treat it as terminal: no retry until the claimable balance has been re-read.

packages/rs-platform-wallet/src/wallet/masternode_withdrawal.rs
QuantumExplorer: Done in fbdcffc121: `masternode_withdrawal_keys` and both FFIs (`…_withdrawal_keys`, `…_withdraw`) take an optional owner-key index hint (`has_owner_key_index_hint` + `owner_key_index_hint`; Swift `ownerKeyIndexHint:`). It is verified, not trusted — Rust derives the key at that index and compares hash160s first — and only then falls back to the pool-depth scan, so a restored wallet with an empty in-memory pool still resolves an owner key above index 19 when the host supplies the persisted index (`PersistentMasternode.ownerKeyIndex`) or its own address join.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Views/MasternodeDetailView.swift
QuantumExplorer: Done in fbdcffc121: `MasternodeDetailView` now loads `masternodeWithdrawalKeys(…, ownerKeyIndexHint: persisted index)` in `.task`, gates Claim on `keys.canWithdraw` (not the persisted `ownerInWallet`), signs with `keys.preferredSigningKey`, offers an editable destination (default payout address) when `canChooseDestination`, shows which key applies, and keeps Confirm disabled after an unconfirmed outcome until the balance is refreshed.

PR #4452, opened 2026-08-22
feat(drive)!: raise the daily withdrawal limit to 4000 Dash

QuantumExplorer:
## Issue being fixed or feature implemented

Platform caps the credits it pools into asset unlock (withdrawal) transactions at a flat **2000 Dash per 24 hours** (`daily_withdrawal_limit` v1, introduced in PV8 by #2287 to match Core v22's `LimitAmountV22`).

Core has since doubled its credit-pool unlock limit to **4000 Dash** (`LimitAmountV24`, [DIP-0165](https://github.com/dashpay/dips/pull/165), [dashpay/dash#6662](https://github.com/dashpay/dash/pull/6662)). This PR raises Platform's daily limit to match, gated on the next protocol version (PV14).

> [!NOTE]
> **Core coordination — none required.** Core's 4000 limit is gated on its `DEPLOYMENT_V24` hard fork, which as of Core v23.1.8 (and `develop`) is still `NEVER_ACTIVE` on mainnet and testnet (live only on devnets/regtest). That does **not** block this change: pre-V24 Core enforces `LimitAmountV22` (2000 Dash) **per block** — `currentLimit = min(credit_pool, 2000)` in `CCreditPoolManager::ConstructCreditPool`, with no sliding-window subtraction (the window only arrives with V24) — and the amount is checked only at block level (`CheckCreditPoolDiffForBlock` / the miner's skip), never at mempool acceptance. Platform pools at most 4 × 500 Dash = 2000 Dash per block, so a day's 4000 Dash is always fully minable; if two batches land in the same Core block interval, the excess simply waits in Core's mempool for the next block (well inside the 48-block unlock expiry). Once V24 activates, Core enforces 4000 Dash per 576-block window, which matches this limit exactly.

## What was done?

The limit is now data in the versioning system rather than a hard-coded method version:

* `rs-platform-version`: new `SystemLimits::daily_withdrawal_limit` (credits), next to `max_withdrawal_amount` / `min_withdrawal_amount` / `withdrawal_transactions_per_block_limit`. `SYSTEM_LIMITS_V1`–`V3` carry 2000 Dash (unchanged behavior for PV8–13; unused by PV1–7, whose method v0 derives the limit from total credits). New `SYSTEM_LIMITS_V4` = `V3` with 4000 Dash, used by PV14 only (`V3` shipped with PV13). PV14's doc header lists the change as its fourth consensus change.
[...]

PR #4457, opened 2026-08-22
feat(drive)!: make the daily withdrawal limit 15% of the total credits held a day ago

QuantumExplorer:

The daily withdrawal limit is a flat amount (2000 Dash since PV8, 4000 Dash in the unreleased PV14 after #4452). With the current volume of Platform → Core withdrawals the flat cap is hit routinely, transfers sit "pending" for hours, and a flat number needs a protocol bump every time Platform grows. It exists as a guardrail against an undiscovered inflation bug, so it should scale with what Platform actually holds — but lag behind it, so a sudden jump in the total cannot lift the limit immediately.

This makes the limit relative for PV14: **Platform pools at most 15% of the total credits it held a day ago into asset unlock transactions per 24 hours**, never below one maximal withdrawal (`max_withdrawal_amount`, 500 Dash), never above Core's unlock capacity per day (`max_daily_withdrawal_amount`, 4000 Dash = `LimitAmountV24`), and with the flat 2000 Dash still applying for the first day after activation while no recorded total is a day old yet. Amounts pooled in the last 24 hours keep counting against the maximum exactly as before; only the base of the maximum changes.

## What was done?

**Rule (`rs-dpp`)** — `daily_withdrawal_limit` v2 takes the day-old total as an `Option`: `Some(total)` → `min(max(daily_withdrawal_limit_percent × total, max_withdrawal_amount), max_daily_withdrawal_amount)` (percent and cap are new `SystemLimits` fields, `Some(15)` / `Some(4000 Dash)` in `SYSTEM_LIMITS_V4`, `None` before; the floor guarantees every accepted withdrawal eventually fits and cannot block the FIFO pooling queue behind it; the cap is Core's credit-pool unlock capacity per day — pooling more than Core mines only cycles unlocks through expiry and re-signing — and is raised together with Core); `None` (no recorded total is a day old yet, i.e. the first day after activation) → the flat 2000 Dash of v1, so the 24h lag cannot be skipped by inflating the total before or at activation. v1 goes back to a frozen `const fn` of 2000 Dash, since that flat value is now history and never changes again (`SystemLimits::daily_withdrawal_limit` from #4452 is removed). `DPP_METHOD_VERSIONS_V3` (new, PV14) selects v2.

**Base (`rs-drive`)** — Platform keeps no history of its total credits, so PV14 adds one under the withdrawals root tree (`WITHDRAWAL_TOTAL_CREDITS_HISTORY_KEY = [4]`): key = block time (ms, big-endian), value = `TOTAL_SYSTEM_CREDITS` at the end of that block's state transitions, written **only when the total changed** — the limit reads the entry in force a day ago and an entry describes the total until the next one, so a block that leaves the total untouched (document traffic and fees don't move it; asset locks, withdrawals and epoch core rewards do) costs one reverse `limit 1` read and no write.
* `Drive::record_total_credits_history` — compares with the latest entry, and on a change inserts this block's entry and prunes, bounded, every entry older than the one the limit reads this block (so the reference entry is never pruned, even across a chain halt).
* `Drive::fetch_total_credits_in_platform_a_day_ago` — the entry recorded at the latest block at least 24h before the given time (`RangeToInclusive(..=now-24h)`, reverse, limit 1); `None` while no entry is that old (no younger fallback).
* `calculate_current_withdrawal_limit` v1 — hands that lagged total (or `None`) to `daily_withdrawal_limit`, keeps `available = daily_maximum - pooled_in_last_24h`. The dispatcher now takes `&BlockInfo`.
* The tree is created at genesis for PV ≥ 14 (`add_initial_withdrawal_state_structure_operations`) and by `transition_to_version_14` for upgrading networks.

**Per-block event (`rs-drive-abci`)** — `record_total_credits_history_for_withdrawals` (`OptionalFeatureVersion`, `Some(0)` in the new `DRIVE_ABCI_METHOD_VERSIONS_V10`, `None` before) runs in `run_block_proposal` after `process_block_fees_and_validate_sum_trees` — the last thing in a block that can move the total (epoch Core rewards land there) — and before the app hash is taken, so the entry carries the block's final total and is part of its state; it only writes when the total moved. `DRIVE_ABCI_WITHDRAWAL_CONSTANTS_V3` adds `total_credits_history_prune_limit: 64`.

**Tables** — `DRIVE_VERSION_V9` (PV14's own) bumps `calculate_current_withdrawal_limit` to 1 and gains the two new slots as `OptionalFeatureVersion` (`Some(0)` in V2, `None` in V1, whose dispatchers return `VersionNotActive` — the history subtree does not exist before PV14); PV13 and earlier keep every table they shipped with, so replay is unchanged. PV14's doc header describes the rule.

### Why a day-old base
`TOTAL_SYSTEM_CREDITS` is the tracked total (it grows with asset locks and epoch core rewards, shrinks when withdrawal state transitions execute). Reading it live would let an inflated total raise the limit at once; the 24h lag gives a day to notice — including across activation, which is why the first day keeps the old flat limit instead of using a younger total. Same-day inflows do not raise the limit either; the percent is meant to be generous enough (today ~30k Dash held → ~4500 Dash/day, up from 2000).

### Core
Unchanged and not a dependency: pre-V24 Core caps unlocks at `LimitAmountV22` (2000 Dash) **per block** (`min(credit_pool, 2000)`, no sliding window — the window only arrives with V24) and checks the amount only at block level, so any daily total is minable across blocks; after V24 Core enforces 4000 Dash per 576-block window, which `max_daily_withdrawal_amount` never exceeds. Today's mainnet total (~30k Dash) would give 4500 Dash/day uncapped, so the effective limit there is the 4000 cap until Core raises its capacity; the percent governs below ~26.7k Dash.

## How Has This Been Tested?

* `cargo test -p dpp --all-features daily_withdrawal_limit` — v2 formula, `None` percent error, PV13 flat vs PV14 relative through the dispatcher.
* `cargo test -p drive` (withdrawals) — history record (incl. no write when the total is unchanged)/prune bounds, a-day-ago lookup incl. oldest fallback and the inclusive 24h boundary, `calculate_current_withdrawal_limit` v1 lag behaviour.
* `cargo test -p drive-abci --lib` — event records/no-ops by slot, `transition_to_version_14` creates the tree idempotently.
* `cargo test -p drive-abci --test strategy_tests should_record_the_total_credits_history_after_epoch_core_rewards` — production-path regression: hourly blocks, one-day epochs, +1 Core height per block; the history entry keyed by the epoch-change block (block 25) holds the post-reward total and the only earlier entry is genesis — written before fee processing, the reward would be keyed a block later and this fails.
* `cargo test -p drive-abci --test strategy_tests should_cap_withdrawals_at_the_relative_daily_limit` — new PV14 chain test cloned from the flat-limit one: 80 withdrawals of 50 Dash against 10,100 Dash held; day one still runs under the flat 2000 Dash (40 pool, 40 queue, 2000 locked); once a recorded total is a day old the reference is the 6,100 Dash left → 915 Dash/day, and after the first locks expire 18 more pool (58 broadcast / 22 queued / 900 locked); all 80 broadcast over the following 250 hourly blocks. The existing TEST_PLATFORM_V3 withdrawal tests are unchanged.
* `cargo check --workspace --all-targets`, clippy `-D warnings` on the touched crates.

## Breaking Changes

Consensus change gated on PV14: the amount of withdrawals pooled per day changes, and the withdrawals root tree gains a child (state root differs from PV14 builds before this PR). PV14 is unreleased.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Protocol version 14 introduces a dynamic daily withdrawal limit based on 15% of platform credits recorded at least 24 hours earlier.
  * Limits include a 2,000-Dash minimum bootstrap amount and a 4,000-Dash maximum cap.
  * Credit history is recorded after fees and rewards are processed, with automatic retention management.

* **Bug Fixes**
  * Corrected withdrawal-limit calculations to use historical rather than current credit totals.
  * Improved handling when historical data or limit configuration is unavailable.

* **Tests**
  * Added coverage for bootstrap behavior, rounding, minimum and maximum limits, history recording, and withdrawal processing.

### DIFF
--- a/packages/rs-dpp/src/withdrawal/daily_withdrawal_limit/v2/mod.rs
+++ b/packages/rs-dpp/src/withdrawal/daily_withdrawal_limit/v2/mod.rs
@@ -47,13 +47,22 @@ pub fn daily_withdrawal_limit_v2(
             )
         })?;
 
+    let max_withdrawal_amount = platform_version.system_limits.max_withdrawal_amount;
+    if max_daily_withdrawal_amount < max_withdrawal_amount {
+        // A cap below one maximal withdrawal would let an accepted withdrawal never fit the
+        // daily maximum; that is a contradictory configuration, not a limit to apply.
+        return Err(ProtocolError::CorruptedCodeExecution(format!(
+            "daily_withdrawal_limit v2 requires system_limits.max_daily_withdrawal_amount ({max_daily_withdrawal_amount}) to be at least max_withdrawal_amount ({max_withdrawal_amount})"
+        )));
+    }
+
     // u128 keeps `total * percent` from overflowing for any u64 total.
     let relative_limit = (total_credits_a_day_ago as u128) * (percent as u128) / 100;
     let relative_limit = Credits::try_from(relative_limit)
         .map_err(|_| ProtocolError::Overflow("daily withdrawal limit overflow"))?;
 
     Ok(relative_limit
-        .max(platform_version.system_limits.max_withdrawal_amount)
+        .max(max_withdrawal_amount)
         .min(max_daily_withdrawal_amount))
 }
 
@@ -158,5 +167,28 @@ mod tests {
             daily_withdrawal_limit_v2(Some(dash_to_credits!(100)), &platform_version),
             Err(ProtocolError::CorruptedCodeExecution(_))
         ));
+    }
+
+    #[test]
+    fn should_fail_when_the_cap_is_below_one_maximal_withdrawal() {
+        let mut platform_version = platform_version_with(Some(15));
+        platform_version.system_limits.max_daily_withdrawal_amount =
+            Some(dash_to_credits!(500) - 1);
+
+        // Whatever the total, a cap below the floor is a contradictory configuration.
+        for total in [0, dash_to_credits!(100), dash_to_credits!(30000)] {
+            assert!(matches!(
+                daily_withdrawal_limit_v2(Some(total), &platform_version),
+                Err(ProtocolError::CorruptedCodeExecution(_))
+            ));
+        }
+
+        // Exactly the floor is allowed and the limit is that floor.
+        platform_version.system_limits.max_daily_withdrawal_amount = Some(dash_to_credits!(500));
+        assert_eq!(
+            daily_withdrawal_limit_v2(Some(dash_to_credits!(30000)), &platform_version)
+                .expect("expected limit"),
+            dash_to_credits!(500)
+        );
     }
 }
--- a/packages/rs-dpp/src/withdrawal/daily_withdrawal_limit/v2/mod.rs
+++ b/packages/rs-dpp/src/withdrawal/daily_withdrawal_limit/v2/mod.rs
@@ -8,10 +8,13 @@ use platform_version::version::PlatformVersion;
 /// Using a day-old base means a sudden jump in the total credits does not raise
 /// the limit for a day.
 ///
-/// Two guards keep it usable:
+/// Three guards keep it usable:
 /// * it is never below `max_withdrawal_amount`, so every withdrawal Platform
 ///   accepts eventually fits the daily maximum and cannot block the pooling
 ///   queue behind it;
+/// * it is never above `max_daily_withdrawal_amount`, Core's credit-pool unlock
+///   capacity per day: pooling more than Core will mine only cycles those
+///   unlocks through expiry and re-signing;
 /// * while the total credits a day ago are not known (`None`: the history is
 ///   younger than a day, i.e. right after this rule activates), the flat limit
 ///   of version 1 applies, so the lag cannot be skipped by inflating the total
@@ -34,12 +37,24 @@ pub fn daily_withdrawal_limit_v2(
             )
         })?;
 
+    let max_daily_withdrawal_amount = platform_version
+        .system_limits
+        .max_daily_withdrawal_amount
+        .ok_or_else(|| {
+            ProtocolError::CorruptedCodeExecution(
+                "daily_withdrawal_limit v2 requires system_limits.max_daily_withdrawal_amount"
+                    .to_string(),
+            )
+        })?;
+
     // u128 keeps `total * percent` from overflowing for any u64 total.
     let relative_limit = (total_credits_a_day_ago as u128) * (percent as u128) / 100;
     let relative_limit = Credits::try_from(relative_limit)
         .map_err(|_| ProtocolError::Overflow("daily withdrawal limit overflow"))?;
 
-    Ok(relative_limit.max(platform_version.system_limits.max_withdrawal_amount))
+    Ok(relative_limit
+        .max(platform_version.system_limits.max_withdrawal_amount)
+        .min(max_daily_withdrawal_amount))
 }
 
 #[cfg(test)]
@@ -53,6 +68,7 @@ mod tests {
             .system_limits
             .daily_withdrawal_limit_percent = percent;
         platform_version.system_limits.max_withdrawal_amount = dash_to_credits!(500);
+        platform_version.system_limits.max_daily_withdrawal_amount = Some(dash_to_credits!(4000));
         platform_version
     }
 
@@ -61,14 +77,15 @@ mod tests {
         let platform_version = platform_version_with(Some(15));
 
         assert_eq!(
-            daily_withdrawal_limit_v2(Some(dash_to_credits!(30000)), &platform_version)
+            daily_withdrawal_limit_v2(Some(dash_to_credits!(20000)), &platform_version)
                 .expect("expected limit"),
-            dash_to_credits!(4500)
+            dash_to_credits!(3000)
         );
+        // Rounds down to whole credits.
         assert_eq!(
-            daily_withdrawal_limit_v2(Some(Credits::MAX), &platform_version)
+            daily_withdrawal_limit_v2(Some(dash_to_credits!(500) + 7), &platform_version)
                 .expect("expected limit"),
-            ((Credits::MAX as u128) * 15 / 100) as Credits
+            dash_to_credits!(500)
         );
     }
 
@@ -95,6 +112,29 @@ mod tests {
     }
 
     #[test]
+    fn should_never_exceed_cores_unlock_capacity_per_day() {
+        let platform_version = platform_version_with(Some(15));
+
+        // 15% of 30000 Dash is 4500 Dash, above what Core mines per day.
+        assert_eq!(
+            daily_withdrawal_limit_v2(Some(dash_to_credits!(30000)), &platform_version)
+                .expect("expected limit"),
+            dash_to_credits!(4000)
+        );
+        assert_eq!(
+            daily_withdrawal_limit_v2(Some(Credits::MAX), &platform_version)
+                .expect("expected limit"),
+            dash_to_credits!(4000)
+        );
+        // Just under the boundary the percent still applies.
+        assert_eq!(
+            daily_withdrawal_limit_v2(Some(dash_to_credits!(26666)), &platform_version)
+                .expect("expected limit"),
+            dash_to_credits!(3999.9)
+        );
+    }
+
+    #[test]
     fn should_keep_the_flat_limit_until_the_lagged_total_is_known() {
         let platform_version = platform_version_with(Some(15));
 
@@ -105,9 +145,15 @@ mod tests {
     }
 
     #[test]
-    fn should_fail_when_the_percent_is_not_configured() {
+    fn should_fail_when_the_percent_or_the_cap_is_not_configured() {
         let platform_version = platform_version_with(None);
+        assert!(matches!(
+            daily_withdrawal_limit_v2(Some(dash_to_credits!(100)), &platform_version),
+            Err(ProtocolError::CorruptedCodeExecution(_))
+        ));
 
+        let mut platform_version = platform_version_with(Some(15));
+        platform_version.system_limits.max_daily_withdrawal_amount = None;
         assert!(matches!(
             daily_withdrawal_limit_v2(Some(dash_to_credits!(100)), &platform_version),
             Err(ProtocolError::CorruptedCodeExecution(_))
--- a/packages/rs-dpp/src/withdrawal/daily_withdrawal_limit/v2/mod.rs
+++ b/packages/rs-dpp/src/withdrawal/daily_withdrawal_limit/v2/mod.rs
@@ -1,16 +1,29 @@
 use crate::fee::Credits;
+use crate::withdrawal::daily_withdrawal_limit::v1::daily_withdrawal_limit_v1;
 use crate::ProtocolError;
 use platform_version::version::PlatformVersion;
 
 /// Relative daily withdrawal limit: `daily_withdrawal_limit_percent` (from the
-/// protocol version's system limits) of the total credits Platform held a day
-/// ago. The caller passes that lagged total as `total_credits_in_platform_a_day_ago`;
-/// using a day-old base means a sudden jump in the total credits does not raise
+/// protocol version's system limits) of the total credits Platform held a day ago.
+/// Using a day-old base means a sudden jump in the total credits does not raise
 /// the limit for a day.
+///
+/// Two guards keep it usable:
+/// * it is never below `max_withdrawal_amount`, so every withdrawal Platform
+///   accepts eventually fits the daily maximum and cannot block the pooling
+///   queue behind it;
+/// * while the total credits a day ago are not known (`None`: the history is
+///   younger than a day, i.e. right after this rule activates), the flat limit
+///   of version 1 applies, so the lag cannot be skipped by inflating the total
+///   before or at activation.
 pub fn daily_withdrawal_limit_v2(
-    total_credits_in_platform_a_day_ago: Credits,
+    total_credits_in_platform_a_day_ago: Option<Credits>,
     platform_version: &PlatformVersion,
 ) -> Result<Credits, ProtocolError> {
+    let Some(total_credits_a_day_ago) = total_credits_in_platform_a_day_ago else {
+        return Ok(daily_withdrawal_limit_v1());
+    };
+
     let percent = platform_version
         .system_limits
         .daily_withdrawal_limit_percent
@@ -22,9 +35,11 @@ pub fn daily_withdrawal_limit_v2(
         })?;
 
     // u128 keeps `total * percent` from overflowing for any u64 total.
-    let limit = (total_credits_in_platform_a_day_ago as u128) * (percent as u128) / 100;
+    let relative_limit = (total_credits_a_day_ago as u128) * (percent as u128) / 100;
+    let relative_limit = Credits::try_from(relative_limit)
+        .map_err(|_| ProtocolError::Overflow("daily withdrawal limit overflow"))?;
 
-    Credits::try_from(limit).map_err(|_| ProtocolError::Overflow("daily withdrawal limit overflow"))
+    Ok(relative_limit.max(platform_version.system_limits.max_withdrawal_amount))
 }
 
 #[cfg(test)]
@@ -32,38 +47,69 @@ mod tests {
     use super::*;
     use crate::dash_to_credits;
 
-    #[test]
-    fn should_return_the_configured_percent_of_the_lagged_total() {
+    fn platform_version_with(percent: Option<u8>) -> PlatformVersion {
         let mut platform_version = PlatformVersion::latest().clone();
         platform_version
             .system_limits
-            .daily_withdrawal_limit_percent = Some(15);
+            .daily_withdrawal_limit_percent = percent;
+        platform_version.system_limits.max_withdrawal_amount = dash_to_credits!(500);
+        platform_version
+    }
+
+    #[test]
+    fn should_return_the_configured_percent_of_the_lagged_total() {
+        let platform_version = platform_version_with(Some(15));
 
         assert_eq!(
-            daily_withdrawal_limit_v2(dash_to_credits!(30000), &platform_version)
+            daily_withdrawal_limit_v2(Some(dash_to_credits!(30000)), &platform_version)
                 .expect("expected limit"),
             dash_to_credits!(4500)
         );
-        // Rounds down to whole credits.
         assert_eq!(
-            daily_withdrawal_limit_v2(7, &platform_version).expect("expected limit"),
-            1
-        );
-        assert_eq!(
-            daily_withdrawal_limit_v2(Credits::MAX, &platform_version).expect("expected limit"),
+            daily_withdrawal_limit_v2(Some(Credits::MAX), &platform_version)
+                .expect("expected limit"),
             ((Credits::MAX as u128) * 15 / 100) as Credits
         );
     }
 
     #[test]
+    fn should_never_go_below_one_maximal_withdrawal() {
+        let platform_version = platform_version_with(Some(15));
+
+        // 15% of 2000 Dash is 300 Dash, below the 500 Dash a single withdrawal may carry.
+        assert_eq!(
+            daily_withdrawal_limit_v2(Some(dash_to_credits!(2000)), &platform_version)
+                .expect("expected limit"),
+            dash_to_credits!(500)
+        );
+        assert_eq!(
+            daily_withdrawal_limit_v2(Some(0), &platform_version).expect("expected limit"),
+            dash_to_credits!(500)
+        );
+        // Exactly at the boundary the percent takes over.
+        assert_eq!(
+            daily_withdrawal_limit_v2(Some(dash_to_credits!(4000)), &platform_version)
+                .expect("expected limit"),
+            dash_to_credits!(600)
+        );
+    }
+
+    #[test]
+    fn should_keep_the_flat_limit_until_the_lagged_total_is_known() {
+        let platform_version = platform_version_with(Some(15));
+
+        assert_eq!(
+            daily_withdrawal_limit_v2(None, &platform_version).expect("expected limit"),
+            dash_to_credits!(2000)
+        );
+    }
+
+    #[test]
     fn should_fail_when_the_percent_is_not_configured() {
-        let mut platform_version = PlatformVersion::latest().clone();
-        platform_version
-            .system_limits
-            .daily_withdrawal_limit_percent = None;
+        let platform_version = platform_version_with(None);
 
         assert!(matches!(
-            daily_withdrawal_limit_v2(dash_to_credits!(100), &platform_version),
+            daily_withdrawal_limit_v2(Some(dash_to_credits!(100)), &platform_version),
             Err(ProtocolError::CorruptedCodeExecution(_))
         ));
     }
--- a/packages/rs-dpp/src/withdrawal/daily_withdrawal_limit/v2/mod.rs
+++ b/packages/rs-dpp/src/withdrawal/daily_withdrawal_limit/v2/mod.rs
@@ -1,0 +1,70 @@
+use crate::fee::Credits;
+use crate::ProtocolError;
+use platform_version::version::PlatformVersion;
+
+/// Relative daily withdrawal limit: `daily_withdrawal_limit_percent` (from the
+/// protocol version's system limits) of the total credits Platform held a day
+/// ago. The caller passes that lagged total as `total_credits_in_platform_a_day_ago`;
+/// using a day-old base means a sudden jump in the total credits does not raise
+/// the limit for a day.
+pub fn daily_withdrawal_limit_v2(
+    total_credits_in_platform_a_day_ago: Credits,
+    platform_version: &PlatformVersion,
+) -> Result<Credits, ProtocolError> {
+    let percent = platform_version
+        .system_limits
+        .daily_withdrawal_limit_percent
+        .ok_or_else(|| {
+            ProtocolError::CorruptedCodeExecution(
+                "daily_withdrawal_limit v2 requires system_limits.daily_withdrawal_limit_percent"
+                    .to_string(),
+            )
+        })?;
+
+    // u128 keeps `total * percent` from overflowing for any u64 total.
+    let limit = (total_credits_in_platform_a_day_ago as u128) * (percent as u128) / 100;
+
+    Credits::try_from(limit).map_err(|_| ProtocolError::Overflow("daily withdrawal limit overflow"))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::dash_to_credits;
+
+    #[test]
+    fn should_return_the_configured_percent_of_the_lagged_total() {
+        let mut platform_version = PlatformVersion::latest().clone();
+        platform_version
+            .system_limits
+            .daily_withdrawal_limit_percent = Some(15);
+
+        assert_eq!(
+            daily_withdrawal_limit_v2(dash_to_credits!(30000), &platform_version)
+                .expect("expected limit"),
+            dash_to_credits!(4500)
+        );
+        // Rounds down to whole credits.
+        assert_eq!(
+            daily_withdrawal_limit_v2(7, &platform_version).expect("expected limit"),
+            1
+        );
+        assert_eq!(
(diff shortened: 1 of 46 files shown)


packages/rs-platform-version/src/version/system_limits/v4.rs
```diff
@@ -18,9 +19,9 @@ pub const SYSTEM_LIMITS_V4: SystemLimits = SystemLimits {
     max_transitions_in_documents_batch: 1,
     withdrawal_transactions_per_block_limit: 4,
     retry_signing_expired_withdrawal_documents_per_block_limit: 1,
-    max_withdrawal_amount: 50_000_000_000_000,   //500 Dash
-    daily_withdrawal_limit: 400_000_000_000_000, //4000 Dash (Core v24 limit; raised from 2000 in v14)
-    min_withdrawal_amount: 1_000_000,            //1000 duffs (raised from 190 in v12)
+    max_withdrawal_amount: 50_000_000_000_000, //500 Dash
+    daily_withdrawal_limit_percent: Some(15), // 15% of the total credits a day ago (replaces the flat 2000 Dash in v14)
```
thepastaclaw, line 25: **🔴 Blocking: The relative cap can permanently strand a valid withdrawal**

PV14 accepts individual withdrawals up to 500 Dash, but its daily maximum can be lower than that without any corresponding admission check. For example, a valid 400-Dash withdrawal against a 2,000-Dash reference total exceeds the 300-Dash daily maximum. Executing the transition has already removed the 400 Dash from `TOTAL_SYSTEM_CREDITS`, so without unrelated inflows the later reference total falls to 1,600 Dash and the maximum falls further to 240 Dash; the document never becomes eligible. `pool_withdrawals_into_transactions_queue_v1` also breaks when the oldest document exceeds the available limit, so this document prevents every later, smaller withdrawal from being pooled. Ensure every accepted withdrawal can eventually fit the daily maximum, support splitting, or change queue handling and admission rules so an unpoolable FIFO head cannot permanently block withdrawals.

<sub>source: ['codex']</sub>

packages/rs-drive-abci/src/execution/engine/run_block_proposal/v0/mod.rs
```diff
@@ -353,6 +353,14 @@ where
         // Prune anchors older than the configured retention depth
         self.prune_shielded_pool_anchors(block_proposal.height, transaction, platform_version)?;
 
+        // Record the total credits in Platform if this block changed it: the daily withdrawal
+        // limit is a share of the total credits Platform held a day ago, read from this history
+        self.record_total_credits_history_for_withdrawals(
+            &block_info,
+            transaction,
+            platform_version,
+        )?;
```
thepastaclaw: **🔴 Blocking: History is recorded before epoch rewards change total credits**

This records `TOTAL_SYSTEM_CREDITS` before `process_block_fees_and_validate_sum_trees` runs at lines 403-409. On an epoch change, that later path calls `add_distribute_fees_from_oldest_unpaid_epoch_pool_to_proposers_operations_v1`, which adds the epoch's Core rewards through `SystemOperationType::AddToSystemCredits`. The reward is therefore associated with the next block's timestamp rather than the block in which Platform received it. This produces the wrong reference total for the interval between those timestamps; after a chain halt longer than 24 hours, the pre-reward value remains eligible while the actual post-reward total is timestamped only at restart and delayed for another day. Record the final snapshot after all operations that mutate total credits have been applied. Pooling can still run before that snapshot because the empty-history fallback supplies the current total on the first activation block.

<sub>source: ['codex']</sub>

packages/rs-drive/src/drive/identity/withdrawals/fetch_total_credits_in_platform_a_day_ago/v0/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+            platform_version,
+        )? {
+            return Ok(Some(entry));
+        }
+
+        // The history is younger than a day: fall back to the oldest entry we have.
+        let mut oldest = Query::new();
+        oldest.insert_all();
+
+        self.fetch_first_recorded_total_credits(oldest, transaction, platform_version)
```
thepastaclaw: **🔴 Blocking: Bootstrap fallback removes the 24-hour inflation delay**

When no snapshot is at least 24 hours old, this uses the oldest available entry. The PV14 migration creates an empty history tree, while `run_block_proposal_v0` processes untrusted state transitions before recording the first entry. An inflation flaw exercised in the activation block therefore installs the already-inflated total as the reference immediately, defeating the guardrail's stated purpose of delaying a sudden increase for 24 hours. Inflation immediately before activation has the same result. Keep the frozen flat limit until a genuinely day-old snapshot exists, or populate historical state before activating the relative rule.

<sub>source: ['codex']</sub>

packages/rs-platform-version/src/version/drive_abci_versions/drive_abci_method_versions/v10.rs
```diff
@@ -0,0 +1,167 @@
...
+//       input/output balances, applied chargeable-failure credits) that pre-v13
+//       nodes never tracked (they passed no map, and PaidConsensusError had no
+//       balance-changes field). Only this helper changed — `process_raw_state_transitions`
+//       (the outer loop) stays at v0 — and it lives in a real _v1, so the _v0
+//       stays byte-identical to old nodes with no version conditional.
+// The storage method `store_address_balances_to_recent_block_storage` is
+// deliberately UNCHANGED (stays Some(0)) — its _v0 implementation did not change;
+// only which changes feed its map did. The expanded recorded set changes the
+// committed state root, so both bumps MUST stay inactive on v12 (see
+// DRIVE_ABCI_METHOD_VERSIONS_V8, which keeps both fields at 0).
```
thepastaclaw: **🟡 Suggestion: Remove the stale PV13 version-table commentary**

The header says this table was introduced in protocol version 13 and describes `process_validation_result` and `record_added_balance_outputs` as its changes. Both fields already have these values in `DRIVE_ABCI_METHOD_VERSIONS_V9`; the only PV14 delta in V10 is enabling `record_total_credits_history_for_withdrawals`. The correct V10 documentation at lines 35-38 is immediately preceded by this contradictory header, and the inherited `changed` comments at lines 130-145 repeat the same ambiguity. Remove the copied V9 header and describe inherited fields as unchanged where commentary is still useful.

thepastaclaw (changes requested):

The versioned implementation and bounded history queries are generally well structured, but three in-scope blockers remain: low balances can permanently strand valid withdrawals and block the FIFO queue, activation bootstrap bypasses the intended 24-hour inflation delay, and epoch rewards are timestamped one block late in the history. The version-table comments also incorrectly present inherited PV13 changes as the introduction of the PV14 table.
Source: gpt-5.6-sol (Codex general, security-auditor, and rust-quality reviewer lanes) and gpt-5.6-sol (final verifier); CodeRabbit inline evidence was also validated. openclaw-agent/cliproxy/gpt-5.6-sol was orchestration-only and is not reviewer evidence.

### Review provenance

<details>

```

In `packages/rs-platform-version/src/version/system_limits/v4.rs`:
- [BLOCKING] packages/rs-platform-version/src/version/system_limits/v4.rs:20-23: The relative cap can permanently strand a valid withdrawal
  PV14 accepts individual withdrawals up to 500 Dash, but its daily maximum can be lower than that without any corresponding admission check. For example, a valid 400-Dash withdrawal against a 2,000-Dash reference total exceeds the 300-Dash daily maximum. Executing the transition has already removed the 400 Dash from `TOTAL_SYSTEM_CREDITS`, so without unrelated inflows the later reference total falls to 1,600 Dash and the maximum falls further to 240 Dash; the document never becomes eligible. `pool_withdrawals_into_transactions_queue_v1` also breaks when the oldest document exceeds the available limit, so this document prevents every later, smaller withdrawal from being pooled. Ensure every accepted withdrawal can eventually fit the daily maximum, support splitting, or change queue handling and admission rules so an unpoolable FIFO head cannot permanently block withdrawals.

In `packages/rs-drive-abci/src/execution/engine/run_block_proposal/v0/mod.rs`:
- [BLOCKING] packages/rs-drive-abci/src/execution/engine/run_block_proposal/v0/mod.rs:356-362: History is recorded before epoch rewards change total credits
  This records `TOTAL_SYSTEM_CREDITS` before `process_block_fees_and_validate_sum_trees` runs at lines 403-409. On an epoch change, that later path calls `add_distribute_fees_from_oldest_unpaid_epoch_pool_to_proposers_operations_v1`, which adds the epoch's Core rewards through `SystemOperationType::AddToSystemCredits`. The reward is therefore associated with the next block's timestamp rather than the block in which Platform received it. This produces the wrong reference total for the interval between those timestamps; after a chain halt longer than 24 hours, the pre-reward value remains eligible while the actual post-reward total is timestamped only at restart and delayed for another day. Record the final snapshot after all operations that mutate total credits have been applied. Pooling can still run before that snapshot because the empty-history fallback supplies the current total on the first activation block.

In `packages/rs-drive/src/drive/identity/withdrawals/fetch_total_credits_in_platform_a_day_ago/v0/mod.rs`:
- [BLOCKING] packages/rs-drive/src/drive/identity/withdrawals/fetch_total_credits_in_platform_a_day_ago/v0/mod.rs:36-40: Bootstrap fallback removes the 24-hour inflation delay
  When no snapshot is at least 24 hours old, this uses the oldest available entry. The PV14 migration creates an empty history tree, while `run_block_proposal_v0` processes untrusted state transitions before recording the first entry. An inflation flaw exercised in the activation block therefore installs the already-inflated total as the reference immediately, defeating the guardrail's stated purpose of delaying a sudden increase for 24 hours. Inflation immediately before activation has the same result. Keep the frozen flat limit until a genuinely day-old snapshot exists, or populate historical state before activating the relative rule.

In `packages/rs-platform-version/src/version/drive_abci_versions/drive_abci_method_versions/v10.rs`:
- [SUGGESTION] packages/rs-platform-version/src/version/drive_abci_versions/drive_abci_method_versions/v10.rs:15-34: Remove the stale PV13 version-table commentary
  The header says this table was introduced in protocol version 13 and describes `process_validation_result` and `record_added_balance_outputs` as its changes. Both fields already have these values in `DRIVE_ABCI_METHOD_VERSIONS_V9`; the only PV14 delta in V10 is enabling `record_total_credits_history_for_withdrawals`. The correct V10 documentation at lines 35-38 is immediately preceded by this contradictory header, and the inherited `changed` comments at lines 130-145 repeat the same ambiguity. Remove the copied V9 header and describe inherited fields as unchanged where commentary is still useful.
```
</details>

packages/rs-platform-version/src/version/system_limits/v4.rs
QuantumExplorer: Fixed in 6780d2a93e: the daily maximum is now floored at `max_withdrawal_amount` (`daily_withdrawal_limit` v2 returns `max(15% × day-old total, max_withdrawal_amount)`), so any accepted withdrawal eventually fits once the locks ahead of it expire and the FIFO head can no longer strand the queue; covered by `should_never_go_below_one_maximal_withdrawal`.

packages/rs-drive-abci/src/execution/engine/run_block_proposal/v0/mod.rs
QuantumExplorer: Fixed in 6780d2a93e: `record_total_credits_history_for_withdrawals` now runs right after `process_block_fees_and_validate_sum_trees` (so epoch Core rewards are in the total it records) and before the app hash is taken; pooling no longer depends on the current block's entry because the limit uses the flat 2000 Dash until a recorded total is a day old.

packages/rs-drive/src/drive/identity/withdrawals/fetch_total_credits_in_platform_a_day_ago/v0/mod.rs
QuantumExplorer: Fixed in 6780d2a93e: the oldest-entry fallback is gone — `fetch_total_credits_in_platform_a_day_ago` returns `None` until an entry is at least 24h old, and `daily_withdrawal_limit` v2 then applies the flat 2000 Dash of v1 (`should_keep_the_flat_limit_until_the_lagged_total_is_known`; the strategy test's first day now runs under the flat limit), so inflation before or at activation cannot become the base without the lag.

packages/rs-platform-version/src/version/drive_abci_versions/drive_abci_method_versions/v10.rs
QuantumExplorer: Fixed in 6780d2a93e: removed the V9 header copied into v10.rs and marked `process_validation_result` / `record_added_balance_outputs` as unchanged from V9 (v1 since v13); the doc comment now only describes the v14 delta.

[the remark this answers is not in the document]
QuantumExplorer: Fixed in 6780d2a93e: removed the stale v13 header and reworded the inherited `changed:` comments as unchanged-from-V9; the test was also renamed to `should_cap_withdrawals_at_the_relative_daily_limit` per the nit.

packages/rs-dpp/src/withdrawal/daily_withdrawal_limit/v2/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+                    .to_string(),
+            )
+        })?;
+
+    // u128 keeps `total * percent` from overflowing for any u64 total.
+    let relative_limit = (total_credits_a_day_ago as u128) * (percent as u128) / 100;
+    let relative_limit = Credits::try_from(relative_limit)
+        .map_err(|_| ProtocolError::Overflow("daily withdrawal limit overflow"))?;
+
+    Ok(relative_limit.max(platform_version.system_limits.max_withdrawal_amount))
```
thepastaclaw: **🔴 Blocking: The relative limit can exceed Core V24's unlock capacity**

This percentage is bounded only from below, so any day-old total above roughly 26,667 Dash yields more than 4000 Dash per day; the test's 30,000-Dash example already returns 4500 Dash. Core V24 independently computes `LimitAmountV24 - latelyUnlocked`, where `LimitAmountV24` is 4000 Dash and the rolling period is 576 blocks, and `CCreditPoolDiff::Unlock` rejects blocks exceeding that available amount. Core's mempool does not enforce the credit-pool capacity, so Platform can successfully broadcast transactions that miners cannot include; those documents remain BROADCASTED and Platform expires/re-signs them after 48 Core blocks, well before the 576-block window clears. Sustained Platform pooling above 4000 Dash therefore creates a growing retry backlog instead of reducing pending withdrawals. Cap the effective Platform rate to Core's active rolling-window capacity, or coordinate a Core rule with matching relative capacity before activating this behavior.

<sub>source: ['codex']</sub>

packages/rs-platform-version/src/version/drive_versions/drive_identity_method_versions/mod.rs
```diff
@@ -20,6 +20,8 @@ pub struct DriveIdentityWithdrawalMethodVersions {
     pub document: DriveIdentityWithdrawalDocumentMethodVersions,
     pub transaction: DriveIdentityWithdrawalTransactionMethodVersions,
     pub calculate_current_withdrawal_limit: FeatureVersion,
+    pub record_total_credits_history: FeatureVersion,
+    pub fetch_total_credits_in_platform_a_day_ago: FeatureVersion,
```
thepastaclaw: **🟡 Suggestion: Represent the PV14-only history methods as optional version slots**

These methods depend on the total-credits history subtree introduced in PV14, but their slots are plain `FeatureVersion`s and the V1 table used by PV1-PV13 assigns both version 0. Their public dispatchers consequently advertise the methods as active for historical protocol versions and route calls into storage paths that did not exist under those versions. Use `OptionalFeatureVersion`, set historical tables to `None`, set PV14 to `Some(0)`, and return `VersionNotActive` from the Drive dispatchers when the slot is absent.

<sub>source: ['codex']</sub>

packages/rs-drive-abci/src/execution/engine/run_block_proposal/v0/mod.rs
```diff
@@ -402,6 +402,16 @@ where
...
 
+        // Record the total credits in Platform if this block changed it: the daily withdrawal
+        // limit is a share of the total credits Platform held a day ago, read from this history.
+        // This runs after fees and epoch rewards, the last things in a block that can move the
+        // total, and before the app hash so the entry is part of this block's state.
+        self.record_total_credits_history_for_withdrawals(
+            &block_info,
+            transaction,
+            platform_version,
+        )?;
```
thepastaclaw, line 413: **🟡 Suggestion: Add a regression test for recording after epoch rewards**

The corrected ordering is consensus-visible, but the added event test manually changes system credits before directly invoking the event, and the strategy test does not assert the history entry produced by an epoch-boundary block with a nonzero Core reward. Add a production-path `run_block_proposal` test that crosses an epoch boundary and verifies that the entry keyed by that block contains the post-reward total. Without that assertion, moving the event before fee processing again would pass the current tests while restoring the previously identified incorrect withdrawal base.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The four prior findings are fixed: the relative limit is floored at one maximal withdrawal, bootstrap waits for genuinely day-old history, history is recorded after epoch rewards, and the stale V10 commentary is gone. One blocking compatibility issue remains because the uncapped percentage can exceed Core V24's fixed 4000-Dash rolling-window capacity; three non-blocking versioning, documentation, and regression-test issues also remain.
Source: gpt-5.6-sol (Codex general and rust-quality reviewers; final verifier). openclaw-agent/cliproxy/gpt-5.6-sol was orchestration-only and is not reviewer evidence.

### Review provenance

<details>

```

In `packages/rs-dpp/src/withdrawal/daily_withdrawal_limit/v2/mod.rs`:
- [BLOCKING] packages/rs-dpp/src/withdrawal/daily_withdrawal_limit/v2/mod.rs:38-42: The relative limit can exceed Core V24's unlock capacity
  This percentage is bounded only from below, so any day-old total above roughly 26,667 Dash yields more than 4000 Dash per day; the test's 30,000-Dash example already returns 4500 Dash. Core V24 independently computes `LimitAmountV24 - latelyUnlocked`, where `LimitAmountV24` is 4000 Dash and the rolling period is 576 blocks, and `CCreditPoolDiff::Unlock` rejects blocks exceeding that available amount. Core's mempool does not enforce the credit-pool capacity, so Platform can successfully broadcast transactions that miners cannot include; those documents remain BROADCASTED and Platform expires/re-signs them after 48 Core blocks, well before the 576-block window clears. Sustained Platform pooling above 4000 Dash therefore creates a growing retry backlog instead of reducing pending withdrawals. Cap the effective Platform rate to Core's active rolling-window capacity, or coordinate a Core rule with matching relative capacity before activating this behavior.

In `packages/rs-platform-version/src/version/drive_versions/drive_identity_method_versions/mod.rs`:
- [SUGGESTION] packages/rs-platform-version/src/version/drive_versions/drive_identity_method_versions/mod.rs:23-24: Represent the PV14-only history methods as optional version slots
  These methods depend on the total-credits history subtree introduced in PV14, but their slots are plain `FeatureVersion`s and the V1 table used by PV1-PV13 assigns both version 0. Their public dispatchers consequently advertise the methods as active for historical protocol versions and route calls into storage paths that did not exist under those versions. Use `OptionalFeatureVersion`, set historical tables to `None`, set PV14 to `Some(0)`, and return `VersionNotActive` from the Drive dispatchers when the slot is absent.

In `packages/rs-drive-abci/src/execution/engine/run_block_proposal/v0/mod.rs`:
- [SUGGESTION] packages/rs-drive-abci/src/execution/engine/run_block_proposal/v0/mod.rs:395-413: Add a regression test for recording after epoch rewards
  The corrected ordering is consensus-visible, but the added event test manually changes system credits before directly invoking the event, and the strategy test does not assert the history entry produced by an epoch-boundary block with a nonzero Core reward. Add a production-path `run_block_proposal` test that crosses an epoch boundary and verifies that the entry keyed by that block contains the post-reward total. Without that assertion, moving the event before fee processing again would pass the current tests while restoring the previously identified incorrect withdrawal base.

In `packages/rs-platform-version/src/version/drive_versions/drive_identity_method_versions/v2.rs`:
- [SUGGESTION] packages/rs-platform-version/src/version/drive_versions/drive_identity_method_versions/v2.rs:16-26: Update the identity V2 table header to include the withdrawal-limit change
  The header says V2 differs from V1 in exactly one behavior-neutral query-builder flip, but line 170 also changes `calculate_current_withdrawal_limit` from 0 to 1 and selects the consensus-visible day-lagged relative rule. Version-table comments are used to audit protocol behavior, so this currently hides one of PV14's substantive changes.
```
</details>

packages/rs-dpp/src/withdrawal/daily_withdrawal_limit/v2/mod.rs
QuantumExplorer: Fixed in 5db4b3e3b9: the relative limit is now capped at `SystemLimits::max_daily_withdrawal_amount` (`Some(4000 Dash)` in `SYSTEM_LIMITS_V4`, i.e. Core's `LimitAmountV24` per 576-block window) — `daily_withdrawal_limit` v2 returns `min(max(15% × day-old total, max_withdrawal_amount), max_daily_withdrawal_amount)`, covered by `should_never_exceed_cores_unlock_capacity_per_day`; raised together with Core.

packages/rs-platform-version/src/version/drive_versions/drive_identity_method_versions/mod.rs
QuantumExplorer: Fixed in 5db4b3e3b9: `record_total_credits_history` and `fetch_total_credits_in_platform_a_day_ago` are now `OptionalFeatureVersion` slots — `None` in `DRIVE_IDENTITY_METHOD_VERSIONS_V1` (PV1–13; the dispatchers return `DriveError::VersionNotActive`), `Some(0)` in V2 (PV14).

packages/rs-drive-abci/src/execution/engine/run_block_proposal/v0/mod.rs
QuantumExplorer: Fixed in 5db4b3e3b9: added the production-path strategy test `should_record_the_total_credits_history_after_epoch_core_rewards` — hourly blocks, one-day epochs, +1 Core height per block; it asserts the history entry keyed by the epoch-change block (block 25) holds the post-reward total and that the only earlier entry is the genesis one, so recording before fee processing would fail it.

packages/rs-dpp/src/withdrawal/daily_withdrawal_limit/v2/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+        })?;
+
+    // u128 keeps `total * percent` from overflowing for any u64 total.
+    let relative_limit = (total_credits_a_day_ago as u128) * (percent as u128) / 100;
+    let relative_limit = Credits::try_from(relative_limit)
+        .map_err(|_| ProtocolError::Overflow("daily withdrawal limit overflow"))?;
+
+    Ok(relative_limit
+        .max(platform_version.system_limits.max_withdrawal_amount)
+        .min(max_daily_withdrawal_amount))
```
thepastaclaw, line 66: **🟡 Suggestion: Validate that the daily cap preserves the withdrawal floor**

The v2 contract says its result is never below `max_withdrawal_amount`, and `SystemLimits` documents that the daily cap must be at least that floor. However, `.max(floor).min(cap)` returns `cap` whenever a supplied platform version has `max_daily_withdrawal_amount < max_withdrawal_amount`. The shipped V4 constants are coherent, but these version tables are manually assembled and `PlatformVersion` can be cloned with public fields; a contradictory snapshot would silently reintroduce an unpoolable maximal withdrawal instead of reporting corrupted configuration. Reject the mismatch explicitly and add a focused test for it.

```suggestion
let max_withdrawal_amount = platform_version.system_limits.max_withdrawal_amount;
    let max_daily_withdrawal_amount = platform_version
        .system_limits
        .max_daily_withdrawal_amount
        .ok_or_else(|| {
            ProtocolError::CorruptedCodeExecution(
                "daily_withdrawal_limit v2 requires system_limits.max_daily_withdrawal_amount"
                    .to_string(),
            )
        })?;

    if max_daily_withdrawal_amount < max_withdrawal_amount {
        return Err(ProtocolError::CorruptedCodeExecution(
            "system_limits.max_daily_withdrawal_amount must be at least system_limits.max_withdrawal_amount"
                .to_string(),
        ));
    }

    // u128 keeps `total * percent` from overflowing for any u64 total.
    let relative_limit = (total_credits_a_day_ago as u128) * (percent as u128) / 100;
    let relative_limit = Credits::try_from(relative_limit)
        .map_err(|_| ProtocolError::Overflow("daily withdrawal limit overflow"))?;

    Ok(relative_limit
        .max(max_withdrawal_amount)
        .min(max_daily_withdrawal_amount))
```

<sub>source: ['codex']</sub>

thepastaclaw (commented):

The relative withdrawal cap, activation lag, PV14-only slots, and post-reward history regression are implemented correctly, and the targeted DPP and strategy tests pass. Two nonblocking issues remain: the V2 identity table header omits its substantive withdrawal/history deltas, and v2 silently violates its floor if the configured daily cap is below the maximum single withdrawal. Source: gpt-5.6-sol (Codex general, security-auditor, rust-quality, and final verifier); openclaw-agent/cliproxy/gpt-5.6-sol was orchestration-only and not reviewer evidence.

### Review provenance

🟡 2 suggestion(s)

<details>

```

In `packages/rs-platform-version/src/version/drive_versions/drive_identity_method_versions/v2.rs`:
- [SUGGESTION] packages/rs-platform-version/src/version/drive_versions/drive_identity_method_versions/v2.rs:16-26: Update the identity V2 table header to include the withdrawal-limit change
  The header still says V2 differs from V1 in exactly one behavior-neutral query-builder flip. The table also changes `calculate_current_withdrawal_limit` to version 1 and activates both total-credits-history methods at lines 170–172. Those three slots select the consensus-visible day-lagged withdrawal rule and its storage support, so the current header is misleading when auditing or backporting protocol-version behavior.

In `packages/rs-dpp/src/withdrawal/daily_withdrawal_limit/v2/mod.rs`:
- [SUGGESTION] packages/rs-dpp/src/withdrawal/daily_withdrawal_limit/v2/mod.rs:40-57: Validate that the daily cap preserves the withdrawal floor
  The v2 contract says its result is never below `max_withdrawal_amount`, and `SystemLimits` documents that the daily cap must be at least that floor. However, `.max(floor).min(cap)` returns `cap` whenever a supplied platform version has `max_daily_withdrawal_amount < max_withdrawal_amount`. The shipped V4 constants are coherent, but these version tables are manually assembled and `PlatformVersion` can be cloned with public fields; a contradictory snapshot would silently reintroduce an unpoolable maximal withdrawal instead of reporting corrupted configuration. Reject the mismatch explicitly and add a focused test for it.
```
</details>

packages/rs-dpp/src/withdrawal/daily_withdrawal_limit/v2/mod.rs
QuantumExplorer: Fixed in f91070d252: `daily_withdrawal_limit_v2` now returns `CorruptedCodeExecution` when `max_daily_withdrawal_amount < max_withdrawal_amount` instead of applying the cap, with `should_fail_when_the_cap_is_below_one_maximal_withdrawal` covering the mismatch and the exact-floor boundary.

thepastaclaw (approved):

Both prior suggestions are fixed at the exact head: the V2 identity-method header now documents all PV14 withdrawal deltas, and the relative-limit implementation rejects a daily cap below one maximal withdrawal. The targeted DPP withdrawal-limit suite passes all 8 tests, and no remaining in-scope defects were reported or identified.
Source: gpt-5.6-sol (Codex general, security-auditor, and rust-quality reviewers and final verifier); openclaw-agent/cliproxy/gpt-5.6-sol was orchestration-only and not reviewer evidence.

### Review provenance

shumkov (approved):

PR #4462, opened 2026-08-23
fix(drive-abci): execute a different block at the same height/round instead of serving the stale context

QuantumExplorer:

App-side half of dashpay/tenderdash#1413.

`process_proposal` keeps its block execution context keyed by height and round only. When a **second, different block** arrives for the same height and round, it does one of two wrong things:

* proposer path (`block_hash` set, `proposer_results` cached): it returns the cached result **for any request hash** — i.e. the app hash of *our own* proposal is served for the network's block;
* otherwise: it returns `Err(BadRequest("received a process proposal request twice with different hash"))`, which reaches Tenderdash as an ABCI exception.

Tenderdash turns the first into a `mustValidate` panic (`wrong Block.Header.AppHash …`) and the second into a `mustEnsureProcess` panic, and its WAL replay re-creates the same situation on every restart, so the node never comes back without discarding the Tenderdash data.

This is exactly what happens to a validator that hands over from block sync to consensus while still behind: it proposes at a historical height (it is in that era's quorum), and the block the network really committed at that height and round then reaches it through consensus catch-up — same height, same round, different block. Observed on a mainnet evonode re-syncing from scratch (stall at 24174, own proposal at `h=24175 r=1`, the collision at 24176 where it was the genuine round-0 proposer).

## What was done?

* `process_proposal`: key the cache by block hash. Same hash → unchanged (proposer cache served; non-proposer re-run with the existing warning). Different hash at the same height/round → log a warning, drop the stale context, and execute the block we were asked about. Never serve the other block's result, never error.
* New strategy tests `test_cases/process_proposal_collision_tests.rs`:
  * `process_proposal_for_a_different_block_must_not_be_served_from_the_proposer_cache` — PrepareProposal + ProcessProposal of our own block at (H, 0), then ProcessProposal of a different block at (H, 0): the answer must be that block's app hash (checked against a clean execution of the same request). Before the fix: `Accept` with our own block's app hash.
  * `process_proposal_for_a_second_different_block_in_the_same_round_must_execute` — two different blocks at (H, 0), neither ours: the second must execute. Before the fix: the ABCI `BadRequest` above.

With this change the collision is harmless: the genuine block executes and commits and the node keeps catching up. The Tenderdash side (not proposing while behind, not evicting the catch-up peer) is tracked in dashpay/tenderdash#1413 and dashpay/tenderdash#1414.

## How Has This Been Tested?

* `cargo test -p drive-abci --test strategy_tests process_proposal_collision` — both tests red on the base, green with the fix.
* `cargo test -p drive-abci --test strategy_tests` — 95 passed, 0 failed, 4 ignored (pre-existing).
* `cargo test -p drive-abci --lib abci::handler` — 52 passed.
* `cargo clippy -p drive-abci --tests --features mocks` clean; `cargo fmt --all -- --check` clean.

## Breaking Changes

None. Valid flows (same block re-sent, next round, next height) are unchanged; only the "different block at the same height/round" case changes, from a wrong answer / error to execution.

## Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved proposal processing when different blocks arrive at the same height and round.
  * Prevented stale cached results from blocking execution of a new proposal.
  * Continued returning cached results safely when the same block is submitted again.

* **Tests**
  * Added regression coverage confirming competing proposals are executed independently, accepted, and assigned distinct application hashes.

### DIFF
--- a/packages/rs-drive-abci/tests/strategy_tests/test_cases/process_proposal_collision_tests.rs
+++ b/packages/rs-drive-abci/tests/strategy_tests/test_cases/process_proposal_collision_tests.rs
@@ -1,0 +1,256 @@
+//! Two proposals for the same height and round.
+//!
+//! A node that hands over from block sync to consensus while still behind becomes a
+//! proposer at a historical height (it is in that era's quorum) and builds a block of its own
+//! there. The committed block for that height then reaches it through consensus catch-up:
+//! same height, same round, a different block. Drive's `process_proposal` keeps an execution
+//! context keyed only by height and round, so it answers the second block from the first
+//! block's cache, or refuses it outright. Either answer wedges Tenderdash: a cached app hash
+//! fails `ValidateBlockWithRoundState` (panic in `ApplyCommit`), and an ABCI error panics in
+//! `mustEnsureProcess`. The WAL replays both proposals on restart, so the node never recovers.
+//!
+//! These tests assert the behaviour Drive needs: a proposal whose hash differs from the one
+//! in the execution context must be executed afresh, never served from the cache and never
+//! rejected with an error.
+#[cfg(test)]
+mod tests {
+    use crate::execution::run_chain_for_strategy;
+    use crate::strategy::{ChainExecutionOutcome, NetworkStrategy};
+    use dpp::block::block_info::BlockInfo;
+    use dpp::block::epoch::Epoch;
+    use dpp::dashcore::hashes::Hash;
+    use drive_abci::config::{ExecutionConfig, PlatformConfig, PlatformTestConfig};
+    use drive_abci::mimic::MimicExecuteBlockOptions;
+    use drive_abci::platform_types::platform_state::PlatformStateV0Methods;
+    use drive_abci::test::helpers::setup::TestPlatformBuilder;
+    use platform_version::version::PlatformVersion;
+    use strategy_tests::{IdentityInsertInfo, StartAddresses, StartIdentities, Strategy};
+    use tenderdash_abci::proto::abci::response_process_proposal::ProposalStatus;
+    use tenderdash_abci::proto::abci::RequestProcessProposal;
+    use tenderdash_abci::proto::google::protobuf::Timestamp;
+    use tenderdash_abci::proto::version::Consensus;
+    use tenderdash_abci::Application;
+
+    const BLOCK_SPACING_MS: u64 = 3000;
+
+    fn strategy() -> NetworkStrategy {
+        NetworkStrategy {
+            strategy: Strategy {
+                start_contracts: vec![],
+                operations: vec![],
+                start_identities: StartIdentities::default(),
+                start_addresses: StartAddresses::default(),
+                identity_inserts: IdentityInsertInfo::default(),
+                identity_contract_nonce_gaps: None,
+                signer: None,
+            },
+            total_hpmns: 50,
+            extra_normal_mns: 0,
+            validator_quorum_count: 10,
+            chain_lock_quorum_count: 10,
+            upgrading_info: None,
+            proposer_strategy: Default::default(),
+            rotate_quorums: false,
+            failure_testing: None,
+            query_testing: None,
+            verify_state_transition_results: false,
+            ..Default::default()
+        }
+    }
+
+    fn config() -> PlatformConfig {
+        PlatformConfig {
+            execution: ExecutionConfig {
+                verify_sum_trees: true,
+                ..Default::default()
+            },
+            block_spacing_ms: BLOCK_SPACING_MS,
+            testing_configs: PlatformTestConfig::default_minimal_verifications(),
+            ..Default::default()
+        }
+    }
+
+    /// A `RequestProcessProposal` for the next block after `outcome`, attributed to the
+    /// proposer at `proposer_index`, with `hash` as its block hash. `time_offset_ms` shifts
+    /// the block time so that two requests built from different offsets execute to
+    /// different app hashes.
+    fn next_block_request(
+        outcome: &ChainExecutionOutcome,
+        proposer_index: usize,
+        hash: [u8; 32],
+        time_offset_ms: u64,
+    ) -> RequestProcessProposal {
+        let platform_state = outcome.abci_app.platform.state.load();
+        let height = platform_state.last_committed_block_height() + 1;
+        let core_height = platform_state.last_committed_core_height();
+        let time_ms = outcome.end_time_ms + BLOCK_SPACING_MS + time_offset_ms;
+        let proposer = outcome.proposers[proposer_index].pro_tx_hash();
+
+        RequestProcessProposal {
+            txs: vec![],
+            proposed_last_commit: None,
+            misbehavior: vec![],
+            hash: hash.to_vec(),
+            height: height as i64,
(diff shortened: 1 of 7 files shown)


packages/rs-drive-abci/tests/strategy_tests/test_cases/process_proposal_collision_tests.rs
```diff
@@ -0,0 +1,256 @@
...
+        RequestProcessProposal {
+            txs: vec![],
+            proposed_last_commit: None,
+            misbehavior: vec![],
+            hash: hash.to_vec(),
+            height: height as i64,
+            time: Some(Timestamp {
+                seconds: (time_ms / 1000) as i64,
+                nanos: ((time_ms % 1000) * 1000) as i32,
+            }),
```
thepastaclaw: **🟡 Suggestion: Convert milliseconds to nanoseconds with the correct scale**

`Timestamp.nanos` is expressed in nanoseconds, but the millisecond remainder is multiplied by 1,000 instead of 1,000,000. The current test timestamps have no fractional-second remainder, so the existing assertions are unaffected, but a future fractional offset would silently lose most or all of that offset and could weaken timing-sensitive proposal tests.

```suggestion
nanos: ((time_ms % 1000) * 1_000_000) as i32,
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The new hash comparison correctly handles collisions after a block hash has been recorded, but the immediately post-PrepareProposal cache remains unbound and can still return a stale app hash for a different block at the same height and round. The regression test establishes the local block hash before submitting the competing block, so it does not cover this remaining path.
Source: reviewer evidence from codex-general, codex-security-auditor, and codex-rust-quality (exact backend model IDs were not included in the supplied projection); final verifier is Anthropic Claude Agent SDK (exact model ID not exposed); orchestration-only and not reviewer evidence: openclaw-agent/cliproxy/gpt-5.6-sol.

### Review provenance

<details>

```

In `packages/rs-drive-abci/src/abci/handler/process_proposal.rs`:
- [BLOCKING] packages/rs-drive-abci/src/abci/handler/process_proposal.rs:109-149: Hashless proposer cache can still answer a different block
  `PrepareProposal` stores `proposer_results` while leaving `block_hash` as `None`. The first `ProcessProposal` request at that round is therefore accepted from this cache whenever only the retained transaction count and Core chain-lock update match. The code does not compare the transaction bytes or other execution inputs such as block time, proposer, Core height, quorum, and app version. A competing empty block can consequently receive the prepared block's app hash, after which its own hash is attached to the stale context. If the two blocks execute differently, Tenderdash receives an app hash that does not belong to the requested block and can enter the same persistent failure this PR is intended to prevent. The new test first processes the local proposal through `mimic_execute_block`, which populates `block_hash`, so it bypasses this branch. Before reusing a hashless proposer result, compare the complete execution-relevant proposal identity, including exact transaction bytes; otherwise discard the context and execute the incoming request.

In `packages/rs-drive-abci/tests/strategy_tests/test_cases/process_proposal_collision_tests.rs`:
- [SUGGESTION] packages/rs-drive-abci/tests/strategy_tests/test_cases/process_proposal_collision_tests.rs:96-98: Convert milliseconds to nanoseconds with the correct scale
  `Timestamp.nanos` is expressed in nanoseconds, but the millisecond remainder is multiplied by 1,000 instead of 1,000,000. The current test timestamps have no fractional-second remainder, so the existing assertions are unaffected, but a future fractional offset would silently lose most or all of that offset and could weaken timing-sensitive proposal tests.
```
</details>

packages/rs-drive-abci/tests/strategy_tests/test_cases/process_proposal_collision_tests.rs
QuantumExplorer: Fixed in a2edbade78 — `nanos` now uses `* 1_000_000` so the sub-second part of `time_ms` round-trips through `Timestamp` (BlockProposal converts back via `to_millis()`; the old `* 1000` produced microseconds and truncated it to zero). Checked the rest of the PR diff: this was the only occurrence. Both collision tests pass with the corrected scale.

thepastaclaw (changes requested):

The hash-aware branch now correctly discards a stale execution context and executes a different block at the same height and round. However, the hashless context left immediately after PrepareProposal is still matched using only transaction count and the Core chain-lock update, so the first competing ProcessProposal can receive the prepared block's stale app hash; this leaves the PR's core failure mode unresolved.
Source: Codex general, security-auditor, and rust-quality reviewer backend: `gpt-5.6-sol`; final verifier backend: Anthropic Claude Agent SDK (exact model ID not exposed). Orchestration-only and not reviewer evidence: `openclaw-agent/cliproxy/gpt-5.6-sol`.

### Review provenance

🔴 1 blocking

<details>

```

In `packages/rs-drive-abci/src/abci/handler/process_proposal.rs`:
- [BLOCKING] packages/rs-drive-abci/src/abci/handler/process_proposal.rs:109-149: Hashless proposer cache can still answer a different block
  `prepare_proposal` executes the local candidate with `block_hash: None` and stores its response in `proposer_results`. On the first `ProcessProposal`, this branch reuses that response when only the retained transaction count and `core_chain_lock_update` match; it does not compare the accepted transaction bytes, block time, proposer, Core height, proposed app version, validator quorum, or validate the incoming consensus app version. These fields are consumed by `BlockProposal` and `run_block_proposal_v0` and can change execution or make the proposal invalid. A competing request with the same transaction count and chain-lock update can therefore receive the prepared candidate's app hash, after which lines 134-141 attach the competing request's hash to that stale context. This directly preserves the wrong-AppHash failure the PR is intended to eliminate. The regression test calls `mimic_execute_block` for the local proposal first, and that helper invokes `ProcessProposal` and records its hash before the competing request, so the test exercises only the corrected hash-present branch. Retain and compare a normalized identity covering every execution-relevant PrepareProposal/ProcessProposal input, including byte-for-byte accepted transactions, before serving this cache; otherwise discard the context and execute the incoming request.
```
</details>

packages/rs-drive-abci/tests/strategy_tests/test_cases/process_proposal_collision_tests.rs
thepastaclaw: Resolved in `a2edbad` — _Convert milliseconds to nanoseconds with the correct scale_ no longer present.

thepastaclaw (approved):

At exact head a5cab13bf4b73f6079807bc856417574c95f30ad, the hashless proposer-cache path now compares every input retained and consumed by BlockProposal execution, including exact accepted transaction bytes, before reusing the prepared result; mismatches discard the stale context and execute the incoming proposal. The prior blocking finding is fixed, and validation passed with all 5 targeted collision tests, cargo fmt, and drive-abci clippy; no in-scope issues remain.
Source: Codex reviewers: gpt-5.6-sol; final verifier: Anthropic Claude Agent SDK (exact model ID not exposed). Orchestration-only and not reviewer evidence: openclaw-agent/cliproxy/gpt-5.6-sol.

### Review provenance

lklimek (approved):
Conceptually approved, as we correctly treat tenderdash as source of truth for the block. Assuming the bot reviews are green.

PR #4463, opened 2026-08-23
fix(kotlin-sdk): typed retryable device-locked keystore errors + createWallet fail-fast

bfoss765:
## Field failure

During QA wallet creation (`PlatformWalletManager.createWallet` → `WalletStorage.storeMnemonic` → `KeystoreManager.encrypt`), Android Keystore2 denied the `MASTER_ALIAS` AES-256-GCM operation with `UserNotAuthenticatedException` on **two QA devices** — no engine anomaly, the same code byte-identical across AARs. `MASTER_ALIAS` carries `setUnlockedDeviceRequired(true)` on lock-screen devices and **no** `setUserAuthenticationRequired`, so that exception on this alias can only mean Keystore's device-locked gate. The environmental verdict: **Keystore2's internal lock-state tracking misreported "locked"** — the devices were demonstrably unlocked at the time. The denial surfaced as an opaque wallet-creation failure after the full native create + rollback dance (the existing rollback worked correctly).

## Changes

1. **Typed retryable error** — device-locked denials on the non-auth-gated master-alias `encrypt`/`decrypt` paths are classified (narrow, name/message-based, JVM-testable — the `isNoSecureLockScreenKeyGenFailure` discipline: any `UserNotAuthenticatedException` in the cause chain, or a `KeyStoreException`/`InvalidKeyException` whose message names the locked device) and rethrown as **`KeystoreDeviceLockedException`**, carrying the alias, the operation, and `KeyguardManager.isDeviceLocked` + `isKeyguardLocked` **sampled at throw time** — so logs distinguish a genuinely-locked device from the false-locked defect. Documented as retryable-after-unlock. The auth-gated identity-key aliases are deliberately untouched: their `UserNotAuthenticatedException` means "auth window closed" and keeps the `BiometricGate` prompt-and-retry contract.

2. **`createWallet` fail-fast pre-check** — if `KeyguardManager.isDeviceLocked` is true at entry, throw the typed exception **before** the native create: no native wallet, no Room rows, nothing to roll back. Lock state is threaded exactly the way the file already threads `isDeviceSecure`: a probe captured against the application context in `WalletStorage`'s companion, handed to `KeystoreManager`.

3. **Bounded false-locked retry in `storeMnemonic`** — a denial whose sampled state says the device is NOT actually locked retries up to 3 times over ~2s (250/750/1000 ms) before propagating. A genuinely-locked device fails fast with no retry (waiting 2 s cannot unlock a phone; the caller retries after unlock).

Explicitly **not** done (per scope): no `MASTER_ALIAS` rotation/regeneration, no non-lock-bound fallback alias, no key-generation parameter changes.

4. **Ride-along (separate commit): diagnosable file-logging install** — the field line "SDK tracing file logging (INFO)… NOT installed (subscriber already set or dir unwritable)" conflated two failure modes behind one native boolean. New `Sdk.installFileLogging` returns `INSTALLED` / `ALREADY_SET` / `SESSION_ROOT_UNWRITABLE` and logs the distinguishing condition (unwritable: the exact path; already-set: the in-process claimant when it went through this API — console `enableLogging` first is the common cause — plus the fix: install file logging first). The writability probe runs before any native call. The `tracing` crate cannot attach layers to an already-installed global subscriber, so a native re-route is not possible; `enableFileLogging` keeps its Boolean signature as a compat wrapper. (The kotlin-sdk module has no pre-existing fallback log directory to try, so none was invented.)

## Test evidence

`JAVA_HOME=$(brew JDK17) ./gradlew :sdk:testDebugUnitTest` (packages/kotlin-sdk, JDK 17.0.19):

- **Full module suite: 51 classes, 326 tests, 0 failures, 0 errors** (includes all pre-existing keystore/WalletStorage tests).
- New `KeystoreDeviceLockedDenialTest` (7 tests, plain JVM): classifier matrix (UNAE direct + wrapped in cause chain; KeyStoreException/InvalidKeyException naming the locked device; unrelated crypto/Keystore faults must NOT classify), mapping + at-throw-time sampling for both the false-locked and genuinely-locked classes, non-denials rethrown unchanged (same instance).
- New `WalletStorageDeviceLockedRetryTest` (7 tests, Robolectric): pre-check throws on `isDeviceLocked` without touching the master key; passes when unlocked and when the keyguard shows without a secure lock; retry succeeds on the 2nd attempt with the mnemonic round-tripping; gives up after the full 3-retry schedule (4 encrypt attempts) with nothing stored; genuinely-locked denial → exactly 1 attempt, immediate rethrow.
- New `SdkFileLoggingInstallTest` (3 tests, Robolectric): unwritable session root reported pre-native; writable/nested-dir probe true; file-as-root probe false.

## Swift parity

Checked `packages/swift-sdk`'s createWallet persist path (`Sources/SwiftDashSDK/Core/Wallet/WalletStorage.swift`): the mnemonic is a Keychain item stored with `kSecAttrAccessibleWhenUnlockedThisDeviceOnly` — Keychain semantics differ and there is no `setUnlockedDeviceRequired` equivalent in use. **The false-locked defect is Keystore2-specific; Swift needs nothing for this fix.** One honest gap noted for a possible follow-up (not required, genuinely-locked class only): a locked iOS device would surface as an untyped `WalletStorageError.keychainError(errSecInteractionNotAllowed)` — Swift has no typed locked-device error and no pre-check (`isProtectedDataAvailable`) today; grep shows `errSecInteractionNotAllowed` is handled nowhere in `Sources/`.

## App side

The app-side retry/cutover fix is dash-wallet#1543.

## Summary by CodeRabbit

- **New Features**
  - Added clearer file-logging installation results, including successful setup, existing logging ownership, and unavailable log destinations.
  - Added retryable device-lock errors with lock-state details for wallet security operations.
  - Wallet creation now verifies master-key availability before proceeding.

- **Bug Fixes**
  - Improved handling of temporary Keystore lock failures with bounded retries.
  - Prevented unnecessary retries when the device is genuinely locked.
  - Ensured sensitive mnemonic data is cleared after storage attempts, including failures and cancellation.

### DIFF
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/Sdk.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/Sdk.kt
@@ -9,6 +9,7 @@ import org.dashfoundation.dashsdk.errors.mapNativeErrors
 import org.dashfoundation.dashsdk.ffi.NativeCleaner
 import org.dashfoundation.dashsdk.ffi.NativeLoader
 import org.dashfoundation.dashsdk.ffi.SdkNative
+import java.io.File
 import java.net.HttpURLConnection
 import java.net.URI
 import java.util.concurrent.atomic.AtomicLong
@@ -155,6 +156,35 @@ class Sdk private constructor(
         ERROR(0), WARN(1), INFO(2), DEBUG(3), TRACE(4)
     }
 
+    /**
+     * Outcome of [installFileLogging]. The native installer returns a
+     * single boolean whose `false` conflates its two failure modes —
+     * field logs showed exactly that: "NOT installed (subscriber already
+     * set or dir unwritable)" with no way to tell which. This type keeps
+     * them apart so the failure is diagnosable from a log line.
+     */
+    enum class FileLoggingInstall {
+        /** This call installed the file-logging subscriber. */
+        INSTALLED,
+
+        /**
+         * A process-global tracing subscriber was already set (first init
+         * wins — e.g. console logging via [enableLogging] ran first, or
+         * another in-process library installed one). The `tracing` API has
+         * no way to attach layers to an already-installed subscriber, so
+         * file logging cannot be added after the fact; install file
+         * logging FIRST if both are wanted.
+         */
+        ALREADY_SET,
+
+        /**
+         * The session root could not be created or written — file logging
+         * was not attempted (the native installer would have failed on the
+         * same directory). The path is named in the logged warning.
+         */
+        SESSION_ROOT_UNWRITABLE,
+    }
+
     @Serializable
     private data class MasternodesEnvelope(
         val success: Boolean,
@@ -197,23 +227,110 @@ class Sdk private constructor(
 
         private val json = Json { ignoreUnknownKeys = true }
 
+        private const val LOG_TAG = "DashSdk"
+
         /** One-time native library load + `dash_sdk_init`. Idempotent. */
         fun initialize() = NativeLoader.ensureLoaded()
+
+        /**
+         * Best-effort record of which call claimed the process-global
+         * tracing subscriber, for the [ALREADY_SET][FileLoggingInstall.ALREADY_SET]
+         * diagnostic. In-process bookkeeping only — a subscriber installed
+         * outside this companion (another library) is invisible here, so a
+         * `null` value means "not through this API", not "none".
+         */
+        @Volatile
+        private var subscriberClaimedBy: String? = null
 
         /** Enable console (logcat) logging for SDK operations. */
         fun enableLogging(level: LogLevel = LogLevel.DEBUG) {
             initialize()
             SdkNative.enableLogging(level.value)
+            // First-wins bookkeeping: the global tracing subscriber can only
+            // be installed once per process, and console logging installs
+            // one — the most common reason a later enableFileLogging finds
+            // the slot taken.
+            if (subscriberClaimedBy == null) {
+                subscriberClaimedBy = "console logging (enableLogging)"
+            }
         }
 
         /**
          * Route the global tracing subscriber to per-bucket files under
-         * [sessionRoot]. Returns false if a subscriber was already
-         * installed or the path is unwritable.
+         * [sessionRoot]. Returns true only when THIS call installed it —
+         * boolean-compat wrapper around [installFileLogging], which callers
+         * should prefer: it reports WHICH condition failed (subscriber
+         * already set vs. session root unwritable, with the path) instead
+         * of an undiagnosable false.
          */
-        fun enableFileLogging(level: LogLevel = LogLevel.DEBUG, sessionRoot: String): Boolean {
+        fun enableFileLogging(level: LogLevel = LogLevel.DEBUG, sessionRoot: String): Boolean =
+            installFileLogging(level, sessionRoot) == FileLoggingInstall.INSTALLED
+
+        /**
+         * [enableFileLogging] with a diagnosable outcome. On failure the
+         * distinguishing condition is also logged as a warning (tag
+         * `DashSdk`), so field logs no longer show the ambiguous
+         * "NOT installed (subscriber already set or dir unwritable)":
+         *
+         * - [FileLoggingInstall.SESSION_ROOT_UNWRITABLE] — [sessionRoot]
+         *   could not be created/written; checked BEFORE touching the
+         *   native installer, and the logged warning names the exact path.
+         * - [FileLoggingInstall.ALREADY_SET] — the directory IS writable
+         *   but a global tracing subscriber already exists (first init
+         *   wins). The `tracing` API cannot re-route an installed
+         *   subscriber, so the fix is ordering: install file logging
+         *   before [enableLogging]. The warning names the in-process
+         *   claimant when it went through this API.
+         */
+        fun installFileLogging(
+            level: LogLevel = LogLevel.DEBUG,
+            sessionRoot: String,
+        ): FileLoggingInstall {
+            val root = File(sessionRoot)
+            if (!sessionRootWritable(root)) {
+                android.util.Log.w(
+                    LOG_TAG,
+                    "SDK file logging NOT installed: session root cannot be " +
+                        "created/written: ${root.absolutePath}",
+                )
+                return FileLoggingInstall.SESSION_ROOT_UNWRITABLE
+            }
             initialize()
-            return SdkNative.enableFileLogging(level.value, sessionRoot)
+            return if (SdkNative.enableFileLogging(level.value, sessionRoot)) {
+                subscriberClaimedBy = "file logging (enableFileLogging)"
+                FileLoggingInstall.INSTALLED
+            } else {
+                val claimant = subscriberClaimedBy
+                    ?: "something outside this API (another in-process library, " +
+                        "or a subscriber surviving from an earlier init)"
+                android.util.Log.w(
+                    LOG_TAG,
+                    "SDK file logging NOT installed at ${root.absolutePath}: the " +
+                        "directory is writable, so a global tracing subscriber was " +
+                        "already set — by $claimant. First init wins and the tracing " +
+                        "API cannot re-route an installed subscriber; call " +
+                        "enableFileLogging before enableLogging to get file logs.",
+                )
+                FileLoggingInstall.ALREADY_SET
+            }
+        }
+
+        /**
+         * Whether [root] exists (or can be created) as a directory we can
+         * write a file into — probed with a real create-and-delete, the
+         * same operation the native installer's `open_file` will perform.
+         * Factored out (and `internal`) so the pre-native gate is
+         * JVM-testable.
+         */
+        internal fun sessionRootWritable(root: File): Boolean = try {
+            root.mkdirs()
+            val probe = File(root, ".dash_sdk_write_probe")
+            probe.delete()
+            val created = probe.createNewFile()
+            probe.delete()
+            created && root.isDirectory
+        } catch (_: Exception) {
+            false
         }
 
         /**
--- a/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/Sdk.kt
+++ b/packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/Sdk.kt
@@ -178,9 +178,12 @@ class Sdk private constructor(
         ALREADY_SET,
 
         /**
-         * The session root could not be created or written — file logging
-         * was not attempted (the native installer would have failed on the
-         * same directory). The path is named in the logged warning.
+         * The session root — or one of the fixed log destinations the
+         * native installer creates/opens under it (`dash_sdk/run.log`,
+         * `build_info.txt`, …) — could not be created or written; file
+         * logging was not attempted (the native installer would have
+         * failed on the same path). The blocked path is named in the
+         * logged warning.
          */
         SESSION_ROOT_UNWRITABLE,
     }
@@ -233,25 +236,42 @@ class Sdk private constructor(
         fun initialize() = NativeLoader.ensureLoaded()
 
         /**
-         * Best-effort record of which call claimed the process-global
-         * tracing subscriber, for the [ALREADY_SET][FileLoggingInstall.ALREADY_SET]
-         * diagnostic. In-process bookkeeping only — a subscriber installed
-         * outside this companion (another library) is invisible here, so a
-         * `null` value means "not through this API", not "none".
+         * Best-effort record of which call is KNOWN to have claimed the
+         * process-global tracing subscriber, for the
+         * [ALREADY_SET][FileLoggingInstall.ALREADY_SET] diagnostic. Only
+         * [installFileLogging] can assert this: its native installer
+         * reports whether THIS call installed the subscriber. In-process
+         * bookkeeping only — a subscriber installed outside this companion
+         * (another library) is invisible here, so a `null` value means
+         * "not through this API", not "none".
          */
         @Volatile
         private var subscriberClaimedBy: String? = null
+
+        /**
+         * Best-effort record of the first [enableLogging] call, kept apart
+         * from [subscriberClaimedBy] because it is only an ATTEMPT: the
+         * native console installer returns nothing and deliberately
+         * swallows a lost `try_init()` race, so the call proves the console
+         * path ran — not that it won the subscriber slot. A library outside
+         * this API may already have held it.
+         */
+        @Volatile
+        private var subscriberAttemptedBy: String? = null
 
         /** Enable console (logcat) logging for SDK operations. */
         fun enableLogging(level: LogLevel = LogLevel.DEBUG) {
             initialize()
             SdkNative.enableLogging(level.value)
-            // First-wins bookkeeping: the global tracing subscriber can only
-            // be installed once per process, and console logging installs
-            // one — the most common reason a later enableFileLogging finds
-            // the slot taken.
-            if (subscriberClaimedBy == null) {
-                subscriberClaimedBy = "console logging (enableLogging)"
+            // First-wins bookkeeping, recorded as an ATTEMPT only, never as
+            // the claimant: console logging installs a subscriber when the
+            // slot is free — the most common reason a later
+            // enableFileLogging finds it taken — but the native call cannot
+            // report whether it actually won, so asserting a claim here
+            // would name the wrong claimant whenever another library got in
+            // first.
+            if (subscriberAttemptedBy == null) {
+                subscriberAttemptedBy = "console logging (enableLogging)"
             }
         }
 
@@ -272,26 +292,37 @@ class Sdk private constructor(
          * `DashSdk`), so field logs no longer show the ambiguous
          * "NOT installed (subscriber already set or dir unwritable)":
          *
-         * - [FileLoggingInstall.SESSION_ROOT_UNWRITABLE] — [sessionRoot]
-         *   could not be created/written; checked BEFORE touching the
-         *   native installer, and the logged warning names the exact path.
-         * - [FileLoggingInstall.ALREADY_SET] — the directory IS writable
-         *   but a global tracing subscriber already exists (first init
-         *   wins). The `tracing` API cannot re-route an installed
-         *   subscriber, so the fix is ordering: install file logging
-         *   before [enableLogging]. The warning names the in-process
-         *   claimant when it went through this API.
+         * - [FileLoggingInstall.SESSION_ROOT_UNWRITABLE] — [sessionRoot],
+         *   or one of the fixed log destinations the native installer
+         *   creates/opens under it, could not be created/written; checked
+         *   BEFORE touching the native installer, and the logged warning
+         *   names the exact blocked path.
+         * - [FileLoggingInstall.ALREADY_SET] — every fixed log destination
+         *   IS writable, yet the native installer reported failure. With
+         *   its I/O legs ruled out up front, that is attributed to a global
+         *   tracing subscriber already existing (first init wins). The
+         *   `tracing` API cannot re-route an installed subscriber, so the
+         *   fix is ordering: install file logging before [enableLogging].
+         *   The warning names the in-process claimant when the subscriber
+         *   verifiably went through this API, or the first in-process
+         *   ATTEMPT ([enableLogging]) as "most likely". Diagnosed by
+         *   elimination, not reported by the native boundary (a bare
+         *   boolean): an I/O failure racing in between the probe and the
+         *   install — or a destination a future native version adds —
+         *   would still be misattributed here.
          */
         fun installFileLogging(
             level: LogLevel = LogLevel.DEBUG,
             sessionRoot: String,
         ): FileLoggingInstall {
             val root = File(sessionRoot)
-            if (!sessionRootWritable(root)) {
+            val blocked = firstUnwritableLogDestination(root)
+            if (blocked != null) {
                 android.util.Log.w(
                     LOG_TAG,
-                    "SDK file logging NOT installed: session root cannot be " +
-                        "created/written: ${root.absolutePath}",
+                    "SDK file logging NOT installed: session-root log " +
+                        "destination cannot be created/written: " +
+                        blocked.absolutePath,
                 )
                 return FileLoggingInstall.SESSION_ROOT_UNWRITABLE
             }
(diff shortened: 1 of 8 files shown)


packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/Sdk.kt
```diff
@@ -197,23 +227,110 @@ class Sdk private constructor(
...
+        internal fun sessionRootWritable(root: File): Boolean = try {
+            root.mkdirs()
+            val probe = File(root, ".dash_sdk_write_probe")
+            probe.delete()
+            val created = probe.createNewFile()
+            probe.delete()
+            created && root.isDirectory
+        } catch (_: Exception) {
+            false
         }
```
thepastaclaw: **🔴 Blocking: The writability probe can delete a caller-owned entry**

The probe uses the fixed name `.dash_sdk_write_probe` and deletes that path before creating its own file. If the session root already contains a regular file with that name, the check destroys the file and its contents; it also removes a pre-existing empty directory with that name. Concurrent calls can delete each other's probe and return inconsistent results. Create a uniquely named temporary file atomically and clean up only the file created by this invocation.

```suggestion
internal fun sessionRootWritable(root: File): Boolean = try {
    root.mkdirs()
    if (!root.isDirectory) {
        false
    } else {
        val probe = File.createTempFile(".dash_sdk_write_probe_", null, root)
        try {
            true
        } finally {
            probe.delete()
        }
    }
} catch (_: Exception) {
    false
}
```

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/security/KeystoreManager.kt
```diff
@@ -318,13 +345,50 @@ open class KeystoreManager(
...
+        operation: String,
+    ): Nothing {
+        if (isDeviceLockedKeystoreDenial(e)) {
+            throw KeystoreDeviceLockedException(
+                alias = alias,
+                operation = operation,
+                lockState = sampleDeviceLockState(),
+                cause = e,
+            )
+        }
```
thepastaclaw: **🔴 Blocking: Restrict device-lock classification to the master alias**

The generic AES branches of `encrypt` and `decrypt` accept arbitrary aliases, and `secretKey` reuses an existing `SecretKey` under such an alias. If a host has provisioned a custom AES alias with `setUserAuthenticationRequired(true)`, an expired authentication window produces `UserNotAuthenticatedException`; this code incorrectly converts it to `KeystoreDeviceLockedException`, preventing the caller's normal authentication prompt-and-retry handling. The PR's classification is valid only for `MASTER_ALIAS`, whose key has no user-authentication gate, so enforce that condition in the mapper and add a regression test showing that a denial for another alias is rethrown unchanged.

```suggestion
internal fun rethrowClassifyingDeviceLockedDenial(
    e: Exception,
    alias: String,
    operation: String,
): Nothing {
    if (alias == MASTER_ALIAS && isDeviceLockedKeystoreDenial(e)) {
        throw KeystoreDeviceLockedException(
            alias = alias,
            operation = operation,
            lockState = sampleDeviceLockState(),
            cause = e,
        )
    }
    throw e
}
```

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/Sdk.kt
```diff
@@ -197,23 +227,110 @@ class Sdk private constructor(
...
+                        "or a subscriber surviving from an earlier init)"
+                android.util.Log.w(
+                    LOG_TAG,
+                    "SDK file logging NOT installed at ${root.absolutePath}: the " +
+                        "directory is writable, so a global tracing subscriber was " +
+                        "already set — by $claimant. First init wins and the tracing " +
+                        "API cannot re-route an installed subscriber; call " +
+                        "enableFileLogging before enableLogging to get file logs.",
+                )
+                FileLoggingInstall.ALREADY_SET
```
thepastaclaw, line 353: **🟡 Suggestion: A writable root does not prove native failure means ALREADY_SET**

The native boolean remains ambiguous after probing only the session root. `platform_wallet_enable_file_logging` separately creates and opens fixed destinations including `dash_sdk/run.log`, `platform_wallet/run.log`, other child paths, and `build_info.txt`. For example, if `$sessionRoot/dash_sdk` is a regular file, the root probe succeeds but native `create_dir_all($sessionRoot/dash_sdk)` fails; this branch then reports `ALREADY_SET` even when no subscriber exists. That contradicts the new API's diagnostic contract. Return a structured I/O-versus-subscriber result from the native boundary, or validate every destination required by the native installer before interpreting `false` as `ALREADY_SET`.

<sub>source: ['codex']</sub>

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/Sdk.kt
```diff
@@ -197,23 +227,110 @@ class Sdk private constructor(
...
         fun enableLogging(level: LogLevel = LogLevel.DEBUG) {
             initialize()
             SdkNative.enableLogging(level.value)
+            // First-wins bookkeeping: the global tracing subscriber can only
+            // be installed once per process, and console logging installs
+            // one — the most common reason a later enableFileLogging finds
+            // the slot taken.
+            if (subscriberClaimedBy == null) {
+                subscriberClaimedBy = "console logging (enableLogging)"
+            }
```
thepastaclaw, line 275: **🟡 Suggestion: enableLogging records a claimant even when native installation fails**

`SdkNative.enableLogging` returns `Unit`, while `dash_sdk_enable_logging` deliberately ignores a failed `try_init()` when another global subscriber is already installed. Calling `enableLogging()` after another library has claimed the subscriber therefore records `console logging (enableLogging)` even though that call installed nothing. A later file-logging failure names the wrong claimant. Make the native console installer return whether `try_init()` succeeded, or record this value as an attempted API call rather than asserting that it claimed the subscriber.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The master-alias wallet pre-check and bounded retry are well covered, but the new mapper does not actually enforce its master-alias-only contract. The logging ride-along can delete caller-owned data and can still misidentify both the native failure category and the subscriber claimant, leaving two blocking issues and two diagnostic correctness issues. Source: reviewer gpt-5.6-sol (general and security-auditor); CodeRabbit inline review (backend model undisclosed); final verifier grok-4.5. Orchestration-only: openclaw-agent/cliproxy/gpt-5.6-sol (not reviewer evidence).

### Review provenance

<details>

```

In `packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/Sdk.kt`:
- [BLOCKING] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/Sdk.kt:325-334: The writability probe can delete a caller-owned entry
  The probe uses the fixed name `.dash_sdk_write_probe` and deletes that path before creating its own file. If the session root already contains a regular file with that name, the check destroys the file and its contents; it also removes a pre-existing empty directory with that name. Concurrent calls can delete each other's probe and return inconsistent results. Create a uniquely named temporary file atomically and clean up only the file created by this invocation.
- [SUGGESTION] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/Sdk.kt:299-314: A writable root does not prove native failure means ALREADY_SET
  The native boolean remains ambiguous after probing only the session root. `platform_wallet_enable_file_logging` separately creates and opens fixed destinations including `dash_sdk/run.log`, `platform_wallet/run.log`, other child paths, and `build_info.txt`. For example, if `$sessionRoot/dash_sdk` is a regular file, the root probe succeeds but native `create_dir_all($sessionRoot/dash_sdk)` fails; this branch then reports `ALREADY_SET` even when no subscriber exists. That contradicts the new API's diagnostic contract. Return a structured I/O-versus-subscriber result from the native boundary, or validate every destination required by the native installer before interpreting `false` as `ALREADY_SET`.
- [SUGGESTION] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/Sdk.kt:246-255: enableLogging records a claimant even when native installation fails
  `SdkNative.enableLogging` returns `Unit`, while `dash_sdk_enable_logging` deliberately ignores a failed `try_init()` when another global subscriber is already installed. Calling `enableLogging()` after another library has claimed the subscriber therefore records `console logging (enableLogging)` even though that call installed nothing. A later file-logging failure names the wrong claimant. Make the native console installer return whether `try_init()` succeeded, or record this value as an attempted API call rather than asserting that it claimed the subscriber.

In `packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/security/KeystoreManager.kt`:
- [BLOCKING] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/security/KeystoreManager.kt:383-390: Restrict device-lock classification to the master alias
  The generic AES branches of `encrypt` and `decrypt` accept arbitrary aliases, and `secretKey` reuses an existing `SecretKey` under such an alias. If a host has provisioned a custom AES alias with `setUserAuthenticationRequired(true)`, an expired authentication window produces `UserNotAuthenticatedException`; this code incorrectly converts it to `KeystoreDeviceLockedException`, preventing the caller's normal authentication prompt-and-retry handling. The PR's classification is valid only for `MASTER_ALIAS`, whose key has no user-authentication gate, so enforce that condition in the mapper and add a regression test showing that a denial for another alias is rethrown unchanged.
```
</details>

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/Sdk.kt
bfoss765: Fixed in 16c4afb9e6. The probe now creates a uniquely named temp file — `File.createTempFile(".dash_sdk_write_probe", null, dir)` — and deletes only the file this invocation created. A pre-existing caller-owned entry named `.dash_sdk_write_probe` (file or directory) is never touched, and concurrent probes can no longer delete each other's file and misreport a writable root. Pinned by `SdkFileLoggingInstallTest.shouldNotDeleteACallerOwnedEntryNamedLikeTheProbe` (a caller file with content survives the probe with its content intact) and `shouldLeaveNoProbeResidueBehind`.

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/Sdk.kt
```diff
@@ -197,23 +227,110 @@ class Sdk private constructor(
...
+         * Factored out (and `internal`) so the pre-native gate is
+         * JVM-testable.
+         */
+        internal fun sessionRootWritable(root: File): Boolean = try {
+            root.mkdirs()
+            val probe = File(root, ".dash_sdk_write_probe")
+            probe.delete()
+            val created = probe.createNewFile()
+            probe.delete()
+            created && root.isDirectory
```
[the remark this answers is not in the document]
bfoss765: Fixed in 16c4afb9e6 (same finding as thepastaclaw's thread on this hunk). The fixed `.dash_sdk_write_probe` name is gone: the probe is created atomically with a unique name via `File.createTempFile` in the target directory, and only the file this invocation created is deleted — a pre-existing caller-owned entry is never destroyed, and concurrent probes no longer race. Regression tests: `shouldNotDeleteACallerOwnedEntryNamedLikeTheProbe`, `shouldLeaveNoProbeResidueBehind`.

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/security/KeystoreManager.kt
bfoss765: Fixed in 16c4afb9e6. Classification is now gated on `alias == MASTER_ALIAS` inside `rethrowClassifyingDeviceLockedDenial` itself — every other alias rethrows the original exception unchanged, so the restriction is enforced in the mapper rather than left to call-site discipline (the generic AES branches of encrypt/decrypt do accept arbitrary aliases, as you note). Regression tests added: `shouldRethrowAuthGatedAliasUserNotAuthenticatedUnclassified` pins that `KEYS_ALIAS_AUTH_GATED`'s `UserNotAuthenticatedException` passes through untouched (assertSame) even while the device reports locked — preserving the BiometricGate prompt-and-retry contract — and `shouldRethrowCustomAliasDenialUnclassified` covers a host-provisioned custom alias. KDocs on the mapper and on `isDeviceLockedKeystoreDenial` updated to state the MASTER_ALIAS-only guarantee.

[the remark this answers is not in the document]
bfoss765: Fixed in 16c4afb9e6 (same finding as thepastaclaw's thread on this hunk), taking the proposed `alias == MASTER_ALIAS &&` gate in the mapper: only MASTER_ALIAS — whose contract guarantees no `setUserAuthenticationRequired` gate — classifies to `KeystoreDeviceLockedException`; any other alias rethrows the original exception unchanged, keeping an auth-gated custom alias's `UserNotAuthenticatedException` available for the caller's prompt-and-retry path. Regression tests: `shouldRethrowAuthGatedAliasUserNotAuthenticatedUnclassified` (auth-gated identity alias) and `shouldRethrowCustomAliasDenialUnclassified` (host-provisioned alias).

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/Sdk.kt
bfoss765: Addressed in 16c4afb9e6, taking your second option — validating every destination the native installer requires — since the native boundary itself is out of scope for this Kotlin-side PR. The pre-native gate now probes the full fixed destination set `platform_wallet_enable_file_logging` opens (`NATIVE_LOG_DESTINATIONS`: dash_sdk/{run,metrics}.log, platform_wallet/{run,metrics}.log, dash_spv/run.log, key_wallet/run.log, grpc/run.log, build_info.txt — with a keep-in-sync pointer at rs-platform-wallet-ffi/src/logging.rs): parent directories are pre-created exactly as the native `create_dir_all` would, each probed with a unique-name create-and-delete, and an existing entry at a destination path must be a writable regular file (a directory squatting on run.log fails the native append-open, so it fails the probe too). Your example — a regular file at $sessionRoot/dash_sdk — now returns SESSION_ROOT_UNWRITABLE naming that exact path instead of ALREADY_SET; pinned by `shouldReportABlockedFixedLogDestinationWithoutTouchingTheNativeInstaller`, `shouldNameTheBlockedDestinationNotJustTheRoot`, and `shouldRejectADirectorySquattingOnADestinationFile`. The docs are also now honest about the residual limit: ALREADY_SET is a diagnosis by elimination over a bare native boolean, so an I/O failure racing in between the probe and the install — or a destination a future native version adds — would still be misattributed. A structured I/O-vs-subscriber result from the native boundary remains the complete fix; happy to file it as a follow-up on the FFI side.

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/Sdk.kt
bfoss765: Fixed in 16c4afb9e6, taking your second option (the native console installer's signature is out of scope for this Kotlin-side PR). `enableLogging` no longer records itself as the subscriber CLAIMANT: it now writes a separate `subscriberAttemptedBy`, documented as an attempt only — the native side deliberately swallows a lost `try_init()` race, so the call proves the console path ran, not that it won the slot. `subscriberClaimedBy` is reserved for the one verifiable case: a successful `installFileLogging`, whose native call does report whether THIS call installed the subscriber. The ALREADY_SET warning now asserts a claimant only in that verifiable case; after a console attempt it says "most likely by console logging (enableLogging) — recorded as an attempt only (its native installer cannot report whether it won the slot), so an earlier claim from outside this API is also possible", and otherwise falls back to the outside-this-API wording. Making `dash_sdk_enable_logging` return `try_init()`'s outcome would upgrade the attempt to a real claim — worth a native-side follow-up.

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/PlatformWalletManager.kt
```diff
@@ -743,13 +743,29 @@ class PlatformWalletManager(
...
     ): ManagedPlatformWallet = withContext(Dispatchers.IO) {
+        // Fail-fast pre-check BEFORE the native create: on a genuinely
+        // locked device (KeyguardManager.isDeviceLocked) the storeMnemonic
+        // step below is guaranteed to be denied by the lock-bound
+        // MASTER_ALIAS key (setUnlockedDeviceRequired), which would force
+        // the full rollback dance. Failing here instead means no native
+        // wallet was created, no Room rows were written, and there is
+        // nothing to roll back — the typed KeystoreDeviceLockedException
+        // tells the caller to retry after unlock.
+        walletStorage.ensureDeviceUnlocked(operation = "createWallet")
```
thepastaclaw: **🔴 Blocking: Check the master key's actual lock binding before failing fast**

This pre-check rejects wallet creation whenever the device is currently locked, but an existing `MASTER_ALIAS` key is not necessarily lock-bound. `generateWithLockScreenDegradation` intentionally creates that key without `setUnlockedDeviceRequired(true)` when no secure lock screen exists, and existing keys are never regenerated. If the user later adds a PIN and invokes `createWallet` while locked, this check throws even though encryption with the existing unbound key would succeed. Base the fail-fast decision on the existing key's effective policy, or perform a preflight master-key encryption before native wallet creation so the Keystore itself determines whether the operation is denied.

<sub>source: ['codex']</sub>

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/security/WalletStorage.kt
```diff
@@ -189,11 +201,73 @@ class WalletStorage(
...
+                    TAG,
+                    "storeMnemonic: Keystore denied encrypt as device-locked but " +
+                        "KeyguardManager reports UNLOCKED (${e.lockState}) — the false-locked " +
+                        "Keystore2 defect; retry $attempt/" +
+                        "${DEVICE_FALSE_LOCKED_RETRY_DELAYS_MS.size} in ${delayMs}ms",
+                    e,
+                )
+                delay(delayMs)
+            }
+        }
```
thepastaclaw, line 319: **🟡 Suggestion: Scrub the mnemonic retry buffer on every exit**

The retry path now retains a separate plaintext `ByteArray` containing the complete mnemonic across the backoff schedule, but never clears it. Success, a final Keystore denial, and coroutine cancellation during `delay` all leave this avoidable mutable copy intact until garbage collection. Wrap the retry loop in `try/finally` and clear the buffer, matching this class's existing handling of other raw secret arrays.

```suggestion
suspend fun storeMnemonic(walletId: ByteArray, mnemonic: String) {
    val plaintext = mnemonic.encodeToByteArray()
    try {
        var attempt = 0
        while (true) {
            try {
                val blob = keystore.encrypt(plaintext)
                store.edit { it[mnemonicKey(walletId)] = encode(blob) }
                return
            } catch (e: KeystoreDeviceLockedException) {
                if (e.deviceReportsLocked || attempt >= DEVICE_FALSE_LOCKED_RETRY_DELAYS_MS.size) {
                    throw e
                }
                val delayMs = DEVICE_FALSE_LOCKED_RETRY_DELAYS_MS[attempt]
                attempt++
                Log.w(
                    TAG,
                    "storeMnemonic: Keystore denied encrypt as device-locked but " +
                        "KeyguardManager reports UNLOCKED (${e.lockState}) — the false-locked " +
                        "Keystore2 defect; retry $attempt/" +
                        "${DEVICE_FALSE_LOCKED_RETRY_DELAYS_MS.size} in ${delayMs}ms",
                    e,
                )
                delay(delayMs)
            }
        }
    } finally {
        plaintext.fill(0)
    }
}
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The latest commit correctly fixes the caller-owned probe deletion, master-alias classification, and unverified console-claimant issues. One blocking regression remains: the unconditional lock-state pre-check rejects creation when the existing master key was generated without lock binding; the logging result can also still misclassify native write failures, and the retry buffer should be scrubbed.

Source: Codex reviewer backend gpt-5.6-sol (general and security-auditor); CodeRabbit backend model undisclosed; final verifier backend Claude Agent SDK (exact model ID not exposed by the runtime); orchestration-only openclaw-agent/cliproxy/gpt-5.6-sol (not reviewer evidence).

### Review provenance

<details>

```

In `packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/PlatformWalletManager.kt`:
- [BLOCKING] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/PlatformWalletManager.kt:760-768: Check the master key's actual lock binding before failing fast
  This pre-check rejects wallet creation whenever the device is currently locked, but an existing `MASTER_ALIAS` key is not necessarily lock-bound. `generateWithLockScreenDegradation` intentionally creates that key without `setUnlockedDeviceRequired(true)` when no secure lock screen exists, and existing keys are never regenerated. If the user later adds a PIN and invokes `createWallet` while locked, this check throws even though encryption with the existing unbound key would succeed. Base the fail-fast decision on the existing key's effective policy, or perform a preflight master-key encryption before native wallet creation so the Keystore itself determines whether the operation is denied.

In `packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/security/WalletStorage.kt`:
- [SUGGESTION] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/security/WalletStorage.kt:246-270: Scrub the mnemonic retry buffer on every exit
  The retry path now retains a separate plaintext `ByteArray` containing the complete mnemonic across the backoff schedule, but never clears it. Success, a final Keystore denial, and coroutine cancellation during `delay` all leave this avoidable mutable copy intact until garbage collection. Wrap the retry loop in `try/finally` and clear the buffer, matching this class's existing handling of other raw secret arrays.

In `packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/Sdk.kt`:
- [SUGGESTION] packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/Sdk.kt:330-353: A writable root does not prove native failure means ALREADY_SET
  (existing thread: https://github.com/dashpay/platform/pull/4463#discussion_r3839571621)
  The expanded probe now catches blocked child paths, but it still cannot establish that every native I/O operation will succeed before interpreting `false` as `ALREADY_SET`. The probe creates zero-length temporary files and checks existing destinations with `canWrite`, while the native installer writes non-empty build metadata using `fs::write(build_info.txt)`. A full filesystem or app quota can therefore permit the probes while rejecting the native metadata write, causing `false` with no subscriber installed to be reported as `ALREADY_SET`. Access checks can likewise differ from the native truncate/write operation. Return a structured native result that distinguishes I/O failure from `try_init` failure instead of inferring the category from a preflight probe.
```
</details>

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/security/KeystoreManager.kt
thepastaclaw: Resolved in `16c4afb` — _Restrict device-lock classification to the master alias_ no longer present.

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/Sdk.kt
thepastaclaw: Resolved in `16c4afb` — _enableLogging records a claimant even when native installation fails_ no longer present.

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/Sdk.kt
thepastaclaw: Resolved in `16c4afb` — _The writability probe can delete a caller-owned entry_ no longer present.

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/wallet/PlatformWalletManager.kt
bfoss765: Fixed in 72d2800771. Valid — an existing MASTER_ALIAS key generated by `generateWithLockScreenDegradation` before a PIN was enrolled carries no `setUnlockedDeviceRequired` binding, so the KeyguardManager-only check over-rejected. Took the preflight-encryption option so the Keystore itself renders the verdict: when the sampled state says locked, the pre-check (renamed `ensureMasterKeyNotLockBlocked` to match its real contract) performs one master-alias encrypt of a non-secret probe byte (ciphertext discarded, nothing persisted, never prompts — the master alias has no auth gate). A denial classified device-locked by the existing MASTER_ALIAS-gated mapper proves the key is lock-bound → rethrown as the typed retryable exception re-labeled `operation="createWallet"` with the classified denial as cause; a successful probe proves the key is unbound → creation proceeds. Unclassified probe failures log and proceed — best-effort, the real store renders the final verdict on the pre-existing rollback path. No key rotation, no new aliases, no parameter changes; a fresh-install probe provisions the key exactly as the first `storeMnemonic` would have (and a lock-bound fresh key on a locked device still fails fast typed, since generation is not usage-gated). Red-then-green: `shouldPassPreCheckWhenDeviceIsLockedButMasterKeyIsNotLockBound` (the exact scenario here) failed at 16c4afb9e6 and passes now; `shouldFailFastWhenDeviceIsGenuinelyLocked` updated to pin the probe-derived verdict (exactly one probe encrypt, classified cause) — its old zero-Keystore-calls assertion was invalidated by design. 337 module tests green.

packages/kotlin-sdk/sdk/src/main/kotlin/org/dashfoundation/dashsdk/security/WalletStorage.kt
bfoss765: Fixed in 72d2800771 — retry loop wrapped in `try/finally { plaintext.fill(0) }` per the suggestion (adapted to the current revision), covering success, the final denial propagating, and coroutine cancellation during a backoff delay. All three exits pinned by tests: `shouldScrubMnemonicBufferAfterSuccessfulStore`, `shouldScrubMnemonicBufferWhenFinalDenialPropagates`, `shouldScrubMnemonicBufferWhenCancelledDuringRetryBackoff` (cancellation was deterministically testable: `cancelAndJoin` during the backoff delay runs the finally before join returns; the fake captures the exact buffer reference). All failed at 16c4afb9e6, all green now.

shumkov (approved):

PR #4467, opened 2026-08-24
feat(dpp)!: rebalance the shielded fee constants for protocol 14

QuantumExplorer:
## Issue being fixed or feature implemented

The flat shielded fee's components are not independently sufficient. `compute_minimum_shielded_fee` reserves 100M credits for Halo 2 proof verification and 22M/action for per-action compute, but allots only 344 bytes/action (9.43M credits) of storage — while the GROVE_V4 fixed per-append model (grovedb #828–#830) meters a 1-action transfer at **17,882,707 credits**, 14,337,000 of it storage (= **523 bytes** at the full 27,400 credits/byte rate: the declared 344-byte payload plus Merk framing, dense path records and the amortized chunk framing). The compute budget silently subsidizes database work — the fee-floor blocker raised on #4382.

## What was done?

Protocol 14 only (`DriveAbciValidationConstants` v10 is protocol-14-only; protocol ≤13 values are locked and byte-identical):

- **`shielded_proof_verification_fee` 100M → 40M** — ~5 ms of Halo 2 at the fee model's ~8M credits/ms of CPU (~27× a BLS signature verification), reserved for compute alone.
- **New table field `shielded_storage_bytes_per_action`** — 344 (the declared physical payload) locked in v1–v9, **550** at v10: the measured GROVE_V4 footprint with headroom, so the storage component alone pays for the database work.
- `compute_minimum_shielded_fee_v0` reads the allowance from the table (the bare `SHIELDED_STORAGE_BYTES_PER_ACTION` const is retired); no method version bump — logic unchanged, released versions replay byte-for-byte.
- Wallet-FFI fee goldens updated; the mainnet-halt-repro test pins itself to protocol 13, whose constants its measured funding band belongs to.

Net flat fees at protocol 14 (2 actions): transfer 162.85M → **114.14M** (−30%), unshield → 120.22M, withdrawal → 226.48M.

## Sequencing

Lands **before** #4382 (the grovedb pin bump). #4382 then rebases onto this and strengthens its fee-floor tests to assert the compute and database components independently against the measured fixed per-append cost — closing the blocker thread there with these constants.

## Breaking Changes

Protocol 14 shielded fees change (protocol 14 is unreleased). Protocol ≤13 is byte-for-byte untouched.

## Checklist:
- [x] I have performed a self-review of my own code
- [x] I have added or updated relevant unit/integration/functional/e2e tests

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

[...]

PR #4470, opened 2026-08-24
fix(platform-wallet): estimate shielded fees at the network's active protocol version

QuantumExplorer:
`platform_wallet_shielded_estimate_fee` pinned `PlatformVersion::latest()` and took no manager handle, so once a client ships the protocol-14 fee rebalance (#4467) its fee preview under-quotes any network still running protocol 13: 114,140,000 vs the 162,851,200 credits the consensus gate actually validates for a 2-action transfer. The shielded builders don't have this problem — they carve fees at the manager's network-tracked `sdk.version()`. Raised by review on #4467 (resolved there as a follow-up).

## What was done?
- `platform_wallet_shielded_estimate_fee` now takes the manager `Handle` and resolves the fee formula's version through the manager's SDK (`sdk().version()`), exactly like the builders. An unknown handle is a hard `ErrorInvalidHandle` — a versionless fallback would silently mis-quote. The kind → formula mapping is factored into a version-explicit selector so the boundary is unit-testable.
- Swift: `estimateShieldedFee` becomes an instance method on `PlatformWalletManager` (it needs the handle). SwiftExampleApp's `SendViewModel` reads view-pushed estimates (same pattern as `platformMinOutputAmount` — the view model holds no wallet handle).
- Kotlin/JNI: the JNI export takes the manager handle; the wrapper moves from the process-global `ShieldedProver` onto `PlatformWalletManager`, matching the other manager-handle entry points. Both example-app screens now estimate through the active manager.
- Boundary tests pin the formula table at protocol 13 (162,851,200 / 168,934,000 / 275,191,200 credits at 2 actions) and protocol 14 (114,140,000 / 120,222,800 / 226,480,000), plus hard-error tests for unknown kind and unknown handle.

## How Has This Been Tested?
- `cargo test -p platform-wallet-ffi --all-features shielded_send` — 14 pass, including both protocol-boundary pins.
- `cargo clippy -p platform-wallet-ffi -p rs-unified-sdk-jni --all-features --all-targets` — clean.
- Kotlin: `:sdk:compileDebugKotlin` and `:app:compileDebugKotlin` — clean.
- Swift: `SwiftDashSDK` builds for the iOS simulator against the regenerated `platform-wallet-ffi` header; SwiftExampleApp sources compile (final link needs the rebuilt xcframework, as usual for FFI signature changes).

## Breaking Changes
None on-chain. Host-facing FFI/SDK signature change: `platform_wallet_shielded_estimate_fee` gains a leading manager-handle parameter; the Swift method is now an instance method and the Kotlin wrapper moved to `PlatformWalletManager`.

## Checklist:

## Summary by CodeRabbit

* **New Features**
  * Shielded transaction fee estimates now reflect the wallet’s tracked network version.
  * Added wallet-manager-based fee estimation for Kotlin and Swift integrations.
  * Send transaction screens resolve shielded fees automatically for transfers, unshielding, and withdrawals.
* **Bug Fixes**
  * Improved fee accuracy across supported platform versions.
  * Fees now refresh when the network protocol version changes.
  * Added safer handling for unavailable estimates, invalid wallet configurations, invalid parameters, and estimation failures.
  * Transaction screens fall back to available estimates when dynamic values cannot be resolved.

### DIFF
--- a/packages/rs-platform-wallet-ffi/src/shielded_send.rs
+++ b/packages/rs-platform-wallet-ffi/src/shielded_send.rs
@@ -156,6 +156,24 @@ pub unsafe extern "C" fn platform_wallet_shielded_prover_is_ready() -> bool {
     CachedOrchardProver::new().is_ready()
 }
 
+/// Map a fee `kind` byte to its consensus fee formula, or `None` for an
+/// unknown kind. All three formulas share the `(num_actions, version) →
+/// credits` shape, so the selection is version-independent and testable
+/// against any explicit [`PlatformVersion`].
+#[allow(clippy::type_complexity)]
+fn shielded_fee_formula(
+    kind: u8,
+) -> Option<
+    fn(usize, &dpp::version::PlatformVersion) -> Result<dpp::fee::Credits, dpp::ProtocolError>,
+> {
+    match kind {
+        0 => Some(compute_minimum_shielded_fee),
+        1 => Some(compute_shielded_unshield_fee),
+        2 => Some(compute_shielded_withdrawal_fee),
+        _ => None,
+    }
+}
+
 /// Estimate the consensus-pinned flat shielded fee (in credits) for a
 /// pool-paid shielded transition.
 ///
@@ -169,40 +187,47 @@ pub unsafe extern "C" fn platform_wallet_shielded_prover_is_ready() -> bool {
 ///   the flat Core withdrawal-document cost).
 ///
 /// `num_actions` is the Orchard action count of the bundle the host will
-/// build (a single-note spend with change is 2 actions). The version is
-/// pinned to [`PlatformVersion::latest()`] — the same version the shielded
-/// builders in `platform-wallet` resolve via `sdk.version()`, so the
-/// estimate can't drift from the fee the builder carves and the consensus
-/// gate validates.
+/// build (a single-note spend with change is 2 actions). The fee is
+/// computed at `handle`'s manager's network-tracked platform version
+/// (`sdk.version()`) — the same version the shielded builders in
+/// `platform-wallet` resolve — so the estimate can't drift from the fee
+/// the builder carves and the consensus gate validates, even when the
+/// connected network hasn't activated the client's latest protocol
+/// version yet.
 ///
-/// Pure computation: no wallet handle, no network. Writes the fee to
-/// `out_fee` and returns `ok()`. An unknown `kind` returns
-/// `ErrorInvalidParameter`; a fee-formula overflow returns
-/// `ErrorArithmeticOverflow`.
+/// No network round-trip and no wallet resolution — just the handle →
+/// version lookup and a pure computation. Writes the fee to `out_fee` and
+/// returns `ok()`. An unknown `kind` returns `ErrorInvalidParameter` (and
+/// is checked before the handle, so it fails the same way regardless of
+/// handle validity); an unknown `handle` returns `ErrorInvalidHandle`; a
+/// fee-formula overflow returns `ErrorArithmeticOverflow`.
 ///
 /// # Safety
 /// `out_fee` must point to 8 writable bytes (a `u64`).
 #[no_mangle]
 pub unsafe extern "C" fn platform_wallet_shielded_estimate_fee(
+    handle: Handle,
     kind: u8,
     num_actions: usize,
     out_fee: *mut u64,
 ) -> PlatformWalletFFIResult {
     check_ptr!(out_fee);
 
-    let platform_version = dpp::version::PlatformVersion::latest();
-    let fee = match kind {
-        0 => compute_minimum_shielded_fee(num_actions, platform_version),
-        1 => compute_shielded_unshield_fee(num_actions, platform_version),
-        2 => compute_shielded_withdrawal_fee(num_actions, platform_version),
-        other => {
-            return PlatformWalletFFIResult::err(
-                PlatformWalletFFIResultCode::ErrorInvalidParameter,
-                format!("unknown shielded fee kind {other} (expected 0/1/2)"),
-            );
-        }
+    let Some(formula) = shielded_fee_formula(kind) else {
+        return PlatformWalletFFIResult::err(
+            PlatformWalletFFIResultCode::ErrorInvalidParameter,
+            format!("unknown shielded fee kind {kind} (expected 0/1/2)"),
+        );
     };
-    match fee {
+    let Some(platform_version) =
+        PLATFORM_WALLET_MANAGER_STORAGE.with_item(handle, |manager| manager.sdk().version())
+    else {
+        return PlatformWalletFFIResult::err(
+            PlatformWalletFFIResultCode::ErrorInvalidHandle,
+            format!("invalid manager handle: {handle}"),
+        );
+    };
+    match formula(num_actions, platform_version) {
         Ok(credits) => {
             *out_fee = credits;
             PlatformWalletFFIResult::ok()
@@ -1637,47 +1662,90 @@ mod tests {
         );
     }
 
-    /// Pin the fee estimator to the on-chain ground-truth values observed at the current platform
-    /// version with 2 actions (single-note spend + change). These are the exact credits the
-    /// builder carves and the consensus gate validates, so the host's "Estimated Fee" must match.
+    /// Resolve the 2-action fee for `kind` at an explicit protocol version
+    /// through the same formula table the FFI dispatches on.
+    fn estimate_at(kind: u8, protocol_version: u32) -> u64 {
+        let version = dpp::version::PlatformVersion::get(protocol_version)
+            .expect("protocol version must exist");
+        shielded_fee_formula(kind).expect("known kind")(2, version)
+            .expect("fee formula must not overflow at 2 actions")
+    }
+
+    /// Pin the fee estimator to the on-chain ground-truth values observed at protocol 13 (the
+    /// released fee constants) with 2 actions (single-note spend + change). The estimator resolves
+    /// the version from the manager's network-tracked `sdk.version()`, so on a protocol-13 network
+    /// these are the exact credits the builder carves and the consensus gate validates.
     #[test]
-    fn estimate_fee_matches_observed_onchain_values_for_2_actions() {
-        unsafe {
-            let estimate = |kind: u8| {
-                let mut fee: u64 = 0;
-                let result = platform_wallet_shielded_estimate_fee(kind, 2, &mut fee);
-                assert_eq!(
-                    result.code,
-                    PlatformWalletFFIResultCode::Success,
-                    "kind {kind} must succeed"
-                );
-                fee
-            };
-            // kind 0 — ShieldedTransfer / Shield base.
-            assert_eq!(
-                estimate(0),
-                114_140_000,
-                "shielded transfer fee (2 actions)"
-            );
-            // kind 1 — Unshield.
-            assert_eq!(estimate(1), 120_222_800, "unshield fee (2 actions)");
-            // kind 2 — ShieldedWithdrawal.
-            assert_eq!(
-                estimate(2),
-                226_480_000,
-                "shielded withdrawal fee (2 actions)"
-            );
-        }
+    fn estimate_fee_matches_observed_onchain_values_at_protocol_13() {
+        // kind 0 — ShieldedTransfer / Shield base.
+        assert_eq!(
+            estimate_at(0, 13),
+            162_851_200,
+            "shielded transfer fee (2 actions, protocol 13)"
+        );
+        // kind 1 — Unshield.
+        assert_eq!(
+            estimate_at(1, 13),
+            168_934_000,
+            "unshield fee (2 actions, protocol 13)"
+        );
+        // kind 2 — ShieldedWithdrawal.
+        assert_eq!(
+            estimate_at(2, 13),
+            275_191_200,
+            "shielded withdrawal fee (2 actions, protocol 13)"
+        );
+    }
+
+    /// The protocol-14 side of the boundary: the rebalanced constants
+    /// (40M proof verification + 550 storage bytes/action). A network that
+    /// has activated protocol 14 must quote these, and a network still on
+    /// protocol 13 must NOT — the pre-fix estimator pinned
+    /// `PlatformVersion::latest()` and silently under-quoted protocol-13
+    /// networks by ~30%.
+    #[test]
+    fn estimate_fee_matches_rebalanced_values_at_protocol_14() {
+        assert_eq!(
+            estimate_at(0, 14),
+            114_140_000,
+            "shielded transfer fee (2 actions, protocol 14)"
+        );
+        assert_eq!(
+            estimate_at(1, 14),
+            120_222_800,
+            "unshield fee (2 actions, protocol 14)"
+        );
+        assert_eq!(
+            estimate_at(2, 14),
+            226_480_000,
+            "shielded withdrawal fee (2 actions, protocol 14)"
+        );
     }
 
     #[test]
     fn estimate_fee_rejects_unknown_kind() {
         unsafe {
             let mut fee: u64 = 0;
-            let result = platform_wallet_shielded_estimate_fee(7, 2, &mut fee);
+            // The kind check runs before handle resolution, so a bogus kind
+            // fails identically with or without a live manager handle.
+            let result = platform_wallet_shielded_estimate_fee(0, 7, 2, &mut fee);
             assert_eq!(
                 result.code,
                 PlatformWalletFFIResultCode::ErrorInvalidParameter
+            );
+        }
+    }
+
+    #[test]
+    fn estimate_fee_rejects_unknown_manager_handle() {
+        unsafe {
+            let mut fee: u64 = 0;
+            let result = platform_wallet_shielded_estimate_fee(0, 0, 2, &mut fee);
+            assert_eq!(
+                result.code,
+                PlatformWalletFFIResultCode::ErrorInvalidHandle,
+                "a versionless fallback here would silently mis-quote — an \
+                 unknown handle must be a hard error"
             );
         }
     }
--- a/packages/rs-platform-wallet-ffi/src/shielded_send.rs
+++ b/packages/rs-platform-wallet-ffi/src/shielded_send.rs
@@ -1722,6 +1722,85 @@ mod tests {
         );
     }
 
+    /// The heart of the fix, exercised end-to-end through the exported
+    /// estimator: the version is resolved from the manager handle's
+    /// network-tracked `sdk.version()`, so managers pinned to protocol 13
+    /// and protocol 14 quote different fees through the SAME entry point.
+    /// Reverting the lookup to `PlatformVersion::latest()` fails the
+    /// protocol-13 half of this test.
+    #[test]
+    fn estimate_fee_resolves_version_through_manager_handle() {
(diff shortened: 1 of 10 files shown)


packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerShieldedSync.swift
```diff
@@ -465,18 +465,26 @@ extension PlatformWalletManager {
...
+            throw PlatformWalletError.invalidHandle(
+                "PlatformWalletManager not configured"
+            )
+        }
         var fee: UInt64 = 0
         // `num_actions` is `usize` on the Rust side → imported as `UInt`.
         try platform_wallet_shielded_estimate_fee(
+            handle,
             kind.rawValue,
             UInt(numActions),
```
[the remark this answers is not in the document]
QuantumExplorer: Added a `numActions >= 0` guard that throws `PlatformWalletError.invalidParameter` before the `UInt` conversion (92f2505d96), matching the JNI bridge's sign check.

packages/rs-platform-wallet-ffi/src/shielded_send.rs
```diff
@@ -169,40 +187,47 @@ pub unsafe extern "C" fn platform_wallet_shielded_prover_is_ready() -> bool {
...
 ///
 /// # Safety
 /// `out_fee` must point to 8 writable bytes (a `u64`).
 #[no_mangle]
 pub unsafe extern "C" fn platform_wallet_shielded_estimate_fee(
+    handle: Handle,
     kind: u8,
     num_actions: usize,
     out_fee: *mut u64,
 ) -> PlatformWalletFFIResult {
```
thepastaclaw, line 213: **🔴 Blocking: Preserve the released estimator ABI when adding manager context**

`platform_wallet_shielded_estimate_fee` was released in v4.0.0 with `(kind, num_actions, out_fee)`, but this changes the same exported symbol to `(handle, kind, num_actions, out_fee)`. A new wrapper loading an older library therefore passes `num_actions` where the old function expects `out_fee`, potentially writing through a small integer address; an older wrapper loading the new library similarly supplies shifted or undefined arguments. The JNI export at `packages/rs-unified-sdk-jni/src/funding.rs:98-104` has the same problem: its released v4.1.0 short JNI symbol is unchanged while a `jlong` is inserted into the descriptor, so the JVM can resolve the symbol despite incompatible arguments. Keep the released C and JNI entry points intact, add manager-aware exports under new names, and route the updated Swift and Kotlin APIs through those new exports.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet-ffi/src/shielded_send.rs
```diff
@@ -1637,47 +1662,90 @@ mod tests {
...
-    /// Pin the fee estimator to the on-chain ground-truth values observed at the current platform
-    /// version with 2 actions (single-note spend + change). These are the exact credits the
-    /// builder carves and the consensus gate validates, so the host's "Estimated Fee" must match.
+    /// Resolve the 2-action fee for `kind` at an explicit protocol version
+    /// through the same formula table the FFI dispatches on.
+    fn estimate_at(kind: u8, protocol_version: u32) -> u64 {
+        let version = dpp::version::PlatformVersion::get(protocol_version)
+            .expect("protocol version must exist");
+        shielded_fee_formula(kind).expect("known kind")(2, version)
+            .expect("fee formula must not overflow at 2 actions")
```
thepastaclaw, line 1671: **🟡 Suggestion: Test version resolution through the public FFI estimator**

The protocol-13 and protocol-14 tests call `shielded_fee_formula` directly with explicit versions. They never create a manager or exercise the successful `platform_wallet_shielded_estimate_fee` path that resolves `manager.sdk().version()`. Consequently, replacing the lookup with `PlatformVersion::latest()` would still leave every added test passing. Create live managers backed by SDKs pinned to protocols 13 and 14, call the exported estimator through each handle, and assert the corresponding fee values.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The estimator now derives fees from the manager SDK version, but it changes two released native ABIs in place, allowing mismatched published wrappers and libraries to invoke the same symbols with shifted arguments. The formula tests also bypass the successful handle-to-version path, leaving the central behavior of this fix unprotected.
Source: Codex general reviewer (`codex-general`, exact backend model ID not supplied), Codex FFI reviewer (`codex-ffi-engineer`, exact backend model ID not supplied), and Claude Agent SDK verifier (exact backend model ID not exposed); orchestration-only: `openclaw-agent/cliproxy/gpt-5.6-sol`.

### Review provenance

<details>

```

- [BLOCKING] packages/rs-platform-wallet-ffi/src/shielded_send.rs:208-213: Preserve the released estimator ABI when adding manager context
  `platform_wallet_shielded_estimate_fee` was released in v4.0.0 with `(kind, num_actions, out_fee)`, but this changes the same exported symbol to `(handle, kind, num_actions, out_fee)`. A new wrapper loading an older library therefore passes `num_actions` where the old function expects `out_fee`, potentially writing through a small integer address; an older wrapper loading the new library similarly supplies shifted or undefined arguments. The JNI export at `packages/rs-unified-sdk-jni/src/funding.rs:98-104` has the same problem: its released v4.1.0 short JNI symbol is unchanged while a `jlong` is inserted into the descriptor, so the JVM can resolve the symbol despite incompatible arguments. Keep the released C and JNI entry points intact, add manager-aware exports under new names, and route the updated Swift and Kotlin APIs through those new exports.
- [SUGGESTION] packages/rs-platform-wallet-ffi/src/shielded_send.rs:1665-1671: Test version resolution through the public FFI estimator
  The protocol-13 and protocol-14 tests call `shielded_fee_formula` directly with explicit versions. They never create a manager or exercise the successful `platform_wallet_shielded_estimate_fee` path that resolves `manager.sdk().version()`. Consequently, replacing the lookup with `PlatformVersion::latest()` would still leave every added test passing. Create live managers backed by SDKs pinned to protocols 13 and 14, call the exported estimator through each handle, and assert the corresponding fee values.
```
</details>

packages/rs-platform-wallet-ffi/src/shielded_send.rs
QuantumExplorer: Done in d9c62e2c38: `estimate_fee_resolves_version_through_manager_handle` creates live mock-SDK managers pinned via `SdkBuilder::with_version` to protocols 13 and 14 and asserts 162,851,200 vs 114,140,000 through the exported estimator — a `latest()` regression now fails the protocol-13 half.

packages/rs-platform-wallet-ffi/src/shielded_send.rs
QuantumExplorer: Not changing the ABI: these FFI crates are internal bridges whose wrappers ship lockstep with the native binary (the Swift package embeds the xcframework and the Kotlin AAR bundles the .so, both built from the same source revision), so a mismatched wrapper/library pair cannot occur in supported packaging. Direct repo precedent: between released v4.0.0 and v4.1.1, `mnemonic_resolver_handle` was inserted mid-signature into four released `platform_wallet_manager_shielded_*` exports the same way. A back-compat shim keeping the old 3-arg symbol would also have to keep quoting `PlatformVersion::latest()` — silently preserving the exact under-quote bug this PR fixes.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Views/SendTransactionView.swift
```diff
@@ -515,6 +520,28 @@ struct SendTransactionView: View {
...
+    /// fee its consensus gate actually validates. A kind that fails to
+    /// resolve is simply absent — `estimateFee(for:)` falls back to the
+    /// static placeholder.
+    private func resolveShieldedFees() {
+        guard viewModel.shieldedFeeEstimates.isEmpty else { return }
+        var fees: [PlatformWalletManager.ShieldedFeeKind: UInt64] = [:]
+        for kind: PlatformWalletManager.ShieldedFeeKind in [.transfer, .unshield, .withdrawal] {
+            fees[kind] = try? walletManager.estimateShieldedFee(kind: kind, numActions: 2)
+        }
+        viewModel.shieldedFeeEstimates = fees
```
thepastaclaw, line 562: **🟡 Suggestion: Recompute cached fees after the SDK protocol version ratchets**

`AppState` publishes a newly created SDK before its detached `refreshProtocolVersion()` finishes. An unpinned SDK initially uses protocol 11 on mainnet or protocol 12 on the other networks, so this method can cache the older 162,851,200-credit transfer fee before a protocol-14 network ratchets the shared SDK to the 114,140,000-credit fee used by builders. `resolveShieldedFees()` runs only on appearance and its nonempty guard prevents replacement, even though `platformState.platformProtocolVersion` publishes refresh completion. The Kotlin producers have the same stale-cache path because `SendTransactionScreen.kt:292-309` keys only on `flow` and `manager`, while `ShieldedFundScreen.kt:82-88` keys only on `manager`. Observe `platformProtocolVersion` in all three screens and replace the estimates whenever it changes so previews remain aligned with the version builders read at submission time.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The manager-handle lookup and new end-to-end test correctly pin protocol-13 and protocol-14 behavior, and the Swift negative-action crash is fixed. The released C/JNI estimator symbols still have incompatible signatures, while the example applications can retain estimates calculated before the asynchronous protocol-version ratchet completes.
Source: Codex general reviewer (backend model ID not supplied), Codex FFI reviewer (backend model ID not supplied), and Claude Agent SDK verifier (backend model ID not exposed); orchestration-only: openclaw-agent/cliproxy/gpt-5.6-sol.

### Review provenance

<details>

```

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Views/SendTransactionView.swift`:
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Views/SendTransactionView.swift:531-537: Recompute cached fees after the SDK protocol version ratchets
  `AppState` publishes a newly created SDK before its detached `refreshProtocolVersion()` finishes. An unpinned SDK initially uses protocol 11 on mainnet or protocol 12 on the other networks, so this method can cache the older 162,851,200-credit transfer fee before a protocol-14 network ratchets the shared SDK to the 114,140,000-credit fee used by builders. `resolveShieldedFees()` runs only on appearance and its nonempty guard prevents replacement, even though `platformState.platformProtocolVersion` publishes refresh completion. The Kotlin producers have the same stale-cache path because `SendTransactionScreen.kt:292-309` keys only on `flow` and `manager`, while `ShieldedFundScreen.kt:82-88` keys only on `manager`. Observe `platformProtocolVersion` in all three screens and replace the estimates whenever it changes so previews remain aligned with the version builders read at submission time.

- [BLOCKING] packages/rs-platform-wallet-ffi/src/shielded_send.rs:208-213: Preserve the released estimator ABI when adding manager context
  (existing thread: https://github.com/dashpay/platform/pull/4470#discussion_r3845329252)
  `platform_wallet_shielded_estimate_fee` shipped in v4.0.0 as `(kind, num_actions, out_fee)`, but the same exported C symbol now expects `(handle, kind, num_actions, out_fee)`. A new caller loading the old library supplies `num_actions` where the library expects `out_fee`, allowing the old implementation to write through a small integer address; the inverse mismatch also invokes the new implementation with shifted and missing arguments. The crate explicitly produces `staticlib` and `cdylib` artifacts, generates a public C header, and documents direct C/C++ use, so lockstep Swift packaging does not make this ABI change safe. The v4.1.0 JNI export at `packages/rs-unified-sdk-jni/src/funding.rs:98-104` likewise retains its short JNI symbol while inserting `manager_handle`, allowing a mismatched Kotlin wrapper and native library to resolve the same symbol with different descriptors. Preserve the released entry points and add manager-aware symbols under new names; updated Swift and Kotlin wrappers can use the new symbols while the legacy estimator retains its previous versionless behavior.
```
</details>

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Views/SendTransactionView.swift
QuantumExplorer: Done in 4e7df3ee6f: Swift re-runs `resolveShieldedFees()` (empty-cache guard dropped — the recompute is a pure handle lookup) via `.onChange(of: platformState.platformProtocolVersion)`, and both Kotlin screens now key their fee `produceState` on the collected `appState.platformProtocolVersion`, so estimates are replaced when the async ratchet completes.

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Views/SendTransactionView.swift
```diff
@@ -346,10 +346,23 @@ struct SendTransactionView: View {
...
             .onChange(of: viewModel.detectedAddressType) { _, _ in
                 autoSelectSource()
             }
+            .onChange(of: platformState.platformProtocolVersion) { _, _ in
+                // The SDK learns the network's protocol version on a
+                // detached task after `AppState` publishes it; estimates
+                // resolved before that ratchet completed were computed at
+                // the seed version. Re-resolve so the preview matches the
+                // version the builders will read at submission time.
+                resolveShieldedFees()
```
thepastaclaw, line 373: **🟡 Suggestion: Refresh estimates after metadata-driven protocol ratchets**

This observer reruns fee estimation only when `AppState.platformProtocolVersion` changes, but that property is written solely by the one startup/network-switch refresh. Rust's `refresh_protocol_version()` catches a failed proven fetch and returns the current seed version, so the property can publish that unchanged seed. Any later proof-verified query can independently ratchet the SDK's shared `AtomicU32` through response metadata without publishing another app-state value. After a transient startup failure—or a protocol activation while the app remains open—the preview therefore remains based on the old version while submission builders read the newer `sdk.version()`. The Kotlin screens have the same gap because their `protocolVersion` key comes from the equivalent one-shot `AppState` flow. Expose metadata-driven SDK version changes through an observable value, retry the refresh, or otherwise trigger estimation whenever the SDK's effective version advances.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The manager-backed estimator and protocol-boundary tests correctly select fees from the manager SDK's active version, and the prior asynchronous startup-cache issue is fixed. The PR still changes released C and JNI symbols in place, creating an unsafe ABI mismatch, and the example applications do not refresh cached estimates after later metadata-driven version ratchets.
Source: Codex general, FFI-engineer, and Rust-quality reviewers (exact backend model IDs not supplied); final verifier: Claude Agent SDK (exact backend model ID not exposed); orchestration-only: openclaw-agent/cliproxy/gpt-5.6-sol.

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — ffi-engineer (completed), `gpt-5.6-sol` — rust-quality (completed)

<details>

```

In `packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Views/SendTransactionView.swift`:
- [SUGGESTION] packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Views/SendTransactionView.swift:358-364: Refresh estimates after metadata-driven protocol ratchets
  This observer reruns fee estimation only when `AppState.platformProtocolVersion` changes, but that property is written solely by the one startup/network-switch refresh. Rust's `refresh_protocol_version()` catches a failed proven fetch and returns the current seed version, so the property can publish that unchanged seed. Any later proof-verified query can independently ratchet the SDK's shared `AtomicU32` through response metadata without publishing another app-state value. After a transient startup failure—or a protocol activation while the app remains open—the preview therefore remains based on the old version while submission builders read the newer `sdk.version()`. The Kotlin screens have the same gap because their `protocolVersion` key comes from the equivalent one-shot `AppState` flow. Expose metadata-driven SDK version changes through an observable value, retry the refresh, or otherwise trigger estimation whenever the SDK's effective version advances.

- [BLOCKING] packages/rs-platform-wallet-ffi/src/shielded_send.rs:208-213: Preserve the released estimator ABI when adding manager context
  (existing thread: https://github.com/dashpay/platform/pull/4470#discussion_r3845329252)
  The v4.0.0 and v4.1.1 releases export `platform_wallet_shielded_estimate_fee(kind, num_actions, out_fee)`, while this PR changes that same unmangled symbol to `(handle, kind, num_actions, out_fee)`. The crate ships `staticlib` and `cdylib` artifacts, generates a C header, and explicitly documents direct C/C++ integration, so independently compiled callers can resolve this symbol while using the old calling convention. For example, a new caller loading the old library makes the old implementation interpret `num_actions` as `out_fee`, potentially writing through a small integer address; the inverse mismatch also invokes Rust with shifted or missing arguments. The released short JNI symbol at `packages/rs-unified-sdk-jni/src/funding.rs:98-104` is changed similarly by inserting `manager_handle`, and JNI short-name lookup does not encode the descriptor. Preserve the released C and JNI entry points with their original signatures, add separately named manager-aware exports, and route the current Swift and Kotlin wrappers through those new symbols. The legacy estimator can retain its previous versionless behavior for binary compatibility and be deprecated.
```
</details>

packages/swift-sdk/SwiftExampleApp/SwiftExampleApp/Core/Views/SendTransactionView.swift
QuantumExplorer: Fixed in 6e95fe7c14 — no new FFI export; the estimator already resolves the live version on every call.

`platform_wallet_shielded_estimate_fee` reads `manager.sdk().version()` per invocation with no round-trip and no caching (`rs-platform-wallet-ffi/src/shielded_send.rs`), so the apps do not need an observable for the SDK version — they only need to re-read at the points where the value is consumed. `AppState.platformProtocolVersion` stays as the cheap early trigger; it is no longer the only one.

**Swift** (`SendTransactionView.swift`): `resolveShieldedFees()` now also runs in the Send action, synchronously before the `Task` that calls `executeSend`. That read is the one taken closest to submission, so a preview built at the seed version cannot survive into a send whose builders carve at a ratcheted `sdk.version()`. Covers both paths you flagged: the swallowed proven-fetch failure that republishes the unchanged seed, and an activation landing while the sheet stays open.

**Kotlin** (`SendTransactionScreen.kt`, `ShieldedFundScreen.kt`): the fee is display-only on these screens (`canSend` never reads it), so a submit-time re-read would only refresh a value the user is navigating away from. Instead both fee producers gained a resume epoch key — `LifecycleResumeEffect` bumps it, so the estimate is re-read on navigation to the screen and on app foreground, which is where a stale number is actually visible. `LifecycleResumeEffect` comes from `lifecycle-runtime-compose`, the same artifact already supplying `collectAsStateWithLifecycle`.

Verification: Kotlin and Swift were both verified by reading rather than compiling. The Swift build needs `DashSDKFFI.xcframework`, which is not present in this worktree (a 10-20 min rebuild); the Gradle compile needs an Android SDK location that is not configured here. I did confirm the two new API usages against the actual cached artifacts rather than from memory — `LifecycleResumeEffect(key1, …)` with the `onPauseOrDispose` scope exists in `lifecycle-runtime-compose-android` 2.9.0, and `produceState`'s vararg-keys overload exists in `compose-runtime` 1.8.1, which is what the four-key call now resolves to.

PR #4312, opened 2026-08-06
feat(platform-wallet): multi-output shielded transfers + output-aware fee predictor

bfoss765:
Continues #4301 — moved from a fork branch to an in-repo branch (rebased onto v4.2-dev post-#4305) so maintainers can push changes directly, per review request. Full review history on #4301.

---

## What this closes

A shielded invite is funded as **one** note. When that note is later spent to
claim the invite, Orchard's `BundleType::DEFAULT` pads the single-spend bundle
up to `MIN_ACTIONS = 2`, and the padding action's dummy nullifier is **randomly
generated** (`orchard` `builder.rs:37,76-99`; `note.rs:227-243`;
`nullifier.rs:53-55`).

The new identity's id is `double_sha256(sorted PUBLISHED nullifiers)` —
`identity_id_from_nullifiers`, computed over *every* action's nullifier,
padding included, because consensus re-derives it the same way and dummies are
indistinguishable by design.

So with one real spend the claim's identity id contains fresh randomness. It
cannot be predicted before the build, and — the part that actually hurts — it
cannot be **re-derived on a retry**: a second attempt builds a different dummy
and therefore a different id. Any idempotent claim-recovery step that asks "did
my earlier attempt already create this identity?" has no expected id to compare
against and must fail closed.

With **two or more real spends** no padding action is added, every published
nullifier is the deterministic nullifier of a real note, and the id is a pure
function of the spent note set — predictable and reproducible.

## The fix: fund with two notes, not one

Fund the one-time address with **two notes that each hold less than the
target**, in one atomic transfer — `D` split as `floor(D/2) + ceil(D/2)`.

Note selection is greedy largest-first and breaks as soon as the accumulated
value covers the target (`note_selection.rs:117-135`). Neither half covers `D`
on its own, so **both are structurally forced into the claim bundle**. There is
no heuristic to tune and no way for the selector to pick just one.

This is why it is two sub-target notes rather than "a main note plus a small
anchor": with a main note that already covers the target, the selector would
stop after it and the padding action would come straight back.

**The claim path needs no change.** It already branches on
`selected_notes.len() >= 2` when deciding whether an expected identity id can
be derived. This PR changes the note *layout* so that branch is always taken;
the recovery logic itself is untouched.

## The fee predictor — why it must ship in the same PR

`build_shielded_transfer_transition` sized its fee from
`spends.len().max(2)`, **ignoring the output count**.

An Orchard action is a joined spend/output slot: the on-wire action count is
[...]

PR #4472, opened 2026-08-24
feat(platform-wallet): shield Platform credits to an external Orchard recipient

QuantumExplorer:
Hosts cannot pay a third-party shielded address from Platform Payment credits in one transition: the Type 15 shield builder (`build_shield_transition`) has always taken an arbitrary recipient, but the wallet layer pinned it to the account's own default Orchard address. dashwallet-ios surfaces this as a dead-end error ("This address can't be paid from your Platform balance") on the Platform balance-row Send sheet.

## What was done?
- New `operations::shield_to` takes an optional `recipient: Option<&PaymentAddress>` plus a 36-byte memo; `operations::shield` keeps its existing signature as a self-shield front over it. `None` keeps today's shield-to-self behavior exactly (Shield/In activity, empty memo). `Some(addr)` builds the note for that address and records the live activity as **Sent/Out** with the recipient's raw-43 counterparty and the memo — the same classification the scan deriver already produces for an OVK-recovered send to a non-own address, so a restored wallet derives the same row and the live/scan activity ids stay aligned (visible-cmx hashing unchanged).
- The recipient must actually be a third party: `resolve_shield_recipient` rejects a `Some` address the account's own IVK recognizes (default or any diversified index — the same `diversifier_index` test the scan's `is_own_orchard_recipient` uses), because its note would be spendable here and a live Sent/Out row would fork from the self-pay row a restore's scan derives.
- `PlatformWallet::shielded_shield_from_account_to_recipient` parses the raw-43 recipient like `shielded_transfer_to` and shares the existing selection/single-flight/preflight body with `shielded_shield_from_account`.
- New **additive** FFI `platform_wallet_manager_shielded_shield_to_recipient` (wallet id, shielded/payment account, recipient_raw_43, amount, memo_text, signer), guarded by the standard `catch_spend_panic` split (helpers mirrored from #4312 — whichever PR merges second dedupes them). The existing shield extern is untouched, so the JNI binding keeps compiling.
- Swift SDK `shieldedShieldToRecipient`, mirroring `shieldedTransfer`'s recipient/memo handling; it and `shieldedShield` now pin the signer with `withExtendedLifetime` across the whole detached FFI call.

No consensus change: the shield transition already carries the recipient inside the opaque Orchard action; `shielded_shield_preflight` is recipient-agnostic and unchanged.

## How Has This Been Tested?
- New round-trip test `shield_to_external_recipient_decrypts_for_recipient_and_recovers_for_sender`: the built bundle's real output IVK-decrypts only for the recipient (at the sent amount), never for the sender, and OVK-recovers for the sender with the recipient address and memo intact.
- Four unit tests on `resolve_shield_recipient` (self default → Shield/In; external → Sent/Out + raw-43 counterparty; own default rejected; own diversified rejected) and two on the FFI `catch_spend_panic` guard (pass-through; panic → `ErrorShieldedSpendUnconfirmed`).
- `cargo fmt` / `clippy --workspace --all-targets --all-features -D warnings` / `cargo check --workspace --all-features` all clean; wallet-crate tests: 897 passed, 1 failed — `regression_reports_max_from_usable_suffix_not_total_account_balance`, which fails identically on a clean v4.2-dev checkout (fixture invalidated by the #4467 fee-constant rebalance; unrelated to this change).
- `./build_ios.sh --target sim` incl. the example app with warnings-as-errors; dashwallet-ios builds and smoke-tests against this branch (app PR to follow, linked once open).

## Breaking Changes
None — the new FFI entry point and wallet/SDK methods are additive; existing signatures are untouched.

## Summary by CodeRabbit

* **New Features**
  * Added support for shielding funds from a Platform Payment account to an external shielded recipient.
  * Added optional memo support for recipient-directed shielded transfers.
  * Added Swift SDK access for initiating recipient-directed shielding operations.
  * Transactions now distinguish incoming shields from outgoing transfers in activity records.

* **Bug Fixes**
  * Added validation for recipient information and transfer inputs.

* **Tests**
  * Added coverage confirming recipient privacy, transfer amounts, addresses, and memos are handled correctly.

### DIFF
--- a/packages/rs-platform-wallet/src/wallet/shielded/operations.rs
+++ b/packages/rs-platform-wallet/src/wallet/shielded/operations.rs
@@ -446,17 +446,108 @@ fn reserve_shield_fee_on_input_0(
     Ok(inputs)
 }
 
+/// The resolved Orchard output and live-activity classification for a
+/// shield — see [`resolve_shield_recipient`].
+#[derive(Debug)]
+struct ShieldRecipient {
+    /// The address the note is built for.
+    address: OrchardAddress,
+    /// Raw 43-byte recipient for the activity row (`Some` only for a
+    /// third-party recipient).
+    counterparty: Option<Vec<u8>>,
+    kind: ShieldedActivityKind,
+    direction: ShieldedDirection,
+}
+
+/// Resolve a shield's Orchard output address and live-activity
+/// classification from the optional third-party `recipient`.
+///
+/// `None` is the internal shield-to-self: the note goes to the
+/// account's default address and the live entry is `Shield`/`In` with
+/// no counterparty. `Some` pays a THIRD-PARTY address: `Sent`/`Out`
+/// with the raw 43-byte address as counterparty — the exact
+/// classification the scan deriver produces for an OVK-recovered send
+/// to a non-own address, so a restore derives the same row.
+///
+/// A `Some` address the account's own IVK recognizes (default or any
+/// diversified index — the same `diversifier_index` test the scan's
+/// `is_own_orchard_recipient` uses) is rejected instead of classified:
+/// its note WOULD be spendable here, so it is not a send, and
+/// recording it live as `Sent`/`Out` while a restore scan-derives a
+/// self-pay row would fork the two histories. Self-shields take the
+/// `None` path.
+fn resolve_shield_recipient(
+    keys: &AccountViewingKeys,
+    recipient: Option<&PaymentAddress>,
+) -> Result<ShieldRecipient, PlatformWalletError> {
+    match recipient {
+        Some(payment_address) => {
+            if keys
+                .incoming_viewing_key
+                .diversifier_index(payment_address)
+                .is_some()
+            {
+                return Err(PlatformWalletError::ShieldedBuildError(
+                    "recipient belongs to this shielded account; use the self-shield \
+                     entry point (no recipient) instead"
+                        .to_string(),
+                ));
+            }
+            Ok(ShieldRecipient {
+                address: payment_address_to_orchard(payment_address)?,
+                counterparty: Some(payment_address.to_raw_address_bytes().to_vec()),
+                kind: ShieldedActivityKind::Sent,
+                direction: ShieldedDirection::Out,
+            })
+        }
+        None => Ok(ShieldRecipient {
+            address: default_orchard_address(keys)?,
+            counterparty: None,
+            kind: ShieldedActivityKind::Shield,
+            direction: ShieldedDirection::In,
+        }),
+    }
+}
+
+/// Shield credits from transparent platform addresses into the
+/// shielded pool, with the resulting note assigned to `account`'s
+/// default Orchard payment address derived from `keys`.
+///
+/// Self-shield front for [`shield_to`], preserving the pre-recipient
+/// signature for existing callers.
+#[allow(clippy::too_many_arguments)]
+pub async fn shield<S: ShieldedStore, Sig: Signer<PlatformAddress>, P: OrchardProver>(
+    sdk: &Arc<dash_sdk::Sdk>,
+    store: &Arc<RwLock<S>>,
+    persister: Option<&WalletPersister>,
+    wallet_id: WalletId,
+    keys: &AccountViewingKeys,
+    account: u32,
+    inputs: BTreeMap<PlatformAddress, Credits>,
+    amount: u64,
+    signer: &Sig,
+    prover: &P,
+) -> Result<(), PlatformWalletError> {
+    shield_to(
+        sdk, store, persister, wallet_id, keys, account, None, inputs, amount,
+        [0u8; 36], // empty memo
+        signer, prover,
+    )
+    .await
+}
+
 /// Shield credits from transparent platform addresses into the
 /// shielded pool. `recipient` selects the note's Orchard payment
 /// address: `None` assigns it to `account`'s default address derived
 /// from `keys` (the internal shield-to-self); `Some` pays a
 /// third-party address — the note funds THAT wallet's pool and never
-/// becomes spendable here. Either way the output is encrypted under
-/// our own OVK, so the scan recovers the send from chain data and the
-/// live and scan-derived activity ids line up (same convention as
-/// [`transfer`], which also does not special-case a self recipient).
+/// becomes spendable here (a `Some` address this account's own IVK
+/// recognizes is rejected — see [`resolve_shield_recipient`]). Either
+/// way the output is encrypted under our own OVK, so the scan recovers
+/// the send from chain data and the live and scan-derived activity ids
+/// line up.
 #[allow(clippy::too_many_arguments)]
-pub async fn shield<S: ShieldedStore, Sig: Signer<PlatformAddress>, P: OrchardProver>(
+pub async fn shield_to<S: ShieldedStore, Sig: Signer<PlatformAddress>, P: OrchardProver>(
     sdk: &Arc<dash_sdk::Sdk>,
     store: &Arc<RwLock<S>>,
     persister: Option<&WalletPersister>,
@@ -470,13 +561,12 @@ pub async fn shield<S: ShieldedStore, Sig: Signer<PlatformAddress>, P: OrchardPr
     signer: &Sig,
     prover: &P,
 ) -> Result<(), PlatformWalletError> {
-    let (recipient_addr, external_counterparty) = match recipient {
-        Some(payment_address) => (
-            payment_address_to_orchard(payment_address)?,
-            Some(payment_address.to_raw_address_bytes().to_vec()),
-        ),
-        None => (default_orchard_address(keys)?, None),
-    };
+    let ShieldRecipient {
+        address: recipient_addr,
+        counterparty: external_counterparty,
+        kind,
+        direction,
+    } = resolve_shield_recipient(keys, recipient)?;
     let id = SubwalletId::new(wallet_id, account);
 
     // Reserve the flat shielded fee `F` on top of `amount` in the input
@@ -559,19 +649,12 @@ pub async fn shield<S: ShieldedStore, Sig: Signer<PlatformAddress>, P: OrchardPr
     trace!("Shield credits: state transition built, broadcasting...");
     let network = sdk.network;
 
-    // Live activity. Shield-to-self is `Shield`/`direction in` (the note
-    // value enters our pool); an external recipient is `Sent`/`direction
-    // out` with the recipient's raw 43-byte Orchard address as
-    // counterparty — the exact classification the scan deriver produces
-    // for an OVK-recovered send to a non-own address, so a restore
-    // derives the same row. Fee = the flat shielded fee reserved above.
-    // The visible output cmx is the recipient note (OVK-keyed either
-    // way), so the live and scan ids line up.
-    let (kind, direction) = if external_counterparty.is_some() {
-        (ShieldedActivityKind::Sent, ShieldedDirection::Out)
-    } else {
-        (ShieldedActivityKind::Shield, ShieldedDirection::In)
-    };
+    // Live activity. Kind / direction / counterparty were resolved
+    // alongside the recipient above (see `resolve_shield_recipient` for
+    // why the rows match what a restore's scan derives). Fee = the flat
+    // shielded fee reserved above. The visible output cmx is the
+    // recipient note (OVK-keyed either way), so the live and scan ids
+    // line up.
     let pending_entry = record_pending_activity(
         store,
         persister,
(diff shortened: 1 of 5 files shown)


packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerShieldedSync.swift
```diff
@@ -721,6 +721,91 @@ extension PlatformWalletManager {
...
+                        ).check()
+                    }
+                    if let memo, !memo.isEmpty {
+                        try memo.withCString { try send($0) }
+                    } else {
+                        try send(nil)
+                    }
+                }
+            }
+        }.value
```
thepastaclaw, line 814: **🔴 Blocking: Pin the signer across the entire detached FFI call**

`KeychainSigner` registers `self` as an `Unmanaged.passUnretained` callback context and destroys the Rust signer handle from `deinit`. The standalone `_ = addressSigner` is only a last use; optimized ARC may release the object before the subsequent FFI call completes. Rust re-materializes the raw handle as `&VTableSigner` inside its synchronously awaited proof worker and may invoke the Swift callback through it, so an operation-local signer can be deallocated while Rust still holds or uses the pointer. Wrap the complete marshalling and FFI call in `withExtendedLifetime(addressSigner)`.

```suggestion
try await Task.detached(priority: .userInitiated) {
    try withExtendedLifetime(addressSigner) {
        try walletId.withUnsafeBytes { widRaw in
            guard let widPtr = widRaw.baseAddress?.assumingMemoryBound(to: UInt8.self)
            else {
                throw PlatformWalletError.invalidParameter("walletId baseAddress is nil")
            }
            try recipientRaw43.withUnsafeBytes { recipientRaw in
                guard let recipientPtr = recipientRaw.baseAddress?
                    .assumingMemoryBound(to: UInt8.self)
                else {
                    throw PlatformWalletError.invalidParameter(
                        "recipient baseAddress is nil"
                    )
                }
                let send: (UnsafePointer<CChar>?) throws -> Void = { memoCStr in
                    try platform_wallet_manager_shielded_shield_to_recipient(
                        handle, widPtr, shieldedAccount, paymentAccount,
                        recipientPtr, amount, memoCStr, signerHandle
                    ).check()
                }
                if let memo, !memo.isEmpty {
                    try memo.withCString { try send($0) }
                } else {
                    try send(nil)
                }
            }
        }
    }
}.value
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/shielded/operations.rs
```diff
@@ -457,12 +463,20 @@ pub async fn shield<S: ShieldedStore, Sig: Signer<PlatformAddress>, P: OrchardPr
     wallet_id: WalletId,
     keys: &AccountViewingKeys,
     account: u32,
+    recipient: Option<&PaymentAddress>,
     inputs: BTreeMap<PlatformAddress, Credits>,
     amount: u64,
+    memo: [u8; 36],
     signer: &Sig,
     prover: &P,
```
thepastaclaw: **🟡 Suggestion: Preserve the existing public shield Rust API**

`wallet::shielded` and its `operations` module are public, making this `pub async fn shield` reachable by downstream Rust consumers even though it is not re-exported at the crate root. Adding mandatory `recipient` and `memo` parameters therefore breaks existing callers, contrary to the PR's additive/no-breaking-change contract. Keep the previous signature as a wrapper that supplies the default recipient and empty memo, and expose the new behavior through a separately named function or private implementation helper.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/shielded/operations.rs
```diff
@@ -457,12 +463,20 @@ pub async fn shield<S: ShieldedStore, Sig: Signer<PlatformAddress>, P: OrchardPr
...
 ) -> Result<(), PlatformWalletError> {
-    let recipient_addr = default_orchard_address(keys)?;
+    let (recipient_addr, external_counterparty) = match recipient {
+        Some(payment_address) => (
+            payment_address_to_orchard(payment_address)?,
+            Some(payment_address.to_raw_address_bytes().to_vec()),
+        ),
+        None => (default_orchard_address(keys)?, None),
+    };
```
thepastaclaw: **🟡 Suggestion: Do not infer an external recipient from Option::Some**

A valid `PaymentAddress` passed through the new recipient API can belong to the selected shielded account, including a diversified address. This branch classifies every `Some` value as `Sent/Out`, while restoration tests ownership with `incoming_viewing_key.diversifier_index` and classifies an own output as incoming or a self-transfer. That makes live and restored activity semantics diverge for an input the public raw-address API currently accepts. Enforce the method's documented third-party invariant by rejecting addresses recognized by the source account and directing callers to the self-shield API.

```suggestion
let (recipient_addr, external_counterparty) = match recipient {
    Some(payment_address) => {
        if keys
            .incoming_viewing_key
            .diversifier_index(payment_address)
            .is_some()
        {
            return Err(PlatformWalletError::ShieldedBuildError(
                "recipient belongs to the source shielded account; use shield-to-self"
                    .to_string(),
            ));
        }
        (
            payment_address_to_orchard(payment_address)?,
            Some(payment_address.to_raw_address_bytes().to_vec()),
        )
    }
    None => (default_orchard_address(keys)?, None),
};
```

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/wallet/shielded/sync/shield_decrypt_tests.rs
```diff
@@ -150,3 +150,111 @@ async fn shield_built_note_is_trial_decryptable_by_own_ivk() {
...
+        0,
+        &&prover,
+        memo,
+        // Production config (`operations::shield`): OVK-keyed to the
+        // SENDER, so the sender's scan can recover the send.
+        Some(sender_keys.outgoing_viewing_key.clone()),
+        PlatformVersion::latest(),
+    )
+    .await
+    .expect("shield transition build should succeed");
```
thepastaclaw, line 199: **🟡 Suggestion: Exercise the new live-activity branch in tests**

The new test invokes the pre-existing `build_shield_transition` primitive directly, so it does not exercise this PR's wallet or `operations::shield` changes. It would still pass if the wallet ignored the recipient, dropped the memo before calling the builder, or recorded the external payment as `Shield/In`. Add focused coverage through the new wallet path, or extract its activity-parameter preparation into a pure helper, and assert recipient and memo forwarding plus `Sent/Out`, raw-43 counterparty, and live-versus-scan activity-ID alignment.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The recipient, memo, and OVK plumbing is sound, but the new Swift entry point can release its pass-unretained signer before the synchronous Rust FFI call finishes, creating an in-scope use-after-free boundary. The Rust change also breaks a public free-function signature, misclassifies wallet-owned recipients, and lacks coverage of the newly added wallet/activity path.
Source: Codex general, security-auditor, rust-quality, and ffi-engineer reviewers — gpt-5.6-sol; final verifier — gpt-5.6-sol; orchestration-only, not reviewer evidence — openclaw-agent/cliproxy/gpt-5.6-sol.

### Review provenance

<details>

```

In `packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerShieldedSync.swift`:
- [BLOCKING] packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerShieldedSync.swift:773-806: Pin the signer across the entire detached FFI call
  `KeychainSigner` registers `self` as an `Unmanaged.passUnretained` callback context and destroys the Rust signer handle from `deinit`. The standalone `_ = addressSigner` is only a last use; optimized ARC may release the object before the subsequent FFI call completes. Rust re-materializes the raw handle as `&VTableSigner` inside its synchronously awaited proof worker and may invoke the Swift callback through it, so an operation-local signer can be deallocated while Rust still holds or uses the pointer. Wrap the complete marshalling and FFI call in `withExtendedLifetime(addressSigner)`.

- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/operations.rs:459-471: Preserve the existing public shield Rust API
  `wallet::shielded` and its `operations` module are public, making this `pub async fn shield` reachable by downstream Rust consumers even though it is not re-exported at the crate root. Adding mandatory `recipient` and `memo` parameters therefore breaks existing callers, contrary to the PR's additive/no-breaking-change contract. Keep the previous signature as a wrapper that supplies the default recipient and empty memo, and expose the new behavior through a separately named function or private implementation helper.
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/operations.rs:473-479: Do not infer an external recipient from Option::Some
  A valid `PaymentAddress` passed through the new recipient API can belong to the selected shielded account, including a diversified address. This branch classifies every `Some` value as `Sent/Out`, while restoration tests ownership with `incoming_viewing_key.diversifier_index` and classifies an own output as incoming or a self-transfer. That makes live and restored activity semantics diverge for an input the public raw-address API currently accepts. Enforce the method's documented third-party invariant by rejecting addresses recognized by the source account and directing callers to the self-shield API.

In `packages/rs-platform-wallet/src/wallet/shielded/sync/shield_decrypt_tests.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/wallet/shielded/sync/shield_decrypt_tests.rs:183-199: Exercise the new live-activity branch in tests
  The new test invokes the pre-existing `build_shield_transition` primitive directly, so it does not exercise this PR's wallet or `operations::shield` changes. It would still pass if the wallet ignored the recipient, dropped the memo before calling the builder, or recorded the external payment as `Shield/In`. Add focused coverage through the new wallet path, or extract its activity-parameter preparation into a pure helper, and assert recipient and memo forwarding plus `Sent/Out`, raw-43 counterparty, and live-versus-scan activity-ID alignment.
```
</details>

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerShieldedSync.swift
QuantumExplorer: Fixed in 766d4e48e4 — both `shieldedShieldToRecipient` and `shieldedShield` (same weaker pattern) now wrap the whole marshalling + FFI call in `withExtendedLifetime(addressSigner)`, matching `shieldedTransfer`.

packages/swift-sdk/Sources/SwiftDashSDK/PlatformWallet/PlatformWalletManagerShieldedSync.swift
```diff
@@ -721,6 +721,91 @@ extension PlatformWalletManager {
...
+                "recipient must be exactly 43 raw Orchard bytes"
+            )
+        }
+
+        let handle = self.handle
+        let signerHandle = addressSigner.handle
+
+        try await Task.detached(priority: .userInitiated) {
+            // Keepalive — same rationale as `shieldedShield`.
+            _ = addressSigner
```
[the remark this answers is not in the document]
QuantumExplorer: Fixed in 766d4e48e4 — took the "fix both" option: `shieldedShieldToRecipient` and `shieldedShield` now pin the signer with `withExtendedLifetime` across the whole detached FFI call.

packages/rs-platform-wallet/src/wallet/shielded/operations.rs
QuantumExplorer: Done in 766d4e48e4: `operations::shield` keeps its pre-recipient signature as a self-shield front, and the recipient/memo behavior moved to the new `pub async fn shield_to` — mirroring the wallet-level `shielded_shield_from_account` / `_to_recipient` split. The PR description's "existing signatures are untouched" claim holds again.

packages/rs-platform-wallet/src/wallet/shielded/operations.rs
QuantumExplorer: Done in 766d4e48e4: recipient resolution now lives in a pure `resolve_shield_recipient` helper that rejects a `Some` address the account's own IVK recognizes (`diversifier_index`, the same test the scan's `is_own_orchard_recipient` uses — so diversified own addresses are caught too), directing callers to the self-shield entry point. The third-party contract is now documented at the wallet method, the FFI export, and the Swift method.

packages/rs-platform-wallet/src/wallet/shielded/sync/shield_decrypt_tests.rs
QuantumExplorer: Addressed via the extraction option in 766d4e48e4: the activity-parameter preparation is now the pure `resolve_shield_recipient` helper with four tests — `None` → default address + Shield/In + no counterparty, external → Sent/Out + raw-43 counterparty, own default rejected, own diversified rejected. Recipient/memo forwarding into the built bundle stays covered by the round-trip decrypt test (`shield_to_external_recipient_decrypts_for_recipient_and_recovers_for_sender`); a full pass through `operations::shield_to` would need a mockable SDK broadcast seam this crate doesn't have.

PR #4473, opened 2026-08-24
fix(platform): preserve tracked masternode ABI and persistence

QuantumExplorer:

Follow up on #4465 to preserve the established masternode C ABI and make tracked-masternode persistence safe across failures, concurrency, SwiftData migration, and unrelated changeset rollbacks.

## What was done?

- Restored the frozen `MasternodeEntryFFI` layout and legacy wallet-only behavior; added `MasternodeEntryV2FFI` plus versioned list/free APIs for provenance and labels.
- Reworked callback-extension size gates to compute field sizes from types before reading optional trailing fields.
- Made Rust whole-set registry mutations linearizable and restored their complete in-memory before-image when persistence fails.
- Added `DashSchemaV2` with a lightweight V1 migration for `PersistentTrackedMasternode`.
- Isolated tracked-masternode SwiftData writes in a dedicated context so unrelated changeset rollback cannot discard them.
- Added ABI layout/stride/free, short-allocation, failure rollback, migration, and changeset-isolation regression tests.

## How Has This Been Tested?

- `cargo test -p platform-wallet --lib` — 730 passed.
- `cargo test -p platform-wallet-ffi --lib` — 277 passed.
- `swift test` — 364 passed, 12 skipped, 0 failed.
- `./build_ios.sh --target sim --profile dev` — simulator XCFramework generated and SwiftExampleApp warnings-as-errors build succeeded.
- `cargo fmt --all -- --check` and `git diff --check` — passed.

## Breaking Changes

None. The pre-#4465 C ABI and behavior remain available through the unversioned entry points; additive fields use V2 APIs.

## Checklist:

### DIFF
--- a/packages/rs-platform-wallet-ffi/src/core_wallet_types.rs
+++ b/packages/rs-platform-wallet-ffi/src/core_wallet_types.rs
@@ -916,7 +916,14 @@ fn transaction_type_to_u8(
 /// [`MasternodeRecord`], built by [`masternode_entry_ffi`] and
 /// returned by `platform_wallet_manager_list_masternodes`. Inline
 /// fixed-size hashes with `has_*` gates (mirroring `TransactionRecordFFI`)
-/// keep heap ownership to the three C strings.
+/// keep heap ownership to the C strings.
+///
+/// # ABI stability
+///
+/// This is the original, frozen layout returned by the unversioned
+/// `platform_wallet_manager_list_masternodes` entry point. Do not add, remove,
+/// or reorder fields. New projections belong in a versioned wrapper such as
+/// [`MasternodeEntryV2FFI`].
 #[repr(C)]
 pub struct MasternodeEntryFFI {
     /// proTxHash (32 wire bytes) — group key; also the registration txid.
@@ -994,12 +1001,6 @@ pub struct MasternodeEntryFFI {
     pub platform_in_wallet: bool,
     pub platform_account_type: u8,
     pub platform_key_index: u32,
-    /// Where this record came from: 0 = one of the wallet's own masternodes
-    /// (aggregated from its provider transactions), 1 = tracked by the user
-    /// independently of every wallet.
-    pub source: u8,
-    /// User label of a tracked masternode, or null.
-    pub label: *mut c_char,
     /// Whether the platform-node ownership check was actually *possible* for
     /// this query: `true` when the wallet's derived platform-node index had
     /// entries to compare against, `false` when it was empty/unavailable (no
@@ -1009,6 +1010,20 @@ pub struct MasternodeEntryFFI {
     /// key rotation to an external node) from "couldn't check yet", so it
     /// never leaves stale ownership set. See `MasternodeSync`.
     pub platform_ownership_checked: bool,
+}
+
+/// Version 2 masternode projection. The frozen V1 entry remains the first
+/// field, preserving one canonical definition for all established fields;
+/// V2 adds record provenance and the optional tracked-node label.
+#[repr(C)]
+pub struct MasternodeEntryV2FFI {
+    pub v1: MasternodeEntryFFI,
+    /// Where this record came from: 0 = one of the wallet's own masternodes
+    /// (aggregated from its provider transactions), 1 = tracked by the user
+    /// independently of every wallet.
+    pub source: u8,
+    /// User label of a tracked masternode, or null.
+    pub label: *mut c_char,
 }
 
 /// Encode a hash160 as a network-specific base58 P2PKH address string
@@ -1138,14 +1153,26 @@ pub(crate) fn masternode_entry_ffi(
         platform_in_wallet,
         platform_account_type,
         platform_key_index,
+        platform_ownership_checked: mn.platform_ownership_checked,
+    }
+}
+
+/// Flatten one record into the additive V2 C-ABI entry.
+pub(crate) fn masternode_entry_v2_ffi(
+    mn: &MasternodeRecord,
+    network: dashcore::Network,
+) -> MasternodeEntryV2FFI {
+    use std::ffi::CString;
+
+    MasternodeEntryV2FFI {
+        v1: masternode_entry_ffi(mn, network),
         source: mn.source.as_u8(),
         label: mn
             .label
             .clone()
-            .and_then(|l| CString::new(l).ok())
+            .and_then(|label| CString::new(label).ok())
             .map(CString::into_raw)
             .unwrap_or(std::ptr::null_mut()),
-        platform_ownership_checked: mn.platform_ownership_checked,
     }
 }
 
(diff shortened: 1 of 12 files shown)


packages/rs-platform-wallet/src/masternode/tracked.rs
```diff
@@ -771,24 +781,14 @@ impl TrackedMasternodes {
...
-                Some(live) => {
-                    live.snapshot = tracked.snapshot.clone();
-                    tracked.label = live.label.clone();
-                    true
-                }
-                None => false,
+        let label = self.mutate_and_persist(|guard| match guard.get_mut(pro_tx_hash) {
+            Some(live) => {
+                live.snapshot = tracked.snapshot.clone();
+                Ok(live.label.clone())
```
thepastaclaw: **🔴 Blocking: Concurrent refreshes can overwrite newly learned snapshot data**

Each refresh clones the current row at line 673 before awaiting several network requests, while this mutation later replaces the entire live snapshot. Two refreshes of the same masternode can therefore start from the same state: refresh A can learn and persist a registration, while refresh B receives `Ok(None)` and retains its original `registration == None`. If B finishes last, this assignment replaces A's snapshot with B's stale clone and persists the loss. The same race affects list and Platform metadata. The service is explicitly cloneable and `Send + Sync`, and the FFI dispatches refreshes through a multithreaded runtime, so callers are not serialized. Serialize refreshes per masternode, merge only fields actually learned by each refresh into the live snapshot, or use a revision/CAS retry, and add an interleaved-refresh regression test.

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The ABI restoration, callback size gating, persistence rollback, and SwiftData migration changes are coherent, but one blocking concurrency defect remains in tracked-masternode refresh persistence. Concurrent refreshes can commit stale whole snapshots and discard metadata learned and persisted by another refresh.
Source: reviewer backend gpt-5.6-sol (Codex general, FFI engineer, Rust quality, and security auditor lanes); final verifier backend gpt-5.6-sol. Orchestration-only: openclaw-agent/cliproxy/gpt-5.6-sol (not reviewer evidence).

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — ffi-engineer (completed), `gpt-5.6-sol` — rust-quality (completed), `gpt-5.6-sol` — security-auditor (completed)

🔴 1 blocking

<details>

```

In `packages/rs-platform-wallet/src/masternode/tracked.rs`:
- [BLOCKING] packages/rs-platform-wallet/src/masternode/tracked.rs:784-787: Concurrent refreshes can overwrite newly learned snapshot data
  Each refresh clones the current row at line 673 before awaiting several network requests, while this mutation later replaces the entire live snapshot. Two refreshes of the same masternode can therefore start from the same state: refresh A can learn and persist a registration, while refresh B receives `Ok(None)` and retains its original `registration == None`. If B finishes last, this assignment replaces A's snapshot with B's stale clone and persists the loss. The same race affects list and Platform metadata. The service is explicitly cloneable and `Send + Sync`, and the FFI dispatches refreshes through a multithreaded runtime, so callers are not serialized. Serialize refreshes per masternode, merge only fields actually learned by each refresh into the live snapshot, or use a revision/CAS retry, and add an interleaved-refresh regression test.
```
</details>

packages/rs-platform-wallet/src/masternode/tracked.rs
QuantumExplorer: Confirmed and fixed.

The race was real and as described: `refresh` cloned the row up front, then the final `mutate_and_persist` assigned the whole `live.snapshot`, so two passes over one node could start from the same state and the one finishing last would put its pre-read clone back — losing the other's `registration`, Platform key hashes, `list`/`ever_listed` and `refreshed_at` — and the whole-set write right after made the loss durable. Nothing serialized the callers: each FFI call builds its own `TrackedMasternodes` from the manager and dispatches through the shared multithreaded runtime.

Fix — serialize per masternode. The manager now holds a `TrackedMasternodeRegistry` (rows + a per-proTxHash gate map, `std::sync::Mutex<BTreeMap<[u8; 32], Arc<tokio::sync::Mutex<()>>>>`) behind one `Arc`, so every handle it hands out shares the gates as well as the rows. `refresh_row_and_persist` takes the node's gate FIRST, then reads the row, runs the network half, and writes back: the gate spans the read and the write, so a second refresh of the same node waits for the one in flight (and skips the network work it already did) instead of racing it. The `std` lock is held only long enough to clone the `Arc` — released before the `await`, so the future stays `Send` and nothing blocks the runtime; gates whose only reference is the map are swept on the next acquisition, so it stays bounded. The network steps moved into `TrackedMasternodes::learn`, which merges into the snapshot it is handed rather than into a clone it took earlier, so "came back empty" now means "keep what is there".

Regression test: `interleaved_refresh_passes_keep_what_the_other_learned` — two real `refresh_row_and_persist` passes over one node with scripted network halves, interleaved so the pass that learns nothing writes last; it asserts the registration the other pass learned survives both in the registry and in the persisted set. It fails on the unserialized ordering (drop the gate line and `registration` comes back `None`) and passes with it.

packages/rs-platform-wallet-ffi/src/core_wallet_types.rs
```diff
@@ -1491,4 +1518,97 @@ mod tests {
...
+        );
+        assert_eq!(
+            std::mem::offset_of!(MasternodeEntryFFI, platform_key_index),
+            284
+        );
+        assert_eq!(
+            std::mem::offset_of!(MasternodeEntryFFI, platform_ownership_checked),
+            288
+        );
+        assert_eq!(std::mem::offset_of!(MasternodeEntryV2FFI, v1), 0);
```
thepastaclaw, line 1551: **🟡 Suggestion: The new V2 C ABI layout is not pinned**

The layout regression test freezes V1 but only verifies that V2 begins with its nested V1 member. The stride test constructs and reads the array through the same current Rust type, so it remains green if `source` or `label` is reordered, padding changes, or another field is inserted. Because `MasternodeEntryV2FFI` is now a public versioned C array element, pin its 64-bit size, alignment, and additive-field offsets so incompatible changes require a V3 API.

```suggestion
        assert_eq!(std::mem::size_of::<MasternodeEntryV2FFI>(), 312);
        assert_eq!(std::mem::align_of::<MasternodeEntryV2FFI>(), 8);
        assert_eq!(std::mem::offset_of!(MasternodeEntryV2FFI, v1), 0);
        assert_eq!(std::mem::offset_of!(MasternodeEntryV2FFI, source), 296);
        assert_eq!(std::mem::offset_of!(MasternodeEntryV2FFI, label), 304);
```

<sub>source: ['codex']</sub>

thepastaclaw (commented):

The prior blocking stale-snapshot race is fixed: refreshes now share a per-masternode gate that spans snapshot read, learning, mutation, and persistence, and the targeted regression test passes. Two non-blocking test-coverage gaps remain around freezing the new V2 ABI layout and deterministically exercising the refresh interleaving.
Source: reviewer backend gpt-5.6-sol (Codex general, FFI engineer, Rust quality, and security auditor lanes); final verifier backend gpt-5.6-sol. Orchestration-only: openclaw-agent/cliproxy/gpt-5.6-sol (not reviewer evidence).

### Review provenance
- Codex reviewers: `gpt-5.6-sol` — general (completed), `gpt-5.6-sol` — ffi-engineer (completed), `gpt-5.6-sol` — rust-quality (completed), `gpt-5.6-sol` — security-auditor (completed)

🟡 2 suggestion(s)

<details>

```

- [SUGGESTION] packages/rs-platform-wallet-ffi/src/core_wallet_types.rs:1551: The new V2 C ABI layout is not pinned
  The layout regression test freezes V1 but only verifies that V2 begins with its nested V1 member. The stride test constructs and reads the array through the same current Rust type, so it remains green if `source` or `label` is reordered, padding changes, or another field is inserted. Because `MasternodeEntryV2FFI` is now a public versioned C array element, pin its 64-bit size, alignment, and additive-field offsets so incompatible changes require a V3 API.

In `packages/rs-platform-wallet/src/masternode/tracked.rs`:
- [SUGGESTION] packages/rs-platform-wallet/src/masternode/tracked.rs:1346-1347: The refresh regression test does not guarantee the intended interleaving
  The regression test relies on one `yield_now()` polling the spawned blind refresh before the learner is released, but Tokio does not guarantee which runnable task is polled after a yield. If the test task or learner proceeds before the blind refresh is first polled, the learner can persist its registration before the blind pass reads the row; the test then passes even if the per-node gate is removed because the blind pass starts from the updated snapshot. Add explicit first-poll synchronization so the blind refresh is known to have reached a pending point before releasing the learner.
```
</details>

packages/rs-platform-wallet/src/masternode/tracked.rs
```diff
@@ -1049,6 +1168,207 @@ mod tests {
...
+                    move |snapshot| async move {
+                        release_blind_rx.await.expect("the test releases this");
+                        (snapshot, RefreshOutcome::default())
+                    },
+                )
+                .await
+            })
+        };
+        // Let it get as far as it can while the learner still holds the row.
+        tokio::task::yield_now().await;
```
thepastaclaw, line 1347: **🟡 Suggestion: The refresh regression test does not guarantee the intended interleaving**

The regression test relies on one `yield_now()` polling the spawned blind refresh before the learner is released, but Tokio does not guarantee which runnable task is polled after a yield. If the test task or learner proceeds before the blind refresh is first polled, the learner can persist its registration before the blind pass reads the row; the test then passes even if the per-node gate is removed because the blind pass starts from the updated snapshot. Add explicit first-poll synchronization so the blind refresh is known to have reached a pending point before releasing the learner.

<sub>source: ['codex']</sub>

packages/rs-platform-wallet/src/masternode/tracked.rs
thepastaclaw: Resolved in `874859d` — _Concurrent refreshes can overwrite newly learned snapshot data_ no longer present.

PR #4166, opened 2026-07-20
fix(drive): bind proof-verifier queries to trusted context

QuantumExplorer:
## Summary

- require vote proofs to bind a full serialized poll to its expected ID, contract, document type, index, and exact contested-index arity
- reject WASM known-contract entries whose map key differs from the embedded contract ID
- stop using unsigned response metadata to select epoch-dependent proof queries; ambiguous current-epoch requests now fail closed

This addresses security review findings DS-CAND-314, 368, and 374.

## Compatibility

The two WASM vote-proof exports now require both an expected poll ID and the full serialized vote poll.

Proved epoch queries no longer derive their bound from unsigned metadata:

- `ExtendedEpochInfo::fetch_current` (a descending epoch query with no explicit start) now fails closed with `proved descending epoch queries require an explicit start epoch`. There is no authenticated current-epoch marker to bound the query, so a node could otherwise cap the proof below the chain tip and pass off a stale epoch as current. Callers must fetch a specific epoch by explicit index until such a marker exists.
- The proved evonode-proposed-epoch-blocks queries likewise require an explicit epoch.

A contested vote poll must now carry exactly as many index values as the index has properties; under-specified polls are rejected instead of silently resolving a partial path.

## Validation

- cargo check -p drive-proof-verifier --lib
- cargo check -p wasm-drive-verify --lib
- cargo test -p drive-proof-verifier --lib (242 passed)
- cargo test -p drive --lib contenders_path_rejects_missing_index_values
- cargo test -p wasm-drive-verify --lib (4 passed)
- cargo test -p dash-sdk --test main fetch::epoch:: fetch::contested_resource_voters:: (7 passed) — realigned `test_epoch_fetch_current` and `test_contested_resource_voters_for_identity_not_found` to assert the new fail-closed behavior
- cargo clippy -p drive -p drive-proof-verifier -p wasm-drive-verify --lib --no-deps -- -D warnings
- cargo fmt --all
- git diff --check

PR #4480, opened 2026-08-25
fix(sdk): backport proved current-epoch fetch fix to v4.1

PastaPastaPasta:

`epoch.current()` / `ExtendedEpochInfo::fetch_current` always fails in the released v4.1.0 and v4.1.1 with:

```
decode request: proved descending epoch queries require an explicit start epoch
```

The proof-context binding hardening (#4166) shipped in v4.1.0 and rejects proved descending epoch queries without an explicit start epoch, but the SDK's own current-epoch fetch still issued exactly that query shape. The fix (#4231) was merged to v4.2-dev on Aug 2 and never reached the 4.1 line, so v4.1.1 (published Aug 18, current npm latest for @dashevo/evo-sdk) ships with the regression. Reproducible against testnet with the published npm artifact.

## What was done?

Clean cherry-pick of 65b5bae25b (#4231) onto v4.1-dev: restores the proved current-epoch fetch as a two-step, guard-compliant query (proved anchor first, then an explicit-start descending epochs query), including the updated offline test vectors.

## How Has This Been Tested?

- `cargo check -p dash-sdk -p drive --locked` on this branch: clean
- `cargo test -p dash-sdk epoch` (offline vectors): 5 passed, 0 failed
- The unpatched failure is demonstrated live by the published 4.1.1 bundle at https://pastapastapasta.github.io/evo-sdk-delivery-bench/before/r1/ (temporary benchmark page)

## Breaking Changes

None — restores intended behavior of an existing API.

## Checklist:

- [x] I have added or updated relevant unit/integration/functional/e2e tests (carried over from #4231)


packages/rs-sdk/src/platform/types/epoch.rs
```diff
@@ -1,57 +1,183 @@
...
+        // than by position.
+        let mut started: BTreeMap<EpochIndex, ExtendedEpochInfo> = epochs
+            .into_iter()
+            .filter_map(|(index, info)| info.map(|info| (index, info)))
+            .collect();
+
+        match started.keys().next_back().copied() {
+            // Nothing at or above the candidate has started: the query landed in
+            // Drive's pre-created empty epoch window.
+            None => return Err(Error::EpochNotFound),
```
thepastaclaw, line 127: **🔴 Blocking: Retry when the confirmation lands on an honest node behind the probe**

The probe and confirmation are separate fetch operations, and each request independently chooses a random live address through `AddressList::get_live_address`. At an epoch boundary, the probe can observe epoch `n` from one node while the confirmation reaches an honest node still at `n - 1`. The second node returns a valid proof with no started epoch at or above candidate `n`, and its one-block-lower metadata passes the SDK's default height tolerance of 1. This branch then creates `EpochNotFound` outside the fetch retry mechanism, so the restored API fails intermittently during ordinary network convergence. The same failure can occur after a proof advances `candidate` and the next refinement reaches an older node. Retry empty confirmations with a bounded policy that preserves the highest proven candidate, or keep the multi-step operation on one node; do not lower the candidate from unsigned metadata.

<sub>source: ['codex']</sub>

packages/rs-sdk-ffi/src/system/queries/platform_status.rs
```diff
@@ -64,40 +67,33 @@ fn get_platform_status(sdk_handle: *const SDKHandle) -> Result<String, String> {
...
+                // No epoch to read a version off, so fall back to the highest
+                // version the SDK has verified so far.
+                let json = format!(
+                    r#"{{"version":{},"network":"{}","blockHeight":0,"coreHeight":0}}"#,
+                    sdk.protocol_version_number(),
+                    network_str
+                );
+                Ok(json)
             }
             Err(e) => Err(format!("Failed to fetch platform status: {}", e)),
```
thepastaclaw, line 98: **🔴 Blocking: Do not turn an untrusted inflated epoch hint into successful zero status**

`EpochNotFound` from the new resolver does not authenticate that Platform has no epochs. The metadata epoch is omitted from the Tenderdash `StateId` constructed in `rs-drive-proof-verifier/src/verify.rs`, so a server can inflate that field without invalidating the quorum signature. The confirmation then validly proves that the attacker-selected future range is empty and returns `EpochNotFound`. This arm converts that fail-closed error into a successful response claiming zero block and core heights; the same false success can result from the honest cross-node race in the SDK resolver. Since an initialized Platform necessarily has a genesis epoch, propagate this error rather than representing it as valid status data.

```suggestion
            Err(e) => Err(format!("Failed to fetch platform status: {}", e)),
```

<sub>source: ['codex']</sub>

packages/rs-sdk/src/error.rs
```diff
@@ -386,6 +387,19 @@ pub enum StaleNodeError {
...
+    /// signature, so `ExtendedEpochInfo::fetch_current` only uses it to shape a
+    /// proved query and then checks it against the proof. This error means the
+    /// check kept failing: every proof showed a newer epoch already started.
+    #[error("received epoch is outdated: hinted {hinted_epoch}, proven started epoch {proven_epoch}; try another server")]
+    Epoch {
+        /// Epoch index the server reported as current in unsigned response metadata
+        hinted_epoch: EpochIndex,
+        /// Newer epoch index that the server's own proof showed as already started
+        proven_epoch: EpochIndex,
+    },
```
thepastaclaw, line 402: **🟡 Suggestion: Adding a public error variant breaks 4.1 source compatibility**

`StaleNodeError` was already a public, exhaustive enum in the released 4.1.1 API. Downstream Rust code can legally exhaustively match its `Height` and `Time` variants; adding `Epoch` makes that code fail to compile with a non-exhaustive-pattern error. This conflicts with the patch backport's claim that it has no breaking changes. Represent the condition using an existing compatible error shape or explicitly treat the release as source-breaking. Adding `#[non_exhaustive]` in this patch would not restore compatibility because that annotation also breaks existing exhaustive matches.

<sub>source: ['codex']</sub>

packages/rs-sdk-ffi/src/system/queries/platform_status.rs
```diff
@@ -117,13 +177,55 @@ mod tests {
...
+    fn test_get_platform_status_without_epoch_expectations_errors() {
         let handle = create_mock_sdk_handle();
         unsafe {
-            let _result = dash_sdk_get_platform_status(handle);
-            // Result depends on mock implementation
-            crate::test_utils::test_utils::destroy_mock_sdk_handle(handle);
+            let result = dash_sdk_get_platform_status(handle);
+            assert!(!result.error.is_null());
+            assert_eq!(result.data_type, DashSDKResultDataType::NoData);
+            destroy_mock_sdk_handle(handle);
```
thepastaclaw, line 228: **💬 Nitpick: Free the FFI error returned by the new test**

On this path, `dash_sdk_get_platform_status` allocates both a boxed `DashSDKError` and its `CString` message. Ownership is returned through `DashSDKResult.error`, but the new test destroys only the SDK handle, leaking both allocations and modeling the foreign caller's ownership contract incorrectly. Return the error through `dash_sdk_error_free` before destroying the handle.

```suggestion
            let result = dash_sdk_get_platform_status(handle);
            assert!(!result.error.is_null());
            assert_eq!(result.data_type, DashSDKResultDataType::NoData);
            crate::dash_sdk_error_free(result.error);
            destroy_mock_sdk_handle(handle);
```

<sub>source: ['codex']</sub>

thepastaclaw (changes requested):

The explicit-start proof flow correctly avoids accepting unsigned epoch metadata as the current epoch, but two blocking paths remain: independently routed confirmation requests can produce false `EpochNotFound` errors during normal node convergence, and the FFI status wrapper converts that fail-closed result into successful zero-valued status. The patch also introduces a source-compatibility break in a public exhaustive Rust enum and leaks the error allocation in a new FFI test.
Source: Codex reviewer lanes (general, security-auditor, rust-quality, and ffi-engineer; exact backend model IDs were not present in the supplied evidence); Claude final verifier (exact backend model ID was not present in the supplied evidence). openclaw-agent/cliproxy/gpt-5.6-sol is orchestration-only and is not reviewer evidence.

### Review provenance

🔴 2 blocking | 🟡 1 suggestion(s) | 💬 1 nitpick(s)

<details>

```

In `packages/rs-sdk/src/platform/types/epoch.rs`:
- [BLOCKING] packages/rs-sdk/src/platform/types/epoch.rs:124-127: Retry when the confirmation lands on an honest node behind the probe
  The probe and confirmation are separate fetch operations, and each request independently chooses a random live address through `AddressList::get_live_address`. At an epoch boundary, the probe can observe epoch `n` from one node while the confirmation reaches an honest node still at `n - 1`. The second node returns a valid proof with no started epoch at or above candidate `n`, and its one-block-lower metadata passes the SDK's default height tolerance of 1. This branch then creates `EpochNotFound` outside the fetch retry mechanism, so the restored API fails intermittently during ordinary network convergence. The same failure can occur after a proof advances `candidate` and the next refinement reaches an older node. Retry empty confirmations with a bounded policy that preserves the highest proven candidate, or keep the multi-step operation on one node; do not lower the candidate from unsigned metadata.

In `packages/rs-sdk-ffi/src/system/queries/platform_status.rs`:
- [BLOCKING] packages/rs-sdk-ffi/src/system/queries/platform_status.rs:88-98: Do not turn an untrusted inflated epoch hint into successful zero status
  `EpochNotFound` from the new resolver does not authenticate that Platform has no epochs. The metadata epoch is omitted from the Tenderdash `StateId` constructed in `rs-drive-proof-verifier/src/verify.rs`, so a server can inflate that field without invalidating the quorum signature. The confirmation then validly proves that the attacker-selected future range is empty and returns `EpochNotFound`. This arm converts that fail-closed error into a successful response claiming zero block and core heights; the same false success can result from the honest cross-node race in the SDK resolver. Since an initialized Platform necessarily has a genesis epoch, propagate this error rather than representing it as valid status data.
- [NITPICK] packages/rs-sdk-ffi/src/system/queries/platform_status.rs:225-228: Free the FFI error returned by the new test
  On this path, `dash_sdk_get_platform_status` allocates both a boxed `DashSDKError` and its `CString` message. Ownership is returned through `DashSDKResult.error`, but the new test destroys only the SDK handle, leaking both allocations and modeling the foreign caller's ownership contract incorrectly. Return the error through `dash_sdk_error_free` before destroying the handle.

In `packages/rs-sdk/src/error.rs`:
- [SUGGESTION] packages/rs-sdk/src/error.rs:390-402: Adding a public error variant breaks 4.1 source compatibility
  `StaleNodeError` was already a public, exhaustive enum in the released 4.1.1 API. Downstream Rust code can legally exhaustively match its `Height` and `Time` variants; adding `Epoch` makes that code fail to compile with a non-exhaustive-pattern error. This conflicts with the patch backport's claim that it has no breaking changes. Represent the condition using an existing compatible error shape or explicitly treat the release as source-breaking. Adding `#[non_exhaustive]` in this patch would not restore compatibility because that annotation also breaks existing exhaustive matches.
```
</details>

PR #4400, opened 2026-08-13
feat(platform)!: required document fields via contract updates (requiredSince)

QuantumExplorer:
## Issue being fixed or feature implemented

Contract owners cannot add new **required** fields to an existing document type: the `required` set is frozen in both directions by the PV14 compatibility rules, for a structural reason — requiredness is baked into the document wire format (required properties serialize raw, optional ones carry a presence flag), so changing it desynchronizes every stored document's bytes from the schema used to read them.

This PR makes it possible. A contract update may add a new required property by annotating it with `requiredSince` equal to the contract version the update creates:

```json
"properties": {
  "newField": { "type": "string", "maxLength": 63, "position": 4, "requiredSince": 3 }
},
"required": ["existingField", "newField"]
```

Documents are stamped with the contract version their bytes conform to (serialization format 3). Deserialization resolves each property's layout by comparing its `requiredSince` against the stamp — so **the latest contract alone reconstructs every stamp's byte layout**. No historical contract lookups are needed anywhere: contract history stays opt-in, and proof verifiers and SDK context providers keep resolving contracts by id at the current version.

Design doc with full semantics, invariants, and alternatives considered: https://claude.ai/code/artifact/1c0ea7e4-9029-4f49-861d-ed8d8a79981e

## What was done?

- **Schema layer**: `requiredSince` property keyword admitted by meta-schema v3 (PV14-only, editable per its header comment), parsed onto `DocumentProperty.required_since` behind a new `apply_required_since` version slot — `None` on pre-v14 tables so frozen parsers stay byte-identical (the `refersTo` / `apply_property_reference` pattern). Parse rules: top-level properties only, must be listed in `required`, value ≥ 1.
- **Wire format 3** (`serialize_v3` / `from_bytes_v3`): a contract-version stamp varint after the format prefix; everything else identical to format 2. A property whose `requiredSince` exceeds the stamp keeps the presence-flagged layout it was written with. New `DOCUMENT_VERSIONS_V4` table (default 3) wired into `v14.rs` only; read dispatch stays purely prefix-driven, so formats 0–2 deserialize exactly as before.
- **Legacy formats 0–2** now read *and* write user properties with `required_at(None)`: byte-identical for every schema without annotations (i.e. all data that exists on any network), and it keeps old-format bytes readable under a schema that later gained a required field — the key migration path.
[...]

PR #4484, opened 2026-08-26
refactor(dpp): split document serialization formats into per-generation files

QuantumExplorer:

Follow-up to #4400, deferred there deliberately so that PR's inline diffs of the legacy serialization formats stayed reviewable. `packages/rs-dpp/src/document/v0/serialize.rs` had grown to 3,815 lines holding four frozen wire-format generations plus ~1,700 lines of tests.

## What was done?

Pure relocation — no semantic change anywhere:

- Each serialization format generation moves to its own file: `document/v0/serialize/{v0,v1,v2,v3}.rs`. Method bodies are byte-for-byte identical, relocated into inherent `impl DocumentV0` blocks (signatures gain only `pub(super)`).
- `mod.rs` keeps the three trait impls: the serialization/deserialization ones become one-line dispatch shims into the per-format inherent methods (inherent methods take precedence, so the same-name delegation is unambiguous via `DocumentV0::method(...)` UFCS); the conversion impl (prefix-driven dispatchers) is unchanged.
- Tests move verbatim (one indentation level) into `serialize/tests.rs`, with the four imports they previously inherited from the file header made explicit.
- Import blocks were copied to each file and pruned to what each format uses; feature-gate attributes were kept correct (verified by building `--no-default-features --features state-transitions` in addition to `--all-features`).

**Why:** consensus hygiene. A shipped format's file must never change, so after this split a diff touching `serialize/v1.rs` is immediately suspect in review — the same physical isolation the version tables give generation modules.

## How Has This Been Tested?

- `git diff --color-moved=zebra --color-moved-ws=allow-indentation-change` classifies all but **85 of the ~4,000 added lines as moves**; the 85 are module plumbing (per-file doc headers, `impl` wrappers, `pub(super)` markers, the 8 shims, 4 test imports).
- Full `dpp` suite passes (3,942 tests), including the entire serialization matrix — any delegation mistake would recurse or misdispatch instantly there.
- `cargo clippy --all-targets --all-features -- -D warnings` clean; `cargo check --no-default-features --features state-transitions` clean; `cargo fmt` applied.

## Breaking Changes

None — pure file reorganization; public API and wire behavior untouched.

## Checklist:


QuantumExplorer (commented):
Nice refactor

